PR #1168, opened 2015-02-11
Add initial support for a toolchain based module naming scheme

ocaisa:
where software is installed in a directory specific to each toolchain, and each toolchain module extends the module path variable. 

*NB* This won't work unless toolchains properly inherit (iccifort -> iimpi ->intel) and/or dependencies are listed appropriately in toolchains.

The other required change in easyblock.py changed the order of statements written in the module file so that the 'module use' statement appears after loading the dependencies. This allows the module path to be updated in the correct order in this naming scheme.


ocaisa:
Part of the reason for doing this is that, in a hierarchy, `module spider` returns pretty complicated ways to get to a particular module. Since we (choose to) expose users to our toolchains, this conflicts a little with how we explain to people how to access installed software.

boegel:
Jenkins: ok to tst

boegel:
Jenkins: ok to test

easybuild/tools/module_naming_scheme/toolchain_mns.py
```diff
@@ -0,0 +1,90 @@
...
+
+        return paths
+
+    def expand_toolchain_load(self):
+        """
+        Determine whether load statements for a toolchain should be expanded to load statements for its dependencies.
+        This is useful when toolchains are not exposed to users.
+        """
+#        return True
+# In our case we have to load the toolchains because they are explicitly exposed when extending the module path
```
boegel: indent comment

easybuild/tools/module_naming_scheme/toolchain_mns.py
```diff
@@ -0,0 +1,90 @@
...
+            paths.append(os.path.join(TOOLCHAIN,  ec['name'], fullver))
+
+        return paths
+
+    def expand_toolchain_load(self):
+        """
+        Determine whether load statements for a toolchain should be expanded to load statements for its dependencies.
+        This is useful when toolchains are not exposed to users.
+        """
+#        return True
```
boegel: drop this one

easybuild/tools/module_naming_scheme/toolchain_mns.py
```diff
@@ -0,0 +1,90 @@
...
+    def det_modpath_extensions(self, ec):
+        """
+        Determine module path extensions, if any.
+        Examples: Toolchain/intel/2014.12 (for intel/2014.12 module)
+        """
+        modclass = ec['moduleclass']
+
+        paths = []
+
+
```
boegel: not too verbose on empty lines please :)

easybuild/tools/module_naming_scheme/toolchain_mns.py
```diff
@@ -0,0 +1,90 @@
...
+	# Take care of the GCC corner case since it is both a compiler and a toolchain
+        if modclass == MODULECLASS_TC or ec['name'] == 'GCC':
```
boegel: change the last bit to `ec['name'] in ['GCC']`, so the list can be easily extended with other compilers that are toolchains

easybuild/tools/module_naming_scheme/toolchain_mns.py
```diff
@@ -0,0 +1,90 @@
...
+        Examples: Core, Toolchain/gpsolf/2015.02
+        """
+
+        if ec.toolchain.name == DUMMY_TOOLCHAIN_NAME:
+
+            # toolchain is dummy/dummy, put in Core
+            subdir = CORE
+        else:
+            subdir = os.path.join(TOOLCHAIN,ec.toolchain.name,ec.toolchain.version)
+
```
boegel: get rid of the empty lines in between here, they don't help in making things more readable

easybuild/tools/module_naming_scheme/toolchain_mns.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+
+import os
+import re
+from vsc.utils import fancylogger
+
+from easybuild.tools.module_naming_scheme.hierarchical_mns import HierarchicalMNS
+from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME
+
+CORE = 'Core'
```
boegel: why not reuse the `CORE` constant available in `hierarchical_mns`?

easybuild/tools/module_naming_scheme/toolchain_mns.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+Implementation of an example hierarchical module naming scheme.
+
+@author: Kenneth Hoste (Ghent University)
+@author: Markus Geimer (Forschungszentrum Juelich GmbH)
+"""
+
+import os
+import re
+from vsc.utils import fancylogger
```
boegel: drop any unused imports here

easybuild/tools/module_naming_scheme/toolchain_mns.py
```diff
@@ -0,0 +1,90 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+Implementation of an example hierarchical module naming scheme.
+
+@author: Kenneth Hoste (Ghent University)
+@author: Markus Geimer (Forschungszentrum Juelich GmbH)
```
boegel: fix name(s)?

I had nothing to do with this one ;-)

easybuild/framework/easyblock.py
```diff
@@ -1642,8 +1642,8 @@ def make_module_step(self, fake=False):
 
         txt = ''
         txt += self.make_module_description()
-        txt += self.make_module_extend_modpath()
         txt += self.make_module_dep()
+        txt += self.make_module_extend_modpath()
```
boegel: let's see if the unit tests are happy with this bit, I'm not sure...

I seem to remember I had a good reason to include the `use` statements before the `load`s, but I'm not 100% sure anymore.

ocaisa: Yeah I thought there may well be a reason for this, if so I can always handle it with an `if` statement based on the MNS

boegel:
we would also need to have this tested by the unit tests, to ensure it doesn't get broken by changes in the framework in the future

easybuild/framework/easyblock.py
boegel: from the top of my head, I don't think it matters... if used to matter, but that's because we weren't handling things correctly, that was fixed post v1.14

let's see if the unit tests prove me wrong ;)

ocaisa:
I'll take a look at the tests you mentioned in #1176 and see how this might be done. I'm a little surprised it passed the tests because without modifying intel or goolf to include a dependency on iimpi/gompi (and then iteratively), you almost certainly can't install software within this MNS. It will work for gpsolf and intel-para though because I've already defined them hierarchically.

I'll be able to resolve this more completely once we implement the method for subtoolchains, since I can enforce the dependency then (given this MNS is in use).

boegel:
The tests pass because they're not using this new MNS.. :-)

boegel:
@ocaisa: ignore the failing test, Jenkins must be drunk

boegel:
Jenkins: test this please

boegel:
@ocaisa: what's going on with your last commit? I'm not sure how useful an empty pull request is... ;)

boegel:
@ocaisa: last tests have failed due to an issue on Jenkins, I'm looking into it

boegel:
Jenkins: test this please

ocaisa:
Argh, I was tidying up everything for the subtoolchain stuff and I deleted the MNS from the branch history to keep things clean...but I never  forked the original devel branch so things got complicated.

boegel:
@ocaisa: `git revert` is your friend

boegel:
@ocaisa any idea why we ended up not including `toolchain_mns.py`? It was removed in 8e18877d94e4b9899516276e5e3891f146947626 .

@bartoldeman was hitting a problem with https://github.com/hpcugent/easybuild-easyconfigs/pull/2899, basically because `ToolchainMNS` as it was is unaware of PGI (I think...)

Do you have an updated version of `ToolchainMNS` somewhere?

ocaisa:
It couldn't work properly because it relied on knowing which toolchain it was inheriting from  (you need to have `iimpi` loaded to  correctly load `intel` since things built with `iimpi` are only available by the path extension provided by the `iimpi` module). Because eb requires components of a toolchain to be expanded (you can switch this resolution off but the framework complains then), this meant you had to explicitly include the subtoolchain as a dependency for it to work. 
With the features introduced by `minimal-toolchains` this should no longer be necessary except that the toolchain hierarchy is only available to the easyconfig class and not the mns class (at present).

ocaisa:
Annoyingly I pulled directly from my develop branch so I need to open a new PR from a fresh  branch

boegel:
@ocaisa I was mainly asking because I was figuring out the bug that got fixed with #1754; if you know it doesn't work (well), there's probably little point in contributing it back?

Although, I have to admit, this is a scheme we would consider using, so maybe it is useful to have a PR for it that makes it clear what the known problems are.

PR #1169, opened 2015-02-11
Skip checksum verification when tweaking software version

geimer:
Using `--try-software-version` on an easyconfig which contains checksums currently fails with a checksum mismatch (different tarball, different checksum). This PR addresses this issue by clearing the checksum list when tweaking the software version.

Note: unit test missing...

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -42,6 +42,7 @@ from distutils.version import LooseVersion
 from vsc.utils import fancylogger
 from vsc.utils.missing import nub
 
+from easybuild.framework.easyconfig.default import get_easyconfig_parameter_default
 from easybuild.framework.easyconfig.easyconfig import EasyConfig, create_paths, process_easyconfig
 from easybuild.tools.filetools import read_file, write_file
 from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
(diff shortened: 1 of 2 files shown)


boegel:
Jenkins: ok to test

easybuild/framework/easyconfig/tweak.py
```diff
@@ -185,7 +194,7 @@ def __repr__(self):
                     _log.debug("Overwriting %s with %s" % (key, fval))
                 ectxt = regexp.sub("%s = %s" % (res.group('key'), newval), ectxt)
                 _log.info("Tweaked %s list to '%s'" % (key, newval))
-            else:
+            elif get_easyconfig_parameter_default(key) != val:
```
boegel: with this change, injecting values for keys that are not known easyconfig parameters doesn't work anymore (cfr. the failing tests), see below;

So, either the tests are broken (since it doesn't make a lot of sense to define `foo` if EasyBuild isn't going to pick up on it), or we need to restore the ability to do something like `--try-amend=foo=bar`

```
ERROR [0.633s]: test obtaining an easyconfig file given certain specifications
  File "test/framework/easyconfig.py", line 520, in test_obtain_easyconfig
    res = obtain_ec_for(specs, [self.ec_dir], None)
  File "easybuild/framework/easyconfig/tweak.py", line 589, in obtain_ec_for
    res = select_or_generate_ec(fp, paths, specs)
  File "easybuild/framework/easyconfig/tweak.py", line 561, in select_or_generate_ec
    tweak_one(selected_ec_file, fp, specs)
  File "easybuild/framework/easyconfig/tweak.py", line 223, in tweak_one
    elif get_easyconfig_parameter_default(key) != val:
  File "easybuild/framework/easyconfig/default.py", line 183, in get_easyconfig_parameter_default
    _log.error("Unkown easyconfig parameter: %s (known: %s)" % (param, sorted(DEFAULT_CONFIG.keys())))
  File "easybuild/tools/build_log.py", line 113, in error
    raise EasyBuildError(newMsg)
EasyBuildError: "EasyBuild crashed with an error (at easybuild/framework/easyconfig/default.py:183 in get_easyconfig_parameter_default): Unkown easyconfig parameter: foo (known: [...])
```

boegel:
Going in, thanks @geimer!

PR #1170, opened 2015-02-11
Added first version of experimental support for package_step 

sntgluca:
(based on code by Marc Litherland)

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -77,7 +77,6 @@ from easybuild.tools.systemtools import det_parallelism, use_group
 from easybuild.tools.utilities import remove_unwanted_chars
 from easybuild.tools.version import this_is_easybuild, VERBOSE_VERSION, VERSION
 
-
 _log = fancylogger.getLogger('easyblock')
 
 
@@ -1444,20 +1443,18 @@ class EasyBlock(object):
         self.clean_up_fake_module(fake_mod_data)
 
     def package_step(self):
-        """Package software (e.g. into an RPM)."""
+        """Prepare package software (e.g. into an RPM) with fpm."""
         rpmname = "HPCBIOS.20150211-%s-%s" % (self.name, self.version)
         os.chdir(os.environ['TMPDIR'])
         
-        
-        toolchaindep = "=".join([self.toolchain.name, self.toolchain.version])
-        if toolchaindep == "dummy=":
-            deplist = ""
+        if self.toolchain.name == "dummy":
+            dependencies = []
         else:
-            deplist = " ".join([" --dependency ", toolchaindep])
-        deplist += " ".join([" --dependency " + i 
-            for i in (map(lambda x:x["name"] + "=" + x["version"], self.cfg.dependencies()))
-        ])
-        cmd = "fpm --workdir $TMPDIR -t rpm --name %s -s dir %s -C %s" % (rpmname, deplist, self.installdir)
+            dependencies = [ "=".join([ self.toolchain.name, self.toolchain.version ]) ]
+        dependencies.extend([ "=".join([ dep['name'], dep['version'] ]) for dep in self.cfg.dependencies() ])
+        depstring = '--dependency ' + ' --dependency '.join(dependencies)
+
+        cmd = "fpm --workdir $TMPDIR -t rpm --name %s -s dir %s -C %s" % (rpmname, depstring, self.installdir)
         (out, _) = run_cmd(cmd, log_all=True, simple=False)
 
     def post_install_step(self):
@@ -1797,11 +1794,11 @@ class EasyBlock(object):
         # part 3: post-iteration part
         steps_part3 = [
             ('extensions', 'taking care of extensions', [lambda x: x.extensions_step()], False),
-            ('package', 'packaging', [lambda x: x.package_step()], True),
             ('postproc', 'postprocessing', [lambda x: x.post_install_step()], True),
             ('sanitycheck', 'sanity checking', [lambda x: x.sanity_check_step()], False),
             ('cleanup', 'cleaning up', [lambda x: x.cleanup_step()], False),
             ('module', 'creating module', [lambda x: x.make_module_step()], False),
+            # ('package', 'packaging', [lambda x: x.package_step()], True),
         ]
 
         # full list of steps, included iterated steps
@@ -1812,6 +1809,14 @@ class EasyBlock(object):
                 lambda x: x.load_module(),
                 lambda x: x.test_cases_step(),
             ], False))
+
+        ## CHANGE TRUE TO build_option, and 
+        ## ADD build-pkg to all the configuration dicts
+        ## # if build_option('build-pkg'):
+        if True:
+            steps.append(('package', 'packaging', [lambda x: x.package_step()], True))
+        else:
+            self.log.debug('Skipping package step')
 
         return steps
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1447,9 +1447,11 @@ class EasyBlock(object):
         """Package software (e.g. into an RPM)."""
         rpmname = "nxprog-%s-%s" % (self.name, self.version)
         os.chdir("/tmp")
-        deplist=""
-        deplist+=" ".join(["--dependency " + i for i in (map(">".join, [(self.toolchain.name, self.toolchain.version)]))])
-        deplist+=" ".join(["--dependency " + i for i in (map(lambda x:x["name"] + ">" + x["version"], self.cfg.dependencies()))])
+        toolchaindep = "=".join([self.toolchain.name, self.toolchain.version])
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1445,7 +1445,13 @@ def extensions_step(self, fetch=False):
 
     def package_step(self):
         """Package software (e.g. into an RPM)."""
-        pass
+        rpmname = "nxprog-%s-%s" % (self.name, self.version)
```
fgeorgatos: @sntgluca : nxprog looks like a hardwired prefix; can we make it generic by picking it up from a configuration option?

easybuild/framework/easyblock.py
```diff
@@ -1445,7 +1445,13 @@ def extensions_step(self, fetch=False):
...
+        os.chdir("/tmp")
```
fgeorgatos: same thing on configurability...

boegel:
Jenkins: ok to test

easybuild/framework/easyblock.py
wpoely86: `$TMPDIR` is set in the beginning of running `eb`. Use that.

wpoely86: I mean: use `tempfile.mkdtemp`.

easybuild/framework/easyblock.py
```diff
@@ -1445,7 +1445,20 @@ def extensions_step(self, fetch=False):
 
     def package_step(self):
         """Package software (e.g. into an RPM)."""
-        pass
+        rpmname = "HPCBIOS.20150211-%s-%s" % (self.name, self.version)
```
stdweird: variable name should not be `rpmname` but rather `pkgname` (so it's neutral).
the `HPCBIOS.20150211` should come from a variable (`pkgprefix`). but you might better use a template with named templates (it will make more sense this way, and then we read this in from a configuration option.

so something like
```python
pkgtemplate = "HPCBIOS.20150211-%(name)s-%(version)s"

pkgname=pkgtemplate % {
    'name:' self.name,
    'version' : self.version,
}
```

easybuild/framework/easyblock.py
```diff
@@ -1445,7 +1445,20 @@ def extensions_step(self, fetch=False):
...
+        os.chdir(os.environ['TMPDIR'])
+
```
stdweird: 1 newline is enough

easybuild/framework/easyblock.py
```diff
@@ -1445,7 +1445,20 @@ def extensions_step(self, fetch=False):
...
+        
+        toolchaindep = "=".join([self.toolchain.name, self.toolchain.version])
+        if toolchaindep == "dummy=":
+            deplist = ""
+        else:
+            deplist = " ".join([" --dependency ", toolchaindep])
+        deplist += " ".join([" --dependency " + i 
+            for i in (map(lambda x:x["name"] + "=" + x["version"], self.cfg.dependencies()))
+        ])
+        cmd = "fpm --workdir $TMPDIR -t rpm --name %s -s dir %s -C %s" % (rpmname, deplist, self.installdir)
```
stdweird: the `rpm` must be set via variable (and later also from a config option).

you should also test if fpm is available (there's a which function somewhere in framework already taht can be used for this)

easybuild/framework/easyblock.py
```diff
@@ -1445,7 +1445,20 @@ def extensions_step(self, fetch=False):
 
     def package_step(self):
```
stdweird: this method should be moved elsewhere (a `package.py` module perhaps?)  and be called `package_fpm`.
then call this new method from here. the packaging flavour should be selected via configuration option (so might wrap this in a rather abstract `Packaging` class in said `package` module; but a simple function will do for now)

boegel:
to figure out the location of the module file in `package_step`:

```python
path_to_module_file = os.path.join(install_path('mod'), build_option('suffix_modules_path'), self.full_mod_name)
```

easybuild/framework/easyblock.py
boegel: +1 on this, since this is most likely only one way of packaging that we'll support in the (near) future, other candidates may include Docker containers, Nix packages (?), ...

@sntgluca: let me know if you're up for tackling @stdweird's remark

sntgluca: @boegel: @stdweird: @fgeorgatos: Agree about all that. 
I don't think I am ready to implement it. Suggestion: you or Fotis can handle the project and I collaborate and learn in the process.

boegel: @sntgluca: ok, maybe we can set up a conf call on this once we've had the chance to take a detailed look at this, or you can join one of the future EasyBuild conf calls (see https://github.com/hpcugent/easybuild/wiki/Conference-calls)

if you're not on the EB mailing list yet, do subscribe (https://lists.ugent.be/wws/info/easybuild)

rjeschmi: Hoping to help with this

Looking at this idea, is there an interface to pass the config state to such a function (or at least an example in the current code?)

I want to pass a lot of the EasyBlock object into package_fpm, but it seems a bit overkill to pass the whole thing...

easybuild/framework/easyblock.py
```diff
@@ -1444,8 +1443,19 @@ def extensions_step(self, fetch=False):
...
+        os.chdir(os.environ['TMPDIR'])
+        
+        if self.toolchain.name == "dummy":
+            dependencies = []
+        else:
+            dependencies = [ "=".join([ self.toolchain.name, self.toolchain.version ]) ]
+        dependencies.extend([ "=".join([ dep['name'], dep['version'] ]) for dep in self.cfg.dependencies() ])
+        depstring = '--dependency ' + ' --dependency '.join(dependencies)
+
+        cmd = "fpm --workdir $TMPDIR -t rpm --name %s -s dir %s -C %s" % (rpmname, depstring, self.installdir)
```
jordansissel: Hope you don't mind me dropping in on the review!

When invoking fpm (or subprocesses, in general), I highly recommend using array-oriented execution instead of one string. This allows you to avoid any problems that might occur with spaces in the filename or other similar problems.

I had a look at the code under `run_cmd` and it uses `subprocess.Popen` which takes either a string or a list/tuple as program args, I think you can do this like:

```python
cmd = ("fpm", "--workdir", os.environ["TMPDIR"], "-t", "rpm", "--name", rpmname, "-s", "dir") + dependency_flags + ("-C", self.installdir)
```

The above assumes 'dependency_flags' is a tuple of flags, not a string.

This will help you if depstring or installdir or $TMPDIR have spaces or other possible hazards. Hope this helps! :)

fgeorgatos: Thanks, this is an interesting suggestion, as anything that removes troubles with fancy namings!

easybuild/framework/easyblock.py
rjeschmi: @stdweird Isn't TMPDIR set by easybuild to be correct? I can't find a good way to grab it from python (the variable eb_tmpdir is set in main.py, but not into a parameter I think I can see from where package_fpm would run

looking at set_tmpdir in tools/config.py

```
    for var in ['TMPDIR', 'TEMP', 'TMP']:
        env.setvar(var, current_tmpdir)
```

stdweird:
one other thing that is ignored here is the license information. it would be nice if package respect the license (and if no license is defined in the easyconfig, you should assume so restricted/private license, so no rights to redistribute in a package either).

easybuild/framework/easyblock.py
stdweird: @jordansissel running everything in subshell was a historical mistake we made in EB long long ago. it's very convenient, maybe even required, for some cirtical parts of EB (like conifgure and make), but we shouldn't be using it were it can be easily avoided like in this case. so we should already try to write code that assumes we will move to non-subshell (as one of my earlier comments suggested).

jordansissel: If it would be useful, I would be open to exploring some kind of rpc
interface (jsonrpc?) to fpm instead of requiring programs invoke the cli
with flags. Not required for this PR, but just putting the idea out there :)

easybuild/framework/easyblock.py
boegel: `self` is the EasyBlock instance in this context, `self.cfg` is the parsed easyconfig file

easybuild/framework/easyblock.py
boegel: @jordansissel: if that's useful for other use cases of FPM too, sure, makes sense imho

wpoely86: @stdweird care to elaborate what the issue is with subshells?

boegel:
This can be closed since #1224 has been merged in, where @rjeschmi continued this effort.

Thanks for kickstarting this @sntgluca, support for `--package` with FPM will be included in EB v2.2.0.

PR #1171, opened 2015-02-11
 Use different logger classes depending on prefix of requested logger

riccardomurri:
As the title says: this allows external libraries (e.g., GC3Pie) to use stdlib logging, while EB continues to use its own logging facility.

It's a hack, however, that involves directly setting a module-level private variable...

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -31,6 +31,7 @@ EasyBuild logger and log utilities, including our own EasybuildError class.
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
 """
+import logging
 import os
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


riccardomurri:
Jenkins: ok to test

boegel:
Jenkins: ok to test

boegel:
Things are very broken now, it seems... Somehow EasyBuild isn't using an EasyBuildLog instance anymore?

```
  File "test/framework/options.py", line 1065, in test_deprecated
    log.deprecated('x', str(orig_value))
TypeError: deprecated() takes at least 4 arguments (3 given)
```

```
  File "test/framework/options.py", line 1028, in test_experimental
    log.experimental('x')
AttributeError: FancyLogger instance has no attribute 'experimental'
```

```
  File "test/framework/easyconfigparser.py", line 75, in test_v20
    ec = ecp.get_config_dict()
  File "easybuild/framework/easyconfig/parser.py", line 148, in get_config_dict
    return self._formatter.get_config_dict()
  File "easybuild/framework/easyconfig/format/two.py", line 99, in get_config_dict
    self.log.experimental(self.__class__.__name__)
AttributeError: FancyLogger instance has no attribute 'experimental'
```

riccardomurri:
@boegel Fixed now, thanks for the heads-up.

riccardomurri:
A lighter-weight alternative to this is in #1179

boegel:
same here as in #1179; this is a very complex workaround while https://github.com/hpcugent/easybuild-framework/pull/1008/files#diff-30c0c724e8b33847f917f22eca0af7a0R57 does the trick as well, and the `raiseError` aspect of `EasyBuildLog` has been deprecated (to disappear in EB v3.0), so this is no longer useful?

riccardomurri:
Yes, I agree to closing w/out merging

PR #1173, opened 2015-02-12
Suppress creation of module symlinks for hierarchical MNS

geimer:
In the context of a hierarchical module naming scheme, the default symlink farm prefixing the module path with the `moduleclass` easyconfig parameter is of very limited use. First, it creates "strange" modulefile layouts such as
```
numlib/
    Compiler/GCC/4.9.2/
        OpenBLAS
    MPI/GCC/4.9.2/OpenMPI/1.8.4/
        FFTW
        ScaLAPACK
```
which are hard to use in a sensible way. And second, the compiler and MPI modules only extend `$MODULEPATH` for the regular hierarchy, resulting in weird effects when loading these modules. Therefore, it's probably best to suppress the generation of these symlinks entirely.

### DIFF
--- a/easybuild/tools/module_naming_scheme/hierarchical_mns.py
+++ b/easybuild/tools/module_naming_scheme/hierarchical_mns.py
@@ -133,6 +133,13 @@ class HierarchicalMNS(ModuleNamingScheme):
 
         return subdir
 
+    def det_module_symlink_paths(self, ec):
+        """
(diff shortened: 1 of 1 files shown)


boegel:
@geimer: makes perfect sense, and if we merge this in before EB v2.0 we have a good excuse to break this ;)

boegel:
Jenkins: ok to test

boegel:
Thanks @geimer, going in!

PR #1082, opened 2014-10-30
	added support for checking out svn revisions to download_file

JensTimmerman:
This works for easyconfig files looking like this (if pysvn and tar are available on the system):

```
name = 'FDS'
version = '20564'

homepage = 'https://code.google.com/p/fds-smv/'
description = """
 Fire Dynamics Simulator (FDS) is a large-eddy simulation (LES) code for low-speed flows, with an emphasis on smoke and
 heat transport from fires.
"""
toolchain = {'name': 'foss', 'version': '2014b'}
toolchainopts = {'pic': True, 'usempi': True}

source_urls = ['svn+http://fds-smv.googlecode.com/svn/trunk/FDS/trunk/']

sources = ['%(version)s.tar.gz']
moduleclass = 'phys'
```

note the 'svn+http' part in the source_urls, this is needed to tell the download function to use svn. version is used as the revision number to check out, 
sources are currently always '%(version)s.tar.gz'
Sine this code checks out the svn repo at the given revision, creates a tarball, and removes it again.

PR #1008, opened 2014-08-13
Support parallel builds with GC3Pie (REVIEW)

riccardomurri:
This is meant to be the implementation of #1006.

Basic functionality works. Configuration for the GC3Pie part is still hard-coded; I need your input as to how you would like this to be exposed in EB.

At the moment, it requires the "trunk" version of GC3Pie, soon to be released as GC3Pie 2.3.

PR #1174, opened 2015-02-12
Fix `log.error` varargs

riccardomurri:
Make `EasyBuildLog.error` treat first string as a format string, and interpolate additional argument into it.

(As discussed briefly in comments to #1082 and #1008)

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -107,8 +107,9 @@ class EasyBuildLog(fancylogger.FancyLogger):
 
     def error(self, msg, *args, **kwargs):
         """Print error message and raise an EasyBuildError."""
-        newMsg = "EasyBuild crashed with an error %s: %s" % (self.caller_info(), msg)
-        fancylogger.FancyLogger.error(self, newMsg, *args, **kwargs)
+        newMsg = ("EasyBuild crashed with an error %s: %s"
+                  % (self.caller_info(), (msg % args)))
+        fancylogger.FancyLogger.error(self, newMsg, **kwargs)
         if self.raiseError:
             raise EasyBuildError(newMsg)
 


wpoely86:
ok to test

easybuild/tools/build_log.py
```diff
@@ -107,8 +107,9 @@ def nosupport(self, msg, ver):
 
     def error(self, msg, *args, **kwargs):
         """Print error message and raise an EasyBuildError."""
-        newMsg = "EasyBuild crashed with an error %s: %s" % (self.caller_info(), msg)
-        fancylogger.FancyLogger.error(self, newMsg, *args, **kwargs)
+        newMsg = ("EasyBuild crashed with an error %s: %s"
+                  % (self.caller_info(), (msg % args)))
```
stdweird: you only needed the interpolation for the `raise`, `fancylogger.FancyLogger.error` can handle `args` (but this is not a blocker for me)

riccardomurri: Yes, but since args have already been interpolated into `newMsg`, there's no need to pass them to the logger: `newMsg` no longer contains any `%s` format specifiers.

easybuild/tools/build_log.py
```diff
@@ -107,8 +107,9 @@ def nosupport(self, msg, ver):
...
+        fancylogger.FancyLogger.error(self, newMsg, **kwargs)
         if self.raiseError:
             raise EasyBuildError(newMsg)
```
boegel: @riccardomurri: isn't this still a problem for GC3Pie?

we should look into a way of only raising when the logger instance is used inside EB

riccardomurri: @boegel that's a separate discussion :-)  

The current `gc3pie.py` has a workaround, but yes it would be nice to have a more general solution.

boegel:
This has been fixed in #1218, so good to close?

riccardomurri:
I guess so.

PR #1176, opened 2015-02-12
Hierarchical module naming scheme categorizing modulefiles by 'moduleclass'

geimer:
**NOTE: This PR depends on https://github.com/hpcugent/easybuild-framework/pull/1163**

This is an enhanced hierarchical module naming scheme which uses the `moduleclass` easyconfig parameter to further categorize modulefiles on each level of the hierarchy.

@fgeorgatos, @pforai: here you go :-)


fgeorgatos:
hooray! nice contribution Markus, thanks; 

My assumption is that most phys/chem HPC shops would much prefer a namespace of this type.

(and now we need to convince @ocaisa for sharing the other alternative toolchain namespace, so that we can cook hybrids of these ideas)

easybuild/tools/module_naming_scheme/categorized_hmns.py
```diff
@@ -0,0 +1,106 @@
...
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+##
```
fgeorgatos: as I see it, this is 3-clause BSD, no? http://en.wikipedia.org/wiki/BSD_licenses

It would be nice if it was stated as such somewhere (just me pipedreaming here that all devs would do the same)

fgeorgatos:
visual review, all looks good. TBC.

easybuild/tools/module_naming_scheme/categorized_hmns.py
geimer: @fgeorgatos: Yes, it is 3-clause BSD. I'm so used to it that I thought it was obvious ;-) Not sure where to best add such a statement, though.

boegel: just add an extra comment line, mentioning something like:

```python
# License: 3-clause BSD
# Author: Markus Geimer (JSC)
```

boegel:
We need to get this integrated in the unit tests somehow, to make sure we don't break it in the future.

There are several test modules that now test `HierachicalMNS`:

```
test/framework//easyblock.py:520:        Make sure that modules under the HierarchicalMNS are correct,
test/framework//easyblock.py:532:        os.environ['EASYBUILD_MODULE_NAMING_SCHEME'] = 'HierarchicalMNS'
test/framework//easyblock.py:543:        # example: for imkl on top of iimpi toolchain with HierarchicalMNS, no module load statements should be included
test/framework//module_generator.py:399:        os.environ['EASYBUILD_MODULE_NAMING_SCHEME'] = 'HierarchicalMNS'
test/framework//modules.py:258:        os.environ['EASYBUILD_MODULE_NAMING_SCHEME'] = 'HierarchicalMNS'
test/framework//options.py:439:            'module-naming-schemes': ['EasyBuildMNS', 'HierarchicalMNS'],
test/framework//options.py:723:            '--module-naming-scheme=HierarchicalMNS',
test/framework//options.py:1219:        for extra_args in [[], ['--module-naming-scheme=HierarchicalMNS']]:
test/framework//toy_build.py:524:            '--module-naming-scheme=HierarchicalMNS',
test/framework//utilities.py:215:        # a flat scheme like EasyBuildMNS and a hierarhical one like HierarchicalMNS doesn't work
```

Not all of these need to also test this new MNS, but it should be tested in one or two places at least.

@geimer: Are you up for that? If not, I can look into it (not sure how soon though).

boegel:
Jenkins: ok to test

easybuild/tools/module_naming_scheme/categorized_hmns.py
boegel: ok, you already have the `author` bit below, so just an extra comment for the license part is sufficient

easybuild/tools/module_naming_scheme/categorized_hmns.py
```diff
@@ -0,0 +1,106 @@
...
+        Examples:
+        Core/<moduleclasses>
+        """
+        known_module_classes = build_option('valid_module_classes')
+        basepaths = super(CategorizedHMNS, self).det_init_modulepaths(ec)
+
+        paths = []
+        for path in basepaths:
+            for moduleclass in known_module_classes:
+                paths.extend([os.path.join(path, moduleclass)])
```
boegel: this is pretty much copy-paste of `det_modpath_extensions`, so the body should be fleshed out in another method or function like `categorise_paths(paths)`, which is simply called twice

we already have enough LoC to take care of ;)

boegel: or, alternatively, collapse this into a single line using a Python list comprehension:

```python
paths = [os.path.join(p, mc) for p in basepaths for mc in known_module_classes]
```

boegel: and then you don't even need `paths` or a separate `categorize_paths()` anymore, just return the list straight away :)

fgeorgatos: Lovely meeting of Python with the functional languages world ;-)

ocaisa:
@fgeorgatos It's sitting there waiting for attention #1168

boegel:
@ocaisa: I've just reviewed #1168

easybuild/tools/module_naming_scheme/categorized_hmns.py
geimer: To be honest, your suggestion looks more like perl than Python... looks awful...

Since I'm in favor of readable code, I opt for the separate method (which in fact makes sense -- and I don't know why I didn't think of it myself).

boegel: this is a pretty clean list comprehension, we have many others... ;-)

but sure, go ahead and do it via a separate method then, I don't feel strongly about making it a list comprehension, since I won't have to maintain this bit ;)

boegel:
@geimer: the best location for a test is probably `test/framework/module_generator.py`, see `test_hierarchical_mns`

geimer:
@boegel: That's what I thought, too. Just staring at it and trying to understand. The first part is quite clear, but what is the final part after the
```
os.environ['EASYBUILD_MODULE_NAMING_SCHEME'] = self.orig_module_naming_scheme
```
good for?

boegel:
@geimer: leave that in, it restores whatever module naming scheme was active before running the tests

geimer:
That's understood. I was actually wondering why the test is verifying some things again afterwards, assuming that the original naming scheme is the "traditional" one.

boegel:
@geimer: it's checking whether the default naming scheme setting is being honored, which is why the tests shouldn't be picking up any user config files :)

geimer:
Shouldn't this be in a separate test method then? I mean, `test_hierarchical_mns` somehow implies that the test is testing the hierarchical module naming scheme. But apparently it's doing more. So why not have a separate `test_easybuild_mns`? The nested `test_ec` should then probably be moved outside to avoid code duplication.

boegel:
@geimer: yes, you're probably right, that could be cleaned up, indeed; I guess it's also checking whether restoring the default MNS actually works

geimer:
@boegel: I now integrated a test into the existing method. Assuming that @ocaisa adds a test there too, refactoring should probably be done in a separate PR afterwards to simplify merging.

boegel:
@geimer: just gave this another review, looks great.

The only thing, as discussed off-PR: let's isolate the test modules specific to `CategorizedHMNS` in a subdirectory of `test/framework/modules`?

You can enhance `setup_hierarchical_modules` in `test/framework/utilities.py` as follows:

```python
def setup_hierarchical_modules(self, subdir=None):
    """Setup hierarchical modules to run tests on."""
    if subdir is not None:
        mod_prefix = os.path.join(self.test_installpath, 'modules', 'all', subdir)
    else:
        mod_prefix = os.path.join(self.test_installpath, 'modules', 'all')
```

and then call `setup_hierarchical_modules(subdir='CategorizedHMNS')` before testing your MNS

geimer:
@boegel: There is a little bit more to be done than just modifying `mod_prefix`. Thus, creating a `setup_categorized_modules` method.

boegel:
@geimer: that should be `setup_categorized_hmns_modules`?

geimer:
@boegel: Then `setup_hierarchical_modules` should have been `setup_hierarchical_mns_modules`, no? ;-)

I'm happy to rename `setup_categorized_modules` as suggested. Though renaming `setup_hierarchical_modules` should be done in a separate PR.

boegel:
I'm just saying you're using a hierarchical MNS too, so only calling it categorised is misleading.

So, `setup_categorized_hierarchical_modules` is another option. ;-)

geimer:
Ah, got your point. Makes sense.

geimer:
@boegel: Unless you see a need for enhancing even more unit tests for the CategorizedHMNS, I'd say this PR is good for review.

boegel:
@geimer: this looks great, thanks for all the effort! Going in!

@ocaisa: if you need inspiration for adding tests for your PR #1168, checkout the diff here

PR #1177, opened 2015-02-12
add support for --update-lmod-caches (disabled by default) and --compile-lmod-caches (WIP)

boegel:
@rtmclay: it would be great if you could give this a review, especially w.r.t. the FIXME I have in there

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -48,11 +48,14 @@ from vsc.utils.patterns import Singleton
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option, get_modules_tool, install_path
 from easybuild.tools.environment import restore_env
-from easybuild.tools.filetools import convert_name, mkdir, read_file, path_matches, which
+from easybuild.tools.filetools import convert_name, mkdir, read_file, path_matches, which, write_file
 from easybuild.tools.module_naming_scheme import DEVEL_MODULE_SUFFIX
 from easybuild.tools.run import run_cmd
 from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME, DUMMY_TOOLCHAIN_VERSION
 from vsc.utils.missing import nub
+
+# relative to user's home directory
+LMOD_USER_CACHE_RELDIR = '.lmod.d/.cache'
 
 # software root/version environment variable name prefixes
 ROOT_ENV_VAR_NAME_PREFIX = "EBROOT"
@@ -832,32 +835,32 @@ class Lmod(ModulesTool):
         return correct_real_mods
 
     def update(self):
-        """Update after new modules were added."""
-        spider_cmd = os.path.join(os.path.dirname(self.cmd), 'spider')
-        cmd = [spider_cmd, '-o', 'moduleT', os.environ['MODULEPATH']]
-        self.log.debug("Running command '%s'..." % ' '.join(cmd))
+        """Update Lmod cache after new modules were added."""
+        cache_dir = build_option('update_lmod_cache')
 
-        proc = subprocess.Popen(cmd, stdout=PIPE, stderr=PIPE, env=os.environ)
-        (stdout, stderr) = proc.communicate()
+        if cache_dir is not None:
+            # run spider to create cache contents
+            spider_cmd = os.path.join(os.path.dirname(self.cmd), 'spider')
+            # FIXME support also updating dbT and reverseMapT
+            cmd = [spider_cmd, '-o', 'moduleT', os.environ['MODULEPATH']]
+            self.log.debug("Running command '%s'..." % ' '.join(cmd))
 
-        if stderr:
-            self.log.error("An error occured when running '%s': %s" % (' '.join(cmd), stderr))
+            # we can't use run_cmd, since we need to separate stdout from stderr
+            proc = subprocess.Popen(cmd, stdout=PIPE, stderr=PIPE, env=os.environ)
+            (cache_txt, stderr) = proc.communicate()
 
-        if self.testing:
-            # don't actually update local cache when testing, just return the cache contents
-            return stdout
-        else:
-            try:
-                cache_filefn = os.path.join(os.path.expanduser('~'), '.lmod.d', '.cache', 'moduleT.lua')
-                self.log.debug("Updating Lmod spider cache %s with output from '%s'" % (cache_filefn, ' '.join(cmd)))
-                cache_dir = os.path.dirname(cache_filefn)
-                if not os.path.exists(cache_dir):
-                    mkdir(cache_dir, parents=True)
-                cache_file = open(cache_filefn, 'w')
-                cache_file.write(stdout)
-                cache_file.close()
-            except (IOError, OSError), err:
-                self.log.error("Failed to update Lmod spider cache %s: %s" % (cache_filefn, err))
+            if stderr:
+                self.log.error("An error occured when running '%s': %s" % (' '.join(cmd), stderr))
+
+            if self.testing:
+                # don't actually update local cache when testing, just return the cache contents
+                return cache_txt
+            else:
+                cache_fp = os.path.join(cache_dir, 'moduleT.lua')
+                self.log.debug("Updating Lmod spider cache %s with output from '%s'" % (cache_fp, ' '.join(cmd)))
+                # FIXME: backup existing cache
+                # FIXME: also support compiling cache
+                write_file(cache_fp, cache_txt)
 
     def prepend_module_path(self, path):
         # Lmod pushes a path to the front on 'module use'
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -48,7 +48,7 @@ from vsc.utils.patterns import Singleton
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option, get_modules_tool, install_path
 from easybuild.tools.environment import restore_env
-from easybuild.tools.filetools import convert_name, mkdir, read_file, path_matches, which, write_file
+from easybuild.tools.filetools import convert_name, copy_file, mkdir, read_file, path_matches, which, write_file
 from easybuild.tools.module_naming_scheme import DEVEL_MODULE_SUFFIX
 from easybuild.tools.run import run_cmd
 from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME, DUMMY_TOOLCHAIN_VERSION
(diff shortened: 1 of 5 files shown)


easybuild/tools/modules.py
```diff
@@ -831,33 +834,54 @@ def available(self, mod_name=None):
 
         return correct_real_mods
 
+    def compile_cache(self, cache_fp):
+        """Compile Lmod cache file using luac."""
+        if build_option('compile_lmod_caches'):
+            # FIXME: ask Lmod to compile, to ensure that luac that matches the lua used by Lmod is used?
```
stdweird: :+1: on letting lmod compile the cache. then we also don't need to test if luac is in PATH etc etc.

easybuild/tools/modules.py
```diff
@@ -831,33 +834,54 @@ def available(self, mod_name=None):
...
-                cache_file.write(stdout)
-                cache_file.close()
-            except (IOError, OSError), err:
-                self.log.error("Failed to update Lmod spider cache %s: %s" % (cache_filefn, err))
+                # run spider to create cache contents
+                spider_cmd = os.path.join(os.path.dirname(self.cmd), 'spider')
+                cmd = [spider_cmd, '-o', cache_type, os.environ['MODULEPATH']]
+                self.log.debug("Running command '%s'..." % ' '.join(cmd))
+
+                # we can't use run_cmd, since we need to separate stdout from stderr
```
stdweird: what's so special about stderr? there's no proper exitcode?

easybuild/tools/modules.py
```diff
@@ -831,33 +834,54 @@ def available(self, mod_name=None):
...
+                else:
+                    cache_fp = os.path.join(cache_dir, '%s.lua' % cache_type)
+                    self.log.debug("Updating Lmod spider cache %s with output from '%s'" % (cache_fp, ' '.join(cmd)))
+
+                    # back existing cache by copying it to <cache>T.old.lua
+                    # Lmod also considers this filename for the cache in case <cache>T.lua isn't found
+                    if os.path.exists(cache_fp):
+                        old_cache_fp = '%s.old.lua' % os.path.splitext(cache_fp)[0]
+                        copy_file(cache_fp, old_cache_fp)
+                        self.compile_cache(old_cache_fp)
```
stdweird: why recompile and not move the old cache (if available)?

easybuild/tools/modules.py
```diff
@@ -831,33 +834,54 @@ def available(self, mod_name=None):
...
-                cache_dir = os.path.dirname(cache_filefn)
-                if not os.path.exists(cache_dir):
-                    mkdir(cache_dir, parents=True)
-                cache_file = open(cache_filefn, 'w')
-                cache_file.write(stdout)
-                cache_file.close()
-            except (IOError, OSError), err:
-                self.log.error("Failed to update Lmod spider cache %s: %s" % (cache_filefn, err))
+                # run spider to create cache contents
+                spider_cmd = os.path.join(os.path.dirname(self.cmd), 'spider')
```
stdweird: why not use https://github.com/TACC/Lmod/blob/master/contrib/BuildSystemCacheFile/createSystemCacheFile.sh? isn't it part of a lmod distribution?

easybuild/tools/options.py
```diff
@@ -177,6 +178,7 @@ def override_options(self):
             'allow-modules-tool-mismatch': ("Allow mismatch of modules tool and definition of 'module' function",
                                             None, 'store_true', False),
             'cleanup-builddir': ("Cleanup build dir after successful installation.", None, 'store_true', True),
+            'compile-lmod-caches': ("Compile Lmod cache files with luac after updating them", None, 'store_true', False),
```
stdweird: i don't like the lmod option in the name. there are other module tools that can have caches

stdweird: what does this do? should be just a toggle, that if you run `--update-cache` also compiles it (but i don't like the idea. lmod should do what is best, i.e. compile if luac is avail).

easybuild/tools/options.py
```diff
@@ -197,10 +199,12 @@ def override_options(self):
             'set-gid-bit': ("Set group ID bit on newly created directories", None, 'store_true', False),
             'sticky-bit': ("Set sticky bit on newly created directories", None, 'store_true', False),
             'skip-test-cases': ("Skip running test cases", None, 'store_true', False, 't'),
-            'umask': ("umask to use (e.g. '022'); non-user write permissions on install directories are removed",
-                      None, 'store', None),
             'optarch': ("Set architecture optimization, overriding native architecture optimizations",
                         None, 'store', None),
+            'umask': ("umask to use (e.g. '022'); non-user write permissions on install directories are removed",
+                      None, 'store', None),
+            'update-lmod-caches': ("Update Lmod cache files in specified directory", None, 'store_or_None',
```
stdweird: should be module tool flavour neutral.

easybuild/tools/options.py
```diff
@@ -405,6 +409,11 @@ def postprocess(self):
             if token is None:
                 self.log.error("Failed to obtain required GitHub token for user '%s'" % self.options.github_user)
 
+        # options w.r.t. updating Lmod cache only make sense when Lmod is selected modules tool
+        if self.options.update_lmod_caches:
+            if self.options.modules_tool != Lmod.__name__:
```
stdweird: should be a call to the modules tool class like `klass.can_create_cache` or something like that

easybuild/tools/modules.py
boegel: that's just a side-effect, but yes, that's good

it's important that Lmod uses the `luac` that matches the `lua` it's using, since the Lua version has to be part of the filename of the compiled cache (I still need to fix that here)

easybuild/tools/modules.py
boegel: `spider` acts like `lmod` and `modulecmd`, I think: the cache contents to `stdout`, any messages (also non-error) to `stderr`

easybuild/tools/modules.py
boegel: Lmod also picks up `moduleT.old.luac_<luaver>` and `moduleT.old.lua`, if they're there, to cover the rare case where `moduleT.lua` is not in place yet, but the previous version is still around

easybuild/tools/modules.py
boegel: I don't think it gets installed as a part of Lmod, and it needs to be tweaked anyway, it has hardcoded stuff in it specific to TACC.

I'm just redoing what that script implements, with proper logging (coming up) and error handling in Python, etc.

easybuild/tools/options.py
boegel: `spider` just dumps the cache to `stdout`, there is no tool included in Lmod (yet?) that takes care of the `.old`, compiling, etc. I'm making the compilation optional here, since `luac` may not be around, and for older Lmod version the compiled cache is useless (since Lmod doesn't pick it up).

easybuild/tools/options.py
boegel: there is no caching in Tcl(/C) env mods... suggestion for a better name?

at least that makes it clear from the name of the command line option that this is specific to Lmod

easybuild/tools/options.py
boegel: hmm, good suggestion

or maybe `is_lmod()` ;-)

pforai:
Can we look into drop using $LMOD_IGNORE_CACHE=1 globally?

easybuild/tools/modules.py
```diff
@@ -48,12 +48,15 @@
...
 from easybuild.tools.environment import restore_env
-from easybuild.tools.filetools import convert_name, mkdir, read_file, path_matches, which
+from easybuild.tools.filetools import convert_name, copy_file, mkdir, read_file, path_matches, which, write_file
 from easybuild.tools.module_naming_scheme import DEVEL_MODULE_SUFFIX
 from easybuild.tools.run import run_cmd
 from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME, DUMMY_TOOLCHAIN_VERSION
 from vsc.utils.missing import nub
 
+# relative to user's home directory
+LMOD_USER_CACHE_RELDIR = '.lmod.d/.cache'
```
wpoely86:
```
  --with-useDotFiles=yes/no
                          If yes use ~/.lmod.d/.cache, if no use
                          ~/.lmod.d/__cache__
```

easybuild/tools/modules.py
wpoely86: Shouldn't we then extand `run_cmd` for this use case?

boegel: we should start using the `run` commands that are provided by `vsc.utils.run`, and the function in there should be enhanced to allow spltiting stdout/stderr

that's a long-term effort though, since either we redefine `run_cmd` to use those, or we drop/deprecate `run_cmd` all together

easybuild/tools/modules.py
boegel: ah, yes, but you're right, I should just move the old cache, rather than recompile is, just realised what you intended to point out

easybuild/tools/modules.py
boegel: this is just a default, `--update-lmod-caches` takes an argument if you want to update caches in another location

since `.lmod.d/.cache` is the default in Lmod, I think just keeping this as is makes sense

easybuild/tools/options.py
boegel: added a static method `is_lmod` in `ModulesTool`

boegel:
@pforai: OK, I'll try and tackle that bit in here, as discussed... It should boil down to always updating the cache before every build is started. Although this will take a little while, it will avoid having to scan the module tree multiple times when running, since `module avail`, which is done for each dependency, etc., will become very fast...

boegel:
@pforai: I took a look, and there's more needed to make sure we can not ignore the cache...

We need to make sure that we can/should update the cache actively being used by Lmod in the environment we're running in. This could be a system cache (where we may not have permissions to to update it), or just our local user cache.

So, it seems like we need the following cmdline options:

* `--ignore-lmod-caches` (enabled by default, to make sure?)
* `--update-lmod-caches[=<dir>` (which is there now)

Iff `--ignore-lmod-caches` is set to `False`, we need to somehow query Lmod which cache it will be using, an try to update it. If that works, we can continue. If not, we need to scream bloody murder and end it before it goes bad somehow (e.g. accidentally reinstalling something because we don't see it in `ml av`).

Auto-updating the 'live' Lmod cache is probably also not a good idea in a general sense, since we don't know for which `$MODULEPATH` it was built... Although it may be fine to assume that the `$MODULEPATH` is correct when `eb` is being asked to update the `live` cache.

In short, I need to give this more thought, so we should probably stay away from *also* tackling that in this PR.

boegel:
@stdweird: please re-review?

boegel:
@pforai: One thing we can do is generate our own Lmod cache in `/tmp`, and then configure Lmod to use that cache via `$LMOD_RC`. With that in place, we can disable `LMOD_IGNORE_CACHE`...

rtmclay:
This looks O.K. to me.  You might want to add support for -D with the spider command as a way to track down Lmod problems when building the spider cache.

easybuild/tools/options.py
```diff
@@ -197,10 +198,13 @@ def override_options(self):
             'set-gid-bit': ("Set group ID bit on newly created directories", None, 'store_true', False),
             'sticky-bit': ("Set sticky bit on newly created directories", None, 'store_true', False),
             'skip-test-cases': ("Skip running test cases", None, 'store_true', False, 't'),
-            'umask': ("umask to use (e.g. '022'); non-user write permissions on install directories are removed",
-                      None, 'store', None),
             'optarch': ("Set architecture optimization, overriding native architecture optimizations",
                         None, 'store', None),
+            'umask': ("umask to use (e.g. '022'); non-user write permissions on install directories are removed",
+                      None, 'store', None),
+            'update-lmod-cache': ("Update Lmod cache files in specified dir after generating module file",
```
stdweird: (again) module tool neutral option name

easybuild/tools/modules.py
```diff
@@ -832,32 +874,72 @@ def available(self, mod_name=None):
...
-                cache_dir = os.path.dirname(cache_filefn)
-                if not os.path.exists(cache_dir):
-                    mkdir(cache_dir, parents=True)
-                cache_file = open(cache_filefn, 'w')
-                cache_file.write(stdout)
-                cache_file.close()
-            except (IOError, OSError), err:
-                self.log.error("Failed to update Lmod spider cache %s: %s" % (cache_filefn, err))
+        """Update Lmod cache after new modules were added."""
+        cache_dir = None
```
stdweird: again, why is this not simply calling https://github.com/TACC/Lmod/blob/master/contrib/BuildSystemCacheFile/createSystemCacheFile.sh
@rtmclay is the createSystemCacheFile.sh not part of the supported lmod release?

easybuild/tools/modules.py
```diff
@@ -137,6 +140,11 @@ class ModulesTool(object):
 
     __metaclass__ = Singleton
 
+    @staticmethod
+    def is_lmod():
```
stdweird: neutral names pls

easybuild/tools/options.py
```diff
@@ -405,6 +409,11 @@ def postprocess(self):
             if token is None:
                 self.log.error("Failed to obtain required GitHub token for user '%s'" % self.options.github_user)
 
+        # options w.r.t. updating Lmod cache only make sense when Lmod is selected modules tool
+        modules_tool_class = modules_tool(name=self.options.modules_tool, return_class=True)
+        if self.options.update_lmod_cache and not modules_tool_class.is_lmod():
+            self.log.error("Options related to Lmod cache are only supported if Lmod is used as a modules tool.")
```
stdweird: there shouldn't be any options related to only one modules tool. ever.

rtmclay:
the script createSystemCacheFile.sh is part of the standard Lmod distribution.  But it has to be modified to be used at each site.

easybuild/tools/modules.py
wpoely86: because `createSystemCacheFile.sh` does not get installed and it needs manual tweaking before you can use it. It's an example how to create the spider cache (hence why it is in the `contrib` directory).

boegel:
@rtmclay: is there any reason why there is no dedicated option in Lmod to update the cache(s) at a specific location?

stdweird:
@rtmclay what needs customisation? it would be better if EB can just use it instead trying to reproduce it in python (it would be even better if lmod had a `lmod --make-cache` option).

easybuild/tools/options.py
boegel: OK, I'll make this more neutral, something like `update-modules-tool-cache`.

boegel:
@stdweird: the location of the cache and timestamp files, potentially also the `$MODULEPATH` for which the cache is built, the `luac` being used (which has to match the one Lmod is using), etc.

But, I agree that there should be a dedicated cmdline option in Lmod for doing this. The only reason I'm reproducing it now is because there isn't (yet). If there was, I would use it for sure.

easybuild/tools/modules.py
stdweird: @wpoely86 why wouldn't it get installed? sure EB can do that (and package maintainers should ship it too)
what needs custiomisation? is it only `LMOD_DIR`?

rtmclay:
The script createSystemCacheFile.sh can be modified to take arguments but having the lmod command do it is just wrong.   I'm not going to take bug reports that updating the system cache fails when the user doesn't have permission to write to that directory.   I can modify it when I get back.  Or I'll take a PR on that.

easybuild/tools/modules.py
boegel: `createSystemCacheFile.sh` doesn't solve our problem.

a dedicated cmdline option in Lmod would be great though, so let's see if that can works out

This PR can be on hold until it's there (or until we get final word it won't happen), it's not going in for EB v2.0 anyway (needs more testing using `--only-module`, which is not ready yet).

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

akesandgren:
@boegel Relevance?

PR #1178, opened 2015-02-12
Fix "AttributeError: 'bool' object has no attribute 'info'" in calls to ...

riccardomurri:
...`print_msg()` and `print_error()`

Functions `print_msg` and `print_error` have an argument called `log`, which should apparently cause both functions to echo the message to the EB log file and not just to the screen.  However, the logger object is also called `log` (module-level global), and the local parameter binding shadows the module-level one.

As a result, `print_msg("foo", log=True)` errors out with `AttributeError: 'bool' object has no attribute 'info`.

This commit changes the argument name to `do_log` to avoid the clash altogether.

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -174,11 +174,11 @@ def get_log(name=None):
     log.nosupport("Use of get_log function", '2.0')
 
 
-def print_msg(msg, log=None, silent=False, prefix=True):
+def print_msg(msg, do_log=False, silent=False, prefix=True):
     """
     Print a message to stdout.
     """
-    if log:
+    if do_log:
         log.info(msg)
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

boegel:
@riccardomurri: you will need to check all uses of `print_msg` and `print_error` throughout the EB codebase too (there shouldn't be any uses of these in easyblocks)

riccardomurri:
@boegel there can't be any uses with `log=...` otherwise you've hit this
bug pretty immediately ;-)

riccardomurri:
ah, no -- sorry I see that the intended usage of the `log=...` parameter is
*not* a boolean as I had thought.

Then please discard this PR.

PR #1179, opened 2015-02-13
Make `EasyBuildLog` customizable depending on logger name.

riccardomurri:
Currently, the only need for this is to turn off `raiseError` for
GC3Pie loggers, but it could have more general applicability,
especially if more 3rd-party libraries are used.

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -67,11 +67,55 @@ class EasyBuildError(Exception):
 class EasyBuildLog(fancylogger.FancyLogger):
     """
     The EasyBuild logger, with its own error and exception functions.
+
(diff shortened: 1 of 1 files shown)


riccardomurri:
This is lighter-weight alternative to #1171 as suggested by @stdweird

boegel:
@riccardomurri: this is not needed with the easier 'hack' included in #1008, i.e. https://github.com/hpcugent/easybuild-framework/pull/1008/files#diff-30c0c724e8b33847f917f22eca0af7a0R57 ?

Also, we deprecated the `raiseError`  in `EasyBuildLog`, so in the long term this will be useless anyway?

riccardomurri:
@boegel Indeed, I think this could now be closed without merging.

PR #1180, opened 2015-02-16
Eb install script

pescobar:


easybuild/scripts/install-easybuild.sh
```diff
@@ -0,0 +1,61 @@
...
+    mkdir $INSTALL_DIR
+fi
+
+for repo in framework easyconfigs easyblocks;
+do
+    echo "downloading http://github.com/hpcugent/easybuild-${repo}/archive/${BRANCH}.tar.gz"
+    wget http://github.com/hpcugent/easybuild-${repo}/archive/${BRANCH}.tar.gz -O /tmp/eb-${repo}-${BRANCH}_$(date +"%d_%m_%Y").tar.gz >> $LOGFILE 2>&1
+    
+    echo -e "uncompressing ${repo}\n"
+    tar xf /tmp/eb-${repo}-${BRANCH}_$(date +"%d_%m_%Y").tar.gz -C $INSTALL_DIR --strip-components=1
```
stdweird: save the date in a variabel once, and use it instead of the date command

easybuild/scripts/install-easybuild.sh
```diff
@@ -0,0 +1,61 @@
...
+
+echo -e "starting installation. Logging to ${LOGFILE}\n"
+
+if [ ! -d "$INSTALL_DIR" ]; then
+    mkdir $INSTALL_DIR
+fi
+
+for repo in framework easyconfigs easyblocks;
+do
+    echo "downloading http://github.com/hpcugent/easybuild-${repo}/archive/${BRANCH}.tar.gz"
```
stdweird: set url in variabel, and use it to make sure this line and the next are the same url

stdweird:
@pescobar what different/better then the bootstrap script?

pescobar:
@stdweird I did the changes you asked for

@stdweird this is doing the same than bootstrap but for me it's more convenient when I need to do a quick test with the devel branch of easybuild because 

1) Doesn't create a module file which it's something I don't want when doing some quick test of some feature in devel
2) The bootstrap script is more complex and when you get an error it's more difficult to debug. This script is really simple, just doing few wget and creating a setup-env file to setup the environment

In any case, I wrote this script for myself because it's more convenient for how I use easybuild. If you prefer to don't merge it and keep just the bootstrap script it's fine. I will keep my script in my github account https://github.com/pescobar/random-scripts  I sent the PR just because @fgeorgatos asked for it

stdweird:
@pescobar so we need option for branches in the bootstrap script and option to disable creating module files? 

i don't know about merging in. one of the PRs for 2.0 will remove vsc-base and thus break this script

@boegel any idea if this is hard to do?

pescobar:
@stdweird I am aware this script will break when vsc-base is merged in 2.0, @boegel already told me,  but hopefully this should be fixed in this script just adding another wget line

boegel:
@pescobar: is this still relevant?

akesandgren:
@pescobar Is this still relevant? You didn't answer the last time :-)

PR #1182, opened 2015-02-16
make EasyBuildLog behave according to standard Logger API, except for EasyBuild loggers (WIP)

boegel:
replacement for #1171, #1179 by @riccardomurri 

I still need to make all other EasyBuild modules use `get_eb_logger` rather than `fancylogger.getLogger`; I like how this also avoids having to import `fancylogger` everywhere, the modules in the `easybuild` namespace shouldn't care which logger is being used (except for `build_log`)

@stdweird, @riccardomurri: thoughts?

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -69,9 +69,25 @@ class EasyBuildLog(fancylogger.FancyLogger):
     The EasyBuild logger, with its own error and exception functions.
     """
 
-    # self.raiseError can be set to False disable raising the exception which is
-    # necessary because logging.Logger.exception calls self.error
-    raiseError = True
+    def __init__(self, name):
+        """
+        Returns EasyBuildLog instance.
+        Definition of 'error' method is determine based on specified name; for loggers containing 'easybuild',
+        the 'error' method will raise an EasyBuildError next to logging an error
+        @param name: name for this logger instance
+        @return: EasyBuildLog instance
+        """
+        super(EasyBuildLog, self).__init__(name)
+
+        # self.raiseError can be set to False disable raising the exception which is
+        # necessary because logging.Logger.exception calls self.error
+        self.raiseError = True
+
+        # only use custom error method for EasyBuild loggers, adher to standard behaviour for other loggers
+        if 'easybuild' in name:
+            self.error = self.easybuild_error
+        else:
+            self.error = self.standard_error
 
     def caller_info(self):
         """Return string with caller info."""
@@ -105,12 +121,16 @@ class EasyBuildLog(fancylogger.FancyLogger):
         """Print error message for no longer supported behaviour, and raise an EasyBuildError."""
         self.error("NO LONGER SUPPORTED since v%s: %s; see %s for more information" % (ver, msg, DEPRECATED_DOC_URL))
 
-    def error(self, msg, *args, **kwargs):
+    def standard_error(self, *args, **kwargs):
+        """Standard error method, just called error method of parent."""
+        super(EasyBuildLog, self).error(*args, **kwargs)
+
+    def easybuild_error(self, msg, *args, **kwargs):
         """Print error message and raise an EasyBuildError."""
-        newMsg = "EasyBuild crashed with an error %s: %s" % (self.caller_info(), msg)
-        fancylogger.FancyLogger.error(self, newMsg, *args, **kwargs)
+        new_msg = "EasyBuild crashed with an error %s: %s" % (self.caller_info(), msg)
+        super(EasyBuildLog, self).error(new_msg, *args, **kwargs)
         if self.raiseError:
-            raise EasyBuildError(newMsg)
+            raise EasyBuildError(new_msg)
 
     def exception(self, msg, *args):
         """Print exception message and raise EasyBuildError."""
@@ -158,6 +178,15 @@ def init_logging(logfile, logtostdout=False, testing=False):
     log = fancylogger.getLogger(fname=False)
 
     return log, logfile
+
+
+def get_eb_logger(name=None, *args, **kwargs):
+    """Get logger instance."""
+    # prefix name of all EasyBuild loggers (except root logger) so we can recognize them easily
+    if name is not None:
+        name = 'easybuild.%s' % name
+
+    return fancylogger.getLogger(*args, name=name, **kwargs)
 
 
 def stop_logging(logfile, logtostdout=False):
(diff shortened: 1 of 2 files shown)


easybuild/tools/build_log.py
```diff
@@ -69,9 +69,25 @@ class EasyBuildLog(fancylogger.FancyLogger):
...
+        super(EasyBuildLog, self).__init__(name)
+
+        # self.raiseError can be set to False disable raising the exception which is
+        # necessary because logging.Logger.exception calls self.error
+        self.raiseError = True
+
+        # only use custom error method for EasyBuild loggers, adher to standard behaviour for other loggers
+        if 'easybuild' in name:
+            self.error = self.easybuild_error
+        else:
```
stdweird: the else block is not needed

easybuild/tools/build_log.py
```diff
@@ -105,12 +121,16 @@ def nosupport(self, msg, ver):
         """Print error message for no longer supported behaviour, and raise an EasyBuildError."""
         self.error("NO LONGER SUPPORTED since v%s: %s; see %s for more information" % (ver, msg, DEPRECATED_DOC_URL))
 
-    def error(self, msg, *args, **kwargs):
+    def standard_error(self, *args, **kwargs):
```
stdweird: not needed

easybuild/tools/build_log.py
```diff
@@ -160,6 +180,15 @@ def init_logging(logfile, logtostdout=False, testing=False):
     return log, logfile
 
 
+def get_eb_logger(name=None, *args, **kwargs):
```
stdweird: if you are going to trust the `get_eb_logger`, you might as well not need trust it and return the EasyBuildLog equivalent of getLogger (and not modify EasyBuildLog)

easybuild/tools/build_log.py
riccardomurri: why not? without it, `EasyBuildLog.error` is not defined for loggers not in the `easybuild.` hierarchy

easybuild/tools/build_log.py
riccardomurri: ditto: see comment to line 89 above

easybuild/tools/build_log.py
riccardomurri: ah, I stand corrected: `EasyBuildLog` inherits from `FancyLogger` which in turn inherits from `logging.Logger`.  you're right: the else branch is not needed and neither is the definition of `standard_error`

easybuild/tools/build_log.py
```diff
@@ -69,9 +69,25 @@ class EasyBuildLog(fancylogger.FancyLogger):
...
+        @return: EasyBuildLog instance
+        """
+        super(EasyBuildLog, self).__init__(name)
+
+        # self.raiseError can be set to False disable raising the exception which is
+        # necessary because logging.Logger.exception calls self.error
+        self.raiseError = True
+
+        # only use custom error method for EasyBuild loggers, adher to standard behaviour for other loggers
+        if 'easybuild' in name:
```
riccardomurri: you might want to change this into::

    if name.startswith('easybuild'):

otherwise a name like `gc3pie.aux.easybuild` would match, and it should not.

easybuild/tools/build_log.py
```diff
@@ -69,9 +69,25 @@ class EasyBuildLog(fancylogger.FancyLogger):
...
+        Definition of 'error' method is determine based on specified name; for loggers containing 'easybuild',
+        the 'error' method will raise an EasyBuildError next to logging an error
+        @param name: name for this logger instance
+        @return: EasyBuildLog instance
+        """
+        super(EasyBuildLog, self).__init__(name)
+
+        # self.raiseError can be set to False disable raising the exception which is
+        # necessary because logging.Logger.exception calls self.error
+        self.raiseError = True
```
riccardomurri: I would only define this if error == easybuild_error; otherwise it's unused but not evidently so.

easybuild/tools/build_log.py
```diff
@@ -105,12 +121,16 @@ def nosupport(self, msg, ver):
...
-    def error(self, msg, *args, **kwargs):
+    def standard_error(self, *args, **kwargs):
+        """Standard error method, just called error method of parent."""
+        super(EasyBuildLog, self).error(*args, **kwargs)
+
+    def easybuild_error(self, msg, *args, **kwargs):
         """Print error message and raise an EasyBuildError."""
-        newMsg = "EasyBuild crashed with an error %s: %s" % (self.caller_info(), msg)
-        fancylogger.FancyLogger.error(self, newMsg, *args, **kwargs)
+        new_msg = "EasyBuild crashed with an error %s: %s" % (self.caller_info(), msg)
```
riccardomurri: Note that this still has the same problem with varargs that Jens reported.  You should interpolate `args` into `msg`, e.g.::

    
+ new_msg = "EasyBuild crashed with an error %s: %s" % (self.caller_info(), (msg % args))

easybuild/tools/build_log.py
```diff
@@ -105,12 +121,16 @@ def nosupport(self, msg, ver):
...
+        super(EasyBuildLog, self).error(new_msg, *args, **kwargs)
```
riccardomurri: If you interpolate `args` into `msg` as suggested above, `args` is not needed here.

easybuild/tools/build_log.py
riccardomurri: why introduce this function? if you are going to change all source files and replace calls to `fancylogger.getLogger(name) with calls to `get_eb_logger(name)`, then you can just replace them with `fancylogger.getLogger('easybuild.' + name)` and spare yourself adding another function...

easybuild/tools/filetools.py
```diff
@@ -41,15 +41,13 @@
...
 
-import easybuild.tools.environment as env
-from easybuild.tools.build_log import print_msg  # import build_log must stay, to activate use of EasyBuildLog
+from easybuild.tools.build_log import get_eb_logger, print_msg  # this import must stay, to activate use of EasyBuildLog
 from easybuild.tools.config import build_option
 from easybuild.tools import run
 
 
-_log = fancylogger.getLogger('filetools', fname=False)
+_log = get_eb_logger('filetools', fname=False)
```
riccardomurri: see above: why not just::

    _log = fancylogger.getLogger('easybuild.filetools', fname=False)

boegel:
closed, going through with #1183 instead

PR #1184, opened 2015-02-17
enhance bootstrap script to allow bootstrapping using supplied tarballs

boegel:
Adds support for bootstrapping from supplied tarballs, by setting `$EASYBUILD_BOOTSTRAP_TARBALLS` to a location where source tarballs named `easybuild-framework.tar.gz`, `vsc-base.tar.gz` etc. are provided before using the `bootstrap_eb.py` script.

Other enhancements:

* allow skipping stage 0 (temporary installation of `distribute` which delivers `easy_install`) by defining `$EASYBUILD_BOOTSTRAP_SKIP_STAGE0`
* enabling/disabling debug logging via `$EASYBUILD_BOOTSTRAP_DEBUG`

This requires https://github.com/hpcugent/easybuild-easyblocks/pull/559 in order to make the installation using local tarballs to work correctly.

The 'normal' bootstrap from PyPI should be unaffected (i.e. it still works for me ;-)). Installing both with and without providing `vsc-base.tar.gz` works.

After playing around with this, I'm close to rewriting `bootstrap_eb.py` in `bootstrap_eb.sh`, to make it simpler and easier to read...


boegel:
@stdweird, @wpoely86: please review/test?

pforai:
Works like a charm for me on MacOS 10.10.2!

easybuild/scripts/bootstrap_eb.py
```diff
@@ -448,16 +511,18 @@ def main():
 toolchain = {'name': 'dummy', 'version': 'dummy'}
 
 source_urls = [
-               'http://pypi.python.org/packages/source/e/easybuild-framework/',
-               'http://pypi.python.org/packages/source/e/easybuild-easyblocks/',
-               'http://pypi.python.org/packages/source/e/easybuild-easyconfigs/',
-              ]
+    'https://pypi.python.org/packages/source/v/vsc-base/',
+    'http://pypi.python.org/packages/source/e/easybuild-framework/',
```
pforai: Can you bump the http to httpS for all URLs?

boegel: excellent suggestion, thanks!

easybuild/scripts/bootstrap_eb.py
```diff
@@ -50,8 +51,7 @@
 from distutils.version import LooseVersion
 
 # set print_debug to True for detailed progress info
-print_debug = False
-#print_debug = True
+print_debug = os.environ.get('EASYBUILD_BOOTSTRAP_DEBUG', False)
```
stdweird: can you check if this is false if `EASYBUILD_BOOTSTRAP_DEBUG=0`?

easybuild/scripts/bootstrap_eb.py
```diff
@@ -221,37 +226,71 @@ def stage0(tmpdir):
...
     os.environ['PYTHONPATH'] = ''
 
-    versions = {}
+    # template string to inject in template easyconfig
+    templates = {}
 
     pkg_egg_dir_framework = None
-    for pkg in ['easyconfigs', 'easyblocks', 'framework']:
-        pkg_egg_dir = find_egg_dir_for(targetdir_stage1, 'easybuild-%s' % pkg)
+    for pkg in ['vsc-base', 'easybuild-framework', 'easybuild-easyblocks', 'easybuild-easyconfigs']:
```
stdweird: make the list of pkgs a constant (and use it also above)

easybuild/scripts/bootstrap_eb.py
```diff
@@ -221,37 +226,71 @@ def stage0(tmpdir):
...
+        templates.update({pkg: ''})
+
+        pkg_egg_dir = find_egg_dir_for(targetdir_stage1, pkg)
+        if pkg_egg_dir is None:
+            if pkg == 'vsc-base':
+                # vsc-base is optional in older EasyBuild versions
```
stdweird: ? (use the older bootstrap to install older EB?)

boegel: no, we need a single bootstrap for both versions (and at this point, we don't know the version being installed yet)

easybuild/scripts/bootstrap_eb.py
```diff
@@ -259,24 +298,34 @@ def stage1(tmpdir):
...
         res = version_regex.search(pkg_egg_dirname)
         if res is not None:
             pkg_version = res.group(1)
-            versions.update({'%s_version' % pkg: pkg_version})
             debug("Found version for easybuild-%s: %s" % (pkg, pkg_version))
+            if sources_path is None:
+                # downloaded tarballs should be versioned
+                templates.update({pkg: "'%s-%s.tar.gz'," % (pkg, pkg_version)})
+            else:
+                # names of specified source tarballs should not contain a version
```
stdweird: why not? you allow it in the glob above?

easybuild/scripts/bootstrap_eb.py
```diff
@@ -259,24 +298,34 @@ def stage1(tmpdir):
...
+                templates.update({pkg: "'%s.tar.gz'," % pkg})
         else:
-            error("Failed to determine version for easybuild-%s package from %s with %s" % (pkg, pkg_egg_dirname, version_regex.pattern))
+            # vsc-base is only required for EasyBuild v2.x or higher
```
stdweird: and that's what this bootstrap will try to install.

easybuild/scripts/bootstrap_eb.py
```diff
@@ -372,6 +425,12 @@ def main():
         error("Usage: %s <install path>" % sys.argv[0])
     install_path = os.path.abspath(sys.argv[1])
 
+    sources_path = os.environ.get('EASYBUILD_BOOTSTRAP_TARBALLS')
```
stdweird: teh environment variable name should refelctthis is a path of some sort

easybuild/scripts/bootstrap_eb.py
```diff
@@ -372,6 +425,12 @@ def main():
...
+    if sources_path is not None:
+        info("Fetching sources from %s..." % sources_path)
+
+    skip_stage0 = os.environ.get('EASYBUILD_BOOTSTRAP_SKIP_STAGE0', False)
```
stdweird: same check behaviour of `=0`

easybuild/scripts/bootstrap_eb.py
```diff
@@ -448,16 +511,18 @@ def main():
 toolchain = {'name': 'dummy', 'version': 'dummy'}
 
 source_urls = [
-               'http://pypi.python.org/packages/source/e/easybuild-framework/',
-               'http://pypi.python.org/packages/source/e/easybuild-easyblocks/',
-               'http://pypi.python.org/packages/source/e/easybuild-easyconfigs/',
-              ]
+    'https://pypi.python.org/packages/source/v/vsc-base/',
```
stdweird: use a variable for the common url `https://pypi.python.org/packages/source`

stdweird: i would make these also templates, so it's easier to support download from github branch in a next PR :smile:

easybuild/scripts/bootstrap_eb.py
boegel: hmm, no, it's not, since the `0` is taken as a string value, and a non-empty string is equivalent to `True`

not sure how to tackle this one...

but since this is disabled by default, it's probably not a disaster if setting this to `0` enables it?

easybuild/scripts/bootstrap_eb.py
boegel: not until it's released, and it should be possible to bootstrap older versions too (by supplying the tarballs for v1.16.1, for example)

easybuild/scripts/bootstrap_eb.py
boegel: hmm, probably doesn't matter indeed, I'll check on that

easybuild/scripts/bootstrap_eb.py
boegel: changed to `EASYBUILD_BOOTSTRAP_SOURCEPATH`

easybuild/scripts/bootstrap_eb.py
boegel: that'll make the names of the source tarballs trickier... we need separate tarballs for each of the packages, and naming them all `develop.tar.gz` is not going to work

boegel:
@stdweird: remarks fixed, good to go in?

stdweird:
also, can you show me the link to the diocs that explain this new functionality?

boegel:
updated docs: https://github.com/hpcugent/easybuild/pull/87

boegel:
updated docs merged, so this is going in too, thanks for the review @stdweird, and @pforai for retesting the bootstrap script!

PR #1185, opened 2015-02-19
disable updating of Lmod cache by default

boegel:
simplified version of #1177 that should go into EasyBuild v2.0, since it changes default behaviour

this only disables updating the Lmod user cache by default, and allows to re-enable it via `--update-modules-tool-cache`; #1177 goes a lot further, adding support for updating the cache in other locations, compiling the Lmod caches, etc.

@stdweird: please review


easybuild/tools/modules.py
```diff
@@ -63,6 +63,8 @@
 # see e.g., https://bugzilla.redhat.com/show_bug.cgi?id=719785
 LD_LIBRARY_PATH = os.getenv('LD_LIBRARY_PATH', '')
 
+LMOD_USER_CACHE_DIR = os.path.join(os.path.expanduser('~'), '.lmod.d', '.cache')
```
stdweird: the cache location should be a class constant of each module tool (class constants should be neutral).

easybuild/tools/modules.py
```diff
@@ -833,31 +835,32 @@ def available(self, mod_name=None):
 
     def update(self):
         """Update after new modules were added."""
-        spider_cmd = os.path.join(os.path.dirname(self.cmd), 'spider')
-        cmd = [spider_cmd, '-o', 'moduleT', os.environ['MODULEPATH']]
-        self.log.debug("Running command '%s'..." % ' '.join(cmd))
+        if build_option('update_modules_tool_cache'):
```
stdweird: skip the unnecssary indentation with 
```python
if not build_option('update_modules_tool_cache'):
    self.log.debug("not updating cache")
    return
```

easybuild/tools/options.py
```diff
@@ -199,8 +201,8 @@ def override_options(self):
             'skip-test-cases': ("Skip running test cases", None, 'store_true', False, 't'),
             'umask': ("umask to use (e.g. '022'); non-user write permissions on install directories are removed",
                       None, 'store', None),
-            'optarch': ("Set architecture optimization, overriding native architecture optimizations",
-                        None, 'store', None),
+            'update-modules-tool-cache': ("Update modules tool cache file(s) after generating module file",
```
stdweird: why default false?

boegel: because now the user cache of the user running EasyBuild is being updated, only to then be totally ignored by EasyBuild itself (since `$LMOD_IGNORE_CACHE` is defined)

this should never have been enabled to start with...

easybuild/tools/modules.py
boegel: we've tried to stay away from in-line `return` statements, so I'm not in favour of this...

wpoely86:
looks fine

boegel:
Jenkins: test this please

easybuild/tools/options.py
stdweird: so updating lmod cache should done separately from installing software with EB? link to the docs wrt best policy would be nice :smile:

easybuild/tools/modules.py
stdweird: ah, ok then.

easybuild/tools/options.py
boegel: in general, yes, although with #1177 we will try and add support to update the cache via EB too, I'll look into detailed documentation on this then

for now, mentioning in the release notes that updating the *user* cache by default requires using `--update-modules-tool-cache` should be sufficient

that's a breaking change, so it makes sense to include this for EasyBuild v2.0

boegel:
going in, thanks for the review @wpoely86 and @stdweird!

PR #1190, opened 2015-02-25
Some python utility scripts that may be of use.

ebgregory:
Some python utility scripts that may be of use for keeping track of what has been installed by easybuild.
make_sw_database.py walks the directory structure looking for *.eb files and prints out a database that can be copied to an accessible location. Then query this database with the other scripts. Good luck.


JensTimmerman:
First of, thank you for sharing!Could you please add a bit of documentation or an example on how to use this?

PR #1191, opened 2015-02-26
drop custom setUp/tearDown in toolchain tests which are hard modifying $MODULEPATH

boegel:
The toolchain unit tests hard modifying `$MODULEPATH` caused some issues with these tests on a system where only `modulecmd.tcl` was available; the tests seemed to get stuck in an infinite loop.

Simply dropping the custom `setUp` and `tearDown` in `test/framework/toolchain.py` fixes the issues, since (re)setting `$MODULEPATH` is already properly done in `test/framework/utilities.py`.

Thanks to @ocaisa to help figure this out.


boegel:
only affects the (toolchain) tests, and only removal of code, so nothing much to review

thanks again for the help in figuring this out @ocaisa!

PR #1192, opened 2015-02-27
use available which() function, rather than running 'which' via run_cmd

boegel:
avoid dependency on external `which`

@wpoely86: please review


wpoely86:
looks fine

boegel:

PR #1193, opened 2015-02-27
fix install-EasyBuild-develop.sh script w.r.t. vsc-base dependency

boegel:
@besserox: please review?

### DIFF
--- a/easybuild/scripts/install-EasyBuild-develop.sh
+++ b/easybuild/scripts/install-EasyBuild-develop.sh
@@ -10,11 +10,11 @@ set -e
 # Print script help
 print_usage()
 {
-    echo "Usage: $0 github_username install_directory"
+    echo "Usage: $0 <github_username> <install_dir>"
     echo
-    echo "    github_username:     username on GitHub for which the easybuild repositories should be cloned"
+    echo "    github_username:     username on GitHub for which the EasyBuild repositories should be cloned"
     echo
-    echo "    install_directory:   directory were all the EasyBuild files will be installed"
+    echo "    install_dir:         directory were all the EasyBuild files will be installed"
     echo
 }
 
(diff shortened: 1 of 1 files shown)


boegel:
Hmm, need to look into this:

```
=== Cloning boegel/easybuild-easyconfigs ...
...
=== Checking out the 'develop' branch
fatal: A branch named 'develop' already exists.
```

besserox:
I tried the new version of the script. I forked the vsc repository on github account and the installation was fine. Then when I tried to run EasyBuild, I get the following error:

```
$ eb -h
  File "/home/xavier/tmp/eb/install/easybuild-framework/easybuild/main.py", line 45, in <module>
    from easybuild.tools.build_log import EasyBuildError, init_logging, print_msg, print_error, stop_logging
  File "/home/xavier/tmp/eb/install/easybuild-framework/easybuild/tools/build_log.py", line 38, in <module>
ImportError: No module named vsc.utils

$ echo $PYTHONPATH
/home/xavier/tmp/eb/install/easybuild-easyconfigs:/home/xavier/tmp/eb/install/easybuild-easyblocks:/home/xavier/tmp/eb/install/easybuild-framework:/home/xavier/tmp/eb/install/vsc-base
```

If I use `$root/vsc-base/lib` instead of `$root/vsc-base` in my PYTHONPATH, it seems to work.

easybuild/scripts/install-EasyBuild-develop.sh
```diff
@@ -69,6 +69,7 @@ conflict    EasyBuild
 
 prepend-path    PATH            "\$root/easybuild-framework"
 
+prepend-path    PYTHONPATH      "\$root/vsc-base"
```
besserox: It looks like it should be `prepend-path PYTHONPATH "\$root/vsc-base/lib"` here.

boegel: You're right, will fix.

I was still testing the updated script myself as well, but bumped into another issue (that I still need to figure out).

besserox:
It looks good to me now. I tested and it works fine.

boegel:
@besserox: thanks for the feedback!

Issue #1195, opened 2015-02-27
New Unit Test are Picking up EasyBlocks from $PYTHONPATH


pforai:
This should have not been the case with the develop branch of framework as of today, but on my testing environment I was working on the boost block for Cray and the test have been failing because the block was visible.

boegel:
fix in #1196, so closing this

thanks for reporting this @pforai!

PR #1196, opened 2015-03-02
isolate unit tests from available easyblocks other than those included in the tests

boegel:
fix for #1195 

@JensTimmerman, @pforai: please review


JensTimmerman:
can't we just do local imports now to import the correct file?

boegel:
no, this change is required to for example isolate the test for `--list-easyblocks` from easyblocks other than the ones included in the tests themselves

we want to test the functionality that supports `--list-easyblocks`, not whether the easyblocks we happen to pick up have no syntax errors...

JensTimmerman:
ok

boegel:
Thanks @JensTimmerman!

PR #1146, opened 2015-01-23
fix hardcoding of /tmp in mpi_cmd_for

boegel:
@wpoely86: please review

PR #1200, opened 2015-03-03
keep $I_MPI_MPD_TMPDIR short to avoid 'socket.error: AF_UNIX path too long

boegel:
This fixes problems with the use of `mpi_cmd_for`, that manifest themselves like:

```
mpdboot_node2025.delcatty.os (handle_mpd_output 980): failed to ping mpd on localhost; received output={}
mpdboot_node2025.delcatty.os (handle_mpd_output 982): Please examine the /tmp/mpd2.logfile_vsc40023 log file on each node of the ring
```

Apparently, the value for `I_MPI_MPD_TMPDIR` must be kept *very* short; see also https://software.intel.com/sites/products/documentation/hpc/ics/impi/41/lin/Reference_Manual/index.htm#Environment_Variables_Multipurpose_Daemon_Commands.htm .

@wpoely86: please review

### DIFF
--- a/easybuild/tools/toolchain/mpi.py
+++ b/easybuild/tools/toolchain/mpi.py
@@ -197,7 +197,7 @@ class Mpi(Toolchain):
                 self.log.warning("$I_MPI_MPD_TMPDIR should be (very) short to avoid problems: %s" % mpd_tmpdir)
 
             # temporary location for mpdboot and nodes files
-            tmpdir = tempfile.mkdtemp(prefix='eb-mpi_cmd_for-')
+            tmpdir = tempfile.mkdtemp(prefix='mpi_cmd_for-')
 
             # set PBS_ENVIRONMENT, so that --file option for mpdboot isn't stripped away
             env.setvar('PBS_ENVIRONMENT', "PBS_BATCH_MPI")
--- a/easybuild/tools/toolchain/mpi.py
+++ b/easybuild/tools/toolchain/mpi.py
@@ -188,11 +188,16 @@ class Mpi(Toolchain):
         # Intel MPI mpirun needs more work
         if mpi_family == toolchain.INTELMPI:  # @UndefinedVariable
 
-            tmpdir = tempfile.mkdtemp(prefix='eb-mpi_cmd_for-')
-
             # set temporary dir for mdp
             # note: this needs to be kept *short*, to avoid mpirun failing with "socket.error: AF_UNIX path too long"
+            # exact limit is unknown, but ~20 characters seems to be OK
             env.setvar('I_MPI_MPD_TMPDIR', tempfile.gettempdir())
+            mpd_tmpdir = os.environ['I_MPI_MPD_TMPDIR']
+            if len(mpd_tmpdir) > 20:
+                self.log.warning("$I_MPI_MPD_TMPDIR should be (very) short to avoid problems: %s" % mpd_tmpdir)
+
+            # temporary location for mpdboot and nodes files
+            tmpdir = tempfile.mkdtemp(prefix='eb-mpi_cmd_for-')
 
             # set PBS_ENVIRONMENT, so that --file option for mpdboot isn't stripped away
             env.setvar('PBS_ENVIRONMENT', "PBS_BATCH_MPI")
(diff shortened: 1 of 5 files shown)


boegel:
These problems are caused by the change made in #1146

easybuild/tools/toolchain/mpi.py
```diff
@@ -191,7 +191,8 @@ def mpi_cmd_for(self, cmd, nr_ranks):
             tmpdir = tempfile.mkdtemp(prefix='eb-mpi_cmd_for-')
 
             # set temporary dir for mdp
-            env.setvar('I_MPI_MPD_TMPDIR', tmpdir)
+            # note: this needs to be kept *short*, to avoid mpirun failing with "socket.error: AF_UNIX path too long"
+            env.setvar('I_MPI_MPD_TMPDIR', tempfile.gettempdir())
```
wpoely86: isn't not an option to use a very short prefix?

easybuild/tools/toolchain/mpi.py
```diff
@@ -191,7 +191,8 @@ def mpi_cmd_for(self, cmd, nr_ranks):
             tmpdir = tempfile.mkdtemp(prefix='eb-mpi_cmd_for-')
```
wpoely86: is this still needed then?

boegel: yes, it's used in other places, for files we produce ourselves (`mpdboot`, `nodes`)

easybuild/tools/toolchain/mpi.py
boegel: well, actually this could be too long as well, depending on what `$TMPDIR` or `--tmpdir` was set to...

We want to set this to something that avoids clashing with other users on the same system (so the default, or previously hard-set, `/tmp` isn't a good option), but it should also be kept as short as possible...

wpoely86: do we know the limit? Else a warning should be given if the path is longer then a certain amount of chars?

boegel: no, and it seems like the limit is system-dependent:

```
The mpd2.console_* file full path length can be limited in some operating systems. You hit
this limitation if you get the following diagnostic message: socket.error: AF_UNIX path
too long. Decrease the length of the <directory name> string to avoid this issue. 
```

boegel: how about something like:

```python
mpd_tmpdir_env_var = 'I_MPI_MPD_TMPDIR'
mpd_tmpdir_limit = 10
# set sensible value for $I_MPI_MPD_TMPDIR, unless it's set already
if os.getenv(mpd_tmpdir_env_var) is None:
  env.setvar(mpd_tmpdir_env_var, tempfile.gettempdir())
# check whether value for $I_MPI_MPD_TMPDIR isn't too long, to avoid problems
# e.g., mpdboot failing with "failed to ping mpd on localhost" due to "socket.error: AF_UNIX path too long"
if len(os.environ[mpd_tmpdir_env_var]) > mpd_tmpdir_limit:
    tup = (mpd_tmpdir_limit, os.environ[mpd_tmpdir_env_var])
    self.log.warning("Value for $%s must be (very) short to avoid problems, current value is longer than %d: %s" % tup)
```

boegel: just checking, using `tempfile.gettmpdir()` rather than `tmpdir` fixes the problem, i.e. a value like `/tmp/easybuild-F_B4M3` *is* short enough

boegel: on second thought, just using `$I_MPI_MPD_TMPDIR` if it's already defined makes `eb` more vulnerable to outside influence, which is probably not a good thing

easybuild/tools/toolchain/mpi.py
```diff
@@ -188,10 +188,16 @@ def mpi_cmd_for(self, cmd, nr_ranks):
         # Intel MPI mpirun needs more work
         if mpi_family == toolchain.INTELMPI:  # @UndefinedVariable
 
-            tmpdir = tempfile.mkdtemp(prefix='eb-mpi_cmd_for-')
-
             # set temporary dir for mdp
-            env.setvar('I_MPI_MPD_TMPDIR', tmpdir)
+            # note: this needs to be kept *short*, to avoid mpirun failing with "socket.error: AF_UNIX path too long"
+            # exact limit is unknown, but ~20 characters seems to be OK
```
wpoely86: maybe add the link you send me?

boegel: can't include it without breaking the column limit :package: 

it's mentioned in this PR, and it turns up pretty quickly when googling `AF_UNIX path too long` mentioned here, so should be fine without it

boegel:

PR #1206, opened 2015-03-05
remove fetch_parameter_from_easyconfig_file rather than deprecating it

boegel:
no need to deprecate something if it's internal to the framework (i.e., not meant to be used in easyblocks), since that means dragging it along (and documenting it) for quite some time to come...

@wpoely86: please review?


wpoely86:
looks fine

boegel:
Thanks @wpoely86!

PR #1209, opened 2015-03-06
add 'bundle' default moduleclass

boegel:


easybuild/tools/config.py
```diff
@@ -146,6 +146,7 @@ def mk_full_default_path(name, prefix=DEFAULT_PREFIX):
 # https://wickie.hlrs.de/platforms/index.php/Application_software_packages
 DEFAULT_MODULECLASSES = [
     ('base', "Default module class"),
+    ('bundle', "Bundle of software"),
```
wpoely86: `software/modules`?

fgeorgatos: could be also as wpoely86 comments; 

+1 for the addition of the module class

boegel:
nevermind, after a brief discussion with @wpoely86 I agree this has little meaning, so closing

PR #1212, opened 2015-03-06
make bootstrap script robust against having 'vsc-base' already available in Python search path

boegel:
bootstrapping EasyBuild v2.0.0 from PyPi fails without this, if the `vsc` namespace is available in the Python search path

### DIFF
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -461,6 +461,9 @@ def main():
         # exclude path if it's potentially an EasyBuild/VSC package, providing the 'easybuild'/'vsc' namespace, resp.
         if any([os.path.exists(os.path.join(path, pkg, '__init__.py')) for pkg in ['easybuild', 'vsc']]):
             include_path = False
+        # exclude any .egg paths
+        if path.endswith('.egg'):
+            include_path = False
         # exclude any path that contains an easy-install.pth file
         if os.path.exists(os.path.join(path, 'easy-install.pth')):
             include_path = False
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -458,12 +458,10 @@ def main():
     sys.path = []
     for path in orig_sys_path:
         include_path = True
-        # exclude path if it's potentially an EasyBuild package
-        if 'easybuild' in path or os.path.exists(os.path.join(path, 'easybuild')):
+        # exclude path if it's potentially an EasyBuild/VSC package, providing the 'easybuild'/'vsc' namespace, resp.
+        if any([os.path.exists(os.path.join(path, pkg, '__init__.py')) for pkg in ['easybuild', 'vsc']]):
             include_path = False
-        if '/vsc_' in path or os.path.exists(os.path.join(path, 'vsc')):
-            include_path = False
-        # exclude path if it contain an easy-install.pth file
+        # exclude any path that contains an easy-install.pth file
         if os.path.exists(os.path.join(path, 'easy-install.pth')):
             include_path = False
 
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -454,13 +454,10 @@ def main():
         # exclude path if it's potentially an EasyBuild package
         if 'easybuild' in path or os.path.exists(os.path.join(path, 'easybuild')):
             include_path = False
-        elif '/vsc_' in path or os.path.exists(os.path.join(path, 'vsc')):
-            include_path = False
-        # exclude path if it provides 'distribute' package
-        elif 'distribute' in path or os.path.exists(os.path.join(path, 'distribute')):
+        if '/vsc_' in path or os.path.exists(os.path.join(path, 'vsc')):
             include_path = False
         # exclude path if it contain an easy-install.pth file
-        elif os.path.exists(os.path.join(path, 'easy-install.pth')):
+        if os.path.exists(os.path.join(path, 'easy-install.pth')):
             include_path = False
 
         if include_path:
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

wpoely86:
looks fine

easybuild/scripts/bootstrap_eb.py
```diff
@@ -58,6 +59,10 @@
 # set print_debug to True for detailed progress info
 print_debug = os.environ.get('EASYBUILD_BOOTSTRAP_DEBUG', False)
 
+# don't add user site directory to sys.path (equivalent to python -s), see https://www.python.org/dev/peps/pep-0370/
```
stdweird: this also applies to regular eb usage, no?

easybuild/scripts/bootstrap_eb.py
```diff
@@ -519,6 +532,8 @@ def main():
 # EasyBuild is a (set of) Python packages, so it depends on Python
 # usually, we want to use the system Python, so no actual Python dependency is listed
 allow_system_deps = [('Python', SYS_PYTHON_VERSION)]
+
+preinstallopts = '%(preinstallopts)s'
```
stdweird: where's the PR for the regualr EB easyconfigs? (and why are we still not using those?)

easybuild/scripts/bootstrap_eb.py
```diff
@@ -448,7 +459,9 @@ def main():
     for path in orig_sys_path:
         include_path = True
         # exclude path if it's potentially an EasyBuild package
-        if 'easybuild' in path:
+        if 'easybuild' in path or os.path.exists(os.path.join(path, 'easybuild')):
```
stdweird: what is the `'easybuild' in path` trying to achieve?

easybuild/scripts/bootstrap_eb.py
```diff
@@ -448,7 +459,9 @@ def main():
...
+            include_path = False
+        if '/vsc_' in path or os.path.exists(os.path.join(path, 'vsc')):
```
stdweird: and what's the `/vsc_`  trying to solve?

easybuild/scripts/bootstrap_eb.py
boegel: not sure what you mean?

these?

v1.16.2: https://github.com/hpcugent/easybuild-easyconfigs/pull/1457
v2.0.0: https://github.com/hpcugent/easybuild-easyconfigs/pull/1456

boegel: there are a couple of reasons why we can't (easily) use the readily available EasyBuild easyconfig files; this `preinstallopts` part, which is specific to the bootstrap, is one

only knowing the version to include in here *after* stage1 (which installs the latest EasyBuild packages (+ vsc-base) from PyPI), is another

easybuild/scripts/bootstrap_eb.py
boegel: excluding any entries in `sys.path` that look like they have something to do with EasyBuild

it's OK if we exclude more (since we don't rely on stuff being there already), but we can't easily drop the entire `sys.path`

easybuild/scripts/bootstrap_eb.py
boegel: excluding any entries in sys.path that look like they have something to do with VSC tools (vsc-base, or something else)

the `/` and `_` are there specifically to recognize `.egg` dirs, mostly

`vsc` by itself seems a bit too generic to exclude paths, it *may* be a partial match for a username, for example

easybuild/scripts/bootstrap_eb.py
boegel: maybe, yes... should probably also go in `PythonPackage`

rjeschmi:
Was failing on my docker builds, but this PR has resolved that issue

easybuild/scripts/bootstrap_eb.py
stdweird: i guess it's not ok when user `easybuild` installs EB in the homedir?

easybuild/scripts/bootstrap_eb.py
stdweird: well, good luck we didn't mount our filesystems under `/vsc_mnt` then..... (my point is, this is just excluding some random strings, have nothing to do with python modules)

easybuild/scripts/bootstrap_eb.py
stdweird: i was talking about the easyconfig for easybuild itself. i don't see why the bootrap easyconfig for EB is different from the one in the repos (i also don't see why don't use the one from the repos)

boegel: We need to include a (template) easyconfig for EasyBuild in the bootstrap script because:
(i) we (now) also support *offline* bootstrapping (by providing tarballs for each of the EasyBuild components)
(ii) we take control over the `easy_install`/`setuptools`/`distribute` being used by the bootstrap, to have more control

We currently don't do (ii) when installing EasyBuild with EasyBuild outside of the bootstrap, since things are less tricky there. Maybe we should look into that too in that situation, but that shouldn't block this PR, imho.

For other Python packages being installed using EasyBuild, you're usually using a Python module installed with EasyBuild, which includes a setuptools, so there it's typically not an issue.

easybuild/scripts/bootstrap_eb.py
boegel: OK, I'll look into fixing this... Just checking for `vsc/__init__.py` should do it, actually.

easybuild/scripts/bootstrap_eb.py
boegel: see https://github.com/hpcugent/easybuild-easyblocks/pull/577

boegel:
Going in, this should make the bootstrap significantly more robust.

Thanks for the reviewing @stdweird, @wpoely86 and testing @rjeschmi.

PR #1214, opened 2015-03-09
Sort the results of a filesearch

wpoely86:
Seems more logical to me and it shouldn't break anything...

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -366,8 +366,9 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False):
             # see http://stackoverflow.com/questions/13454164/os-walk-without-hidden-folders
             dirnames[:] = [d for d in dirnames if not d in ignore_dirs]
 
+        hits = sorted(hits)
+
         if hits:
-            hits = sorted(hits)
             common_prefix = det_common_path_prefix(hits)
             if short and common_prefix is not None and len(common_prefix) > len(var) * 2:
                 var_lines.append("%s=%s" % (var, common_prefix))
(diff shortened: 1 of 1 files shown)


wpoely86:
@boegel Jenkins says:
```
Failed to find EasyBuild vsc-base package
```

easybuild/tools/filetools.py
```diff
@@ -367,6 +367,7 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False):
             dirnames[:] = [d for d in dirnames if not d in ignore_dirs]
 
         if hits:
+            hits.sort()
```
boegel: maybe better use `hits = sorted(...)`, wherever `hits` is defined?

wpoely86:
retest this please

easybuild/tools/filetools.py
wpoely86: why? Does that make any difference?

boegel: it ensures that `hits` is always sorted, may be used deeper down or higher up (maybe not now, but later, who knows)

also, using `.sort()` means side-effects (`hits` is modified in-place, no explicit reassignment to `hits`), which makes it less easier to understand the code (it may be more efficient, but that'll be in the noise here, most likely)

wpoely86:
@boegel fixed

easybuild/tools/filetools.py
```diff
@@ -367,6 +367,7 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False):
             dirnames[:] = [d for d in dirnames if not d in ignore_dirs]
 
         if hits:
+            hits = sorted(hits)
```
boegel: isn't `hits` defined higher up? do the `sorted` then there, outside of this `if hits` block

wpoely86: why? it only makes sense to sort if there is something to sort?

boegel: again, `hits` may be used again deeper down or higher up, who knows

`sorted([])` is pretty cheap anyway ;)

boegel:
going in, sorry for being so picky @wpoely86 ;)

PR #1215, opened 2015-03-11
make bootstrap script more robust w.r.t. vsc-base component

boegel:
purposely targeting `master` branch to include #1193

should be fine to include this change in `master` without bumping the version or re-releasing, since the scripts are not part of an actual installation


boegel:
already reviewed via #1193, going into `master`...

PR #1224, opened 2015-03-17
Add support for packaging with FPM (REVIEW)

rjeschmi:
This is very probably broken right now, but shows some of the reorganization suggested in https://github.com/hpcugent/easybuild-framework/pull/1170

Since it is based on that branch it could be merged back that way too if we come up with a plan.

edit: preliminary documentation at https://rjeschmi-eb-draft.readthedocs.org/en/latest/Creating_binary_packages.html


easybuild/framework/package.py
```diff
@@ -0,0 +1,65 @@
...
+A place for packaging functions
+
+"""
+
+import os
+from easybuild.tools.run import run_cmd
+
+def package_fpm(easyblock, modfile_path ):
+    rpmname = "HPCBIOS.20150211-%s-%s" % (easyblock.name, easyblock.version)
+    workdir = "/tmp"
```
rjeschmi: this needs to be picked out of easybuild somehow

boegel: just use `tempfile.mkdtemp` to obtain a temporary directory

This will use the temporary directory set in the EasyBuild configuration.

easybuild/framework/package.py
```diff
@@ -0,0 +1,65 @@
...
+    os.chdir(workdir)
+
+    pkgtemplate = "HPCBIOS.20150211-%(name)s-%(version)s"
```
rjeschmi: I'm aware this name is lame, but will change with a config variable I expect.

easybuild/framework/package.py
```diff
@@ -0,0 +1,65 @@
...
+
+    pkgname=pkgtemplate % {
+        'name' : easyblock.name,
+        'version' : easyblock.version,
+    }
+
+    if easyblock.toolchain.name == "dummy":
+        dependencies = []
+    else:
+        dependencies = [ "=".join([ easyblock.toolchain.name, easyblock.toolchain.version ]) ]
```
boegel: toolchain is already part of list returned by `easyblock.cfg.dependencies()`

easybuild/framework/package.py
```diff
@@ -0,0 +1,65 @@
...
+        '-t', 'rpm', # target
+        '-s', 'dir', # source
+        '-C', easyblock.installdir,
+    ]
+    cmdlist.extend([ depstring ])
+    cmdlist.extend([
+        easyblock.installdir,
+    ])
+
+    (out, _) = run_cmd(cmdlist, log_all=True, simple=False)
```
boegel: just use `simple=True` (which is the default) if you're not going to do anything with `out`:


```python
run_cmd(cmdlist, log_all=True_
```

easybuild/framework/package.py
```diff
@@ -0,0 +1,65 @@
...
+
+"""
+
+import os
+from easybuild.tools.run import run_cmd
+
+def package_fpm(easyblock, modfile_path ):
+    rpmname = "HPCBIOS.20150211-%s-%s" % (easyblock.name, easyblock.version)
+    workdir = "/tmp"
+    os.chdir(workdir)
```
boegel: should be wrapped in `try... except OSError`, and you'll need to return to the dir you were in

easybuild/framework/package.py
```diff
@@ -0,0 +1,65 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+# #
+
+"""
+A place for packaging functions
+
```
boegel: add yourself as an author (probably @sntgluca too)

easybuild/framework/easyblock.py
```diff
@@ -1800,6 +1803,14 @@ def prepare_step_spec(initial):
                 lambda x: x.test_cases_step(),
             ], False))
 
+        ## CHANGE TRUE TO build_option, and 
+        ## ADD build-pkg to all the configuration dicts
+        ## # if build_option('build-pkg'):
```
boegel: see #1185 for inspiration on how to add a new build option (look for `update-modules-tool-cache` and `update_modules_tool_cache`)

easybuild/framework/easyblock.py
```diff
@@ -1784,11 +1787,11 @@ def prepare_step_spec(initial):
         # part 3: post-iteration part
         steps_part3 = [
             ('extensions', 'taking care of extensions', [lambda x: x.extensions_step()], False),
-            ('package', 'packaging', [lambda x: x.package_step()], True),
             ('postproc', 'postprocessing', [lambda x: x.post_install_step()], True),
             ('sanitycheck', 'sanity checking', [lambda x: x.sanity_check_step()], False),
             ('cleanup', 'cleaning up', [lambda x: x.cleanup_step()], False),
             ('module', 'creating module', [lambda x: x.make_module_step()], False),
+            # ('package', 'packaging', [lambda x: x.package_step()], True),
```
boegel: just remove

easybuild/framework/easyblock.py
```diff
@@ -1444,8 +1444,11 @@ def extensions_step(self, fetch=False):
         self.clean_up_fake_module(fake_mod_data)
 
     def package_step(self):
-        """Package software (e.g. into an RPM)."""
-        pass
+        """Prepare package software (e.g. into an RPM) with fpm."""
+
+        path_to_module_file = os.path.join(install_path('mod'), build_option('suffix_modules_path'), self.full_mod_name)
+
+        package_fpm(self, path_to_module_file)
```
boegel: this should be steered via the added build option

the `easybuild.tools.package` module should probably have a function named `package_with`, that gets passed a string (the value for `build_option(...)`), which tens pick the right `package_foo` function to package with

easybuild/framework/easyblock.py
```diff
@@ -56,6 +56,7 @@
 from easybuild.framework.easyconfig.easyconfig import get_easyblock_class, get_module_path, resolve_template
 from easybuild.framework.easyconfig.tools import get_paths_for
 from easybuild.framework.easyconfig.templates import TEMPLATE_NAMES_EASYBLOCK_RUN_STEP
+from easybuild.framework.package import package_fpm
```
boegel: make it `easybuild.tools.package`

boegel: maybe `easybuild.tools.packaging`?

boegel:
Jenkins: ok to test

boegel:
@rjeschmi: looking good as further work on top of #1170 

Feel free to work on this off and on, let me know if you have any questions or if you're not sure on how to tackle something (either via this PR, or via IRC, etc.).

rjeschmi:
thanks @boegel

easybuild/framework/package.py
fgeorgatos: it's a template name, each for his own ;-)
this belongs to an external definition (possibly eb configfile), be my guest

easybuild/framework/package.py
fgeorgatos: i' interested to be part of this, minimum as a tester

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,67 @@
...
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+# #
+
+"""
+A place for packaging functions
+
+@author: Marc Litherland
+@author: Gianluca Santarossa 
+@author: Robert Schmidt (Ottawa Hospital Research Institute)
```
fgeorgatos: @author: Fotis Georgatos

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,67 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+# #
+
+"""
+A place for packaging functions
+
+@author: Marc Litherland
+@author: Gianluca Santarossa
```
boegel: mention company (Novartis)

easybuild/tools/packaging.py
boegel: +1

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,67 @@
...
+@author: Marc Litherland
+@author: Gianluca Santarossa 
+@author: Robert Schmidt (Ottawa Hospital Research Institute)
+"""
+
+import os
+from easybuild.tools.run import run_cmd
+
+def package_fpm(easyblock, modfile_path ):
+    rpmname = "HPCBIOS.20150211-%s-%s" % (easyblock.name, easyblock.version)
```
stdweird: remove since no longer needed

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,67 @@
...
+
+@author: Marc Litherland
+@author: Gianluca Santarossa 
+@author: Robert Schmidt (Ottawa Hospital Research Institute)
+"""
+
+import os
+from easybuild.tools.run import run_cmd
+
+def package_fpm(easyblock, modfile_path ):
```
stdweird: add docstring

stdweird: and is this really easyblock? or is thei sthe easyconfig instance.

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,67 @@
...
+    rpmname = "HPCBIOS.20150211-%s-%s" % (easyblock.name, easyblock.version)
+    workdir = tempfile.mkdtemp()
+    try:
+        os.chdir(workdir)
+    except OSError, err:
+        _log.error("Failed to chdir into workdir: %s : %s" % (workdir, err))
+
+    pkgtemplate = "HPCBIOS.20150211-%(name)s-%(version)s"
```
stdweird: the pkgtemplate should be read from an option

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,67 @@
...
+
+    pkgname=pkgtemplate % {
+        'name' : easyblock.name,
+        'version' : easyblock.version,
+    }
+    dependencies = []
+    dependencies.extend([ "=".join([ dep['name'], dep['version'] ]) for dep in easyblock.cfg.dependencies() ])
+    depstring = '--depends ' + ' --depends '.join(dependencies)
```
stdweird: don't make a string. use lists like
```python
for dep in easyblock.cfg.dependencies():
    cmdlist.extend(['--depends',"=".join([ dep['name'], dep['version'] ])])
```

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,67 @@
...
+    cmdlist=[
+        'fpm',
+        '--workdir', workdir,
+        '--name', pkgname,
+        '-t', 'rpm', # target
+        '-s', 'dir', # source
+        '-C', easyblock.installdir,
```
stdweird: add the software license

easybuild/framework/easyblock.py
```diff
@@ -1444,8 +1444,11 @@ def extensions_step(self, fetch=False):
         self.clean_up_fake_module(fake_mod_data)
 
     def package_step(self):
-        """Package software (e.g. into an RPM)."""
-        pass
+        """Prepare package software (e.g. into an RPM) with fpm."""
+
```
stdweird: step 1 should be: error if the license does not allow you to redistribute.

fgeorgatos: please don't do that!

Explanation:
rpms are a delivery mechanism, not a publication mechanism (think scp or, tar for that matter!)

if you insist too much on a license check, at least make it a list for conditional exit.

rjeschmi:
A working command line looked something like this for those interested:

`eb foss-2014b.eb --prefix=/software/easybuild --package-tool=fpm --packagepath=/export/easybuild/packages --sourcepath=/export/easybuild/sources -dfr`

boegel:
@rjeschmi: any chance you can share the RPMs you've obtained? Just to take a quick look at their structure.

rjeschmi:
@boegel sure, no prob. I'll link to them later (they are most definitely broken for deps, but give some ideas)

easybuild/framework/easyblock.py
rjeschmi: I share concerns of @stdweird (and think about it occasionally when building for docker, etc). Do we know what the license is in EB? I don't mind including some advice to the user when packaging things that are restricted.

rjeschmi:
And thanks to @jordansissel for advice in the previous PR for integrating FPM into EasyBuild (I've taken this up now and have tried to integrate you suggestions).

I may have some questions to improve things later, but it seems to be working well enough for our purposes.

rjeschmi:
It feels a bit like we should support SCL

https://access.redhat.com/documentation/en-US/Red_Hat_Software_Collections/1/html/Packaging_Guide/index.html

We already have a package prefix, and that should be maintained for the depends and provides.

easybuild/framework/easyblock.py
boegel: We don't know typically, but everything is in place to start tracking licenses (see the `software_license` easyconfig parameter). We can assume the license is the most restrictive possible when `software_license` is set to `None`.

boegel:
@rjeschmi: with `fpm`, we need to think broader than only RPMs...

rjeschmi:
some sample packages available here:

http://ottbioinfo.ca/eb-packages/

will add hpl

rjeschmi:
some dep problems still unfortunately (looks like toolchain is missing)

easybuild/framework/package.py
rjeschmi: @boegel The toolchain doesn't seem to be in the list of deps in the case I have. Is there anything in particular I need to do to get toolchain included?

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,91 @@
...
+    pkgtemplate = "%(prefix)s-%(name)s"
+    #"HPCBIOS.20150211-%(name)s-%(version)s"
+
+    pkgname=pkgtemplate % {
+        'prefix' : pkgprefix,
+        'name' : easyblock.name,
+    }
+    _log.debug("The dependencies to be added to the package are: " + pprint.pformat(easyblock.cfg.dependencies()))
+    depstring = ""    
+    for dep in easyblock.cfg.dependencies():
```
rjeschmi: @boegel I don't seem to be getting the toolchain returned here...

rjeschmi:
There are some issues with package collisions. I will need to add the toolchain to the package name.

boegel:
Including the toolchain in the package name only makes sense to me...

easybuild/tools/packaging.py
boegel: That's very weird... Which easyconfig/toolchain is this?

rjeschmi: It would be everything below foss-2015a. Like say FFTW/3.3.4-gompi-2015a

boegel: OK, I was confusing an `EasyConfig` instance with the dictionary that is constructed by `process_easyconfig` (which it does via an EasyConfig instance); only the latter would include the toolchain in the value that matches the `dependencies` key (mainly so that the robot also considers the toolchain as a dep to resolve)

You can figure out the toolchain here using `easyblock.toolchain`.

Something like this should work:

```python
for dep in [easyblock.toolchain.as_dict()] + easyblock.cfg.dependencies():
```

rjeschmi:
Just rebased the commits and squashed the original commits, hopefully I didn't break anything :)

rjeschmi:
I have updated the example packages here:

http://ottbioinfo.ca/eb-packages/

Note that the prefix needs to be unique to allow parallel install

boegel:
@rjeschmi: ignore the failed tests, I cancelled the Jenkins job for your last push

Let me know if you need help retriggering the tests (but they should get triggered on your next push in any case).

easybuild/tools/config.py
```diff
@@ -65,7 +66,7 @@
 }
 DEFAULT_PREFIX = os.path.join(os.path.expanduser('~'), ".local", "easybuild")
 DEFAULT_REPOSITORY = 'FileRepository'
-
+DEFAULT_PACKAGE_TEMPLATE = "eb-%(name)s-%(version)s-%(toolchain)s"
```
fgeorgatos: this is hardwired, can it become a tunable?

if someone needs to embed, say, -centos5- in the name, this is not too easy. similarly, if there is a suffix, it helps.

easybuild/tools/options.py
```diff
@@ -243,6 +243,14 @@ def config_options(self):
                                None, 'store_or_None', None, {'metavar': "PATH"}),
             'modules-tool': ("Modules tool to use",
                              'choice', 'store', DEFAULT_MODULES_TOOL, sorted(avail_modules_tools().keys())),
+            'package-tool': ("Packaging tool to use",
+                             None, 'store_or_None', None),
+            'package-type': ("Packaging type to output to",
+                             None, 'store_or_None', None),
+            'package-template': ("A template string to name the package",
+                             None, 'store', DEFAULT_PACKAGE_TEMPLATE),
```
rjeschmi: The package template can be overridden here @fgeorgatos

easybuild/framework/easyblock.py
fgeorgatos: Perhaps, there is one sane direction to propose about this; let's deliver built rpms on a per dict basis:

`output_subdirectory = ['GPL': 'public/', 'MIT': 'public/', 'JavaCP': 'private/', '': 'private']`

People then may choose, if they fence it via permissions or some other mechanism.

If the idea passes, this could well be part of easybuild configuration, for extensibility (alike moduleclasses).

easybuild/tools/config.py
boegel: this is just the default, `--package-template` can be used to define it

fgeorgatos: oops, this is `config.py` indeed. Where is the cancel button for the above?! -P
Anyway, the need for potential suffix still applies, but is not as urgent, it can be done anytime later.

easybuild/tools/packaging.py
```diff
@@ -0,0 +1,116 @@
...
+from vsc.utils import fancylogger
+
+from easybuild.tools.run import run_cmd
+from easybuild.tools.config import install_path, package_template
+from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
+from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME
+
+_log = fancylogger.getLogger('tools.packaging')
+
+def package_fpm(easyblock, modfile_path, package_type="rpm" ):
```
fgeorgatos: what if this was also tunable via `package_type`? L55/L113 could be similarly sync'ed to this.

rjeschmi:
Just to document a potential "bug" is that fpm itself will not be loaded by easybuild (and it isn't detected early in the build so it can take a long time to die).

At the very least the build should die right away. Better, it would be nice to load Ruby with module.

boegel:
Notes after conf call with @rtmclay for things to look into:

* [ ] synergy with `--skip`: skip (re)installation, only (re)generate package
* [ ] template for RPM `version` 
* [ ] template keys for EasyBuild version, toolchain name/version (separately), versionsuffix (and -prefix)
* [ ] support a way to bump/set the RPM `release` number
* [ ] dependency specifications should also take version of dependency into account

boegel:
Some info on one of the RPMs built by TACC for their systems (provided by @rtmclay), may be useful to figure out good defaults for the pkg name/version/release:

```
Sstaff(3005)% rpm -qil phdf5-intel13-mvapich2_1_9-1.8.9-3.x86_64
Name        : phdf5-intel13-mvapich2_1_9   Relocations: (not relocatable)
Version     : 1.8.9                             Vendor: NCSA
Release     : 3                             Build Date: Fri 07 Dec 2012 04:14:58 PM CST
Install Date: Mon 10 Dec 2012 04:39:22 PM CST      Build Host: build.stampede.tacc.utexas.edu
Group       : Development/Libraries         Source RPM: hdf5-1.8.9-3.src.rpm
Size        : 24500151                         License: see included Copyright
Signature   : (none)
Packager    : TACC - mclay@tacc.utexas.edu
URL         : http://hdf.ncsa.uiuc.edu/HDF5/
Summary     : HDF5 library
Description :
HDF5 is:
* A versatile data model that can represent very complex data objects and a wide variety of metadata.
* A completely portable file format with no limit on the number or size of data objects in the collection.
* A software library that runs on a range of computational platforms, from laptops to massively parallel systems, and implements a high-level API with C and Fortran 90 interfaces.
* A rich set of integrated performance features that allow for access time and storage space optimizations.
* Tools and applications for managing, manipulating, viewing, and analyzing the data in the collection.
/opt/apps/intel13/mvapich2_1_9/modulefiles/phdf5
/opt/apps/intel13/mvapich2_1_9/modulefiles/phdf5/.version.1.8.9
/opt/apps/intel13/mvapich2_1_9/modulefiles/phdf5/1.8.9.lua
```

easybuild/tools/config.py
```diff
@@ -188,13 +191,13 @@ class ConfigurationVariables(FrozenDictKnownKeys):
         'buildpath',
         'config',
         'installpath',
-        'installpath_modules',
-        'installpath_software',
```
rjeschmi: I wonder how this got missed in the merge...

boegel: that's my bad when resolving conflicts, please re-add these

rjeschmi:
Add the experimental check and fpm check to 

https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/tools/options.py#L405

(perhaps add it in a way that calls out to the packaging.py so that merges later on will be easy/possible.

rjeschmi:
- [x] require use of --experimental
- [x] include (very) early check for fpm
- [x] add check in place to make sure an existing package file is not overwritten without use of --force

boegel:
@rjeschmi: the error message I'm getting when generating a package while one already exists is a bit cryptic (and the error occurs quite late too), maybe we can do better here?

```
$ rm /homeappl/home/hoste/appl_taito/modules/all/bzip2/1.0.6
rm: remove regular file `/homeappl/home/hoste/appl_taito/modules/all/bzip2/1.0.6'? y
$ eb --experimental --package-tool=fpm bzip2-1.0.6.eb -d
== temporary log file in case of crash /tmp/hoste/eb-SX0z04/easybuild-5zZulT.log
== processing EasyBuild easyconfig /homeappl/home/hoste/EasyBuild/easybuild-easyconfigs/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb
== building and installing bzip2/1.0.6...
== fetching files...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== testing...
== installing...
== postprocessing...
== sanity checking...
== cleaning up...
== creating module...
== packaging...
== FAILED: Installation ended unsuccessfully (build directory: /tmp/hoste/easybuild_build/bzip2/1.0.6/dummy-dummy): build failed (first 300 chars): Unable to copy file, dest already exists. Look in src for packages dest: /wrk/hoste/EasyBuild/packages/eb-bzip2-1.0.6-dummy-dummy-eb-1.x86_64.rpm src: /tmp/hoste/eb-SX0z04/tmp3m2Dfv/eb-bzip2-1.0.6-dummy-dummy-eb-1.x86_64.rpm 
== Results of the build can be found in the log file /tmp/hoste/eb-SX0z04/easybuild-bzip2-1.0.6-20150617.215148.UoQss.log
ERROR: Build of /homeappl/home/hoste/EasyBuild/easybuild-easyconfigs/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb failed (err: 'build failed (first 300 chars): Unable to copy file, dest already exists. Look in src for packages dest: /wrk/hoste/EasyBuild/packages/eb-bzip2-1.0.6-dummy-dummy-eb-1.x86_64.rpm src: /tmp/hoste/eb-SX0z04/tmp3m2Dfv/eb-bzip2-1.0.6-dummy-dummy-eb-1.x86_64.rpm ')
```

rjeschmi:
the package includes the build reports, this makes it get bigger and bigger each time:

```
rob@ogic:~/eb/eb-devel-install/export/packages$ rpm -qlp eb-foss-2015a-dummy-dummy-2015a-3.x86_64.rpm
/export/easybuild/modules/all/foss/2015a
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.021511.log
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.021511_test_report.md
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.022857.log
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.022857_test_report.md
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.022918.log
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.022918_test_report.md
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.023120.log
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.023120_test_report.md
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.023229.log
/export/easybuild/software/foss/2015a/easybuild/easybuild-foss-2015a-20150623.023229_test_report.md
/export/easybuild/software/foss/2015a/easybuild/foss-2015a-easybuild-devel
/export/easybuild/software/foss/2015a/easybuild/foss-2015a.eb
```

rjeschmi:
- [x] require use of --experimental
- [x] include (very) early check for fpm
- [x] add check in place to make sure an existing package file is not overwritten without use of --force
- [x] support a way to bump/set the RPM release number
- [x] dependency specifications should also take version of dependency into account

rjeschmi:
I think the only thing left is:

- [ ] dependency specifications should also take version of dependency into account

I'm not sure this is important, but I'll see if I can make it work. right now the requirement is the "name" which includes the version by default.

rjeschmi:
I will implement at least one more example package naming scheme as a test that the feature set for package-naming-scheme is sufficient.

I'll use the TACC sample as an example since it has some interesting features (no version in name).

rjeschmi:
added eb default version so name looks like:

eb2.2.0dev-foss-2015a-dummy-dummy-2015a-3.x86_64.rpm

getting pretty long :)

boegel:
@rjeschmi: let's include https://github.com/rjeschmi/easybuild-framework/pull/12, and then merge this beast for inclusion in EasyBuild v2.2.0...

boegel:
OK, going in after careful review and some helping out with unit tests & code style mostly.

Thanks *a lot* for your hard work on this @rjeschmi, and @sntgluca for kicking this off.

Note: for now, `--experimental` is required to use this, since we anticipate that changes may need to be made after we get feedback from people who have played around with this. It should work as expected though.

I will look into getting the documentation that covers this updated as well, see https://github.com/hpcugent/easybuild/pull/122.

PR #1226, opened 2015-03-17
Support system modfiles

pforai:
Initial support for loading non EB-generated module files in configs. 
Example:
```
name = 'HDF5'
version = '1.8.13'

homepage = 'http://www.hdfgroup.org/HDF5/'
description = """HDF5 is a unique technology suite that makes possible the management of
 extremely large and complex data collections."""

toolchain = {'name': 'cpeg', 'version': '5.2.25'}

source_urls = ['http://www.hdfgroup.org/ftp/HDF5/releases/hdf5-%(version)s/src']

patches = [
    'HDF5-%(version)s_configure_intel.patch',
    'configure_libtool.patch',
]

buildopts = 'CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK"'

dependencies = [
    ('zlib', '1.2.8'),
    ('Szip', '2.1'),
]

system_modules = [('craypkg-gen')]
postinstallcmds = ['craypkg-gen -p %(installdir)s/../']

moduleclass = 'data'
```


easybuild/framework/easyconfig/default.py
```diff
@@ -137,7 +137,7 @@
     'dependencies': [[], "List of dependencies", DEPENDENCIES],
     'hiddendependencies': [[], "List of dependencies available as hidden modules", DEPENDENCIES],
     'osdependencies': [[], "OS dependencies that should be present on the system", DEPENDENCIES],
-
+    'system_modules': [[], "System module dependencies that should be present on the system", DEPENDENCIES],
```
boegel: there should be a matching configuration option `--system-modules` too for this, to list modules that should *always* be loaded

easybuild/framework/easyblock.py
```diff
@@ -1306,6 +1306,7 @@ def prepare_step(self):
         """
         self.cfg['unwanted_env_vars'] = env.unset_env_vars(self.cfg['unwanted_env_vars'])
         self.toolchain.prepare(self.cfg['onlytcmod'])
+        self.modules_tool.load(self.cfg['system_modules'])
```
boegel: I think the `system_modules` should be loaded *first*, before anything else...

Would that break anything as you're using them now?

easybuild/framework/easyconfig/default.py
pforai: Agreed, can you add this boilerplate here? I'm sure it takes you much less time than if I go digging :)

pforai:
Dont get what happend here. Extra files from old PR with same branch name were included. Opening a new one.

boegel:
replaced by #1227

PR #1227, opened 2015-03-17
Support system module files

pforai:
Initial support for system module files that were not generated via EB itself.

Example:
```python
name = 'HDF5'
version = '1.8.13'

homepage = 'http://www.hdfgroup.org/HDF5/'
description = """HDF5 is a unique technology suite that makes possible the management of
 extremely large and complex data collections."""

toolchain = {'name': 'cpeg', 'version': '5.2.25'}

source_urls = ['http://www.hdfgroup.org/ftp/HDF5/releases/hdf5-%(version)s/src']

patches = [
    'HDF5-%(version)s_configure_intel.patch',
    'configure_libtool.patch',
]

buildopts = 'CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK"'

dependencies = [
    ('zlib', '1.2.8'),
    ('Szip', '2.1'),
]

system_modules = [('craypkg-gen')]
postinstallcmds = ['craypkg-gen -p %(installdir)s/../']

moduleclass = 'data'
```


easybuild/framework/easyblock.py
```diff
@@ -1305,7 +1305,8 @@ def prepare_step(self):
         Pre-configure step. Set's up the builddir just before starting configure
         """
         self.cfg['unwanted_env_vars'] = env.unset_env_vars(self.cfg['unwanted_env_vars'])
-        self.toolchain.prepare(self.cfg['onlytcmod'])
+        self.modules_tool.load(self.cfg['system_modules']) 
+	self.toolchain.prepare(self.cfg['onlytcmod'])
```
boegel: no tabs pls

stdweird:
@pforai can you explain what's up with the 
```
system_modules = [('craypkg-gen')]
postinstallcmds = ['craypkg-gen -p %(installdir)s/../']
```

if all easyconfigs require this, we need a better solution. it also makes that all easyconfigs need to be adapted to support cray. (and this should be refelcetd in the version.)

easybuild/framework/easyconfig/default.py
```diff
@@ -137,7 +137,7 @@
     'dependencies': [[], "List of dependencies", DEPENDENCIES],
     'hiddendependencies': [[], "List of dependencies available as hidden modules", DEPENDENCIES],
     'osdependencies': [[], "OS dependencies that should be present on the system", DEPENDENCIES],
-
+    'system_modules': [[], "System module dependencies that should be present on the system", DEPENDENCIES],
```
stdweird: can we call this `external_modules` or `external_dependencies`? `system` is a vague / misleading word. (i can understand that the compiler provided by the OS gets the label `system`, but this is something different, esp if these are actual modules).

boegel: it should certainly be `foo_modules`, since they're only modules, not EB dependency specifications

`external_modules` works for me

boegel:
@stdweird: the example that @pforai gives is better handles via a configuration option `--system-modules` (I'm working on a PR to @pforai's branch for that), together with an extra option specifying that the easyblock should run `craypkg-gen` as a part of the post install commands

`system_modules` in easyconfigs will only be used in easyconfigs for Cray toolchains, I think, but @pforai will correct me if I'm mistaken

easybuild/framework/easyconfig/default.py
stdweird: but they are dependency specifications

boegel: no, these are just module names: `GCC/4.7.2` (bad example), a dependency specification is `('GCC', '4.7.2')`

stdweird: i meant dependecy as in "without it, it won't work". dependencies in the form of external modules, like osdependencies but with modules.

boegel:
@pforai: PR for `--system-modules` at https://github.com/pforai/easybuild-framework/pull/12, it's probably a good idea to rename it to `external_modules` as @stdweird suggests

easybuild/framework/easyconfig/default.py
boegel: ok, I see...

well, I don't feel too strongly about it, so `external_dependencies` is fine by me too; up until now we haven't been using any `_` for `*dependencies` though, so that would make it `externaldependencies`...

or maybe `external_dependency_modules`?

stdweird: i'm ok `externaldependencies`, the help can say something like `dependencies that are modules provided by external sources`

pforai:
The example I provided is for a piece of software that doesnt generate pkg-config .pc files as part of the build process, but I need the .pc files for other purposes later on. On the Cray they provide a tool that generates .pc files given a software installation directory, but this is not available per default in the environment. This example shows how to use a binary that would otherwise not be available for the build. 
This is only needed on a hand full of builds where no .pc files are generated, so nothing universal. I do like the fact that it is explicitly visible that for this build pkg-config files are going to be generated.

The 'external_modules' have a lot more use cases than just the simple one I provided with that example.

Also on the cray the code generator's backend can be specified via a module that can be loaded, so I have two use cases for external_modules right now:

* code generator backend (ivy-bridge vs haswell for example)
* Cray utilities that are needed for certain builds

stdweird:
well, given that this is a general request, are the `system_modules`/`external_dependencies` build dependencies, runtime or both?

pforai:
They can actually be both!

pforai:
@stdweird can you tell me what issues and problems you see from having this? 

I will refactor the naming to be external_dependencies later on.

One more example of usefulness is the definition for the Cray toolchains itself, as they wrap existing modules.

```python
easyblock = 'Toolchain'

name = '%(name)s'
version = '%(version)s'

homepage = 'http://hpcugent.github.io/easybuild/'
description = """This is a shim module for having EB pick up the Cray
Programming Environment (PrgEnv-*) modules. This module implements the EB
toolchain module for each of the cray modules."""

source_urls = []
sources = []
dependencies = []

system_modules=[('PrgEnv-gnu',5.2.25'),('craype-haswell')]]

moduleclass = 'toolchain'
```

stdweird:
also, if 
```
system_modules = [('craypkg-gen')]
postinstallcmds = ['craypkg-gen -p %(installdir)s/../']
```
is used alot, it might make sense to introduce a cray toolchain

stdweird:
@pforai sorry, wrong button

pforai:
Dunno how to do it nicely - any suggestions? Have the block run this as an additional step if some var is exposed from the config and it's a cray machine?
It should not be too frequently as a lot of libs are generating .pc files - most of software that is running autotools will do it nowadays.

boegel:
@pforai: it would be nice to have a more-or-less complete overview of use cases where 'external modules' are involved; @stdweird agrees that support for loading external modules would be useful, even on non-Cray systems, but for Cray support specifically, we probably can take a better approach

Here are the ones I know about:

* loading of `PrgEnv-cray` modules; this could be tackled by the toolchain support itself in the framework, no need to list this module anywhere?
* loading of `craype-foo` module to specify the code generator backend; we could use the existing `--optarch` for this, and have the Cray toolchain support in framework handle this correctly (i.e. by loading the module(s) specified that way, other than tweaking the part of `$CFLAGS` that matches having `optarch=True` set in the toolchain options
  * can this be a *list* of modules, rather than a single one? and can this depend on the software being built, or is it (always) dictated by the system? this is important w.r.t. robot builds, where multiple software pkgs are being installed "in one go"
* occasionally loading a particular module, depending on the software being built
  * `craypkg-gen` is one example, but there we may need to make this a custom easyconfig parameter, so you can just add `craypkg_gen = True` in the easyconfig file, since `craypkg-gen` now needs to be speciifed twice, once as a module, once in a command (that will always be the same?)
  * any other examples you know about?

Any other use cases I'm forgetting about?

It seems like with only these in mind, we don't really need `--external-modules` or an easyconfig parameter `external-modules` (or whatever it's named), since there seem to be better alternatives for each of them

pforai:
@boegel @stdweird 
I agree with the remarks here. Feel free to reuse at any later point in time if the feature makes sense for other use cases.

boegel:
I like the general idea, so I've opened #1230 to see if we can get this generic feature in.

The stuff required for Cray specifically will be handled in a separate PR to contribute Cray toolchains, I'll work together with @pforai on that.

Issue #1181, opened 2015-02-16
Please rename the `dummy` toolchain


riccardomurri:
I understand that `dummy` is a sensible name from the programmers' point of view: if one is conversant with the way EasyBuild works, it just translates to "do not build any tool, just use what's available".

However, for newcomers and "simple" users `dummy` is kind of misleading, as it suggests no-op or a "fake" toolchain.

May I suggest that it be renamed to "system" (use system tools/compilers) or "os"?

boegel:
I feel that's a valid request, but we'll need to phase out the `dummy` name, rather than kill it hard.

We can deprecate `dummy`, and drop support for it in EasyBuild v3.0 all together.

@stdweird: thoughts?

malcook:
FWIW: +1 - good idea - was a bit baffling to me too - of course it is
documented somewhere - I myself briefly had the thought that it might be
called 'base' or 'host' when I came to understand its purpose.  The word
"vendor" also came to mind, but only briefly.  A minor change I know but
one that might lend a little superficial maturity to the project.

~ Malcolm Cook
h: 913.642.5257
m: 913.226.1542
w: 816.926.4449

stdweird:
to be clear, 'dummy' never meant 'use system tools'. it's a different discussion, but it might as well get cleaned up together.

2 usecases exists: 
'dummy' / 'noop' : simply unpack the sources, no compilation required
'dummy' / 'system': use the system compiler, not a full EB controlled toolchain  (can be system/OS, or vendor provided,...)

noop should stay the same obscure code path as dummy today :smile: 
for system, there should be a valid system toolchain.

pforai:
+1 for renaming the dummy TC to something more sensible. Or clarify what exactly the dummy is, how it works and what the use case is.  There is some general knowledge among EBers how and when to use dummy. Mostly it boils down to use dummy for "TC independent"  (the EB way is to roll everything from EB and have reproducible builds and all nice things we have when rolling the EB way) builds.

According to https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/toolchains/compiler/dummycompiler.py#L40 
it should _not_ even try to use a system GCC. But it certainly does that.

boegel:
@pforai: no, it doesn't, EasyBuild even sets `$CC` to `dummy` in the environment

to actually build stuff when using a `dummy` toolchain, you need to use `prebuildopts = "CC=gcc"` to get around that

pforai:
[...]

PR #1228, opened 2015-03-17
Make dummy dummy resolve deps

pforai:
I find this to be more intuitive and it works they way you expect EB to work. 
@fgeorgatos, @pescobar, @wpoely86 opinions?


wpoely86:
@boegel what is the historical reason that `dummy-dummy` doesn't load deps?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -333,10 +333,13 @@ def prepare(self, onlymod=None):
         if not self._toolchain_exists():
             self.log.error("No module found for toolchain: %s" % self.mod_short_name)
 
-        if self.name == DUMMY_TOOLCHAIN_NAME:
+        
+	if self.name == DUMMY_TOOLCHAIN_NAME:
```
stdweird: spaces, no tabs for indentation

easybuild/tools/toolchain/toolchain.py
```diff
@@ -333,10 +333,13 @@ def prepare(self, onlymod=None):
...
             if self.version == DUMMY_TOOLCHAIN_VERSION:
                 self.log.info('prepare: toolchain dummy mode, dummy version; not loading dependencies')
-            else:
+		#@todo I keep this is as, but want dummy to do the same as a regular toolchain. Need to think this through.
```
stdweird: at the very least, also cleanup the log message.
wrt the todo, any progress? i think this requires an extra option to have this behaviour. (as it's very backwards incompatible).

stdweird:
@wpoely86 i think that's the only difference between `dummy` and `dummy-dummy`. there must be some docs about this somewhere

wpoely86:
@stdweird as fars as I know, it's the only difference (it's in the readthedocs). But why did you make this difference in the first place? There must have been some reason?

stdweird:
@wpoely86 don't remember immediately, but one thing @boegel is going to add are runtime deps, which is basicallly what the original `dummy-dummy` hack is all about.

stdweird:
@pforai do you need `dummy-dummy` or were you not aware of this difference?

boegel:
The only reason why `dummy` with `dummy` version doesn't load deps is because it has been a crippled toolchain from the very start. The idea behind `dummy` was that it should only be used to build compilers with, to kick of a toolchain, and then build everything else with a proper toolchain (which is potentially compiler-only).

That's not how people have been using it however. There has been a call for proper support for building with the system-provided tools, so we should have a proper `system` toolchain instead (cfr. #1181).

So, we should:

* deprecate the usage of `dummy` (both with and without `dummy` version)
* support a `system` toolchain to replace it (where we don't have this 'don't load deps' trickery)
* switch from using `dummy` to `system` in all easyconfigs where needed, and then simply remove `dummy` in EB v3.0

pforai:
@boegel agreed. I guess providing a SystemGCC toolchain is quite easy to do. But I would make dummy dummy behave as all the others until it is removed in 3.0.

boegel:
@pforai: not a big fan of modifying behaviour of something from one minor version to another, especially if we're going to deprecate it...

Shouldn't be too much work to deprecate `dummy`.

PR #1229, opened 2015-03-18
unpack_options should default to '' instead of None

JensTimmerman:
fixes
```
== 2015-03-18 15:26:24,010 main.EasyConfig ERROR EasyBuild crashed with an error (at easybuild/framework/easyconfig/easyconfig.py:218 in update): Can't update configuration value for unpack_options, because it's not a string or list.
== 2015-03-18 15:26:24,011 main.easyblock WARNING build failed (first 300 chars): EasyBuild crashed with an error (at easybuild/framework/easyconfig/easyconfig.py:218 in update): Can't update configuration value for unpack_options, because it's not a string or list.
```


boegel:
looks fine, will merge if Jenkins is OK with it

boegel:
Going in, thanks @jgtimmer.

boegel:
Hmm, hold on... `self.cfg.update` isn't tested at all right now, which is bad

I'll add a small unit test for this, using `unpack_options`, to get it included here (I'll issue a PR your way in a couple of minutes).

boegel:
@JensTimmerman: https://github.com/JensTimmerman/easybuild-framework/pull/10

easybuild/framework/easyconfig/default.py
```diff
@@ -116,7 +116,7 @@
     'stop': [None, 'Keyword to halt the build process after a certain step.', BUILD],
     'tests': [[], ("List of test-scripts to run after install. A test script should return a "
                    "non-zero exit status to fail"), BUILD],
-    'unpack_options': [None, "Extra options for unpacking source", BUILD],
+    'unpack_options': ['', "Extra options for unpacking source", BUILD],
```
stdweird: `options` should be list, no? (and what wrong with default `None`?)

boegel: a list would work too

`None` is not a list or a string, and so `self.cfg.update` doesn't work

We should also avoid using `[]` as a default value, since that's mutable... Those values should all be `None`, and then be later transformed into `[]`.

stdweird: what's the error you get with `None`? 

but i don't get your remark wrt the empty list. (but changing the default type is probably not the best idea without some more deprecation code).

boegel: if we use `[]` here as a default value, we open the door for very weird/hard to debug bugs where somewhere deep in the code we're updating this list; same argument as why using `[]` as a default value for a function argument is a bad idea, it can cause global side-effects

as to why not `None`:

```
Can't update configuration value for unpack_options, because it's not a string or list.")
```
(the `update` method in `EasyConfig` only supports updating list or string-typed values, because it appends (and doesn't want to guess the type a value should be if the previous value was `None`)

stdweird: wrt `[]`, you mean as opposed to hmm, i don't know `unwanted_env_vars` below?

wrt `None`, ok, makes sense

boegel: I wasn't saying this isn't used anywhere right now. I'll open an issue to clean this up.

boegel: see #1231

boegel:
going in, cleanup of `[]` can be done later, thanks for tackling this @JensTimmerman (and thanks for the great tests! ;-))

PR #1230, opened 2015-03-18
add support for loading 'external' modules (REVIEW)

boegel:
follow-up of #1227, taking over from @pforai

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -114,7 +114,7 @@ class EasyConfig(object):
         self.log = fancylogger.getLogger(self.__class__.__name__, fname=False)
 
         if path is not None and not os.path.isfile(path):
-            self.log.error("EasyConfig __init__ expected a valid path")
+            raise EasyBuildError("EasyConfig __init__ expected a valid path")
 
         # read easyconfig file contents (or use provided rawtxt), so it can be passed down to avoid multiple re-reads
         self.path = None
@@ -215,7 +215,7 @@ class EasyConfig(object):
         elif isinstance(prev_value, list):
             self[key] = prev_value + value
         else:
-            self.log.error("Can't update configuration value for %s, because it's not a string or list." % key)
+            raise EasyBuildError("Can't update configuration value for %s, because it's not a string or list.", key)
 
     def parse(self):
         """
@@ -228,7 +228,8 @@ class EasyConfig(object):
             # build a new dictionary with only the expected keys, to pass as named arguments to get_config_dict()
             arg_specs = self.build_specs
         else:
-            self.log.error("Specifications should be specified using a dictionary, got %s" % type(self.build_specs))
+            raise EasyBuildError("Specifications should be specified using a dictionary, got %s",
+                                 type(self.build_specs))
         self.log.debug("Obtained specs dict %s" % arg_specs)
 
         self.log.info("Parsing easyconfig file %s with rawcontent: %s" % (self.path, self.rawtxt))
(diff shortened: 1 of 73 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -562,12 +568,25 @@ def _parse_dependency(self, dep, hidden=False):
...
+            if dep and dep[-1] == EXTERNAL_MODULE_MARKER:
+                if len(dep) == 2:
+                    dependency['external_module'] = True
+                    dependency['short_mod_name'] = dep[0]
+                    dependency['full_mod_name'] = dep[0]
+                else:
+                    raise EasyBuildError("Incorrect external dependency specification: %s", dep)
+            else:
+                # non-external dependency: tuple (or list) that specifies name/version(/versionsuffix(/toolchain))
+                dependency.update(dict(zip(attr, dep)))
```
wpoely86: the convert to list is not needed anymore?

wpoely86: why not use `dependency.update(dict(zip(attr, dep)))` for both cases and set `external_module` to `True` for the external deps?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -562,12 +568,25 @@ def _parse_dependency(self, dep, hidden=False):
...
+
         else:
-            raise EasyBuildError('Dependency %s of unsupported type: %s.', dep, type(dep))
+            raise EasyBuildError("Dependency %s of unsupported type: %s", dep, type(dep))
+
+        if dependency['external_module']:
+            self.log.debug("Returning parsed external dependency: %s", dependency)
+            return dependency
```
wpoely86: why is there a return here?

easybuild/framework/easyconfig/easyconfig.py
boegel: it was never needed, a `zip` works fine on list/tuple mix

the `zip` doesn't make sense in the case of an external module, since we can't define `name` or `version` then, we only know the module name (the name/version can't be derived easily, or at least we shouldn't assume we can)

easybuild/framework/easyconfig/easyconfig.py
boegel: because the full below doesn't make sense for an external dependency (for which name/version is not defined)

easybuild/framework/easyconfig/tools.py
```diff
@@ -111,13 +111,23 @@ def find_resolved_modules(unprocessed, avail_modules, retain_all_deps=False):
         new_ec = ec.copy()
         deps = []
         for dep in new_ec['dependencies']:
-            full_mod_name = ActiveMNS().det_full_module_name(dep)
+            full_mod_name = dep.get('full_mod_name', None)
```
stdweird: and why would all external module have the same naming scheme?

easybuild/framework/easyconfig/tools.py
```diff
@@ -111,13 +111,23 @@ def find_resolved_modules(unprocessed, avail_modules, retain_all_deps=False):
...
+                full_mod_name = ActiveMNS().det_full_module_name(dep)
+
             dep_resolved = full_mod_name in new_avail_modules
             if not retain_all_deps:
                 # hidden modules need special care, since they may not be included in list of available modules
                 dep_resolved |= dep['hidden'] and modtool.exist([full_mod_name])[0]
+
             if not dep_resolved:
-                deps.append(dep)
+                # treat external modules as resolved when retain_all_deps is enabled (e.g., under --dry-run)
```
stdweird: please explain why external modules don't need to be resolved.

easybuild/framework/easyconfig/tools.py
boegel: the naming scheme of external modules shouldn't matter, EB just needs to know about the module name so it can i) load the module, ii) include a `module load` statement in the generated module file

external modules are specified by *module name* only, no separate name/version (or suffix/toolchain) details, just a module name

easybuild/framework/easyconfig/tools.py
boegel: external modules have no matching easyconfig file, they're just there

I'll clarify the comment

boegel:
Going in, thanks for the review @wpoely86 and @stdweird, and @pforai for initiating this!

Issue #741, opened 2013-10-31
also try and resolve dependencies using subtoolchains


boegel:
Feature proposal by @berndmohr:

For resolving dependencies specified in a `goolf` toolchain build, we could also try and resolve them using a `gompi` toolchain and maybe even a `GCC` toolchain.

This would allow that tools like `CMake` are built with a minimal toolchain (e.g. `CMake-2.8.4-GCC-4.7.2`), and still can be used in a `goolf` easyconfig as build dependency.

This is less strict than what is being done now (i.e. only use the specified toolchain, either the one in `toolchain` or in the custom toolchain in the dependency specification).

I'm confident @besserox would support this feature proposal too, let's see what @stdweird thinks about it. :-)

ocaisa:
I struggle to see how you can get this to work without having some kind of equivalency classes defined between toolchains (somewhat like what Fotis was describing but not so fine-grained) and well defined inheritance in a toolchain (gmpolf inherits gompi inherits GCC) so that searching back down the toolchain hierarchy is straightforward. 

Not to worry, this is what I plan to work on next week (along with a new HMNS that would leverage hierarchically defined toolchains)!

boegel:
@ocaisa: that's indeed indispensable, but it could be as simple as adding something like this in the toolchain definitions that are now already in the framework:

```python
# in goolf
subtoolchains = ['gompi', 'GCC']
```

`dummy` would then be an implied subtoolchains for all others

ocaisa:
That might be enough...but then which version of gompi matches the current toolchain? If it's not listed somehow in the easyconfig, wouldn't it be a struggle to figure out uniquely?

boegel:
The versioning can be tricky, yes. But we can do that best effort...

For several toolchains (e.g. `goolf`), subtoolchains (e.g. `gompi`) that are *not* individual toolchain components (e.g. `GCC`) have the same version as their parent toolchain.

So, if a subtoolchain is included as an actual toolchain component, use the component version, if not, use the toolchain version itself.

For example, when using `goolf/1.4.10`, consider `gompi/1.4.10` (use toolchain version), and `GCC/4.7.2` (use `GCC` component version), and eventually also `dummy`.

That way, several toolchains we're using already will work with this procedure, and others can be made compatible with this in future versions of the toolchain.

boegel:
@ocaisa: fixed with #1306?

ocaisa:
Yes, I think so

PR #1232, opened 2015-03-18
Move module path extension in module file

ocaisa:
This is something that is required if you are using a Toolchain based name space. An example would be gpsmpi, where GCC is a dependency, this ensure that the path is extended by GCC before being extended by gpsmpi

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1642,8 +1642,8 @@ class EasyBlock(object):
 
         txt = ''
         txt += self.make_module_description()
-        txt += self.make_module_extend_modpath()
         txt += self.make_module_dep()
+        txt += self.make_module_extend_modpath()
         txt += self.make_module_req()
         txt += self.make_module_extra()
         txt += self.make_module_footer()


boegel:
Jenkins: ok to test

boegel:
partial duplicate of #1168

@ocaisa: this simply results in reordering `module use` vs `module load` for deps to `module load`'s first, and `module use` after, correct?

Why is that needed, exactly (I may have already asked in the context of #1168, but couldn't find an answer there either)? Can you give an specific example of what goes wrong if this is not included?

I seem to recall that the order mattered at some point w.r.t. `HierarchicalMNS`, but the tests results in #1168 suggest it doesn't (all green). This is probably from back when we were doing things the wrong way, while working on #953. I may be mistaken (I trust the tests more than my memory).

ocaisa:
The problem is that a toolchain based naming scheme would have other toolchains as dependencies:
```gpsolf -> gpsmpi -> GCC```
With the current way the module path is extended the path would be extended by gpsolf first, then gpsmpi, then GCC resulting in a ```module avail``` with a reverse view of the hierarchy (GCC at the top). Moving the module path extension gives a more logical hierarchy (in this case).

For the current default HMNS, it's the expansion of the toolchain modules that controls how the path is extended. Since this is guaranteed to go ````compiler -> mpi``` it shouldn't cause any problems

ocaisa:
To be honest, I'm creating this pull request partly from #1168 because we are already using the Toolchain based naming scheme in production and with this change maintaining that until we can support it properly in the framework would only require the missing naming scheme python module (and not modifying something within the framework).
For those who are interested, the issue with supporting it this naming scheme currently is that it requires the subtoolchain to be explicitly listed as a dependency. This is not very satisfactory since it can't be verified that the toolchain works at build time (if you leave out the dependency, the toolchain would build but you could not install anything with it because the module path would not be correctly extended). The solution is to add subtoolchain support to the framework #741, this could then be backwards compatible and would also give us  something to grab on to so we can kick an error if necessary.

boegel:
@ocaisa: ok, thanks for clarifying that, good to hear that it only fixes a 'cosmetic' issue with the order in `$MODULEPATH`, going in

PR #1233, opened 2015-03-18
define 'system' toolchain, deprecate 'dummy' toolchain (WIP)

boegel:
TODO:
- make sure all unit tests pass (at least one still fails, because of pulling in easyconfigs using `dummy` via `--from-pr`)
- add dedicated test to ensure that using `dummy` still works (and will keep working until v3.0)


easybuild/toolchains/compiler/systemcompiler.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class SystemCompiler(Compiler):
+    """Use system compilers."""
+    COMPILER_MODULE_NAME = []
+    COMPILER_FAMILY = TC_CONSTANT_SYSTEM
+
+    # deliberately not picking particular compilers
+    COMPILER_CC = '%sCC' % TC_CONSTANT_SYSTEM
+    COMPILER_CXX = '%sCXX' % TC_CONSTANT_SYSTEM
```
wpoely86: So, EB will not set `$CC` or `$CXX` right? The user can choose which system compiler to use?

pforai: Yes, and I'm not sure if we shouldn't add a SystemGCC TC to specific pick GCC and populate $CC and friends and set them to the same values as for the GCC TC.

boegel: With this, `EB` will set `$CC` to `SYSTEMCC`, which mimics with `dummy` was doing. Maybe we shouldn't set it at all.

Introducing a separate `SystemGCC` may be interesting, but wouldn't buy us much I think (people have no trouble in using `dummy` even though EB sets `$CC` to `DUMMYCC`, so in practice setting `$CC` makes little difference, apparently).

wpoely86: I would expect that the system toolchain doesn't define `$CC` or `$CXX` at all. Any code build with it, should try to find it's own compiler using whatever build system it has. The end user should be able to choose what to do by setting `$CC` (choosing between clang and gcc for example) and EB will not touch that.

I'm not a fan of having a `SystemGCC`.

pforai: We could have the system TC expose a module that a site can customize to fit the needs to set $CC et al. ?

damianam:
While you are at it, I'd argue too that the new `system` toolchain should load dependencies too (or have a mechanism to allow it). Example: Java packages that need to be compiled with `javac` and nothing else. Those depend on Java, but Java is not being loaded for dummy, which pushes the dependency up to `GCCcore`, which is not too bad, but not really necessary.

boegel:
@damianam To get dependencies loaded when using the `dummy` toolchain, use an empty version rather than `dummy` as version, see also http://easybuild.readthedocs.io/en/latest/Concepts_and_Terminology.html#dummy-toolchain

damianam:
That's very intuitive, I don't know how I missed that ;-P. Thanks!

boegel:
replaced by #2877

PR #1234, opened 2015-03-18
Add support for Cray PrgEnv compiler wrappers (WIP)

pforai:
Thanks @boegel for his work to clean this up and get it in the current shape. This is the basic stuff that can actually build at least things like HDF5. 

craypkg-gen support will be followed up shortly. This is needed to get the wrappers to pick up things that dont generate pkg-config .pc files that are read by the drivers to generate -I and -L flags. Anything that creates .pc files out of the box will be picked up by the driver anyway.


easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    COMPILER_CC = 'cc'
+    COMPILER_CXX = 'CC'
+
+    COMPILER_F77 = 'ftn'
+    COMPILER_F90 = 'ftn'
+
+    # FIXME (kehoste) hmmmm, really? then how do you control optimisation, precision when using the Cray wrappers?
+    COMPILER_FLAGS = []  # we dont have this for the wrappers
```
pforai: The defaults are supposed to be good enough. Except for the GNU case it is recommended to not specify those.

see: NERSC https://www.nersc.gov/users/computational-systems/hopper/programming/compiling-codes-on-hopper/

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+        """Load craype module specified via 'optarch' build option."""
+        optarch = build_option('optarch')
+        if optarch is None:
+            self.log.error("Don't know which 'craype' module to load, 'optarch' build option is unspecified.")
+        else:
+            self.modules_tool.load([self.CRAYPE_MODULE_NAME_TEMPLATE % {'optarch': optarch}])
+
+    # FIXME: (kehoste) is it really needed to customise this?
+    # this looks like a workaround for setting the COMPILER_*_FLAGS lists empty?
+    def _set_compiler_flags(self):
```
pforai: it kinda is, because the current Compiler base class is too specific for GCC like compilers.

pforai:
Ping @stdweird  - waiting for your comments and insight!

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+Support for the Cray Programming Environment Wrappers (aka cc, CC, ftn).
+The Cray compiler wrappers are actually way more than just a compiler drivers.
```
stdweird: `a compiler driver` or ` compiler drivers`

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+
+The basic concept is that the compiler driver knows how to invoke the true underlying
+compiler with the compiler's specific options tuned to Cray systems.
+
+That means that certain defaults are set that are specific to Cray's computers.
+
+The compiler wrappers are quite similar to EB toolchains as they include
+linker and compiler directives to use the Cray libraries for their MPI (and network drivers)
```
stdweird: comma missng at the end?

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+class CrayPEWrapper(Compiler):
+    """Generic support for using Cray compiler wrappers"""
+
+    # no toolchain components, so no modules to list here (empty toolchain definition w.r.t. components)
+    # the PrgEnv and craype are loaded, but are not considered actual toolchain components
+    COMPILER_MODULE_NAME = []
+    COMPILER_FAMILY = TC_CONSTANT_CRAYPEWRAPPER
+
+    COMPILER_UNIQUE_OPTS = {
+        # FIXME: (kehoste) how is this different from the existing 'shared' toolchain option? just map 'shared' to '-dynamic'? (already done)
```
stdweird: :+1:

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+        'dynamic': (True, "Generate dynamically linked executables and libraries."),
+        'mpich-mt': (False, "Directs the driver to link in an alternate version of the Cray-MPICH library which \
```
stdweird: hmmmm, i'm not in favour of such specific options. you  could probably generalise this to `mpi-mt` and use 
```
Directs the driver to link in an alternate version of the MPI library which provides fine-grained multi-threading support to applications that perform MPI operations within threaded regions (currently only supports Cray-MPICH).
```
(and then figure out a way that this only makes sense when using `Cray-MPICH` 
@boegel "regular" mpi also has this, so either this is already available or can be generalised.

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+                             provides fine-grained multi-threading support to applications that perform \
+                             MPI operations within threaded regions."),
+        'usewrappedcompiler': (False, "Use the embedded compiler instead of the wrapper"),
```
stdweird: hmm, odd option name, how about `bypasscompilerwrapper` or `directcompiler`? (and is this only relevant to the compiler, or does it bypass the wrapper completely (i mean wrt e.g. tuned MPI linking etc)

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    COMPILER_UNIQUE_OPTION_MAP = {
+        'pic': 'shared',
```
stdweird: add comment that there is no way to generate shared non-PIC binaries.

easybuild/toolchains/compiler/craypewrappers.py
stdweird: except when you do `usewrappedcompile`?

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+    COMPILER_F90 = 'ftn'
+
+    # FIXME (kehoste) hmmmm, really? then how do you control optimisation, precision when using the Cray wrappers?
+    COMPILER_FLAGS = []  # we dont have this for the wrappers
+    COMPILER_OPT_FLAGS = []  # or those
+    COMPILER_PREC_FLAGS = []  # and those for sure not !
+
+    # template and name suffix for PrgEnv module that matches this toolchain
+    # e.g. 'gnu' => 'PrgEnv-gnu/<version>'
+    PRGENV_MODULE_NAME_TEMPLATE = 'PrgEnv-%(suffix)s/%(version)s'
```
stdweird: i'd keep the suffix for an actual suffix and make this e.g.
```python
PRGENV_MODULE_NAME = None
PRGENV_MODULE_NAME_TEMPLATE = 'PrgEnv-%(name)s%(suffix)s/%(version)s'
PRGENV_MODULE_NAME_SUFFIX = ''
```

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+    PRGENV_MODULE_NAME_SUFFIX = None
+
+    # template for craype module (determines code generator backend of Cray compiler wrappers)
+    CRAYPE_MODULE_NAME_TEMPLATE = 'craype-%(optarch)s'
```
stdweird: also name,suffix and version (the fact that the `name` comes from `optarch` option, doesn't require that you use it like that in the template)

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+            self.log.error("Don't know which 'craype' module to load, 'optarch' build option is unspecified.")
+        else:
+            self.modules_tool.load([self.CRAYPE_MODULE_NAME_TEMPLATE % {'optarch': optarch}])
+
+    # FIXME: (kehoste) is it really needed to customise this?
+    # this looks like a workaround for setting the COMPILER_*_FLAGS lists empty?
+    def _set_compiler_flags(self):
+        """Collect the flags set, and add them as variables too"""
+
+        flags = [self.options.option(x) for x in self.COMPILER_FLAGS if self.options.get(x, False)]
```
stdweird: use `None` instead of `False` to inidicate missing option

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+        cflags = [self.options.option(x) for x in self.COMPILER_C_FLAGS + self.COMPILER_C_UNIQUE_FLAGS \
```
stdweird: remove the `\`

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+                  if self.options.get(x, False)]
+        fflags = [self.options.option(x) for x in self.COMPILER_F_FLAGS + self.COMPILER_F_UNIQUE_FLAGS \
```
stdweird: same here

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def _set_compiler_vars(self):
+        """Set compiler variables, either for the compiler wrapper, or the underlying compiler."""
+        if self.options.option('usewrappedcompiler'):
+            self.log.info("Using underlying compiler, as specified by the %s class" % Gcc)
+
+            comp_attrs = ['UNIQUE_OPTS', 'UNIQUE_OPTION_MAP', 'CC', 'CXX', 'C_UNIQUE_FLAGS',
+                          'F77', 'F90', 'F_UNIQUE_FLAGS']
+            for attr_name in ['COMPILER_%s' % a for a in comp_attrs]:
+                setattr(self, attr_name, getattr(IntelIccIfort, attr_name))
```
stdweird: from `IntelIccIfort`?

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+
+        super(CrayPEWrapperGNU,self)._set_compiler_vars()
```
stdweird: space after comma

easybuild/toolchains/compiler/craypewrappers.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+# Gcc's base is Compiler
+class CrayPEWrapperGNU(CrayPEWrapper):
+    """Support for using the Cray GNU compiler wrappers."""
+    TC_CONSTANT_CRAYPEWRAPPER = TC_CONSTANT_CRAYPEWRAPPER + '_GNU'
+
+    PRGENV_MODULE_NAME_SUFFIX = 'gnu'  # PrgEnv-gnu
+
+    def _set_compiler_vars(self):
```
stdweird: make a new private method in `CrayPEWrapper` that does this so you can share it with Intel and call it here like
```python
def _set_compiler_vars(self):
    self._new_method(Gcc, [])
    super(CrayPEWrapperGNU,self)._set_compiler_vars()
```
and for intel
def _set_compiler_vars(self):
    self._new_method(IntelIccIfort, ['LINKER_TOGGLE_STATIC_DYNAMIC'])
    super(CrayPEWrapperGNU,self)._set_compiler_vars()
```
(generating and reviewing almost exact copy/paste code is hard :smile:)

File: easybuild/toolchains/craycce.py [hunk 17]
```diff
@@ -0,0 +1,34 @@
+##
+# Copyright 2014 Petar Forai
```
stdweird: 2015

easybuild/toolchains/craygnu.py
```diff
@@ -0,0 +1,34 @@
+##
+# Copyright 2014 Petar Forai
```
stdweird: 2015

easybuild/toolchains/craycce.py
stdweird: @boegel do we insist on copyright for ugent in framework?

easybuild/toolchains/crayintel.py
```diff
@@ -0,0 +1,34 @@
+##
+# Copyright 2014 Petar Forai
```
stdweird: same issue

easybuild/tools/toolchain/toolchain.py
```diff
@@ -317,6 +317,10 @@ def is_dep_in_toolchain_module(self, name):
         """Check whether a specific software name is listed as a dependency in the module for this toolchain."""
         return any(map(lambda m: self.mns.is_short_modname_for(m, name), self.toolchain_dep_mods))
 
+    def _pre_prepare(self):
+        """Toolchain-specific preparations thay should be done first."""
```
stdweird: `thay`?

stdweird: either rename the method to indicate what is claimed  in the docstring (toolchain specific preparation in the prepare method of the toolchain module, who would have tought that...), but better just override the `prepare` method in the `CrayPEWrapper`.

stdweird:
@pforai you asked for it...

easybuild/toolchains/compiler/craypewrappers.py
pforai: Yeah, then you basically get a bare call to 'gcc' and friends.

easybuild/toolchains/compiler/craypewrappers.py
pforai: Yeah, total bypass.

boegel:
OK, I'm calling this.

What we have in place now is pretty solid and supports building HPL with all three toolchains (see https://github.com/hpcugent/easybuild-easyconfigs/pull/1538) and a couple of major HPC apps (CP2K, GROMACS, WRF) + Python (incl. numpy and scipy) on top of CrayGNU (see https://github.com/hpcugent/easybuild-easyconfigs/pull/1558).

We've pushed this under `--experimental` for now, so we can get feedback from experienced users of Cray systems on what we have in place now, and how to proceed.

Thanks a lot for working with me on this @pforai, and @stdweird for the reviewing and feedback.

Going in!

PR #1238, opened 2015-03-24
also copy rotated logs

boegel:


boegel:
@stdweird: please review

easybuild/tools/filetools.py
```diff
@@ -875,6 +876,36 @@ def rmtree2(path, n=3):
         _log.info("Path %s successfully removed." % path)
 
 
+def move_logs(from_path, target_path):
```
stdweird: `from_path` is not really a path but a logfile, maybe call it `main_logfile`?

stdweird: `target_path` is also no directory?

boegel: ok, will rename

easybuild/tools/filetools.py
```diff
@@ -875,6 +876,36 @@ def rmtree2(path, n=3):
...
+            # move log to target path
+            shutil.move(app_log, new_log_path)
+            _log.info("Moved log file %s to %s" % (from_path, new_log_path))
+
+    except (IOError, OSError), err:
+        _log.error("Failed to move log file(s) %s* to new log file %s*: %s" % (from_path, target_path, err))
+
+
 def cleanup(logfile, tempdir, testing):
     """Cleanup the specified log file and the tmp directory"""
```
stdweird: and this method should also support multiple logfiles

boegel:
remarks fixed, Jenkins happy, going in

Issue #1239, opened 2015-03-24
easybuild not checking the existence of the configuration file


pescobar:
This is a minor problem but when I try to use a non-existent easybuild configuration file the error doesn't say "config file not found". Easybuild just tries to run with the default parameters and as I use Lmod it complains about not finding EnvironmentModulesC

It would be nice if the error message is "config file /tmp/asdfasdfasdf not found"

```
[soft@compilator ~]$ eb --configfiles=/tmp/asdfasdfasdf
== temporary log file in case of crash /tmp/eb-kisXvE/easybuild-owPmud.log
ERROR: EasyBuild crashed with an error (at easybuild/tools/modules.py:231 in check_cmd_avail): EnvironmentModulesC modules tool can not be used, 'modulecmd' command is not available.
```

boegel:
@stdweird: thoughts? This is not good behaviour imho, `eb` should indeed be complaining that the specified config file isn't around?

stdweird:
hmm, i'm a bit surprised by this. will check generaloption to make this configurable if need be

boegel:
why make it configurable? specifying a non-existing config file is wrong, and you should be made aware?

stdweird:
https://github.com/hpcugent/vsc-base/blob/master/lib/vsc/utils/generaloption.py#L1108
so you can set it to True in EB

boegel:
OK, I'll look into that...

This will be tricky w.r.t. default config file locations, I think...

boegel:
see #1240, just a little bit tricky w.r.t. config unit tests (closing this, let's follow up in the PR)

boegel:
Thanks for reporting this @pescobar!

PR #1240, opened 2015-03-24
complain if specified config files are missing

boegel:
fix for #1239

@stdweird: please review

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -79,9 +79,10 @@ class EasyBuildOptions(GeneralOption):
     VERSION = this_is_easybuild()
 
     DEFAULT_LOGLEVEL = 'INFO'
-    DEFAULT_CONFIGFILES = DEFAULT_SYS_CFGFILES + [DEFAULT_USER_CFGFILE]
+    DEFAULT_CONFIGFILES = [p for p in DEFAULT_SYS_CFGFILES + [DEFAULT_USER_CFGFILE] if os.path.exists(p)]
 
     ALLOPTSMANDATORY = False  # allow more than one argument
+    CONFIGFILES_RAISE_MISSING = True  # don't allow non-existing config files to be specified
 
     def __init__(self, *args, **kwargs):
         """Constructor."""
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -48,7 +48,8 @@ from easybuild.framework.easyconfig.licenses import license_documentation
 from easybuild.framework.easyconfig.templates import template_documentation
 from easybuild.framework.easyconfig.tools import get_paths_for
 from easybuild.framework.extension import Extension
-from easybuild.tools import build_log, config, run  # @UnusedImport make sure config is always initialized!
+from easybuild.tools import build_log, config, run  # build_log should always stay there, to ensure EasyBuildLog
+from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import DEFAULT_LOGFILE_FORMAT, DEFAULT_MNS, DEFAULT_MODULES_TOOL, DEFAULT_MODULECLASSES
 from easybuild.tools.config import DEFAULT_PATH_SUBDIRS, DEFAULT_PREFIX, DEFAULT_REPOSITORY
 from easybuild.tools.config import get_pretend_installpath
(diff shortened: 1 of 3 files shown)


easybuild/tools/options.py
```diff
@@ -79,9 +80,10 @@ class EasyBuildOptions(GeneralOption):
     VERSION = this_is_easybuild()
 
     DEFAULT_LOGLEVEL = 'INFO'
-    DEFAULT_CONFIGFILES = DEFAULT_SYS_CFGFILES + [DEFAULT_USER_CFGFILE]
+    DEFAULT_CONFIGFILES = [p for p in DEFAULT_SYS_CFGFILES + [DEFAULT_USER_CFGFILE] if os.path.exists(p)]
```
stdweird: only test the `DEFAULT_USER_CFGFILE` (the other comes from a `glob`)

stdweird:
@boegel the main difference is that `--help` won't show the locations of the default config files anymore. so maybe that needs an option (is there a `--show-defaults` btw?)?

boegel:
@stdweird: that was a problem already, since `DEFAULT_SYS_CFGFILES` uses a `glob`...

There's no `--show-defaults` no, that's the job of `GeneralOption`, no?

boegel:
@stdweird: the default paths being considered for configuration files are clearly documented at http://easybuild.readthedocs.org/en/latest/Configuration.html#default-configuration-files, I doubt a lot of users rely on what `--help` lists as a 'default'

stdweird:
`--show-defaults` here means: show all paths EB considers. go already shows the defautl sin the help. but in this case EB, preprocesses the defaults (and that's probably the reason why the default was set to allow missing config files). config files that are specirfied on the commandline / environment shouldn't be missing tough

boegel:
Do you mean specifically for `--configfiles`, or across all path-related options?

stdweird:
config files, there is no such thing as `path-related options` in generaloption

stdweird:
@boegel i gave it some more thought. i'm going to implement what i described above: an option to only fail on missing config files that have been set explicitly via command or environment (and not fail on the default ones; unless the default ones are specified on commandline/environment)
then you can still choose which one to enforce.

but generaloption should fail when someone does `--configfiles=/wrong/path` by default imho. there was no existing code path to do that, but i think the required bit are available since `store_or_None`, so not that much work

boegel:
@stdweird: makes sense, that would make things significantly simpler here

While you're at it, how about adding support for specifying config files via a glob pattern (that may contain environment variables)?
Then I can make the default paths in EasyBuild truly static, by using (the equivalent of) `$XDG_CONFIG_DIR/easybuild.d/*.cfg` and letting generaloption figuring it out.

Maybe I'm asking too much here though, especially the env var part. ;-)

But, to make the default list of configuration files truly static (and make the default mentioned in `--help` output clear), that's what I would need...

boegel:
Added `--show-default-configfiles` as requested by @stdweird, example output:

```
$ XDG_CONFIG_DIRS=/tmp/etc:/tmp/moaretc eb --show-default-configfiles
Couldn't import dot_parser, loading of dot files will not be possible.
Default list of configuration files:

[with $XDG_CONFIG_HOME: (not set), $XDG_CONFIG_DIRS: /tmp/etc:/tmp/moaretc]

* user-level: ${XDG_CONFIG_HOME:-$HOME/.config}/easybuild/config.cfg
  -> /Users/kehoste/.config/easybuild/config.cfg => found
* system-level: ${XDG_CONFIG_DIRS:-/etc}/easybuild.d/*.cfg
  -> {/tmp/etc, /tmp/moaretc}/easybuild.d/*.cfg => /tmp/etc/easybuild.d/config.cfg, /tmp/moaretc/easybuild.d/bar.cfg, /tmp/moaretc/easybuild.d/foo.cfg

Default list of existing configuration files (4): /tmp/etc/easybuild.d/config.cfg, /tmp/moaretc/easybuild.d/bar.cfg, /tmp/moaretc/easybuild.d/foo.cfg, /Users/kehoste/.config/easybuild/config.cfg
```

boegel:
I'm not going to hold this back any longer. @stdweird: if you end up implementing the support for only complaining if specified config files are non-existent, we can update the code here accordingly

The only 'issue' now is that the default value for `--configfiles` as shown by `eb --help` as preprocessed in the sense that only *existing* config files are chosen; this is remedied by the documentation available at http://easybuild.readthedocs.org/en/latest/Configuration.html#default-configuration-files, and now also with `--show-default-configfiles`. So I don't consider that a blocker (especially since this PR doesn't change this, it was already like that).

PR #1242, opened 2015-03-26
Problem with timeout being considered a String

rjeschmi:
I needed this quick patch to get the download-timeout to work

should it be an int? the error was download_timeout not a Float was the only reason I did it this way.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -255,7 +255,7 @@ def download_file(filename, url, path):
 
     _log.debug("Trying to download %s from %s to %s", filename, url, path)
 
-    timeout = build_option('download_timeout')
+    timeout = float(build_option('download_timeout'))
     if timeout is None:
         # default to 10sec timeout if none was specified
         # default system timeout (used is nothing is specified) may be infinite (?)


boegel:
Jenkins: ok to test

boegel:
@rjeschmi: the Python docs (https://docs.python.org/2/library/urllib2.html#urllib2.urlopen) aren't very clear of the type, but casting it to a float is the better option imho, to avoid tripping over a values like `1.5` being passed

The unit tests should be enhanced to trigger the (now fixed) problem, I'll send a PR your way to tackle that.

easybuild/tools/filetools.py
```diff
@@ -255,7 +255,7 @@ def download_file(filename, url, path):
 
     _log.debug("Trying to download %s from %s to %s", filename, url, path)
 
-    timeout = build_option('download_timeout')
+    timeout = float(build_option('download_timeout'))
```
boegel: this doesn't work, since now you'll run into a crash without specifying a timeout (since the default value is `None`)

fixed in https://github.com/rjeschmi/easybuild-framework/pull/1

boegel:
Going in, thanks for the fix @rjeschmi!

PR #1246, opened 2015-03-27
make bootstrap script robust against having 'vsc-base' already available in Python search path

boegel:
somehow #1212 disappeared from `develop`...

not sure how that happened, may be a screwup of mine

this PR fixes the problem (by replacing the changes that were made in #1212, same branch)


boegel:
wait, no, this diff doesn't make any sense, since https://github.com/hpcugent/easybuild-framework/blob/develop/easybuild/scripts/bootstrap_eb.py is showing the contents as they should be *after* this PR

@github: wtf is going on here? PR diff views are showing the wrong things...

PR #1252, opened 2015-04-02
add support for appending/prepending to --robot-paths to avoid destroying the default (REVIEW)

boegel:
fix for feature request in https://github.com/hpcugent/easybuild/issues/101#issuecomment-78284371

depends on the merge of https://github.com/hpcugent/vsc-base/pull/162

note: WIP!

### DIFF
--- a/test/framework/config.py
+++ b/test/framework/config.py
@@ -527,10 +527,10 @@ class EasyBuildConfigTest(EnhancedTestCase):
         eb_go = eboptions.parse_options(args=[])
         self.assertEqual(eb_go.options.robot_paths, ['/foo', tmp_ecs_dir, '/bar/baz'])
 
-        # combining $EASYBUILD_ROBOT_PATHS and --robot-paths: all paths are retained in the order to be expected
+        # --robot-paths overrides $EASYBUILD_ROBOT_PATHS
         os.environ['EASYBUILD_ROBOT_PATHS'] = '/foobar::/barbar/baz/baz'
         eb_go = eboptions.parse_options(args=['--robot-paths=/one::/last'])
-        self.assertEqual(eb_go.options.robot_paths, ['/one', '/foobar', tmp_ecs_dir, '/barbar/baz/baz', '/last'])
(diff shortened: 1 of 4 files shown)


test/framework/config.py
```diff
@@ -428,6 +428,96 @@ def test_XDG_CONFIG_env_vars(self):
...
+        config_file = os.path.join(self.tmpdir, 'testconfig.cfg')
+        cfgtxt = '\n'.join([
+            '[config]',
+            'robot-paths=/cfgfirst::/cfglast',
+        ])
+        write_file(config_file, cfgtxt)
+        eb_go = eboptions.parse_options(args=['--configfiles=%s' % config_file])
+        self.assertEqual(eb_go.options.robot_paths, ['/cfgfirst', tmp_ecs_dir, '/cfglast'])
+
+        # cfgfile entry is lost when env var and/or cmdline options are used
```
boegel: @stdweird: Is this the intended behaviour? Using the prepend/append stuff on cfgfile + env or cmdline causes the cfgfile values to be lost?

stdweird: ofcourse. for the nth time, commandline precedes environment preceds configfiles.

boegel: Of course, but then why doesn't command line override environment, but complement it?

```
export EASYBUILD_ROBOT_PATHS='/foo::/bar/baz'
eb --robot-paths /one::/last
```

gives `['/one', '/foo', <DEFAULT>, '/bar/baz', '/last']` as robot search path instead of `['/one', <DEFAULT>, '/last']` (see test above). Hence my confusion...

stdweird: looks like a bug. please add the tests to vsc-base instead EB (EB shouldn't test dependencies in such detail).

boegel: dedicated unit tests for this: https://github.com/stdweird/vsc-base/pull/5

boegel:
Going in, thanks for the review @stdweird, and for tackling the required support in vsc-base for this!

PR #1060, opened 2014-10-15
Support Lua modulefiles (REVIEW)

pforai:
This branch ONLY supports lua module files at the moment and is very early. More to come in the following days.

PR #1255, opened 2015-04-03
[bugfix] restore newlines at the end of setenv/prepend-path/set-alias statements

boegel:
The newlines at the end of various statements generated by the module generator were removed in #1060, but shouldn't have been, since easyblocks using the methods that generate these statements assume that the newlines are there.

@wpoely86: please review
cc @ehiggs

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -144,7 +144,7 @@ class ModuleGeneratorTcl(ModuleGenerator):
 
     def comment(self, msg):
         """Return string containing given message as a comment."""
-        return "# %s" % msg
+        return "# %s\n" % msg
 
     def conditional_statement(self, condition, body, negative=False):
         """Return formatted conditional statement, with given condition and body."""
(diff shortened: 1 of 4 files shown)


easybuild/tools/module_generator.py
```diff
@@ -204,23 +204,23 @@ def load_module(self, mod_name):
             # not wrapping the 'module load' with an is-loaded guard ensures recursive unloading;
             # when "module unload" is called on the module in which the depedency "module load" is present,
             # it will get translated to "module unload"
-            load_statement = ["module load %(mod_name)s"]
+            load_statement = [self.LOAD_TEMPLATE, '']
         else:
             load_statement = [self.conditional_statement("is-loaded %(mod_name)s", self.LOAD_TEMPLATE, negative=True)]
-        return '\n'.join([''] + load_statement + ['']) % {'mod_name': mod_name}
```
wpoely86: why are we dropping the empty line here?

boegel: We're not. The string returned by `self.conditional_statement` already contains a newline, and for a non-conditional load the `''` will make sure it's followed by a newline

wpoely86:
I think it's okay but it's quite hard to follow with all the newlines coming and going...

boegel:
@wpoely86: it took me a while too, but the tests are checking things quite strictly (especially with the added test)

boegel:
Hooray for Jenkins!

Thanks for reviewing @wpoely86!

PR #1257, opened 2015-04-03
add support for Lmod-specific module family and module properties features

boegel:
* `--module-family` command line option to include family statement
* `--module-properties` command line option to include property statements (note: requires that the properties are listed in `$LMODRC` file, see e.g. https://github.com/TACC/Lmod/blob/master/init/lmodrc.lua)

both options work with Tcl *and* Lua module file syntax

Example (Tcl):

```
export LMODRC=$PWD/lmodrc.lua
export EASYBUILD_MODULE_PROPERTIES=lmod:sticky
eb bzip2-1.0.6.eb -f --module-family test --module-properties=state:testing,arch:mic:offload
```

```
#%Module
...
if { [ string match "*tcl2lua.tcl" $env(_) ] } {
    family "test"
}
if { [ string match "*tcl2lua.tcl" $env(_) ] } {
    add-property "lmod" "sticky"
    add-property "state" "testing"
    add-property "arch" "mic:offload"
}
...
```

Example (Lua):

```
export LMODRC=$PWD/lmodrc.lua
export EASYBUILD_MODULE_PROPERTIES=lmod:sticky
export EASYBUILD_MODULE_SYNTAX=Lua
eb bzip2-1.0.6.eb -f --module-family test --module-properties=state:testing,arch:mic:offload
```

```
...
family("test")
add_property("lmod", "sticky")
add_property("state", "testing")
add_property("arch", "mic:offload")
...
```

@rtmclay: any comments?


rtmclay:
The modulefiles for both TCL and Lua look good to me.

easybuild/framework/easyblock.py
```diff
@@ -188,15 +188,27 @@ def __init__(self, ec):
...
         self.group = None
         if group_name is not None:
             self.group = use_group(group_name)
 
+        # module features
+        self.module_family = build_option('module_family')
+        self.module_properties = None
+        if build_option('module_properties') is not None:
+            for prop_spec in build_option('module_properties'):
+                prop_spec_parts = prop_spec.split(':')
```
boegel: @stdweird: I'm not terribly happy with this bit. What I really need is a `listdict` type for configuration options.

Then when using `--module-properties=foo:bar;baz,one:two` would yield a value like this for `options.module_properties` (and hence also `build_option('module_properties')`):

```python
{'foo': ['bar', 'baz'], 'one': ['two']}
```

boegel:
feature request by @pescobar for this: also add `module_family` and `module_properties` easyconfig parameters

pescobar:
having the possibility to include module_family and module_properties in the easyconfigs would be nice to be able to rebuild a complete software stack in an automated way. Having the option only as a cli flag doesn´t allow an automated rebuild.

Also, I think it would be nice to define a list of Lmod properties which are supported by easybuild the same way we have now a list of supported moduleclasses. Maybe the properties included in the default file provided with Lmod https://github.com/TACC/Lmod/blob/master/init/lmodrc.lua is enough? or should be add few others? any suggestion on this?

easybuild/framework/easyblock.py
```diff
@@ -188,15 +188,28 @@ def __init__(self, ec):
...
         if group_name is not None:
             self.group = use_group(group_name)
 
+        # module features
+        self.module_family = build_option('module_family')
+        self.module_properties = None
+        if build_option('module_properties') is not None:
+            self.module_properties = []
+            for prop_spec in build_option('module_properties'):
+                prop_spec_parts = prop_spec.split(':')
```
boegel: this would benefit from https://github.com/hpcugent/vsc-base/pull/166

stdweird: no it won't. it will make it very easy to pass invalid data instead properly supported/checked options.

easybuild/framework/easyblock.py
```diff
@@ -839,6 +852,19 @@ def make_module_description(self):
         """
         return self.module_generator.get_description()
 
+    def make_module_features(self):
+        """Create features section of module file (module family, properties, ...)."""
+        txt = ''
+
+        if self.module_family is not None:
+            txt = self.module_generator.family(self.module_family)
```
stdweird: `+=`

easybuild/tools/module_generator.py
```diff
@@ -146,14 +186,17 @@ def comment(self, msg):
...
+    def conditional_statement(self, condition, body, negative=False, indent_body=True):
         """Return formatted conditional statement, with given condition and body."""
         if negative:
             lines = ["if { ![ %s ] } {" % condition]
         else:
             lines = ["if { [ %s ] } {" % condition]
 
-        lines.append('    ' + body)
+        if indent_body:
+            body = '    ' + body
```
stdweird: can we use a class constant `INDENTATION` or at least `' ' * 4` so we can read this?

easybuild/tools/module_generator.py
```diff
@@ -279,6 +322,21 @@ def set_alias(self, key, value):
...
 
+    def lmod_only(self, statements):
+        """Put provided statements under Lmod-only condition."""
+        # $_ contains full path to program being used to parse module file
+        # when using Lmod, this should always be the tcl2lua.tcl script
+        return self.conditional_statement('string match "*tcl2lua.tcl" $env(_)', statements, indent_body=False)
+
+    def family(self, family):
+        """Generate family statement."""
+        return self.lmod_only('    family "%s"' % family)
```
stdweird: use fixed indentation

easybuild/tools/module_generator.py
```diff
@@ -279,6 +322,21 @@ def set_alias(self, key, value):
...
+
+    def properties(self, properties):
+        """Generate property statements."""
+        property_statements = '\n'.join(['    add-property "%s" "%s"' % prop for prop in properties])
```
stdweird: fixed indentation

easybuild/tools/module_generator.py
```diff
@@ -303,15 +361,17 @@ def comment(self, msg):
...
         """Return formatted conditional statement, with given condition and body."""
         if negative:
             lines = ["if not %s then" % condition]
         else:
             lines = ["if %s then" % condition]
 
-        for line in body.split('\n'):
-            lines.append('    ' + line)
+        if indent_body:
+            body = '    ' + body
```
stdweird: fixed indentation

easybuild/tools/options.py
```diff
@@ -239,8 +239,10 @@ def config_options(self):
             # purposely take a copy for the default logfile format
             'logfile-format': ("Directory name and format of the log file",
                                'strtuple', 'store', DEFAULT_LOGFILE_FORMAT[:], {'metavar': 'DIR,FORMAT'}),
+            'module-family': ("Include family statement in generated module file", 'str', 'store', None),
```
stdweird: no checks on what is supported?
can you give an example what could/should be added here?

easybuild/tools/options.py
```diff
@@ -239,8 +239,10 @@ def config_options(self):
...
             'module-naming-scheme': ("Module naming scheme",
                                      'choice', 'store', DEFAULT_MNS, sorted(avail_module_naming_schemes().keys())),
+            'module-properties': ("Include property statements in generated module file", 'strlist', 'extend', None),
```
stdweird: same remark: no checks, so what are you trying to achieve here?

test/framework/options.py
```diff
@@ -1536,6 +1536,88 @@ def test_robot(self):
...
             self.assertTrue(ec_regex.search(outtxt), "Pattern %s found in %s" % (ec_regex.pattern, outtxt))
 
+    def test_module_features(self):
+        """Test use of --module-family and --module-properties."""
+        # use toy-0.0.eb easyconfig file that comes with the tests
+        eb_file = os.path.join(os.path.dirname(__file__), 'easyconfigs', 'toy-0.0.eb')
+
+        # put lmodrc file in place (properties must be known)
+        lmodrc = os.path.join(self.test_prefix, 'lmodrc.lua')
+        lmodrc_txt = '\n'.join([
```
stdweird: worst misusage of join ever. put this in a multiline module constant (it's fixed/readinly text  anyway).

easybuild/framework/easyblock.py
boegel: if you pass invalid properties, Lmod will complain and the `module load` done during the sanity check will not work, so you'll know

geimer:
Regarding families: Instead of guarding the `family` statement in Tcl, the corresponding functionality could also be emulated using environment variables:
```
  switch [module-info mode] {
    load {
      if [info exists env(EB_FAMILY_FOO)] {
        puts "Conflicting module of 'foo' family detected!"
        puts "Please unload module \"$env(EB_FAMILY_FOO)\" first."
        exit
      } else {
        setenv EB_FAMILY_FOO "my_pkg"
      }
    }
  }
```
Not sure how much effort to put into it, though.

boegel:
@geimer: that's interesting... I'd love to hear @rtmclay's thoughts on that.

geimer:
@boegel It's a poor man's approach to prevent loading two modules of the same family. I've used it in the past and it works pretty well. Lmod is actually going beyond that by doing a switch.  However, exiting with an error is probably the best one can do for Tmod as it doesn't automatically reload modules.

fgeorgatos:
IMHO, exiting should be the default, in the name of safety, with the
automated  approach being a tunable
(fi. we might have even distinct policies about what kind of swap policy
kicks in, upon module swaps)

Also, I had a related idea in the past based on the concept of a var
holding: FAMILY1:FAMILY2:FAMILY3 etc
If we need to store individual software features though, then that gets
more interesting..


On 14 November 2015 at 22:04, geimer <notifications@github.com> wrote:

> @boegel <https://github.com/boegel> It's a poor man's approach to prevent
> loading two modules of the same family. I've used it in the past and it
> works pretty well. Lmod is actually going beyond that by doing a switch.
> However, exiting with an error is probably the best one can do for Tmod as
> it doesn't automatically reload modules.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hpcugent/easybuild-framework/pull/1257#issuecomment-156749280>
> .
>



-- 


echo "sysadmin know better bash than english"|sed s/min/mins/ \
  | sed 's/better bash/bash better/' # signal detected in a CERN forum

easybuild/tools/module_generator.py
```diff
@@ -325,6 +362,21 @@ def set_alias(self, key, value):
         # quotes are needed, to ensure smooth working of EBDEVEL* modulefiles
         return 'set-alias\t%s\t\t%s\n' % (key, quote_str(value))
 
+    def lmod_only(self, statements):
+        """Put provided statements under Lmod-only condition."""
+        # $_ contains full path to program being used to parse module file
+        # when using Lmod, this should always be the tcl2lua.tcl script
+        return self.conditional_statement('string match "*tcl2lua.tcl" $env(_)', statements, indent_body=False)
```
boegel: quoting @rtmclay:

```
I like the trick with using $_ to figure out whether Lmod or Tmod is parsing a TCL modulefile.   This trick works on Linux and Mac OS X and Solaris.  It would probably be good to check that all the various version of Unix you are trying to run Lmod also support $_.

HOWEVER, the internals of Lmod have changed.  The program tcl2lua.tcl is no longer called directly.  Instead it is /path/to/tclsh /path/to/tcl2lua.tcl  is called.  So I suggested that you use this instead:

     if { [string match "*modulecmd*" $env(_)]} {
       #Tmod
    } else {
       #Lmod
    }

The string *modulecmd* will match both modulecmd and modulecmd.tcl.  This if block will work with older and newer version of Lmod.
```

(cfr. https://sourceforge.net/p/lmod/mailman/message/35396987/)

JackPerdue:
This might not be where to request this (given the age of the issue) but I'd like to have an option to simply add another Lmod "conflict" statement to the generated module.  In the case of Anaconda, I would want it to conflict with both Anaconda (other versions) and Python (any version).  I don't want to do this for all builds, just select packages.  If there was a "additional_conflicts" tag I could add to an EasyConfig, that would probably best (and simplest) for now.

akesandgren:
@boegel Status?

boegel:
Someone interested in this should pick up on this, but it's likely most of the effort here will have to be re-done due to all the merge conflicts...

I'll leave this open as a reminder, this is probably a feasible thing to work on during a hackathon/user meeting.

boegel:
W.r.t. @JackPerdue's question: let's open a separate issue on that, since that's a feature request orthogonal to the "family" & "properties" concepts.

PR #1258, opened 2015-04-05
add --installpath-modules and --installpath-software configuration options

boegel:
fix for https://github.com/hpcugent/easybuild/issues/105

Required to be able to install software and modules in totally different paths; this will come in useful when adding support for only (re)installing modules, so modules in using a different naming scheme can be installed in a different path, without having to reinstall the software.

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -334,8 +334,7 @@ def install_path(typ=None):
     res = variables[key]
     if res is None:
         key = 'subdir_%s' % typ
-        subdir = variables[key]
-        res = os.path.join(variables['installpath'], subdir)
+        res = os.path.join(variables['installpath'], variables[key])
         _log.debug("%s install path as specified by 'installpath' and '%s': %s", typ, key, res)
     else:
         _log.debug("%s install path as specified by '%s': %s", typ, key, res)
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -324,14 +324,21 @@ def install_path(typ=None):
     elif typ == 'mod':
         typ = 'modules'
 
+    known_types = ['modules', 'software']
+    if typ not in known_types:
+        raise EasyBuildError("Unknown type specified in install_path(): %s (known: %s)", typ, ', '.join(known_types))
+
     variables = ConfigurationVariables()
 
-    res = None
-    if variables.get('installpath_%s' % typ, None) is None:
-        suffix = variables['subdir_%s' % typ]
-        res = os.path.join(variables['installpath'], suffix)
+    key = 'installpath_%s' % typ
+    res = variables[key]
+    if res is None:
+        key = 'subdir_%s' % typ
+        subdir = variables[key]
+        res = os.path.join(variables['installpath'], subdir)
+        _log.debug("%s install path as specified by 'installpath' and '%s': %s", typ, key, res)
     else:
-        res = variables['installpath_%s' % typ]
+        _log.debug("%s install path as specified by '%s': %s", typ, key, res)
 
     return res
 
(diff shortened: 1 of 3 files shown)


easybuild/tools/config.py
```diff
@@ -323,8 +325,15 @@ def install_path(typ=None):
         typ = 'modules'
```
stdweird: add check that only valid `typ` are `software` and `modules`

easybuild/tools/config.py
```diff
@@ -323,8 +325,15 @@ def install_path(typ=None):
...
 
     variables = ConfigurationVariables()
-    suffix = variables['subdir_%s' % typ]
-    return os.path.join(variables['installpath'], suffix)
+
+    res = None
+    if variables.get('installpath_%s' % typ, None) is None:
```
stdweird: you mean
```
ip = 'installpath_%s' % typ
if ip in variables:
```

easybuild/tools/config.py
```diff
@@ -323,8 +325,15 @@ def install_path(typ=None):
...
+        suffix = variables['subdir_%s' % typ]
```
stdweird: `suffix` -> `subdir`

easybuild/tools/options.py
```diff
@@ -236,6 +236,8 @@ def config_options(self):
                             'strlist', 'store', ['.git', '.svn']),
             'installpath': ("Install path for software and modules",
                             None, 'store', mk_full_default_path('installpath')),
+            'installpath-modules': ("Install path for modules", None, 'store', None),
```
stdweird: make it clear what the default is (and if this a relative or absolute or both path)

easybuild/tools/options.py
```diff
@@ -364,23 +366,35 @@ def unittest_options(self):
...
-        error_cnt = 0
+        error_msgs = []
 
         for opt in ['software', 'try-software', 'toolchain', 'try-toolchain']:
             val = getattr(self.options, opt.replace('-', '_'))
             if val and len(val) != 2:
-                self.log.warning('--%s requires NAME,VERSION (given %s)' % (opt, ','.join(val)))
-                error_cnt += 1
+                msg = "--%s requires NAME,VERSION (given %s)" % (opt, ','.join(val))
+                self.log.warning(msg)
```
stdweird: don't warn here

easybuild/tools/options.py
```diff
@@ -364,23 +366,35 @@ def unittest_options(self):
...
+            subdir_opt = 'subdir_%s' % typ
+            val = getattr(self.options, subdir_opt)
+            if os.path.isabs(getattr(self.options, subdir_opt)):
+                msg = "Configuration option '%s' must specify a *relative* path (use 'installpath-%s' instead?): '%s'"
+                msg = msg % (subdir_opt, typ, val)
+                self.log.warning(msg)
+                error_msgs.append(msg)
+
+        if error_msgs:
+            raise EasyBuildError("Found problems validating the options, treating warnings in log file as fatal: %s",
```
stdweird: warn here, one by one.

stdweird: or better yet, don't warn at all. the messages are in the raise message

easybuild/tools/config.py
boegel: no, that won't work, since the `installpath_software` key will always be there

I'll change it to not using `.get` and inverse the logic

easybuild/tools/config.py
boegel: done

boegel:
documentation update worked out in https://github.com/hpcugent/easybuild/pull/114, merging this in (remarks are fixed, Jenkins is happy)

thanks for the review @stdweird!

PR #1261, opened 2015-04-09
fix vsc-base dep version

boegel:


boegel:
2.2.0 comes with https://github.com/hpcugent/vsc-base/pull/162, which is not merged yet

PR #1263, opened 2015-04-21
fix dealing with empty variables in Variables.join method

boegel:
the tests failed with the patch in variables.py

this popped up when using `BLAS_LIB = []` and `LAPACK_IS_BLAS = True` in the support for the CrayPE compiler drivers (cfr. https://github.com/hpcugent/easybuild-framework/pull/1234)

@stdweird: please review
cc @pforai

### DIFF
--- a/easybuild/tools/variables.py
+++ b/easybuild/tools/variables.py
@@ -481,6 +481,10 @@ class Variables(dict):
                 else it is nappend-ed
         """
         self.log.debug("join name %s others %s" % (name, others))
+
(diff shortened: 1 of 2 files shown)


pforai:
Looks good to me, also we need exactly this semantics - ie push empty env vars, but have it defined for compatibility in the blocks to not have to do checks around which TC is in use and then remove the vars in the Cray cases.

stdweird:
@boegel yeah, LGTM :smile:

boegel:
\o/

PR #1264, opened 2015-04-23
get rid of $PROFILEREAD hack when running commands, not needed anymore

boegel:
We used to make sure that `/etc/profile.d/modules.sh` was being resources on every command being executed when `$PROFILEREAD` was defined (i.e. on SuSE Linux systems).

The only reason why this was needed was to make sure that the `module` function is defined in the environment in which the command is being executed.

However, we don't rely on `module` at all, so there's no longer any need for this.

@pforai: feedback, since you guys are running SuSE?

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -60,29 +60,6 @@ ERROR = 'error'
 strictness = WARN
 
 
-def adjust_cmd(func):
-    """Make adjustments to given command, if required."""
-
-    def inner(cmd, *args, **kwargs):
-        # SuSE hack
-        # - profile is not resourced, and functions (e.g. module) is not inherited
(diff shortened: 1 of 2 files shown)


pforai:
I think it should stay there, but be disabled by default on SLES. You _might_ need this at some point in time. So just wrap it in a default disabled configure option would be preferable, imho.

boegel:
The only reason we would ever need this (again) is if we need the `module` function defined, right?

Besides, removing it from the code doesn't mean we lose it, it's in the git history after all.

pforai:
Yes, also that it assumes that modules.sh is the file that needs to be sourced. OK, lets get rid of this then.

wpoely86:
looks fine

boegel:
Thanks for the feedback! Let's get rid of this cruft...

PR #1265, opened 2015-04-23
add support for specifying and using external modules metadata (REVIEW)

boegel:


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -549,7 +551,8 @@ def _parse_dependency(self, dep, hidden=False):
             'version': None,
             'versionsuffix': '',
             'hidden': hidden,
-            'external_module': False
+            'external_module': False,
+            'root': None,
```
stdweird: add a comment to explain what this is about (and maybe also the ones above). it seems like magic here

easybuild/tools/options.py
```diff
@@ -54,6 +54,7 @@
 from easybuild.tools.config import DEFAULT_MODULECLASSES, DEFAULT_PATH_SUBDIRS, DEFAULT_PREFIX, DEFAULT_REPOSITORY
 from easybuild.tools.config import get_pretend_installpath
 from easybuild.tools.config import mk_full_default_path
+from easybuild.tools.configobj import ConfigObj
```
stdweird: start with regular configparser

easybuild/tools/options.py
```diff
@@ -235,6 +236,7 @@ def config_options(self):
             'avail-repositories': ("Show all repository types (incl. non-usable)",
                                    None, "store_true", False,),
             'buildpath': ("Temporary build path", None, 'store', mk_full_default_path('buildpath')),
+            'external-modules-metadata': ("File specifying metadata for external modules", None, 'store', None),
```
stdweird: use a comma spearated list of files, also specify the format (eg `INI File(s) ...`)

easybuild/tools/options.py
stdweird: nvm, keep ConfigObj for easy access; just don't mention it's configobj :smile:

easybuild/tools/options.py
```diff
@@ -444,6 +446,12 @@ def postprocess(self):
             if token is None:
                 raise EasyBuildError("Failed to obtain required GitHub token for user '%s'", self.options.github_user)
 
+        # parse file specifying metadata for external modules
+        if self.options.external_modules_metadata:
```
stdweird: no need for `if/else`; `ConfigObj` instance is empty dict

easybuild/tools/options.py
```diff
@@ -444,6 +446,12 @@ def postprocess(self):
...
+            self.options.external_modules_metadata = ConfigObj(self.options.external_modules_metadata)
```
stdweird: i don't know what error to expect here with incalid syntax, but catch them here and reraise them with proper message saying they come from external_modules_metadata

easybuild/framework/easyconfig/easyconfig.py
boegel: I'll rename this to `prefix` (and add a comment)

boegel:
@stdweird: ready for re-review; this works as expected, I was able to build WRF on top of the Cray-provided `cray-hdf5-parallel` and `cray-netcdf` when combining this with #1234

pforai:
Totally like it!

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -544,12 +546,13 @@ def _parse_dependency(self, dep, hidden=False):
...
             'short_mod_name': None,  # short module name
-            'name': None,  # software name
+            'name': None,  # software name (may be a list of names, e.g. for external modules)
             'toolchain': None,
             'version': None,
             'versionsuffix': '',
             'hidden': hidden,
-            'external_module': False
+            'external_module': False,
+            'prefix': None,  # installation prefix (only relevant for external modules)
```
stdweird: then call it `external_prefix` (way too general otehrwise)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -544,12 +546,13 @@ def _parse_dependency(self, dep, hidden=False):
...
             'full_mod_name': None,  # full module name
             'short_mod_name': None,  # short module name
-            'name': None,  # software name
+            'name': None,  # software name (may be a list of names, e.g. for external modules)
             'toolchain': None,
             'version': None,
             'versionsuffix': '',
             'hidden': hidden,
-            'external_module': False
+            'external_module': False,
```
stdweird: add comment (and also the ones above)

easybuild/tools/environment.py
```diff
@@ -84,10 +84,11 @@ def setvar(key, value):
     put key in the environment with value
     tracks added keys until write_changes has been called
     """
+    oldval = os.environ.get(key, 'None (not defined)')
```
stdweird: which one is it: `None` or `(not defined)`

boegel: This is just for the debug message below; it's a vain attempt to make it clear that the environment variable is not defined. But I'll do a better job here, will fix.

easybuild/tools/options.py
```diff
@@ -444,8 +447,26 @@ def postprocess(self):
...
                 raise EasyBuildError("Failed to obtain required GitHub token for user '%s'", self.options.github_user)
 
+        self._postprocess_external_modules_metadata()
+
         self._postprocess_config()
 
+    def _postprocess_external_modules_metadata(self):
+        """Parse file(s) specifying metadata for external modules."""
+        parsed_external_modules_metadata = ConfigObj()
+        for path in self.options.external_modules_metadata:
```
stdweird: just pass the whole list to configobj in one go? (unless configObj doesn't raise error on missing files)

easybuild/tools/toolchain/toolchain.py
```diff
@@ -317,6 +318,43 @@ def is_dep_in_toolchain_module(self, name):
...
+        # load modules for all dependencies
+        dep_mods = [dep['short_mod_name'] for dep in self.dependencies]
+        self.log.debug("Loading modules for dependencies: %s" % dep_mods)
+        self.modules_tool.load(dep_mods)
+
+        # define $EBROOT* and $EBVERSION* for external modules, if metadata is available
+        for dep in [d for d in self.dependencies if d['external_module']]:
+            self.log.debug("Defining $EB* environment variables for external module %s", dep['short_mod_name'])
+
+            names = dep['name'] or []
```
stdweird: `dep.get('name', [])`

easybuild/tools/toolchain/toolchain.py
```diff
@@ -317,6 +318,43 @@ def is_dep_in_toolchain_module(self, name):
...
+            if isinstance(names, basestring):
```
stdweird: ? what other types are expected here?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -415,9 +453,20 @@ def _add_dependency_variables(self, names=None, cpp=None, ld=None):
...
             deps = [{'name': name} for name in names if name is not None]
 
-        for root in self.get_software_root([dep['name'] for dep in deps]):
-            self.variables.append_subdirs("CPPFLAGS", root, subdirs=cpp_paths)
-            self.variables.append_subdirs("LDFLAGS", root, subdirs=ld_paths)
+        dep_names = []
+        for dep in deps:
+            # name may be None or a list of names for external modules
+            if dep['name'] is not None:
+                if isinstance(dep['name'], list):
```
stdweird: use 
```python
if isinstance(dep['name'], basestring):
    dep['name']=[dep['name']]
dep_names.extend(dep['name'])
```
but why can the `name` have multiple types?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -415,9 +453,20 @@ def _add_dependency_variables(self, names=None, cpp=None, ld=None):
         else:
             deps = [{'name': name} for name in names if name is not None]
 
-        for root in self.get_software_root([dep['name'] for dep in deps]):
-            self.variables.append_subdirs("CPPFLAGS", root, subdirs=cpp_paths)
-            self.variables.append_subdirs("LDFLAGS", root, subdirs=ld_paths)
+        dep_names = []
+        for dep in deps:
+            # name may be None or a list of names for external modules
```
stdweird: then check that, all other modules should raise error (and stop doing magic/hackish things like dynamic types)

easybuild/tools/toolchain/toolchain.py
```diff
@@ -415,9 +453,20 @@ def _add_dependency_variables(self, names=None, cpp=None, ld=None):
...
+            if dep['name'] is not None:
+                if isinstance(dep['name'], list):
+                    dep_names.extend(dep['name'])
+                else:
+                    dep_names.append(dep['name'])
+
+        for root in self.get_software_root(dep_names):
+            # software install prefix may be None for external modules
```
stdweird: then check that

stdweird:
@boegel not ok until there's documentation for the added options

easybuild/tools/options.py
boegel: ConfigObj just ignores missing files:

```
>>> from easybuild.tools.configobj import ConfigObj
>>> ConfigObj('/no/such/file')
ConfigObj({})
```

if a user specifies a non-existing file, we should complain imho, because something fishy is going on

also, it doesn't accept a list of files (I think)

boegel:
Documentation update via https://github.com/hpcugent/easybuild/pull/116 about to be merged, so this is good to go.

Thanks for the review @stdweird, and for the suggestions for improvements @wpoely86!

PR #1268, opened 2015-04-26
check whether craype exists before loading it

boegel:
just to make sure a clear error message is shown


pforai:
LGM!

boegel:
Thanks for the review @pforai!

PR #1271, opened 2015-04-29
Add support for 'whatis' easyconfig parameter

geimer:
Currently, the `module whatis` output includes the full package description. This PR adds support for a new (optional) easyconfig parameter `whatis`. If given, the value will be used when generating the modulefile; otherwise it will fall back to using the value of `description` for backwards compatibility.


boegel:
Jenkins: ok to test

geimer:
While looking at this again, I noticed that the lua version includes the name and version twice: once as a prefix to the output and once as the output itself.  Does it make sense to rectify this?  Maybe we should also get @pforai on board?

geimer:
Here is an example to clarify what I meant by my last comment:
```
$ module whatis GCC/4.9.2
GCC/4.9.2           : Name: GCC
GCC/4.9.2           : Version: 4.9.2
GCC/4.9.2           : Description: The GNU Compiler Collection includes front ends for C, C++, Objective-C, Fortran, Java, and Ada,
 as well as libraries for these languages (libstdc++, libgcj,...). - Homepage: http://gcc.gnu.org/
GCC/4.9.2           : Homepage: http://gcc.gnu.org/

$ module whatis GCC/4.8.2
GCC/4.8.2           : Description: The GNU Compiler Collection includes front ends for C, C++, Objective-C, Fortran, Java, and Ada,  as well as libraries for these languages (libstdc++, libgcj,...). - Homepage: http://gcc.gnu.org/
```
The GCC/4.9.2 module is in Lua format, GCC/4.8.2 is Tcl.

boegel:
@geimer: let's follow this up in August, tagging this for EB v2.3

geimer:
@boegel: Do you think we can finalize this in time for EB v2.4.0?

boegel:
@geimer: doesn't look like a lot of work, but I don't want to rush it in either, especially since it's likely we'll be doing some work on the module generator part at the hackathon in Austin, cfr. #1358

Also, I would really like to try and get #1306 ready to merge (needs code cleanup & unit tests), and to get  #1388 merged as well. I'm not sure how I'll ever pull that off, so I'm not keen on adding other stuff on top of it. ;-)

How about we revisit this together with what needs to be done for #1358?

geimer:
@boegel: I just wanted to get it off the list ;-) But your proposal makes sense. I was planning to look into improving the documentation abilities anyway (the stuff we discussed during the f2f meeting in Ghent) and there seems to be a significant overlap with #1358.

boegel:
@geimer: moved to milestone 2.5.0 (which should be there by the end of 2015), let's try and work on this during the hackathon in Austin

geimer:
The latest version of this PR removes the inconsistency between Lua and Tcl modules mentioned above.  This also addresses the desire of TACC to be able to specify, for example, the version explicitly (see https://github.com/hpcugent/easybuild-framework/issues/1358) , which may conflict with the automatically generated string.

boegel:
@geimer: please take a look at https://github.com/geimer/easybuild-framework/pull/4

boegel:
Going in (finally), thanks for your efforts on this @geimer!

PR #1275, opened 2015-05-06
only use environment variable to find module command binary if command can't be found in $PATH

boegel:
If `lmod` is found in `$PATH`, we shouldn't be using the value of `$LMOD_CMD` to determine the path to `lmod`.

When Lmod is available both on the system and in the user's account, we may end up using the wrong `lmod` binary that way (cfr. Taito @ CSC.fi).

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -122,7 +122,8 @@ class ModulesTool(object):
     TERSE_OPTION = (0, '--terse')
     # module command to use
     COMMAND = None
-    # environment variable that may refer to module command
+    # environment variable to determine path to module command;
+    # used as fallback in case command is not available in $PATH
     COMMAND_ENVIRONMENT = None
     # run module command explicitly using this shell
     COMMAND_SHELL = None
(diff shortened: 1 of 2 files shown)


easybuild/tools/modules.py
```diff
@@ -122,7 +122,7 @@ class ModulesTool(object):
     TERSE_OPTION = (0, '--terse')
     # module command to use
     COMMAND = None
-    # environment variable to determine the module command (instead of COMMAND)
+    # environment variable that may refer to module command
```
stdweird: missing documentation how it resolves it now.

easybuild/tools/modules.py
```diff
@@ -156,8 +156,8 @@ def __init__(self, mod_paths=None, testing=False):
 
         # actual module command (i.e., not the 'module' wrapper function, but the binary)
         self.cmd = self.COMMAND
-        if self.COMMAND_ENVIRONMENT is not None and self.COMMAND_ENVIRONMENT in os.environ:
-            self.log.debug('Set command via environment variable %s' % self.COMMAND_ENVIRONMENT)
+        if which(self.cmd) is None and self.COMMAND_ENVIRONMENT in os.environ:
```
stdweird: i would log the case where `which(self.cmd)` and `self.COMMAND_ENVIRONMENT` are not equal (debug is enough, but can help debugging with messed up setups)

easybuild/tools/modules.py
boegel: docs update at https://github.com/hpcugent/easybuild/pull/119

easybuild/tools/modules.py
boegel: OK, makes sense

boegel:

Issue #1274, opened 2015-05-05
modules generated with --module-only are incomplete


boegel:
`load` statements for dependencies are missing when additional modules are generated using `--module-only`

boegel:
fixed in #1276

PR #1276, opened 2015-05-06
fix issue with missing load statements when --module-only is used, don't skip ready/prepare steps

boegel:
fix for #1274

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1738,8 +1738,10 @@ class EasyBlock(object):
             self.log.info("Skipping %s step (skip: %s, skipsteps: %s)", step, self.skip, self.cfg['skipsteps'])
             skip = True
 
-        # skip step when only generating module file; still run sanity check without use of force
-        elif module_only and not step in ['sanitycheck', 'module']:
+        # skip step when only generating module file
+        # * still run sanity check without use of force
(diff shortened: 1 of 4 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1738,8 +1738,10 @@ def _skip_step(self, step, skippable):
             self.log.info("Skipping %s step (skip: %s, skipsteps: %s)", step, self.skip, self.cfg['skipsteps'])
             skip = True
 
-        # skip step when only generating module file; still run sanity check without use of force
-        elif module_only and not step in ['sanitycheck', 'module']:
+        # skip step when only generating module file
+        # * still run sanity check without use of force
+        # * always run ready & prepare step to set up toolchain + deps
+        elif module_only and not step in ['ready', 'prepare', 'sanitycheck', 'module']:
```
stdweird: need modifications in the help message of module only. so move the list of steps to constants and genereate the help from it

boegel: OK, will do, nice catch

boegel:
@wpoely86: please (re)review, I've fixed Stijn's remarks, but added quite a few of extra changes by introducing constants for the step names

stdweird:
LGTM, :+1: for switching to constants

boegel:

PR #1277, opened 2015-05-07
fix location of module_only build option w.r.t. default value

boegel:
trivial change, more for cleanup purposes and avoid confusion in the future than anything else


boegel:
no-brainer, going in

Issue #1279, opened 2015-05-11
problem when defining hidden deps in easyconfig and in command line at the same time


pescobar:
I found and issue when defining hidden deps in a easyconfig and also on command line (or in a easybuild config file)

As example, if this is in the easyconfig:

```
hiddendependencies = [
    ('libreadline', '6.3'),
    ('ncurses', '5.9'),
    ('libpng', '1.6.16'),  # for plotting in R
    ('libjpeg-turbo', '1.4.0'),  # for plottting in R
    ('LibTIFF', '4.0.3'),  # for plottting in R
    ('cairo', '1.12.14'),  # for plottting in R
]
```

and then I run easybuild like this:

``` eb --hide-deps=libreadline,ncurses,libpng```

I get this error:

```
ERROR: Failed to process easyconfig R-3.2.0-goolf-1.4.10.eb: Hidden dependencies with visible module names ['li
breadline/6.3-goolf-1.4.10', 'ncurses/5.9-goolf-1.4.10', 'libpng/1.6.16-goolf-1.4.10', 'libjpeg-turbo/1.4.0-goolf-1.4.10'] not in list of dependen
cies: ['libreadline/.6.3-goolf-1.4.10', 'ncurses/.5.9-goolf-1.4.10', 'libpng/.1.6.16-goolf-1.4.10', 'libjpeg-turbo/.1.4.0-goolf-1.4.10', 'LibTIFF/
4.0.3-goolf-1.4.10', 'cairo/1.12.14-goolf-1.4.10', 'Java/1.7.0_75', 'zlib/.1.2.8-goolf-1.4.10', 'cURL/7.40.0'] 
```

it seems like EB is trying to hide dependencies which are already hidden. See this detail in the error message:
```
not in list of dependencies: ['libreadline/.6.3-goolf-1.4.10'
```

boegel:
Thanks for reporting this @pescobar, fixed in #1280, should become part of EasyBuild v2.1.1

pescobar:
great @boegel !!!   thanks for the quick fix..........some day you will be as fast as @rtmclay   :P

boegel:
@pescobar: well, I can try ;)

PR #1280, opened 2015-05-11
fix combined use of --hide-deps and hiddendependencies

boegel:
fix for bug #1279, reported by @pescobar


wpoely86:
looks fine I think

boegel:
Thanks for the review @wpoely86. The unit test confirms it's the proper fix, before this patch you'd run into an error message as mentioned in #1279

PR #1281, opened 2015-05-11
enhance --search: only consider actual filename (not entire path), use regex syntax

boegel:
This change is motivated by the output of `--search` that was produced by `eb --search python`, which in practice resulted in *all* easyconfigs to be listed (since it's not unlikely that 'python' is part of the full path).

With this change, this works as expected:

```
eb -S '^Python-[23].*eb' 
```

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -342,6 +342,8 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False):
         raise EasyBuildError("search_file: ignore_dirs (%s) should be of type list, not %s",
                              ignore_dirs, type(ignore_dirs))
 
+    query = re.compile(query.lower())
+
     var_lines = []
     hit_lines = []
     var_index = 1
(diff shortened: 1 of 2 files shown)


easybuild/tools/filetools.py
```diff
@@ -349,18 +351,16 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False):
...
         hit_in_path = False
-        print_msg("Searching (case-insensitive) for '%s' in %s " % (query, path), log=_log, silent=silent)
+        print_msg("Searching (case-insensitive) for '%s' in %s " % (query.pattern, path), log=_log, silent=silent)
 
-        query = query.lower()
         for (dirpath, dirnames, filenames) in os.walk(path, topdown=True):
             for filename in filenames:
-                filename = os.path.join(dirpath, filename)
-                if filename.lower().find(query) != -1:
+                if query.search(filename.lower()):
```
wpoely86: add  `re.IGNORECASE` to the regex?

easybuild/tools/filetools.py
```diff
@@ -342,6 +342,8 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False):
         raise EasyBuildError("search_file: ignore_dirs (%s) should be of type list, not %s",
                              ignore_dirs, type(ignore_dirs))
 
+    query = re.compile(query.lower())
```
wpoely86: why not do it trough the regex support of options system? And why the `.lower()`?

easybuild/tools/filetools.py
boegel: good point, totally forgot about this option :+1:

boegel:

boegel:
@martbhell: this resolves the problem you ran into during the EasyBuild hackathon @ CSC.fi

PR #1282, opened 2015-05-14
remove log handlers that were added during tests, to ensure effective cleanup of log files

boegel:
Although each test is properly cleaning up the tmp dir that was assigned to it in `tearDown`, the disk space used by the log files was not being freed because of open file handles on the log files.

I looked into properly cleaning up those file handles to trigger the disk space being freed, but this proved to be far from trivial (I concluded this after quite a couple of hours of digging around, accompanied with lots of swearing).

Alternative solution (thanks to @stdweird for the tip): quickly rotate the log files, so that only small bits (max 1MB per log) are left behind in the overall tmp directory used by the running test suite.

This effectively trims the required disk space in /tmp (or equivalent) down from ~3GB to less than 100MB.

### DIFF
--- a/test/framework/utilities.py
+++ b/test/framework/utilities.py
@@ -48,7 +48,7 @@ from easybuild.main import main
 from easybuild.tools import config
 from easybuild.tools.config import module_classes, set_tmpdir
 from easybuild.tools.environment import modify_env
-from easybuild.tools.filetools import mkdir, read_file, write_file
+from easybuild.tools.filetools import mkdir, read_file
 from easybuild.tools.module_naming_scheme import GENERAL_CLASS
 from easybuild.tools.modules import modules_tool
 from easybuild.tools.options import CONFIG_ENV_VAR_PREFIX, EasyBuildOptions
@@ -152,9 +152,9 @@ class EnhancedTestCase(_EnhancedTestCase):
         # restore original Python search path
         sys.path = self.orig_sys_path
 
-        # cleanup: remove test directory, but restores log files (empty)
+        # cleanup: remove test directory, but restore log files (as empty files)
         # log rotation only kicks in when *all* log handles are out of scope (when this TestCase object is removed);
-        # log files should still be in place at that time
+        # log files must still be in place at that time
         try:
             logfiles = []
             for root, _, filenames in os.walk(self.test_prefix):
@@ -165,9 +165,12 @@ class EnhancedTestCase(_EnhancedTestCase):
             shutil.rmtree(self.test_prefix)
 
             for logfile in logfiles:
-                mkdir(os.path.dirname(logfile), parents=True)
-                write_file(logfile, '')
-        except OSError, err:
+                os.makedirs(os.path.dirname(logfile))
+                f = open(logfile, 'w')
+                f.write('')
+                f.close()
+                sys.stderr.write('Restored %s\n' % logfile)
+        except (OSError, IOError), err:
             pass
 
         # restore original 'parent' tmpdir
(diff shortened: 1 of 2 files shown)


boegel:
Not out of the woods yet:

```
  File "/var/lib/jenkins/.local/lib/python2.6/logging/handlers.py", line 77, in emit
    self.doRollover()
  File "/var/lib/jenkins/.local/lib/python2.6/logging/handlers.py", line 137, in doRollover
    self.stream = self._open()
  File "/var/lib/jenkins/.local/lib/python2.6/logging/__init__.py", line 846, in _open
    stream = open(self.baseFilename, self.mode)
IOError: [Errno 2] No such file or directory: '/var/tmp/eb-SFKPKF/eb-RIx5GT/eb-MZB8sc/eb-26IEHo/eb-cCFCzg/easybuild-toy-0.0-20150514.162546.AjYeE.log'
```

```
  File "/var/lib/jenkins/.local/lib/python2.6/logging/handlers.py", line 76, in emit
    if self.shouldRollover(record):
  File "/var/lib/jenkins/.local/lib/python2.6/logging/handlers.py", line 150, in shouldRollover
    self.stream.seek(0, 2)  #due to non-posix-compliant Windows feature
ValueError: I/O operation on closed file
```

boegel:
Hmm, problems mentioned above seemed to be caused by a known bug, see http://bugs.python.org/issue10949 (Jenkins is running Python 2.6).

boegel:
OK, found the actual culprit of the log files not being removed: the log handlers were not being cleaned up...

Simple fix, huge impact.

boegel:
with this patch: Ran 284 tests in 313.337s

without this patch: Ran 284 tests in 907.145s

wpoely86:
looks fine

test/framework/utilities.py
```diff
@@ -98,10 +102,6 @@ def setUp(self):
         # keep track of original environment/Python search path to restore
         self.orig_sys_path = sys.path[:]
 
-        self.orig_paths = {}
```
wpoely86: this can go because we restore the original env anyway, right?

boegel: yep, no need to treat this separately

pforai:
@boegel can you explain why this makes such a huge difference in performance?!

boegel:
@pforai: Not removing the log handlers means that they stay in place, and also keep getting used. Every time `main()` is called from the tests, one or two handlers are added.
If these don't get removed, they will be used every time a `log` method is called (even though the log file they are supposed to write to is already gone, or at least the inode pointing to the data in the file is).

Near the end of the test suite, over 100 handlers were still active, all of which were still writing to disk (even though the inode of the log file was already removed via `shutil.rmtree(self.test_prefix)` in `tearDown`).

Before this patch, about 3GB of disk space was required in `/tmp` because the log files of previous tests were still being written to. Now, about 100MB is more than enough...

That's actually was caused us diving into this: `/tmp` on the Jenkins server is quite limited (~6GB free space), and two framework test suites running at the same was enough to fill up `/tmp` and let all things go to hell. Quite annoying (since all subsequent tests, including PR tests, failed hard then), and I was (finally) fed up with it, which motivated me to get to the bottom of this. ;-)

Some other changes were also done in this PR, i.e. not reconfiguring in `tearDown`, only to then reconfigure again in `setUp` of the next step.  That helps too, but just a little bit.

boegel:
Time to merge this in, and forget about it.

Thanks for the review @wpoely86, and @stdweird for helping to figure this out (a bit ;-)).

stdweird:
@boegel nice!

PR #1283, opened 2015-05-15
define $CRAYPE_LINK_TYPE if 'dynamic' toolchain option is enabled

boegel:
@pforai: not sure what to do with the `-dynamic` flag being passed to the compiler; should we keep that in place?

And should we also define `$CRAYPE_LINK_TYPE` if the `shared` toolchain option is defined?

Does it still make sense to have both `dynamic` and `shared` as separate toolchain options?

### DIFF
--- a/easybuild/toolchains/compiler/craype.py
+++ b/easybuild/toolchains/compiler/craype.py
@@ -68,10 +68,9 @@ class CrayPECompiler(Compiler):
     }
 
     COMPILER_UNIQUE_OPTION_MAP = {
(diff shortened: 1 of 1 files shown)


pforai:
@boegel 
I would drop -shared and -dynamic  and just expose $CRAYPE_LINK_TYPE for now.

boegel:
And then remove the `dynamic` toolchain option entirely? If they want a dynamically linked build, they should enable `shared`, so `$CRAYPE_LINK_TYPE` is defined?

boegel:
Jenkins: test this please

boegel:
going in, thanks @pforai for testing/reviewing this!

PR #1286, opened 2015-05-18
use https for PyPI URLs

boegel:
downloading over http seems to be broken hard now?

@wpoely86: please review


wpoely86:
okay

boegel:
Tested with all affected easyconfigs, all is well.

PR #1288, opened 2015-05-21
Cray module swap for craype-<cpu-target> (WIP)

pforai:
Some refactoring and enhancement for auto swapping the craype-<cpu-target> module instead of relying on the modules tool to just bail on conflict.

Same logic will be used in follow-up commits for PrgEnv-<gnu/intel/cray> modules. As discussed with @boegel this will probably land before the call to super() in the prepare() method.


boegel:
@pforai: what's the status on this? I need to close down for 2.2 in the next couple of days, it seems like this will need more love and should be targeted to 2.3 (early Sept'15)?

boegel:
@pforai: I won't have the time required to give this a thorough look in time for EB v2.2, so I'm moving the milestone to v2.3 on this one;

let's rediscuss this in August?

pforai:
Rebased on newer develop. /cc @gppezzi can you give this a review?

easybuild/tools/modules.py
```diff
@@ -469,6 +479,19 @@ def get_value_from_modulefile(self, mod_name, regex):
         else:
             raise EasyBuildError("Can't get value from a non-existing module %s", mod_name)
 
+    def get_conflicts_from_modfile(self,mod_name):
+        conflicts_re = re.compile('^conflict\s*(\S+).*$',re.M)
```
boegel: does this also work for modules in Lua syntax?

please add unit tests for this...

easybuild/tools/modules.py
```diff
@@ -450,6 +450,16 @@ def show(self, mod_name):
         """
         return self.run_module('show', mod_name, return_output=True)
 
+    def swap(self, mod_name1, mod_name2):
+        """
+        Run 'module swap' for mod_name1 to mod_name2
+        :param mod_name1: The module that should be swapped
+        :param mod_name2: to the module that should be swapped with
+        :return: 0 if the module that should be swapped to is loaded after the swap
```
boegel: no `return` in the implementation?

easybuild/toolchains/compiler/craype.py
```diff
@@ -103,6 +103,31 @@ def __init__(self, *args, **kwargs):
         # copy unique option map, since we fiddle with it later
         self.COMPILER_UNIQUE_OPTION_MAP = copy.deepcopy(self.COMPILER_UNIQUE_OPTION_MAP)
 
+    # this is specific to the semantics of craype-modules. There are craype modules for steering the compiler's backend code gen,
+    # while others define run time behaviour like the use of hugepages.    
+    def load_or_swap_module(self, mod_name_prefix, mod_name):
+	self.log.debug("Using load or swap")
+        if self.modules_tool.exist([mod_name], skip_avail=True)[0]:
+            if any(mod.startswith(mod_name_prefix) for mod in self.modules_tool.loaded_modules()):
+                loaded_mods_same_prefix = [mod for mod in self.modules_tool.loaded_modules() if
```
boegel: shorter/cleaner:

```python
loaded_mods = self.modules_tool.loaded_modules()
if any(mod.startswith(mod_name_prefix) for mod in loaded_mods):
    loaded_mods_same_prefix = [mod for mod in loaded_mods if mod.startswith(mod_name_prefix)]

File: easybuild/toolchains/compiler/craype.py [hunk 4]
```diff
@@ -103,6 +103,31 @@ def __init__(self, *args, **kwargs):
         # copy unique option map, since we fiddle with it later
         self.COMPILER_UNIQUE_OPTION_MAP = copy.deepcopy(self.COMPILER_UNIQUE_OPTION_MAP)
 
+    # this is specific to the semantics of craype-modules. There are craype modules for steering the compiler's backend code gen,
+    # while others define run time behaviour like the use of hugepages.    
+    def load_or_swap_module(self, mod_name_prefix, mod_name):
+	self.log.debug("Using load or swap")
+        if self.modules_tool.exist([mod_name], skip_avail=True)[0]:
+            if any(mod.startswith(mod_name_prefix) for mod in self.modules_tool.loaded_modules()):
```
boegel: I don't think you need this check at all...

If no modules with the same prefix are loaded, `loaded_mods_same_prefix` is going to be an empty list below, and you're good?

So, OK to drop and fix the indent, imho.

easybuild/toolchains/compiler/craype.py
```diff
@@ -103,6 +103,31 @@ def __init__(self, *args, **kwargs):
...
+                for loaded_mod in loaded_mods_same_prefix:
+                    conflicts = self.modules_tool.get_conflicts_from_modfile(loaded_mod)
+                    if mod_name in conflicts:
+                        self.log.debug("List of relevant conflicts is %s" % conflicts)
+                        #special case for craype-<cpu-target> modules as they contain themselfs as conflicts
+                        if loaded_mod == mod_name:
+                            pass
+                        else:
+                            self.modules_tool.swap(loaded_mod, mod_name)
+                else:
```
boegel: you have tabs here somewhere... please remove *all* tabs, use 4 spaces instead

easybuild/toolchains/compiler/craype.py
```diff
@@ -103,6 +103,31 @@ def __init__(self, *args, **kwargs):
...
+                self.log.debug("Loaded modules of same prefix %s " % loaded_mods_same_prefix)
+                for loaded_mod in loaded_mods_same_prefix:
+                    conflicts = self.modules_tool.get_conflicts_from_modfile(loaded_mod)
+                    if mod_name in conflicts:
+                        self.log.debug("List of relevant conflicts is %s" % conflicts)
+                        #special case for craype-<cpu-target> modules as they contain themselfs as conflicts
+                        if loaded_mod == mod_name:
+                            pass
+                        else:
+                            self.modules_tool.swap(loaded_mod, mod_name)
```
boegel:
```python
if loaded_mod != mod_name:
    self.modules_tool.swap(loaded_mod, mod_name)
```

easybuild/toolchains/compiler/craype.py
```diff
@@ -103,6 +103,31 @@ def __init__(self, *args, **kwargs):
...
+    # this is specific to the semantics of craype-modules. There are craype modules for steering the compiler's backend code gen,
+    # while others define run time behaviour like the use of hugepages.    
+    def load_or_swap_module(self, mod_name_prefix, mod_name):
+	self.log.debug("Using load or swap")
+        if self.modules_tool.exist([mod_name], skip_avail=True)[0]:
+            if any(mod.startswith(mod_name_prefix) for mod in self.modules_tool.loaded_modules()):
+                loaded_mods_same_prefix = [mod for mod in self.modules_tool.loaded_modules() if
+                                      mod.startswith(mod_name_prefix)]
+                self.log.debug("Loaded modules of same prefix %s " % loaded_mods_same_prefix)
+                for loaded_mod in loaded_mods_same_prefix:
```
boegel: the `for` loop doesn't make sense, you can only `swap` once...

so, I think you need to check how many loaded modules with the same prefix there are; if there's more than one, something is very wrong...

boegel:
@pforai: is this still relevant?

cc @gppezzi

boegel:
@pforai, @gppezzi ping on this? Is it worth reviving this?

pforai:
I think it would make sense to revive as it is itsn't too hard to do and offers some convenience.

boegel:
@pforai OK, let's start with syncing this with current `develop`: https://github.com/pforai/easybuild-framework/pull/41

PR #1253, opened 2015-04-02
enable detection of use of unknown $EASYBUILD-prefixed env vars

boegel:
depends on https://github.com/hpcugent/vsc-base/pull/162 (with https://github.com/stdweird/vsc-base/pull/4 included)

PR #1289, opened 2015-05-27
exclude 'easyblocks' pkg from sys.path, make sure $EASYBUILD_BOOTSTRAP* env vars are undefined via pop

boegel:
The `setup.py` of easybuild-easyblocks import directly from the `easyblocks` package (see below), so any paths providing this namespace should also be excluded from the Python search path prior to initiating the bootstrap procedure.

```python
sys.path.append('easybuild')                                                                                            
from easyblocks import VERSION  
```

The `$EASYBUILD_BOOTSTRAP*` environment variables must be unset, since they are not valid EasyBuild configuration options, which EasyBuild trips over since v2.1.0 (see #1253).


berndmohr:
@boegel I can confirm that the script fixes the issues I had with bootstrapping 2.1.1 on my laptop

wpoely86:
looks fine

boegel:
Thanks for the review @wpoely86, and @berndmohr for confirming the fix!

Going in...

PR #1294, opened 2015-06-05
reset keep toolchain component class 'constants' every time

boegel:
This fixes a problem that often occurs when testing large easyconfig PRs that contain easyconfigs with a large variety on toolchains using `--from-pr`, for example https://github.com/hpcugent/easybuild-easyconfigs/pull/1337#issuecomment-108479308 .

If old and new version of Intel compilers are used in the same session, things break because constants aren't exactly constant.

This fixes the problem; I took care of checking the other toolchain components too.

@wpoely86, @pforai, @stdweird: please review?

### DIFF
--- a/easybuild/toolchains/compiler/inteliccifort.py
+++ b/easybuild/toolchains/compiler/inteliccifort.py
@@ -91,7 +91,7 @@ class IntelIccIfort(Compiler):
 
     def __init__(self, *args, **kwargs):
         """Toolchain constructor."""
-        self.CLASS_CONSTANTS_TO_RESTORE.append('LIB_MULTITHREAD')
+        self.add_class_constants_to_restore(['LIB_MULTITHREAD'])
         super(IntelIccIfort, self).__init__(*args, **kwargs)
 
     def _set_compiler_vars(self):
--- a/easybuild/toolchains/compiler/inteliccifort.py
+++ b/easybuild/toolchains/compiler/inteliccifort.py
@@ -91,7 +91,8 @@ class IntelIccIfort(Compiler):
 
     def __init__(self, *args, **kwargs):
         """Toolchain constructor."""
-        self.add_class_constants_to_restore(['LIB_MULTITHREAD'])
+        class_constants = kwargs.setdefault('class_constants', [])
+        class_constants.append('LIB_MULTITHREAD')
         super(IntelIccIfort, self).__init__(*args, **kwargs)
 
     def _set_compiler_vars(self):
--- a/easybuild/toolchains/compiler/inteliccifort.py
+++ b/easybuild/toolchains/compiler/inteliccifort.py
@@ -88,8 +88,11 @@ class IntelIccIfort(Compiler):
     }
 
     LIB_MULTITHREAD = ['iomp5', 'pthread']  # iomp5 is OpenMP related
-    # keep track of original value, needs to be restored every time since we append to class 'constant' LIB_MULTITHREAD
-    _INIT_LIB_MULTITHREAD = LIB_MULTITHREAD[:]
+
+    def __init__(self, *args, **kwargs):
+        """Toolchain constructor."""
+        self.CLASS_CONSTANTS_TO_RESTORE.append('LIB_MULTITHREAD')
+        super(IntelIccIfort, self).__init__(*args, **kwargs)
 
     def _set_compiler_vars(self):
         """Intel compilers-specific adjustments after setting compiler variables."""
@@ -106,9 +109,6 @@ class IntelIccIfort(Compiler):
             raise EasyBuildError("_set_compiler_vars: mismatch between icc version %s and ifort version %s",
                                  icc_version, ifort_version)
 
-        # reset LIB_MULTITHREAD every time,
-        # to avoid problems when multiple Intel compilers versions are used in a single session
-        self.LIB_MULTITHREAD = self._INIT_LIB_MULTITHREAD[:]
         if LooseVersion(icc_version) < LooseVersion('2011'):
             self.LIB_MULTITHREAD.insert(1, "guide")
 
--- a/easybuild/toolchains/compiler/inteliccifort.py
+++ b/easybuild/toolchains/compiler/inteliccifort.py
@@ -87,7 +87,9 @@ class IntelIccIfort(Compiler):
         'dynamic':'-Bdynamic',
     }
 
-    LIB_MULTITHREAD = None
+    LIB_MULTITHREAD = ['iomp5', 'pthread']  # iomp5 is OpenMP related
+    # keep track of original value, needs to be restored every time since we append to class 'constant' LIB_MULTITHREAD
+    _INIT_LIB_MULTITHREAD = LIB_MULTITHREAD[:]
 
     def _set_compiler_vars(self):
         """Intel compilers-specific adjustments after setting compiler variables."""
@@ -106,7 +108,7 @@ class IntelIccIfort(Compiler):
 
         # reset LIB_MULTITHREAD every time,
         # to avoid problems when multiple Intel compilers versions are used in a single session
-        self.LIB_MULTITHREAD = ['iomp5', 'pthread']  # iomp5 is OpenMP related
+        self.LIB_MULTITHREAD = self._INIT_LIB_MULTITHREAD[:]
         if LooseVersion(icc_version) < LooseVersion('2011'):
             self.LIB_MULTITHREAD.insert(1, "guide")
 
--- a/easybuild/toolchains/compiler/inteliccifort.py
+++ b/easybuild/toolchains/compiler/inteliccifort.py
@@ -87,7 +87,7 @@ class IntelIccIfort(Compiler):
         'dynamic':'-Bdynamic',
     }
 
-    LIB_MULTITHREAD = ['iomp5', 'pthread']  ## iomp5 is OpenMP related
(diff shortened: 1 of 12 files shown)


easybuild/toolchains/compiler/inteliccifort.py
```diff
@@ -104,6 +104,9 @@ def _set_compiler_vars(self):
             raise EasyBuildError("_set_compiler_vars: mismatch between icc version %s and ifort version %s",
                                  icc_version, ifort_version)
 
+        # reset LIB_MULTITHREAD every time,
+        # to avoid problems when multiple Intel compilers versions are used in a single session
+        self.LIB_MULTITHREAD = ['iomp5', 'pthread']  # iomp5 is OpenMP related
```
stdweird: hmmmm, this is not inheritance-safe. need to think about this one

stdweird: maybe
```python
if self._init_LIB_MULTITHREAD is None:
    self._init_LIB_MULTITHREAD = self.LIB_MULTITHREAD[:]
else:
    self.LIB_MULTITHREAD = self._init_LIB_MULTITHREAD[:]
```

stdweird: and add `self._init_... = None` to the `__init__`? (or use a dict to store all such non-constant class constants)

wpoely86: On another note, why do we manually add `libiomp5`? Using `-fopenmp` should take care of this?

boegel: should, we had a very good reason at some point for this, and I'm not keen on breaking stuff that works just to remove a handful of characters ;)

stdweird: these are libraries you can pass via `LDFLAGS`, doesn't always work with compiler options. there's probably some crappy code out there that wants library names.

ideally, EB should figure out what libraries `-fopenmp` links in and use those instead of hardcoding.

boegel: @stdweird: good point on inheritance

How about simply this in `__init__.py`:

```python
self._init_LIB_MULTITHREAD = self.LIB_MULTITHREAD[:]
```

and this here:

```python
self.LIB_MULTITHREAD = self._init_LIB_MULTITHREAD[:]
```

That should work, no?

Toolchain instances are recreated for different versions.

stdweird: then maybe also make 2 private methods `_copy_some_class_constants` and `_restore_some_class_constants` and call/extend those.

boegel: I wouldn't generalise this, it should be considered an exception, not encouraged.

boegel:
@stdweird: fixed the inheritance issue you pointed out, please rereview?

easybuild/toolchains/compiler/inteliccifort.py
```diff
@@ -87,7 +87,9 @@ class IntelIccIfort(Compiler):
         'dynamic':'-Bdynamic',
     }
 
-    LIB_MULTITHREAD = ['iomp5', 'pthread']  ## iomp5 is OpenMP related
+    LIB_MULTITHREAD = ['iomp5', 'pthread']  # iomp5 is OpenMP related
+    # keep track of original value, needs to be restored every time since we append to class 'constant' LIB_MULTITHREAD
```
stdweird: no, please hide this in the `__init__`, people will start to manipultae this

boegel:
@stdweird: central solution implemented, unit test enhanced to make sure things work as expected when multiple toolchains and toolchain versions are in play

easybuild/toolchains/linalg/intelmkl.py
```diff
@@ -28,7 +28,7 @@
 @author: Stijn De Weirdt (Ghent University)
 @author: Kenneth Hoste (Ghent University)
 """
-
+import copy
```
stdweird: not needed?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -95,6 +99,10 @@ def __init__(self, name=None, version=None, mns=None):
 
         self.vars = None
 
+        self.add_class_constants_to_restore([])  # make sure self.CLASS_CONSTANTS_TO_RESTORE is initialised
```
stdweird: this is a bit odd, no? why is the default not `[]`?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -108,6 +116,13 @@ def __init__(self, name=None, version=None, mns=None):
                 self.mod_short_name = self.mns.det_short_module_name(tc_dict)
                 self.init_modpaths = self.mns.det_init_modulepaths(tc_dict)
 
+    def add_class_constants_to_restore(self, names):
+        """Add given constants to list of class constants to restore with each new instance."""
+        if self.CLASS_CONSTANTS_TO_RESTORE is None:
```
stdweird: i don't understand the benefit of None. initial `CLASS_CONSTANTS_TO_RESTORE` could be `[]`, and you wouldn't need a method at all.

easybuild/tools/toolchain/toolchain.py
boegel: Defining a class constant as a value of a mutable type is a problem, especially if you update it (see add_class_constants_to_restore).

I need to make sure a new list gets created for every toolchain class (`Ictce` vs `Goalf` vs `Intel`, etc.).

That doesn't happen if you set it to `[]` in `Toolchain` (outside of `__init__`).

boegel: This was the reason the previous test by Jenkins failed btw (see last commit).

easybuild/toolchains/linalg/intelmkl.py
boegel: will remove, thx

easybuild/tools/toolchain/toolchain.py
boegel: no, it can't, that's what I had initially... we update `self.CLASS_CONSTANTS_TO_RESTORE`, potentially with many different toolchains in play, so we need a new list instance for each toolchain (not each instance, but each toolchain class, e.g. `Ictce`, `Goolf`, `Intel`, `Foss`, etc.)

easybuild/tools/toolchain/toolchain.py
stdweird: ofcourse.

easybuild/tools/toolchain/toolchain.py
stdweird: ok

easybuild/tools/toolchain/toolchain.py
```diff
@@ -95,6 +99,10 @@ def __init__(self, name=None, version=None, mns=None):
 
         self.vars = None
 
+        self.add_class_constants_to_restore([])  # make sure self.CLASS_CONSTANTS_TO_RESTORE is initialised
+        self._copy_class_constants()
+        self._restore_class_constants()
```
stdweird: make a new method 
```
def _init_class_constants(self):
  self.add_class_constants_to_restore([])
  self._copy_class_constants()  
  self._restore_class_constants()
``` 

because this looks so strange otherwise

easybuild/toolchains/linalg/intelmkl.py
```diff
@@ -69,11 +68,16 @@ class IntelMKL(LinAlg):
...
     SCALAPACK_LIB_MT = ["mkl_scalapack%(lp64_sc)s"]
-    SCALAPACK_LIB_MAP = {"lp64_sc":"_lp64"}
+    SCALAPACK_LIB_MAP = {'lp64_sc': '_lp64'}
     SCALAPACK_REQUIRES = ['LIBBLACS', 'LIBBLAS']
     SCALAPACK_LIB_GROUP = True
     SCALAPACK_LIB_STATIC = True
 
+    def __init__(self, *args, **kwargs):
+        """Toolchain constructor."""
+        self.add_class_constants_to_restore(['BLAS_LIB_MAP', 'SCALAPACK_LIB', 'SCALAPACK_LIB_MT', 'SCALAPACK_LIB_MAP'])
```
stdweird: support this via the init kwargs. this just looks too weird.
```python
    csts = kwargs.setdefault('class_constants', [])
    csts.extend(['BLAS_LIB_MAP', 'SCALAPACK_LIB', 'SCALAPACK_LIB_MT', 'SCALAPACK_LIB_MAP'])
    super(IntelMKL, self).__init__(*args, **kwargs)
```

stdweird:
although the code looks good, it just feel weird looking at it. but my alternative with the `__init__` isn't much better :smile: 
so feel free to modify or merge

boegel:
The last test report for https://github.com/hpcugent/easybuild-easyconfigs/pull/1337, available at https://github.com/hpcugent/easybuild-easyconfigs/pull/1337#issuecomment-109446243, still shows problems, so I need to look at it more closely, and see why this change isn't fixing the issues there...

boegel:
retested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1337, works as intended and fixes the problem it intends to fix (previous failing test was due to issues with Intel license server, not due to EB itself)

Going in, thanks for the review @stdweird!

PR #1295, opened 2015-06-05
make --strict also a build option

boegel:
@wpoely86: please review

Useful for https://github.com/hpcugent/easybuild-easyblocks/pull/622

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -44,6 +44,7 @@ from vsc.utils.missing import FrozenDictKnownKeys
 from vsc.utils.patterns import Singleton
 
 import easybuild.tools.environment as env
+from easybuild.tools import run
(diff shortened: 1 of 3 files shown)


wpoely86:
I'm probably missing something here but what exactly does this change?

boegel:
@wpoely86: it enables using `build_option('strict')` without running into a `KeyError`

easybuild/tools/config.py
```diff
@@ -119,6 +121,9 @@ def mk_full_default_path(name, prefix=DEFAULT_PREFIX):
     True: [
         'cleanup_builddir',
     ],
+    DEFAULT_STRICT: [
+        'strict',
+    ],
```
boegel: @wpoely86: key change is here ^

boegel:

PR #1296, opened 2015-06-12
Question answer

npklein:
Part of adding question/answer  functionallity during command line installation to install Anaconda

### DIFF
--- a/easybuild/framework/easyconfig/default.py
+++ b/easybuild/framework/easyconfig/default.py
@@ -115,7 +115,8 @@ DEFAULT_CONFIG = {
     'versionprefix': ['', ('Additional prefix for software version '
                            '(placed before version and toolchain name)'), BUILD],
     'versionsuffix': ['', 'Additional suffix for software version (placed after toolchain name)', BUILD],
-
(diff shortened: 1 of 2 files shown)


boegel:
@npklein: I'm not sure this is the right location for this (the `qa` variable you're adding), since you need a way to tie this to a particular call to `run_cmd(_qa)`.
What you really need, I think, is to add a custom easyconfig parameter for the Binary easyblock (in `extra_options`, see https://github.com/hpcugent/easybuild-easyblocks/blob/master/easybuild/easyblocks/generic/binary.py#L53), and then use that value in the `run_cmd` calling in `Binary.install_step` (at https://github.com/hpcugent/easybuild-easyblocks/blob/master/easybuild/easyblocks/generic/binary.py#L109).

It's also not clear what impact the changes have to `run_cmd_qa`, we need to be very careful here not to break anything that works...

boegel:
Jenkins: ok to test

boegel:
@npklein: ping on this?

npklein:
@boegel did not manage to get this finished

boegel:
@npklein: I'm just wondering whether it makes sense at all, see remarks above

PR #1299, opened 2015-06-15
stop making ModulesTool class a singleton

boegel:
With the `ModulesTool` class being a singleton, things go horribly wrong when builds are performed with different toolchains in a single session, when an alternative module naming scheme like `HierarchicalMNS` is used. This issue was reported by @geimer.

The problem is that the `ModulesTool` class remember the `$MODULEPATH` values (via the `mod_paths` class variable), which is not compatible with the concept of a 'dynamic' `$MODULEPATH` that is required when using a hierarchical module naming scheme.

Disabling the singleton property for `ModulesTool` should only have a (very minor) impact on performance, and has to adverse effects on functionality.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1029,8 +1029,7 @@ class EasyBlock(object):
         fake_mod_path = self.make_module_step(True)
 
         # load fake module
-        modtool = modules_tool()
-        modtool.prepend_module_path(fake_mod_path)
+        self.modules_tool.prepend_module_path(fake_mod_path)
         self.load_module(purge=purge)
 
         return (fake_mod_path, env)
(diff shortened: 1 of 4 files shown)


wpoely86:
looks fine, I think?

How does this solve the issue you mention?

geimer:
The aforementioned issue showed up when trying to generate a partial (only GNU-based toolchains) test report for https://github.com/hpcugent/easybuild-easyconfigs/pull/1690. It consistently failed for one build, where `$MODULEPATH` was messed up.

Using the changes from this PR, I was able to successfully build all 15 GNU-based libxml2 configurations, i.e., it seems to solve the issue.

boegel:
@wpoely86: the `ModulesTool` class is now no longer a singleton, i.e. multiple instances are created, thus state is not kept in between builds

boegel:
I need to look into a unit test dedicated to the problem that is resolved by this change...

boegel:
unit test is there, not holding this back any further

Thanks for the feedback @geimer and @wpoely86!

PR #1300, opened 2015-06-16
don't modify values of 'paths' list passed as argument to prepend_paths

boegel:
I ran into this issue when using the updated `PythonPackage` easyblock from https://github.com/hpcugent/easybuild-easyblocks/pull/628....

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -226,18 +226,21 @@ class ModuleGeneratorTcl(ModuleGenerator):
             self.log.debug("Wrapping %s into a list before using it to prepend path %s" % (paths, key))
             paths = [paths]
 
-        for i, path in enumerate(paths):
+        abspaths = []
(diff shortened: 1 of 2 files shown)


easybuild/tools/module_generator.py
```diff
@@ -226,18 +226,21 @@ def prepend_paths(self, key, paths, allow_abs=False):
             self.log.debug("Wrapping %s into a list before using it to prepend path %s" % (paths, key))
             paths = [paths]
 
-        for i, path in enumerate(paths):
```
wpoely86: so, we were changing the list while iterating over it? :-1:

boegel: we were updating elements in a list while iterating over it, which is OK

What's not OK is that we're modifying the list that is being passed to this method...

boegel:
Going in, thanks @wpoely86 for the review!

PR #1301, opened 2015-06-18
add support for --include-* (REVIEW)

boegel:
Add support to easily include additional custom easyblocks, module naming schemes and toolchains/toolchain components, without having to fiddle with `$PYTHONPATH` or Python packaging magic (`__init__.py`), etc.

TODO list:
* [x] `--include-easyblocks`
* [x] `--include-module-naming-schemes`
* [x] `--include-toolchains`
* [x] unit tests
  * [x] `include_*` functions
  * [x] `--include-*` options combined with `--list-toolchains`, `--list-easyblocks`, `--avail-module-naming-schemes`
* [x] updated documentation (see https://github.com/hpcugent/easybuild/pull/129)

note: requires https://github.com/hpcugent/easybuild-easyblocks/pull/617 in order to work correctly


boegel:
Currently only the `--include-easyblocks` part works:

```
$ eb --list-easyblocks | grep foo

$ eb --include-easyblocks=$PWD/foo.py,$PWD/generic/foobar.py --list-easyblocks | grep foo
|-- EB_foo
|-- foobar
```

If also enhanced the output of `--list-easyblocks=detailed`:

```
$ eb --include-easyblocks=$PWD/foo.py,$PWD/generic/foobar.py --list-easyblocks=detailed | grep foo
|-- EB_foo (easybuild.easyblocks.foo @ /var/folders/8s/_frgh9sj6m744mxt5w5lyztr0000gn/T/eb-yFeTNf/included-easyblocks/easybuild/easyblocks/foo.py)
|-- foobar (easybuild.easyblocks.generic.foobar @ /var/folders/8s/_frgh9sj6m744mxt5w5lyztr0000gn/T/eb-yFeTNf/included-easyblocks/easybuild/easyblocks/generic/foobar.py)
```

easybuild/main.py
```diff
@@ -163,9 +163,6 @@ def main(testing_data=(None, None, None)):
         new_umask = int(options.umask, 8)
         old_umask = os.umask(new_umask)
 
-    # set temporary directory to use
-    eb_tmpdir = set_tmpdir(options.tmpdir)
```
wpoely86: why this change?

boegel:
When combined with https://github.com/hpcugent/easybuild-easyblocks/pull/617, this also reflects easyblocks being overridden:

```
$ eb --include-easyblocks=$PWD/foo.py,$PWD/gcc.py,$PWD/generic/\*.py --list-easyblocks=detailed | grep -C 2 GCC
|   |-- EB_g2clib (easybuild.easyblocks.g2clib @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/g2clib.py)
|   |-- EB_g2lib (easybuild.easyblocks.g2lib @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/g2lib.py)
|   |-- EB_GCC (easybuild.easyblocks.gcc @ /var/folders/8s/_frgh9sj6m744mxt5w5lyztr0000gn/T/eb-5Zqzj1/included-easyblocks/easybuild/easyblocks/gcc.py)
|   |-- EB_GHC (easybuild.easyblocks.ghc @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/ghc.py)
|   |-- EB_Go (easybuild.easyblocks.go @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/go.py)
[kehoste:easybuild-framework] include+ $ eb --list-easyblocks=detailed | grep GCC 
|   |-- EB_GCC (easybuild.easyblocks.gcc @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/gcc.py)

$ eb --list-easyblocks=detailed | grep -C 2 GCC
|   |-- EB_g2clib (easybuild.easyblocks.g2clib @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/g2clib.py)
|   |-- EB_g2lib (easybuild.easyblocks.g2lib @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/g2lib.py)
|   |-- EB_GCC (easybuild.easyblocks.gcc @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/gcc.py)
|   |-- EB_GHC (easybuild.easyblocks.ghc @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/ghc.py)
|   |-- EB_Go (easybuild.easyblocks.go @ /Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/g/go.py)
```

easybuild/main.py
boegel: I need a temporary directory to symlink the included easyblocks into the correct namespace.

When `--include-easyblocks` is combined with `--list-easyblocks`, this point is never reached (`--list-easyblocks` is handled via `tools/options.py`, and so we don't return back to `main.py` after parsing options).

I now set the temporary directory as a part of postprocessing the parsed options, so earlier than I was here. Should be fine (can't hurt to do it earlier).

easybuild/tools/include.py
```diff
@@ -0,0 +1,113 @@
...
+
+_log = fancylogger.getLogger('tools.include', fname=False)
+
+
+PKG_INIT_BODY = """
+from pkgutil import extend_path
+__path__ = extend_path(__path__, __name__)
+"""
+
+
```
fgeorgatos: any double space to trim above, all is intentional?

boegel: we separate constants from functions (and top-level functions between each other) with a double newline

fgeorgatos:
first visual review ok. very interesting PR, I am sure it will get quickly popular.

I hope the lack of `--include-module-naming-schemes`, `--include-toolchains` will not delay merging it in before 2.2.0 (if so, just turn the latter two into experimental facilities).

boegel:
@fgeorgatos: the idea is to get some early feedback, finishing the other two options shouldn't be that hard, with what's in place already (it pretty analogous)

boegel:
implementation finished, first stab at dedicated unit tests for `include_*` functions included

should also add tests for `--include-*` options combined with `--list-toolchains`, `--list-easyblocks`, `--avail-module-naming-schemes`

easybuild/tools/include.py
```diff
@@ -0,0 +1,210 @@
...
+        # note: can't use mkdir, since that required build options to be initialised
+        if not os.path.exists(path):
+            os.makedirs(path)
+
+        # put __init__.py files in place, with required pkgutil.extend_path statement
+        # note: can't use write_file, since that required build options to be initialised
+        handle = open(init_path, 'w')
+        handle.write(PKG_INIT_BODY)
+        handle.close()
+    except (IOError, OSError) as err:
```
JensTimmerman: this is fine, 
but it's sometimes easier to have
`with open(newfile, 'w') as handle:`

easybuild/tools/include.py
```diff
@@ -0,0 +1,210 @@
...
+        for subpkg in subpkgs:
+            create_pkg(os.path.join(pkgpath, subpkg))
+
+    # creata package dirs on each level
+    while pkgpath != parent_path:
+        create_pkg(pkgpath)
+        pkgpath = os.path.dirname(pkgpath)
+
+
+def expand_glob_paths(glob_paths):
```
JensTimmerman: does this belong here?

easybuild/tools/include.py
```diff
@@ -0,0 +1,210 @@
...
+    """Expand specified glob paths to a list of unique non-glob paths to only files."""
+    paths = []
+    for glob_path in glob_paths:
+        paths.extend([f for f in glob.glob(glob_path) if os.path.isfile(f)])
+
+    return nub(paths)
+
+
+def safe_symlink(source_path, symlink_path):
```
JensTimmerman: does this belong here, in this module?

easybuild/tools/include.py
```diff
@@ -0,0 +1,210 @@
...
+    """Create a symlink at the specified path for the given path."""
+    try:
+        os.symlink(os.path.abspath(source_path), symlink_path)
+        _log.info("Symlinked %s to %s", source_path, symlink_path)
+    except OSError as err:
+        raise EasyBuildError("Symlinking %s to %s failed: %s", source_path, symlink_path, err)
```
JensTimmerman: I don't get what this function is supposed to do?
I'd put in the docstring something like: `Creates a symlink but hides the OSError that might occur  and masks it with a more general EasyBuildError. `

easybuild/tools/options.py
```diff
@@ -568,6 +588,13 @@ def _postprocess_list_avail(self):
             self.log.info(msg)
         else:
             print msg
+
+        # cleanup tmpdir
+        try:
+            shutil.rmtree(self.options.tmpdir)
+        except OSError as err:
+            raise EasyBuildError("Failed to clean up temporary directory %s: %s", self.options.tmpdir, err)
```
JensTimmerman: is it really needed to raise an error here? Can't we just emit a warning and continue?

easybuild/tools/utilities.py
```diff
@@ -99,5 +102,9 @@ def import_available_modules(namespace):
                 mod_name = module.split(os.path.sep)[-1].split('.')[0]
                 modpath = '.'.join([namespace, mod_name])
                 _log.debug("importing module %s" % modpath)
-                modules.append(__import__(modpath, globals(), locals(), ['']))
+                try:
+                    mod = __import__(modpath, globals(), locals(), [''])
+                except ImportError as err:
```
JensTimmerman: why do you want to mask this ImportError?
Isn't it better to just rais the ImportError, so the function that uses this method can try to handle it?
Now it has to Catch an EasyBuildError, and it doesn't know what that means, anything could have gone wrong, since EasyBuildErrors are raised all over the place, so it can't try to recover from the error but is forced to quit?

easybuild/tools/include.py
boegel: I don't want to get a traceback for something like this, just a clear error suffices. Tracebacks confuse regular users, and have no extra value in this case (symlinking would only fail here in case of some system error).

easybuild/tools/options.py
boegel: If this goes wrong, something is very wrong, you'd want to know (a warning would only pop up in the log, so it's very likely you'll miss it).

easybuild/tools/utilities.py
boegel: I'm trying to avoid that i) I get an ugly traceback, ii) that I have to use `except ImportError` all over the place. If this goes wrong, something is very wrong (and it's not an EasyBuild bug), and we're done, so we throw up an EasyBuildError that will be catched higher up (to avoid a traceback, and only print the error message and exit).

easybuild/tools/include.py
boegel: I'll move this to `filetools.py`

easybuild/tools/include.py
boegel: will move to `filetools.py`

easybuild/tools/include.py
boegel: yeah, I keep forgetting we're no longer stuck to Python 2.4 ;)

easybuild/tools/include.py
JensTimmerman: But you are writing a library here, the 'user' you're talking about is another programmer writing a library or a binary.
A programmer won't be confused by a stacktrace, he will actually want one, so he can debug things.

When you write your binary you should check exceptions, and give meaningfull messages to users, but masking exceptions in a library just confuses other programmers.

easybuild/tools/options.py
JensTimmerman: so don't catch the exception? it will bubble up to the script/gui calling your function here, and that script/gui can then inform the user about the problem, and quit what it's doing.
And maybe that script can handle certain errors, but not others, but now you're masking OSErrors and other errors alike to EasyBuildErrors, so it's hard for the calling script to distinguish between them, and to know what to do.

easybuild/tools/utilities.py
JensTimmerman: Not all over the place, just on the very top, in your binary, where you can handle things. and print output to the user.
if it goes wrong, it will be very wrong, and bubble up all to the top.

If somewhere in that path up to the top you have a `except EasyBuildError`, which can happen, because it's starting to look like everything that can fail starts to raise EasyBuildErrors, so if I call a function and can recover from certain failures I have to catch EasyBuildError, it won't stop and continue in a weird state.

And how do you know that something is very wrong and you're done? You're writing a library here, you have no idea of where it will end up being used.
If you want it to be generic and reusable, then you don't want to mask errors, and you can't assume every error is fatal.

if you're not catching this error, it will be raised aswell, so things will stop, no need to catch it and mask it with another error?

boegel:
@JensTimmerman: please rereview

easybuild/tools/utilities.py
boegel: The general (unwritten) rule-of-thumb in EasyBuild is that errors that may be expected (e.g., caused by system issues) don't yield a traceback. Problems that do yield a traceback point to a bug in EasyBuild.

The functions we're discussing here are not intended to be of general use (at least not when including here), and are considered to be fatal when a problem occur caused by an external issue (e.g. filesystem, etc.).

For those type of issues, I want the user to get a clear error message (what was tried, what went wrong), not a full traceback (that's simply too daunting for end-users).

So: this idiom is applied all over the EB codebase.

If we want this changed, it should be i) discussed in a separate issue, ii) handled via a separate PR. I'm just sticking to the way of doing things here.

JensTimmerman: Ok if they're not for general use, but you might want to write more general functions and include them in vsc-base, so everybody benefits ;-)

JensTimmerman:
ok

boegel:
OK, I'm calling this, merging to be included in EB v2.2.0.

All TODO's are ticked off, extensive unit tests are in place, and a documentation update is ready to go as well (see https://github.com/hpcugent/easybuild/pull/129).

Thanks for the feedback on this @fgeorgatos, @JensTimmerman, @wpoely86 and @rjeschmi!

PR #1304, opened 2015-06-26
add --read-only-installdir and --group-writable-installdir configuration options

### LINKED ISSUE #1303: Group write permissions for shared easybuild installation
We'd like to share the easybuild installation directory with write permissions at group level: everyone from the group should be able to install, remove  and update existing apps & modules.

I managed to configure that (thanks to Alan's @ JSC example), by using --set-umask=002 & --set-gid-bit, however at some point EB decides to remove the group permissions:

https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/framework/easyblock.py#L1533

As a workaround I commented that line out on my EB and it seems to work as I expect, so I'm opening an issue to see if we can find a better solution (not sure if the problem comes from my permissions).

boegel:
fixes #1303


ocaisa:
I looked at this and the code looks ok to me, haven't tested it though...

boegel:
@JensTimmerman (or maybe @wpoely86): quick review?

@ocaisa: I want to look into adding a simply test for verifying that this does what it's supposed to before I merge this in.

easybuild/tools/config.py
```diff
@@ -448,16 +450,6 @@ def get_log_filename(name, version, add_salt=False):
     return filepath
 
 
-def read_only_installdir():
-    """
-    Return whether installation dir should be fully read-only after installation.
-    """
```
JensTimmerman: why doesn't this function return the result of the config option?

boegel: it could, but we already have the `build_option` function that serves this purpose, this placeholder function predates `build_option`

JensTimmerman: but this breaks the api?
Shouldn't you deprecate this function, and let it return the result of the config build option?

boegel: I could, but this is not considered to be part of the 'external' API, i.e. what easyblocks should be using. So it's OK to drop this.

JensTimmerman:
ok

rjeschmi:
looks fine

boegel:
Going in, thanks for the review @JensTimmerman and @rjeschmi!

Issue #1305, opened 2015-06-26
Easyconfig Dump method not robust enough


ocaisa:
The [easyconfig dump() method](https://github.com/hpcugent/easybuild-framework/blob/d1387c5847ef3e6d4ffe4c3fba2a0eb8e02ebb46/easybuild/framework/easyconfig/easyconfig.py#L475) is not robust enough to handle easyconfigs which include python code (such as for [libint](https://github.com/hpcugent/easybuild-easyconfigs/blob/e2de9e479830843a2644c08189edadb039f42d5e/easybuild/easyconfigs/l/Libint/Libint-1.1.4-intel-2015a.eb))

boegel:
Any error traceback to share?

ocaisa:
```
  File "/home/alanc/EasyBuild_Git/easybuild-framework/easybuild/framework/easyconfig/easyconfig.py", line 508, in dump
    ebtxt.append("%s = %s" % (key1, to_str(val)))
  File "/home/alanc/EasyBuild_Git/easybuild-framework/easybuild/framework/easyconfig/easyconfig.py", line 485, in to_str
    return "%s" % x
TypeError: not all arguments converted during string formatting```

**Comment by boegel** [comment]
@caylo: please take a look at this

**Comment by boegel** [comment]
@caylo: to make sure `bump` works, and keeps working, after fixing the issue reported here, we should consider including a check for it in https://github.com/hpcugent/easybuild-easyconfigs/blob/master/test/easyconfigs/easyconfigs.py

**Comment by boegel** [comment]
@caylo: to reproduce:

```
$ python
Python 2.7.6 (default, Sep  9 2014, 15:04:36) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from easybuild.framework.easyconfig.easyconfig import EasyConfig
>>> from test.framework.utilities import init_config
Couldn't import dot_parser, loading of dot files will not be possible.
>>> cfg = init_config()
>>> ec = EasyConfig('/Users/kehoste/work/easybuild-easyconfigs/easybuild/easyconfigs/l/Libint/Libint-1.1.4-goolf-1.4.10.eb')
>>> ec.dump('test.eb')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "easybuild/framework/easyconfig/easyconfig.py", line 508, in dump
[...]

Issue #1307, opened 2015-06-29
`dump()` method in EasyConfig class should be cleaned up and enhanced to produce pretty-printed result


boegel:
for @caylo

see https://github.com/hpcugent/easybuild-framework/blob/d1387c5847ef3e6d4ffe4c3fba2a0eb8e02ebb46/easybuild/framework/easyconfig/easyconfig.py#L469

* [x] replace internal function `to_str` with `quote_str` which is available in `easybuild.tools.utilities` (after checking they produce the same results) (see ~~https://github.com/hpcugent/easybuild-framework/pull/1308~~)
* [x] add unit tests for `dump()` method in `test/framework/easyconfig.py` (using some of the test easyconfigs in `test/framework/easyconfigs`) (see ~~https://github.com/hpcugent/easybuild-framework/pull/1308~~)

* [x] test `dump()` method on every easyconfig file via unit tests in easybuild-easyconfigs (see ~~https://github.com/hpcugent/easybuild-easyconfigs/pull/1761~~)

* [x] make sure templated values are dumped unresolved (`"%(version)s"` should remain exactly that, rather than "1.2.3" for example) (by setting `self.enable_templating = True` when `dump()` starts, and restoring the original value of `self.enable_templating` before `dump()` completes) (see ~~https://github.com/hpcugent/easybuild-framework/pull/1314~~)
* [x] check whether certain values can be replaced with a constant, e.g. replace `"%(name)s-%(version)s.tar.gz"` with `SOURCE_TAR_GZ`, see also http://easybuild.readthedocs.org/en/latest/Writing_easyconfig_files.html#dynamic-values-for-easyconfig-parameters (see ~~https://github.com/hpcugent/easybuild-framework/pull/1314~~)
  * [x] fix https://github.com/hpcugent/easybuild-framework/issues/1322

* [x] make sure comments from original easyconfig file are retained in dumped easyconfig file (~~#1327~~ + ~~#1330~~ + ~~1345~~)

* [x] check what needs to be changed to make dumped result adher to easyconfigs style (use https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/TEMPLATE.eb as initial reference) (~~#1327~~ + ~~#1330~~ + ~~1345~~)
  * [x] avoid long lines (< 120 characters)
  * [x] indentation of list/dictionary values

boegel:
closing this, since all list items have been checked off

thanks @Caylo!

PR #1308, opened 2015-06-29
Fix for to_str in dump method of easyconfig.py

Caylo:
Resolves issue #1305.


boegel:
Jenkins: ok to test

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -505,14 +493,14 @@ def to_str(x):
                 for key2, [def_val, _, _] in DEFAULT_CONFIG.items():
                     # only print parameters that are different from the default value
                     if key1 == key2 and val != def_val:
-                        ebtxt.append("%s = %s" % (key1, to_str(val)))
+                        ebtxt.append("%s = %s" % (key1, quote_str(val, True)))
```
boegel: use `force_str=True` (don't specify optional arguments unnamed)

but, the whole `force_str` thing can be dropped anyway (yes, I know I suggested it ;-)), since the return value is only used in a string interpolation context anyway (and it's already part of a tuple, so no need to force it into a string anymore)

easybuild/tools/utilities.py
```diff
@@ -68,14 +68,14 @@ def quote_str(x):
...
-        if "'" in x and '"' in x:
+        if '\n' in x or ("'" in x and '"' in x):
             return '"""%s"""' % x
         elif '"' in x:
             return "'%s'" % x
         else:
             return '"%s"' % x
     else:
-        return x
+	return x
```
boegel: no tabs please, see http://legacy.python.org/dev/peps/pep-0008/#tabs-or-spaces

add this to your `~/.vimrc`:
```
:set expandtab tabstop=4 shiftwidth=4
```

test/framework/easyconfig.py
```diff
@@ -1132,6 +1133,38 @@ def test_hide_hidden_deps(self):
...
+	self.assertEqual(quote_str('foo\'bar"baz'), '"""foo\'bar"baz"""')
+	self.assertEqual(quote_str("foo'bar\"baz"), '"""foo\'bar"baz"""')
+	self.assertEqual(quote_str("foo \n bar"), '"""foo \n bar"""')	
+
+	""" Non-string values """
+	n = 42
+	self.assertEqual(quote_str(n), 42)
+	l = ["foo", "bar"]
+        self.assertEqual(quote_str(l), ["foo", "bar"])
+	t = ('foo', 'bar')
```
boegel: no need to the `t` variable, just call `quote_str(("foo", "bar"))` ?

test/framework/easyconfig.py
```diff
@@ -1132,6 +1133,38 @@ def test_hide_hidden_deps(self):
...
+        test_ecs_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs')
+
+        test_ec = os.path.join(self.test_prefix, 'test.eb')
+
+        ec = EasyConfig(os.path.join(test_ecs_dir, 'toy-0.0.eb'))
+        ec.dump(test_ec)
+        ectxt = read_file(test_ec)
+        version_regex = re.compile('^version = "0.0"', re.M)
+        self.assertTrue(version_regex.search(ectxt))
+        # FIXME: add more test cases (using easyconfigs available in test/framework/easyconfigs)
```
boegel: flesh this out too please

test/framework/module_generator.py
```diff
@@ -249,8 +249,8 @@ def test_load_msg(self):
             tcl_load_msg = '\n'.join([
                 '',
                 "if { [ module-info mode load ] } {",
-                "    puts stderr \"test \\$test \\$test",
-                "test \\$foo \\$bar\"",
+                "    puts stderr \"\"\"test \\$test \\$test",
+                "test \\$foo \\$bar\"\"\"",
```
boegel: hmm, is this what we want? why do quotes get wrapped in quotes now?

it seems like `to_str` differs from `quote_str` in this?

test/framework/modules/imkl/10.3.12.361
```diff
@@ -13,7 +13,7 @@ module-whatis {Intel Math Kernel Library is a library of highly optimized,
  applications that require maximum performance. Core math functions include
  BLAS, LAPACK, ScaLAPACK, Sparse Solvers, Fast Fourier Transforms, Vector Math, and more. - Homepage: http://software.intel.com/en-us/intel-mkl/}
 
-set root	/var/folders/6y/x4gmwgjn5qz63b7ftg4j_40m0000gn/T/tmpZg_7A8
+set root	/tmp/eb-bI0pBy/eb-DmuEpJ/eb-leoYDw/eb-UtJJqp/tmp8P3FOY
```
boegel: I need to get this fixed, this shouldn't happen (it's a problem of a test that fiddles with this file in place failing, and then not restoring this file to what it was)

test/framework/module_generator.py
Caylo: It should indeed behave differently on ```\n``` - should i put the to_str method back in?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -505,18 +498,42 @@ def to_str(x):
...
             if not key in printed_keys and val != self._config[key][0]:
-                ebtxt.append("%s = %s" % (key, to_str(self._config[key][0])))
+                ebtxt.append("%s = %s" % (key, quote_str(self._config[key][0], esc_newline=True)))
 
-        eb_file.write('\n'.join(ebtxt))
+        eb_file.write(self.wrap('\n'.join(ebtxt), width=120))
         eb_file.close()
 
+    
+    def wrap(self, text, width=80):
```
stdweird: what is missing in `textwrap`?

easybuild/tools/utilities.py
```diff
@@ -56,11 +56,11 @@ def flatten(lst):
     return res
 
 
-def quote_str(x):
+def quote_str(x, esc_newline=False):
     """
     Obtain a new value to be used in string replacement context.
 
-    For non-string values, it just returns the exact same value.
+    For non-string values, it just returns the exact same value.
```
stdweird: remove trailing whitespace

easybuild/framework/easyconfig/easyconfig.py
boegel: we'll revisit `textwrap` vs this custom wrap function in an upcoming PR (to tackle the tasks mentioned in #1307)

for now, we'll look into getting this PR merged, since it fixes #1305

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -39,6 +39,7 @@
 import difflib
 import os
 import re
+import textwrap
```
boegel: @Caylo: kick this out for now

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -466,34 +467,27 @@ def toolchain(self):
             self.log.debug("Initialized toolchain: %s (opts: %s)" % (tc_dict, self['toolchainopts']))
         return self._toolchain
 
+
```
boegel: not need (but whatever ;-))

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -466,34 +467,27 @@ def toolchain(self):
...
             ['name', 'version', 'versionprefix', 'versionsuffix'],
             ['homepage', 'description'],
             ['toolchain', 'toolchainopts'],
-            ['source_urls', 'sources'],
+            ['sources', 'source_urls'],
             ['patches'],
             ['builddependencies', 'dependencies', 'hiddendependencies'],
             ['parallel', 'maxparallel'],
-            ['osdependencies']
+            ['osdependencies'],
```
boegel: this should go right below the `dependencies` group

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -466,34 +467,27 @@ def toolchain(self):
...
+            ['prebuildopts', 'preinstallopts', 'preconfigopts'],
```
boegel: funky order here

also, this should actually be 3 separate groups:

```python
['preconfigopts', 'configopts'],
['prebuildopts', 'buildopts'],
['preinstallopts', 'installopts'],
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -466,34 +467,27 @@ def toolchain(self):
...
         grouped_keys = [
+            ['easyblock'],
             ['name', 'version', 'versionprefix', 'versionsuffix'],
             ['homepage', 'description'],
             ['toolchain', 'toolchainopts'],
-            ['source_urls', 'sources'],
+            ['sources', 'source_urls'],
             ['patches'],
             ['builddependencies', 'dependencies', 'hiddendependencies'],
             ['parallel', 'maxparallel'],
```
boegel: move this down, below `buildopts` (& co)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -466,34 +467,27 @@ def toolchain(self):
...
-            ['osdependencies']
+            ['osdependencies'],
+            ['prebuildopts', 'preinstallopts', 'preconfigopts'],
+            ['sanity_check_paths'],
+            ['moduleclass']
```
boegel: don't include these two here, they should always be at the end, so write these out (with a newline in between) after the for loop that tackles the 'other' parameters

easybuild/tools/utilities.py
```diff
@@ -56,7 +56,7 @@ def flatten(lst):
     return res
 
 
-def quote_str(x):
+def quote_str(x, esc_newline=False):
```
boegel: if you want to keep this short and non-descriptive, go with `esc_nl`, otherwise use `escape_newline` (not a mix of both)

easybuild/tools/utilities.py
```diff
@@ -56,7 +56,7 @@ def flatten(lst):
...
     """
     Obtain a new value to be used in string replacement context.
```
boegel: extend docstring with:

```python
@param escape_newline: wrap strings that include a newline in triple quotes

File: test/framework/easyconfig.py [hunk 17]
```diff
@@ -1132,6 +1133,49 @@ def test_hide_hidden_deps(self):
         self.assertEqual(ec['hiddendependencies'][0]['full_mod_name'], 'toy/.0.0-deps')
         self.assertEqual(ec['dependencies'], [])
 
+    def test_quote_str(self):
+        """Test quote_str function."""
+        self.assertEqual(quote_str('foo'), '"foo"')
+        self.assertEqual(quote_str('foo\'bar'), '"foo\'bar"')
+        self.assertEqual(quote_str('foo\'bar"baz'), '"""foo\'bar"baz"""')
+        self.assertEqual(quote_str("foo'bar\"baz"), '"""foo\'bar"baz"""')
```
boegel: add test cases with `\n` in the string, both without specifying `escape_newline` (to test the default), and with setting it to `True` and `False`

test/framework/easyconfig.py
```diff
@@ -1132,6 +1133,49 @@ def test_hide_hidden_deps(self):
...
+        test_ec2 = os.path.join(self.test_prefix, 'test2.eb')
+
+        ec = EasyConfig(os.path.join(test_ecs_dir, 'toy-0.0.eb'))
+        ec.dump(test_ec)
+        ectxt = read_file(test_ec)
+        version_regex = re.compile('^version = "0.0"', re.M)
+        self.assertTrue(version_regex.search(ectxt))
+        description_regex = re.compile('^description = "Toy C program."', re.M)
+        self.assertTrue(description_regex.search(ectxt))
+        """ Parse the result again """
```
boegel: don't use docstring syntax inline, make this a comment:

```python
# parse result again
```

test/framework/easyconfig.py
```diff
@@ -1132,6 +1133,49 @@ def test_hide_hidden_deps(self):
...
+        """ Parse the result again """ 
+        dumped_ec = EasyConfig(test_ec)
+
+        ec2 = EasyConfig(os.path.join(test_ecs_dir, 'goolf-1.4.10.eb'))
+        ec2.dump(test_ec2)
+        ectxt2 = read_file(test_ec2)
+        name_regex = re.compile('^name = "goolf"', re.M)
+        self.assertTrue(name_regex.search(ectxt2))
+        moduleclass_regex = re.compile('^moduleclass = "toolchain"', re.M)
+        self.assertTrue(moduleclass_regex.search(ectxt2))
```
boegel: this is copy paste, so make this whole thing a for loop over `['toy-0.0.eb', 'goolf-1.4.10.eb']

test/framework/modules/imkl/10.3.12.361
boegel: ok to ignore here, I'll make sure this remains unchanged

test/framework/easyconfig.py
```diff
@@ -1132,6 +1133,51 @@ def test_hide_hidden_deps(self):
...
+        test_ecs_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs')
+
+        for f in ['toy-0.0.eb', 'goolf-1.4.10.eb', 'ScaLAPACK-2.0.2-gompi-1.4.10-OpenBLAS-0.2.6-LAPACK-3.4.2.eb']:
+            test_ec = os.path.join(self.test_prefix, 'test.eb')
+
+            ec = EasyConfig(os.path.join(test_ecs_dir, f))
+            ec.dump(test_ec)
+            ectxt = read_file(test_ec)
+            name_regex = re.compile('^name = "[^"]*"', re.M)
+            description_regex = re.compile('^description = "[^"]*"', re.M)
```
boegel: the regex here is kind of flawed... `""` would match too, for example, and since `[^"]*` also matches the empty string, you're not checking the full string anyway...

maybe just go with `'^description = ["\']'` ?

boegel: also, use 'raw' strings for regexes: `r'^description = ["\']'`

test/framework/easyconfig.py
```diff
@@ -1132,6 +1133,51 @@ def test_hide_hidden_deps(self):
...
+            version_regex = re.compile('^version = [0-9\.]*', re.M)
+            self.assertTrue(name_regex.search(ectxt))
```
boegel: you can make this another loop (easier to extend later):

```python
patterns = [r'^name = ["\']', r'^version = [0-9]', r'^description = ["\']']
for pattern in patterns:
    regex = re.compile(pattern, re.M)
    self.assertTrue(regex.search(ectxt), "Pattern '%s' found in: %s" % (regex.pattern, ectxt))
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -472,47 +472,57 @@ def dump(self, fp):
...
+
+        def check_and_print(keyset):
+            """
+            Internal function checking for default values
+            """
+            for group in keyset:
+                printed = False
+                for key1 in group:
+                    val = self._config[key1][0]
+                    for key2, [def_val, _, _] in DEFAULT_CONFIG.items():
```
boegel: the original code is pretty dumb here, there's no need for another for loop at all:

```python
for key in group:
  val = self._config[key][0]
  if key in DEFAULT_CONFIG:  
    def_val = DEFAULT_CONFIG[key]
    if val != def_val:
      ...
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -472,47 +472,57 @@ def dump(self, fp):
...
-                        ebtxt.append("%s = %s" % (key1, to_str(val)))
-                        printed_keys.append(key1)
-            ebtxt.append("")
+        check_and_print(grouped_keys)
 
         # print other easyconfig parameters at the end
         for key, [val, _, _] in DEFAULT_CONFIG.items():
-            if not key in printed_keys and val != self._config[key][0]:
-                ebtxt.append("%s = %s" % (key, to_str(self._config[key][0])))
+            if not key in printed_keys and not key in [k for sublist in last_keys for k in sublist] and val != self._config[key][0]:
```
boegel: don't use `not key in`, use `key not in` (which is not the same thing)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -472,47 +472,57 @@ def dump(self, fp):
...
+                ebtxt.append("%s = %s" % (key, quote_str(self._config[key][0], escape_newline=True)))
+
+        # print last two parameters
+        check_and_print(last_keys)
```
boegel: maybe make `last_keys` a flat list, and change this to

```python
check_and_print([[k] for k in last_keys])
```

Just something that popped to mind, `last_keys` isn't going to change most likely?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -472,47 +472,57 @@ def dump(self, fp):
...
-                for key2, [def_val, _, _] in DEFAULT_CONFIG.items():
-                    # only print parameters that are different from the default value
-                    if key1 == key2 and val != def_val:
-                        ebtxt.append("%s = %s" % (key1, to_str(val)))
-                        printed_keys.append(key1)
-            ebtxt.append("")
+        check_and_print(grouped_keys)
 
         # print other easyconfig parameters at the end
         for key, [val, _, _] in DEFAULT_CONFIG.items():
```
boegel: fix the order in which the keys are being considered here (the order of the elements in `.items()` on a dict-like object may differ, you can't make assumptions on it's order here); maybe used `sorted`?

easybuild/framework/easyconfig/easyconfig.py
Caylo: The whole check is unnecessary, `key` is always in ```DEFAULT_CONFIG```.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -472,47 +472,53 @@ def dump(self, fp):
...
+            ['preconfigopts', 'configopts'],
+            ['prebuildopts', 'buildopts'],
+            ['preinstallopts', 'installopts'],
             ['parallel', 'maxparallel'],
-            ['osdependencies']
         ]
 
+        last_keys = ['sanity_check_paths', 'moduleclass']
+
+        def check_and_print(keyset):
```
boegel: last remark (for this PR, at least ;-)): rename to `include_defined_parameters`, and change docstring to

```python
"""
Internal function to include parameters in the dumped easyconfig file which have a non-default value.
"""
```

boegel:
@Caylo: looks good now, going in, thanks!

Congrats on your first contribution to EasyBuild! ;-)

PR #1311, opened 2015-07-02
Fixed extra_options bug in dump()

Caylo:


boegel:
Jenkins: ok to test

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -498,22 +503,20 @@ def include_defined_parameters(keyset):
...
-
         # print easyconfig parameters ordered and in groups specified above
         ebtxt = []
         printed_keys = []
         include_defined_parameters(grouped_keys)
 
         # print other easyconfig parameters at the end
-        for key, [val, _, _] in sorted(DEFAULT_CONFIG.items()):
+        for (key, val) in default_values.items():
             if key not in printed_keys and key not in last_keys and val != self._config[key][0]:
```
boegel: let's make this a little bit more readable (and also faster, by not using `.items()`):

```python
keys_to_ignore = printed_keys + last_keys
for key in default_values:
    if key not in keys_to_ignore and default_values[key] != self[key]:
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -498,22 +503,20 @@ def include_defined_parameters(keyset):
                 printed = False
                 for key in group:
                     val = self._config[key][0]
-                    def_val = DEFAULT_CONFIG[key]
-                    if val not in def_val:
+                    if val != default_values[key]:
```
boegel: replace `val` with `self[key]`, should work (see the `__getitem__` defined in the `EasyConfig` class)

boegel:
@Caylo: let's try and add a unit test that verifies this fix too...

The easiest way is to parse/dump/parse an easyconfig that uses the `EB_foo` easyblock that is included in the tests. You can use the `rawtxt` argument to avoid having to actually generate an easyconfig file first:

```python
    def test_dump_extra(self):
       """ ... """

       rawtxt = '\n'.join([
           "easyblock = 'EB_foo'",
           "name = 'foo'",
           ...
       ])
       ec = EasyConfig(None, rawtxt=rawtxt)
       ...
```

test/framework/easyconfig.py
```diff
@@ -1177,6 +1177,28 @@ def test_dump(self):
...
+            "toolchain = {'name': 'dummy', 'version': 'dummy'}",
+            "foo_extra1 = 'foobar'",
+        ])
+
+        testec = os.path.join(self.test_prefix, 'test.eb')
+
+        ec = EasyConfig(None, rawtxt=rawtxt)
+        ec.dump(testec)
+        extra_regex = re.compile(r'^foo_extra1 = "foobar"', re.M)
+        self.assertTrue(extra_regex.search(read_file(testec)))
```
boegel: maybe here you can do an exact comparison between rawtxt and what got dumped?

you'll need to make sure that rawtxt includes the necessary empty lines (and you may have to fiddle with single vs double quotes)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -497,25 +502,23 @@ def include_defined_parameters(keyset):
...
         printed_keys = []
         include_defined_parameters(grouped_keys)
 
         # print other easyconfig parameters at the end
-        for key, [val, _, _] in sorted(DEFAULT_CONFIG.items()):
-            if key not in printed_keys and key not in last_keys and val != self._config[key][0]:
-                ebtxt.append("%s = %s" % (key, quote_str(self._config[key][0], escape_newline=True)))
+        keys_to_ignore = printed_keys + last_keys
+        for key in default_values:
+            if key not in keys_to_ignore and default_values[key] != self[key]:
```
boegel: replace `default_values[key] != self[key]` with `self[key] != default_values[key]`, so that it matches the check above, and it makes more sense too: "is value equal to the default value"

test/framework/easyconfig.py
```diff
@@ -1177,6 +1177,28 @@ def test_dump(self):
...
+            "easyblock = 'EB_foo'",
+            "name = 'foo'",
+            "version = '0.0.1'",
+            "description = 'foo description'",
+            "homepage = 'http://foo.com/'",
+            "toolchain = {'name': 'dummy', 'version': 'dummy'}",
+            "foo_extra1 = 'foobar'",
+        ])
+
+        testec = os.path.join(self.test_prefix, 'test.eb')
```
boegel: make this a random file:

```python
testec, handle = tempfile.mkstemp(prefix=self.test_prefix, suffix='.eb')
os.close(handle)

boegel:
Excellent, going in, thanks @Caylo!

PR #1314, opened 2015-07-03
Dump template and constant values

Caylo:
Fixes point 4 and 5 of issue #1307.


boegel:
Jenkins: ok to test

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -64,7 +64,7 @@
 from easybuild.framework.easyconfig.parser import DEPRECATED_PARAMETERS, REPLACED_PARAMETERS
 from easybuild.framework.easyconfig.parser import EasyConfigParser, fetch_parameters_from_easyconfig
 from easybuild.framework.easyconfig.templates import template_constant_dict
-
+from easybuild.framework.easyconfig.templates import TEMPLATE_CONSTANTS
```
boegel: join this with the `import` line above, leave two blank lines under last `import` (that's intentional)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +491,31 @@ def dump(self, fp):
 
         last_keys = ['sanity_check_paths', 'moduleclass']
 
+        tmp = self.enable_templating
```
boegel: replace `tmp` with a proper name like `orig_enable_templating`

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +491,31 @@ def dump(self, fp):
...
+        self.enable_templating = False
```
boegel: explain why you're disabling it

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +491,31 @@ def dump(self, fp):
...
+
         # build dict of default values
         default_values = dict([(key, DEFAULT_CONFIG[key][0]) for key in DEFAULT_CONFIG])
         default_values.update(dict([(key, self.extra_options[key][0]) for key in self.extra_options]))
 
+        template_values = dict([(const[1], const[0]) for const in TEMPLATE_CONSTANTS])
+
+        def replace_templates(value):
```
boegel: but this right next to `resolve_templates` as a top-level function, rather than making it an inner function (inner functions should only be used by exception)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +491,31 @@ def dump(self, fp):
...
+            """ Internal function to replace certain values with constants"""
+            if isinstance(value, basestring):
+                value = template_values.get(value.lower(), value)
+
```
boegel: needs more whitespace ;)

easybuild/framework/easyconfig/easyconfig.py
boegel: this also helps to add a unit test for `replace_templates` specifically

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +491,31 @@ def dump(self, fp):
...
         # build dict of default values
         default_values = dict([(key, DEFAULT_CONFIG[key][0]) for key in DEFAULT_CONFIG])
         default_values.update(dict([(key, self.extra_options[key][0]) for key in self.extra_options]))
 
+        template_values = dict([(const[1], const[0]) for const in TEMPLATE_CONSTANTS])
+
+        def replace_templates(value):
+            """ Internal function to replace certain values with constants"""
+            if isinstance(value, basestring):
+                value = template_values.get(value.lower(), value)
```
boegel: what's with the `lower`?

test/framework/easyconfig.py
```diff
@@ -1178,7 +1178,7 @@ def test_dump(self):
             dumped_ec = EasyConfig(test_ec)
 
     def test_dump_extra(self):
-        """Test EasyConfig's dump() method for files containing extra values"""
+        """Test EasyConfig's dump() method for files containing extra values and templates"""
```
boegel: ah no, since you have a separate test :)

test/framework/easyconfig.py
```diff
@@ -1204,6 +1204,32 @@ def test_dump_extra(self):
...
+            '',
+            'name = "foo"',
+            'version = "0.0.1"',
+            '',
+            'homepage = "http://foo.com/"',
+            'description = "foo description"',
+            '',
+            'toolchain = {\'version\': \'dummy\', \'name\': \'dummy\'}',
+            '',
+            'sources = [\'%(namelower)s-%(version)s.TAR.gz\']',
```
boegel: this probably relates to the `lower` above?

a file ending in `TAR.gz` is very different from a file ending in `tar.gz`; casing matters (we're not using Windows ;-))

boegel: what happens if you make this:

```python
sources = ['foo-0.0.1.tar.gz']
```

In a first pass, this should become `'%(name)s-%(version)s.tar.gz'`, which could then be translated to `SOURCE_TAR_GZ`...

That would mean doing multiple passes with `replace_templates` though, until there are no more changes...

Is that feasible?

test/framework/easyconfig.py
```diff
@@ -1204,6 +1204,32 @@ def test_dump_extra(self):
...
+        rawtxt = '\n'.join([
+            'easyblock = "EB_toy"',
+            '',
+            'name = "foo"',
+            'version = "0.0.1"',
+            '',
+            'homepage = "http://foo.com/"',
+            'description = "foo description"',
+            '',
+            'toolchain = {\'version\': \'dummy\', \'name\': \'dummy\'}',
```
boegel: avoid escaping unless it's really needed:

```python
"toolchain = {'version': 'dummy', 'name': 'dummy'}",
```

boegel:
@Caylo: add another unit test dedicated to (only) `replace_templates` too, which probably begs for testing `resolve_templates` at the same time (since they're each others inverse, more or less)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +491,20 @@ def dump(self, fp):
...
         # build dict of default values
         default_values = dict([(key, DEFAULT_CONFIG[key][0]) for key in DEFAULT_CONFIG])
         default_values.update(dict([(key, self.extra_options[key][0]) for key in self.extra_options]))
 
+        self.generate_template_values()
+        templ_const = dict([(const[1], const[0]) for const in TEMPLATE_CONSTANTS])
+        templ_const.update([(self.template_values[key], key) for key in self.template_values if key != ''])
+
+        for i in templ_const:
+            print i
```
boegel: :)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -918,8 +931,27 @@ def resolve_template(value, tmpl_dict):
 
     return value
 
+def replace_templates(value, templ_const):
+    """
+    Given a value, try to substitute constants where possible.
```
boegel: s/constants/template strings/g

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -918,8 +931,27 @@ def resolve_template(value, tmpl_dict):
...
+        - value can be a string, list, tuple, dict or combination thereof
+        - templ_const is a dictionary of constants
+    """
+    if isinstance(value, basestring):
+        # TODO replace template values, not only constants
```
boegel: please fix

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -918,8 +931,27 @@ def resolve_template(value, tmpl_dict):
...
+        if isinstance(value, list):
+            value = [replace_templates(v, templ_const) for v in value]
+        elif isinstance(value, tuple):
+            value = tuple(replace_templates(list(value), templ_const))
+        elif isinstance(value, dict):
+            value = dict([(key, replace_templates(v, templ_const)) for key, v in value.items()])
+    return value
+
 
-def process_easyconfig(path, build_specs=None, validate=True, parse_only=False, hidden=None):
```
boegel: why? :P

test/framework/easyconfig.py
```diff
@@ -1204,6 +1204,32 @@ def test_dump_extra(self):
...
+        ])
+
+        handle, testec = tempfile.mkstemp(prefix=self.test_prefix, suffix='.eb')
+        os.close(handle)
+
+        ec = EasyConfig(None, rawtxt=rawtxt)
+        ec.dump(testec)
+        ectxt = read_file(testec)
+
+        regex = re.compile(r'sources \= \[\'SOURCELOWER_TAR_GZ\'\]', re.M)
```
boegel: avoid escaping `'`

test/framework/easyconfig.py
```diff
@@ -1204,6 +1204,32 @@ def test_dump_extra(self):
...
+        self.assertTrue(regex.search(ectxt))
```
boegel: it's useful to make `assertTrue` produce meaningful error messages:

```python
self.assertTrue(regex.search(ectxt), "Pattern '%s' found in: %s" % (regex.pattern, ectxt))
```

easybuild/framework/easyconfig/easyconfig.py
Caylo: Something went wrong there. :P

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +491,19 @@ def dump(self, fp):
...
+        orig_enable_templating = self.enable_templating
+        self.enable_templating = False # templated values should be dumped unresolved
+
         # build dict of default values
         default_values = dict([(key, DEFAULT_CONFIG[key][0]) for key in DEFAULT_CONFIG])
         default_values.update(dict([(key, self.extra_options[key][0]) for key in self.extra_options]))
 
+        self.generate_template_values()
+        templ_const = dict([(const[1], const[0]) for const in TEMPLATE_CONSTANTS])
+        templ_val = dict([(self.template_values[key], key) for key in self.template_values if len(self.template_values[key]) > 2])
```
boegel: shorter (and add a comment):

```python
# reverse map of templates longer than 2 characters, to inject template values where possible
templ_vals = dict([(val, key) for key, val in self.template_values.items() if len(val) > 2])
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +491,19 @@ def dump(self, fp):
...
+
+        exclude_keys = ['name', 'version', 'description', 'homepage', 'toolchain'] # values will not be templated for these keys
```
boegel: move comment above this line

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -63,7 +63,7 @@
 from easybuild.framework.easyconfig.licenses import EASYCONFIG_LICENSES_DICT, License
 from easybuild.framework.easyconfig.parser import DEPRECATED_PARAMETERS, REPLACED_PARAMETERS
 from easybuild.framework.easyconfig.parser import EasyConfigParser, fetch_parameters_from_easyconfig
-from easybuild.framework.easyconfig.templates import template_constant_dict
+from easybuild.framework.easyconfig.templates import template_constant_dict, TEMPLATE_CONSTANTS
```
boegel: put constant first in import line (to keep import lines consistent)

```python
from easybuild.framework.easyconfig.templates import TEMPLATE_CONSTANTS, template_constant_dict
```

did you check how `template_constant_dict` relates to `TEMPLATE_CONSTANTS`? sure you can't just use `template_constant_dict` (the names seems to suggest you can)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -502,8 +511,11 @@ def include_defined_parameters(keyset):
             for group in keyset:
                 printed = False
                 for key in group:
-                    if self[key] != default_values[key]:
-                        ebtxt.append("%s = %s" % (key, quote_str(self[key], escape_newline=True)))
+                    val = self[key]
+                    if val != default_values[key]:
+                        if key not in exclude_keys:
+                            val = replace_templates(val, templ_const, templ_val)
```
boegel: add some logging here, so we can debug replacements that shouldn't be happening:

```python
self.log.debug("Original value before replacing matching template values: %s", val)
val = replace_templates(val, templ_const, templ_val)
self.log.debug("New value after replacing matching template values: %s", val)
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -527,6 +539,8 @@ def include_defined_parameters(keyset):
         eb_file.write(('\n'.join(ebtxt)).strip()) # strip for newlines at the end
         eb_file.close()
 
+        self.enable_tamplating = orig_enable_templating
```
boegel: typo: `tamplating` vs `templating`

this shows you need a better unit test ;)

check whether values containing templates are still being properly 'expanded' after calling dump in the unit test

easybuild/framework/easyconfig/easyconfig.py
boegel: also, this should either move to `replace_templates`, or be defined as a constant list (`EXCLUDED_KEYS_REPLACE_TEMPLATES` or somesuch) on top of this module

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -918,6 +932,33 @@ def resolve_template(value, tmpl_dict):
...
+    """
+    if isinstance(value, basestring):
+        old_value = ""
+        while value != old_value:
+            old_value = value
+            if value in templ_const:
+                value = templ_const[value]
+            else:
+                # check for template values - longest strings first
+                for v in sorted(templ_val, key=lambda v: len(v), reverse=True):
```
boegel: you're sorting `templ_val` over and over again here, but it doesn't change over time

at least move the sorting bit out of the while loop, and maybe even out of `replace_templates` into `dump`? sorting it once should suffice ;)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -918,6 +932,33 @@ def resolve_template(value, tmpl_dict):
...
+def replace_templates(value, templ_const, templ_val):
+    """
+    Given a value, try to substitute template strings where possible.
+        - value can be a string, list, tuple, dict or combination thereof
+        - templ_const is a dictionary of template strings (constants)
+        - templ_val is a dictionary of template strings specific for this easyconfig file
+    """
+    if isinstance(value, basestring):
+        old_value = ""
+        while value != old_value:
```
boegel: doing the while loop here means you assume that each template value is a string

that's probably true today, but may not be in the future

so, why not make this a recursive function?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -918,6 +932,33 @@ def resolve_template(value, tmpl_dict):
...
 
+def replace_templates(value, templ_const, templ_val):
+    """
+    Given a value, try to substitute template strings where possible.
+        - value can be a string, list, tuple, dict or combination thereof
+        - templ_const is a dictionary of template strings (constants)
+        - templ_val is a dictionary of template strings specific for this easyconfig file
+    """
+    if isinstance(value, basestring):
+        old_value = ""
```
boegel: make this `None`?

test/framework/easyconfig.py
```diff
@@ -1204,6 +1204,32 @@ def test_dump_extra(self):
...
+
+        handle, testec = tempfile.mkstemp(prefix=self.test_prefix, suffix='.eb')
+        os.close(handle)
+
+        ec = EasyConfig(None, rawtxt=rawtxt)
+        ec.dump(testec)
+        ectxt = read_file(testec)
+
+        regex = re.compile(r"sources \= \['SOURCELOWER_TAR_GZ'\]", re.M)
+        self.assertTrue(regex.search(ectxt), "Pattern '%s' found in: %s" % (regex.pattern, ectxt))
```
boegel: extend this a bit:

* check that `description` is not templated with `%(name)s`
* also check values of other types, i.e. a string (`configopts`?), a dict (`sanity_check_paths`?), with some other templates (`namelower`?)

easybuild/framework/easyconfig/easyconfig.py
Caylo: They're not interchangeable -  TEMPLATE_CONSTANTS contains constant values like `('SOURCELOWER_TGZ', '%(namelower)s-%(version)s.tgz')` while template_constant_dict will generate templates specifically for this ec file (like `'version': '2.1.0'`)

boegel: ok, nvm then :)

test/framework/easyconfig.py
```diff
@@ -1204,6 +1203,49 @@ def test_dump_extra(self):
...
+            'description = "foo description"',
+            '',
+            "toolchain = {'version': 'dummy', 'name': 'dummy'}",
+            '',
+            "sources = ['foo-0.0.1.tar.gz']",
+            '',
+            'preconfigopts = "--opt1=%s" % name',
+            'configopts = "--opt2=0.0.1"',
+            '',
+            "sanity_check_paths = {'files': ['files/foo/bar'], 'dirs':[] }",
```
boegel: change to `'files/foo/foobar'`, which should result in `'files/%(namelower)s/foobar'`, i.e. the `foo` in `'foobar'` does not get replaced with the matching template?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -918,6 +939,33 @@ def resolve_template(value, tmpl_dict):
...
+    if isinstance(value, basestring):
+        old_value = None
+        while value != old_value:
+            old_value = value
+            if value in templ_const:
+                value = templ_const[value]
+            else:
+                # check for template values
+                for k, v in templ_val.items():
+                    value = re.sub(r"\b" + re.escape(k) + r"\b", r'%(' + v + ')s', value)
```
boegel: add a comment here explaining the `\b`, something like:

```python
# only replace full words with templates, not substrings, by using \b in regex

File: easybuild/framework/easyconfig/easyconfig.py [hunk 27]
```diff
@@ -918,6 +939,33 @@ def resolve_template(value, tmpl_dict):
...
+    """
+    if isinstance(value, basestring):
+        old_value = None
+        while value != old_value:
+            old_value = value
+            if value in templ_const:
+                value = templ_const[value]
+            else:
+                # check for template values
+                for k, v in templ_val.items():
```
boegel: don't use single-letter variables outside of a list comprehension, `key, val` is fine

boegel: or make it `templ_val, templ_name`, that's more descriptive

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -918,6 +939,33 @@ def resolve_template(value, tmpl_dict):
...
+                    value = re.sub(r"\b" + re.escape(k) + r"\b", r'%(' + v + ')s', value)
+
+    else:
+        if isinstance(value, list):
+            value = [replace_templates(v, templ_const, templ_val) for v in value]
+        elif isinstance(value, tuple):
+            value = tuple(replace_templates(list(value), templ_const, templ_val))
+        elif isinstance(value, dict):
+            value = dict([(key, replace_templates(v, templ_const, templ_val)) for key, v in value.items()])
```
boegel: either use `k, v` or `key, val`

`k, v` is probably less confusing here

test/framework/easyconfig.py
```diff
@@ -1204,6 +1203,49 @@ def test_dump_extra(self):
...
+
+        handle, testec = tempfile.mkstemp(prefix=self.test_prefix, suffix='.eb')
+        os.close(handle)
+
+        ec = EasyConfig(None, rawtxt=rawtxt)
+        ec.enable_templating = True
+        ec.dump(testec)
+        ectxt = read_file(testec)
+
+        self.assertTrue(ec.enable_templating)
```
boegel: add comment: `# templating should still be enabled after calling dump()`

test/framework/easyconfig.py
```diff
@@ -1204,6 +1203,49 @@ def test_dump_extra(self):
...
+
+        patterns = [
+            r"sources = \['SOURCELOWER_TAR_GZ'\]",
+            r'description = "foo description"',  # no templating for description
```
boegel: also check homepage?

boegel: and check that `name = 'Foo'` isn't templated to `name = '%(name)s'`, since that wouldn't work :)

test/framework/easyconfig.py
```diff
@@ -1204,6 +1203,49 @@ def test_dump_extra(self):
...
+            "sanity_check_paths = {'files': ['files/foo/bar'], 'dirs':[] }",
+        ])
+
+        handle, testec = tempfile.mkstemp(prefix=self.test_prefix, suffix='.eb')
+        os.close(handle)
+
+        ec = EasyConfig(None, rawtxt=rawtxt)
+        ec.enable_templating = True
+        ec.dump(testec)
+        ectxt = read_file(testec)
```
boegel: try reparsing the dumped easyconfig file:

```python
# reparsing dumped easyconfig file should work
ecbis = EasyConfig(testec)
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -74,6 +75,9 @@
 # set of configure/build/install options that can be provided as lists for an iterated build
 ITERATE_OPTIONS = ['preconfigopts', 'configopts', 'prebuildopts', 'buildopts', 'preinstallopts', 'installopts']
 
+# values for these keys will not be templated in dump()
+EXCLUDED_KEYS_REPLACE_TEMPLATES = ['name', 'version', 'description', 'homepage', 'toolchain']
```
boegel: add `easyblock` to this list, and check it in the unit tests

test/framework/easyconfig.py
```diff
@@ -1204,6 +1203,49 @@ def test_dump_extra(self):
 
         dumped_ec = EasyConfig(testec)
 
+    def test_dump_template(self):
+        """ Test EasyConfig's dump() method for files containing templates"""
+        rawtxt = '\n'.join([
+            'easyblock = "EB_toy"',
```
boegel: use `EB_foo` as value, this should *not* be templated to `EB_%(namelower)s`

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -491,10 +495,20 @@ def dump(self, fp):
...
+
         # build dict of default values
         default_values = dict([(key, DEFAULT_CONFIG[key][0]) for key in DEFAULT_CONFIG])
         default_values.update(dict([(key, self.extra_options[key][0]) for key in self.extra_options]))
 
+        self.generate_template_values()
+        templ_const = dict([(const[1], const[0]) for const in TEMPLATE_CONSTANTS])
+        # reverse map of templates longer than 2 characters, to inject template values where possible, sorted on length
+        templ_val = OrderedDict([(self.template_values[k], k)
+                        for k in sorted(self.template_values, key=lambda k:len(self.template_values[k]), reverse=True)
```
boegel: introduce a variable for `sorted(...)`:

```python
keys = sorted(self.template_values, key=lambda k:len(self.template_values[k]), reverse=True)
templ_val = OrderedDict([(self.template_values[k], k) for k in keys if len(self.template_values[k]) > 2])
```

test/framework/easyconfig.py
boegel: tackled lower down, thx @caylo

boegel:
Going in, thanks @Caylo!

PR #1317, opened 2015-07-08
Generate docs for generic easyblocks (REVIEW)

Caylo:
Fixes https://github.com/hpcugent/easybuild-easyblocks/issues/636


boegel:
Jenkins: ok to test

easybuild/tools/docs.py
```diff
@@ -160,3 +147,115 @@ def avail_easyconfig_params(easyblock, output_format):
...
+    if classname in common_params:
+        commonly_used = 'Commonly used easyconfig parameters with ' + backtick(classname) + ' easyblock'
+        lines.extend([commonly_used, '-' * len(commonly_used)])
+
+        for opt in common_params[classname]:
+            param = '* ' + backtick(opt) + ' - ' + DEFAULT_CONFIG[opt][1]
+            lines.append(param)
+
+
+    if classname + '.eb' in os.listdir(os.path.join(os.path.dirname(__file__), 'doc_examples')):
```
boegel: these need to come from somewhere else, i.e. a location specified as an argument to `doc_easyblock` (which may need another name, although I can't think of one in 1-2-3)

also, don't include example .eb files here; at best, you can include some in the tests (next to the test easyconfigs we have already), so you can add a unit test for this code

easybuild/tools/docs.py
```diff
@@ -160,3 +147,115 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        lines.extend(['', 'Example', '-' * 8, '', '::', ''])
+        f = open(os.path.join(os.path.dirname(__file__), 'doc_examples', classname+'.eb'), "r")
```
boegel: use `read_file(filepath).split('\n')`

easybuild/tools/docs.py
```diff
@@ -160,3 +147,115 @@ def avail_easyconfig_params(easyblock, output_format):
...
+    return toc + sorted(docs)
+
+
+def doc_easyblock(eb_class):
+    """
+    Compose overview of one easyblock given class object of the easyblock in rst format
+    """
+    classname = eb_class.__name__
+
+    common_params = {
```
boegel: this should be a constant, defined on top of this module

easybuild/tools/docs.py
```diff
@@ -160,3 +147,115 @@ def avail_easyconfig_params(easyblock, output_format):
...
+    for i in range(0, len(values[0])):
+        table.append(line_tmpl.format(*[v[i] for v in values], c=' '))
+
+    table.extend([table_line, ''])
+
+    return table
+
+
+def backtick(string):
+    return '``' + string + '``'
```
boegel: missing docstring (and the function is rather silly, not sure if it's worth to define a function for this?

easybuild/tools/docs.py
```diff
@@ -34,25 +34,27 @@
...
 from easybuild.tools.ordereddict import OrderedDict
-from easybuild.tools.utilities import quote_str
-
+from easybuild.tools.utilities import quote_str, import_available_modules
+from easybuild.tools.filetools import read_file
 
 FORMAT_RST = 'rst'
 FORMAT_TXT = 'txt'
 
+def det_col_width(entries, title):
```
boegel: keep two blank lines between imports, constants, and top-level functions (so add one above here)

easybuild/tools/docs.py
```diff
@@ -34,25 +34,27 @@
...
+import inspect
+import os
 
 from easybuild.framework.easyconfig.default import DEFAULT_CONFIG, HIDDEN, sorted_categories
 from easybuild.framework.easyconfig.easyconfig import get_easyblock_class
 from easybuild.tools.ordereddict import OrderedDict
-from easybuild.tools.utilities import quote_str
-
+from easybuild.tools.utilities import quote_str, import_available_modules
+from easybuild.tools.filetools import read_file
```
boegel: two blank lines here please

easybuild/tools/docs.py
```diff
@@ -65,29 +67,14 @@ def det_col_width(entries, title):
...
-        line_tmpl = "{0:{c}<%s}   {1:{c}<%s}   {2:{c}<%s}" % (nw, dw, dfw)
-        table_line = line_tmpl.format('', '', '', c='=', nw=nw, dw=dw, dfw=dfw)
-
-        # table header
-        lines.append(table_line)
-        lines.append(line_tmpl.format(name_title, descr_title, dflt_title, c=' '))
-        lines.append(line_tmpl.format('', '', '', c='-'))
+        titles = ["**Parameter name**", "**Description**", "**Default value**"]
+        values = [
+            ['``' + name + '``' for name in grouped_params[grpname].keys()],
```
boegel: add comment at the end of this line: `# parameter name` (two spaces before `#` with same-line comments); similar below

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
         FORMAT_TXT: avail_easyconfig_params_txt,
     }
     return avail_easyconfig_params_functions[output_format](title, grouped_params)
+
```
boegel: two blank lines between top-level functions

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+def generic_easyblocks(path_to_examples, common_params={}, doc_functions=[]):
```
boegel: rename function name to make it clear this yields something in rst format

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+    """
+    Compose overview of all generic easyblocks
```
boegel: mention rst

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+    """
+    modules = import_available_modules('easybuild.easyblocks.generic')
```
boegel: the package name `easybuild.easyblocks.generic` should become a parameter of this function, so we can generate an overview of software-specific (i.e. non-generic) easyblocks too

maybe rename function to `gen_easyblocks_overview_rst`?

boegel: or, maybe just add a `generic=True` named argument, so you can include a matching title

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        for name,obj in inspect.getmembers(m, inspect.isclass):
+            eb_class = getattr(m, name)
+            # skip imported classes that are not easyblocks
+            if eb_class.__module__.startswith('easybuild.easyblocks.generic') and eb_class not in all_blocks:
+                all_blocks.append(eb_class)
+
+    for eb_class in all_blocks:
+        docs.append(doc_easyblock(eb_class, path_to_examples, common_params, doc_functions, all_blocks))
+
+    toc = ['.. contents:: Available generic easyblocks', '    :depth: 1', '']
```
boegel: drop the `Available generic easyblocks`, we should have a page title in place instead (above the `.. contents::`)

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+
+    return toc + sorted(docs)
+
+def doc_easyblock(eb_class, path_to_examples, common_params, doc_functions, all_blocks):
```
boegel: rename to `gen_easyblock_doc_section_rst`

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+    """
+    Compose overview of one easyblock given class object of the easyblock in rst format
```
boegel: mention rst

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+
+    lines = [
+        '``' + classname + '``',
+        '=' * (len(classname)+4),
+        '',
+    ]
+
+    bases = []
+    for b in eb_class.__bases__:
+        base = b.__name__ + '_' if b in all_blocks else b.__name__
```
boegel: make these references to other sections, via:

```
.. section_label:
```

```
:ref:`section_label`
```

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+            f = eb_class.__dict__[func]
+        elif func in eb_class.__bases__[0].__dict__:
+            f = eb_class.__bases__[0].__dict__[func]
+            inh = ' (inherited)'
+
+        if f.__doc__:
+            custom.append('* ``' + func + '`` - ' + f.__doc__.strip() + inh)
+
+    if custom:
+        title = 'Customised steps'
```
boegel: add `in <name> easyblock`

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        lines.extend([title, '-' * len(title)] + custom)
+        lines.append('')
+
+    # Add example if available
+    if classname + '.eb' in os.listdir(os.path.join(path_to_examples)):
+        lines.extend(['', 'Example', '-' * 8, '', '::', ''])
```
boegel: make section title `Example for <name> easyblock`

boegel: and don't hardcode `8`, put the section title in a variable and use `len`

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        for line in read_file(os.path.join(path_to_examples, classname+'.eb')).split('\n'):
```
boegel: what if no example is available? this will make it crash hard, it should just skip the `Example` section instead?

boegel: we maybe only include examples for some easyblocks, especially with non-generic ones

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+            lines.append('    ' + line.strip())
+        lines.append('') # empty line after literal block
+
+    return '\n'.join(lines)
+
```
boegel: double blank line please

easybuild/tools/docs.py
```diff
@@ -160,3 +147,134 @@ def avail_easyconfig_params(easyblock, output_format):
...
+    table.append(line_tmpl.format(*titles, c=' '))
+    table.append(table_line)
+
+    for i in range(0, len(values[0])):
+        table.append(line_tmpl.format(*[v[i] for v in values], c=' '))
+
+    table.extend([table_line, ''])
+
+    return table
+
```
boegel: drop emtpy line

easybuild/tools/docs.py
Caylo: With `if classname + '.eb' in os.listdir(os.path.join(path_to_examples)):` it won't crash, or am i overlooking something? I tested it with only a few examples as well, it should work

boegel: ah, yes, missed that

easybuild/tools/docs.py
```diff
@@ -160,3 +149,136 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        if f.__doc__:
+            custom.append('* ``' + func + '`` - ' + f.__doc__.strip() + inh)
+
+    if custom:
+        title = 'Customised steps'
+        lines.extend([title, '-' * len(title)] + custom)
+        lines.append('')
+
+    # Add example if available
+    if classname + '.eb' in os.listdir(os.path.join(path_to_examples)):
```
boegel: use `os.path.exists(os.path.join(path_to_examples, '%s.eb' % classname))` instead?

boegel:
@Caylo: to test the generated overview of easyblocks in rst format:

* I gave you write access to my fork of the `easybuild` repo where the docs are housed
* clone that repo: `git clone git@github.com:boegel/easybuild.git`
* check out the `generic_easyblocks` branch: `git checkout -b generic_easyblocks origin/generic_easyblocks`
* edit `docs/Generic_easyblocks.rst`, and commit the change
* push to the `generic_easyblocks` branch in my repo: `git push origin generic_easyblocks`
* http://boegel-eb.readthedocs.org/en/generic_easyblocks/Generic_easyblocks.html should refresh automagically shortly after your push
* you should also see that https://github.com/hpcugent/easybuild/pull/130 is updated after the push (readthedocs will follow a couple of minutes later)

boegel:
@Caylo: can you create a small (Python) script that taps into this, and that can be included in a `scripts` subdirectory in https://github.com/hpcugent/easybuild/tree/master/docs ?

easybuild/tools/docs.py
```diff
@@ -34,25 +34,29 @@
...
 """
 import copy
+import inspect
+import os
 
 from easybuild.framework.easyconfig.default import DEFAULT_CONFIG, HIDDEN, sorted_categories
 from easybuild.framework.easyconfig.easyconfig import get_easyblock_class
 from easybuild.tools.ordereddict import OrderedDict
-from easybuild.tools.utilities import quote_str
+from easybuild.tools.utilities import quote_str, import_available_modules
```
boegel: sort imported functions alphabetically

easybuild/tools/docs.py
```diff
@@ -34,25 +34,29 @@
...
+from easybuild.tools.filetools import read_file
```
boegel: sort alphabetically with other `from ... import` lines

easybuild/tools/docs.py
```diff
@@ -65,29 +69,14 @@ def det_col_width(entries, title):
...
-
-        # figure out column widths
-        nw = det_col_width(grouped_params[grpname].keys(), name_title) + 4  # +4 for raw format ("``foo``")
-        dw = det_col_width([x[0] for x in grouped_params[grpname].values()], descr_title)
-        dfw = det_col_width([str(quote_str(x[1])) for x in grouped_params[grpname].values()], dflt_title)
+        titles = ["**Parameter name**", "**Description**", "**Default value**"]
+        values = [
+            ['``' + name + '``' for name in grouped_params[grpname].keys()],  # parameter name
+            [x[0] for x in grouped_params[grpname].values()],  # description
+            [str(quote_str(x[1])) for x in grouped_params[grpname].values()]  #default value
```
boegel: space after `#` ;-)

easybuild/tools/docs.py
```diff
@@ -160,3 +149,139 @@ def avail_easyconfig_params(easyblock, output_format):
...
+                all_blocks.append(eb_class)
+
+    for eb_class in all_blocks:
+        docs.append(gen_easyblock_doc_section_rst(eb_class, path_to_examples, common_params, doc_functions, all_blocks))
+
+    title = 'Overview of generic easyblocks'
+
+    heading = ['=' * len(title), title, '=' * len(title), '', '.. contents::', '    :depth: 2', '']
+    return heading + sorted(docs)
+
```
boegel: insert extra blank line

easybuild/tools/docs.py
```diff
@@ -160,3 +149,139 @@ def avail_easyconfig_params(easyblock, output_format):
...
+def gen_easyblocks_overview_rst(package_name, path_to_examples, common_params={}, doc_functions=[]):
+    """
+    Compose overview of all easyblocks in the given package in rst format
+    """
+    modules = import_available_modules(package_name)
+    docs = []
+    all_blocks = []
+
+    # get all blocks
+    for m in modules:
```
boegel: no single-letter variable names outside of list comprehensions, use `mod` or `module` (make sure `module` is not a keyword in Python)

easybuild/tools/docs.py
```diff
@@ -160,3 +149,139 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        for name,obj in inspect.getmembers(m, inspect.isclass):
+            eb_class = getattr(m, name)
+            # skip imported classes that are not easyblocks
+            if eb_class.__module__.startswith(package_name) and eb_class not in all_blocks:
+                all_blocks.append(eb_class)
+
+    for eb_class in all_blocks:
```
boegel: sort `all_blocks` here using `sorted(all_blocks, key=...)`

easybuild/tools/docs.py
```diff
@@ -160,3 +149,139 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        docs.append(gen_easyblock_doc_section_rst(eb_class, path_to_examples, common_params, doc_functions, all_blocks))
+
+    title = 'Overview of generic easyblocks'
+
+    heading = ['=' * len(title), title, '=' * len(title), '', '.. contents::', '    :depth: 2', '']
+    return heading + sorted(docs)
```
boegel: don't sort `docs` here, should already be sorted via for loop above

easybuild/tools/docs.py
```diff
@@ -160,3 +149,139 @@ def avail_easyconfig_params(easyblock, output_format):
...
+            # skip imported classes that are not easyblocks
+            if eb_class.__module__.startswith(package_name) and eb_class not in all_blocks:
+                all_blocks.append(eb_class)
+
+    for eb_class in all_blocks:
+        docs.append(gen_easyblock_doc_section_rst(eb_class, path_to_examples, common_params, doc_functions, all_blocks))
+
+    title = 'Overview of generic easyblocks'
+
+    heading = ['=' * len(title), title, '=' * len(title), '', '.. contents::', '    :depth: 2', '']
```
boegel: define this list over multiple lines, to make the indent clear:

```python
heading = [
    '=' * len(title),
    title,
    '=' * len(title),
    '',
    '.. contents::',
    '    :depth: 2',
    '',
]
```

easybuild/tools/docs.py
```diff
@@ -160,3 +149,139 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        ]
+
+        lines.extend(mk_rst_table(titles, values))
+
+    # Add commonly used parameters
+    if classname in common_params:
+        commonly_used = 'Commonly used easyconfig parameters with ``' + classname + '`` easyblock'
+        lines.extend([commonly_used, '-' * len(commonly_used)])
+
+        for opt in common_params[classname]:
```
boegel: use `mk_rst_table` here too?

easybuild/tools/docs.py
```diff
@@ -160,3 +149,139 @@ def avail_easyconfig_params(easyblock, output_format):
...
+
+    # Add docstring for custom steps
+    custom = []
+    inh = ''
+    for func in doc_functions:
+        if func in eb_class.__dict__:
+            f = eb_class.__dict__[func]
+        elif func in eb_class.__bases__[0].__dict__:
+            f = eb_class.__bases__[0].__dict__[func]
+            inh = ' (inherited)'
```
boegel: hmm, not sure about this, my preference would be to not included these inherited methods and only mention the ones custom to the easyblock being documented in this section

test/framework/docs.py
```diff
@@ -0,0 +1,65 @@
...
+from test.framework.utilities import EnhancedTestCase, init_config
+from unittest import TestLoader, main
+from easybuild.tools.docs import mk_rst_table
+
+class DocsTest(EnhancedTestCase):
+
+    def test_rst_table(self):
+        """ Test mk_rst_table function """
+        entries_1 = [['one', 'two', 'three']]
+        t1 = 'This title is long'
```
boegel: no it's not :P

test/framework/docs.py
```diff
@@ -0,0 +1,65 @@
...
+
+from test.framework.utilities import EnhancedTestCase, init_config
+from unittest import TestLoader, main
+from easybuild.tools.docs import mk_rst_table
+
+class DocsTest(EnhancedTestCase):
+
+    def test_rst_table(self):
+        """ Test mk_rst_table function """
+        entries_1 = [['one', 'two', 'three']]
```
boegel: what's with all the `_1`? there's no `_2`...

boegel:
@Caylo: also include a test for `gen_easyblocks_overview_rst`? There are test easyblocks/easyconfigs included in `test/framework/sandbox` and `test/framework/easyconfigs` that you can use.

easybuild/tools/docs.py
boegel: maybe also use `backlinks` and `local`?

The title in the contents block looks weird, I *think* `local` may avoid that.

boegel: see http://docutils.sourceforge.net/docs/ref/rst/directives.html#table-of-contents

easybuild/tools/docs.py
```diff
@@ -160,3 +151,154 @@ def avail_easyconfig_params(easyblock, output_format):
...
+    for eb_class in sorted(all_blocks, key=lambda c: c.__name__):
+        docs.append(gen_easyblock_doc_section_rst(eb_class, path_to_examples, common_params, doc_functions, all_blocks))
+
+    title = 'Overview of generic easyblocks'
+
+    heading = [
+        '=' * len(title),
+        title,
+        '=' * len(title),
+        '',
```
boegel: insert a line that makes it clear that this page is auto-generated, something like:

```python
'',
'*(this page was generated automatically using* ``easybuild.tools.docs.gen_easyblocks_overview_rst()`` *)*',
'',
```

easybuild/tools/docs.py
```diff
@@ -160,3 +151,154 @@ def avail_easyconfig_params(easyblock, output_format):
...
+        '.. contents::',
```
boegel: the contents block is pretty long due to the number of sections, I think a long list of links to subsections separated by ` - ` would be nicer?

that does imply that you make sure that each section gets a label: `.. _this_is_a_label:`, which you can refer to using:

```
:ref:`this_is_a_label`
```

boegel: ah, you already have labels in place I see below, good

test/framework/docs.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertTrue(check_configuremake in ebdoc)
+
+        for mod in modules:
+            for name, obj in inspect.getmembers(mod, inspect.isclass):
+                eb_class = getattr(mod, name)
+                # skip imported classes that are not easyblocks
+                if eb_class.__module__.startswith(module):
+                    self.assertTrue(name in ebdoc)
+
+        patterns = [
```
boegel: check for the full list of (test) easyblocks here:

```
... - :ref:`ConfigureMake` - :ref:`DummyExtension` - ...
```

boegel:
Going in, thanks @Caylo!

PR #1320, opened 2015-07-13
fix more bugs in EasyConfig.dump()

boegel:
With the easybuild-easyconfigs unit tests now also using `dump()`, several shortcomings in it are revealed (cfr. https://jenkins1.ugent.be/view/EasyBuild%20(develop)/job/easybuild-easyconfigs_unit-test_hpcugent_develop/1428/consoleFull).

This PR fixes these issues, such that the easyconfig unit tests are fail-free again:

* avoid templated values referring to the key they define, e.g. `versionsuffix = '_foo'` becomes `versionsuffix = '%(versionsuffix)s'`
* 'unparse' dependency easyconfigs parameters, rather than spitting out the `dict` representing the parsed dependency specs in tuple format, and handle dependencies marked as external module as a special case
* don't quote string containing a single quote using single quotes, since that yields `SyntaxError`s

cc @Caylo

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -506,7 +506,7 @@ class EasyConfig(object):
         templ_const = dict([(quote_py_str(const[1]), const[0]) for const in TEMPLATE_CONSTANTS])
 
         # reverse map of templates longer than 2 characters, to inject template values where possible, sorted on length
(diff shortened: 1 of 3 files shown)


boegel:
Going in as a hotfix, since this is blocking progress on EasyBuild v2.2.0.

@Caylo: can you take a look this, see if you have any remarks/thoughts?

PR #1327, opened 2015-07-17
Preserve comments in dump() method

Caylo:
At the moment, there's no support for comments on lines that do not contain keys because for that to work, dump's indentation needs to be fixed first

for example, right now, the comment in

    dependencies = [
        ('JasPer', '1.900.1'),
        ('grib_api', '1.9.18'),
        ('netCDF', '4.1.3'),  # can't use 4.2, because Fortran stuff is in seperate netCDF-Fortran install
    ]

will be ignored. This will have to be fixed when I take care of the indentation.

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -46,7 +46,7 @@ from vsc.utils.patterns import Singleton
 import easybuild.tools.environment as env
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option, get_module_naming_scheme
-from easybuild.tools.filetools import decode_class_name, encode_class_name, read_file
+from easybuild.tools.filetools import decode_class_name, encode_class_name, read_file, write_file
 from easybuild.tools.module_naming_scheme import DEVEL_MODULE_SUFFIX
 from easybuild.tools.module_naming_scheme.utilities import avail_module_naming_schemes, det_full_ec_version
 from easybuild.tools.module_naming_scheme.utilities import det_hidden_modname, is_valid_module_name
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyconfig/parser.py
```diff
@@ -131,6 +137,46 @@ def _read(self, filename=None):
...
+            'header' : [],
+            'inline' : dict(),
+            'above' : dict(),
+         }
+
+        raw = self.rawcontent.split('\n')
+        header = True
+
+        i = 0
+        while i<len(raw):
```
JensTimmerman: why not use a for loop?
precompute len(raw) so it's not computed on every itteration.

Caylo: it's not a for loop because the counter needs to be incremented more if there are multiline comments (the nested loop on line 164)

JensTimmerman:
jenkins: ok to test

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -537,29 +538,43 @@ def include_defined_parameters(keyset):
...
+                        add_key_and_comments(key, val)
+
                         printed_keys.append(key)
                         printed = True
                 if printed:
                     ebtxt.append('')
 
+        def add_key_and_comments(key, val):
+            """ Adds key, value pair and comments (if there are any) to the dump file """
+            if key in self.comments['inline']:
```
JensTimmerman: I think it's ugly to use the self from the outer scope,
can't you make this a private method instead of an inner method, and pass self along?

boegel: 'outer scope' is overstating things a bit, this inner function (in a method) is just referring to a 'variable' that's available in that methods scope, which is fine

moving that function out of the method, either as private method or a stand-alone function, would be messier imho (since it'll be split of from its context)

boegel:
going in, follow-up in #1330

PR #1333, opened 2015-07-30
add rst support for --avail-easyconfig-templates

Caylo:
@JensTimmerman please review

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -259,7 +259,7 @@ def template_documentation_rst():
     instance_title = 'Template names/values derived from easyconfig instance'
     doc.extend([instance_title, '-' * len(instance_title)])
     table_values = [
-        ['``' + name[0] + '``' for name in TEMPLATE_NAMES_EASYCONFIG],
+        ['``%s``' % name[0] for name in TEMPLATE_NAMES_EASYCONFIG],
         [name[1] for name in TEMPLATE_NAMES_EASYCONFIG],
     ]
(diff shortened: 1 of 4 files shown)


JensTimmerman:
ok to test

easybuild/framework/easyconfig/templates.py
```diff
@@ -203,8 +208,17 @@ def template_constant_dict(config, ignore=None, skip_lower=True):
     return template_values
 
 
-def template_documentation():
+def template_documentation(output_format):
```
JensTimmerman: set a default to TXT, so you don't have to change it everywhere it's not passed on yet.

JensTimmerman:
ok

easybuild/framework/easyconfig/templates.py
```diff
@@ -235,4 +249,51 @@ def template_documentation():
...
+    config_title = 'Template names/values as set in easyconfig'
+    doc.extend([config_title, '-' * len(config_title)])
+    for name in TEMPLATE_NAMES_CONFIG:
+        doc.append('* ``' + name + '``')
+    doc.append('')
+
+    lower_title = 'Lowercase values of template values'
+    doc.extend([lower_title, '-' * len(lower_title)])
+    table_values = [
+        ['``' + TEMPLATE_NAMES_LOWER_TEMPLATE % {'name': name} + '``' for name in TEMPLATE_NAMES_LOWER],
```
JensTimmerman: come to think of it, I think I prefer
'``%s``' % foo over '`' + foo + `'`

PR #1334, opened 2015-07-30
added rst support for --list-easyblocks

Caylo:
Merge https://github.com/hpcugent/easybuild-framework/pull/1333 first

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -608,7 +608,7 @@ class EasyBuildOptions(GeneralOption):
 
         # dump available easyblocks
         if self.options.list_easyblocks:
-            msg += self.avail_easyblocks()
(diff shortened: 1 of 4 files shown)


easybuild/tools/options.py
```diff
@@ -324,7 +325,7 @@ def informative_options(self):
                                          "easyblock-specific ones by using -e)"),
                                         'choice', 'store_or_None', FORMAT_TXT, [FORMAT_RST, FORMAT_TXT], 'a'),
             'avail-easyconfig-templates': (("Show all template names and template constants "
```
JensTimmerman: this should use the global output format, so just boolean here? not store or none

JensTimmerman:
ok to test

PR #1335, opened 2015-07-30
added rst support for --avail-cfgfile-constants

Caylo:
Merge https://github.com/hpcugent/easybuild-framework/pull/1333 first


easybuild/tools/options.py
```diff
@@ -324,7 +326,7 @@ def informative_options(self):
                                          "easyblock-specific ones by using -e)"),
                                         'choice', 'store_or_None', FORMAT_TXT, [FORMAT_RST, FORMAT_TXT], 'a'),
             'avail-easyconfig-templates': (("Show all template names and template constants "
```
JensTimmerman: use global output format here

Caylo: fixed this in https://github.com/hpcugent/easybuild-framework/pull/1336

JensTimmerman: ok

PR #1330, opened 2015-07-28
Dump formatting

Caylo:
~~https://github.com/hpcugent/easybuild-framework/pull/1327~~ will have to be merged first.

Takes care of formatting and inline comments for dump() method

PR #1344, opened 2015-08-07
fix key order of toolchain parameter definition in dumped easyconfig (WIP)

boegel:
continues where #1330 left off (which should be merged first)


akesandgren:
@boegel At the very least this is now broken due to "dummy" toolchain having been dropped.

akesandgren:
Ping @boegel

Issue #1329, opened 2015-07-23
Bug in packaging, module_generator in fake state before packaging


rjeschmi:
We missed the module path in testing the package feature and it is accidentally left in the "fake" state before the package step runs.

I'm looking into a PR for this, but it isn't trivial since the state isn't isolated (sanity check step seems to rely on the state already being fake for module_generator).

rjeschmi:
reported by @dirkdedraak

boegel:
fix available in #1348, let's continue the discussion there

PR #1348, opened 2015-08-11
rework module_generator to avoid keeping state w.r.t. fake modules

boegel:
fix (+ enhanced unit test) for #1329, reported by @dirkdedraak and @rjeschmi

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1028,7 +1028,7 @@ class EasyBlock(object):
         env = copy.deepcopy(os.environ)
 
         # create fake module
-        fake_mod_path = self.make_module_step(True)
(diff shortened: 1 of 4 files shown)


boegel:
@rjeschmi: up for reviewing this? maybe @pforai can take a look too, since he has played around in `module_generator.py` quite a bit in the past

rjeschmi:
Yeah, this is generally what I thought needed to be done (but didn't really know how to do). Most of the state information wasn't helpful and was ripe for side-effects.

I'll try testing it, but otherwise looks good.

rjeschmi:
tested it and it seems to work fine with packaging.

JensTimmerman:
ok

boegel:
Thanks for the reviewing/testing @rjeschmi and @JensTimmerman!

Issue #1328, opened 2015-07-21
Problems with hiddendependencies and --job


ocaisa:
It seems that the gc3pie integration doesn't currently trigger the ```--hidden``` option to the submitted ```eb``` command when it encounters a hidden dependency (though that may have been true before too, I just didn't know it).

boegel:
fixed in #1356

PR #1356, opened 2015-08-14
include --hidden in job script when needed

boegel:
fix for #1328 

This also fixes an other issue I ran into, i.e. dependencies marked as external modules were not being filtered when determining job dependencies, which manifested itself as:

```
ERROR: test_build_easyconfigs_in_parallel_pbs_python (__main__.ParallelBuildTest)
Test build_easyconfigs_in_parallel(), using (mocked) pbs_python as backend for --job.
  File "test/framework/parallelbuild.py", line 121, in test_build_easyconfigs_in_parallel_pbs_python
    jobs = build_easyconfigs_in_parallel("echo '%(spec)s'", ordered_ecs, prepare_first=False)
  File "/Users/kehoste/work/easybuild-framework/easybuild/tools/parallelbuild.py", line 97, in build_easyconfigs_in_parallel
    new_job = create_job(active_job_backend, build_command, ec, output_dir=output_dir)
  File "/Users/kehoste/work/easybuild-framework/easybuild/tools/parallelbuild.py", line 174, in create_job
    if easyconfig['ec']['hidden']:
  File "/Users/kehoste/work/easybuild-framework/easybuild/framework/easyconfig/easyconfig.py", line 102, in new_ec_method
    return ec_method(self, key, *args, **kwargs)
  File "/Users/kehoste/work/easybuild-framework/easybuild/framework/easyconfig/easyconfig.py", line 694, in __getitem__
    raise EasyBuildError("Use of unknown easyconfig parameter '%s' when getting parameter value", key)
EasyBuildError: "Use of unknown easyconfig parameter 'hidden' when getting parameter value"

Ran 2 tests in 3.140s
```

cc @ocaisa

### DIFF
--- a/easybuild/tools/parallelbuild.py
+++ b/easybuild/tools/parallelbuild.py
@@ -121,6 +121,7 @@ def submit_jobs(ordered_ecs, cmd_line_opts, testing=False, prepare_first=True):
     @param ordered_ecs: list of easyconfigs, in the order they should be processed
     @param cmd_line_opts: list of command line options (in 'longopt=value' form)
     @param testing: If `True`, skip actual job submission
+    @param prepare_first: prepare by runnning fetch step first for each easyconfig
     """
     curdir = os.getcwd()
 
@@ -133,7 +134,7 @@ def submit_jobs(ordered_ecs, cmd_line_opts, testing=False, prepare_first=True):
     # compose string with command line options, properly quoted and with '%' characters escaped
     opts_str = subprocess.list2cmdline(opts).replace('%', '%%')
 
-    command = "unset TMPDIR && cd %s && eb %%(spec)s %s %%(add_opts)s--testoutput=%%(output_dir)s" % (curdir, opts_str)
+    command = "unset TMPDIR && cd %s && eb %%(spec)s %s %%(add_opts)s --testoutput=%%(output_dir)s" % (curdir, opts_str)
     _log.info("Command template for jobs: %s" % command)
     job_info_lines = []
     if testing:
--- a/easybuild/tools/parallelbuild.py
+++ b/easybuild/tools/parallelbuild.py
@@ -96,8 +96,10 @@ def build_easyconfigs_in_parallel(build_command, easyconfigs, output_dir='easybu
         _log.info("creating job for ec: %s" % str(ec))
         new_job = create_job(active_job_backend, build_command, ec, output_dir=output_dir)
 
-        # sometimes unresolved_deps will contain things, not needed to be build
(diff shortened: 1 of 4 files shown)


boegel:
@wpoely86, @JensTimmerman, @rjeschmi: please review?

easybuild/tools/parallelbuild.py
```diff
@@ -131,13 +133,13 @@ def submit_jobs(ordered_ecs, cmd_line_opts, testing=False):
     # compose string with command line options, properly quoted and with '%' characters escaped
     opts_str = subprocess.list2cmdline(opts).replace('%', '%%')
 
-    command = "unset TMPDIR && cd %s && eb %%(spec)s %s --testoutput=%%(output_dir)s" % (curdir, opts_str)
+    command = "unset TMPDIR && cd %s && eb %%(spec)s %s %%(add_opts)s--testoutput=%%(output_dir)s" % (curdir, opts_str)
```
wpoely86: why the unset TMPDIR?

wpoely86: missing space between `add_opts` and `--testoutput`?

easybuild/tools/parallelbuild.py
```diff
@@ -113,7 +115,7 @@ def build_easyconfigs_in_parallel(build_command, easyconfigs, output_dir='easybu
     return jobs
 
 
-def submit_jobs(ordered_ecs, cmd_line_opts, testing=False):
+def submit_jobs(ordered_ecs, cmd_line_opts, testing=False, prepare_first=True):
```
wpoely86: extend docstring too

wpoely86:
looks fine

easybuild/tools/parallelbuild.py
boegel: I'm not sure the `unset $TMPDIR &&` is still needed (since `eb` set it itself), but it was at some point...

Tried to figure it out with `git log -L136,136:easybuild/tools/parallelbuild.py`, but no luck, it's been there since forever.

Probably since before EB starting setting `$TMPDIR` and taking control over what directory `tempfile.*` is using, so it's likely it can go. But I'm not going to do that in this PR. :-)

boegel:

boegel:
I've should've known better than to merge this without having a unit test for `dep_graph`, since switching away from `unresolved_deps` broke `--dep-graph` (it resulted in an unconnected set of nodes being dumped, rather than an actual graph).

#1359 fixes the issue, and adds a unit test for `dep_graph`

PR #1357, opened 2015-08-14
use 'python -O' in 'eb' command & add support for running unit tests with eb --test (WIP)

boegel:
This is interesting because of significant performance improvements for (mainly) the logger via https://github.com/hpcugent/vsc-base/pull/182, but needs to be evaluated carefully to make sure this doesn't break anything.

First step is to see whether all the unit tests still pass under `python -O`.

See also http://stackoverflow.com/questions/2055557/python-optimized-mode .

edit: now requires https://github.com/hpcugent/vsc-base/pull/188


boegel:
TODO:

* [ ] fix broken tests
* [ ] add support for `eb --test`, to make sure unit tests are run exactly like `eb` is (e.g., with `python -O`)
* [ ] update documentation on running unit tests

test/framework/build_log.py
```diff
@@ -65,7 +65,7 @@ def test_easybuilderror(self):
         self.assertErrorRegex(EasyBuildError, 'BOOM', raise_easybuilderror, 'BOOM')
         logToFile(tmplog, enable=False)
 
-        log_re = re.compile("^%s :: BOOM \(at .*:[0-9]+ in [a-z_]+\)$" % getRootLoggerName(), re.M)
+        log_re = re.compile("^%s :: BOOM( \(at .*:[0-9]+ in [a-z_]+\))?$" % getRootLoggerName(), re.M)
```
boegel: not sure how happy I am with this, since this basically indicates that the location where errors occur are no longer available...

@JensTimmerman: maybe you've been a little bit too eager with putting the `if __debug__`s in place?

I'll try and see if this can be restored by rolling back one of the `if __debug__`s without big impact on performance...

boegel:
Needs more work:

```
$ eb --test
== Running framework unit tests...

Running tests...
.......................F...............................(skipping SvnRepository test)
.................................................................................................................................................................................................................................................................GC3Pie not available, skipping test
...........
ERROR [0.326s]: test_generate_cmd_line (test.framework.options.CommandLineOptionsTest)
Test for generate_cmd_line.
  File "/Users/kehoste/work/easybuild-framework/test/framework/options.py", line 1673, in test_generate_cmd_line
    self.assertEqual(ebopts.generate_cmd_line(), [])
AssertionError: Lists differ: ['--test'] != []

First list contains 1 additional elements.
First extra element 0:
--test

- ['--test']
+ []

Ran 323 tests in 544.241s

FAILED (errors=1)

Generating XML reports...
ERROR: Not all tests were successful.
Log available at /var/folders/8s/_frgh9sj6m744mxt5w5lyztr0000gn/T/eb-59xoyb/eb-F68gCF/eb-9Vdxio/easybuild-tests-PSz63l.log , XML output of tests available in test-reports directory
ERROR: One or more tests failed!
```

boegel:
Jenkins: test this please

test/framework/build_log.py
JensTimmerman: No I have not been to eager, the location is only visible in debug mode.

eb
```diff
@@ -82,5 +82,5 @@ then
     echo "Make sure you PYTHONPATH setting is correct."
     exit 1
 else
-    python $easybuild_main $@
+    python -O $easybuild_main $@
```
boegel: supporting a way to turn of the use of `-O` might make sense...

maybe the wrapper should check whether `-d` or `--debug` was passed on the command line (may be tricky for `-d`...) + check whether `$EASYBUILD_DEBUG` is set? this doesn't catch the case where `debug` is enabled via a config file though...

boegel:
@wpoely86: thoughts on this, especially w.r.t. being able to turn off the use of `python -O`? Still hardwired now.

wpoely86:
The `-O` is fine by me. But running the unit tests from the main program seems a bit weird to me? That's something that's usually clearly separated?

JensTimmerman:
agreed, unit tests are not usually shipped with the packaged version of an application, and it's not something you would run a lot, only when developing, so I don't see a reason to add it, but now it's there, oh well.
However, --test is probably a bad option to specify, because to me this would imply you are going to test the build of the easyconfig you've given as argument.

the --test option doesn't seem to be an option at all, but more of an alternate argument? 
so it becomes clear tha t`eb test` has nothing to do with the options you give when you run
`eb easyconfig.eb` 
and since an easyconfing can actually be called test already, maybe it would be less confusing and clear if you just ad a new `eb-test` command?

boegel:
I removed the `--test` bit, since it wasn't allowing to run single tests, which is actually what I'm after...

@wpoely86: please (re)review?

This makes a major difference in terms of speed, I really want this merged ASAP.

wpoely86:
This all looks fairly trivial?

boegel:
@wpoely86: yep, the only change needed to make the tests happy is the thing in `build_log`.

wpoely86:
lgtm

boegel:

PR #1359, opened 2015-08-15
fix --dep-graph, use all_dependencies EasyConfig class variable rather than 'dependencies' parameter

boegel:
fix for bug introduced in #1356, see also https://github.com/hpcugent/easybuild-easyconfigs/pull/1870


boegel:
This needs a unit test to catch that bug...

boegel:
@wpoely86, @pforai, @rjeschmi: please review?

wpoely86:
looks fine

boegel:

PR #1360, opened 2015-08-15
fix ActiveMNS.det_full_module_name for external modules

boegel:
this is required to fix the issue with https://github.com/hpcugent/easybuild-easyconfigs/pull/1870, since `det_full_module_name` is also called for dependencies that are marked as external modules now

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1105,8 +1105,13 @@ class ActiveMNS(object):
     def det_full_module_name(self, ec, force_visible=False):
         """Determine full module name by selected module naming scheme, based on supplied easyconfig."""
         self.log.debug("Determining full module name for %s (force_visible: %s)" % (ec, force_visible))
-        mod_name = self._det_module_name_with(self.mns.det_full_module_name, ec, force_visible=force_visible)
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86 or @pforai: please review?

pforai:
LGTM! How did not notice this earlier?

boegel:
We filter external modules in `resolve_dependencies`, so they don't appear in the output of `--dry-run` currently (I plan to change that).

And before https://github.com/hpcugent/easybuild-easyconfigs/pull/1853, the external modules were never considered by the easyconfigs unit test, so `det_full_module_name` was never called for external modules. :-)

rjeschmi:
looks fine, is this going to affect packaging? Maybe we need to look at these cases just to be sure they don't cause too many problems

boegel:
For packaging, we just need to make sure we consider both `dependencies` and `hiddendependencies` (since the latter are not included in the former after the easyconfig is parsed).

Maybe we also need to worry about `builddependencies`, not sure if/how we can/should make that distinction when packaging.

If you feel something is wrong there, please open an issue/PR for it, let's not discuss it here; keeping sort of off-topic discussions going in a closed PR (which this will be soon) is not a good idea imho.

boegel:
Thanks for the review @pforai and @rjeschmi!

Issue #1340, opened 2015-08-04
installation of R fails when `libpng` is a hidden dependency


riccardomurri:
(I am not sure if this report belongs into `easybuild-framework` or
into `easybuild-easyblocks`; please excuse me if I used the wrong
tracker.)

I am trying to build `R-3.2.1-foss-2015a`; I have successfully built
it already once, and am now trying with `--hide-deps=libpng`.
However, when `libpng` is hidden, the build step fails::

        == 2015-08-04 08:17:44,145 main ERROR add_dependencies: no module 'libpng/.1.6.17-foss-2015a' found for dependency {'external_module_metadata': {}, 'dummy': False, 'toolchain': {'version': '2015a', 'name': 'foss'}, 'name': 'libpng', 'short_mod_name': 'libpng/.1.6.17-foss-2015a', 'external_module': False, 'versionsuffix': '', 'full_mod_name': 'libpng/.1.6.17-foss-2015a', 'version': '1.6.17', 'hidden': True} (at easybuild/software/EasyBuild/2.2.0/lib/python2.7/site-packages/easybuild_framework-2.2.0-py2.7.egg/easybuild/tools/toolchain/toolchain.py:339 in add_dependencies)

The module file however is there::

        $ ls /apps/redhatenterpriseserver-7.1-x86_64/modules/all/
        Autoconf  Automake  FFTW  foss  GCC  gompi  hwloc  Java  libjpeg-turbo  libpng  libtool  NASM  numactl  OpenBLAS  OpenMPI  R  ScaLAPACK

        $ ls /apps/redhatenterpriseserver-7.1-x86_64/modules/all/libpng/ -A
        .1.6.17-foss-2015a.lua

and EasyBuidl is instructed to find it::

        $ env | fgrep EASYBUILD_
        EASYBUILD_INSTALLPATH=/apps/redhatenterpriseserver-7.1-x86_64
        EASYBUILD_CONFIGFILES=/apps/etc/easybuild.cfg

For completeness' sake I'm showing also the entire config::

    $ cat $EASYBUILD_CONFIGFILES
    [config]

    # module files configuration
    modules-tool = Lmod
    module-syntax = Lua

    # enable dependency resolution; search for `.eb` files here
    robot = /apps/easybuild/configs

    # temporary build path
    buildpath = /scratch/rmurri

    # use a shared dir for temp log storage (works better with --job)
    tmpdir = /scratch/rmurri
    tmp-logdir = /scratch/rmurri

[...]

PR #1364, opened 2015-08-18
fix 'exist' for hidden Lua module files

boegel:
fix for #1340

Existence of hidden modules is checked via `module show`, and then checking whether the module file path matches with the module name.

The regex wasn't taking into account that the module file path may have a `.lua` extension for hidden Lua module files.


boegel:
@pforai: up for giving this a review, since you're familiar with this?

pforai:
LGTM!

boegel:
Thanks for the review @pforai, and @riccardomurri for reporting the bug!

Issue #1367, opened 2015-08-19
include-module-naming-schemes doesn't seem to work


ocaisa:
I tried to use the option `--include-module-naming-schemes=toolchain_mns.py` with the naming scheme of #1168 but I get the following:
```
-bash-4.2$ eb --include-module-naming-schemes=~/toolchain_mns.py --module-naming-scheme=ToolchainMNS
Usage: eb [options] easyconfig [...]

eb: error: option --module-naming-scheme: invalid choice: 'ToolchainMNS' (choose from 'CategorizedHMNS', 'EasyBuildMNS', 'HierarchicalMNS', 'MigrateFromEBToHMNS')

```

boegel:
@ocaisa: can you try giving it a full path, e.g. `$PWD/toolchain_mns.py`?

Does that file exist?

ocaisa:
Doesn't help, the new MNS doesn't seem to appear in the list of possible options (am I getting the option syntax wrong?). For example, if I try to use an MNS I know doesn't exist, I should get the list of possibilities and the new MNS is not included:
```
-bash-4.2$ eb --include-module-naming-schemes=/homeb/zam/swmanage/MNS_TEST/toolchain_mns.py --module-naming-scheme=test
Usage: eb [options] easyconfig [...]

eb: error: option --module-naming-scheme: invalid choice: 'test' (choose from 'CategorizedHMNS', 'EasyBuildMNS', 'HierarchicalMNS', 'MigrateFromEBToHMNS')

```

boegel:
Can you try enabling 'early debugging' by defining `$DEBUG_EASYBUILD_OPTIONS=1` and provide the output you get?

ocaisa:
Ok the issue occurs when the environment variable EASYBUILD_MODULE_NAMING_SCHEME is set to point to the included MNS:
```
-bash-4.2$  eb --include-module-naming-schemes=/homeb/zam/swmanage/MNS_TEST/toolchain_mns.py --avail-module-naming-schemesList of supported module naming schemes:
	CategorizedHMNS
	EasyBuildMNS
	MigrateFromEBToHMNS
	HierarchicalMNS
	ToolchainMNS
-bash-4.2$ EASYBUILD_MODULE_NAMING_SCHEME=EasyBuildMNS eb --include-module-naming-schemes=/homeb/zam/swmanage/MNS_TEST/toolchain_mns.py --avail-module-naming-schemes
List of supported module naming schemes:
[...]

PR #1370, opened 2015-08-19
fix using included module naming scheme

boegel:
fix for #1367

`--include-module-naming-schemes` has to be processed before checking whether the specified naming scheme `--module-naming-scheme` is available or not, i.e. don't let the option parser check it by making `--module-naming-scheme` a choice option


boegel:
@wpoely86: please review?

wpoely86:
looks fine

boegel:
Thanks for the review @wpoely86, and @ocaisa for reporting the bug!

PR #1375, opened 2015-08-31
add support for extracting .iso files using 7z (p7zip)

boegel:
see also http://unix.stackexchange.com/a/70740

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -554,6 +554,9 @@ def extract_cmd(filepath, overwrite=False):
         else:
             cmd_tmpl = "unzip -qq %(filepath)s"
 
+    elif exts[-1] in ['iso']:
+        cmd_tmpl = "7z x %(filepath)s"
+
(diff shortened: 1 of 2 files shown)


wpoely86:
looks fine

fgeorgatos:
thanks about this; fyi. we were considering this issue while working back then for MATLAB PRs
* https://github.com/hpcugent/easybuild-easyconfigs/pull/233
* https://github.com/hpcugent/easybuild-easyconfigs/pull/234
will dig some more info and pass it along...

boegel:
Works as intended.

I did run into this problem: https://github.com/hpcugent/easybuild-framework/issues/1376, but that's out of scope for this PR.

Thanks for the review @wpoely86 & @fgeorgatos!

fgeorgatos:
for the record: 
it would be great to consider how this can be generalized, 
fi. invoke bsdtar in certain other systems or expand that logic for future archivers/decompressors etc...

PR #1382, opened 2015-09-03
Refactoring conditional allergy

sfranky:


boegel:
Jenkins: ok to test

boegel:
@sfranky: nice contribution, thanks!

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,42 @@ def get_local_dirs_purged():
...
+        # bzipped or bzipped tarball
+        '.tar.bz2': "tar xjf %(filepath)s",
+        '.bz2':     "bunzip2 %(filepath)s",
+        '.tbz':     "tar xjf %(filepath)s",
+        '.tbz2':    "tar xjf %(filepath)s",
+        '.tb2':     "tar xjf %(filepath)s",
+        # xzipped or xzipped tarball
+        '.tar.xz':  "unxz %(filepath)s --stdout | tar x",
+        '.xz':      "unxz %(filepath)s",
+        '.txz':     "unxz %(filepath)s --stdout | tar x",
```
boegel: how about ordering these alphabetically per block, i.e. `tar.xz`, `.txz`, `.xz` here

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,42 @@ def get_local_dirs_purged():
...
+        '.tar.gz':  "tar xzf %(filepath)s",
+        '.gz':      "gunzip -c %(filepath)s > %(target)s",
+        '.tgz':     "tar xzf %(filepath)s",
+        '.gtgz':    "tar xzf %(filepath)s",
+        # bzipped or bzipped tarball
+        '.tar.bz2': "tar xjf %(filepath)s",
+        '.bz2':     "bunzip2 %(filepath)s",
+        '.tbz':     "tar xjf %(filepath)s",
+        '.tbz2':    "tar xjf %(filepath)s",
+        '.tb2':     "tar xjf %(filepath)s",
```
boegel: start this block with `.bz2`, end with `tbz2`

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,42 @@ def get_local_dirs_purged():
...
+        raise EasyBuildError('Unknown file type for file %s (%s)', filepath, ext)
+
+    target = filename.rstrip(ext)
+
+    extract_cmds = {
+        # gzipped or gzipped tarball
+        '.tar.gz':  "tar xzf %(filepath)s",
+        '.gz':      "gunzip -c %(filepath)s > %(target)s",
+        '.tgz':     "tar xzf %(filepath)s",
+        '.gtgz':    "tar xzf %(filepath)s",
```
boegel: reorder these too please

fgeorgatos: N.B. 
the true ordering aspect is defined in the regex, as opposed to dict (regular python dicts are kinda useless for that)
So, is this still needed? what is the rule then?

boegel: It's just a style thing, it's easier on the eyes, or to make sure things aren't overlooked easily.

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,42 @@ def get_local_dirs_purged():
...
-    # zip file
-    elif exts[-1] in ['zip']:
-        if overwrite:
-            cmd_tmpl = "unzip -qq -o %(filepath)s"
-        else:
-            cmd_tmpl = "unzip -qq %(filepath)s"
-
-    elif exts[-1] in ['iso']:
-        cmd_tmpl = "7z x %(filepath)s"
+    pat = r'\.(?P<ext>tar\.gz|gz|tgz|gtgz|tar.bz2|bz2|tbz|tbz2|tb2|tar.xz|xz|txz|tar|zip|iso)$'
```
boegel: the list of recognised extensions is now in two places: here, and as keys in `extract_cmds` below

can we derive this regex from the keys in `extract_cmds`?

```python
pat = r'\.(?P<ext>%s)$' % '|'.join([ext.replace('.', '\\.') for ext in extract_cmds.keys()])
```

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,42 @@ def get_local_dirs_purged():
...
+    try:
+        ext = '.' + re.search(pat, filename, flags=re.IGNORECASE).group('ext')
```
boegel: `try`/`except` should be used for exceptional stuff, here we can anticipate that we may run into an unknown extension

so, without `try`/`except`:

```python
res = re.search(pat, filename, flags=re.IGNORECASE)
if res:
    ext = '.' + res.group('ext')
else:
    raise EasyBuildError("Unknown file type for file %s (%s)", filepath, ext)
```

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,42 @@ def get_local_dirs_purged():
...
+        '.tar':     "tar xf %(filepath)s",
+        # zip file
+        '.zip':     "unzip -qq -o %(filepath)s" if overwrite else "unzip -qq %(filepath)s",
+        # iso file
+        '.iso':     "7z x %(filepath)s"
+    }
 
-    if cmd_tmpl is None:
-        raise EasyBuildError('Unknown file type for file %s (%s)', filepath, exts)
+    cmd_tmpl = extract_cmds[ext.lower()]
```
boegel: this is another reason why the regex pattern should be derived from `extract_cmds.keys()`

if someone even only adds another extension in the regex but not in `extract_cmds`, this may result in a `KeyError``

easybuild/tools/filetools.py
boegel: it's a little bit trickier than this, maybe, since we need to make sure `tar.gz` is considered before `gz`...

so, either somehow enforce 'longest match first', or order the keys by length

fgeorgatos: longest match first is an interesting take and probably the right one.
Would something like the following list ordering work?

```
>>> d = {'what':1, 'the':2, 'heck':3, 'is':4, 'this':5}
>>> zip(*sorted(zip(map(len,d.keys()), d.keys()), reverse=True))[1]
('what', 'this', 'heck', 'the', 'is')
```
Then, you iterate over the list of keys like visible above?

easybuild/tools/filetools.py
fgeorgatos: agreed, it is potential trap for creating bugs; 
if the len() operator is sufficient deterrent to avoid that, let's go that way.

easybuild/tools/filetools.py
fgeorgatos: let's go the len() way for this, too...

easybuild/tools/filetools.py
boegel: why make things so complicated? ;-)
This will do:

```python
sorted(d.keys(), key=len, reverse=True)
```

easybuild/tools/filetools.py
boegel: I'd prefer keeping this alphabetical, since it's just a matter of style

we shouldn't give the impression that the order of keys here is preserved, since it's not...

easybuild/tools/filetools.py
fgeorgatos: argh, sweet ;-)

easybuild/tools/filetools.py
fgeorgatos: ok, pass - I guess this should be clarified via a comment line, then!

sfranky: sorry guys i was travelling yesterday. Love what you've done with it !!

boegel: @sfranky: are you up for fixing the remarks yourself (just push extra commits to your branch, this PR will be updated automagically)

if not, let us know, so we can handle it and get this merged

sfranky: Will do, I should be back to civilisation tonight!

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,44 @@ def get_local_dirs_purged():
...
+        '.txz':     "unxz %(filepath)s --stdout | tar x",
+        '.xz':      "unxz %(filepath)s",
+        # tarball
+        '.tar':     "tar xf %(filepath)s",
+        # zip file
+        '.zip':     "unzip -qq -o %(filepath)s" if overwrite else "unzip -qq %(filepath)s",
+        # iso file
+        '.iso':     "7z x %(filepath)s"
+    }
+    pat = r'(?P<ext>%s)$' % '|'.join([ext.replace('.', '\\.') for ext in sorted(extract_cmds.keys(), key=len, reverse=True)])
```
fgeorgatos: 1) fix line length of L534 (fi. by creating the list `suffixes`)
2) leave a space after L533
3) make sure that `myuntar.gz` falls in the right clause - a test case?

sfranky:
@boegel ready for review, pls check!

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,46 @@ def get_local_dirs_purged():
...
-            cmd_tmpl = "unzip -qq -o %(filepath)s"
-        else:
-            cmd_tmpl = "unzip -qq %(filepath)s"
+    suffixes = sorted(extract_cmds.keys(), key=len, reverse=True)
+    pat = r'(?P<ext>%s)$' % '|'.join([ext.replace('.', '\\.') for ext in suffixes])
+    res = re.search(pat, filename, flags=re.IGNORECASE)
+    if res:
+        ext = res.group('ext')
+    else:
+        ext = filename.split('.')[-1]  # best guess?
```
boegel: I wouldn't include this in the error message, since `split` isn't the way in which the extension is actually determined, it will only lead to confusion?

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,46 @@ def get_local_dirs_purged():
...
-    elif exts[-1] in ['txz']:
-        cmd_tmpl = "unxz %(filepath)s --stdout | tar x"
-
-    # tarball
-    elif exts[-1] in ['tar']:
-        cmd_tmpl = "tar xf %(filepath)s"
+    extract_cmds = {
+        # gzipped or gzipped tarball
+        '.gz':      "gunzip -c %(filepath)s > %(target)s",
+        '.gtgz':    "tar xzf %(filepath)s",
```
boegel: hmm, in my alphabet, `gtgz` comes before `gz` ;)

easybuild/tools/filetools.py
```diff
@@ -503,62 +504,46 @@ def get_local_dirs_purged():
...
+        ext = filename.split('.')[-1]  # best guess?
+        raise EasyBuildError('Unknown file type for file %s (%s)', filepath, ext)
 
-    elif exts[-1] in ['iso']:
-        cmd_tmpl = "7z x %(filepath)s"
+    target = filename.rstrip(ext)
 
-    if cmd_tmpl is None:
-        raise EasyBuildError('Unknown file type for file %s (%s)', filepath, exts)
+    cmd_tmpl = extract_cmds[ext.lower()]
```
boegel: do the `lower()` part where `ext` is defined?

boegel:
Looks a lot cleaner now, just a couple more very minor remarks... Sorry about being so picky. ;-)

easybuild/tools/filetools.py
sfranky: makes sense, I'll just print the whole filename then.

easybuild/tools/filetools.py
sfranky: errr, was too eager apparently!

easybuild/tools/filetools.py
sfranky: @boegel could you please rephrase, I don't understand this!

boegel: use this above:

```python
ext = res.group('ext').lower()
```

and change this line to

```python
cmd_tmpl = extract_cmds[ext]
```

sfranky: ok, but in this case, if ext is uppercase, 
`target = filename.rstrip(ext)`
will not work and return the whole filename.
This is needed in the extract_cmd dict, for the value of '.gz'.

boegel: Ok, good point, I missed that. Leave it as is then, just tackle the other remarks.

boegel:
Thanks a lot for this cleanup @sfranky, and for baring with us w.r.t. the remarks being made on your work.

Congratulations on your first contribution to EasyBuild, hopefully the first of many, and welcome to the wonderful world of EasyBuild.

Going in!

boegel:
And thanks @fgeorgatos for the support on this one!

sfranky:
glad to have contributed!

fgeorgatos:
good job, folks, let's keep it going  ;-)

PR #1384, opened 2015-09-05
fix typo

boegel:
reported by @gppezzi


boegel:
Thanks for pointing this out @gppezzi!

Issue #1219, opened 2015-03-11
defining $EBVAR env vars should be deprecated/dropped


boegel:
the motivation for having these around is pretty broken:

from `easybuild/tools/toolchain/toolchain.py`:
```
442             # also set unique named variables that can be used in Makefiles
443             # - so you can have 'CFLAGS = $(EBVARCFLAGS)'
444             # -- 'CLFLAGS = $(CFLAGS)' gives  '*** Recursive variable `CFLAGS'
445             # references itself (eventually).  Stop' error
446             setvar("EBVAR%s" % key, val)
```

wpoely86:
The `:=` assignment expanded immediately so when you do something like:
```
A =
CFLAGS = -Wall ${A}
```
where `A` gets filled in later in a makefile, you cannot switch to `:=`. To avoid recursive issues, you need a `EBVAR` variable.

boegel:
@wpoely86 Do you mind issuing a PR to clarify the comment why `$EBVAR` variables are also being defined, and to add a test to check whether they're being set (so we don't break this accidentally)?

Worth mentioning: `$EBVARCFLAGS` & co are required for the patch for FSL added in https://github.com/hpcugent/easybuild-easyconfigs/pull/4614

PR #1388, opened 2015-09-11
add support for --extended-dry-run/-x (REVIEW)

boegel:
TODO:

* [x] add warning that actual build procedure may diverge from output shown, due to conditional checks in the easyblock
* [x] catch `EasyBuildError` for each method being called, simply report the error and continue?
* [x] add `(DRY RUN)` tag for `== ` line printed for every step
* [x] auto-enable `--force`?
* [x] sort env vars + avoid `$EBVAR*` (see also #1219)
* [x] verbose dry run of `sanity_check_step`
* [x] verbose dry run of `make_module_step`
* [x] more verbose dry run version of `fetch_step`
* [x] avoid excessive output from `extensions_step`
* [x] simulate `module load` for deps/toolchain if module is missing
* [x] update documentation (see https://github.com/hpcugent/easybuild/pull/159)

Example output:

```
$ eb make-3.82-GCC-4.8.2.eb -x
== temporary log file in case of crash /tmp/eb-ksVC07/easybuild-MmAugb.log

== processing EasyBuild easyconfig /home/example/eb/easybuild-easyconfigs/easybuild/easyconfigs/m/make/make-3.82-GCC-4.8.2.eb

Important note: the actual build & install procedure that will be performed may diverge
(slightly) from what is outlined below, due to conditions in the easyblock which are
incorrectly handled in a dry run.
Any errors that may occur are ignored and reported as warnings, on a per-step basis.
Please be aware of this, and only use the information below for quick debugging purposes.

*** DRY RUN using 'ConfigureMake' easyblock (easybuild.easyblocks.generic.configuremake @ /home/example/eb/easybuild-easyblocks/easybuild/easyblocks/generic/configuremake.pyc) ***

== building and installing make/3.82-GCC-4.8.2...
fetching files... [DRY RUN]

[fetch_step method]
Available download URLs for sources/patches:
  * http://ftpmirror.gnu.org/make/$source

List of sources:
  * make-3.82.tar.bz2 will be downloaded to /home/example/eb/sources/m/make/make-3.82.tar.bz2

List of patches:
(none)

creating build dir, resetting environment... [DRY RUN]

[check_readiness_step method]

[make_builddir method]

[reset_changes method]

[handle_iterate_opts method]

unpacking... [DRY RUN]

[checksum_step method]
* expected checksum for make-3.82.tar.bz2: (none)

[extract_step method]
  running command "tar xjf /home/example/eb/sources/m/make/make-3.82.tar.bz2"
  (in /home/example/eb/build/make/3.82/GCC-4.8.2)

patching... [DRY RUN]

[patch_step method]

preparing... [DRY RUN]

[prepare_step method]
Defining build environment, based on toolchain (options) and specified dependencies...

Loading toolchain module...

module load GCC/4.8.2

Loading modules for dependencies...


Full list of loaded modules:
  1) GCC/4.8.2

Defining build environment...

  export CC="gcc"
  export CFLAGS="-O2"
  export CXX="g++"
  export CXXFLAGS="-O2"
  export F77="gfortran"
  export F90="gfortran"
  export F90FLAGS="-O2"
  export FFLAGS="-O2"
  export FLIBS="-lgfortran"
  export LDFLAGS="-L/home/example/eb/software/GCC/4.8.2/lib"
  export LIBS="-lm -lpthread"
  export OPTFLAGS="-O2"
  export PRECFLAGS=""

configuring... [DRY RUN]

[configure_step method]
  running command " ./configure --prefix=/home/example/eb/software/make/3.82-GCC-4.8.2 "
  (in /home/example/eb/build/make/3.82/GCC-4.8.2/make-3.82)

building... [DRY RUN]

[build_step method]
  running command " make -j 4 "
  (in /home/example/eb/build/make/3.82/GCC-4.8.2/make-3.82)

testing... [DRY RUN]

[test_step method]

installing... [DRY RUN]

[stage_install_step method]

[make_installdir method]

[install_step method]
  running command " make install "
  (in /home/example/eb/build/make/3.82/GCC-4.8.2/make-3.82)

taking care of extensions... [DRY RUN]

[extensions_step method]

postprocessing... [DRY RUN]

[post_install_step method]

sanity checking... [DRY RUN]

[sanity_check_step method]
Sanity check paths - file ['files']
  * bin/make
Sanity check paths - (non-empty) directory ['dirs']
  (none)
Sanity check commands
  (none)

cleaning up... [DRY RUN]

[cleanup_step method]

creating module... [DRY RUN]

[make_module_step method]
Generating module file /home/example/eb/modules/all/make/3.82-GCC-4.8.2, with contents:

    #%Module
    proc ModulesHelp { } {
        puts stderr { make-3.82: GNU version of make utility - Homepage: http://www.gnu.org/software/make/make.html
        }
    }
    
    module-whatis {Description: make-3.82: GNU version of make utility - Homepage: http://www.gnu.org/software/make/make.html}
    
    set root /home/example/eb/software/make/3.82-GCC-4.8.2
    
    conflict make
    
    if { ![ is-loaded GCC/4.8.2 ] } {
        module load GCC/4.8.2
    }
    
    setenv	EBROOTMAKE		"$root"
    setenv	EBVERSIONMAKE		"3.82"
    setenv	EBDEVELMAKE		"$root/easybuild/make-3.82-GCC-4.8.2-easybuild-devel"
    
    # Built with EasyBuild version 2.4.0
    

permissions... [DRY RUN]

[permissions_step method]

packaging... [DRY RUN]

[package_step method]

== COMPLETED: Installation ended successfully

Important note: the actual build & install procedure that will be performed may diverge
(slightly) from what is outlined above, due to conditions in the easyblock which are
incorrectly handled in a dry run.
Any errors that may occur are ignored and reported as warnings, on a per-step basis.
Please be aware of this, and only use the information above for quick debugging purposes.

(no ignored errors during dry run)

== Build succeeded for 1 out of 1
== Temporary log file(s) /tmp/eb-ksVC07/easybuild-MmAugb.log* have been removed.
== Temporary directory /tmp/eb-ksVC07 has been removed.
```


pforai:
woot! great visibility improvement. totally dig this. :+1:

verdurin:
Looks very useful.

Occurs to me that showing where files would be fetched from would be a good addition.

boegel:
@verdurin: you mean download URL and path where it'll be downloaded to?

verdurin:
@boegel - exactly, yes.

pescobar:
this looks really useful.

I would add to verdurin comment that if the source tarball is found locally it would be nice to show the full path to the source file.

boegel:
I think I'll go with something like:

```
[fetch_step method]
List of required files:
* make-3.82.tar.bz2 found at /home/kehoste/easybuild/sources/make-3.82.tar.bz2
```

vs

```
[fetch_step method]
List of required files:
* make-3.82.tar.bz2 downloaded from http://ftpmirror.gnu.org/make/make-3.82.tar.bz2 to /home/kehoste/easybuild/sources/make-3.82.tar.bz2
```

verdurin:
Yes, that's what's needed here.

fgeorgatos:
@boegel: nice one, this will help a lot plenty of people, good job!

gppezzi:
+1 for this feature!

Can we use the output from this for reproducing the build environment on the current shell? It would be nice for debugging/tweaking the parameters of a failing build.

boegel:
@gppezzi: i can tweak the format of the output a bit so you can just copy-paste this, yeah...

E.g.:

```
export FOO="bar baz"
module load dep1/1.2.3
```

boegel:
@pescobar, @verdurin: hmm, while looking into making the output for `fetch` more verbose as requested, I realised that things are more complicated: if multiple source URLs are listed, they're cycled over when trying to download a file... So, we don't know where files will be downloaded from until we actually try it.

I could make the download mechanism itself support a 'dry run' mode, where only the HTTP return code is checked rather than actually doing the download, but that implies that you need to be online to use `-x` if you don't have the files available yet.

So, counter proposal:

```
[fetch_step method]
Download URLs (for files not available yet):
* http://ftpmirror.gnu.org/make/
List of required files:
* make-3.82.tar.bz2 (found at /home/kehoste/easybuild/sources/make-3.82.tar.bz2)
```

vs

```
[fetch_step method]
Download URLs (for files not available yet):
* http://ftpmirror.gnu.org/make/
List of required files:
* make-3.82.tar.bz2 (will be downloaded to /home/kehoste/easybuild/sources/make-3.82.tar.bz2)
```

Thoughts?

pescobar:
@boegel I think printing the list of source urls would be fine. I would not go so far as doing a "fake http request" I don't think it's worth adding that extra complexity which can trigger errors which confuse the user. IMHO printing the list of source_urls is fine

Being picky, I would append the filename to the url so it prints something like (easier in case you want to copy& paste to try if the url works)

```
[fetch_step method]
Download URLs (for files not available yet):
* http://ftpmirror.gnu.org/make/make-3.82.tar.bz2
```

boegel:
@pescobar: glueing the file names to the source URLs is going to be a problem if there are multiple URLs and multiple source files...

pescobar:
@boegel I didn't realize about that. Then just ignore my suggestion I think it's fine as you suggest.

verdurin:
@boegel - understood about this extra wrinkle. Your proposal to deal with it makes sense to me.

Just as a comparison, with RPMs you can use spectool to do things like verifying source URLs. Perhaps something like that would be useful for EB, which would be separate from this sort of dry-run test.

boegel:
@pforai, @fgeorgatos, @verdurin, @gppezzi, @pescobar: updated the PR and description to reflect the current state

I still need to handle a couple of issues w.r.t. extensions (environment is printed over and over again when extensions are being handled, which is not nice when doing a dry run for a non-bare Perl easyconfig...)

feedback, suggestions?

easybuild/framework/easyblock.py
```diff
@@ -1239,15 +1272,27 @@ def fetch_step(self, skip_checksums=False):
                 raise EasyBuildError("EasyBuild-version %s is newer than the currently running one. Aborting!",
                                      easybuild_version)
 
+        if build_option('extended_dry_run'):
+
+            print_msg("Available download URLs for missing sources/patches:", silent=self.silent, prefix=False)
```
verdurin: I would change this to:

"Available download URLs for required sources/patches:"

I think I know why you've phrased it as "missing", but it sounds odd.

boegel: @verdurin: I see your point. I'll drop the 'required' part, and just make it "Available download URLs for sources/patches:"

boegel:
For me, the framework side of this is covered now, I'm happy with how this works and what it spits out (I updated the example output in the description).
More work will be needed to maximise the usefulness of the output we get, by doing some minor tweaks in the easyblocks (cfr. https://github.com/hpcugent/easybuild-easyblocks/pull/696 and https://github.com/hpcugent/easybuild-easyblocks/pull/685).

I have documentation for this in the works, will try and finish it ASAP (that's a blocker for getting this merged in).

I think it's time that this gets a thorough review... @wpoely86: are you up for taking a look at this beast?

boegel:
draft documentation is available at http://boegel-eb.readthedocs.org/en/extended_dry_run/Extended_dry_run.html (see https://github.com/hpcugent/easybuild/pull/159)

easybuild/framework/easyblock.py
```diff
@@ -235,8 +243,12 @@ def _init_log(self):
...
 
         this_module = inspect.getmodule(self)
-        self.log.info("This is easyblock %s from module %s (%s)",
-                      self.__class__.__name__, this_module.__name__, this_module.__file__)
+        eb_class = self.__class__.__name__
+        eb_loc = this_module.__file__
+        self.log.info("This is easyblock %s from module %s (%s)", eb_class, this_module.__name__, eb_loc)
+
+        if self.dry_run:
+            dry_run_msg("*** DRY RUN using '%s' easyblock (from %s) ***\n" % (eb_class, eb_loc), silent=self.silent)
```
wpoely86: this is really nitpicking but: `DRY RUN using easyblock '%s' `

And why not print the module name in case of dry run?

easybuild/framework/easyblock.py
```diff
@@ -1253,15 +1293,27 @@ def fetch_step(self, skip_checksums=False):
...
+
+            dry_run_msg("Available download URLs for sources/patches:", silent=self.silent)
+            if self.cfg['source_urls']:
+                for source_url in self.cfg['source_urls']:
+                    dry_run_msg("  * %s/$source" % source_url, silent=self.silent)
+            else:
+                dry_run_msg('(none)', silent=self.silent)
+
+            # actual list of sources is printed via _obtain_file_dry_run method
+            dry_run_msg("\nList of sources:", silent=self.silent)
```
wpoely86: so print this over there?

wpoely86: never mind, I understand :wink:

easybuild/framework/easyblock.py
```diff
@@ -1575,7 +1662,68 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
...
+            else:
+                commands = []
+                self.log.info("Using specified sanity check commands: %s" % commands)
+
+        for i, command in enumerate(commands):
+            # set command to default. This allows for config files with
+            # non-tuple commands
+            if not isinstance(command, tuple):
+                self.log.debug("Setting sanity check command to default")
+                command = (None, None)
```
wpoely86: shouldn't be:
```python
command = (command, None)
```

easybuild/framework/easyblock.py
```diff
@@ -1575,7 +1662,68 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
...
+            dry_run_msg("  (none)", silent=self.silent)
+
+    def _sanity_check_step(self, custom_paths=None, custom_commands=None, extension=False):
+        """Real version of sanity_check_step method."""
+        paths, path_keys_and_check, commands = self._sanity_check_step_common(custom_paths, custom_commands)
+
+        # check sanity check paths
         for key, check_fn in path_keys_and_check.items():
+            if key == 'dirs':
+                typ = "(non-empty) directory"
```
wpoely86: why replace this here but not at the `dry_run` version?

easybuild/framework/easyblock.py
```diff
@@ -1708,19 +1832,28 @@ def make_module_step(self, fake=False):
         txt += self.make_module_footer()
 
         mod_filepath = self.module_generator.get_module_filepath(fake=fake)
-        write_file(mod_filepath, txt)
 
-        self.log.info("Module file %s written: %s", mod_filepath, txt)
+        if self.dry_run:
+            if not fake:
```
wpoely86: I love the mixture of `dry_run` and `fake`. I can only imagine the look on people's faces when they read this :wink:

easybuild/framework/easyblock.py
```diff
@@ -1696,7 +1820,7 @@ def cleanup_step(self):
 
     def make_module_step(self, fake=False):
         """
-        Generate a module file.
+        Generate module file
```
wpoely86: maybe add a comment about the purpose of `fake`?

easybuild/framework/easyblock.py
```diff
@@ -1823,7 +1956,25 @@ def run_step(self, step, methods):
...
             self.log.info("Running method %s part of step %s" % ('_'.join(m.func_code.co_names), step))
-            m(self)
+
+            if self.dry_run:
+
+                dry_run_msg("[%s method]" % m(self).__name__, silent=self.silent)
+
+                # if an known possible error occurs, just report it and continue
+                try:
+                    m(self)()
```
wpoely86: why is the last `()` needed?

easybuild/framework/easyblock.py
```diff
@@ -1823,7 +1956,25 @@ def run_step(self, step, methods):
...
+                    if build_option('extended_dry_run_ignore_errors'):
+                        dry_run_msg("!!! WARNING !!! ignoring error: %s" % err, silent=self.silent)
+                        self.ignored_errors = True
+                    else:
+                        raise
+
+                dry_run_msg('', silent=self.silent)
+
+            else:
+                m(self)()
```
wpoely86: maybe a bit less empty lines in this part?

easybuild/framework/easyblock.py
```diff
@@ -1840,46 +1991,46 @@ def get_step(tag, descr, substeps, skippable, initial=True):
 
         # list of substeps for steps that are slightly different from 2nd iteration onwards
         ready_substeps = [
-            (False, lambda x: x.check_readiness_step()),
-            (True, lambda x: x.make_builddir()),
-            (True, lambda x: env.reset_changes()),
-            (True, lambda x: x.handle_iterate_opts()),
+            (False, lambda x: x.check_readiness_step),
```
wpoely86: okay, this answers my previous question. Time for a new one: why the change?

easybuild/framework/easyblock.py
```diff
@@ -1978,6 +2150,10 @@ def build_and_install_one(ecdict, init_env):
 
     try:
         app_class = get_easyblock_class(easyblock, name=name)
+
+        if dry_run:
+            print_dry_run_warning('below', silent=silent)
```
wpoely86: please add a comment about the why here

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1027,9 +1043,12 @@ def robot_find_easyconfig(name, version):
             if os.path.isfile(easyconfig_path):
                 _log.debug("Found easyconfig file for name %s, version %s at %s" % (name, version, easyconfig_path))
                 _easyconfig_files_cache[key] = os.path.abspath(easyconfig_path)
-                return _easyconfig_files_cache[key]
+                res = _easyconfig_files_cache[key]
```
wpoely86: Why this change?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1014,12 +1028,14 @@ def robot_find_easyconfig(name, version):
     if key in _easyconfig_files_cache:
         _log.debug("Obtained easyconfig path from cache for %s: %s" % (key, _easyconfig_files_cache[key]))
         return _easyconfig_files_cache[key]
+
     paths = build_option('robot_path')
-    if not paths:
-        raise EasyBuildError("No robot path specified, which is required when looking for easyconfigs (use --robot)")
```
wpoely86: why is this error gone?

easybuild/tools/environment.py
```diff
@@ -93,6 +94,12 @@ def setvar(key, value):
     _changes[key] = value
     _log.info("Environment variable %s set to %s (%s)", key, value, oldval_info)
 
+    if verbose and build_option('extended_dry_run'):
```
wpoely86: why no `verbose` when not in dry run?

easybuild/tools/filetools.py
```diff
@@ -154,22 +157,28 @@ def read_file(path, log_error=True):
...
     """Write given contents to file at given path (overwrites current file contents!)."""
-    f = None
+
+    # early exit in 'dry run' mode
+    if not forced and build_option('extended_dry_run'):
+        dry_run_msg("file written: %s" % path, silent=build_option('silent'))
+        return
+
+    handle = None
     # note: we can't use try-except-finally, because Python 2.4 doesn't support it as a single block
```
wpoely86: but now we can!

easybuild/tools/filetools.py
```diff
@@ -604,33 +613,41 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
+        # skip checking of files in dry run mode
+        patch_filename = os.path.basename(patch_file)
+        dry_run_msg("* applying patch file %s" % patch_filename, silent=build_option('silent'))
+
+    elif not os.path.isfile(patch_file):
         raise EasyBuildError("Can't find patch %s: no such file", patch_file)
-        return
 
-    if fn and not os.path.isfile(fn):
+    elif fn and not os.path.isfile(fn):
```
wpoely86: sure this needs to be `elif`?  They all need to be checked?

easybuild/tools/filetools.py
```diff
@@ -604,33 +613,41 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
-        except IOError, err:
-            raise EasyBuildError("Failed to copy %s to dir %s: %s", patch_file, dest, err)
-            return
+        if build_option('extended_dry_run'):
+            dry_run_msg("  %s copied to %s" % (patch_file, dest), silent=build_option('silent'))
+        else:
+            try:
+                shutil.copy2(patch_file, dest)
+                _log.debug("Copied patch %s to dir %s" % (patch_file, dest))
+                return 'ok'
```
wpoely86: say what?

easybuild/tools/filetools.py
```diff
@@ -604,33 +613,41 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
+            except IOError, err:
+                raise EasyBuildError("Failed to copy %s to dir %s: %s", patch_file, dest, err)
 
     # use absolute paths
     apatch = os.path.abspath(patch_file)
     adest = os.path.abspath(dest)
 
-    if not level:
+    patch_level = level
```
wpoely86: why is this?

easybuild/tools/filetools.py
```diff
@@ -604,33 +613,41 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
+    if patch_level is None and build_option('extended_dry_run'):
+        patch_level = '<derived>'
```
wpoely86: shouldn't we simply skip the actual calling of `patch`?

easybuild/tools/filetools.py
```diff
@@ -658,17 +674,32 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
     result = run.run_cmd(patch_cmd, simple=True)
     if not result:
         raise EasyBuildError("Patching with patch %s failed", patch_file)
-        return
 
     return result
 
 
+def apply_regex_substitutions(path, regex_subs):
+    """Apply specified list of regex substitutions."""
```
wpoely86: document the arguments

easybuild/tools/filetools.py
```diff
@@ -658,17 +674,32 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
+
+    # early exit in 'dry run' mode
+    if build_option('extended_dry_run'):
+        dry_run_msg("patching file %s" % path, silent=build_option('silent'))
```
wpoely86: maybe change to `applying regex...` to clearly differentiate between patching from the easyconfig and the easyblock.

easybuild/tools/filetools.py
```diff
@@ -658,17 +674,32 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
+        return
+
+    for i, (regex, subtxt) in enumerate(regex_subs):
+        regex_subs[i] = (re.compile(regex), subtxt)
```
wpoely86: what about specifying `re.M` or such?

easybuild/tools/run.py
```diff
@@ -73,6 +74,22 @@ def run_cmd(cmd, log_ok=True, log_all=False, simple=False, inp=None, regexp=True
     - path is the path run_cmd should chdir to before doing anything
```
wpoely86: all options should be documented

easybuild/tools/config.py
```diff
@@ -517,46 +521,3 @@ def module_classes():
 def read_environment(env_vars, strict=False):
     """NO LONGER SUPPORTED: use read_environment from easybuild.tools.environment instead"""
     _log.nosupport("read_environment has moved to easybuild.tools.environment", '2.0')
-
-
-def set_tmpdir(tmpdir=None, raise_error=False):
```
wpoely86: shouldn't we do a deprecated warning thingie? As good policy?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -363,79 +378,133 @@ def is_dep_in_toolchain_module(self, name):
         """Check whether a specific software name is listed as a dependency in the module for this toolchain."""
         return any(map(lambda m: self.mns.is_short_modname_for(m, name), self.toolchain_dep_mods))
 
-    def _prepare_dependency_external_module(self, dep):
+    def _simulated_load_dependency_module(self, name, version, metadata):
         """Set environment variables picked up by utility functions for dependencies specified as external modules."""
```
wpoely86: document the args

easybuild/framework/easyblock.py
boegel: your rephrasing suggestion makes less sense to me if I read it out loud than what I have now...

I'll print the module name as well, you're right

easybuild/framework/easyblock.py
boegel: no, the 1st `None` is important, it basically results in using `self.name.lower()` as default command (see below)

easybuild/framework/easyblock.py
boegel: to avoid duplicate code? ;)

will check, this could make the output of the dry run sanity_check_step clearer, indeed

easybuild/framework/easyblock.py
boegel: the method as passed by 'name', so `()` is needed to actually *call* the method, rather than just referring to it, I'll add a comment to clarify that

easybuild/framework/easyblock.py
boegel: by returning the method itself, I can access `.func_code.co_names`

easybuild/framework/easyconfig/easyconfig.py
boegel: serves no use, `robot_path` is always defined via `main()`

easybuild/framework/easyconfig/easyconfig.py
boegel: mostly style, returning somewhere 'deep' in the function results in code that's harder to read or understand at a glance (it's not immediately clear if there are other `return`s for example)

easybuild/tools/environment.py
boegel: the `verbose` indicates whether or not defining this environment variables should be printed under `-x` or not

in some places, defining the environment variables is considered 'internal', i.e. it serves a purposes but is mostly irrelevant so shouldn't 'pollute' the dry run output

example: all the `$EBVAR*` environment variables (which we should get rid of), there are others too

boegel: I'll clarify this by updating the `setvar` docstring

easybuild/tools/filetools.py
boegel: well, using `with` (context manager) avoids the need for `try-except-finally` all together, so let me use that instead :)

easybuild/tools/filetools.py
boegel: well, it's just cleaner; if the `elif` above passes, we'll `raise` anyway

easybuild/tools/filetools.py
boegel: hehe, should be `return True`, will fix

easybuild/tools/filetools.py
boegel: there was a `patch_level = level` deeper down, so I moved it up

it is a bit weird though, maybe sticking to `level` is clearer indeed, will fix

easybuild/tools/filetools.py
boegel: that's skipped already (`run_cmd` is modified to not execute any commands unless forced)

trying to derive the patch level should never be tried though, since that requires the patch file to be available, and to find/read it, etc.

easybuild/tools/filetools.py
boegel: this function currently only supports patching by lines, so no multiline patching support yet (it's rather rare to require that at runtime)

we can enhance this function later to support that

easybuild/tools/config.py
boegel: this is internal to framework, if someone is using this in an easyblock, they're doing something weird or something they shouldn't be doing ;)

easybuild/framework/easyblock.py
wpoely86: I was more thing in the case of, what if command is just a string?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -363,79 +378,139 @@ def is_dep_in_toolchain_module(self, name):
...
+            if tcmods_exists:
+                self.modules_tool.load([tc_mod], silent=silent)
+                dry_run_msg("module load %s" % tc_mod, silent=silent)
+            else:
+                dry_run_msg("module load %s [SIMULATED]" % tc_mod, silent=silent)
+                # provide fake prefix so $EBROOT* gets defined
+                fake_prefix = os.path.join(fake_root, tc_mod)
+                self._simulated_load_dependency_module(self.name, self.version, {'prefix': fake_prefix})
+
+                # also simulate loads for toolchain dependencies, if required information is available
```
wpoely86: shouldn't simulated the load of the toolchain deps before loading the toolchain itself?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -363,79 +378,139 @@ def is_dep_in_toolchain_module(self, name):
...
+                    # provide fake prefix so $EBROOT* gets defined
+                    fake_prefix = os.path.join(fake_root, mod_name)
+                    self._simulated_load_dependency_module(dep['name'], dep['version'], {'prefix': fake_prefix})
+        else:
+            # load modules for all dependencies
+            dep_mods = [dep['short_mod_name'] for dep in self.dependencies]
+            self.log.debug("Loading modules for dependencies: %s" % dep_mods)
+            self.modules_tool.load(dep_mods, silent=silent)
 
         # define $EBROOT* and $EBVERSION* for external modules, if metadata is available
```
wpoely86: this comment is no longer accurate, they are already loaded?

wpoely86: a no, just a strange diff here.

test/framework/options.py
```diff
@@ -1898,6 +1904,75 @@ def test_review_pr(self):
...
+        self.mock_stdout(False)
+        self.assertEqual(len(stdout), 0)
+
+        msg_regexs = [
+            re.compile(r"the actual build \& install procedure that will be performed may diverge", re.M),
+            re.compile(r"^\*\*\* DRY RUN using 'EB_toy' easyblock", re.M),
+            re.compile(r"^== COMPLETED: Installation ended successfully", re.M),
+        ]
+        ignoring_error_regex = re.compile(r"!!! WARNING !!! ignoring error", re.M)
+        ignored_error_regex = re.compile(r"!!! WARNING !!! One or more errors were ignored, see warnings above", re.M)
```
boegel: hmm, this is broken

easybuild/framework/easyblock.py
```diff
@@ -246,6 +260,20 @@ def close_log(self):
...
 
     #
+    # INIT UTILITIES
+    #
+    def init_dry_run(self):
+        """Initialise easyblock instance for performing a dry run."""
+
+        # replace build/install dir
+        self.builddir = os.path.join(tempfile.gettempdir(), '__ROOT__', self.builddir.lstrip(os.path.sep))
+        dry_run_msg("Using fake build directory: %s" % self.builddir, silent=self.silent)
```
wpoely86: We should display this more clearly: keep using a 'fake' build and install dir but displaying them nicely like: `$EASYBUILD_BUILDDIR/...`.

wpoely86: Like @rjeschmi asked

easybuild/tools/toolchain/toolchain.py
```diff
@@ -363,79 +378,139 @@ def is_dep_in_toolchain_module(self, name):
...
 
-    def _prepare_dependencies(self):
+    def _load_toolchain_module(self, silent=False):
+        """Load toolchain module."""
+
+        tc_mod = self.det_short_module_name()
+
+        if self.dry_run:
+
+            fake_root = os.path.join(tempfile.gettempdir(), '__fake__')
```
wpoely86: Using both `__root__` and `__fake__` is confusing. Let's show the `$EASYBUILD_INSTALLPATH_MODULES`, as @tbrown-curc asked

easybuild/framework/easyblock.py
boegel: Huh, ok, that makes a lot of sense, good suggestion. :-)

easybuild/tools/toolchain/toolchain.py
boegel: yep, that's better, will change

easybuild/tools/toolchain/toolchain.py
boegel: I think the idea was to show `$EBROOTFOO` instead.

easybuild/framework/easyblock.py
boegel: In every dry run message being printed, the fake build/install dir is now replaced with the original build/install dir, so no more `__ROOT__` in the output, but the build/install dir is still replaced as was the case here to make sure only paths in the tmpdir are used/created/touched

boegel:
@wpoely86, @rjeschmi, @tbrown-curc: remarks w.r.t. `__ROOT__` and `__fake__` paths used in output fixed, updated the example in the PR description and at http://boegel-eb.readthedocs.org/en/extended_dry_run/Extended_dry_run_examples.html, looks a lot better now!

boegel:
I gave this another full review myself, and added some final minor tweaks.

Since the documentation is also update (see https://github.com/hpcugent/easybuild/pull/159), I feel this is ready to merge as soon as Jenkins gives the green light.

Thanks to everyone who contributed to this in one way or another, in my view this is going to be a killer feature.

PR #1389, opened 2015-09-14
add support for specifying alternate name to be part of generated module name (REVIEW)

boegel:


boegel:
@rjeschmi: please review (since this was your idea ;-))

easybuild/framework/easyconfig/default.py
```diff
@@ -155,12 +155,13 @@
     'exts_list': [[], 'List with extensions added to the base installation', EXTENSIONS],
 
     # MODULES easyconfig parameters
+    'modaliases': [{}, "Aliases to be defined in module file", MODULES],
     'modextrapaths': [{}, "Extra paths to be prepended in module file", MODULES],
     'modextravars': [{}, "Extra environment variables to be added to module file", MODULES],
     'modloadmsg': [{}, "Message that should be printed when generated module is loaded", MODULES],
     'modluafooter': ["", "Footer to include in generated module file (Lua syntax)", MODULES],
+    'modname': [None, "Module name to use (rather than using software name", MODULES],
```
boegel: not very happy with this parameter name, since it's not clear this only covers part of the full module name (only the bit that is now covered by `name`)

maybe it should be `name_in_module`?

@rjeschmi, @wpoely86: suggestions?

test/framework/easyconfig.py
```diff
@@ -1491,6 +1491,17 @@ def test_find_related_easyconfigs(self):
...
+        """Test specifying an alternative name for the software name, to use when determining module name."""
+        ec_file = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'toy-0.0-deps.eb')
+        ectxt = read_file(ec_file)
+        modified_ec_file = os.path.join(self.test_prefix, os.path.basename(ec_file))
+        write_file(modified_ec_file, ectxt + "\nmodname = 'notreallyatoy'")
+        ec = EasyConfig(modified_ec_file)
+        self.assertEqual(ec.full_mod_name, 'notreallyatoy/0.0-deps')
+        self.assertEqual(ec.short_mod_name, 'notreallyatoy/0.0-deps')
+        self.assertEqual(ec['name'], 'toy')
+
```
boegel: This test should be extended: make sure that the alternate name is actually used for the generated module file, check module mentioning in `--dry-run` output, etc.

easybuild/framework/easyconfig/default.py
rjeschmi: It doesn't seem too bad. It is the "name" in the modulefile, so it is accurate.

rjeschmi:
looks fine

easybuild/framework/easyconfig/default.py
boegel: we use 'module name' in different places to refer to the whole thing, I'd like to make it crystal clear this is not the 'full' module name, just the software name part of it...

rjeschmi: modmetaname?

IE it is the metadata for the module

boegel: @rjeschmi: quoting you: "It is the "name" in the modulefile", so `name_in_module` is almost a one-to-one match?

'meta' is so... meta. ;-)

rjeschmi: I _don't_ like _ It also makes the option reorder somewhere else alphabetically.

pescobar: modaltsoftname ?

modaltappname ?

boegel: I like `modaltsoftname`: descriptive and long enough so you'll think twice before using it. @rjeschmi?

rjeschmi: Sounds good

boegel:
updated to `modaltsoftname`, seems to work as expected so will merge as soon as Jenkins gives a green light

boegel:
Going in, thanks for the review @rjeschmi!

Issue #1392, opened 2015-09-18
Intel Ifort and FC compiler variable


DirkdeDraak:
On 15/09/15 14:10, Kenneth Hoste  wrote:
My guess is that the reason that the framework currently doesn't define $FC is because it's unclear whether $FC should be set to the value used in $F77 or $F90... 

This is mostly historic though, since I think $F77 and $F90 are now set to the same value for all Fortran compilers we support. 

Maybe open an issue for this via https://github.com/hpcugent/easybuild-framework/issues/new, just so it doesn't get forgotten easily? 

It's a trivial change to also define $FC, but I fear it may come with some fallout (some builds may suddenly start failing with having $FC defined). 

Making a change like this early in the development phase of the next EB version makes sense, since it's likely we'll pick up any issues caused by it well before the 2.4.0 release. 

regards,

Kenneth

On 15/09/15 14:10, Fokko Masselink wrote:
Hi,

I'm building OpenMPI 1.10.0 with EasyBuild against an Intel Toolchain with GCC
4.9.2.
In the configure step gfortran is used as the Fortran compiler, though we want
to use ifort from Intel of course.
Turns out (if I'm not mistaken) that the configure of OpenMPI looks for the FC
environment variable and ignores both F77 and F90.
Can the FC variable be included in the compiler toolchains?

Now I overcome this by adding
preconfigopts = 'FC=$F90 CFLAGS="%s" CXXFLAGS="%s" FCFLAGS="%s" ' % (....)
to the easyconfig file

A very basic manual test also shows this behavior:
$ untar OpenMPI tarball
$ module load intel
$ CC=icc CXX=icpc F90=ifort ./configure
*** Fortran compiler
checking for gfortran... gfortran
checking whether we are using the GNU Fortran compiler... yes
checking whether gfortran accepts -g... yes
checking whether ln -s works... yes
checking if Fortran compiler works... yes

Once you start to specify intel specific FCFLAGS the gfortran compiler will
fail to build software and the configure process is stopped.

If you configure with the following:
$ untar OpenMPI tarball
$ module load intel
$ CC=icc CXX=icpc FC=ifort ./configure
*** Fortran compiler
checking whether we are using the GNU Fortran compiler... no
checking whether ifort accepts -g... yes
checking whether ln -s works... yes
checking if Fortran compiler works... yes

boegel:
fixed with #1394, let's follow up there

PR #1394, opened 2015-09-18
also define $FC and $FCFLAGS in build environment

boegel:
fix for #1392

### DIFF
--- a/easybuild/toolchains/mpi/craympich.py
+++ b/easybuild/toolchains/mpi/craympich.py
@@ -47,6 +47,7 @@ class CrayMPICH(Mpi):
     MPI_COMPILER_MPICXX = CrayPECompiler.COMPILER_CXX
     MPI_COMPILER_MPIF77 = CrayPECompiler.COMPILER_F77
     MPI_COMPILER_MPIF90 = CrayPECompiler.COMPILER_F90
+    MPI_COMPILER_MPIFC = CrayPECompiler.COMPILER_FC
 
     # no MPI wrappers, so no need to specify serial compiler
     MPI_SHARED_OPTION_MAP = {
@@ -54,6 +55,7 @@ class CrayMPICH(Mpi):
         '_opt_MPICXX': '',
         '_opt_MPIF77': '',
         '_opt_MPIF90': '',
+        '_opt_MPIFC': '',
     }
 
     def _set_mpi_compiler_variables(self):
--- a/easybuild/toolchains/mpi/craympich.py
+++ b/easybuild/toolchains/mpi/craympich.py
@@ -31,7 +31,8 @@ MPI support for the Cray Programming Environment (craype).
 from easybuild.toolchains.compiler.craype import CrayPECompiler
 from easybuild.toolchains.mpi.mpich import TC_CONSTANT_MPICH, TC_CONSTANT_MPI_TYPE_MPICH
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.toolchain.constants import COMPILER_VARIABLES, MPI_COMPILER_TEMPLATE, SEQ_COMPILER_TEMPLATE
+from easybuild.tools.toolchain.constants import COMPILER_VARIABLES, MPI_COMPILER_TEMPLATE, MPI_COMPILER_VARIABLES
+from easybuild.tools.toolchain.constants import SEQ_COMPILER_TEMPLATE
 from easybuild.tools.toolchain.mpi import Mpi
 
 
@@ -50,13 +51,7 @@ class CrayMPICH(Mpi):
     MPI_COMPILER_MPIFC = CrayPECompiler.COMPILER_FC
 
     # no MPI wrappers, so no need to specify serial compiler
-    MPI_SHARED_OPTION_MAP = {
-        '_opt_MPICC': '',
-        '_opt_MPICXX': '',
-        '_opt_MPIF77': '',
-        '_opt_MPIF90': '',
-        '_opt_MPIFC': '',
-    }
+    MPI_SHARED_OPTION_MAP = dict([('_opt_%s' % var, '') for var, _ in MPI_COMPILER_VARIABLES])
 
     def _set_mpi_compiler_variables(self):
         """Set the MPI compiler variables"""
--- a/easybuild/toolchains/mpi/mpich.py
+++ b/easybuild/toolchains/mpi/mpich.py
@@ -31,6 +31,7 @@ Support for MPICH as toolchain MPI library.
 @author: Dmitri Gribenko (National Technical University of Ukraine "KPI")
 """
 
+from easybuild.tools.toolchain.constants import COMPILER_VARIABLES, MPI_COMPILER_VARIABLES
 from easybuild.tools.toolchain.mpi import Mpi
 from easybuild.tools.toolchain.variables import CommandFlagList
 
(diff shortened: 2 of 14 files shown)


boegel:
@geimer: can you take a look at this before I trim the edges?

Should we also define `$F77FLAGS`, while we're at it?

boegel:
Jenkins: test this please

geimer:
@boegel: I can take a look, however, this definitely won't happen before next week. And even this may be wishful thinking...

geimer:
@boegel Regarding `$F77FLAGS`: GNU autotools will use `$FFLAGS` for `$F77` and `$FCFLAGS` for `$FC`.  The `$F90` variable and the associated `$F90FLAGS` are somewhat non-standard (if one considers autotools to be standard...) and custom project-specific extensions. I've never seen `$F77FLAGS` anywhere, though that doesn't mean that it isn't used.

easybuild/tools/toolchain/compiler.py
```diff
@@ -237,21 +238,15 @@ def _set_compiler_flags(self):
...
-
-        self.variables.nappend('F90FLAGS', flags)
-        self.variables.nappend('F90FLAGS', fflags)
-        self.variables.join('F90FLAGS', 'OPTFLAGS', 'PRECFLAGS')
+        for var in ['CFLAGS', 'CXXFLAGS']:
+            self.variables.nappend(var, flags)
+            self.variables.nappend(var, cflags)
+            self.variables.join(var, 'OPTFLAGS', 'PRECFLAGS')
+
+        for var in ['FFLAGS', 'F90FLAGS']:  # FIXME F77FLAGS?
```
boegel: @geimer: so, we should also define `$FCFLAGS` and `$F77FLAGS`?

geimer:
@boegel: `$FCFLAGS` for sure.  As I said, I personally haven't seen `$F77FLAGS` in the wild, but others may know more. At this point I would be reluctant to also add `$F77FLAGS`.

boegel:
OK, I guess we'll stick to just `$FCFLAGS` then, thanks for your input @geimer! I'll try and finish this up in time for inclusion in EB v2.4.0

easybuild/toolchains/mpi/mpich.py
```diff
@@ -46,19 +47,18 @@ class Mpich(Mpi):
 
     MPI_LIBRARY_NAME = 'mpich'
 
+    # FIXME version-dependent? see http://www.mpich.org/static/docs/v3.1.4/www1/mpifort.html
+    MPI_COMPILER_MPIFC = 'mpif90'
+    #MPI_COMPILER_MPIFC = 'mpifort'
```
boegel: @geimer: it seems like MPICH switched from `mpif90` to `mpifort` in v3.1.2? Any idea about MPICH2?

geimer: That's right. They suddenly switch the name in a **patch release**! This also makes the sanity check fail when using a v3.1.1 easyconfig with `--try-software-version`. With respect to MPICH2, there is no need to worry as the entire release series provided `mpif90`.

Note that the `mpifort` thing also applies to recent versions of Open MPI (see https://www.open-mpi.org/faq/?category=mpi-apps#mpifort-vs-mpif77-and-mpif90), though they automatically create symlinks to `mpif77` and `mpif90` for backwards compatibility.

boegel:
Summary of the changes made in here (since quite a bit of cleanup was done too):

* also define `$FC` (same as `$F90`) in build environment (same for `$MPIFC`)
* also define `$FCFLAGS` (same as `$F90FLAGS)` in build environment
* define `$MPICF*` for MPICH depending on MPICH version
  * `mpif77`/`mpifort` for MPICH v3.1.2 and newer
  * `mpif77`/`mpif90` for older MPCIH versions
* define `$MPICF*` for OpenMPI depending on OpenMPI version
  * `mpifort` for OpenMPI v1.7 and newer
  * `mpif77`/`mpif90` for older OpenMPI versions

@geimer: how does this sound?

easybuild/toolchains/mpi/craympich.py
```diff
@@ -54,6 +55,7 @@ class CrayMPICH(Mpi):
         '_opt_MPICXX': '',
         '_opt_MPIF77': '',
         '_opt_MPIF90': '',
+        '_opt_MPIFC': '',
```
geimer: See `easybuild/toolchains/mpi/mpich.py` below: There you use a dict and `MPI_COMPILER_VARIABLES`. Consistency would be nice and using a dict is probably less error prone.

geimer: Forget it.  In `mpich.py` it's defined at runtime...

easybuild/tools/toolchain/constants.py
```diff
@@ -39,11 +39,13 @@
...
     ('F77', 'Fortran 77 compiler'),
     ('F90', 'Fortran 90 compiler'),
+    ('FC', 'Fortran compiler'),
 ]
 
 COMPILER_FLAGS = [
     ('CFLAGS', 'C compiler flags'),
     ('CXXFLAGS', 'C++ compiler flags'),
+    ('FCFLAGS', 'Fortran compiler flags'),
     ('FFLAGS', 'Fortran compiler flags'),
```
geimer: FFLAGS = Fortran 77 compiler flags?

easybuild/tools/toolchain/constants.py
```diff
@@ -72,10 +74,12 @@
         ('CUDA_CXX', 'CUDA C++ compiler command'),
         ('CUDA_F77', 'CUDA Fortran77 compiler command'),
         ('CUDA_F90', 'CUDA Fortran90 compiler command'),
```
geimer: CUDA Fortran 77/90 compiler command (space after 'Fortran')?

easybuild/tools/toolchain/constants.py
```diff
@@ -72,10 +74,12 @@
...
+        ('CUDA_FC', 'CUDA Fortran compiler command'),
     ],
     FlagList: [
         ('CUDA_CFLAGS', 'CUDA C compiler flags'),
         ('CUDA_CXXFLAGS', 'CUDA C++ compiler flags'),
+        ('CUDA_FCFLAGS', 'CUDA Fortran compiler flags'),
         ('CUDA_FFLAGS', 'CUDA Fortran compiler flags'),
```
geimer: CUDA Fortran 77 compiler flags?

test/framework/toolchain.py
```diff
@@ -189,7 +169,7 @@ def test_validate_pass_by_value(self):
     def test_optimization_flags(self):
         """Test whether optimization flags are being set correctly."""
 
-        flag_vars = ['CFLAGS', 'CXXFLAGS', 'FFLAGS', 'F90FLAGS']
+        flag_vars = ['CFLAGS', 'CXXFLAGS', 'FCFLAGS', 'FFLAGS', 'F90FLAGS']
```
geimer: It's only a test so it doesn't really matter, but at basically all other places you used the order F77, F90, FC. Maybe adopt it here (and 6 more times below) as well?

geimer:
@boegel: Visually looks good to me; my comments are really nitpicking ;-)  I've also tried to build Open MPI 1.10.0 using the Intel compilers on top of this PR w/o having FC set in `configopts` (like in https://github.com/hpcugent/easybuild-framework/issues/1392) and the build worked out fine.

easybuild/toolchains/mpi/craympich.py
boegel: No, I think this is a valid remark.

test/framework/toolchain.py
boegel: Like you said, it doesn't really matter. In fact, the order here matches the order of the `COMPILER_FLAGS` list.

boegel:
Going in, thanks for the review @geimer!

PR #1395, opened 2015-09-18
add --copy-easyconfigs and --copy-patches configure options to specify locations to copy easyconfigs/patches to (WIP)

boegel:


boegel:
@ocaisa, @geimer: this is clearly WIP, since the options aren't taken into account yet anywhere, but I was wondering whether this is what you had in mind

Basically, it allows you to specify something like `--copy-patches=installdir,repository,sourcedir`.

I'm not sure whether:

* there should be an extra possible value: any (existing) filepath to copy easyconfigs/patches to (e.g., `$HOME/mystuff`)
* whether also supporting `--copy-logs` makes sense (maybe it does, when combined with the above)

boegel:
Jenkins: test this please

ocaisa:
I've been trying to see if/how we can also store the inheritance tree for the easyblock as well to ensure reproducability but my expertise is in googling not python :P

Would it work to use `inspect.mro` on the easyblock class to be used with `inspect.getfile` on the result to store the entire class history for the easyblock (maybe in a subdir of the installation and maybe as an option)?

ocaisa:
And to be really pedantic...what about automatically adding an md5sum check for the sources to the archived easyconfig to ensure you have the same sources.

boegel:
never got merged, and I managed to overwrite the existing branch, so closing this...

PR #1399, opened 2015-09-22
fix issues to make EB framework installed as zipped egg work (WIP)

### LINKED ISSUE #1369: make EB installation as zipped eggs work
The major advantage of this is significant speedup when EasyBuild is installed on a shared filesystem.

The major problem with this is that there's no direct access to the Python modules or easyconfig files anymore as files.
Via `pkg_resources.resource_string` for example, the contents of a specific Python module or data file (like easyconfig files) can be retrieved from a zipped egg, see https://pythonhosted.org/setuptools/pkg_resources.html#resourcemanager-api .

It also introduces a dependency on `setuptools`, since `pkg_resources` is a part of `setuptools`.

However, since the performance impact of this is potentially huge (especially for things like `--search`), this may be worth the trouble...

boegel:
fixes #1369

### DIFF
--- a/easybuild/tools/utilities.py
+++ b/easybuild/tools/utilities.py
@@ -29,11 +29,11 @@ Module with various utility functions
 """
 import glob
 import os
+import pkgutil
 import string
 import sys
 from vsc.utils import fancylogger
 
-import easybuild.tools.environment as env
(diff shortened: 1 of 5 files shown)


boegel:
Here's a good reason to make this work: https://github.com/mbachry/exxo/

Issue #196, opened 2012-08-29
add a dynamic libraries section to sanityCheckPaths


boegel:
(old internal ticket 262)

Next to the obligatory `files` and `dirs` keys in the `sanityCheckPaths` dictionary, we should also support on optional `dynlibs` key.

This would support specifying the names on the dynamic libraries we expect to find (without extension), making it more portable. On Linux, dynamic libraries end in `.so`, on OS X they end in `.dylib`.

Micket:
We deal with .so vs .dylib via the SHLIB_EXT variable now in lots of places, and rather than checking bits on binaries I think sanity_check_commands serves this purpose well, so I don't think there is anything more that needs solving here.

PR #1405, opened 2015-09-28
[SUGGESTION] patch sanity_check_paths['files'] to look in lib/ and lib64/

### LINKED ISSUE #1404: provide generic way for sanity_check to look in both `lib` and `lib64` for library files
AFAIK the only way of checking whether library files ended up in `lib/` or `lib64/` is via

``` .py
sanity_check_paths = {
  'files': [
    ('lib/myLib.so', 'lib64/myLib.so'),
    ('lib/myLib.a', 'lib64/myLib.a')
  ]
}
```

Wouldn't it be nice having this collapsing into something like

``` .py
sanity_check_paths = {
  'files': [
    LibraryPathBuilder('myLib', shared=True, static=True)
  ]
}
```

It's less error prone to typos and way less tedious.

torbjoernk:
This is a suggestion to fix #1404.
Don't know whether it breaks something elsewhere ...

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1578,7 +1578,10 @@ class EasyBlock(object):
         for key, check_fn in path_keys_and_check.items():
             for xs in paths[key]:
                 if isinstance(xs, basestring):
-                    xs = (xs,)
+                    if xs.startswith('lib/'):
+                        xs = (xs, xs.replace('lib/', 'lib64/'))
+                    else:
+                        xs = (xs,)
                 elif not isinstance(xs, tuple):
                     raise EasyBuildError("Unsupported type '%s' encountered in %s, not a string or tuple",
                                          key, type(xs))


boegel:
Jenkins: ok to test

boegel:
@torbjoernk: this is an interesting idea... Did @berndmohr put you up to this? ;-)

I had something more elaborate in mind, i.e. add support for additional keys like `libs` or `libraries` to be specified in `sanity_check_paths`, which then enables us to check both in `lib` and `lib64` automagically; this has been an idea for a *long* time (see #196), but it has never materialised.

Nevertheless, I like this approach, since it's a quick-and-easy fix for something that has annoyed a lot of people (mostly on SuSE and derivatives), and shouldn't do any harm. We already consider both `lib` and `lib64` for `$LD_LIBRARY_PATH` and `$LIBRARY_PATH`, for example.

@wpoely86, @rjeschmi, @fgeorgatos, @pescobar, @ocaisa, @JensTimmerman: thoughts?

wpoely86:
I feel more for the #196 approach... It cannot be that hard to implement?

boegel:
@wpoely86: I don't feel this excludes the need for #196. It's a quick and imho harmless fix to a real problem. Implementing #196 would still require updating *a lot* of easyconfig files...

wpoely86:
#196 can perfectly be done in a backwards compatibly way. Adding a libraries section doesn't invalidate the current `files`.

boegel:
@wpoely86: sure, but introduces a `libraries` key is only adding support for a fix (which requires reworking lots of easyconfig files), while this trick is an actual fix for the problem at hand (and doesn't rule out working in `libraries` in the coming future)

akesandgren:
Is this still valid or should it be closed?
If still valid the code needs a slight adjustment, don't use "/", use os.pathsep (or whatever it is called)
And it should go both ways...

boegel:
@akesandgren This is no longer relevant, so I'll go ahead and close it.

Fallback to also consider `lib64` for `lib` paths (and vice versa) was implemented in #2477 (and enabled by default, with a switch to disable it) .

This PR did serve as inspiration for that change though, so I'd like to thank @torbjoernk for this...

PR #1408, opened 2015-09-30
[.yeb] initial support for parsing easyconfig files in YAML syntax (REVIEW)

Caylo:


boegel:
Jenkins: ok to test

boegel:
Jenkins: test this please

easybuild/framework/easyconfig/parser.py
```diff
@@ -63,7 +63,10 @@ def fetch_parameters_from_easyconfig(rawtxt, params):
     """
     param_values = []
     for param in params:
-        regex = re.compile(r"^\s*%s\s*=\s*(?P<param>\S.*?)\s*$" % param, re.M)
+        if eb_format == '.yeb':
+            regex = re.compile(r"^\s*%s\s*: \s*(?P<param>\S.*?)\s*$" % param, re.M)
+        else:
+            regex = re.compile(r"^\s*%s\s*=\s*(?P<param>\S.*?)\s*$" % param, re.M)
```
boegel: these are now *very* close to each other, basically it's only `=` vs `:`

also, I think in YAML `=` can also be used?

so, rather than making this regex specific to the format in which the easyconfig is written in, I would just make it a tiny bit more generic, and accept both `=` and `:` as 'assignment' operators, in a single regex:

```
regex = re.compile(r"^\s*%s\s*[=:]\s*(?P<param>\S.*?)\s*$" % param, re.M)
```

also, when touching an existing function like `fetch_parameters_from_easyconfig`, check whether there's a unit test for it, and:

i) if there is, enhance it to also check for the new behaviour
ii) if not, add one, and try to cover all things it should be able to deal with

easybuild/framework/easyconfig/parser.py
```diff
@@ -89,7 +92,8 @@ def __init__(self, filename=None, format_version=None, rawcontent=None):
 
         self.format_version = format_version
         self._formatter = None
-
+        if filename and os.path.splitext(filename)[-1] == '.yeb':
```
boegel: don't use `.yeb` all over the place, introduce a constant for it: `YEB_EXTENSION`

and here specifically, I would change this to:

```python
if filename and is_yeb_format(filename):
    ....
```

also, maybe this should move to `_set_formatter`? and we make need a way of detecting whether the `rawcontent` being passed is `.yeb` or `.eb` format (not sure how easy/useful/necessary that is)

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def dump(self, ecfg, default_values, templ_const, templ_val):
+        #TODO
+        pass
+
+    def extract_comments(self,txt):
+        #TODO
+        pass
+
+def retrieve_blocks_in_spec(spec, only_blocks, silent=False):
```
boegel: this whole thing is probably no longer relevant/useful for `.yeb` easyconfigs...

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        f = read_file(self.filename)
+        return yaml.load(f)
+
+    def parse(self, txt):
+        """
+        Pre-process txt to extract header, docstring and pyheader, with non-indented section markers enforced.
+        """
+        #TODO
+        pass
```
boegel: `yaml.load` should go here?

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -0,0 +1,165 @@
...
+    def validate(self):
+        """Format validation"""
+        # TODO for YAML
+
+    def get_config_dict(self):
+        """
+        Return parsed easyconfig as a dictionary, based on specified arguments.
+        """
+        f = read_file(self.filename)
+        return yaml.load(f)
```
boegel: this should simply return `self.parsed_yeb` (which is set in `parse`)?

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -0,0 +1,165 @@
...
+    """Support for easyconfig YAML format"""
+    USABLE = True
+
+    def __init__(self, filename):
+        self.filename = filename
+        super(FormatYeb, self).__init__()
+
+    def validate(self):
+        """Format validation"""
+        # TODO for YAML
```
boegel: just return `True` for now, add a `log.info` statement making it clear that there's no real validation (yet)?

setup.py
```diff
@@ -107,5 +107,5 @@ def find_rel_test():
     provides=["eb"] + easybuild_packages,
     test_suite="test.framework.suite",
     zip_safe=False,
-    install_requires=["vsc-base >= 2.2.4"],
+    install_requires=["vsc-base >= 2.2.4", "PyYAML >= 3.11"],
```
boegel: for now, we should probably make `PyYAML` an optional dependency (see http://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies)

that allows us to include whatever support we have for `.yeb` in the next release under `--experimental`, if it can be useful to let people start playing around with it and provide feedback

fgeorgatos: +1.

I happen to play with a another code tree these days which relies on yaml, and basically it proved to be a major pain on systems with python2.6 as a default, since it is routinely just not there (even if, it is easily obtainable in one way or another)

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -0,0 +1,141 @@
...
+        self.filename = filename
+        super(FormatYeb, self).__init__()
+        self.log.experimental("Parsing .yeb easyconfigs")
+
+    def validate(self):
+        """Format validation"""
+        _log.info("yaml format validation isn't implemented (yet) - validate always returns true")
+        return True
+
+    @requires_yaml
```
boegel: move this to where `yaml.*` is used

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def __init__(self, filename):
+        """FormatYeb constructor"""
+        self.filename = filename
+        super(FormatYeb, self).__init__()
+        self.log.experimental("Parsing .yeb easyconfigs")
+
+    def validate(self):
+        """Format validation"""
+        _log.info("yaml format validation isn't implemented (yet) - validate always returns true")
```
boegel: s/yaml/.yeb/

boegel:
Going in, thanks @Caylo!

Issue #1410, opened 2015-10-02
Easybuild on Archlinux


f-koehler:
I am on Archlinux where the default `python` is `python3`. What is the recommended way of installing EasyBuild in such an environment.

I tried to create a `virtualenv` for `python2` but then I get this output from the boostrap script:

```
[[INFO]] Found module command 'lmod' (Lmod), so using it.
[[INFO]] 

+++ STAGE 0: installing distribute via included (patched) distribute_setup.py...


Downloading http://pypi.python.org/packages/source/d/distribute/distribute-0.6.34.tar.gz
Extracting in /tmp/tmpUjz6AO
Now working in /tmp/tmpUjz6AO/distribute-0.6.34
Installing Distribute
Something went wrong during the installation.
See the error message above.
[[ERROR]] Installing distribute which should deliver easy_install failed?
```

This is not very helpful. Calling it explicitly leads a bit further but still fails:
```
[[INFO]] Found module command 'lmod' (Lmod), so using it.
[[INFO]] 

+++ STAGE 0: installing distribute via included (patched) distribute_setup.py...


Downloading http://pypi.python.org/packages/source/d/distribute/distribute-0.6.34.tar.gz
Extracting in /tmp/tmpkEgqdD
Now working in /tmp/tmpkEgqdD/distribute-0.6.34
Installing Distribute
[[INFO]] 

+++ STAGE 1: installing EasyBuild in temporary dir with easy_install...


[[INFO]] installing EasyBuild with 'easy_install --quiet --upgrade --prefix=/tmp/tmpb8cBVa/eb_stage1 easybuild'
Installing with setuptools.setup...
Installing version 2.3.0
warning: install_lib: 'build/lib' does not exist -- no Python modules to install

zip_safe flag not set; analyzing archive contents...
Installing with setuptools.setup...
Installing version 2.3.0 (API version 2)
Installing with setuptools.setup...
Installing version 2.3.0 (required versions: API >= 2)
Installing with setuptools.setup...
Installing version 2.3.0 (required versions: API >= 2, easyblocks >= 2.3)
warning: install_lib: 'build/lib' does not exist -- no Python modules to install

[[ERROR]] Stage 1 failed, could not determine EasyBuild version (txt:   File "<string>", line 1
    from easybuild.tools.version import this_is_easybuild; print this_is_easybuild()
                                                                                 ^
SyntaxError: invalid syntax
).
```
[...]

PR #1411, opened 2015-10-02
fix issues with 'python' being 'python3' rather than 'python2'

boegel:
cfr. #1410

### DIFF
--- a/easybuild/tools/toolchain/__init__.py
+++ b/easybuild/tools/toolchain/__init__.py
@@ -50,6 +50,10 @@ which contains toolchain related modules.
 @author: Stijn De Weirdt (Ghent University)
 @author: Kenneth Hoste (Ghent University)
 """
+from pkgutil import extend_path
+
+# we're not the only ones in this namespace
+__path__ = extend_path(__path__, __name__)  #@ReservedAssignment
 
 # name/version for dummy toolchain
 # if name==DUMMY_TOOLCHAIN_NAME and version==DUMMY_TOOLCHAIN_VERSION, do not load dependencies
--- a/eb
+++ b/eb
@@ -39,15 +39,25 @@
 # @author: Pieter De Baets (Ghent University)
 # @author: Jens Timmerman (Ghent University)
 
+PYTHON=python
+
 # Python 2.6 or more recent 2.x required
 REQ_MAJ_PYVER=2
 REQ_MIN_PYVER=6
 REQ_PYVER=${REQ_MAJ_PYVER}.${REQ_MIN_PYVER}
 
 # make sure Python version being used is compatible
-pyver=`python -V 2>&1 | cut -f2 -d' '`
+pyver=`$PYTHON -V 2>&1 | cut -f2 -d' '`
 pyver_maj=`echo $pyver | cut -f1 -d'.'`
 pyver_min=`echo $pyver | cut -f2 -d'.'`
+
+if [ $pyver_maj -ne $REQ_MAJ_PYVER ]
+then
+    PYTHON=${PYTHON}$REQ_MAJ_PYVER}
+    pyver=`$PYTHON -V 2>&1 | cut -f2 -d' '`
+    pyver_maj=`echo $pyver | cut -f1 -d'.'`
+    pyver_min=`echo $pyver | cut -f2 -d'.'`
+fi
 
 if [ $pyver_maj -ne $REQ_MAJ_PYVER ]
 then
@@ -60,27 +70,4 @@ then
     exit 2
 fi
 
-
-main_script_base_path="easybuild/main.py"
-python_search_path_cmd="python -c \"import sys; print ' '.join(sys.path)\""
-
-easybuild_main=
-for path in `eval $python_search_path_cmd`;
-do
-    test_path="$path/$main_script_base_path"
-    if [ -f $test_path ]
-    then
-        easybuild_main=$test_path
-        break;
-    fi
-done
-
-if [ -z $easybuild_main ]
-then
-    echo "ERROR: Failed to locate EasyBuild's main script $main_script_base_path" \
-         "in Python's search path, obtained with '$python_search_path_cmd' ."
-    echo "Make sure you PYTHONPATH setting is correct."
-    exit 1
-else
-    python $easybuild_main $@
-fi
+$PYTHON -m easybuild.main $@
--- a/eb
+++ b/eb
@@ -49,14 +49,12 @@ REQ_PYVER=${REQ_MAJ_PYVER}.${REQ_MIN_PYVER}
 # make sure Python version being used is compatible
 pyver=`$PYTHON -V 2>&1 | cut -f2 -d' '`
 pyver_maj=`echo $pyver | cut -f1 -d'.'`
-pyver_min=`echo $pyver | cut -f2 -d'.'`
 
 if [ $pyver_maj -ne $REQ_MAJ_PYVER ]
 then
     PYTHON=${PYTHON}$REQ_MAJ_PYVER}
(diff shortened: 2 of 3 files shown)


boegel:
@wpoely86: please review?

eb
```diff
@@ -39,48 +39,35 @@
...
 REQ_MIN_PYVER=6
 REQ_PYVER=${REQ_MAJ_PYVER}.${REQ_MIN_PYVER}
 
 # make sure Python version being used is compatible
-pyver=`python -V 2>&1 | cut -f2 -d' '`
+pyver=`$PYTHON -V 2>&1 | cut -f2 -d' '`
 pyver_maj=`echo $pyver | cut -f1 -d'.'`
-pyver_min=`echo $pyver | cut -f2 -d'.'`
+
+if [ $pyver_maj -ne $REQ_MAJ_PYVER ]
```
wpoely86: isn't it simpler to simply test if python2 works and use it if so?

eb
```diff
@@ -39,48 +39,35 @@
...
-do
-    test_path="$path/$main_script_base_path"
-    if [ -f $test_path ]
-    then
-        easybuild_main=$test_path
-        break;
-    fi
-done
-
-if [ -z $easybuild_main ]
```
wpoely86: and this is remove because why?

eb
boegel: that would kind of break backwards compatibility... what if `python` is there, but `python2` is not?

eb
boegel: all this crap is no longer needed with Py2.6 and newer, since `-m easybuild.main` just works

eb
wpoely86: Use python2 is present, else python then do the version check once?

boegel: OK, that would work too.

boegel: Well, except that it's different from what we're doing now, and may lead to surprises...

What if `python` (Py2.x) is available in one place, and python2 in another location (and they're not from the same Python installation)?

Right now, we would use `python`, but with your proposed change we would use `python2`...

wpoely86: yes, and it seems to me like the proper thing to do. We want python 2.x.  We shouldn't directly call python and assume it is 2.x. And if you have to different pythons (python and python2), then you're in a world of shit. We cannot account for every oddball case in the world? Most (if not all) distros  have a python2 and a python3 these days.

boegel: Yeah, maybe I'm being overly careful here, you're right I guess.

OK, I'll modify this to check for `python2`, fall back to `python`, and only check the version once.

boegel:
@wpoely86: remarks fixed, please recheck?

wpoely86:
lgtm

boegel:
Hmm, apparently something got broken:

```
$ eb --review-pr 2034
== temporary log file in case of crash /tmp/eb-5ChE_F/easybuild-YArLfV.log
ERROR: Failed to process easyconfig /tmp/eb-5ChE_F/tmp0heXYn/spglib-1.7.4-intel-2015b.eb: import_available_modules: Failed to import easybuild.toolchains.cgmpich: No module named compiler
```

@wpoely86: are you seeing this too on top of this PR?

wpoely86:
Yeah, got it too
```
== 2015-10-06 15:53:00,598 runpy ERROR import_available_modules: Failed to import easybuild.toolchains.cgmpich: No module named compiler (at easybuild/tools/utilities.py:122 in import_available_modules)
== 2015-10-06 15:53:00,670 runpy ERROR Failed to process easyconfig /tmp/eb-HzetzO/tmpQtasCt/spglib-1.7.4-intel-2015b.eb: import_available_modules: Failed to import easybuild.toolchains.cgmpich: No module named compiler (at easybuild/framework/easyconfig/easyconfig.py:943 in process_easyconfig)
ERROR: Failed to process easyconfig /tmp/eb-HzetzO/tmpQtasCt/spglib-1.7.4-intel-2015b.eb: import_available_modules: Failed to import easybuild.toolchains.cgmpich: No module named compiler
```

wpoely86:
Okay, this is weird: `eb -d --review-pr 2034` does work...

wpoely86:
@boegel using `$OURPYTHON` instead of `$PYTHON` fixes it for me

boegel:
Hmm, weird, but I don't see `$PYTHON` mentioned anywhere at https://docs.python.org/2/using/cmdline.html#environment-variables .

Clearly, we need a unit test that tries to do something with `eb`...

boegel:
@wpoely86: apparently, `python easybuild/main.py` vs `python -m easybuild.main` is more different than I think it is...

https://docs.python.org/2/using/cmdline.html should explain why (but I haven't figured it out exactly yet).

@JensTimmerman: do you know the exact difference?

eb
```diff
@@ -44,43 +44,34 @@ REQ_MAJ_PYVER=2
...
-pyver=`python -V 2>&1 | cut -f2 -d' '`
+pyver=`$PYTHON -V 2>&1 | cut -f2 -d' '`
 pyver_maj=`echo $pyver | cut -f1 -d'.'`
 pyver_min=`echo $pyver | cut -f2 -d'.'`
 
 if [ $pyver_maj -ne $REQ_MAJ_PYVER ]
 then
     echo "ERROR: EasyBuild is currently only compatible with Python v${REQ_MAJ_PYVER}.x, found v${pyver}" 1>&2
-    exit 1
+    exit 2
```
JensTimmerman: why do these exit codes have to change? can't you make the new one exit 3?

JensTimmerman:
`-m` will  import the module with `__name__` set to the module name, `python file.py` will import it with `__name__` set to `__main__`
`-m` will also take the first occurence of the module in the pythonpath, `python file.py` will take the file.

boegel:
@JensTimmerman: that's not the actual problem here, it's an issue with `sys.path`...

easybuild/tools/toolchain/__init__.py
```diff
@@ -50,6 +50,10 @@
 @author: Stijn De Weirdt (Ghent University)
 @author: Kenneth Hoste (Ghent University)
 """
+from pkgutil import extend_path
+
+# we're not the only ones in this namespace
+__path__ = extend_path(__path__, __name__)  #@ReservedAssignment
```
boegel: @wpoely86, @JensTimmerman: putting this in place fixes the `import` problem that was caused by switching to `-m easybuild.main` rather than `easybuild/main.py`, but I'm not 100% sure why...

When I added a `print __file__` here, I'm seeing that this module (and `easybuild.tools`) is being imported as `easybuild/tools/toolchain/__init__.pyc`, while others (e.g. `easybuild.toolchain.cgmpich`) that are being imported print an absolute path.
There are other `__init__.py` files that do not have an `extend_path` though (e.g. `easybuild/framework/__init__.py`), they seem to be fine.

This may be related to the calls to the `import_available_modules` function made by the option parser, which is called for `easybuild.tools.repository`, `easybuild.tools.module_naming_scheme`, etc. to figure out the possible values for `--module-naming-scheme` for example.

eb
boegel: I want to preserve 'order', that's all.

boegel:
I'm merging this in, despite not fully understanding why the `extend_path` in `easybuiod/tools/toolchain/__init__.py` is really needed. It shouldn't cause problems adding it, and this is an important fix given the raise of Python 3.x being the default `python`.

eb
JensTimmerman: but you're changing the meaning of the exit codes, scripts that relied on them will start acting weird? 
so it's better to keep the meaning of the codes the same, or just always exit 1 since the meaning can change arbitrarily anyway?

PR #1412, opened 2015-10-06
Support for SLES 12 and kernel 3.12.x

smith1511:
Adding support for SLES 12.

Based on this document (and what I see on the VMs) I've added support for SLES 12 and kernel versions 3.12.x.  SLES 12 kernel versions range from 3.12.28 - .43 for the initial release.

https://www.novell.com/support/kb/doc.php?id=3594951

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -338,6 +338,9 @@ def get_os_version():
                     ('2.6.32', '_SP1'),
                     ('3.0', '_SP2'),
                 ],
+                "12": [
+                    ('3.12', ''),
+                ],
             }
 
             # append suitable suffix to system version
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -336,8 +336,14 @@ def get_os_version():
                 "11": [
                     ('2.6.27', ''),
                     ('2.6.32', '_SP1'),
+                    ('3.0.101-63', '_SP4'),
+                    # not 100% correct, since early SP3 had 3.0.76 - 3.0.93, but close enough?
+                    ('3.0.101', '_SP3'),
+                    # SP2 kernel versions range from 3.0.13 - 3.0.101
                     ('3.0', '_SP2'),
                 ],
+
+                # Once SLES 12 SP1 comes out we'll need to make this stricter
                 "12": [
                     ('3.12', ''),
                 ],


smith1511:
Yes

boegel:
Jenkins: ok to test

smith1511:
Ahh, that's how you do it ;-)

easybuild/tools/systemtools.py
```diff
@@ -338,6 +338,9 @@ def get_os_version():
                     ('2.6.32', '_SP1'),
                     ('3.0', '_SP2'),
```
boegel: Do you mind updating these too, while we're at it, since it's unaware of SP3 or SP4, and `3.0` doesn't necessarily mean SLES 11 SP2.

according to https://wiki.microfocus.com/index.php/Kernel_versions, things got more complicated though...

The kernel versions for SLES 11 SP2 now range from 3.0.13-0.27.1 to 3.0.101-0.7.17.1, while SP3 ranges from 3.0.76-0.11.1 to 3.0.101-0.47.67.2, and SP4 is 3.0.101-63.1.

So, it seems the logic we were using got broken. :-(
So, I'm not sure how to handle this...

Maybe like this (since order matters):

```
"11": [
    ('2.6.27', ''),
    ('2.6.32', '_SP1'),
    ('3.0.101-63', '_SP4'),
    # not 100% correct, since early SP3 had 3.0.76 - 3.0.93, but close enough?
    ('3.0.101', '_SP3'),
    # SP2 kernel versions range from 3.0.13 - 3.0.101
    ('3.0', '_SP2'),
],
```

easybuild/tools/systemtools.py
```diff
@@ -338,6 +338,9 @@ def get_os_version():
...
                 ],
+                "12": [
+                    ('3.12', ''),
```
boegel: good enough for now, but maybe this will have to be made stricter once SLES 12 SP1 is out...

maybe add a comment on that?

boegel:
@smith1511: triggering the bot to test your PR only works for people with 'admin' access to our Jenkins instance (sort of a security thing)

smith1511:
Sorry for the delay.  I've added your suggestions.  Can this be merged now?  Would be great to get this in as it's a blocker on our side.

boegel:
Jenkins: ok to test

boegel:
@smith1511: looks fine now, thanks for coming back to this

I'll get this in as soon as Jenkins is happy (it's not now, but it's not your fault, there's an issue with the tests I'm working on that).

boegel:
Jenkins: test this please

boegel:
Going in, thanks @smith1511!

PR #1413, opened 2015-10-07
Ignore hidden directories when returning the finaldir

wpoely86:

PR #1415, opened 2015-10-07
fix guess_start_dir in case we're already in start_dir

boegel:
follow-up for #1413, since that was a non-backwards compatible change (I realised that the second after hitting the shiny green `Merge` button)

@wpoely86: can you review and maybe look into an enhanced unit test for this?


easybuild/framework/easyblock.py
```diff
@@ -1136,19 +1136,28 @@ def guess_start_dir(self):
...
+                topdir = self.builddir
+
+            abs_start_dir = os.path.join(topdir, start_dir)
+            if topdir.endswith(start_dir) and not os.path.exists(abs_start_dir):
+                self.cfg['start_dir'] = topdir
+            else:
+                if os.path.exists(abs_start_dir):
+                    self.cfg['start_dir'] = abs_start_dir
+                else:
+                    raise EasyBuildError("Specified start dir %s does not exist", abs_start_dir)
```
wpoely86: print `self.cfg['start_dir']` here as that is the dir that is actually specified?

wpoely86:
Can this not be done simpler by letting the extract_step not change the directory but simple setting the `finaldir`?

boegel:
@wpoely86: you're working on a test for this? PR pending?

wpoely86:
@boegel I've sent you a PR with it: boegel/easybuild-framework#10

wpoely86:
Btw, it can not be done simpler. This is probably the best solution.

easybuild/framework/easyblock.py
boegel: no, this error applies to the failing `if os.path.exists(abs_start_dir)`, but I'll include an extra debug log message printing the `start_dir` we end up with

boegel:
@wpoely86: I fleshed out the test you wrote a bit more, please rererereview?

Should be good to go in otherwise, imho.

wpoely86:
looks good

easybuild/tools/filetools.py
```diff
@@ -840,6 +840,31 @@ def expand_glob_paths(glob_paths):
...
+    # strip path1 for use in comparisons
+    path1s = path1.strip(os.path.sep)
+
+    # init part2 head/tail/parts
+    path2_head = path2.strip(os.path.sep)
+    path2_tail = ''
+    path2_parts = path2.split(os.path.sep)
+
+    while path2_parts and not path1s.endswith(path2_head):
+        path2_tail = os.path.join(path2_parts.pop(-1), path2_tail)
```
wpoely86: the `-1` is not needed?

test/framework/filetools.py
```diff
@@ -429,12 +429,26 @@ def test_multidiff(self):
...
+        self.assertEqual(ft.weld_paths('/foo/bar', 'foobar/baz'), '/foo/bar/foobar/baz/')
+        self.assertEqual(ft.weld_paths('foo', 'bar/'), 'foo/bar/')
+        self.assertEqual(ft.weld_paths('foo/', '/bar'), '/bar/')
+
+        # overlap is taken into account
+        self.assertEqual(ft.weld_paths('foo/bar', 'bar/baz'), 'foo/bar/baz/')
+        self.assertEqual(ft.weld_paths('foo/bar/baz', 'bar/baz'), 'foo/bar/baz/')
+        self.assertEqual(ft.weld_paths('foo/bar', 'foo/bar/baz'), 'foo/bar/baz/')
+        self.assertEqual(ft.weld_paths('foo/bar', 'foo/bar'), 'foo/bar/')
+        self.assertEqual(ft.weld_paths('/foo/bar', 'foo/bar'), '/foo/bar/')
```
wpoely86: add two absolute paths with and without overlap?

wpoely86:
lgtm

boegel:

PR #1416, opened 2015-10-07
add and use only_if_module_is_available decorator function to guard functionality that uses optional dependencies

boegel:
This results in removal of some duplicate code, and clean/uniform error messages on missing Python modules, like:

```
ImportError: No module named gc3libs.core; required module 'gc3libs.core' is not available (provided by Python package gc3pie, available from https://pypi.python.org/pypi/gc3pie)
```

### DIFF
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -55,34 +55,28 @@ from easybuild.tools.multidiff import multidiff
 from easybuild.tools.ordereddict import OrderedDict
 from easybuild.tools.run import run_cmd
 from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME
-from easybuild.tools.utilities import quote_str
+from easybuild.tools.utilities import only_if_module_is_available, quote_str
 
 # optional Python packages, these might be missing
 # failing imports are just ignored
 # a NameError should be catched where these are used
 
-# PyGraph (used for generating dependency graphs)
-graph_errors = []
 try:
-    from pygraph.classes.digraph import digraph
-except ImportError, err:
-    graph_errors.append("Failed to import pygraph-core: try easy_install python-graph-core")
-
-try:
+    # PyGraph (used for generating dependency graphs)
+    # https://pypi.python.org/pypi/python-graph-core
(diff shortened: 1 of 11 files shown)


wpoely86:
Isn't this a lot of overhead for things like GitPython, GC3Pie etc? The entire file is not useful is you don't have those modules. Why not check it once for those and be done with it.

boegel:
@wpoely86: remarks fixed, please rereview?

easybuild/tools/repository/gitrepo.py
```diff
@@ -54,7 +55,7 @@
 # failing imports are just ignored
 # a NameError should be catched where these are used
 
-# GitPython
+# GitPython (http://gitorious.org/git-python)
```
wpoely86: why not add this url to the decorator below?

boegel: the decorator will print the PyPI URL

easybuild/framework/easyconfig/tools.py
```diff
@@ -55,34 +55,28 @@
...
     import pygraph.readwrite.dot as dot
-except ImportError, err:
-    graph_errors.append("Failed to import pygraph-dot: try easy_install python-graph-dot")
-
-# graphviz (used for creating dependency graph images)
-try:
+    # graphviz (used for creating dependency graph images)
     sys.path.append('..')
     sys.path.append('/usr/lib/graphviz/python/')
     sys.path.append('/usr/lib64/graphviz/python/')
```
wpoely86: why do we do this?

boegel: because it doesn't work without that, see also http://stackoverflow.com/questions/27302795/how-to-get-past-import-gv-in-python-graph

boegel:
Jenkins: test this please

boegel:
4bce8db should fix the issue with the test suite hanging when testing this PR...

apparently, calling Python 2.7 from a subshell of a Python 2.6 session may lead to weird things, e.g. `import pysvn` segfaulting and resulting in a `run_cmd` call that never returns (which is what was really going on)

boegel:

PR #1417, opened 2015-10-07
give easyblocks the possibility to choose maxhits for run_cmd_qa

molden:
This will give easyblocks the possibility to choose maxhits for run_cmd_qa. This is necessary for the CUDA easyblock fix (to be submitted).

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -227,8 +227,6 @@ def run_cmd_qa(cmd, qa, no_qa=None, log_ok=True, log_all=False, simple=False, re
     else:
         runLog = None
 
-    maxHitCount = maxhits 
-
     try:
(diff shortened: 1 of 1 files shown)


easybuild/tools/run.py
```diff
@@ -227,7 +227,7 @@ def check_answers_list(answers):
     else:
         runLog = None
 
-    maxHitCount = 50
+    maxHitCount = maxhits
```
boegel: do you mind renaming this to `max_hit_count` while you're at it (we're trying to get rid of CamelCase throughout the whole codebase)?

or rather, just use `maxhits` wherever `maxHitCount` is used now, there's no need to define another variable for the same thing (since the value doesn't change afaik?)

boegel:
Jenkins: ok to test

boegel:
@molden: ignore the failed test, the run was cancelled due to issues with Jenkins, I'll retrigger the test

boegel:
Jenkins: test this please

boegel:
Looking great, thanks @molden!

PR #1418, opened 2015-10-08
use class name (string) rather than License instances as values for software license constants

boegel:
@Caylo ran into this when working on support for easyconfigs files in YAML syntax (.yeb), see #1408, where constants have to be injected into the easyconfig file being parsed; injecting `License` instances into YAML doesn't work

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -214,6 +214,7 @@ class EasyConfig(object):
         self.short_mod_name = mns.det_short_module_name(self)
         self.mod_subdir = mns.det_module_subdir(self)
 
+        self.software_license = None
 
     def copy(self):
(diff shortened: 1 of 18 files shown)


rjeschmi:
It looks fine, but I'm not sure I understand what the original issue was

boegel:
@rjeschmi: originally, the license constants (e.g., `GPLv3`) corresponded to *instances* of a `License` subclass (e.g. `<instance of LicenseGPLv3 class>`), and those can't be injected into YAML to trick it into resolving the constants; it does work with strings (e.g. `'LicenseGPLv3'`)though, hence this fix

rjeschmi:
thanks. Yeah it seems like a reasonable standard too when we find other cases :)

boegel:
Going in, thanks for the review @rjeschmi!

PR #1422, opened 2015-10-10
relax sanity_check_paths in EasyBuild bootstrap script to deal with possible zipped .egg

boegel:
If the vsc-base dependency gets installed as a zipped `.egg`, the sanity check will fail because the EasyBuild easyblock checks for files *in* the vsc-base .egg directory.

This is a hotfix, the problem has to be resolved in the EasyBuildMeta easyblock (see https://github.com/hpcugent/easybuild-easyblocks/issues/710), but this change is required right now to ensure the bootstrap keeps working until the next EasyBuild release is out that includes the fixes EasyBuildMeta easyblock.


boegel:
Going in as a hotfix for the bootstrap script, actual fix is to resolve https://github.com/hpcugent/easybuild-easyblocks/issues/710.

PR #1423, opened 2015-10-10
don't allow passing of values that are look very similar to an existing option

boegel:
fix for https://github.com/hpcugent/easybuild/issues/113

### DIFF
--- a/test/framework/options.py
+++ b/test/framework/options.py
@@ -49,7 +49,7 @@ from easybuild.tools.environment import modify_env
 from easybuild.tools.filetools import mkdir, read_file, write_file
 from easybuild.tools.github import fetch_github_token
 from easybuild.tools.modules import modules_tool
-from easybuild.tools.options import EasyBuildOptions
+from easybuild.tools.options import EasyBuildOptions, parse_options
 from easybuild.tools.toolchain.utilities import TC_CONST_PREFIX
 from easybuild.tools.version import VERSION
 from vsc.utils import fancylogger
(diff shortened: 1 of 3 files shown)


test/framework/options.py
```diff
@@ -1898,6 +1898,13 @@ def test_review_pr(self):
         self.mock_stdout(False)
         self.assertTrue(re.search(r"^Comparing zlib-1.2.8\S* with zlib-1.2.8", txt))
 
+    def test_allow_values_that_are_probably_meant_to_be_options(self):
+        """Test detecting typos or ambiguities on the eb command line."""
+        common_args = []
+
+        eb_go = parse_options(args=common_args + ['-r', 'f'])
+        self.assertFalse(eb_go.options.robot == 'f')
```
boegel: @stdweird: for same reason, this doesn't work... `-r f` should be rejected by generaloption because there's an option `-f`, but it's not

I checked a bit, and for some reason `is_value_a_commandline_option` is getting `()` as a value passed in, rather than `f` which would lead to the reject

any idea? `--robot`/`-r` is a `store_or_None` option, could that be the reason?

stdweird: can you add a test that is not store_or_none? just to make sure the vsc-base is up to date? (or is this with `python setup.py test`?)

boegel: I should be testing `f in eb_go.options.robot` to actually make the test fail, but fine :-)

boegel: @stdweird: reproduced this problem with a test case for vsc-base, see https://github.com/hpcugent/vsc-base/issues/200

boegel:
blocked by https://github.com/hpcugent/vsc-base/issues/200

akesandgren:
@boegel relevance?

PR #1424, opened 2015-10-12
generic sanity-checking of libraries

torbjoernk:
A new key (`'libs'`) is introduced potentially holding a list of maps specifying libraries to be checked in a generic way. Each library map must have a `'name'` and a `'kind'` while the latter must either be `'shared'` or `'static'`. The name is only the library name without path and suffix.

Example:
To test for a library `libmy.so`, which might either be in `lib/` or in `lib64/`:

```
sanity_check_paths = {
  'libs': [{'name': 'libmy', 'kind': 'shared'}]
}
```

This will look for the following files:

```
lib/libmy.so
lib/libmy.dyn
lib64/libmy.so
lib64/libmy.dyn
```

This is done by generating a tuple of these, which is appended to the `sanity_check_paths['files']` list.

Todo:
- [ ] make the library directory and library suffixes configurable
- [ ] think of better way to specify the libraries in the easyconfig file

BTW: The first commit is a potential bug fix. The lambdas specified in `path_keys_and_check` where never used.


easybuild/framework/easyblock.py
```diff
@@ -1546,6 +1546,10 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
         - if *any* of the files/subdirectories in the installation directory listed
           in sanity_check_paths are non-existent (or empty), the sanity check fails
         """
+        # constants for library checking (TODO: should probably be defined somewhere else/configurab;e)
+        LIB_DIRS = ['lib', 'lib64']
+        LIB_SUFFIXES = {'static': ['.a'], 'shared': ['.so', '.dyn']}
```
boegel: constant should be defined at the top of this file, below imports

rather than hardcoding `.so` or `.dyn`, use `get_shared_lib_ext()` from `easybuild.tools.systemtools`

easybuild/framework/easyblock.py
```diff
@@ -1566,6 +1570,24 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
         else:
             self.log.info("Using specified sanity check paths: %s" % paths)
 
+            # transform specified libs into simple file paths
+            if 'libs' in paths.keys():
+                for lib in paths['libs']:
```
boegel: use `for lib in paths.get('libs', [])` to avoid deep indentation

easybuild/framework/easyblock.py
```diff
@@ -1566,6 +1570,24 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
...
+                    if 'name' in lib.keys() and 'kind' in lib.keys():
+                        if lib['kind'] in LIB_SUFFIXES.keys():
+                            # TODO: also check possible library version numbers (e.g. libawesome.so.3.0.1)
+                            libpaths = ()
+                            for dir in LIB_DIRS:
+                                for suffix in LIB_SUFFIXES[lib['kind']]:
+                                    libpaths += (os.path.join(dir, lib['name'] + suffix), )
+                            paths['files'] += [libpaths]
+                        else:
+                            raise EasyBuildError("Library kind must either be static or shared.")
```
boegel: don't hardcode the available keys in the message, use `LIB_SUFFIXES.keys()`:

```python
raise EasyBuildError("Unknown kind of library, only known kinds are: %s" % ' '.join(LIB_SUFFIXES.keys()))
```

easybuild/framework/easyblock.py
```diff
@@ -1566,6 +1570,24 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
...
+                    else:
+                        raise EasyBuildError("Libraries are specified via 'name' and 'kind'.")
```
boegel: s/are/must be/

easybuild/framework/easyblock.py
```diff
@@ -1566,6 +1570,24 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
...
+            # transform specified libs into simple file paths
+            if 'libs' in paths.keys():
+                for lib in paths['libs']:
+                    if 'name' in lib.keys() and 'kind' in lib.keys():
+                        if lib['kind'] in LIB_SUFFIXES.keys():
+                            # TODO: also check possible library version numbers (e.g. libawesome.so.3.0.1)
+                            libpaths = ()
+                            for dir in LIB_DIRS:
+                                for suffix in LIB_SUFFIXES[lib['kind']]:
+                                    libpaths += (os.path.join(dir, lib['name'] + suffix), )
```
boegel: use a list for `libpaths`, which is easier to update, and then use `tuple(libpaths)` below

easybuild/framework/easyblock.py
```diff
@@ -1566,6 +1570,24 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
...
+                            paths['files'] += [libpaths]
```
boegel: use:
```python
paths['files'].append(tuple(libpaths))
```

easybuild/framework/easyblock.py
```diff
@@ -1566,6 +1570,24 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
...
+                        else:
+                            raise EasyBuildError("Library kind must either be static or shared.")
+                    else:
+                        raise EasyBuildError("Libraries are specified via 'name' and 'kind'.")
+                # remove the libs
+                del paths['libs']
```
boegel: why does this need to be removed?

easybuild/framework/easyblock.py
```diff
@@ -1585,7 +1607,7 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, extension=F
                 found = False
                 for name in xs:
                     path = os.path.join(self.installdir, name)
-                    if os.path.exists(path):
+                    if check_fn(path):
```
boegel: this is indeed a bug, this fix is also included in #1388 (which doesn't mean it shouldn't be included here)

maybe we should make this a separate PR though, to make the bug fix stand out; it may have some fallout, i.e. causing sanity checks to fail on empty directories where they didn't before (they should have though)

boegel:
@torbjoernk: currently, this will be a bit verbose in the easyconfig due to requiring the `name` and `kind` keys; two alternatives I have in mind are using a tuple rather than a dictionary for the `libs` entries, or using `shared_libs` and `static_libs` keys.

Also, doesn't it make sense to omit the `lib` part too, since that will always be there?

Example comparison:

```python
sanity_check_paths = {
    'libs': [{'name': 'foo', 'kind': 'shared'}, {'name': 'foo', 'kind': 'static'},
             {'name': 'bar', 'kind': 'shared'}, {'name': 'bar', 'kind': 'static'}],
}
```

vs

```python
sanity_check_paths = {
    'libs': [('foo', 'shared'), ('foo', 'static'), ('bar', 'shared'), ('bar', 'static')],
}
```

vs

```python
sanity_check_paths = {
    'shared_libs': ['foo', 'bar'],
    'static_libs': ['foo', 'bar'],
}
```

boegel:
Jenkins: ok to test

easybuild/framework/easyblock.py
torbjoernk: Awesome. I was looking for such a method but couldn't find it. Thanks.

easybuild/framework/easyblock.py
torbjoernk: Not necessarily, just for cleanup purposes. I can omit that.

torbjoernk:
I personally like the last suggested version with separate fields for shared and static libraries. That seems the least verbose, most expressive and least typo-prone version.
I'll try to implement that one.

easybuild/framework/easyblock.py
torbjoernk: Otherwise the following exception will be thrown:
```{.py}
if not ks == req_keys or sum(valnottypes) > 0 or sum(lenvals) == 0:
  raise EasyBuildError("Incorrect format for sanity_check_paths (should (only) have %s keys, "
                       "values should be lists (at least one non-empty)).", ','.join(req_keys))
```

This might be circumvented by implementing the library checks as a lambda in `path_keys_and_check` (similar for files and dirs) though that would make it impossible to treat the libraries as regular files (and reusing the logic for these).

easybuild/framework/easyblock.py
boegel: even worse, this check will now pass on directories that are actually files

this distinction can be important, for example if you want to make sure that a Python package is installed as a zipped egg, i.e., that the `.egg` is a file rather than a directory

so, we really need a separate PR for this

boegel: see #1436

easybuild/framework/easyblock.py
```diff
@@ -100,6 +100,11 @@
 
 MODULE_ONLY_STEPS = [MODULE_STEP, PREPARE_STEP, READY_STEP, SANITYCHECK_STEP]
 
+# constants for library checking (TODO: should probably be configurable)
+LIB_DIRS = ['lib', 'lib64']
+LIB_STATIC_SUFFIXES = ['a']
+LIB_SHARED_SUFFIXES = [get_shared_lib_ext()]
```
boegel: Shouldn't one suffix suffice? I don't see it happening soon that we'll have multiple suffixes?

It will also simplify things below a bit.

easybuild/framework/easyblock.py
```diff
@@ -100,6 +100,11 @@
 
 MODULE_ONLY_STEPS = [MODULE_STEP, PREPARE_STEP, READY_STEP, SANITYCHECK_STEP]
 
+# constants for library checking (TODO: should probably be configurable)
+LIB_DIRS = ['lib', 'lib64']
```
boegel: I agree, making this configurable makes a lot of sense.

boegel:
@torbjoernk: we also need enhanced unit tests for this... Not sure how easy that would be, shout if you need help.

torbjoernk:
I've implement a simple test case for the new library specifiers. Is that the kind of test you had in mind?

torbjoernk:
I have no idea why this fails on your Jenkins. :confused:

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
@torbjoernk: the problem is that you're changing the `toy-0.0.eb` test easyconfig, and specify that `libtoy.so` and `libtoy.a` must be there

In the unit test you add, you make sure they're there, but the sanity check for that easyconfig fails in other tests where it is used.

You should avoid changing the `toy-0.0.eb` test easyconfig, and just modify the `sanity_check_paths` parameter in your test, something like this:

```python
# specify that libtoy.a and libtoy.so should also be there, either in lib or lib64
ec['ec']['sanity_check_paths'].update({
    'shared_libs': ['libtoy'],
    'static_libs': ['libtoy'],
})
eb = EB_toy(ec['ec'])
...
```

akesandgren:
@boegel should we move forward with this?

boegel:
@akesandgren It may be worth picking this up again, but it should be revived and re-evaluated, it's been too long...

PR #1428, opened 2015-10-14
add support for auto-converting to expected value type for easyconfig parameters

boegel:

PR #1430, opened 2015-10-14
[.yeb] no (more) need for quotes in version definition in gzip.yeb (WIP)

boegel:
We have support now to automatically convert wrongly parsed values, so use it to fix the `version` issue in `gzip.yeb`.

### DIFF
--- a/easybuild/framework/easyconfig/parser.py
+++ b/easybuild/framework/easyconfig/parser.py
@@ -79,11 +79,21 @@ class EasyConfigParser(object):
         Can contain references to multiple version and toolchain/toolchain versions
     """
 
-    def __init__(self, filename=None, format_version=None, rawcontent=None):
-        """Initialise the EasyConfigParser class"""
+    def __init__(self, filename=None, format_version=None, rawcontent=None,
+                 auto_convert_value_types=True):
+        """
+        Initialise the EasyConfigParser class
+        @param filename: path to easyconfig file to parse (superseded by rawcontent, if specified)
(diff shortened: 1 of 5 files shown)


boegel:
cc @Caylo (clearly not finished yet)

boegel:
@Caylo: can you look into getting this to work? We should use the (auto-)convert stuff that was added in #1428, but I'm not sure where exactly... Always (auto-)converting is maybe not a good idea, haven't made up my mind yet.

easybuild/framework/easyconfig/parser.py
```diff
@@ -114,9 +114,11 @@ def check_values_types(self, cfg):
         """
         wrong_type_msgs = []
         for key in cfg:
-            type_ok, _ = check_type_of_param_value(key, cfg[key])
+            type_ok, newval = check_type_of_param_value(key, cfg[key], auto_convert=True)
             if not type_ok:
                 wrong_type_msgs.append("value for '%s' should be of type '%s'" % (key, TYPES[key].__name__))
+            else:
```
boegel: @Caylo: maybe this would be better?

```python
elif newval != cfg[key]:
    self.log.warning("Value for '%s' easyconfig parameter was converted from %s (type: %s) to %s (type: %s)",
                     key, cfg[key], type(cfg[key], newval, type(newval))
    cfg[key] = newval
```

boegel:
@Caylo: maybe add a named argument to `check_values_types` so you can disable `auto_convert` from the outside (e.g., from the tests)?

boegel:
Thanks @Caylo for finishing this!

PR #1432, opened 2015-10-15
minor typo

truatpasteurdotfr:
starting easy :P

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -941,7 +941,7 @@ def cleanup(logfile, tempdir, testing):
                     os.remove(log)
             except OSError, err:
                 raise EasyBuildError("Failed to remove log file(s) %s*: %s", logfile, err)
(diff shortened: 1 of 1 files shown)


boegel:
This is going to sound very silly for a tiny typo fix like this, but... pull requests should target the `develop` branch (and the target branch can not be changed for open PRs).

The only time pull requests to `master` are actually merged is when a new release is being made.

Also, I'm afraid this fix is already included in `develop` currently, you're late to the party. ;-)
So, this will be fixed in the next EasyBuild release.

I hope you understand that we prefer sticking to this policy. For something like this it would obviously be OK to merge it, but drawing a line at "no PRs get merged to master" and sticking to it is easier if *no* exceptions are made (well, except for a 'new release' PR :p).
Besides, `master` always corresponds to the latest EasyBuild release, which is a certainty we don't want to part from.

Please let us know what you think.

I won't close this PR (another thing we try to stick to), I'll leave that up to you.

truatpasteurdotfr:
no problem with me, I wasn't aware of your policy. I will swith to the develop branch.

PR #1435, opened 2015-10-20
Added new option --rebuild

cfenoy:
The --force option skips osdependencies check so we need a mechanism to rebuild an application without skipping any check


easybuild/tools/options.py
```diff
@@ -134,6 +134,7 @@ def basic_options(self):
             'job': ("Submit the build as a job", None, 'store_true', False),
             'logtostdout': ("Redirect main log to stdout", None, 'store_true', False, 'l'),
             'only-blocks': ("Only build listed blocks", None, 'extend', None, 'b', {'metavar': 'BLOCKS'}),
+            'rebuild': ("Rebuild the software without skipping dependency tests",None, 'store_true', False),
```
boegel: fix the small style issue here: space after `,`

boegel:
@cfenoy: are you up for enhancing the `test_force` test in test/framework/options.py to verify that `--rebuild` works as expected?

boegel:
Jenkins: ok to test

cfenoy:
@boegel you mean adding something like this?:

        # check that --rebuild works
        args = [
                eb_file,
                '--rebuild',
                '--debug',
               ]
        outtxt = self.eb_main(args)

        self.assertTrue(not re.search(already_msg, outtxt), "Already installed message not there with --rebuild")

boegel:
@cfenoy: basically, yes

boegel:
@cfenoy: the other way of fixing this is to *not* imply `--ignore-osdeps` when `--force` is used, if that's considered a bug (as suggested by @pescobar).

Thoughts on that?

cfenoy:
From my point of view --force should always install the application, even if the os dependencies are not met. So having a new flag to rebuild seems more logical from my point of view, but I'm open to discuss this.

easybuild/tools/options.py
```diff
@@ -134,6 +134,7 @@ def basic_options(self):
             'job': ("Submit the build as a job", None, 'store_true', False),
             'logtostdout': ("Redirect main log to stdout", None, 'store_true', False, 'l'),
             'only-blocks': ("Only build listed blocks", None, 'extend', None, 'b', {'metavar': 'BLOCKS'}),
+            'rebuild': ("Rebuild the software without skipping dependency tests", None, 'store_true', False),
```
boegel: I would change the help message to:

`"Rebuild software, even if module already exists (don't skip OS dependencies checks)"

and maybe revisit the help message for `--force` and clarify it if needed?

test/framework/options.py
```diff
@@ -198,6 +198,16 @@ def test_force(self):
 
         self.assertTrue(not re.search(already_msg, outtxt), "Already installed message not there with --force")
 
+        # check that --rebuild works
+        args = [
+                eb_file,
+                '--rebuild',
+                '--debug',
+               ]
```
boegel: this is basically copy-paste from above, how about this to clean this whole block (and the stuff for `--force` above) up a bit:

```python
# check that --force and --rebuild work
for arg in ['--force', '--rebuild']:
    outtxt = self.eb_main([eb_file, '--debug', arg])
    self.assertTrue(not re.search(already_msg, outtxt), "Already installed message not there with %s" % arg)
```

boegel:
@cfenoy: merge is broken because of (probably minor) conflicts with current `develop`; I'll send a PR your way to deal with that

boegel:
@cfenoy: https://github.com/cfenoy/easybuild-framework/pull/1 should fix the conflict with current develop

boegel:
Looks good, but, left to do:

* [ ] dry run mentions `[F]` for forced rebuilds, which doesn't really match `--rebuild` (see also http://easybuild.readthedocs.org/en/latest/Using_the_EasyBuild_command_line.html#get-an-overview); can we do better? maybe `[R]` if `--rebuild` is used, while sticking to `[F]` when `--force` is used?
* [ ] documentation needs to be updated
  * it should mention `--rebuild` everywhere (??) where `--force` is mentioned now
  * a clear section on explaining the difference between `--rebuild` and `--force` should be added (basically, `--force` == `--rebuild --ignore-osdeps`)

easybuild/tools/config.py
```diff
@@ -109,6 +109,7 @@ def mk_full_default_path(name, prefix=DEFAULT_PREFIX):
         'only_blocks',
         'optarch',
         'parallel',
+        'rebuild',
```
boegel: this should be added to the `False` list rather than the `None` list

in practice, it's not going to make a difference probably, but it's better to do it 'right'

@cfenoy: please fix this, by moving `rebuild` into the `False` list below?

easybuild/tools/options.py
```diff
@@ -138,11 +138,12 @@ def basic_options(self):
             'extended-dry-run': ("Print build environment and (expected) build procedure that will be performed",
                                  None, 'store_true', False, 'x'),
             'extended-dry-run-ignore-errors': ("Ignore errors that occur during dry run", None, 'store_true', True),
-            'force': ("Force to rebuild software even if it's already installed (i.e. if it can be found as module)",
+            'force': ("Force to rebuild software even if it's already installed (i.e. if it can be found as module) skipping OS dependencies checks",
```
boegel: line too long (should be 120 characters or less)

easybuild/tools/options.py
```diff
@@ -138,11 +138,12 @@ def basic_options(self):
...
                       None, 'store_true', False, 'f'),
             'job': ("Submit the build as a job", None, 'store_true', False),
             'logtostdout': ("Redirect main log to stdout", None, 'store_true', False, 'l'),
             'only-blocks': ("Only build listed blocks", None, 'extend', None, 'b', {'metavar': 'BLOCKS'}),
+            'rebuild': ("Rebuild software, even if module already exists (don't skip OS dependencies checks)", None, 'store_true', False),
```
boegel: line too long?

easybuild/tools/config.py
boegel: @cfenoy: fixed this (and the other remarks) in https://github.com/cfenoy/easybuild-framework/pull/2, please merge to get it included here?

boegel:
Going in, thanks @cfenoy!

PR #1437, opened 2015-10-21
enable auto-converting type of easyconfig parameter values

boegel:
fleshed out the relevant bits from #1430 

This shouldn't make a difference with `.eb` files, especially since only the type of `name` and `version` is being checked and converted if needed.

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -109,7 +109,8 @@ class EasyConfig(object):
     Class which handles loading, reading, validation of easyconfigs
     """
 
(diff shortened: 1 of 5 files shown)


boegel:
@wpoely86, @Caylo: please review?

Caylo:
looks good to me :)

wpoely86:
looks fine but test missing?

boegel:
@wpoely86: good point, I'll enhance the tests to check for the default behaviour

We already have a test that check whether an error is produced in case a broken easyconfig is passed with auto-convert disabled, so should be trivial to add.

boegel:
Thanks for the review @Caylo and @wpoely86!

PR #1439, opened 2015-10-27
Skip symlinked files when fixing permissions

geimer:
Either the symlinks will point to other files in the installation prefix, so permissions for those files are being handled twice in that case, or the symlinks point to somewhere else (like for PDT), in which case we shouldn't touch the files they point to at all (not only to avoid errors, it's just plain wrong to do so)...

The old behavior is still available via `skip_symlinks=False`.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -701,15 +701,16 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
             paths = []
             if not onlydirs:
                 if not skip_symlinks:
-                    paths += files
+                    paths.extend(files)
                 else:
                     for path in files:
-                        fname = os.path.join(root, path)
-                        if not os.path.islink(fname):
+                        if not os.path.islink(os.path.join(root, path)):
                             paths.append(path)
+                        else:
+                            _log.debug("Not adjusting permissions for symlink %s", path)
             if not onlyfiles:
                 # os.walk skips symlinked dirs by default, i.e., no special handling needed here
-                paths += dirs
+                paths.extend(dirs)
 
             for path in paths:
                 allpaths.append(os.path.join(root, path))
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -700,14 +700,14 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
         for root, dirs, files in os.walk(name):
             paths = []
             if not onlydirs:
-                if not skip_symlinks:
-                    paths.extend(files)
-                else:
+                if skip_symlinks:
                     for path in files:
-                        if not os.path.islink(os.path.join(root, path)):
-                            paths.append(path)
-                        else:
+                        if os.path.islink(os.path.join(root, path)):
                             _log.debug("Not adjusting permissions for symlink %s", path)
+                        else:
+                            paths.append(path)
+                else:
+                    paths.extend(files)
             if not onlyfiles:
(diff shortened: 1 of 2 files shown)


easybuild/tools/filetools.py
```diff
@@ -700,8 +700,15 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
         for root, dirs, files in os.walk(name):
             paths = []
             if not onlydirs:
-                paths += files
+                if not skip_symlinks:
+                    paths += files
```
boegel: let's switch to `paths.extend(files)` here (matches better with the `.append` used below)

easybuild/tools/filetools.py
```diff
@@ -700,8 +700,15 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
...
+                else:
+                    for path in files:
+                        fname = os.path.join(root, path)
+                        if not os.path.islink(fname):
+                            paths.append(path)
             if not onlyfiles:
+                # os.walk skips symlinked dirs by default, i.e., no special handling needed here
                 paths += dirs
```
boegel: please change to `paths.extend(dirs)`

easybuild/tools/filetools.py
```diff
@@ -700,8 +700,15 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
         for root, dirs, files in os.walk(name):
             paths = []
             if not onlydirs:
-                paths += files
+                if not skip_symlinks:
+                    paths += files
+                else:
+                    for path in files:
+                        fname = os.path.join(root, path)
+                        if not os.path.islink(fname):
```
boegel: fname is only used in one place, so collapse?

```python
if not os.path.islink(os.path.join(root, path)):
```

easybuild/tools/filetools.py
```diff
@@ -700,8 +700,15 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
...
+                            paths.append(path)
```
boegel: let's add an `else:` here with some logging?

```python
else:
    _log.debug("Not adjusting permissions for symlink %s", path)

File: easybuild/tools/filetools.py (hunk 3 above)
Comment (geimer) [inline easybuild/tools/filetools.py]: Right. This is a leftover from an intermediate version. Initially I appended `fname` rather than `path`, so it was actually used twice. But that was plain wrong...

File: easybuild/tools/filetools.py [hunk 5]
```diff
@@ -700,9 +700,17 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
         for root, dirs, files in os.walk(name):
             paths = []
             if not onlydirs:
-                paths += files
+                if not skip_symlinks:
```
boegel: hmm, maybe it's better to avoid 'double negation' (sort of):

```python
if skip_symlinks:
   ...
else:
   # ...
   paths.extend(dirs)
```

boegel:
Jenkins: ok to test

easybuild/tools/filetools.py
geimer: Makes some sense.  But after switching this, we have yet another negation inside the for loop -- which has now moved up.  Though I would suggest to keep it that way to have the part where something has to be done first.

geimer: Thinking about it, it may actually make sense to switch that too (If we are skipping symlinks and if it *is* a link, issue a debug message; otherwise do it)

boegel:
Going in, thanks @geimer!

PR #1440, opened 2015-10-27
Fix HMNS to always use full version number

geimer:
Fixes an issue encountered when building `GNU-4.9.3-2.25` using `GCC-4.9.3-binutils-2.25`.  The `GCC/4.9.3-binutils-2.25` module file extended the modulepath by `<installdir>/Compiler/GCC/4.9.3-binutils-2.25`, but software was (incorrectly) installed to `<installdir>/Compiler/GCC/4.9.3`.


boegel:
Jenkins: ok to test

boegel:
Thanks @geimer!

PR #1443, opened 2015-10-28
added sqlite and python test files

Caylo:
(dependencies are commented out in the test files to avoid having to add them all - we test those with other files already, so it shouldn't be a problem)


test/framework/easyconfigs/yeb/python-2.7.10.yeb
```diff
@@ -0,0 +1,132 @@
...
+#    [zlib, 1.2.8],
+#    [libreadline, '6.3'],
+#    [ncurses, '5.9'],
+#    [SQLite, 3.8.10.2],
+#    [Tk, 8.6.4, -no-X11],
+#    [OpenSSL, 1.0.1m],  # OS dependency should be preferred if the os version is more recent then this version, its
+#    nice to have an up to date openssl for security reasons
+]
+
+# fixme: tuple values
```
boegel: include a reference to the issue you opened in this comment

test/framework/easyconfigs/yeb/python-2.7.10.yeb
```diff
@@ -0,0 +1,132 @@
...
+    more effectively.
+
+toolchain: {name: ictce, version: 4.1.13}
+toolchainopts: {pic: True, opt: True, optarch: True}
+
+source_urls: ["http://www.python.org/ftp/python/2.7.10/"]
+sources: [*SOURCE_TGZ]
+
+# python needs bzip2 to build the bz2 package
+dependencies: [
```
boegel: add a comment why these dependencies are commented out (also in the `.eb` file)

boegel:
@Caylo: please rename `python-2.7.10.eb` to `Python-2.7.10-ictce-4.1.13.eb` (stick to the 'official' naming scheme for easyconfigs)

For the `.yeb` files, it's less of an issue, but just use the equivalent `.yeb` name (i.e. include the toolchain name in the filename as well).

While you're at it, also fix the `gzip.yeb` and `bzip2.yeb` filenames, since they're version and toolchain specific (for now).

boegel:
Jenkins: ok to test

test/framework/easyconfigs/yeb/python-2.7.10.yeb
```diff
@@ -0,0 +1,132 @@
...
+
+homepage: http://python.org/
+description: |
+    Python is a programming language that lets you work more quickly and integrate your systems
+    more effectively.
+
+toolchain: {name: ictce, version: 4.1.13}
+toolchainopts: {pic: True, opt: True, optarch: True}
+
+source_urls: ["http://www.python.org/ftp/python/2.7.10/"]
```
boegel: use !join or `%(version)s` here, don't hardcode the version

test/framework/easyconfigs/yeb/python-2.7.10.yeb
```diff
@@ -0,0 +1,132 @@
+_internal_variables_:
+    - &numpyversion 1.9.2
+    - &scipyversion 0.15.1
+    - &name Python
+    - &version 2.7.10
+
+easyblock: ConfigureMake
+
+name: Python
+version: 2.7.10
```
boegel: use `*version` here too, or don't include `version` in the list above; same for `name`

test/framework/easyconfigs/yeb/sqlite-3.8.10.2.yeb
```diff
@@ -0,0 +1,31 @@
+_internal_variables_:
+    - &versionstr 3081002
```
boegel: this is now hardcoding the version in two places, just in a different way, which we should try to avoid...

doesn't look trivial to change though, I must admit

test/framework/easyconfigs/yeb/Python-2.7.10-ictce-4.1.13.yeb
```diff
@@ -0,0 +1,132 @@
...
+
+homepage: http://python.org/
+description: |
+    Python is a programming language that lets you work more quickly and integrate your systems
+    more effectively.
+
+toolchain: {name: ictce, version: 4.1.13}
+toolchainopts: {pic: True, opt: True, optarch: True}
+
+source_urls: [!join ["http://www.python.org/ftp/python/", *version, /]]
```
boegel: use `%(version)s` here, and drop the `version` anchor/alias, there's no reason that this should be different from the `.eb` easyconfig?

boegel: see https://github.com/Caylo/easybuild-framework/pull/21

boegel:
Thanks @Caylo!

PR #1444, opened 2015-10-28
[WIP] yeb versionmajor function

Caylo:


easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -63,9 +63,19 @@ def yaml_join(loader, node):
     seq = loader.construct_sequence(node)
     return ''.join([str(i) for i in seq])
 
-# register the tag handler
-yaml.add_constructor('!join', yaml_join)
+def yaml_versionmajorminor(loader, node):
+    """
+    defines custom YAML function to get major version.
+    @param loader: the YAML Loader
+    @param node: the YAML (scalar) node
```
boegel: it's not a scalar currently, but a sequence?

test/framework/yeb.py
```diff
@@ -117,6 +117,24 @@ def test_join(self):
...
 
+    def test_majmin(self):
+        """ Test yaml_versionmajorminor function """
+        stream = [
+            "variables:",
+            "   - &version 2.7.65",
+            "   - &comp_version 1.4.8.7",
+            "",
+            "mm1: !majmin [*version]",
+            "mm2: !majmin [*comp_version]",
```
boegel: there must be a way to avoid the need for passing a sequence if an alias like `*version*` is used

can you include an extra test case with a hardcoded version, like:

```
mm3: !majmin 1.23.456
```

does that work as expected (without requiring a list)? You'll need to update the function supporting this to differentiate between a string and list value...

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -63,9 +63,19 @@ def yaml_join(loader, node):
...
+    """
+    pyver = loader.construct_sequence(node)
```
boegel: add this:

```python
if isinstance(pyver, list):
    pyver = pyver[0]
```

and maybe add a debug log message too in that case?

anyway, this shouldn't be needed, there must be a better way

easybuild/framework/easyconfig/format/yeb.py
Caylo: but i want it to be a scalar :(

boegel:
Jenkins: ok to test

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -63,9 +63,26 @@ def yaml_join(loader, node):
...
 
-# register the tag handler
-yaml.add_constructor('!join', yaml_join)
+def yaml_versionmajorminor(loader, node):
+    """
+    defines custom YAML function to get major version.
+    @param loader: the YAML Loader
+    @param node: the YAML (sequence) node - for now, ideally this would be a scalar
+    """
+    # this should be fixed
```
boegel: yes please :)

boegel: or maybe it's ok

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -63,9 +63,26 @@ def yaml_join(loader, node):
...
+    if isinstance(node, yaml.nodes.ScalarNode):
+        pyver = loader.construct_scalar(node)
+    elif isinstance(node, yaml.nodes.SequenceNode):
+        pyver = loader.construct_sequence(node)
+        pyver = pyver[0]
```
boegel: this needs a check: if the `pyver` sequence has more than one element, we're in trouble here (that is, we shouldn't be simply ignored all the other elements, we should report it as a problem/syntax error)

boegel:
@Caylo: fixed conflict issue in https://github.com/Caylo/easybuild-framework/pull/22

boegel:
I don't think this is very useful anymore, we have a better way of handling that via templates like `%(pyshortver)s`, cfr. #1595

So, I'll close this, we can revisit later if needed.

PR #1429, opened 2015-10-14
[.yeb] easyconfig files in YAML syntax working with dependencies

Caylo:

PR #1447, opened 2015-10-28
[.yeb] add initial (experimental) support for easyconfigs in YAML syntax

boegel:
mostly implemented by @Caylo, mature enough to be included in EB v2.4.0 under `--experimental`

see http://easybuild.readthedocs.org/en/latest/Writing_yeb_easyconfig_files.html for detailed documentation on what is already supported

requires that https://pypi.python.org/pypi/PyYAML is available (but only when parsing `.yeb` easyconfigs)


boegel:
Already reviewed via the PRs to the `feature/yaml` branch (see #1408, #1429, #1430, #1443), going in.

Thanks @caylo!

PR #1306, opened 2015-06-26
Subtoolchain searching

ocaisa:
Add support for automatically matching the lowest possible member of a toolchain hierarchy when resolving dependencies using a command line option ```--minimal-toolchains```.
Requires a ```SUBTOOLCHAIN``` to be defined for each level in the hierarchy (see examples inside)

WIP @boegel ... please review
FYI @geimer

PR #1451, opened 2015-10-30
Add GCCcore toolchain

ocaisa:
GCCcore is a reworking of the GCC/GNU concept. GCCcore is an intermediary GCC build that allows us to build binutils with a non-system compiler. 
The proposed implementation is that GCCcore will sit underneath all compilers including GCC. This will alow us to give GCC equivalence to intel, PGI and place them all in the same lmod family.
The added advantage is that one can build common applications like CMake in GCCcore and have the same installation available in all naming schemes to all compilers.
It also allows to have different versions of gcc for GCCcore and GCC. 

This PR is paired with hpcugent/easybuild-easyconfigs#2108


boegel:
Jenkins: ok to test

easybuild/toolchains/gcc.py
```diff
@@ -28,9 +28,10 @@
...
-from easybuild.toolchains.compiler.gcc import Gcc
+from easybuild.toolchains.gcccore import GCCcore
 
-
-class GccToolchain(Gcc):
+class GccToolchain(GCCcore):
     """Simple toolchain with just the GCC compilers."""
     NAME = 'GCC'
+    COMPILER_MODULE_NAME = [NAME]
+    SUBTOOLCHAIN = GCCcore.NAME
```
boegel: having this here only makes sense if #1306 is merged (which we plan to do soon, so feel free to leave this here)

easybuild/toolchains/gcccore.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+EasyBuild support for GCC compiler toolchain.
+
+@author: Kenneth Hoste (Ghent University)
+"""
+
+from easybuild.toolchains.compiler.gcc import Gcc
+from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME
+
+class GCCcore(Gcc):
```
boegel: should be the other way around, i.e. `Gcc` deriving from `GCCcore`?

easybuild/toolchains/iccifort.py
```diff
@@ -30,8 +30,10 @@
 """
 
 from easybuild.toolchains.compiler.inteliccifort import IntelIccIfort
-
+# Need to import the GCCbase class so I can get the name from there
```
boegel: base vs core

easybuild/toolchains/gcccore.py
boegel: ah, no, nvm, I'm mixing up compilers and toolchains here

boegel:
@geimer, @wpoely86: thoughts?

boegel:
Jenkins: test this please

boegel:
@ocaisa: please take a look at the failing tests (no broken Jenkins now)

ocaisa:
Argh! This is a complicated one. The problem is the toolchain hierarchy becomes incomplete for the existing tests. That's fairly easily fixed by migrating in the changes in hpcugent/easybuild-easyconfigs#2108 into the tests. It raises a big issue though, minimal toolchains will not be compatible with the toolchain defs unless you use GCCcore (so none of the legacy stuff would be compatible). 

I don't really know how to deal with this. Either we accept the limitation on minimal toolchains or hard code in the exception into `get_toolchain_hierarchy` (if not found GCCcore version, go straight to dummy).

What do you think?

ocaisa:
Added some logic to get_toolchain_hierarchy` so that it ignores `GCCcore` if it can't find a version when trying to generate the hierarchy. This will mean that older toolchains will still work with `--minimal-toolchains`

boegel:
finally found some time to take a good look at this, cfr. https://github.com/ocaisa/easybuild-framework/pull/26

the changes made here seem harmless in the sense that they shouldn't break backward compatibility

boegel:
Going in, thanks for the effort on this @ocaisa!

To use this in easyconfigs, I suggest the approach taken in https://github.com/hpcugent/easybuild-easyconfigs/pull/2214 (rather than https://github.com/hpcugent/easybuild-easyconfigs/pull/2108).

PR #1452, opened 2015-10-30
Revert "fix issues with 'python' being 'python3' rather than 'python2'"

JensTimmerman:
Reverts hpcugent/easybuild-framework#1411

not all remarks were adressed yet, should be fixed first?


boegel:
@JensTimmerman Reverting the whole PR is nuts since it actually fixes a problem, just fix the unresolved remark...

PR #1454, opened 2015-11-01
fix MPI compiler wrapper env vars for toolchains using psmpi

boegel:
bug introduced via #1394, manifests itself with error below, for example when building ScaLAPACK with gpsmpi/2014.12:

```
mpifort -c -fPIC -O2 -march=native icopy.f
make[2]: mpifort: Command not found
```

### DIFF
--- a/easybuild/toolchains/mpi/psmpi.py
+++ b/easybuild/toolchains/mpi/psmpi.py
@@ -1,4 +1,4 @@
-##
+# #
 # Copyright 2012-2015 Ghent University
 #
 # This file is part of EasyBuild,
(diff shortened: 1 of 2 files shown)


boegel:
@ocaisa, @geimer: please review?

easybuild/toolchains/mpi/psmpi.py
```diff
@@ -0,0 +1,48 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+Support for MPICH2 as toolchain MPI library.
```
geimer: Should be `ParaStation MPI` instead of `MPICH2`.

easybuild/toolchains/mpi/psmpi.py
```diff
@@ -0,0 +1,48 @@
...
+@author: Stijn De Weirdt (Ghent University)
+@author: Kenneth Hoste (Ghent University)
+@author: Jens Timmerman (Ghent University)
+"""
+
+from easybuild.toolchains.mpi.mpich import Mpich
+
+
+class Psmpi(Mpich):
+    """MPICH2 MPI class"""
```
geimer: Fix docstring (see above)

geimer:
The proposed changes look reasonable to me, though it probably also works without.  The ParaStation version we are currently using is based on MPICH 3.1.4 and provides `mpifort`.  In fact, both `mpif77` and `mpif90` are just symlinks to the `mpifort` script (like in vanilla MPICH >= 3.1.2, AFAIR).

easybuild/toolchains/ipsmpi.py
```diff
@@ -27,13 +27,12 @@
 
 """
 
-from easybuild.toolchains.impich import Impich
+from easybuild.toolchains.compiler.inteliccifort import IntelIccIfort
+from easybuild.toolchains.mpi.psmpi import Psmpi
```
ocaisa: Why is it necessary to do this? I deliberately inherited from Impich

ocaisa: Sorry, ok I get it.

ocaisa:
Yeah, looks good.

boegel:
@geimer: it doesn't work without, I ran into `mpifort: Command not found` when using `gpsmpi/2014.12`

Thanks for the review both!

boegel:
@geimer: basically, the Parastation MPI version is being used in by the `Mpich` class, which picks `mpifort` vs `mpif90` based on the version (>= 3.2.1).

That's why this change is needed: going with `mpifort` with psmpi 5.1.0 doesn't work (yet).

boegel:
Jenkins: test this please

boegel:
fix confirmed, thanks for the review @geimer and @ocaisa!

PR #1456, opened 2015-11-04
[.yeb] alternate toolchain format

Caylo:
yeb files now work with `toolchain: name,version` or `toolchain: [name,version]`; there's also still support for the old syntax.

docs: https://github.com/hpcugent/easybuild/pull/165

### DIFF
--- a/easybuild/framework/easyconfig/format/yeb.py
+++ b/easybuild/framework/easyconfig/format/yeb.py
@@ -95,9 +95,6 @@ class FormatYeb(EasyConfigFormat):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
-        # toolchain param value check, convert to dict if alt format is used
-        check_type_of_param_value('toolchain', self.parsed_yeb['toolchain'], auto_convert=True)
-
 
     def _inject_constants_dict(self, txt):
         """Inject constants so they are resolved when actually parsing the YAML text."""
--- a/easybuild/framework/easyconfig/format/yeb.py
+++ b/easybuild/framework/easyconfig/format/yeb.py
@@ -35,6 +35,7 @@ from vsc.utils import fancylogger
 from easybuild.framework.easyconfig.format.format import INDENT_4SPACES, EasyConfigFormat
 from easybuild.framework.easyconfig.format.pyheaderconfigobj import build_easyconfig_constants_dict
 from easybuild.framework.easyconfig.format.pyheaderconfigobj import build_easyconfig_variables_dict
+from easybuild.framework.easyconfig.types import check_type_of_param_value
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import read_file
 from easybuild.tools.utilities import only_if_module_is_available, quote_str
@@ -94,12 +95,9 @@ class FormatYeb(EasyConfigFormat):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
-        # make dict out of toolchain
-        if not isinstance(self.parsed_yeb['toolchain'], dict):
-            self.parsed_yeb['toolchain'] = {
-                'name': self.parsed_yeb['toolchain'].split(',')[0].strip(),
-                'version': self.parsed_yeb['toolchain'].split(',')[1].strip(),
-            }
+        # toolchain param value check, convert to dict if alt format is used
+        check_type_of_param_value('toolchain', self.parsed_yeb['toolchain'], auto_convert=True)
+
 
     def _inject_constants_dict(self, txt):
         """Inject constants so they are resolved when actually parsing the YAML text."""
--- a/easybuild/framework/easyconfig/format/yeb.py
+++ b/easybuild/framework/easyconfig/format/yeb.py
@@ -35,7 +35,6 @@ from vsc.utils import fancylogger
 from easybuild.framework.easyconfig.format.format import INDENT_4SPACES, EasyConfigFormat
 from easybuild.framework.easyconfig.format.pyheaderconfigobj import build_easyconfig_constants_dict
 from easybuild.framework.easyconfig.format.pyheaderconfigobj import build_easyconfig_variables_dict
-from easybuild.framework.easyconfig.types import check_type_of_param_value
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import read_file
 from easybuild.tools.utilities import only_if_module_is_available, quote_str
--- a/easybuild/framework/easyconfig/format/yeb.py
+++ b/easybuild/framework/easyconfig/format/yeb.py
@@ -94,6 +94,12 @@ class FormatYeb(EasyConfigFormat):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
+        # make dict out of toolchain
+        if not isinstance(self.parsed_yeb['toolchain'], dict):
+            self.parsed_yeb['toolchain'] = {
+                'name': self.parsed_yeb['toolchain'].split(',')[0].strip(),
+                'version': self.parsed_yeb['toolchain'].split(',')[1].strip(),
+            }
 
     def _inject_constants_dict(self, txt):
         """Inject constants so they are resolved when actually parsing the YAML text."""
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -111,33 +111,28 @@ def convert_value_type(val, typ):
 
     return res
 
-def to_dict(tc):
+def to_toolchain(tcspec):
     """
     Convert a toolchain string "foo, v0" to a dictionary {'name':'foo', 'version':'v0'}
 
-    @param toolchain_str: a toolchain string
+    @param tcspec: a toolchain in the form of a string or a list
     """
     # check if tc is a string or a list of two values; else, it can not be converted
-    if isinstance(tc, basestring):
-        if tc.count(',') != 1:
-            raise EasyBuildError("Can not convert string %s to toolchain dict. Expected format: name,version" % tc)
-        else:
-            tc_split = tc.split(',')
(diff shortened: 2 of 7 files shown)


boegel:
Jenkins: ok to test

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,34 @@ def convert_value_type(val, typ):
         raise EasyBuildError("No conversion function available (yet) for target type %s", typ)
 
     return res
+
+def to_dict(tc):
```
boegel: this is `to_toolchain`, not a generic `to_dict` conversion function

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,34 @@ def convert_value_type(val, typ):
...
+    """
+    Convert a toolchain string "foo, v0" to a dictionary {'name':'foo', 'version':'v0'}
+
+    @param toolchain_str: a toolchain string
```
boegel: or list, and there's no such thing as `toolchain_str`

according to PEP8, two-letter variables are not a good idea, so maybe use `tcspec`?

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,34 @@ def convert_value_type(val, typ):
...
+    """
+    # check if tc is a string or a list of two values; else, it can not be converted
+    if isinstance(tc, basestring):
+        if tc.count(',') != 1:
```
boegel: check using `len(...) == 2` after the `.split(',') instead`

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,34 @@ def convert_value_type(val, typ):
...
+            raise EasyBuildError("Can not convert string %s to toolchain dict. Expected format: name,version" % tc)
+        else:
+            tc_split = tc.split(',')
+            return {'name':tc_split[0].strip(), 'version':tc_split[1].strip()}
+
+    elif isinstance(tc, list):
```
boegel: avoid duplicate code; use `split` if it's a string originally, and then require that's it's a list for the remainder; that way, you only need to check the length once

```python
if isinstance(tcspec, basestring):
    tcspec = tcspec.split(',')

if isinstance(tcspec, list):
    if len(tcspec) == 2:
      ...
    else:
      raise ...

else:
      raise ...

File: test/framework/easyconfigs/yeb/bzip-bad-toolchain.yeb [hunk 5]
```diff
@@ -0,0 +1,25 @@
...
+version: 1.0.6
+
+homepage: 'http://www.bzip.org/'
+# preserve newlines using '|' to ensure match with description from .eb easyconfig
+description: |
+    bzip2 is a freely available, patent free, high-quality data compressor. It typically
+    compresses files to within 10% to 15% of the best available techniques (the PPM family of statistical
+    compressors), whilst being around twice as fast at compression and six times faster at decompression.
+
+toolchain: GCC,4.9,2
```
boegel: use spaces after `,` for readability?

test/framework/yeb.py
```diff
@@ -119,6 +120,15 @@ def test_join(self):
             self.assertEqual(loaded.get(key), 'foobar')
 
 
+    def test_toolchain_alt_format(self):
```
boegel: rename test name, include 'bad' in it since you're only tested those cases here

boegel:
@Caylo: let's see a docs update too for this please :)

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,29 @@ def convert_value_type(val, typ):
         raise EasyBuildError("No conversion function available (yet) for target type %s", typ)
 
     return res
+
```
boegel: double empty line between 'top-level' function/class definitions

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,29 @@ def convert_value_type(val, typ):
...
+        tcspec = tcspec.split(',')
+
+    if isinstance(tcspec, list):
+        if len(tcspec) == 2:
+            return {'name':tcspec[0].strip(), 'version':tcspec[1].strip()}
+        else:
+            raise EasyBuildError("Can not convert list %s to toolchain dict. Expected 2 elements")
+    else:
+        raise EasyBuildError("Conversion of type %s to toolchain dict is not supported" % type(tcspec))
+
```
boegel: double empty line pls

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,29 @@ def convert_value_type(val, typ):
         raise EasyBuildError("No conversion function available (yet) for target type %s", typ)
 
     return res
+
+def to_toolchain(tcspec):
+    """
+    Convert a toolchain string "foo, v0" to a dictionary {'name':'foo', 'version':'v0'}
```
boegel: use an actual toolchain as example, e.g. `intel, 2015a` and `{'name': 'intel', 'version': '2015a'}`

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,29 @@ def convert_value_type(val, typ):
...
+
+    @param tcspec: a toolchain in the form of a string or a list
+    """
+    # check if tc is a string or a list of two values; else, it can not be converted
+    if isinstance(tcspec, basestring):
+        tcspec = tcspec.split(',')
+
+    if isinstance(tcspec, list):
+        if len(tcspec) == 2:
+            return {'name':tcspec[0].strip(), 'version':tcspec[1].strip()}
```
boegel: use `res = {...}` here, and `return res` at the end of the function (we try to stay away from 'internal' returns)

also, space after `:`

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,29 @@ def convert_value_type(val, typ):
...
+        else:
+            raise EasyBuildError("Can not convert list %s to toolchain dict. Expected 2 elements")
```
boegel: this is missing `, tcspec`

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,29 @@ def convert_value_type(val, typ):
...
+    else:
+        raise EasyBuildError("Conversion of type %s to toolchain dict is not supported" % type(tcspec))
```
boegel: include the value `tcspec` in the error message

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +110,29 @@ def convert_value_type(val, typ):
...
+
+TYPE_CONVERSION_FUNCTIONS = {
```
boegel: add a comment explaining why this is at the bottom rather than the top of the module, i.e. because `to_toolchain` needs to be defined before it can be included here

test/framework/easyconfigs/yeb/bzip-bad-toolchain.yeb
```diff
@@ -0,0 +1,25 @@
...
+version: 1.0.6
+
+homepage: 'http://www.bzip.org/'
+# preserve newlines using '|' to ensure match with description from .eb easyconfig
+description: |
+    bzip2 is a freely available, patent free, high-quality data compressor. It typically
+    compresses files to within 10% to 15% of the best available techniques (the PPM family of statistical
+    compressors), whilst being around twice as fast at compression and six times faster at decompression.
+
+toolchain: GCC, 4.9, 2
```
boegel: add a comment above his line to indicate that it's purposely wrong

boegel:
@Caylo: please add an extensive test for `to_toolchain` in `test/framework/type_checking.py`

test/framework/type_checking.py
```diff
@@ -85,6 +85,22 @@ class Foo():
...
+
+        # wrong type
+        self.assertErrorRegex(EasyBuildError, r"Conversion of .* \(type .*\) to toolchain dict is not supported",
+            to_toolchain, ('intel', '2015a'))
+
+        # wrong number of elements
+        self.assertErrorRegex(EasyBuildError, "Can not convert list .* to toolchain dict. Expected 2 elements",
+            to_toolchain, "intel, 2015, a")
+        self.assertErrorRegex(EasyBuildError, "Can not convert list .* to toolchain dict. Expected 2 elements",
+            to_toolchain, "intel")
```
wpoely86: add a list version here too

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -94,6 +95,9 @@ def parse(self, txt):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
+        # toolchain param value check, convert to dict if alt format is used
+        check_type_of_param_value('toolchain', self.parsed_yeb['toolchain'], auto_convert=True)
```
boegel: you're not using the return value here, so if the value type was converted, you're missing it

why is this not popping up in the tests?

boegel: actually, just drop this, the auto convert is already done in the parser, on all keys

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -35,6 +35,7 @@
 from easybuild.framework.easyconfig.format.format import INDENT_4SPACES, EasyConfigFormat
 from easybuild.framework.easyconfig.format.pyheaderconfigobj import build_easyconfig_constants_dict
 from easybuild.framework.easyconfig.format.pyheaderconfigobj import build_easyconfig_variables_dict
+from easybuild.framework.easyconfig.types import check_type_of_param_value
```
boegel: drop this import, not used

boegel:
Thanks @Caylo!

pforai:
@Caylo did you have a look at the YAML structure they have in Spack? /cc @boegel

Caylo:
@pforai i did! it looks good, @boegel mentioned a possible collaboration between Spack and EB to make some kind of base library we can both use. we'll probably discuss it more next week :)

PR #1460, opened 2015-11-10
Use diff --git lines to determine the files in PR

wpoely86:
Instead of searching for +++ lines, search for diff --git. This should
be more foolproof and lead to less false positives.

hpcugent/easybuild-easyconfigs#2126 is an example where the previous regex failed.


wpoely86:
@boegel please review

wpoely86:
jenkins please retest

wpoely86:
Jenkins retest this please

wpoely86:
@boegel can you have a look at these tests? I don't understand what is causing the problem.

boegel:
Jenkins: test this please

boegel:
Going in, thanks @wpoely86!

Issue #1462, opened 2015-11-11
suggestion: add url and description when packaging rpms


pescobar:
@rjeschmi I noticed rpms generated with easybuild have empty url and description. This is just cosmetics but you have this info in every easyconfig and FPM supports --description and --url so adding those when generating the rpms should be easy

rjeschmi:
thanks @pescobar this is a good issue and can be fixed.

boegel:
fixed in #1469

PR #1469, opened 2015-11-12
adding description and url

rjeschmi:
Adding description and url to fpm commandline 

Will resolve: #1462

### DIFF
--- a/test/framework/package.py
+++ b/test/framework/package.py
@@ -42,16 +42,18 @@ from easybuild.tools.filetools import adjust_permissions, read_file, write_file
 from easybuild.tools.package.utilities import ActivePNS, avail_package_naming_schemes, check_pkg_support, package
 from easybuild.tools.version import VERSION as EASYBUILD_VERSION
 
-
+DEBUG = False
+DEBUG_FPM_FILE = "debug_fpm_mock"
 MOCKED_FPM = """#!/bin/bash
 
-DEBUG=  #put something here if you want to debug 
+DEBUG=%(debug)s  #put something here if you want to debug
+
 debug_echo () {
     if [ -n "$DEBUG" ]; then
-        echo "$@" >> /tmp/debug_fpm_mock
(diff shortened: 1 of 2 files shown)


boegel:
@rjeschmi: how about an (enhanced) unit test? see `test/framework/package.py`

boegel:
Jenkins: ok to test

test/framework/package.py
```diff
@@ -44,16 +44,70 @@
 
 
 MOCKED_FPM = """#!/bin/bash
+
+DEBUG=  #put something here if you want to debug 
+debug_echo () {
+    if [ -n "$DEBUG" ]; then
+        echo "$@" >> /tmp/debug_fpm_mock
```
boegel: don't hardcode `/tmp/debug_fpm_mock`, use `$TMPDIR` which will be a unique directory when the script it run (you can echo where you're debugging to), or at the very least use `/tmp/$USER/debug_fpm_mock` (but I strongly prefer using `$TMPDIR`)

test/framework/package.py
```diff
@@ -44,16 +44,70 @@
...
+            ;;
+        --*)
+            debug_echo "got a unhandled opption"
+            shift 2
+            ;;
+        *)
+            debug_echo "got the rest of the output"
+            installdir="$1"
+            modulefile="$2"
+            shift 2
```
boegel: since you're always shifting 2, do this outside of the `case`?

test/framework/package.py
```diff
@@ -44,16 +44,70 @@
...
+        "-t")
+            target="$2"
+            shift 2
+            ;;
+        "-s")
+            source="$2"
+            shift 2
+            ;;
+        --*)
+            debug_echo "got a unhandled opption"
```
boegel: `option`

rjeschmi:
Looks good. Tests all pass now anyway...

boegel:
Thanks a lot @rjeschmi, great work on enhancing the tests!

PR #1470, opened 2015-11-12
Initial support for IBM XL compilers on Power7 and PowerPC (BlueGene)

JackPerdue:
This is just a framework to get started.  The compiler flags still need tweaking, but work for now.

### DIFF
--- a/easybuild/toolchains/compiler/ibmxl.py
+++ b/easybuild/toolchains/compiler/ibmxl.py
@@ -24,70 +24,36 @@ class IBMXL(Compiler):
     COMPILER_FAMILY = TC_CONSTANT_IBMCOMP
     COMPILER_UNIQUE_OPTS = {
         'ibm-static': (False, "Link IBM XL provided libraries statically"),
-#        'no-icc': (False, "Don't set Intel specific macros"),
         'error-unknown-option': (False, "Error instead of warning for unknown options"),
     }
 
     COMPILER_UNIQUE_OPTION_MAP = {
-#        'i8': '',
-#        'r8': '',
-#        'optarch': ['qtune=auto', 'cpu=Power7'],  # IBM XL - later
-        'optarch': 'qtune=auto',  # IBM XL - later
-#        'optarch': 'mcpu=native',  # IBM XL - later
-        'openmp': 'qsmp=omp',    # IBM XL - later
+        'optarch': 'qtune=auto',
+        'openmp': 'qsmp=omp',
         'strict': ['', ''],
         'precise': [''],
-#        'defaultprec': ['ftz', 'fp-speculation=safe', 'fp-model source'],
         'defaultprec': ['', '', ''],
         'loose': [''],
         'veryloose': [''],
-        'ibm-static': 'qstaticlink=xllibs',  # IBM XL - later
-#        'no-icc': 'no-icc',
-        'error-unknown-option': 'we10006',  # error at warning #10006: ignoring unknown option
-        'pic': 'qpic',  # override?
-        'shared': 'qmkshrobj',  # override?
+        'ibm-static': 'qstaticlink=xllibs',
+        'pic': 'qpic',
+        'shared': 'qmkshrobj',
     }
 
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
         systemtools.IBM: ['qtune=auto', 'qmaxmem=-1'],
-        #systemtools.PPC: 'mcpu=Power7',
-        #systemtools.PPC: '',
     }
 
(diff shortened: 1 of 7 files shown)


easybuild/toolchains/compiler/ibmxl.py
```diff
@@ -0,0 +1,93 @@
...
+        #xlc_root, _ = self.get_software_root(self.COMPILER_MODULE_NAME)
+        #xlc_version, xlf_version = self.get_software_version(self.COMPILER_MODULE_NAME)
+
+        libpaths = ['xlsmp/4.1.0', 'xlmass/8.1.0']  # not sure how to get versions
+        if self.options.get('32bit', None):
+            libpaths = ['%s/lib' % x for x in libpaths]
+        else:
+            libpaths = ['%s/lib64' % x for x in libpaths]
+
+        self.variables.append_subdirs("LDFLAGS", '/opt/ibm', subdirs=libpaths)
```
boegel: as discussed, this is probably not needed at all, since your testing indicates it seems to work even though the hardcoded versions don't match up with the compiler versions you were tested with

easybuild/toolchains/compiler/ibmxl.py
```diff
@@ -0,0 +1,93 @@
...
+
+    #LIB_MULTITHREAD = ['xlsmp', 'pthread']  # iomp5 is OpenMP related
+    LIB_MULTITHREAD = ['xlsmp']  # iomp5 is OpenMP related
+
+    def _set_compiler_vars(self):
+        """IBM XL compilers-specific adjustments after setting compiler variables."""
+        super(IBMXL, self)._set_compiler_vars()
+
+        if not ('xlc' in self.COMPILER_MODULE_NAME and 'xlf' in self.COMPILER_MODULE_NAME):
+            self.log.raiseException("_set_compiler_vars: missing xlc and/or xlf from COMPILER_MODULE_NAME %s" % self.COMPILER_MODULE_NAME)
```
boegel: this check is not needed here; it makes sense for ClangGCC, but here it's pretty useless

easybuild/toolchains/compiler/ibmxl.py
```diff
@@ -0,0 +1,93 @@
...
+
+    LINKER_TOGGLE_STATIC_DYNAMIC = {
+        'static': '-Bstatic',
+        'dynamic': '-Bdynamic',
+    }
+
+    #LIB_MULTITHREAD = ['xlsmp', 'pthread']  # iomp5 is OpenMP related
+    LIB_MULTITHREAD = ['xlsmp']  # iomp5 is OpenMP related
+
+    def _set_compiler_vars(self):
```
boegel: you can drop this whole block, see remarks below

easybuild/toolchains/compiler/ibmxl.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class IBMXL(Compiler):
+
+    COMPILER_MODULE_NAME = ['xlc', 'xlf']
+
+    COMPILER_FAMILY = TC_CONSTANT_IBMCOMP
+    COMPILER_UNIQUE_OPTS = {
+        'ibm-static': (False, "Link IBM XL provided libraries statically"),
+#        'no-icc': (False, "Don't set Intel specific macros"),
```
boegel: please drop the commented out lines, they're irrelevant?

boegel:
Jenkins: ok to test

JackPerdue:
Note... in a way this depends upon https://github.com/hpcugent/easybuild-easyblocks/pull/741 since the XL compilers require a multilib GCC to be configured (ala .http://siliconslick.com/easybuild/ebfiles_repo_cleaned/curie/xlc/xlc-13.1.0.1.484.eb [which I need to rework given some recent changes in EB])

boegel:
@JackPerdue: this looks OK, but I'm a bit reluctant to merge this without being able to try this myself...

boegel:
@JackPerdue: as discussed off-PR, let's merge this in. It's pretty clear that this doesn't affect anything else, and you've tested this at length, as you've showed us.

The try and ensure that we don't break things in the future, it would be useful to have access to a system to test this on, but we're working on that, so good.

Going in, thanks a lot for all of your efforts on this, let's see where this goes...

boegel:
Hmm, not sure what triggered that failed test... Weird.

It's a false negative though, the tests basically didn't run.

JackPerdue:
Thanks Sir Boegel!  I don't see the issue in the error log either but I promise it works on Power7 and BG/Q (mostly... AFAICT).

PR #1471, opened 2015-11-12
Generic compilation support 

pescobar:


easybuild/toolchains/compiler/gcc.py
```diff
@@ -63,11 +64,21 @@ class Gcc(Compiler):
         'veryloose': ['mrecip=all', 'mno-ieee-fp'],
     }
 
-    COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
-        systemtools.AMD : 'march=native',
-        systemtools.INTEL : 'march=native',
-        systemtools.POWER: 'mcpu=native',  # no support for march=native on POWER
-    }
+    optarch = build_option('optarch')
+    if optarch == 'GENERIC':
```
boegel: this should be a constant named `GENERIC`, that you import from `easybuild.tools.toolchain.toolchain`

easybuild/toolchains/compiler/inteliccifort.py
```diff
@@ -68,11 +69,20 @@ class IntelIccIfort(Compiler):
         'no-icc': 'no-icc',
         'error-unknown-option': 'we10006',  # error at warning #10006: ignoring unknown option
     }
+    
+    optarch = build_option('optarch')
+    if optarch == 'GENERIC':
```
boegel: constant pls

boegel:
Jenkins: test this please

boegel:
@pescobar: this still needs:

* [ ] enhanced unit test (see `test_override_optarch` in `test/framework/toolchain.py`)
* [ ] docs update

rjeschmi:
Is it worth supporting something in between generic and native?

pescobar:
@rjeschmi I tried to build a software stack "in between" to use it as my generic software stack and I got problems.

My specific case was that I was compiling my "generic" software stack in the oldest cpu I had around which was a nehalem. I was building in the nehalem cpu with the default gcc option in easybuild of "-march=native"  and I supposed that this would work in any other cpu but I found problems when trying to use this software stack in AMD cpus (magny-cours I think) and also when using it in xenserver virtual machines.

That's why I would suggest to build totally generic. I would focus this feature in portability and not in performance

pescobar:
I did some tests and seems to be working for GCC. See this three build logs:

This one is a normal build without using --optarch 
https://gist.github.com/61a9a47dc513e5bb4af2

This one is with --optarch=GENERIC and toolchainopts = {'optarch': True}
https://gist.github.com/6e1bf60e5e51a4ace7f5

This one is with --optarch=GENERIC and toolchainopts = {'optarch': False}
https://gist.github.com/e53fec626cae9c75d2ad

If you grep for "march" in those build logs all them seem fine for me. 

Opinions @boegel ?

boegel:
@pescobar: seems to be working indeed, so go ahead with the docs update

easybuild/toolchains/compiler/inteliccifort.py
```diff
@@ -94,6 +94,20 @@ def __init__(self, *args, **kwargs):
         """Toolchain constructor."""
         class_constants = kwargs.setdefault('class_constants', [])
         class_constants.append('LIB_MULTITHREAD')
+
+        optarch = build_option('optarch', default=None)
+        if optarch == OPTARCH_GENERIC:
+            # do generic build if --optarch=GENERIC
+            COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
```
boegel: this should be `self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {`

easybuild/toolchains/compiler/inteliccifort.py
```diff
@@ -94,6 +94,20 @@ def __init__(self, *args, **kwargs):
...
+                systemtools.INTEL : 'xSSE2',
+                systemtools.AMD : 'xSSE2',
+            }
+        else: # do optimized build (default)
+            COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
```
boegel: missing `self.`

boegel:
@pescobar: unit test + cleanup in https://github.com/pescobar/easybuild-framework/pull/16

boegel:
Going in, thanks @pescobar!

docs update at https://github.com/hpcugent/easybuild/pull/172 will be included soon too

PR #1472, opened 2015-11-12
Allow user-local modules with hierarchical naming schemes

geimer:
If one uses EB's default MNS, user-local modules are no problem, as a site only needs to 'module use' something like '$HOME/.local/modules' in their modules setup script.  However, when using an HMNS, this per-user location needs to be taken into account by the EB-generated modules which extend
$MODULEPATH.

This PR introduces a new option 'subdir-user-modules' (empty by default) which allows to define a location relative to the user's `$HOME` in which a user-local modules hierarchy (according to the module naming scheme, configurable via `det_user_modpath_extensions`) is searched. The corresponding modules will then include an additional `module use` statement to pick up modules from the this location. To prevent errors if the directory does not exist (C modules), this `module use` is guarded by a check.


boegel:
Jenkins: ok to test

easybuild/framework/easyblock.py
```diff
@@ -1019,7 +1019,18 @@ def make_module_extend_modpath(self):
...
+            quoted_modpath_exts = [quote_str(ext) for ext in full_path_modpath_extensions]
+            txt = self.module_generator.use(quoted_modpath_exts)
+
+            # add user-specific module path; use statement will be guarded so no need to create the directories
+            user_modpath = build_option('subdir_user_modules')
+            if user_modpath:
+                user_modpath_exts = ActiveMNS().det_user_modpath_extensions(self.cfg)
+                self.log.debug("Including user module path extensions returned by naming scheme: %s"
+                               % user_modpath_exts)
+                quoted_user_modpath_exts = [quote_str(os.path.join(user_modpath, ext)) for ext in user_modpath_exts]
```
boegel: maybe we should be quoting in `module_generator.use` instead, if it's so important to do this?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1226,6 +1226,13 @@ def det_modpath_extensions(self, ec):
         self.log.debug("Obtained modulepath extensions: %s" % modpath_extensions)
         return modpath_extensions
 
+    def det_user_modpath_extensions(self, ec):
+        """Determine user-specific modulepath extensions according to module naming scheme."""
+        self.log.debug("Determining user modulepath extensions for %s" % ec)
+        modpath_extensions = self.mns.det_user_modpath_extensions(self.check_ec_type(ec))
+        self.log.debug("Obtained user modulepath extensions: %s" % modpath_extensions)
```
boegel: use `self.log.debug("...", foo)` rather than `"..." % foo` (the former is slightly faster, string templating will only be done when the `log.debug` is actually 'performed')

easybuild/tools/module_generator.py
```diff
@@ -250,13 +250,22 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
         """
         use_statements = []
         for path in paths:
-            use_statements.append("module use %s\n" % path)
+            if prefix:
+                full_path = "[file join %s %s]" % (prefix, path)
+            else:
+                full_path = path
+            if guarded:
+                use_statements.append("if {[file isdirectory %s]} {\n" % full_path)
```
boegel: use the `conditional_statement` method available

easybuild/tools/module_generator.py
```diff
@@ -250,13 +250,22 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
+                use_statements.append("    module use %s\n" % full_path)
+                use_statements.append("}\n")
+            else:
+                use_statements.append("module use %s\n" % full_path)
         return ''.join(use_statements)
```
boegel: use `'\n'.join` here, and don't include `\n` everywhere above

easybuild/tools/module_generator.py
```diff
@@ -289,6 +298,11 @@ def set_alias(self, key, value):
         # quotes are needed, to ensure smooth working of EBDEVEL* modulefiles
         return 'set-alias\t%s\t\t%s\n' % (key, quote_str(value))
 
+    def det_home(self):
+        """
+        Return module-syntax specific code to determine the user's home directory.
+        """
+        return '$env(HOME)'
```
boegel: maybe this should be generalised to `def get_env(self)`, and then `self.module_generator.get_env('HOME')` should be used?

easybuild/tools/module_generator.py
```diff
@@ -414,12 +428,24 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
         """
-        return ''.join([self.PREPEND_PATH_TEMPLATE % ('MODULEPATH', quote_str(p)) for p in paths])
+        use_statements = []
+        for path in paths:
+            if prefix:
+                full_path = "pathJoin(%s, %s)" % (prefix, path)
+            else:
+                full_path = path
+            if guarded:
+                use_statements.append('if (isDir(%s)) then\n' % full_path)
```
boegel: same here, reuse the existing `self.conditional_statement`

easybuild/tools/module_generator.py
```diff
@@ -414,12 +428,24 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
+                use_statements.append('    ' + self.PREPEND_PATH_TEMPLATE % ('MODULEPATH', full_path))
+                use_statements.append('end\n')
+            else:
+                use_statements.append(self.PREPEND_PATH_TEMPLATE % ('MODULEPATH', full_path))
+        return ''.join(use_statements)
```
boegel: `'\n'.join` please

easybuild/tools/module_generator.py
```diff
@@ -447,6 +473,12 @@ def set_alias(self, key, value):
         # quotes are needed, to ensure smooth working of EBDEVEL* modulefiles
         return 'setalias("%s", %s)\n' % (key, quote_str(value))
 
+    def det_home(self):
```
boegel: same here, make this a more general `get_env`

easybuild/tools/module_naming_scheme/categorized_hmns.py
```diff
@@ -83,6 +83,16 @@ def det_modpath_extensions(self, ec):
...
 
+    def det_user_modpath_extensions(self, ec):
+        """
+        Determine user module path extensions, if any. As typical users are not expected to have many local modules,
+        further categorizing them using module classes is considered overkill. Thus, we are using a plain hierarchical
+        scheme for user modules instead.
+
+        Examples: Compiler/GCC/4.8.3 (for GCC/4.8.3 module), MPI/GCC/4.8.3/OpenMPI/1.6.5 (for OpenMPI/1.6.5 module)
+        """
+        return super(CategorizedHMNS, self).det_modpath_extensions(ec)
```
boegel: this is exactly the same as the default implementation? what's the point of defining it then, just let inheritance take care of it?

test/framework/easyblock.py
```diff
@@ -611,7 +611,7 @@ def test_exclude_path_to_top_of_module_tree(self):
             for imkl_dep in excluded_deps:
                 tup = (imkl_dep, modfile_path, modtxt)
                 failmsg = "No 'module load' statement found for '%s' not found in module %s: %s" % tup
-                self.assertFalse(re.search("module load %s" % imkl_dep, modtxt), failmsg)
+                self.assertFalse(re.search('module load "%s"' % imkl_dep, modtxt), failmsg)
```
boegel: this is wrong, twice... :-)

* a `module load` statement in a module file in Tcl syntax will not include quotes
* this check is useful in case the module file is in Lua syntax, where it would be `load("foo")`

test/framework/module_generator.py
```diff
@@ -227,18 +228,37 @@ def test_prepend_paths(self):
...
         if self.MODULE_GENERATOR_CLASS == ModuleGeneratorTcl:
+            # Test regular 'module use' statements
             expected = ''.join([
-                "module use /some/path\n",
-                "module use /foo/bar/baz\n",
+                'module use "/some/path"\n',
+                'module use "/foo/bar/baz"\n',
             ])
-            self.assertEqual(self.modgen.use(["/some/path", "/foo/bar/baz"]), expected)
+            self.assertEqual(self.modgen.use([quote_str(p) for p in "/some/path", "/foo/bar/baz"]), expected)
```
boegel: another sign that the `quote_str` should be done by `use` itself, if you ask me

test/framework/module_generator.py
```diff
@@ -227,18 +228,37 @@ def test_prepend_paths(self):
...
+
+            # Test guarded 'module use' statements using prefix
+            expected = ''.join([
+                'if {[file isdirectory [file join "/foo" "/some/path"]]} {\n',
+                '    module use [file join "/foo" "/some/path"]\n',
+                '}\n',
+            ])
+            self.assertEqual(self.modgen.use([quote_str("/some/path")], prefix=quote_str("/foo"), guarded=True),
```
boegel: again, let `use` quote the path being passed

test/framework/module_generator.py
```diff
@@ -227,18 +228,37 @@ def test_prepend_paths(self):
...
+                             expected)
         else:
+            # Test regular 'module use' statements
             expected = ''.join([
                 'prepend_path("MODULEPATH", "/some/path")\n',
                 'prepend_path("MODULEPATH", "/foo/bar/baz")\n',
             ])
-            self.assertEqual(self.modgen.use(["/some/path", "/foo/bar/baz"]), expected)
+            self.assertEqual(self.modgen.use([quote_str(p) for p in "/some/path", "/foo/bar/baz"]), expected)
```
boegel: let `use` quote stuff

test/framework/module_generator.py
```diff
@@ -227,18 +228,37 @@ def test_prepend_paths(self):
...
 
+            # Test guarded 'module use' statements using prefix
+            expected = ''.join([
+                'if (isDir(pathJoin("/foo", "/some/path"))) then\n',
+                '    prepend_path("MODULEPATH", pathJoin("/foo", "/some/path"))\n',
+                'end\n',
+            ])
+            self.assertEqual(self.modgen.use([quote_str("/some/path")], prefix=quote_str("/foo"), guarded=True),
```
boegel: `quote_str` -> `use`

boegel:
@geimer: sorry it took me so long to give this a proper review, you should have pinged me earlier

There's a couple of loose ends here, so too late to make it in for EB v2.5.0 I'm afraid, so I'm going to put 2.6.0 as milestone on this...

Next to the remarks inline:

* a test should be added for the `--subdir-user-modules` option somewhere, to verify that it's actually picked up
* a test should be added for `module_generator.det_home`, or actually for `get_env` (see other remarks)

Other than that, this looks really reasonable to me, you have to stop pretending to not know Python. ;-)

Let me know if you need any help in tackling the remarks (or just throw your hands up and let me take over, that's fine by me).

easybuild/tools/module_naming_scheme/categorized_hmns.py
geimer: It is called *because* of inheritance.  `det_modpath_extensions` has been overwritten in this class to take the `moduleclass` into account, but this seems overkill for user module path extensions.  Thus, the implementation provided by the superclass is called (intentionally).

easybuild/framework/easyblock.py
geimer: I think quoting only gets important if the path contains some crazy characters that are evaluated by Tcl/Lua.  That is, the quotes are more a "safety net" rather than required.  AFAIR, I did the quoting outside of `module_generator.use` to allow both `prefix` and `path` to be expressions that are evaluated when loading the modulefile.  But I agree that this may be over-engineered.

easybuild/framework/easyconfig/easyconfig.py
geimer: Copy/paste engineering ;-)  The `"..." % foo` version is used all over the place in this file, in particular in the methods surrounding this new one.  I'll change the two instances here but leave the others untouched.  I guess this should be addressed in a separate PR.

geimer:
@boegel: With respect to moving `quote_str` inside of `use` or not: I don't believe that quoting is really necessary unless paths contain "strange" characters such as spaces.  However, this probably causes havoc at other places, too.  So let's assume for a moment that we want to use quotes to be on the safe side.  Now we have two parameters to consider: `prefix` and `path`.  As `prefix` may include an statement that is executed when loading the modulefile (e.g., `$env(HOME)`), it can't be quoted in the general case.  And to avoid an inconsistency in the handling of `prefix` and `path`, I added the quoting of `path` to the call sites.

At the moment, I see three possible options on how to proceed:

 - Remove the quoting.  This is probably closest to the previous code, but may hit us when whitespaces in paths come into play, as mentioned before.
 - Move the quoting of `path` inside of `use`, as you suggested.  This would introduce the inconsistency in the parameter handling.
 - Leave it as is.

I don't have a strong preference for any of these, so I leave it up to you to decide.  What do you think?

boegel:
@geimer: OK, I now understand better why you implemented it the way you do.

How about this alternative: move the quoting of both `prefix` and `path` inside of `use`, with a check to only quote using `quote_str` if no magic like `$env(...)` is used?

That would basically be something like this inside of `use`:

```python
if `$env` not in prefix:
    prefix = quote_str(prefix)
```

You could take this one step further and add this below:

```
elif ' ' in prefix:
    raise EasyBuildError("Use of both '$env' and spaces detected in %s", prefix)
```

easybuild/tools/module_naming_scheme/categorized_hmns.py
boegel: OK, maybe make that clear with a comment in the code?

boegel:
@geimer: small conflict with `develop`, fixed with https://github.com/hpcugent/easybuild-framework/pull/1551

boegel:
@geimer: ignore the failing test for now, Jenkins is having issues... The tests get stuck, I'm looking into it.

boegel:
Jenkins: test this please

test/framework/easyblock.py
```diff
@@ -166,6 +166,59 @@ def test_fake_module_load(self):
...
+        init_config(build_options=build_options)
+        eb = EasyBlock(EasyConfig(self.eb_file))
+        eb.installdir = config.install_path()
+
+        txt = eb.make_module_extend_modpath()
+        if get_module_syntax() == 'Tcl':
+            regexs = [r'^module use ".*/modules/all/Compiler/pi/3.14/%s"$' % c for c in modclasses]
+            regexs.extend([
+                # extension for user modules is guarded
+                r'if {\[file isdirectory \[file join \$env\(HOME\) "%s/Compiler/pi/3.14"\]\]} {$' % subdir_user_modules,
```
boegel: @geimer: this test is now failing because you switched to using `conditional_statement` (slightly different formatting), I'll look into fixing it

boegel:
@geimer: broken test fixed with https://github.com/geimer/easybuild-framework/pull/7; note that it actually revealed a bug you introduced while fixing the remarks ;)

geimer:
@boegel: Did I say that I was done with addressing your remarks? ;-)

boegel:
@geimer: OK, I made some assumptions here, but it was pretty easy to fix...

boegel:
@geimer: let me know if you need help with fixing the broken test(s), or if you don't have time to follow up in the coming days (I'd like to have this merged in for EB v2.6)

test/framework/easyblock.py
geimer: Now that I've looked at this again, I fully agree.  I have no clue what led me to this change...

But... This test fails since https://github.com/geimer/easybuild-framework/commit/2335cce791cad9fb251c9ffaf4be94fa9ab74ff9 and I have no idea yet why.

geimer:
@boegel: Getting closer... The `exclude_path_to_top_of_module_tree` test still fails for me, but only for `TEST_EASYBUILD_MODULE_SYNTAX=Tcl`.  Interestingly, Lua works...  Does that make sense?  Any ideas?  I'm a bit clueless at this point...

geimer:
@boegel: OK, I believe the culprit is in `ModulesTool.modpath_extensions_for`...  Hard-coded regex to search for `module use`... I don't think this ever worked for Lua syntax...  Shouldn't the regrex be better queried from `module_generator`?  Also, the regex has to become more complicated to account for `prefix`, which can also be an expression. Urgh...

Seems like this is getting more and more unrelated to the original topic of this PR, just because it triggers already existing bugs in other areas of the code...  What do you think?

boegel:
@geimer: it sounds like you're hitting the problem that @t-brown was working on during the hackathon in Austin, cfr. PR #1474

geimer:
@boegel: Well, not really.  I just noticed as well that the Lua syntax was missing here.

The more difficult issue is that with user-local modules, the argument given to `module use` may be an expression in Lua/Tcl syntax that is evaluated at *module load time*.  My current understanding of the code is that it assumes absolute paths in plain text, that it can then pass to `path_matches`, which implicitly calls `os.path.exists` and `os.path.samefile`.  This won't work for the Lua/Tcl expression...

boegel:
@geimer: it sounds like we'll need to rework that to parse the output of `module use` rather than parsing the module file themselves?

geimer:
@boegel: Something like this, yes.  But since `module use` doesn't output anything, you probably need to load the module, see whether it extends `$MODULEPATH`, and unload it again if not. But I'm still trying to figure out what exactly `modpath_extensions_for` and `path_to_top_of_module_tree` are doing...

boegel:
@geimer: I meant to say `module show` rather than `module use`; the output of `module show` should always include the modifications to `$MODULEPATH` (unless the directory is not there in `$HOME`, in which case we don't care, I think)

geimer:
@boegel: Looks like `module show` doesn't seem to work for Tcl modules... :-(

boegel:
@geimer In what sense, exactly? And do you mean with module files written in Tcl, or using `modulecmd.tcl`?

geimer:
@boegel: I tried `module show` on a modulefile including a `module use` statement using `modulecmd.tcl`, and the output didn't include anything w.r.t. `$MODULEPATH`.  May be a non-issue, though, as module hierarchies only work reasonably well with Lmod. But still...

boegel:
@geimer: it doesn't even include the `module use` as is?

boegel:
Seems like you're right, urgh.

test module:
```
$ cat ~/.local/easybuild/modules/all/toy/0.0 
#%Module
module use /tmp
```

'`show`' outputs shows no clue of `use` or `$MODULEPATH`
```
$ tclsh tcl-modulecmd/tcl/modulecmd.tcl bash show toy/0.0                           
/Users/kehoste/.local/easybuild/modules/all/toy/0.0:

```

'`load`' does what it should though:
```
$ tclsh tcl-modulecmd/tcl/modulecmd.tcl bash load toy/0.0
_LMFILES__modshare=/Users/kehoste/.local/easybuild/modules/all/toy/0.0:1; export _LMFILES__modshare;
LOADEDMODULES_modshare=toy/0.0:1; export LOADEDMODULES_modshare;
MODULEPATH_modshare=/tmp:1:/Users/kehoste/.local/easybuild/modules/all:1; export MODULEPATH_modshare;
_LMFILES_=/Users/kehoste/.local/easybuild/modules/all/toy/0.0; export _LMFILES_;
LOADEDMODULES=:toy/0.0; export LOADEDMODULES;
MODULEPATH=/tmp:/Users/kehoste/.local/easybuild/modules/all; export MODULEPATH;
```

'`show`' output with Lmod:
```
$ lmod bash show toy/0.0
LMOD_DEFAULT_MODULEPATH="/Users/kehoste/.local/easybuild/modules/all";
export LMOD_DEFAULT_MODULEPATH;
MODULEPATH="/Users/kehoste/.local/easybuild/modules/all";
export MODULEPATH;
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
   /Users/kehoste/.local/easybuild/modules/all/toy/0.0:
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
prepend_path("MODULEPATH","/tmp")

LMOD_DEFAULT_MODULEPATH="/Users/kehoste/.local/easybuild/modules/all";
export LMOD_DEFAULT_MODULEPATH;
MODULEPATH="/Users/kehoste/.local/easybuild/modules/all";
export MODULEPATH;
_ModuleTable001_="X01vZHVsZVRhYmxlXz17WyJhY3RpdmVTaXplIl09MCxiYXNlTXBhdGhBPXsiL1VzZXJzL2tlaG9zdGUvLmxvY2FsL2Vhc3lidWlsZC9tb2R1bGVzL2FsbCIsfSxbImNfcmVidWlsZFRpbWUiXT1mYWxzZSxbImNfc2hvcnRUaW1lIl09ZmFsc2UsZmFtaWx5PXt9LGluYWN0aXZlPXt9LG1UPXt9LG1wYXRoQT17Ii9Vc2Vycy9rZWhvc3RlLy5sb2NhbC9lYXN5YnVpbGQvbW9kdWxlcy9hbGwiLH0sWyJzeXN0ZW1CYXNlTVBBVEgiXT0iL1VzZXJzL2tlaG9zdGUvLmxvY2FsL2Vhc3lidWlsZC9tb2R1bGVzL2FsbCIsWyJ2ZXJzaW9uIl09Mix9";
export _ModuleTable001_;
_ModuleTable_Sz_="1";
export _ModuleTable_Sz_;
```

'`show`' output with Tcl/C env mods:
```
$ modulecmd bash show toy/0.0
/Users/kehoste/.local/easybuild/modules/all/toy/0.0:

module		 use /tmp 
```

geimer:
@boegel: I've adjusted the regex in `ModulesTool.modpath_extensions_for` to account for quoted `module use` statements.  But this is obviously not the full solution.  I suggest to fix the `modpath_extensions_for` routine and its unit test (which doesn't check Lua module syntax) in a separate PR.

W.r.t. `''.join(...)` vs. `'\n'.join(...)`: The latter feels more natural, agreed.  But it introduces a number of extra newlines as, for example, `conditional_statement` already includes a trailing newline.  As fixing this is again something that can have wide-spread impacts, addressing this in a separate PR would IMHO be preferred.

Apart from this, I believe all your comments have been addressed.  Please take another look.

easybuild/tools/modules.py
```diff
@@ -588,7 +588,7 @@ def modpath_extensions_for(self, mod_names):
         modpath_exts = {}
         for mod_name in mod_names:
             modtxt = self.read_module_file(mod_name)
-            useregex = re.compile(r"^\s*module\s+use\s+(\S+)", re.M)
+            useregex = re.compile(r'^\s*module\s+use\s+"?([^" \t\n\r\f\v]+)"?', re.M)
```
boegel: @geimer: please use `\s` instead of ` \t\n\r\f\v`, cfr. https://docs.python.org/2/library/re.html

```
\s
When the UNICODE flag is not specified, it matches any whitespace character,
this is equivalent to the set [ \t\n\r\f\v]. 
```

boegel:
@geimer: one single small style remark, looks excellent otherwise!

W.r.t. `modpath_extensions_for`: can you clarify what's left to fix there? Just that it's blissfully unaware of Lua syntax (which is basically what #1474 is about), or is there more to it?

easybuild/tools/modules.py
geimer: @boegel: I wasn't aware that one can use character classes within character classes.  But makes sense.

boegel:
tests failed due to a temporary issue with GitHub, it seems, let's try again

Jenkins: test this please

geimer:
@boegel: W.r.t. `modpath_extensions_for`: With the current usage model, I believe things should be fine.  However, if at some point in the future someone uses the `prefix` argument of the `use` method also for module paths that need to be considered in the `modpath_extensions_for` routine, things will break (for a potential use case, see below).  Thus it's more about being future-proof.

And now the potential use case:  Assume that we have a system-wide software installation using EB in `<prefix>`, using a hierarchical module naming scheme.  In the long run, I would like to be able to extend it with my own local software/module collection in `$HOME` (this is what this PR is good for), but also use EB for installing this stuff (e.g., using `--installpath=$HOME/.local`).  One example might be to install a special debugging version or a version with a specific patch of an MPI library, and then some software on top of it.  Then, the module paths in my `$HOME` and the modpath extensions of modules installed there need to be taken into account in the `modpath_extensions_for`, right?

I guess a bunch of other things may need to be resolved before the use case above will work with EB.  But it is something I would like to see at some point (and I know at least one other guy in our group who already asked whether this would be possible).

boegel:
@geimer: OK, I see what you're saying.

The current implementation of `modpath_extensions_for` only looks for `module use "..."`, so it will miss the `module use [ file join ... ]` entries we have there for user-local module path extensions.

For your use case to work, we will need a way of dealing with those too.
I'm not sure how we can do that since `module show` doesn't seem to be an option that always works, but we can figure something out (e.g. push what we get through Tcl first to evaluate it, or whatever, or just look at how `$MODULEPATH` changes if we fake-load the module).

Can you please open a separate issue on this with largely a copy-paste of your last comment, and a pointer to this PR?

This is good to go btw, thanks *a lot* for your work and dedicated follow-up on this, much appreciated.

easybuild/tools/module_naming_scheme/mns.py
```diff
@@ -126,6 +126,18 @@ def det_modpath_extensions(self, ec):
         # by default: an empty list of subdirectories to extend $MODULEPATH with
         return []
 
+    def det_user_modpath_extensions(self, ec):
```
ocaisa: @geimer  @boegel This does not seem to be respecting `EASYBUILD_SUFFIX_MODULES_PATH` (which defaults to `all`):
```
conflict("GCCcore")
prepend_path("MODULEPATH", "/usr/local/software/jureca/Stages/Devel-2017a/modules/all/Compiler/GCCcore/5.4.0")
if isDir(pathJoin(os.getenv("HOME"), ".local/EasyBuild/modules/Compiler/GCCcore/5.4.0")) then
    prepend_path("MODULEPATH", pathJoin(os.getenv("HOME"), ".local/EasyBuild/modules/Compiler/GCCcore/5.4.0"))
end
```
Notice the missing `all` in the local path.

boegel: @ocaisa what was the full EasyBuild configuration here (`eb --show-config`)?

ocaisa:
```
-bash-4.2$ eb --show-config
#
# Current EasyBuild configuration
# (C: command line argument, D: default value, E: environment variable, F: configuration file)
#
buildpath                      (E) = /dev/shm/ocaisa
experimental                   (E) = True
fixed-installdir-naming-scheme (E) = True
group-writable-installdir      (E) = True
hide-deps                      (E) = ANTLR, APR, APR-util, AT-SPI2-ATK, AT-SPI2-core, ATK, Autoconf, Automake, adwaita-icon-theme, ant, assimp, Bison, babl, binutils, byacc, bzip2, CUSP, Coreutils, cairo, configurable-http-proxy, DB, DBus, DocBook-XML, Dyninst, dbus-glib, damageproto, ETSF_IO, Exiv2, eudev, expat, FFmpeg, FLTK, FTGL, fixesproto, fontsproto, fontconfig, freeglut, freetype, GCCcore, GDAL, GEGL, GL2PS, GLEW, GLib, GLPK, GPC, GObject-Introspection, GTI, GTK+, GTS, Gdk-Pixbuf, Ghostscript, GraphicsMagick, GtkSourceView, g2clib, g2lib, gc, gexiv2, glproto, gperf, guile, grib_api, gsettings-desktop-schemas, gettext, HarfBuzz, icc, ifort, inputproto, intltool, itstool, JUnit, JSON-C, JSON-GLib, JasPer, jhbuild, kbproto, LZO, LibTIFF, LibUUID, Libint, LittleCMS, libGLU, libICE, libSM, libX11, libXau, libXaw, libXcursor, libXdamage, libXdmcp, libXext, libXfixes, libXfont, libXft, libXi, libXinerama, libXmu, libXpm, libXrandr, libXrender, libXt, libXtst, libcerf, libcroco, libctl, libdap, libdrm, libdwarf, libelf, libepoxy, libevent, libffi, libfontenc, libgd, libgeotiff, libglade, libidn, libjpeg-turbo, libmatheval, libmypaint, libpng, libpciaccess, libpthread-stubs, libreadline, librsvg, libtool, libunistring, libunwind, libyaml, libxcb, libxkbcommon, libxml2, libxslt, libyuv, M4, Mesa, makedepend, motif, msgpack-c, NASM, ncurses, nettle, nodejs, nvenc_sdk, nvidia, OPARI2, OTF2, PCRE, PDT, PROJ, Pango, Pmw, PnMPI, PyCairo, PyGObject, Python-Xpra, pixman, pkg-config, pkgconfig, popt, pscom, Qhull, Qt, Qt5, qrupdate, randrproto, recordproto, renderproto, S-Lang, SCons, SIP, SQLite, SWIG, Serf, Szip, scrollkeeper, Tcl, Tk, texinfo, UDUNITS, util-linux, vpx, wxPropertyGrid, wxWidgets, XML-Parser, XZ, XKeyboardConfig, x264, x265, xbitmaps, xcb-proto, xcb-util, xcb-util-image, xcb-util-keysyms, xcb-util-renderutil, xcb-util-wm, xextproto, xineramaproto, xorg-macros, xprop, xproto, xtrans, YAXT, Yasm, zlib
hide-toolchains                (E) = GCCcore
include-easyblocks             (E) = /work/zam/swmanage/EasyBuild/Custom_EasyBlocks/Devel-2017a/*.py, /work/zam/swmanage/EasyBuild/Custom_EasyBlocks/Devel-2017a/generic/*.py
include-module-naming-schemes  (E) = /work/zam/swmanage/EasyBuild/Custom_MNS/Devel-2017a/*.py
include-toolchains             (E) = /work/zam/swmanage/EasyBuild/Custom_Toolchains/Devel-2017a/*.py
installpath                    (E) = /usr/local/software/jureca/Stages/Devel-2017a
job-backend-config             (E) = /usr/local/software/FZJ/gc3pie.cfg
job-cores                      (E) = 48
job-max-walltime               (E) = 1
minimal-toolchains             (E) = True
module-naming-scheme           (E) = CustomHierarchicalMNS
packagepath                    (E) = /usr/local/software/jureca/Stages/Devel-2017a/packages
prefix                         (E) = /usr/local/software/jureca/Stages/Devel-2017a
repositorypath                 (E) = /usr/local/software/jureca/Stages/Devel-2017a/eb_repo
robot                          (E) = /work/zam/swmanage/EasyBuild/Golden_Repo/Devel-2017a, /usr/local/software/jureca/Stages/Devel-2017a/eb_repo, /work/zam/swmanage/EasyBuild/Golden_Repo/Devel-2017a
robot-paths                    (E) = /work/zam/swmanage/EasyBuild/Golden_Repo/Devel-2017a, /usr/local/software/jureca/Stages/Devel-2017a/eb_repo, /work/zam/swmanage/EasyBuild/Golden_Repo/Devel-2017a
set-gid-bit                    (E) = True
sourcepath                     (E) = /work/zam/swmanage/EasyBuild/sources
subdir-user-modules            (E) = .local/EasyBuild/modules
umask                          (E) = 002
use-existing-modules           (E) = True
```

ocaisa: I get the same effect with using `HierarchicalMNS`

ocaisa: This occurs with a vanilla instance and the command line:
```
eb --module-naming-scheme=HierarchicalMNS --subdir-user-modules=.local/EasyBuild/modules GCCcore-5.4.0.eb -x
```

boegel: fix implemented in #2250

Issue #1473, opened 2015-11-13
Modules modpath_extensions_for doesn't find lmod dependencies


t-brown:
The regex within the modules modpath_extensions_for function to find the module dependencies when using lua module files.

boegel:
@t-brown: we can close this since it's being handled in #1474?

t-brown:
Closing issues as it is being addressed in #1474.

PR #1474, opened 2015-11-13
Bug fix for modpath_extensions_for function.

t-brown:
Updating the regex within the modules modpath_extensions_for
function to find the module dependencies when using lua module
files.

This addresses issue #1473

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -587,9 +587,15 @@ class ModulesTool(object):
 
         modpath_exts = {}
         for mod_name in mod_names:
-            modtxt = self.read_module_file(mod_name)
-            useregex = re.compile(r"^\s*module\s+use\s+(\S+)", re.M)
-            exts = useregex.findall(modtxt)
+            # Should be module show so as to catch conditional loads.
+            modtxt = self.show(mod_name)
+            useregex = re.compile(r"^\s*module\s+use\s+(\S+)|^\s*prepend_path\(\"MODULEPATH\",\s*\"(\S+)\"", re.M)
+            exts = []
+            for i in  useregex.finditer(modtxt):
+                    if i.group(1):
+                        exts.append(i.group(1))
+                    elif i.group(2):
+                        exts.append(i.group(2))
 
             self.log.debug("Found $MODULEPATH extensions for %s: %s" % (mod_name, exts))
             modpath_exts.update({mod_name: exts})


boegel:
Jenkins: ok to test

easybuild/tools/modules.py
```diff
@@ -587,9 +587,15 @@ def modpath_extensions_for(self, mod_names):
 
         modpath_exts = {}
         for mod_name in mod_names:
-            modtxt = self.read_module_file(mod_name)
-            useregex = re.compile(r"^\s*module\s+use\s+(\S+)", re.M)
-            exts = useregex.findall(modtxt)
+            # Should be module show so as to catch conditional loads.
```
boegel: not loads, use statements

easybuild/tools/modules.py
```diff
@@ -587,9 +587,15 @@ def modpath_extensions_for(self, mod_names):
...
+            modtxt = self.show(mod_name)
+            useregex = re.compile(r"^\s*module\s+use\s+(\S+)|^\s*prepend_path\(\"MODULEPATH\",\s*\"(\S+)\"", re.M)
```
boegel: hmm, do we have to consider single quotes here too in the `prepend_path`?

easybuild/tools/modules.py
```diff
@@ -587,9 +587,15 @@ def modpath_extensions_for(self, mod_names):
...
+            exts = []
+            for i in  useregex.finditer(modtxt):
+                    if i.group(1):
+                        exts.append(i.group(1))
+                    elif i.group(2):
+                        exts.append(i.group(2))
```
boegel: hmm, `findall` doesn't do the job anymore? there must be a better/easier way than this...

boegel:
@t-brown: this is only a partial fix, right? there also was the issue where the `.lua` extension was not being taken into account to determine the relative subdirectory in which the module file is located?

easybuild/tools/modules.py
t-brown: Yes, although since we are creating the line, we might be able to get away with only double.

boegel:
@t-brown: what's the status of this? you mentioned that there still was some problem here?

boegel:
This will need to be revisited thoroughly when #1472 is merged in, since they're getting in the way of each other.

@t-brown: I propose I take this over from you, since things just got more complicated (only a little bit though)?
Can you provide details on what exact problem you ran into that resulted in blocking this?

t-brown:
It seems that when using `recursive-module-unload` with hierarchical modules:
    [config]
    buildpath=/dev/shm
    module-naming-scheme=HierarchicalMNS
    modules-tool=Lmod
    module-syntax=Lua
    recursive-module-unload=1

The module files generated at the second hierarchical level include it's parent. For example, `/GCC/4.9.3-binutils-2.25/Autoconf/2.69` contains:
    load("GCC/4.9.3-binutils-2.25")
    load("binutils/2.25")
    load("M4/1.4.17")
and `GCC/4.9.3-binutils-2.25/Automake/1.15` contains:
    load("GCC/4.9.3-binutils-2.25")
    load("binutils/2.25")
    load("Autoconf/2.69")

So when build `foss-2015b.eb` (for example), loads `GCC/4.9.3-binutils-2.25` then loads `Autoconf/2.69` and then when trying to load `Automake/1.15`, hits the `load("GCC/4.9.3-binutils-2.25")` which causes an unload, then the load fails:
    Lmod has detected the following error:  These module(s) exist but cannot be loaded as requested: "Autoconf/2.69", "binutils/2.25", "GCC/4.9.3-binutils-2.25"

If you comment out the `load("GCC/4.9.3-binutils-2.25")` from all modules under that hierarchy, it loads and works.

boegel:
@t-brown: hah! OK, that is very wrong indeed, the `load("GCC/...")` shouldn't be there for a module installed using `GCC` as a toolchain

This is supposed to be handled by https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/framework/easyblock.py#L901, which (through `path_to_top_of_module_tree`) relies on `modpath_extensions_for`.

So, that means the `use` regex must be correct.

What does a debug log for e.g. `Autoconf` include for `List of retained dependencies:`?

boegel:
@t-brown: ping?

boegel:
@t-brown: did you have any time for this? If not, I'll probably try and get it fixed myself (on top of your work/feedback).

boegel:
@t-brown final ping before I fork this off myself to get it fixed for EB v2.8.0 (i.e., this week)

boegel:
fixed via #1766, thanks for your work on this @t-brown!

PR #1479, opened 2015-11-18
[.yeb] alternate dependencies format

Caylo:
builds on #1456. 

new format as specified in the [docs](http://caylo-easybuild.readthedocs.org/en/yeb-docs/Writing_yeb_easyconfig_files.html#dependencies) (PR https://github.com/hpcugent/easybuild/pull/165)


boegel:
Jenkins: ok to test

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -94,6 +95,11 @@ def parse(self, txt):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
+        # toolchain and dependecy param value check, convert to dict if alt format is used
+        check_type_of_param_value('toolchain', self.parsed_yeb['toolchain'], auto_convert=True)
+        for dep in self.parsed_yeb.get('dependencies', ''):
+            check_type_of_param_value('dependencies', dep, auto_convert=True)
```
boegel: no need, auto convert is already done in parser

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +111,64 @@ def convert_value_type(val, typ):
...
+    else:
+        errstr = "Conversion of %s (type %s) to toolchain %s is not supported"
+        raise EasyBuildError(errstr % (tcspec, type(tcspec), convert_to))
+
+    return res
+
+
+def to_dependency(dep):
+    """
+    Convert a dependency dict from the new YAML format {'name': 'lib, 1.2.8', 'toolchain': 'gcc, 4.8.2'}
```
boegel: it comes out as `name: 'lib, 1.2.8'`? that's wrong, since it also contains `version`?

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +111,64 @@ def convert_value_type(val, typ):
...
+    to a tuple in the dependency format
+    """
+    tup = ()
+    if isinstance(dep, dict):
+        if 'name' in dep:
+            tup += to_toolchain(dep['name'], convert_to='tuple')
```
boegel: to_toolchain?

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +111,64 @@ def convert_value_type(val, typ):
...
+
+    return res
+
+
+def to_dependency(dep):
+    """
+    Convert a dependency dict from the new YAML format {'name': 'lib, 1.2.8', 'toolchain': 'gcc, 4.8.2'}
+    to a tuple in the dependency format
+    """
+    tup = ()
```
boegel: only convert to a tuple at the end, every time you 'append' to a tuple with `+=`, you're making a new copy

use a list, and `foo.append` or `foo.extend`, and then return `tuple(foo)` at the end

easybuild/framework/easyconfig/types.py
```diff
@@ -116,3 +111,64 @@ def convert_value_type(val, typ):
...
+            tup += to_toolchain(dep['name'], convert_to='tuple')
+        else:
+            raise EasyBuildError("Can not parse dependency without name and version: %s", dep)
+
+        tup += (dep.get('versionsuffix', ''),)
+        if 'toolchain' in dep:
+            tcspec = to_toolchain(dep['toolchain'], convert_to='tuple')
+            tup += (tcspec,)
+        else:
+            tup += ('',)
```
boegel: what's this about? empty versionsuffix?

the toolchain is the last entry in the tuple, so this doesn't make sense, you need to do this if there's no versionsuffix set

test/framework/type_checking.py
```diff
@@ -85,6 +85,42 @@ class Foo():
...
+        self.assertErrorRegex(EasyBuildError, r"Conversion of .* \(type .*\) to toolchain dict is not supported",
+            to_toolchain, ('intel', '2015a'))
+
+        # wrong number of elements
+        errstr = "Can not convert .* to name and version .*. Expected 2 elements"
+        self.assertErrorRegex(EasyBuildError, errstr, to_toolchain, "intel, 2015, a")
+        self.assertErrorRegex(EasyBuildError, errstr, to_toolchain, "intel")
+        self.assertErrorRegex(EasyBuildError, errstr, to_toolchain, ['gcc', '4', '7'])
+
+    def to_dependency(self):
```
boegel: you need to call this `test_to_dependency`, otherwise the test is not run (only stuff that starts with `test` is run as a test)

easybuild/framework/easyconfig/types.py
Caylo: no, this is the input. yaml parses the new format for dependencies to a dict :)

test/framework/type_checking.py
```diff
@@ -85,6 +85,42 @@ class Foo():
...
+        """ Test dependency dict to tuple conversion """
+        # normal cases
+        self.assertEqual(to_dependency({'name': 'lib, 1.2.8'}), ('lib', '1.2.8'))
+        self.assertEqual(to_dependency({'name': 'lib, 1.2.8', 'toolchain': 'gcc, 4.8.2'}),
+            ('lib', '1.2.8', '', ('gcc', '4.8.2')))
+        self.assertEqual(to_dependency({'name': ['lib', '1.2.8'], 'toolchain': ['gcc', '4.8.2']}),
+            ('lib', '1.2.8', '', ('gcc', '4.8.2')))
+        self.assertEqual(to_dependency({'name': 'foo, 1.3', 'versionsuffix': '-bar'}), ('foo', '1.3', '-bar', ()))
```
boegel: this shouldn't have `()` for toolchain, it should be a tuple with 3 elements

test/framework/type_checking.py
```diff
@@ -85,6 +85,42 @@ class Foo():
...
+
```
boegel: please add a test case with both toolchain and versionsuffix specified

easybuild/framework/easyconfig/types.py
boegel: well, the key says 'name', which is wrong, it should be 'name_version' ?

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -94,6 +95,9 @@ def parse(self, txt):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
+        # parse dependencies upfront; this cant be done with simple type checking
```
boegel: `s/cant/can't/` (or "can not")

also mention why it can not be done with simple type checking?

easybuild/framework/easyconfig/types.py
```diff
@@ -34,7 +34,7 @@
 
 
 # easy types, that can be verified with isinstance
-EASY_TYPES = [basestring, int, dict]
+EASY_TYPES = [basestring, int, dict, tuple, list]
```
boegel: sort these alphabetically, order doesn't matter for this list?

easybuild/framework/easyconfig/types.py
```diff
@@ -112,7 +112,7 @@ def convert_value_type(val, typ):
     return res
 
 
-def to_toolchain(tcspec):
+def to_nameversion_dict(tcspec, convert_to='dict'):
```
boegel: don't use magic strings, introduce a constant `TYPE_DICT` to `'dict'`

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +124,63 @@ def to_toolchain(tcspec):
 
     if isinstance(tcspec, list):
         if len(tcspec) == 2:
-            res = {'name': tcspec[0].strip(), 'version': tcspec[1].strip()}
+            if convert_to == 'tuple':
```
boegel: `TYPE_TUPLE`

easybuild/framework/easyconfig/types.py
boegel: also, depending on `convert_to`, this may yield a `tuple` rather than a `dict`, so rename this to `to_name_version`?

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +124,63 @@ def to_toolchain(tcspec):
...
+                res = (tcspec[0].strip(), tcspec[1].strip())
+            else:
+                res = {'name': tcspec[0].strip(), 'version': tcspec[1].strip()}
         else:
-            raise EasyBuildError("Can not convert list %s to toolchain dict. Expected 2 elements", tcspec)
+            errstr = "Can not convert list %s to name and version %s. Expected 2 elements"
+            raise EasyBuildError(errstr % (tcspec, convert_to))
```
boegel: use

```python
raise EasyBuildError(errstr, tcspec, convert_to)
```

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +124,63 @@ def to_toolchain(tcspec):
...
     else:
-        raise EasyBuildError("Conversion of %s (type %s) to toolchain dict is not supported", tcspec, type(tcspec))
+        errstr = "Conversion of %s (type %s) to name and version %s is not supported"
+        raise EasyBuildError(errstr % (tcspec, type(tcspec), convert_to))
```
boegel: same here:

```python
raise EasyBuildError(errstr, tcspec, type(tcspec), convert_to)
```

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +124,63 @@ def to_toolchain(tcspec):
...
+    Convert a list of dependencies in yeb format to a list of dependency tuples
+    """
+    dep_list = []
+    for el in dep:
+        dep_list.append(to_dependency(el))
+
+    return dep_list
+
+
+def to_dependency(dep):
```
boegel: move this up, above `to_dependencies`?

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +124,63 @@ def to_toolchain(tcspec):
...
-# this uses to_toolchain, so it needs to be at the bottom of the module
+def to_dependencies(dep):
+    """
+    Convert a list of dependencies in yeb format to a list of dependency tuples
+    """
+    dep_list = []
+    for el in dep:
+        dep_list.append(to_dependency(el))
+
+    return dep_list
```
boegel: easier/shorter:

```python
def to_dependencies(dep_spec_list):
    """..."""
    return [to_dependency(dep_spec) for dep_spec in dep_spec_list]
```

easybuild/framework/easyconfig/types.py
boegel: ping?

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +124,63 @@ def to_toolchain(tcspec):
...
+
+
+def to_dependency(dep):
+    """
+    Convert a dependency dict from the new YAML format {'name': 'lib, 1.2.8', 'toolchain': 'gcc, 4.8.2'}
+    to a tuple in the dependency format
+    """
+    depspec = []
+    if isinstance(dep, dict):
+        if 'name' in dep:
```
boegel: `s/'name'/'name_version'/` ?

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +124,63 @@ def to_toolchain(tcspec):
...
+            # if toolchain is specified, versionsuffix is empty element
+            depspec.append('')
+
+        if 'toolchain' in dep:
+            tcspec = to_nameversion_dict(dep['toolchain'], convert_to='tuple')
+            depspec.append(tcspec)
+
+    else:
+        raise EasyBuildError("Can not convert %s (type %s) to dependency tuple", dep, type(dep))
+    return tuple(depspec)
```
boegel: insert blank link above

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +124,63 @@ def to_toolchain(tcspec):
...
+        if 'versionsuffix' in dep:
+            depspec.append(dep['versionsuffix'])
+        elif 'toolchain' in dep:
+            # if toolchain is specified, versionsuffix is empty element
+            depspec.append('')
+
+        if 'toolchain' in dep:
+            tcspec = to_nameversion_dict(dep['toolchain'], convert_to='tuple')
+            depspec.append(tcspec)
+
```
boegel: too much whitespace?

test/framework/easyconfigs/yeb/goolf-1.4.10.yeb
```diff
@@ -3,7 +3,7 @@ _internal_variables_:
 
     - &comp_name GCC
     - &comp_version 4.7.2
-    - &comp [*comp_name, *comp_version]
+    - &comp !join [*comp_name, ', ', *comp_version]
```
boegel: this looks very weird now...

maybe it's better to drop `*comp`, and just use `*comp_name, *comp_version` below?

or use this here (does that work?):

```
&comp *comp_name, *comp_version
```

test/framework/easyconfigs/yeb/goolf-1.4.10.yeb
```diff
@@ -28,12 +28,17 @@ toolchain: {name: dummy, version: dummy}
...
-    [OpenMPI, 1.6.4, '', *comp],
-    [*blaslib, *blasver, *blas_suff, *comp_mpi_tc],
-    [FFTW, 3.3.3, '', *comp_mpi_tc],
-    [ScaLAPACK, 2.0.2, !join [-, *blas, *blas_suff], *comp_mpi_tc]
-]
+dependencies:
+    - name: *comp
+    - name: OpenMPI, 1.6.4
+      toolchain: *comp
+    - name: [*blaslib, *blasver]
```
boegel: are the `[`/`]` needed?

test/framework/type_checking.py
```diff
@@ -85,23 +86,43 @@ class Foo():
             pass
         self.assertErrorRegex(EasyBuildError, "No conversion function available", convert_value_type, None, Foo)
 
-    def test_to_toolchain(self):
+    def test_to_nameversion_dict(self):
         """ Test toolchain string to dict conversion """
         # normal cases
-        self.assertEqual(to_toolchain("intel, 2015a"), {'name': 'intel', 'version': '2015a'})
-        self.assertEqual(to_toolchain(['gcc', '4.7']), {'name': 'gcc', 'version': '4.7'})
+        self.assertEqual(to_nameversion_dict("intel, 2015a"), {'name': 'intel', 'version': '2015a'})
```
boegel: add comment that default is to convert to a name/version dict

boegel:
Jenkins: test this please

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +127,59 @@ def to_toolchain(tcspec):
 
     if isinstance(tcspec, list):
         if len(tcspec) == 2:
-            res = {'name': tcspec[0].strip(), 'version': tcspec[1].strip()}
+            if convert_to == TYPE_TUPLE:
+                res = (tcspec[0].strip(), tcspec[1].strip())
+            else:
+                res = {'name': tcspec[0].strip(), 'version': tcspec[1].strip()}
```
boegel: make this:

```python
if convert_to == TYPE_DICT:
    ...
elif convert_to == TYPE_TUPLE:
   ...
else:
    raise EasyBuildError("Unknown target type specified: %s", convert_to)
```

test/framework/easyconfigs/yeb/goolf-1.4.10.yeb
```diff
@@ -28,12 +28,17 @@ toolchain: {name: dummy, version: dummy}
...
-    *comp,
-    [OpenMPI, 1.6.4, '', *comp],
-    [*blaslib, *blasver, *blas_suff, *comp_mpi_tc],
-    [FFTW, 3.3.3, '', *comp_mpi_tc],
-    [ScaLAPACK, 2.0.2, !join [-, *blas, *blas_suff], *comp_mpi_tc]
-]
+dependencies:
+    - name_version: *comp
+    - name_version: OpenMPI, 1.6.4
+      toolchain: *comp
```
boegel: @Caylo: please check if we can make it work as follows:

```yaml
- *comp_name: *comp_version
- OpenMPI: 1.6.4
  toolchain: *comp
...
```

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +127,59 @@ def to_toolchain(tcspec):
...
 
 
-# this uses to_toolchain, so it needs to be at the bottom of the module
+def to_dependency(dep):
+    """
+    Convert a dependency dict from the new YAML format {'name_version': 'lib, 1.2.8', 'toolchain': 'gcc, 4.8.2'}
+    to a tuple in the dependency format
+    """
+    depspec = []
+    if isinstance(dep, dict):
```
boegel: the `_parse_dependency` method that will eventually parse all the values in `dependencies` can handle both a tuple value as the one you're creating here, or a `dict` value that contains `name`, `version`, `versionsuffix` and `toolchain` keys, so you should be able to simplify this significantly by just taking the dict and making sure `name` and `version` are what they should be; there's no need to mimic the tuple value we're usually using in `.eb` files

easybuild/framework/easyconfig/types.py
```diff
@@ -124,20 +127,59 @@ def to_toolchain(tcspec):
 
     if isinstance(tcspec, list):
         if len(tcspec) == 2:
-            res = {'name': tcspec[0].strip(), 'version': tcspec[1].strip()}
+            if convert_to == TYPE_TUPLE:
```
boegel: if you rework `to_depenency` and no longer need this case, kill it with fire (and maybe rename this function back to `to_name_version_dict` ;-))

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -94,6 +95,10 @@ def parse(self, txt):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
+        # parse dependencies upfront; this can't be done with simple type checking as it is already a list,
+        # but needs to be a list of tuples instead of dicts.
```
boegel: wasn't the actual reason more like making sure that the conversion happens early enough?

we can detect a list of dicts and acts accordingly in the type conversion stuff, that's not the real problem?

easybuild/framework/easyconfig/types.py
```diff
@@ -126,18 +126,53 @@ def to_toolchain(tcspec):
...
+    to a dependency dictionary
+    """
+    depspec = {}
+    if isinstance(dep, dict):
+        for key, value in dep.items():
+            if key == 'versionsuffix':
+                depspec['versionsuffix'] = value
+            elif key == 'toolchain':
+                depspec['toolchain'] = to_name_version_dict(value)
+            elif 'name' not in depspec:
```
boegel: I think you need a separate boolean value here to check, something like:

```python
if not found_name_version:
    depspec.update({'name': key, 'version': value})
    found_name_version = True
```

to protect against situations like:

```yaml
- name: foo
  version: 1.2.3
  toolchain: GCC, 4,9.2
  versionsuffix: -blah
```

easybuild/framework/easyconfig/types.py
```diff
@@ -126,18 +126,53 @@ def to_toolchain(tcspec):
...
+                depspec.update({'name': key, 'version': value})
+            else:
+                raise EasyBuildError("Found unexpected key, value pair: %s, %s", key, value)
+
+        if 'name' not in depspec:
```
boegel:
```python
if not found_name_version:
```

easybuild/framework/easyconfig/types.py
```diff
@@ -126,18 +126,53 @@ def to_toolchain(tcspec):
...
+            raise EasyBuildError("Can not parse dependency without name and version: %s", dep)
```
boegel: `"Could not find name/version in dependency specification %s" % dep)`

easybuild/framework/easyconfig/format/yeb.py
boegel: this?

easybuild/framework/easyconfig/types.py
```diff
@@ -126,18 +126,53 @@ def to_toolchain(tcspec):
...
+
+
+# this uses functions defined in this module, so it needs to be at the bottom of the module
 TYPE_CONVERSION_FUNCTIONS = {
     basestring: str,
     float: float,
     int: int,
     str: str,
-    dict: to_toolchain,
+    dict: to_name_version_dict,
```
boegel: hmm, this implies that the auto-conversion will try to convert everything that should be a `dict` but is not using `to_name_version_dict`, which is probably not what we want in the long term... we need a custom type `name_version_dict`?

test/framework/easyconfigs/yeb/goolf-1.4.10.yeb
```diff
@@ -28,12 +28,17 @@ toolchain: {name: dummy, version: dummy}
...
+    - OpenMPI: 1.6.4
+      toolchain: *comp
+    - *blaslib: *blasver
+      versionsuffix: *blas_suff
+      toolchain: *comp_mpi_tc
+    - FFTW: 3.3.3
+      toolchain: *comp_mpi_tc
+    - ScaLAPACK: 2.0.2
+      versionsuffix: !join [-, *blas, *blas_suff]
+      toolchain: *comp_mpi_tc
```
boegel: this looks a lot clear, especially keeping in mind that it doesn't get nastier than this (specifying both `versionsuffix` and `toolchain` for almost all dependencies is definitely the exception, not the rule)

boegel:
@wpoely86: can you give this a review, since it now includes quite a bit of my code too?

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -94,6 +95,10 @@ def parse(self, txt):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
+        # make sure dependencies are in a format that can be handled by the EasyConfig class
+        # detecting that auto-converting the list-of-dicts value that results from parsing a .yeb is needed is tricky
+        if 'dependencies' in self.parsed_yeb:
+            self.parsed_yeb['dependencies'] = to_dependencies(self.parsed_yeb['dependencies'])
```
boegel: @Caylo: can you look into a follow-up PR that implements type checking for `dependencies`?

With the new functionality added in here (`is_value_of_type`), it should be fairly easy to get that working.

Then we can remove these two lines, and apply the same conversion trivially to also `builddependencies` and `hiddendependencies` (see http://easybuild.readthedocs.org/en/latest/Writing_easyconfig_files.html#dependencies)

easybuild/framework/easyconfig/format/yeb.py
```diff
@@ -94,6 +95,10 @@ def parse(self, txt):
         """
         txt = self._inject_constants_dict(txt)
         self.parsed_yeb = yaml.load(txt)
+        # make sure dependencies are in a format that can be handled by the EasyConfig class
+        # detecting that auto-converting the list-of-dicts value that results from parsing a .yeb is needed is tricky
+        if 'dependencies' in self.parsed_yeb:
+            self.parsed_yeb['dependencies'] = [to_dependency(d) for d in self.parsed_yeb['dependencies']]
```
boegel: @Caylo: can you look into a follow-up PR that implements type checking for `dependencies`?

With the new functionality added in here (`is_value_of_type`), it should be fairly easy to get that working.

Then we can remove these two lines, and apply the same conversion trivially to also `builddependencies` and `hiddendependencies` (see http://easybuild.readthedocs.org/en/latest/Writing_easyconfig_files.html#dependencies)

boegel:
Going in, thanks @Caylo!

PR #1486, opened 2015-11-24
Change polling loops so they keep going only if error code is None

valtandor:
This small but important fix deals with the situation where the UNIX command generates an exit code less than zero, as happens if the UNIX command dies by reason of receiving a signal. While I would expect that receiving such an exit code causes the loop to exit, previous behaviour had it looping endlessly until the Python process was interrupted by a user or killed via some other means. Now, the loop should only keep running if, in fact, the UNIX command has not yet exited (by means of a signal or otherwise).

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -122,7 +122,7 @@ def run_cmd(cmd, log_ok=True, log_all=False, simple=False, inp=None, regexp=True
 
     ec = p.poll()
     stdouterr = ''
-    while ec < 0:
+    while ec is None:
         # need to read from time to time.
         # - otherwise the stdout/stderr buffer gets filled and it all stops working
         output = p.stdout.read(readSize)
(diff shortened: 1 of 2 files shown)


valtandor:
Yes.

valtandor:
For more information about the return value from a poll() call, please see https://docs.python.org/2/library/subprocess.html#subprocess.Popen.returncode.

boegel:
Jenkins: ok to test

boegel:
@valtandor: waw, nice catch! I'm surprised this hasn't popped up earlier...

The only reason why this went unnoticed is because `None < 0` results in `True` in Python >_<

boegel:
@valtandor: please merge https://github.com/valtandor/easybuild-framework/pull/1 to include the unit test that checks for this problem, to avoid that we re-introduce it

valtandor:
@boegel: Merge done.

boegel:
Going in, thanks @valtandor!

PR #1487, opened 2015-11-25
Add support for module_classes.

blappm:
It's now possible to specify multiple classes instead of just one. This change still allows the old naming
scheme:

Old:	moduleclass = 'data'
New:	moduleclasses = ['tools', 'data']


boegel:
@blappm: thoughts after a quick review:

* don't use tabs please, stick to 4-space indent (see https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces)
* rather than introducing a new `moduleclasses` easyconfig parameter, I'd prefer enhancing the existing `moduleclass` one, by making it accept not only a single string value, but also a list of strings

Thoughts on the latter?

boegel:
Jenkins: ok to test

geimer:
I'm not yet sure how well multiple module classes work in combination with the CategorizedHMNS.  One may end up with ambiguities.  Specifying additional categories as `whatis` text in conjunction with Lmod's `keyword` command (for an example, see https://www.tacc.utexas.edu/research-development/tacc-projects/lmod/user-guide) may be a superior solution. But I may be missing the point here.

boegel:
@geimer: just taking the 1st category into account (assuming that's the 'best fit' category) in CategorizedHMNS may be sufficient?
Sometimes it's hard to assign a single module class, which is where supporting a list of classes comes in. But there typically is one class that is the major one (which should be listed first).

blappm:
Hi, what do you mean with CategorizedHMNS ?

We use the module_classes support for different user groups, which are defined by us - to have a better overview with the 'ml avail command'. The normal view with 300 packages, is just a mess.

[config]
modules-tool = Lmod
moduleclasses= .all,python2,python3, group1, group2, group3 [...]

Like this it's then possible to do something like:

ml use /EasyBuild-Apps/modules/.all/
ml use /EasyBuild-Apps/modules/group1
ml use /EasyBuild-Apps/modules/python2
ml use /EasyBuild-Apps/modules/tools

geimer:
@blappm See https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/tools/module_naming_scheme/categorized_hmns.py
This implements a hierarchical module naming scheme which further categorizes the modules using the module class.  The categories are basically just subdirectories named after the `moduleclass` which are `module use`'d and thus provide a separation of modules in the `avail` output by category.

blappm:
@geimer  Ahh this one. Yes - we looked at it. IMHO from a user point of view it's not really userfriendly, since you have a bad overview on the first view.

geimer:
@blappm As usual, it's a matter of taste and site policies.  Personally, I like it much more than having all modules in a single directory.  But I agree that it can be confusing if the categorization is too fine-grained.  That is, I probably would restrict the number of categories to <10 or so.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -579,8 +579,13 @@ def _validate(self, attr, values):  # private method
         """
         if values is None:
             values = []
-        if self[attr] and self[attr] not in values:
-            raise EasyBuildError("%s provided '%s' is not valid: %s", attr, self[attr], values)
+        if isinstance(self[attr], (list, tuple)):
+                for val in self[attr]:
+                    if val and val not in values:
+                        raise EasyBuildError("%s provided '%s' is not valid: %s", attr, val, values)
```
boegel: the indentation is off here; should be a single level of 4 spaces...

boegel:
@blappm: sorry for not getting back to you on this...

We need enhanced unit tests before we can merge this in.

Let me know if you're up for that (if not, I can try and find time to tackle it myself).

PR #1488, opened 2015-11-25
Patch new files

blappm:
Allow patches which have '/dev/null' in the first line:

--- /dev/null   2015-07-10 16:14:12.096153473 +0200
+++ conifer_v0.2.2/setup.py    2015-11-24 17:59:09.000000000 +0100

Reasons:

- It's a valid patch naming
- There are tons of patches like this in the FreeBSD ports repo, this makes migrating easier
- If this is not working, one has to write much more easyblocks, just to touch or create a file. IMHO this helps to avoid 1/3 of all existing custom easyblocks.

### DIFF
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -51,6 +51,14 @@ import sys
 import tempfile
 from distutils.version import LooseVersion
 
+# argparse preferrred, optparse deprecated >=2.7
+HAVE_ARGPARSE=False
+try: 
+    import argparse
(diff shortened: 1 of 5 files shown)


boegel:
@blappm: the `patches` easyconfig parameter also accepts something like this:

```python
patches = [('foo.txt', 'bar')]
```

This will copy 'foo.txt' to the subdirectory 'bar' of the unpacked sources directory.

Which doesn't mean this change is useless (but I'll take a look if you open a new PR targetting `develop`).

PR #1493, opened 2015-11-26
Add easy_install to setup.cfg

valtandor:


boegel:
@valtandor: do you know what this implies (I'm not sure, just asking whether you have checked what kind of effects this may have)?

valtandor:
@boegel: As far as I can tell, because I don't specify any options to the easy_install command, it will do nothing. The purpose of the setup.cfg file is, so far as I can tell, to provide a place where options to the install commands can be specified once instead of having to respecify them each time; see https://docs.python.org/2/distutils/configfile.html.

It got put into my file by some automatic process when I rebuilt EasyBuild with the new setuptools. I figured that it probably wouldn't hurt, and having asked the question of the list and received no responses, I thought I would upload it and see if anyone else recognised it as a good, bad or indifferent thing.

boegel:
Should be harmless indeed, I'll know soon enough if it's not.

PR #1495, opened 2015-12-02
[.yeb] auto-convert of dependencies (REVIEW)

Caylo:


easybuild/framework/easyconfig/types.py
```diff
@@ -39,14 +39,16 @@
 # specific type: dict with only name/version as keys, and with string values
 # additional type requirements are specified as tuple of tuples rather than a dict, since this needs to be hashable
 NAME_VERSION_DICT = (dict, (('only_keys', ('name', 'version')), ('value_types', (str,))))
+DEPENDENCIES = (list, (('element_types', (tuple,)),))
```
boegel: `dependencies` are converted to `dict` values, no? which have a set of allowed keys (name/version/toolchain/versionsuffix)?

easybuild/framework/easyconfig/types.py
```diff
@@ -76,6 +78,16 @@ def is_value_of_type(value, typ_spec):
                 # check whether all values have allowed types
                 'value_types': lambda val: all([type(el) in extra_reqs['value_types'] for el in val.values()]),
```
boegel: hmm, maybe we should make this a recursive call to `is_value_of_type` (which should be made aware of `EASY_TYPES`)?

easybuild/framework/easyconfig/types.py
```diff
@@ -76,6 +78,16 @@ def is_value_of_type(value, typ_spec):
...
             }
+        elif parent_type == list:
+            extra_req_checkers = {
+                # check required type
+                'element_types': lambda val: all([type(el) in extra_reqs['element_types'] for el in val]),
```
boegel: we should flesh out a small function that can check the type of list elements and dict values, and use that for both `value_types` and `element_types` (which should be merged together probably)

easybuild/framework/easyconfig/types.py
```diff
@@ -76,6 +78,16 @@ def is_value_of_type(value, typ_spec):
...
+            }
+
+        else:
+            raise EasyBuildError("Don't know how to check value with parent type %s", parent_type)
+
+        if extra_req_checkers:
```
boegel: this check is probably not needed? if `extra_req_checkers` is not defined, something is very wrong

easybuild/framework/easyconfig/types.py
```diff
@@ -83,9 +95,7 @@ def is_value_of_type(value, typ_spec):
                     type_ok &= check_ok
                     _log.debug("Check for %s requirement (%s) %s for %s", er_key, extra_reqs[er_key], msg, value)
                 else:
-                    raise EasyBuildError("Unknown type requirement specified: %s", er_key)
-        else:
-            raise EasyBuildError("Don't know how to check value with parent type %s", parent_type)
+                   raise EasyBuildError("Unknown type requirement specified: %s", er_key)
```
boegel: what happened with the indent here? not a multiple of 4-space indent anymore (or maybe it wasn't before)?

easybuild/framework/easyconfig/types.py
```diff
@@ -206,6 +216,8 @@ def to_name_version_dict(spec):
 
     return res
 
+def to_dependencies(dep_list):
+    return [to_dependency(dep) for dep in dep_list]
```
boegel: two newlines in between top level functions

easybuild/framework/easyconfig/types.py
```diff
@@ -206,6 +216,8 @@ def to_name_version_dict(spec):
...
```
boegel: docstring pls

easybuild/framework/easyconfig/types.py
boegel: also, we should probably make this a more generic type, since we will also be able to use it for `sanity_check_paths`

easybuild/framework/easyconfig/types.py
```diff
@@ -32,27 +32,16 @@
...
-
 _log = fancylogger.getLogger('easyconfig.types', fname=False)
 
 
-def is_value_of_type(value, typ_spec):
+def dict2tuple(dict_value):
+    """
+    Helper function, converts dict to (nested) tuples
+    """
+    return tuple((k, v) for k, v in sorted(dict_value.items()))
```
boegel: change this to:

```python
def dict2tuple(dict_value):
    """Help function, convert dict value to hashable equivalent via tuples."""
    res = []
    for key, val in sorted(dict_value.items()):
        if isinstance(val, list):
            val = tuple(val)
        res.append((key, val))
    return tuple(res)

File: easybuild/framework/easyconfig/types.py [hunk 9]
```diff
@@ -60,22 +49,47 @@ def is_value_of_type(value, typ_spec):
...
+        # first step: check parent type
+        type_ok = isinstance(value, parent_type)
+        if type_ok:
+            _log.debug("Parent type of value %s matches %s, going in...", value, parent_type)
+            # second step: check additional type requirements
+            if parent_type == dict:
+                extra_req_checkers = {
+                    # check whether all keys have allowed types
+                    'key_types': lambda val:
+                        all([any(is_value_of_type(el, t) for t in extra_reqs['key_types']) for el in val.keys()]),
```
boegel: maybe we should define small helper functions rather than using lambda's?

it'll certainly help in making this more readable :)

easybuild/framework/easyconfig/types.py
```diff
@@ -60,22 +49,47 @@ def is_value_of_type(value, typ_spec):
...
+                    # check whether only allowed keys are used
+                    'only_keys': lambda val: set(val.keys()) == set(extra_reqs['only_keys']),
+                    # check whether all values have allowed types
+                    'value_types': lambda val:
+                        all([any(is_value_of_type(el, t) for t in extra_reqs['value_types']) for el in val.values()]),
+                    'required_keys': lambda val: all([k in val.keys() for k in extra_reqs['required_keys']]),
+                    'opt_keys': lambda val:
+                        all([k in [extra_reqs['required_keys'], extra_reqs['opt_keys']] for k in val.keys()])
```
boegel: hmm, we now have `only_keys`, `required_keys` and `opt_keys`, which seems a bit much...

`only_keys` and `required_keys` are actually used for the same thing (making sure that name/version keys are there)?

easybuild/framework/easyconfig/types.py
```diff
@@ -85,9 +99,10 @@ def is_value_of_type(value, typ_spec):
                 else:
                     raise EasyBuildError("Unknown type requirement specified: %s", er_key)
```
boegel: please add an extra log message that mentions whether the overal type check passed, like there was in `check_type_of_param_value`

easybuild/framework/easyconfig/types.py
```diff
@@ -207,6 +210,14 @@ def to_name_version_dict(spec):
     return res
 
 
+def to_dependencies(dep_list):
```
boegel: move below `to_dependency`? just a matter of style

easybuild/framework/easyconfig/types.py
```diff
@@ -236,17 +247,47 @@ def to_dependency(dep):
         if not found_name_version:
             raise EasyBuildError("Can not parse dependency without name and version: %s", dep)
 
+    # also deal with dependencies in the "old" format
+    elif isinstance(dep, tuple):
+        depspec = dep
```
boegel: this is a not a dict value (cfr. docstring)?

maybe use this?

```python
if isinstance(dep, (tuple, list)):
    if len(dep) >= 2:
        depspec.update({
            'name': dep[0],
            'version': dep[0],
        })
    if len(dep) >= 3:
        depspec.update({'versionsuffix': dep[2]})
    if len(dep) >= 4:
        depspec.update({'toolchain': dep[3]})
```

The rest of this function (see above) should handle the rest?

easybuild/framework/easyconfig/types.py
```diff
@@ -236,17 +247,47 @@ def to_dependency(dep):
...
 
 
-# this uses functions defined in this module, so it needs to be at the bottom of the module
+# these constants use functions defined in this module, so they needs to be at the bottom of the module
+
+# specific type: dict with only name/version as keys, and with string values
+# additional type requirements are specified as tuple of tuples rather than a dict, since this needs to be hashable
+NAME_VERSION_DICT = (dict, dict2tuple({
+    'only_keys': ('name', 'version'),
+    'value_types': (str,),
```
boegel: with the proposed changes to `dict2tuple`, this can be:

```python
NAME_VERSION_DICT = (dict, dict2tuple({
    'only_keys': ['name', 'version'],
    'value_types': [str],
})
```

easybuild/framework/easyconfig/types.py
```diff
@@ -236,17 +247,47 @@ def to_dependency(dep):
...
+}))
+DEPENDENCY_DICT = (dict, dict2tuple({
+    'required_keys': ('name','version'),
+    'opt_keys': ('versionsuffix', 'toolchain'),
```
boegel: alphabet ;)

easybuild/framework/easyconfig/types.py
```diff
@@ -60,22 +49,47 @@ def is_value_of_type(value, typ_spec):
...
+                extra_req_checkers = {
+                    # check whether all keys have allowed types
+                    'key_types': lambda val:
+                        all([any(is_value_of_type(el, t) for t in extra_reqs['key_types']) for el in val.keys()]),
+                    # check whether only allowed keys are used
+                    'only_keys': lambda val: set(val.keys()) == set(extra_reqs['only_keys']),
+                    # check whether all values have allowed types
+                    'value_types': lambda val:
+                        all([any(is_value_of_type(el, t) for t in extra_reqs['value_types']) for el in val.values()]),
+                    'required_keys': lambda val: all([k in val.keys() for k in extra_reqs['required_keys']]),
```
boegel: order keys alphabetically pls

easybuild/framework/easyconfig/types.py
```diff
@@ -236,17 +247,47 @@ def to_dependency(dep):
...
+
+# easy types, that can be verified with isinstance
+EASY_TYPES = [basestring, dict, int, list, tuple]
+
+# type checking is skipped for easyconfig parameters names not listed in TYPES
+TYPES = {
+    'name': basestring,
+    'version': basestring,
+    'toolchain': NAME_VERSION_DICT,
+    'dependencies': DEPENDENCIES,
```
boegel: alphabet ;)

boegel:
Jenkins: ok to test

boegel:
@Caylo: please merge https://github.com/Caylo/easybuild-framework/pull/25

We will probably need to roll back my suggestion for dependency values that are of type tuple/list (see `to_dependency`), and let that be handled by `EasyConfig._parse_dependency` after all...
There's a lot more going on there than I remembered (see also failing tests).

Another thing that is not handled yet for the `.yeb` format is specifying external modules as dependencies, that's currently only handled in case the dependency is specified as a tuple (see `_parse_dependency` and http://easybuild.readthedocs.org/en/latest/Using_external_modules.html).

There may be other things I missed, we should check again on running the full easyconfig test suite against this branch, to make sure we're not breaking current `.eb`s (but just letting dependency specs as tuple values pass through to `_parse_dependency` should be sufficient).

To be followed up Dec 9th! ;-)

easybuild/framework/easyconfig/types.py
```diff
@@ -209,16 +290,25 @@ def to_name_version_dict(spec):
...
     Example:
         {'foo': '1.2.3', 'toolchain': 'GCC, 4.8.2'}
         to
         {'name': 'foo', 'version': '1.2.3', 'toolchain': {'name': 'GCC', 'version': '4.8.2'}}
+
+        or
+
+        ('foo', '1.2.3', '-test', ('GCC', '4.8.2'))
+        to
+        {'name': 'foo', 'version': '1.2.3', 'toolchain': {'name': 'GCC', 'version': '4.8.2'}, 'versionsuffix': '-test'}
```
boegel: drop this, tuples are being passed down as is

easybuild/framework/easyconfig/types.py
```diff
@@ -237,16 +327,58 @@ def to_dependency(dep):
             raise EasyBuildError("Can not parse dependency without name and version: %s", dep)
 
     else:
-        raise EasyBuildError("Can not convert %s (type %s) to dependency dict", dep, type(dep))
+        _log.warning("to_dependency failed to recognise type; passing value down as is")
```
boegel: include `dep`  in the message:

```python
_log.warning("to_dependency failed to recognise type; passing value down as is: %s", dep)
```

boegel: actually, better:

```python
else:
    # pass down value untouched, let EasyConfig._parse_dependency handle it
    depspec = dep
    if isinstance(dep, (tuple, list)):
        _log.debug("Passing down dependency value of type %s without touching it: %s", type(dep), dep)
    else:
        _log.warning("Unknown type of value in to_dependency %s; passing value down as is: %s", type(dep), dep)
```

test/framework/type_checking.py
```diff
@@ -128,6 +141,10 @@ def test_to_dependency(self):
         self.assertEqual(to_dependency({'lib': '1.2.8'}), {'name': 'lib', 'version': '1.2.8'})
         self.assertEqual(to_dependency({'lib': '1.2.8', 'toolchain': 'GCC, 4.8.2'}), lib_dict)
         self.assertEqual(to_dependency({'lib': '1.2.8', 'toolchain': ['GCC', '4.8.2']}), lib_dict)
+        lib_dict.update({'versionsuffix': ''})
+
+        self.assertEqual(to_dependency(('foo', '1.3')), ('foo','1.3'))
```
boegel: add a comment here:

```python
# to_dependency doesn't touch values of non-dict type
```

 and add a test case to check that non-dict/tuple/list values are passed untouched (i.e., no error):

```python
self.assertEqual(to_dependency('foo/1.3'), 'foo/1.3')
```

test/framework/type_checking.py
```diff
@@ -157,16 +174,44 @@ def test_to_dependency(self):
         self.assertErrorRegex(EasyBuildError, "Found unexpected \(key, value\) pair: .*",
             to_dependency, {'lib': '1.2.8', 'foo':'1.3', 'toolchain': 'lib, 1.2.8', 'versionsuffix': 'suff'})
 
+    def test_to_dependencies(self):
+        """Test to_dependencies function."""
+        self.assertEqual(to_dependencies([]), [])
+        deps = [
+            ('foo', '1.2.3'),
```
boegel: add a `foo/1.2.3` in here too, that is left untouched

PR #1498, opened 2015-12-04
[bugfix] make sure that extra custom easyconfig parameters are known for extensions

boegel:
This is required to fix installing `numpy` as an extension, due to the custom easyconfig parameter that was defined in https://github.com/hpcugent/easybuild-easyblocks/pull/757


rjeschmi:
Looks good

boegel:
Thanks for the review @rjeschmi!

PR #1499, opened 2015-12-04
don't load fake module in sanity_check_step during a dry run

boegel:


boegel:
@wpoely86: please review?

wpoely86:
seems fine. What's the reason this pops up now?

boegel:
I was doing a dry run for `numpy`, and noticed this in the output:

```
[sanity_check_step method]

!!!
!!! WARNING: ignoring error 'Changing environment as dictated by module failed: name \'false\' is not defined (stdout: os.environ["LMOD_DEFAULT_MODULEPATH"] = "/private/var/folders/8s/_frgh9sj6m744mxt5w5lyztr0000gn/T/eb-ohnu2j/tmpZCthJv/all:/Users/kehoste/.local/easybuild/modules/all";\nos.environ["MODULEPATH"] = "/private/var/folders/8s/_frgh9sj6m744mxt5w5lyztr0000gn/T/eb-ohnu2j/tmpZCthJv/all:/Users/kehoste/.local/easybuild/modules/all";\nfalse\n, stderr: Lmod has detected the following error: The following module(s) are unknown:\n"numpy/1.10.1-intel-2015b-Python-2.7.10"\n\n   Please check the spelling or version number. Also try "module spider ..."\n\n)'
!!!
```

It occurs because the numpy module (or anything that can be installed as an extension) is not actually created under dry run.

There's more issues like this, it's basically about making everything dry-run aware.

With this fix and https://github.com/hpcugent/easybuild-easyblocks/pull/760 combined, I now get this for `eb numpy.eb -x`:

```
sanity checking... [DRY RUN]

[sanity_check_step method]
  running command "python -c "import numpy""
  (in /Users/kehoste/.local/easybuild/build)
Sanity check paths - file ['files']
  * lib/python/site-packages/numpy/__init__.py
Sanity check paths - (non-empty) directory ['dirs']
  (none)
Sanity check commands
  * nm lib/python/site-packages/numpy/core/multiarray.so | grep blas_dot -h
  * python -c "import numpy"
```

boegel:

PR #1502, opened 2015-12-07
hoist string value into a list in make_module_req

boegel:
minor enhancement to avoid weird problems when passing a string value rather than a list

without this, you may run into an error like:

```
Absolute path / passed to prepend_paths which only expects relative paths
```

caused by `make_module_req` looping over the string value as if it were a list...

cc @wpoely86: please review?


wpoely86:
It looks fine. Do we want to warn about it being a string instead of a list of just silently convert?

boegel:
Adding a warning doesn't hurt I guess, will do.

boegel:

PR #1504, opened 2015-12-09
[.yeb] external modules

Caylo:
specification:

    dependencies: 
        - name: fftw/3.3.4.2
        - external: true  # true / True / TRUE all work

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -589,10 +589,19 @@ class EasyConfig(object):
             'external_module_metadata': {},
         }
         if isinstance(dep, dict):
-            dependency.update(dep)
             # make sure 'dummy' key is handled appropriately
             if 'dummy' in dep and not 'toolchain' in dep:
                 dependency['toolchain'] = dep['dummy']
+            if 'external' in dep and dep['external']:
+                dependency['external_module'] = True
+                dependency['short_mod_name'] = dep['name']
+                dependency['full_mod_name'] = dep['name']
+                if dep['name'] in self.external_modules_metadata:
+                        dependency['external_module_metadata'].update(self.external_modules_metadata[dep['name']])
+                        self.log.info("Updated dependency info with available metadata for external module %s: %s",
+                                      dep['name'], dependency['external_module_metadata'])
+            else:
+                dependency.update(dep)
 
         elif isinstance(dep, Dependency):
             dependency['name'] = dep.name()
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -549,6 +549,21 @@ class EasyConfig(object):
         if self[attr] and self[attr] not in values:
             raise EasyBuildError("%s provided '%s' is not valid: %s", attr, self[attr], values)
 
+    def handle_external_module_metadata(self, dep_name):
+        """
+        helper function for _parse_dependency
+        handles metadata for external module dependencies
+        """
+        dependency = {}
+        if dep_name in self.external_modules_metadata:
+            dependency['external_module_metadata'].update(self.external_modules_metadata[dep_name])
+            self.log.info("Updated dependency info with available metadata for external module %s: %s",
+                          dep_name, dependency['external_module_metadata'])
+        else:
+            self.log.info("No metadata available for external module %s", dep_name)
+
+        return dependency
+
     # private method
     def _parse_dependency(self, dep, hidden=False):
         """
@@ -592,16 +607,10 @@ class EasyConfig(object):
             # make sure 'dummy' key is handled appropriately
             if 'dummy' in dep and not 'toolchain' in dep:
                 dependency['toolchain'] = dep['dummy']
-            if 'external' in dep and dep['external']:
-                dependency['external_module'] = True
-                dependency['short_mod_name'] = dep['name']
-                dependency['full_mod_name'] = dep['name']
-                if dep['name'] in self.external_modules_metadata:
-                        dependency['external_module_metadata'].update(self.external_modules_metadata[dep['name']])
-                        self.log.info("Updated dependency info with available metadata for external module %s: %s",
-                                      dep['name'], dependency['external_module_metadata'])
-            else:
-                dependency.update(dep)
+
+            dependency.update(dep)
+            if dep.get('full_mod_name', False):
+                dependency.update(self.handle_external_module_metadata(dep['name']))
 
         elif isinstance(dep, Dependency):
             dependency['name'] = dep.name()
@@ -619,12 +628,7 @@ class EasyConfig(object):
                     dependency['external_module'] = True
                     dependency['short_mod_name'] = dep[0]
                     dependency['full_mod_name'] = dep[0]
-                    if dep[0] in self.external_modules_metadata:
-                        dependency['external_module_metadata'].update(self.external_modules_metadata[dep[0]])
-                        self.log.info("Updated dependency info with available metadata for external module %s: %s",
-                                      dep[0], dependency['external_module_metadata'])
-                    else:
-                        self.log.info("No metadata available for external module %s", dep[0])
+                    dependency.update(self.handle_external_module_metadata(dep[0]))
                 else:
                     raise EasyBuildError("Incorrect external dependency specification: %s", dep)
             else:
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -556,7 +556,7 @@ class EasyConfig(object):
         """
         dependency = {}
         if dep_name in self.external_modules_metadata:
-            dependency['external_module_metadata'].update(self.external_modules_metadata[dep_name])
+            dependency['external_module_metadata'] = self.external_modules_metadata[dep_name]
             self.log.info("Updated dependency info with available metadata for external module %s: %s",
                           dep_name, dependency['external_module_metadata'])
         else:
@@ -609,8 +609,8 @@ class EasyConfig(object):
                 dependency['toolchain'] = dep['dummy']
 
             dependency.update(dep)
-            if dep.get('full_mod_name', False):
-                dependency.update(self.handle_external_module_metadata(dep['name']))
+            if dep.get('external_module', False):
+                dependency.update(self.handle_external_module_metadata(dep['full_mod_name']))
 
         elif isinstance(dep, Dependency):
             dependency['name'] = dep.name()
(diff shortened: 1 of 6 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -589,10 +589,19 @@ def _parse_dependency(self, dep, hidden=False):
...
-            dependency.update(dep)
             # make sure 'dummy' key is handled appropriately
             if 'dummy' in dep and not 'toolchain' in dep:
                 dependency['toolchain'] = dep['dummy']
+            if 'external' in dep and dep['external']:
+                dependency['external_module'] = True
+                dependency['short_mod_name'] = dep['name']
+                dependency['full_mod_name'] = dep['name']
+                if dep['name'] in self.external_modules_metadata:
+                        dependency['external_module_metadata'].update(self.external_modules_metadata[dep['name']])
```
boegel: indenting once is sufficient ;)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -589,10 +589,19 @@ def _parse_dependency(self, dep, hidden=False):
...
+                        self.log.info("Updated dependency info with available metadata for external module %s: %s",
+                                      dep['name'], dependency['external_module_metadata'])
```
boegel: this is almost a copy-paste from stuff below, so this should be fleshed out in a separate function (one without side-effects), something like:

```python
if dep.get('external', False):
    dependency.update(self.handle_external_module_dep(dep['name']))
```

and then similar below with `dep[0]` as argument

the tricky bit is going to be the `dependency['external_module_metadata'].update` ?

easybuild/framework/easyconfig/types.py
```diff
@@ -33,10 +33,10 @@
 
 from easybuild.tools.build_log import EasyBuildError
 
-
 _log = fancylogger.getLogger('easyconfig.types', fname=False)
 
 
+
```
boegel: need MOAR whitespace? ;-)

two empty lines above/below `_log` is ok, leave it as is

easybuild/framework/easyconfig/types.py
```diff
@@ -301,23 +301,31 @@ def to_dependency(dep):
...
-        depspec = {}
-        found_name_version = False
-        for key, value in dep.items():
-            if key in ['name', 'version', 'versionsuffix']:
-                depspec[key] = value
-            elif key == 'toolchain':
-                depspec['toolchain'] = to_name_version_dict(value)
-            elif not found_name_version:
-                depspec.update({'name': key, 'version': value})
+        if 'external' in dep and dep['external']:
```
boegel: `if dep.get('external', False)`

easybuild/framework/easyconfig/types.py
```diff
@@ -301,23 +301,31 @@ def to_dependency(dep):
...
+            if len(dep) == 2 and 'name' in dep:
+                depspec = dep
+                depspec.update({'version': None}) # required key
             else:
-                raise EasyBuildError("Found unexpected (key, value) pair: %s, %s", key, value)
+                raise EasyBuildError("Can not parse external module: %s", dep)
```
boegel: `"Unexpected format for dependency marked as external module"`

easybuild/framework/easyconfig/types.py
```diff
@@ -301,23 +301,31 @@ def to_dependency(dep):
...
-        found_name_version = False
-        for key, value in dep.items():
-            if key in ['name', 'version', 'versionsuffix']:
-                depspec[key] = value
-            elif key == 'toolchain':
-                depspec['toolchain'] = to_name_version_dict(value)
-            elif not found_name_version:
-                depspec.update({'name': key, 'version': value})
+        if 'external' in dep and dep['external']:
+            if len(dep) == 2 and 'name' in dep:
```
boegel: this is clearer imho:

```python
expected_keys = ['external_module', 'name']
if sorted(dep.keys()) == expected_keys:
```

and then you can use `', '.join(expected_keys)` in the error message

easybuild/framework/easyconfig/types.py
```diff
@@ -301,23 +301,31 @@ def to_dependency(dep):
...
+                depspec = dep
```
boegel: what's the use of this?

just use

```python
depspec.update({
    'name': dep['name'],
    'version': None,
}
```

`external_module` should no longer be there?

easybuild/framework/easyconfig/types.py
```diff
@@ -301,23 +301,31 @@ def to_dependency(dep):
...
+                depspec.update({'version': None}) # required key
```
boegel: where is `version` required exactly?

easybuild/framework/easyconfig/types.py
boegel: also, update the docstring for this case?

```python
{'name': 'fftw/3.3.4.1', 'external_module': True}
=>
{'name': 'fftw/3.3.4.1', 'version': None}
```

easybuild/framework/easyconfig/types.py
boegel: or it should *always* be there (and `False` by default, of course)

test/framework/easyconfigs/yeb/CrayCCE-5.1.29.yeb
```diff
@@ -0,0 +1,18 @@
...
+
+toolchain: dummy, dummy
+
+dependencies:
+    - name: PrgEnv-cray/5.1.29  # also loads cray-libsci
+      external: true
+    - name: fftw/3.3.4.0
+      external: true
+
+
```
boegel: drop an empty line?

test/framework/easyconfigs/yeb/CrayCCE-5.1.29.yeb
```diff
@@ -0,0 +1,18 @@
...
+name: CrayCCE
+version: 5.1.29
+
+homepage: (none)
+description: Toolchain using Cray compiler wrapper, using PrgEnv-cray module.
+
+toolchain: dummy, dummy
+
+dependencies:
+    - name: PrgEnv-cray/5.1.29  # also loads cray-libsci
```
boegel: move comment above this line, to make it easier to read?

easybuild/framework/easyconfig/types.py
Caylo: specified below, 'name' and 'version' are required parameters for a dependency dict, see https://github.com/Caylo/easybuild-framework/blob/yeb-external-module-fix/easybuild/framework/easyconfig/types.py#L360

easybuild/framework/easyconfig/types.py
boegel: and use `external_module` rather than `external`, more descriptive

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -589,10 +589,19 @@ def _parse_dependency(self, dep, hidden=False):
             'external_module_metadata': {},
         }
         if isinstance(dep, dict):
-            dependency.update(dep)
             # make sure 'dummy' key is handled appropriately
             if 'dummy' in dep and not 'toolchain' in dep:
                 dependency['toolchain'] = dep['dummy']
+            if 'external' in dep and dep['external']:
```
```python
if dep.get('external', False):
```

easybuild/framework/easyconfig/types.py
boegel: ah, yeah, it should be there, `_parse_dependency` picks up on it

then make sure it's always there

easybuild/framework/easyconfig/types.py
boegel: with the existing code in `_parse_dependency`, *both* name/version stay defined as `None`, stick to that here, and just define `depspec['full_mod_name']` (and `short_mod_name` too) as `dep['name']`

boegel: this will simplify things a bit in `_parse_dependency`

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -549,6 +549,21 @@ def _validate(self, attr, values):  # private method
...
             raise EasyBuildError("%s provided '%s' is not valid: %s", attr, self[attr], values)
 
+    def handle_external_module_metadata(self, dep_name):
+        """
+        helper function for _parse_dependency
+        handles metadata for external module dependencies
+        """
+        dependency = {}
+        if dep_name in self.external_modules_metadata:
+            dependency['external_module_metadata'].update(self.external_modules_metadata[dep_name])
```
boegel: this can never work, since the `'external_module_metadata'` is not defined

also, you need to update what is there already

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -589,11 +604,14 @@ def _parse_dependency(self, dep, hidden=False):
...
         }
         if isinstance(dep, dict):
-            dependency.update(dep)
             # make sure 'dummy' key is handled appropriately
             if 'dummy' in dep and not 'toolchain' in dep:
                 dependency['toolchain'] = dep['dummy']
 
+            dependency.update(dep)
+            if dep.get('full_mod_name', False):
+                dependency.update(self.handle_external_module_metadata(dep['name']))
```
boegel: name or full_mod_name?

easybuild/framework/easyconfig/types.py
```diff
@@ -296,28 +295,48 @@ def to_dependency(dep):
...
+
+        if dep.get('external_module', False):
+            expected_keys = ['external_module', 'name']
+            if sorted(dep.keys()) == expected_keys:
+                depspec.update({
+                    'external_module': True,
+                    'name': None,
+                    'version': None,
+                    'full_mod_name': dep['name'],
+                    'short_mod_name': dep['name']
```
boegel: sort alphabetically?

easybuild/framework/easyconfig/easyconfig.py
boegel: use `None` here as fallback value for `full_mod_name`

boegel: or actually, use

```python
if dep.get('external_module', False):
    dependency.update(self.handle_external_module_metadata(dep['full_mod_name']))
else:
    dependency.update(dep)
```

boegel:
Jenkins: ok to test

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -589,11 +604,14 @@ def _parse_dependency(self, dep, hidden=False):
             'external_module_metadata': {},
         }
         if isinstance(dep, dict):
-            dependency.update(dep)
             # make sure 'dummy' key is handled appropriately
             if 'dummy' in dep and not 'toolchain' in dep:
                 dependency['toolchain'] = dep['dummy']
 
+            dependency.update(dep)
```
boegel: hmm, moving this down may change the value for the `toolchain` key, if it's defined in `dep`?

boegel:
Jenkins: test this please

boegel:
merged via #1514, thanks @Caylo!

PR #1505, opened 2015-12-10
PKG_CONFIG_PATH search extended to {lib,lib64,lib32,share}/pkgconfig

hajgato:


easybuild/framework/easyblock.py
```diff
@@ -1065,7 +1065,7 @@ def make_module_req_guess(self):
             'LIBRARY_PATH': ['lib', 'lib64', 'lib32'],
             'CPATH': ['include'],
             'MANPATH': ['man', 'share/man'],
-            'PKG_CONFIG_PATH': ['lib/pkgconfig', 'share/pkgconfig'],
+            'PKG_CONFIG_PATH': ['%s/pkgconfig' % x for x in ['lib', 'lib64', 'lib32', 'share']],
```
boegel: I'd prefer this (use `os.path.join` + alphabetical order):

```python
    'PKG_CONFIG_PATH': [os.path.join(x, 'pkgconfig') for x in ['lib', 'lib32', 'lib64', 'share']],
```

hajgato: @boegel Why alphabetic? then should not be `'LIBRARY_PATH': ['lib', 'lib64', 'lib32']` also alphabetic?
And os.path.join for manpaht as well?

boegel: @hajgato: style, mostly, the order should be changed for other stuff too, indeed

for `'share/man'`, it's a single string, not a composed string, but it doesn't hurt to also use `os.path.join` there

hajgato:
@boegel done

boegel:
@hajgato: thanks, look good, will merge when Jenkins gives the green light

boegel:
Going in, thanks @hajgato!

PR #1506, opened 2015-12-11
add support for generating swap statements for dependencies (REVIEW)

boegel:
required for stable Cray support (cfr. https://github.com/hpcugent/easybuild-framework/issues/1390)

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -868,6 +868,8 @@ class EasyBlock(object):
     def make_module_dep(self, swap_info=None):
         """
         Make the dependencies for the module file.
+
+        @param swap_info: dictionary with full module names as keys and module name to unload first as corr. value
         """
         deps = []
         mns = ActiveMNS()
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -865,12 +865,13 @@ class EasyBlock(object):
         # cleanup: unload fake module, remove fake module dir
         self.clean_up_fake_module(fake_mod_data)
 
-    def make_module_dep(self):
+    def make_module_dep(self, swap_info=None):
         """
         Make the dependencies for the module file.
         """
         deps = []
         mns = ActiveMNS()
+        swap_info = swap_info or {}
 
         # include load statements for toolchain, either directly or for toolchain dependencies
         if self.toolchain.name != DUMMY_TOOLCHAIN_NAME:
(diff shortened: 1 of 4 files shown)


boegel:
@wpoely86: still needs tests, but can you give this an early review?

@gppezzi: this is required for the `CrayToolchain` easyblock

boegel:
cc @pforai

easybuild/framework/easyblock.py
```diff
@@ -865,12 +865,13 @@ def make_devel_module(self, create_in_builddir=False):
         # cleanup: unload fake module, remove fake module dir
         self.clean_up_fake_module(fake_mod_data)
 
-    def make_module_dep(self):
+    def make_module_dep(self, swap_info=None):
```
wpoely86: document arguments

easybuild/framework/easyblock.py
```diff
@@ -865,12 +865,13 @@ def make_devel_module(self, create_in_builddir=False):
...
         """
         Make the dependencies for the module file.
         """
         deps = []
         mns = ActiveMNS()
+        swap_info = swap_info or {}
```
wpoely86: Can you just not as well make the default arg `{}`?

boegel: no, mutable values as default value are of the devil, see http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

easybuild/tools/module_generator.py
```diff
@@ -128,6 +128,25 @@ def conditional_statement(self, condition, body, negative=False):
...
+        raise NotImplementedError
+
+    def swap_module(self, mod_names, recursive_unload=False):
+        """Generate swap statement for specified modules."""
+        # swap via unload/load
+        # 'swap' command is not used, because:
+        # i) it would fail if the module being swapped out is not loaded
+        # ii) it would do weird stuff when the module in which it is used in gets unloaded (?)
+        comment = self.comment("'safe' swap: unload %s when loaded, then load %s" % mod_names).strip()
+        unload = self.unload_module(mod_names[0]).strip()
```
wpoely86: add check if `mod_names` is a tuple?

easybuild/tools/module_generator.py
```diff
@@ -128,6 +128,25 @@ def conditional_statement(self, condition, body, negative=False):
...
 
+    def load_module(self, mod_name, recursive_unload=False):
+        """Generate load statement for specified module."""
+        raise NotImplementedError
+
+    def unload_module(self, mod_name, conditional=True):
+        """Generate unload statement for specified module."""
+        raise NotImplementedError
+
+    def swap_module(self, mod_names, recursive_unload=False):
```
wpoely86: doc arguments

wpoely86: ( everywhere )

easybuild/tools/module_generator.py
boegel: nah, that's like an assert in the code, we don't do that anywhere?

easybuild/framework/easyblock.py
boegel: done

boegel:

PR #1514, opened 2015-12-15
[.yeb] support for external modules

boegel:
wrapper PR for @Caylo's PR #1504

* sync with develop
* fix trivially broken test due to added test easyconfig
* restore order of updating 'dependency' dict in _parse_dependency


boegel:
easyblocks/easyconfigs unit tests on top of this still pass, so good to go into `feature/yaml`, and to merge that into `develop`

PR #1519, opened 2015-12-16
[bugfix] only use glob in make_module_req on non-empty strings

boegel:
fix for *ancient* bug that popped up in https://github.com/hpcugent/easybuild-easyblocks/pull/758, which was introduced in #898 (March 2014!)

it managed to go undetected for a long time because only the CUDA easyblock passed empty string values via `make_module_req_guess` to `make_module_req`...

cc @damianam


wpoely86:
lgtm

boegel:
Jenkins: test this please

boegel:

PR #1528, opened 2015-12-30
implement support for --new-pr and --update-pr (REVIEW)

boegel:
adds support for `eb --new-pr` and `eb --update-pr`

results in PR like https://github.com/boegel/easybuild-easyconfigs/pull/15

target account/repository can be controlled via `--github-target-account` and `--github-target-repo` (defaults to `hpcugent` and `easybuild-easyconfigs`)

title/description for new PR can be controlled using `--pr-title` and `--pr-descr`, resp.

commit message for new/updated PR can be controlled using `--pr-commit-msg`


boegel:
fully functional, but may need some more options (e.g. `--github-target-pr-branch` which defaults to `develop`)

@wpoely86: up for giving this a test/review?

boegel:
cc @pescobar

easybuild/tools/options.py
```diff
@@ -364,25 +365,43 @@ def informative_options(self):
         self.log.debug("informative_options: descr %s opts %s" % (descr, opts))
         self.add_group_parser(opts, descr)
 
+    def github_options(self):
+        """GitHub integration configuration options."""
+        descr = ("GitHub integration options", "Integration with GitHub")
+
+        opts = OrderedDict({
+            'dump-test-report': ("Dump test report to specified path", None, 'store_or_None', 'test_report.md'),
```
wpoely86: nit picking but maybe sort?

easybuild/tools/github.py
```diff
@@ -58,6 +64,12 @@
     _log.warning("Failed to import from 'vsc.utils.rest' Python module: %s" % err)
     HAVE_GITHUB_API = False
 
+try:
+    import git
+    from git import GitCommandError
+except ImportError as err:
+    _log.warning("Failed to import 'git' Python module: %s", err)
```
wpoely86: didn't we have a decorator for this?

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
         raise EasyBuildError("Failed to create comment in PR %s#%d; status %s, data: %s", repo, issue, status, data)
 
 
+def _copy_easyconfigs_to_repo(paths, target_dir):
```
wpoely86: any reason why start with `_`? I'm not a big fan of such things.

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+        'paths_in_repo': [],
+        'new': [],
+    }
+
+    a_to_z = [chr(i) for i in range(ord('a'), ord('z') + 1)]
+    subdir = os.path.join('easybuild', 'easyconfigs')
+
+    if os.path.exists(os.path.join(target_dir, subdir)):
+        for path in paths:
+            ecs = process_easyconfig(path)
```
wpoely86: what happens if `path` is a directory?

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+            if len(ecs) == 1:
+                file_info['ecs'].append(ecs[0]['ec'])
+                name = file_info['ecs'][-1].name
+                ec_filename = '%s-%s.eb' % (name, det_full_ec_version(file_info['ecs'][-1]))
+
+                letter = name.lower()[0]
+                if letter not in a_to_z:
+                    raise EasyBuildError("Don't know which letter subdir to use for %s", name)
+
+                target_path = os.path.join(subdir, name.lower()[0], name, ec_filename)
```
wpoely86: use `letter` here

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+
+                    mkdir(os.path.dirname(full_target_path), parents=True)
+                    shutil.copy2(path, full_target_path)
+                    _log.info("%s copied to %s", path, full_target_path)
+                except OSError as err:
+                    raise EasyBuildError("Failed to copy %s to %s: %s", path, target_path, err)
+
+                file_info['paths_in_repo'].append(target_path)
+            else:
+                raise EasyBuildError("Multiple EasyConfig instances obtained from easyconfig file %s", path)
```
wpoely86: when does that happen? When you give lists to `buildopts`? That means that FFTW doesn't work with this?

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+
+
+@only_if_module_is_available('git', pkgname='GitPython')
+def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_account=None, commit_msg=None):
+    """
+    Common code for new_pr and update_pr functions
+
+    @paths: list of paths that will be used to create/update PR
+    @start_branch: name of branch to start from
+    @pr_branch: name of branch to push to GitHub
```
wpoely86: doc all the args

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+    else:
+        raise EasyBuildError("Subdirectory %s not found in %s", subdir, target_dir)
+
+    return file_info
+
+
+@only_if_module_is_available('git', pkgname='GitPython')
+def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_account=None, commit_msg=None):
+    """
+    Common code for new_pr and update_pr functions
```
wpoely86: yeah, but what does it do?

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+
+    # git checkout -b <branch>; git pull
+    if hasattr(origin.refs, start_branch):
+        origin_branch = getattr(origin.refs, start_branch)
+    else:
+        raise EasyBuildError("Branch '%s' not found at %s", start_branch, github_url)
+
+    _log.debug("Checking out branch %s from remote %s", start_branch, github_url)
+    git_repo.create_head(start_branch, origin_branch).checkout()
+    origin.pull(start_branch)
```
wpoely86: why the `pull`?

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+
+    # copy files to right place
+    file_info = _copy_easyconfigs_to_repo(paths, git_working_dir)
+
+    # create target branch, if it was specified
```
wpoely86: I would do this *before* copying the files?

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+    # commit all staged in one go if a commit message is specified
+    if commit_msg is not None:
+        git_repo.index.commit(commit_msg)
+
+    # push to GitHub
+    github_user = build_option('github_user')
+    github_url = 'git@github.com:%s/%s.git' % (github_user, github_target_repo)
+    my_remote = git_repo.create_remote(github_user, github_url)
+    _log.info("Pushing branch %s to remote %s", pr_branch, my_remote)
+    my_remote.push(pr_branch)
```
wpoely86: what if this fails?

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+        raise EasyBuildError("No paths specified")
+
+    # create working directory
+    git_working_dir = tempfile.mkdtemp(prefix='git-working-dir')
+    git_repo = git.Repo.init(git_working_dir)
+    _log.debug("git working directory initialized at %s", git_working_dir)
+
+    # add remote to pull from
+    github_target_repo = build_option('github_target_repo')
+    github_url = 'git@github.com:%s/%s.git' % (target_account, github_target_repo)
```
wpoely86: if we only need read only, why not use `https`?

easybuild/tools/github.py
```diff
@@ -395,6 +407,235 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+
+@only_if_module_is_available('git', pkgname='GitPython')
+def new_pr(paths, title=None, descr=None, commit_msg=None):
+    """Open new pull request using specified files."""
+    _log.info("Opening new pull request for  with %s", paths)
+
+    github_target_account = build_option('github_target_account')
+    github_target_repo = build_option('github_target_repo')
+    if github_target_repo != GITHUB_EASYCONFIGS_REPO:
+        raise EasyBuildError("Don't know how to open a pull request to the %s repository yet", github_target_repo)
```
wpoely86: what's the issue? If a token with apprioate right is being used?

easybuild/tools/options.py
boegel: will do, thanks

easybuild/tools/github.py
boegel: we do, and I'm using it lower down, see `@only_if_module_is_available`

but `git` still needs to be imported, the decorator doesn't take care of that (it maybe could though, but I'll keep that for when I have time to spare once ;-))

easybuild/tools/github.py
boegel: it's an indication that this function is to be considered private, i.e. shouldn't be imported in other places... it basically provides the code common to both `--new-pr` and `--update-pr`, but is pretty useless in a different context

easybuild/tools/github.py
boegel: You should run into something like `IOError: Failed to read file ...`.

It can check here (or in process_easyconfig) whether it's a file though, doesn't hurt...

boegel: Actually, adding a check in `process_easyconfig` with `os.path.isfile` may be expensive, so it's probably better to just try and access the file assuming it's an existing file, and letting the IOError bubble up in case it's not (non-existing, not a file, etc.)

boegel: You'll get a clear error message like this, even without the check:

```
>>> try:
...   read_file('/tmp')
... except EasyBuildError as err:
...   print err.msg
... 
Failed to read /tmp: [Errno 21] Is a directory: '/tmp'
```

easybuild/tools/github.py
boegel: nice catch, thx

easybuild/tools/github.py
boegel: `process_easyconfig` takes into account that several 'blocks' may be defined in an easyconfig, but this is a really old & unused feature which is not used in any of the easyconfigs we ship in the easybuild-easyconfigs repository...

it something entirely different than using a list of values for `configopts`, like we do for FFTW

at some point, we need to clean things up to make `process_easyconfig` only return a single `EasyConfig` instance, and throw an error if we detect the use of 'blocks'

easybuild/tools/github.py
boegel: ok

easybuild/tools/github.py
boegel: fleshed out the docstring

easybuild/tools/github.py
boegel: hmm... good point :)

should be fine to drop this

easybuild/tools/github.py
boegel: doesn't matter as long as nothing is changed?

this is basically `git checkout -b <branch>`

easybuild/tools/github.py
boegel: checking the return value of `push` now

if this fails (and it did because of the `git fetch --depth=1`, you can't push a shallow copy apparently), opening the PR just fails

but, a problem with `--update-pr` would have gone undetected, this is handled now

easybuild/tools/github.py
boegel: yeah, using `https` now, although it matters little

easybuild/tools/github.py
boegel: the main problem is that we don't know how to move the files supplied to the right place;

also, the way we determine the default branch name assumes we're dealing with easyconfigs, etc.

so, maybe this check should be moved to `_easyconfigs_pr_common`, I'll do that

easybuild/tools/github.py
wpoely86: I don't know about Python, but in C/C++, this is evil. It seems to indicate that it is a system function.

easybuild/tools/github.py
wpoely86: I consider it the cleaner workflow to first create the branch, then make the changes?

easybuild/tools/github.py
boegel: It's merely a convention that indicates that this is a 'private' function, i.e. that it shouldn't be used outside of this module. It's nothing more than a naming convention though, it'll work either way.

boegel: However, let me move this function (and make it non-private), it doesn't really belong in `github.py`.
I'll move it to `easyconfig.py`, where related functions like `robot_find_easyconfig` are already located.

easybuild/tools/github.py
boegel: OK, makes sense, will move.

boegel: Ah, hold on. I can't.

I'm using information returned by the function that copies the easyconfigs to determine the (default) name of the PR branch, i.e. the software name/version of the 1st easyconfig.

So, leaving as is.

boegel:
Jenkins: test this please

boegel:
some more (minor) enhancements + added unit tests + require `--experimental`

I believe this is ready to go in (still working on the docs update to go with this)...

@wpoely86: final review?

boegel:
tests failed on Jenkins because of outdated `GitPython` installation on Jenkins, now updated to 1.0.1

Jenkins: test this please

boegel:
Jenkins: test this please

easybuild/tools/github.py
```diff
@@ -395,6 +408,279 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+            if not os.path.exists(path):
+                non_existing_paths.append(path)
+
+        if non_existing_paths:
+            raise EasyBuildError("One or more non-existing paths specified: %s", ', '.join(non_existing_paths))
+    else:
+        raise EasyBuildError("No paths specified")
+
+    tmp_git_working_dir = tempfile.mkdtemp(prefix='git-working-dir')
+    os.rmdir(tmp_git_working_dir)
```
wpoely86: should this be in a `try`?

easybuild/tools/github.py
```diff
@@ -395,6 +408,279 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+    if build_option('git_working_dirs_path'):
+        workdir = os.path.join(git_working_dirs_path, pr_target_repo)
+        if os.path.exists(workdir):
+            try:
+                print_msg("copying %s..." % workdir)
+                shutil.copytree(workdir, tmp_git_working_dir)
+            except OSError as err:
+                raise EasyBuildError("Failed to copy git working dir %s to %s: %s", workdir, tmp_git_working_dir, err)
+
+    if not os.path.exists(tmp_git_working_dir):
```
wpoely86: you deleted it on L444?

wpoely86: okay, just put the `rmdir` in the `try` at line 452?

easybuild/tools/github.py
```diff
@@ -395,6 +408,279 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+        raise EasyBuildError("Failed to fetch branch '%s' from %s: %s", start_branch, github_url, err)
+    if res:
+        if res[0].flags & res[0].ERROR:
+            raise EasyBuildError("Fetching branch '%s' from remote %s failed: %s", start_branch, origin, res[0].note)
+        else:
+            _log.debug("Fetched branch '%s' from remote %s (note: %s)", start_branch, origin, res[0].note)
+    else:
+        raise EasyBuildError("Fetching branch '%s' from remote %s failed: empty result", start_branch, origin)
+
+    # git checkout -b <branch>; git pull
```
wpoely86: no git pull here

easybuild/tools/github.py
```diff
@@ -395,6 +408,279 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+    if hasattr(origin.refs, start_branch):
```
wpoely86: wouldn't it fail in the above code if the branch isn't present?

easybuild/tools/github.py
boegel: good point, thanks!

easybuild/tools/github.py
boegel: this is sort of a sanity check for the `fetch` done above, I'll leave it in, doesn't hurt

easybuild/tools/github.py
boegel: I removed it at som point, but I think it should still be there, readding it...

wpoely86: why, you just did a fetch?

boegel: yeah, well, I'm not sure yet...

The test is failing on Jenkins, and I'm not sure why. I'm just checking now if the `pull` makes a difference

boegel:
I've opened/updated over a dozen of PRs with this already, and with the unit tests in place and the thorough review by @wpoely86, this is good to go.

To be clear: using `--new-pr` or `--update-pr` currently requires `--experimental`, since I'm expecting the interface to slightly chance as people provide feedback on this and when extending this to also easyblocks/framework is added.

Going in!

PR #1529, opened 2015-12-30
Fix devel install script 

pescobar:
the module file was not created because the folder doesn't exists


boegel:
Thanks @pescobar!

PR #1530, opened 2015-12-30
updating quotes for url

### LINKED ISSUE #1527: packaging fails when homepage = '(none)'
I just found that packaging fails with some easyconfigs which have ```homepage = '(none)'```. As example this one fails
https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/g/goolf/goolf-1.7.20.eb

Here you have a debug log:
https://gist.github.com/02ce3baf6fe8afbc46ce

As temporary workaround I just defined a homepage for this toolchain using url http://easybuild.readthedocs.org/en/latest/eb_list_toolchains.html

Maybe the proper solution is to check the content of the ```homepage``` variable and skip using it in case it's not valid? 

CC: @rjeschmi

rjeschmi:
a PR to fix #1527

### DIFF
--- a/easybuild/tools/package/utilities.py
+++ b/easybuild/tools/package/utilities.py
@@ -108,7 +108,7 @@ def package_with_fpm(easyblock):
     for dep in deps:
         _log.debug("The dep added looks like %s ", dep)
         dep_pkgname = package_naming_scheme.name(dep)
-        depstring += " --depends '%s'" % quote_str(dep_pkgname)
+        depstring += " --depends %s" % quote_str(dep_pkgname)
 
     cmdlist = [
         PKG_TOOL_FPM,
--- a/easybuild/tools/package/utilities.py
+++ b/easybuild/tools/package/utilities.py
@@ -113,7 +113,7 @@ def package_with_fpm(easyblock):
     cmdlist = [
         PKG_TOOL_FPM,
         '--workdir', workdir,
-        '--name', pkgname,
+        '--name', quote_str(pkgname),
         '--provides', pkgname,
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

boegel:
@rjeschmi: should we also use `quote_str` on other stuff? The software name could be something like `Open|SpeedShop` (yes, that's a thing, cfr. https://openspeedshop.org/) for example, which would also cause problems...

boegel:
@rjeschmi: ping?

easybuild/tools/package/utilities.py
```diff
@@ -113,14 +113,14 @@ def package_with_fpm(easyblock):
     cmdlist = [
         PKG_TOOL_FPM,
         '--workdir', workdir,
-        '--name', pkgname,
+        '--name', quote_str(pkgname),
         '--provides', pkgname,
```
boegel: also quote here?

rjeschmi: yes, and also quoted the dep pkgname above.

boegel: :+1:

boegel:
@rjeschmi: did you retest using `--package` with this in place? Just making very sure nothing broke here, since the unit tests only cover the EB side of things with a mocked `fpm`...

rjeschmi:
I tested it on goolf-1.7.20 which errored originally due to having homepage = '(none)'

boegel:
Going in, thanks @rjeschmi and @pescobar for reporting the issue!

PR #1537, opened 2016-01-08
trying to fix 1342_pgisupport

jgphpc:
See issue: https://github.com/hpcugent/easybuild-framework/issues/1536


easybuild/tools/toolchain/craytoolchain.py
```diff
@@ -0,0 +1 @@
+craytoolchain.py-with-swap-fix-gpp
```
boegel: this is a symlink? any particular reason?

boegel:
@jgphpc: it seems like you included the PGI support from #1342 in here?

It's good to know that others are playing around with it, especially on Cray systems, but we should get #1342 merged first...

From what I can tell, the `CraypPGI` side of things in here looks OK, except for the symlink I pointed out.

easybuild/tools/toolchain/craytoolchain.py
gppezzi: I created this symlink for our production installation in order to avoid overwriting the stock craytoolchain.py with our modified version (that it is still under testing).

I suppose that craytoolchain.py should come from PR [#2222](https://github.com/hpcugent/easybuild-easyconfigs/pull/2222) and the file+symlink could be removed from this PR.

jgphpc:
Would it be ok if i close this PR and commit my files to https://github.com/hpcugent/easybuild-framework/pull/1342/files instead ?

boegel:
@jgphpc I'm not sure what you're trying to do... #1342 is not your PR, so how will you add your changes there?

I think we should get #1342 merged in first (we're close to that), and then revisit this PR (by updating the branch you used here with `develop`).

Keeping the `CrayPGI` changes separate from general PGI support makes sense to me?

jgphpc:
Will continue in https://github.com/hpcugent/easybuild-framework/pull/1729

Issue #1540, opened 2016-01-08
View EasyBuild log files without cut-n-paste


citibeth:
I frequently get EasyBuild errors.  Then I have to cut-n-paste out of my command window, in order to see the errors.  In fact, I have to cut-n-paste the RIGHT filename out of a bunch of text.  For example:

```
gs611-tyndall:app rpfische$ eb UDUNITS-2.2.19-GCC-4.9.3.eb
PYTHONPATH= /Users/rpfische/ebx/pylib
== temporary log file in case of crash /var/folders/n2/0ll2lwm50b91f5ppmyggd_g4cdc3y3/T/eb-yNa2wa/easybuild-Ab31tE.log
== resolving dependencies ...
== processing EasyBuild easyconfig /Users/rpfische/git/usermacports/ebx/easyconfigs/app/UDUNITS-2.2.19-GCC-4.9.3.eb
== building and installing UDUNITS/2.2.19-GCC-4.9.3...
== fetching files...
== FAILED: Installation ended unsuccessfully (build directory: /Users/rpfische/eb/easybuild/build/UDUNITS/2.2.19/GCC-4.9.3): build failed (first 300 chars): Couldn't find file udunits-2.2.19.tar.gz anywhere, and downloading it didn't work either... Paths attempted (in order): /Users/rpfische/git/usermacports/ebx/easyconfigs/app/u/UDUNITS/udunits-2.2.19.tar.gz, /Users/rpfische/git/usermacports/ebx/easyconfigs/app/UDUNITS/udunits-2.2.19.tar.gz, /Users/rpf
== Results of the build can be found in the log file /var/folders/n2/0ll2lwm50b91f5ppmyggd_g4cdc3y3/T/eb-yNa2wa/easybuild-UDUNITS-2.2.19-20160108.114133.xMUeN.log
ERROR: Build of /Users/rpfische/git/usermacports/ebx/easyconfigs/app/UDUNITS-2.2.19-GCC-4.9.3.eb failed (err: "build failed (first 300 chars): Couldn't find file udunits-2.2.19.tar.gz anywhere, and downloading it didn't work either... Paths attempted (in order): /Users/rpfische/git/usermacports/ebx/easyconfigs/app/u/UDUNITS/udunits-2.2.19.tar.gz, /Users/rpfische/git/usermacports/ebx/easyconfigs/app/UDUNITS/udunits-2.2.19.tar.gz, /Users/rpf")
```

[...]

PR #1541, opened 2016-01-11
add support for eb --last-log

boegel:
implementation of feature request #1540

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -549,10 +549,15 @@ def find_last_log(curlog):
     except OSError as err:
         raise EasyBuildError("Failed to locate/select/order log files matching '%s': %s", glob_pattern, err)
 
-    # log of current session is typically listed last, should be taken into account
-    res = sorted_paths[-1]
-    if os.path.samefile(res, curlog):
-        res = sorted_paths[-2]
+    try:
+        # log of current session is typically listed last, should be taken into account
+        res = sorted_paths[-1]
+        if os.path.exists(curlog) and os.path.samefile(res, curlog):
+            res = sorted_paths[-2]
+
+    except IndexError:
+        _log.debug("No last log file found (sorted retained paths: %s)", sorted_paths)
+        res = None
 
     _log.debug("Picked %s as last log file (current: %s) from %s", res, curlog, sorted_paths)
     return res
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -489,9 +489,15 @@ def get_build_log_path():
     return res
 
 
-def get_log_filename(name, version, date=None, timestamp=None, add_salt=False):
+def get_log_filename(name, version, add_salt=False, date=None, timestamp=None):
     """
     Generate a filename to be used for logging
+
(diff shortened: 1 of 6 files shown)


boegel:
@wpoely86: please review

Basically, this allows having something like this in your bash history for easy access to the last log file:

```
vim `eb --last-log`
```

boegel:
Hmm, actually more like this, since there's some extra output getting in the way...

```
vim `eb --last-log | grep -v '^=='`
```

easybuild/tools/config.py
```diff
@@ -517,6 +520,32 @@ def get_log_filename(name, version, add_salt=False):
...
+    try:
+        # sorted by modification time, most recent last
+        sorted_paths = sorted(glob.glob(glob_pattern), key=lambda f: os.stat(f).st_mtime)
+    except OSError as err:
+        raise EasyBuildError("Failed to locate log files matching '%s': %s", glob_pattern, err)
+
+    # log of current session is typically listed last, should be taken into account
+    res = sorted_paths[-1]
+    if os.path.samefile(res, curlog):
+        res = sorted_paths[-2]
```
ocaisa: You could also check if the current user is the owner of the file...there could be a corner case where two EB-ers are using the same system at the same time.

boegel: Hmm, good idea, thanks.

wpoely86: :+1:

easybuild/tools/config.py
```diff
@@ -517,6 +520,32 @@ def get_log_filename(name, version, add_salt=False):
     return filepath
 
 
+def find_last_log(curlog):
+    """Find location to last log file that is still available."""
```
wpoely86: doc args

boegel:
@wpoely86, @ocaisa: remarks fixed, and I also got rid of the `== `-prefixed lines in the output of `--last-log`, please rereview?

You may still need to use ```vim `eb --last-log | grep 'log$'` ``` though, in case running `eb` spits out warnings to stdout through Python modules being loaded (depends on your setup)

ocaisa:
Logic looks good to me

easybuild/tools/config.py
```diff
@@ -488,18 +489,20 @@ def get_build_log_path():
     return res
 
 
-def get_log_filename(name, version, add_salt=False):
+def get_log_filename(name, version, date=None, timestamp=None, add_salt=False):
```
wpoely86: doc all args. And isn't nicer to append new args?

easybuild/tools/filetools.py
```diff
@@ -971,7 +971,7 @@ def move_logs(src_logfile, target_logfile):
                              src_logfile, target_logfile, err)
 
 
-def cleanup(logfile, tempdir, testing):
+def cleanup(logfile, tempdir, testing, silent=False):
     """Cleanup the specified log file and the tmp directory, if desired."""
```
wpoely86: doc args

boegel:
Last remarks fixed, going in.

Thanks for the review @ocaisa and @wpoely86!

PR #1544, opened 2016-01-13
Fix packaging to exclude logs and test reports

rjeschmi:
Also clean up some pylint complaints

### DIFF
--- a/easybuild/tools/package/utilities.py
+++ b/easybuild/tools/package/utilities.py
@@ -198,8 +198,8 @@ class ActivePNS(object):
         if sel_pns in avail_pns:
             self.pns = avail_pns[sel_pns]()
         else:
-            raise EasyBuildError("Selected package naming scheme %s could not "
-                                 "be found in %s", sel_pns, avail_pns.keys())
+            raise EasyBuildError("Selected package naming scheme %s could not be found in %s",
+                                 sel_pns, avail_pns.keys())
 
     def name(self, easyconfig):
         """Determine package name"""
--- a/easybuild/tools/package/utilities.py
+++ b/easybuild/tools/package/utilities.py
@@ -118,9 +118,8 @@ def package_with_fpm(easyblock):
     ]
     # stripping off leading / to match expected glob in fpm
     exclude_files_glob = [
-        '--exclude %s' % quote_str(os.path.join(easyblock.installdir[1:], x))
-        for x
-        in exclude_files_glob
+        '--exclude %s' % quote_str(os.path.join(easyblock.installdir.lstrip(os.sep), x))
+        for x in exclude_files_glob
     ]
     _log.debug("exclude_glob: %s", exclude_files_glob)
     cmdlist = [
@@ -139,9 +138,7 @@ def package_with_fpm(easyblock):
     cmdlist.extend(exclude_files_glob)
 
     if build_option('debug'):
-        cmdlist.extend([
-            '--debug',
-        ])
+        cmdlist.append('--debug')
 
     cmdlist.extend([
         depstring,
@@ -211,5 +208,5 @@ class ActivePNS(object):
 
     def release(self, easyconfig):
         """Determine package release"""
-        release = self.pns.release()
+        release = self.pns.release(easyconfig)
         return release
--- a/easybuild/tools/package/utilities.py
+++ b/easybuild/tools/package/utilities.py
@@ -28,7 +28,7 @@ Various utilities related to packaging support.
 
 @author: Marc Litherland (Novartis)
 @author: Gianluca Santarossa (Novartis)
-@author: Robert Schmidt (Ottawa Hospital Research Institute)
+@author: Robert Schmidt (The Ottawa Hospital, Research Institute)
 @author: Fotis Georgatos (Uni.Lu, NTUA)
 @author: Kenneth Hoste (Ghent University)
 """
@@ -40,7 +40,8 @@ from vsc.utils import fancylogger
 from vsc.utils.missing import get_subclasses
 from vsc.utils.patterns import Singleton
 
-from easybuild.tools.config import PKG_TOOL_FPM, PKG_TYPE_RPM, build_option, get_package_naming_scheme
+from easybuild.tools.config import PKG_TOOL_FPM, PKG_TYPE_RPM, build_option, \
+    get_package_naming_scheme
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import which
 from easybuild.tools.package.package_naming_scheme.pns import PackageNamingScheme
@@ -49,12 +50,13 @@ from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME
 from easybuild.tools.utilities import import_available_modules, quote_str
 
 
-_log = fancylogger.getLogger('tools.package')
+_LOG = fancylogger.getLogger('tools.package')
 
 
 def avail_package_naming_schemes():
     """
-    Returns the list of valed naming schemes that are in the easybuild.package.package_naming_scheme namespace
+    Returns the list of valed naming schemes
+    They are loaded from the easybuild.package.package_naming_scheme namespace
     """
     import_available_modules('easybuild.tools.package.package_naming_scheme')
     class_dict = dict([(x.__name__, x) for x in get_subclasses(PackageNamingScheme)])
@@ -80,7 +82,7 @@ def package_with_fpm(easyblock):
     """
     workdir = tempfile.mkdtemp(prefix='eb-pkgs-')
     pkgtype = build_option('package_type')
-    _log.info("Will be creating %s package(s) in %s", pkgtype, workdir)
+    _LOG.info("Will be creating %s package(s) in %s", pkgtype, workdir)
 
     try:
         origdir = os.getcwd()
@@ -94,7 +96,7 @@ def package_with_fpm(easyblock):
     pkgver = package_naming_scheme.version(easyblock.cfg)
     pkgrel = package_naming_scheme.release(easyblock.cfg)
 
-    _log.debug("Got the PNS values for (name, version, release): (%s, %s, %s)", pkgname, pkgver, pkgrel)
+    _LOG.debug("Got the PNS values for (name, version, release): (%s, %s, %s)", pkgname, pkgver, pkgrel)
     deps = []
     if easyblock.toolchain.name != DUMMY_TOOLCHAIN_NAME:
         toolchain_dict = easyblock.toolchain.as_dict()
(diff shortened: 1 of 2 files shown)


easybuild/tools/package/utilities.py
```diff
@@ -40,7 +40,8 @@
 from vsc.utils.missing import get_subclasses
 from vsc.utils.patterns import Singleton
 
-from easybuild.tools.config import PKG_TOOL_FPM, PKG_TYPE_RPM, build_option, get_package_naming_scheme
+from easybuild.tools.config import PKG_TOOL_FPM, PKG_TYPE_RPM, build_option, \
+    get_package_naming_scheme, log_path
```
boegel: no need to break the import line here? if it's less than 120 chars, keep it on the same line

easybuild/tools/package/utilities.py
```diff
@@ -49,12 +50,13 @@
 from easybuild.tools.utilities import import_available_modules, quote_str
 
 
-_log = fancylogger.getLogger('tools.package')
+_LOG = fancylogger.getLogger('tools.package')
```
boegel: we usually don't use caps here, but it does make sense, so fine

boegel: it would be nicer to make this consistent across all modules though...

boegel: don't do that in this PR pls :)

easybuild/tools/package/utilities.py
```diff
@@ -102,14 +104,20 @@ def package_with_fpm(easyblock):
...
-        _log.debug("The dep added looks like %s ", dep)
+        _LOG.debug("The dep added looks like %s ", dep)
         dep_pkgname = package_naming_scheme.name(dep)
         depstring += " --depends %s" % quote_str(dep_pkgname)
 
+    exclude_files_glob = [
+        os.path.join(log_path(), "*.log"),
+        os.path.join(log_path(), "*.md"),
+    ]
+    exclude_files_glob = [os.path.join(easyblock.installdir[1:], x) for x in exclude_files_glob]
```
boegel: what's the `[1:]` about?

easybuild/tools/package/utilities.py
```diff
@@ -102,14 +104,20 @@ def package_with_fpm(easyblock):
...
     depstring = ''
     for dep in deps:
-        _log.debug("The dep added looks like %s ", dep)
+        _LOG.debug("The dep added looks like %s ", dep)
         dep_pkgname = package_naming_scheme.name(dep)
         depstring += " --depends %s" % quote_str(dep_pkgname)
 
+    exclude_files_glob = [
+        os.path.join(log_path(), "*.log"),
+        os.path.join(log_path(), "*.md"),
```
boegel: add comments here to make it clear that these are (EasyBuild) logs and test report

easybuild/tools/package/utilities.py
```diff
@@ -121,15 +129,26 @@ def package_with_fpm(easyblock):
         '--iteration', pkgrel,
         '--description', quote_str(easyblock.cfg["description"]),
         '--url', quote_str(easyblock.cfg["homepage"]),
+    ]
+    excludes = ['--exclude %s' % quote_str(x) for x in exclude_files_glob]
```
boegel: collapse this with the 2nd `exclude_files_glob =`:

```python
excludes = ['--exclude %s' % quote_str(os.path.join(self.installdir[1:], x)) for x in exclude_files_glob]
```

Still not sure what the `[1:]` is about

easybuild/tools/package/utilities.py
```diff
@@ -140,19 +159,20 @@ def package_with_fpm(easyblock):
...
         # rpmbuild is required for generating RPMs with FPM
         if pkgtool == PKG_TOOL_FPM and build_option('package_type') == PKG_TYPE_RPM:
             rpmbuild_path = which('rpmbuild')
             if rpmbuild_path:
-                _log.info("Required tool 'rpmbuild' found at %s", rpmbuild_path)
+                _LOG.info("Required tool 'rpmbuild' found at %s", rpmbuild_path)
             else:
-                raise EasyBuildError("rpmbuild is required when generating RPM packages with FPM, but was not found")
+                raise EasyBuildError("rpmbuild is required when generating RPM "
+                                     "packages but was not found")
```
boegel: don't wrap the string across two lines, it's < 120 chars?

easybuild/tools/package/utilities.py
```diff
@@ -173,20 +193,20 @@ def __init__(self):
         if sel_pns in avail_pns:
             self.pns = avail_pns[sel_pns]()
         else:
-            raise EasyBuildError("Selected package naming scheme %s could not be found in %s",
-                                 sel_pns, avail_pns.keys())
+            raise EasyBuildError("Selected package naming scheme %s could not "
+                                 "be found in %s", sel_pns, avail_pns.keys())
```
boegel: same here, why wrap?

if you need to wrap, don't break the string:

```python
raise EasyBuildError("Selected package naming scheme %s could not be found in %s",
                     sel_pns, avail_pns.keys())
```

easybuild/tools/package/utilities.py
```diff
@@ -173,20 +193,20 @@ def __init__(self):
...
 
-    def version(self, ec):
+    def version(self, easyconfig):
         """Determine package version"""
-        version = self.pns.version(ec)
+        version = self.pns.version(easyconfig)
         return version
 
-    def release(self, ec):
+    def release(self, easyconfig):  # pylint: disable=W0613
```
boegel: what are you disabling here? what doesn't pylint like about this one?

easybuild/tools/package/utilities.py
rjeschmi: It was a line length issue, both pep8 and pylint complain about it. 

If we want to make a standard longer line I'm fine with that.

boegel: our code style allows lines up until 120 chars, the default in pep8/pylint is 80 I think (which is a nightmare imho)

easybuild/tools/package/utilities.py
rjeschmi: I don't use the variable easyconfig, but it is probably useful to include it as a sort of documentation.

boegel: shouldn't it be passed down to `self.pns.release` (which can then choose to ignore it)?

boegel:
Jenkins: ok to test

easybuild/tools/package/utilities.py
```diff
@@ -110,6 +111,18 @@ def package_with_fpm(easyblock):
...
         depstring += " --depends %s" % quote_str(dep_pkgname)
 
+    # Excluding the EasyBuild logs and test reports that might be in the installdir
+    exclude_files_glob = [
+        os.path.join(log_path(), "*.log"),
+        os.path.join(log_path(), "*.md"),
+    ]
+    # stripping off leading / to match expected glob in fpm
+    exclude_files_glob = [
+        '--exclude %s' % quote_str(os.path.join(easyblock.installdir[1:], x))
```
boegel: use `easyblock.installdir.lstrip(os.pathsep)`

you have no idea what you're stripping off with `[1:]`

easybuild/tools/package/utilities.py
```diff
@@ -110,6 +111,18 @@ def package_with_fpm(easyblock):
...
+        for x
+        in exclude_files_glob
+    ]
```
boegel: no need to split across multiple lines? it looks very weird to me this way

easybuild/tools/package/utilities.py
```diff
@@ -121,10 +134,20 @@ def package_with_fpm(easyblock):
...
         '--description', quote_str(easyblock.cfg["description"]),
         '--url', quote_str(easyblock.cfg["homepage"]),
+    ]
+    _log.debug("excludes list: %s", exclude_files_glob)
+    cmdlist.extend(exclude_files_glob)
+
+    if build_option('debug'):
+        cmdlist.extend([
+            '--debug',
+        ])
```
boegel: do this single line, with `.append`:

```python
cmdlist.append('--debug')
```

boegel:
@rjeschmi: can we enhance the unit test(s) we have for the packaging support to check that logs/test report are *not* included?

test/framework/package.py
```diff
@@ -195,11 +216,18 @@ def test_package(self):
         self.assertTrue(os.path.isfile(pkgfile), "Found %s" % pkgfile)
 
         pkgtxt = read_file(pkgfile)
-        pkgtxt_regex = re.compile("Contents of installdir %s" % easyblock.installdir)
+        pkgtxt_regex = re.compile("STARTCONTENTS of installdir %s" % easyblock.installdir)
         self.assertTrue(pkgtxt_regex.search(pkgtxt), "Pattern '%s' found in: %s" % (pkgtxt_regex.pattern, pkgtxt))
 
+        no_logfiles_regex = re.compile(r'STARTCONTENTS.*\.(log|md)$.*ENDCONTENTS', re.DOTALL|re.MULTILINE)
+        self.assertFalse(no_logfiles_regex.search(pkgtxt), "Pattern '%s' found in: %s" % (no_logfiles_regex.pattern, pkgtxt))
```
boegel: this should say `"Pattern '%s' not found in: %s"`

easybuild/tools/package/utilities.py
```diff
@@ -121,10 +133,18 @@ def package_with_fpm(easyblock):
         '--iteration', pkgrel,
         '--description', quote_str(easyblock.cfg["description"]),
         '--url', quote_str(easyblock.cfg["homepage"]),
+    ]
+    _log.debug("excludes list: %s", exclude_files_glob)
```
boegel: you're logging this twice?

easybuild/tools/package/utilities.py
```diff
@@ -110,6 +111,17 @@ def package_with_fpm(easyblock):
...
+    exclude_files_glob = [
+        os.path.join(log_path(), "*.log"),
+        os.path.join(log_path(), "*.md"),
+    ]
+    # stripping off leading / to match expected glob in fpm
+    exclude_files_glob = [
+        '--exclude %s' % quote_str(os.path.join(easyblock.installdir.lstrip(os.sep), x))
+        for x in exclude_files_glob
+    ]
+    _log.debug("exclude_glob: %s", exclude_files_glob)
```
boegel: what's `exclude_glob`?

test/framework/package.py
rjeschmi: Well it is the error message, so the error is that the pattern is found, no?

boegel: you'll get an error like:

```AssertionError: Pattern '...' found in ...`

The assertion you're making is that the pattern is *not* found, however, and that's what's being reported as an error.

You can interpret it both ways though, but we generally mention what *should* be there in the error message.

boegel:
Going in, thanks @rjeschmi!

Issue #1535, opened 2016-01-06
Switching between naming schemes


ocaisa:
One major hurdle when switching between naming schemes is that the default installation directory is [dependent on the naming scheme](https://github.com/hpcugent/easybuild-framework/blob/bbd3b0ce82b4d67319d7cd934ead9c45e1cc5ff7/easybuild/tools/module_naming_scheme/mns.py#L97). This is unnecessary since the module files themselves don't care where the installation is. Changing the default to a unique value, for example that used by [easybuild_mns.py](https://github.com/hpcugent/easybuild-framework/blob/541556488193cc917c2478f2134ae1e51bbde7d4/easybuild/tools/module_naming_scheme/easybuild_mns.py), would make switching naming schemes relatively trivial via a simple script:
* Search for all installed easyconfigs in the installation space (no need to know what MNS was used to install them)
* Parse them
* Check existing module from old naming scheme to see if it is hidden, set flag if it is
* Rebuild all easyconfigs using `--module-only` with new MNS

I even think the default naming scheme tree should be created regardless of the naming scheme actually used just so you only need to indicate the new MNS when creating a new module tree.

ocaisa:
Changing this would be backwards compatible for everyone already using the new default and would work also work seamlessly for other supported naming schemes (but they would not be able to change naming scheme via the new script since only the new software would be in the *correct* installation path)

boegel:
@ocaisa: supporting using a fixed naming scheme for the installation directories makes a lot of sense indeed, I'm just not sure it's a good idea to make this the *default*

Although EasyBuild itself or existing modules/installations won't mind, people may have made assumptions on the naming scheme used for the installation in other places, e.g. a script that crawls the installation directories (for whatever reason).

So, your backward compatibility argument is not 100% correct.

That said, we do have some other options if we do want to change the default to using a fixed naming scheme:

* make sure people can opt-out easily, e.g. with `--disable-fixed-install-dir-naming-scheme` or whatever
* be very verbal about changing the default (mentioning it on the ML in a separate post, in the release notes, and in the release announcement, comments in the code where the name of the install directory is decided on, etc.)
* indicate that the naming scheme of the modules and the install directories doesn't match every time `eb` is run, with an option to silence that warning (maybe even with a generic `--slience-warnings=module-vs-install-dir-naming-scheme-mismatch`); maybe that's a bit over the top though, I'm not sure

[...]

PR #1546, opened 2016-01-14
Allow fixed install dir scheme

ocaisa:
This is fix for #1535 that defaults to keeping things as they are but adds the flexibility to use a fixed install scheme which would make it trivial to switch/trial other MNS

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1233,13 +1233,13 @@ class ActiveMNS(object):
 
     def det_install_subdir(self, ec):
         """Determine name of software installation subdirectory."""
+        self.log.debug("Determining software installation subdir for %s", ec)
         if build_option('fixed_installdir_naming_scheme'):
             subdir = os.path.join(ec['name'], det_full_ec_version(ec))
-            self.log.debug("Using fixed naming software installation subdir: %s (ec: %s)", subdir, ec)
+            self.log.debug("Using fixed naming software installation subdir: %s", subdir)
         else:
-            self.log.debug("Determining software installation subdir for %s", ec)
             subdir = self.mns.det_install_subdir(self.check_ec_type(ec))
-        self.log.debug("Obtained subdir %s", subdir)
+            self.log.debug("Obtained subdir %s", subdir)
         return subdir
 
     def det_devel_module_filename(self, ec, force_visible=False):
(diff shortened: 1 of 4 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1234,7 +1234,11 @@ def det_full_module_name(self, ec, force_visible=False):
     def det_install_subdir(self, ec):
         """Determine name of software installation subdirectory."""
         self.log.debug("Determining software installation subdir for %s", ec)
-        subdir = self.mns.det_install_subdir(self.check_ec_type(ec))
+        if build_option('fixed_installdir_naming_scheme'):
+            self.log.debug("Using fixed naming software installation subdir for %s", ec)
```
boegel: move this down, also mention the subdir itself in the log message? Also, `ec` is a dictionary, this will be a big/unreadable log message?

Both remarks fixed with this:

```python
self.log.debug("Using fixed naming software installation subdir: %s (ec: %s)", subdir, ec)
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1233,8 +1233,12 @@ def det_full_module_name(self, ec, force_visible=False):
 
     def det_install_subdir(self, ec):
         """Determine name of software installation subdirectory."""
-        self.log.debug("Determining software installation subdir for %s", ec)
-        subdir = self.mns.det_install_subdir(self.check_ec_type(ec))
+        if build_option('fixed_installdir_naming_scheme'):
+            subdir = os.path.join(ec['name'], det_full_ec_version(ec))
+            self.log.debug("Using fixed naming software installation subdir: %s (ec: %s)", subdir, ec)
```
boegel: move those outside the if/else? doesn't hurt to always print the subdir?

ocaisa: Just thought the repetition was annoying

boegel: this makes more sense to me:

```python
if build_option('fixed_installdir_naming_scheme'):
    self.log.debug("Using fixed naming for software installation subdir")
    subdir = os.path.join(ec['name'], det_full_ec_version(ec))
else:
    subdir = self.mns.det_install_subdir(self.check_ec_type(ec))
self.log.debug("Determined software installation subdir: %s (ec: %s)", subdir, ec)
```

boegel:
Jenkins: ok to test

ocaisa:
Download timed out

boegel:
Jenkins: test this please

boegel:
Going in, thanks @ocaisa!

PR #1550, opened 2016-01-16
skip dependencies marked as external modules when packaging

boegel:
fix for issue reported in https://github.com/hpcugent/easybuild/issues/188

@rjeschmi: please review?

### DIFF
--- a/easybuild/tools/package/utilities.py
+++ b/easybuild/tools/package/utilities.py
@@ -106,9 +106,12 @@ def package_with_fpm(easyblock):
                pprint.pformat([easyblock.toolchain.as_dict()] + easyblock.cfg.dependencies()))
     depstring = ''
     for dep in deps:
-        _log.debug("The dep added looks like %s ", dep)
-        dep_pkgname = package_naming_scheme.name(dep)
-        depstring += " --depends %s" % quote_str(dep_pkgname)
+        if dep.get('external_module', False):
+            _log.debug("Skipping dep marked as external module: %s", dep['name'])
+        else:
+            _log.debug("The dep added looks like %s ", dep)
+            dep_pkgname = package_naming_scheme.name(dep)
+            depstring += " --depends %s" % quote_str(dep_pkgname)
 
     cmdlist = [
         PKG_TOOL_FPM,


boegel:
original report by @DirkDeDraak:

When an EasyConfig file has an EXTERNAL MODULE as a dependency, it fails to package.
The reason is that the ec object has no toolchain set (None) and thus cannot retrieve the toolchain name.
The error occurs in easybuild/tools/module_naming_scheme/utilities.py in the det_full_ec_version function on:
```
    # determine main install version based on toolchain
    if ec['toolchain']['name'] == DUMMY_TOOLCHAIN_NAME:
        ecver = ec['version']
```


Here's a possible fix:
```
--- easybuild/tools/module_naming_scheme/utilities.py.orig      2016-01-13 16:23:57.962731609 +0100
+++ easybuild/tools/module_naming_scheme/utilities.py   2016-01-14 09:27:28.245027212 +0100
@@ -53,6 +53,9 @@

     ecver = None

+    # EXTERNAL MODULES don't have toolchain set
+    if ec.get('external_module') == True and ec['toolchain'] == None:
+        return None
     # determine main install version based on toolchain
     if ec['toolchain']['name'] == DUMMY_TOOLCHAIN_NAME:
         ecver = ec['version']
```

boegel:
@DirkDeDraak: thanks for reporting this

I don't think your proposed suggestion is the best one long-term, however. With your approach, you'll end up with argument to `fpm` like `--depends "some/external/module-None"`.
I'm not sure that even works, since the dependency can never be resolved?

My proposed change in this PR just skips dependencies that are marked as external modules, which seems more appropriate. Note that this does not affect the generated module which is included in the package, the `module load` statements for the external modules will still be there.

easybuild/tools/package/utilities.py
```diff
@@ -106,9 +106,12 @@ def package_with_fpm(easyblock):
                pprint.pformat([easyblock.toolchain.as_dict()] + easyblock.cfg.dependencies()))
     depstring = ''
     for dep in deps:
-        _log.debug("The dep added looks like %s ", dep)
-        dep_pkgname = package_naming_scheme.name(dep)
-        depstring += " --depends %s" % quote_str(dep_pkgname)
+        if dep.get('external_module', False):
```
rjeschmi: is get('external_module', default=False) also valid? I find this a bit unintuitive, but understand it now.

rjeschmi:
Looks good

easybuild/tools/package/utilities.py
boegel: No, see:

```
>> help({}.get)
Help on built-in function get:

get(...)
    D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.
```

```
>>> {}.get('external_module', False)
False
>>> {}.get('external_module', default=False)
TypeError: get() takes no keyword arguments
```

I'm using this instead of `dep['external_module']` because the `external_module` key is not defined for the toolchain (yet). And it's safer, anyway.

boegel:
Jenkins: test this please

easybuild/tools/package/utilities.py
rjeschmi: Yeah that's too bad. Anyway, it looks good. Thanks for fixing this.

I'll think about test cases at some point.

boegel:
Thanks for the review @rjeschmi!

PR #1552, opened 2016-01-17
check whether dependencies marked as external module are hidden

boegel:
fixes issue reported in https://github.com/hpcugent/easybuild/issues/185

@DirkDeDraak: can you check whether the problem you reported is fixed with this patch (which is a better place to handle it than the patch you proposed imho)?


boegel:
@wpoely86: please review?

wpoely86:
lgtm

wpoely86:
Maybe add test? :smiling_imp:

boegel:
@wpoely86: unit test (case) added

boegel:

@DirkDeDraak: do keep us posted whether this solves the problem you reported

PR #1555, opened 2016-01-19
add first version of script to check code style in easyconfigs (WIP)

boegel:
Currently only contains two checks, but we can gradually extend it.

```
Running style checks...

* checking for tab characters... OK
* checking for trailing whitespace... OK

All style checks passed.
```

My plan is to define an additional PR checker on Jenkins which only runs this (or to include it in the existing checker).

Related easyconfig PRs to make these two tests pass:
- https://github.com/hpcugent/easybuild-easyconfigs/pull/2302
- https://github.com/hpcugent/easybuild-easyconfigs/pull/2341

### DIFF
--- a/easybuild/scripts/style_check.py
+++ b/easybuild/scripts/style_check.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python
+##
+# Copyright 2016-2016 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of the University of Ghent (http://ugent.be/hpc).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+
+"""
+Script to check code style in easyconfig files
+
+@author: Kenneth Hoste (Ghent University)
+"""
+
+import os
+import re
+import sys
+from vsc.utils.generaloption import simple_option
+
+from easybuild.tools.filetools import find_easyconfigs, read_file
+
+#######################################################################################################################
+#
+# Each check should be implemented as a function, with the following characteristics:
+# * its name must start with 'check_'
+# * it takes one argument: a list of (path, lines) tuples for easyconfig files
+# * it prints a single-line message starting with '* ' and ending with 'OK' or 'FAILED (<reason>)'
+# * it returns a boolean value indicating whether the check pass
+#
+#######################################################################################################################
+
+
+def _check_by_line(ecs, line_check, ignore_keys=None):
+    """Template check performed on a per-line basis."""
+
+    if ignore_keys is None:
+        ignore_keys = []
+
+    comment_regex = re.compile('^\s*#', re.M)
+
+    last_key = None
+    key_line_regex = re.compile('^(?P<key>[a-z_]+)\s*=\s*', re.M)
+
+    faulty_ecs = []
+    for path, lines in ecs:
+        faulty_lines = []
+        for i, line in enumerate(lines):
+            # take keys to ignore into account (if any)
+            if ignore_keys:
+                res = key_line_regex.match(line)
+                if res:
+                    last_key = res.group('key')
+
+                if last_key in ignore_keys:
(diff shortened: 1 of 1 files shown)


boegel:
@rjeschmi asked me why I'm re-inventing PyLint here.

In some sense I am, but checking code style in easyconfig files is very different from checking regular Python code. Although there's significant overlap, PyLint does a lot more (complicated) checks that are irrelevant for easyconfigs. I'm not sure how easy/hard it would be to convince PyLint to check `.eb` files rather than `.py` files, btw...
Also, we need highly custom checks like: order of parameter definitions, grouping of parameter definitions, use of templates like `%(version)s`, exceptions like allowing trailing whitespace in `description`, etc.

PyLint supports writing custom checkers as plugins (cfr. http://docs.pylint.org/plugins.html), so maybe that's an avenue worth looking into.

I can certainly look into a separate "partial PyLint" check which calls out to PyLint to check stuff like tabs, space after `,`, maybe even indentation, etc, to avoid reinventing what PyLint already supports and is relevant for easyconfigs. I'd prefer doing that rather than running a custom PyLint checker.

Another aspects is speed: this scripts now crunches through all ~5k easyconfigs in 1.5s on my laptop; I'm pretty sure running PyLint would take longer (even when it's trimmed down to just a couple of relevant checks). I do have to check on the latter though...

With this in mind: do you consider this OK to go in as a starting point? It's certainly my intention to let this evolve, but it's about time we have an automated style review in place for easyconfigs, we've lost way too much human time on this already.

cc @wpoely86

wpoely86:
Just as reference:
```
$ time find -name '*.eb' -print0 | xargs -0  pylint -j 4 > /dev/null 

real	1m2.049s
user	3m42.890s
sys	0m7.692s
```
and
```$ time find -name '*.eb' -print0 | xargs -0  pep8 > /dev/null 

real	0m18.388s
user	0m18.084s
sys	0m0.238s
```

rjeschmi:
@wpoely86 that is useful, thanks. I'm not concerned abut pylint use particularly, just would rather reuse as much code as we can from elsewhere.

In the context of this analyzing easyconfigs, I see it making sense to build our own checker that can enforce standards like order and such.

wpoely86:
@rjeschmi yeah, I think we should take a look at pep8. It's already telling a lot of useful stuff:
```
./easybuild/easyconfigs/r/Ruby/Ruby-2.1.0-goolf-1.4.10.eb:2:2: W291 trailing whitespace
./easybuild/easyconfigs/r/Ruby/Ruby-2.1.0-goolf-1.4.10.eb:15:60: W291 trailing whitespace
./easybuild/easyconfigs/r/Ruby/Ruby-2.1.0-goolf-1.4.10.eb:16:74: W291 trailing whitespace
./easybuild/easyconfigs/r/RAxML/RAxML-8.1.1-goolf-1.4.10-mpi-sse3.eb:10:16: E225 missing whitespace around operator
./easybuild/easyconfigs/r/RAxML/RAxML-8.1.1-goolf-1.4.10-mt-avx.eb:10:16: E225 missing whitespace around operator
./easybuild/easyconfigs/r/RNAz/RNAz-2.1-ictce-4.0.6.eb:29:23: E126 continuation line over-indented for hanging indent
./easybuild/easyconfigs/r/RNAz/RNAz-2.1-ictce-4.0.6.eb:31:22: E126 continuation line over-indented for hanging indent
./easybuild/easyconfigs/r/R/R-3.1.2-intel-2015a.eb:277:14: E241 multiple spaces after ','
./easybuild/easyconfigs/r/R/R-3.1.2-foss-2015a.eb:277:14: E241 multiple spaces after ','
./easybuild/easyconfigs/r/R/R-3.1.2-foss-2015a.eb:378:1: W391 blank line at end of file
./easybuild/easyconfigs/r/R/R-3.1.1-intel-2014b.eb:36:5: E101 indentation contains mixed spaces and tabs
./easybuild/easyconfigs/r/R/R-3.1.1-intel-2014b.eb:36:5: W191 indentation contains tabs
./easybuild/easyconfigs/r/R/R-3.1.1-intel-2014b.eb:37:5: E101 indentation contains mixed spaces and tabs
./easybuild/easyconfigs/r/R/R-3.1.1-intel-2014b.eb:37:5: W191 indentation contains tabs
```

boegel:
@wpoely86: is this on current develop? there should be no more tabs there...

But yes, this looks useful. Question is whether it's also going to complain about stuff we don't want it to complain about. :-)

Also, how do we condense the output from `pep8` to make it easily digestible and dump it in a comment in a PR (not sure we need to).

wpoely86:
@boegel no this was not synced. 

With pep8 you can specify which checks (W191, E101, ...) have to be reported and which one can be ignored.

boegel:
It would be better to integrate this into the EasyBuild framework itself, rather than having it as a separate script; the main reason being that the automated style check can then be performed whenever `--new-pr` (or `--update-pr`) is used... Marked this as WIP for now.

boegel:
superseded by #1618 and #2038

PR #1558, opened 2016-01-20
implement support for --modules-header

boegel:
fixes https://github.com/hpcugent/easybuild/issues/186

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -158,11 +158,16 @@ class EasyBlock(object):
         # module generator
         self.module_generator = module_generator(self, fake=True)
 
-        # modules footer
+        # modules footer/header
         self.modules_footer = None
         modules_footer_path = build_option('modules_footer')
         if modules_footer_path is not None:
             self.modules_footer = read_file(modules_footer_path)
+
+        self.modules_header = None
+        modules_header_path = build_option('modules_header')
+        if modules_header_path is not None:
+            self.modules_header = read_file(modules_header_path)
 
         # easyconfig for this application
         if isinstance(ec, EasyConfig):
@@ -823,7 +828,7 @@ class EasyBlock(object):
         # load fake module
         fake_mod_data = self.load_fake_module(purge=True)
 
-        header = self.module_generator.MODULE_HEADER
+        header = self.module_generator.MODULE_SHEBANG
         if header:
             header += '\n'
 
@@ -1889,7 +1894,14 @@ class EasyBlock(object):
         """
         modpath = self.module_generator.prepare(fake=fake)
 
-        txt = self.make_module_description()
+        txt = self.module_generator.MODULE_SHEBANG
+        if txt:
+            txt += '\n'
+
+        if self.modules_header is not None:
+            txt += self.modules_header + '\n'
+
+        txt += self.make_module_description()
         txt += self.make_module_dep()
         txt += self.make_module_extend_modpath()
         txt += self.make_module_req()
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1898,7 +1898,7 @@ class EasyBlock(object):
         if txt:
             txt += '\n'
 
-        if self.modules_header is not None:
+        if self.modules_header:
             txt += self.modules_header + '\n'
 
         txt += self.make_module_description()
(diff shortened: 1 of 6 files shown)


boegel:
@wpoely86: please review?

easybuild/framework/easyblock.py
```diff
@@ -158,12 +158,17 @@ def __init__(self, ec):
...
-        # modules footer
+        # modules footer/header
         self.modules_footer = None
         modules_footer_path = build_option('modules_footer')
         if modules_footer_path is not None:
             self.modules_footer = read_file(modules_footer_path)
 
+        self.modules_header = None
+        modules_header_path = build_option('modules_header')
+        if modules_header_path is not None:
```
wpoely86: why the `not None`? if it's empty, that's also trouble?

easybuild/framework/easyblock.py
```diff
@@ -823,7 +828,7 @@ def make_devel_module(self, create_in_builddir=False):
         # load fake module
         fake_mod_data = self.load_fake_module(purge=True)
 
-        header = self.module_generator.MODULE_HEADER
+        header = self.module_generator.MODULE_SHEBANG
         if header:
```
wpoely86: is `self.module_generator.MODULE_SHEBANG` ever going to be empty?

easybuild/framework/easyblock.py
```diff
@@ -1889,7 +1894,14 @@ def make_module_step(self, fake=False):
         """
         modpath = self.module_generator.prepare(fake=fake)
 
-        txt = self.make_module_description()
+        txt = self.module_generator.MODULE_SHEBANG
+        if txt:
+            txt += '\n'
+
+        if self.modules_header is not None:
```
wpoely86: just `if self.modules_header` is better? Skip also if empty?

easybuild/framework/easyblock.py
wpoely86: ok, found the answer already

easybuild/tools/module_generator.py
```diff
@@ -187,7 +187,6 @@ def get_description(self, conflict=True):
             whatis = [description]
 
         lines = [
-            self.MODULE_HEADER.replace('%', '%%'),
```
wpoely86: why is this gone?

test/framework/options.py
```diff
@@ -972,20 +984,23 @@ def test_footer(self):
...
         toy_module = os.path.join(self.test_installpath, 'modules', 'all', 'toy', '0.0')
         if get_module_syntax() == 'Lua':
             toy_module += '.lua'
         toy_module_txt = read_file(toy_module)
-        footer_regex = re.compile(r'%s$' % module_footer_txt.replace('(', '\\(').replace(')', '\\)'), re.M)
-        msg = "modules footer '%s' is present in '%s'" % (module_footer_txt, toy_module_txt)
-        self.assertTrue(footer_regex.search(toy_module_txt), msg)
+        for txt in [modules_header_txt, modules_footer_txt]:
+            regex = re.compile(r'%s$' % txt.replace('(', '\\(').replace(')', '\\)'), re.M)
+            msg = "modules footer '%s' is present in '%s'" % (txt, toy_module_txt)
```
wpoely86: footer => footer/header

wpoely86:
lgtm

easybuild/framework/easyblock.py
boegel: the default value is `None`; if it's empty, you're doing something weird (e.g. use `eb --modules-header=''`, and you deserve to know you did something wrong?

You will get a proper error via `read_file` in that case, which is certainly better than just ignoring this and leaving people wondering why it doesn't work.

easybuild/framework/easyblock.py
boegel: yes, it's empty for module files in Lua syntax

easybuild/framework/easyblock.py
boegel: ok, makes sense here indeed, to avoid an extra newline that serves no purpose

easybuild/tools/module_generator.py
boegel: it's moved to `make_module_step`, we add there shebang there to start with, then add the `--modules-header` (if any), then the description

boegel:

PR #1559, opened 2016-01-21
add support to specify 'else' body for conditional statements in modules

boegel:


boegel:
@wpoely86: please review

I have no urgent need for this (yet), I fleshed it out because I thought we would need it for https://github.com/hpcugent/easybuild-easyblocks/pull/766, but it turns out we don't (really) need it.

Regardless: this works, and includes some minor cleanup + an extra unit test, so let's get it in if it's considered good to go.

wpoely86:
lgtm

boegel:

PR #1561, opened 2016-01-21
Added wrappers for shutil.copytree and os.chdir (WIP)

wpoely86:
To be used for the extended dry run.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1119,6 +1119,7 @@ def change_to(dst):
         return
 
     try:
+        _log.debug("Changing to directory %s" % dst)
         os.chdir(dst)
     except OSError, err:
         raise EasyBuildError("Failed to move to directory %s: %s", dst, err)
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1106,20 +1106,42 @@ def det_size(path):
 
 def copy_files(src, dst, symlinks=False, ignore=None):
     """
-    Wrapper around shutil.copytree to nicely handle extended dry run
-    Arguments are the same as for the original function
-    The destination directory must not exist
+    Copy the list of files or directories to a destintation
+    @src list or string of files and directories to copy
+    @dst the destination directory
+    @symlinks copy symlinks as symlinks
+    @ignore files to ignore in case the src is a directory
     """
     if build_option('extended_dry_run'):
         dry_run_msg("Copying directory %s to %s" % (src, dst), silent=build_option('silent'))
         return
 
     if os.path.exists(dst):
-        raise EasyBuildError("Destination directory %s of copy already exists.", dst)
+        if not os.path.isdir(dst):
+            raise EasyBuildError("File with the same name as destination directory %s \
+                                  already exists.", dst)
+        elif not os.path.listdir(dst):
+            raise EasyBuildError("Destination directory %s of copy already exists.", dst)
+        else:
+            try:
+                _log.debug("Removing empty directory %s." % dst)
+                os.rmdir(dst)
+            except OSError, err:
+                raise EasyBuildError("Failed to remove %s: %s", dst, err)
+
+    if isinstance(src, basestring):
+        src = [src]
 
     try:
-        _log.debug("Copying directory %s to %s" % (src, dst))
-        shutil.copytree(src, dst, symlinks=symlinks, ignore=ignore)
+        for item in src:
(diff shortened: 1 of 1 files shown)


easybuild/tools/filetools.py
```diff
@@ -1003,80 +1006,9 @@ def cleanup(logfile, tempdir, testing, silent=False):
...
-        shutil.copystat(src, dst)
-    except OSError, why:
-        if WindowsError is not None and isinstance(why, WindowsError):
-            # Copying file access times may fail on Windows
-            pass
-        else:
-            errors.extend((src, dst, str(why)))
-    if errors:
-        raise Error(errors)
+    """Has been replaced by copy_files"""
```
boegel: include `DEPRECATED` in the docstring please

easybuild/tools/filetools.py
```diff
@@ -1158,3 +1090,39 @@ def det_size(path):
         _log.warn("Could not determine install size: %s" % err)
 
     return installsize
+
+
+def copy_files(src, dst, symlinks=False, ignore=None):
+    """
+    Wrapper around shutil.copytree to nicely handle extended dry run
+    Arguments are the same as for the original function
+    The destination directoy must not exist
```
boegel: typo: `directory`

easybuild/tools/filetools.py
```diff
@@ -1158,3 +1090,39 @@ def det_size(path):
...
+    """
+    if build_option('extended_dry_run'):
+        dry_run_msg("Copying directory %s to %s" % (src, dst), silent=build_option('silent'))
+        return
+
+    if os.path.exists(dst):
+        raise EasyBuildError("Destination directory %s of copy already exists.", dst)
```
boegel: why is this needed?

do we know why `copytree` does this?

it would be useful if we're able to cope with this...

easybuild/tools/filetools.py
```diff
@@ -1158,3 +1090,39 @@ def det_size(path):
...
+
+
+def copy_files(src, dst, symlinks=False, ignore=None):
+    """
+    Wrapper around shutil.copytree to nicely handle extended dry run
+    Arguments are the same as for the original function
+    The destination directoy must not exist
+    """
+    if build_option('extended_dry_run'):
+        dry_run_msg("Copying directory %s to %s" % (src, dst), silent=build_option('silent'))
```
boegel: `copy_files` should be able to copy both files and directories imho...

easybuild/tools/filetools.py
wpoely86: it's in the docs that it must be so. We could integrate it here to delete the directory if it exists but I'm not sure if that is what we want

easybuild/tools/filetools.py
wpoely86: yes, it should. I will look if I find some time

easybuild/tools/filetools.py
boegel: I don't see why we shouldn't, if the directory is empty, that is.
Otherwise the caller is simply going to do exactly that anyway?

easybuild/tools/filetools.py
boegel: please make this `(WIP)` in the mean time

easybuild/tools/filetools.py
wpoely86: well, it the dir is empty, sure

wpoely86:
@boegel an attempt at including files, still needs proper testing

boegel:
@wpoely86: keep me posted if you'v tested this through; are you up for adding tests for these too?

boegel:
@wpoely86: please take into account https://github.com/hpcugent/easybuild-easyblocks/issues/492 here

easybuild/tools/filetools.py
```diff
@@ -1158,3 +1102,61 @@ def det_size(path):
...
+                raise EasyBuildError("Failed to remove %s: %s", dst, err)
+
+    if isinstance(src, basestring):
+        src = [src]
+
+    try:
+        for item in src:
+            if os.path.isfile(item):
+                _log.debug("Copying file %s to %s" % (item, dst))
+                shutil.copy2(item, dst)
```
boegel: you should take into account that individual files can also be symlinks, since `copy2` resolves the symlink rather than copying it

cfr. http://stackoverflow.com/questions/4847615/copying-a-symbolic-link-in-python

wpoely86:
This is going to stay WIP forever

PR #1564, opened 2016-01-22
Update gcc.py

ocaisa:
The old way has unintended consequences like including libs from GCCCore compiler as well as GCC (although I wonder if that comes GCCCore being a dependency of GCC)

### DIFF
--- a/easybuild/toolchains/gcc.py
+++ b/easybuild/toolchains/gcc.py
@@ -28,9 +28,10 @@ EasyBuild support for GCC compiler toolchain.
 @author: Kenneth Hoste (Ghent University)
 """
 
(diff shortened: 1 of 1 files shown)


ocaisa:
This is as much a question as a PR :P

boegel:
Jenkins: ok to test

ocaisa:
Answer is no, the issue that led to this is that if I list GCCcore as a dependency of GCC this has a knock on effect when I use GCC as a toolchain since loading the GCCcore module during a build gives precedence to GCCcore due to the LD_LIBRARY_PATH changes...in other words GCCCore can't be a dependency of GCC!

PR #1565, opened 2016-01-22
Allow user to define the default compiler optimsation level

ocaisa:


easybuild/tools/options.py
```diff
@@ -213,6 +213,9 @@ def override_options(self):
             'cleanup-builddir': ("Cleanup build dir after successful installation.", None, 'store_true', True),
             'cleanup-tmpdir': ("Cleanup tmp dir after successful run.", None, 'store_true', True),
             'color': ("Allow color output", None, 'store_true', True),
+            "default-optimisation": ("Set default optimization level, overriding easybuild optimization default "
+                                     "(possibilities are: noopt, lowopt, defaultopt, opt)",
```
boegel: hmm, we shouldn't hardcode the options here, we should try and obtain them via `easybuild.tools.toolchain.compiler`?

also, the option parser supports a `choice` type of option, so do:

```python
'default-opt-level': ("Specify default optimisation level", 'choice', 'store', 'defaultopt', known_opt_levels),
```

I like how you use different spelling for `optimisation` vs `optimization` in the option name and the help, so let's use `opt` instead ;)

easybuild/tools/toolchain/compiler.py
```diff
@@ -231,9 +231,19 @@ def _set_compiler_flags(self):
         fflags = [self.options.option(x) for x in self.COMPILER_F_FLAGS + self.COMPILER_F_UNIQUE_FLAGS \
                   if self.options.get(x, False)]
 
+        # Allow a user-defined default optimisation
+        if build_option('default_optimisation'):
+            if build_option('default_optimisation') in self.COMPILER_OPT_FLAGS:
+                default_opt = self.options.option(build_option('default_optimisation'))
```
boegel: you can simplify this a bit by only calling out to `build_option` once:

```python
default_opt_level = build_option('default_opt_level')
if default_opt_level in self.COMPILER_OPT_FLAGS:
    default_opt = default_opt_level
else:
    raise ...
```

Note that the `default_opt` default setting has just moved into `options.py` (see my other remark), no need to hardcode it anymore here

boegel:
Jenkins: ok to test

easybuild/tools/options.py
boegel: better:

```
from easybuild.tools.toolchain.compiler import Compiler
...
    'default-opt-level': ("Specify default optimisation level", 'choice', 'store', 'defaultopt', Compiler.COMPILER_OPT_FLAGS),
```

ocaisa: `opt` it is, copy paste is the enemy

easybuild/tools/toolchain/compiler.py
```diff
@@ -231,9 +231,17 @@ def _set_compiler_flags(self):
...
                   if self.options.get(x, False)]
 
+        # Allow a user-defined default optimisation
+        default_opt_level = build_option('default_opt_level')
+        if default_opt_level in self.COMPILER_OPT_FLAGS:
+            default_opt = default_opt_level
+        else:
+            raise EasyBuildError("Unknown value for default optimisation: %s (possibilities are %s)" %
+                                 (default_opt_level, self.COMPILER_OPT_FLAGS))
+
```
ocaisa: Invalid option already picked up by the option parser...could remove the if/else?

boegel: Hmm, leave it there, it's pretty cheap, and we may change things later such that `default_opt_level` comes from somewhere else

easybuild/tools/toolchain/compiler.py
```diff
@@ -231,9 +231,17 @@ def _set_compiler_flags(self):
         fflags = [self.options.option(x) for x in self.COMPILER_F_FLAGS + self.COMPILER_F_UNIQUE_FLAGS \
                   if self.options.get(x, False)]
 
+        # Allow a user-defined default optimisation
+        default_opt_level = build_option('default_opt_level')
+        if default_opt_level in self.COMPILER_OPT_FLAGS:
+            default_opt = default_opt_level
+        else:
+            raise EasyBuildError("Unknown value for default optimisation: %s (possibilities are %s)" %
+                                 (default_opt_level, self.COMPILER_OPT_FLAGS))
```
boegel: how about

```python
default_opt_level = build_option('default_opt_level')
if default_opt_level not in self.COMPILER_OPT_FLAGS:
    raise ...
```

and then also use `default_opt_level` below

boegel:
@ocaisa: fix for the failing tests in https://github.com/ocaisa/easybuild-framework/pull/28

boegel:
Going in, thanks a lot @ocaisa!

PR #1568, opened 2016-01-22
undefine unwanted env vars *after* setting up build environment via toolchain mechanism

boegel:
(fix for issue reported by @ocaisa during EB telco, cfr. https://github.com/hpcugent/easybuild/wiki/Conference-call-notes-20160120#other)

This is a backwards incompatible change, so I'd love feedback on this.

Currently, `unwanted_env_vars` is only used in two places: the easyblocks for GCC and imkl; neither should be affected by this:

* GCC is normally built with dummy toolchain, where no build env is set up by `toolchain.prepare()`
* for `imkl` only `$IMKLROOT` is undefined, which should not get defined/changed by `toolchain.prepare()`

I couldn't find any easyconfig files where this is used, but some EasyBuild users may be using `unwanted_env_vars` in their own 'private' easyconfig files...

Feedback from @wpoely86 would be nice, since he originally implemented the support for `unwanted_env_vars` (in fact, it was his first (framework?) contribution), cfr. https://github.com/hpcugent/easybuild-framework/pull/673.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -190,6 +190,10 @@ class EasyBlock(object):
         # list of loaded modules
         self.loaded_modules = []
 
+        # dictionary of unwanted environment variables that were unset and should be restored
+        # defined when env vars are unset in prepare_step, used when env vars are restored in cleanup_step
+        self.unwanted_env_vars = {}
+
         # iterate configure/build/options
         self.iter_opts = {}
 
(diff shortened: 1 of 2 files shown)


wpoely86:
Well, this changes the function of this. It was supposed to be used for external env variables, not the ones set up by EB. I don't think this is a proper solution for the mpich issue. Are we sure it's a good idea to allow this to unset EB vars from the easyconfig? I would assume you should always use a easyblock for something so special.

ocaisa:
I had my suspicions that that was the original intention for this. I do think this really should be handled in an easyblock because I see the potential pain that this could cause.

boegel:
@wpoely86/@ocaisa: thanks for the feedback

We should probably just update the help message for `unwanted_env_vars` then?

@ocaisa and @damianam are working on a proper solution for the MPICH issue (cfr. https://github.com/hpcugent/easybuild-easyconfigs/issues/2334) through an easyblock.

boegel:
@wpoely86, @ocaisa: thoughts on proper 'fix' for this?

ocaisa:
I'm not sure it needs a fix other than to add something to the help message
```
a list of external environment variables to be unset during the EasyBuild instance. This does not affect environment variables set by EasyBuild as part of the build process.
```

PR #1569, opened 2016-01-22
sanitize environment before each installation by undefining $PYTHON*

boegel:


pforai:
LGTM

rjeschmi:
generally a good idea. I wonder if we might want to relocate it and generalize it eventually?

The length of that particular file is a bit of a problem, but more of a general refactoring issue.

rjeschmi:
This is some of the nix env santization steps. I think it is generally a good idea to sanitize as much as possible. Might talk about namespace builds in the future (no root required isolated builds). Would need to whitelist os libraries we would include, but is probably an interesting feature

https://github.com/NixOS/nix/blob/a478e8a7bb8c24da0ac91b7100bd0e422035c62f/src/libstore/build.cc#L1622

rtmclay:
I found that a sanitized environment to be a great help in my XALT package to protect XALT from the crazy things that users do.

riccardomurri:
LGTM as well

boegel:
Thanks for the review everyone!

Issue #1539, opened 2016-01-08
EasyBuild installation test failed


JudyGardiner:
I'm fairly new to easybuild and wanted to evaluate the latest version for suitability on our system.  I followed your installation procedures successfully, but the unit test failed.  Here's the recommended troubleshooting information followed by the test output.  I would appreciate any insight you can provide.  --Judy

```
[oakley02]$ python -V
Python 2.6.6
[oakley02]$ type module
module is a function
module () 
{ 
    eval `$LMOD_CMD bash "$@"`
}
[oakley02]$ type -f module
-bash: type: module: not found
[oakley02]$ module --version

Modules based on Lua: Version 5.9.3  2015-04-04 06:40
    by Robert McLay mclay@tacc.utexas.edu

[oakley02]$ module av EasyBuild

------------------------------------------- /nfs/14/judithg/sciapps/eb/oak/modules/all -------------------------------------------
   EasyBuild/2.5.0

Use "module spider" to find all possible modules.
Use "module keyword key1 key2 ..." to search for all possible modules matching any of the "keys".

[oakley02]$ which -a eb
~/sciapps/eb/oak/software/EasyBuild/2.5.0/bin/eb
[oakley02]$ eb --version
This is EasyBuild 2.5.0 (framework: 2.5.0, easyblocks: 2.5.0) on host oakley02.osc.edu.
[oakley02]$ python -V
Python 2.6.6
[oakley02]$ type module
module is a function
module () 
{ 
    eval `$LMOD_CMD bash "$@"`
}
[oakley02]$ type -f module
-bash: type: module: not found
[oakley02]$ module --version

Modules based on Lua: Version 5.9.3  2015-04-04 06:40
    by Robert McLay mclay@tacc.utexas.edu

[oakley02]$ module av EasyBuild

------------------------------------------- /nfs/14/judithg/sciapps/eb/oak/modules/all -------------------------------------------
   EasyBuild/2.5.0

Use "module spider" to find all possible modules.
Use "module keyword key1 key2 ..." to search for all possible modules matching any of the "keys".

[oakley02]$ which -a eb
~/sciapps/eb/oak/software/EasyBuild/2.5.0/bin/eb
[oakley02]$ eb --version
This is EasyBuild 2.5.0 (framework: 2.5.0, easyblocks: 2.5.0) on host oakley02.osc.edu.

[oakley01]$ python -m test.framework.suite
WARNING: xmlrunner module not available, falling back to using unittest...

[...]

PR #1570, opened 2016-01-23
fix test_modules_tool_stateless unit test for stateless ModulesTool with Lmod as modules tool

boegel:
fix for #1539

I'd love to hear @rtmclay's thoughts on this, I'm not 100% sure this is correct.

The problem reported in #1539 doens't seem to be very reproducible.... Suddenly, it becomes very hard to reproduce, seemingly after Lmod throwing the expected error once.

I'm not sure why that happens, it seems like Lmod is remembering stuff somewhere?

### DIFF
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -334,9 +334,6 @@ class ModulesTest(EnhancedTestCase):
         """Check whether ModulesTool instance is stateless between runs."""
         test_modules_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'modules')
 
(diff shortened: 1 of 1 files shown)


boegel:
OK, found the issue, it is related to `$LMOD_DEFAULT_MODULEPATH` after all, it was leaking into the test, and Lmod was fiddling with `$MODULEPATH` because of it.

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

PR #1577, opened 2016-01-27
add support for --search-filename and --terse (REVIEW)

boegel:
This can be useful to support bash completion of easyconfig files.

cc @damienfrancois

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -178,7 +178,7 @@ def main(args=None, logfile=None, do_build=None, testing=False):
 
     # initialise logging for main
     global _log
-    _log, logfile = init_logging(logfile, logtostdout=options.logtostdout, silent=testing or options.last_log)
+    _log, logfile = init_logging(logfile, logtostdout=options.logtostdout, silent=testing or options.terse)
 
     # disallow running EasyBuild as root
     if os.getuid() == 0:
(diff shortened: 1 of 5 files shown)


easybuild/tools/filetools.py
```diff
@@ -367,8 +371,12 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False):
             else:
                 hit_lines.extend([" * %s" % fn for fn in hits])
 
-    for line in var_lines + hit_lines:
-        print_msg(line, log=_log, silent=silent, prefix=False)
+    if terse:
+        for line in hits:
+            print line
```
wpoely86: `print()` (for py3)

easybuild/tools/options.py
```diff
@@ -566,6 +569,10 @@ def postprocess(self):
             if not HAVE_AUTOPEP8:
                 raise EasyBuildError("Python 'autopep8' module required to reformat dumped easyconfigs as requested")
 
+        # some options imply enabling --terse
```
wpoely86: explain why in the comment because this two seem unrelated

easybuild/main.py
```diff
@@ -223,8 +223,6 @@ def main(args=None, logfile=None, do_build=None, testing=False):
         # print location to last log file, and exit
         last_log = find_last_log(logfile) or '(none)'
         print_msg(last_log, log=_log, prefix=False)
-        cleanup(logfile, eb_tmpdir, testing, silent=True)
```
wpoely86: so, this changes functionally? You can now combine `--last-log` with `--search`?

boegel: hmm, that's an unintended side-effect, but does it matter?

you can indeed combine these, but I don't think it makes a lot of sense to do it

doesn't hurt, so fine?

wpoely86: I don't know. Might bite us in the ass later?

boegel: This should be fine, there's lots of weird combo's of options you can make now.

Using `--last-log` and `--search` together won't break anything (I checked).

boegel:

PR #1579, opened 2016-01-28
Search for available modules in reverse order... 

ocaisa:
(i.e., highest level toolchain first) in case you have built the dependency at higher level for performance reasons (zlib is faster with Intel compilers for example)


boegel:
Jenkins: ok to test

ocaisa:
Addressed in #1614

PR #1581, opened 2016-01-28
changed default for dynamic (true) 

gppezzi:


pforai:
LGTM

We basically should have included this already earlier.

boegel:
Jenkins: ok to test

easybuild/toolchains/compiler/craype.py
```diff
@@ -59,12 +59,12 @@ class CrayPECompiler(Compiler):
...
 
     COMPILER_UNIQUE_OPTS = {
-        'dynamic': (False, "Generate dynamically linked executable"),
+        'dynamic': (True, "Generate dynamically linked executable"),
         'mpich-mt': (False, "Directs the driver to link in an alternate version of the Cray-MPICH library which \
                              provides fine-grained multi-threading support to applications that perform \
                              MPI operations within threaded regions."),
         'optarch': (False, "Enable architecture optimizations"),
-        'verbose': (True, "Verbose output"),
+        'verbose': (False, "Verbose output"),
```
boegel: why disable this? isn't it good to be a little bit verbose, so we know wth the wrappers are doing?

gppezzi: The argument was that verbose was needed while TC was experimental, but now we could make it optional.

But you definitely have a point here since we lose the actual command line parameters behind the wrappers. Wrappers arguments are usually minimal and if we disable verbose I suppose we won't see on the logs all the includes, links and so on... 

@pforai what do you think?

pforai:
Same as you said, we could keep verbose but still leave dynamic set to True. I personally like verbosity in logs :)

easybuild/toolchains/compiler/craype.py
boegel: OK, so let's revert `verbose` back to `True`, and merge this :shipit:

boegel:
Thanks @gppezzi!

PR #1586, opened 2016-01-29
Supply a working config for external meta to be used for Cray.

pforai:
This is a copy of the file as in use at CSCS as of 29/01/2016. We should ship this as part of the distro so people can actually get started. We can discuss things like removing GCC or something. /cc @gppezzi


gppezzi:
@pforai: gcc was added as external module for a non-Cray installation, you can remove it from here.

cray_external_modules_metadata.cfg
```diff
@@ -0,0 +1,64 @@
...
+name = PETSc
+version = 3.5.3.1
+prefix = PETSC_DIR
+
+[cray-petsc-64/3.5.3.1]
+name = PETSc
+version = 3.5.3.1
+prefix = PETSC_DIR
+
+[gcc/4.9.2]
```
boegel: @pforai: please remove this, see remark by @gppezzi

boegel:
@pforai: I would move the file in the `easybuild

boegel:
@pforai: I would move this file into the `easybuild` directory (in order not to clutter the root dir)

Also, `setup.py` should be updated so that the file is listed in `data_files`, and the default location for `--external-modules-metadata` should be updated to include this location by default

gppezzi:
Also we need to add the entries for fftw (which were removed from the toolchain). 

This is the updated file https://gist.github.com/gppezzi/d89f0545b01c8b024c74

boegel:
@gppezzi: can you send a PR to @pforai's branch to update (and maybe also move) the file?

boegel:
@pforai: please reopen to target `develop`

easybuild/cray_external_modules_metadata.cfg
```diff
@@ -0,0 +1,69 @@
...
+
+[fftw/3.3.4.5]
+name = FFTW
+version = 3.3.4.5
+prefix = FFTW_INC
+
+[fftw/3.3.4.3]
+name = FFTW
+version = 3.3.4.3
+prefix = FFTW_INC
```
boegel: this is not 100% correct, right, since it points to the `include` dir?

boegel: please change to `FFTW_INC/..`, cfr. https://github.com/hpcugent/easybuild-framework/pull/1637

easybuild/cray_external_modules_metadata.cfg
```diff
@@ -0,0 +1,69 @@
...
+
+[cray-petsc-64/3.5.3.1]
+name = PETSc
+version = 3.5.3.1
+prefix = PETSC_DIR
+
+[fftw/3.3.4.5]
+name = FFTW
+version = 3.3.4.5
+prefix = FFTW_INC
```
boegel: please change to `FFTW_INC/..`, cfr. https://github.com/hpcugent/easybuild-framework/pull/1637

boegel:
replaced by #1638, which adds corrected metadata (w.r.t. `fftw`), updates setup.py and makes sure the provided metadata is picked up by default

PR #1591, opened 2016-02-06
easybuild is a namespace package

boegel:
fix for #1588

PR #1592, opened 2016-02-06
Revert "easybuild is a namespace package"

boegel:
This reverts commit f6f96be40b53cdf4cde6fe0e044d6f107eebbd44 (cfr. #1591)

When `python setup.py install` is used, this error is thrown:

```
error: Namespace package problem: easybuild is a namespace package, but its
__init__.py does not call declare_namespace()! Please fix it.
(See the setuptools manual under "Namespace Packages" for details.)
```


boegel:
The PR tester should have catched this, it'll need to be enhanced.

PR #1593, opened 2016-02-06
easybuild is a namespace package (2nd attempt) (REVIEW)

boegel:
another attempt at #1591 (after that was reverted in #1592)

PR tester has been enhanced, and should now see the problem reported in #1592 that warranted a rollback of #1591...

documentation on namespace packages is available at https://pythonhosted.org/setuptools/pkg_resources.html#namespace-package-support

### DIFF
--- a/easybuild/tools/include.py
+++ b/easybuild/tools/include.py
@@ -65,16 +65,17 @@ pkg_resources.declare_namespace(__name__)
 # repository, i.e. in first-letter subdirectories
 EASYBLOCKS_PKG_INIT_BODY = """
 import pkg_resources
+import pkgutil
 
 # extend path so Python finds our easyblocks in the subdirectories where they are located
 subdirs = [chr(l) for l in range(ord('a'), ord('z') + 1)] + ['0']
 for subdir in subdirs:
-    pkg_resources.declare_namespace('%s.%s' % (__name__, subdir))
+    __path__ = pkgutil.extend_path(__path__, '%s.%s' % (__name__, subdir))
 
 # extend path so Python knows this is not the only place to look for modules in this package
 pkg_resources.declare_namespace(__name__)
 
-del subdir, subdirs, l
+del l, subdir, subdirs
 """
 
 
--- a/easybuild/tools/include.py
+++ b/easybuild/tools/include.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python
 # #
-# Copyright 2015-2015 Ghent University
+# Copyright 2015-2016 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
 # with support of Ghent University (http://ugent.be/hpc),
 # the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
-# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
 # and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
 #
 # http://github.com/hpcugent/easybuild
--- a/easybuild/tools/include.py
+++ b/easybuild/tools/include.py
@@ -30,6 +30,7 @@ Support for including additional Python modules, for easyblocks, module naming s
 """
 import os
 import sys
+from pkg_resources import fixup_namespace_packages
 from vsc.utils import fancylogger
 
 from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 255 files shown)


wpoely86:
I don't understand all the dark magic involved but it looks fine? At least, when jenkins agrees.

boegel:
@wpoely86: please re-review?

test/framework/systemtools.py
```diff
@@ -387,7 +387,7 @@ def test_os_version(self):
     def test_gcc_version_native(self):
         """Test getting gcc version."""
         gcc_version = get_gcc_version()
-        self.assertTrue(isinstance(gcc_version, basestring) or gcc_version == UNKNOWN)
+        self.assertTrue(isinstance(gcc_version, basestring) or gcc_version == UNKNOWN or gcc_version is None)
```
wpoely86: when is it `None`? Should be `UNKNOWN` in that case no?

boegel: There's no (real) GCC on OS X, then it's `None` (see also the tests below).

If `gcc --version` fails to run (exit code != 0), it's set to unknown.

setup.py
```diff
@@ -110,8 +110,12 @@ def find_rel_test():
     provides=["eb"] + easybuild_packages,
     test_suite="test.framework.suite",
     zip_safe=False,
-    install_requires=["vsc-base >= 2.2.6"],
+    install_requires=[
+        'setuptools',
```
boegel: this should specify a minimal version

cfr. https://github.com/hpcugent/vsc-install/pull/31

boegel: the minimal required version is quite old, basically all we need is `pkg_resources.declare_namespace` and `pkg_resources.fixup_namespace_packages`; both are available in 0.6:

```
-bash-4.1$ python -c "import setuptools; print setuptools.__version__; from pkg_resources import declare_namespace, fixup_namespace_packages"
0.6
-bash-4.1$ 
```

boegel:
Going in, thanks for the feedback @wpoely86!

easybuild/tools/include.py
```diff
@@ -56,27 +57,26 @@
...
 
 # extend path so Python finds our easyblocks in the subdirectories where they are located
 subdirs = [chr(l) for l in range(ord('a'), ord('z') + 1)] + ['0']
 for subdir in subdirs:
-    __path__ = extend_path(__path__, '%s.%s' % (__name__, subdir))
+    __path__ = pkgutil.extend_path(__path__, '%s.%s' % (__name__, subdir))
 
 # extend path so Python knows this is not the only place to look for modules in this package
-__path__ = extend_path(__path__, __name__)
+pkg_resources.declare_namespace(__name__)
```
ocaisa: Shouldn't this be `__path__`?

ocaisa: Maybe  not, I don't really understand what's going on here

wpoely86: Yeah, it's a bit of black magic.

boegel: no, `__name__` is correct

easybuild/tools/include.py
```diff
@@ -170,12 +170,23 @@ def include_easyblocks(tmpdir, paths):
...
+    reload(easybuild.easyblocks)
+    import easybuild.easyblocks.generic
+    reload(easybuild.easyblocks.generic)
+
+    # hard inject location to included (generic) easyblocks into Python search path
+    # only prepending to sys.path is not enough due to 'declare_namespace' in easybuild/easyblocks/__init__.py
+    new_path = os.path.join(easyblocks_path, 'easybuild', 'easyblocks')
+    easybuild.easyblocks.__path__.insert(0, new_path)
+    new_path = os.path.join(new_path, 'generic')
+    easybuild.easyblocks.generic.__path__.insert(0, new_path)
```
boegel: @ocaisa I think something extra needs to be done after this...

Can you print `easybuild.easyblocks.__path__` and `easybuild.easyblocks.__path__` after this, and then trigger the bug again?

easybuild/tools/include.py
```diff
@@ -170,12 +170,23 @@ def include_easyblocks(tmpdir, paths):
...
     _log.debug("Included software-specific easyblocks: %s", included_ebs)
 
-    # inject path into Python search path, and reload modules to get it 'registered' in sys.modules
+    # prepend new location to Python search path
     sys.path.insert(0, easyblocks_path)
-    reload(easybuild)
-    if 'easybuild.easyblocks' in sys.modules:
-        reload(easybuild.easyblocks)
-        reload(easybuild.easyblocks.generic)
+    fixup_namespace_packages(easyblocks_path)
```
boegel: in the tests we're making sure `easybuild.easyblocks` and `easybuild.easyblocks.generic` are imported before fiddling with `sys.path` (see above); do we need to do that here too?

boegel: see https://github.com/hpcugent/easybuild-framework/pull/1593/files#diff-4ac837acfb87e279976a2cf620d61620R139

boegel: hmm, these `import`s are already done at the top of `include.py`...

easybuild/tools/include.py
ocaisa: I'm not getting to that point, it's exiting before

boegel: That's weird, that's not what I'm seeing...

ocaisa: For me (with `--include-easyblock`) it was bailing out  at `fixup_namespace_packages(easyblocks_path)`

boegel: Is this the same problem as in https://gist.github.com/ocaisa/2294ee06303e5b5c3016 which you mentioned in #1677?

Because there it clearly mentions `Failed to import easyblock`, which is an error thrown by `verify_imports`...

ocaisa: In those cases I had undefined `EASYBUILD_INCLUDE_EASYBLOCKS` because I thought there was something specifically related to the easyblocks in the error

boegel: follow-up on this problem in #1678

PR #1596, opened 2016-02-07
Adjusted setup.py to include default Cray meta data config

pforai:
Related to https://github.com/hpcugent/easybuild-framework/pull/1586


boegel:
@pforai: please do this in #1586, keep things together in there

boegel:
replacd by #1638

PR #1602, opened 2016-02-09
use git rather than https in --new-pr/--update-pr

boegel:
using `https://` is very broken on some systems, cfr. https://bugs.launchpad.net/ubuntu/+source/git/+bug/1397685

and since access through SSH (`git://`) is required anyway to push, there's little point in using `https://`

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -684,6 +684,9 @@ def update_pr(pr, paths, commit_msg=None):
     _log.experimental("Updating pull request #%s with %s", pr, paths)
 
     github_user = build_option('github_user')
+    if github_user is None:
+        raise EasyBuildError("GitHub user must be specified to use --new-pr")
+
     pr_target_account = build_option('pr_target_account')
     pr_target_repo = build_option('pr_target_repo')
 
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -455,7 +455,10 @@ def setup_repo(git_repo, github_url, target_account, branch_name):
     except GitCommandError as err:
         alt_branch = 'pr_start_branch_%s_%s' % (branch_name, salt)
         _log.debug("Trying to work around checkout error ('%s') by using different branch name '%s'", err, alt_branch)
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86: please review

boegel:
Jenkins: test this please

wpoely86:
lgtm

boegel:
Tests are failing because the switch to `git://` now also requires an SSH pubkey to be registered on GitHub for `eb --new-pr -D`, while it didn't before...

Maybe I should try the `https://` URL first, and fall back to trying `git://` if that fails? Best of both worlds?

Thoughts @wpoely86?

wpoely86:
Yeah, I prefer https and only ssh when it's really needed. Bugs in other libraries should be solved there, not by EB.

boegel:
@wpoely86: please rereview? Now using `git` URL as fallback to `https`, rather than replacing `https`

easybuild/tools/github.py
```diff
@@ -408,6 +408,56 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+    else:
+        raise EasyBuildError("Branch '%s' not found at %s", branch_name, github_url)
+
+    _log.debug("Checking out branch '%s' from remote %s", branch_name, github_url)
+    try:
+        origin_branch.checkout(b=branch_name)
+    except GitCommandError as err:
+        alt_branch = 'pr_start_branch_%s_%s' % (branch_name, salt)
+        _log.debug("Trying to work around checkout error ('%s') by using different branch name '%s'", err, alt_branch)
+        origin_branch.checkout(b=alt_branch)
```
wpoely86: and if this fails too?

boegel: I can wrap this into a try/except too, but if this fails, there's nothing much we can do except throw an exception, because then something is very wrong/weird.

wpoely86: yes, we should throw a nice error message.

boegel:

PR #1608, opened 2016-02-11
stop running 'module purge', only restore environment

boegel:
This shouldn't break anything, and should be equivalent.

Even if modules were loaded before `eb` was started, since restoring the environment will also restore modules that were loaded in the environment where `eb` is being run.

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -392,11 +392,13 @@ class ModulesTool(object):
         if mod_paths is None:
             mod_paths = []
 
-        # purge all loaded modules if desired
+        # purge all loaded modules if desired by restoring initial environment
+        # actually running 'module purge' is futile (and wrong/broken on some systems, e.g. Cray)
         if purge:
-            self.purge()
             # restore initial environment if provided
-            if init_env is not None:
+            if init_env is None:
+                raise EasyBuildError("Initial environment required when purging before loading, but not available")
+            else:
                 restore_env(init_env)
 
         # make sure $MODULEPATH is set correctly after purging


boegel:
@pforai, @gppezzi: please review

pforai:
Looks trivial and the purge seems redundant, but I'd prefer an opt-in for this new behaviour (config option).

boegel:
@pforai: why an opt-in? Right now, we're always restoring the environment on `purge` (we have to in some cases to avoid breaking stuff), so there's no use in purging at all... The purge is basically 'undone' (in some sense) by restoring the environment.

boegel:
@wpoely86: thoughts?

easybuild/tools/modules.py
```diff
@@ -392,11 +392,13 @@ def load(self, modules, mod_paths=None, purge=False, init_env=None):
...
+        # purge all loaded modules if desired by restoring initial environment
+        # actually running 'module purge' is futile (and wrong/broken on some systems, e.g. Cray)
         if purge:
-            self.purge()
             # restore initial environment if provided
-            if init_env is not None:
+            if init_env is None:
+                raise EasyBuildError("Initial environment required when purging before loading, but not available")
+            else:
                 restore_env(init_env)
```
boegel: `sanitize_env` should also be run?

wpoely86:
should be okay

gppezzi:
I've used this PR to build Python (including numpy, which uses libsci) on Cray (Piz Dora) and didn't notice any side effects of not running purge.

easybuild/tools/modules.py
boegel: no need to sanitise again

the environment we're restoring here is the starting environment when the easyblock was instantiated (which is done *after* sanitising), so no need to sanitise again

boegel:
Going in, I'm pretty confident this shouldn't have any fallout.

Thanks for the feedback everyone!

PR #1613, opened 2016-02-13
RPATH support using ld wrapper (WIP)

rjeschmi:
I am working to getting this working to spec, but I thought I'd start the PR with a version using the wrapper. It works well enough, but I haven't removed the module deps yet (so LD_LIBRARY_PATH is still populated)

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -73,6 +73,7 @@ from easybuild.tools.module_generator import ModuleGeneratorLua, ModuleGenerator
 from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
 from easybuild.tools.modules import ROOT_ENV_VAR_NAME_PREFIX, VERSION_ENV_VAR_NAME_PREFIX, DEVEL_ENV_VAR_NAME_PREFIX
 from easybuild.tools.modules import get_software_root, modules_tool
+from easybuild.tools.rpath import prepare_ld_wrapper
 from easybuild.tools.package.utilities import package
 from easybuild.tools.repository.repository import init_repository
 from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME
@@ -1523,6 +1524,12 @@ class EasyBlock(object):
         # prepare toolchain: load toolchain module and dependencies, set up build environment
         self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent)
 
+        self.log.debug("prepare_step: PATH %s" % os.environ['PATH'])
+        if build_option('rpath'):
+            # Setup the environment and copy wrapper script into path
+            prepare_ld_wrapper()
+
+        self.log.debug("prepare_step after rpath : PATH %s" % os.environ['PATH'])
         # guess directory to start configure/build/install process in, and move there
         self.guess_start_dir()
 
--- a/easybuild/scripts/ld_wrapper.sh
+++ b/easybuild/scripts/ld_wrapper.sh
@@ -12,9 +12,10 @@
 # rpaths to binary.
 
 SCRIPTNAME=$(basename $0)
-
+ALLSCRIPTS=( $( which -a $SCRIPTNAME ) )
+$EB_LD_VERBOSE && echo "found scriptname: $SCRIPTNAME allscripts: ${ALLSCRIPTS[1]}"
+LDORIG=${ALLSCRIPTS[1]}
 LDWRAPPER=$EB_LD_FLAG
-LDORIG=${EB_LD:-/bin/$SCRIPTNAME}
 LINKER=$EB_LINKER_NAME   ## not yet implemented
 EB_LD_VERBOSE=${EB_LD_VERBOSE:-false}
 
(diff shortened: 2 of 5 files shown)


easybuild/tools/options.py
```diff
@@ -433,6 +433,16 @@ def package_options(self):
         self.log.debug("package_options: descr %s opts %s" % (descr, opts))
         self.add_group_parser(opts, descr)
 
+    def rpath_options(self):
+        # rpath Options
+        descr = ("RPATH options", "changes linking to include rpaths")
+
+        opts = OrderedDict({
+            'rpath': ("Enable RPATH support", None, 'store_true', False),
```
boegel: do you expect other options here?

if not, this shouldn't be an option group by itself?

easybuild/framework/easyblock.py
```diff
@@ -1523,6 +1524,12 @@ def prepare_step(self):
         # prepare toolchain: load toolchain module and dependencies, set up build environment
         self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent)
 
+        self.log.debug("prepare_step: PATH %s" % os.environ['PATH'])
+        if build_option('rpath'):
+            # Setup the environment and copy wrapper script into path
+            prepare_ld_wrapper()
```
boegel: hmm, shouldn't we do this in the `toolchain.prepare()` itself? keep things nicely together?

easybuild/tools/rpath.py
```diff
@@ -0,0 +1,57 @@
+"""A module to setup and use an ld wrapper to support RPATH"""
```
boegel: missing license header + author

easybuild/tools/rpath.py
```diff
@@ -0,0 +1,57 @@
...
+
+import os
+import copy
+import shutil
+import stat
+import tempfile
+from vsc.utils import fancylogger
+from easybuild.tools.filetools import adjust_permissions
+
+_log = fancylogger.getLogger('tools.package')
```
boegel: not package, rpath

easybuild/tools/rpath.py
```diff
@@ -0,0 +1,57 @@
...
+
+orig_os_environ = copy.deepcopy(os.environ)
```
boegel: we have copies of the original environment in various places already, no need to keep another?

easybuild/tools/rpath.py
```diff
@@ -0,0 +1,57 @@
...
+# RPATH is linux only
+ld_wrapper_script_loc = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "scripts", "ld_wrapper.sh"))
```
boegel: this is a constant, so name it `LD_WRAPPER_SCRIPT_LOC`

easybuild/tools/rpath.py
```diff
@@ -0,0 +1,57 @@
...
+    os.environ['EB_LD_FLAG'] = "1"
+    if DEBUG:
+        os.environ['EB_LD_VERBOSE'] = 'true'
+
+
+def teardown_ld_wrapper():
+    """
+    Return environment
+    """
+    os.environ = copy.deepcopy(orig_os_environ)
```
boegel: the environment is reset for every new build already, I don't think we need to worry about this?

easybuild/tools/rpath.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def check_rpath_support():
```
boegel: not called anywhere?

easybuild/tools/rpath.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    _log.debug("rpath: preparing ld wrapper script")
+    wrapper_dir = tempfile.mkdtemp(prefix='eb-ldwrapper-')
+    # copy wrapper script from framework
+    wrapper_ld = os.path.join(wrapper_dir, "ld")
+    wrapper_ld_gold = os.path.join(wrapper_dir, "ld.gold")
+    shutil.copy(ld_wrapper_script_loc, wrapper_ld)
+    shutil.copy(ld_wrapper_script_loc, wrapper_ld_gold)
+    adjust_permissions(wrapper_ld, stat.S_IXUSR, add=True)
+    adjust_permissions(wrapper_ld_gold, stat.S_IXUSR, add=True)
```
boegel:
```python
for linker in ['ld', 'ld.gold']:
    wrapper = os.path.join(wrapper_dir, linker)
    shutil.copy(ld_wrapper_script_loc, wrapper)
    adjust_permissions(wrapper, stat.S_IXUSR, add=True)
```

boegel:
@rjeschmi: this looks pretty basic (good!), but I think wrapper `ld` is not what we plan to do in the end, right?

rjeschmi:
I'll probably leave it as an "option"

I feel like having a wrapper available with some more generic hooks is a good thing and might lead to helpful debugging or special case handling in the future.

The way it is injected makes it low risk for inclusion even if it is not the default method

Ultimately for some cases this might work better than setting environment variables and is a good option to have around.

easybuild/scripts/ld_wrapper.sh
```diff
@@ -0,0 +1,97 @@
...
+## More aggressive rpath-ing
+    exclude_lib_paths=("/tmp" "/opt")
+  fi
+
+  L=""
+  lib_array=()
+  dir=`pwd`
+  for (( i=${#@}; i >= 0; i-- )); do
+    if [[ ${!i} == "--enable-new-dtags" ]]; then
+	## we are removing this flag if passed as it creates a copy of rpath to runpath.
```
ocaisa: Aren't we at the mercy of what the default is set to here? We probably should (always) be adding `--disable-new-dtags` to the arguments as well as removing `--enable-new-dtags`

easybuild/framework/easyblock.py
boegel: see `easybuild/tools/toolchain/toolchain.py:prepare`

boegel:
TODO:

* add sanity check (using pyreadelf)
* allow rpath methods: env, ld wrapper, compilers wrappers (multiple options at the same)
* no `$LD_LIBRARY_PATH` in modules (and check for it when loading dep modules)

PR #1576, opened 2016-01-27
fix get_toolchain_hierarchy w.r.t. finding subtoolchain version (WIP)

boegel:
cc @ocaisa

PR #1614, opened 2016-02-15
A simplified version of minimal toolchains that does the toolchain re…

ocaisa:
…solution as the easyconfig deps are being parsed.


easybuild/framework/easyconfig/tools.py
```diff
@@ -46,8 +46,8 @@
 from vsc.utils import fancylogger
 
 from easybuild.framework.easyconfig import EASYCONFIGS_PKG_SUBDIR
-from easybuild.framework.easyconfig.easyconfig import ActiveMNS, create_paths, process_easyconfig
-from easybuild.framework.easyconfig.easyconfig import robot_find_easyconfig
+from easybuild.framework.easyconfig.easyconfig import ActiveMNS, create_paths, process_easyconfig, module_is_available
+from easybuild.framework.easyconfig.easyconfig import robot_find_easyconfig, get_toolchain_hierarchy
```
boegel: are these actually still used here?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -707,7 +810,8 @@ def _parse_dependency(self, dep, hidden=False, build_only=False):
                     raise EasyBuildError("Found toolchain spec as dict with wrong keys (no name/version): %s", tc_spec)
             else:
                 raise EasyBuildError("Unsupported type for toolchain spec encountered: %s (%s)", tc_spec, type(tc_spec))
-
+        elif build_option('minimal_toolchains'):
```
boegel: change the logic to make this more readable:

```python
if tc_spec is None:
    if build_option('minimal_toolchains'):
        ...
    else:
        ... # ??
else:
    # (true) boolean value simply indicates that a dummy toolchain is used
    ...
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -707,7 +810,8 @@ def _parse_dependency(self, dep, hidden=False, build_only=False):
...
+            tc = robot_find_minimal_toolchain_of_dependency(dependency, tc)
```
boegel: empty line below to separate line that jumps back one indent level

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1137,6 +1241,49 @@ def robot_find_easyconfig(name, version):
 
     return res
 
+def robot_find_minimal_toolchain_of_dependency(dependency, parent_toolchain):
```
boegel: we try to use two empty lines between top-level definitions

easybuild/tools/robot.py
```diff
@@ -131,6 +131,7 @@ def resolve_dependencies(easyconfigs, retain_all_deps=False, minimal_toolchains=
     @param use_existing_modules: boolean for whether to prioritise the reuse of existing modules (works in
                                      combination with minimal_toolchains)
     """
+    minimal_toolchains = False
```
boegel: add a `FIXME` here that this needs to be cleaned up (or just clean it up ;-))

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1137,6 +1241,49 @@ def robot_find_easyconfig(name, version):
...
+    @dependency: dependency target dict (long and short module names may not exist yet)
+    @parent_toolchain: toolchain from which to derive the toolchain hierarchy to search
+    @return: minimal toolchain for which an easyconfig exists for this dependency (and matches build_options)
+    """
+    if build_option('use_existing_modules') and not build_option('retain_all_deps'):
+        existing_modules = modules_tool().available()
+    modtool = modules_tool()
+
+    newdep = copy.deepcopy(dependency)
+    toolchain_hierarchy = get_toolchain_hierarchy(parent_toolchain)
```
boegel: shorter variable names may make this easier to read

* `dependency` -> `dep`
* `toolchain` -> `tc`

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1137,6 +1241,49 @@ def robot_find_easyconfig(name, version):
...
+def robot_find_minimal_toolchain_of_dependency(dependency, parent_toolchain):
+    """
+    Find the minimal toolchain of a dependency
+
+    @dependency: dependency target dict (long and short module names may not exist yet)
+    @parent_toolchain: toolchain from which to derive the toolchain hierarchy to search
+    @return: minimal toolchain for which an easyconfig exists for this dependency (and matches build_options)
+    """
+    if build_option('use_existing_modules') and not build_option('retain_all_deps'):
+        existing_modules = modules_tool().available()
```
boegel: make sure that `existing_modules` is *always* defined (also in case of `else`)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -105,6 +105,109 @@ def new_ec_method(self, key, *args, **kwargs):
     return new_ec_method
 
 
+def module_is_available(full_mod_name, modtool, avail_modules, hidden):
```
boegel: we should probably move this inside `robot_find_minimal_toolchain_of_dependency` as an inner function,

or just simply it to this in `robot_find_minimal_toolchain_of_dependency`:

```python
module_exists = (full_mod_name in avail_modules) or (hidden and modtool.exist([full_mod_name])[0])
```

ocaisa: This function is needed by tools.py

boegel: well, you can replace it with the same single line there too?

having a function like this here is kind of weird

if you want to keep it, it should be moved to `easybuild.tools.modules` instead

ocaisa:
Includes the fix of #1576 and satisfies #1579

boegel:
@ocaisa: so both #1576 and #1579 should be closed?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -105,6 +105,123 @@ def new_ec_method(self, key, *args, **kwargs):
...
+        # parse the easyconfig
+        parsed_ec = process_easyconfig(path, validate=False)[0]
+        # search the dependencies for the version of the subtoolchain
+        #dep_tcs = [dep_toolchain['toolchain'] for dep_toolchain in parsed_ec['dependencies']
+        #                                      if dep_toolchain['toolchain']['name'] == subtoolchain_name]
+
+
+        dep_tcs = []
+        for dep in parsed_ec['dependencies']:
+            # dep == icc
```
boegel: remove this comment?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -105,6 +105,123 @@ def new_ec_method(self, key, *args, **kwargs):
...
+            ec = robot_find_easyconfig(dep['name'], det_full_ec_version(dep))
+            ec = process_easyconfig(ec, validate=False)[0]
+            #print ec
+            # ec == icc.eb
+            # dummy, GCCcore, binutils
```
boegel: remove these comments?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -51,22 +51,22 @@
...
 from easybuild.framework.easyconfig.default import DEFAULT_CONFIG
 from easybuild.framework.easyconfig.format.convert import Dependency
 from easybuild.framework.easyconfig.format.one import retrieve_blocks_in_spec
-from easybuild.framework.easyconfig.licenses import EASYCONFIG_LICENSES_DICT, License
+from easybuild.framework.easyconfig.licenses import EASYCONFIG_LICENSES_DICT
 from easybuild.framework.easyconfig.parser import DEPRECATED_PARAMETERS, REPLACED_PARAMETERS
 from easybuild.framework.easyconfig.parser import EasyConfigParser, fetch_parameters_from_easyconfig
 from easybuild.framework.easyconfig.templates import TEMPLATE_CONSTANTS, template_constant_dict
-
+from easybuild.toolchains.gcccore import GCCcore
```
boegel: please keep imports alphabetically sorted

ocaisa: You say that, but I'm pretty sure f comes before t in the alphabet...

boegel: oops :)

boegel:
@ocaisa: some style cleanup/minor refactoring in https://github.com/ocaisa/easybuild-framework/pull/30

There's one issue though... With this approach, the easyconfig file that is dumped in the install dir will no longer have the minimal toolchains that were used for the dependencies.

We should probably get that fixed before merging this PR?

ocaisa:
Ok, this method now creates a new subdir in the `easybuild` directory of the installation for a dump of parsed easyconfig (which describes exactly which toolchains were used for dependencies). This dump is also the one that is submitted to the repository so that the build can be reproduced (without using `--minimal-toolchains`).

ocaisa:
lgtm?

boegel:
Good to go, thanks @ocaisa!

easybuild/framework/easyblock.py
```diff
@@ -2333,7 +2349,7 @@ def build_and_install_one(ecdict, init_env):
         move_logs(app.logfile, application_log)
 
         try:
-            newspec = os.path.join(new_log_dir, "%s-%s.eb" % (app.name, det_full_ec_version(app.cfg)))
+            newspec = os.path.join(new_log_dir, ec_filename)
```
boegel: `ec_filename` is out of scope here... >_<

fixed in #1625

PR #1616, opened 2016-02-17
add support for --install-github-token and --check-github (REVIEW)

boegel:
Example output:

```
$ eb --check-github
== temporary log file in case of crash /var/folders/8s/_frgh9sj6m744mxt5w5lyztr0000gn/T/eb-UwwYW2/easybuild-Y_GLrS.log

Checking status of GitHub integration...

* GitHub user... boegel => OK
* GitHub token... a3e..9fc (len: 40) => OK (validated)
* git command... OK ("git version 2.5.4 (Apple Git-61); ")
* GitPython module... OK
* push access to boegel/easybuild-easyconfigs repo @ GitHub... OK
* creating gists... OK
* location to Git working dirs... OK (/Users/kehoste/work)

All checks PASSed!

Status of GitHub integration:
* --from-pr: OK
* --new-pr: OK
* --review-pr: OK
* --update-pr: OK
* --upload-test-report: OK
```


boegel:
Jenkins: test this please

boegel:
I'm happy with this now...

@wpoely86: please review?

easybuild/main.py
```diff
@@ -152,6 +152,31 @@ def build_and_install_software(ecs, init_session_state, exit_on_failure=True):
     return res
 
 
+def handle_github_options(options, orig_paths):
+    """Handle options related to GitHub integration, if any are set."""
```
wpoely86: doc args

easybuild/tools/github.py
```diff
@@ -204,7 +209,7 @@ class GithubError(Exception):
     pass
 
 
-def github_api_get_request(request_f, github_user=None, **kwargs):
+def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
     """
     Helper method, for performing get requests to GitHub API.
     @param request_f: function that should be called to compose request, providing a RestClient instance
```
wpoely86: new arg, new doc line

easybuild/tools/github.py
```diff
@@ -227,7 +234,7 @@ def github_api_get_request(request_f, github_user=None, **kwargs):
     return (status, data)
 
 
-def fetch_latest_commit_sha(repo, account, branch='master'):
+def fetch_latest_commit_sha(repo, account, branch='master', github_user=None, token=None):
```
wpoely86: idem

easybuild/tools/github.py
```diff
@@ -408,14 +416,45 @@ def post_comment_in_issue(issue, txt, repo=GITHUB_EASYCONFIGS_REPO, github_user=
...
+    @param silent: keep quiet (don't print any messages)
+    """
+    # copy or init git working directory
+    git_working_dirs_path = build_option('git_working_dirs_path')
+    if git_working_dirs_path:
+        workdir = os.path.join(git_working_dirs_path, repo_name)
+        if os.path.exists(workdir):
+            try:
+                print_msg("copying %s..." % workdir, silent=silent)
+                os.rmdir(path)
```
wpoely86: needs to be empty?

easybuild/tools/github.py
```diff
@@ -453,13 +492,44 @@ def setup_repo(git_repo, github_url, target_account, branch_name):
...
+        github_url = tmpl_github_url % (target_account, target_repo)
+        try:
+            repo = setup_repo_from(git_repo, github_url, target_account, branch_name, silent=silent)
+            break
+
+        except EasyBuildError as err:
+            errors.append("Checking out branch '%s' from %s failed: %s" % (branch_name, github_url, err))
+
+    if not repo:
+        raise EasyBuildError('; '.join(errors))
```
wpoely86: why not newlines?

easybuild/tools/github.py
```diff
@@ -718,12 +761,181 @@ def update_pr(pr, paths, commit_msg=None):
...
+    """
+    Check status of GitHub integration, and report back.
+    * check whether GitHub username is available
+    * check whether a GitHub token is available, and whether it works
+    * check whether git and GitPython are available
+    * check whether push access to own GitHub repositories works
+    * check whether creating gists works
+    * check whether location to local working directories for Git repositories is available (not strictly needed)
+    """
+    # start by assuming that everything works, individual checks will disable action that won't work
```
wpoely86: isn't it safer to do the inverse?

easybuild/tools/github.py
```diff
@@ -718,12 +761,181 @@ def update_pr(pr, paths, commit_msg=None):
...
+
+    # check whether we're online; if not, half of the checks are going to fail...
+    try:
+        urllib2.urlopen(GITHUB_URL, timeout=5)
+    except urllib2.URLError as err:
+        sys.stderr.write("\nERROR: checking status of GitHub integration must be done online.\n\n")
+        sys.exit(1)
+
+    # GitHub user
+    print "* GitHub user...",
```
wpoely86: why not print_msg? And do `print()` so we have less work when going to python 3

easybuild/tools/github.py
```diff
@@ -718,12 +761,181 @@ def update_pr(pr, paths, commit_msg=None):
...
+
+    print check_res
+
+    # check GitHub token
+    print "* GitHub token...",
+    github_token = fetch_github_token(github_user)
+    if github_token is None:
+        check_res = "(no token found) => FAIL"
+    else:
+        # don't print full token, should be kept secret!
```
wpoely86: why print it (or part of it) in the first place?

easybuild/tools/github.py
```diff
@@ -718,12 +761,181 @@ def update_pr(pr, paths, commit_msg=None):
...
+    for action in ['--from-pr', '--new-pr', '--review-pr', '--upload-test-report', '--update-pr']:
+        status[action] = True
+
+    print "\nChecking status of GitHub integration...\n"
+
+    # check whether we're online; if not, half of the checks are going to fail...
+    try:
+        urllib2.urlopen(GITHUB_URL, timeout=5)
+    except urllib2.URLError as err:
+        sys.stderr.write("\nERROR: checking status of GitHub integration must be done online.\n\n")
```
wpoely86: why not use an `EasyBuildError`?

easybuild/tools/github.py
```diff
@@ -718,12 +761,181 @@ def update_pr(pr, paths, commit_msg=None):
...
+        status['--new-pr'] = status['--update-pr'] = False
+
+    print check_res
+
+    # cleanup: delete test branch that was pushed to GitHub
+    if git_repo:
+        try:
+            git_repo.remotes.origin.push(branch_name, delete=True)
+        except GitCommandError:
+            pass
```
wpoely86: print at least a `_log.debug` when it fails

easybuild/tools/github.py
```diff
@@ -718,12 +761,181 @@ def update_pr(pr, paths, commit_msg=None):
...
+    if git_working_dirs_path:
+        print "OK (%s)" % git_working_dirs_path
+    else:
+        print "not found (suboptimal)"
+
+    # report back
+    if all(status.values()):
+        print "\nAll checks PASSed!\n"
+    else:
+        print '\n'.join([
```
wpoely86: why not just do a multi line string?

easybuild/tools/github.py
```diff
@@ -718,12 +761,181 @@ def update_pr(pr, paths, commit_msg=None):
...
+        else:
+            print "not supported"
+    print ''
+
+
 class GithubToken(object):
     """Representation of a GitHub token."""
 
-    # singleton metaclass: only one instance is created
-    __metaclass__ = Singleton
```
wpoely86: an repercussions of dropping this?

easybuild/tools/github.py
```diff
@@ -754,3 +966,73 @@ def __init__(self, user):
...
     """Fetch GitHub token for specified user from keyring."""
     return GithubToken(user).token
+
+
+def install_github_token(token, github_user, silent=False):
+    """
+    Install specified GitHub token for specified user.
+
+    @param token: GitHub token to install
+    @param github_user: GitHub user to install token for
```
wpoely86: don't forget silent

easybuild/tools/github.py
```diff
@@ -754,3 +966,73 @@ def __init__(self, user):
...
+    """
+    if github_user is None:
+        raise EasyBuildError("GitHub user must be specified to install GitHub token")
+
+    if HAVE_KEYRING:
+        # check if there's a token available already
+        current_token = fetch_github_token(github_user)
+
+        if current_token:
+            current_token = '%s..%s' % (current_token[:3], current_token[-3:])
```
wpoely86: there is no point in printing it?

easybuild/tools/github.py
```diff
@@ -754,3 +966,73 @@ def __init__(self, user):
...
+
+def validate_github_token(token, github_user):
+    """
+    Check GitHub token:
+    * see if it conforms expectations (only [a-f]+[0-9] characters, length of 40)
+    * see if it can be used for authenticated access
+    """
+    sha_regex = re.compile('^[0-9a-f]{40}')
+
+    # token should be 40 characters long, and only contain characters in [0-9a-f]
```
wpoely86: are we sure about this? It's officially written down by github?

easybuild/tools/github.py
```diff
@@ -754,3 +966,73 @@ def __init__(self, user):
...
+            print_msg("Token seems to be valid, installing it.", prefix=False, silent=silent)
+        else:
+            raise EasyBuildError("Token validation failed, not installing it. Please verify your token and try again.")
+
+        # install token
+        keyring.set_password(KEYRING_GITHUB_TOKEN, github_user, token)
+        print_msg("Token '%s..%s' installed!" % (token[:3], token[-3:]), prefix=False, silent=silent)
+
+    else:
+        msg = "Failed to obtain GitHub token from keyring, "
```
wpoely86: we have a decorator for this?

easybuild/tools/options.py
```diff
@@ -429,10 +429,12 @@ def github_options(self):
         descr = ("GitHub integration options", "Integration with GitHub")
 
         opts = OrderedDict({
+            'check-github': ("Check status of GitHub integration, and report back", None, 'store_true', False),
             'dump-test-report': ("Dump test report to specified path", None, 'store_or_None', 'test_report.md'),
             'from-pr': ("Obtain easyconfigs from specified PR", int, 'store', None, {'metavar': 'PR#'}),
             'git-working-dirs-path': ("Path to Git working directories for EasyBuild repositories", str, 'store', None),
-            'github-user': ("GitHub username", None, 'store', None),
+            'github-user': ("GitHub username", str, 'store', None),
+            'install-github-token': ("Install GitHub token (requires --github-user)", str, 'store', None),
```
wpoely86: I don't like this. We should use ask password like we do now in the docs

easybuild/tools/github.py
boegel: makes the implementation less clean imho, now I can just set one (or more) of the `status` entries to `False` if a check fails...

for some of these, multiple checks must pass

easybuild/tools/github.py
boegel: could be useful to spot silly mistakes, like not including the actual token but something else

easybuild/tools/github.py
boegel: I... I don't know. :p

Fixing that, thanks

easybuild/tools/github.py
boegel: I was just testing you here ;)

easybuild/tools/github.py
boegel: ugly because of indenting...

easybuild/tools/github.py
boegel: There shouldn't be, there can only be one token available for a particular user anyway

It has a slight performance impact in case the token is fetched multiple times, but that's negligible.

if anything, it has advantages, certainly for testing...

easybuild/tools/github.py
boegel: sure there is, may help with understanding why it doesn't allow authenticated access even though there's *some* value there

Say it prints this:

```
boe...gel
```

Then you probably know you switched token and username (or something like that)

easybuild/tools/github.py
boegel: All SHAs are 40 chars, and that's basically what it is

easybuild/tools/github.py
boegel: good point

easybuild/tools/options.py
boegel: You mean make it interactive with `getpass.getpass()`?

easybuild/tools/github.py
boegel: don

easybuild/tools/github.py
wpoely86: well, then you need to print all of it...

I would avoid printing it.

easybuild/tools/options.py
wpoely86: yes. It's a one time thing, so it's acceptable.

easybuild/tools/github.py
boegel: Fully printing it is a bad idea, since it's basically a password.

There's no harm done in a partial print though, since this info is not enough to steal/copy the token.

It can still be useful for debugging issues though, so I'm in favour of keeping this in.

easybuild/tools/github.py
```diff
@@ -754,3 +971,71 @@ def __init__(self, user):
...
+        current_token = '%s..%s' % (current_token[:3], current_token[-3:])
+        if build_option('force'):
+            msg = "WARNING: overwriting installed token '%s' for user '%s'..." % (current_token, github_user)
+            print_msg(msg, prefix=False, silent=silent)
+        else:
+            raise EasyBuildError("Installed token '%s' found for user '%s', not overwriting it without --force",
+                                 current_token, github_user)
+
+    # get token to install
+    token = getpass.getpass(prompt="Token: ")
```
wpoely86: you this be wrapped in a try? What if the user simple presses enter?

easybuild/tools/github.py
```diff
@@ -754,3 +971,71 @@ def __init__(self, user):
...
+    # validate token before installing it
+    print_msg("Validating token...", prefix=False, silent=silent)
+    valid_token = validate_github_token(token, github_user)
+    if valid_token:
+        print_msg("Token seems to be valid, installing it.", prefix=False, silent=silent)
+    else:
+        raise EasyBuildError("Token validation failed, not installing it. Please verify your token and try again.")
+
+    # install token
+    keyring.set_password(KEYRING_GITHUB_TOKEN, github_user, token)
```
wpoely86: The user also needs to set up the keyring to use plaintext (maybe not always, gnome keyring should also do it but that's not available on a hpc node). Maybe refer to the docs if this doesn't work?

easybuild/tools/github.py
boegel: then they'll try to install an invalid token, and the sanity check on it will fail

easybuild/tools/github.py
boegel: Yes, I'll try and make that clear in the docs.

See also https://github.com/hpcugent/easybuild/wiki/Review-process-for-contributions#setting-things-up

boegel:
Jenkins: test this please

boegel:
Jenkins PR builder plugin was updated...

Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Not holding this back any longer, it's time this goes in so it can be included in EasyBuild v2.7.0.

The documentation update at https://github.com/hpcugent/easybuild/pull/199 is still in the works, I'll try and get that done ASAP.

PR #1618, opened 2016-02-17
Add style tests

wpoely86:


wpoely86:
The first Jenkins test should fail.

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,115 @@
...
+
+_log = fancylogger.getLogger('easyconfig.style', fname=False)
+
+
+# any function starting with eb_check_ will be added to the tests
+# if the test number is added to the select list. The test number is
+# definied as AXXX where A is either E or W and XXX is a 3 digit number.
+# It should be mentioned in the docstring as a single word.
+# Read the pep8 docs to understand the arguments of these functions
+def eb_check_trailing_whitespace(physical_line, lines, line_number, total_lines):
```
boegel: this needs `@only_if_module_is_available('pep8')` too?

test/framework/style.py
```diff
@@ -0,0 +1,71 @@
...
+class StyleTest(TestCase):
+    log = fancylogger.getLogger("StyleTest", fname=False)
+
+    def test_style_conformance(self):
+        """Check the easyconfigs for style"""
+        if 'pep8' not in sys.modules:
+            print "Skipping style checks (no pep8 available)"
+            return
+
+
```
boegel: needs more whitespace?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,115 @@
...
+
+@only_if_module_is_available('pep8')
+def style_conformance(lst_easyconfigs, verbose=False):
+    """
+    Check the given list of easyconfigs for style
+    @param lst_easyconfigs list of file paths to easyconfigs
+    @param verbose print our statistics and be verbose about the errors and warning
+    @return the number of warnings and errors
+    """
+    # importing autopep8 fucks up pep8. We reload it just to be sure
```
boegel: some people may be offended by `f***` (I'm not, but I'm not the only one looking at this codebase ;-))

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+# any function starting with eb_check_ will be added to the tests
+# if the test number is added to the select list. The test number is
+# definied as AXXX where A is either E or W and XXX is a 3 digit number.
+# It should be mentioned in the docstring as a single word.
+# Read the pep8 docs to understand the arguments of these functions
+def eb_check_trailing_whitespace(physical_line, lines, line_number, total_lines):
+    """
+    W299
```
boegel: why `299`?

shouldn't be define our own range like `9xy` (or whatever other range that is not used by PEP8)?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,115 @@
...
+    Warn about trailing whitespace, expect for the description and comments.
+    This differs from the standard trailing whitespace check as that
+    will will warn for any trailing whitespace.
```
boegel: will will will will?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,115 @@
...
+    # we ignore some tests
+    # note that W291 has be replaced by our custom W299
+    options.ignore = ('E402',  # import not on top
+                      'W291',  # replaced by W299
+                      'E501',  # line too long
+                      )
+    if verbose:
+        options.verbose = 1
+    else:
+        options.verbose = 0
```
boegel:
```python
options.verbose = int(verbose)
```

or

```python
options.verbose = (0, 1)[verbose]
```

boegel:
@wpoely86: it fails, hooray! Now fix it? ;)

easybuild/framework/easyconfig/style.py
wpoely86: well, this should be a 'private' function. Never ever be called outside this module.

easybuild/framework/easyconfig/style.py
wpoely86: Well, there is a structure in the codes: https://pep8.readthedocs.org/en/latest/intro.html#error-codes

easybuild/framework/easyconfig/style.py
boegel: sure, but you still need the guard imho (since it uses `pep8`)

If you want to make it a private function, rename it to `_eb_check...`

boegel:
Jenkins: test this please

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+from vsc.utils import fancylogger
+
+from easybuild.tools.utilities import only_if_module_is_available
+
+try:
+    import pep8
+except ImportError:
+    pass
+
+_log = fancylogger.getLogger('easyconfig.style', fname=False)
```
boegel: why add a logger if you're not using it?

(don't remove it, use it below :))

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
+##
+# Copyright 2016 Ghent University
```
boegel: 2016-2016 (makes updating trivial next year)

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
```
boegel: FWO ;)

test/framework/style.py
```diff
@@ -0,0 +1,70 @@
+##
+# Copyright 2016 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
```
boegel: FWO

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+    import pep8
+except ImportError:
+    pass
+
+_log = fancylogger.getLogger('easyconfig.style', fname=False)
+
+
+# any function starting with _eb_check_ will be added to the tests
+# if the test number is added to the select list. The test number is
+# definied as AXXX where A is either E or W and XXX is a 3 digit number.
```
boegel: the "A is either E or W" bit is confusing :)

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+# It should be mentioned in the docstring as a single word.
```
boegel: on a separate line?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+# Read the pep8 docs to understand the arguments of these functions
+def _eb_check_trailing_whitespace(physical_line, lines, line_number, total_lines):
+    """
+    W299
+    Warn about trailing whitespace, expect for the description and comments.
+    This differs from the standard trailing whitespace check as that
+    will warn for any trailing whitespace.
+    """
```
boegel: doc args?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+    pass
+
+_log = fancylogger.getLogger('easyconfig.style', fname=False)
+
+
+# any function starting with _eb_check_ will be added to the tests
+# if the test number is added to the select list. The test number is
+# definied as AXXX where A is either E or W and XXX is a 3 digit number.
+# It should be mentioned in the docstring as a single word.
+# Read the pep8 docs to understand the arguments of these functions
```
boegel: pointer to pep8 docs?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+def _eb_check_trailing_whitespace(physical_line, lines, line_number, total_lines):
+    """
+    W299
+    Warn about trailing whitespace, expect for the description and comments.
+    This differs from the standard trailing whitespace check as that
+    will warn for any trailing whitespace.
+    """
+    comment_re = re.compile('^\s*#')
+    if comment_re.match(physical_line):
+        return None
```
boegel: not a big fan of inline returns (we rarely do this in the current codebase, I'd prefer sticking to that)

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+
+    result = pep8.trailing_whitespace(physical_line)
+    if result:
+        result = (result[0], result[1].replace("W291", "W299"))
```
boegel: ideally, we only have the `"W299"` string in a single place

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+    # we will not issue a warning
+    keys_re = re.compile("^(?P<key>[a-z_]+)\s*=\s*")
+
+    for line in reversed(lines[0:line_number]):
+        res = keys_re.match(line)
+        if res:
+            if res.group("key") == "description":
+                return None
+            else:
+                break
```
boegel: so, always break? and use `result = None` above => one less inline return :)

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+
+    result = pep8.trailing_whitespace(physical_line)
+    if result:
+        result = (result[0], result[1].replace("W291", "W299"))
+
+    # if the warning is about the multiline string of description
+    # we will not issue a warning
+    keys_re = re.compile("^(?P<key>[a-z_]+)\s*=\s*")
+
+    for line in reversed(lines[0:line_number]):
```
boegel: the `0` is useless, use `[:line_number]`

boegel: why reversed btw?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+    @return the number of warnings and errors
+    """
+    # importing autopep8 changes some pep8 functions.
+    # We reload it to be sure to get the real pep8 functions.
+    reload(pep8)
+
+    # register the extra checks before using pep8:
+    # any function in this module starting with `_eb_check_` will be used.
+    cands = globals()
+    for check_function in sorted([cands[f] for f in cands if callable(cands[f]) and f.startswith('_eb_check_')]):
```
boegel: `eb_check_` should be a constant?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+        pep8.register_check(check_function)
+
+    pep8style = pep8.StyleGuide(quiet=False, config_file=None)
+    options = pep8style.options
+    options.max_line_length = 120
```
boegel: add a comment above to explain why (and what the default is)

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+    # we ignore some tests
+    # note that W291 has be replaced by our custom W299
+    options.ignore = ('E402',  # import not on top
+                      'W291',  # replaced by W299
+                      'E501',  # line too long
```
boegel: please fix style, indent by 4 spaces

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+                      )
+    options.verbose = int(verbose)
```
boegel: just make the default value `0`?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+                break
+
+    return result
+
+
+@only_if_module_is_available('pep8')
+def style_conformance(lst_easyconfigs, verbose=False):
+    """
+    Check the given list of easyconfigs for style
+    @param lst_easyconfigs list of file paths to easyconfigs
```
boegel: why `lst_`?

test/framework/easyconfigs/GCC-4.6.3.eb
```diff
@@ -1,29 +1,29 @@
...
+sources = ['%s-%s.tar.gz' % (name.lower(), version),
+           'gmp-5.0.4.tar.bz2',
+           'mpfr-3.0.1.tar.gz',
+           'mpc-0.9.tar.gz',
            ]
+source_urls = ['http://ftpmirror.gnu.org/%(name)s/%(name)s-%(version)s' % {'name': name.lower(), 'version': version},  # GCC auto-resolving HTTP mirror
+               'http://ftpmirror.gnu.org/gmp',  # idem for GMP
+               'http://ftpmirror.gnu.org/mpfr',  # idem for MPFR
+               'http://www.multiprecision.org/mpc/download',  # MPC official
+               ]
```
boegel: fix with 'correct' indent style?

test/framework/easyconfigs/Python-2.7.10-ictce-4.1.13.eb
```diff
@@ -19,14 +19,14 @@ sources = [SOURCE_TGZ]
...
 dependencies = [
-#    ('bzip2', '1.0.6'),
-#    ('zlib', '1.2.8'),
-#    ('libreadline', '6.3'),
-#    ('ncurses', '5.9'),
-#    ('SQLite', '3.8.10.2'),
-#    ('Tk', '8.6.4', '-no-X11'),
-#    ('OpenSSL', '1.0.1m'),  # OS dependency should be preferred if the os version is more recent then this version, it's
-#    nice to have an up to date openssl for security reasons
+    #    ('bzip2', '1.0.6'),
```
boegel: single space after `#`?

test/framework/easyconfigs/SQLite-3.8.10.2-GCC-4.7.2.eb
```diff
@@ -27,8 +27,8 @@ sources = ['sqlite-autoconf-%s.tar.gz' % version_str]
 
 # commented out for testing to avoid having to add them all - dependencies are tested in other files
 dependencies = [
-#    ('libreadline', '6.3'),
-#    ('Tcl', '8.6.4'),
+    #    ('libreadline', '6.3'),
+    #    ('Tcl', '8.6.4'),
```
boegel: single space after `#`

test/framework/easyconfigs/SQLite-3.8.10.2-goolf-1.4.10.eb
```diff
@@ -27,8 +27,8 @@ sources = ['sqlite-autoconf-%s.tar.gz' % version_str]
 
 # commented out for testing to avoid having to add them all - dependencies are tested in other files
 dependencies = [
-#    ('libreadline', '6.3'),
-#    ('Tcl', '8.6.4'),
+    #    ('libreadline', '6.3'),
+    #    ('Tcl', '8.6.4'),
```
boegel: single space after `#`

test/framework/easyconfigs/gzip-1.4-GCC-4.6.3.eb
```diff
@@ -18,10 +18,10 @@ homepage = "http://www.gzip.org/"
 description = "gzip (GNU zip) is a popular data compression program as a replacement for compress"
 
 # test toolchain specification
-toolchain = {'name': 'GCC','version': '4.6.3'}
+toolchain = {'name': 'GCC', 'version': '4.6.3'}
 
 # source tarball filename
-sources = ['%s-%s.tar.gz'%(name,version)]
+sources = ['%s-%s.tar.gz' % (name, version)]
```
boegel: `%(version)s` etc? ;)

test/framework/style.py
```diff
@@ -0,0 +1,70 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+Style tests for easyconfig files. Uses pep8.
```
boegel: uses `pep8` is irrelevant here; it uses whatever the framework functions we're calling are using?

test/framework/style.py
```diff
@@ -0,0 +1,70 @@
...
+except ImportError:
+    pass
+
+
+class StyleTest(TestCase):
+    log = fancylogger.getLogger("StyleTest", fname=False)
+
+    def test_style_conformance(self):
+        """Check the easyconfigs for style"""
+        if 'pep8' not in sys.modules:
```
boegel: we can enhance this to fail the test if we're on Jenkins

boegel: we should probably have a method in `EnhancedTestCase` named `jenkins` (which could check `$CONTINUOUS_INTEGRATION` or `$CI` (like https://github.com/auchenberg/volkswagen does, and/or check for `$JENKINS_URL`)

and then do this:

```python
if 'pep8' not in sys.modules:
    if self.jenkins():
        self.assertTrue(False, "pep8 not available")
    else:
        print "Skipping style checks (no pep8 available)"
        return
```

test/framework/style.py
```diff
@@ -0,0 +1,70 @@
...
+
+from easybuild.framework.easyconfig.style import style_conformance
+
+try:
+    import pep8
+except ImportError:
+    pass
+
+
+class StyleTest(TestCase):
```
boegel: use `EnhancedTestCase` (like the other test modules do)

test/framework/style.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_style_conformance(self):
+        """Check the easyconfigs for style"""
+        if 'pep8' not in sys.modules:
+            print "Skipping style checks (no pep8 available)"
+            return
+
+        # all available easyconfig files
+        test_easyconfigs_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs')
+        specs = glob.glob('%s/*.eb' % test_easyconfigs_path)
+        specs = sorted(specs)
```
boegel: why sort?

test/framework/style.py
```diff
@@ -0,0 +1,70 @@
...
+
+        result = style_conformance(specs)
+
+        self.assertEqual(result, 0, "Found code style errors (and/or warnings).")
```
boegel: the message should say `"No code style errors (and/or warnings) found"`

and then you'd get:

```
AssertionError: No code style errors (and/or warnings) found
```

test/framework/suite.py
```diff
@@ -106,8 +107,8 @@
 
 # call suite() for each module and then run them all
 # note: make sure the options unit tests run first, to avoid running some of them with a readily initialized config
-tests = [gen, bl, o, r, ef, ev, ebco, ep, e, mg, m, mt, f, run, a, robot, b, v, g, tcv, tc, t, c, s, l, f_c, sc, tw,
-         p, i, pkg, d, env, et, y]
+tests = [gen, bl, o, r, ef, ev, ebco, ep, e, mg, m, mt, f, run, a, robot, b, v, g, tcv, tc, t, c, s, l, f_c, sc, st,
```
boegel: just add it at the end? these are not sorted alphabetically, they have some kind of order

easybuild/framework/easyconfig/style.py
wpoely86: I can copy & paste the pep8 docs but I prefer not to. The names are already quite clear and the precise meaning should be looked up in the url above

easybuild/framework/easyconfig/style.py
wpoely86: It's much more readable in this way. Inline returns are not bad.

easybuild/framework/easyconfig/style.py
wpoely86: it should be in the docs as well, so it's gonna be here twice.

easybuild/framework/easyconfig/style.py
boegel: please refer to the right place in the pep8 docs in the docstring; the comments won't show up in the auto-generated API documentation for EB we'll have at easybuild.readthedocs.org someday...

easybuild/framework/easyconfig/style.py
boegel: can we string template a docstring? :P

easybuild/framework/easyconfig/style.py
wpoely86: list

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,113 @@
...
+
+# any function starting with _eb_check_ will be added to the tests
+# if the test number is added to the select list. The test number is
+# definied as AXXX where A is either E or W and XXX is a 3 digit number.
+# It should be mentioned in the docstring as a single word.
+# Read the pep8 docs to understand the arguments of these functions
+def _eb_check_trailing_whitespace(physical_line, lines, line_number, total_lines):
+    """
+    W299
+    Warn about trailing whitespace, expect for the description and comments.
```
ocaisa: except

easybuild/framework/easyconfig/style.py
wpoely86: it is 0?

easybuild/framework/easyconfig/style.py
wpoely86: if you really want too but I don't see the point. These functions should 'private' to this file.

easybuild/framework/easyconfig/style.py
wpoely86: good question, I think to have a hit quicker? Not sure anymore.

wpoely86: a no, it's because of the break

easybuild/framework/easyconfig/style.py
wpoely86: what this does is: find the last `key = value` going back from the current line to the first one. It's a `description`, we pass the test else we return the default pep8 test.

easybuild/framework/easyconfig/style.py
boegel: easyconfig*s* makes it clear it's a list?

let's not start using Hungarian syntax, we don't do it anywhere else in the codebase...

easybuild/framework/easyconfig/style.py
boegel: sure, but `eb_check_` is what we call a 'magic string'

right now, it's sort of OK, since it's only used here (and in the function names, but there you can't use a constant anyway)

but we should definitely not start using it in other places, like log messages

in fact... I'd like to see a log message, or maybe even a print, for a list of checks functions being performed?

easybuild/framework/easyconfig/style.py
boegel: well, since even you're forgetting it easily, add a comment above this line why the search is reversed...

test/framework/style.py
wpoely86: fine by me

test/framework/style.py
wpoely86: why not?

boegel: well, why not reorder them randomly?

if there's no reason to sort, then don't sort?

also: they're already sorted in the result of `glob.glob`, I think

wpoely86: no `glob` is giving a random order.

boegel: well, ok, if this affects the output of the test results in any way, then yes, sorting them makes sense

wpoely86:
This still needs a test for every custom style check we have added (currently just one).

boegel:
I need to get back to this, but rather than cooking up our own approach, we can also rely on an external service that integrates with GitHub, like http://docs.sonarqube.org/display/PLUG/Python+Plugin

wpoely86:
Why? We are replying on pep8 and pylint? I think style is something which we want to control ourself?

boegel:
@wpoely86 as far as I can tell, you can get full control over how you define your code style via SonarQube too, you can supply a custom`pylintrc`, and it integrates well with Travis (see for example https://github.com/singularityware/singularity/pull/341)

And it avoids that we have to write & maintain our own stuff (although we still may have to do that for easyconfigs, I'm mostly talking out the actual Python code in framework/easyblocks here).

wpoely86:
This PR is only about style testing for easyconfigs?

test/framework/style.py
```diff
@@ -0,0 +1,76 @@
...
+    def test_style_conformance(self):
+        """Check the easyconfigs for style"""
+        if 'pep8' not in sys.modules:
+            print "Skipping style checks (no pep8 available)"
+            return
+            # if self.jenkins():
+            #    self.assertTrue(False, "pep8 not available")
+            # else:
+            #    print "Skipping style checks (no pep8 available)"
+            #    return
```
boegel: @wpoely86 please remove this

test/framework/style.py
```diff
@@ -0,0 +1,76 @@
...
+        specs = sorted(specs)
+
+        result = style_conformance(specs)
+
+        self.assertEqual(result, 0, "No code style errors (and/or warnings) found.")
+
+
+def suite():
+    """Return all style tests for easyconfigs."""
+    return TestLoader().loadTestsFromTestCase(StyleTest)
```
boegel: @wpoely86 please use `return TestLoaderFiltered().loadTestsFromTestCase(StyleTest, sys.argv[1:])  ` to allow filtering of tests

test/framework/style.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+if __name__ == '__main__':
+    main()
```
boegel: switch to `TextTestRunner(verbosity=1).run(suite())`, see #1828

test/framework/style.py
```diff
@@ -0,0 +1,76 @@
...
+            print "Skipping style checks (no pep8 available)"
+            return
+            # if self.jenkins():
+            #    self.assertTrue(False, "pep8 not available")
+            # else:
+            #    print "Skipping style checks (no pep8 available)"
+            #    return
+
+        # all available easyconfig files
+        test_easyconfigs_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs')
```
boegel: test easyconfigs are in a different place now, see #1970

easybuild/framework/easyconfig/style.py
boegel: just tested this, templating on a docstring doesn't really work...

easybuild/framework/easyconfig/style.py
boegel: I understand, but stick to a single `return`:

```python
for line in reversed(lines[:line_number]):
    res = keys_re.match(line)
    if res and res.group("key") == "description":
        result = None
        break

return result
```

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,128 @@
...
+            if res.group("key") == "description":
+                return None
+            else:
+                break
+
+    return result
+
+
+@only_if_module_is_available('pep8')
+def style_conformance(easyconfigs, verbose=False):
```
boegel: rename to `check_easyconfigs_style` ?

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Check the given list of easyconfigs for style
+    :param easyconfigs list of file paths to easyconfigs
+    :param verbose print our statistics and be verbose about the errors and warning
+    :return the number of warnings and errors
```
boegel: missing `:`

easybuild/framework/easyconfig/style.py
boegel: @wpoely86 let's flesh out `_eb_check_` as a constant

easybuild/framework/easyconfig/style.py
```diff
@@ -0,0 +1,128 @@
...
+    options = pep8style.options
+    # we deviate from standard pep8 and allow 120 chars
+    # on a line: the default of 79 is too narrow.
+    options.max_line_length = 120
+    # we ignore some tests
+    # note that W291 has be replaced by our custom W299
+    options.ignore = (
+        'E402',  # import not on top
+        'W291',  # replaced by W299
+        'E501',  # line too long
```
boegel: Why are we ignoring E402 and E501? If there's a good reason, please mention why.

wpoely86: E402 because some easyconfigs imported stuff (not on top)
E501 because some easyconfigs have very long lines

boegel: both of them are valid reason to fail the style check?

test/framework/easyconfigs/test_ecs/g/gzip/gzip-1.6-GCC-4.9.2.eb
```diff
@@ -13,7 +13,8 @@
 easyblock = 'ConfigureMake'
 
 name = 'gzip'
-version = '1.6'  # FIXME bug: quotes are required here to make sure this is parsed a a string, not a floating point value
+# FIXME bug: quotes are required here to make sure this is parsed a a string, not a floating point value
```
boegel: @wpoely86 please drop this comment, it doesn't make any sense here (it's copy-pasted from the equivalent `.yeb` easyconfig)

test/framework/style.py
```diff
@@ -0,0 +1,70 @@
...
+    log = fancylogger.getLogger("StyleTest", fname=False)
+
+    def test_style_conformance(self):
+        """Check the easyconfigs for style"""
+        if 'pep8' not in sys.modules:
+            print "Skipping style checks (no pep8 available)"
+            return
+
+        # all available easyconfig files
+        test_easyconfigs_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'test_ecs')
```
boegel: @wpoely86 based on the tweaked test easyconfigs, we should probably revert testing only the easyconfigs in `test_ecs`, and test *all* test easyconfigs (including the `.yeb` ones)? I missed that earlier, sorry...

boegel: oh wait, `.yeb` easyconfigs aren't going to pass a PEP8 test... :)

so, leave as is

boegel:
lgtm, thanks for all the effort on this @wpoely86!

wpoely86:
Easyconfig use is in hpcugent/easybuild-easyconfigs#2506

Issue #1606, opened 2016-02-11
--minimal-toolchains seems to don't work with ToolchainMNS


damianam:
Apparently ToolchainMNS needs an easyconfig file for each dependency, so --minimal-toolchains doesn't work, it crashes even before trying to use minimal toolchains. The attached log shows an debug example. @ocaisa can probably share more details.

[crash.zip](https://github.com/hpcugent/easybuild-framework/files/126694/crash.zip)

boegel:
Is this 'just' a matter of doing the mininal-toolchains-aware-robot-search earlier, i.e. when the module naming scheme needs to figure out the module name for the dependencies (and why does it even need to do that)?

ocaisa:
Should be solved with #1614

boegel:
a test that checks for this is added in #1619; it should pass when #1614 is merged in

boegel:
@damianam: ok to close?

damianam:
@boegel ok to close

boegel:
@damianam: then close it ;)

I prefer not to close issues opened by others (except when merging stuff, of course).

PR #1619, opened 2016-02-17
add end-to-end test for --minimal-toolchains

boegel:
This reproduces the issue in #1606.

Without #1614 included in `develop`, this test will fail.

cc @ocaisa, @damianam: please review?


boegel:
Test fails as it should:

```
ERROR [0.365s]: End-to-end test for --minimal-toolchains.
  File "test/framework/options.py", line 2104, in test_minimal_toolchains
    self.eb_main(args, do_build=True, raise_error=True, testing=False)
  File "test/framework/utilities.py", line 242, in eb_main
    raise myerr
EasyBuildError: "Failed to process easyconfig /var/tmp/eb-_sxVRh/eb-JRorAV/eb-YKAHI_/eb-rmtr9f/eb-6H7qEN/eb-iSOisD/eb-8K9O5o/eb-T4sxWA/test_minimal_toolchains.eb: Failed to find easyconfig file 'hwloc-1.6.2-gompi-1.4.10.eb' when determining module name for: {'external_module_metadata': {}, 'full_mod_name': None, 'build_only': False, 'versionsuffix': '', 'external_module': False, 'dummy': False, 'toolchain': {'version': '1.4.10', 'name': 'gompi'}, 'name': 'hwloc', 'short_mod_name': None, 'version': '1.6.2', 'hidden': False}"

Ran 387 tests in 366.827s

FAILED (errors=1)
```

boegel:
let's retest now that #1614 is merged...

Jenkins: test this please

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -162,7 +162,7 @@ def get_toolchain_hierarchy(parent_toolchain):
         for dep in parsed_ec['dependencies']:
             ec = robot_find_easyconfig(dep['name'], det_full_ec_version(dep))
             ec = process_easyconfig(ec, validate=False)[0]
-            alldeps = [ec['ec']['toolchain']] + ec['ec']['dependencies']
+            alldeps = [ec['ec']['toolchain']] + parsed_ec['dependencies']
```
boegel: @ocaisa: the fix in #1576 was not 100% correct, as I discovered when adding a test for it...

cc @damianam

test/framework/options.py
```diff
@@ -2070,6 +2070,43 @@ def test_set_tmpdir(self):
...
+        ectxt = '\n'.join([
+            "easyblock = 'ConfigureMake'",
+            "name = 'test'",
+            "version = '1.2.3'",
+            "homepage = 'http://example.com'",
+            "description = 'this is just a test'",
+            "toolchain = {'name': 'gompi', 'version': '1.4.10'}",
+            # hwloc-1.6.2-gompi-1.4.10.eb is *not* available, but hwloc-1.6.2-GCC-4.7.2.eb is,
+            # and GCC/4.7.2 is a subtoolchain of gompi/1.4.10
+            "dependencies = [('hwloc', '1.6.2')]",
```
boegel: TODO: also add SQLite as dep (available with *both* GCC and gompi...)

test/framework/robot.py
```diff
@@ -567,6 +567,14 @@ def test_get_toolchain_hierarchy(self):
             {'name': 'gompi', 'version': '1.4.10'},
```
ocaisa: gompi is a bit of a broken choice when you try to add dummy, because the underlying GCC doesn't actually have any dependencies on anything built with `dummy` which is (currently) required. I guess we should hard-include `dummy` when the relevant option is there

ocaisa: This would solve the first failing test

ocaisa: Wait, this problem is also the cause of the second test, it must be related to the change you made in the toolchain hierarchy generator

easybuild/framework/easyconfig/easyconfig.py
ocaisa: You sure this is correct? I think this was originally correct - what was the error? I think it should probably be
```
alldeps = [ec['ec']['toolchain']] + ec['ec']['dependencies']['toolchain']
```

ocaisa: Argh, I'm confused by this!

boegel: @ocaisa: the problem is that the previous line was taking the dependencies of the dependencies...

Say we're parsing `GCC-4.9.3-2.25.eb`; the dependencies are `GCCcore` and `binutils`, which are listed in `parsed_ec['dependencies']`

We also want to take into account the toolchain for each of these dependencies, so `[ec['ec']['toolchain']]`

We're adding the same list of deps over and over again now, so I need to change this to:

```
deps_tcs = [d['name'] for d in parsed_ec['dependencies']]
for dep in parsed_ec['dependencies']:
    # also add toolchains of each dep
```

or something like that (I'll look into it later)

ocaisa: Ok, but you're going backwards. The reason this was changed originally in #1576  was for iccifort built on GCCcore. iccifort depends on icc/ifort which then depend on GCCcore/binutils. Your proposal wouldn't cover this case. You need toolchains of deps *and* deps of deps *and* toolchains of deps of deps. Yes this is overkill but who cares, you've already parsed the deps anyway and we select only the toolchain we're interested in later and then make sure we have uniqueness.

test/framework/robot.py
boegel: It's the caching for toolchain hierarchy screwing up the tests... We'll need to reset the cache.

boegel: I don't think `gompi` is a broken choice, because we always include the *toolchain* too, not only the deps.

So, `dummy` is considered as a subtoolchain because `GCC` is built with `dummy`

easybuild/framework/easyconfig/easyconfig.py
boegel: OK, I'll add a testcase for `iccifort` and make sure that works too.

test/framework/robot.py
ocaisa: Agreed, my mistake

On 18 Feb 2016 21:06, "Kenneth Hoste" <notifications@github.com<mailto:notifications@github.com>> wrote:

In test/framework/robot.py<https://github.com/hpcugent/easybuild-framework/pull/1619#discussion_r53374477>:

> @@ -567,6 +567,14 @@ def test_get_toolchain_hierarchy(self):
>              {'name': 'gompi', 'version': '1.4.10'},


I don't think gompi is a broken choice, because we always include the toolchain too, not only the deps.

So, dummy is considered as a subtoolchain because GCC is built with dummy

—
Reply to this email directly or view it on GitHub<https://github.com/hpcugent/easybuild-framework/pull/1619/files#r53374477>.


------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------
Forschungszentrum Juelich GmbH
52425 Juelich
Sitz der Gesellschaft: Juelich
Eingetragen im Handelsregister des Amtsgerichts Dueren Nr. HR B 3498
Vorsitzender des Aufsichtsrats: MinDir Dr. Karl Eugen Huthmacher
Geschaeftsfuehrung: Prof. Dr.-Ing. Wolfgang Marquardt (Vorsitzender),
Karsten Beneke (stellv. Vorsitzender), Prof. Dr.-Ing. Harald Bolt,
Prof. Dr. Sebastian M. Schmidt
------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------

boegel:
@ocaisa: can you give this another look? Should be good to go.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -151,14 +156,14 @@ def get_toolchain_hierarchy(parent_toolchain):
...
-            raise EasyBuildError("Could not find easyconfig for %(name)s toolchain version %(version)s",
+            raise EasyBuildError("Could not find easyconfig for %s toolchain version %s",
                                  current_tc_name, current_tc_version)
 
         # parse the easyconfig
         parsed_ec = process_easyconfig(path, validate=False)[0]
 
         # search the toolchain+dependencies for the version of the subtoolchain
-        dep_tcs = []
+        dep_tcs = [d for d in parsed_ec['dependencies'] if d['name'] == subtoolchain_name]
```
ocaisa: I know this is a little pedantic since we have solved the particular problem with `icc/ifort` but...
Originally this function only dealt with toolchains from the depenencies, now it is dealing with dependencies *and* the toolchains of dependencies but these are not the same. Dependencies can have a `versionsuffix` which toolchains don't, so if we are adding dependencies we need to add them as
```
{'name': d['name'], ''version': '%s%s' % (d['version'], d['versionsuffix'])}
```
Same is true when adding dependencies below (basically need a repeat of this line). And I still think we should also be adding toolchains of deps of deps, since we've already parsed the deps so it comes at zero cost.

boegel: OK, makes sense to me, thanks for the feedback.

boegel: @ocaisa: fixed now, please rererererererererereview?

ocaisa:
lgtm 
If it passes the tests, I'm happy

boegel:
Going in, thanks for the excellent feedback @ocaisa!

PR #1623, opened 2016-02-18
New method to detect total available RAM on the building machine

valtandor:
Required by https://github.com/hpcugent/easybuild-easyblocks/pull/834


easybuild/tools/systemtools.py
```diff
@@ -105,6 +106,25 @@ def get_core_count():
     """NO LONGER SUPPORTED: use get_avail_core_count() instead"""
     _log.nosupport("get_core_count() is replaced by get_avail_core_count()", '2.0')
 
+def get_total_memory():
+
+    # Return total memory as an integer, a number of megabytes.
+
+    memtotal = 0
```
boegel: start with `None`?

easybuild/tools/systemtools.py
```diff
@@ -105,6 +106,25 @@ def get_core_count():
...
+    os_type = get_os_type()
+
+    if os_type == LINUX and os.path.exists(PROC_MEMINFO_FP):
+        with open(PROC_MEMINFO_FP, 'r') as meminfo:
+            for memline in meminfo:
+                memline_sub = re.sub(r'^MemTotal:\s*(\d+)\s*kB$', r'\1', memline)
+                if memline_sub != memline:
+                    memtotal = int(memline_sub) / 1024
+
+    if memtotal == 0:
```
boegel: `if memtotal is None`?

valtandor:
@boegel: Done.

boegel:
@valtandor: are you up for writing a simply unit test for this (in `test/framework/systemtools.py`)?

Just something that checks whether a non-zero memory size is returned on Linux, and raises the exception on Darwin?

If not, that's fine, I can help with that.

Where are you planning to use this?

Maybe also add it to `get_system_info`?

valtandor:
@boegel: This was used in the NWChem EasyBlock that I wrote. Part of the compilation involves trying to get the amount of available system memory. If `get_system_info` is a better place for it, I'm OK with that.

Where do the unit tests go?

boegel:
In `get_system_info`, you would just add a call to the function you wrote here, the function itself looks perfect.

Unit tests for functions in `systemtools.py` go in https://github.com/hpcugent/easybuild-framework/blob/master/test/framework/systemtools.py .

Take a look at the `test_glibc_version*` tests, you could write a group of tests for your version in a very similar way.

boegel:
Jenkins: ok to test

easybuild/tools/systemtools.py
```diff
@@ -105,6 +106,25 @@ def get_core_count():
     """NO LONGER SUPPORTED: use get_avail_core_count() instead"""
     _log.nosupport("get_core_count() is replaced by get_avail_core_count()", '2.0')
 
+def get_total_memory():
+
+    # Return total memory as an integer, a number of megabytes.
```
wpoely86: make this a docstring for the function:
```python
"""@returns Return total memory as an integer, a number of megabytes."""
```

easybuild/tools/systemtools.py
```diff
@@ -105,6 +106,25 @@ def get_core_count():
...
+
+    memtotal = None
+    os_type = get_os_type()
+
+    if os_type == LINUX and os.path.exists(PROC_MEMINFO_FP):
+        with open(PROC_MEMINFO_FP, 'r') as meminfo:
+            for memline in meminfo:
+                memline_sub = re.sub(r'^MemTotal:\s*(\d+)\s*kB$', r'\1', memline)
```
wpoely86: why work with a `sub`? Just look for the pattern and if it's match, extract the group?

boegel: yeah, this is a bit strange indeed

should be something like:

```python
res = re.search(r'^MemTotal:\s*(\d+)\s*kB$', memline)
if res:
    memtotal = int(res.group(1)) / 1024
```

but do test this, of course

boegel:
@valtandor: please take a look at https://github.com/valtandor/easybuild-framework/pull/2, which updates this PR a bit (implement support for Darwin + enhances tests)

easybuild/tools/systemtools.py
```diff
@@ -106,6 +107,28 @@ def get_core_count():
     _log.nosupport("get_core_count() is replaced by get_avail_core_count()", '2.0')
 
 
+def get_total_memory():
+    """
+    Try to ascertain this node's total memory
+    
+    @return: total memory as an integer, specifically a number of megabytes
```
fgeorgatos: If possible, it would be nice, if there was the ability to tune things from the outside about this, such as $BC_MEM_PER_NODE:
http://centers.hpc.mil/consolidated/bc/policies.php?choice=environment

The meaning of this is to be able to modify builds behavior dynamically, by adjusting such a variable.

easybuild/tools/systemtools.py
```diff
@@ -106,6 +107,28 @@ def get_core_count():
...
+    os_type = get_os_type()
+
+    if os_type == LINUX and os.path.exists(PROC_MEMINFO_FP):
+        with open(PROC_MEMINFO_FP, 'r') as meminfo:
+            for memline in meminfo:
+                mem_mo = re.match(r'MemTotal:\s*(\d+)\s*kB$', memline)
+                if mem_mo:
+                    memtotal = int(mem_mo.group(1)) / 1024
+
+    if memtotal is None:
```
fgeorgatos: Allowing $BC_MEM_PER_NODE would permit to overcome the limitation on missing or unmounted /proc/meminfo,
so that people that hit this conditional may pass it succesfully...

easybuild/tools/systemtools.py
boegel: If we do that, we should do it consistently (also for other system features), and it should be well documented which environment variables EB picks up on...

So, not in this PR.

boegel:
@valtandor ping on merging https://github.com/valtandor/easybuild-framework/pull/2 in here?

boegel:
Jenkins: test this please

boegel:
@valtandor broken test is fixed by https://github.com/valtandor/easybuild-framework/pull/3, let's get this merged...

boegel:
Going in, thanks @valtandor!

Issue #1584, opened 2016-01-29
Ideas to improve extended dry run


gppezzi:
I'm opening the issue just to avoid forgetting these ideas we've been discussing to improve -x:
- Dumping script to setup build environment (exports, modules loads)
- or going further: dumping a shell script that actually performs the whole build
- Pointers to where stuff happens (easyblocks, framework)

boegel:
Your first bullet point is being implemented in $1624.

As I briefly mentioned there, your 2nd bullet point is hard to do in practice I think, especially w.r.t. making sure the script is complete (i.e. covers all things done in the easyblock).

gppezzi:
:+1:

gppezzi:
As for the second point, if the framework can provide something that helps to debug, even if it is not complete, it will be particularly useful for new users. 

The most common complain I hear when people move to EB is "it works when I do it manually but not when done by EB". I think that this dumped script might help people to spot the differences between EB & the manual build.

I know that everything is on the log file, but new users are usually overwhelmed by all the output they see there.

If doing the script as a dry-run is complicated, how about doing it during the actual build?

boegel:
The main issue is that just the commands that are executed are not sufficient...

There's also other things going on in the Python code (creating files, moving to a different directory, copying files, ...) that would need to be translated to the bash equivalent.

It's not impossible, but a lot of work, and hard to get it 100% correct.

For some things, it may matter whether you're doing a dry run or not, but that's an orthogonal issue.

But, we can work on it, sure.

The first step would be to get the output of `--extended-dry-run` to be a valid bash script. It's still going to be incomplete though.

PR #1624, opened 2016-02-19
add support for --dump-env-script

boegel:
Example usage:

```
$ eb --dump-env-script GCC-4.9.2.eb OpenMPI-1.6.4-GCC-4.7.2.eb FFTW-3.3.3-gompi-1.4.10.eb
== temporary log file in case of crash /tmp/eb-yDmmIN/easybuild-ZFlM8G.log
Script to set up build environment for /example/GCC-4.9.2.eb dumped to GCC-4.9.2.env
Script to set up build environment for /example/OpenMPI-1.6.4-GCC-4.7.2.eb dumped to OpenMPI-1.6.4-GCC-4.7.2.env
Script to set up build environment for /example/FFTW-3.3.3-gompi-1.4.10.eb dumped to FFTW-3.3.3-gompi-1.4.10.env
== Temporary log file(s) /tmp/eb-yDmmIN/easybuild-ZFlM8G.log* have been removed.
== Temporary directory /tmp/eb-yDmmIN has been removed.
```

```
$ eb --dump-env-script GCC-4.9.2.eb                                                      
== temporary log file in case of crash /tmp/eb-_x72Fd/easybuild-ZvClAa.log
ERROR: One or more scripts already exists, not overwriting them (unless forced): GCC-4.9.2.env

$ eb --dump-env-script GCC-4.9.2.eb --force
== temporary log file in case of crash /tmp/eb-C4mBUJ/easybuild-_tFgst.log
Script to set up build environment for /example/GCC-4.9.2.eb dumped to GCC-4.9.2.env
== Temporary log file(s) /tmp/eb-C4mBUJ/easybuild-_tFgst.log* have been removed.
== Temporary directory /tmp/eb-C4mBUJ has been removed.
```

```
$ cat OpenMPI-1.6.4-GCC-4.7.2.env
#!/bin/bash
# script to set up build environment as defined by EasyBuild v2.7.0.dev0 for OpenMPI-1.6.4-GCC-4.7.2.eb
# usage: source OpenMPI-1.6.4-GCC-4.7.2.env

# toolchain & dependency modules
module load GCC/4.7.2
module load hwloc/1.6.2-GCC-4.7.2

# build environment
export CC='gcc'
export CFLAGS='-O2'
export CPPFLAGS=''
export CXX='g++'
export CXXFLAGS='-O2'
export F77='gfortran'
export F90='gfortran'
export F90FLAGS='-O2'
export FC='gfortran'
export FCFLAGS='-O2'
export FFLAGS='-O2'
export FLIBS='-lgfortran'
export LDFLAGS=''
export LIBS='-lm -lpthread'
export OPTFLAGS='-O2'
export PRECFLAGS=''
```


boegel:
@molden: is this what you had in mind?

@wpoely86: please review?

molden:
So this will only set the environment if I'm not mistaken. What about a 'dump-installation-script'? Will this also be included?

boegel:
@molden: which would do what exactly?

The procedure as spit out by extended dry run is always going to be incomplete, that's why we have easyblocks...

With some effort, you may be able to get it to work for 80% of the installations, but making it perfect... You might as well rewrite everything in bash imho.

molden:
@boegel : true, I maybe misunderstood that the extended dry run function would be rewritten in script format.

boegel:
@molden: well, that's doable, but what would you be using that for? You'd still have to go and check the easyblock (potentially) to verify that all actions are actually reflected in the output of extended dry run.

Making that airtight is very hard, I think.

molden:
@boegel I see, maybe these issues should be discussed here: #1584

easybuild/framework/easyblock.py
```diff
@@ -1510,7 +1510,7 @@ def patch_step(self, beginpath=None):
             if not apply_patch(patch['path'], src, copy=copy_patch, level=level):
                 raise EasyBuildError("Applying patch %s failed", patch['name'])
 
-    def prepare_step(self):
+    def prepare_step(self, start_dir=True):
```
wpoely86: doc args

easybuild/framework/easyconfig/tools.py
```diff
@@ -465,3 +467,61 @@ def review_pr(pr, colored=True, branch='develop'):
             lines.extend(['', "(no related easyconfigs found for %s)\n" % os.path.basename(ec['spec'])])
 
     return '\n'.join(lines)
+
+
+def dump_env_script(easyconfigs):
+    """
+    Dump source scripts that set up build environment for specified easyconfigs."""
```
wpoely86: doc args

easybuild/framework/easyconfig/tools.py
```diff
@@ -465,3 +467,61 @@ def review_pr(pr, colored=True, branch='develop'):
...
+        script_path = '%s.env' % os.path.splitext(os.path.basename(easyconfig['spec']))[0]
+        ecs_and_script_paths.append((easyconfig['ec'], script_path))
+
+    # don't just overwrite existing scripts
+    existing_scripts = [s for (_, s) in ecs_and_script_paths if os.path.exists(s)]
+    if existing_scripts:
+        if build_option('force'):
+            _log.info("Found existing scripts, overwriting them: %s", ' '.join(existing_scripts))
+        else:
+            raise EasyBuildError("One or more scripts already exists, not overwriting them (unless forced): %s",
```
wpoely86: `unless --force is used`

easybuild/framework/easyconfig/tools.py
```diff
@@ -465,3 +467,61 @@ def review_pr(pr, colored=True, branch='develop'):
...
 
     return '\n'.join(lines)
+
+
+def dump_env_script(easyconfigs):
+    """
+    Dump source scripts that set up build environment for specified easyconfigs."""
+    ecs_and_script_paths = []
+    for easyconfig in easyconfigs:
+        script_path = '%s.env' % os.path.splitext(os.path.basename(easyconfig['spec']))[0]
```
wpoely86: `.env.txt` ?

easybuild/framework/easyconfig/tools.py
```diff
@@ -465,3 +467,61 @@ def review_pr(pr, colored=True, branch='develop'):
...
+        app.dry_run = app.silent = app.toolchain.dry_run = True
+
+        # prepare build environment (in dry run mode)
+        app.check_readiness_step()
+        app.prepare_step(start_dir=False)
+
+        # compose script
+        ecfile = os.path.basename(ec.path)
+        script_lines = [
+            "#!/bin/bash",
```
wpoely86: this header is not needed it it's meant as sourcing script.

easybuild/framework/easyconfig/tools.py
```diff
@@ -465,3 +467,61 @@ def review_pr(pr, colored=True, branch='develop'):
...
+            script_lines.append("# (no modules loaded)")
+
+        script_lines.extend(['', "# build environment"])
+        if app.toolchain.vars:
+            env_vars = sorted(app.toolchain.vars.items())
+            script_lines.extend(["export %s='%s'" % (var, val.replace("'", "\\'")) for (var, val) in env_vars])
+        else:
+            script_lines.append("# (no build environment defined)")
+
+        write_file(script_path, '\n'.join(script_lines))
```
wpoely86: are all the env vars set in the easyblock also part of this?

easybuild/framework/easyconfig/tools.py
boegel: not sure about this...

it's a script, so `.txt` doesn't make sense to me, but I also don't want to use `.sh` to make it clear it should be *sourced*

is there a standard for something like this?

easybuild/framework/easyconfig/tools.py
boegel: it's not, but I included it anyway as a kind of self-documentation (to make clear it's a bash script)

easybuild/framework/easyconfig/tools.py
boegel: only if they are set in `prepare_step`, as it is right now...

Maybe this can be extended to work together with `--stop`...

boegel:
Jenkins: test this please

boegel:
I feel this is good to go, it can be enhanced/extended later (e.g. to cooperate with `--stop`), thanks for the feedback @wpoely86 and @molden!

PR #1625, opened 2016-02-19
fix bug triggered with use of --stop + add test that triggers it

boegel:
fix for bug introduced in #1614...


boegel:
considered trivial enough to go in as is

PR #1630, opened 2016-02-24
enhance ModulesTool.exist to also recognize partial module names

boegel:
required to be able to list partial module names as dependency when using external modules (cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/2554)

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -358,6 +358,18 @@ class ModulesTool(object):
         """
         Check if modules with specified names exists.
         """
+        def mod_exists_via_show(mod_name, partial=False):
+            """Helper function to check whether specified module name exists through 'module show'."""
+            if partial:
+                mod_exists_regex = mod_exists_regex_template % ('%s.*' % re.escape(mod_name))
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86: please review?

easybuild/tools/modules.py
```diff
@@ -354,10 +354,25 @@ def available(self, mod_name=None, extra_args=None):
...
         """
+        def mod_exists_via_show(mod_name):
+            """
+            Helper function to check whether specified module name exists through 'module show'.
+
+            @param mod_name: module name
+            """
+            mod_exists_regex = mod_exists_regex_template % re.escape(mod_name)
+            txt = self.show(mod_name)
+            mod_exists_regex = re.compile(mod_exists_regex, re.M)
```
wpoely86: why the compile? That has no purpose here?

boegel: yeah, I guess it doesn't, will change

boegel:
going in, thanks for the review @wpoely86!

PR #1631, opened 2016-02-25
Better error message for toolchain definition errors

pforai:
Tripped a number of times over this error and it was never clear what rhs and lhs were.


boegel:
lgtm, thanks @pforai, will merge when Jenkins agrees

boegel:
Going in, thanks @pforai!

PR #1632, opened 2016-02-25
make default is_short_modname_for check less strict to support versionless external modules as deps

boegel:
required for https://github.com/hpcugent/easybuild-easyconfigs/pull/2554


gppezzi:
I confirm that this new regex enables us to use versionless PrgEnv modules. 

Thanks @boegel

wpoely86:
lgtm

boegel:
Going in, thanks for the review/test @gppezzi and @wpoely86!

PR #1633, opened 2016-02-25
add function to find FlexLM licenses

boegel:
cfr. discussion in https://github.com/hpcugent/easybuild-easyblocks/pull/837

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1171,3 +1171,72 @@ def det_size(path):
         _log.warn("Could not determine install size: %s" % err)
 
     return installsize
+
+
+def find_flexlm_license(custom_env_vars=None, lic_specs=None):
+    """
+    Find FlexLM license.
+
+    Considered specified list of environment variables;
+    checks for path to existing license file or valid license server specification.
+
+    If no license if found through environment variables, also consider 'lic'.
+
+    @param custom_env_vars: list of environment variables to considered (if None, only consider $LM_LICENSE_FILE)
+    @param lic_specs: list of license specifications
+    @return: tuple with list of valid license specs found and name of first valid environment variable
+    """
+    valid_lic_specs = []
+    lic_env_var = None
+
+    # regex for license server spec; format: <port>@<server>
+    server_port_regex = re.compile(r'^[0-9]+@\S+$')
+
+    # always consider $LM_LICENSE_FILE
+    default_lic_env_var = 'LM_LICENSE_FILE'
+    if custom_env_vars is None:
+        lic_env_vars = [default_lic_env_var]
+    else:
+        lic_env_vars = custom_env_vars + [default_lic_env_var]
+
+    # grab values for defined environment variables
+    cand_lic_specs = {}
+    for env_var in lic_env_vars:
+        if env_var in os.environ:
+            cand_lic_specs[env_var] = os.environ[env_var].split(os.pathsep)
+
+    # also consider provided license spec (last)
+    if lic_specs:
+        cand_lic_specs[None] = lic_specs
+
+    _log.debug("Candidate license specs: %s", cand_lic_specs)
+
+    # check for valid license specs
+    # order matters, so loop over original list of environment variables to consider
+    for env_var in lic_env_vars + [None]:
+        values = cand_lic_specs.get(env_var, None) or []
+        _log.info("Considering $%s to find FlexLM license specs: %s", env_var, values)
+
+        for value in values:
+            if server_port_regex.match(value) or os.path.isfile(value):
+                valid_lic_specs.append(value)
+
+            elif os.path.isdir(value):
+                # consider all *.dat and *.lic files in specified directory
+                lic_files = glob.glob(os.path.join(value, '*.dat')) + glob.glob(os.path.join(value, '*.lic'))
+                if lic_files:
+                    valid_lic_specs = sorted(lic_files)
+
+        # stop after finding valid license specs
+        if valid_lic_specs:
+            lic_env_var = env_var
+            break
+
+    if lic_env_var:
+        via_msg = '$%s' % lic_env_var
+    else:
+        via_msg = "provided license spec"
+
+    _log.info("Found valid license specs via %s: %s", via_msg, valid_lic_specs)
+
+    return (valid_lic_specs, lic_env_var)
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1193,13 +1193,11 @@ def find_flexlm_license(custom_env_vars=None, lic_specs=None):
     server_port_regex = re.compile(r'^[0-9]+@\S+$')
 
     # always consider $LM_LICENSE_FILE
-    default_lic_env_var = 'LM_LICENSE_FILE'
-    if custom_env_vars is None:
-        lic_env_vars = [default_lic_env_var]
-    elif isinstance(custom_env_vars, basestring):
-        lic_env_vars = [custom_env_vars, default_lic_env_var]
-    else:
-        lic_env_vars = custom_env_vars + [default_lic_env_var]
+    lic_env_vars = ['LM_LICENSE_FILE']
+    if isinstance(custom_env_vars, basestring):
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86: please review?

@damianam, @ocaisa: thoughts?

ocaisa:
Great, this should help simplify other easyblocks too

easybuild/tools/filetools.py
```diff
@@ -1171,3 +1171,72 @@ def det_size(path):
...
+
+
+def find_flexlm_license(custom_env_vars=None, lic_specs=None):
+    """
+    Find FlexLM license.
+
+    Considered specified list of environment variables;
+    checks for path to existing license file or valid license server specification.
+
+    If no license if found through environment variables, also consider 'lic'.
```
damianam: Small typo ("If no license is")

easybuild/tools/filetools.py
```diff
@@ -1171,3 +1171,72 @@ def det_size(path):
...
+
+    # regex for license server spec; format: <port>@<server>
+    server_port_regex = re.compile(r'^[0-9]+@\S+$')
+
+    # always consider $LM_LICENSE_FILE
+    default_lic_env_var = 'LM_LICENSE_FILE'
+    if custom_env_vars is None:
+        lic_env_vars = [default_lic_env_var]
+    else:
+        lic_env_vars = custom_env_vars + [default_lic_env_var]
```
damianam: My Python-foo is weak, but wouldn't this fail if in the easyblock somebody passes `custom_env_vars='something'`, instead of `custom_env_vars=['something']` ?

easybuild/tools/filetools.py
```diff
@@ -1171,3 +1171,72 @@ def det_size(path):
...
+
+    # also consider provided license spec (last)
+    if lic_specs:
+        cand_lic_specs[None] = lic_specs
+
+    _log.debug("Candidate license specs: %s", cand_lic_specs)
+
+    # check for valid license specs
+    # order matters, so loop over original list of environment variables to consider
+    for env_var in lic_env_vars + [None]:
```
damianam: If I understood this right, the provided license in the easyconfig will be the last one to consider. I would assume that manually provided licenses should have precedence.

easybuild/tools/filetools.py
boegel: it would, yes

but then they didn't check the docstring of the function... ;)

I'll add this, so you can pass both a string and a list:

```python
if isinstance(custom_env_vars, basestring):
    custom_env_vars = [custom_env_vars]
```

easybuild/tools/filetools.py
boegel: Yes, that's the way it is now too in `IntelBase` (and in the code you copy-pasted in the PGI easyblock in https://github.com/hpcugent/easybuild-easyblocks/pull/837).

This is the order in `IntelBase` (and also in this function)

* `$INTEL_LICENSE_FILE` (`custom_env_vars`)
* `$LM_LICENSE_FILE` (`default_lic_env_var`)
* `license_file` (`lic_specs`)

Note that this is very similar to how EasyBuild configuration options are handled: `$EASYBUILD_*` environment variables win over entries in EasyBuild `*.cfg` configuration files.

For these license environment variables, this makes a lot of sense imho, since they're usually defined anyway already.

I guess it all depends on what you mean with 'manually provided' here...

easybuild/tools/filetools.py
```diff
@@ -1171,3 +1171,74 @@ def det_size(path):
...
+    @return: tuple with list of valid license specs found and name of first valid environment variable
+    """
+    valid_lic_specs = []
+    lic_env_var = None
+
+    # regex for license server spec; format: <port>@<server>
+    server_port_regex = re.compile(r'^[0-9]+@\S+$')
+
+    # always consider $LM_LICENSE_FILE
+    default_lic_env_var = 'LM_LICENSE_FILE'
```
wpoely86: why not put `lic_env_vars` equal to `LM_LICENSE_FILE` and prepend?

easybuild/tools/filetools.py
```diff
@@ -1171,3 +1171,74 @@ def det_size(path):
...
+
+    # grab values for defined environment variables
+    cand_lic_specs = {}
+    for env_var in lic_env_vars:
+        if env_var in os.environ:
+            cand_lic_specs[env_var] = os.environ[env_var].split(os.pathsep)
+
+    # also consider provided license spec (last)
+    if lic_specs:
+        cand_lic_specs[None] = lic_specs
```
wpoely86: `None`?

easybuild/tools/filetools.py
```diff
@@ -1171,3 +1171,74 @@ def det_size(path):
...
+        _log.info("Considering %s to find FlexLM license specs: %s", env_var, values)
+
+        for value in values:
+            if server_port_regex.match(value) or os.path.isfile(value):
+                valid_lic_specs.append(value)
+
+            elif os.path.isdir(value):
+                # consider all *.dat and *.lic files in specified directory
+                lic_files = glob.glob(os.path.join(value, '*.dat')) + glob.glob(os.path.join(value, '*.lic'))
+                if lic_files:
```
wpoely86: can we read the file?

easybuild/tools/filetools.py
damianam: What I meant is that something set explicitly in an easyconfig should have precedence over other options that might go unnoticed. Imagine that you want to install a beta of one tool, with an special temporal license. You will specify that license in the easyconfig, but you might not want to unset `LM_LICENSE_FILE`. In any case, it is a minor thing.

boegel: Note that the `IntelBase` easyblock will define `$INTEL_LICENSE_FILE` according to whatever license was used at installation time...

I think you can argue both ways for this, and since we already have it in place with `$INTEL_LICENSE_FILE` being preferred over `$LM_LICENSE_FILE` being preferred over `license_file`...

damianam:
@boegel do we have an ETA for this PR? It is needed for this one: https://github.com/hpcugent/easybuild-easyblocks/pull/837

boegel:
@damianam: I need to tackle @wpoely86's last remarks... but I think you can pretty much assume the function is going to be left unchanged in terms of how it'll be used

I tackle this ASAP.

easybuild/tools/filetools.py
boegel: ok

easybuild/tools/filetools.py
boegel: well, I need a key that doesn't clash with any of the provided env vars...

this way, I'm sure there is no clash, *and* I can detect how the retained license was obtained (through environment or not)

since the keys in `cand_lic_specs` are the name of the environment variable they were obtained with, it's also semantically correct

boegel: I'll clarify with a comment in the code

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins PR builder plugin was updated...

Jenkins: test this please

boegel:
Going in, thanks for the feedback everyone!

easybuild/tools/filetools.py
```diff
@@ -1171,3 +1171,87 @@ def det_size(path):
...
+    valid_lic_specs = []
+    for env_var in lic_env_vars + [None]:
+        values = cand_lic_specs.get(env_var, None) or []
+        _log.info("Considering %s to find FlexLM license specs: %s", env_var, values)
+
+        for value in values:
+
+            # license files to consider
+            lic_files = None
+            if os.path.isfile(value):
```
damianam: The function will fail with something like:

```
    if os.path.isfile(value):
  File "/usr/lib64/python2.7/genericpath.py", line 29, in isfile
    st = os.stat(path)
TypeError: coercing to Unicode: need string or buffer, NoneType found
```

if none of the expected environment variables are not defined AND the easyconfig has no `license_file` defined.

The problem is that `values` becomes `[None]`, then `value` is `None`, and `os.path.isfile` breaks.

PR #1634, opened 2016-02-26
fix availability check for external modules with partial module name

boegel:
required to make `--robot --from-pr` work on https://github.com/hpcugent/easybuild-easyconfigs/pull/2554

without this, you run into this:

```
kehoste@daint102:~> eb --from-pr 2554 -dfr
== temporary log file in case of crash /tmp/eb-SCGaNv/easybuild-nBdPrL.log
== resolving dependencies ...
ERROR: Missing modules for one or more dependencies marked as external modules: ['PrgEnv-gnu', 'PrgEnv-gnu', 'PrgEnv-cray', 'PrgEnv-cray', 'PrgEnv-intel', 'PrgEnv-intel']
```

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1364,7 +1364,7 @@ class EasyBlock(object):
         # - if a current module can be found, skip is ok
         # -- this is potentially very dangerous
         if self.cfg['skip']:
-            if self.modules_tool.exist([self.full_mod_name])[0]:
+            if self.modules_tool.exist([self.full_mod_name], skip_avail=True)[0]:
                 self.skip = True
                 self.log.info("Module %s found." % self.full_mod_name)
                 self.log.info("Going to skip actual main build and potential existing extensions. Expert only.")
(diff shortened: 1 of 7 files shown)


boegel:
cc @pforai, @gppezzi

wpoely86:
So the check is now independent whether it's hidden or not? lgtm

boegel:
@wpoely86: yes, it basically fall backs to checking with `module show` in case the exact module name is not listed in the output of `module avail`

This occurs for hidden modules (e.g. `.foo/1.2.3`) and external modules with a partial module name (e.g. checking if `PrgEnv-gnu` is available, while `avail` spits out `PrgEnv-gnu/5.2.40`).

There's no harm in always falling back to `show`, except for maybe making things slower (`modtool.exist` will also check via `avail` first).

Maybe I should add an extra named argument to `modtool.exist` to specify that it should skip the `avail` check?

boegel:
@wpoely86: done, please rereview? Using `skip_avail=True` makes sense anywhere where only a single module name is provided (`show` is likely faster than `avail` in the case)

gppezzi:
I confirm that this PR is working (see reports on https://github.com/hpcugent/easybuild-easyconfigs/pull/2554)

thx @boegel

wpoely86:
:+1:

boegel:
Going in, thanks for the review @wpoely86 and feedback @gppezzi!

PR #1637, opened 2016-02-26
support providing additional relative path for prefix in external module metadata

boegel:
add support for entries like:

```
[fftw/3.3.4.3]
name = FFTW
version = 3.3.4.3
prefix = FFTW_INC/..
```

which results in `$EBROOTFFTW` being defined as the equivalent of `$FFTW_INC/../` (if `$FFTW_INC` is defined)

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -402,13 +402,23 @@ class Toolchain(object):
         # define $EBROOT env var for install prefix, picked up by get_software_root
         prefix = metadata.get('prefix')
         if prefix is not None:
-            if prefix in os.environ:
-                val = os.environ[prefix]
-                self.log.debug("Using value of $%s as prefix for software named %s: %s", prefix, name, val)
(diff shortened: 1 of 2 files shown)


boegel:
@gppezzi, @pforai, @wpoely86: review/feedback?

gppezzi:
Ideally Cray should fix that on their module files.  But maybe they consider not setting a variable for fftw root as a feature, so it's good to be able to workaround that.

thx @boegel

boegel:
Well, I'm guessing they wouldn't be willing to change existing module files...

And this is a useful feature regardless.

wpoely86:
looks good but maybe name it `root` instead of `prefix` to be consistent?

boegel:
@wpoely86: well, `root` is confusing (could mean the `root` user). You could argue the same for `$EBROOT...` of course, but there it's kind of clear from the context, I hope.

Also, we can't just rename that now, the support for providing metadata for external modules is not experimental by itself, that's a stable feature.

boegel:

PR #1638, opened 2016-02-27
add external modules metadata for Cray systems

boegel:
replaces #1586 and #1596

requires ~~#1637~~ for the provided metadata to be handled correctly (cfr. the metadata for `fftw`)

cc @pforai, @gppezzi


gppezzi:
looks good!

Now that we are arriving into final details: how much effort it would require to support versionless entries on the metadata file? that would avoid lots of copy&pasting when supporting other PE versions

pforai:
LGTM!

boegel:
@gppezzi: how would a versionless entry look like?

gppezzi:
how about:
[fftw]
 name = FFTW
 prefix = FFTW_INC/..

the metadata file would be much easier to maintain this way and the version would be needed only for exceptions (or when cray changes things for new versions)

boegel:
@gppezzi: we could take that one step further, and support something like:

```ini
[fftw/%(version)s]
name = FFTW
version = %(version)s
prefix = FFTW_INC/..
```

Just to also have the version info available (which is sometimes used in easyblocks for version checks on the dependencies).

pforai:
that looks pretty nice!

gppezzi:
@boegel can this work in a way that if you use explicit version it would override the entry with %(version)s?

boegel:
@gppezzi: it would have to, yes. But, I'd say we look into that in a future PR, since it doesn't add any extra feature or fixes an important bug. Let's focus on the other things required for stable Cray support first... So: please open an issue with this as a feature request.

gppezzi:
Issue opened https://github.com/hpcugent/easybuild-framework/issues/1639

As for this PR, I don' t have any other remarks. If this MD file is loaded by default then people should be able to easily try out the Cray TCs.

easybuild/tools/options.py
```diff
@@ -1107,6 +1070,59 @@ def process_software_build_specs(options):
...
+            _log.info("No default external modules metadata found")
+
+    # leave external_modules_metadata untouched if no files are provided
+    if not cfgs:
+        _log.debug("No metadata provided for external modules.")
+        return {}
+
+    parsed_metadata = ConfigObj()
+    for cfg in cfgs:
+        if os.path.exists(cfg):
```
wpoely86: `isfile`?

wpoely86:
:+1:

easybuild/tools/options.py
boegel: ok, thanks

boegel:
Going in, thanks for the review @wpoely86, and @pforai and @gppezzi for helping out with this!

Issue #1636, opened 2016-02-26
Different behaviour on second build if the same easyconfig file is passed twice to the eb command line


gppezzi:
I pasted twice on the command line (by mistake) the list of ebs that I wanted to build  and I've noticed that the second build of the same package was failing (while the first one passed).

I'm giving here the logs of a failing build and a working one for debugging, but my feeling is that this could be fixed by just filtering the list to remove duplicates (or do you have an use case where it make sense to build stuff twice on the same command line?).

* Failed build: https://gist.github.com/gppezzi/eee4c9799dda57b9c87c
* Build OK: https://gist.github.com/gppezzi/ed7f602b71cae48f13fa

Here's the command line I've used to trigger this behaviour.
```
eb HPL-2.1-CrayCCE-2015.06-XC.eb HPL-2.1-CrayCCE-2015.11-XC.eb HPL-2.1-CrayGNU-2015.06-XC.eb HPL-2.1-CrayGNU-2015.11-XC.eb HPL-2.1-CrayIntel-2015.06-XC.eb HPL-2.1-CrayIntel-2015.11-XC.eb HPL-2.1-CrayCCE-2015.06-XC.eb HPL-2.1-CrayCCE-2015.11-XC.eb HPL-2.1-CrayGNU-2015.06-XC.eb HPL-2.1-CrayGNU-2015.11-XC.eb HPL-2.1-CrayIntel-2015.06-XC.eb
```

boegel:
We could simply filter the provided list of easyconfigs for unique values, but the fact that the 2nd build failed worries me.

We do quite a bit of effort to try and ensure multiple builds done in the same session are isolated from each other. This suggests we missed something...

boegel:
Filtered log shows that `$CFLAGS` is incorrectly defined in the 2nd round...

For some reason, the `-ftz` & co flags stick around.
This is a bug in the toolchain mechanism somewhere... :(

```
EB_HPL INFO building and installing HPL/2.1-CrayCCE-2015.06-XC...
environment INFO Environment variable CFLAGS set to -craype-verbose -O2 (previously undefined)
environment INFO Environment variable CFLAGS set to -craype-verbose -O2 (previously undefined)
environment INFO Environment variable CFLAGS set to -craype-verbose -O2 (previously undefined)
easyblock INFO COMPLETED: Installation ended successfully
EB_HPL INFO building and installing HPL/2.1-CrayCCE-2015.11-XC...
[...]

PR #1641, opened 2016-02-28
fix independency of Cray toolchains w.r.t. toolchain build environment

boegel:
fix for bug reported in #1636 by @gppezzi


boegel:
@pforai: please review?

pforai:
Good catch aktuelly! Merge now! :)

boegel:
I'm looking into a cleaner fix, the `deepcopy` feels to much like dancing around the problem...

boegel:
Hmm, nevermind, the alternate solution I have in mind is way worse... :D

Going in, thanks for the review @pforai!

PR #1646, opened 2016-02-29
No need to care about multiple versions of dummy

ocaisa:
...and also need to handle case where dependency easyconfig does not exist

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -209,7 +209,7 @@ def get_toolchain_hierarchy(parent_toolchain):
         else:
             if subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
                 # Don't care about multiple versions of dummy, take whatever comes first
-                _log.info("Ignoring multiple versions of %s in hierarchy", DUMMY_TOOLCHAIN_NAME)
+                _log.info("Ignoring multiple versions of %s in toolchain hierarchy", DUMMY_TOOLCHAIN_NAME)
                 subtoolchain_version = dep_tcs[0]['version']
             else:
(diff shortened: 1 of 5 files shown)


boegel:
@ocaisa: is it worth enhancing the tests for the multiple-dummy-versions issue?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -204,8 +207,13 @@ def get_toolchain_hierarchy(parent_toolchain):
                 raise EasyBuildError("No version found for subtoolchain %s in dependencies of %s",
                                      subtoolchain_name, current_tc_name)
         else:
-            raise EasyBuildError("Multiple versions of %s found in dependencies of toolchain %s: %s",
-                                 subtoolchain_name, current_tc_name, uniq_subtc_versions)
+            if subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
+                # Don't care about multiple versions of dummy, take whatever comes first
+                _log.info("Ignoring multiple versions of %s in toolchain hierarchy", DUMMY_TOOLCHAIN_NAME)
+                subtoolchain_version = dep_tcs[0]['version']
```
boegel: hmm, rather than leaving this to chance, wouldn't it be better to use `''` as version? we know that using `dummy` as version has implications (i.e. deps not being loaded at build time)

ocaisa: Can do but it actually has no impact, it's only used for resolving the easyconfig file name which in the case of dummy doesn't matter

boegel:
lgtm, ok to go in when Jenkins gives green light

boegel:
Going in, thanks @ocaisa!

PR #1648, opened 2016-03-03
avoid that fancylogger tries to import mpi4py to determine MPI rank

boegel:


boegel:
@stdweird: there's no other way right (and this is quite OK imho)

cc @pforai, @gppezzi

pforai:
It has to happen before importing as far as I understand it and that is quite a simple solution.

gppezzi:
:+1:

stdweird:
not an easy way, no. in  principle, it should only be set on system with broken mpi4py, but i guess there's no harm in disabling it like this for EB.

boegel:
I agree @stdweird, no harm done with disabling this for EB.

Going in, thanks for the feedback everyone!

PR #1651, opened 2016-03-03
Update develop install script

besserox:
I'm not sure that someone except me actually uses this script. 
However, I believe it is still good to keep it updated.


boegel:
Jenkins: ok to test

boegel:
@besserox: I agree, makes sense to update it, thanks for looking into this!

PR #1652, opened 2016-03-04
handle allowed system deps during prepare_step rather than during parsing of easyconfig

boegel:
fix for bug reported by @ocaisa 

If handling of system deps happens too early, it causes problems.

@ocaisa: can you detail on the issue you ran into?


ocaisa:
I was trying out hpcugent/easybuild-easyconfigs#2621 and I ran into an issue with HMNS. Deep in it's list of dependencies there is a build dependency on `xcb-proto`, which inside has a system dependency for Python. For a hierarchical naming scheme the robot is the default so the easyconfig of xcb-proto will always get parsed, even though it is not part of the build process for Python. `EBROOTPYTHON` was being set as the file was parsed, which was causing the build to stop hard because it thought the module was loaded.

boegel:
@ocaisa: can you retry that on top of this PR? It should fix the problem.

ocaisa:
Fix works, thanks @boegel !

boegel:
Great, thanks for the confirmation @ocaisa, going in!

PR #1656, opened 2016-03-05
fix issue in bootstrap script with vsc-base being picked up from the OS

boegel:
Fix for an issue that only occurs if `vsc-base` is already installed system-wide (reported by @ehiggs).

This replicates the part of https://github.com/hpcugent/easybuild-easyblocks/pull/813 where `vsc-base` is installed again, in stage 1 of the bootstrap procedure (where we're installing EasyBuild in a temporary location, so not using `eb` yet).


boegel:
@wpoely86: please review?

wpoely86:
:+1:

boegel:

ehiggs:
Thanks for the fix @boegel!

PR #1667, opened 2016-03-12
add sanity check in easybuild/__init__.py w.r.t. current working dir

boegel:
With #1593 included, running `eb` from the `easybuild-framework` cloned repo working directory fails with errors like:

```
import_available_modules: Failed to import easybuild.toolchains.cgmpich: No module named cgmpich
```

This seems to occur when the path to `easybuild-framework` and `easybuild-easyblocks` is listed in `$PYTHONPATH`, *and* you're running `eb` from either of those locations.

Since the current directory is always part of the Python search path, this basically leads to having the same path twice in the Python search path, under a different 'name' (once absolute, once relative), which seems to be screwing up things.

With a simple check in `easybuild/__init__.py`, we can exit early and explain that running `eb` from this directories doesn't work; if `eb` is being outside of these directory, `__path__` will contain an absolute path rather than only `easybuild`.

### DIFF
--- a/easybuild/__init__.py
+++ b/easybuild/__init__.py
@@ -28,5 +28,14 @@ Declares EasyBuild namespace, in an extendable way.
 @author: Jens Timmerman (Ghent University)
 @author: Kenneth Hoste (Ghent University)
 """
+import os
 import pkg_resources
+import sys
+
+# check whether EasyBuild is being run from a directory that contains easybuild/__init__.py;
+# that doesn't work (fails with import errors), due to weirdness to Python packaging/setuptools/namespaces
+if __path__[0] == 'easybuild':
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86: please review?

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

wpoely86:
:+1:

boegel:
I briefly tried to fix the actual problem, but couldn't get it to work...

There a problem with using `__import__`, it doesn't seem to be coping well with packages/`__init__.py`s that use `declare_namespace`.

Not a big issue for now imho, so let's merge this and see if it sufficiently annoys us in the future to spend time on fixing this...

boegel:

PR #1669, opened 2016-03-16
bump version to 2.7.0 and update release notes

boegel:


wpoely86:
:+1:

RELEASE_NOTES
```diff
@@ -3,6 +3,64 @@ For more detailed information, please see the git log.
 
 These release notes can also be consulted at http://easybuild.readthedocs.org/en/latest/Release_notes.html.
 
+v2.7.0 (March 18th 2016)
+------------------------
+
+feature + bugfix release
+- stabilize Cray support
+    - enable 'dynamic' toolchain option by default for Cray* toolchains (#1581)
+    - remove FFTW from the Cray Toolchain Definition (#1585)
```
fgeorgatos: maybe say: "this is the preferred default for now" 
(in the future, there might be toolchains with libsci etc perhaps..)

RELEASE_NOTES
```diff
@@ -3,6 +3,64 @@ For more detailed information, please see the git log.
...
+    - add external modules metadata for Cray systems (#1638)
+    - fix independency of Cray toolchains w.r.t. toolchain build environment (#1641, #1647)
```
fgeorgatos: s/fix/improve/g

RELEASE_NOTES
boegel: not sure what you mean here...

We were basically forced to exclude FFTW from the Cray toolchains, because we were running into problems with the Cray-provided `fftw` module is it's always around.

And `libsci` *is* already part of the Cray toolchains.

RELEASE_NOTES
boegel: no, it's actually *fix*, the independency between Cray toolchains was broken, in the sense that using stuff was leaking from one Cray toolchain into the other, if you were using two different toolchains in a single EB session

RELEASE_NOTES
```diff
@@ -3,6 +3,64 @@ For more detailed information, please see the git log.
...
+    - use git rather than https in --new-pr/--update-pr (#1602)
+    - add -u as shorthand for --upload-test-report (#1605)
+    - fix --from-pr for PRs that include renamed/deleted files (#1615)
+    - add support for --install-github-token and --check-github (#1616)
+    - fix fetch_easyconfigs_from_pr w.r.t. duplicate files in PRs (#1628)
+- various other enhancements, including:
+    - add support for --search-filename and --terse (REVIEW) (#1577)
+    - support complete bash completion (#1580)
+    - add support for %(*ver)s and %(*shortver)s templates (#1595, #1604)
+        - incl. %(javaver)s, %(javashortver)s, %(perlver)s, %(perlshortver)s, %(pyver)s, %(pyshortver)s, %(rver)s, $(rshortver)s
```
fgeorgatos: no change for you yet, just a comment:
* this seems like the beginning of some templating need that I have in mind since a long time; think "common deps"...
(no idea how this would really grow in real life though, yet)

fgeorgatos:
ok, lgtm!

sorry for the delay, I was actually reading many of the PRs... very impressive collection of work!

gppezzi:
Please mention on the release notes that recommended way of using custom easyblocks is by using:
 *  [--include-easyblocks](http://easybuild.readthedocs.org/en/latest/Including_additional_Python_modules.html#including-additional-easyblocks-include-easyblocks)

On our setup we were adding custom easyblocks by manually modifying $PYTHONPATH and this method stops working with EB 2.7.0 (well it may still work by tweaking the `__init__.py` file but using `--include-easyblocks` makes more sense).

BTW did I mention already that 2.7 is a great release? :+1:

boegel:
@gppezzi Done, thanks for the suggestion!

Issue #1678, opened 2016-03-18
--include-easyblocks broken with recent setuptools (19.4 or newer)


ocaisa:
With the latest code in develop, `--include-easyblocks` becomes broken when I load GC3Pie (which includes a module for setuptools 20.3, CentOS 7 setuptools is 0.9.8):
```
-bash-4.2$ eb --search GC3
== temporary log file in case of crash /tmp/eb-RT6Y3q/easybuild-B3kHwG.log
== Searching (case-insensitive) for 'GC3' in /work/zam/swmanage/EasyBuild/Golden_Repo/Stage1 
== Searching (case-insensitive) for 'GC3' in /usr/local/software/jureca/Stage1/eb_repo 
== Searching (case-insensitive) for 'GC3' in /homeb/zam/swmanage/software/easybuild-easyconfigs/easybuild/easyconfigs 
 * /work/zam/swmanage/EasyBuild/Golden_Repo/Stage1/GC3Pie-2.4.2.eb
 * /usr/local/software/jureca/Stage1/eb_repo/GC3Pie/GC3Pie-2.4.2.eb
 * /homeb/zam/swmanage/software/easybuild-easyconfigs/easybuild/easyconfigs/g/GC3Pie/GC3Pie-2.2.3.eb
 * /homeb/zam/swmanage/software/easybuild-easyconfigs/easybuild/easyconfigs/g/GC3Pie/GC3Pie-2.3.eb
 * /homeb/zam/swmanage/software/easybuild-easyconfigs/easybuild/easyconfigs/g/GC3Pie/GC3Pie-2.4.0.eb
 * /homeb/zam/swmanage/software/easybuild-easyconfigs/easybuild/easyconfigs/g/GC3Pie/GC3Pie-2.4.1.eb
 * /homeb/zam/swmanage/software/easybuild-easyconfigs/easybuild/easyconfigs/g/GC3Pie/GC3Pie-2.4.2.eb
== Temporary log file(s) /tmp/eb-RT6Y3q/easybuild-B3kHwG.log* have been removed.
== Temporary directory /tmp/eb-RT6Y3q has been removed.
-bash-4.2$ module load GC3Pie
-bash-4.2$ eb --search GC3
ERROR: Failed to parse configuration options: '/homeb/zam/swmanage/software/easybuild-easyblocks/easybuild' is not in list

```

ocaisa:
May be connected to #1677

ocaisa:
pdb output is
```
[...]

PR #1680, opened 2016-03-19
remove unnecessary fixup_namespace_packages in include.py

boegel:
fix for #1678

from what I can tell, using `fixup_namespace_packages` is not needed at all in `include.py` (tests and manual experiments with the various `--include-*` options work just fine, with different setuptools versions)

also includes workaround for `setuptools` bug (https://bitbucket.org/pypa/setuptools/issues/520) in `setUp` of tests, and some cleanup in the tests

use of `fixup_namespace_packages` is only needed in the setup of the tests, to make sure all modules in `sandbox` are picked up correctly

### DIFF
--- a/easybuild/tools/include.py
+++ b/easybuild/tools/include.py
@@ -30,7 +30,6 @@ Support for including additional Python modules, for easyblocks, module naming s
 """
 import os
 import sys
-from pkg_resources import fixup_namespace_packages
 from vsc.utils import fancylogger
 
 from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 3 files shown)


boegel:
@ocaisa, @chrisacbr: up for reviewing/testing this?

gppezzi:
I've tested this with `--include-easyblocks` and the following versions of setuptools:
* 20.2.2
* 17.1.1
* 0.8b2 

and it works.

boegel:
Thanks for the confirmation @gppezzi, I see no reason to hold this back any further, going in!

PR #1683, opened 2016-03-20
Allow hiding toolchains (REVIEW)

geimer:
Fix for https://github.com/hpcugent/easybuild-framework/issues/1682

Unit tests still succeed (since there is no functional change if `--hide-deps` is not used) and the example mentioned in https://github.com/hpcugent/easybuild-framework/issues/1682 works like a charm, both a dry-run and an actual build.

One or the other unit test should probably use the `--hide-deps` option, though I don't have a clear idea yet where this would be most effective.

**Edit:** After a discussion during an EB conf call (cfr. https://github.com/hpcugent/easybuild/wiki/Conference-call-notes-20160706), the approach taken has slightly changed.  This PR will tackle
 - [x] Hiding toolchains via `--hide-toolchains=foo,bar` command-line option (as well as env var and config file parameter)
 - [x] Extend `toolchain` easyconfig parameter to handle `hidden` key (as outlined in https://github.com/hpcugent/easybuild-framework/pull/1683#issuecomment-205492486)
 - [x] Unit tests

To be done in a separate PR:
 - Handle `hidden` parameter in `--try-toolchain` option


boegel:
Jenkins: ok to test

pforai:
+1 on this!

geimer:
@boegel Could you please take a quick look at this and let me know what is missing?  I still don't know where to test this in the unit tests. As far as I can see, there is no test for `as_dict` yet and the result would only have an impact at a higher level anyway.

easybuild/tools/toolchain/toolchain.py
```diff
@@ -108,6 +108,7 @@ def __init__(self, name=None, version=None, mns=None, class_constants=None, tcde
 
         # toolchain instances are created before initiating build options sometimes, e.g. for --list-toolchains
         self.dry_run = build_option('extended_dry_run', default=False)
+        self.hide_deps = build_option('hide_deps', default=None)
```
boegel: I would use an empty list as a fallback here (not as default value for the build option, that's something else)

so:

```python
self.hide_deps = build_option('hide_deps', default=None) or []
```

easybuild/tools/toolchain/toolchain.py
```diff
@@ -269,6 +270,10 @@ def as_dict(self, name=None, version=None):
             name = self.name
         if version is None:
             version = self.version
+        if self.hide_deps:
```
boegel: no need for this check if you make sure `self.hide_deps` is always a list (by using `[]` as a fallback, see above)

boegel:
@geimer: there is a test for `--hide-deps` already, see `test/framework/options.py`, look for `test_hide_deps`; you may want to consider a dedicated new test (`test_hide_toolchain` or whatever)

However, I'm not sure this is the right approach here...

With the current proposed changes, a hidden toolchain can only be used if the toolchain name is listed in `--hide-deps`. I'm not sure if that is enough...

Shouldn't there be a more explicit way of using a hidden toolchain as well?
For example, by supporting this:

```python
toolchain = {'name': 'foss', 'version': '2016a', 'hidden': True}
```

Or even through the module naming scheme somehow (maybe that's already possible though, and it doesn't need any changes in the framework, not sure).

I think we need to double check that we're covering all bases here, since there are also multiple ways in which a dependency can be listed as hidden (`--hide-deps`, `hiddendependencies`, `eb --hidden`), ...

geimer:
@boegel: Thanks, I'll take a look at the tests in `test/framework/options.py`.

> Shouldn't there be a more explicit way of using a hidden toolchain as well?
For example, by supporting this:
```
toolchain = {'name': 'foss', 'version': '2016a', 'hidden': True}
```

I'd like to avoid this approach, as it would require users to touch all easyconfigs that use this toolchain.  What would be the advantage?

> Or even through the module naming scheme somehow (maybe that's already possible though, and it doesn't need any changes in the framework, not sure).

How does the naming scheme fit into the picture???

> I think we need to double check that we're covering all bases here, since there are also multiple ways in which a dependency can be listed as hidden (`--hide-deps`, `hiddendependencies`, `eb --hidden`), ...

`eb --hidden` should be no problem, as it only affects the package that is currently being built, which can't be it's own toolchain.  W.r.t. `--hide-deps` and `hiddendependencies`: I thought they were affecting the same setting, i.e., `hiddendependencies` being the easyconfig variant of the `--hide-deps` command-line parameter...

boegel:
@geimer: you're right about the relation between `--hide-deps` and `hiddendependencies`, which is sort of my point...

With your current approach, hiding a toolchain can only be done via `--hide-deps`; including it in `hiddendependencies` will not work (the `toolchain` is not a regular dependency), and there's no equivalent alternative to making a toolchain hidden in an easyconfig file.

I wouldn't like having to touch all the easyconfigs either, but it is a good way of making very sure the toolchain will always be installed hidden, regardless of the EasyBuild configuration that is in place (similar to `hiddendependencies`).

So, I'm wondering whether supporting with something like `--hide-toolchains` and `hidden_toolchain = True` would be better, rather than 'half' mixing dependencies and toolchains...

ocaisa:
Personally, I found it painful to use `hiddendependencies` and have abandoned it in favour of `--hide-deps` (or at least the equivalent env-var which I configure for all our installers). Unless it's something is consistently hidden for all configs coming from the main repo (and there are currently 0 like this), it's very tedious to remember...or more particularly to get everyone else to remember.

geimer:
@boegel: Components such as `GCCcore` can be both: a toolchain (e.g., for `binutils-2.25-GCCcore-4.9.3`) and a dependency (e.g., for `GCC-4.9.3-2.25` whose toolchain is `dummy`).  Thus, introducing `--hide-toolchains` would still require to list `GCCcore` in `--hide-deps` for a robot build to do the right thing.  That is, you would have to list it twice.  Another source for errors.  Urgh...

The more I think about it, the more I'm inclined to say that the whole `hidden` business is designed the wrong way around.  In the end, being hidden is a property of a package.  So why should all packages depending on it (either as a dependency or a toolchain) care whether it is hidden or not?  They probably shouldn't -- and leave it up to the dependency resolution mechanism of `--robot` to select the right module.  This would mean to only support the `--hidden` command-line option as well as a `hidden = True` easyconfig parameter.  An environment variable interface is tricky, though, as it may only be desired to hide a module built with a certain toolchain (e.g., one may want the `Bison-3.0.4` module used for bootstrapping to be hidden, whereas the `Bison-3.0.4-GCCcore-4.9.3` module should stay visible).

ocaisa:
I agree that supporting `hidden = True` in the easyconfig might be a simple solution but Kenneth argued against that when we first suggested it. Maybe he can be convinced now...

On a similar note, I ran into an issue yesterday that our back-end has no `libuuid` but our frontend does. If you're building on the back-end then you need to provide the dep, but then on the front-end this causes problems because basic system libs are expecting the system libuuid which has other symbols inside. The solution was actually simple since the particular builds that needed libuuid had rpath-ed it in, so I could safely leave it as a build dep so that on the front-end it resolves to the system provided version for the other tools. The massive amount of vis libs that we currently installing makes me worried about the environment exported to slurm getting too large, and I see this approach as a potential solution. We could sidestep having to load all the dependencies of (for example) libX11 at runtime by having an `rpathed = True`  in the easyconfig

boegel:
@geimer this is still WIP, right, since you're planning to add support for marking a toolchain as `hidden` in the easyconfig file too? if so, please update the PR title (and also mention you're adding a `hidden` easyconfig parameter)

Are you planning to make the robot hidden-aware in this PR too, i.e. make it able to always fall back to checking for a hidden module too? Maybe it's better to do that in a separate PR?

geimer:
@boegel: I believe the parts that are already *done* in this PR are ready for review.  That said, I got lost while investigating how things are passed from the option parsing to the easyconfig tweaking to address `--try-toolchain` handling...  But maybe it would be better to handle this in a separate PR anyhow to keep things in bite-sized chunks?  What do you think?

geimer:
@boegel: As discussed during the conf call yesterday, `--try-toolchain` handling should be addressed in a separate PR. Thus renamed this PR and opened it for review.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1252,6 +1252,8 @@ def process_easyconfig(path, build_specs=None, validate=True, parse_only=False,
                 dep = ec.toolchain.as_dict()
                 _log.debug("Adding toolchain %s as dependency for app %s." % (dep, name))
                 easyconfig['dependencies'].append(dep)
+                if dep['hidden']:
+                    easyconfig['hiddendependencies'].append(dep)
```
boegel: @geimer: I don't think this is required? I even think we should *not* do this, since it may have unwanted side-effects...

I'm working on another PR to your branch, where this is removed, but I figured I should double-check with you first why this was added.

geimer: @boegel: To be honest, I don't remember.  I almost believe that this is a leftover from some hacking and got committed incidentally...

geimer: @boegel: Maybe to create a correct `module load` statement for the hidden module in the modulefile? Cross-checking it right now...

boegel:
@geimer some minor style fixes and added tests for `--hide-toolchains` and hiding aspect of `toolchain` easyconfig parameter in https://github.com/geimer/easybuild-framework/pull/9

easybuild/framework/easyconfig/easyconfig.py
boegel: No, I checked where `hiddendependencies` is used, and it's not used to determine the module name to use.

I dropped this in https://github.com/geimer/easybuild-framework/pull/9, I'm fairly sure it's not needed at all.

geimer: @boegel: I tried various things and couldn't reproduce a situation where is needed. So, yes, let's drop it.

geimer:
@boegel: Does it make sense to rename the `to_name_version_dict` function to something like `to_toolchain_dict`?  As far as I can see it is really only used for toolchain specifications and it's now more than just name/version.

boegel:
@geimer: sure, I think it makes sense... Although we should keep `to_name_version_dict` around for now as a wrapper around `to_toolchain_dict`, and deprecate it (using a `log.deprecated` call), just to be safe in case someone uses this somewhere else (e.g. in a script that hooks into the framework).

geimer:
@boegel: WIP, stay tuned...  Is there any way to also deprecate the `NAME_VERSION_DICT` constant?

geimer:
@boegel Do we need to extend `test_check_type_of_param_value_toolchain` as well?

geimer:
@boegel Please take another look at my last commit which does the renaming/deprecation. The two questions from the previous comments still hold, though.

easybuild/framework/easyconfig/types.py
```diff
@@ -272,22 +275,33 @@ def to_name_version_dict(spec):
...
     else:
-        raise EasyBuildError("Conversion of %s (type %s) to name and version dict is not supported", spec, type(spec))
+        raise EasyBuildError("Conversion of %s (type %s) to toolchain dict is not supported", spec, type(spec))
 
     return res
 
 
+def to_name_version_dict(spec):
+    """Deprecated in favor of to_toolchain_dict."""
+    _log.deprecated("to_name_version_dict; use to_toolchain_dict instead.", '2.9')
```
boegel: the version number indicates when the warning will be transformed into an error, so use `3.0`

geimer: @boegel: Are you sure?  The output I get when modifying one of the unit tests is
```
DEPRECATED (since v2.9) functionality used: to_name_version_dict; use to_toolchain_dict instead.; see http://easybuild.readthedocs.org/en/latest/Deprecated-functionality.html for more information
```
This suggests that it is the version since when it is deprecated, no?

boegel:
@geimer: I wouldn't worry about deprecating `NAME_VERSION_DICT`, I consider that internal to the framework, and I'd be quite surprised if someone is importing that directly; so, there's no need to retain `NAME_VERSION_DICT`, just renaming it is fine

also enhancing `test_check_type_of_param_value_toolchain` is probably a good idea, yes

easybuild/framework/easyconfig/types.py
boegel: as soon as you include a `log.deprecated` message, you're deprecating something

the version indicates when the deprecation warning becomes a deprecation error, so it should be 3.0

boegel:
Going in, thanks @geimer!

boegel:
Ah crap, there are conflicts with current `develop` that will need to be tackled first... :(

@geimer: can you look into that? If not, I'll tackle it myself with a wrapper PR later this week.

boegel:
bootstrap test fails due to unrelated issue, cfr. #1869, so good to go

Thanks @geimer!

PR #1698, opened 2016-03-25
Add option to daemonize EB

riccardomurri:
With the new option `--daemonize` implemented here, EB can detach itself from the controlling terminal and continue running in the background.

A PID file is written so one can kill the running daemon. Note that compilation processes, etc. will go on for a while when the daemon is killed.

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -44,7 +44,7 @@ import traceback
 
 # IMPORTANT this has to be the first easybuild import as it customises the logging
 #  expect missing log output when this not the case!
-from easybuild.tools.build_log import EasyBuildError, init_logging, print_msg, print_error, stop_logging
+from easybuild.tools.build_log import EasyBuildError, NonTerminatingLoggerProxy, init_logging, print_msg, print_error, stop_logging
 
 import easybuild.tools.config as config
 import easybuild.tools.options as eboptions
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -182,9 +182,12 @@ def handle_github_options(options, ec_paths):
     return done
 
 
-def main(args=None, logfile=None, do_build=None, testing=False):
+def main_init(args=None, logfile=None, do_build=None, testing=False):
     """
-    Main function: parse command line options, and act accordingly.
+    Set up environment, possibly daemonize and give control to main function.
+
+    Any arguments will be forwarded unchanged to `main_body`.
+
     @param args: command line arguments to use
     @param logfile: log file to use
     @param do_build: whether or not to actually perform the build
@@ -196,7 +199,6 @@ def main(args=None, logfile=None, do_build=None, testing=False):
     # initialise options
     eb_go = eboptions.parse_options(args=args)
     options = eb_go.options
-    orig_paths = eb_go.args
 
     # set umask (as early as possible)
     if options.umask is not None:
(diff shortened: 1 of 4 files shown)


boegel:
Jenkins: ok to test

easybuild/tools/build_log.py
```diff
@@ -168,6 +168,21 @@ def exception(self, msg, *args):
         raise EasyBuildError(ebmsg, *args)
 
 
+class NonTerminatingLoggerProxy(object):
+    """
+    Monkeypatch an existing `EasyBuildLog` object so that calling
+    `.error()` does not raise an exception and terminate the program.
+    """
+    def __init__(self, logger):
+        self._logger = logger
```
boegel: @riccardomurri should we also include a `logger.deprecated` call in here, to make sure we trash this when we get to making `log.error` behave as expected?

easybuild/main.py
```diff
@@ -44,7 +44,7 @@
 
 # IMPORTANT this has to be the first easybuild import as it customises the logging
 #  expect missing log output when this not the case!
-from easybuild.tools.build_log import EasyBuildError, init_logging, print_msg, print_error, stop_logging
+from easybuild.tools.build_log import EasyBuildError, NonTerminatingLoggerProxy, init_logging, print_msg, print_error, stop_logging
```
boegel: our coding style prescribes a column limit of 120 chars, so please change to:

```python
from easybuild.tools.build_log import EasyBuildError, NonTerminatingLoggerProxy
from easybuild.tools.build_log import init_logging, print_msg, print_error, stop_logging
```

easybuild/main.py
```diff
@@ -219,9 +221,68 @@ def main(args=None, logfile=None, do_build=None, testing=False):
     eb_cmd_line = eb_go.generate_cmd_line() + eb_go.args
     log_start(eb_cmd_line, eb_tmpdir)
 
+    # report on umask here so we don't have to forward this info to `main`
     if options.umask is not None:
         _log.info("umask set to '%s' (used to be '%s')" % (oct(new_umask), oct(old_umask)))
 
+    # sanity check
+    if options.daemonize and options.logtostdout:
```
boegel: there are probably others... basically anything where the whole point of using the option is to see output

at least `--search`, `--dry-run`, `--extended-dry-run`, `--review-pr`, `--show-config`, ...

easybuild/main.py
```diff
@@ -219,9 +221,68 @@ def main(args=None, logfile=None, do_build=None, testing=False):
...
+        raise EasyBuildError(
+            "Option `--daemonize` is incompatible with `--logtostdout`."
+            " Please use only one of the two.")
```
boegel: collapse this to a single line?

easybuild/main.py
boegel: all `--list-*` and `--avail-*` options too

riccardomurri: > all `--list-*` and `--avail-*` options too

Is there a way I can extract them from an options data structure, I
should I rather hard-code values here?

easybuild/main.py
```diff
@@ -219,9 +221,68 @@ def main(args=None, logfile=None, do_build=None, testing=False):
...
+            # local state
+            init_session_state, eb_go, eb_tmpdir, eb_cmd_line,
+            # original `main` args
+            args, logfile, do_build, testing)
+    else:
+        main_body(
+            # local state
+            init_session_state, eb_go, eb_tmpdir, eb_cmd_line,
+            # original `main` args
+            args, logfile, do_build, testing)
```
boegel: I don't like the copy-pasting of the arguments here... how about this?

```python
if options.daemonize:
    main_continued = main_daemonize
else:
    main_continued = main_body

main_continued(...)
```

easybuild/main.py
```diff
@@ -219,9 +221,68 @@ def main(args=None, logfile=None, do_build=None, testing=False):
...
+
+
+def main_daemonize(init_session_state, eb_go, eb_tmpdir, eb_cmd_line,
+                   args=None, logfile=None, do_build=None, testing=False):
```
boegel: please include a docstring

easybuild/main.py
```diff
@@ -219,9 +221,68 @@ def main(args=None, logfile=None, do_build=None, testing=False):
...
+    pidfile = os.path.join(eb_tmpdir, 'easybuild.pid')
+    print_msg("Backgrounding EasyBuild process,"
+              " will record PID in file '%s' ..." % (pidfile,))
+    from daemonize import Daemonize
```
boegel: please move to the top, protected with a `try`/`except ImportError`, and use our fancy `only_if_module_is_available` decorator, which issues clear error messages

easybuild/main.py
```diff
@@ -219,9 +221,68 @@ def main(args=None, logfile=None, do_build=None, testing=False):
...
+    daemon = Daemonize(
+        app='EasyBuild',
```
boegel: hmm, maybe we should define this as a constant somewhere?

easybuild/main.py
```diff
@@ -219,9 +221,68 @@ def main(args=None, logfile=None, do_build=None, testing=False):
...
+        pid=pidfile,
+        chdir=os.getcwd(),
+        auto_close_fds=False,
+        logger=NonTerminatingLoggerProxy(_log),
+        verbose=True,
+        action=(lambda: main_body(
```
boegel: use `*args, **kwargs`?

easybuild/main.py
```diff
@@ -219,9 +221,68 @@ def main(args=None, logfile=None, do_build=None, testing=False):
...
+            args, logfile, do_build, testing)))
+    daemon.start()
+
+
+def main_body(init_session_state, eb_go, eb_tmpdir, eb_cmd_line,
+              args=None, logfile=None, do_build=None, testing=False):
+    """
+    Main function: parse command line options, and act accordingly.
+
+    @param init_session_state: Part of the context set up by `main1`
```
boegel: there's no `main1` (anymor)

setup.py
```diff
@@ -113,6 +113,7 @@ def find_rel_test():
     install_requires=[
         'setuptools >= 0.6',
         "vsc-base >= 2.4.18",
+        'daemonize >= 2.4.6',
```
boegel: hmm, should we include this as an optional dependency instead (see `extras_require`)?

easybuild/main.py
boegel: Since all of the options are attributes in `self.options`, I think you need to resort to something like:

```python
list_avail_option_attrs = [a for a in dir(self.options) if a.startswith('list_') or a.startswith('avail_')]
if any([getattr(self.options, a) for a in list_avail_option_attrs]):
    ....
```

easybuild/tools/build_log.py
riccardomurri: Bcc:
Reply-To:
> @riccardomurri should we also include a `logger.deprecated` call in here, to make sure we trash this when we get to making `log.error` behave as expected?

Either that or a FIXME comment.

(But where would you place a `logger.deprecated` call exactly?)

easybuild/main.py
riccardomurri: > > +        app='EasyBuild',
>
> hmm, maybe we should define this as a constant somewhere?

Looks like overengineering to me :-)

If you're ever going to rename the project, this line will be the least
of your problems :-)

easybuild/main.py
riccardomurri: > > +        action=(lambda: main_body(
>
> use `*args, **kwargs`?

No, the function passed as `action` is called by `daemon.start()` and is
called without any arguments.

setup.py
riccardomurri: (Kenneth Hoste, Fri, Mar 25, 2016 at 08:46:59AM -0700:)
> > @@ -113,6 +113,7 @@ def find_rel_test():
> >      install_requires=[
> >          'setuptools >= 0.6',
> >          "vsc-base >= 2.4.18",
> > +        'daemonize >= 2.4.6',
>
> hmm, should we include this as an optional dependency instead (see `extras_require`)?

That's your call as release manager :-) to assess whether this would be
a generally-useful feature that many people could profit from, or a
niche functionality (those who want it will have to go the extra mile).

That said, the `daemonize` Python library is a single file of ~200 lines
with no additional dependencies so I do not see any problem in depending
on it even for the default EB.

PR #1699, opened 2016-03-25
Fix running test suite `test.framework.options` on my EB install

riccardomurri:
Actual patch by @boegel but I'm commiting it since it's on my laptop.

edit (by @boegel): to clarify: this fixes the problem where the easyblocks from the EasyBuild installation were being picked up by the `test_include_easyblocks*` tests, even though the `setUp` of each tests tries to isolate the tests from non-test easyblocks...

The problem was that the original `sys.path` is being restored in the middle of the `test_include_easyblocks*` tests, which potentially reinstated the location to non-test easyblocks in the Python search path, breaking the 2nd part of these tests.

### DIFF
--- a/test/framework/options.py
+++ b/test/framework/options.py
@@ -1766,6 +1766,8 @@ class CommandLineOptionsTest(EnhancedTestCase):
 
     def test_include_easyblocks(self):
         """Test --include-easyblocks."""
+        orig_local_sys_path = sys.path[:]
+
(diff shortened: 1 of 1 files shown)


boegel:
Thanks @riccardomurri!

Jenkins: ok to test

boegel:
Going in, thanks again @riccardomurri!

PR #1701, opened 2016-03-25
Change option `--color` choices to auto/always/never

riccardomurri:
The option values are as in `ls --color` and `grep --color`,
except that "auto" is the default which colorizes logs and output
whenever EB's STDOUT is connected to a terminal.

**Note that this requires PR ~~https://github.com/hpcugent/vsc-base/pull/225~~ to be merged!**
(Otherwise `fancylogger.logToScreen()` does not have a `color` parameter.)


setup.py
```diff
@@ -112,7 +112,8 @@ def find_rel_test():
     zip_safe=False,
     install_requires=[
         'setuptools >= 0.6',
-        "vsc-base >= 2.4.18",
+        "vsc-base >= 2.5.0",
+        'humanfriendly',  # determine whether terminal supports ANSI color
```
boegel: this should be moved down into `extras_require`, only the packages that are strictly required to make `eb` work  should be listed in `install_requires`

setup.py
```diff
@@ -112,7 +112,8 @@ def find_rel_test():
     zip_safe=False,
     install_requires=[
         'setuptools >= 0.6',
-        "vsc-base >= 2.4.18",
+        "vsc-base >= 2.5.0",
```
boegel: what you specify here depends on how you bump the `vsc-base` version in https://github.com/hpcugent/easybuild-framework/pull/225, which should most likely be a minor bump, although I guess bumping to 2.5.0 wouldn't hurt anyone (it's just that there's nothing shocking being added that may warrant a bump from 2.4.x to 2.5.0...)

easybuild/tools/options.py
```diff
@@ -258,7 +258,7 @@ def override_options(self):
                                             None, 'store_true', False),
             'cleanup-builddir': ("Cleanup build dir after successful installation.", None, 'store_true', True),
             'cleanup-tmpdir': ("Cleanup tmp dir after successful run.", None, 'store_true', True),
-            'color': ("Allow color output", None, 'store_true', True),
+            'color': ("Colorize output", 'choice', 'store', 'auto', ['auto', 'always', 'never'], {'metavar':'WHEN'}),
```
boegel: since enabling this requires having `humanfriendly` installed, I wouldn't enable this by default...

also, changing this may break people's configuration files, since they may specify `False` or `True` now, so we'll need to be careful with changing this to string, and at least translate `False` to `never` and `True` to `auto` (I wouldn't go with `always` since using that is probably rare...)

easybuild/tools/build_log.py
```diff
@@ -186,10 +186,10 @@ def exception(self, msg, *args):
 _init_easybuildlog = fancylogger.getLogger(fname=False)
 
 
-def init_logging(logfile, logtostdout=False, silent=False):
+def init_logging(logfile, logtostdout=False, silent=False, color=None):
```
boegel: the default for `color` here should be a string value? or `None` should be handled well in `logToScreen`, maybe that's better?

easybuild/main.py
```diff
@@ -182,6 +188,25 @@ def handle_github_options(options, ec_paths):
     return done
 
 
+def _use_color(colorize, stream=sys.stdout):
```
boegel: this should probably be moved into `tools/options.py`, and be called from the `postprocess` there to determine whether `options.color` should be defined as `True` or `False`, based on the provided (string) value?

easybuild/main.py
```diff
@@ -64,6 +64,12 @@
 from easybuild.tools.testing import create_test_report, overall_test_report, regtest, session_module_list, session_state
 from easybuild.tools.version import this_is_easybuild
 
+try:
+    from humanfriendly.terminal import terminal_supports_colors
+except ImportError:
+    # provide an approximation that should work in most cases
```
boegel: so, when would it not work? do we really needed `humanfriendly` at all?

boegel:
Jenkins: ok to test

easybuild/main.py
```diff
@@ -208,7 +233,8 @@ def main(args=None, logfile=None, do_build=None, testing=False):
 
     # initialise logging for main
     global _log
-    _log, logfile = init_logging(logfile, logtostdout=options.logtostdout, silent=testing or options.terse)
+    _log, logfile = init_logging(logfile, logtostdout=options.logtostdout,
+                                 silent=(testing or options.terse)), color=options.color)
```
boegel: the `))` here should be `)`, syntax error

@riccardomurri you didn't test this, did you? Did you?! ;-)

pforai:
+1 for this!

setup.py
riccardomurri: (Kenneth Hoste, Fri, Mar 25, 2016 at 03:41:57PM -0700:)
> > @@ -112,7 +112,8 @@ def find_rel_test():
> >      zip_safe=False,
> >      install_requires=[
> >          'setuptools >= 0.6',
> > -        "vsc-base >= 2.4.18",
> > +        "vsc-base >= 2.5.0",
>
> what you specify here depends on how you bump the `vsc-base` version
> in https://github.com/hpcugent/easybuild-framework/pull/225

yes, of course they should be synchronized; see comments on that PR for
the versioning debate (short version: I don't care, just tell me the number)

easybuild/tools/options.py
riccardomurri: (Kenneth Hoste, Fri, Mar 25, 2016 at 03:44:00PM -0700:)
> > @@ -258,7 +258,7 @@ def override_options(self):
> >                                              None, 'store_true', False),
> >              'cleanup-builddir': ("Cleanup build dir after successful installation.", None, 'store_true', True),
> >              'cleanup-tmpdir': ("Cleanup tmp dir after successful run.", None, 'store_true', True),
> > -            'color': ("Allow color output", None, 'store_true', True),
> > +            'color': ("Colorize output", 'choice', 'store', 'auto', ['auto', 'always', 'never'], {'metavar':'WHEN'}),
>
> since enabling this requires having `humanfriendly` installed,

No it doesn't -- if you don't have the required packages, it will
silently fall back to "no colorization".

> also, changing this may break people's configuration files, since they
> may specify `False` or `True` now, so we'll need to be careful with
> changing this to string, and at least translate `False` to `never` and
> `True` to `auto`

Good point, but how to I do that with vsc's option parsing framework?

easybuild/tools/build_log.py
riccardomurri: (Kenneth Hoste, Fri, Mar 25, 2016 at 03:44:54PM -0700:)
> > @@ -186,10 +186,10 @@ def exception(self, msg, *args):
> >  _init_easybuildlog = fancylogger.getLogger(fname=False)
> >
> >
> > -def init_logging(logfile, logtostdout=False, silent=False):
> > +def init_logging(logfile, logtostdout=False, silent=False, color=None):
>
> the default for `color` here should be a string value?

Yep, changed default to `auto`

easybuild/main.py
riccardomurri: (Kenneth Hoste, Fri, Mar 25, 2016 at 03:47:26PM -0700:)
> > @@ -64,6 +64,12 @@
> >  from easybuild.tools.testing import create_test_report, overall_test_report, regtest, session_module_list, session_state
> >  from easybuild.tools.version import this_is_easybuild
> >
> > +try:
> > +    from humanfriendly.terminal import terminal_supports_colors
> > +except ImportError:
> > +    # provide an approximation that should work in most cases
>
> so, when would it not work?

`os.isatty()` tells you if the stream is connected to a terminal, it
does not say whether the terminal supports ANSI colors.

I guess it's OK to presume that everyone nowadays is using a terminal
emulator with ANSI color support (hence "most cases") but you never
know...  (I have seen people still run FVWM as their window manager in
2015; I would not be surprised to hear that somebody is stil using a
pre-ANSI `xterm` ...)


> do we really needed `humanfriendly` at all?

It's a dependency of `coloredlogs` (the Python package), so if you are
going to have colored logs, `humanfriendly` is going to be there anyway.

easybuild/main.py
riccardomurri: (Kenneth Hoste, Fri, Mar 25, 2016 at 04:07:15PM -0700:)
> > @@ -208,7 +233,8 @@ def main(args=None, logfile=None, do_build=None, testing=False):
> >
> >      # initialise logging for main
> >      global _log
> > -    _log, logfile = init_logging(logfile, logtostdout=options.logtostdout, silent=testing or options.terse)
> > +    _log, logfile = init_logging(logfile, logtostdout=options.logtostdout,
> > +                                 silent=(testing or options.terse)), color=options.color)
>
> the `))` here should be `)`, syntax error
>
> @riccardomurri you didn't test this, did you? Did you?! ;-)

Damn Jenkins, you caught me!

easybuild/main.py
riccardomurri: Bcc:
Reply-To:
(Kenneth Hoste, Fri, Mar 25, 2016 at 03:46:34PM -0700:)
> > @@ -182,6 +188,25 @@ def handle_github_options(options, ec_paths):
> >      return done
> >
> >
> > +def _use_color(colorize, stream=sys.stdout):
>
> this should probably be moved into `tools/options.py`,

Done.

> and be called from the `postprocess` there to determine whether `options.color` should be defined as `True` or `False`, based on the provided (string) value?

Not done: `fancylogger.logToScreen()` expects a 3-choice string value so
I'm keeping it this way.

easybuild/tools/options.py
riccardomurri: >> also, changing this may break people's configuration files, since they
>> may specify `False` or `True` now, so we'll need to be careful with
>> changing this to string, and at least translate `False` to `never` and
>> `True` to `auto`
>
> Good point, but how to I do that with vsc's option parsing framework?

Actually, I would translate the other way round:

* `True` -> `auto`
* `False` -> `never`

easybuild/tools/options.py
```diff
@@ -142,6 +153,24 @@ def pretty_print_opts(opts_dict):
...
+    """
+    Return ``True`` or ``False`` depending on whether ANSI color
+    escapes are to be used when printing to `stream`.
+
+    The `colorize` argument can take the three string values
+    ``'auto'``/``'always'``/``'never'``, see the ``--color`` option
+    for their meaning.
+    """
+    # turn color=auto/yes/no into a boolean value
+    if options.color == 'auto':
```
boegel: @riccardomurri this should use `colorize` rather than `options.color`?

riccardomurri:
@boegel I cannot reproduce the failure that Jenkins reported on my laptop, and I cannot understand what part of the code has actually failed from the traceback.... help?

boegel:
@riccardomurri The failing test is probably because the PR is not being tested against the updated `vsc-base`, i.e. `fancylogger.logToScreen` doesn't consume a named argument `colorize` yet

In other words: https://github.com/hpcugent/easybuild-framework/pull/225 needs to be merged first.

boegel:
@riccardomurri please merge https://github.com/riccardomurri/easybuild-framework/pull/26 to sync this with current `develop`

riccardomurri:
@boegel done.

setup.py
```diff
@@ -126,6 +126,11 @@ def find_rel_test():
     ],
     extras_require = {
         'yeb': ["PyYAML >= 3.11"],
+        'coloredlogs': [
+            'vsc-base[coloredlogs] >= 2.5.0',
```
boegel: @riccardomurri please bump to 2.5.3 as minimal required version

riccardomurri:
@boegel bumped `vsc-base` requirement in commit https://github.com/hpcugent/easybuild-framework/pull/1701/commits/544563fcd6539f1a6cc760046afc977b8b64e202

boegel:
lgtm, thanks @riccardomurri!

Issue #1685, opened 2016-03-21
Test failure of EB 2.7.0


jhein32:
HI Kenneth,

Following your message I upgraded our cluster to EB 2.7.0 and got a few failures.  If I understand the errors correctly, this one is strange.  If I do eb --list-toolchains and eb --avail-module-naming-schemes on the command line, the commands behave "normal".  

Here is a screenshot from the test:

![screen shot 2016-03-21 at 18 35 08](https://cloud.githubusercontent.com/assets/15908212/13928194/08f0f202-ef95-11e5-9655-449108c20e0f.png)

Thanks for looking.

Best wishes
  Joachim

boegel:
@jhein32 There's a new requirement in the tests, i.e. that `eb` is available through `$PATH` (that wasn't needed before to make the tests pass).

However, since you're loading the `EasyBuild` module, you *have* `eb` available through `$PATH` in this session, right?

Do you have anything particular in your `.bashrc` that may be getting in the way? These tests use `run_cmd` to run `eb`, which is done in a subshell...

jhein32:
On 21 Mar 2016, at 23:08, Kenneth Hoste <notifications@github.com<mailto:notifications@github.com>> wrote:


@jhein32<https://github.com/jhein32> There's a new requirement in the tests, i.e. that eb is available through $PATH (that wasn't needed before to make the tests pass).

However, since you're loading the EasyBuild module, you have eb available through $PATH in this session, right?


eb was available.  I could issue the commands at the prompt and they succeeded.

Do you have anything particular in your .bashrc that may be getting in the way? These tests use run_cmd to run eb, which is done in a subshell...


I don’t think there is anything particular in my .bashrc or similar.  Are these the only tests running the eb command via “run_cmd”?

Joachim

boegel:
@jhein32 Yes, only these two tests require having `eb` available in `$PATH`...

I think the problem is related to the use of `module purge` by the tests, which is fixed in #1702.

boegel:
@jhein32 Please try applying the simple patch from https://github.com/hpcugent/easybuild-framework/pull/1702/files, which should resolve the problem with `eb` when running the tests.

jhein32:
Hi,

I finally managed to test this (sorry that it took time) and, unless I made a mistake, this does not seem to resolve the issue :(

I hand edited the file `utilities.py` in the directory
[...]

PR #1702, opened 2016-03-25
Remove purge from the tests 

pforai:
in Analogy to https://github.com/hpcugent/easybuild-framework/pull/1608


boegel:
I can't come up with any style remarks for this, and Jenkins is happy about it, so I guess I'll need to just go ahead and merge this... ;-)

This makes perfect sense, since the original environment is being restored in `tearDown` after each test, and `module purge` is 'dangerous' in the sense that it may break the environment, cfr. the problem with `eb` not being found in #1685.

So, going in, thanks @pforai!

Issue #1697, opened 2016-03-25
qt486+dummy: easyblock qt.py


jgphpc:
Hi,

As i am working on https://github.com/hpcugent/easybuild-easyconfigs/issues/2777, 
eb ./Qt-4.8.6.eb will fail if

```
EASYBUILD_INCLUDE_EASYBLOCKS=$FORKJG/easybuild-easyblocks/easybuild/easyblocks/*/*.py
```
but not if
```
EASYBUILD_INCLUDE_EASYBLOCKS=$FORKJG/easybuild-easyblocks/easybuild/easyblocks/q/qt.py
```

The error message is:
```
eb ./Qt-4.8.6.eb
ERROR: Failed to parse configuration options: 'Symlinking /apps/common/UES/sandbox/jgp/
easybuild.forkjg/easybuild-easyblocks/easybuild/easyblocks/0/__init__.py 
to /tmp/eb-2S5tfS/included-easyblocks/easybuild/easyblocks/__init__.py failed: 
[Errno 17] File exists'
```

The easyblock looks like:
```.py
class EB_Qt(ConfigureMake):
    """
    Support for building and installing Qt.
    """

    @staticmethod
    def extra_options():
        extra_vars = {
             'platform': [None, "Target platform to build for (e.g. linux-g++-64, linux-icc-64)", CUSTOM],
        }
        return ConfigureMake.extra_options(extra_vars)

    def configure_step(self):
        """Configure Qt using interactive `configure` script."""

        self.cfg.update('configopts', '-release')

        platform = None
        comp_fam = self.toolchain.comp_family()
        if self.cfg['platform']:
                platform = self.cfg['platform']
        # if no platform is specified, try to derive it based on compiler in toolchain
        elif comp_fam in [toolchain.GCC]:  #@UndefinedVariable
                platform = 'linux-g++-64'
        elif comp_fam in [toolchain.INTELCOMP]:  #@UndefinedVariable
                platform = 'linux-icc-64'

        if platform:
                self.cfg.update('configopts', "-platform %s" % platform)
        else:
                raise EasyBuildError("Don't know which platform to set based on compiler family.")
```

In my easyconfig, i have:
```
configopts += ' -platform linux-g++-64'
```

I also get another error:
```
 'platform': [None, "Target platform to build for (e.g. linux-g++-64, linux-icc-64)", CUSTOM],
NameError: global name 'CUSTOM' is not defined
```

jg.

boegel:
[...]

PR #1704, opened 2016-03-26
ignore __init__.py files in --include-*

boegel:
fix for problem reported in #1697

cc @jgphpc


boegel:
@wpoely86 please review?

wpoely86:
looks good but maybe also protect against `__init__.pyc`?

boegel:
Having an `__init__.pyc` is irrelevant; the non-init `*.py` files are symlinked in a temporary directory, and then `eb` adds it's own `__init__.py` along with the symlinks. If the `__init__.py` files are symlinked too, things break.

So, good to go imho, thanks for the review @wpoely86!

PR #1715, opened 2016-04-07
Clarify boolean options documentation --disable-cleanup-builddir

rjeschmi:
Since some options have defaults which are true, it isn't easy to find how to disable them (and the command line option is by default a noop).

Also, the python default seems to be "--no-*" so --no-cleanup-builddir or --no-cleanup-tmpdir

The options used in easybuild are documented at the end of the help but are --disable-cleanup-builddir and --disable-cleanup-tmpdir

This PR adds those options to their so a search will find them. In the worst case this PR might get indexed and save someone some time.


boegel:
@rjeschmi If we do this, i.e. mention the `--disable-*` option for boolean options that are enabled by default, we should do it consistently.

And the only way to do that is to automagically amend the help messages for those options, rather than doing it manually.

boegel:
closing this, we did this in a better way in #3151

Issue #1717, opened 2016-04-08
Option openmp deprecated


ocaisa:
This is going to be interesting, Intel have deprecated the `-openmp` option and you should now use `-qopenmp`...I think we can update to `-fopenmp` which has backward compatibility and is still valid.
https://software.intel.com/en-us/node/525020

boegel:
follow-up in #1718, thanks @ocaisa!

PR #1718, opened 2016-04-08
openmp option deprecated for intel compiler

ocaisa:
`fopenmp` still valid so switch to that, see #1717

### DIFF
--- a/easybuild/toolchains/compiler/inteliccifort.py
+++ b/easybuild/toolchains/compiler/inteliccifort.py
@@ -58,7 +58,7 @@ class IntelIccIfort(Compiler):
         'i8': 'i8',
         'r8': 'r8',
         'optarch': 'xHost',
-        'openmp': 'openmp',  # both -openmp/-fopenmp are valid for enabling OpenMP
+        'openmp': 'fopenmp',  # both -openmp/-fopenmp are valid for enabling OpenMP
         'strict': ['fp-speculation=strict', 'fp-model strict'],
         'precise': ['fp-model precise'],
         'defaultprec': ['ftz', 'fp-speculation=safe', 'fp-model source'],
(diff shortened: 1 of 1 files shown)


boegel:
@ocaisa Can you double-check whether `-fopenmp` is supported properly by the oldest Intel compilers we employ (v11.x)?

Other than that, good to go, thanks for looking into this!

ocaisa:
@boegel I don't have access to sources for v11.x and there's nothing in the documentation I could find about `-fopenmp` so I can't be sure either way

boegel:
@ocaisa I gave it a test myself, `-fopenmp` doesn't work with Intel 11.x...

```
$ icc --version
icc (ICC) 11.1 20100806
Copyright (C) 1985-2010 Intel Corporation.  All rights reserved.

$ icc -fopenmp omp_hello.c -o omp_hello
icc: command line warning #10006: ignoring unknown option '-fopenmp'
omp_hello.c(22): warning #161: unrecognized #pragma
  #pragma omp parallel private(nthreads, tid)
          ^
/tmp/iccRDDbxa.o: In function `main':
omp_hello.c:(.text+0x27): undefined reference to `omp_get_thread_num'
omp_hello.c:(.text+0x40): undefined reference to `omp_get_num_threads'
```

It seems to be OK in v12.x though...

I don't want to spend too much effort on making sure things still work with 11.x, but since we haven't officially deprecated that yet, we should take it into account.

I'll take a look on how to do that in a non-intrusive way; it's not straight-forward since here it's being defined as a part of a constant...

boegel:
@ocaisa see https://github.com/ocaisa/easybuild-framework/pull/36

boegel:
Going in, thanks @ocaisa!

PR #1720, opened 2016-04-13
Add cuda 7.0 entry for the version available on Titan.

uhaehner:

PR #1721, opened 2016-04-14
Add cuda 7.0 entry for the version available on Titan.

uhaehner:


boegel:
Validated by Jenkins (although the message went to #1720 because the same commit is reused).

So going in, thanks @uhaehner!

PR #1722, opened 2016-04-14
make sure user write permissions are set after failed removal attempt

boegel:
Sometimes performing a forced reinstallation fails because the install directory can not be removed due to some files/directories not having write permissions enabled (see for example https://github.com/hpcugent/easybuild-easyconfigs/pull/2879#issuecomment-209831991).

This change should resolve that issue.


boegel:
@wpoely86 please review?

wpoely86:
:+1:

boegel:
fix verified with https://github.com/hpcugent/easybuild-easyconfigs/pull/2879

PR #1730, opened 2016-04-20
remove check whether 'easybuild' is being imported from dir that contains easybuild/__init__.py

boegel:
This reverts #1667, which was needed because of #1593, but is no longer relevant after the use of `fixup_namespace_packages ` proved not be required (cfr. #1680).


boegel:
Jenkins: test this please

PR #1731, opened 2016-04-21
Add read permission to all installed files for everybody

wpoely86:
If nothing is specified, all installed files will have read permissions
for everybody.

edit: we ran into this with VarScan, where the `.jar` was only readable for the install user

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1998,13 +1998,22 @@ class EasyBlock(object):
             perms = stat.S_IWGRP | stat.S_IWOTH
             adjust_permissions(self.installdir, perms, add=False, recursive=True, relative=True, ignore_errors=True)
             self.log.info("Successfully removed write permissions recursively for group/other on install dir.")
-            # add read permissions for everybody on all files
-            perms = (stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH)
-            if build_option('umask') is not None:
-                perms &= build_option('umask')
-                self.log.debug("Taking umask into account %s when adding read permissions to all files.")
+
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1998,6 +1998,10 @@ def permissions_step(self):
             perms = stat.S_IWGRP | stat.S_IWOTH
             adjust_permissions(self.installdir, perms, add=False, recursive=True, relative=True, ignore_errors=True)
             self.log.info("Successfully removed write permissions recursively for group/other on install dir.")
+            # add read permissions for everybody on all files
```
boegel: more like: `make sure read permissions for everyone are in place`, because usually they're already there...

wpoely86: that boils down to exactly the same?

boegel:
Should be fine imho, because that's what you'd usually expect.
It will make the installation a bit slower for large installations though, since this implies `stat`'ing *all* files (again)...

@pescobar, @fgeorgatos, @ocaisa, @rjeschmi: thoughts?

easybuild/framework/easyblock.py
boegel: yeah, fine :)

rjeschmi:
Sounds good to me. I wonder what a system would look like that restricted permissions. You'd probably have to do a lot more throughout the module hierarchy to make it work usefully.

boegel:
@wpoely86 broken tests are fixed with https://github.com/wpoely86/easybuild-framework/pull/20

boegel:
Going in, thanks @wpoely86!

boegel:
oops, didn't mean to close, finger slipped... :)

PR #1733, opened 2016-04-22
add test configuration for Travis

boegel:


boegel:
@wpoely86 please review? ^_^

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+      - git
+      # for GC3Pie (optional dep for EasyBuild)
+      - time
+before_install:
+    # keyring is required to provide GitHub token to EasyBuild;
+    # keyring v5.7.1 is last version to be compatible with py2.6;
+    # for recent versions of keyring, keyrings.alt must be installed too
+    - if [ "x$TRAVIS_PYTHON_VERSION" == 'x2.6' ]; then pip install keyring==5.7.1; else pip install keyring keyrings.alt; fi
+    # optional Python packages for EasyBuild
+    - pip install autopep8 GC3Pie GitPython python-graph-dot python-hglib PyYAML
```
wpoely86: hg also as package? And pep8 itself for the style tests

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+    - if [ ! -z $LMOD_VERSION ]; then source $INSTALL_DEP Lmod-${LMOD_VERSION} $HOME; fi
+    - if [ ! -z $ENV_MOD_TCL_VERSION ]; then source $INSTALL_DEP modules-tcl-${ENV_MOD_TCL_VERSION} $HOME; fi
+script:
+    # set up environment for modules tool (if $MOD_INIT is defined)
+    - if [ ! -z $MOD_INIT ]; then source $MOD_INIT; fi
+    # set up environment for EasyBuild framework tests
+    - export PATH=$TRAVIS_BUILD_DIR:$PATH
+    - export PYTHONPATH=$TRAVIS_BUILD_DIR
+    # install GitHub token
+    - if [ ! -z $GITHUB_TOKEN ]; then
```
wpoely86: let the tests which require a token be skipped if this is not set.

wpoely86:
Looks great

.travis.yml
boegel: seems like hg/Mercurial is just there?

we will add pep8 when we actually need it

.travis.yml
boegel: well, the tests that require a token are skipped if no token is available, see some of the tests that were tweaked

it's not checking `$GITHUB_TOKEN` though, since EasyBuild doesn't pick that up

.travis.yml
boegel: I included `mercurial` in the list of packages to install above.

boegel:

I'll leave both Jenkins and Travis enabled for a while, before we switch to Travis-only...

PR #1735, opened 2016-04-22
stop creating ModulesTool instances over and over again

### LINKED ISSUE #1661: Too many calls to "module use ..."
During a dry run of `Qt` I did a debug log and counted over 1100 calls to `module use` with a HMNS...and the eb command took 3.5 minutes to complete. With a flat naming scheme it was still almost 700 calls and 2.7 minutes. I'm pretty sure that this is because `module use` is pretty expensive for lmod.

Can we reduce the number of calls? Or configure our lmod use so that it's not so expensive? Thoughts @boegel @wpoely86 @rtmclay ?

boegel:
`ModulesTool` used to be a singleton class; back then, it was OK to call `modules_tool()` over and over again.

However, since #1299 `ModulesTool` is no longer a singleton class, and creating `ModulesTool` instances is rather expensive, since there's quite a bit of overhead involved with them (checking modules tool version, making sure `$MODULEPATH` is correct via `module use`/`unuse`, etc.)

This patch fixes that, by creating much less instances of `ModulesTool`: only once during `main`, and then once again for every easyconfig file being installed (which is necessary to avoid reintroducing the bug that was fixed in #1299).

In addition, a thorough cleanup was done in the tests, where `ModulesTool` instances were being created profusely as well...

These changes should result in a significant speedup of EasyBuild overall, and in particular of the test suite (which now runs over 2x faster!).

This fixes #1661.

cc @wpoely86, @ocaisa


boegel:
@wpoely86 please review?

easybuild/tools/robot.py
```diff
@@ -117,7 +116,7 @@ def dry_run(easyconfigs, short=False):
     return '\n'.join(lines)
 
 
-def resolve_dependencies(easyconfigs, retain_all_deps=False):
+def resolve_dependencies(easyconfigs, modtool, retain_all_deps=False):
```
boegel: this change in function signature breaks the easyconfig tests...

easybuild/framework/easyblock.py
```diff
@@ -154,11 +154,6 @@ def __init__(self, ec):
         self.skip = None
         self.module_extra_extensions = ''  # extra stuff for module file required by extensions
 
-        # modules interface with default MODULEPATH
-        self.modules_tool = modules_tool()
-        # module generator
-        self.module_generator = module_generator(self, fake=True)
-
         # modules footer/header
```
wpoely86: maybe move this part to below the `modules_tool`? Not required but it's more of a logical order to me?

easybuild/tools/modules.py
```diff
@@ -192,6 +196,11 @@ def modules(self):
 
     def set_and_check_version(self):
         """Get the module version, and check any requirements"""
+        if self.COMMAND in MODULE_VERSIONS:
+            self.version = MODULE_VERSIONS[self.COMMAND]
+            self.log.info("Found cached version for %s: %s", self.COMMAND, self.version)
```
wpoely86: info? More like debug to me?

easybuild/tools/modules.py
```diff
@@ -115,6 +115,9 @@
 _log = fancylogger.getLogger('modules', fname=False)
 
 
+MODULE_VERSIONS = {}
```
wpoely86: add a comment what this does

easybuild/tools/modules.py
```diff
@@ -192,6 +196,11 @@ def modules(self):
 
     def set_and_check_version(self):
         """Get the module version, and check any requirements"""
+        if self.COMMAND in MODULE_VERSIONS:
+            self.version = MODULE_VERSIONS[self.COMMAND]
+            self.log.info("Found cached version for %s: %s", self.COMMAND, self.version)
+            return
```
wpoely86: I'm glad to see that you also now use inline returns :smiley:

easybuild/tools/modules.py
```diff
@@ -273,13 +284,14 @@ def check_module_function(self, allow_mismatch=False, regex=None):
...
-            for mod_path in self.mod_paths:
-                self.prepend_module_path(mod_path)
-        else:
+        if mod_paths is None:
             # no paths specified, so grab list of (existing) module paths from $MODULEPATH
             self.mod_paths = [p for p in nub(curr_module_paths()) if os.path.exists(p)]
+        else:
+            for mod_path in nub(mod_paths):
+                self.prepend_module_path(mod_path, set_mod_paths=False)
+            self.mod_paths = nub(mod_paths)
```
wpoely86: why this change in the order?

easybuild/tools/modules.py
```diff
@@ -292,24 +304,32 @@ def unuse(self, path):
         """Remove module path via 'module unuse'."""
         self.run_module(['unuse', path])
 
-    def add_module_path(self, path):
+    def add_module_path(self, path, set_mod_paths=True):
```
wpoely86: General remark: document all arguments!

easybuild/tools/modules.py
```diff
@@ -520,6 +543,7 @@ def run_module(self, *args, **kwargs):
 
         full_cmd = ' '.join(cmdlist + args)
         self.log.debug("Running module command '%s' from %s" % (full_cmd, os.getcwd()))
+
         proc = subprocess.Popen(cmdlist + args, stdout=PIPE, stderr=PIPE, env=environ)
```
wpoely86: why not `run_cmd`? Because of the `pipe`?

easybuild/tools/modules.py
```diff
@@ -864,10 +890,14 @@ def update(self):
                 except (IOError, OSError), err:
                     raise EasyBuildError("Failed to update Lmod spider cache %s: %s", cache_fp, err)
 
-    def prepend_module_path(self, path):
+    def prepend_module_path(self, path, set_mod_paths=True):
```
wpoely86: do we really need a different version of this for Tmod and Lmod?

easybuild/tools/modules.py
boegel: positive conditions (`is None`) are easier to comprehend (especially if there's also an else), we try to do this throughout the whole codebase

easybuild/tools/modules.py
boegel: no, because we need `stdout` and `stderr` separately...

easybuild/tools/modules.py
boegel: yes, for Lmod the `unuse` is not required (since `use` pushes the path to the front), so this is an optimised version that only works for Lmod...

easybuild/tools/modules.py
boegel: renamed to `MODULE_VERSION_CACHE` instead, self-explanatory

ocaisa:
Took this for a test drive...it worked and roughly a 2X improvement
https://gist.github.com/ocaisa/d2382ecd9b37bba3a5678fb40ac21477

easybuild/tools/robot.py
boegel: fixed in https://github.com/hpcugent/easybuild-easyconfigs/pull/2952

boegel:
Thanks for the confirmation @ocaisa, and @wpoely86 for the review.

I just gave this another (self-)review, I feel this is good to go.

Likely to be followed up with more improvements, cfr. #1739

PR #1746, opened 2016-04-29
change to install dir rather than buildpath in sanity check of extension, latter may not exist

boegel:
ran into this when installing something with `buildininstalldir = True` on a system where the parent build path in /tmp was not created yet...

we also change to the install dir in the `sanity_check_step` in `easyblock.py`


wpoely86:
:+1:

PR #1749, opened 2016-05-02
translate PyPI download URL to alternate URL with a hash

boegel:
PyPI was changed to use totally differently formatted URLs for new package versions, see also https://bitbucket.org/pypa/pypi/issues/438 .

This change implements auto-translating URLs like https://pypi.python.org/packages/source/e/easybuild/easybuild-2.7.0.tar.gz to https://pypi.python.org/packages/5b/03/e135b19fadeb9b1ccb45eac9f60ca2dc3afe72d099f6bd84e03cb131f9bf/easybuild-2.7.0.tar.gz .

This new style of URLs works for both old and new versions of Python packages, so this change should be transparent.

That is, until they utterly break things again, of course...

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -625,6 +625,8 @@ class EasyBlock(object):
                         self.log.warning("Source URL %s is of unknown type, so ignoring it." % url)
                         continue
 
+                    # PyPI URLs may need to be converted due to change in format of these URLs,
+                    # cfr. https://bitbucket.org/pypa/pypi/issues/438
                     if PYPI_PKG_URL_PATTERN in fullurl and not is_alt_pypi_url(fullurl):
                         alt_url = derive_alt_pypi_url(fullurl)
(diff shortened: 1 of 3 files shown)


boegel:
@wpoely86 please review?

boegel:
cc @pescobar

wpoely86:
:+1:

boegel:
slightly reworked to parse `/simple/` HTML page using `xml` module, rather than 'parsing' it directly, as recommended at https://wiki.python.org/moin/PyPISimple

I also considered using the alternate JSON (https://wiki.python.org/moin/PyPIJSON) or XML-RPC (https://wiki.python.org/moin/PyPIXmlRpc) APIs, but these are less straightforward to use in this particular case because we don't have the package version readily available (unless we rework more of the code); we can easily extract the package name from the available URL though, which is sufficient for the Simple API.

boegel:

PR #1086, opened 2014-11-05
fix HierarchicalMNS for CUDA with iccifort, push comp_versions definition out of for loop

boegel:

PR #1753, opened 2016-05-07
hierarchical_mns.py: fix backtrace with CUDA

thiell:
Hierarchical MNS is broken with CUDA easyconfigs. This patch fixes a backtrace and installs CUDA in Core. We have this patch since a while but I forgot to submit a pull request before.

Without the patch:

```
$ eb CUDA-7.5.18.eb  --robot $PWD
== temporary log file in case of crash /lstor/sthiell/38870/eb-SjcWO4/easybuild-1o9wBR.log
== resolving dependencies ...
== processing EasyBuild easyconfig /global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_easyconfigs-2.7.0-py2.6.egg/easybuild/easyconfigs/c/CUDA/CUDA-7.5.18.eb
== building and installing Core/CUDA/7.5.18...
== fetching files...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== testing...
== installing...
== postprocessing...
== sanity checking...
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/main.py", line 113, in build_and_install_software
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyblock.py", line 2292, in build_and_install_one
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyblock.py", line 2208, in run_all_steps
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyblock.py", line 2087, in run_step
    step_method(self)()
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_easyblocks-2.7.0-py2.6.egg/easybuild/easyblocks/c/cuda.py", line 137, in sanity_check_step
    super(EB_CUDA, self).sanity_check_step(custom_paths=custom_paths)
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyblock.py", line 1731, in sanity_check_step
    self._sanity_check_step(*args, **kwargs)
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyblock.py", line 1851, in _sanity_check_step
    fake_mod_data = self.load_fake_module(purge=True)
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyblock.py", line 1148, in load_fake_module
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyblock.py", line 1938, in make_module_step
    txt += self.make_module_extend_modpath()
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyblock.py", line 1044, in make_module_extend_modpath
    modpath_exts = ActiveMNS().det_modpath_extensions(self.cfg)
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/framework/easyconfig/easyconfig.py", line 1517, in det_modpath_extensions
    modpath_extensions = self.mns.det_modpath_extensions(self.check_ec_type(ec))
  File "/global/easytests/software/Core/EasyBuild/2.7.0/lib/python2.6/site-packages/easybuild_framework-2.7.0-py2.6.egg/easybuild/tools/module_naming_scheme/hierarchical_mns.py", line 177, in det_modpath_extensions
KeyError: 'GCC'
```


boegel:
Hmm, just removing the `CUDA` & `GCC` combo doesn't seem like a good idea to me...

How does this relate to #1086?

thiell:
@boegel: ok but this code is broken and hard to understand to be honest. I feel like having a dependency on gcc adds a lot of complexity as nvcc has its own checks in headers, so it's clear for the user if there is a mismatch. Also, at our site, a same CUDA installation works with several toolchains based on GCC, and it's easier, I guess.

#1086 is related to this yes, but also deals with CUDA-aware MPI which is another thing.

ocaisa:
We had issues with this too, in the end we felt it was too much trouble to add a CUDA layer before the MPI layer in a HMNS and we just did both in one shot (compiler -> compiler+CUDA+MPI) to keep the hierarchy. We hacked a custom MNS to side-step the problem you ran in to (which also introduced some other different conventions).

boegel:
@thiell I understand your point, but just removing that entry doesn't seem like a good approach to fixing the problem...

Shouldn't we come up with a way to fix the `KeyError` occurring in `det_modpath_extensions` instead, maybe just catching/ignoring it?

That's basically what's being done in https://github.com/hpcugent/easybuild-framework/pull/1086/files#diff-f24b21ac5017831f5937b3acaba305aeR169; there was a remark on that 'fix' which ended up blocking the whole PR though...

So, maybe we should reconsider that part (separate from the CUDA-aware MPI and CUDA+intel aspects), and work out a fix?

I'm actually quite OK with just ignoring the `KeyError` like we're doing there, although we should probably log stuff.

How about something replacing:

```python
```

with

```python
if comp_versions  in comp_ver_tmpl:
    break
```

Can you check if that would fix your problem? If so, we can let our bot/Travis check whether it break anything else (it shouldn't, imho).

thiell:
Your code as is doesn't work but I got the idea, so here is this new patch proposal that will raise a handled exception when CUDA is installed against the dummy toolchain, and also compiler versions are built from a list of compiler keys.

easybuild/tools/module_naming_scheme/hierarchical_mns.py
```diff
@@ -47,10 +47,10 @@
...
 # note: names in keys are ordered alphabetically
 COMP_NAME_VERSION_TEMPLATES = {
-    'icc,ifort': ('intel', '%(icc)s'),
-    'Clang,GCC': ('Clang-GCC', '%(Clang)s-%(GCC)s'),
-    'CUDA,GCC': ('GCC-CUDA', '%(GCC)s-%(CUDA)s'),
-    'xlc,xlf': ('xlcxlf', '%(xlc)s'),
+    'icc,ifort': ('intel', ('icc',)),
+    'Clang,GCC': ('Clang-GCC', ('Clang', 'GCC')),
+    'CUDA,GCC': ('GCC-CUDA', ('GCC', 'CUDA')),
+    'xlc,xlf': ('xlcxlf', ('xlc',)),
```
boegel: @thiell I think we should try and avoid changing the format here, since other module naming schemes may be relying on this...

The only reason you did this is to have access to the names used to resolve the templates with, and make sure they're all available before resolving the templates, right?

In that case, why not use the names in the `key.split(',')` instead? Then there's no need to change the format here?

boegel:
@thiell please take a look at https://github.com/thiell/easybuild-framework/pull/1, which tackles my last remark by grabbing the required keys from the template using a regex (`re.findall(...)`), and add a test case for CUDA w/ `dummy`

The test case I added now fails with:

```
ERROR: test_hierarchical_mns (__main__.LuaModuleGeneratorTest)
Test hierarchical module naming scheme.
  File "/Users/kehoste/work/easybuild-framework/test/framework/module_generator.py", line 675, in test_hierarchical_mns
    test_ec(ecfile, *mns_vals)
  File "/Users/kehoste/work/easybuild-framework/test/framework/module_generator.py", line 641, in test_ec
    self.assertEqual(ActiveMNS().det_modpath_extensions(ec), modpath_exts)
  File "easybuild/framework/easyconfig/easyconfig.py", line 1519, in det_modpath_extensions
    modpath_extensions = self.mns.det_modpath_extensions(self.check_ec_type(ec))
  File "easybuild/tools/module_naming_scheme/hierarchical_mns.py", line 187, in det_modpath_extensions
    ec['toolchain'], ec['name'], ec['version'])
EasyBuildError: "Required compilers not available in toolchain {'version': 'dummy', 'name': 'dummy'} for CUDA v5.5.22"
```

Does it make sense to retain this, i.e. to not support installing CUDA in `Core`?
See also question on the EB mailing list today: https://lists.ugent.be/wws/arc/easybuild/2016-05/msg00044.html

boegel:
@thiell ping on the question w.r.t. not supporting `CUDA` in `Core`? I feel that should be supported, no?

boegel:
>> @thiell ping on the question w.r.t. not supporting CUDA in Core? I feel that should be supported, no?

reping @thiell?

akesandgren:
An shorter fix for this is to combine PR #1982 with adding
'CUDA': ('CUDA', '%(CUDA)s'),
to COMP_NAME_VERSION_TEMPLATES in module_naming_scheme/hierarchical_mns.py

That seems to do the right thing. @boegel any thoughts on that?

thiell:
Sorry for the late reply! We have a GPU cluster with hierarchical mns and CUDA in Core (with the patch from this PR) because we don't want CUDA library to be integrated as part of a compiler toolchain. CUDA is very easy to upgrade when in Core but when it's part of a toolchain, it's a long process to rebuild and for users to adopt the new toolchain. Also we think explicit is better than implicit in that case.

One other thing, hierarchical mns has been complicated for our users, especially to understand the concept of hidden modules. I don't think that we will use it again.

boegel:
@thiell I guess that motivation makes sense...

I looked into what was needed to finally make this ready to be merged via a wrapper PR, see #1986.
It basically boils down to making an exception in `det_module_subdir` for CUDA installed with a `dummy` toolchain (I'm not sure how that could've worked for you with the version of `hierarchical_mns.py` in this PR...).
I also added some comments and renamed some variables to hopefully make the implementation a bit easier to digest.

If you're attending SC16, you should definitely try and meet up with @damianam to discuss the issues with HMNS your users have been struggling with, since they have a very similar setup. See also Damian's talk at HUST16 (http://hust16.github.io/#program, paper available at http://hpcugent.github.io/easybuild/files/eb-jsc-hust16.pdf).

boegel:
Merged via #1986; GitHub should realise that, but it doesn't because of the last merge commit.

All relevant commits (those with actual code changes) are included in #1986 though,so I'm going to close this one.

PR #1754, opened 2016-05-09
only load modules using short module names

boegel:
This fixes a long-standing bug, which has remain hidden up until now because it only causes actual problems in particular circumstances, e.g. when running the sanity check commands during the sanity check of installing OpenMPI with a PGI toolchain when using a hierarchical module naming scheme, cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/2899. The combination of sanity check commands with a toolchain that actually corresponds to a software package (rather than just a bundle of other modules) and a hierarchical module naming scheme is required to trigger this bug...

The problem is that the temporary module that is generated and loaded at the start of the sanity check was being loaded with  the *full* name (e.g. `Compiler/GCC/4.7.2/OpenMPI/1.6.4`) rather than the *short* name (e.g. `OpenMPI/1.6.4`).

The latter requires that the module for the toolchain used to perform must be loaded first, to make sure that `$MODULEPATH` is extended properly for the load with the short name to be work, and it's exactly this missing load that was causing problems when performing the sanity check commands for OpenMPI installed on top of PGI, i.e. that the PGI compiler commands (e.g. `pgcc`) were not available in `$PATH` (because the `PGI` module wasn't actually loaded during the sanity check).

This change fixes this problem, by making sure that all `loads` performed by EasyBuild are done with the *short* module name and that `$MODULEPATH` is prepended as needed to make the temporary (fake) module loadable.

(WIP because I want to look into a dedicated test to catch this...)

cc @bartoldeman, @ocaisa, @damianam, @geimer

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -357,7 +357,13 @@ class Toolchain(object):
     def add_dependencies(self, dependencies):
         """ Verify if the given dependencies exist and add them """
         self.log.debug("add_dependencies: adding toolchain dependencies %s" % dependencies)
-        dep_mod_names = [dep['short_mod_name'] for dep in dependencies]
+
+        # use *full* module name to check existence of dependencies, since the modules may not be avaialble in the
+        # current $MODULEPATH without loading the prior dependencies in a module hierarchy
+        # (e.g. OpenMPI module may only be available after loading GCC module);
+        # when actually loading the modules for the dependencies, the *short* module name is used,
+        # see _load_dependencies_modules()
+        dep_mod_names = [dep['full_mod_name'] for dep in dependencies]
 
         # check whether modules exist
         if self.dry_run:
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -357,7 +357,7 @@ class Toolchain(object):
     def add_dependencies(self, dependencies):
         """ Verify if the given dependencies exist and add them """
         self.log.debug("add_dependencies: adding toolchain dependencies %s" % dependencies)
-        dep_mod_names = [dep['full_mod_name'] for dep in dependencies]
+        dep_mod_names = [dep['short_mod_name'] for dep in dependencies]
 
         # check whether modules exist
         if self.dry_run:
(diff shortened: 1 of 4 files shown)


easybuild/tools/toolchain/toolchain.py
```diff
@@ -357,6 +357,12 @@ def get_dependency_version(self, dependency):
     def add_dependencies(self, dependencies):
         """ Verify if the given dependencies exist and add them """
         self.log.debug("add_dependencies: adding toolchain dependencies %s" % dependencies)
+
+        # use *full* module name to check existence of dependencies, since the modules may not be avaialble in the
```
geimer: Nitpicking: avaialble? ;-)

boegel: nice catch :)

boegel:
dedicated unit test added that triggers the bug, i.e. without the fix implemented here, it fails with:

```
ERROR: test_toy_sanity_check_commands (__main__.ToyBuildTest)
Test toy build with extra sanity check commands.
  File "test/framework/toy_build.py", line 1115, in test_toy_sanity_check_commands
    self.eb_main(args, logfile=self.dummylogfn, do_build=True, verbose=True, raise_error=True)
  File "/Users/kehoste/work/easybuild-framework/test/framework/utilities.py", line 276, in eb_main
    raise myerr
EasyBuildError: "Build of /tmp/eb-95zh3u/eb-EdLoTc/toy-0.0-tweaked.eb failed
(err: 'build failed (first 300 chars): Sanity check failed:
sanity check command env | grep EBROOTGCC  exited with code 1 (output: ),
sanity check command env | grep EBROOTGOOLF  exited with code 1 (output: )')"

```

wpoely86:
:+1:

boegel:
This is good to go imho; @geimer took a look at it and didn't seem to have any objections (other than a silly typo), @wpoely86 gave it thumbs up and @bartoldeman confirmed that this fixes the problem he was having with https://github.com/hpcugent/easybuild-easyconfigs/issues/2899 (which is confirmed with the test report that uses `HierarchicalMNS`)

I'm still somewhat surprised that this didn't pop up earlier, but it only comes forward under very specific circumstances...

PR #1755, opened 2016-05-10
set $LMOD_REDIRECT to 'no' when initialising Lmod

boegel:
cc @wpoely86 

@rtmclay: always setting to `$LMOD_REDIRECT` to `no` to avoid Lmod redirecting output to `stdout` by issuing `echo` commands makes sense, right?

In EasyBuild, we *want* output messages to go to stderr since then we can separate them from the changes to the environment easily...

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -887,6 +887,8 @@ class Lmod(ModulesTool):
         os.environ['LMOD_QUIET'] = '1'
         # make sure Lmod ignores the spider cache ($LMOD_IGNORE_CACHE supported since Lmod 5.2)
         os.environ['LMOD_IGNORE_CACHE'] = '1'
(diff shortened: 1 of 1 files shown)


wpoely86:
:+1:

boegel:
I don't see how this can break anything, tests are happy, and this is required if Lmod is configured to do redirection to avoid EB tipping over...

So, going in, thanks for the review @wpoely86!

(@rtmclay: please don't hesitate to drop in a comment even post-merge with your opinion on this)

rtmclay:
Yes, setting LMOD_REDIRECT=no will force Lmod to send module list etc to stderr independent of whether Lmod is configured to redirect or not.

boegel:
Thanks for the feedback @rtmclay!

Issue #1756, opened 2016-05-11
Have a unit test against borken `module use`


fgeorgatos:

boegel:
```
module use /tmp/modules
echo $MODULEPATH
bash
echo $MODULEPATH
```

fgeorgatos:
example of the bug against recent fedora, patched last month:
https://bugzilla.redhat.com/show_bug.cgi?id=1326075

fgeorgatos:
Example of the bug:
```
[root@bright-lmod ~]# wget ftp://195.220.108.108/linux/epel/6/x86_64/Lmod-6.0.15-2.el6.noarch.rpm
[...]
[root@bright-lmod ~]# md5sum Lmod-6.0.15-2.el6.noarch.rpm
36d074c7875a02d32f227abb4a553a78  Lmod-6.0.15-2.el6.noarch.rpm
[root@bright-lmod ~]# yum localinstall !$
yum localinstall Lmod-6.0.15-2.el6.noarch.rpm
[...]
Installed:
  Lmod.noarch 0:6.0.15-2.el6

Complete!
[root@bright-lmod ~]# su - fgeorgatos
Last login: Fri May  6 14:32:36 CEST 2016 from 46.17.164.10 on pts/3
[fgeorgatos@bright-lmod ~]$ echo $MODULEPATH
/etc/modulefiles:/usr/share/modulefiles:/usr/share/Modules/modulefiles:/usr/share/modulefiles/Linux:/usr/share/modulefiles/Core:/usr/share/lmod/lmod/modulefiles/Core
[fgeorgatos@bright-lmod ~]$ module use $HOME/.local/easybuild/modules.all
[fgeorgatos@bright-lmod ~]$ echo $MODULEPATH
/home/fgeorgatos/.local/easybuild/modules.all:/etc/modulefiles:/usr/share/modulefiles:/usr/share/Modules/modulefiles:/usr/share/modulefiles/Linux:/usr/share/modulefiles/Core:/usr/share/lmod/lmod/modulefiles/Core
[fgeorgatos@bright-lmod ~]$ bash
[fgeorgatos@bright-lmod ~]$ echo $MODULEPATH
/etc/modulefiles:/usr/share/modulefiles:/usr/share/Modules/modulefiles
[fgeorgatos@bright-lmod ~]$
```

btw. the bug is expected to have vanished for most new installs, since the fresher 6.3.1 rollouts:
[...]

PR #1758, opened 2016-05-11
add test for broken modules tool setup affecting 'module use'

boegel:
fix for #1756

### DIFF
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -45,6 +45,7 @@ from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import mkdir, read_file, write_file
 from easybuild.tools.modules import Lmod, get_software_root, get_software_version, get_software_libdir
 from easybuild.tools.modules import invalidate_module_caches_for, modules_tool
+from easybuild.tools.run import run_cmd
 
 
 # number of modules included for testing purposes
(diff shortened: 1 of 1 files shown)


boegel:
@fgeorgatos please review?

test/framework/modules.py
```diff
@@ -465,6 +466,23 @@ def test_module_caches(self):
...
+        # this test is here as check for a nasty bug in how the modules tool is deployed
+        # cfr. https://github.com/hpcugent/easybuild-framework/issues/1756,
+        # https://bugzilla.redhat.com/show_bug.cgi?id=1326075
+        modules_dir = os.path.abspath(os.path.join(self.test_prefix, 'modules'))
+        self.assertFalse(modules_dir in os.environ['MODULEPATH'])
+
+        mkdir(modules_dir, parents=True)
+        self.modtool.use(modules_dir)
+        modulepath = os.environ['MODULEPATH']
+        self.assertTrue(modules_dir in modulepath)
```
fgeorgatos: nice

test/framework/modules.py
```diff
@@ -465,6 +466,23 @@ def test_module_caches(self):
...
+
+        out, _ = run_cmd("bash -c 'echo MODULEPATH: $MODULEPATH'", simple=False)
+        self.assertEqual(out.strip(), "MODULEPATH: %s" % modulepath)
```
fgeorgatos: nice

test/framework/modules.py
```diff
@@ -465,6 +466,23 @@ def test_module_caches(self):
...
+        self.assertTrue(modules_dir in out)
```
fgeorgatos: strictly speaking, I think this is redundant, no? (I mean, how would you pass the above 2 and this fails).
I'm curious if you have a particular failure mode in mind or, it's just verbose because it's testing...

otherwise really good, ready for merging! that was fast!

boegel: You're right, it's redundant, but it makes it very explicit what we're checking for exactly.

Doesn't hurt to have it, it's not like it's going to slow down the test noticeably. ;-)

boegel:
Thanks for the feedback @fgeorgatos, going in!

PR #1761, opened 2016-05-12
include '-ldl' in $LIBBLAS*/* & co when linking with Intel MKL 11.x or newer

boegel:
This change results in `-lm -ldl` being added to `$LIBBLAS`, `$LIBLAPACK`, etc. for Intel MKL 11.0 and newer, following the discussion in https://github.com/hpcugent/easybuild-easyconfigs/pull/2900, and as instructed by the Intel MKL Link Advisor.

This is required in particular for building HPL with PGI, cfr. https://github.com/hpcugent/easybuild-easyconfigs/issues/3046.

### DIFF
--- a/easybuild/toolchains/linalg/intelmkl.py
+++ b/easybuild/toolchains/linalg/intelmkl.py
@@ -83,17 +83,17 @@ class IntelMKL(LinAlg):
     def set_variables(self):
         """Set the variables"""
 
-        # for recent versions of Intel MKL, -lm -ldl should be used for linking
-        # according to the Intel MKL Link Advisor, this should always be done,
-        # but it only seems to be strictly required for certain compilers
+        # for recent versions of Intel MKL, -ldl should be used for linking
+        # the Intel MKL Link Advisor specifies to always do this,
+        # but it is only needed when statically linked with Intel MKL,
+        # and only strictly needed for some compilers (e.g. PGI)
         mkl_version = self.get_software_version(self.BLAS_MODULE_NAME)[0]
         if LooseVersion(mkl_version) >= LooseVersion('11'):
-        #if LooseVersion(mkl_version) >= LooseVersion('11') and self.COMPILER_FAMILY in [TC_CONSTANT_PGI]:
-            math_libs = ['m', 'dl']
+            self.log.info("Adding -ldl as extra library when linking with Intel MKL libraries (for v11.x and newer)")
             if self.LIB_EXTRA is None:
-                self.LIB_EXTRA = math_libs
-            else:
-                self.LIB_EXTRA.extend([l for l in math_libs if l not in self.LIB_EXTRA])
+                self.LIB_EXTRA = ['dl']
+            elif 'dl' not in self.LIB_EXTRA:
+                self.LIB_EXTRA.append('dl')
 
         super(IntelMKL, self).set_variables()
 
--- a/easybuild/toolchains/linalg/intelmkl.py
+++ b/easybuild/toolchains/linalg/intelmkl.py
@@ -84,13 +84,16 @@ class IntelMKL(LinAlg):
         """Set the variables"""
 
         # for recent versions of Intel MKL, -lm -ldl should be used for linking
+        # according to the Intel MKL Link Advisor, this should always be done,
+        # but it only seems to be strictly required for certain compilers
         mkl_version = self.get_software_version(self.BLAS_MODULE_NAME)[0]
         if LooseVersion(mkl_version) >= LooseVersion('11'):
+        #if LooseVersion(mkl_version) >= LooseVersion('11') and self.COMPILER_FAMILY in [TC_CONSTANT_PGI]:
             math_libs = ['m', 'dl']
-            if self.LIB_MATH is None:
-                self.LIB_MATH = math_libs
+            if self.LIB_EXTRA is None:
+                self.LIB_EXTRA = math_libs
             else:
-                self.LIB_MATH.extend([l for l in math_libs if l not in self.LIB_MATH])
+                self.LIB_EXTRA.extend([l for l in math_libs if l not in self.LIB_EXTRA])
 
         super(IntelMKL, self).set_variables()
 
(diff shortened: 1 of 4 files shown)


easybuild/toolchains/linalg/intelmkl.py
```diff
@@ -80,6 +80,23 @@ def __init__(self, *args, **kwargs):
...
 
+    def set_variables(self):
+        """Set the variables"""
+
+        # for recent versions of Intel MKL, -lm -ldl should be used for linking
+        # according to the Intel MKL Link Advisor, this should always be done,
+        # but it only seems to be strictly required for certain compilers
+        mkl_version = self.get_software_version(self.BLAS_MODULE_NAME)[0]
+        if LooseVersion(mkl_version) >= LooseVersion('11'):
+        #if LooseVersion(mkl_version) >= LooseVersion('11') and self.COMPILER_FAMILY in [TC_CONSTANT_PGI]:
```
boegel: I'm a little bit reluctant to do this for *any* compiler, but since this is what is advised by the Intel MKL Link Advisor, it should be fine I think.

Thoughts, @bartoldeman, @ocaisa, @damianam, @wpoely86?

wpoely86: if it's not needed, the linker will discard them?

boegel: Sure, but my biggest concern is: what if they're not around? I'm not sure if that's realistic though, they're probably always there...

wpoely86: The math library and the dynamic linking library? If these are not around, we will have bigger concerns that this. Sure, `glibc-devel` is needed but that's a requirement anyway.

boegel: Why is `glibc-devel` a requirement already (just to be clear)?

wpoely86:
```
$ rpm -qR gcc-5.3.1-6.fc23.x86_64
/bin/sh
/bin/sh
/bin/sh
/sbin/install-info
/sbin/install-info
binutils >= 2.24
cpp = 5.3.1-6.fc23
glibc-devel >= 2.2.90-12
...
Like for example `libpthread.so` is also in there.
```

bartoldeman:
Conservatively I think it's only really needed for PGI + static linking only, and then only -ldl (not -lm).

Progressively the link advisor also suggests -lpthread (though perhaps not everywhere?). Either way I don't have a strong opinion on this, whatever works works...

boegel:
I did see this pop up, when compiling something with the Intel compilers that was using `$LIBBLAS` in a funny place (quite early in the link line):

```
ifort: warning #10315: specifying -lm before files may supercede the Intel(R) math library and affect performance
```

That makes me want to reconsider either doing this for all compilers or only slapping on `-ldl`.

@bartoldeman why would `-lm` not be required?

bartoldeman:
lm and lpthread are not required for PGI because they are already included by pgcc:

```
pgcc -v hello.c
...
/software/CentOS-6/eb/software/Toolchain/GCCcore/4.9.3/binutils/2.25/bin/ld /usr/lib64/crt1.o /usr/lib64/crti.o /sb/software/CentOS-6/eb/software/Core/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/lib/trace_init.o /sb/software/CentOS-6/eb/software/Core/GCCcore/4.9.3/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.9.3/crtbegin.o /sb/software/CentOS-6/eb/software/Core/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/lib/initmp.o --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 /sb/software/CentOS-6/eb/software/Core/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/lib/pgi.ld -L/sb/software/CentOS-6/eb/software/Core/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/lib -L/usr/lib64 -L/sb/software/CentOS-6/eb/software/Core/GCCcore/4.9.3/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.9.3 /tmp/pgccspCb24NhjivM.o -rpath /sb/software/CentOS-6/eb/software/Core/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/lib -lpgmp -lnuma -lpthread -lnspgc -lpgc -lm -lgcc -lc -lgcc /sb/software/CentOS-6/eb/software/Core/GCCcore/4.9.3/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.9.3/crtend.o /usr/lib64/crtn.o
```

bartoldeman:
Similarly, icc includes -lm and -ldl by default.

boegel:
So, is PGI not including `-ldl` a bug?

Anyway, it seems like w can switch this to only including `-ldl` for PGI, and maybe *always* (not just with Intel MKL), similar to how we consistently include `-lm` for GCC (which may only be needed for older versions of GCC): https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/toolchains/compiler/gcc.py#L89

bartoldeman:
No it's not a bug, not including -ldl. Many programs compile without using -ldl.

It's just static MKL that needs it.

It will probably not hurt to add -ldl everywhere (as icc does it by default for some reason) but from a minimalistic point of view it is not necessary. I'm tending to favour the most minimalistic change (only use -ldl for statically linking MKL).

boegel:
@bartoldeman OK, thanks for the feedback, I changed things to only slap on `-ldl`.

boegel:
fix verified with https://github.com/hpcugent/easybuild-easyconfigs/pull/3044 and https://github.com/hpcugent/easybuild-easyconfigs/pull/3046

boegel:
Good to go now, thanks for the feedback @bartoldeman and @wpoely86!

bartoldeman:
Thanks, I can confirm it works.

boegel:
I had to make the inclusion of `-ldl` conditional on using PGI after all, since it breaking backwards compatibility, cfr. https://github.com/hpcugent/easybuild-framework/pull/1764

PR #1763, opened 2016-05-13
fix modpath_extensions_for method: take into account modules in Lua syntax and conditional $MODULEPATH extensions (WIP)

boegel:
wrapper PR for #1474 by @t-brown


boegel:
@geimer: can you give this a quick review? it's not finished yet, a couple of tests are broken it seems so I may have missed something (or those tests are broken themselves somehow ;)), but the overall approach is there, i.e. use the output from '`lmod python load <mod>`' rather than reading the module file directly

not only does this help in supporting both Tcl/Lua modules, it also deals with conditional extensions (guarded by a check on whether the directory exists, which is done in modules generated when using `--subdir-user-modules`)

boegel:
@damianam, @ocaisa: I'm a bit puzzled as to why you're not hit by this...
Are you using module files in Lua syntax together with a hierarchical module naming scheme?

In particular, are the modules that extend `$MODULEPATH` (i.e. `Core/*` or `Compiler/*`) in Lua syntax?

easybuild/tools/modules.py
```diff
@@ -681,13 +681,42 @@ def modpath_extensions_for(self, mod_names):
...
+            # module must be available to (fake) load
+            if not self.exist([mod_name])[0]:
+                raise EasyBuildError("Can not determine $MODULEPATH extensions for non-existing module %s", mod_name)
+
+            # grab current value for $MODULEPATH, so we can figure out which paths get added to it
+            curr_modulepath = [p for p in os.environ.get('MODULEPATH', '').split(os.pathsep) if p]
+
+            # fake load for module: just obtain changes that would be made to environment, don't apply them (yet)
+            # this is prefrred over parsing the module file contents, to take into account conditional extensions
+            mod_load_txt = self.run_module('load', mod_name, return_output=True)
```
boegel: I switched to a 'fake load' rather than using `module show` (as was done in #1474) to dance around the fact that `modulecmd.tcl` is not showing changes to `$MODULEPATH` if the module file includes `module use` statements, although it *does* properly extend `$MODULEPATH` when the module is loaded...

This approach was taken over just 'parsing' the module file itself because some of the `module use` statements may be conditional, i..e when `--subdir-user-modules` is used (cfr. `make_module_extend_modpath`); also, it makes parsing easier, since the output is the same regardless of Tcl or Lua module files.

However, neither the `show` nor the 'fake load' approach work, because they're too dependent on the environment in which they're being run...

It's possible that the `$MODULEPATH` extensions performed by a module were already done in one way or another in the current environment, which will cause us to miss extensions; resetting `$MODULEPATH` is not going to work out, since then we may break the `load` of not only the module being inspected, but also of `load` statements included in that module...

Also, strictly speaking this implementation gives wrong results for `icc`/`ifort` (cfr. new `modpath_extensions_for` test), since it'll result in an empty list of `$MODULEPATH` extensions for the `ifort` module, because the `icc` module that contains the same extensions will be loaded by the time the `ifort` module is inspected.

I need to think about a proper way to deal with all of this... It seems like we'll need to go back to parsing the actual module file, and somehow figure out ourselves whether the `use` statements are conditional... Or, maybe we don't need to bother: does it matter whether or not extensions that may not actually happen are included?

@geimer: thoughts?

boegel: PR updated, I reverted to actually parsing the module file itself, after figuring out that including `$MODULEPATH` extensions that may not actually happen at runtime due to a condition not being met is fine...

boegel:
nice, I managed to screw up this PR with a force push... >_<

new PR in #1766

PR #1774, opened 2016-05-20
Expand ~ in include paths

tobbez:
Paths/globs specified to e.g. `--include-easyblocks` are currently not expanded if they reference a home directory using ~ (e.g. `--include-easyblocks=~/eb/block.py` or `--include-easyblocks=~user/eb/block.py`).

This can get a bit confusing since specifying the path in the same way using the environment variable (`EASYBUILD_INCLUDE_EASYBLOCKS`) works, because the ~ is expanded by the shell in that case.


boegel:
@tobbez patch looks great, thanks!

To make this complete, we should add a test in `test/framework/filetools.py` to make sure we don't break this in the future, and maybe also in `test/framework/include.py` to make sure `--include-*` keeps honouring it

Please let us know if you're up for this; if not, I can help out.

boegel:
@tobbez I implemented tests that check for this change in https://github.com/tobbez/easybuild-framework/pull/1, please review/merge.

tobbez:
@boegel Done.

boegel:
@tobbez thanks! Travis is reporting that one of the new tests fails, due to an oversight by me, which is fixed by https://github.com/tobbez/easybuild-framework/pull/2, can you please merge that too?

Then this is good to go, I'll include it in the upcoming EasyBuild v2.8.1 (which will be released because of fixing https://github.com/hpcugent/easybuild-framework/issues/1775).

boegel:
Good to go, thanks @tobbez!

Issue #1775, opened 2016-05-20
fatal error Invalidating the module cache with 2.8


brettbode:
Hi,
   I am hitting a Python Error and traceback at the end of creating a module with EasyBuild 2.8. I retested and 2.7 doesn't hit the issue so it appears to be new in 2.8. The package and module seem to build fine and indeed it doesn't appear to depend on what software you are actually building as the error happens after the module is created. This is on a Cray XE system with the traditional modules install. Below is the end of the debug log. The full debug log is attached and the traceback is below. I can get past it by commenting out the invalidate cache code, which occasionally means the cache is out of date, but otherwise works. However, I am not sure how to fix it properly.

Thanks,
Brett

== 2016-05-20 13:02:18,852 build_log.py:152 ERROR Traceback (most recent call last):
  File "/mnt/abc/u/staff/bbode/EasyBuild-2.8/software/EasyBuild/2.8.0/lib64/python2.6/site-packages/easybuild_framework-2.8.0-py2.6.egg/easybuild/main.py", line 114, in build_and_install_software
    (ec_res['success'], app_log, err) = build_and_install_one(ec, init_env)
  File "/u/staff/bbode/EasyBuild-2.8/software/EasyBuild/2.8.0/lib64/python2.6/site-packages/easybuild_framework-2.8.0-py2.6.egg/easybuild/framework/easyblock.py", line 2358, in build_and_install_one
    result = app.run_all_steps(run_test_cases=run_test_cases)
  File "/u/staff/bbode/EasyBuild-2.8/software/EasyBuild/2.8.0/lib64/python2.6/site-packages/easybuild_framework-2.8.0-py2.6.egg/easybuild/framework/easyblock.py", line 2274, in run_all_steps
    self.run_step(step_name, step_methods)
  File "/u/staff/bbode/EasyBuild-2.8/software/EasyBuild/2.8.0/lib64/python2.6/site-packages/easybuild_framework-2.8.0-py2.6.egg/easybuild/framework/easyblock.py", line 2153, in run_step
    step_method(self)()
  File "/u/staff/bbode/EasyBuild-2.8/software/EasyBuild/2.8.0/lib64/python2.6/site-packages/easybuild_framework-2.8.0-py2.6.egg/easybuild/framework/easyblock.py", line 2005, in make_module_step
    invalidate_module_caches_for(path)
[...]

PR #1777, opened 2016-05-22
avoid changing $MODULEPATH when prepending with symlink of path already at head of $MODULEPATH

boegel:
Although this should be fine in theory, it sometimes has repercussions in practice, even beyond performance, cfr. #1775.


boegel:
@wpoely86 please review?

wpoely86:
:+1:

boegel:

PR #1779, opened 2016-05-25
filter out duplicates in find_flexlm_license

boegel:
This is probably enough to fix the regression reported in https://github.com/hpcugent/easybuild-easyblocks/issues/933, but it seems like there's more going on (in `IntelBase`, that is).

@damianam: you're somewhat familiar with this, please review? ;)

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1231,7 +1231,8 @@ def find_flexlm_license(custom_env_vars=None, lic_specs=None):
     Find FlexLM license.
 
     Considered specified list of environment variables;
-    checks for path to existing license file or valid license server specification.
+    checks for path to existing license file or valid license server specification;
+    duplicate paths are not retained in the returned list of license specs.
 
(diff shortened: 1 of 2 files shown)


boegel:
awaiting feedback from @MalkinE to see whether this helps with the reported issue

MalkinE:
This fixes the mentioned issue. Thanks @boegel!

damianam:
LGTM. But anybody reading this should probably know that this fix hides a potentially weirder problem (on Intel's side or in particular user configurations, where exactly is not clear) on the installation of intel tools. In particular, `exist_lic` should work with `$INTEL_LICENSE_FILE`, without further specifying a license file entry in `silent.cfg`.

This fix worked in the reported issue because with it we simply avoid setting `exist_lic` (unnecessarily) in favour of `license_server` or `license_file`.

boegel:
Details on the latter are available at https://github.com/hpcugent/easybuild-easyblocks/issues/933

Thanks for the review @damianam, and @MalkinE for confirming the fix!

PR #1783, opened 2016-05-29
don't use LooseVersion to define version_major/version_minor

boegel:
don't use `LooseVersion` to define `version_major` and especially `version_minor` template values, since a verson like `4.09` isn't handled correctly in that case; we expect `version_minor` to be `09` rather than `9`

cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/3141

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -171,13 +171,13 @@ def template_constant_dict(config, ignore=None, skip_lower=True):
             if version is not None:
 
                 _log.debug("version found in easyconfig is %s", version)
-                version = LooseVersion(version).version
+                version = version.split('.')
                 try:
-                    major = str(version[0])
+                    major = version[0]
                     template_values['version_major'] = major
-                    minor = str(version[1])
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/templates.py
```diff
@@ -171,13 +171,13 @@ def template_constant_dict(config, ignore=None, skip_lower=True):
...
 
                 _log.debug("version found in easyconfig is %s", version)
-                version = LooseVersion(version).version
+                version = version.split('.')
                 try:
-                    major = str(version[0])
+                    major = version[0]
                     template_values['version_major'] = major
-                    minor = str(version[1])
+                    minor = version[1]
```
wpoely86: and what about `3.14.15`?

boegel: what about it?

in that case:

* `version_major=3`
* `version_minor=14`

(and we don't have a template for the `15`)

wpoely86: okay fine

boegel:

PR #1788, opened 2016-05-31
ensure nice error message when non-existing path is passed to apply_regex_substitutions

boegel:


boegel:
@wpoely86 please review

easybuild/tools/filetools.py
```diff
@@ -741,6 +741,9 @@ def apply_regex_substitutions(path, regex_subs):
     else:
         _log.debug("Applying following regex substitutions to %s: %s", path, regex_subs)
 
+        if not os.path.exists(path):
```
wpoely86: why not wrap in a try? The file can exists without proper permissions?

boegel: yeah, ok, will do

boegel:

PR #1789, opened 2016-06-01
fix installation of Lua in install_eb_dep.sh script

boegel:
The `lua-5.1.4.8.tar.gz` at https://sourceforge.net/projects/lmod/files was updated (to also support `make uninstall`, cfr. https://sourceforge.net/p/lmod/mailman/message/35128697/).

However, the new version of the tarball is 'dirty', i.e. it ships `.so` files etc., which turns out to be an issue on Travis: `posix.so` for example doesn't get rebuilt, but then Lmod fails to pick it up because it's probably incompatible with the system it's being used on.

This is fixed by running `make clean` before configuring-building-installing Lua from the tarball provided in the Lmod repo. Doing this also requires to add the current working directory to `$PATH`, such that the `config.status` script is found.

We need to take care that `make clean` is only run for Lua (it fails hard to Lmod & other tools because of a missing target).

### DIFF
--- a/easybuild/scripts/install_eb_dep.sh
+++ b/easybuild/scripts/install_eb_dep.sh
@@ -12,6 +12,9 @@ PREFIX=$2
 PKG_NAME=`echo $PKG | sed 's/-[^-]*$//g'`
 PKG_VERSION=`echo $PKG | sed 's/.*-//g'`
 
+CONFIG_OPTIONS=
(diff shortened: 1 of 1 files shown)


boegel:
cc @rtmclay

wpoely86:
:+1:

boegel:

Issue #1787, opened 2016-05-30
Minimal toolchains picking up the wrong module when it exists in both dummy and GCCcore


damianam:
It seems like EB picks up the wrong module during building. For instance, when `binutils` or `ncurses` are both available in `dummy` and `GCCcore`, `--dry-run` and `--extended-dry-run` will work properly and show the correct dependency. However, during the build, the referenced paths are the ones of the `dummy` packages, not the ones in `GCCcore`. This seems like a silent bug in most cases.

Attached are a dry run, an extended dry run, and a debug log. Note in the debug log how at some point early in the process it load the correct module (the one in `GCCcore`), but later on the wrong one replaces it.

In the dry run you can see how `binutils` seems to not be installed in `dummy`. That is due to the same bug. We have removed the module in `dummy` to force EB to load the correct one.

The lack of `flex` in `dummy` is just because we have hide it there, so it is safe to ignore.

[easybuild-Python-3.5.1-20160530.183325.DnIed.log.txt](https://github.com/hpcugent/easybuild-framework/files/289893/easybuild-Python-3.5.1-20160530.183325.DnIed.log.txt)
[Python-3.5.1-GCC-5.3.0-2.26.dry_run.txt](https://github.com/hpcugent/easybuild-framework/files/289894/Python-3.5.1-GCC-5.3.0-2.26.dry_run.txt)
[Python-3.5.1-GCC-5.3.0-2.26.extended_dry_run.txt](https://github.com/hpcugent/easybuild-framework/files/289895/Python-3.5.1-GCC-5.3.0-2.26.extended_dry_run.txt)

boegel:
@damianam would you mind copy-pasting the contents of the `Core/GCC/5.3.0-2.26` module?

`binutils` gets loaded via that module, and since `GCCcore` gets (probably) loaded first, the `load binutils` picks up the module from `Compiler/GCCcore/5.3.0`.

However, isn't that by design? You want to use the binutils that was built with `GCCcore/5.3.0` whenever you're loading `GCC/5.3.0-2.25`, no?

damianam:
> However, isn't that by design? You want to use the binutils that was built with GCCcore/5.3.0 whenever you're loading GCC/5.3.0-2.25, no?

Exactly. The problem is that this is not the binutils that gets picked up, the binutils that gets picked up is the one in `dummy`.

The GCC module seems correct to me, and behaves correctly when loading it manually:

```
[...]

PR #1792, opened 2016-06-03
Protect hierarchical modules from MODULEPATH reordering

ocaisa:
Addresses #1787 

Don't allow subdirs of the target space for module files to be moved around, hierarchical schemes rely on their placement.

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -388,9 +388,8 @@ class ModulesTool(object):
             eb_modpath = os.path.join(install_path(typ='modules'), build_option('suffix_modules_path'))
             shufflepaths = \
                 (mod_path for mod_path in self.mod_paths[::-1] if
-                 (os.path.commonprefix([os.path.realpath(mod_path), os.path.realpath(eb_modpath)]) != eb_modpath)
-                 and (os.path.realpath(mod_path) != os.path.realpath(eb_modpath))
-                 )
+                 (os.path.realpath(mod_path) == os.path.realpath(eb_modpath)) or
+                 (os.path.commonprefix([os.path.realpath(mod_path), os.path.realpath(eb_modpath)]) != eb_modpath))
             for mod_path in shufflepaths:
                 self.prepend_module_path(mod_path)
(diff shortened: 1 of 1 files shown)


ocaisa:
This is not good enough, apart from style I think it will exclude `eb_modpath` which is explicitly not what we want.

ocaisa:
This is ugly, feel free to come up with a cleverer/shorter version.

ocaisa:
@damianam Can you test this when you get a chance and see if this fixes the problem?

boegel:
@ocaisa can you make the PR title more descriptive?

boegel:
@ocaisa, @damianam: I took a detailed look at this, and I think this is the actual bug:

```diff
diff --git a/easybuild/tools/modules.py b/easybuild/tools/modules.py
index 4b64d8b..2452f60 100644
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -485,8 +485,9 @@ class ModulesTool(object):
             else:
                 restore_env(init_env)
 
-        # make sure $MODULEPATH is set correctly after purging
-        self.check_module_path()
+            # make sure $MODULEPATH is set correctly after purging
+            self.check_module_path()
+
         # extend $MODULEPATH if needed
         for mod_path in mod_paths:
             full_mod_path = os.path.join(install_path('mod'), build_option('suffix_modules_path'), mod_path)
```


I.e., `check_module_path` should not be called before a `load` unless a '`purge`' was done (actually a reset of the environment to the starting state, but it has the same effect).

Basically, this translates to: don't fiddle with `$MODULEPATH` when modules may have been loaded that define `$MODULEPATH`.

Thoughts?

ocaisa:
That looks like it should work if that's where the problem is really triggered

boegel:
@ocaisa as far I can tell, this is the actual problem... Shall I open a separate PR for it, so @damianam can give it a test?

ocaisa:
Yeah sure, I'll close this if things are fixed there

boegel:
@ocaisa done in #1795

ocaisa:
Solved in #1795 so closing

PR #1793, opened 2016-06-03
add check for module output, empty stdout is a sign of trouble with Lmod

boegel:
motivated by https://lists.ugent.be/wws/arc/easybuild/2016-06/msg00037.html

The current implementation is too simple, but it's a starts and serves as a base for a fully fledged check.

@rtmclay any feedback on this? What would be a good way of checking whether Lmod ran correctly, based on the stdout & stderr produced by the `module` command?

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -929,10 +929,11 @@ class Lmod(ModulesTool):
 
     def check_module_output(self, cmd, stdout, stderr):
         """Check output of 'module' command, see if if is potentially invalid."""
+        fail_regex = re.compile('^stack traceback:', re.M)
         if stdout:
             self.log.debug("Output found in stdout, seems like '%s' ran fine", cmd)
-        else:
-            raise EasyBuildError("No output found in stdout, seems like '%s' failed", cmd)
+        elif fail_regex.search(stderr):
+            raise EasyBuildError("Empty stdout, '%s' found in stderr, seems like '%s' failed", fail_regex.pattern, cmd)
 
     def available(self, mod_name=None):
         """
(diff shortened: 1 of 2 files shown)


rtmclay:
I have two ideas that you might want to use to see if Lmod is "working".  First is to use

     $ $LMOD_CMD python --version

This will give you no output to stdout but it will give you the following written to stderr:

      Modules based on Lua: Version 6.4.1 (6.4-3-ge6438e8) 2016-06-04 08:48
          by Robert McLay mclay@tacc.utexas.edu

You could use the "Module based on Lua" string as check.  If you didn't get that then you know that Lmod is not working.  The printing of the version doesn't go very far into the program's execution.  So if you wanted something that went farther but was "cheap" to do then you could do:

     $ $LMOD_CMD python list

And check to see if stdout contained:

    os.environ["_ModuleTable001_"] ="...."

There will always be at least one of the _ModuleTable001_ variables in the environment even if you have no modules loaded at the time.

I hope this helps.

boegel:
@rtmclay so, only for `--version` the `stdout` can be empty, is that right? In all other cases, there should be `os.environ` in `stdout`, even for commands that don't make any changes to the environment, like `list`?

rtmclay:
Well there are several version type commands --gitversion, --dumpversion, --version will have an empty stdout.  I believe that in all other cases there will be an os.environ in stdout.  I haven't checked them all.

boegel:
OK, thanks for the feedback @rtmclay; I've made an exception for `--version` (we're not using the others).

@wpoely86 please review?

wpoely86:
:+1:

boegel:

PR #1794, opened 2016-06-06
Cray OpenMP wrapper flag

gppezzi:
First attempt to fix the "-True" flag that is added when using openmp as TC option on Cray. 
@boegel @pforai

### DIFF
--- a/easybuild/toolchains/compiler/craype.py
+++ b/easybuild/toolchains/compiler/craype.py
@@ -76,7 +76,6 @@ class CrayPECompiler(Compiler):
         'dynamic': '',
         'verbose': 'craype-verbose',
         'mpich-mt': 'craympich-mt',
-        'openmp': 'fopenmp',
     }
 
     COMPILER_CC = 'cc'
--- a/easybuild/toolchains/compiler/craype.py
+++ b/easybuild/toolchains/compiler/craype.py
@@ -76,6 +76,7 @@ class CrayPECompiler(Compiler):
         'dynamic': '',
         'verbose': 'craype-verbose',
         'mpich-mt': 'craympich-mt',
+        'openmp': 'fopenmp',
     }
 
     COMPILER_CC = 'cc'
--- a/easybuild/toolchains/compiler/craype.py
+++ b/easybuild/toolchains/compiler/craype.py
@@ -137,6 +137,7 @@ class CrayPEGCC(CrayPECompiler):
     def __init__(self, *args, **kwargs):
         """CrayPEGCC constructor."""
         super(CrayPEGCC, self).__init__(*args, **kwargs)
+        self.COMPILER_UNIQUE_OPTION_MAP.update({'openmp': 'fopenmp'})
         for precflag in self.COMPILER_PREC_FLAGS:
             self.COMPILER_UNIQUE_OPTION_MAP[precflag] = Gcc.COMPILER_UNIQUE_OPTION_MAP[precflag]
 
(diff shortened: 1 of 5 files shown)


boegel:
@gppezzi looks perfect to me... Did you test it? Any example easyconfigs where you have `openmp` enabled?

gppezzi:
I've tested with GROMACS GPU but this is WIP, I will try with other simpler cases and let you know.

boegel:
@gppezzi maybe try building `STREAM` with the Cray* toolchains to test this?

pforai:
@gppezzi, @boegel 

The OpenMP flags are sadly inconsistent between the drivers for CCE, Intel, GNU and PGI:

```bash
OpenMP support:
   -homp		Enables OpenMP and links in OpenMP libraries when
			possible using CCE.
			(This is the default.)

   -hnoomp		Disables OpenMP and links in non-OpenMP libraries
			when using CCE.

   -fopenmp		Enables OpenMP and links in OpenMP libraries when
			possible using GNU.

   -openmp		Enables OpenMP and links in OpenMP libraries when
			possible when using Intel.

   -mp			Enables OpenMP and links in OpenMP libraries when
			possible using PGI.

   -Mnoopenmp		Disables OpenMP and links in non-OpenMP libraries
			when using PGI.
```

gppezzi:
If the new implementation makes sense to you, I'll apply the same fix accordingly to Intel, CCE and PGI. Let me know.

``` 
eb ~/GROMACS-5.1.2-CrayGNU-2016.03-cuda-7.0.eb -x | grep CFLAGS
```
* Before the fix
```
  export CFLAGS="-craype-verbose -True -O2"
  export FCFLAGS="-craype-verbose -True -O2"
```
* After
```
  export CFLAGS="-craype-verbose -fopenmp -O2"
  export FCFLAGS="-craype-verbose -fopenmp -O2"
```

pforai:
Yeah, this looks good! 

Update the unique compiler map for each of the subclasses accordingly.

gppezzi:
I've updated the other subclasses. 

So far I've tested this for GNU and Intel. CCE and PGI still need to be checked.

easybuild/toolchains/crayintel.py
```diff
@@ -38,3 +38,4 @@ class CrayIntel(CrayPEIntel, CrayMPICH, LibSci):
     """Compiler toolchain for Cray Programming Environment for Intel compilers (PrgEnv-intel)."""
     NAME = 'CrayIntel'
     SUBTOOLCHAIN = DUMMY_TOOLCHAIN_NAME
+    CrayPEIntel.COMPILER_UNIQUE_OPTION_MAP.update({'openmp': 'openmp'})
```
boegel: `openmp` is deprecated for recent Intel versions, better stick to `fopenmp` here too, cfr. #1718

boegel:
@gppezzi I think you'll need to tackle this in `easybuild/toolchains/compiler/craype.py` instead, where we already fiddle with `self.COMPILER_UNIQUE_OPTION_MAP` for the precision flags (which are also compiler specific)

gppezzi:
@boegel now tested with CCE, Intel and GNU (I don't have any test for PGI)
cc @pforai @jgphpc 

 * CCE
```
export CFLAGS="-craype-verbose -homp -O2"
```
* Intel
```
export CFLAGS="-craype-verbose -fopenmp -O2 -ftz -fp-speculation=safe -fp-model"
```
* GNU
```
export CFLAGS="-craype-verbose -fopenmp -O2"
```

easybuild/toolchains/compiler/craype.py
```diff
@@ -137,6 +137,7 @@ class CrayPEGCC(CrayPECompiler):
     def __init__(self, *args, **kwargs):
         """CrayPEGCC constructor."""
         super(CrayPEGCC, self).__init__(*args, **kwargs)
+        self.COMPILER_UNIQUE_OPTION_MAP.update({'openmp': 'fopenmp'})
```
boegel: @gppezzi I would go with this instead, just to match the style below:

```python
self.COMPILER_UNIQUE_OPTION_MAP['openmp'] = 'fopenmp'
```

other than that: should be good to go

gppezzi:
@boegel  any other remark?

boegel:
lgtm, thanks @gppezzi

PR #1795, opened 2016-06-07
only reset $MODULEPATH before loading a module if environment was reset

boegel:
fix for #1787, alternative for #1792

Always redefining `$MODULEPATH` according to `self.mod_paths` for every `load` is wrong, since the `$MODULEPATH` may have been tweaked by previous loads with a module hierarchy.

Besides, it's fairly expensive to do so as well (lots of unnecessary `module use` commands...).

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -485,8 +485,9 @@ class ModulesTool(object):
             else:
                 restore_env(init_env)
 
-        # make sure $MODULEPATH is set correctly after purging
-        self.check_module_path()
+            # make sure $MODULEPATH is set correctly after purging
+            self.check_module_path()
+
         # extend $MODULEPATH if needed
         for mod_path in mod_paths:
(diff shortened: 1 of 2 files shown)


boegel:
@damienam please verify whether this fixes #1787?

damianam:
I've just checked it (I needed the bug fix for something, so timing was really convenient). It works now for me.

damianam:
Attached a debug log for those interested. ncurses was previously picked up from the wrong level (`dummy`)

[easybuild-Python-2.7.11-20160607.151121.log.gz](https://github.com/hpcugent/easybuild-framework/files/302665/easybuild-Python-2.7.11-20160607.151121.log.gz)

boegel:
@damianam Thanks for the confirmation!

I'll see if this needs a test before it can be merged in...

boegel:
Test added, Travis approves, going in, thanks for reporting this @damianam and for verifying the fix.

Issue #1481, opened 2015-11-19
No FFT_MT


ocaisa:
With the MKL, there are _MT versions of BLAS and LAPACK environment variables but no corresponding _MT variable for FFT. Is there a good reason for this? It looks like it is required when building  a threaded version of ABINIT

boegel:
I don't think there's a very good reason.

It seems like some of the required logic is there though, see https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/toolchains/fft/fftw.py#L52.

I.e., we're basically defining `$LIBFFT` with the thread-safe `libfftw` if the `usempi` toolchain option is `True`.

That's not exactly consistent with what we do for `BLAS` and `LAPACK` though, so letting the framework also define `$LIBFFT_MT` makes sense.

Are you up for that? It should be pretty straightforward based on the logic for `$LIBBLAS_MT`.

Take a look at `easybuild/tools/toolchain/linalg.py`, look for `_MT`, and then adjust `easybuild/tools/toolchain/fft.py` + the modules in `easybuild/toolchains/linalg/`

ocaisa:
This bit us again yesterday with Quantum Espresso and the foss toolchain. Part of the problem is that the quantum espresso easyblock is not (yet) smart enough to use the correct libraries when things are threaded. It needs some additional logic in https://github.com/hpcugent/easybuild-easyblocks/blob/master/easybuild/easyblocks/q/quantumespresso.py#L136 to ensure the MT libraries are used when building the hybrid mode. Even if you do that, you come back to the issue that `LIBFFT_MT` does not exist (yet) in the framework, without it you end with linking problems (when not using MKL).

boegel:
fixed with #1802, thanks @ocaisa!

PR #1802, opened 2016-06-10
Add multi-threaded FFT to toolchain

ocaisa:
This is required in some cases or you're going to get linking problems. Addresses #1481

### DIFF
--- a/easybuild/toolchains/fft/fftw.py
+++ b/easybuild/toolchains/fft/fftw.py
@@ -53,8 +53,12 @@ class Fftw(Fft):
         fftw_libs = ["fftw%s" % suffix]
         if self.options['usempi']:
             fftw_libs.insert(0, "fftw%s_mpi" % suffix)
+        fftw_libs_mt = ["fftw%s" % suffix]
+        if self.options['openmp']:
+            fftw_libs_mt.insert(0, "fftw%s_omp" % suffix)
 
         self.FFT_LIB = fftw_libs
+        self.FFT_LIB_MT = fftw_libs_mt
 
     def _set_fft_variables(self):
         self._set_fftw_variables()
@@ -66,3 +70,5 @@ class Fftw(Fft):
         self.variables.join('FFTW_LIB_DIR', 'FFT_LIB_DIR')
         if 'FFT_STATIC_LIBS' in self.variables:
             self.variables.join('FFTW_STATIC_LIBS', 'FFT_STATIC_LIBS')
+        if 'FFT_STATIC_LIBS_MT' in self.variables:
+            self.variables.join('FFTW_STATIC_LIBS_MT', 'FFT_STATIC_LIBS_MT')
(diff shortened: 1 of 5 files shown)


easybuild/toolchains/fft/fftw.py
```diff
@@ -53,8 +53,12 @@ def _set_fftw_variables(self):
         fftw_libs = ["fftw%s" % suffix]
         if self.options['usempi']:
             fftw_libs.insert(0, "fftw%s_mpi" % suffix)
+        fftw_libs_mt = ["fftw%s" % suffix]
+        if self.options['openmp']:
```
boegel: better use `self.options.get('openmp', False)`, it may not always be defined

easybuild/toolchains/fft/fftw.py
```diff
@@ -53,8 +53,12 @@ def _set_fftw_variables(self):
         fftw_libs = ["fftw%s" % suffix]
         if self.options['usempi']:
             fftw_libs.insert(0, "fftw%s_mpi" % suffix)
+        fftw_libs_mt = ["fftw%s" % suffix]
+        if self.options['openmp', False]:
```
ocaisa: @boegel What am I doing wrong here? The tests are still failing...should the default be none?

boegel: @ocaisa not `[...]` but `(...)`:

```python
if self.options('openmp', False):
```

boegel: oh, sorry, this:

```python
if self.options.get('openmp', False):
```

ocaisa: How come that's not needed above for `usempi`?

boegel: That's a very good question.

Apparently `usempi` is *always* defined, but `openmp` is not. It's subtle, and hard to pinpoint (the toolchain mechanism is... trickey).

It's probably better to do the same for `usempi`

boegel:
@ocaisa We should also make sure `$LIBFFT_MT` is defined correctly for Intel MKL?

And maybe add tests to verify that, we have a bunch of similar ones already (for `$LIBBLAS`, etc.), see `test/framework/toolchain.py`

ocaisa:
@boegel I've checked all over the web and I can't see that you need to do anything special for threaded MKL, you just need to set variables like `OMP_NUM_THREADS` in your environment at runtime. In that sense then the code already covers that case since it just sets `LIBFFT = LIBFFT_MT` by default if there's no other option.
I'll look into the tests next week.

ocaisa:
@boegel Test included

boegel:
@ocaisa Thanks for the clarification on `$LIBFFT` for Intel MKL, and for tackling the test.

It's good enough as is, but it has to *pass* of course:

```
AssertionError: '-lfftw3 -lpthread' != '-lfftw3_omp -lfftw3 -lpthread'
```

Any idea what's wrong? Forgot to enable `openmp` toolchain option?

ocaisa:
Yup, that's it, but I don't know how I can switch on `openmp` toolchain option for the toolchain instance

ocaisa:
Ok, figured it out, test passes now

boegel:
Great, thanks @ocaisa, going in!

PR #1806, opened 2016-06-13
cache $PATH & $PYTHONPATH in test setUp, restore them in tests where 'eb' is used

boegel:
(verified) fix for #1685


boegel:
@wpoely86 please review

@jhein32: can you verify whether this fixes the installation on your end? It should, since I was able to reproduce the problem reported in #1685, and then fix it by putting these changes in place...

wpoely86:
:+1:

boegel:
Fix confirmed bi @jhein32 in #1685

PR #1808, opened 2016-06-13
avoid special characters like '[', ']' in path to temporary directory

boegel:
I've run into this a couple of times when submitting array jobs with `eb` commands on our system using Torque, since then you end up in an environment where `$TMPDIR` is set to something like:

```
`/local/12345[3200].master15.delcatty.gent.vsc`
```

Sometimes, this breaks the build/installation for no good reason. I've seen this occur primarily when installing Python packages (it seems like `setuptools` has trouble with it), and when using `nvcc` which picks up `$TMPDIR` and uses it as a 'scratch' space, cfr. https://github.com/hpcugent/easybuild-easyblocks/pull/953#issuecomment-225613062

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -63,6 +63,7 @@ from easybuild.tools.config import get_pretend_installpath, mk_full_default_path
 from easybuild.tools.configobj import ConfigObj, ConfigObjError
 from easybuild.tools.docs import FORMAT_RST, FORMAT_TXT, avail_easyconfig_params
 from easybuild.tools.environment import restore_env, unset_env_vars
+from easybuild.tools.filetools import mkdir
 from easybuild.tools.github import GITHUB_EB_MAIN, GITHUB_EASYCONFIGS_REPO, HAVE_GITHUB_API, HAVE_KEYRING
 from easybuild.tools.github import fetch_github_token
 from easybuild.tools.include import include_easyblocks, include_module_naming_schemes, include_toolchains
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86 please review

easybuild/tools/options.py
```diff
@@ -1174,6 +1175,18 @@ def set_tmpdir(tmpdir=None, raise_error=False):
     except OSError, err:
         raise EasyBuildError("Failed to create temporary directory (tmpdir: %s): %s", tmpdir, err)
 
+    # avoid having special characters like '[' and ']' in the tmpdir pathname,
+    # it is known to cause problems (e.g., with Python install tools, CUDA's nvcc, etc.);
+    # only common characteris like alphanumeric, '_', '-', '.' and '/' are retained; others are converted to '_'
+    special_chars_regex = r'[^\w/.-]'
+    if re.search(special_chars_regex, current_tmpdir):
+        current_tmpdir = re.sub(special_chars_regex, '_', current_tmpdir)
```
wpoely86: nitpicking but I would convert to `X` ?

boegel: any particular reason?

wpoely86: cause `X` is just an ordinary letter and `_` might be 'special' for some codes.

boegel: well, ok, I'll change it

wpoely86:
:+1:

boegel:

Issue #1804, opened 2016-06-13
Can't install via bootstrapper script


eburgueno:
Hi, I'm trying to install easytools 2.8.1 in our cluster where we already use environment modules (Tcl) for some of the tools we provide.

The [bootstrap](http://easybuild.readthedocs.io/en/latest/Installation.html#bootstrapping-easybuild) script currently fails at STAGE 2 with the following traceback:
```
$ EASYBUILD_PREFIX=/<prefix path>/easybuild-2.8.1
$ python bootstrap_eb.py $EASYBUILD_PREFIX
[[INFO]] Installation prefix /<prefix path>/easybuild-2.8.1
[[INFO]] Found module command 'modulecmd' (EnvironmentModulesC), so using it.
[[INFO]] 
(...)
+++ STAGE 2: installing EasyBuild in /<MY_PREFIX>/easybuild-2.8.1 with EasyBuild from stage 1...

== temporary log file in case of crash /tmp/eb-wzg4zg/easybuild-mKcPfI.log
Traceback (most recent call last):
  File "bootstrap_eb.py", line 787, in <module>
    main()
  File "bootstrap_eb.py", line 590, in main
    stage2(tmpdir, templates, install_path, distribute_egg_dir, sourcepath)
  File "bootstrap_eb.py", line 508, in stage2
    easybuild_main()
  File "/tmp/tmpj6LFnx/eb_stage1/lib/python2.7/site-packages/easybuild_framework-2.8.1-py2.7.egg/easybuild/main.py", line 255, in main
    modtool = modules_tool(testing=testing)
  File "/tmp/tmpj6LFnx/eb_stage1/lib/python2.7/site-packages/easybuild_framework-2.8.1-py2.7.egg/easybuild/tools/modules.py", line 1126, in modules_tool
    return modules_tool_class(mod_paths=mod_paths, testing=testing)
  File "/tmp/tmpj6LFnx/eb_stage1/lib/python2.7/site-packages/easybuild_framework-2.8.1-py2.7.egg/easybuild/tools/modules.py", line 192, in __init__
    self.check_module_path()
  File "/tmp/tmpj6LFnx/eb_stage1/lib/python2.7/site-packages/easybuild_framework-2.8.1-py2.7.egg/easybuild/tools/modules.py", line 379, in check_module_path
    self.prepend_module_path(eb_modpath)
[...]

PR #1810, opened 2016-06-14
don't reset $MODULEPATH in stage 2 of bootstrap script, support forced installation during stage 2

boegel:
fix for problem reported in #1804

### DIFF
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -67,6 +67,9 @@ EASYBUILD_PACKAGES = [VSC_BASE, 'easybuild-framework', 'easybuild-easyblocks', '
 # set print_debug to True for detailed progress info
 print_debug = os.environ.pop('EASYBUILD_BOOTSTRAP_DEBUG', False)
 
+# install with --force in stage2?
+forced_install = os.environ.pop('EASYBUILD_BOOTSTRAP_FORCED', False)
+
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 please review

easybuild/scripts/bootstrap_eb.py
```diff
@@ -469,13 +472,13 @@ def stage2(tmpdir, templates, install_path, distribute_egg_dir, sourcepath):
     handle.write(EASYBUILD_EASYCONFIG_TEMPLATE % templates)
     handle.close()
 
-    # unset $MODULEPATH, we don't care about already installed modules
-    os.environ['MODULEPATH'] = ''
```
wpoely86: why?

boegel: why what? not reset it?

because
* i) it's not needed (I can't think of a reason why wiping `$MODULEPATH` would matter, other than having `EasyBuild` module available already)
* ii) wrong, especially in combination with Lmod (since it may trigger unloading of modules because they're deemed inactive because they're not longer available)

wpoely86: okay

boegel:
@eburgueno mentioned in #1804 that this doesn't solve his problem, and I'm unsure why, so I'll hold this off until more feedback is given via #1804...

boegel:
fix in bootstrap script confirmed in #1804 by @eburgueno, so merging this (even though this by itself is not sufficient to fix the problem reported in #1804)

PR #1811, opened 2016-06-15
Added option for targeting Xeon Phi arch

RalphTheBest:
This is the bare minimum to compile an app for Xeon Phi architecture. As I said in the mail list, it's not enough in most of the cases, but I haven't been able to find any general workarounds. 
Some extra work will be needed to make each application work properly, but anyway,  this provides the common step.

Please take a look and tell me if any changes are needed.


RalphTheBest:
Oops, wrong commit. I forgot to add some details.

boegel:
@RalphTheBest you can just push additional commits to the same branch to update a pull request

Issue #1402, opened 2015-09-25
HierarchicalMNS cannot parse easyconfig


ocaisa:
For some reason, `binutils-2.25-GCC-4.9.3-binutils-2.25.eb` cannot be parsed when using `HierarchicalMNS`:
```
$ eb  binutils-2.25-GCC-4.9.3-binutils-2.25.eb  --dry-run --module-naming-scheme=HierarchicalMNS
== temporary log file in case of crash /tmp/eb-IBa7go/easybuild-7uT1Np.log
ERROR: Failed to process easyconfig /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/binutils/binutils-2.25-GCC-4.9.3-binutils-2.25.eb: Failed to find easyconfig file 'flex-2.5.39-GCC-4.9.3-binutils-%(version)s.eb' when determining module name for: {'external_module_metadata': {}, 'dummy': False, 'toolchain': {'version': '4.9.3-binutils-%(version)s', 'name': 'GCC'}, 'name': 'flex', 'full_mod_name': None, 'short_mod_name': None, 'versionsuffix': '', 'version': '2.5.39', 'hidden': False, 'external_module': False}
```

Ideas @geimer ?

ocaisa:
Is this an easyconfig problem or a parser? It seems to be trivially fixed by replacing
```
toolchain = {'name': 'GCC', 'version': '4.9.3-binutils-%(version)s' }
```
with
```
toolchain = {'name': 'GCC', 'version': '4.9.3-binutils-%s' % version}
```

boegel:
@ocaisa: this indicates a problem in the framework somewhere; the `%(version)s` template is not being resolved

seems like a chicken-egg situation (the version is only available after parsing the easyconfig)

boegel:
I'll merge https://github.com/hpcugent/easybuild-easyconfigs/pull/2104 as a workaround for this until we can figure out a proper fix.

boegel:
Seems like this problem is more extensive than I realised, see https://github.com/hpcugent/easybuild-easyconfigs/issues/3210 reported by @verdurin... I'll try and fix this by the next release.

boegel:
This is fixed with #1812, @ocaisa: please verify?

boegel:
fixed with #1812

PR #1812, opened 2016-06-16
fix issue with templates defined by deps being required while still parsing deps

boegel:
This should fix the problems being reported in #1402 and https://github.com/hpcugent/easybuild-easyconfigs/issues/3210.

cc @verdurin, @ocaisa


boegel:
@verdurin, @ocaisa: I'll look into an test that covers this, but it would be great if you could verify this fix (and review the changes).

verdurin:
@boegel having manually applied the new version of `easyconfig.py` then reload the module, I tried this:

```
eb cnvkit-0.7.3-foss-2015b-Python-2.7.10.eb -dr --force --minimal-toolchains --robot ./
```

and it worked, without generating the errors I reported in https://github.com/hpcugent/easybuild-easyconfigs/issues/3210

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -806,19 +811,8 @@ def _parse_dependency(self, dep, hidden=False, build_only=False):
...
-                orig_tc = tc
-                tc = robot_find_minimal_toolchain_of_dependency(dependency, self.modules_tool, parent_tc=tc)
-                if tc is None:
-                    raise EasyBuildError("No easyconfig for %s that matches toolchain hierarchy generated by %s",
-                                         dependency, orig_tc)
-                else:
-                    self.log.debug("Obtained minimal toolchain: %s", tc)
-            else:
-                self.log.debug("Inheriting parent toolchain %s for dependency %s", tc, dependency)
+            self.log.debug("Inheriting parent toolchain %s for dependency %s (for now)", tc, dependency)
```
verdurin: @boegel - can you clarify what you mean by "(for now)" here? This is an interim fix, or just "for now" as part of the dependency resolution process?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -843,28 +837,70 @@ def _parse_dependency(self, dep, hidden=False, build_only=False):
...
+
+        for key in DEPENDENCY_PARAMETERS:
+            # loop over a *copy* of dependency dicts (with resolved templates);
+            # to update the original dep dict, we need to index with idx into self._config[key][0]...
+            for idx, dep in enumerate(self[key]):
+
+                # reference to original dep dict, this is the one we should be updating
+                orig_dep = self._config[key][0][idx]
+
+                # minimize toolchains for dependencies with inherite toolchain, if requested
```
verdurin: Should be "inherited"?

ocaisa:
@boegel Tested with `Boost-1.60.0-intel-2016a-Python-2.7.11.eb` from the main easyconfig repo and parses everything correctly.

easybuild/framework/easyconfig/easyconfig.py
ocaisa: With `minimal-toolchains` you might be replacing this toolchain with one further down the hierarchy when the dependencies are finalised. Probably could do with replacing `(for now)` by `(until dependencies are finalised)`

ocaisa:
lgtm

wpoely86:
:+1:

boegel:
Thanks for the confirmation of the fix @verdurin and @ocaisa, and for the review of the changes!

Going in!

PR #1813, opened 2016-06-20
skip unneeded unuse/use commands on tail of $MODULEPATH in check_module_path

boegel:
fix for the issue reported in #1804, we should avoid unneeded `unuse` commands, since they can have unwanted side-effects


boegel:
@wpoely86 please review?

wpoely86:
:+1:

boegel:

Issue #1796, opened 2016-06-08
Overhauling the EasyBuild logging


ocaisa:
It might be interesting to consider yaml output for EB logging. This approach is used, for example, by [BigDFT](http://bazaar.launchpad.net/~luigi-genovese/+junk/1.8/view/head:/bigdft/tests/DFT/cubic/C/C.out.ref.yaml).

This would potentially allow us to use log files as input to allow reproduction of bugs, and also facilitate post-processing of log files (different views:minimal, detailed, debug,...). 

BigDFT have tips for how to still allow the output to be human readable and grep-able and said they'd be happy to share them

verdurin:
This came up again in the Maintainer Summit, March 2023.

@smoors may be able to share the Vim syntax colouring setup he created, for easier navigation.

smoors:
here is my vim syntax file for inspiration
can certainly be improved upon
https://gist.github.com/smoors/0b4012a1109acbc6edb8e41852eff648

terjekv:
Is there a reason to use YAML over JSON for logging? https://www.structlog.org/en/stable/ is very nice to work with, and integrates very smoothly into python.

Issue #1353, opened 2015-08-13
Log files are intimidating!


ocaisa:
It would be nice to at least have the option to split log files on a per-step basis. Log files are currently full of EB specific info that (at least to a new user) hide useful information. For me the most important things when building an app are being able to study the output from the configure step to make sure all requested features are there an then studying the output of the build process to fix any errors.

boegel:
We've had this idea for a while, but it never materialised.

Making this happen should not be too hard, I think. It's a matter of redefining `self.log` before starting each step (see `run_step` in `easyblock.py`), and then making sure all those separate log files are collapsed into a single one at the end.

Creating a separate log in case of errors may also be possible/easy to do.

PR #1815, opened 2016-06-22
add support for 'devel' log level

boegel:
Currently, debug logs are _very_ verbose, with lots of information being logged that is only really useful when developing EasyBuild features.

One particular example in the logging by the toolchain mechanism.

With this `devel` log level in place, we could avoid including stuff like this in `debug` logs, while avoiding having to remove the `log.debug` statements entirely.

Once a review of all current `log.debug` statements is done and the relevant ones are 'downgraded' to `log.devel`, we can consider make debug logging the default in EasyBuild v3.0.0 (potentially combined with zipping logs by default).

cc @wpoely86, @ocaisa


boegel:
relates to #1353 and #1796

boegel:
@wpoely86 please review?

The 2nd step here would be to actually switch to using `log.devel` in some places in the framework (e.g. a large part of the `toolchain` mechanism), to make debug logs more lightweight/focused.

easybuild/tools/build_log.py
```diff
@@ -58,6 +59,10 @@
 DRY_RUN_MODULES_INSTALL_DIR = None
 
 
+DEVEL_LOG_LEVEL = logging.DEBUG - 1
```
wpoely86: not +1 ?

boegel: no, it's a lower logging level; `log.devel` messages should not be included when using `--debug`, but `log.debug` messages *should* be included when using `--devel` (see also the added test)

boegel: @wpoely86 so, is this good to go?

boegel:

PR #1821, opened 2016-06-28
add support for --extra-modules

boegel:


boegel:
@wpoely86 please review

easybuild/framework/easyblock.py
```diff
@@ -1560,6 +1560,11 @@ def prepare_step(self, start_dir=True):
             # version is expected to be something that makes sense
             env.setvar(get_software_version_env_var_name(name), version)
 
+        extra_modules = build_option('extra_modules')
+        if extra_modules:
+            self.log.info("Loading extra modules: %s", extra_modules)
+            self.modules_tool.load(extra_modules)
```
wpoely86: remind me, is this before or after loading the toolchain and deps?

boegel: after, I use it for XALT, for example (where order matters a lot)

wpoely86: perfect. We can use it too for `ccache` than.

wpoely86:
:+1:

boegel:

PR #1825, opened 2016-06-29
fix format of supported easyconfig templates in help output

boegel:
This ensures that `%(name)s`, `%(namelower)s` is used rather than just `name`, `namelower` in the help output for templates...


boegel:
@wpoely86 please review

wpoely86:
:+1:

boegel:

PR #1826, opened 2016-06-30
[yeb] type-convert checksum and patch values

Caylo:
(to use in yeb, simply replace tuples by lists)

### DIFF
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -401,10 +401,10 @@ def to_dependencies(dep_list):
     return [to_dependency(dep) for dep in dep_list]
 
 def to_checksums(checksums):
-    if any(isinstance(checksum, list) for checksum in checksums):
-        return [to_list_of_strings_and_tuples(checksums) for cs in checksums]
+    if not any(isinstance(checksum, list) for checksum in checksums):
+        return to_list_of_strings_and_tuples(checksums)
     else:
-        return to_list_of_strings_and_tuples(cs)
+        return [to_list_of_strings_and_tuples(cs) for cs in checksums]
 
 
 # these constants use functions defined in this module, so they needs to be at the bottom of the module
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -400,6 +400,11 @@ def to_dependencies(dep_list):
     """
     return [to_dependency(dep) for dep in dep_list]
 
+def to_checksums(checksums):
+    # make sure we have a nested list of checksums
+    if not isinstance(checksums[0], list):
+        checksums = [checksums]
+    return [to_list_of_strings_and_tuples(cs) for cs in checksums]
 
 # these constants use functions defined in this module, so they needs to be at the bottom of the module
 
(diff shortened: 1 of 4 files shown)


easybuild/framework/easyconfig/types.py
```diff
@@ -400,6 +400,12 @@ def to_dependencies(dep_list):
     """
     return [to_dependency(dep) for dep in dep_list]
 
+def to_checksums(checksums):
+    if not isinstance(checksums[0], list):
```
wpoely86: why the `[0]`?

Caylo: checksums can be a list of lists, see https://github.com/hpcugent/easybuild-framework/blob/master/test/framework/easyconfigs/toy-0.0.eb#L10

boegel:
(open/close to trigger Travis)

easybuild/framework/easyconfig/types.py
```diff
@@ -433,16 +439,20 @@ def to_dependencies(dep_list):
     'opt_keys': [],
     'req_keys': ['files', 'dirs'],
 }))
-CHECKABLE_TYPES = [DEPENDENCIES, DEPENDENCY_DICT, NAME_VERSION_DICT, SANITY_CHECK_PATHS_DICT, STRING_OR_TUPLE_LIST,
-                   TUPLE_OF_STRINGS]
+CHECKSUMS = (list, as_hashable({'elem_types': [STRING_OR_TUPLE_LIST]}))
+
+CHECKABLE_TYPES = [CHECKSUMS, DEPENDENCIES, DEPENDENCY_DICT, NAME_VERSION_DICT, SANITY_CHECK_PATHS_DICT,
+                    STRING_OR_TUPLE_LIST, TUPLE_OF_STRINGS]
```
boegel: style nit-picking: fix indent to align with line above

easybuild/framework/easyconfig/types.py
boegel: I'm not sure why this is needed either, it looks wrong to only consider the 1st element.

easybuild/framework/easyconfig/types.py
```diff
@@ -400,6 +400,12 @@ def to_dependencies(dep_list):
     """
     return [to_dependency(dep) for dep in dep_list]
 
+def to_checksums(checksums):
+    if not any(isinstance(checksum, list) for checksum in checksums):
+        return to_list_of_strings_and_tuples(checksums)
+    else:
+        return [to_list_of_strings_and_tuples(cs) for cs in checksums]
```
boegel: positive logic (and a docstring) makes this easier to comprehend:

```python
def to_checksums(checksums):
    """Ensure correct element types for list of checksums: convert each element to a tuple."""
    if all(isinstance(cs, basestring) for cs in checksums):
        # if all elements are strings, we convert each individual element to a tuple
        res = [to_list_of_strings_and_tuples(cs) for cs in checksums]
    else:
        res = to_list_of_strings_and_tuples(checksums)

    return res
```

(do check my comment/docstring, I may confuse between list/tuple...)

boegel:
Going in, thanks @Caylo!

Issue #1831, opened 2016-07-06
relative path for -r doesn't work


Caylo:

boegel:
Clarification: this is only an issue in df7e6f3d45e8b1b8cf359ad278c8a00f66dfda61 because of the `chdir` being done there (without returning back to the original directory).

Regardless, it's probably wise to resolve the path(s) passed to robot before going on.

@Caylo Can you tackle that yourself, in `easybuild/tools/options.py` in `_postprocess_config`, where `self.options.robot` is defined, use `os.path.abspath`.

And while you're at it, also do this for the paths to the easyconfig files, i.e. at the end of `det_easyconfig_paths` in `easybuild.framework.easyconfig.tools`, right before `ec_files` is returned.

boegel:
fixed in #1834, thanks @Caylo

PR #1834, opened 2016-07-07
use abspath for robot and easyconfig files

Caylo:
ref #1831


boegel:
lgtm, ok to go if Travis is happy

boegel:
Going in, thanks @Caylo!

PR #1837, opened 2016-07-07
Add support for 'hidden=True' easyconfig parameter

geimer:
This complements the '--hidden' command-line option.

### DIFF
--- a/easybuild/framework/easyconfig/default.py
+++ b/easybuild/framework/easyconfig/default.py
@@ -88,6 +88,7 @@ DEFAULT_CONFIG = {
     'easyblock': [None, "EasyBlock to use for building; if set to None, an easyblock is selected "
                         "based on the software name", BUILD],
     'easybuild_version': [None, "EasyBuild-version this spec-file was written for", BUILD],
+    'hidden': [False, "Install module file as 'hidden' by prefixing its name with '.'", BUILD],
     'installopts': ['', 'Extra options for installation', BUILD],
     'maxparallel': [None, 'Max degree of parallelism', BUILD],
     'parallel': [None, ('Degree of parallelism for e.g. make (default: based on the number of '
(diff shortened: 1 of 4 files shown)


boegel:
@geimer: `build_option('hidden')` is also used in `process_easyconfig`, do we need to check the `hidden` parameter there too?

Looks tricky at first sight, since it's used before parsing the easyconfig, and I'm not sure it's strictly needed there...

geimer:
@boegel: Something like `hidden = hidden or ec['hidden']` in the `if not parse_only:` block may do the trick.  What do you think?

geimer:
Of course I meant `easyconfig['hidden']` instead of `ec['hidden']` ;-)

geimer:
@boegel: I believe the easyconfig update for `hidden` in L1230 isn't necessary. The `build_option` is passed to the `EasyConfig` constructor, which will correctly initialize the value based on `build_option` and the `hidden` easyconfig parameter.  Does that make sense?

geimer:
@boegel: Still trying to understand the code...

Shouldn't the update in https://github.com/geimer/easybuild-framework/blob/5d9a56a543eb7f8e078d6aa737077cfc23be40ed/easybuild/framework/easyconfig/easyconfig.py#L1230 be `'hidden': ec.hidden`?  Then the first query to `build_option('hidden')` only affects the cache lookup, which should be OK as the data in the cache should already include parsed values.  The update then takes the actual parsed value into account, determined in the `EasyConfig` constructor based on the build option and the easyconfig parameter.

The thing is that the `easyconfig` unit tests all pass, regardless of what the update does.  I can even hard-code it to `True` or `False`...

BTW: There is a docstring for the `parse_only` parameter missing.

boegel:
@geimer I think you're right with your last comment. Up until now it didn't really matter, but with the introduction of the `hidden` easyconfig parameter, it does matter.

I'd have to dive in and figure out where that `'hidden'` key is actually used, if at all, it may be "dead code", but it certainly doesn't hurt to get it right (and using `ec.hidden` is correct imho).

The cache aspect is a bit weird, but only taking into account `--hidden` in the cache key is correct imho, since the `hidden` easyconfig parameter only comes into play when the easyconfig is already parsed.

geimer:
@boegel: IMHO, this is ready for review.

boegel:
@geimer looks good, but I couldn't resist making a few minor tweaks, see https://github.com/geimer/easybuild-framework/pull/8

boegel:
Going in, thanks @geimer!

PR #1838, opened 2016-07-08
merge with develop when using --from-pr

Caylo:
see https://github.com/hpcugent/easybuild-framework/pull/1838#issuecomment-231350392


easybuild/tools/filetools.py
```diff
@@ -655,7 +655,7 @@ def guess_patch_level(patched_files, parent_dir):
     return patch_level
 
 
-def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
+def apply_patch(patch_file, dest, fn=None, copy=False, level=None, log_ok=True, from_pr=None, obliged=False):
```
boegel: I don't think `log_ok` is very useful, we should always run the `patch` command with `log_ok=False` since we check the return value of `run_cmd` anyway?

boegel: also, I don't think we should make `apply_patch` aware whether is was called in the context of `--from-pr` or not, just catch a possible patching failure where you call `apply_patch` and react to it?

easybuild/tools/github.py
```diff
@@ -324,6 +336,24 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
+        raise EasyBuildError("Downloading of develop branch failed: not found in %s", develop)
+
+    patch_name = 'patch%s' % pr
+    patch_url = URL_SEPARATOR.join([GITHUB_URL, GITHUB_EB_MAIN, GITHUB_EASYCONFIGS_REPO, GITHUB_PR, '%s.patch' % pr])
+    patch_path = os.path.join(path, patch_name)
+    patch = download_file(patch_name, patch_url, patch_path)
+    if patch is None:
+        raise EasyBuildError("Failed to download file %s to %s", patch_url, patch_path)
+
+    result = apply_patch(patch, develop, level=1, log_ok=False, from_pr=pr, obliged=force_patch)
```
boegel: `force_patch` doens't make sense, you're actually just ignoring a patch failure...

easybuild/tools/github.py
```diff
@@ -324,6 +336,24 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
+    if not result:
+        # patching with develop failed, remove downloaded files
+        shutil.rmtree(develop)
+        os.remove(os.path.join(path, GITHUB_EB_MAIN, 'develop.tar.gz'))
+        os.remove(patch_path)
```
boegel: wrap this in `try`/`except OSError`

boegel:
@Caylo I briefly discussed this with @wpoely86, and he made a good point that we shouldn't just try to blindly apply the patch and fall back to using the PR branch as is, since we can actually *check* whether it makes sense to try and apply the patch...

If the PR is not merged yet, applying the patch to `develop` should work; if not, it means the PR is broken due to a conflict with `develop`, in which case it's OK to 'fail'. You can even detect the latter via the GitHub API too.

If the PR is already merged, there's no point in trying to apply the patch to `develop` at all, so then you could fall back to using the files in the PR as-is. Or we could (try to) use the latest version from `develop` instead (but maybe not by default).

Long story short: this needs a bit more thought... :)

wpoely86:
- Merged -> develop downloaden
- Open & Mergeable -> develop + patch
- Other: files in PR downloaden (like current) + big fat warning

easybuild/tools/filetools.py
```diff
@@ -718,9 +718,10 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
         _log.debug("Using specified patch level %d for patch %s" % (level, patch_file))
 
     patch_cmd = "patch -b -p%s -i %s" % (level, apatch)
-    result = run.run_cmd(patch_cmd, simple=True, path=adest)
+    result = run.run_cmd(patch_cmd, simple=True, path=adest, log_ok=False)
+
     if not result:
-        raise EasyBuildError("Patching with patch %s failed", patch_file)
+        raise EasyBuildError("Couldn't apply patch")
```
boegel: keep mentioning name of patch file, and maybe also grab the output from the `patch` command and include it in the error message (you'll need to use `simple=False` in `run_cmd` and do `out, ec = run_cmd(...)`

easybuild/tools/github.py
```diff
@@ -332,14 +334,32 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
         raise EasyBuildError("Failed to get data for PR #%d from %s/%s (status: %d %s)",
                              pr, GITHUB_EB_MAIN, GITHUB_EASYCONFIGS_REPO, status, pr_data)
 
-    # 'clean' on successful (or missing) test, 'unstable' on failed tests
+    dl_dev = False
+    # if PR is open and mergable, download develop and patch
     stable = pr_data['mergeable_state'] == GITHUB_MERGEABLE_STATE_CLEAN
+    closed = pr_data['state'] == GITHUB_STATE_CLOSED and pr_data['merged'] == False
+    # 'clean' on successful (or missing) test, 'unstable' on failed tests
```
boegel: also unstable on merge conflict?

easybuild/tools/github.py
```diff
@@ -332,14 +334,32 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
         raise EasyBuildError("Failed to get data for PR #%d from %s/%s (status: %d %s)",
                              pr, GITHUB_EB_MAIN, GITHUB_EASYCONFIGS_REPO, status, pr_data)
 
-    # 'clean' on successful (or missing) test, 'unstable' on failed tests
+    dl_dev = False
+    # if PR is open and mergable, download develop and patch
     stable = pr_data['mergeable_state'] == GITHUB_MERGEABLE_STATE_CLEAN
+    closed = pr_data['state'] == GITHUB_STATE_CLOSED and pr_data['merged'] == False
```
boegel: this looks weird... merged implies closed (sort of), so what's with the last condition which implies merge == not closed?

easybuild/tools/github.py
```diff
@@ -332,14 +334,32 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
+    # 'clean' on successful (or missing) test, 'unstable' on failed tests
     if not stable:
         _log.warning("Mergeable state for PR #%d is not '%s': %s.",
                      pr, GITHUB_MERGEABLE_STATE_CLEAN, pr_data['mergeable_state'])
 
-    for key, val in sorted(pr_data.items()):
-        _log.debug("\n%s:\n\n%s\n" % (key, val))
+    if stable and not closed:
+        # wether merged or not, download develop
```
boegel: whether :)

easybuild/tools/github.py
```diff
@@ -332,14 +334,32 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
+        # not merged: download patch
+        if not pr_data['merged']:
+            patch_name = 'patch%s' % pr
+            patch_url = URL_SEPARATOR.join([GITHUB_URL, GITHUB_EB_MAIN, GITHUB_EASYCONFIGS_REPO, GITHUB_PR, '%s.patch' % pr])
+            patch_path = os.path.join(path, patch_name)
+            patch = download_file(patch_name, patch_url, patch_path)
+            if patch is None:
+                raise EasyBuildError("Failed to download file %s to %s", patch_url, patch_path)
+
+            patched = apply_patch(patch, develop, level=1)
```
boegel: don't be silent on else, at least include some debug logging (maybe also printing?)

easybuild/tools/github.py
```diff
@@ -332,14 +334,32 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
-        _log.debug("\n%s:\n\n%s\n" % (key, val))
+    if stable and not closed:
+        # wether merged or not, download develop
+        develop = download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='develop', path=path)
+        if not os.path.isdir(develop):
+            raise EasyBuildError("Downloading of develop branch failed: not found in %s", develop)
+        dl_dev = True
+
+        # not merged: download patch
+        if not pr_data['merged']:
```
boegel: `not closed` implies `pr_data['merged'] == Ture` currently, so how...?

easybuild/tools/github.py
boegel: also, I think you mean `not pr_data['merged']` ;)

easybuild/tools/github.py
```diff
@@ -332,14 +334,32 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
+            patch_name = 'patch%s' % pr
+            patch_url = URL_SEPARATOR.join([GITHUB_URL, GITHUB_EB_MAIN, GITHUB_EASYCONFIGS_REPO, GITHUB_PR, '%s.patch' % pr])
+            patch_path = os.path.join(path, patch_name)
+            patch = download_file(patch_name, patch_url, patch_path)
+            if patch is None:
+                raise EasyBuildError("Failed to download file %s to %s", patch_url, patch_path)
```
boegel: we're already downloading the patch via `diff_url` below, avoid duplicate code please :)

and use `pr_date['diff_url']` rather than composing the patch URL yourself, i.e. use the GitHub API :)

easybuild/tools/github.py
```diff
@@ -332,14 +334,32 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
         _log.warning("Mergeable state for PR #%d is not '%s': %s.",
                      pr, GITHUB_MERGEABLE_STATE_CLEAN, pr_data['mergeable_state'])
 
-    for key, val in sorted(pr_data.items()):
-        _log.debug("\n%s:\n\n%s\n" % (key, val))
+    if stable and not closed:
+        # wether merged or not, download develop
+        develop = download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='develop', path=path)
+        if not os.path.isdir(develop):
+            raise EasyBuildError("Downloading of develop branch failed: not found in %s", develop)
```
boegel: also mention repo name in error

easybuild/tools/github.py
```diff
@@ -332,14 +334,32 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
     if not stable:
         _log.warning("Mergeable state for PR #%d is not '%s': %s.",
                      pr, GITHUB_MERGEABLE_STATE_CLEAN, pr_data['mergeable_state'])
 
-    for key, val in sorted(pr_data.items()):
-        _log.debug("\n%s:\n\n%s\n" % (key, val))
+    if stable and not closed:
+        # wether merged or not, download develop
+        develop = download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='develop', path=path)
+        if not os.path.isdir(develop):
```
boegel: can't `download_repo` return `None`? if so, `os.path.isdir` is not going to like that...

easybuild/tools/github.py
```diff
@@ -351,6 +371,9 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
     patched_files = det_patched_files(txt=diff_txt, omit_ab_prefix=True, github=True, filter_deleted=True)
     _log.debug("List of patched files: %s" % patched_files)
 
+    for key, val in sorted(pr_data.items()):
+        _log.debug("\n%s:\n\n%s\n" % (key, val))
```
boegel: if this will be retained, make the log messages a bit more descriptive (mention what is being printed here)

easybuild/tools/github.py
```diff
@@ -360,14 +383,16 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
-    # obtain most recent version of patched files
-    for patched_file in patched_files:
-        # path to patch file, incl. subdir it is in
-        fn = os.path.sep.join(patched_file.split(os.path.sep)[-2:])
-        sha = last_commit['sha']
-        full_url = URL_SEPARATOR.join([GITHUB_RAW, GITHUB_EB_MAIN, GITHUB_EASYCONFIGS_REPO, sha, patched_file])
-        _log.info("Downloading %s from %s" % (fn, full_url))
-        download_file(fn, full_url, path=os.path.join(path, fn), forced=True)
+    if not stable or closed:
+        print "WARNING: Using unstable/closed PR on easybuild installation!"
```
boegel: the `on easybuild installation` is a bit weird, maybe go with `\n*** WARNING: Using easyconfigs from unstable/closed PR #<id>***\n` (where you replace `<id>` with the PR ID)

easybuild/tools/github.py
```diff
@@ -376,10 +401,21 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
+            raise EasyBuildError("Couldn't find file in %s: %s", path, patched)
+
+    if dl_dev:
+        # if we merged with develop, the paths to look for are different
+        path = develop
+        all_files = patched_files
 
-    ec_files = [os.path.join(path, f) for f in tmp_files]
+    ec_files = []
+    for f in all_files:
```
boegel: don't use single-letter variable names outside of list comprehensions please

easybuild/tools/github.py
Caylo: `state` is either open or closed - merged always means closed, but closed doesn't always mean merged. unmerged closed prs need to be treated diffently, that's why the extra check is there.

easybuild/tools/github.py
Caylo: `closed` implies not merged **and** closed. this is when the pr is not merged, but still open

easybuild/tools/filetools.py
```diff
@@ -718,10 +718,12 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
         _log.debug("Using specified patch level %d for patch %s" % (level, patch_file))
 
     patch_cmd = "patch -b -p%s -i %s" % (level, apatch)
-    result = run.run_cmd(patch_cmd, simple=True, path=adest)
-    if not result:
-        raise EasyBuildError("Patching with patch %s failed", patch_file)
+    out, ec = run.run_cmd(patch_cmd, simple=False, path=adest, log_ok=False)
 
+    if ec:
+        raise EasyBuildError("Couldn't apply patch file %s. Process exited with code %s: %s", patch_file, out, result)
```
boegel: `result` is undefined here?

this points out that the tests should be enhanced to check for a failing patch, which should raise this error (rather than a traceback)

easybuild/tools/github.py
```diff
@@ -332,14 +334,22 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
                      pr, GITHUB_MERGEABLE_STATE_CLEAN, pr_data['mergeable_state'])
 
-    for key, val in sorted(pr_data.items()):
-        _log.debug("\n%s:\n\n%s\n" % (key, val))
+    if (stable or pr_data['merged']) and not closed:
+        # whether merged or not, download develop
+        develop = download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='develop', path=path)
+        if not os.path.isdir(develop):
+            raise EasyBuildError("Downloading of %s/develop branch failed: not found in %s",
+                                  GITHUB_EASYCONFIGS_REPO, develop)
```
boegel: wouldn't it be better to hoist this check into `download_repo`?
or, if a check like this is already there, just drop this check here

easybuild/tools/github.py
boegel: ok

easybuild/tools/github.py
```diff
@@ -360,14 +373,17 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
-    for patched_file in patched_files:
-        # path to patch file, incl. subdir it is in
-        fn = os.path.sep.join(patched_file.split(os.path.sep)[-2:])
-        sha = last_commit['sha']
-        full_url = URL_SEPARATOR.join([GITHUB_RAW, GITHUB_EB_MAIN, GITHUB_EASYCONFIGS_REPO, sha, patched_file])
-        _log.info("Downloading %s from %s" % (fn, full_url))
-        download_file(fn, full_url, path=os.path.join(path, fn), forced=True)
+    if not(pr_data['merged']):
+        if not stable or closed:
+            print "\n*** WARNING: Using easyconfigs from unstable/closed PR #%s***\n" % pr
```
boegel: nitpicking: add space before last `***` to make message more readable

easybuild/tools/github.py
```diff
@@ -376,14 +392,22 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
 
-    ec_files = [os.path.join(path, f) for f in tmp_files]
+    ec_files = []
+    for patched in patched_files:
+        if os.path.exists(os.path.join(path, patched)):
+            ec_files.append(os.path.join(path, patched))
+        else:
+            raise EasyBuildError("Coudln't find path to patched file %s", os.path.join(path, f))
 
+    print ec_files
```
boegel: do we want to print this?

easybuild/tools/github.py
```diff
@@ -376,14 +392,22 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
+        if not patched in tmp_files:
+            raise EasyBuildError("Couldn't find file in %s: %s", path, patched)
 
-    ec_files = [os.path.join(path, f) for f in tmp_files]
+    ec_files = []
+    for patched in patched_files:
+        if os.path.exists(os.path.join(path, patched)):
+            ec_files.append(os.path.join(path, patched))
+        else:
+            raise EasyBuildError("Coudln't find path to patched file %s", os.path.join(path, f))
```
boegel: isn't this the same check as above? checking once is enough (keep this one, drop the separate loop above?)

test/framework/github.py
```diff
@@ -132,7 +132,9 @@ def test_fetch_easyconfigs_from_pr(self):
         try:
             ec_files = gh.fetch_easyconfigs_from_pr(2481, path=tmpdir, github_user=GITHUB_TEST_ACCOUNT)
             self.assertEqual(all_ecs, sorted([os.path.basename(f) for f in ec_files]))
-            self.assertEqual(all_ecs, sorted([os.path.basename(f) for f in glob.glob(os.path.join(tmpdir, '*', '*'))]))
+            tmp_files = []
+            for (dirpath, _, filenames) in os.walk(tmpdir):
+                tmp_files.extend([f for f in filenames])
```
boegel: aren't we missing an assert here? why is `tmp_files` constructed?

test/framework/sandbox/sources/toy/toy-0.0_broken.patch
```diff
@@ -0,0 +1,10 @@
+--- a/toy-0.0/toy.source.orig   2013-11-30 14:43:48.000000000 +0100
```
boegel: can you add a line on top describing why this patch is broken, and make clear it was done deliberately (to test apply_patch)

test/framework/filetools.py
```diff
@@ -694,6 +694,22 @@ def test_derive_alt_pypi_url(self):
         url = 'https://pypi.python.org/packages/source/n/nosuchpackageonpypiever/nosuchpackageonpypiever-0.0.0.tar.gz'
         self.assertEqual(ft.derive_alt_pypi_url(url), None)
 
+    def test_apply_patch(self):
+        """ Test apply_patch """
+        testdir = os.path.dirname(os.path.abspath(__file__))
+        tmpdir = tempfile.mkdtemp()
```
boegel: use `self.test_prefix` (which is cleaned up automagically after the test runs, whether it succeeds or not)

test/framework/filetools.py
```diff
@@ -694,6 +694,22 @@ def test_derive_alt_pypi_url(self):
...
+        toy_patch = os.path.join(testdir, 'sandbox', 'sources', 'toy', 'toy-0.0_typo.patch')
+        toy_patch_broken = os.path.join(testdir, 'sandbox', 'sources', 'toy', 'toy-0.0_broken.patch')
+
+        self.assertTrue(ft.apply_patch(toy_patch, path))
+        patched = ft.read_file(os.path.join(path, 'toy.source'))
+        pattern = "I'm a toy, and very proud of it"
+        self.assertTrue(pattern in patched)
+
+        # faulty patch file
+        self.assertErrorRegex(EasyBuildError, "Couldn't apply patch file", ft.apply_patch, toy_patch_broken, path)
```
boegel: do you actually need an additional patch file? just try applying the same patch twice, 2nd time will/should fail?

easybuild/tools/github.py
boegel: ok now, thx

easybuild/tools/github.py
```diff
@@ -332,14 +335,18 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
+    # 'clean' on successful (or missing) test, 'unstable' on failed tests or merge conflict
     if not stable:
         _log.warning("Mergeable state for PR #%d is not '%s': %s.",
                      pr, GITHUB_MERGEABLE_STATE_CLEAN, pr_data['mergeable_state'])
 
-    for key, val in sorted(pr_data.items()):
-        _log.debug("\n%s:\n\n%s\n" % (key, val))
+    if (stable or pr_data['merged']) and not closed:
+        # whether merged or not, download develop
+        path = download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='develop', path=path)
```
boegel: hold on, where are we applying the patch from the PR on `develop`?

boegel: ok, fixed below now

easybuild/tools/github.py
```diff
@@ -360,30 +369,34 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
-        _log.info("Downloading %s from %s" % (fn, full_url))
-        download_file(fn, full_url, path=os.path.join(path, fn), forced=True)
-
-    # sanity check: make sure all patched files are downloaded
-    all_files = [os.path.sep.join(f.split(os.path.sep)[-2:]) for f in patched_files]
-
-    tmp_files = []
-    for (dirpath, _, filenames) in os.walk(path):
-        tmp_files.extend([os.path.join(os.path.basename(dirpath), f) for f in filenames])
+    if not(pr_data['merged']):
```
boegel: and what if the PR *is* merged?

then you should also just use what's in `develop` (but without printing a warning)

boegel: ah, wait, nvm, then you are just picking from `develop` already, fine

boegel:
lgtm, going in, thanks @Caylo!

PR #1830, opened 2016-07-04
add --avail-toolchain-opts

Caylo:
see https://github.com/hpcugent/easybuild/issues/66

**eb --avail-toolchain opts intel**
```
Available options for intel toolchain:
    32bit: Compile 32bit target (default: False)
    cciscxx: Use CC as CXX (default: False)
    cstd: Specify C standard (default: None)
    debug: Enable debug (default: False)
    defaultopt: Default compiler optimizations (default: False)
    defaultprec: Default precision (default: False)
    error-unknown-option: Error instead of warning for unknown options (default: False)
    i8: Integers are 8 byte integers (default: False)
    intel-static: Link Intel provided libraries statically (default: False)
    loose: Loose precision (default: False)
    lowopt: Low compiler optimizations (default: False)
    no-icc: Don't set Intel specific macros (default: False)
    noopt: Disable compiler optimizations (default: False)
    openmp: Enable OpenMP (default: False)
    opt: High compiler optimizations (default: False)
    optarch: Enable architecture optimizations (default: True)
    packed-linker-options: Pack the linker options as comma separated list (default: False)
    pic: Use PIC (default: False)
    precise: High precision (default: False)
    r8: Real is 8 byte real (default: False)
    shared: Build shared library (default: False)
    static: Build static library (default: False)
    strict: Strict (highest) precision (default: False)
    unroll: Unroll loops (default: False)
    usempi: Use MPI compiler as default compiler (default: False)
    verbose: Verbose output (default: False)
    veryloose: Very loose precision (default: False)
```

**eb --avail-toolchain opts intel --output-format=rst**
```
Available options for intel toolchain:
--------------------------------------

=========================    ===============================================    =========
option                       description                                        default  
=========================    ===============================================    =========
``32bit``                    Compile 32bit target                               ``False``
``cciscxx``                  Use CC as CXX                                      ``False``
``cstd``                     Specify C standard                                 ``None`` 
``debug``                    Enable debug                                       ``False``
``defaultopt``               Default compiler optimizations                     ``False``
``defaultprec``              Default precision                                  ``False``
``error-unknown-option``     Error instead of warning for unknown options       ``False``
``i8``                       Integers are 8 byte integers                       ``False``
``intel-static``             Link Intel provided libraries statically           ``False``
``loose``                    Loose precision                                    ``False``
``lowopt``                   Low compiler optimizations                         ``False``
``no-icc``                   Don't set Intel specific macros                    ``False``
[...]

PR #1839, opened 2016-07-08
add attribute check for toolchainopts documentation

Caylo:
fixes a bug in #1830 for toolchains without MPI options

### DIFF
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -564,7 +564,14 @@ def avail_toolchain_opts(name, output_format=FORMAT_TXT):
         raise EasyBuildError("Couldn't find toolchain: '%s'. To see available toolchains, use --list-toolchains" % name)
     tc = tc_class(version='1.0') # version doesn't matter here, but needs to be defined
 
-    options = [tc.COMPILER_SHARED_OPTS, tc.COMPILER_UNIQUE_OPTS, tc.MPI_SHARED_OPTS, tc.MPI_UNIQUE_OPTS]
+    mpishared = []
+    mpiunique = []
+    if hasattr(tc, 'MPI_SHARED_OPTS'):
+        mpishared = tc.MPI_SHARED_OPTS
+    if hasattr(tc, 'MPI_UNIQUE_OPTS'):
+        mpiunique = tc.MPI_UNIQUE_OPTS
+
+    options = [tc.COMPILER_SHARED_OPTS, tc.COMPILER_UNIQUE_OPTS, mpishared, mpiunique]
 
     tc_dict = {}
     for opt in options:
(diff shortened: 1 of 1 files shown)


easybuild/tools/docs.py
```diff
@@ -564,7 +564,14 @@ def avail_toolchain_opts(name, output_format=FORMAT_TXT):
         raise EasyBuildError("Couldn't find toolchain: '%s'. To see available toolchains, use --list-toolchains" % name)
     tc = tc_class(version='1.0') # version doesn't matter here, but needs to be defined
 
-    options = [tc.COMPILER_SHARED_OPTS, tc.COMPILER_UNIQUE_OPTS, tc.MPI_SHARED_OPTS, tc.MPI_UNIQUE_OPTS]
+    mpishared = []
+    mpiunique = []
```
boegel: this won't work, since you use these values in `.update(..)`, so they should be `{}` instead?

easybuild/tools/docs.py
```diff
@@ -564,7 +564,14 @@ def avail_toolchain_opts(name, output_format=FORMAT_TXT):
...
+    if hasattr(tc, 'MPI_SHARED_OPTS'):
+        mpishared = tc.MPI_SHARED_OPTS
+    if hasattr(tc, 'MPI_UNIQUE_OPTS'):
+        mpiunique = tc.MPI_UNIQUE_OPTS
+
+    options = [tc.COMPILER_SHARED_OPTS, tc.COMPILER_UNIQUE_OPTS, mpishared, mpiunique]
```
boegel: I would do:

```python
options = []
for cst in ['COMPILER_SHARED_OPTS', 'COMPILER_UNIQUE_OPTS', 'MPI_SHARED_OPTS', 'MPI_UNIQUE_OPTS']:
    if hasattr(tc, cst):
        options.append(getattr(tc, cst))
```

or even:

```python
tc_dict = {}
for cst in ['COMPILER_SHARED_OPTS', 'COMPILER_UNIQUE_OPTS', 'MPI_SHARED_OPTS', 'MPI_UNIQUE_OPTS']:
    if hasattr(tc, cst):
        opts = getattr(tc, cst)
        if opts is not None:
            tc_dict.update(opts)
```

easybuild/tools/docs.py
```diff
@@ -564,19 +564,21 @@ def avail_toolchain_opts(name, output_format=FORMAT_TXT):
         raise EasyBuildError("Couldn't find toolchain: '%s'. To see available toolchains, use --list-toolchains" % name)
     tc = tc_class(version='1.0') # version doesn't matter here, but needs to be defined
 
-    options = [tc.COMPILER_SHARED_OPTS, tc.COMPILER_UNIQUE_OPTS, tc.MPI_SHARED_OPTS, tc.MPI_UNIQUE_OPTS]
-
+    mpishared = []
+    mpiunique = []
```
boegel: @Caylo these's are longer used, so please remove?

boegel:
Thanks @Caylo!

PR #1841, opened 2016-07-12
add support for --use-ccache (WIP)

boegel:
TODO:

* [ ] allow controlling location of ccache cache (argument to `--use-ccache`?)
* [ ] tests

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -148,6 +148,7 @@ BUILD_OPTIONS_CMDLINE = {
         'sticky_bit',
         'upload_test_report',
         'update_modules_tool_cache',
+        'use_ccache',
         'use_existing_modules',
     ],
     True: [
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -305,6 +305,7 @@ class EasyBuildOptions(GeneralOption):
                       None, 'store', None),
             'update-modules-tool-cache': ("Update modules tool cache file(s) after generating module file",
                                           None, 'store_true', False),
+            'use-ccache': ("Enable use of ccache to speed up compilation", None, 'store_true', False),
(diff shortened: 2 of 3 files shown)


boegel:
@Caylo: ccache only supports C/C++, not Fortran, but there are alternatives available, for example http://people.irisa.fr/Edouard.Canot/f90cache

So, `--use-ccache` should probably be renamed to `--use-compiler-cache`, and for Fortran90 we should use f90cache instead.

Do note that it works differently: rather than simply symlinking the original compiler commands, you need to prefix them, so you'll need to create a wrapper script that does this...

boegel:
@Caylo just ran into this while playing around with this WIP branch, please fix? :)

```
  File "/data/EasyBuild/easybuild-framework/easybuild/main.py", line 114, in build_and_install_software
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/framework/easyblock.py", line 2363, in build_and_install_one
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/framework/easyblock.py", line 2279, in run_all_steps
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/framework/easyblock.py", line 2158, in run_step
    step_method(self)()
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/framework/easyblock.py", line 1554, in prepare_step
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/tools/toolchain/toolchain.py", line 647, in prepare
    self.symlink_compilers(ccache)
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/tools/toolchain/toolchain.py", line 602, in symlink_compilers
    os.symlink(path, os.path.join(tmpdir, comp))
OSError: [Errno 17] File exists
```

PR #1844, opened 2016-07-13
ccache with argument + support f90cache

Caylo:
(ref #1841 )


easybuild/tools/options.py
```diff
@@ -305,6 +305,7 @@ def override_options(self):
                       None, 'store', None),
             'update-modules-tool-cache': ("Update modules tool cache file(s) after generating module file",
                                           None, 'store_true', False),
+            'use-compiler-cache': ("Enable use of ccache to speed up compilation", str, 'store', False),
```
boegel: also mention f90cache in description

boegel:
@Caylo make this WIP in the PR title since this doesn't cover `f90cache` yet?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -48,6 +49,9 @@
 
 _log = fancylogger.getLogger('tools.toolchain', fname=False)
 
+CCAHCE = 'ccache'
```
boegel: typo: `CCACHE` :)

easybuild/tools/toolchain/toolchain.py
```diff
@@ -620,6 +649,20 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
                 self.generate_vars()
                 self._setenv_variables(onlymod, verbose=not silent)
 
+        if build_option('use_compiler_cache'):
+            paths = {}
+            for cache in [CCACHE, F90CACHE]:
+                cache_path = which(cache)
+                if cache_path is None:
+                    raise EasyBuildError("%s binary not found in $PATH", cache)
+                paths[cache] = cache_path
```
boegel: use `else`?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -620,6 +649,20 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
                 self.generate_vars()
                 self._setenv_variables(onlymod, verbose=not silent)
 
+        if build_option('use_compiler_cache'):
+            paths = {}
+            for cache in [CCACHE, F90CACHE]:
+                cache_path = which(cache)
+                if cache_path is None:
+                    raise EasyBuildError("%s binary not found in $PATH", cache)
```
boegel: mention: `, require by `--use-compiler-cache`

easybuild/tools/toolchain/toolchain.py
```diff
@@ -620,6 +649,20 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
...
+                paths[cache] = cache_path
+
+            for cachename in ["CCACHE", "F90CACHE"]:
+                os.environ["%s_DIR" % cachename] = build_option('use_compiler_cache')
```
boegel: only call out to `build_option` once, so do `compiler_cache = build_option('use_compiler_cache')` above and then use `compiler_cache` instead

easybuild/tools/toolchain/toolchain.py
```diff
@@ -620,6 +649,20 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
...
+                os.environ["%s_TEMPDIR" % cachename] = os.path.join(build_option('use_compiler_cache'), 'tmp')
```
boegel: I would use `tempfile.mkdtemp()` here (which gets cleaned up automagically)

easybuild/tools/toolchain/toolchain.py
```diff
@@ -584,6 +588,31 @@ def _verify_toolchain(self):
...
+        """
+        tmpdir = tempfile.mkdtemp()
+        compilers = {}
+
+        if self.name == DUMMY_TOOLCHAIN_NAME:
+            compilers[CCACHE] = ['gcc', 'g++']
+            compilers[F90CACHE] = ['gfortran']
+        else:
+            compilers[CCACHE] = [self.COMPILER_CC, self.COMPILER_CXX]
+            compilers[F90CACHE] = [self.COMPILER_F77, self.COMPILER_F90, self.COMPILER_FC]
```
boegel: this should be made more generic, since now `symlink_compilers` is specific to the compiler caching...

I think the argument passed to `symlink_compilers` should contain all information of which commands should be linked to which path, so something like:

```python
{
    'ccache': (`/path/to/ccache`, ['gcc', 'g++']),
    'f90cache': (`/path/to/f90cache`, ['gfortran']),
}
```

so constructing this should be done in the `if compiler_cache` in prepare, where we should call out to a separate function rather than blowing up the `prepare` method (especially since more stuff will be added there later for rpath too)

easybuild/tools/toolchain/toolchain.py
```diff
@@ -584,6 +586,24 @@ def _verify_toolchain(self):
...
+        @param paths: dictionary containing mapping from types of caches (ccache, f90cache) to
+                      tuple ('path/to/cache', [commands to symlink to this cache])
+        """
+        tmpdir = tempfile.mkdtemp()
+
+        for key, links in paths.items():
+            for comp in links[1]:
+                comp_s = os.path.join(tmpdir, comp)
+                if not os.path.exists(comp_s):
+                    os.symlink(links[0], comp_s)
```
boegel: @Caylo please wrap this in a `try`/`except`:

```python
try:
    os.symlink(...)
except OSError as err:
    raise EasyBuildError("Failed to symlink %s to %s: %s", links[0], comp_s, err)
```

easybuild/tools/toolchain/toolchain.py
```diff
@@ -620,6 +640,39 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
                 self.generate_vars()
                 self._setenv_variables(onlymod, verbose=not silent)
 
+        if build_option('use_compiler_cache'):
+            self.prepare_compiler_cache()
+
+    def prepare_compiler_cache(self):
```
boegel: missing docstring

easybuild/tools/toolchain/toolchain.py
```diff
@@ -584,6 +586,24 @@ def _verify_toolchain(self):
...
+    def symlink_compilers(self, paths):
+        """
+        Create a symlink for each compiler to binary/script at specified path.
+
+        @param paths: dictionary containing mapping from types of caches (ccache, f90cache) to
+                      tuple ('path/to/cache', [commands to symlink to this cache])
+        """
+        tmpdir = tempfile.mkdtemp()
+
+        for key, links in paths.items():
```
boegel: I would do `for _, (path, comps) in paths.items()` to make this more readable?

also, key is not used?!

easybuild/tools/toolchain/toolchain.py
```diff
@@ -584,6 +586,24 @@ def _verify_toolchain(self):
             raise EasyBuildError("List of toolchain dependency modules and toolchain definition do not match "
                                  "(found %s vs expected %s)", self.toolchain_dep_mods, toolchain_definition)
 
+    def symlink_compilers(self, paths):
+        """
+        Create a symlink for each compiler to binary/script at specified path.
+
+        @param paths: dictionary containing mapping from types of caches (ccache, f90cache) to
```
boegel: `:param` ;)

easybuild/tools/toolchain/toolchain.py
```diff
@@ -620,6 +640,39 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
...
+        if self.name == DUMMY_TOOLCHAIN_NAME:
+            gcc = ['gcc', 'g++']
+            fortr =  ['gfortran']
+        else:
+            gcc = [self.COMPILER_CC, self.COMPILER_CXX]
+            fortr = [self.COMPILER_F77, self.COMPILER_F90, self.COMPILER_FC]
+
+        compilers = {
+            CCACHE : gcc,
+            F90CACHE : fortr,
```
boegel: please change `gcc` to `c_comps` and `fortr` to `fortran_comps`

easybuild/tools/toolchain/toolchain.py
```diff
@@ -620,6 +640,39 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
...
+
+        for cache in [CCACHE, F90CACHE]:
+            cache_path = which(cache)
+            if cache_path is None:
+                raise EasyBuildError("%s binary not found in $PATH, required by --use-compiler-cache", cache)
+            else:
+                paths[cache] = (cache_path, compilers[cache])
+
+        # set paths for caches
+        compiler_cache = build_option('use_compiler_cache')
```
boegel: make it clear this is a path, so maybe name this `comp_cache_path`

test/framework/toolchain.py
```diff
@@ -816,6 +820,48 @@ def test_pgi_imkl(self):
...
+            os.chmod(fn, st.st_mode | stat.S_IEXEC)
+            setvar('PATH', '%s:%s' % (path, os.getenv('PATH')))
+
+
+        eb_file = os.path.join(os.path.dirname(__file__), 'easyconfigs', 'toy-0.0.eb')
+
+        args = [
+            eb_file,
+            "--use-compiler-cache",
+            self.test_prefix,
```
boegel: maybe change this to `"--use-compiler-cache=%s" % self.test_prefix` which makes it more explicit that `--use-compiler-cache` takes an argument

test/framework/toolchain.py
```diff
@@ -816,6 +820,48 @@ def test_pgi_imkl(self):
...
+            "--force",
+            "--debug",
+            ]
+
+        out = self.eb_main(args, raise_error=True, do_build=True)
+
+        patterns = [
+            "This is a ccache wrapper",
+            "Command %s found at %s" % (cachename, path)
```
boegel: careful with possible symlinks, make this:

```python
"Command %s found at .*%s" % (cachename, os.path.dirname(path))
```

test/framework/toolchain.py
```diff
@@ -816,6 +820,48 @@ def test_pgi_imkl(self):
...
+        ]
+        self.assertTrue(all([pattern in out for pattern in patterns]))
```
boegel: replace `pattern in out` with `re.search(pattern, out)` since you'll be using `.*`...

boegel:
lgtm, will merge when Travis is happy with it, thanks @Caylo!

Issue #1757, opened 2016-05-11
--check-github fails with old GitPython versions


pescobar:
I tried `eb --github-user=pescobar --check-github` and I was getting this error:

```
Checking status of GitHub integration...

Making sure we're online... OK

* GitHub user... pescobar => OK
* GitHub token... fa1..5f5 (len: 40) => OK (validated)
* git command... OK ("git version 1.8.5.6; ")
* GitPython module... OK
* push access to pescobar/easybuild-easyconfigs repo @ GitHub... FAIL (unexpected exception: 'Remote' object has no attribute 'exists')
* creating gists... OK
* location to Git working dirs...  not found (suboptimal)
```

I was using python 2.6 and GitPython from centos6 epel repository `GitPython-0.3.2-0.6.RC1.el6.noarch`

After upgrading my GitPython installation doing `easy_install -U --prefix=$HOME/.local GitPython` the issue is fixed. It seems there is an problem with old GitPython versions. After the upgrade it's working:

```
All checks PASSed!

Status of GitHub integration:
* --from-pr: OK
* --new-pr: OK
* --review-pr: OK
* --update-pr: OK
* --upload-test-report: OK
```

boegel:
for future reference: latest GitPython version is 2.0.2 (i.e. the one that @pescobar is using) at this time

boegel:
fixed with #1847

PR #1847, opened 2016-07-15
better error handling for outdated gitpython module

Caylo:
ref #1757

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -899,7 +899,7 @@ def check_github():
             check_res = "OK"
     elif github_user:
         if LooseVersion(git.__version__) < LooseVersion('1.0'):
-            check_res = "FAIL (outdated gitpython module, try updating first)"
+            check_res = "FAIL (outdated GitPython module, try updating first)"
         else:
             check_res = "FAIL (unexpected exception: %s)" % push_err
     else:
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -898,10 +898,9 @@ def check_github():
         else:
             check_res = "OK"
     elif github_user:
-        req_gp_ver = '1.0'
-        if LooseVersion(git.__version__) < LooseVersion(req_gp_ver):
-            check_res = "FAIL (GitPython version %s is too old, should be version %s or newer)" % (git.__version__,
-                         req_gp_ver)
+        gp_ver, req_gp_ver = git.__version__, '1.0'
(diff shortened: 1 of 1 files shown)


easybuild/tools/github.py
```diff
@@ -896,7 +896,10 @@ def check_github():
         else:
             check_res = "OK"
     elif github_user:
-        check_res = "FAIL (unexpected exception: %s)" % push_err
+        if sys.modules['git'].__version__ < "1.0.0":
```
boegel: please use `LooseVersion` when comparing versions

also, just use `git.__version__`?

```
if LooseVersion(git.__version__) < LooseVersion('1.0'):
```

Did you verify whether things (should) work with GitPython 1.0 and up?

Caylo: Yup, verified that. Everything above 1.0 works.

easybuild/tools/github.py
```diff
@@ -896,7 +898,10 @@ def check_github():
         else:
             check_res = "OK"
     elif github_user:
-        check_res = "FAIL (unexpected exception: %s)" % push_err
+        if LooseVersion(git.__version__) < LooseVersion('1.0'):
+            check_res = "FAIL (outdated gitpython module, try updating first)"
```
boegel: can you clarify 'outdated' in the error, i.e. mention what the minimal version is? and also mention the what version was found

something like

```python
check_res = "FAIL (GitPython version %s is too old, should be version %s or newer)" % (git.__version__, req_gp_ver)
```

easybuild/tools/github.py
```diff
@@ -896,7 +898,12 @@ def check_github():
         else:
             check_res = "OK"
     elif github_user:
-        check_res = "FAIL (unexpected exception: %s)" % push_err
+        req_gp_ver = '1.0'
+        if LooseVersion(git.__version__) < LooseVersion(req_gp_ver):
+            check_res = "FAIL (GitPython version %s is too old, should be version %s or newer)" % (git.__version__,
+                         req_gp_ver)
```
boegel: bleh on the wrapping here...

maybe like this?

```python
gp_ver, req_gp_ver = git.__version__, '1.0'
if LooseVersion(gp_ver) < LooseVersion(req_gp_ver):
    check_res = "FAIL (GitPython version %s is too old, should be version %s or newer)" % (gp_ver, req_gp_ver)
```

boegel:
Going in, thanks @Caylo!

Issue #1674, opened 2016-03-17
--new-pr could be smarter


ocaisa:
It would be nice if `--new-pr` was smart enough to also pull in any (new or altered) dependencies as well as the easyconfig on the command line

boegel:
@ocaisa I enhanced the description a bit, to include other ideas to make `--new-pr` smarter/better, and also made it a checklist...

Let's make this a project issue to track ideas/feature requests for `--new-pr` and `--update-pr`...

boegel:
although not all boxes are checked here yet, the support for `--new-pr` and `--update-pr` is now considered stable (cfr. #1979)

the remaining things to implement will be enhancements to what is supported already, there should be no changes in semantics of the existing functionality of `--new-pr` and `--update-pr`

boegel:
I've enhanced this a bit to include other ideas for improving the GitHub integration beyond `--new-pr` and `--updat-pr`.

akesandgren:
Would be nice if --new-pr would automatically include any required patch files.

PR #1852, opened 2016-07-25
also include patch files in --new-pr and --update-pr

Caylo:
building on #1751 
cfr #1674


easybuild/tools/github.py
```diff
@@ -595,8 +595,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                if patch_file in ec['patches']:
+                    soft_name = ec.name
+                    break
+
+            # FIXME: try harder if software name wasn't found yet?
+
+            if soft_name:
+                patch_specs.append((patch_path, soft_name))
+            else:
+                raise EasyBuildError("Failed to determine software name to which patch file %s relates", patch_path)
```
boegel: @Caylo please hoist this blob of code that determines `patch_specs` into a separate function, it'll help make things clearer since it's rather big (and will get bigger if the FIXME is handled)

easybuild/tools/github.py
```diff
@@ -595,8 +595,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+            soft_name = None
+            patch_file = os.path.basename(patch_path)
+
+            # consider patch lists of easyconfigs being provided
+            for ec in file_info['ecs']:
+                if patch_file in ec['patches']:
+                    soft_name = ec.name
+                    break
+
+            # FIXME: try harder if software name wasn't found yet?
```
boegel: @Caylo hmm, ideally, this should be tackled too...

As it is, only the easyconfig files that are being tweaked are being considered when trying to figure out which software the patch belongs too, but this may not be enough in practice

For example:

```
eb --new-pr foo.eb
# oops, forgot patch
eb --update-pr 1234 foo.patch
```

So, I think we should fall back to scanning all easyconfigs, and check which patches they require... This may take a while, but it's the best we can do imho (we can't rely on the patch name).

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1359,6 +1359,39 @@ def robot_find_minimal_toolchain_of_dependency(dep, modtool, parent_tc=None):
...
+        try:
+            mkdir(os.path.dirname(full_target_path), parents=True)
+            shutil.copy2(path, full_target_path)
+            _log.info("%s copied to %s", path, full_target_path)
+        except OSError as err:
+            raise EasyBuildError("Failed to copy %s to %s: %s", path, target_path, err)
+    else:
+        raise EasyBuildError("Subdirectory %s not found in %s", subdir, target_dir)
+
+    return target_path
```
boegel: this should be `return full_target_path`?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1373,42 +1406,36 @@ def copy_easyconfigs(paths, target_dir):
...
 
-                letter = name.lower()[0]
-                if letter not in a_to_z:
-                    raise EasyBuildError("Don't know which letter subdir to use for %s", name)
+            target_path = _copy_file_to_easyconfigs_dir(path, target_dir, soft_name, ec_filename)
+            full_target_path = os.path.join(target_dir, target_path)
 
-                target_path = os.path.join(subdir, letter, name, ec_filename)
-                _log.debug("Target path for %s: %s", path, target_path)
+            file_info['new'].append(not os.path.exists(full_target_path))
```
boegel: hmm, the `exists` check will always return `True` here, since it was just copied?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1373,42 +1406,36 @@ def copy_easyconfigs(paths, target_dir):
...
+    """
+    Copy patch files to specified directory, in the 'right' location according to the software name they relate to.
 
-    return file_info
+    @param patch_specs: list of tuples with patch file location and name of software they are for
+    @param target_dir: target directory
+    """
+    patched_files = []
+    for patch_path, soft_name in patch_specs:
+        patched_files.append(_copy_file_to_easyconfigs_dir(patch_path, target_dir, soft_name, os.path.basename(patch_path)))
```
boegel: line too long?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1365,6 +1365,34 @@ def robot_find_minimal_toolchain_of_dependency(dep, modtool, parent_tc=None):
...
 
+def det_easyconfig_location(path, target_dir, soft_name, target_file):
+    """
+    Determine path to easyconfigs directory for specified software name, using specified target file name.
+
+    @param path: path of file to copy
+    @param target_dir: (parent) target directory, should contain easybuild/easyconfigs subdirectory
+    @param soft_name: software name (to determine location to copy to)
+    @param target_file: target file name
+    @return: full path to the right location
```
boegel: `s/@param/:param/`

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1379,41 +1407,59 @@ def copy_easyconfigs(paths, target_dir):
...
+    @param patch_specs: list of tuples with patch file location and name of software they are for
+    @param target_dir: target directory
+    """
+    patched_files = {
+        'files': [],
+        'paths_in_repo': [],
+        'new': [],
+    }
+    for patch_path, soft_name in patch_specs:
+        target_path = det_easyconfig_location(patch_path, target_dir, soft_name, os.path.basename(patch_path))
```
boegel: since `det_easyconfig_location` is also used for patch files, we should rename it to `det_location_for`?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1379,41 +1407,59 @@ def copy_easyconfigs(paths, target_dir):
...
+        patched_files['new'].append(not os.path.exists(target_path))
+
+        try:
+            mkdir(os.path.dirname(target_path), parents=True)
+            shutil.copy2(patch_path, target_path)
+            _log.info("%s copied to %s", patch_path, target_path)
+
+        except OSError as err:
+            raise EasyBuildError("Failed to copy %s to %s: %s", patch_path, target_path, err)
```
boegel: this whole block is copy-pasted, let's make a `copy_file` function in `filetools` for this?

I'm sure it can be used in other places as well (but that's beyond the scope of this PR)...

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1379,41 +1407,59 @@ def copy_easyconfigs(paths, target_dir):
...
+
+        patched_files['files'].append(os.path.splitext(os.path.basename(patch_path))[0])
```
boegel: hmm, what's the purpose of this `files` entry?

I'm fully aware that I may have been the one who added this, but it's no longer clear to me what this is about... ;-)

easybuild/tools/github.py
```diff
@@ -617,14 +617,53 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                patch_specs.append((patch_path, soft_name))
+            else:
+                # fall back on scanning all eb files for patches
+                print "Matching easyconfig not found: determining where patch file belongs (this may take a while)..."
+                soft_name = scan_all_easyconfigs(patch_file)
+                if soft_name:
+                    patch_specs.append((patch_path, soft_name))
+                else:
+                    # still nothing found
+                    raise EasyBuildError("Failed to determine software name to which patch file %s relates", patch_path)
```
boegel: this whole block that determines to which software each patch file belongs should be fleshed out in a separate function, including this like this makes `_easyconfigs_pr_common` too big

easybuild/tools/github.py
```diff
@@ -617,14 +617,53 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+            for ec in file_info['ecs']:
+                if patch_file in ec['patches']:
+                    soft_name = ec.name
+                    break
+
+            if soft_name:
+                patch_specs.append((patch_path, soft_name))
+            else:
+                # fall back on scanning all eb files for patches
+                print "Matching easyconfig not found: determining where patch file belongs (this may take a while)..."
```
boegel: mention for which patch file the matching easyconfig was not found, and mention "on the first try" and that you're taking *all* available easyconfigs into account on a 2nd attempt

easybuild/framework/easyconfig/easyconfig.py
Caylo: i needed a replacement for 'ecs' as these are not easyconfig files. for the purpose of the pr title. it can probably be scratched now

easybuild/tools/github.py
```diff
@@ -684,6 +727,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
     return file_info, git_repo, pr_branch, diff_stat
 
 
+def scan_all_easyconfigs(patch_name):
+    """
+    Scan all easyconfigs in the robot path to determine which software a patch file belongs to
+
+    @param patch_name: name of the patch file
+    """
+    robot_path = build_option('robot_path')[0]
```
boegel: why the `[0]`? you're only considering the first entry in the robot search path? why?

easybuild/tools/github.py
```diff
@@ -684,6 +727,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+    soft_name = ""
+    found = False
+    counter = 0
+    for (dirpath, _, filenames) in os.walk(robot_path):
+        if found:
+            break
+        for filename in [fn for fn in filenames if ".eb" in fn]:
+            rawtxt = read_file(os.path.join(dirpath, filename))
+            if "patches" in rawtxt and "TEMPLATE" not in filename:
```
boegel: why skip `TEMPLATE`, it'll never result in a match anyway?

easybuild/tools/github.py
```diff
@@ -684,6 +727,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+    @param patch_name: name of the patch file
+    """
+    robot_path = build_option('robot_path')[0]
+    soft_name = ""
+    found = False
+    counter = 0
+    for (dirpath, _, filenames) in os.walk(robot_path):
+        if found:
+            break
+        for filename in [fn for fn in filenames if ".eb" in fn]:
```
boegel: better check the extension via `os.splitext` or via `.endswith`, and what if it's a `.yeb` easyconfig?

easybuild/tools/github.py
```diff
@@ -684,6 +727,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+            rawtxt = read_file(os.path.join(dirpath, filename))
+            if "patches" in rawtxt and "TEMPLATE" not in filename:
+                try:
+                    ec = EasyConfig(os.path.join(dirpath, filename), validate=False)
+                    ec._generate_template_values()
+                    if patch_name in ec.asdict()['patches']:
+                        soft_name = ec.asdict()['name']
+                        found = True
+                except EasyBuildError:
+                    pass
```
boegel: at least log a debug message that parsing this easyconfig failed:

```python
path = os.path.join(dirpath, filename)
...
except EasyBuildError as err:
    _log.debug("Ignoring easyconfig %s that fails to parse: %s", path, err)

File: easybuild/tools/github.py [hunk 16]
```diff
@@ -684,6 +727,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+            sys.stdout.write('\r%s easyconfigs checked' % counter)
```
boegel: also mention how many easyconfigs will be checked in total (so construct a list of paths to all easyconfig files first via `os.walk`, and then simply loop over that list

easybuild/tools/github.py
```diff
@@ -684,6 +727,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+            sys.stdout.flush()
+            counter = counter+1
```
boegel: you could use `for idx, ec_path in enumerate(ec_paths)` here, and then use `idx + 1` as counter

easybuild/tools/github.py
```diff
@@ -684,6 +727,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
     return file_info, git_repo, pr_branch, diff_stat
 
 
+def scan_all_easyconfigs(patch_name):
```
boegel: please add a test for this new function

easybuild/tools/github.py
Caylo: TEMPLATE has the "patches"  parameter, so it will match, but can't be parsed as an EasyConfig in the line below
i could change the check to `patches = [.+]` to make sure it's not an empty list?

easybuild/tools/github.py
```diff
@@ -684,6 +727,38 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+    found = False
+    counter = 0
+    for (dirpath, _, filenames) in os.walk(robot_path):
+        if found:
+            break
+        for filename in [fn for fn in filenames if ".eb" in fn]:
+            rawtxt = read_file(os.path.join(dirpath, filename))
+            if "patches" in rawtxt and "TEMPLATE" not in filename:
+                try:
+                    ec = EasyConfig(os.path.join(dirpath, filename), validate=False)
```
boegel: don't use `EasyConfig` directly, use `process_easyconfig` instead (which implements caching, so the 2nd time you'd bypass actually reparsing the easyconfig file)

easybuild/framework/easyconfig/easyconfig.py
boegel: we shouldn't auto-guess the PR title if only patch files are being added, so yes, I would drop this

easybuild/tools/github.py
boegel: OK, but then skip it with `filename != 'TEMPLATE.eb'`

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1365,55 +1365,96 @@ def robot_find_minimal_toolchain_of_dependency(dep, modtool, parent_tc=None):
...
+
+        target_path = os.path.join(target_dir, target_path)
+
+    else:
+        raise EasyBuildError("Subdirectory %s not found in %s", subdir, target_dir)
+
+    return target_path
+
+
+def copy_file(path, target_dir, soft_name, target_name):
```
boegel: @Caylo I'd move this to `easybuild.tools.filetools`, it's a useful function in general

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1365,55 +1365,96 @@ def robot_find_minimal_toolchain_of_dependency(dep, modtool, parent_tc=None):
...
+
+    target_path = det_location_for(path, target_dir, soft_name, target_name)
+    new = os.path.exists(target_path)
```
boegel: move this out of this function, this should b a generic function to copy files (also: add a docstring)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1365,55 +1365,96 @@ def robot_find_minimal_toolchain_of_dependency(dep, modtool, parent_tc=None):
...
+
+    try:
+        mkdir(os.path.dirname(target_path), parents=True)
+        shutil.copy2(path, target_path)
+        _log.info("%s copied to %s", path, target_path)
+    except OSError as err:
+        raise EasyBuildError("Failed to copy %s to %s: %s", path, target_path, err)
+
+    return target_path, new
```
boegel: I don't think this should return anything, you should pass is both the source and target path (and the `new` thing is something very specific, not relevant in general)

easybuild/tools/github.py
```diff
@@ -627,6 +627,22 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
 
     _log.debug("git status: %s", git_repo.git.status())
 
+    # seperate easyconfigs and patch files
+    ec_paths = [p for p in existing_paths if p.endswith('.eb') or p.endswith('.yeb') or not is_patch_file(p)]
```
boegel: the two `endswith` checks are useless if you also check `not is_patch_file`, since non-easyconfig files will fall through on the last check anyway

so, just use `not is_patch_file(p))` (and assume only easyconfigs are passed in)

easybuild/tools/github.py
```diff
@@ -716,6 +736,74 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+    """
+    Scan all easyconfigs in the robot path to determine which software a patch file belongs to
+
+    @param patch_name: name of the patch file
+    """
+    robot_paths = build_option('robot_path')
+    soft_name = ""
+    found = False
+    for robot_path in robot_paths:
+        all_ecs = []
```
boegel: move this out of this loop, and construct a *full* set of easyconfigs first before iterating over them?

```python
all_ecs = []
for robot_path in robot_paths:
    ...

for idx, filename in enumerate(all_ecs):
    ...

File: easybuild/tools/github.py [hunk 25]
```diff
@@ -716,6 +736,74 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+        for (dirpath, _, filenames) in os.walk(robot_path):
+            all_ecs.extend([os.path.join(dirpath, fn) for fn in filenames])
+
+        for idx, filename in enumerate(all_ecs):
+            if found:
+                break
+            if os.path.splitext(filename)[-1] in ['.eb', '.yeb']:
```
boegel: use the `EB_FORMAT_EXTENSION` and `YEB_FORMAT_EXTENSION` constants

easybuild/tools/github.py
```diff
@@ -716,6 +736,74 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                path = os.path.join(dirpath, filename)
+                rawtxt = read_file(path)
+                if "patches" in rawtxt and filename != "TEMPLATE.eb":
```
boegel: filter out `TEMPLATE.eb` when you build the `all_ecs` list

easybuild/tools/github.py
boegel: and maybe also filter on the presence of `patches` already? that'll make this function a bit easier to follow

it does mean reading all files which contain 'patches' twice, but thanks to filesystem caching you'll probably not notice

easybuild/tools/github.py
```diff
@@ -716,6 +736,74 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                    try:
+                        ecs = process_easyconfig(path, validate=False)
+                        for ec in ecs:
+                            ec['ec']._generate_template_values()
+                            if patch_name in ec['ec'].asdict()['patches']:
+                                soft_name = ec['ec'].asdict()['name']
+                                found = True
+                    except EasyBuildError as err:
+                        _log.debug("Ignoring easyconfig %s that fails to parse: %s", path, err)
+                        pass
```
boegel: no need for the `pass` since you have the `log.debug`

easybuild/tools/github.py
```diff
@@ -716,6 +736,74 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+            if found:
+                break
+            if os.path.splitext(filename)[-1] in ['.eb', '.yeb']:
+                path = os.path.join(dirpath, filename)
+                rawtxt = read_file(path)
+                if "patches" in rawtxt and filename != "TEMPLATE.eb":
+                    try:
+                        ecs = process_easyconfig(path, validate=False)
+                        for ec in ecs:
+                            ec['ec']._generate_template_values()
```
boegel: why is this needed?

easybuild/tools/github.py
```diff
@@ -716,6 +736,74 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                            if patch_name in ec['ec'].asdict()['patches']:
```
boegel: just do `ec['ec']['patches']`?

easybuild/tools/github.py
```diff
@@ -716,6 +736,74 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                                soft_name = ec['ec'].asdict()['name']
+                                found = True
```
boegel: add a `break` here, we're done searching

easybuild/tools/github.py
```diff
@@ -716,6 +736,74 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                    except EasyBuildError as err:
+                        _log.debug("Ignoring easyconfig %s that fails to parse: %s", path, err)
+                        pass
+                sys.stdout.write('\r%s of %s easyconfigs checked' % (idx+1, len(all_ecs)))
```
boegel: just do `len(all_ecs)` once outside of the `for` loop, the length won't change

easybuild/tools/filetools.py
```diff
@@ -1324,3 +1330,13 @@ def find_flexlm_license(custom_env_vars=None, lic_specs=None):
     _log.info("Found valid license specs via %s: %s", via_msg, valid_lic_specs)
 
     return (valid_lic_specs, lic_env_var)
+
+
+def copy_file(path, target_path):
+
```
boegel: docstring please!

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1365,55 +1365,85 @@ def robot_find_minimal_toolchain_of_dependency(dep, modtool, parent_tc=None):
...
-                if letter not in a_to_z:
-                    raise EasyBuildError("Don't know which letter subdir to use for %s", name)
+            target_path = det_location_for(path, target_dir, soft_name, ec_filename)
+            new = os.path.exists(target_path)
 
-                target_path = os.path.join(subdir, letter, name, ec_filename)
-                _log.debug("Target path for %s: %s", path, target_path)
+            copy_file(path, target_path)
+            file_info['paths_in_repo'].append(target_path)
+            file_info['new'].append(new)
```
boegel: no need for a `new` variable, it's only used in one place so let's do `file_info['new'].append(os.path.exists(target_path))`

easybuild/tools/github.py
```diff
@@ -627,6 +629,22 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
     _log.debug("git status: %s", git_repo.git.status())
 
+    # seperate easyconfigs and patch files
+    ec_paths = [p for p in existing_paths if not is_patch_file(p)]
+    patch_paths = [p for p in existing_paths if p not in ec_paths]
+
+    # copy easyconfig files to right place
+    target_dir = os.path.join(git_working_dir, pr_target_repo)
+    print_msg("copying easyconfigs to %s..." % target_dir)
+    file_info = copy_easyconfigs(ec_paths, target_dir)
```
boegel: add guard `if ec_paths:` ?

easybuild/tools/github.py
```diff
@@ -716,6 +738,77 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                raise EasyBuildError("Failed to determine software name to which patch file %s relates", patch_path)
+
+    return patch_specs
+
+
+def scan_all_easyconfigs(patch_name):
+    """
+    Scan all easyconfigs in the robot path to determine which software a patch file belongs to
+
+    @param patch_name: name of the patch file
```
boegel: `:param`, and also cover `:return:`

easybuild/tools/github.py
```diff
@@ -716,6 +738,77 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+    """
+    robot_paths = build_option('robot_path')
+    soft_name = ""
```
boegel: use `None` as default

easybuild/tools/github.py
```diff
@@ -716,6 +738,77 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+            for fn in filenames:
+                if fn != 'TEMPLATE.eb':
+                    path = os.path.join(dirpath, fn)
+                    rawtxt = read_file(path)
+                    if 'patches' in rawtxt:
+                        all_ecs.append(path)
+
+    nr_of_ecs = len(all_ecs)
+    for idx, filename in enumerate(all_ecs):
+        if found:
```
boegel: drop `found`, check with `if soft_name`

easybuild/tools/github.py
```diff
@@ -716,6 +738,77 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+        for (dirpath, _, filenames) in os.walk(robot_path):
+            for fn in filenames:
+                if fn != 'TEMPLATE.eb':
+                    path = os.path.join(dirpath, fn)
+                    rawtxt = read_file(path)
+                    if 'patches' in rawtxt:
+                        all_ecs.append(path)
+
+    nr_of_ecs = len(all_ecs)
+    for idx, filename in enumerate(all_ecs):
```
boegel: `s/filename/path`, since it's actually a full path

it also helps in avoiding silly mistakes to use the same variable name as above ;)

easybuild/tools/github.py
```diff
@@ -716,6 +738,77 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+            if soft_name:
+                patch_specs.append((patch_path, soft_name))
+            else:
+                # still nothing found
+                raise EasyBuildError("Failed to determine software name to which patch file %s relates", patch_path)
+
+    return patch_specs
+
+
+def scan_all_easyconfigs(patch_name):
```
boegel: rename to `find_software_name_for_patch`?

easybuild/tools/github.py
```diff
@@ -716,6 +738,77 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+
+    nr_of_ecs = len(all_ecs)
+    for idx, filename in enumerate(all_ecs):
+        if found:
+            break
+        rawtxt = read_file(filename)
+        try:
+            ecs = process_easyconfig(filename, validate=False)
+            for ec in ecs:
+                if patch_name in ec['ec']['patches']:
```
boegel: I just realised this is too simplistic...

The entries in `patches` can also be tuples like `('foo.patch', 2)`, where the `2` is the patch level

So, I'd create an inner function in `find_software_name_for_patch` and use that:

```python
def is_patch_for(patch_name, ec):
    """Check whether specified patch matches any patch in the provided EasyConfig instance."""
    res = False
    for patch in ec['patches']:
        if isinstance(patch, (tuple, list)):
            patch = patch[0]
        if patch == patch_name:
            res = True
            break

    return res
````

and then check with `if is_patch_for(patch_name, ec['ec'])`

File: easybuild/tools/github.py [hunk 41]
```diff
@@ -716,6 +738,77 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+                    soft_name = ec['ec']['name']
+                    found = True
+                    break
+        except EasyBuildError as err:
+            _log.debug("Ignoring easyconfig %s that fails to parse: %s", path, err)
```
boegel: see why you should use `path` rather than `filename`? ;)

easybuild/tools/github.py
```diff
@@ -716,6 +738,77 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+        sys.stdout.write('\r%s of %s easyconfigs checked' % (idx+1, nr_of_ecs))
+        sys.stdout.flush()
+
+    print ''
```
boegel: use `sys.stdout.write('\n')` here for consistency

test/framework/filetools.py
```diff
@@ -672,6 +672,12 @@ def test_find_flexlm_license(self):
         del os.environ['LM_LICENSE_FILE']
         self.assertEqual(ft.find_flexlm_license(lic_specs=[None]), ([], None))
 
+    def test_is_patch_file(self):
+        """Test for is_patch_file() function."""
+        testdir = os.path.dirname(os.path.abspath(__file__))
+        self.assertFalse(ft.is_patch_file(os.path.join(testdir, 'easyconfigs', 'toy-0.0.eb')))
+        self.assertTrue(ft.is_patch_file(os.path.join(testdir, 'sandbox', 'sources', 'toy', 'toy-0.0_typo.patch')))
+
```
boegel: @Caylo: please add a test for `copy_file` here too

easybuild/tools/github.py
```diff
@@ -659,9 +659,6 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
             else:
                 raise EasyBuildError("Path doesn't exist or file to delete isn't found in target branch: %s", fn)
 
-    # copy edited/added files to right place
-    file_info = copy_easyconfigs(ec_paths, repo_path)
```
boegel: @Caylo: oops?

test/framework/github.py
```diff
@@ -240,6 +242,28 @@ def test_validate_github_token(self):
...
+            'allow_modules_tool_mismatch': True,
+            'minimal_toolchains': True,
+            'use_existing_modules': True,
+            'external_modules_metadata': ConfigObj(),
+            'robot_path': [ec_path],
+            'valid_module_classes': module_classes(),
+            'validate': False,
+        })
+        self.mock_stdout(True)
+        ec = gh.scan_all_easyconfigs('toy-0.0_typo.patch')
```
boegel: this won't work :)

easybuild/tools/github.py
boegel: nvm

boegel:
Going in, thanks @Caylo!

PR #1853, opened 2016-07-25
handle deleted files in --new-pr

Caylo:
usage: `eb --new-pr [path to deleted file] `
cfr #1674

includes #1867


boegel:
I would go with `eb --new-pr :gzip1-6.eb` (or `eb --new-pr -gzip1-6.eb`, and why not support both)

easybuild/tools/github.py
```diff
@@ -44,7 +44,7 @@
 from vsc.utils import fancylogger
 from vsc.utils.missing import nub
 
-from easybuild.framework.easyconfig.easyconfig import copy_easyconfigs
+from easybuild.framework.easyconfig.easyconfig import copy_easyconfigs, process_easyconfig
```
boegel: why import `process_easyconfig` if it's not being used?

boegel:
@Caylo I made some changes to this, see https://github.com/Caylo/easybuild-framework/pull/30

easybuild/tools/github.py
```diff
@@ -617,12 +627,30 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+        fullpath = os.path.join(repo_path, fn)
+        if os.path.exists(fullpath):
+            deleted_paths.append(fullpath)
+        else:
+            # if no existing relative path is specified, assume just the easyconfig file name is provided
+            hits = glob.glob(os.path.join(repo_path, 'easybuild', 'easyconfigs', '*', '*', fn))
+            if len(hits) == 1:
+                deleted_paths.append(hits[0])
+            else:
+                raise EasyBuildError("Path doesn't exist or deleted file isn't found in target branch: %s", fn)
```
boegel: `s/deleted file/file to delete/g`

easybuild/tools/github.py
```diff
@@ -721,14 +753,19 @@ def new_pr(paths, title=None, descr=None, commit_msg=None):
...
+                main_title = ', '.join(names_and_versions)
+            else:
+                main_title = ', '.join(names_and_versions[:3] + ['...'])
+
+            title = "{%s}[%s] %s" % (class_label, toolchain_label, main_title)
 
-        title = "{%s}[%s] %s" % (class_label, toolchain_label, main_title)
+        else:
+            raise EasyBuildError("Don't know how to make a PR title for this PR. "
+                                 "Please include a title (use --pr-title)")
```
boegel: I would also require to use `--pr-title` whenever existing easyconfigs are modified

the auto-guessed title really only makes sense if only new easyconfigs are added (which is the most common use of --new-pr, probably)

Caylo: ok, see this check here https://github.com/hpcugent/easybuild-framework/pull/1853/files#diff-ccf1e273d8604bc208900aa7330c80aaR756

easybuild/tools/github.py
```diff
@@ -721,14 +753,19 @@ def new_pr(paths, title=None, descr=None, commit_msg=None):
     class_label = ','.join([tc for (cnt, tc) in classes_counted if cnt == classes_counted[-1][0]])
 
     if title is None:
-        # mention software name/version in PR title (only first 3)
-        names_and_versions = ["%s v%s" % (ec.name, ec.version) for ec in file_info['ecs']]
-        if len(names_and_versions) <= 3:
-            main_title = ', '.join(names_and_versions)
-        else:
-            main_title = ', '.join(names_and_versions[:3] + ['...'])
+        if file_info['ecs'] and any(file_info['new']):
```
boegel: @Caylo this should be `all(file_info['new'])`, no? We only want to auto-guess the title if all easyconfigs are new?

test/framework/options.py
```diff
@@ -2365,6 +2366,36 @@ def test_new_update_pr(self):
...
+
+        if self.github_token is None:
+            print "Skipping test_new_pr_delete, no GitHub token available?"
+            return
+
+        args = [
+            '--new-pr',
+            '--experimental',
+            '--github-user=%s' % GITHUB_TEST_ACCOUNT,
+            ':bzip2-1.0.6.eb',
```
boegel: @Caylo is it worth adding another test that both adds an easyconfig file (`toy-0.0.eb`) *and* deletes one at the same time?

Caylo: I don't think this has a lot of extra use. The other two should be enough imo :)

easybuild/tools/github.py
```diff
@@ -721,14 +753,19 @@ def new_pr(paths, title=None, descr=None, commit_msg=None):
     class_label = ','.join([tc for (cnt, tc) in classes_counted if cnt == classes_counted[-1][0]])
 
     if title is None:
-        # mention software name/version in PR title (only first 3)
-        names_and_versions = ["%s v%s" % (ec.name, ec.version) for ec in file_info['ecs']]
-        if len(names_and_versions) <= 3:
-            main_title = ', '.join(names_and_versions)
-        else:
-            main_title = ', '.join(names_and_versions[:3] + ['...'])
+        if file_info['ecs'] and all(file_info['new']):
```
boegel: also add `and not deleted_paths`, i.e. enforce picking a custom title with easyconfigs are added/removed in a single go

boegel:
Going in, thanks @Caylo!

PR #1855, opened 2016-07-25
auto-add dependencies of easyconfig in --new-pr

Caylo:
cfr #1674 
note: searches for dependencies in robot-path, not sure if this is the right way to go


easybuild/framework/easyconfig/tools.py
```diff
@@ -546,3 +545,6 @@ def dump_env_script(easyconfigs):
         print_msg("Script to set up build environment for %s dumped to %s" % (ecfile, script_path), prefix=False)
 
         restore_env(orig_env)
+
+
+from easybuild.tools.github import fetch_easyconfigs_from_pr, download_repo
```
boegel: @Caylo we need to find a way around this (I'm even surprised it help in working around a cyclic import at all)

easybuild/tools/github.py
```diff
@@ -617,9 +619,14 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
 
     _log.debug("git status: %s", git_repo.git.status())
 
-    # copy files to right place
+    deps = resolve_dependencies([process_easyconfig(ec)[0] for ec in paths], modules_tool())
+
     file_info = copy_easyconfigs(paths, os.path.join(git_working_dir, pr_target_repo))
 
+    # dont save info on dependencies
```
boegel: `don't`

easybuild/tools/github.py
```diff
@@ -617,9 +619,14 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
...
+    dep_paths = [dep['spec'] for dep in deps if dep['spec'] not in paths]
```
boegel: are the `paths` specified as absolute paths here? I'd expect not, so the `not in` check isn't going to do what is intended...

test/framework/options.py
```diff
@@ -2352,6 +2352,48 @@ def test_new_update_pr(self):
             regex = re.compile(regex, re.M)
             self.assertTrue(regex.search(txt), "Pattern '%s' found in: %s" % (regex.pattern, txt))
 
+    def test_new_pr_dependencies(self):
+        """Test use of --new-pr to delete easyconfigs."""
```
boegel: fix docstring please :)

test/framework/options.py
```diff
@@ -2352,6 +2352,48 @@ def test_new_update_pr(self):
...
+
+        if self.github_token is None:
+            print "Skipping test_new_pr_dependencies, no GitHub token available?"
+            return
+
+        # copy toy test easyconfig
```
boegel: no toy involved?

test/framework/options.py
```diff
@@ -2352,6 +2352,48 @@ def test_new_update_pr(self):
...
+        self.mock_stdout(True)
+        self.eb_main(args, do_build=True, raise_error=True, testing=False)
+        txt = self.get_stdout()
+        self.mock_stdout(False)
+
+        regexs = [
+            r"^\* overview of changes:",
+            r".*/hwloc-1.6.2-GCC-4.7.2.eb\s+\|\s+[0-9]+\s+\++",
+            r".*/OpenMPI-1.6.4-GCC-4.7.2.eb\s+\|\s+[0-9]+\s+\++",
+            r"^\s*2 files changed",
```
boegel: hold on... this OpenMPI easyconfig is already in `develop`, why would it be added then?

Caylo: @boegel  the test version slightly differs from the develop one (see osdependencies)

easybuild/tools/github.py
```diff
@@ -617,9 +619,14 @@ def _easyconfigs_pr_common(paths, start_branch=None, pr_branch=None, target_acco
 
     _log.debug("git status: %s", git_repo.git.status())
 
-    # copy files to right place
+    deps = resolve_dependencies([process_easyconfig(ec)[0] for ec in paths], modules_tool())
```
boegel: @Caylo make sure a proper error is printed here in case dependencies could not be resolved with the current robot search path

PR #1856, opened 2016-07-26
[WIP] --new-pr for easyblocks repo

Caylo:
use: with `--pr-target-repo=easybuild-easyblocks`
cfr #1674 

(todo: filenames with special chars)


easybuild/tools/github.py
```diff
@@ -77,10 +79,13 @@
     _log.warning("Failed to import 'git' Python module: %s", err)
 
 
+EB_PREFIX = 'EB_'
```
boegel: use the existing `EASYBLOCK_CLASS_PREFIX` constant from `filetools`

boegel:
follow-up in #1876

PR #1860, opened 2016-08-02
fix easyblock documentation

Caylo:
output : http://caylo-easybuild.readthedocs.io/en/fix-easyblocks/version-specific/easyblocks.html


boegel:
Thanks @Caylo!

PR #1861, opened 2016-08-03
update easybuild from command line

Caylo:
use: `eb --update`

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -55,7 +55,7 @@ from easybuild.framework.easyconfig.tools import get_paths_for, parse_easyconfig
 from easybuild.framework.easyconfig.tweak import obtain_ec_for, tweak
 from easybuild.tools.config import find_last_log, get_repository, get_repositorypath, build_option
 from easybuild.tools.filetools import adjust_permissions, cleanup, write_file
-from easybuild.tools.github import check_github, install_github_token, new_pr, update_pr
+from easybuild.tools.github import check_github, find_easybuild_eb, install_github_token, new_pr, update_pr
 from easybuild.tools.modules import modules_tool
 from easybuild.tools.options import parse_external_modules_metadata, process_software_build_specs
 from easybuild.tools.robot import check_conflicts, det_robot_path, dry_run, resolve_dependencies, search_easyconfigs
@@ -271,6 +271,10 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     if search_query:
         search_easyconfigs(search_query, short=options.search_short, filename_only=options.search_filename,
                            terse=options.terse)
+
+    if options.update:
+        eb_file = find_easybuild_eb()
+        orig_paths.append(eb_file)
 
     # GitHub integration
     cleanup_and_exit = handle_github_options(options, orig_paths)
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -147,6 +147,7 @@ BUILD_OPTIONS_CMDLINE = {
         'skip_test_cases',
         'sticky_bit',
         'upload_test_report',
+        'update',
         'update_modules_tool_cache',
         'use_existing_modules',
     ],
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1,4 +1,4 @@
- ##
+##
 # Copyright 2012-2016 Ghent University
 #
 # This file is part of EasyBuild,
@@ -41,6 +41,7 @@ import sys
 import tempfile
 import time
 import urllib2
+from distutils.version import LooseVersion
 from vsc.utils import fancylogger
 from vsc.utils.missing import nub
 
@@ -1067,7 +1068,15 @@ def find_easybuild_eb():
     """
     dev_repo = download_repo(GITHUB_EASYCONFIGS_REPO, branch='develop', account=GITHUB_EB_MAIN)
     eb_parent_path = os.path.join(dev_repo, 'easybuild', 'easyconfigs', 'e', 'EasyBuild')
-    fn = sorted(os.listdir(eb_parent_path))[-1]
+    files = os.listdir(eb_parent_path)
+
+    # find most recent version
+    file_versions = []
+    for eb_file in files:
+        version = eb_file.split('-', 1)[1]
+        file_versions.append((eb_file, version.split('.eb', 1)[0]))
+
+    fn = sorted(file_versions, key=lambda x: LooseVersion(x[1]))[-1][0]
 
     eb_file = os.path.join(eb_parent_path, fn)
     return eb_file
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1,4 +1,4 @@
-##
+ ##
 # Copyright 2012-2016 Ghent University
 #
 # This file is part of EasyBuild,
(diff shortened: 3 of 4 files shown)


easybuild/main.py
```diff
@@ -272,6 +272,10 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         search_easyconfigs(search_query, short=options.search_short, filename_only=options.search_filename,
                            terse=options.terse)
 
+    if options.update:
+        eb_file = find_easybuild_eb()
+        orig_paths.append(eb_file)
```
boegel: I would try and make sure that *only* EasyBuild is installed, and throw an error if other easyconfigs are installed at the same time?

easybuild/tools/config.py
```diff
@@ -147,6 +147,7 @@ def mk_full_default_path(name, prefix=DEFAULT_PREFIX):
         'skip_test_cases',
         'sticky_bit',
         'upload_test_report',
+        'update',
```
boegel: We need a better name than `--update`, since it's too confusing w.r.t. *what* is being updated; it's not clear that this is about updating EasyBuild itself (or actually, installing a module for the latest EasyBuild version, it's not even in-place updating).

easybuild/tools/github.py
```diff
@@ -1059,3 +1060,23 @@ def validate_github_token(token, github_user):
...
+    """
+    dev_repo = download_repo(GITHUB_EASYCONFIGS_REPO, branch='develop', account=GITHUB_EB_MAIN)
+    eb_parent_path = os.path.join(dev_repo, 'easybuild', 'easyconfigs', 'e', 'EasyBuild')
+    files = os.listdir(eb_parent_path)
+
+    # find most recent version
+    file_versions = []
+    for eb_file in files:
+        version = eb_file.split('-', 1)[1]
+        file_versions.append((eb_file, version.split('.eb', 1)[0]))
```
boegel: wouldn't it be better to grab the `version =` line from the easyconfig file itself, rather than relying on the filename? would be cleaner imho...

easybuild/tools/github.py
```diff
@@ -1059,3 +1060,23 @@ def validate_github_token(token, github_user):
...
+
+    fn = sorted(file_versions, key=lambda x: LooseVersion(x[1]))[-1][0]
```
boegel: I think you can simplify this a bit by swapping the order in the tuple and using LooseVersion there already, i.e. `(LooseVersion(<version>), <file>)`, then you don't need to fiddle with `key` at all?

easybuild/tools/config.py
boegel: How about `eb --install-latest` or `eb --update-self` or something like that?

easybuild/tools/github.py
```diff
@@ -1088,3 +1087,26 @@ def validate_github_token(token, github_user):
         _log.info("GitHub token can be used for authenticated GitHub access, validation passed")
 
     return sanity_check and token_test
+
+
+def find_easybuild_eb():
```
boegel: @Caylo please add a unit test for this (just check whether it returns a path an existing easyconfig for EasyBuild)

easybuild/main.py
```diff
@@ -272,6 +272,13 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         search_easyconfigs(search_query, short=options.search_short, filename_only=options.search_filename,
                            terse=options.terse)
 
+    if options.latest_self:
+        if not orig_paths:
+            eb_file = find_easybuild_eb()
+            orig_paths.append(eb_file)
+        else:
+            raise EasyBuildError("Update option isn't compatible with other easyconfig files")
```
boegel: make this `"Installing the latest EasyBuild release can not be combined with installing other easyconfigs"`

boegel:
@Caylo needs a sync with `develop` to resolve conflicts

easybuild/tools/github.py
```diff
@@ -1088,3 +1087,26 @@ def validate_github_token(token, github_user):
...
+    dev_repo = download_repo(GITHUB_EASYCONFIGS_REPO, branch='develop', account=GITHUB_EB_MAIN)
+    eb_parent_path = os.path.join(dev_repo, 'easybuild', 'easyconfigs', 'e', 'EasyBuild')
+    files = os.listdir(eb_parent_path)
+
+    # find most recent version
+    file_versions = []
+    for eb_file in files:
+        txt = read_file(os.path.join(eb_parent_path, eb_file))
+        for line in txt.split('\n'):
+            if re.search(r'^version =', line):
```
boegel: use `\s*` rather than hardcoding the space before the `=`

easybuild/tools/github.py
```diff
@@ -1088,3 +1087,26 @@ def validate_github_token(token, github_user):
...
+                version = line.split('= ')[-1].replace('"', '').replace("'", '')
```
boegel: use `exec` instead:

```python
scope = {}
exec(line, scope)
version = scope['version']
```

easybuild/tools/options.py
```diff
@@ -288,6 +288,7 @@ def override_options(self):
                           "(e.g. --hide-deps=zlib,ncurses)", 'strlist', 'extend', None),
             'hide-toolchains': ("Comma separated list of toolchains that you want automatically hidden, "
                                 "(e.g. --hide-toolchains=GCCcore)", 'strlist', 'extend', None),
+            'latest-self': ("Install latest known version of easybuild", None, 'store_true', False),
```
boegel: hmm, somehow my last comment on this got trashed...

let's go with `--install-latest-eb-release` here

test/framework/github.py
```diff
@@ -240,6 +240,13 @@ def test_validate_github_token(self):
 
         self.assertTrue(gh.validate_github_token(self.github_token, GITHUB_TEST_ACCOUNT))
 
+    def test_find_easybuild_eb(self):
+        """Test for find_easybuild_eb function"""
+        path = gh.find_easybuild_eb()
+        expected = os.path.join('e', 'EasyBuild', 'EasyBuild-[1-9]+\.[1-9]+\.[1-9]+\.eb')
+        regex = re.compile(expected)
+        self.assertTrue(re.search(regex, path))
```
boegel: also make sure the returned path exists: `self.assertTrue(os.path.exists(path))`

easybuild/main.py
```diff
@@ -272,6 +272,14 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         search_easyconfigs(search_query, short=options.search_short, filename_only=options.search_filename,
                            terse=options.terse)
 
+    if options.latest_self:
+        if not orig_paths:
+            eb_file = find_easybuild_eb()
+            orig_paths.append(eb_file)
+        else:
+            raise EasyBuildError("Installing the latest EasyBuild release can not be combined with installing "
```
boegel: use positive logic (easier to understand):

```python
if orig_paths:
    raise EasyBuildError("...")
else:
    orig_paths = [find_easybuild_easyconfig()]
```

easybuild/tools/github.py
boegel: also, rename to `find_easybuild_easyconfig`?

easybuild/tools/github.py
```diff
@@ -1125,3 +1124,28 @@ def validate_github_token(token, github_user):
...
+    for eb_file in files:
+        txt = read_file(os.path.join(eb_parent_path, eb_file))
+        for line in txt.split('\n'):
+            if re.search(r'^version\s*=', line):
+                scope = {}
+                exec(line, scope)
+                version = scope['version']
+                file_versions.append((LooseVersion(version), eb_file))
+
+    fn = sorted(file_versions)[-1][1]
```
boegel: @Caylo let's throw a proper error here if `file_versions` is empty...

boegel:
going in, thanks @Caylo!

PR #1862, opened 2016-08-04
Make sure the robot ignores filtered dependencies 

ocaisa:
when creating the toolchain cache


ocaisa:
This is failing, @boegel can you remind me to how to reference the easyconfig object in the dictionary so I can access the method?

boegel:
@ocaisa I think you want `parsed_ec['ec'].dependencies()`, see https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/framework/easyconfig/easyconfig.py#L1216

boegel:
lgtm, thanks @ocaisa!

PR #1867, opened 2016-08-09
fix broken --from-pr tests

Caylo:


easybuild/tools/github.py
```diff
@@ -383,17 +387,25 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
                 _log.info("Downloading %s from %s" % (fn, full_url))
                 download_file(fn, full_url, path=os.path.join(path, patched_file), forced=True)
         else:
-            apply_patch(diff_filepath, path, level=1)
+            apply_patch(diff_filepath, dl_path, level=1)
+
+    if dl_path != path:
+        for eb_file in glob.glob(os.path.join(dl_path, 'easybuild', 'easyconfigs', '*')):
```
boegel: these are not files, but directories, and better use `os.listdir` since we'll symlink all of them anyway

easybuild/tools/github.py
```diff
@@ -348,11 +349,14 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
 
     if (stable or pr_data['merged']) and not closed:
         # whether merged or not, download develop
-        path = download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='develop', path=path)
+        dl_path = download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='develop', path=path)
```
boegel: I would drop the `path=path`, which will make `download_repo` download `develop` into a separate tmpdir

otherwise we'll have both the `easybuild-easyconfigs-develop/easybuild/easyconfigs` and `a`..`z` dirs intermixed in a single parent directory...

easybuild/tools/github.py
```diff
@@ -383,17 +387,25 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
-            apply_patch(diff_filepath, path, level=1)
+            apply_patch(diff_filepath, dl_path, level=1)
+
+    if dl_path != path:
+        for eb_file in glob.glob(os.path.join(dl_path, 'easybuild', 'easyconfigs', '*')):
+            os.symlink(eb_file, os.path.join(path, os.path.basename(eb_file)))
+        check_path = dl_path
+
+    else:
+        check_path = path
```
boegel: `dl_path` and `check_path` are now exactly the same thing, and I don't like either name :p

how about changing the definition of `dl_path` to `final_path`, and just use that (no need to introduce another variable for exactly the same thing)

easybuild/tools/github.py
```diff
@@ -31,6 +31,7 @@
 """
 import base64
 import getpass
+import glob
```
boegel: @Caylo unused, please remove

Issue #1869, opened 2016-08-10
bootstrap fails on Travis with Python 2.7.12 due to setuptools version requirement


boegel:
For reasons unclear to me currently, the testing of the bootstrap script is failing on Travis when the tests are being run on top of Python 2.7.12; it appears Travis is rolling out an update of Python that is causing this problem...

```
Traceback (most recent call last):
  File "setup.py", line 59, in <module>
    shared_setup.action_target(PACKAGE)
  File "external_dist_only/shared_setup_dist_only.py", line 1354, in action_target
    vsc_setup().action_target(package, *args, **kwargs)
  File "external_dist_only/shared_setup_dist_only.py", line 1342, in action_target
    setupfn(**x)
  File "/opt/python/2.7.12/lib/python2.7/distutils/core.py", line 111, in setup
    _setup_distribution = dist = klass(attrs)
  File "/tmp/tmploR4yQ/lib/python2.7/site-packages/distribute-0.6.49-py2.7.egg/setuptools/dist.py", line 221, in __init__
    self.fetch_build_eggs(attrs.pop('setup_requires'))
  File "/tmp/tmploR4yQ/lib/python2.7/site-packages/distribute-0.6.49-py2.7.egg/setuptools/dist.py", line 245, in fetch_build_eggs
    parse_requirements(requires), installer=self.fetch_build_egg
  File "/tmp/tmploR4yQ/lib/python2.7/site-packages/distribute-0.6.49-py2.7.egg/pkg_resources.py", line 600, in resolve
    raise VersionConflict(dist,req) # XXX put more info here
pkg_resources.VersionConflict: (setuptools 0.6c11 (/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages), Requirement.parse('setuptools>=0.7'))
```

boegel:
example of failing Travis test: https://travis-ci.org/hpcugent/easybuild-framework/jobs/151187422

boegel:
temporary fix applied in #1870

riccardomurri:
I too think this is the same issue as #1887 , i.e., `bootstrap_eb.py` depending on the outdated `distribute` package.

boegel:
fixed in #1946, follow-up in #1887

PR #1871, opened 2016-08-10
add support for hiding toolchains

boegel:
wrapper PR for #1683 by @geimer

no changes were made other than syncing the PR branch with current `develop` such that #1869 is included as well, which is required to make Travis happy and make the PR mergeable


boegel:
going in, review was done via #1683 

Thanks @geimer!

Issue #1814, opened 2016-06-22
github download url could possibly be templated


fgeorgatos:
fi. `https://github.com/git/git/archive` should work out of the box

`source_urls = ['https://github.com/alexdobin/STAR/archive/'` may need `github_org=alexdobin`
similarly, https://github.com/EddyRivasLab/hmmer/tree/3.0 may need `github_org=EddyRivasLab`
(thinking of it, non-default capitalization is an issue here, maybe insert a `github-repo` combo string)

fyi. this is spin-off from https://github.com/hpcugent/easybuild-easyconfigs/issues/3242

boegel:
```
# https://github.com/%(github_account)s/%(name)s/archive
source_urls = [GITHUB_SOURCE]
# or via GitHub API?
github_account = 'foo'
```

fgeorgatos:
yes, that would work! and looks good. 

idea: also templating `github_commitid` might help where upstream does a poor job in tagging a version;
as a bonus, you get decent certainty of downloading what you really meant (remember Rgputools?)

boegel:
fixed in #1872

PR #1872, opened 2016-08-10
template for github source url

Caylo:
cfr. #1814


easybuild/framework/easyconfig/default.py
```diff
@@ -88,6 +88,7 @@
     'easyblock': [None, "EasyBlock to use for building; if set to None, an easyblock is selected "
                         "based on the software name", BUILD],
     'easybuild_version': [None, "EasyBuild-version this spec-file was written for", BUILD],
+    'github_account': [None, "Github account name in case of github source_urls", BUILD],
```
boegel: make it clear this is used to resolve a template:

`"GitHub account name to be used to resolve template values in source URLs"`

easybuild/framework/easyconfig/templates.py
```diff
@@ -88,6 +89,8 @@
      'CRAN (contrib) source url'),
     ('FTPGNOME_SOURCE', 'http://ftp.gnome.org/pub/GNOME/sources/%(namelower)s/%(version_major_minor)s',
      'http download for gnome ftp server'),
+    ('GITHUB_SOURCE', 'https://github.com/%(github_account)s/%(namelower)s/archive',
+     'GitHub source url'),
```
boegel: `"GitHub source URL (requires github_account easyconfig parameter to be specified)"`

easybuild/framework/easyconfig/templates.py
```diff
@@ -88,6 +89,8 @@
      'CRAN (contrib) source url'),
     ('FTPGNOME_SOURCE', 'http://ftp.gnome.org/pub/GNOME/sources/%(namelower)s/%(version_major_minor)s',
      'http download for gnome ftp server'),
+    ('GITHUB_SOURCE', 'https://github.com/%(github_account)s/%(namelower)s/archive',
```
boegel: hmm, shouldn't this be `%(name)s` rather than `%(namelower)s`?!

boegel:
Going in, thanks @Caylo!

PR #1880, opened 2016-08-18
Making the bootstrap script work offline

ocaisa:
Allow people to place the patched `distribute` tarball in `$EASYBUILD_BOOTSTRAP_SOURCEPATH`


easybuild/scripts/bootstrap_eb.py
```diff
@@ -252,7 +252,12 @@ def stage0(tmpdir):
     # We download a custom version of distribute: it uses a newer version of markerlib to avoid a bug (#1099)
     # It's is the source of distribute 0.6.49 with the file _markerlib/markers.py replaced by the 0.6 version of
     # markerlib which can be found at https://pypi.python.org/pypi/markerlib/0.6.0
-    sys.argv.append('--download-base=http://hpcugent.github.io/easybuild/files/')
+    if EASYBUILD_BOOTSTRAP_SOURCEPATH is not None:
+       # Assume we're offline and the file should be available
+       debug("Looking for distribute tarball in EASYBUILD_BOOTSTRAP_SOURCEPATH")
+       sys.argv.append('--download-base=file://' + os.path.abspath(EASYBUILD_BOOTSTRAP_SOURCEPATH) + '/')
```
boegel: @ocaisa currently, not all tarballs *must* be provided in `$EASYBUILD_BOOTSTRAP_SOURCEPATH`, but here you're requiring that they are

How about making this conditional on having the expected tarball there?

boegel:
@ocaisa this is only a partial fix, right? Since you need to make sure `vsc-install` is picked up too?

easybuild/scripts/bootstrap_eb.py
ocaisa: I thought about that but then I said if you were aiming for an offline install then there's no difference, if the file is not there it will fail to get it anyway

boegel: the`$EASYBUILD_BOOTSTRAP_SOURCEPATH` option is not intended exclusively for offline installation, but also for testing the bootstrap mechanism under certain conditions (e.g. with a develop tarball for framework, or with an updated easyblock for EasyBuild itself, etc.)...

ocaisa:
Actually it's a fix, in that it's already stated that `vsc-install` is a requirement of EB in the documentation. If your python doesn't provide it you're not meeting the requirements (and you need to use my super duper triple bootstrap script instead)

boegel:
@ocaisa How does this PR add support to the bootstrap script for picking up a vsc-install tarball? It doesn't yet, right?

ocaisa:
No, that's an install requirement of EasyBuild as I understand things, otherwise you would have implemented it :)

boegel:
Well, no, the `vsc-install` requirement is relatively recent, and since the bootstrap procedure is usually performed online, it hasn't been a bug problem yet.

Enhancing the bootstrap script to also support picking up a vsc-install tarball makes sense.

ocaisa:
So is setuptools, if your underlying python doesn't have setuptools then the final installation of eb also won't work, it will complain about not being able to load `pkg_resources`

boegel:
@ocaisa OK, then that should be taken care of too

ocaisa:
distribute is not included in the installation directory of the bootstrapped EB, but even if it was that's a problem when you want to update EB

ocaisa:
You'd be forced into re-bootstrapping for every EB install

boegel:
@ocaisa Is it worth following up on this, or shall we just close it?

ocaisa:
(tldr) Up to you, I don't mind.

There is no meat to this PR really, just a discussion. Allowing for an offline installation can be required (that was my experience on an OpenPower system which had no internet access), this change was what made it possible for me. 

It would have to be updated based on the current status of the bootstrap script and it's requirements, I don't have the need for that any more so I'm reluctant to commit to trying to make it happen.

boegel:
@ocaisa Do you happen to remember how this is different from the offline support that is already there in the bootstrap script? Cfr. https://easybuild.readthedocs.io/en/latest/Installation.html#offline-bootstrapping-using-supplied-source-tarballs.

There was a fix fairly recently to ensure it fully works offline: https://github.com/easybuilders/easybuild-framework/pull/2517

ocaisa:
So I think basically comes back to whether we need Stage0 or not to be offline, currently it (still) tries to download the patched distribute (see https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/scripts/bootstrap_eb.py#L436) and will fail if we are offline. That makes setuptools a hard requirement for an offline install...and maybe that is ok?

If you do want fully offline capabilities this fix is potentially valid but it means that you can *only* upgrade EB by rebootstrapping.

ocaisa:
Indeed, having setuptools as a hard requirement for an offline install is probably required looking back at my comment above:
> if your underlying python doesn't have setuptools then the final installation of eb also won't work, it will complain about not being able to load `pkg_resources`

ocaisa:
@boegel You will understand this better than me. Currently, in Stage3 setuptools gets pulled into the installation when the bootscript triggers Stage0. Now if I supply an EasyBuild easyconfig to the bootstrapped eb to do an upgrade (`eb EasyBuild-v4.0.0.eb`), does setuptools also get installed with the new version? If not, the eb command will get broken. If that was the case, the only way to do an upgrade would be to do another bootstrap

ocaisa:
With the work that's going to remove the setuptools dependency this is no longer relevant.

PR #1882, opened 2016-08-19
Add a bootstrap script that can bootstrap Python, EasyBuild and Lmod

ocaisa:


ocaisa:
Relies on #1880 for offline reproducability

ocaisa:
@fgeorgatos Your dream has finally come true

easybuild/scripts/Python_EasyBuild_Lmod_bootstrap.sh
```diff
@@ -0,0 +1,342 @@
...
+  INSTALL_PYTHON_VERSION='2.7.12'
+fi
+# There are our python module dependencies
+setuptools_version='25.2.0'
+setuptools_version_url='https://pypi.python.org/packages/9f/32/81c324675725d78e7f6da777483a3453611a427db0145dfb878940469692/setuptools-25.2.0.tar.gz#md5=a0dbb65889c46214c691f6c516cf959c'
+vscinstall_version='0.10.11'
+vscinstall_version_url='https://pypi.python.org/packages/03/d0/291da76d7da921cf8e70dd7db79b0838e0f633655e8f2dd06093d99ce851/vsc-install-0.10.11.tar.gz'
+# Where do we get the EasyBuild bootstrap script
+#eb_bootstrap_url='https://raw.githubusercontent.com/hpcugent/easybuild-framework/develop/easybuild/scripts/bootstrap_eb.py'
+eb_bootstrap_url='https://raw.githubusercontent.com/ocaisa/easybuild-framework/d2734f33613332ff283b24faabd84a727d7883f9/easybuild/scripts/bootstrap_eb.py'
```
ocaisa: This needs to be removed (and the above uncommented) when #1880 is merged

ocaisa:
The usage is simple: `./Python_EasyBuild_Lmod_bootstrap.sh <Target Dir>`

easybuild/scripts/Python_EasyBuild_Lmod_bootstrap.sh
```diff
@@ -0,0 +1,342 @@
...
+  mkdir -p $sources/TCL_MOD
+  cd $sources
+  # Grab the file from sourceforge
+  download_file $tclmod_url
+  # It has a funky name because of the URL, let's rename appropriately
+  mv modulecmd.tcl*raw $sources/TCL_MOD/modulecmd.tcl
+fi
+chmod +x $sources/TCL_MOD/modulecmd.tcl 
+export PATH=$PATH:$sources/TCL_MOD
+eval `tclsh $(which modulecmd.tcl) sh autoinit`
```
ocaisa: Need to ensure MODULEPATH is defined before doing this

ocaisa:
@boegel Can I get your opinion on something. I ran this on an absolutely minimal system and there are a few issues. First up you need `tcl`, `zlib-devel` and `openssl-devel` to get a functional Python and the ability to use the TCL module command. 
Next is that you need a C++ compiler to pass the sanity checks of `ncurses`, which is a dep of Lua. Then `ncurses` needs to be loaded for Lua to work. I can create an ncurses build that doesn't require C++ and statically link that to Lua with a config option which means this script should work on a bare system with those few packages. The problem would be if there are more things in built with dummy that actually require `g++` to exist (specifically the deps of GCCcore), if so then it's a bit of a waste of time getting around it since the end result is not really functional. Opinion?

ocaisa:
I went ahead and tested, I fall over at `binutils`. I could probably get a build out I think but might be a bit of effort.

ocaisa:
Ok, binutils will build if I disable `ld.gold` so it *is* possible to forego a C++ compiler and build `GCCcore`

ocaisa:
Aw forget it, `GCCcore` is failing now, looks like no way around it

boegel:
@ocaisa the latest versions of `GCC` are (partially) written in C++, so I think you'll have a hard time getting that to build without a C++ compiler on the system...

boegel:
@ocaisa unless you're willing to build on old version of `GCC` (that is still C-only), just to provide working C++ compiler, then you can probably get away with only requiring a C compiler on the system

PR #1883, opened 2016-08-19
add support for --list-software

boegel:
TODO:

* [x] figure out a way to speed this up significantly
  * ~~currently takes ~4m to print the full list of software because it involves parsing *all* 6.5k easyconfigs~~
  * ~~options:~~
     * ~~speed up parsing of easyconfigs significantly (involves profiling current codebase in detail)~~
     * ~~do not create `EasyConfig` instances at all, and only parse what we really need (`name`, `version` and `toolchain` lines)~~
     * ~~implement support for caching `EasyConfig` instances into a file on disk (which also involves support for updating the cache, etc.)~~
  * now takes only about 10s with 6.5k easyconfigs by avoiding to create full `EasyConfig` instances
* [x] ~~support different levels of detail: only software names, also incl. versions, also incl. toolchains~~
  * see `--list-software` (equivalent with `--list-software=simple`) vs `--list-software=detailed`
* [x] also support `rst` output format
* [x] also take into account `versionsuffix`
* [x] add test(s)
* [ ] update documentation


boegel:
@wpoely86 do you mind giving this a review before I add tests?

Rendered result of `--list-software=detailed --output-format=rst` is available at http://boegel-eb.readthedocs.io/en/list_software/version-specific/Supported_software.html

migueldiascosta:
@boegel for generating local documentation, would it make sense to add an option to only list installed software, or somehow highlight the ones that are already installed? 

Of course, there's `module avail` :), but it would be nice to have it in `rst` format (ok, one can always post-process the full list)

boegel:
@migueldiascosta that's a good idea, but maybe that should be a separate option like `--list-installed-software` that hooks into the functionality implemented here?

I think it's best to tackle that in a follow-up PR.

easybuild/tools/docs.py
```diff
@@ -514,6 +522,160 @@ def add_class(classes, cls):
...
+            'versionsuffix': ec.get('versionsuffix', '') % template_values,
+        })
+
+    print_msg("Found %d different software packages" % len(software), silent=silent)
+
+    return generate_doc('list_software_%s' % output_format, [software, detailed])
+
+
+def list_software_rst(software, detailed=False):
+    """Return overview of supported software in RST format"""
```
wpoely86: doc args

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -924,7 +924,9 @@ def _generate_template_values(self, ignore=None, skip_lower=True):
         # (eg the run_setp code in EasyBlock)
 
         # step 1-3 work with easyconfig.templates constants
-        template_values = template_constant_dict(self._config, ignore=ignore, skip_lower=skip_lower)
+        self.enable_templating = False
```
boegel: done

easybuild/tools/docs.py
```diff
@@ -514,6 +522,160 @@ def add_class(classes, cls):
     return '\n'.join(txt)
 
 
+def list_software(output_format=FORMAT_TXT, detailed=False):
```
boegel: done

easybuild/tools/docs.py
boegel: done

boegel:

PR #1884, opened 2016-08-22
make sure self.path is passed down in copy method of EasyConfig object

boegel:


wpoely86:
:+1:

boegel:

PR #1886, opened 2016-08-25
print message on which extension is being installed, incl. progress counter

### LINKED ISSUE #1507: Report progress during extentions installation
It would be nice to have some form of feedback during the `== taking care of extensions...` phase. Some easyconfigs spend the majority of time in there so any indication of progress would be nice.

Something like this:

```
...
== testing...
== installing...
== * Working on extension a-1.23
== * Working on extension a-1.23
== * Working on extension a-1.23
== * Working on extension a-1.23
== postprocessing...
...
```

boegel:
fixes #1507

results into output like:

```
...
== installing...
== installing extensions BiocGenerics v0.18.0 (1/109)...
== installing extensions Biobase v2.32.0 (2/109)...
== installing extensions S4Vectors v0.10.3 (3/109)...
== installing extensions IRanges v2.6.1 (4/109)...
...
```

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1651,7 +1651,7 @@ class EasyBlock(object):
         exts_cnt = len(self.exts)
         for idx, ext in enumerate(self.exts):
             self.log.debug("Starting extension %s" % ext['name'])
-            print_msg("installing extensions %s v%s (%d/%d)..." % (ext['name'], ext['version'], idx+1, exts_cnt))
+            print_msg("installing extension %s v%s (%d/%d)..." % (ext['name'], ext['version'], idx+1, exts_cnt))
 
             # always go back to original work dir to avoid running stuff from a dir that no longer exists
             os.chdir(self.orig_workdir)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1651,7 +1651,7 @@ class EasyBlock(object):
         exts_cnt = len(self.exts)
         for idx, ext in enumerate(self.exts):
             self.log.debug("Starting extension %s" % ext['name'])
-            print_msg("installing extension %s v%s (%d/%d)..." % (ext['name'], ext['version'], idx+1, exts_cnt))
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86, @serverhorror  please review, do you think this is *too* verbose?

@jhein32: would this help in raising awareness that e.g. Python also provides a bunch of auxiliary packages like `numpy`, `scipy`, etc.?

wpoely86:
Fine by me. I personally always use `-l` option.

jhein32:
@boegel Hi Kenneth,

For sure this is something.  However I often feel the need for more documentation for a specifig easyconfig (not sure about easyblocks yet).   Typical case, is, if the source tarball needs to be installed by hand.  How is it supposed to be called.  

It might be an idea to offer a documentation feature.  The provider (upgrader) of an easyconfig could write what he things is important for a person installing the package.  Perhaps a field like "install-info" in the config.   One could get the printed with a dedicated eb option or as part of the "dry-run".

In case of the Scipy/Numpy easyconfigs, I would include in the install-info, that this config is essentially obsolete since the underlying Python config has a reasonably modern Scipy/Numpy included.

jhein32:
Since you have changed something quite basic, would that also now change the behaviour of R (another package which spends a lot of time in building extensions)?  Such a thing is useful to monitor progress as well.

boegel:
@jhein32 Can you open a separate issue at https://github.com/hpcugent/easybuild-framework/issues w.r.t. the `install_info` idea? I like it, but we should make sure we can automate it as much as possible, otherwise we'll just end up with another documentation source that can go out-of-date in no time...

This change applies to all types of extensions, including Perl, R, Python, etc., since it's a generic change in the framework. It should get included with the upcoming EasyBuild v2.9.0.

wpoely86:
I'm against another source of info like this. We already use git for the history. If we do a `install_info` thingie, it should use the git logs.

serverhorror:
That looks very nice. Exactly what I had in mind, thank you so much

boegel:
Going in, feedback on @jhein32's idea can be added in #1888

PR #1889, opened 2016-08-26
Target correct path when using bunzip2

sebth:
The correct target path is not used when calling bunzip2 to handle
bzip2 files. This causes easyblocks such as BinariesTarball to fail for single bzip2 files.

To fix this problem, we can call bunzip2 the same way gunzip is called to handle gzip files. The gzip functionality has a test case to catch targeting errors, however the bzip2 functionality has not. We can add a similar test case to catch targeting errors for bzip2.


boegel:
@sebth nice catch! Do you have an example easyconfig where you were hitting this problem with?

sebth:
I have not found any in the repository, but I stumbled upon the bug when I wrote this:
```python
easyblock = 'BinariesTarball'
name = 'XDSSTAT'
version = '20151203'

homepage = 'http://strucbio.biologie.uni-konstanz.de/xdswiki/index.php/Xdsstat'
description = ('A home-brewn program that prints various statistics '
               '(that are not available from XDS itself) in the form of '
               'tables and images.')

source_urls = ['ftp://turn5.biologie.uni-konstanz.de/pub/']
sources = ['%(namelower)s-linux64.bz2']
checksums = ['ddf18f58aa191dec91e112dc14982b36']

# XDSSTAT binary is linked to GCC runtime.
toolchain = {'name': 'goolf', 'version': '1.7.20'}
dependencies = [('CCP4', '7.0')]

# Default sanity_check_paths dirs include both 'bin' and 'lib'.
# XDSSTAT has no 'lib'.
sanity_check_paths = {'files': [], 'dirs': ['bin']}

moduleclass = 'data'
```

boegel:
Going in, thanks @sebth!

PR #1892, opened 2016-08-29
don't resolve path to Lmod command

boegel:
there is no good reason to resolve the path to the Lmod command, which was added in #1793 for the sake of the tests

resolving the path may lead to problem in practice, when `eb` is being run while the Lmod installation is being upgraded

cc @wpoely86


wpoely86:
@
:+1:

boegel:

Issue #1887, opened 2016-08-25
bootstrap script fails on Ubuntu 16.04


riccardomurri:
The bootstrap script apparently fails on Ubuntu 16.04 ("xenial"), because of a
conflict between the "distribute" package included in `bootstrap_eb.py` and the
system-provided setuptools:

```
+++ STAGE 0: installing distribute via included (patched) distribute_setup.py...


[[DEBUG]] preparing environment...
[[DEBUG]] Preparing for path /tmp/tmppD0F_b
[[DEBUG]] os.environ['PYTHONPATH'] after reset:
[[DEBUG]] $EASYBUILD_MODULES_TOOL set to Lmod
[[DEBUG]] distribute_setup.__file__: /tmp/tmppD0F_b/distribute_setup.py
Downloading http://hpcugent.github.io/easybuild/files/distribute-0.6.49-patched1.tar.gz
Extracting in /tmp/tmp_9cDov
Now working in /tmp/tmp_9cDov/distribute-0.6.49
Installing Distribute
Traceback (most recent call last):
  File "setup.py", line 45, in <module>
    exec(init_file.read(), d)
  File "<string>", line 8, in <module>
  File "/tmp/tmp_9cDov/distribute-0.6.49/setuptools/__init__.py", line 2, in <module>
    from setuptools.extension import Extension, Library
  File "/tmp/tmp_9cDov/distribute-0.6.49/setuptools/extension.py", line 5, in <module>
    from setuptools.dist import _get_unpatched
  File "/tmp/tmp_9cDov/distribute-0.6.49/setuptools/dist.py", line 6, in <module>
    from setuptools.command.install import install
  File "/tmp/tmp_9cDov/distribute-0.6.49/setuptools/command/__init__.py", line 8, in <module>
    from setuptools.command import install_scripts
  File "/tmp/tmp_9cDov/distribute-0.6.49/setuptools/command/install_scripts.py", line 3, in <module>
    from pkg_resources import Distribution, PathMetadata, ensure_directory
  File "/tmp/tmp_9cDov/distribute-0.6.49/pkg_resources.py", line 2897, in <module>
    add_activation_listener(lambda dist: dist.activate())
  File "/tmp/tmp_9cDov/distribute-0.6.49/pkg_resources.py", line 712, in subscribe
    callback(dist)
  File "/tmp/tmp_9cDov/distribute-0.6.49/pkg_resources.py", line 2897, in <lambda>
[...]

PR #1893, opened 2016-08-29
Fix issue #1887

riccardomurri:
Fix "TypeError: putenv() argument 2 must be string" when skipping stage 0.

See comments on issue #1887 for more details.


boegel:
Thanks @riccardomurri!

PR #1896, opened 2016-09-02
fix function signature of log.deprecated compared to fancylogger.deprecated

boegel:
cfr. https://github.com/hpcugent/vsc-base/issues/236

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -120,10 +120,20 @@ class EasyBuildLog(fancylogger.FancyLogger):
             msg = 'Experimental functionality. Behaviour might change/be removed later (use --experimental option to enable). ' + msg
             raise EasyBuildError(msg, *args)
 
-    def deprecated(self, msg, max_ver):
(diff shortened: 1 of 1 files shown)


easybuild/tools/build_log.py
```diff
@@ -120,10 +120,20 @@ def experimental(self, msg, *args, **kwargs):
             msg = 'Experimental functionality. Behaviour might change/be removed later (use --experimental option to enable). ' + msg
             raise EasyBuildError(msg, *args)
 
-    def deprecated(self, msg, max_ver):
-        """Print deprecation warning or raise an EasyBuildError, depending on max version allowed."""
-        msg += "; see %s for more information" % DEPRECATED_DOC_URL
-        fancylogger.FancyLogger.deprecated(self, msg, str(CURRENT_VERSION), max_ver, exception=EasyBuildError)
+    def deprecated(self, msg, ver, max_ver=None):
```
JensTimmerman: add an `*args, **kwargs` so you catch everything that could be added in the future.

boegel: done, please recheck?

PR #1897, opened 2016-09-05
avoid needless use of deepcopy, speed up support for templating in easyconfigs

boegel:
These changes are motivated by profiing the `easyconfigs` test suite, which was starting to take prohibitively long (> 10m).

Together with https://github.com/hpcugent/easybuild-easyconfigs/pull/3520, this speeds up the `easyconfigs` test suite by a factor 3x...

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1164,13 +1164,14 @@ def resolve_template(value, tmpl_dict):
         # '%%' -> '%%%%'
         # '%(name)s' -> '%(name)s'
         # '%%(name)s' -> '%%(name)s'
-        value = PRESERVE_TEMPLATE_REGEX.sub(r'\1\1', value)
+        if '%' in value:
+            value = PRESERVE_TEMPLATE_REGEX.sub(r'\1\1', value)
 
-        try:
-            value = value % tmpl_dict
-        except KeyError:
-            _log.warning("Unable to resolve template value %s with dict %s" %
-                             (value, tmpl_dict))
+            try:
+                value = value % tmpl_dict
+            except KeyError:
+                _log.warning("Unable to resolve template value %s with dict %s" %
+                                 (value, tmpl_dict))
     else:
         # this block deals with references to objects and returns other references
         # for reading this is ok, but for self['x'] = {}
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -79,8 +79,6 @@ MANDATORY_PARAMS = ['name', 'version', 'homepage', 'description', 'toolchain']
 # set of configure/build/install options that can be provided as lists for an iterated build
 ITERATE_OPTIONS = ['preconfigopts', 'configopts', 'prebuildopts', 'buildopts', 'preinstallopts', 'installopts']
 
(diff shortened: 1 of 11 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -323,7 +325,7 @@ def __init__(self, path, extra_options=None, build_specs=None, validate=True, hi
 
         # list of *all* dependencies, including hidden/build deps & toolchain, but excluding filtered deps
         self.all_dependencies = copy.deepcopy(self.dependencies())
-        if self.toolchain.name != DUMMY_TOOLCHAIN_NAME:
+        if self['toolchain']['name'] != DUMMY_TOOLCHAIN_NAME:
```
wpoely86: is this a pure style change?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -79,6 +79,8 @@
 # set of configure/build/install options that can be provided as lists for an iterated build
 ITERATE_OPTIONS = ['preconfigopts', 'configopts', 'prebuildopts', 'buildopts', 'preinstallopts', 'installopts']
 
+PRESERVE_TEMPLATE_REGEX = re.compile(r'(%)(?!%*\(\w+\)s)')
```
wpoely86: just make the actual regex a global constant? There is no benefit from storing the compiled version.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -919,9 +926,7 @@ def _generate_template_values(self, ignore=None, skip_lower=True):
         # (eg the run_setp code in EasyBlock)
 
         # step 1-3 work with easyconfig.templates constants
-        # use a copy to make sure the original is not touched/modified
-        template_values = template_constant_dict(copy.deepcopy(self._config),
-                                                 ignore=ignore, skip_lower=skip_lower)
+        template_values = template_constant_dict(self._config, ignore=ignore, skip_lower=skip_lower)
```
wpoely86: dropping the deepcopy has no repercussions?

easybuild/framework/easyconfig/easyconfig.py
boegel: no, it's avoiding that a `Toolchain` instance is created already at this point, which is not needed at all for `dummy` toolchains here...

easybuild/framework/easyconfig/easyconfig.py
boegel: sure there is, compiling takes time too... That's the whole point of precompiling a regex, in case you'll be using it over and over again.

easybuild/framework/easyconfig/easyconfig.py
boegel: no, since `template_constant_dict` only uses the 1st argument read-only

this was a better-safe-than-sorry move, which is not required and turns out to be quite expensive, especially for the `easyconfigs` test suite

easybuild/framework/easyconfig/easyconfig.py
wpoely86: regex are cached. Unless you use it in a lot of places, it has no point.

wpoely86: and it makes the import slower.

boegel: OK, changed to no longer use the constant, thanks!

boegel:

PR #1898, opened 2016-09-05
fix 'colored' argument for review_pr, bump vsc-base requirement, define COLOR_* constants in options.py

boegel:
~~This PR fixes the following problem with current `develop` in case the `vsc-base` version is outdated.~~

~~Since this is basically just a workaround for `vsc-base` being older than version 2.5.3 (cfr. https://github.com/hpcugent/vsc-base/pull/225), maybe we should just bump the `vsc-base` requirement instead?~~

update: this PR now basically just bumps the `vsc-base` version requirement; if you're running into the error message below, update your `vsc-base` installation

cc @riccardomurri, @wpoely86

```
  File "/usr/lib64/python2.6/runpy.py", line 122, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.6/runpy.py", line 34, in _run_code
    exec code in run_globals
  File "/vscmnt/gent_vulpix/_/user/home/gent/vsc400/vsc40003/easybuild_easyinstalled/lib/python2.6/site-packages/easybuild_framework-2.9.0.dev0-py2.6.egg/easybuild/main.py", line 47, in <module>
    from easybuild.tools.build_log import EasyBuildError, init_logging, print_msg, print_error, stop_logging
  File "/user/home/gent/vsc400/vsc40003/easybuild_easyinstalled/lib/python2.6/site-packages/easybuild_framework-2.9.0.dev0-py2.6.egg/easybuild/tools/build_log.py", line 201, in <module>
    def init_logging(logfile, logtostdout=False, silent=False, colorize=fancylogger.Colorize.AUTO):
AttributeError: 'module' object has no attribute 'Colorize'
```
~~

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -100,8 +100,6 @@ XDG_CONFIG_DIRS = os.environ.get('XDG_CONFIG_DIRS', '/etc').split(os.pathsep)
 DEFAULT_SYS_CFGFILES = [f for d in XDG_CONFIG_DIRS for f in sorted(glob.glob(os.path.join(d, 'easybuild.d', '*.cfg')))]
 DEFAULT_USER_CFGFILE = os.path.join(XDG_CONFIG_HOME, 'easybuild', 'config.cfg')
 
-(COLOR_AUTO, COLOR_ALWAYS, COLOR_NEVER) = fancylogger.Colorize
-
 
 _log = fancylogger.getLogger('options', fname=False)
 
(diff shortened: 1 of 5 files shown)


JensTimmerman:
very much in favour of bumping the vsc-base version requirement!

boegel:
@JensTimmerman on second thought, I agree

I also fixed the failing `--review-pr` test, caused by not using `use_color` to determine whether colored output should be used or not.

wpoely86:
:+1:

easybuild/tools/options.py
```diff
@@ -176,7 +171,7 @@ def use_color(colorize, stream=sys.stdout):
     elif colorize == COLOR_ALWAYS:
         return True
     else:
-        assert colorize == COLOR_NEVER, "Argument `colorize` must be one of: %s" % ', '.join(COLOR_OPTIONS)
+        assert colorize == COLOR_NEVER, "Argument `colorize` must be one of: %s" % ', '.join(fancylogger.Colorize)
```
riccardomurri: This should end with `.join([ COLOR_ALWAYS, COLOR_AUTO, COLOR_NEVER ])` if you insist in decoupling this part from `vsc.fancylogger` i.e. not using the same named constants in EB and `vsc.fancylogger`.

Otherwise just drop `COLOR_*` and use `fancylogger.Colorize` everywhere...

boegel: good point, done!

boegel:
Thanks for the review everyone!

PR #1899, opened 2016-09-05
don't pass down 'max_ver' as keyword to fancylogger.deprecated

boegel:
This should have been part of #1896...

@JensTimmerman please review?

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -130,6 +130,9 @@ class EasyBuildLog(fancylogger.FancyLogger):
         :param max_ver: version threshold for warning vs exception (compared to 'ver')
         """
         max_ver = kwargs.pop('max_ver', None)
+        if max_ver is None and args:
+            max_ver = args[0]
+            args = args[1:]
         if max_ver is None:
             msg += "; see %s for more information" % DEPRECATED_DOC_URL
             kwargs['exception'] = EasyBuildError
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -129,13 +129,13 @@ class EasyBuildLog(fancylogger.FancyLogger):
                     else: version to check against max_ver to determine warning vs exception
         :param max_ver: version threshold for warning vs exception (compared to 'ver')
         """
-        max_ver = kwargs.get('max_ver', None)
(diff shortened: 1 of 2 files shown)


JensTimmerman:
ok, can you add a test case that would have triggered this?

boegel:
@JensTimmerman good point, adding tests made me realise there was another issue in `log.deprecated`, since `max_ver` can also be passed unnamed via `*args`

please rerereview?

easybuild/tools/build_log.py
```diff
@@ -129,13 +129,16 @@ def deprecated(self, msg, ver, *args, **kwargs):
                     else: version to check against max_ver to determine warning vs exception
         :param max_ver: version threshold for warning vs exception (compared to 'ver')
         """
-        max_ver = kwargs.get('max_ver', None)
+        max_ver = kwargs.pop('max_ver', None)
+        if max_ver is None and args:
+            max_ver = args[0]
+            args = args[1:]
```
JensTimmerman: you have no way to know that max_ver will be args[0], so please don't do this, if you expect it, 
make the function definition 
`def deprecated(self, msg, ver, max_ver=None,*args, **kwargs):`
`

easybuild/tools/build_log.py
```diff
@@ -129,13 +129,16 @@ def deprecated(self, msg, ver, *args, **kwargs):
...
         if max_ver is None:
             msg += "; see %s for more information" % DEPRECATED_DOC_URL
             kwargs['exception'] = EasyBuildError
             fancylogger.FancyLogger.deprecated(self, msg, str(CURRENT_VERSION), ver, *args, **kwargs)
         else:
-            fancylogger.FancyLogger.deprecated(self, msg, ver, max_ver)
+            fancylogger.FancyLogger.deprecated(self, msg, ver, max_ver, *args, **kwargs)
```
JensTimmerman: your pr title states:

easybuild/tools/build_log.py
boegel: good point, thanks

easybuild/tools/build_log.py
boegel: not as a keyword, I'll clarify the title

JensTimmerman:
looks good to me

boegel:

PR #1901, opened 2016-09-09
Tcl8.6 travis patch

wpoely86:


easybuild/scripts/install_eb_dep.sh
```diff
@@ -53,6 +53,13 @@ if [ $? -ne 0 ] && [ ! -z $BACKUP_PKG_URL ]; then
     wget ${BACKUP_PKG_URL} && tar xfz *${PKG_VERSION}.tar.gz
 fi
 set -e
+
+# environment-modules needs a patch to work with Tcl8.6
+if [ x$PKG_NAME == 'xmodules' ]; then
+    wget -O 'modules-tcl8.6.patch' 'https://sourceforge.net/p/modules/feature-requests/_discuss/thread/4502a6c9/9aa5/attachment/environment-modules-tcl86.patch'
```
boegel: @wpoely86 use the copy you uploaded in https://github.com/hpcugent/easybuild/pull/246 ?

Also, make this conditional on something, we only want to do this if we're actually using Tcl 8.6?

wpoely86: why conditional? The patch fixes it in such a way it works for both 8.5 and 8.6?

boegel:
Thanks @wpoely86!

PR #1902, opened 2016-09-10
fix combining --extended-dry-run with --from-pr

boegel:
Without this patch, combining `--from-pr` with `-x` results in an error like:

```
EasyBuildError: '/tmp/eb-96zkBo/tmpaaRyAU/hpcugent/easybuild-easyconfigs-develop
should exist and contain the repo easybuild-easyconfigs at branch develop'
```

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -171,7 +171,7 @@ def remove_file(path):
         raise EasyBuildError("Failed to remove %s: %s", path, err)
 
 
-def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False):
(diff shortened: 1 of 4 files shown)


boegel:
@wpoely86 please review?

easybuild/tools/filetools.py
```diff
@@ -171,7 +171,7 @@ def remove_file(path):
         raise EasyBuildError("Failed to remove %s: %s", path, err)
 
 
-def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False):
+def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced=False):
```
wpoely86: document the new argument

boegel:

PR #1908, opened 2016-09-15
Don't pass '--try-*' command-line options to EB instance running within job script

geimer:
As tweaking the easyconfigs is already done by the driver `eb` instance on the head node, `--try-*` command-line options don't need to be passed to the build instances running within the job scripts.  Thus, filter them out.  This should ~~fix~~ address https://github.com/hpcugent/easybuild-framework/issues/1475 (a shared `--tmpdir` is still needed as mentioned in https://github.com/hpcugent/easybuild-framework/issues/1355).

Tested with `eb --try-toolchain=gompi,2016a LAPACK-3.4.2-gompi-1.5.12.eb --job`; fails w/o the fix, succeeds with it.

### DIFF
--- a/easybuild/tools/parallelbuild.py
+++ b/easybuild/tools/parallelbuild.py
@@ -127,7 +127,9 @@ def submit_jobs(ordered_ecs, cmd_line_opts, testing=False, prepare_first=True):
     curdir = os.getcwd()
 
     # the options to ignore (help options can't reach here)
-    ignore_opts = ['robot', 'job']
+    ignore_opts = ['robot', 'job', 'try-amend',
+                   'try-software', 'try-software-name', 'try-software-version',
+                   'try-toolchain', 'try-toolchain-name', 'try-toolchain-version']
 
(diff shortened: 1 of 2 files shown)


easybuild/tools/parallelbuild.py
```diff
@@ -127,7 +127,9 @@ def submit_jobs(ordered_ecs, cmd_line_opts, testing=False, prepare_first=True):
     curdir = os.getcwd()
 
     # the options to ignore (help options can't reach here)
-    ignore_opts = ['robot', 'job']
+    ignore_opts = ['robot', 'job', 'try-amend',
+                   'try-software', 'try-software-name', 'try-software-version',
+                   'try-toolchain', 'try-toolchain-name', 'try-toolchain-version']
```
boegel: @geimer let's avoid hardcoding all `--try-*` options here, cfr. https://github.com/geimer/easybuild-framework/pull/11

boegel:
Going in, thanks @geimer!

PR #1909, opened 2016-09-15
convert all_dependencies to a property in EasyConfig class

boegel:
This is mainly to avoid that the `toolchain` property is being accessed when an `EasyConfig` instance is created, since that i) is time-consuming, ii) can cause problems with 'special' toolchains like the `Cray*` toolchains, where a `craype-<optarch>` module *must* be available when the toolchain is instantiated (which is rather silly to do when just parsing the easyconfig file).

`all_dependencies` is only used in specific situations, i.e. when using `--dep-graph`, with `--job --robot`, and in `--check-conflicts`.

The proposed change ensures that `all_dependencies` is correctly defined when it's first accessed.

This will help speed up parsing of easyconfig files a little bit too.

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -321,10 +321,7 @@ class EasyConfig(object):
         # filter hidden dependencies from list of dependencies
         self.filter_hidden_deps()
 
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86 please review

boegel:
Thanks @wpoely86!

PR #1910, opened 2016-09-15
add support for --list-installed-software

boegel:
Equivalent to `--list-software`, but only retains software for which a module is already available (idea by @ migueldiascosta, cfr. https://github.com/hpcugent/easybuild-framework/pull/1883#issuecomment-247210236)

Example output on a system where only a handful of modules are installed:

```
$ time eb --list-installed-software=detailed
== temporary log file in case of crash /tmp/eb-pYOgVv/easybuild-jIlNHN.log
== Processed 7000/7000 easyconfigs... 
== Found 1129 different software packages
== Retained 9 installed software packages

* Autoconf

Autoconf is an extensible package of M4 macros that produce shell scripts  to automatically configure software source code packages. These scripts can adapt the  packages to many kinds of UNIX-like systems without manual user intervention. Autoconf  creates a configuration script for a package from a template file that lists the  operating system features that the package can use, in the form of M4 macro calls.

homepage: http://www.gnu.org/software/autoconf/

  * Autoconf v2.69: dummy

* Automake

Automake: GNU Standards-compliant Makefile generator

homepage: http://www.gnu.org/software/automake/automake.html

  * Automake v1.15: dummy

* Autotools

This bundle collect the standard GNU build tools: Autoconf, Automake and libtool

homepage: http://autotools.io

  * Autotools v20150215: dummy

* bzip2

bzip2 is a freely available, patent free, high-quality data compressor. It typically  compresses files to within 10% to 15% of the best available techniques (the PPM family of statistical  compressors), whilst being around twice as fast at compression and six times faster at decompression.

homepage: http://www.bzip.org/

  * bzip2 v1.0.6: dummy

* EasyBuild

EasyBuild is a software build and installation framework  written in Python that allows you to install software in a structured,  repeatable and robust way.

homepage: http://hpcugent.github.com/easybuild/

  * EasyBuild v1.0.0: dummy
  * EasyBuild v2.8.1: dummy
  * EasyBuild v2.8.2: dummy

* libtool

GNU libtool is a generic library support script. Libtool hides the complexity of using shared libraries  behind a consistent, portable interface.

homepage: http://www.gnu.org/software/libtool

  * libtool v2.4.6: dummy

* M4

GNU M4 is an implementation of the traditional Unix macro processor.  It is mostly SVR4 compatible although it has some extensions   (for example, handling more than 9 positional parameters to macros).  GNU M4 also has built-in functions for including files, running shell commands, doing arithmetic, etc. 

homepage: http://www.gnu.org/software/m4/m4.html

  * M4 v1.4.17: dummy

* S.A.G.E.

 S.A.G.E. (Statistical Analysis for Genetic  Epidemiology) is free software package containing programs for  use in the genetic analysis of family, pedigree and individual data.

homepage: https://code.google.com/archive/p/opensage/

  * S.A.G.E. v6.3: dummy

* zlib

zlib is designed to be a free, general-purpose, legally unencumbered -- that is,  not covered by any patents -- lossless data-compression library for use on virtually any  computer hardware and operating system.

homepage: http://www.zlib.net/

  * zlib v1.2.8: dummy

```

TODO:

* [x] tests
* [ ] docs


boegel:
@wpoely86 please review?

boegel:

PR #1912, opened 2016-09-16
replace --use-compiler-cache with --use-ccache and --use-f90cache

boegel:
The most recent version of `f90cache` doesn't support GCC 5.x, and `ccache` 3.3 supports Fortran(77).

Besides that, it makes sense to split this up in order not to enforce using *both* `ccache` and `f90cache`, especially since the latter *breaks* Fortran compilation for recent GCC versions by doing an `exit` if the GCC version is too new (rather than acting as a fall through , which it should be doing).

Since `--use-compiler-cache` is not included in a released EasyBuild version yet, I feel it's OK to make this change now.


easybuild/tools/toolchain/toolchain.py
```diff
@@ -647,41 +651,64 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
                 self.generate_vars()
                 self._setenv_variables(onlymod, verbose=not silent)
 
-        if build_option('use_compiler_cache'):
-            self.prepare_compiler_cache()
+        # consider f90cache first, since ccache can also wrap Fortran compilers
+        for cache_tool in [F90CACHE, CCACHE]:
```
wpoely86: should the list of all tools also be a global?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -647,41 +651,64 @@ def prepare(self, onlymod=None, silent=False, loadmod=True):
...
+        Prepare for using specified compiler caching tool (e.g., ccache, f90cache)
+
+        :param cache_tool: name of compiler caching tool to prepare for
+        """
+        compilers = self.comp_cache_compilers(cache_tool)
+        self.log.debug("Using compiler cache tool '%s' for compilers: %s", cache_tool, compilers)
+
+        # set paths that should be used by compiler caching tool
+        comp_cache_path = build_option('use_%s' % cache_tool)
+        setvar('%s_DIR' % cache_tool.upper(), comp_cache_path)
```
wpoely86: are these values printed somewhere (for debug)?

easybuild/tools/toolchain/toolchain.py
boegel: it's only used in one place now, and it has to be in a specific order here, so I don't see the point in making it available globally? it'll make it harder to see why it's in a particular order

easybuild/tools/toolchain/toolchain.py
boegel: `setvar` will issue a log message (which is exactly why I switched to using `setvar`)

boegel:
Thanks @wpoely86!

PR #1913, opened 2016-09-16
take into account existing % chars in to_template_str

boegel:
@wpoely86 easyconfig test suite is currently broken because of this (because of `cmds_map` in the `FastTree` easyconfigs), as a side-effect of the templating-related changes in #1883 

It was wrong all along though, we should've been taking into account `%` already...


boegel:
Thanks @wpoely86!

PR #1914, opened 2016-09-16
take into account self.silent in progress messages for extensions

boegel:


boegel:
Trivial enough to merge in as-is; this fixes a minor printing issue with the tests

PR #1919, opened 2016-09-17
Small bugfix for --modules-only

wpoely86:
If the easyconfig specified a 'start_dir', the prepare step would fail
as that directory will not exist (because we skip the extract_step).

@boegel have a look

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1281,7 +1281,7 @@ class EasyBlock(object):
         -- else, treat it as subdir for regular procedure
         """
         start_dir = ''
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1281,7 +1281,9 @@ def guess_start_dir(self):
         -- else, treat it as subdir for regular procedure
         """
         start_dir = ''
-        if self.cfg['start_dir']:
+        # do not use the specifiy 'start_dir' when running as --module-only as
```
boegel: @wpoely86 please fix the typo here: `specified`

boegel:
Thanks @wpoely86!

Issue #1906, opened 2016-09-14
Disabling MPI testing


jhein32:
A number of easy-configs start MPI tests.  Depending on environment this can create issues and prevents the module from being build.  A command-line option to disable MPI testing would be useful.  

EB-users would (should) know that stuff is not properly tested and would need to resort to some other tests.  But at least this way they get executables and modules for further testing.

boegel:
support for `--disable-mpi-tests` is added in #1920, easyblocks are updated to take it into account in https://github.com/hpcugent/easybuild-easyblocks/pull/993, let's follow up there

PR #1920, opened 2016-09-17
add support for --disable-mpi-tests

boegel:
cfr. feature request #1906 by @jhein32

this only adds a new build option named `mpi_tests`, easyblocks where MPI tests are being run need to be made aware of the `mpi_tests` build option (see https://github.com/hpcugent/easybuild-easyblocks/pull/993)

disabling MPI tests can then be done using `--disable-mpi-tests`


boegel:
@wpoely86 please review

PR #1924, opened 2016-09-20
disable cleanupoldbuild if buildininstalldir is specified, since it's enabled by default

boegel:
this completes the fix for `--modules-only` in #1919 for builds that have `buildininstalldir` enabled


boegel:

PR #1925, opened 2016-09-20
don't touch existing build/install dir under --module-only

boegel:
#1919 and #1924 are not sufficient to avoid that an existing install directory is 'removed';

with #1924, the directory doesn't get removed, but *renamed* (cfr. `make_dir`)...

We need to explicitly avoid touching an existing build/install dir under `--module-only`...


boegel:

PR #1933, opened 2016-09-26
remove existing module file under --force/--rebuild

boegel:
We are currently not removing the module file when a forced reinstall is being done.

That's problematic for a number of reasons:

* while the installation is being performed, the existing module provides access to a partial (or empty) installation
* if the installation fails somewhere halfway through, the module is still there even though the installation is partial/broken
* when installing extensions, the existing module may get loaded rather than the temporary module that is generated when Lmod is used as a modules tool, which can cause problems in case dependencies have been tweaked (this occurred with https://github.com/hpcugent/easybuild-easyconfigs/pull/3613)


boegel:

Issue #1935, opened 2016-09-29
`eb --search` fails if `EASYBUILD_TRY_TOOLCHAIN` is set


riccardomurri:
Invocation of `eb --search` fails if the environment variable `EASYBUILD_TRY_TOOLCHAIN` is set::

    $ export EASYBUILD_TRY_TOOLCHAIN=foss,2016b
    $ eb --search gawk
    == temporary log file in case of crash /tmp/eb-30naRL/easybuild-sqnQB9.log
    == Searching (case-insensitive) for 'gawk' in /tmp/eb-30naRL/tweaked_easyconfigs
    == Searching (case-insensitive) for 'gawk' in /opt/easybuild/ebfiles 
    == Searching (case-insensitive) for 'gawk' in /opt/easybuild/software/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_easyconfigs-2.9.0-py2.7.egg/easybuild/easyconfigs 
     * /opt/easybuild/software/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_easyconfigs-2.9.0-py2.7.egg/easybuild/easyconfigs/g/gawk/gawk-4.0.2-goolf-1.4.10.eb
     * /opt/easybuild/software/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_easyconfigs-2.9.0-py2.7.egg/easybuild/easyconfigs/g/gawk/gawk-4.0.2-ictce-5.3.0.eb
    Traceback (most recent call last):
      File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
        "__main__", fname, loader, pkg_name)
      File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
        exec code in run_globals
      File "/opt/easybuild/software/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/main.py", line 442, in <module>
        main()
      File "/opt/easybuild/software/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/main.py", line 333, in main
        easyconfigs = tweak(easyconfigs, build_specs, modtool, targetdir=tweaked_ecs_path)
[...]

Issue #1371, opened 2015-08-25
search breaks when EASYBUILD_TRY_TOOLCHAIN_VERSION is set


dominikborkowski:
It seems that setting the environment variable EASYBUILD_TRY_TOOLCHAIN_VERSION affects negatively the search function (but it works just fine when building software).

Example:
* working:
```
dom@sfxlogin3:/apps/easybuild> eb -S ngs
== temporary log file in case of crash /tmp/eb-tJSfPg/easybuild-TPk5ci.log
== Searching (case-insensitive) for 'ngs' in /apps/easybuild/ebfiles_repo-vbi/shadowfax-sandy_bridge
== Searching (case-insensitive) for 'ngs' in /apps/easybuild/ebfiles_repo-vbi/generic
== Searching (case-insensitive) for 'ngs' in /apps/easybuild/ebfiles_repo/shadowfax-sandy_bridge
== Searching (case-insensitive) for 'ngs' in /apps/easybuild/ebfiles_repo-vbi/shadowfax-sandy_bridge
== Searching (case-insensitive) for 'ngs' in /apps/easybuild/ebfiles_repo-vbi/generic
== Searching (case-insensitive) for 'ngs' in /apps/easybuild/ebfiles_repo/shadowfax-sandy_bridge
== Searching (case-insensitive) for 'ngs' in /apps/easybuild/software/shadowfax-sandy_bridge/EasyBuild/2.2.0/lib64/python2.6/site-packages/easybuild_easyconfigs-2.2.0-py2.6.egg/easybuild/easyconfigs
CFGS1=/apps/easybuild/software/shadowfax-sandy_bridge/EasyBuild/2.2.0/lib64/python2.6/site-packages/easybuild_easyconfigs-2.2.0-py2.6.egg/easybuild/easyconfigs/n/ngspice
 * $CFGS1/ngspice-26-intel-2014b.eb
== temporary log file(s) /tmp/eb-tJSfPg/easybuild-TPk5ci.log* have been removed.
== temporary directory /tmp/eb-tJSfPg has been removed.
```
* broken:
```
dom@sfxlogin3:/apps/easybuild> export EASYBUILD_TRY_TOOLCHAIN_VERSION=1.7.20
dom@sfxlogin3:/apps/easybuild> eb -S ngs
== temporary log file in case of crash /tmp/eb-N_tMcj/easybuild-N5D92K.log
[...]

PR #1937, opened 2016-09-30
fix combining --search and --try-*

boegel:
fix for bug reported in #1371 and #1935

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -94,18 +94,15 @@ def tweak(easyconfigs, build_specs, modtool, targetdir=None):
     # keep track of originally listed easyconfigs (via their path)
     listed_ec_paths = [ec['spec'] for ec in easyconfigs]
 
-    # obtain full dependency graph for specified easyconfigs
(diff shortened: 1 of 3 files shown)


boegel:
@wpoely86 please review

easybuild/framework/easyconfig/tweak.py
```diff
@@ -94,18 +94,15 @@ def tweak(easyconfigs, build_specs, modtool, targetdir=None):
     # keep track of originally listed easyconfigs (via their path)
     listed_ec_paths = [ec['spec'] for ec in easyconfigs]
 
-    # obtain full dependency graph for specified easyconfigs
-    # easyconfigs will be ordered 'top-to-bottom': toolchain dependencies and toolchain first
-    orig_ecs = resolve_dependencies(easyconfigs, modtool, retain_all_deps=True)
```
wpoely86: why is this gone?

boegel: this is actually a bug, this line is overwriting the `orig_ecs` defined above in the `if-else`...

boegel:

PR #1942, opened 2016-10-04
add support for RPATH linking (REVIEW)

boegel:
(continuation of the work started in #1613 by @rjeschmi)

With `--rpath` enabled, EasyBuild will generate a `bash` wrapper script for each compiler and linker command (e.g. `icc`, `icpc`, `ifort`, `ld` and `ld.gold`) during the toolchain `prepare` phase, where a couple of things are hardcoded into:
- the name of the command being wrapped
- the path to the original command
- the path to a Python utility script (`rpath_args.py`)
- the path to a log file specific to the wrapper script

`rpath_args.py` takes care of determining which `-rpath` option should be passed down to the original command `$RPATH`), based on the specified command line arguments for that command.
It may also filter/tweak the specified command line arguments that will actually be passed down to the original command (`$CMD_ARGS`).
The `bash` wrapper script will run `rpath_args.py`, evaluate the statements it prints out (which define `$RPATH` and `$CMD_ARGS`), and then call the original command using `$RPATH` and `$CMD_ARGS`).

The location of these wrapper scripts is prepended to `$PATH` to they get preference over the compiler/linker commands they wrap around.

The sanity check was extended to check the output of `ldd` and `readelf -d` on all dynamically linked executables and libraries (that reside in `bin`, `lib` and `lib64` subdirectories) in case `--rpath` was used.

TODO:

[ ] more testing
- in combination with option that avoids that `$LD_LIBRARY_PATH` is defined by generated modules (TODO)
  [ ] fix the `FIXME`s in `rpath_args.py`
  [ ] end-to-end test for `--rpath`
  [ ] documentation

Feedback welcome!

cc @geimer, @pescobar


boegel:
Hmm, this doesn't look right...

Some builds are failing with errors like:

```
configure:2996: checking whether the C compiler works
configure:3018: gcc    conftest.c  >&5
/tmp/eb-msDPyW/rpath_wrappers/gcc: fork: retry: Resource temporarily unavailable
/tmp/eb-msDPyW/rpath_wrappers/gcc: fork: retry: Resource temporarily unavailable
/tmp/eb-msDPyW/rpath_wrappers/gcc: fork: retry: Resource temporarily unavailable
/tmp/eb-msDPyW/rpath_wrappers/gcc: fork: retry: Resource temporarily unavailable
/tmp/eb-msDPyW/rpath_wrappers/gcc: fork: Resource temporarily unavailable
configure:3022: $? = 254
configure:3060: result: no
```

from `ps axfu`:
```
vsc40023  92965  0.0  0.0 106076  1412 pts/1    S+   19:56   0:00      \_ /bin/bash /user/scratch/gent/vsc400/vsc40023/easybuild_easy_installed/bin/eb --rpath --prefix /tmp/vsc40023/rpathtest HPL-2.1-foss-2016b.
vsc40023  92976 10.1  0.0 252872 24868 pts/1    S+   19:56   0:02          \_ python2 -m easybuild.main --rpath --prefix /tmp/vsc40023/rpathtest HPL-2.1-foss-2016b.eb -dr
vsc40023  99142  0.0  0.0 106080  1480 pts/1    S+   19:56   0:00              \_ /bin/sh ./configure --prefix=/tmp/vsc40023/rpathtest/software/zlib/1.2.8
vsc40023  99161  0.0  0.0 106080   676 pts/1    S+   19:56   0:00                  \_ /bin/sh ./configure --prefix=/tmp/vsc40023/rpathtest/software/zlib/1.2.8
vsc40023  99162  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                      \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99169  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                          \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99176  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                              \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99183  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                  \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99190  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                      \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99197  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                          \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99204  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                              \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99211  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                  \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99218  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                      \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99225  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                          \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99232  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                              \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99239  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                                  \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99246  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                                      \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99253  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                                          \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99260  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                                              \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99267  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                                                  \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
vsc40023  99274  0.0  0.0 106080  1420 pts/1    S+   19:56   0:00                                                                                      \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/gcc -v
...
```

In other words, I've somehow created a fork bomb... ouch

boegel:
similar, with an actual compilation:


```
vsc40023  92965  0.0  0.0 106076  1412 pts/1    S+   19:56   0:00      \_ /bin/bash /user/scratch/gent/vsc400/vsc40023/easybuild_easy_installed/bin/eb --rpath --prefix /tmp/vsc40023/rpathtest HPL-2.1-foss-2016b.eb -dr
vsc40023  92976  0.9  0.0 252872 24868 pts/1    S+   19:56   0:02          \_ python2 -m easybuild.main --rpath --prefix /tmp/vsc40023/rpathtest HPL-2.1-foss-2016b.eb -dr
vsc40023  99142  0.0  0.0 106080  1512 pts/1    S+   19:56   0:00              \_ /bin/sh ./configure --prefix=/tmp/vsc40023/rpathtest/software/zlib/1.2.8
vsc40023 120620  0.0  0.0 101152   880 pts/1    S+   19:59   0:00                  \_ cc -O -D_LARGEFILE64_SOURCE=1 -o ztest99142 ztest99142.c
vsc40023 120623  0.0  0.0 100996   716 pts/1    S+   19:59   0:00                      \_ /usr/libexec/gcc/x86_64-redhat-linux/4.4.7/collect2 --eh-frame-hdr --build-id -m elf_x86_64 --hash-style=gnu -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o ztest99142 /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crt1.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbegin.o -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../.. /tmp/eb-Sas1T3/cc4muTKV.o -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtend.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o
vsc40023 120624  0.0  0.0 106080  1440 pts/1    S+   19:59   0:00                          \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/ld --eh-frame-hdr --build-id -m elf_x86_64 --hash-style=gnu -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o ztest99142 /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crt1.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbegin.o -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../.. /tmp/eb-Sas1T3/cc4muTKV.o -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtend.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o
vsc40023 120631  0.0  0.0 106080  1448 pts/1    S+   19:59   0:00                              \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/ld -rpath=/lib64:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64 --eh-frame-hdr --build-id -m elf_x86_64 --hash-style=gnu -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o ztest99142 /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crt1.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbegin.o -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../.. /tmp/eb-Sas1T3/cc4muTKV.o -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtend.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o
vsc40023 120638  0.0  0.0 106084  1448 pts/1    S+   19:59   0:00                                  \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/ld -rpath=/lib64:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64 -rpath=/lib64:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64 --eh-frame-hdr --build-id -m elf_x86_64 --hash-style=gnu -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o ztest99142 /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crt1.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbegin.o -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../.. /tmp/eb-Sas1T3/cc4muTKV.o -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtend.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o
vsc40023 120645  0.0  0.0 106084  1452 pts/1    S+   19:59   0:00                                      \_ /bin/bash /tmp/eb-Sas1T3/rpath_wrappers/ld -rpath=/lib64:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64 -rpath=/lib64:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64 -rpath=/lib64:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64 --eh-frame-hdr --build-id -m elf_x86_64 --hash-style=gnu -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o ztest99142 /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crt1.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbegin.o -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../.. /tmp/eb-Sas1T3/cc4muTKV.o -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtend.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o
...
```

boegel:
cause of the problem: faulty self-referencing value for `ORIG_CMD` in the generated wrapper script:

```
$ grep ^ORIG_CMD /tmp/eb-Sas1T3/rpath_wrappers/ld
ORIG_CMD=/tmp/eb-Sas1T3/rpath_wrappers/ld
```

boegel:
fork bomb problem is resolved with 91b3589, strict check to avoid reintroducing fork bomb is added in 238b535 + 10f89fc

boegel:
testing this with a rebuild of `HPL-2.1-foss-2016b.eb` from scratch...

RPATH-specific sanity check currently fails on installing `binutils/2.26`, see below

Missing `.so` is available though:

```
$ find /tmp/vsc40023/rpathtest/software/binutils/2.26/ -name libbfd-2.26.20160125.so
/tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libbfd-2.26.20160125.so
```

Do we need to include the `lib` dir of the install path for the software we're installing itself in `-rpath` too?!

```
== 2016-10-14 13:18:49,492 run.py:137 DEBUG run_cmd: running cmd file /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so (in /tmp/vsc40023/rpathtest/software/binutils/2.26)
== 2016-10-14 13:18:49,496 run.py:433 DEBUG cmd "file /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so" exited with exitcode 0 and output:
/tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped

== 2016-10-14 13:18:49,497 run.py:477 DEBUG Using default regular expression: (?<![(,-]|\w)(?:error|segmentation fault|failed)(?![(,-]|\.?\w)
== 2016-10-14 13:18:49,497 run.py:137 DEBUG run_cmd: running cmd ldd /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so (in /tmp/vsc40023/rpathtest/software/binutils/2.26)
== 2016-10-14 13:18:49,505 run.py:433 DEBUG cmd "ldd /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so" exited with exitcode 0 and output:
        linux-vdso.so.1 =>  (0x00002aaaaaacb000)
        libbfd-2.26.20160125.so => not found
        libc.so.6 => /lib64/libc.so.6 (0x00002aaaab086000)
        /lib64/ld-linux-x86-64.so.2 (0x00002aaaaaaab000)

== 2016-10-14 13:18:49,505 run.py:477 DEBUG Using default regular expression: (?<![(,-]|\w)(?:error|segmentation fault|failed)(?![(,-]|\.?\w)
== 2016-10-14 13:18:49,505 easyblock.py:1860 WARNING One or more required libraries not found for /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so:       linux-vdso.so.1 =>  (0x00002aaaaaacb000)
        libbfd-2.26.20160125.so => not found
        libc.so.6 => /lib64/libc.so.6 (0x00002aaaab086000)
        /lib64/ld-linux-x86-64.so.2 (0x00002aaaaaaab000)

== 2016-10-14 13:18:49,505 run.py:137 DEBUG run_cmd: running cmd readelf -d /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so (in /tmp/vsc40023/rpathtest/software/binutils/2.26)
== 2016-10-14 13:18:49,509 run.py:433 DEBUG cmd "readelf -d /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so" exited with exitcode 0 and output:

Dynamic section at offset 0x1a2b08 contains 23 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libbfd-2.26.20160125.so]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000e (SONAME)             Library soname: [libopcodes-2.26.20160125.so]
 0x000000000000000f (RPATH)              Library rpath: [/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/libiberty/pic:/lib64:/tmp/vsc40023/rpathtest/software/Bison/3.0.4/lib:/tmp/vsc40023/rpathtest/software/flex/2.6.0/lib:/tmp/vsc40023/rpathtest/software/zlib/1.2.8/lib:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64:/lib]
 0x000000000000000c (INIT)               0x49a48
 0x000000000000000d (FINI)               0x523a8
 0x000000006ffffef5 (GNU_HASH)           0x1b8
 0x0000000000000005 (STRTAB)             0x6d8
 0x0000000000000006 (SYMTAB)             0x2a0
 0x000000000000000a (STRSZ)              889 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x3a2d08
 0x0000000000000002 (PLTRELSZ)           336 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x498f8
 0x0000000000000007 (RELA)               0xad0
 0x0000000000000008 (RELASZ)             298536 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffe (VERNEED)            0xab0
 0x000000006fffffff (VERNEEDNUM)         1
 0x000000006ffffff0 (VERSYM)             0xa52
 0x000000006ffffff9 (RELACOUNT)          12429
 0x0000000000000000 (NULL)               0x0
```

geimer:
@boegel Of course you also need to include the `lib` dir.  Why should the dynamic linker be able to resolve that one automagically?

boegel:
The plot thickens, since `libopcodes-2.26.20160125.so` *is* found just fine for `objdump`:

```
== 2016-10-14 13:18:49,379 run.py:137 DEBUG run_cmd: running cmd ldd /tmp/vsc40023/rpathtest/software/binutils/2.26/bin/objdump (in /tmp/vsc40023/rpathtest/software/binutils/2.26)                                                                                                                                     
== 2016-10-14 13:18:49,387 run.py:433 DEBUG cmd "ldd /tmp/vsc40023/rpathtest/software/binutils/2.26/bin/objdump" exited with exitcode 0 and output:                                                                                                                                                                     
    linux-vdso.so.1 =>  (0x00002aaaaaacb000)                                                                                                                                                                                                                                                                            
    libopcodes-2.26.20160125.so => /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libopcodes-2.26.20160125.so (0x00002aaaaacce000)                                                                                                                                                                                  
    libbfd-2.26.20160125.so => /tmp/vsc40023/rpathtest/software/binutils/2.26/lib/libbfd-2.26.20160125.so (0x00002aaaab072000)                                                                                                                                                                                          
    libdl.so.2 => /lib64/libdl.so.2 (0x00002aaaab3b0000)                                                                                                                                                                                                                                                                
    libc.so.6 => /lib64/libc.so.6 (0x00002aaaab5b4000)                                                                                                                                                                                                                                                                  
    /lib64/ld-linux-x86-64.so.2 (0x00002aaaaaaab000)   
```

boegel:
linker command for `libopcodes-2.26.20160125.so`:

```
/bin/sh ./libtool --tag=CC   --mode=link gcc -W -Wall -Wstrict-prototypes -Wmissing-prototypes -Wshadow -Werror -g -O2  -release `cat ../bfd/libtool-soversion`   -o libopcodes.la -rpath /tmp/vsc40023/rpathtest/software/binutils/2.26/lib dis-buf.lo disassemble.lo dis-init.lo i386-dis.lo i386-opc.lo -Wl,/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../bfd/.libs/libbfd.so -L/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../libiberty/pic -liberty -Wl,-lc,--as-needed,-lm,--no-as-needed -Wl,-rpath=/usr/lib -Wl,-rpath=/lib -Wl,-rpath=/usr/lib64 -Wl,-rpath=/lib64                                  
```

relevant entries in `rpath_wrapper_gcc.log`:

```
(100154) [2016-10-14 13:18:11] found CMD: gcc | ORIG_CMD: /usr/bin/gcc | orig args: '-shared -fPIC -DPIC .libs/dis-buf.o .libs/disassemble.o .libs/dis-init.o .libs/i386-dis.o .libs/i386-opc.o -L/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../libiberty/pic -liberty -Wl,/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../bfd/.libs/libbfd.so -Wl,-lc -Wl,--as-needed -Wl,-lm -Wl,--no-as-needed -Wl,-rpath=/usr/lib -Wl,-rpath=/lib -Wl,-rpath=/usr/lib64 -Wl,-rpath=/lib64 -Wl,-soname -Wl,libopcodes-2.26.20160125.so -o .libs/libopcodes-2.26.20160125.so'
(100154) [2016-10-14 13:18:11] RPATH: '-Wl,-rpath=/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/libiberty/pic', CMD_ARGS: '-shared -fPIC -DPIC .libs/dis-buf.o .libs/disassemble.o .libs/dis-init.o .libs/i386-dis.o .libs/i386-opc.o -L/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../libiberty/pic -liberty -Wl,/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../bfd/.libs/libbfd.so -Wl,-lc -Wl,--as-needed -Wl,-lm -Wl,--no-as-needed -Wl,-rpath=/usr/lib -Wl,-rpath=/lib -Wl,-rpath=/usr/lib64 -Wl,-rpath=/lib64 -Wl,-soname -Wl,libopcodes-2.26.20160125.so -o .libs/libopcodes-2.26.20160125.so'
(100154) [2016-10-14 13:18:11] running '/usr/bin/gcc -Wl,-rpath=/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/libiberty/pic -shared -fPIC -DPIC .libs/dis-buf.o .libs/disassemble.o .libs/dis-init.o .libs/i386-dis.o .libs/i386-opc.o -L/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../libiberty/pic -liberty -Wl,/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../bfd/.libs/libbfd.so -Wl,-lc -Wl,--as-needed -Wl,-lm -Wl,--no-as-needed -Wl,-rpath=/usr/lib -Wl,-rpath=/lib -Wl,-rpath=/usr/lib64 -Wl,-rpath=/lib64 -Wl,-soname -Wl,libopcodes-2.26.20160125.so -o .libs/libopcodes-2.26.20160125.so'
```

relevant entries in `rpath_wrapper_ld.log`:

```
(100163) [2016-10-14 13:18:11] found CMD: ld | ORIG_CMD: /usr/bin/ld | orig args: '--eh-frame-hdr --build-id -m elf_x86_64 --hash-style=gnu -shared -o .libs/libopcodes-2.26.20160125.so /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbeginS.o -L/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../libiberty/pic -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/tmp/vsc40023/rpathtest/software/zlib/1.2.8/lib -L/tmp/vsc40023/rpathtest/software/Bison/3.0.4/lib -L/tmp/vsc40023/rpathtest/software/flex/2.6.0/lib -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../.. -rpath=/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/libiberty/pic .libs/dis-buf.o .libs/disassemble.o .libs/dis-init.o .libs/i386-dis.o .libs/i386-opc.o -liberty /dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../bfd/.libs/libbfd.so -lc --as-needed -lm --no-as-needed -rpath=/usr/lib -rpath=/lib -rpath=/usr/lib64 -rpath=/lib64 -soname libopcodes-2.26.20160125.so -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtendS.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o'
(100163) [2016-10-14 13:18:11] RPATH: '-rpath=/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/libiberty/pic:/lib64:/tmp/vsc40023/rpathtest/software/Bison/3.0.4/lib:/tmp/vsc40023/rpathtest/software/flex/2.6.0/lib:/tmp/vsc40023/rpathtest/software/zlib/1.2.8/lib:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64', CMD_ARGS: '--eh-frame-hdr --build-id -m elf_x86_64 --hash-style=gnu -shared -o .libs/libopcodes-2.26.20160125.so /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbeginS.o -L/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../libiberty/pic -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/tmp/vsc40023/rpathtest/software/zlib/1.2.8/lib -L/tmp/vsc40023/rpathtest/software/Bison/3.0.4/lib -L/tmp/vsc40023/rpathtest/software/flex/2.6.0/lib -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../.. -rpath=/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/libiberty/pic .libs/dis-buf.o .libs/disassemble.o .libs/dis-init.o .libs/i386-dis.o .libs/i386-opc.o -liberty /dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../bfd/.libs/libbfd.so -lc --as-needed -lm --no-as-needed -rpath=/usr/lib -rpath=/lib -rpath=/usr/lib64 -rpath=/lib64 -soname libopcodes-2.26.20160125.so -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtendS.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o'
(100163) [2016-10-14 13:18:11] running '/usr/bin/ld -rpath=/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/libiberty/pic:/lib64:/tmp/vsc40023/rpathtest/software/Bison/3.0.4/lib:/tmp/vsc40023/rpathtest/software/flex/2.6.0/lib:/tmp/vsc40023/rpathtest/software/zlib/1.2.8/lib:/usr/lib:/usr/lib/gcc/x86_64-redhat-linux/4.4.4:/usr/lib64 --eh-frame-hdr --build-id -m elf_x86_64 --hash-style=gnu -shared -o .libs/libopcodes-2.26.20160125.so /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbeginS.o -L/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../libiberty/pic -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/tmp/vsc40023/rpathtest/software/zlib/1.2.8/lib -L/tmp/vsc40023/rpathtest/software/Bison/3.0.4/lib -L/tmp/vsc40023/rpathtest/software/flex/2.6.0/lib -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../.. -rpath=/dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/libiberty/pic .libs/dis-buf.o .libs/disassemble.o .libs/dis-init.o .libs/i386-dis.o .libs/i386-opc.o -liberty /dev/shm/vsc40023/binutils/2.26/dummy-/binutils-2.26/opcodes/../bfd/.libs/libbfd.so -lc --as-needed -lm --no-as-needed -rpath=/usr/lib -rpath=/lib -rpath=/usr/lib64 -rpath=/lib64 -soname libopcodes-2.26.20160125.so -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtendS.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o'
```

ocaisa:
Yes, but this can be a relative path

boegel:
@geimer that's... urgh. :)

That means you need to inject both the `/lib` and `/lib64` install directories of the thing you're installing to `-rpath` for every compiler/linker command being executed...

Seems like Spack does it too: https://github.com/LLNL/spack/blob/develop/lib/spack/env/cc#L305

geimer:
@boegel Nobody claimed that rpath'ing was fun :)

ocaisa:
I was wrong, you should use `$ORIGIN` not a relative path
http://stackoverflow.com/questions/24573732/difference-between-relative-path-and-using-origin-as-rpath

boegel:
@ocaisa ah, that's useful! It's better explained here: http://stackoverflow.com/a/24740945/1467322

So, basically always rpathing in `'$ORIGIN/../lib'` and `'$ORIGIN/../lib64'` should do the trick.

boegel:
... or maybe not, see http://stackoverflow.com/questions/11316016/gnu-ld-does-not-handle-origin-processing-correctly-is-there-a-workaround

boegel:
No, wait, `$ORIGIN` *can* be used to specify where `ld.so` (the loader) should look for libraries, but the *linker* (`ld`) doesn't expand it (how could it, it wouldn't know what to expand it to).

I'll give it a shot with rpath'ing in `'$ORIGIN/../lib'`, see if that fixes the issue with `binutils`.

boegel:
issue with `binutils` is indeed fixed via rpath'ing in`$ORIGIN/../lib`

on to the next problem, now when building `GCC/5.4.0`

```
g++ -c   -g -O2 -DIN_GCC    -fno-exceptions -fno-rtti -fasynchronous-unwind-tables -W -Wall -Wwrite-strings -Wcast-qual -Wmissing-format-attribute -Woverloaded-virtual -pedantic -Wno-long-long -Wno-variadic-macros -Wno-overlength-strings   -DHAVE_CONFIG_H -DGENERATOR_FILE -I. -Ibuild -I../../gcc -I../../gcc/build -I../../gcc/../include  -I../../gcc/../libcpp/include  \
        -DBASEVER="\"5.4.0\"" -DDATESTAMP="\"\"" \
        -DREVISION="\"\"" \
        -DDEVPHASE="\"\"" -DPKGVERSION="\"(GCC) \"" \
        -DBUGURL="\"<http://gcc.gnu.org/bugs.html>\"" -o build/version.o ../../gcc/version.c
gawk -f ../../gcc/opt-gather.awk ../../gcc/ada/gcc-interface/lang.opt ../../gcc/fortran/lang.opt ../../gcc/go/lang.opt ../../gcc/java/lang.opt ../../gcc/lto/lang.opt ../../gcc/c-family/c.opt ../../gcc/common.opt ../../gcc/config/fused-madd.opt ../../gcc/config/i386/i386.opt ../../gcc/config/gnu-user.opt ../../gcc/config/linux.opt ../../gcc/config/linux-android.opt > tmp-optionlist
g++: ": No such file or directory
<command-line>: warning: missing terminating " character
../../gcc/version.c:35: error: missing terminating " character
../../gcc/version.c:35: error: expected primary-expression before ‘;’ token
make[2]: *** [build/version.o] Error 1
```

I believe this is where @geimer said, "I told you so!"...

ocaisa:
For ld to process `$ORIGIN` I saw this recent post
http://stackoverflow.com/questions/6324131/rpath-origin-not-having-desired-effect
If we go that route then we probably need that extra flag

ocaisa:
Also, not sure if we should worry about this but `xild` exists as well in the Intel toolchains and takes over if `-ipo` is used:
https://software.intel.com/en-us/forums/intel-c-compiler/topic/287357

boegel:
@ocaisa I'm guessing passing `-z origin` as well is only required for very old versions of binutils to make sure that using `$ORIGIN` works, otherwise we would have noticed already? Shouldn't hurt to always pass it though, I guess...

W.r.t. `xild`: so, we should also wrap `xild` if we're using Intel compilers, just to make sure?

boegel:
I tested this a bit more, trying to install `HPL` with `foss/2014b`, `foss/2015a`, `foss/2015b`, `foss/2016a` and `foss/2016b` while having `--rpath` enabled. Here's what I found:

* building `GCC` 4.8.3 and 4.9.2 (with system binutils) works fine with `--rpath` enabled, and the RPATH sanity check passes
* building OpenMPI 1.8.1 and 1.84 with these GCC 4.8.3 and 4.9.2 (resp.) fails with a cryptic error:
```
  CXXLD    libmpi_cxx.la
libtool: link: unsupported hardcode properties
libtool: link: See the libtool documentation for more information.
libtool: link: Fatal configuration error.
make[2]: *** [libmpi_cxx.la] Error 1
``` 
* building `GCC` 4.9.3 (with binutils 2.25) works with `--rpath` enabled, but the sanity check fails on checking for `RPATH` in the output `readelf -f .../bin/g++` (there *is* a `RUNPATH` entry though)
  * same applies to `GCCcore` 4.9.3 (on top of `binutils` 2.25)
  * the problem is basically that the compiler/linker wrapper we put in place has no effect during stage2/stage3 of the bootstrap procedure (thanks @tgamblin for the pointer on this)
  * we should be able to fix this using `--with-boot-ldflags`, see https://gcc.gnu.org/ml/gcc/2011-06/msg00003.html (just like Spack does, see https://github.com/LLNL/spack/blob/develop/var/spack/repos/builtin/packages/gcc/package.py#L94), but it will require an update to the GCC easyblock
* building `GCC` 5.1, 5.2, 5.3, 5.4, 6.1, 6.2 fails all with the same problem, i.e.:
  ```/usr/bin/install: cannot stat `.libs/libcc1.so.0.0.0': No such file or directory```
  * a full log file exhibiting this issue is available at https://gist.github.com/boegel/263ca62858b4785de533a6b50929e0e0
  * from what I can tell, the `libtool` command that is supposed to produce `libcc1.so.0.0.0` is simply being *skipped* for some reason...

On top of these issues, I actually want to revert using `eval "$ORIG_CMD $RPATH $CMD_ARGS"` in the wrapper script; I don't like that approach, and it causes headaches w.r.t. bash expansion.

So, there's clearly quite a bit more work to do to get this stable... I'll try and spend some more time on this, but we're probably looking towards including the support for `--rpath` under `--experimental` (which is better than holding it back entirely, imho).

boegel:
OK, it seems like I've nailed down the problem that broke the GCC builds: the wrapper broke `g++ -v` because it *always* inserted `-Wl,-rpath,$ORIGIN/../lib`, leading to this:

```
$ /tmp/eb-X5Eb4r/tmp0UV637/rpath_wrappers/g++ -v
...
gcc version 4.4.7 20120313 (Red Hat 4.4.7-17) (GCC) 
...
/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
collect2: ld returned 1 exit status
```

boegel:
Another update: made some great progress with this over the last couple of days!

Things that work now that were broken before:

* building `GCC` with `--rpath` enabled: 4.9.3, 5.4.0, 6.2.0 and several versions in between works fine now (incl. the RPATH sanity check), both on top of the system `binutils` and the one provided via EasyBuild
* building OpenMPI with `--rpath` works fine too (incl. RPATH sanity check)

Even better: building HPL on top of `foss/201[456][ab]` + `foss/2016.09` (and everything underneath) now works too using `--rpath` for everything.

The implementation has changed a bit too now, the `eval` I mentioned before is no longer used in the bash wrapper script (which is now fleshed out as an actual script rather than being embedded in `toolchain.py`), and `rpath_args.py` now tries not to reorder the `-rpath` vs `-L` flags being passed to the compiler (which is important to not break stuff).

Some initial checks on the overhead incurred of wrapper very compiler & linker command with a bash script that calls out to a Python script shows about 10-15% slowdown, which I consider acceptable for keeping things as is (i.e. not implementing everything that is now done in `rpath_args.py` directly in the bash wrapper, which would make things a whole lot uglier and harder to maintain).

Currently testing this further by trying to build the whole stack for `OpenFOAM` v3.0.0 with `foss/2016b`... It seems like the build of `CMake` is broken somehow under `--rpath`, I'll dive in deeper and try and figure out what the issue is now.

Other things to test: building `HPL` with `intel/*` from scratch. I expect the RPATH sanity check to barf when doing a binary install (since nothing is being compiled/linked with the wrappers that are put in place), I'll need to find a way to figure out whether or not binaries/libraries in the install dir should pass the RPATH sanity check...

TL;DR: getting very close to making this good-to-go under `--experimental`, so it can be included in EasyBuild v3.0.0 and testing more thoroughly by others!

boegel:
OK, time for one more update...

With the current state combined with https://github.com/hpcugent/easybuild-easyblocks/pull/1031 (and https://github.com/hpcugent/easybuild-easyblocks/pull/1030, but I think that's mostly irrelevant), `eb --rpath` can build the whole `foss/2016a` stack + all dependencies for `OpenFOAM` 3.0.0 (i.e. CMake, Boost, METIS, ...) without issues! \o/
The corresponding list of modules:

```
M4/1.4.17
Autoconf/2.69
Automake/1.15
libtool/2.4.6
flex/2.5.39
Autotools/20150215
Bison/3.0.4
zlib/1.2.8
binutils/2.25
GCCcore/4.9.3
M4/1.4.17-GCCcore-4.9.3
flex/2.5.39-GCCcore-4.9.3
Bison/3.0.4-GCCcore-4.9.3
zlib/1.2.8-GCCcore-4.9.3
binutils/2.25-GCCcore-4.9.3
GCC/4.9.3-2.25
OpenBLAS/0.2.15-GCC-4.9.3-2.25-LAPACK-3.6.0
numactl/2.0.11-GCC-4.9.3-2.25
hwloc/1.11.2-GCC-4.9.3-2.25
OpenMPI/1.10.2-GCC-4.9.3-2.25
gompi/2016a
FFTW/3.3.4-gompi-2016a
ScaLAPACK/2.0.2-gompi-2016a-OpenBLAS-0.2.15-LAPACK-3.6.0
foss/2016a
bzip2/1.0.6-foss-2016a
flex/2.5.39-foss-2016a
zlib/1.2.8-foss-2016a
M4/1.4.17-foss-2016a
ncurses/6.0-foss-2016a
SCOTCH/6.0.4-foss-2016a
CMake/3.4.3-foss-2016a
Boost/1.60.0-foss-2016a
Bison/3.0.4-foss-2016a
libreadline/6.3-foss-2016a
METIS/5.1.0-foss-2016a-32bitIDX
```

The OpenFOAM build itself is still running...

So, unless further testing reveals any serious problems, this is the version that is going to go in under `--experimental` in EasyBuild v3.0.

@wpoely86 do you mind giving this implementation a code review?

Once this is merged, I'll open up an issue to track any problems that pop up with the `--rpath` support that is included in EasyBuild v3.0.

easybuild/framework/easyblock.py
```diff
@@ -1813,6 +1814,76 @@ def sanity_check_step(self, *args, **kwargs):
...
+        for dirpath in dirpaths:
+            if os.path.exists(dirpath):
+                self.log.debug("Sanity checking RPATH for files in %s", dirpath)
+
+                for path in [os.path.join(dirpath, x) for x in os.listdir(dirpath)]:
+                    self.log.debug("Sanity checking RPATH for %s", path)
+
+                    out, ec = run_cmd("file %s" % path, simple=False)
+                    if ec:
+                        fails.append("Failed to run 'find %s': %s" % (path, out))
```
wpoely86: find -> file

easybuild/scripts/rpath_args.py
```diff
@@ -0,0 +1,119 @@
...
+if cmd in ['ld', 'ld.gold']:
+    flag_prefix = ''
+else:
+    flag_prefix = '-Wl,'
+
+if rpath_filter:
+    rpath_filter = rpath_filter.split(',')
+else:
+    # don't RPATH system library locations by default
+    rpath_filter = ['/lib', '/usr/lib']
```
wpoely86: lib64?

easybuild/scripts/rpath_args.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # if command is run in 'version check' mode, make sure we don't include *any* -rpath arguments
+    if arg in ['-v', '-V', '--version', '-dumpversion']:
+        version_mode = True
+        cmd_args.append(arg)
+
+    # FIXME: also consider $LIBRARY_PATH?
+    # FIXME: support to hard inject additional library paths?
+    # FIXME: support to specify list of path prefixes that should not be RPATH'ed into account?
+    # FIXME skip paths in /tmp, build dir, etc.?
```
wpoely86: yeah, this must be done. rpmbuild, CMake etc all do this.

easybuild/scripts/rpath_args.py
```diff
@@ -0,0 +1,119 @@
...
+cmd_args, cmd_args_rpath = [], []
+
+# process list of original command line arguments
+idx = 0
+while idx < len(args):
+
+    arg = args[idx]
+
+    # if command is run in 'version check' mode, make sure we don't include *any* -rpath arguments
+    if arg in ['-v', '-V', '--version', '-dumpversion']:
```
wpoely86: it's an internal script, just stick to one.

easybuild/scripts/rpath_args.py
```diff
@@ -0,0 +1,119 @@
...
+        flag_prefix + '-rpath=$ORIGIN/../lib64',
+        # try to make sure that RUNPATH is not used by always injecting --disable-new-dtags
+        flag_prefix + '--disable-new-dtags',
+    ] + cmd_args
+
+# wrap all arguments into single quotes to avoid further bash expansion
+cmd_args = ["'%s'" % arg.replace("'", "''") for arg in cmd_args]
+
+# output: statement to define $CMD_ARGS and $RPATH_ARGS
+print "CMD_ARGS=(%s)" % ' '.join(cmd_args)
```
wpoely86: print()?

easybuild/scripts/rpath_args.py
```diff
@@ -0,0 +1,119 @@
...
+    # this would result in copying rpath to runpath, meaning that $LD_LIBRARY_PATH is taken into account again
+    elif arg != '--enable-new-dtags':
+        cmd_args.append(arg)
+
+    idx += 1
+
+# add -rpath flags in front
+cmd_args = cmd_args_rpath + cmd_args
+
+if not version_mode:
```
wpoely86: what's the meaning of this version mode anyway?

easybuild/scripts/rpath_wrapper_template.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+set -e
+
+# logging function
+function log {
+    echo "($$) [$(date "+%%Y-%%m-%%d %%H:%%M:%%S")] $1" >> %(rpath_wrapper_log)s
```
wpoely86: explain why the %% are here.

easybuild/scripts/rpath_wrapper_template.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
wpoely86: why both a bash and python wrapper?

easybuild/tools/filetools.py
```diff
@@ -442,6 +442,30 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False, filen
     return var_defs, hits
 
 
+def find_eb_script(script_name):
```
wpoely86: lets cache this?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -591,29 +596,43 @@ def _verify_toolchain(self):
...
+                cmd_path = which(cmd)
+                self.log.debug("which(%s): %s -> %s", cmd, cmd_path, os.path.realpath(cmd_path))
+
+            self.log.info("Commands symlinked to %s via %s: %s", path, symlink_dir, ', '.join(cmds))
+
+    def compilers(self):
+        """Return list of relevant compilers for this toolchain"""
+
+        if self.name == DUMMY_TOOLCHAIN_NAME:
+            c_comps = ['gcc', 'g++']
```
wpoely86: cc, c++?

easybuild/framework/easyblock.py
boegel: nice catch, thanks

easybuild/framework/easyblock.py
```diff
@@ -1797,6 +1798,76 @@ def sanity_check_step(self, *args, **kwargs):
...
+
+                    # only run ldd/readelf on dynamically linked executables/libraries
+                    # example output:
+                    # ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), ...
+                    # ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped
+                    if "dynamically linked" in out:
+                        # check whether all required libraries are found via 'ldd'
+                        out, ec = run_cmd("ldd %s" % path, simple=False)
```
boegel: Sure, but at least the library is either found via `RPATH` or in a default system location.

This is the best we can do in terms of a sanity check for RPATH imho, there's no guarantee here indeed...

I'm open to suggestions on how to check this more strictly (while keeping it relatively lightweight).

easybuild/scripts/rpath_args.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
```
boegel: That doesn't work, you can only pass a single argument to a shebang.

And anyway, this doesn't matter much, we call the script with `python rpath_args.py ...` anyway (i.e. with the `python` that is used to run `eb`, i.e. `sys.executable`).

easybuild/scripts/rpath_args.py
boegel: these are prefixes (checked with `foo.startswith`), so we're covering both `lib` and `lib64` with both of these...

Alternatively, I can make these full regexes, so you'll need to specify `/lib*` instead, and we'll transform it to `^/lib*$` internally before using `re.match`.

Maybe that makes it more clear?

easybuild/scripts/rpath_args.py
boegel: No, we need to cover all possible `gcc` arguments for which we should not inject `-rpath` flags.

`gcc -v` should actually be called as `gcc -v`, not `gcc -Wl,-rpath=$ORIGIN/..lib ... -v`, because is breaks the command...

easybuild/scripts/rpath_args.py
boegel: see above, commands like `gcc -v` should not get broken by injecting `-rpath` flags

easybuild/scripts/rpath_args.py
boegel: Doesn't matter much since most of the rest of EB isn't Py3 compatible yet, but I guess it doesn't hurt looking forward, I'll change this, thanks.

easybuild/scripts/rpath_wrapper_template.sh
boegel: because iterating of the command line arguments in bash is damn nasty to implement imho, it can be done way cleaner in a Python script (which is easier to unit test too).

easybuild/scripts/rpath_wrapper_template.sh
boegel: OK, will do.

easybuild/tools/filetools.py
boegel: Pfft, I don't think that's worth the effort. `find_eb_script` is called once for the bash wrapper script, once for `rpath_args.py` per installation (by `toolchain.prepare()`). Caching isn't going to make a significant difference here (even when multiple installations are being done in one session).

easybuild/scripts/rpath_wrapper_template.sh
boegel: @rjeschmi I like your suggestion to name it `.sh.in`, I'll apply that change.

boegel:
@wpoely86 I consider this good to go now. This is certainly not 100% finished, but that's exactly why we've pushed this under `--experimental`.

It will allow people to easily provide feedback on what we have in place now though, since it will be included in the upcoming EasyBuild v3.0.

Please submit an approved review, so this can be merged?

easybuild/framework/easyblock.py
wpoely86: I would at the minimum print a warning about linkage to non EB paths. We can whitelist all glibc things but any other library needs to be looked after. It's not necessarily an error, but it needs to be checked.

easybuild/scripts/rpath_args.py
wpoely86: `env` passes the arguments. `env python -E` will just work.

easybuild/scripts/rpath_args.py
wpoely86: regex are slower, keep the startswith but add a comment. And I would use `/usr`.

easybuild/scripts/rpath_args.py
wpoely86: okay, and this is the full list? `-E` stuff still works?

easybuild/scripts/rpath_args.py
wpoely86: I don't agree with that. This can give very weird bugs. It's not that hard to add? And `/tmp` should be skipped: you have no guarantees that those files will stay there.

easybuild/scripts/rpath_wrapper_template.sh
wpoely86: so again, why both a bash and python wrapper? Use a purely python one? Forks are costly.

easybuild/tools/toolchain/toolchain.py
wpoely86: Also fine by me. Some projects use `cc`/`c++` before `gcc`/`g++`

easybuild/framework/easyblock.py
boegel: OK, but I won't block this PR for that, we can do that as a follow-up, since it'll take a while before we get the white-listed right probably.

Also, that's sort of orthogonal to `--rpath` anyway, we should always be doing that to catch unintended linking to system libraries (but then we also need to take into account `--filter-deps`, which further complicates things).

I opened an issue for this, see #1991

easybuild/scripts/rpath_args.py
boegel: Hmm, ok, seems like it does work.

But still, this is irrelevant, since we call this script with `python rpath_args.py`, bypassing the shebang entirely...

easybuild/scripts/rpath_args.py
boegel: Bugs how?
If you inject RPATH locations that are no longer there later, they will just be skipped.

I see your concern though, but again, we can't skip paths in `/tmp` blindly, since that would break the case in which someone is installing something with `/tmp/...` as an install path (which is often done during testing).

Things we can do is filter out any location in the build directory `self.builddir`, and in the temporary directory used by EasyBuild (`$TMPDIR`), since we *know* those will be removed once the installation is complete (well, except when we're doing build-in-installdir...).

easybuild/scripts/rpath_wrapper_template.sh
boegel: I'm not sure that calling the original compiler/linker command is going to work when called from within a Python shell, I suspect that may lead to problems.

Total overhead of this bash+Python wrapper approach is 10-15%, which is very acceptable imho, you'll barely notice.

easybuild/scripts/rpath_args.py
boegel: Are you sure they're slower? I'm compiling the regexes before using `.match` on it.
I could even take it one step further and compose on single compiled regex using all supplied entries in `rpath_filter`. I would expect that to be significantly faster than looping over a list and using `startswith`.

Also, using a list of regex's provides more flexibility, since you're not limited to a prefix match, that may come in useful in some situations.

easybuild/scripts/rpath_args.py
boegel: I suggest we follow this up later via #1992, I really want to get this PR merged so we can proceed with smaller enhancements on top.

easybuild/tools/toolchain/toolchain.py
boegel: see #1993

boegel:
@wpoely86 now updated to included build dir and tmpdir in RPATH filter (passed down from `prepare_step` to `toolchain.prepare`)

easybuild/framework/easyblock.py
```diff
@@ -1584,8 +1588,15 @@ def prepare_step(self, start_dir=True):
         # clean environment, undefine any unwanted environment variables that may be harmful
         self.cfg['unwanted_env_vars'] = env.unset_env_vars(self.cfg['unwanted_env_vars'])
 
+        # list of paths to include in RPATH filter;
+        # only include builddir if we're not building in installation directory
+        if os.getenv('TMPDIR'):
```
wpoely86: shouldn't we better use `tempfile.tempdir`?

boegel: I think you mean `tempfile.gettempdir()`, but sure, good point.

boegel:
Thanks for the thorough review @wpoely86 and @rjeschmi!

EricDeveaud:
Hello, 

maybee it will looks a really dumb comment, but why don't include LIBRARY_PATH and LD_RUN_PATH export in modulefiles 

we use this kind of approach on our current module setup. 
egg boost C modulefile defines the following

those environment varaiables are natively used by the linker. 

'''
cixi:~ > module show boost/1_61_0
cixi:~ > show boost
/local/gensoft2/devmodules/boost/1_61_0:

module-whatis    Set environnement for boost (1_61_0) 
module-whatis    topic_3316 
prepend-path     LIBRARY_PATH /local/gensoft2/lib/boost/1_61_0/lib 
prepend-path     LD_RUN_PATH /local/gensoft2/lib/boost/1_61_0/lib 
prepend-path     -d   LDFLAGS -L/local/gensoft2/lib/boost/1_61_0/lib 
prepend-path     CMAKE_LIBRARY_PATH /local/gensoft2/lib/boost/1_61_0/lib 
prepend-path     C_INCLUDE_PATH /local/gensoft2/lib/boost/1_61_0/include 
prepend-path     CPLUS_INCLUDE_PATH /local/gensoft2/lib/boost/1_61_0/include 
prepend-path     CMAKE_INCLUDE_PATH /local/gensoft2/lib/boost/1_61_0/include 
prepend-path     -d   CPPFLAGS -I/local/gensoft2/lib/boost/1_61_0/include 
setenv           BOOST_ROOT /local/gensoft2/lib/boost/1_61_0 
'''


loading boost and compiling versus this one provide binaries with accurate boost rpath 
eg ufasta, part of masurca-3.1.3 (this one was compiled versus boost-1_61_1 and gcc-4.9.0 

'''
/local/gensoft2/exe/masurca/3.1.3/bin/ufasta:
        linux-vdso.so.1 =>  (0x00007ffc461dc000)
        libboost_regex.so.1.61.0 => /local/gensoft2/lib/boost/1_61_0/lib/libboost_regex.so.1.61.0 (0x00007f50efdef000)
        libbz2.so.1 => /lib64/libbz2.so.1 (0x00000034fb800000)
        libz.so.1 => /lib64/libz.so.1 (0x00000034f6800000)
        librt.so.1 => /lib64/librt.so.1 (0x00000034f6000000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00000034f5800000)
        libstdc++.so.6 => /local/gensoft2/exe/gcc/4.9.0/lib64/libstdc++.so.6 (0x00007f50efacc000)
        libm.so.6 => /lib64/libm.so.6 (0x00000034f6400000)
        libgcc_s.so.1 => /local/gensoft2/exe/gcc/4.9.0/lib64/libgcc_s.so.1 (0x00007f50ef8b5000)
        libc.so.6 => /lib64/libc.so.6 (0x00000034f5400000)
        libicudata.so.42 => /usr/lib64/libicudata.so.42 (0x00007f50ee76a000)
        libicui18n.so.42 => /usr/lib64/libicui18n.so.42 (0x00007f50ee3d3000)
        libicuuc.so.42 => /usr/lib64/libicuuc.so.42 (0x00007f50ee081000)
        /lib64/ld-linux-x86-64.so.2 (0x00000034f5000000)
'''

NB there is a "corner case" using this approach, when some rpath instructions are set in upstream compilation mechanism. as ldd in this case ignores LD_RUN_PATH and LD_LIBRARY_PATH (man ld explain it pretty well)

PR #1943, opened 2016-10-04
add support for filtering out setting/updating particular environment variables from generated modules

boegel:
This is intended to be used together with `--rpath` from #1942, i.e.: `eb --rpath --filter-env-vars=LD_LIBRARY_PATH`.

cc @geimer, @pescobar, @rjeschmi


pescobar:
in theory this is not strictly required isn't it? I thought that if the app is compiled with rpath the LD_LIBRARY_PATH is just ignored. Or am I wrong? I have no experience with rpath so please correct me if I am wrong ;)

boegel:
@pescobar that's right, this is orthogonal to RPATH-style linking

But, since one of the motivations of using RPATH is to limit the size of required changes in the environment (by not setting `$LD_LIBRARY_PATH`, by making more dependencies *build*-only dependencies, etc.), this is sort of related; it'll make a lot of sense to use `--rpath` together with `--filter-env-vars=LD_LIBRARY_PATH`.

boegel:

PR #1948, opened 2016-10-07
fix appending to existing buildstats in FileRepository.add_easyconfig

boegel:
fix for bug reporting on EasyBuild mailing list (https://lists.ugent.be/wws/arc/easybuild/2016-10/msg00014.html)


boegel:

PR #1950, opened 2016-10-08
Handle the case when no compiler-specific optarch flag is defined

geimer:
If for a particular CPU family no generic/optimal architecture compiler flag is specified, EasyBuild currently aborts with the following traceback:

```
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/main.py", line 115, in build_and_install_software
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/framework/easyblock.py", line 2376, in build_and_install_one
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/framework/easyblock.py", line 2292, in run_all_steps
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/framework/easyblock.py", line 2171, in run_step
    step_method(self)()
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/framework/easyblock.py", line 1564, in prepare_step
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/tools/toolchain/toolchain.py", line 639, in prepare
    self.set_variables()
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/tools/toolchain/compiler.py", line 151, in set_variables
    self._set_compiler_flags()
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/tools/toolchain/compiler.py", line 257, in _set_compiler_flags
    self.variables.nextend('OPTFLAGS', optflags[:1] + optarchflags)
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/tools/variables.py", line 538, in _passthrough
    res = actual_function(*args, **kwargs)
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/tools/variables.py", line 313, in nextend
    if not self._str_ok(el):
  File "/opt/eb/software/Core/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_framework-2.9.0-py2.7.egg/easybuild/tools/variables.py", line 347, in _str_ok
    test = x is not None and len(x) > 0
TypeError: object of type 'bool' has no len()
 (at easybuild/main.py:148 in build_and_install_software)
```

The culprit is that `ToolchainOptions.option("optarch")` returns the value of the `optarch` toolchain option (a `bool`) if the lookup of the compiler-specific `optarch` flags fails.  This PR handles this and ignores the `optarch` setting in this particular case.

@boegel Please let me know whether you think this is an appropriate solution to tackle this.  To me it feels more like a hack rather than a proper solution, but I couldn't come up with something better yet.  In case you think it's OK, I also need to fix some unit tests to take this situation into account.  Note that these tests don't fail on x86, as we do have an `optarch` setting there.


geimer:
After staring at the code for some time, I believe that I've found a more appropriate solution to address this issue (which no longer feels like a hack...).

geimer:
Of the toolchain tests, only `test_optarch_generic` is still failing on non-x86, as it hardcodes architecture-specific values to compare with.  Apart from this, this PR is now ready for review.

boegel:
@geimer why not simply try and make sure we have an `optarch` setting available for each architecture?

We should definitely avoid spitting out the ugly traceback and confusing error message, and make it emit a proper error message (incl. a suggestion on how to fix the issue), but I'm not sure whether just going with an empty value by default if we run into something new is a good idea...

This feels like hiding a problem under the carpet, where it may go unnoticed for quite a while...

It should be fairly rare that a new architecture suddenly pops up?

test/framework/toolchain.py
```diff
@@ -278,16 +278,19 @@ def test_misc_flags_unique(self):
                 tc = self.get_toolchain("goalf", version="1.1.0-no-OFED")
                 tc.set_options({opt: enable})
                 tc.prepare()
+                flag = None
                 if opt == 'optarch':
-                    flag = '-%s' % tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[tc.arch]
+                    if tc.arch in tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION:
+                        flag = '-%s' % tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[tc.arch]
```
boegel: at the very least, we should make sure that for the major known architectures (mainly x86-64), a matching value for `optarch` *IS* available...

we're making things too liberal here, this may cause us missing an important bug at some time in the future?

test/framework/toolchain.py
```diff
@@ -303,16 +306,18 @@ def test_override_optarch(self):
                 flag = None
                 if optarch_var is not None:
                     flag = '-%s' % optarch_var
-                else:
-                    # default optarch flag
+
+                # default optarch flag if defined
+                elif tc.arch in tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION:
```
boegel: same here, make sure the major architectures have a matching `optarch` option, that should never disappear

geimer:
@boegel I think the discussion comes down to the question of how you define "robustness".  For me, EasyBuild could be considered robust if "it just works" on any platform.  (Well, at least that it tries to do its best to do the job.  There can be platform-specific issues with the packages themselves, of course.)  And on a known platform, you get the target-specific optimization as a bonus.  If you believe that the target-specific optimizations are the most important feature of EB, then aborting with an error may be appropriate.  But I don't believe that this is the major selling point.

What I am missing is a way to emit a big, fat warning even when not in debugging mode, so that such warnings would be much harder to miss.  AFAIK, `log.warning()` also only shows up in the detailed log, right?

boegel:
@geimer I don't think the host-specific optimisation done by default is the major selling point of EasyBuild, but I find it important enough to favour erroring out over silently falling back to suboptimally building. The error message could be as helpful as possible, and suggest how to work around the issue of not knowing how to build with arch-specific optimizations and/or point to http://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html.

The print a big fat warning, you can use the `print_warning` function, see http://easybuild.readthedocs.io/en/latest/api/easybuild.tools.build_log.html#easybuild.tools.build_log.print_warning.

geimer:
@boegel I have another version of this in the pipe that will raise an `EasyBuildError`.  Stay tuned...

geimer:
@boegel Here is the "hard exit" version of it.  I guess the exception message is expressive enough ;-) However, I'm not yet convinced that the `dummy` toolchain is handled in the best way.

On an unknown architecture, unit tests (e.g., from `test.framework.toolchain`) should fail with the corresponding exception messages if the `COMPILER_OPTIMAL_ARCHITECTURE_OPTION` and/or `COMPILER_GENERIC_OPTION` maps do not include a value for the current architecture. I tested this by commenting out settings in the GCC toolchain definition. But...

There is one test (`test_optarch_generic`) which has architecture-specific result values hard-coded! And it doesn't make much sense to check the generic optimization flags of the Intel compiler on anything other than Intel platforms anyway.  Any comments on this?

geimer:
@boegel Hmm... well... I just noticed that not only `dummy` doesn't define `COMPILER_GENERIC_OPTION`, but also `ibmxl`, `cuda`, and `craype`.  And the latter two also don't define `COMPILER_OPTIMAL_ARCHITECTURE_OPTION`.  So what would be the desired approach here?  Add those variables to the compiler toolchain classes and insert "something" (presumably an empty string) for each known architecture, or ignore this setting (as implemented)?

boegel:
@geimer I'm sorry for not replying to this earlier, I haven't had time to take a close look at this again.

Are you OK with moving this to the `3.1` milestone?

geimer:
@boegel Could you please take another look at this?

geimer:
@boegel Thinking about it again, I'm now feeling unsure whether the `--ignore-unknown-optarch-flags` switch is really needed (it was your suggestion anyhow :wink:).  More or less the same effect can also be achieved by specifying `--optarch=''` or passing some other valid compiler flag.  So maybe just mentioning `--optarch` and pointing to http://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html is sufficient.  What do you think?

geimer:
@boegel After several iterations, I think this is now good to go -- really... :wink: The change in `easybuild/tools/options.py` is kind of unrelated and a leftover of some intermediate step, though it fixes the sorting order of the options.

easybuild/tools/toolchain/compiler.py
```diff
@@ -320,8 +322,12 @@ def _set_optimal_architecture(self, default_optarch=None):
...
             self.options.options_map['optarch'] = optarch
 
-        if 'optarch' in self.options.options_map and self.options.options_map.get('optarch', None) is None:
-            raise EasyBuildError("_set_optimal_architecture: don't know how to set optarch for %s", self.arch)
+        if self.options.options_map.get('optarch', None) is None:
+            raise EasyBuildError("_set_optimal_architecture: don't know how to set%s optarch flags for %s/%s!  Use "
+                                 "'--optarch=<flags>' to override (see "
+                                 "http://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html "
+                                 "for details) and consider contributing your settings back.",
+                                 ('', ' generic')[use_generic], self.arch, self.cpu_family)
```
boegel: this looks a bit messy to me... I feel this is easier to read:

```python
optarch_flags_str = "%soptarch flags" % ('', 'generic ')[use_generic]
error_msg = "Don't know how to set %s for %s/%s! " % (optarch_flags_str, self.arch, self.cpu_family)
error_msg += "Use --optarch='<flags>' to override (see "
error_msg += "http://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html "
error_msg += "for details) and consider contributing your settings back (see "
error_msg += "http://easybuild.readthedocs.io/en/latest/Contributing.html)."
```

boegel:
Thanks @geimer! :)

PR #1953, opened 2016-10-09
Improved ARM platform/CPU detection

geimer:
This PR revises several functions in `framework.tools.systemtools`, especially to improve the platform/CPU detection for ARM-based systems.

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -261,8 +261,13 @@ def get_cpu_family():
         _log.debug("Using vendor as CPU family: %s" % family)
 
     else:
+        arch = get_cpu_architecture()
+        if arch in [AARCH32, AARCH64]:
+            # Custom ARM-based designs from other vendors
+            family = ARM
+
         # POWER family needs to be determined indirectly via 'cpu' in /proc/cpuinfo
-        if os.path.exists(PROC_CPUINFO_FP):
+        elif os.path.exists(PROC_CPUINFO_FP):
             cpuinfo_txt = read_file(PROC_CPUINFO_FP)
             power_regex = re.compile(r"^cpu\s+:\s*POWER.*", re.M)
             if power_regex.search(cpuinfo_txt):
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -96,6 +96,17 @@ VENDOR_IDS = {
     'GenuineIntel': INTEL,
     'IBM': IBM,
 }
+ARM_CORTEX_IDS = {
+    '0xc07': 'Cortex-A7',
+    '0xc08': 'Cortex-A8',
+    '0xc09': 'Cortex-A9',
+    '0xc0e': 'Cortex-A17',
+    '0xc0f': 'Cortex-A15',
+    '0xd03': 'Cortex-A53',
+    '0xd07': 'Cortex-A57',
+    '0xd08': 'Cortex-A72',
+    '0xd09': 'Cortex-A73',
+}
 
 
 class SystemToolsException(Exception):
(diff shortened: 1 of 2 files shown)


easybuild/tools/systemtools.py
```diff
@@ -49,11 +49,24 @@
 _log = fancylogger.getLogger('systemtools', fname=False)
 
 # constants
+AARCH32 = 'AArch32'
+AARCH64 = 'AArch64'
+POWER = 'POWER'
+X86_64 = 'x86_64'
+
 AMD = 'AMD'
```
boegel: make it clear that these are vendors, while the above are architectures (with a comment)

easybuild/tools/systemtools.py
```diff
@@ -64,13 +77,37 @@
 PROC_CPUINFO_FP = '/proc/cpuinfo'
 PROC_MEMINFO_FP = '/proc/meminfo'
 
-CPU_FAMILIES = [ARM, AMD, INTEL, POWER]
-VENDORS = {
-    'ARM': ARM,
+CPU_ARCHITECTURES = [AARCH32, AARCH64, POWER, X86_64]
+CPU_FAMILIES = [AMD, ARM, INTEL, POWER]
+CPU_VENDORS = [AMD, APM, ARM, BROADCOM, CAVIUM, DEC, IBM, INTEL, MARVELL, MOTOROLA, NVIDIA, QUALCOMM, MARVELL]
```
boegel: @geimer `MARVELL` is listed twice, your sorting algorithm is broken ;)

easybuild/tools/systemtools.py
```diff
@@ -64,13 +77,37 @@
...
+VENDOR_IDS = {
```
boegel: any reference for this list?

easybuild/tools/systemtools.py
```diff
@@ -64,13 +77,37 @@
...
+    '0x4E': NVIDIA,
+    '0x50': APM,
+    '0x51': QUALCOMM,
+    '0x56': MARVELL,
+    '0x69': INTEL,
     'AuthenticAMD': AMD,
     'GenuineIntel': INTEL,
     'IBM': IBM,
 }
+ARM_CORTEX_IDS = {
```
boegel: ref?

easybuild/tools/systemtools.py
```diff
@@ -137,34 +174,73 @@ def get_total_memory():
...
+    power_regex = re.compile("ppc64.*")
+
+    system, node, release, version, machine, processor = platform.uname()
+    if aarch32_regex.match(machine):
+        arch = AARCH32
+    elif aarch64_regex.match(machine):
+        arch = AARCH64
+    elif power_regex.match(machine):
+        arch = POWER
+    elif machine == 'x86_64':
```
boegel: rather than sorting alphabetically, I would try and rank these by how common they are, to short-circuit early in the most common case (i.e. `x86_64`)?

easybuild/tools/systemtools.py
```diff
@@ -137,34 +174,73 @@ def get_total_memory():
...
+        arch = X86_64
+    else:
+        arch = UNKNOWN
+        _log.warning("Failed to determine CPU architecture, returning %s", arch)
+        return arch
```
boegel: I would start out with `arch = UNKNOWN`, and avoid using `return` twice as follows:

```python
arch = UNKNOWN
if machine == X86_64:
    ...
...

if arch == UNKNOWN:
    _log.warning("Failed to determine CPU architecture, returning %s", arch)
else:
    _log.debug("Determined CPU architecture: %s", arch)

return arch
```

easybuild/tools/systemtools.py
boegel: also, compare with `X86_64` constant here?

easybuild/tools/systemtools.py
```diff
@@ -137,34 +174,73 @@ def get_total_memory():
...
+    _log.debug("Determined CPU architecture: %s", arch)
+    return arch
+
+
 def get_cpu_vendor():
     """
     Try to detect the CPU vendor
 
-    :return: a value from the VENDORS dict
+    :return: a value from the CPU_VENDORS list
```
boegel: dropping `VENDORS` OK since we're working towards EB v3.0, you're lucky ;)

easybuild/tools/systemtools.py
```diff
@@ -137,34 +174,73 @@ def get_total_memory():
...
+    if os_type == LINUX:
+        vendor_regex = None
+
+        arch = get_cpu_architecture()
+        if arch in [AARCH32, AARCH64]:
+            vendor_regex = re.compile(r"CPU implementer\s+:\s*(\S+)")
+        elif arch == POWER:
+            vendor_regex = re.compile(r"model\s+:\s*(\w+)")
+        elif arch == X86_64:
+            vendor_regex = re.compile(r"vendor_id\s+:\s*(\S+)")
```
boegel: same here, sort by most common?

easybuild/tools/systemtools.py
```diff
@@ -186,8 +262,13 @@ def get_cpu_family():
         _log.debug("Using vendor as CPU family: %s" % family)
 
     else:
+        arch = get_cpu_architecture()
+        if arch in [AARCH32, AARCH64]:
```
boegel:
```python
if get_cpu_architecture() in [AARCH32, AARCH64]:
```

since `arch` is only used in one place?

easybuild/tools/systemtools.py
```diff
@@ -210,15 +291,37 @@ def get_cpu_model():
     os_type = get_os_type()
 
     if os_type == LINUX and os.path.exists(PROC_CPUINFO_FP):
-        # we need 'model name' on Linux/x86, but 'model' is there first with different info
-        # 'model name' is not there for Linux/POWER, but 'model' has the right info
-        model_regex = re.compile(r"^model(?:\s+name)?\s+:\s*(?P<model>.*[A-Za-z].+)\s*$", re.M)
         txt = read_file(PROC_CPUINFO_FP)
```
boegel: move this down too, to where it's actually used (i.e. in the `else`), and maybe rename `txt` to `proc_cpuinfo`?

easybuild/tools/systemtools.py
```diff
@@ -210,15 +291,37 @@ def get_cpu_model():
...
+            # we can reverse-map the part number.
+            vendor = get_cpu_vendor()
+            if vendor == ARM:
+                model_regex = re.compile(r"CPU part\s+:\s*(\S+)", re.M)
+                # There can be big.LITTLE setups with different types of cores!
+                res = model_regex.findall(txt)
+                if res:
+                    model_ids = list(set(res))
+                    id_list = []
+                    for model_id in model_ids:
```
boegel:
```python
model_ids = model_regex.findall(txt)
if model_ids:
    id_list = []
    for model_id in set(model_ids):
```

easybuild/tools/systemtools.py
```diff
@@ -210,15 +291,37 @@ def get_cpu_model():
...
+                        if model_id in ARM_CORTEX_IDS:
+                            id_list.append(ARM_CORTEX_IDS[model_id])
+                        else:
+                            id_list.append(UNKNOWN)
```
boegel: collapse to:

```python
id_list.append(ARM_CORTEX_IDS.get(model_id, UNKNOWN))
```

easybuild/tools/systemtools.py
geimer: The `txt` variable is actually used in both branches of the `if` statement...
But I'm fine with renaming.  Makes sense.

easybuild/tools/systemtools.py
geimer: Will the iteration order be the same for every run?  OK, I may have missed a `sort()` call in my first attempt, but that's a different story...

boegel: No, there's no order in a `set`, and creating a `list` from it isn't going to change that (the order you get in the `list` may in theory be different every time).

If you want to guarantee order over unique elements, use `sorted(set(model_ids))`

geimer:
@boegel Would you please take another look at this?  I almost believe that this PR is a prerequisite for further work on https://github.com/hpcugent/easybuild-framework/issues/1952. Thanks!

boegel:
@geimer: thanks for the bump, I lost track of this, sorry

Looks perfect now, so going in, thanks again for all your effort on this one!

PR #1954, opened 2016-10-10
use '0' as letter dir for funky software names that don't start with a letter

boegel:
I noticed this when looking into an easyconfig file to `3to2` (required for https://github.com/hpcugent/easybuild-easyconfigs/pull/3655)...

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1285,6 +1285,19 @@ def process_easyconfig(path, build_specs=None, validate=True, parse_only=False, 
     return easyconfigs
 
 
+def letter_dir_for(name):
+    """
+    Determine 'letter' directory for specified software name.
+    This usually just the 1st letter of the software name (in lowercase),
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1285,6 +1285,19 @@ def process_easyconfig(path, build_specs=None, validate=True, parse_only=False,
     return easyconfigs
 
 
+def letter_dir_for(name):
+    """
+    Determine 'letter' directory for specified software name.
+    This usually just the 1st letter of the software name (in lowercase),
+    except for funky software names, e.g. ones starting with a digit.
```
wpoely86: why not use number instead of something like `4real` or so?

boegel: well, it should be a catch-all, I don't care how it's called exactly...

I can use `_` instead?

wpoely86: `0` is just fine.

boegel:

Issue #1947, opened 2016-10-06
--module-only overwrites buildstats in archived easyconfig


geimer:
As the title already says, regenerating a module file using `--module-only` overwrites the `buildstats` list written to the archived easyconfig.  Usually, this shouldn't be big deal, however, http://easybuild.readthedocs.io/en/latest/Submitting_jobs.html suggests that at least the `build_time` parameter is taken to calculate the maximum walltime when using `--job`.

Now, consider a package such as `GCCcore` - which takes quite some time to build - and regenerate the module file.  The build time stored in the archived easyconfig will now be in the order of a second... and a rebuild of the package is guaranteed to hit the wall-clock limit...

boegel:
@geimer hmm, that's indeed annoying...

So, what *should* it do?

I think we want it to update the archived easyconfig file (since the whole reason of using `--module-only` may be that it got tweaked), but we would also like to have build stats in it... Should those be inherited from an easyconfig that is already in the archive? Or maybe just partially?

geimer:
@boegel I think everything except for the `build_time` (and maybe the `timestamp`?) could be overwritten.  Maybe we could add a comment that a value was inherited from a previous install?

geimer:
@boegel Thinking about it again, maybe the *build*_stats shouldn't change at all, as the build itself didn't change?

boegel:
fixed in #1956, I suggest we follow up there @geimer?

PR #1956, opened 2016-10-12
skip adding of (new) buildstats under --module-only

boegel:
fix for #1947

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -54,7 +54,7 @@ from easybuild.tools import config, filetools
 from easybuild.framework.easyconfig import EASYCONFIGS_PKG_SUBDIR
 from easybuild.framework.easyconfig.easyconfig import ITERATE_OPTIONS, EasyConfig, ActiveMNS
 from easybuild.framework.easyconfig.easyconfig import get_easyblock_class, get_module_path, resolve_template
-from easybuild.framework.easyconfig.parser import fetch_parameters_from_easyconfig
+from easybuild.framework.easyconfig.parser import EasyConfigParser, fetch_parameters_from_easyconfig
 from easybuild.framework.easyconfig.tools import get_paths_for
 from easybuild.framework.easyconfig.templates import TEMPLATE_NAMES_EASYBLOCK_RUN_STEP
 from easybuild.tools.build_details import get_build_stats
@@ -2435,8 +2435,12 @@ def build_and_install_one(ecdict, init_env):
             # collect build stats
             _log.info("Collecting build stats...")
 
-            buildstats = get_build_stats(app, start_time, build_option('command_line'))
-            _log.info("Build stats: %s" % buildstats)
+            if build_option('module_only'):
+                buildstats = None
+                _log.info("Not adding (new) build stats because of --module-only")
+            else:
+                buildstats = get_build_stats(app, start_time, build_option('command_line'))
+                _log.info("Build stats: %s" % buildstats)
 
             if build_option("minimal_toolchains"):
                 # for reproducability we dump out the parsed easyconfig since the contents are affected when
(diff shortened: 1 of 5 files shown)


boegel:
@geimer please review?

easybuild/framework/easyblock.py
```diff
@@ -2435,8 +2435,12 @@ def build_and_install_one(ecdict, init_env):
             # collect build stats
             _log.info("Collecting build stats...")
 
-            buildstats = get_build_stats(app, start_time, build_option('command_line'))
-            _log.info("Build stats: %s" % buildstats)
+            if build_option('module_only'):
```
geimer: @boegel As far as I can remember, I never had the need to use `--skip`.  That is, I first need to look up what it is actually doing ;)  Do you have any particular use case in mind that would cause problems?

easybuild/framework/easyblock.py
```diff
@@ -2453,13 +2457,26 @@ def build_and_install_one(ecdict, init_env):
...
+
+                # grab build stats from archived easyconfig file, if it exists
+                path_in_repo = repo.easyconfig_path_for(repo_spec, app.name, full_ver)
+                if os.path.exists(path_in_repo):
+                    archived_buildstats = EasyConfigParser(path_in_repo).get_config_dict().get('buildstats')
+                    if archived_buildstats:
+                        currentbuildstats.extend(archived_buildstats)
+
+                if app.cfg['buildstats']:
+                    currentbuildstats.extend(app.cfg['buildstats'])
```
geimer: @boegel I haven't tried using this PR yet, but these lines suggest that EB is keeping a list of previous build statistics?  Doesn't this then address the `--skip` issue mentioned above?

easybuild/framework/easyblock.py
boegel: `--skip --force` can be used to install additional extensions added in the easyconfig file, without having the reinstall the whole shebang, cfr. http://easybuild.readthedocs.io/en/latest/Partial_installations.html#partial-installation-skip

easybuild/framework/easyblock.py
boegel: it does, indeed

before, we were already retained the `buildstats` present in the easyconfig file being used for the installation

with this PR, we'll now also retain `buildstats` from the archived easyconfig as well

I need to look into which `buildstats` element is picked up to pick a walltime for `--job`...

easybuild/framework/easyblock.py
geimer: Urgh... In this case you probably want `--job` to use the sum of the `build_time`s of the package and all subsequently installed extensions.

boegel: Yes, but then you need to know which of the build stats was done with `--skip`, and which ones were not.

geimer: ???

The first build generates the `build_stats` entry in the archived easyconfig.  Then you add an extension using `--skip --force`, which will extend `build_stats` with a new list item.  Then you do `--module-only` which simply retains both entries in `build_stats` without touching them.  And then you install another extension with `--skip --force`, which adds a third item to `build_stats`.  In all cases, `--job` takes the sum of the `build_time` entries of all `build_stats` list items and multiply it by two.  So where exactly do you need to know which of the builds was done using `--skip`?  Seems like I'm still missing something...

boegel: What if you do a full forced reinstallation? That'll add another entry to `buildstats`.

geimer: Why?  This is then a completely new build and I would expect it to overwrite `build_stats`.

geimer:
@boegel To summarize, here is what I would expect to happen (w/o knowing the history of this feature):
 - New build: generate build stats
 - Full forced build: overwrite build stats
 - Install extension with `--skip --force`: extend build stats list with new list entry
 - Recreate module with `--module-only`: retain build stats

The wall-clock limit detection for `--job` should iterate over the entries in the build stats and take the sum of all build times.  As far as I understand the code, it currently only considers the last entry.  If someone wants the full history of builds, he can still use a Subversion/git repository for the archived easyconfigs.

Does that make sense?

boegel:
moved this to `3.1`, I'll try and revisit this ASAP after the EasyBuild v3.0 release.

@geimer: feel free to remind me. ;-)

boegel:
bumping to v3.2.0, just don't have the time to finish this now...

Issue #1850, opened 2016-07-22
Listing a build dependency in `hiddendependencies` silently converts it to a real dependency


geimer:
Just noticed this with EasyBuild 2.8.2:  When a package is listed in both `builddependencies` and `hiddendependencies`, it silently gets converted into a regular dependency.  One can use `Bison-3.0.4.eb` as a simple example; this will robot-build `M4-1.4.17.eb` as it is listed in `builddependencies`.  Here is what happens (note that I've configured Lua module syntax):
 * Run `eb Bison-3.0.4.eb -r`.  In this case, everything is fine -- both modules visible and no `load(...)` statements in the `Bison` module file.
 * Run `eb Bison-3.0.4.eb --hide-deps=M4 -r` -- same as above, except that the `M4` module is now hidden.
 * Add `hiddendependencies = [('M4', '1.4.17')]` to the `Bison-3.0.4.eb` file and run `eb Bison-3.0.4.eb -r`. This hides the `M4` module as well, but generates

    ```
    if not isloaded("M4/.1.4.17") then
        load("M4/.1.4.17")
    end
    ```
    in the `Bison` module file.

boegel:
fix is implemented in #1959, let's follow up there

PR #1959, opened 2016-10-12
fix filtering loads for (hidden) build deps from generated module

boegel:
fix for #1850

@geimer please review/verify


geimer:
@boegel Tested with the test case I provided in #1850 (with various combinations of deps listed as normal/build/hidden dependencies).  Looks good to me.

boegel:
Thanks for the verification @geimer!

PR #1960, opened 2016-10-13
add support for 'linkdependencies' easyconfig parameter (WIP)

boegel:
This depends on #1942.

The idea here is that we need to be able to discriminate between dependencies that are only required at build time when `--rpath` is used.

With this in place, we can modify easyconfig files such that they'll work both with and without `--rpath`, by moving dependencies from `dependencies` to `linkdependencies` in case they're only needed at build/link time when `--rpath` is used.

cc @wpoely86, @rjeschmi, @damianam, @geimer

TODO:
* [ ] dedicated test for `linkdependencies`


boegel:
@wpoely86, @rjeschmi, @damianam, @geimer: any feedback on this? Does it make sense to introduce the concept of `linkdependencies`?

boegel:
I briefly discussed this with @jdonners & co earlier today. Their group has been using the (currently still experimental) RPATH support in EasyBuild actively, and feels this is one of the missing features before the RPATH support can be declared stable.

One key aspect that was highlighted is that there's a use case where the user would want that link dependencies *are* being loaded, even tough the software was RPATH'ed, i.e. when other software is built on top of the libraries provided by a particular example. For example: if a netCDF module is loaded to build software on top of it, the user should be able to somehow specify that the corresponding HDF5 module (on top of which netCDF was built) must be loaded too.

It's important that this can be done dynamically, on demand, rather than either excluding the `load` statements for link dependencies, or always loading those link dependencies (since that partially removes some of the advantages provided by using RPATH, e.g. reducing the chance of version conflicts in the required (runtime) dependencies).

This could be done by making `load` statements for link dependencies conditional on a particular environment variable being set, e.g.:

```
if os.getenv("LOAD_LINK_DEPS") ~= nil and not isloaded("HDF5/1.8.19-foss-2017a") then
    load("HDF5/1.8.19-foss-2017a")
end
```

ocaisa:
Shouldn't the decision to apply the additional guard for the load be site specific? Most people probably want to treat a link dep as a dep.

How exactly are we planning to treat this in the repos? Is every site expected to devise themselves what is a link dependency and implement it in a custom repo? Or we planning to add the flag (`linked_dep = True`) to appropriate easyconfigs and have another configuration variable that triggers the guard?

In my opinion the uptake on listing "hidden_dependencies" explicitly in easyconfigs was low because it requires every easyconfig that features the dep to be updated which is tedious and error prone (and since it wasn't done in the main repos it was very time consuming). For us, if we did  start using RPATH, this would probably be most easily implemented in the same way, via a list in an environment variable.

boegel:
@ocaisa Whether or not something is a link dependency is not a site-specific choice, it's a fact. When dependency D is only needed when linking application A (i.e. via `$LD_LIBRARY_PATH` when not using RPATH, or only at installation time when using RPATH), D is a link dependency of A.
Generally speaking, that would be basically all libraries, unless they also provide tools along with it (so `HDF5` should probably not be a link dependency, unless you are sure the tools it provided are not required at runtime).

So, once we have support for indicating which dependencies are `link_dependencies`, we can move things like libraries from `dependencies` to `link_dependencies`. That should be done in the central easyconfigs, since this should not be determined by site policies but by facts.

How these link dependencies are treated can be site-specific though. By default, nothing will change, since unless you're using RPATH there's no difference between regular runtime dependencies and link dependencies: you need to load the modules for both in order to use the software.

When using RPATH, link dependencies should not get loaded by the module, at least by default (imho). Users should be able to trigger the loads though, for example by defining an environment variable like `$LOAD_LINK_DEPS` before loading module, in case they want to build software on top of the provided modules.
This default of whether to modules should load link dependencies or not (without a user specifying desired behaviour) can be made configurable too (in the EasyBuild configuration).

mboisson:
Just 👍 'ing this one ;)

PR #1961, opened 2016-10-13
add support for blas_family() and lapack_family() methods in Toolchain instances

boegel:
This is required to get rid of some deprecated blocks of code in easyblocks like ScaLAPACK which currently use a hardcoded list of BLAS/LAPACK libraries.


boegel:
@wpoely86 please review?

boegel:

PR #1963, opened 2016-10-14
make copy_file dry-run aware

boegel:


boegel:

PR #1969, opened 2016-10-26
fix --list-software by making letter_dir_for aware of '*' wildcard name

boegel:
`--list-software` is broken in `develop` without this, only easyconfig files in the `0` subdir are taken into account.

The tests didn't notice this because the easyconfig files in `test/framework/easyconfigs` are organised differently than the ones in the `easybuild-easyconfigs` repository. I'll look into how much of a PITA it is to reorganise the easyconfigs there to match the directory structure we have in the easybuild-easyconfigs repository.

This was introduced in #1954.


boegel:
@wpoely86 please review?

a PR that reorganises the test easyconfigs to trigger the broken `--list-software` is available at #1970

boegel:

PR #1970, opened 2016-10-27
reorganise test easyconfigs to match structure in easyconfigs repo

boegel:
The fact that `--list-software` is currently broken (fix in #1969) went unnoticed because of the 'flat' organisation of the test easyconfigs, so I reorganised them...

Only the location of test easyconfigs was changed, nothing else.


boegel:
I've verified that this triggers the broken `--list-software` (without #1969) included...

Thanks for the thorough review @wpoely86! ;)

Issue #1952, opened 2016-10-09
Revise ARM platform support


geimer:
While EasyBuild already contains minimal support for ARM platforms, it was most likely only ever tested on a Raspberry Pi.  For other systems, and especially for 64-bit ARMv8 platforms, various enhancements are necessary.

 - [ ] Improve ARM platform/CPU detection in `framework.tools.systemtools`
 - [ ] Define compiler-specific generic/optimal architecture command-line flags (also needs https://github.com/hpcugent/easybuild-framework/pull/1950)
 - [ ] Make framework unit tests pass
 - [ ] Whatever else shows up

boegel:
cc @ericvh, I figured you may be interested in this given your recent work in the OpenHPC repo

boegel:
@geimer OK to move to `3.1` milestone?

geimer:
@boegel Depends pretty much on the release date of `v3.0` ;-)

I haven't found the time yet to address the review comments for #1974, but hope to work on this right after SC. Regarding the framework unit tests, I already know that some of the toolchain tests are failing and need some tweaking. And since my ARM dev board is not the fastest, I haven't run the whole test suite yet, i.e., there may be more failing tests to come...

boegel:
@geimer I plan to release v3.0 in the next couple of days, certainly before the end of SC16, so I'm bumping this to the 3.1 milestone.

Let me know if I can be of any help with this.

PR #1974, opened 2016-10-31
ARM: GCC optimal/generic architecture compiler flags

geimer:
This PR adds support for optimal/generic architecture compiler flags for the GCC compiler.  This required basing the flag selection on the CPU architecture rather than the CPU family.

**AArch32:** The generic option assumes an ARMv7 platform.  Using EasyBuild on older ARM architectures probably isn't of interest anyhow.

**AArch64:** Since the `-mcpu=native` option is only supported since GCC 6, a simple heuristic to target the detected CPU model for vanilla ARM cores has been implemented for older compilers.  This is a "best effort" approach that may fail if the actual compiler does not (yet) support the corresponding CPU model.  However, in this case the user still has the option to manually specify a reasonable option via `--optarch=...`.


easybuild/toolchains/compiler/gcc.py
```diff
@@ -66,15 +69,17 @@ class Gcc(Compiler):
...
+        systemtools.X86_64: 'march=native', # implies -mtune=native
     }
     # used with --optarch=GENERIC
     COMPILER_GENERIC_OPTION = {
-        systemtools.AMD : 'march=x86-64 -mtune=generic',
-        systemtools.INTEL : 'march=x86-64 -mtune=generic',
-        systemtools.POWER: 'mcpu=generic-arch',  # no support for -march on POWER
+        systemtools.AARCH32: 'mcpu=generic-armv7', # implies -march=armv7 and -mtune=generic-armv7
+        systemtools.AARCH64: 'mcpu=generic',       # implies -march=armv8-a and -mtune=generic
+        systemtools.POWER: 'mcpu=generic-arch',    # no support for -march on POWER
```
geimer: @JackPerdue Did you ever test this?  I could not find any indication in the GCC manuals that `generic-arch` is a valid *cpu_type* value for the `-mcpu` option on POWER...

JackPerdue:
To be honest, I haven't looked at Power for a bit, but I
don't think that was me.  e.g.

https://github.com/hpcugent/easybuild-framework/commit/7f44bb66a6172d64a69d001ef1a1be710cbf8f6e

I remember adding -mcpu=native but I don't recall adding
any generic-arch stuff.


Jack Perdue
Lead Systems Administrator
High Performance Research Computing
TAMU Division of Research
j-perdue@tamu.edu    http://hprc.tamu.edu
HPRC Helpdesk: help@hprc.tamu.edu

JackPerdue:
A quick test rebuilding HPL-2.1-foss-2016b.eb with --optarch=GENERIC shows a definite fail.  I'm not used to using the GENERIC flag but, yes, GCC won't work on Power as is.

JackPerdue:
-mcpu=native -mtune=native might be the best way to go for now if the GENERIC needs to be differentiated in some way... but generic-arch is right out (as you suspected)

JackPerdue:
gcc: note: valid arguments to ‘-mtune=’ are: 401 403 405 405fp 440 440fp 464 464fp 476 476fp 505 601 602 603 603e 604 604e 620 630 740 7400 7450 750 801 821 823 8540 8548 860 970 G3 G4 G5 a2 cell e300c2 e300c3 e500mc e500mc64 e5500 e6500 ec603e native power3 power4 power5 power5+ power6 power6x power7 power8 powerpc powerpc64 rs64 titan

I don't have any suggestions beyond those not understanding what "generic" means on Power.

easybuild/toolchains/compiler/gcc.py
boegel: @pescobar: it seems like you added this while implementing support for `--optarch=GENERIC`? cfr. 7f44bb6

Any pointers to where you got this from?

easybuild/toolchains/compiler/clang.py
```diff
@@ -84,14 +84,12 @@ class Clang(Compiler):
 
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
-        systemtools.INTEL : 'march=native',
-        systemtools.AMD : 'march=native',
```
boegel: @geimer it seems unwise to drop these?

easybuild/toolchains/compiler/gcc.py
```diff
@@ -110,3 +115,26 @@ def _set_compiler_vars(self):
...
+                    # Get CPU model and strip off 'ARM ' prefix
+                    cpu_model = systemtools.get_cpu_model()[4:]
+                    # big.LITTLE setups: sort core types to have big core (higher model number) first
+                    core_types = [core.strip().lower() for core in cpu_model.split('+')]
+                    sorted_core_types = sorted([(int(re.search('\d+', i).group(0)), i) for i in core_types],
+                                               reverse=True)
+                    # Construct -mcpu option
+                    option = 'mcpu=%s' % '.'.join([i[1] for i in sorted_core_types])
+
+                self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[systemtools.AARCH64] = option
```
boegel: @geimer don't we want some kind of warning in case `option` is an empty string here?

easybuild/toolchains/compiler/clang.py
boegel: oh, wait, this relates to your change in `compiler.py` where you define `self.arch` differently...

easybuild/toolchains/compiler/gcc.py
```diff
@@ -110,3 +115,26 @@ def _set_compiler_vars(self):
...
+            gcc_version = self.get_software_version(self.COMPILER_MODULE_NAME)[0]
+            if LooseVersion(gcc_version) < LooseVersion('6'):
+                option = ''
+                if systemtools.get_cpu_vendor() == systemtools.ARM:
+                    # Get CPU model and strip off 'ARM ' prefix
+                    cpu_model = systemtools.get_cpu_model()[4:]
+                    # big.LITTLE setups: sort core types to have big core (higher model number) first
+                    core_types = [core.strip().lower() for core in cpu_model.split('+')]
+                    sorted_core_types = sorted([(int(re.search('\d+', i).group(0)), i) for i in core_types],
+                                               reverse=True)
```
boegel: there's lots that could go wrong here, and there's no error handling whatsoever; aren't we making too much assumption here on what `get_cpu_model` returns?

easybuild/toolchains/compiler/gcc.py
```diff
@@ -110,3 +115,26 @@ def _set_compiler_vars(self):
...
+                    # Construct -mcpu option
+                    option = 'mcpu=%s' % '.'.join([i[1] for i in sorted_core_types])
```
boegel: can you provide a couple of examples in a comment of how what value would be specified to `mcpu` with this approach (just in case we get something entirely different at some point)?

easybuild/tools/toolchain/compiler.py
```diff
@@ -146,8 +146,11 @@ def set_options(self, options):
 
     def set_variables(self):
         """Set the variables"""
+        if self.arch is None:
+            self.arch = systemtools.get_cpu_architecture()
```
boegel: hmm, aren't there any other locations where we rely on the `.arch` attribute of a `Toolchain` instance (potentially) that we should worry about?

playing the devil's advocate here, since this approach makes more sense anyway (defining `self.arch` with the result of `get_cpu_family` was already an indication that something wasn't 100% right...)

test/framework/toolchain.py
```diff
@@ -339,6 +353,20 @@ def test_optarch_generic(self):
                         else:
                             self.assertFalse(generic_flags in tc.get_variable(var))
 
+    def test_optarch_aarch64_heuristic(self):
```
boegel: @geimer don't we want to test the default behaviour for `AARCH64` as well here?

you can put a dummy module in place for GCC 6.x to check that

easybuild/toolchains/compiler/inteliccifort.py
```diff
@@ -71,13 +71,11 @@ class IntelIccIfort(Compiler):
 
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
-        systemtools.INTEL : 'xHost',
-        systemtools.AMD : 'xHost',
+        systemtools.X86_64: 'xHost',
```
boegel: @geimer hmm, I'm not so sure we want to drop the distinction between `AMD` and `INTEL` here... right now, it doesn't matter, but it may sometime in the future?

geimer: @boegel Hmm... for ARM, one needs to distinguish by architecture (as both AArch32 and AArch64 are of CPU family `ARM`), for X64_64 you want a distinction by CPU family.  I can see various ways to address this:
 - Use the architecture as dictionary key (like now) and overwrite the value in `_set_optimal_architecture` if needed
 - Do it exactly the other way around (family as key, overwrite on architecture)
 - Use a tuple `(architecture, family)` as dictionary key

If you really want to keep the family distinction, I'm leaning towards the last option.  What do you think?

boegel: Going with the latter makes sense to me!

easybuild/toolchains/compiler/gcc.py
geimer: @boegel To be honest, I copy/paste-engineered this based on some code snippet I found on the web (StackOverflow, IIRC). I have no clue how to add error handling here...

Regarding your concerns about assumptions: This is certainly an issue, yes.  However, I currently see no other way to optimize for the host architecture if `-mcpu=native` is not supported.  If you have any better suggestions, please let me know.

test/framework/toolchain.py
geimer: @boegel This is causing some trouble I don't understand... I've added a module file for GCCcore 6.2.0 and added
```
tc = self.get_toolchain("GCCcore", version="6.2.0")
tc.prepare()
self.assertEqual(...)
```
at the end of this test.  But it is still giving me the modified `mcpu` flag based on the heuristic, even though it is definitely **not** taking the "modify option with heuristic" branch in `_set_optimal_architecture`.  Why don't I get a fresh instance with `COMPILER_OPTIMAL_ARCHITECTURE_OPTION` initialized to the defaults???

boegel: @geimer Hmm, you should get a fresh instance... Do make sure you also call `tc.set_options()` before `tc.prepare()` though, even if you're not setting any custom toolchain options. Maybe that explains it?

easybuild/tools/toolchain/compiler.py
boegel: Not seeing any other use of `.arch` elsewhere is good, but we don't control *all* easyblocks out there, and people have scripts around that hook into the EasyBuild framework, who knows...

That shouldn't hold us back from making this change though, since it makes a whole lot of sense, and if anyone relies on `.arch` somewhere, they probably shouldn't be anyway (they should call the appropriate function from `systemtools` instead).

I see no reason either for lazy init of `self.arch`, so feel free to change that.

boegel:
@geimer Looks like this will have to be moved to the `3.1` milestone, are you OK with that?

test/framework/toolchain.py
geimer: @boegel Nope. Here is my current test (with a `GCCcore/6.2.0` module in place):
```
    def test_optarch_aarch64_heuristic(self):
        """Test whether AArch64 pre-GCC-6 optimal architecture flag heuristic works."""
        st.get_cpu_architecture = lambda: st.AARCH64
        st.get_cpu_model = lambda: 'ARM Cortex-A53'
        st.get_cpu_vendor = lambda: st.ARM
        tc = self.get_toolchain("GCC", version="4.7.2")
        tc.prepare()
        self.assertEqual(tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[tc.arch], 'mcpu=cortex-a53')

        st.get_cpu_model = lambda: 'ARM Cortex-A53 + Cortex-A72'
        tc = self.get_toolchain("GCC", version="4.7.2")
        tc.prepare()
        self.assertEqual(tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[tc.arch], 'mcpu=cortex-a72.cortex-a53')

        tc = self.get_toolchain("GCCcore", version="6.2.0")
        tc.prepare()
        self.assertEqual(tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[tc.arch], 'mcpu=native')
```
I've added
```
    print(self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[systemtools.AARCH64])
```
to `_set_optimal_architecture` in `easybuild-framework/toolchains/compiler/gcc.py` right at the beginning after the check that we're on AArch64 (i.e., before modifying the default value).  And it prints
```
mcpu=native
mcpu=cortex-a53
mcpu=cortex-a72.cortex-a53
```
Calling `tc.set_options({})` right before `tc.prepare()` in the test doesn't change anything.  So it keeps the modified value even when creating a new instance. What am I doing wrong here?  I'm confused...

geimer:
@boegel I still have no clue why this unit test is failing.  I expect to get a fresh `toolchain` instance for each call to `self.get_toolchain` with the values of `COMPILER_OPTIMAL_ARCHITECTURE_OPTION` reset to the defaults.  However, my debugging shows that this is not the case.  Is there some kind of caching going on?

easybuild/toolchains/compiler/gcc.py
boegel: @pescobar ping on the question abov?

test/framework/toolchain.py
boegel: @geimer Taking a look at this right now, I'll keep you posted.

boegel:
@geimer the problem you're running into is fixed with https://github.com/geimer/easybuild-framework/pull/12 (see PR description for details)

JackPerdue:
Again... I don't know who added mcpu=generic-arch for POWER or why and I have no idea what GENERIC is supposed to be but --optarch=GENERIC on Power7 is broken in 3.0.... there is no "generic-arch" option for mcpu (see above).

boegel:
@JackPerdue yeah, the plan is to fix that in this PR as well, since we're fiddling in that area.

Not quite sure what `--optarch=GENERIC` should imply on POWER(7) though...

We could just spit out an error stating this is not supported on POWER.

geimer:
@JackPerdue, @boegel: Looking through the GCC manuals of various versions, these may be reasonable values for POWER:

> -mcpu=powerpc, -mcpu=powerpc64, and -mcpu=powerpc64le specify pure 32-bit PowerPC (either endian), 64-bit big endian PowerPC and 64-bit little endian PowerPC architecture machine types, with an appropriate, generic processor model assumed for scheduling purposes.

As far as I can see, `powerpc64le` is supported since GCC 4.8, the other two were already available in the 3.x series.  However, this means that we have to distinguish between big and little endian POWER (with everything up to POWER7 being big endian and POWER8 being little endian)

test/framework/toolchain.py
```diff
@@ -338,6 +352,30 @@ def test_optarch_generic(self):
...
+        tc = self.get_toolchain("GCC", version="4.7.2")
+        tc.set_options({})
+        tc.prepare()
+        self.assertEqual(tc.options.options_map['optarch'], 'mcpu=cortex-a53')
+        self.assertTrue('-mcpu=cortex-a53' in os.environ['CFLAGS'])
+
+        tc = self.get_toolchain("GCCcore", version="6.2.0")
+        tc.set_options({})
+        tc.prepare()
+        self.assertEqual(tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[tc.arch], 'mcpu=native')
```
geimer: Is it OK to compare with `tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION` here?  In the other cases we compare against `tc.options.options_map['optarch']`.

test/framework/toolchain.py
```diff
@@ -338,6 +352,30 @@ def test_optarch_generic(self):
...
+        tc = self.get_toolchain("GCC", version="4.7.2")
+        tc.set_options({})
+        tc.prepare()
+        self.assertEqual(tc.options.options_map['optarch'], 'mcpu=cortex-a53')
+        self.assertTrue('-mcpu=cortex-a53' in os.environ['CFLAGS'])
+
+        tc = self.get_toolchain("GCCcore", version="6.2.0")
+        tc.set_options({})
+        tc.prepare()
+        self.assertEqual(tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(tc.arch, tc.cpu_family)], 'mcpu=native')
```
geimer: Hmm... One should refresh the page before commenting ;-) So, again:
> Is it OK to compare with tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION here? In the other cases we compare against tc.options.options_map['optarch'].

geimer:
@boegel The failing test seems to be due to an inconsistent behavior of TCL modules. Any suggestions how to deal with this?

geimer:
@boegel I believe this PR is ready for another review.  However, there is still three issues:
 -  One unit test is failing in Travis, but only with the pure TCL modules (looks like an inconsistent behavior).  If you have a suggestion how to fix this, please let me know.
 - The `--optarch=GENERIC` thing on POWER should probably be addressed in a separate PR. See #2064.
 - A couple of unit tests are failing when run natively on ARM due to hard-coded Intel stuff.  I'm up for tackling those, but again probably in a separate PR to keep things manageable.  This is task 3 of #1952.

boegel:
@geimer fix for the broken test in https://github.com/geimer/easybuild-framework/pull/13

easybuild/toolchains/compiler/gcc.py
```diff
@@ -66,15 +69,19 @@ class Gcc(Compiler):
...
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
-        systemtools.AMD : 'march=native',
-        systemtools.INTEL : 'march=native',
-        systemtools.POWER: 'mcpu=native',  # no support for march=native on POWER
+        (systemtools.AARCH32, systemtools.ARM): 'mcpu=native', # implies -march=native and -mtune=native
+        (systemtools.AARCH64, systemtools.ARM): 'mcpu=native', # since GCC 6; implies -march=native and -mtune=native
+        (systemtools.POWER, systemtools.POWER): 'mcpu=native',   # no support for -march on POWER; implies -mtune=native
+        (systemtools.X86_64, systemtools.AMD): 'march=native', # implies -mtune=native
+        (systemtools.X86_64, systemtools.INTEL): 'march=native', # implies -mtune=native
```
boegel: @geimer this makes me wonder, should we also be using `-mcpu=native` on x86_84, since that implies both `-march=native` and `-mtune=native` for AARCH32?

Which one is closer to Intel's `-xHost`, `mcpu=native` or `-march=native`?

test/framework/toolchain.py
boegel: for completeness sake: this problem was fixed in https://github.com/geimer/easybuild-framework/pull/12

easybuild/toolchains/compiler/gcc.py
geimer: @boegel: These kind of options are a mess in GCC, as they are inconsistent between architectures... `-march=native` is the way to go on x86_64.

On x86_64, `-march=native` implies `-mtune=native`, but `-mcpu` is a deprecated synonym for `-mtune`.  So using `-mcpu=native` would only tune for the host, but not enable all CPU features (as `-march` does).

boegel: @geimer That sounds like a mess indeed, thanks for clarifying!

easybuild/toolchains/compiler/gcc.py
```diff
@@ -110,3 +117,43 @@ def _set_compiler_vars(self):
...
+        GCC-specific adjustments for optimal architecture flags.
+
+        :param default_optarch: default value to use for optarch, rather than using default value based on architecture
+                                (--optarch and --optarch=GENERIC still override this value)
+        """
+        if default_optarch is None and self.arch == systemtools.AARCH64:
+            # On AArch64, -mcpu=native is not supported prior to GCC 6.  In this case, try to optimize for the detected
+            # CPU model (vanilla ARM cores only).  Note that this heuristic may fail if the CPU model is not supported
+            # by the GCC version being used.
+            gcc_version = self.get_software_version(self.COMPILER_MODULE_NAME)[0]
```
boegel: @geimer: did you ever test this with a `GCC` module that is actually a bundle of `GCCcore` and `binutils`?

As far as I can tell, grabbing the GCC version will fail in that case, as least with this current implementation (which only takes into account `GCC` as module name)...

easybuild/toolchains/compiler/gcc.py
```diff
@@ -110,3 +117,43 @@ def _set_compiler_vars(self):
...
+            if LooseVersion(gcc_version) < LooseVersion('6'):
+                cpu_vendor = systemtools.get_cpu_vendor()
+                cpu_model = systemtools.get_cpu_model()
+                if cpu_vendor == systemtools.ARM and cpu_model.startswith('ARM '):
+                    self.log.debug("Determining architecture-specific compiler optimization flag for ARM (model: %s)",
```
boegel: @geimer while giving this one last review, this felt really heavy to keep it all under the same method, so I suggest we flesh out the AARCH64 specific stuff into a separate method, see https://github.com/geimer/easybuild-framework/pull/14

easybuild/toolchains/compiler/gcc.py
boegel: Ah, I see why this worked: we usually make sure in the `GCC` bundles that `$EBROOTGCC` and `$EBVERSIONGCC` are also defined for the `GCCcore` module, by defining `altroot`/`altversion` in the `GCC` bundle easyconfig, see https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/g/GCC/GCC-6.2.0-2.27.eb#L21

However, it would be better to take both `GCCcore` (first) and `GCC` (second)` into account for determining the GCC version, like we already do for the installation prefix in `_set_compiler_vars`.

Are you up for making that change @geimer?

geimer: @boegel: I have to admit that I don't remember whether I tested this with a `GCC` bundle.  But...

As far as I can see, `self.COMPILER_MODULE_NAME` corresponds to `GCC` when using the `GCC` toolchain, and to `GCCcore` when using this toolchain. So it should actually do the right thing, no?

boegel: Well, no, since when `GCC` is being used as a toolchain, the compiler is actually provided via `GCCcore`, so we should grab the version directly from there (without relying on a `GCC` bundle` also defining `$EBROOTGCCCORE`...

geimer: @boegel Just to clarify: You are proposing to use `get_software_version` from `easybuild.tools.modules` here rather than `self.get_software_version`, right? I don't think I understand the difference right now (it's been a long day...)

boegel: No, I propose to try obtaining the version using `GCCcore` first, and fall back to use `GCC` if that doesn't work. Whether you use `self.get_software_version` or `get_software_version` doesn't really matter here, that's only relevant when multiple modules are in play (`self.get_software_version` has a way of dealing with that).

So, something like this:

```python
gcc_ver = get_software_version('GCCcore')
if gcc_ver is None:
    gcc_ver = get_software_version('GCC')
```

(switching to `get_software_version` makes sense here since we know we'll never be dealing with multiple names at the same time here)

boegel:
Looks good to go to me.

@geimer: let me know whether or not you want to verify whether Travis would've picked up the failing test you fixed with fc6cad1 before I merge this

geimer:
@boegel If you are OK with knowing that the test result shown above for https://github.com/hpcugent/easybuild-framework/pull/1974/commits/27b53073051e1ee86105fed051e9d07f5fd9db9e is bogus, please go ahead and merge. In the end, it's only the last test result that really matters.

boegel:
@geimer I don't mind at all, I only really care about the test result of the last commit; but it's good to know Travis screws up if you push multiple commits quickly enough.

Going in, thanks!

Issue #1964, opened 2016-10-17
Alternative generic easyblocks are not picked up if they aren't in a generic subfolder


damianam:
The title is kind of self descriptive. If you have a folder where you have all your custom easyblocks, generic easyblocks are not picked up. You have to manually include them with `--include-easyblocks $SOME_PATH/generic/some_easyblock.py`, which is kind of counterintuitive given that all the other easyblocks work just fine.

pescobar:
I have also hitted this one

boegel:
proposed fix in #2432

boegel:
follow-up in #2432

PR #1980, opened 2016-11-08
check that each glob pattern matches at least one file expand_glob_paths (important for --include-*)

boegel:
I was bitten by `--include-easyblocks` silently not including specified easyblocks because I messed up the specified path a couple of times.

This change fixes that, by making `expand_glob_paths`, which is used (only) by the `--include-*` options, stricter to ensure at least one hit per specified glob pattern.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -961,7 +961,11 @@ def expand_glob_paths(glob_paths):
     """Expand specified glob paths to a list of unique non-glob paths to only files."""
     paths = []
     for glob_path in glob_paths:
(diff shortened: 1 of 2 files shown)


boegel:
@damianam I believe you were hit by this too?

@wpoely86 please review?

damianam:
Mmmh, I think my issue was different. I included a generic easyblock as any other easyblock, and that wasn't enough, it needed to be in the generic subdir.

boegel:
@damianam ah, ok, this is about #1964 then

I can see if I can mitigate that issue too, I agree that's pretty confusing as well, especially since EasyBuild should be able to distinguish between generic and software-specific easyblocks itself... 

But I'll do that in a separate PR.

boegel:

PR #1981, opened 2016-11-08
bump page limit in fetch_latest_commit_sha, spit out more meaningful error if too many branches were encountered

boegel:
too many branches may break `fetch_latest_commit_sha`, see e.g. https://travis-ci.org/hpcugent/easybuild-framework/jobs/174153835

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -252,7 +252,7 @@ def fetch_latest_commit_sha(repo, account, branch='master', github_user=None, to
     :return: latest SHA1
     """
     status, data = github_api_get_request(lambda x: x.repos[account][repo].branches,
-                                          github_user=github_user, token=token)
+                                          github_user=github_user, token=token, per_page=GITHUB_MAX_PER_PAGE)
     if not status == HTTP_STATUS_OK:
         raise EasyBuildError("Failed to get latest commit sha for branch %s from %s/%s (status: %d %s)",
                              branch, account, repo, status, data)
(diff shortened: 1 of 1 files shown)


easybuild/tools/github.py
```diff
@@ -264,7 +264,10 @@ def fetch_latest_commit_sha(repo, account, branch='master', github_user=None, to
             break
 
     if res is None:
-        raise EasyBuildError("No branch with name %s found in repo %s/%s (%s)", branch, account, repo, data)
+        error_msg = "No branch with name %s found in repo %s/%s" % (branch, account, repo)
+        if len(data) >= GITHUB_MAX_PER_PAGE:
```
wpoely86: how does this work? Github will only return what you requested?

boegel: well, there shouldn't be more than `GITHUB_MAX_PER_PAGE` hits, indeed, so I could also check with `==`, but it doesn't matter...

boegel: bleh, it's far from trivial to do this, since the header info isn't readily available in the result of `github_api_get_request` (which is used by `fetch_latest_commit_sha`)...

Is it worth the trouble? This is only about a slightly more informative error message...

boegel: OK, but that should be tackled in a separate PR, it shouldn't block this one.

I opened an issue for that, see #1988

boegel:
This is good to go, follow-up in #1988, thanks for the review @wpoely86!

Issue #1983, opened 2016-11-09
TypeError exception while using bootstrap_eb.py


jbd:
Hello,

I'm trying bootstrap_eb.py from develop/e7fed028a529a2f2 as the documentation mentions (curl -O https://raw.githubusercontent.com/hpcugent/easybuild-framework/develop/easybuild/scripts/bootstrap_eb.py) and I've got this error if I'm trying to execute it from a clean docker environment :

```
Traceback (most recent call last):
  File "bootstrap_eb.py", line 953, in <module>
    main()
  File "bootstrap_eb.py", line 756, in main
    stage2(tmpdir, templates, install_path, distribute_egg_dir, sourcepath)
  File "bootstrap_eb.py", line 594, in stage2
    preinstallopts += "%s $(which easy_install) -U --prefix %(installdir)s setuptools && " % sys.executable
TypeError: format requires a mapping
```

The whole output of python bootstrap_eb.py $EASYBUILD_PREFIX is here : https://gist.github.com/jbd/7ddd6a1a665dc2333e9e9d1d073fed94

I've got no problem if I'm using the master (e65a6c9771cb6318) branch (https://raw.githubusercontent.com/hpcugent/easybuild-framework/master/easybuild/scripts/bootstrap_eb.py).

boegel:
can you try changing ` %(installdir)s` to `% %(installdir)s` on the offending line?

jbd:
No more luck, but a different error :

```
python bootstrap_eb.py $EASYBUILD_PREFIX 
[[INFO]] EasyBuild bootstrap script (version 20161104.01, MD5: 769aa7b23f2a66cec5f83072c88d63b5)
[[INFO]] Found Python 2.6.6 (r266:84292, Aug 18 2016, 15:13:37) ; [GCC 4.4.7 20120313 (Red Hat 4.4.7-17)]

[[INFO]] Installation prefix /soft/adm/easybuild
[[INFO]] Found module command '/soft/adm/lmod/lmod/libexec/lmod' via $LMOD_CMD (Lmod), so using it.
[[INFO]] No suitable setuptools installation found, proceeding with stage 0...


[[INFO]] +++ STAGE 0: installing distribute via included (patched) distribute_setup.py...

Downloading http://hpcugent.github.io/easybuild/files/distribute-0.6.49-patched1.tar.gz
Extracting in /tmp/tmpUSuQo1
Now working in /tmp/tmpUSuQo1/distribute-0.6.49
[...]

PR #1946, opened 2016-10-06
enhance bootstrap to auto-skip stage 0 in case a suitable setuptools is already available

boegel:

PR #1984, opened 2016-11-09
fix issues with setuptools provided in step 0 of bootstrap script

boegel:
fix for bugs in bootstrap script reported in #1983, introduced in #1946


boegel:
fixed confirmed in #1983, going in

PR #1985, opened 2016-11-10
change default config to use Lmod/Lua for modules tool/syntax, GC3Pie as job backend

boegel:
This updates the default configuration in preparation of EasyBuild v3.0.

cc @rtmclay, @riccardomurri: thanks for all your help with the support for Lmod and GC3Pie in EasyBuild! :)


boegel:

PR #1986, opened 2016-11-10
fix CUDA-related issues in HierarchicalMNS

boegel:
combination of #1753 and #1982 + relevant part of #1976 + fixing the use of CUDA with `dummy` toolchain

makes #1086 obsolete

I also renamed some variables to try and make the code more readable, and added some clarifying comments.

cc @akesandgren, @thiell


boegel:
This is (finally!) good to go, reviewed via #1753, #1982, #1976.

Thanks for the feedback and help with this @thiell and @akesandgren!

PR #1987, opened 2016-11-10
Simplify GC3Pie version checks.

riccardomurri:
Since GC3Pie migrated to GitHub, we can no longer rely on parsing SVN
revisions.  So:

1) If there is a version number, check that.
2) If GC3Pie is marked as a "development version", trust it to be more recent than 2.4.0.


boegel:
lgtm, thanks @riccardomurri!

Issue #1927, opened 2016-09-21
Suggesting to point towards the --modules-tool solution beyond stating the error


fgeorgatos:
ie. we could say something like:
- Please use `--modules-tool` option or define variable `EASYBUILD_MODULES_TOOL` as appropriate

Example case:

```
[fgeorgatos@demo2 ~]$ eb --from-pr 1410 -Dr
== temporary log file in case of crash /tmp/eb-BIlNA5/easybuild-BLHy3J.log
ERROR: EnvironmentModulesC modules tool can not be used, 'modulecmd' command is not available.

[fgeorgatos@demo2 ~]$ eb -H|grep -i modules-tool=
    --modules-tool=MODULES-TOOL
[fgeorgatos@demo2 ~]$ export EASYBUILD_MODULES_TOOL=Lmod
```

boegel:
implemented this in #1989, thanks for the suggestion @fgeorgatos!

PR #1989, opened 2016-11-10
include suggestion on how to change configuration w.r.t. modules tool/syntax

boegel:
implementation of suggestion made in #1927

Output now looks like this with a wrong configuration:

```
$ eb ...
ERROR: EnvironmentModulesC modules tool can not be used, 'modulecmd' command is not available; use --modules-tool to specify a different modules to use (Lmod, EnvironmentModulesTcl, EnvironmentModulesC)
```

some idea applied to `--module-syntax`:

```
$ eb ...
ERROR: Failed to parse configuration options: 'Generating Lua module files requires Lmod as modules tool; use --module-syntax to specify a different module syntax to use (Lua, Tcl)'
```

```
```

This is very relevant now since we just switched to Lmod/Lua in #1985.

@fgeorgatos please review?

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -253,7 +253,10 @@ class ModulesTool(object):
             self.log.info("Full path for module command is %s, so using it" % self.cmd)
         else:
             mod_tool = self.__class__.__name__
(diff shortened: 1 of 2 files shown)


fgeorgatos:
lgtm! let's merge!

boegel:
Thanks for the feedback @fgeorgatos!

PR #1997, opened 2016-11-15
use https in the "install-EasyBuild-develop.sh" script

pescobar:
The script to install develop was not working for one of my user accounts. It was working fine with another account which I use to work with github

Switching to https works with both accounts


boegel:
Hmm, is this really what you want? This way, you'll need to give a password every time you want to push something to one of your forks?

You should stick to using `git:` imho, and just make sure you forward your SSH private key to the session you're using?

pescobar:
I see your point and I totally agree from a developer point of view

But from a user point of view who only wants to test a feature which is still only available in devel you can do `./install-EasyBuild-develop.sh hpcugent /tmp/eb-devel-installation` and this would only work when using https

boegel:
@pescobar OK, so how about this: try with `git:` first, and fall back to clone over `https:` if clone over `git:` fails?

pescobar:
@boegel that sounds perfect to me. It would address both situtations

PR #2002, opened 2016-11-16
also pick up $EASYBUILD_MODULE_SYNTAX in bootstrap script

boegel:


boegel:
correct fix is in #2003

PR #2003, opened 2016-11-17
fix testing of bootstrap script in Travis config

boegel:


boegel:
This change in the Travis config is actually dancing around a bug in the bootstrap script, see #2007...

PR #2011, opened 2016-11-20
use setvar in modules.py to define environment variables

boegel:
without `setvar`, there are no log entries...


boegel:
@wpoely86 please review? ;)

Issue #2006, opened 2016-11-18
boostraping easybuild-3.0 does not install rpath_wrapper_template.sh.in


EricDeveaud:
Hello, 

after completly wipping the previous easybuild installation and installing easybuil-3.0 the following way 

```
wget https://raw.githubusercontent.com/hpcugent/easybuild-framework/develop/easybuild/scripts/bootstrap_eb.py && \  
    python bootstrap_eb.py /soft/adm/easybuild
```
lead to correct installation, see

```
bigmess:src/easybuild > module av                

---------------------------------------- /soft/adm/easybuild/modules/all ----------------------------------------
   EasyBuild/3.0.0 (L)

bigmess:src/easybuild > module load EasyBuild
bigmess:src/easybuild > eb --version 
This is EasyBuild 3.0.0 (framework: 3.0.0, easyblocks: 3.0.0) on host bigmess.it.pasteur.fr.
```

in order to test rpath support 
I've done the following.

```
eb goolf-1.7.20.eb --robot -l --experimental --rpath
```

that unfortunatly crashed with.

```
2016-11-18 09:54:24,480 build_log.py:149 ERROR EasyBuild crashed with an error (at easybuild/software/EasyBuild/3.0.0/lib/python2.7/site-packages/vsc_base-2.5.5-py2.7.egg/vsc/utils/exceptions.py:124 in __init__): build failed (first 300 chars): Script 'rpath_wrapper_template.sh.in' not found at expected location: /soft/adm/easybuild/software/EasyBuild/3.0.0/lib/python2.7/site-packages/easybuild_framework-3.0.0-py2.7.egg/easybuild/scripts/rpath_wrapper_template.sh.in (at easybuild/software/EasyBuild/3.0.0/lib/python2.7/site-packages/easybuild_framework-3.0.0-py2.7.egg/easybuild/main.py:118 in build_and_install_software)
```

regards

Eric

pescobar:
I have also experienced this same issue

boegel:
@EricDeveaud Thanks for the bug report, I indeed seem to have overlooked this entirely...

I guess it also shows that we should be running the tests differently in Travis, since we're totally missing packaging issues now.

To fix this, you can go to the `easybuild/scripts` directory mentioned in the error message, and pull in https://raw.githubusercontent.com/hpcugent/easybuild-framework/master/easybuild/scripts/rpath_wrapper_template.sh.in using `curl` or `wget`, which should fix the problem.

[...]

PR #2015, opened 2016-11-22
fix packaging issue with non-Python scripts in easybuild/scripts

boegel:
fix for #2006, see change to `MANIFEST.in`

I also changed the Travis config to run the test suite differently so we'll actually notice any packaging issues in the future.


boegel:
@wpoely86 please review

boegel:
Thanks for the review @wpoely86 and @rjeschmi!

PR #2016, opened 2016-11-22
make tests more robust against running headless

boegel:
For some reason unclear to me, a couple of the EasyBuild framework tests have been failing when running from my fork (https://github.com/boegel/easybuild-framework), with errors like this:

```
FAIL: Test multidiff function.
  File "/home/travis/build/boegel/easybuild-framework/test/framework/filetools.py", line 436, in test_multidiff
    self.assertEqual(lines[2], "3 %s- versionsuffix = '-test'%s (1/2) toy-0.0-gompi-1.3.12-test.eb" % (red, endcol))
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/vsc_install-0.10.19-py2.6.egg/vsc/install/testing.py", line 79, in assertEqual
    raise AssertionError("%s:\nDIFF%s:\n%s" % (msg, limit, ''.join(diff[:self.ASSERT_MAX_DIFF])))
AssertionError: "3 \x1b[0;41m- versionsuffix = '-test'\x1b[0m (1/2) toy-0.0-gompi-1.3.12-test\x1b[0m..." != "3 \x1b[0;41m- versionsuffix = '-test'\x1b[0m (1/2) toy-0.0-gompi-1.3.12-test.eb":
DIFF:
- 3 - versionsuffix = '-test' (1/2) toy-0.0-gompi-1.3.12-test...?                                                                       ---- ^^
+ 3 - versionsuffix = '-test' (1/2) toy-0.0-gompi-1.3.12-test.eb
```
```
FAIL: Test det_terminal_size function.
  File "/home/travis/build/boegel/easybuild-framework/test/framework/systemtools.py", line 687, in test_det_terminal_size
    self.assertTrue(isinstance(height, int) and height > 0)
AssertionError
```

I'm not sure why they only fail for my fork and not the central repos, since the test setup is exactly the same. My best guess is that Travis is gradually rolling out some changes which have not yet hit all test environments.

This change makes the affected tests more robust against running headless. For the `multidiff` test, it makes some of the test cases slightly less strict, but that shouldn't affect the capability to catch regression (since the test is quite extensively).

### DIFF
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -514,7 +514,7 @@ class ModulesTest(EnhancedTestCase):
 
         if isinstance(self.modtool, Lmod):
             # GCC/4.6.3 is nowhere to be found (in $MODULEPATH)
-            load_err_msg = r"The following module\(s\) are unknown"
+            load_err_msg = r"The[\s\n]*following[\s\n]*module\(s\)[\s\n]*are[\s\n]*unknown"
         else:
             load_err_msg = "Unable to locate a modulefile"
 
@@ -547,7 +547,10 @@ class ModulesTest(EnhancedTestCase):
         if isinstance(self.modtool, Lmod):
             # OpenMPI/1.6.4 exists, but is not available for load;
             # exact error message depends on Lmod version
-            load_err_msg = r"These module\(s\) exist but cannot be|The following module\(s\) are unknown"
+            load_err_msg = '|'.join([
+                r'These[\s\sn]*module\(s\)[\s\sn]*exist[\s\sn]*but[\s\sn]*cannot[\s\sn]*be',
+                'The[\s\sn]*following[\s\sn]*module\(s\)[\s\sn]*are[\s\sn]*unknown',
+            ])
         else:
             load_err_msg = "Unable to locate a modulefile"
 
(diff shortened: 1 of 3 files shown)


boegel:
@wpoely86 please review

test/framework/modules.py
```diff
@@ -514,7 +514,7 @@ def test_modules_tool_stateless(self):
 
         if isinstance(self.modtool, Lmod):
             # GCC/4.6.3 is nowhere to be found (in $MODULEPATH)
-            load_err_msg = r"The following module\(s\) are unknown"
+            load_err_msg = r"The[\s\n]*following[\s\n]*module\(s\)[\s\n]*are[\s\n]*unknown"
```
wpoely86: this is just weird?

boegel: Why? Lmod is wrapping output lines differently in a headless (0-width) terminal, so the whole message doesn't fit on a single line anymore, it's wrapped.

wpoely86: It shouldn't be wrapping at all if the with is 0?

boegel: well, it does

remember, I'm also testing with old Lmod versions, so this is just something we need to deal with (it's also not a big issue imho)

we're not testing Lmod here, we're checking that some module is not available because EB didn't put it there

boegel:
Thanks for the critical review @wpoely86!

Issue #2010, opened 2016-11-19
Easybuild with rpath support enabled is really slow when building extensions. Easybuild launches many processes


pescobar:
When compiling my old Perl easyconfig which includes more than 400 extensions enabling the rpath support I have noticed that the build takes really long. 

While monitoring the build with htop I have noticed that easybuild is launching a lot of processes when building the extensions. See the attached screenshot. Is this expected?

![many processes](http://i.imgur.com/Y78Ydkh.png)

boegel:
@pescobar For every compilation or linking command, the RPATH wrapper will kick in which means you don't get a single process (e.g. the compiler), but 3 processes (the wrapper, a Python script called by the wrapper, and the compiler)
The wrapper and the Python script are *very* light weight though, and some benchmarking I did didn't show huge overhead (e.g. for OpenFOAM the whole build slows down with less than 10%).

The output you showed seems to suggest something is very wrong though, since they're not all running the same wrapper script, but different ones (the `tmp*` subdir is different every time)...

This looks weird to me, since only a single `gcc` wrapper script should be generated/used for every extension being installed (+ one for Perl itself), and the extensions are installed sequentially, not in parallel.
I think you are hitting a bug that only pops up when building extensions.

One likely possibility is that the `gcc` wrapper script used for building `Perl` itself is being wrapped with an additional wrapper script when the 1st extension is being built, which then gets wrapped *again* when the 2nd extension is being installed, and so on.

So, when you're building the 400th extension, there's actually 400 wrapper scripts being called before the actual compiler gets called... That's of course not the intention. ;-)

This probably happens because the `prepare_step` is called again for every extension, which installs an additional wrapper script that wraps around the current `gcc` 'command' that is first in `$PATH` (which may already be an RPATH wrapper script).

There is a check in place to avoid that the wrapper script calls itself (since that would make it a fork bomb), but it doesn't check that it's wrapping the *real* compiler and not a wrapper script...

boegel:
fixed in #2022

boegel:
follow-up in #2022

PR #2022, opened 2016-11-23
avoid rewrapping already wrapped compiler/linker command with RPATH wrapper script

boegel:
fix for #2010

cc @pescobar

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -57,6 +57,8 @@ _log = fancylogger.getLogger('tools.toolchain', fname=False)
 CCACHE = 'ccache'
 F90CACHE = 'f90cache'
 
+RPATH_WRAPPERS_SUBDIR = 'rpath_wrappers'
+
 
 class Toolchain(object):
     """General toolchain class"""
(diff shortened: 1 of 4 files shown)


boegel:
@wpoely86, @rjeschmi: please review?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -727,6 +729,15 @@ def prepare_compiler_cache(self, cache_tool):
         self.cached_compilers.update(compilers)
         self.log.debug("Cached compilers (after preparing for %s): %s", cache_tool, self.cached_compilers)
 
+    @staticmethod
+    def is_rpath_wrapper(path):
+        """
+        Check whether command at specified location already is an RPATH wrapper script rather than the actual command
+        """
+        in_rpath_wrappers_dir = os.path.basename(os.path.dirname(path)) == RPATH_WRAPPERS_SUBDIR
+        calls_rpath_args = 'rpath_args.py $CMD' in read_file(path)
```
wpoely86: use hash?

boegel: The RPATH wrapper script that is in place is a templated version of `easybuild/scripts/rpath_wrapper_template.sh.in`. That doesn't mean we can't hash it, but I'd rather have a stupid yet safe check in place to really avoid wrapping an existing wrapper. We really *really* REALLY don't want to include a fork bomb in EasyBuild...

Checking a hash is really strict, you just need a single byte changed somehow to have a hash mismatch...

wpoely86: Yes, so it's very very strict, like you want?

boegel: I don't necessarily want the check to see whether the command I'm wrapping is a wrapper script already to be strict. On the contrary, I'd rather do a less strict non-hash check (like I'm doing now), which limits the chance of running into a false negative and wrapping an existing wrapper...

boegel: @wpoely86 So, do we still disagree on this?

boegel:
Thanks for the review @wpoely86@

Issue #2008, opened 2016-11-19
Getting tracebacks from build_log.py


akesandgren:
I'm now and then getting this:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/logging/handlers.py", line 76, in emit
    if self.shouldRollover(record):
  File "/usr/lib/python2.7/logging/handlers.py", line 156, in shouldRollover
    msg = "%s\n" % self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
TypeError: not enough arguments for format string
Logged from file build_log.py, line 149
```

akesandgren:
I'm getting hit by it during build of ParaView 5.1.2 -mpi with intel 2017.01 which hits an internal compiler error on one of the files.

And i never get to see the actual build output, due to the traceback.

I have captured the ebmsg and args in build_log's error() function if it's of any use (quite large...)

ocaisa:
I think this is an old problem with rolling over large log files, the log
file from ParaView is massive. From what I remember this doesn't affect the
installation, just the recording of the build output.

boegel:
The related issue is #1094 , but the traceback is different here, this *may* be a different problem...

boegel:
I've also seen this occurring with fairly small logs, e.g. with a failing HDF5 build.

My suspicion is that it only occurs when `%` is present in the error message, which is screwing up Python string templating done inside of `format` in the loggers...

boegel:
fixed in #2024, let's follow up there

PR #1978, opened 2016-11-05
clean up behaviour that was deprecated for EasyBuild v3.0

boegel:

PR #2024, opened 2016-11-26
fix log.error traceback due to 'raise EasyBuildError' involving a '%s' in error message

boegel:
fix for #2008

The traceback occurs when an `EasyBuildError` is being raised where the error message includes a '`%s`', as sometimes occurs in the output of a failed command that was run via `run_cmd`.

This is a side effect of the switch from the (now removed) `_error_no_raise` log method to (the default) `error` log method for reporting a raised `EasyBuildError` in #1978.

Tests are included to make sure this doesn't get reintroduced.


boegel:
@wpoely86 please review

boegel:

Issue #2025, opened 2016-11-26
original easyconfig is not dumped to archive under --minimal-toolchains


boegel:
cfr. https://lists.ugent.be/wws/arc/easybuild/2016-11/msg00063.html

cc @ocaisa, @JackPerdue

boegel:
fixed in #2028, follow-up there

PR #2028, opened 2016-11-27
only dump easyconfig with modified deps due to --minimal-toolchains to 'reprod' subdir of install dir

boegel:
(fix for #2025)

This restores dumping of the *original* easyconfig used to both the easyconfig archive and the `easybuild` subdir of the install directory.

The tweaked easyconfig, where dependencies may get assigned a particular subtoolchains (and where template values may have been replaced with the actual value) is still dumped to the `easybuild/reprod` subdir of the install directory, as it was before.

The current behaviour, which was introduced in #1614 (EasyBuild v2.7.0) is considered not fully correct, see also https://lists.ugent.be/wws/arc/easybuild/2016-11/msg00063.html .

cc @ocaisa, @JackPerdue, @verdurin


boegel:
Thanks for the review @ocaisa!

PR #2030, opened 2016-11-29
avoid appending '-h' to sanity check commands specified as a string

boegel:
fix for bug reported in https://github.com/hpcugent/easybuild-easyconfigs/pull/3858


boegel:
@wpoely86 please review?

boegel:

PR #2032, opened 2016-11-30
fix bug in list_software_rst: always include versionsuffix regardless of value

boegel:
fix for https://github.com/hpcugent/easybuild/issues/283


boegel:

PR #2047, opened 2016-12-07
add 'rpath' toolchain option to selectively disable use of RPATH wrappers

boegel:
@pescobar this probably works, but needs to be tested. Can you give it a try?

I'll also look into a dedicated test for this.

PR #2052, opened 2016-12-10
skip RPATH sanity check when toolchain did not use RPATH wrappers

boegel:
completes #2047 

cc @pescobar


pescobar:
this seems to fix the bug repoted in https://github.com/hpcugent/easybuild-framework/pull/2047

Here you have the build log for the same easyconfig which was crashing before this patch
https://gist.github.com/428bfe6814608b821b7d3787169e99e9

boegel:
Thanks for the feedback @pescobar!

Issue #2057, opened 2016-12-12
Using environment variables in config.cfg


brevans:
Referring to the environment variables $USER or $HOME in the actual file would be handy, especially since many of the defaults in the documentation refer to those variables. I've tried $HOME and ${HOME} and neither seem to work, instead referring to the literal strings. 

Is there a sensible way to have access to some if not all environment variables from within the config.cfg file? Maybe via a few more supported config-level constants? I understand that these are easily overridden with EASYBUILD_* and command line options, but I'd like to set portable but user-specific defaults for our sites.

boegel:
@brevans Right now, using environment variables in an EasyBuild configuration file is indeed not supported yet, but I agree this would be very useful.

There are a couple of options:

* define selected (commonly) available environment variables as additional template values that can be used in config files; the prime candidates would indeed by `$HOME` and `$USER`, but maybe others make sense too
* define *ALL* environment variables dynamically as template values for config files; this is more 'intrusive' in some sense, but maybe it's OK to just do it this way. The only problem I can think of with this is that it encourages using environment variables that may not always be defined, which may lead to confusing/annoying errors when those config files get parsed in an environment where they are not defined

To implement this, the `go_cfg_constants` dict that is used to set up the 'initial environment' used by the option parser `EasyBuildOptions` needs to be updated.

Are you up for implementing this @brevans?

boegel:
@brevans shall we close this since we're following up in #2063?

boegel:
fixed for `$HOME` and `$USER` in #2063, so this issue can be closed @brevans?

brevans:
yes, all done!

PR #2062, opened 2016-12-19
#2057: environment variables in cfg

brevans:
brought up in [this issue](https://github.com/hpcugent/easybuild-framework/issues/2057). After playing around a bit I didn't see a great reason to introduce other environment variables than HOME and USER.


boegel:
replaced by #2063

PR #2063, opened 2016-12-19
added HOME and USER from env to available cfg file constants

brevans:
Brought up in this [issue](https://github.com/hpcugent/easybuild-framework/issues/2057). After playing around a bit I didn't see a great reason to introduce other environment variables than HOME and USER.

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -215,12 +215,16 @@ class EasyBuildOptions(GeneralOption):
         # set up constants to seed into config files parser, by section
         self.go_cfg_constants = {
             self.DEFAULTSECT: {
-                'DEFAULT_REPOSITORYPATH': (self.default_repositorypath[0], "Default easyconfigs repository path"),
+                'DEFAULT_REPOSITORYPATH': (self.default_repositorypath[0], 
+                                           "Default easyconfigs repository path"),
                 'DEFAULT_ROBOT_PATHS': (os.pathsep.join(self.default_robot_paths),
                                         "List of default robot paths ('%s'-separated)" % os.pathsep),
+                'USER': (os.environ['USER'], "Current username, derived from the os environment"),
+                'HOME': (os.environ['HOME'], 
+                         "Current user's home directory, derived from the os environment")
             }
         }
-
+        
         # update or define go_configfiles_initenv in named arguments to pass to parent constructor
         go_cfg_initenv = kwargs.setdefault('go_configfiles_initenv', {})
         for section, constants in self.go_cfg_constants.items():
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -40,6 +40,7 @@ import re
 import shutil
 import sys
 import tempfile
+import pwd
 import vsc.utils.generaloption
 from distutils.version import LooseVersion
 from vsc.utils import fancylogger
@@ -219,12 +220,13 @@ class EasyBuildOptions(GeneralOption):
                                            "Default easyconfigs repository path"),
                 'DEFAULT_ROBOT_PATHS': (os.pathsep.join(self.default_robot_paths),
                                         "List of default robot paths ('%s'-separated)" % os.pathsep),
-                'USER': (os.environ['USER'], "Current username, derived from the os environment"),
-                'HOME': (os.environ['HOME'], 
-                         "Current user's home directory, derived from the os environment")
+                'USER': (pwd.getpwuid(os.geteuid()).pw_name,
+                         "Current username, translated uid from password file"),
+                'HOME': (os.path.expanduser('~'), 
+                         "Current user's home directory, expanded '~'")
             }
         }
-        
+
         # update or define go_configfiles_initenv in named arguments to pass to parent constructor
         go_cfg_initenv = kwargs.setdefault('go_configfiles_initenv', {})
         for section, constants in self.go_cfg_constants.items():
--- a/test/framework/config.py
+++ b/test/framework/config.py
@@ -279,7 +279,7 @@ class EasyBuildConfigTest(EnhancedTestCase):
             'buildpath = %s' % testpath1,
             'sourcepath = %(DEFAULT_REPOSITORYPATH)s',
             'repositorypath = %(DEFAULT_REPOSITORYPATH)s,somesubdir',
-            'robot-paths=/tmp/foo:%(sourcepath)s:%(DEFAULT_ROBOT_PATHS)s:%(HOME)s:/tmp/%(USER)s',
+            'robot-paths=/tmp/foo:%(sourcepath)s:%(HOME)s:/tmp/%(USER)s:%(DEFAULT_ROBOT_PATHS)s',
             'installpath-modules=%s' % installpath_modules,
         ])
         write_file(config_file, cfgtxt)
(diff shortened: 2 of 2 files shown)


easybuild/tools/options.py
```diff
@@ -215,12 +215,16 @@ def __init__(self, *args, **kwargs):
         # set up constants to seed into config files parser, by section
         self.go_cfg_constants = {
             self.DEFAULTSECT: {
-                'DEFAULT_REPOSITORYPATH': (self.default_repositorypath[0], "Default easyconfigs repository path"),
+                'DEFAULT_REPOSITORYPATH': (self.default_repositorypath[0], 
+                                           "Default easyconfigs repository path"),
                 'DEFAULT_ROBOT_PATHS': (os.pathsep.join(self.default_robot_paths),
                                         "List of default robot paths ('%s'-separated)" % os.pathsep),
+                'USER': (os.environ['USER'], "Current username, derived from the os environment"),
+                'HOME': (os.environ['HOME'],
```
boegel: @brevans please use `os.path.expanduser('~')` here rather than `os.environ['HOME']`

There's a subtle difference, and we use `expanduser` in other places already, so let's be consistent; this is still based on `$HOME` on Linux though, see:

```shell
$ HOME=foo python -c "import os; print os.path.expanduser('~')"
foo
```

easybuild/tools/options.py
```diff
@@ -215,12 +215,16 @@ def __init__(self, *args, **kwargs):
...
+                                           "Default easyconfigs repository path"),
                 'DEFAULT_ROBOT_PATHS': (os.pathsep.join(self.default_robot_paths),
                                         "List of default robot paths ('%s'-separated)" % os.pathsep),
+                'USER': (os.environ['USER'], "Current username, derived from the os environment"),
+                'HOME': (os.environ['HOME'], 
+                         "Current user's home directory, derived from the os environment")
             }
         }
-
+
```
boegel: nitpicking: keep this line blank?

easybuild/tools/options.py
```diff
@@ -215,12 +215,16 @@ def __init__(self, *args, **kwargs):
         # set up constants to seed into config files parser, by section
         self.go_cfg_constants = {
             self.DEFAULTSECT: {
-                'DEFAULT_REPOSITORYPATH': (self.default_repositorypath[0], "Default easyconfigs repository path"),
+                'DEFAULT_REPOSITORYPATH': (self.default_repositorypath[0], 
+                                           "Default easyconfigs repository path"),
```
boegel: I'm ok with this if you do this consistently, i.e. also for `USER`

boegel:
@brevans are you up for also updating the unit test for this? see `test_generaloption_config_file` in `test/framework/config.py`, where we already test `DEFAULT_REPOSITORYPATH`, etc. ?

easybuild/tools/options.py
brevans: I suppose we should also then use `os.getlogin()` for the user name?

easybuild/tools/options.py
brevans: my mistake, will correct

easybuild/tools/options.py
boegel: Hmm, based on the documentation for `getlogin`, no:

```
Return the name of the user logged in on the controlling terminal of the process. For most purposes, it is more useful to use the environment variable LOGNAME to find out who the user is, or pwd.getpwuid(os.getuid())[0] to get the login name of the process’s real user id.
```

I've never heard about `$LOGNAME`, so I wouldn't use that. You could go via `os.getuid`, but I'm not sure it's worth the trouble here...

brevans:
Should we just test for existence of USER and HOME? They could just be invoked in a commented line added to cfgtxt as something like this?:
`'#built by %(USER)s, whose home directory is %(HOME)'`

easybuild/tools/options.py
brevans: After a bit of reading, I think `pwd.getpwuid(os.geteuid()).pw_name` is the best solution on osx and unix. Any strong feelings against?

boegel:
@brevans I don't think template values in a commented out line will be expanded, you'll need to actually use them in a config setting

In the test, I would also verify that they're actually expanded to the value of `$HOME` and `$USER` (which you can grab via `os.environ['HOME']` etc.)

easybuild/tools/options.py
boegel: Sounds ok to me, but in practice this should be the same as `$USER`, right?

So, I would check the templated value against `$USER` in the tests, which is a good sanity check

boegel:
@brevans any idea if you'll be able to finish this soonish? I'm starting to lock this down that will make it in the EasyBuild v3.1.0 release, it would be nice to include this too.

brevans:
I'll have something submitted by end of next week

test/framework/config.py
```diff
@@ -272,13 +272,14 @@ def test_generaloption_config_file(self):
...
         # test with config file passed via environment variable
+        # also test for existence of HOME and USER by adding paths to robot-paths
         installpath_modules = tempfile.mkdtemp(prefix='installpath-modules')
         cfgtxt = '\n'.join([
             '[config]',
             'buildpath = %s' % testpath1,
             'sourcepath = %(DEFAULT_REPOSITORYPATH)s',
             'repositorypath = %(DEFAULT_REPOSITORYPATH)s,somesubdir',
-            'robot-paths=/tmp/foo:%(sourcepath)s:%(DEFAULT_ROBOT_PATHS)s',
+            'robot-paths=/tmp/foo:%(sourcepath)s:%(DEFAULT_ROBOT_PATHS)s:%(HOME)s:/tmp/%(USER)s',
```
boegel: @brevans this looks good, but there are situation (e.g. my setup) where the test would fail without having `%(DEFAULT_ROBOT_PATHS)s` at the end; that's also why we're not checking for the entire list of `robot_paths`, but just the first *N*...

So, please change to:

```python
'robot-paths=/tmp/foo:%(sourcepath)s:%(HOME)s:/tmp/%(USER)s:%(DEFAULT_ROBOT_PATHS)s',
```

test/framework/config.py
```diff
@@ -300,8 +301,10 @@ def test_generaloption_config_file(self):
             '/tmp/foo',
             os.path.join(os.getenv('HOME'), '.local', 'easybuild', 'ebfiles_repo'),
             os.path.join(tmpdir, 'easybuild', 'easyconfigs'),
+            os.getenv('HOME'),
+            os.path.join('/tmp',os.getenv('USER')),
         ]
-        self.assertEqual(options.robot_paths[:3], robot_paths)
+        self.assertEqual(options.robot_paths[:5], robot_paths)
```
boegel: turn out the error is pretty confusing when the comparison is done like this (as in, I couldn't figure it out and had to break this up into separate tests), so let's change this to:

```python
# hardcoded first entry
self.assertEqual(options.robot_paths[0], '/tmp/foo')
# resolved value for %(sourcepath)s template
self.assertEqual(options.robot_paths[1], os.path.join(os.getenv('HOME'), '.local', 'easybuild', 'ebfiles_repo'))
# resolved value for HOME constant
self.assertEqual(options.robot_paths[2], os.getenv('HOME'))
# resolved value that uses USER constant
self.assertEqual(options.robot_paths[3], os.path.join('/tmp', os.getenv('USER')))
# first path in DEFAULT_ROBOT_PATHS
self.assertEqual(options.robot_paths[4], os.path.join(tmpdir, 'easybuild', 'easyconfigs'))
```

test/framework/config.py
```diff
@@ -300,8 +301,20 @@ def test_generaloption_config_file(self):
             '/tmp/foo',
             os.path.join(os.getenv('HOME'), '.local', 'easybuild', 'ebfiles_repo'),
             os.path.join(tmpdir, 'easybuild', 'easyconfigs'),
+            os.getenv('HOME'),
+            os.path.join('/tmp',os.getenv('USER')),
```
boegel: @brevans you can remove this `robot_paths` list too, it's not used anymore?

boegel:
Looks excellent now, thanks for looking into this @brevans! Going in...

Issue #2036, opened 2016-11-30
Error if scaling_max_freq unreadable


ocaisa:
`eb` gets a traceback if `/sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq` is not readable, probably just noting that it is unreadable is enough.

This would solve the problems I have running `eb` on [Bash on Ubuntu on Windows](https://msdn.microsoft.com/en-us/commandline/wsl/about) which I know is probably a big concern for you

boegel:
Hmm, is this the only 'file' we need to be careful with w.r.t. read permissions, or are there others?

Have you run the framework unit tests in the Bash@Windows environment?

Primarily these: `python -O -m test.framework.systemtools`

ocaisa:
Yes, just ran those tests and changing the read permissions allowed them to pass:
```
ocaisa@DESKTOP-1653BCN:~$ python -O -m test.framework.systemtools
INFO: This is (based on) vsc.install.shared_setup 0.10.20
.............E......................E
======================================================================
ERROR: test_cpu_speed_native (__main__.SystemToolsTest)
Test getting CPU speed.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ocaisa/EasyBuild/software/EasyBuild/3.0.1/lib/python2.7/site-packages/easybuild_framework-3.0.1-py2.7.egg/test/framework/systemtools.py", line 403, in test_cpu_speed_native
    cpu_speed = get_cpu_speed()
  File "/home/ocaisa/EasyBuild/software/EasyBuild/3.0.1/lib/python2.7/site-packages/easybuild_framework-3.0.1-py2.7.egg/easybuild/tools/systemtools.py", line 356, in get_cpu_speed
    txt = read_file(MAX_FREQ_FP)
  File "/home/ocaisa/EasyBuild/software/EasyBuild/3.0.1/lib/python2.7/site-packages/easybuild_framework-3.0.1-py2.7.egg/easybuild/tools/filetools.py", line 143, in read_file
    raise EasyBuildError("Failed to read %s: %s", path, err)
EasyBuildError: "Failed to read /sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq: [Errno 13] Permission denied: '/sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq'"

======================================================================
ERROR: test_system_info (__main__.SystemToolsTest)
Test getting system info.
----------------------------------------------------------------------
[...]

PR #2065, opened 2016-12-21
check whether file-like paths are readable before reading them in systemtools module

boegel:
fix for #2036


boegel:
@ocaisa please review?

boegel:
Thanks for the review @ocaisa!

PR #2071, opened 2016-12-29
optarch can now be specified on a toolchain basis

damianam:
This PR enables EB to accept a more powerful `--optarch` syntax. Valid options are:

`--optarch=someflag` -> Same as now
`--optarch=GENERIC` -> Same as now
`--optarch=Intel:xCORE-AVX512,GCC:GENERIC` -> Tells Intel based toolchains to use `-xCORE-AVX512` as a flag, and GCC based toolchains to use a generic set of flags (as specified in the toolchain definition)
`--optarch=Intel:,GCC:someflag` -> Tells Intel based toolchains to have the default behaviour (currently `-xHost`), and GCC based toolchains to use `-someflag`
`--optarch=Intel,GCC:someflag` -> Same as the previous case
`--optarch=Intel:xCORE-AVX512,GCC:GENERIC,Intel:xHost` -> Tells Intel based toolchains to use `-xCORE-AVX512`. The second case is ignored.

This allows to install packages on multiple toolchains using the correct `--optarch` option in a single command, or to have a single definition of `EASYBUILD_OPTARCH` that is correct regardless of the toolchain used.

A test is pending.

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -32,6 +32,7 @@ Command line options for eb
 :author: Jens Timmerman (Ghent University)
 :author: Toon Willems (Ghent University)
 :author: Ward Poelmans (Ghent University)
+:author: Damian Alvarez (Forschungszentrum Juelich GmbH)
 """
 import copy
 import glob
@@ -79,7 +80,8 @@ from easybuild.tools.modules import Lmod
 from easybuild.tools.ordereddict import OrderedDict
 from easybuild.tools.run import run_cmd
 from easybuild.tools.package.utilities import avail_package_naming_schemes
-from easybuild.tools.toolchain.compiler import DEFAULT_OPT_LEVEL, OPTARCH_TOOLCHAIN_DELIMITER, OPTARCH_OPTION_DELIMITER, Compiler
+from easybuild.tools.toolchain.compiler import DEFAULT_OPT_LEVEL, OPTARCH_SEP, \
+        OPTARCH_MAP_CHAR, Compiler
 from easybuild.tools.toolchain.utilities import search_toolchain
 from easybuild.tools.repository.repository import avail_repositories
 from easybuild.tools.version import this_is_easybuild
@@ -717,25 +719,7 @@ class EasyBuildOptions(GeneralOption):
 
         # make sure --optarch has a valid format
         if self.options.optarch:
-            # optarch is in most cases equivalent to strlist, but changing the type breaks compatibility, so we hack it here
-            n_compilers = len(self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER))
-            
-            # if the format is not adequate we refuse continuing
-            if n_compilers != self.options.optarch.count(OPTARCH_OPTION_DELIMITER) and n_compilers > 1:
-                raise EasyBuildError("The optarch option has an incorrect syntax: %s" % self.options.optarch)
-            else:
-                # if there are options for different compilers, we set up a dict
-                if OPTARCH_OPTION_DELIMITER in self.options.optarch:
-                    optarch_dict = dict()
-                    for compiler, compiler_opt in [i.split(OPTARCH_OPTION_DELIMITER) for i in self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER)]:
-                        if compiler in optarch_dict:
-                            raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s" % (compiler, self.options.optarch))
-                        else:
-                            optarch_dict[compiler] = compiler_opt
-                    self.options.optarch = optarch_dict 
-                # we do nothing and keep the string
-                else:
-                    pass
+            self._postprocess_optarch()
 
         # handle configuration options that affect other configuration options
         self._postprocess_config()
@@ -744,6 +728,31 @@ class EasyBuildOptions(GeneralOption):
         if self.options.show_config or self.options.show_full_config:
             self.show_config()
             cleanup_and_exit(self.tmpdir)
+
+    def _postprocess_optarch(self):
+        """Postprocess --optarch option."""
+        optarch_parts = self.options.optarch.split(OPTARCH_SEP)
+        n_compilers = len(optarch_parts)
+        
+        # if the format is not adequate we refuse continuing
+        if n_compilers > 1 and n_compilers != self.options.optarch.count(OPTARCH_MAP_CHAR):
+            raise EasyBuildError("The optarch option has an incorrect syntax: %s" % self.options.optarch)
+        else:
+            logger = fancylogger.getLogger()
+            # if there are options for different compilers, we set up a dict
+            if OPTARCH_MAP_CHAR in optarch_parts[0]:
+                optarch_dict = {}
+                for compiler, compiler_opt in [x.split(OPTARCH_MAP_CHAR) for x in optarch_parts]:
+                    if compiler in optarch_dict:
+                        raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s" \
+                                % (compiler, self.options.optarch))
+                    else:
+                        optarch_dict[compiler] = compiler_opt
+                self.options.optarch = optarch_dict 
+                logger.info("Transforming optarch into a dict: %s" % self.options.optarch)
+            # we do nothing and keep the string
+            else:
+                logger.info("Keeping optarch raw: %s" % self.options.optarch)
 
     def _postprocess_include(self):
         """Postprocess --include options."""
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -734,8 +734,9 @@ class EasyBuildOptions(GeneralOption):
         optarch_parts = self.options.optarch.split(OPTARCH_SEP)
         
         # we expect to find a ':' in every entry in optarch, in case optarch is specified on a per-compiler basis
-        if (len(optarch_parts) > 1 and not all(p.count(OPTARCH_MAP_CHAR) == 1 for p in optarch_parts)) or \
-                (len(optarch_parts) == 1 and optarch_parts[0].count(OPTARCH_MAP_CHAR) > 1):
+        n_parts = len(optarch_parts)
+        map_char_cnts = [p.count(OPTARCH_MAP_CHAR) for p in optarch_parts]
+        if (n_parts > 1 and any(c != 1 for c in map_char_cnts)) or (n_parts == 1 and map_char_cnts[0] > 1):
             raise EasyBuildError("The optarch option has an incorrect syntax: %s", self.options.optarch)
         else:
             # if there are options for different compilers, we set up a dict
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -734,16 +734,17 @@ class EasyBuildOptions(GeneralOption):
         optarch_parts = self.options.optarch.split(OPTARCH_SEP)
         
         # we expect to find a ':' in every entry in optarch, in case optarch is specified on a per-compiler basis
-        if len(optarch_parts) > 1 and not all(OPTARCH_MAP_CHAR in p for p in optarch_parts):
+        if (len(optarch_parts) > 1 and not all(p.count(OPTARCH_MAP_CHAR) == 1 for p in optarch_parts)) or \
+                (len(optarch_parts) == 1 and optarch_parts[0].count(OPTARCH_MAP_CHAR) > 1):
             raise EasyBuildError("The optarch option has an incorrect syntax: %s", self.options.optarch)
         else:
             # if there are options for different compilers, we set up a dict
             if OPTARCH_MAP_CHAR in optarch_parts[0]:
                 optarch_dict = {}
-                for compiler, compiler_opt in [x.split(OPTARCH_MAP_CHAR) for x in optarch_parts]:
+                for compiler, compiler_opt in [p.split(OPTARCH_MAP_CHAR) for p in optarch_parts]:
                     if compiler in optarch_dict:
                         raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s",
-                                compiler, self.options.optarch)
+                                             compiler, self.options.optarch)
                     else:
                         optarch_dict[compiler] = compiler_opt
                 self.options.optarch = optarch_dict 
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -351,7 +351,7 @@ class EasyBuildOptions(GeneralOption):
                                  None, 'store', None),
             'mpi-tests': ("Run MPI tests (when relevant)", None, 'store_true', True),
             'optarch': ("Set architecture optimization, overriding native architecture optimizations",
-                        None, 'store', None),
+                        'strlist', 'extend', None),
             'output-format': ("Set output format", 'choice', 'store', FORMAT_TXT, [FORMAT_TXT, FORMAT_RST]),
             'parallel': ("Specify (maximum) level of parallellism used during build procedure",
                          'int', 'store', None),
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -745,7 +745,7 @@ class EasyBuildOptions(GeneralOption):
                 optarch_dict = {}
                 for compiler, compiler_opt in [x.split(OPTARCH_MAP_CHAR) for x in optarch_parts]:
                     if compiler in optarch_dict:
-                        raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s" \
+                        raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s" 
                                 % (compiler, self.options.optarch))
                     else:
                         optarch_dict[compiler] = compiler_opt
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -79,7 +79,7 @@ from easybuild.tools.modules import Lmod
 from easybuild.tools.ordereddict import OrderedDict
 from easybuild.tools.run import run_cmd
 from easybuild.tools.package.utilities import avail_package_naming_schemes
-from easybuild.tools.toolchain.compiler import DEFAULT_OPT_LEVEL, Compiler
+from easybuild.tools.toolchain.compiler import DEFAULT_OPT_LEVEL, OPTARCH_TOOLCHAIN_DELIMITER, OPTARCH_OPTION_DELIMITER, Compiler
 from easybuild.tools.toolchain.utilities import search_toolchain
 from easybuild.tools.repository.repository import avail_repositories
 from easybuild.tools.version import this_is_easybuild
@@ -350,7 +350,6 @@ class EasyBuildOptions(GeneralOption):
             'mpi-cmd-template': ("Template for MPI commands (template keys: %(nr_ranks)s, %(cmd)s)",
                                  None, 'store', None),
             'mpi-tests': ("Run MPI tests (when relevant)", None, 'store_true', True),
-            # optarch is in most cases equivalent to strlist, but changing the type breaks compatibility
             'optarch': ("Set architecture optimization, overriding native architecture optimizations",
                         None, 'store', None),
             'output-format': ("Set output format", 'choice', 'store', FORMAT_TXT, [FORMAT_TXT, FORMAT_RST]),
@@ -715,6 +714,28 @@ class EasyBuildOptions(GeneralOption):
         # imply --terse for --last-log to avoid extra output that gets in the way
         if self.options.last_log:
             self.options.terse = True
+
+        # make sure --optarch has a valid format
+        if self.options.optarch:
+            # optarch is in most cases equivalent to strlist, but changing the type breaks compatibility, so we hack it here
+            n_compilers = len(self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER))
+            
+            # if the format is not adequate we refuse continuing
+            if n_compilers != self.options.optarch.count(OPTARCH_OPTION_DELIMITER) and n_compilers > 1:
+                raise EasyBuildError("The optarch option has an incorrect syntax: %s" % self.options.optarch)
+            else:
+                # if there are options for different compilers, we set up a dict
+                if OPTARCH_OPTION_DELIMITER in self.options.optarch:
+                    optarch_dict = dict()
+                    for compiler, compiler_opt in [i.split(OPTARCH_OPTION_DELIMITER) for i in self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER)]:
+                        if compiler in optarch_dict:
+                            raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s" % (compiler, self.options.optarch))
+                        else:
+                            optarch_dict[compiler] = compiler_opt
+                    self.options.optarch = optarch_dict 
+                # we do nothing and keep the string
+                else:
+                    pass
 
         # handle configuration options that affect other configuration options
         self._postprocess_config()
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -80,8 +80,7 @@ from easybuild.tools.modules import Lmod
 from easybuild.tools.ordereddict import OrderedDict
 from easybuild.tools.run import run_cmd
 from easybuild.tools.package.utilities import avail_package_naming_schemes
-from easybuild.tools.toolchain.compiler import DEFAULT_OPT_LEVEL, OPTARCH_SEP, \
-        OPTARCH_MAP_CHAR, Compiler
+from easybuild.tools.toolchain.compiler import DEFAULT_OPT_LEVEL, OPTARCH_MAP_CHAR, OPTARCH_SEP, Compiler
 from easybuild.tools.toolchain.utilities import search_toolchain
 from easybuild.tools.repository.repository import avail_repositories
 from easybuild.tools.version import this_is_easybuild
@@ -733,27 +732,25 @@ class EasyBuildOptions(GeneralOption):
     def _postprocess_optarch(self):
         """Postprocess --optarch option."""
         optarch_parts = self.options.optarch.split(OPTARCH_SEP)
-        n_compilers = len(optarch_parts)
         
-        # if the format is not adequate we refuse continuing
-        if n_compilers > 1 and n_compilers != self.options.optarch.count(OPTARCH_MAP_CHAR):
-            raise EasyBuildError("The optarch option has an incorrect syntax: %s" % self.options.optarch)
+        # we expect to find a ':' in every entry in optarch, in case optarch is specified on a per-compiler basis
+        if len(optarch_parts) > 1 and not all(OPTARCH_MAP_CHAR in p for p in optarch_parts):
+            raise EasyBuildError("The optarch option has an incorrect syntax: %s", self.options.optarch)
         else:
-            logger = fancylogger.getLogger()
             # if there are options for different compilers, we set up a dict
             if OPTARCH_MAP_CHAR in optarch_parts[0]:
                 optarch_dict = {}
                 for compiler, compiler_opt in [x.split(OPTARCH_MAP_CHAR) for x in optarch_parts]:
                     if compiler in optarch_dict:
-                        raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s" 
-                                % (compiler, self.options.optarch))
+                        raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s",
+                                compiler, self.options.optarch)
                     else:
                         optarch_dict[compiler] = compiler_opt
                 self.options.optarch = optarch_dict 
-                logger.info("Transforming optarch into a dict: %s" % self.options.optarch)
-            # we do nothing and keep the string
+                self.log.info("Transforming optarch into a dict: %s", self.options.optarch)
+            # if optarch is not in mapping format, we do nothing and just keep the string
             else:
-                logger.info("Keeping optarch raw: %s" % self.options.optarch)
+                self.log.info("Keeping optarch raw: %s", self.options.optarch)
 
     def _postprocess_include(self):
         """Postprocess --include options."""
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -350,8 +350,9 @@ class EasyBuildOptions(GeneralOption):
             'mpi-cmd-template': ("Template for MPI commands (template keys: %(nr_ranks)s, %(cmd)s)",
                                  None, 'store', None),
             'mpi-tests': ("Run MPI tests (when relevant)", None, 'store_true', True),
+            # optarch is in most cases equivalent to strlist, but changing the type breaks compatibility
             'optarch': ("Set architecture optimization, overriding native architecture optimizations",
-                        'strlist', 'extend', None),
+                        None, 'store', None),
             'output-format': ("Set output format", 'choice', 'store', FORMAT_TXT, [FORMAT_TXT, FORMAT_RST]),
             'parallel': ("Specify (maximum) level of parallellism used during build procedure",
                          'int', 'store', None),
--- a/easybuild/tools/toolchain/compiler.py
+++ b/easybuild/tools/toolchain/compiler.py
@@ -298,8 +298,9 @@ class Compiler(Toolchain):
         # --optarch=GENERIC
         elif build_option('optarch') == OPTARCH_GENERIC:
             use_generic = True
-        # no --optarch specified
-        elif self.arch in (self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or []):
+
+        # no --optarch specified or no option found for the current compiler
+        if use_generic == False and optarch is None and self.arch in (self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or []):
             optarch = self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[self.arch]
 
         if use_generic == True:
--- a/easybuild/tools/toolchain/compiler.py
+++ b/easybuild/tools/toolchain/compiler.py
@@ -286,9 +286,9 @@ class Compiler(Toolchain):
                 current_compiler = getattr(self, 'COMPILER_FAMILY', None)
                 for optarch_option in build_option('optarch').split(","):
                     comp = optarch_option.split(":")[0]
-                    comp_optarch = optarch_option.split(":")[1] if len(optarch_option.split(":")) > 1 else ''
+                    comp_optarch = optarch_option.split(":")[1] if len(optarch_option.split(":")) > 1 else None 
                     if comp == current_compiler:
-                        if comp_optarch != '' and comp_optarch != OPTARCH_GENERIC:
+                        if comp_optarch is not None and comp_optarch != OPTARCH_GENERIC:
                             optarch = comp_optarch
                         elif comp_optarch == OPTARCH_GENERIC:
                             use_generic = True
--- a/easybuild/tools/toolchain/compiler.py
+++ b/easybuild/tools/toolchain/compiler.py
@@ -302,7 +302,7 @@ class Compiler(Toolchain):
                 # no option for this compiler
                 else:
                     optarch = None
-                    self.log.info("_set_optimal_architecture: no optarch found for compiler %s. Ignoring option.", \
+                    self.log.info("_set_optimal_architecture: no optarch found for compiler %s. Ignoring option.", 
                             current_compiler)
             else:
                 raise EasyBuildError("optarch is neither an string or a dict. This should never ever happen")
--- a/easybuild/tools/toolchain/compiler.py
+++ b/easybuild/tools/toolchain/compiler.py
@@ -27,6 +27,7 @@ Toolchain compiler module, provides abstract class for compilers.
 
 :author: Stijn De Weirdt (Ghent University)
 :author: Kenneth Hoste (Ghent University)
+:author: Damian Alvarez (Forschungszentrum Juelich GmbH)
 """
 from easybuild.tools import systemtools
 from easybuild.tools.build_log import EasyBuildError
@@ -279,12 +280,19 @@ class Compiler(Toolchain):
         optarch = None
         use_generic = False
         # --optarch is specified with flags to use
-        if build_option('optarch') is not None and build_option('optarch') != OPTARCH_GENERIC:
-            if "," not in build_option('optarch') and ":" not in build_option('optarch'):
-                optarch = build_option('optarch')
-            else:
-                current_compiler = getattr(self, 'COMPILER_FAMILY', None)
-                for optarch_option in build_option('optarch').split(","):
+        if build_option('optarch') is not None:
+            for optarch_option in build_option('optarch'):
+                # No compilers specified, and option is not generic, so simply take what is there
+                if ":" not in optarch_option and optarch_option != OPTARCH_GENERIC:
+                    optarch = optarch_option
+                    break
+                # Option is generic, so mark it as to be set later and stop iterating
+                elif optarch_option == OPTARCH_GENERIC:
+                    use_generic = True
+                    break
+                # Compilers specified, check if it is the one we need
+                else:
+                    current_compiler = getattr(self, 'COMPILER_FAMILY', None)
                     comp = optarch_option.split(":")[0]
                     comp_optarch = optarch_option.split(":")[1] if len(optarch_option.split(":")) > 1 else None 
                     if comp == current_compiler:
@@ -293,17 +301,13 @@ class Compiler(Toolchain):
                         elif comp_optarch == OPTARCH_GENERIC:
                             use_generic = True
                         break
-                if optarch is None:
-                    self.log.info("_set_optimal_architecture: no optarch found for compiler %s. Ignoring option." % current_compiler)
-        # --optarch=GENERIC
-        elif build_option('optarch') == OPTARCH_GENERIC:
-            use_generic = True
+            if optarch is None and use_generic == False:
+                self.log.info("_set_optimal_architecture: no optarch found for compiler %s. Ignoring option." % current_compiler)
 
         # no --optarch specified or no option found for the current compiler
         if use_generic == False and optarch is None and self.arch in (self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or []):
             optarch = self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[self.arch]
-
-        if use_generic == True:
+        elif use_generic == True:
             if self.arch in (self.COMPILER_GENERIC_OPTION or []):
                 optarch = self.COMPILER_GENERIC_OPTION[self.arch]
 
--- a/easybuild/tools/toolchain/compiler.py
+++ b/easybuild/tools/toolchain/compiler.py
@@ -43,8 +43,8 @@ DEFAULT_OPT_LEVEL = 'defaultopt'
 OPTARCH_GENERIC = 'GENERIC'
 
 # Characters that separate toolchains and options in --optarch
-OPTARCH_TOOLCHAIN_DELIMITER = ';'
-OPTARCH_OPTION_DELIMITER = ':'
+OPTARCH_SEP = ';'
+OPTARCH_MAP_CHAR = ':'
 
 def mk_infix(prefix):
     """Create an infix based on the given prefix."""
@@ -281,35 +281,31 @@ class Compiler(Toolchain):
         :param default_optarch: default value to use for optarch, rather than using default value based on architecture
                                 (--optarch and --optarch=GENERIC still override this value)
         """
-        optarch = None
         use_generic = False
+        optarch= build_option('optarch') 
         # --optarch is specified with flags to use
-        if build_option('optarch') is not None:
+        if optarch is not None:
             # optarch has been validated as a simple string
-            if type(build_option('optarch')) is str:
+            if isinstance(optarch, basestring):
                 # Option is generic, so mark it as to be set later
-                if build_option('optarch') == OPTARCH_GENERIC:
+                if optarch == OPTARCH_GENERIC:
                     use_generic = True
-                # No compilers specified, and option is not generic, so simply take what is there
-                else:
-                    optarch = build_option('optarch')
 
             # optarch has been validated as complex string with multiple compilers and converted to a dictionary
-            elif type(build_option('optarch')) is dict:
+            elif isinstance(optarch, dict):
                 current_compiler = getattr(self, 'COMPILER_FAMILY', None)
-                if current_compiler in build_option('optarch'):
-                    if build_option('optarch')[current_compiler] == OPTARCH_GENERIC:
+                if current_compiler in optarch:
+                    if optarch[current_compiler] == OPTARCH_GENERIC:
                         use_generic = True
                     else:
-                        optarch = build_option('optarch')[current_compiler]
+                        optarch = optarch[current_compiler]
                 # no option for this compiler
                 else:
-                    pass
+                    optarch = None
+                    self.log.info("_set_optimal_architecture: no optarch found for compiler %s. Ignoring option.", \
+                            current_compiler)
             else:
                 raise EasyBuildError("optarch is neither an string or a dict. This should never ever happen")
-
-            if optarch is None and use_generic == False:
-                self.log.info("_set_optimal_architecture: no optarch found for compiler %s. Ignoring option." % current_compiler)
 
         if use_generic == True:
             if (self.arch, self.cpu_family) in (self.COMPILER_GENERIC_OPTION or []):
@@ -318,7 +314,7 @@ class Compiler(Toolchain):
         elif default_optarch and optarch is None:
             optarch = default_optarch
         # no --optarch specified, no option found for the current compiler, and no default optarch
-        elif use_generic == False and optarch is None and (self.arch, self.cpu_family) in (self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or []):
+        elif optarch is None and (self.arch, self.cpu_family) in (self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or []):
             optarch = self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(self.arch, self.cpu_family)]
 
         if optarch is not None:
--- a/easybuild/tools/toolchain/compiler.py
+++ b/easybuild/tools/toolchain/compiler.py
@@ -42,6 +42,9 @@ DEFAULT_OPT_LEVEL = 'defaultopt'
 # by doing eb --optarch=GENERIC
 OPTARCH_GENERIC = 'GENERIC'
 
+# Characters that separate toolchains and options in --optarch
+OPTARCH_TOOLCHAIN_DELIMITER = ';'
+OPTARCH_OPTION_DELIMITER = ':'
 
 def mk_infix(prefix):
     """Create an infix based on the given prefix."""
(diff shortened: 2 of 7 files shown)


boegel:
@damianam A couple of quick thoughts:

* The 'list of strings' aspect should be handled by making the `optarch` option of type `strlist` in `tools/options.py`
* What's the point of specifying that the default flags should be used for a particular toolchain? If the toolchain is not listed, that would/should be the default behaviour, so there's no point in listing those? Also, how would you then specify not using any 'optarch' compiler options at all?
* The unit tests should be updated too (see `test_*_optarch` in `test/framework/toolchain.py`)
* Are you up for also updating the documentation (http://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html), via a PR to https://github.com/hpcugent/easybuild/blob/master/docs/Controlling_compiler_optimization_flags.rst ?

damianam:
Now `--optarch=Intel:` and `--optarch=Intel:''` effectively disable any `optarch` flag. `--optarch=Intel` does nothing.

boegel:
@damianam but the behaviour of using `Intel:` changes when combined with `GCC:foo`?

damianam:
@boegel no, it behaves the same:

```
eb darshan-runtime-3.1.1-ipsmpi-2016b.eb darshan-runtime-3.1.1-gpsmpi-2016b.eb --extended-dry-run  --optarch=Intel: | grep "export CFLAGS"
  export CFLAGS="-O2 -ftz -fp-speculation=safe -fp-model source"
  export CFLAGS="-O2 -march=native"
```

```
eb darshan-runtime-3.1.1-ipsmpi-2016b.eb darshan-runtime-3.1.1-gpsmpi-2016b.eb --extended-dry-run  --optarch=Intel:,GCC:GENERIC | grep "export CFLAGS"
  export CFLAGS="-O2 -ftz -fp-speculation=safe -fp-model source"
  export CFLAGS="-O2 -march=x86-64 -mtune=generic"
```

```
eb darshan-runtime-3.1.1-ipsmpi-2016b.eb darshan-runtime-3.1.1-gpsmpi-2016b.eb --extended-dry-run  --optarch=Intel,GCC:GENERIC | grep "export CFLAGS"
  export CFLAGS="-O2 -xHost -ftz -fp-speculation=safe -fp-model source"
  export CFLAGS="-O2 -march=x86-64 -mtune=generic"
```

```
eb darshan-runtime-3.1.1-ipsmpi-2016b.eb darshan-runtime-3.1.1-gpsmpi-2016b.eb --extended-dry-run  --optarch=GCC:GENERIC | grep "export CFLAGS"
  export CFLAGS="-O2 -xHost -ftz -fp-speculation=safe -fp-model source"
  export CFLAGS="-O2 -march=x86-64 -mtune=generic"

```

damianam:
@boegel I changed the option type. I'll take a look to the tests soon[ish].

boegel:
@damianam please remove the `(WIP)` tag in the PR title when you consider this good to go

damianam:
@boegel I think the PR is ready now. Please review

damianam:
I've noticed a few caveats:

* Clang won't accept this syntax because it doesn't set `COMPILER_FAMILY` (!?!?). Maybe we should use the compiler class name instead? That doesn't seem very clean though. I suggest to actually assign a `COMPILER_FAMILY` to Clang
* Currently I am using PGI as a failsafe check in the tests, to check that nothing in particular is done to its settings if it is not explicitly listed. The issue here is that by default the optimal architecture setting is simply not setting anything (ie: `''`), and therefore I can't check that `xFLAGS` are correctly set (`'' in any_string` is always `True`). I'd gladly change the check to clang but we need to fix the first issue first.

boegel:
* We haven't considered Clang a full compiler family up until now because it doesn't include a Fortran compiler (all other compiles we support do), but I guess that's not a very compelling argument... I'm OK with defining `COMPILER_FAMILY` for Clang by itself.

* See inline comment for the check involving an empty string.

easybuild/tools/options.py
```diff
@@ -350,6 +350,7 @@ def override_options(self):
             'mpi-cmd-template': ("Template for MPI commands (template keys: %(nr_ranks)s, %(cmd)s)",
                                  None, 'store', None),
             'mpi-tests': ("Run MPI tests (when relevant)", None, 'store_true', True),
+            # optarch is in most cases equivalent to strlist, but changing the type breaks compatibility
```
boegel: hmm, I think this comment would be more confusing in the long run, it's better to mention this where we split the `optarch` value by `,`?

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +279,40 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+        if build_option('optarch') is not None:
+            for optarch_option in build_option('optarch').split(','):
+                # No compilers specified, and option is not generic, so simply take what is there
+                if ":" not in optarch_option and optarch_option != OPTARCH_GENERIC:
+                    optarch = optarch_option
+                    break
+                # Option is generic, so mark it as to be set later and stop iterating
+                elif optarch_option == OPTARCH_GENERIC:
+                    use_generic = True
+                    break
```
boegel: With this approach, a value like `--optarch=GENERIC,intel:foo` would be equivalent to `--optarch=GENERIC`, while `--optarch=intel:foo,GENERIC` would not be, right? That seems very confusing to me...

I think we should do some validation here, and either except `--optarch=GENERIC`, `--optarch=<flags>` or `--optarch=<flags-by-compiler>`, and not allow mixing; if you detect a dict-like value, we should validate that all entries really specify a mapping, and not make things too flexible (since that makes it too easy to shoot yourself in the foot and end up with a configuration that doesn't do what you think it does).

Another problem I'm concerned with is how to discriminate between `--optarch=<flags>` where the `,` character is included in the list of flags (e.g. `-O2 -Wl,-Bstatic`), and `--optarch=<flags-by-compiler>`, where the entries for the different compiler are separated by a `,`.

So, I think we should go with a more 'obscure' separation character rather than `,`, something like `|` or `@` or `&`, i.e. stuff that is a lot less likely to appear in a set of compiler flags.
The same goes for the `:` character that is not used to separate the compiler name from the set of flags to use for that compiler.

Either that, or let `--optarch` accept the path to a file that specifies to compiler-to-flag mapping in some syntax that's easy to parse (e.g. `.ini` format like we have for external modules metadata, see http://easybuild.readthedocs.io/en/latest/Using_external_modules.html#metadata-for-external-modules).

I actually strongly prefer the latter, since that makes things *a lot* easier; you can just check whether `--optarch` points to an existing file, and fall back to what we do now if it doesn't. Parsing a dict-like value from a string is just painful, and it's probably better to avoid that?

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +279,40 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+                # Compilers specified, check if it is the one we need
+                else:
+                    current_compiler = getattr(self, 'COMPILER_FAMILY', None)
+                    comp = optarch_option.split(":")[0]
+                    comp_optarch = optarch_option.split(":")[1] if len(optarch_option.split(":")) > 1 else None 
+                    if comp == current_compiler:
```
boegel: you're not both parsing the `--optarch` value and checking for a match in one go, it would make more sense to do the parsing a lot earlier (when initialising the toolchain), to catch any formatting problems with the specified `--optarch` value early on

the actual checking of which flags to use from that parsed value should still occur here

test/framework/toolchain.py
```diff
@@ -381,6 +381,42 @@ def test_optarch_aarch64_heuristic(self):
...
+    def test_compiler_dependent_optarch(self):
+        """Test whether specifying optarch on a per compiler basis works."""
+        flag_vars = ['CFLAGS', 'CXXFLAGS', 'FCFLAGS', 'FFLAGS', 'F90FLAGS']
+        for intelflags,intel_expanded_flags in [('intelflag','intelflag'),('GENERIC','xSSE2'),('','')]:
+            for gccflags,gcc_expanded_flags in  [('gccflag','gccflag'),('GENERIC','march=x86-64 -mtune=generic'),('','')]:
+                optarch_var = 'Intel:%s,GCC:%s' % (intelflags,gccflags)
+                build_options = {'optarch': optarch_var}
+                init_config(build_options=build_options)
+                for toolchain, toolchain_version in [('iccifort','2011.13.367'),('GCC','4.7.2'),('PGI','16.7-GCC-5.4.0-2.26')]:
+                    for enable in [True, False]:
```
boegel: wow, a quadruple for-loop is a bit hard to understand...

Can't we construct one large list of test cases outside of the for loop, and then just have a single for loop over those?

That would also help with avoiding the deep indenting here...

test/framework/toolchain.py
```diff
@@ -381,6 +381,42 @@ def test_optarch_aarch64_heuristic(self):
...
+                        tc.set_options({'optarch': enable})
+                        tc.prepare()
+                        flag = None
+                        if toolchain == 'iccifort':
+                            flag = intel_expanded_flags
+                        elif toolchain == 'GCC':
+                            flag = gcc_expanded_flags 
+                        else: # PGI as an example of compiler not set
+                            # default optarch flag
+                            flag = tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(tc.arch,tc.cpu_family)]
```
boegel: This should be handled outside the for loop, where you construct a list of test cases?

Also, hardcode the value for PGI rather than grabbing it from the `Toolchain` instance?

test/framework/toolchain.py
```diff
@@ -381,6 +381,42 @@ def test_optarch_aarch64_heuristic(self):
...
+                        
+                        flags = tc.options.options_map['optarch']
+
+                        self.assertTrue(flag == flags, "optarch: True means '%s' == '%s'" % (flag, flags))
+                        
+                        # Also check that it is correctly passed to xFLAGS, honoring 'enable'
+                        if flag != '':
```
boegel: what if flag *is* `== ''`, no checks at all?

easybuild/tools/options.py
damianam: Alright. I'll move it in the next commit

easybuild/tools/toolchain/compiler.py
damianam: You are right in the first part of the comment. My mind set was to accept the "first match" and properly document it. Even validating every entry you need something like that to deal with things like `--optarch=Intel:xHost,GCC:something,Intel:xSSE2`.

I think that like it is now is not just more powerful, but also easier to implement (the validation requires a couple of lines more of code, even though it is not significant). It comes with a cost, which is _lack of strictness_ (sounds worse than _flexibility_ :-P). However I have no strong feelings about doing it like you suggest. Just let me know which direction you want to take.

Regarding the parsing bit and the separators: Couldn't we use some similar library to parse that? Even without it, it shouldn't be too complicated to accept something like `--optarch=Intel:'O2 -Wl,-Bstatic',GCC:something`.

Beyond that, do we really need that capability? I mean, none of the compilers I know require it (see list below), and if an user wants to put flags in `--optarch` that don't belong there, should we get out of our way to support that? In my opinion we should support that just if a compiler requires it.

* Intel: `-xHost` or `-xAVX2`, etc
* GCC: `-march=x86-64 -mtune=generic`
* PGI: `-tp=haswell`
* clang: `-march=x86-64 -mtune=generic`
* IBM XL: `-qtune=auto`

And just as I finish typing this I see that `nvcc` accepts things like `--gpu-architecture=compute_50 --gpu-code=sm_50,compute_50` and my argument falls apart and goes down the drain. So, coming back to the previous point, do we accept something like `--optarch=Intel:'O2 -Wl,-Bstatic',GCC:something`, or do we change the characters? Changing them is trivial so I am ok with it.

Last but not least: I strongly oppose the configuration file idea. My arguments are:

1. Disrupts our workflow. We do this sort of tuning with environment variables that are set in configuration modules. We just use files for python code (custom easyblocks, toolchains, MNS)
2. It is kind of overkill. The longest string I can think of is something like `Intel:xCORE-AVX2,GCC:-march=x86-64 -mtune=generic,PGI:-tp=haswell,CLANG:-march=x86-64 -mtune=generic,IBMXL:-qtune=auto,CUDA:--gpu-architecture=compute_50 --gpu-code=sm_50,compute_50`. This is an exaggeration, but even exaggerating it, it is pretty manageable.

I would be willing to accept a file as an *additional* option, if it fits better in people's workflow, but not excluding the `--optarch=Compiler1:flag1,Compiler2:flag2` syntax

test/framework/toolchain.py
damianam: You mean manually constructing that list of 54 elements, or construct the list in 4 loops, and then iterate over it?

In any case, they are 4 loops but fairly simple

test/framework/toolchain.py
damianam: If I hardcode it it would break if one day changes. A change like that should be irrelevant for the test.

test/framework/toolchain.py
damianam: The only way I can think of to check when flag is `== ''` is to have a blacklist and check against it, which is kind of overkill I think....

easybuild/tools/toolchain/compiler.py
damianam: Mmmmmh. It looks reasonable in principle but following that approach I think it would make more sense to have an input sanitizing step for all the options. Is there something like that?

Otherwise seems kind of unnatural, this option would be the only one that gets parsed before it is actually processed.

easybuild/tools/toolchain/compiler.py
boegel: I think that being too flexible may be an issue too, since it makes it too easy to have `eb` do something while you're expecting something else. Even if this is clearly documented, we should try and avoid surprises, and do a bit of input validation to protect against silly mistakes (we all make them).

I'm OK with your opposition against using (only) a file, I understand the argument. Also supporting a file can always be added later.

The easiest way to go forward seems to be:

* use different characters than `,` and `:` to specify the mapping of compiler to flags (the `nvcc` flags are an excellent example); wrapping flags in single quotes is not something I'd go for, since you need to be quite careful to make sure bash doesn't eat your single quotes...

I would also input validation, see my previous comment and add to that to *not* allow includes the same compiler twice (like in your `Intel:...,...,Intel:....` example), there's no reason we should allow that, it will only lead to confusion. And it's fairly easy to check.

easybuild/tools/toolchain/compiler.py
boegel: We have type checking support for easyconfig parameters, but not for configuration options.

We do a little bit of option checking in the `postprocess` method in `options.py` (see e.g. the check w.r.t. module syntax), but not to a big extent. That would be the right location for input validation though.

test/framework/toolchain.py
boegel: No, I mean creating 4 separate lists, and then letting Python create a list of 4-tuples with all possible options (seems like `itertools.product` does what you need).

Anyway, this is less important, since this is only a test where code style is less of an issue

test/framework/toolchain.py
boegel: If something changes in framework that affects this test, we can update the test accordingly.

I'd rather be aware of the impact of a change in framework.

test/framework/toolchain.py
boegel: well, we should have an `else` in any case?

damianam: Done

test/framework/toolchain.py
damianam: Done

test/framework/toolchain.py
damianam: Ignored, at least for the moment :-P

easybuild/tools/toolchain/compiler.py
damianam: Ok, done

easybuild/tools/toolchain/compiler.py
damianam: Ok, changed to `:` and `;`, and now it is more strict

easybuild/tools/options.py
damianam: Done

damianam:
@boegel I've implemented most of the discussed changes. From my perspective there is just one thing remaining: Making test to make sure that the parsing is done correctly, not just checking for exceptions.

easybuild/tools/options.py
```diff
@@ -715,6 +715,28 @@ def postprocess(self):
         if self.options.last_log:
             self.options.terse = True
 
+        # make sure --optarch has a valid format
+        if self.options.optarch:
+            # optarch is in most cases equivalent to strlist, but changing the type breaks compatibility, so we hack it here
```
boegel: I'm not sure if we should include this comment, since we now have other reasons to take this approach, e.g. `strlist` only supports using `,` as a separator, while we're using `;`

easybuild/tools/options.py
```diff
@@ -715,6 +715,28 @@ def postprocess(self):
...
+            n_compilers = len(self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER))
```
boegel: just to the `split` once, and save it in a variable you reuse later:

```
optarch_parts = self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER)
```

easybuild/tools/options.py
```diff
@@ -715,6 +715,28 @@ def postprocess(self):
...
+            
+            # if the format is not adequate we refuse continuing
+            if n_compilers != self.options.optarch.count(OPTARCH_OPTION_DELIMITER) and n_compilers > 1:
+                raise EasyBuildError("The optarch option has an incorrect syntax: %s" % self.options.optarch)
+            else:
+                # if there are options for different compilers, we set up a dict
+                if OPTARCH_OPTION_DELIMITER in self.options.optarch:
+                    optarch_dict = dict()
+                    for compiler, compiler_opt in [i.split(OPTARCH_OPTION_DELIMITER) for i in self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER)]:
```
boegel: reuse `optarch_parts` here, and I'd use `x` rather than `i` (`i` is usually an index)

easybuild/tools/options.py
```diff
@@ -715,6 +715,28 @@ def postprocess(self):
...
+        if self.options.optarch:
+            # optarch is in most cases equivalent to strlist, but changing the type breaks compatibility, so we hack it here
+            n_compilers = len(self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER))
+            
+            # if the format is not adequate we refuse continuing
+            if n_compilers != self.options.optarch.count(OPTARCH_OPTION_DELIMITER) and n_compilers > 1:
+                raise EasyBuildError("The optarch option has an incorrect syntax: %s" % self.options.optarch)
+            else:
+                # if there are options for different compilers, we set up a dict
+                if OPTARCH_OPTION_DELIMITER in self.options.optarch:
```
boegel: this should be `if OPTARCH_OPTION_DELIMITER in optarch_parts[0]`?

easybuild/tools/options.py
```diff
@@ -715,6 +715,28 @@ def postprocess(self):
...
+                    optarch_dict = dict()
+                    for compiler, compiler_opt in [i.split(OPTARCH_OPTION_DELIMITER) for i in self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER)]:
+                        if compiler in optarch_dict:
+                            raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s" % (compiler, self.options.optarch))
+                        else:
+                            optarch_dict[compiler] = compiler_opt
+                    self.options.optarch = optarch_dict 
+                # we do nothing and keep the string
+                else:
+                    pass
```
boegel: this `else: pass` is useless, unless you replace the `pass` with a meaningful log message

easybuild/tools/options.py
```diff
@@ -715,6 +715,28 @@ def postprocess(self):
         if self.options.last_log:
             self.options.terse = True
 
+        # make sure --optarch has a valid format
+        if self.options.optarch:
+            # optarch is in most cases equivalent to strlist, but changing the type breaks compatibility, so we hack it here
+            n_compilers = len(self.options.optarch.split(OPTARCH_TOOLCHAIN_DELIMITER))
+            
+            # if the format is not adequate we refuse continuing
+            if n_compilers != self.options.optarch.count(OPTARCH_OPTION_DELIMITER) and n_compilers > 1:
```
boegel: checking for `n_compilers > 1` first makes more sense to me

easybuild/tools/options.py
```diff
@@ -715,6 +715,28 @@ def postprocess(self):
...
+                raise EasyBuildError("The optarch option has an incorrect syntax: %s" % self.options.optarch)
+            else:
+                # if there are options for different compilers, we set up a dict
+                if OPTARCH_OPTION_DELIMITER in self.options.optarch:
+                    optarch_dict = dict()
```
boegel: replace `dict()` with `{}`

easybuild/tools/toolchain/compiler.py
```diff
@@ -27,6 +27,7 @@
 
 :author: Stijn De Weirdt (Ghent University)
 :author: Kenneth Hoste (Ghent University)
+:author: Damian Alvarez (Forschungszentrum Juelich GmbH)
```
boegel: also do this in `options.py`?

easybuild/tools/toolchain/compiler.py
```diff
@@ -41,6 +42,9 @@
 # by doing eb --optarch=GENERIC
 OPTARCH_GENERIC = 'GENERIC'
 
+# Characters that separate toolchains and options in --optarch
+OPTARCH_TOOLCHAIN_DELIMITER = ';'
+OPTARCH_OPTION_DELIMITER = ':'
```
boegel: I would shorten these a bit, e.g. `OPTARCH_SEP` for `;` and `OPTARCH_MAP_CHAR` for `:`

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +282,43 @@ def _set_optimal_architecture(self, default_optarch=None):
                                 (--optarch and --optarch=GENERIC still override this value)
         """
         optarch = None
+        use_generic = False
         # --optarch is specified with flags to use
-        if build_option('optarch') is not None and build_option('optarch') != OPTARCH_GENERIC:
-            optarch = build_option('optarch')
-        # --optarch=GENERIC
-        elif build_option('optarch') == OPTARCH_GENERIC:
+        if build_option('optarch') is not None:
```
boegel: just grab `build_option('optarch')` once:

```python
optarch = build_option('optarch')
if optarch is not None:
    ....
```

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +282,43 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+            # optarch has been validated as a simple string
+            if type(build_option('optarch')) is str:
```
boegel: please change to:

```python
if isinstance(optarch, basestring):
```

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +282,43 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+                # Option is generic, so mark it as to be set later
+                if build_option('optarch') == OPTARCH_GENERIC:
+                    use_generic = True
+                # No compilers specified, and option is not generic, so simply take what is there
+                else:
+                    optarch = build_option('optarch')
+
+            # optarch has been validated as complex string with multiple compilers and converted to a dictionary
+            elif type(build_option('optarch')) is dict:
```
boegel: please change to:

```python
if isinstance(optarch, dict):
```

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +282,43 @@ def _set_optimal_architecture(self, default_optarch=None):
...
-        elif build_option('optarch') == OPTARCH_GENERIC:
+        if build_option('optarch') is not None:
+            # optarch has been validated as a simple string
+            if type(build_option('optarch')) is str:
+                # Option is generic, so mark it as to be set later
+                if build_option('optarch') == OPTARCH_GENERIC:
+                    use_generic = True
+                # No compilers specified, and option is not generic, so simply take what is there
+                else:
+                    optarch = build_option('optarch')
```
boegel: you can drop this if you already did this above the first `if`

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +282,43 @@ def _set_optimal_architecture(self, default_optarch=None):
...
         # --optarch is specified with flags to use
-        if build_option('optarch') is not None and build_option('optarch') != OPTARCH_GENERIC:
-            optarch = build_option('optarch')
-        # --optarch=GENERIC
-        elif build_option('optarch') == OPTARCH_GENERIC:
+        if build_option('optarch') is not None:
+            # optarch has been validated as a simple string
+            if type(build_option('optarch')) is str:
+                # Option is generic, so mark it as to be set later
+                if build_option('optarch') == OPTARCH_GENERIC:
```
boegel:
```python
if optarch == OPTARCH_GENERIC:
```

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +282,43 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+            elif type(build_option('optarch')) is dict:
+                current_compiler = getattr(self, 'COMPILER_FAMILY', None)
+                if current_compiler in build_option('optarch'):
+                    if build_option('optarch')[current_compiler] == OPTARCH_GENERIC:
+                        use_generic = True
+                    else:
+                        optarch = build_option('optarch')[current_compiler]
+                # no option for this compiler
+                else:
+                    pass
```
boegel: this `pass` should be replaced with the `no optarch found` log message?

together with resetting `optarch` to `None`?

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +282,43 @@ def _set_optimal_architecture(self, default_optarch=None):
...
                 optarch = self.COMPILER_GENERIC_OPTION[(self.arch, self.cpu_family)]
-        # specified optarch default value
-        elif default_optarch:
+        # Specified optarch default value
+        elif default_optarch and optarch is None:
             optarch = default_optarch
-        # no --optarch specified, no default value specified
-        elif (self.arch, self.cpu_family) in (self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or []):
+        # no --optarch specified, no option found for the current compiler, and no default optarch
+        elif use_generic == False and optarch is None and (self.arch, self.cpu_family) in (self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or []):
```
boegel: you know `use_generic` is `False`, otherwise you could never reach here

easybuild/tools/toolchain/compiler.py
```diff
@@ -278,18 +282,43 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+                    else:
+                        optarch = build_option('optarch')[current_compiler]
+                # no option for this compiler
+                else:
+                    pass
+            else:
+                raise EasyBuildError("optarch is neither an string or a dict. This should never ever happen")
+
+            if optarch is None and use_generic == False:
+                self.log.info("_set_optimal_architecture: no optarch found for compiler %s. Ignoring option." % current_compiler)
```
boegel:
```python
self.log.info("...", current_compiler)
```

(replace `%` with`, `)

test/framework/options.py
```diff
@@ -2967,6 +2967,19 @@ def test_list_software(self):
         self.assertTrue(re.search('^\s+\* GCC v4.6.3: dummy', txt, re.M))
         self.assertFalse(re.search('gzip', txt, re.M))
 
+    def test_optarch(self):
```
boegel: call this `test_parse_optarch`?

test/framework/options.py
```diff
@@ -2967,6 +2967,19 @@ def test_list_software(self):
...
+        #
+        args = ['--optarch=Intel:something;GCC']
+        error_msg = "The optarch option has an incorrect syntax"
+        self.assertErrorRegex(EasyBuildError, error_msg, self.eb_main, args, raise_error=True)
+
+        args = ['--optarch=Intel:something;GCC:somethingelse;Intel:anothersomething']
+        error_msg = "The optarch option contains duplicated entries for compiler"
+        self.assertErrorRegex(EasyBuildError, error_msg, self.eb_main, args, raise_error=True)
+
+        # TODO: Check for correct parsing and dictionary creation
```
boegel: yes please :)

also check with a list of flags, `GENERIC`, ...

test/framework/toolchain.py
```diff
@@ -381,6 +381,49 @@ def test_optarch_aarch64_heuristic(self):
...
+                            for var in flag_vars:
+                                if enable:
+                                    self.assertTrue(flag in tc.get_variable(var), "optarch: True means '%s' in '%s'" % (flag, tc.get_variable(var)))
+                                else:
+                                    self.assertFalse(flag in tc.get_variable(var), "optarch: False means no '%s' in '%s'" % (flag, tc.get_variable(var)))
```
boegel: several lines are way too long here, partially due to the deep indent...

also, just do the `tc.get_variable(var)s` once, above the `if enable`...

easybuild/tools/toolchain/compiler.py
damianam: What is the policy here? Anybody that touches a file should be listed as an author, or is there a more specific criteria?

easybuild/tools/toolchain/compiler.py
damianam: Using that variable name screws up the logic of the rest of the method. But I get your point, I'll pick a different variable name.

easybuild/tools/toolchain/compiler.py
boegel: There's no strict policy, but if you made beyond trivial changes to a file it makes sense to include yourself as an author; it gives people a good idea on who to talk to in case of questions.

easybuild/tools/toolchain/compiler.py
boegel: Yeah, sorry for the confusion, but you get the idea... :)

damianam:
@boegel I addressed all your comments, even the one that altered the logic for the code avoiding calling too many times `build_option('optarch')`. The only thing missing as far as I can see is completing the tests of `options.py`

easybuild/tools/toolchain/compiler.py
damianam: Done

damianam:
@boegel I think the PR is ready for a final review. I am only wondering if there is a better way to test that EasyBuild doesn't generate exceptions when it shouldn't, than simply run it and see what happens. Is there some sort of `asssertNoError`?

test/framework/options.py
```diff
@@ -2968,6 +2968,51 @@ def test_list_software(self):
...
+        args = ['--optarch=Intel:something;GCC']
+        error_msg = "The optarch option has an incorrect syntax"
+        self.assertErrorRegex(EasyBuildError, error_msg, self.eb_main, args, raise_error=True)
+
+        args = ['--optarch=Intel:something;GCC:somethingelse;Intel:anothersomething']
+        error_msg = "The optarch option contains duplicated entries for compiler"
+        self.assertErrorRegex(EasyBuildError, error_msg, self.eb_main, args, raise_error=True)
+
+        # Check that it doesn't raise an exception when the input is correct
+        test_easyconfig = 'easyconfigs/test_ecs/t/toy/toy-0.0.eb'
```
boegel: @damianam please specify the *full* path here, usually like this:

```python
test_ecs = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'test_ecs')
toy_ec = os.path.join(test_ecs, 't', 'toy', 'toy-0.0.eb')
```

test/framework/options.py
```diff
@@ -2968,6 +2968,51 @@ def test_list_software(self):
...
+                ['--optarch=',test_easyconfig],
+                ['--optarch=something',test_easyconfig],
+                ['--optarch=GENERIC',test_easyconfig],
+                ['--optarch=Intel:something',test_easyconfig],
+                ['--optarch=Intel:something;GCC:somethingelse',test_easyconfig],
+                ['--optarch=Intel:GENERIC;GCC:somethingelse',test_easyconfig],
+                ['--optarch=Intel:;GCC:somethingelse;',test_easyconfig]
+                ]
+        for args in test_cases:
+            self.eb_main(args, raise_error=True)
```
boegel: @damianam this doesn't seem like a very useful test, you're not actually checking anything here (other than the lack of crashing)...

Why not test with an existing test easyconfig that uses `GCC` as a toolchain (which is a lot better than using the `dummy` toolchain where `<toolchain>.prepare()` is never called...), and also include `--extended-dry-run` as an option. That wait, you can check in the output whether you're seeing what you expect to see.

test/framework/toolchain.py
```diff
@@ -381,6 +382,56 @@ def test_optarch_aarch64_heuristic(self):
...
+                    if enable:
+                        self.assertTrue(flag in flags, "optarch: True means '%s' in '%s'" 
+                                % (flag, flags))
+                    else:
+                        self.assertFalse(flag in flags, "optarch: False means no '%s' in '%s'" 
+                                % (flag, flags))
+            # if flag is '' there is not much to check. A more robust approach
+            # would be to check against a blacklist, but that is probably not necessary
+            else:
+                pass
```
boegel: this doesn't do anything at all... :)

damianam: That's not true. It is not completely useless, it can be used as a bad example ;-P. I'll make a blacklist check then.

test/framework/options.py
damianam: Mmmmh, ok, but can I ask why?

test/framework/options.py
damianam: What you say is done in the tests of `toolchain.py`. There I check against `xFLAGS`.

What I wanted to check here is simply that there aren't any issues when the input is correct. Arguably, it doesn't make too much sense, since the affected code is tested right below, calling `options.postprocess()`, where the code is really validated.

The reason I included these apparently useless tests is because the tests below check against just the postprocessing of options, instead of a "full" run. If you think this is not necessary I can simply delete these lines, but a more thorough test is already done in `toolchain.py`, I see no point in repeating it.

test/framework/toolchain.py
damianam: Done

test/framework/options.py
boegel: Doing a 'full' run using `toy-0.0.eb` is meaningless imho, since `toolchain.prepare` will never be called since the `dummy` toolchain is used.

So, yes, I would drop this, since this test is mainly about how the `optarch` value is being parsed.

I would also change the tests where you check for the error on faulty output to not use `eb_main` but `options.postprocess()`?

test/framework/options.py
boegel: The main reason to use the full path to `toy-0.0.eb` is to make sure the test always runs, also if you're running it from a different location.

This test was failing on Travis with

```
EasyBuildError: "Can't find path /home/travis/easyconfigs/test_ecs/t/toy/toy-0.0.eb"
```

It probably worked for you because you were running the tests from the `test/framework` subdirectory?

damianam: Yep

test/framework/options.py
damianam: Ok, done.

easybuild/tools/toolchain/compiler.py
```diff
@@ -277,19 +281,40 @@ def _set_optimal_architecture(self, default_optarch=None):
...
-        optarch = None
+        use_generic = False
+        optarch= build_option('optarch') 
         # --optarch is specified with flags to use
-        if build_option('optarch') is not None and build_option('optarch') != OPTARCH_GENERIC:
-            optarch = build_option('optarch')
-        # --optarch=GENERIC
-        elif build_option('optarch') == OPTARCH_GENERIC:
+        if optarch is not None:
+            # optarch has been validated as a simple string
```
boegel: s/validated/parsed/

easybuild/tools/toolchain/compiler.py
```diff
@@ -277,19 +281,40 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+            if isinstance(optarch, basestring):
+                # Option is generic, so mark it as to be set later
```
boegel: hmm, dunno if this comment has any added value...

test/framework/options.py
```diff
@@ -3011,7 +3042,6 @@ def test_check_style(self):
         for pattern in patterns:
             self.assertTrue(re.search(pattern, stdout, re.M), "Pattern '%s' found in: %s" % (pattern, stdout))
 
-
```
boegel: style nitpicking: there's supposed to be a double empty line here (always between top-level class/functions)

test/framework/options.py
```diff
@@ -2968,6 +2968,37 @@ def test_list_software(self):
...
+        # Check the parsing itself
+        test_cases = [
+                ('',''),
+                ('xHost','xHost'),
+                ('GENERIC','GENERIC'),
+                ('Intel:xHost', {'Intel': 'xHost'}),
+                ('Intel:GENERIC', {'Intel': 'GENERIC'}),
+                ('Intel:xHost;GCC:march=x86-64 -mtune=generic', {'Intel': 'xHost', 'GCC': 'march=x86-64 -mtune=generic'}),
+                ('Intel:;GCC:march=x86-64 -mtune=generic', {'Intel': '', 'GCC': 'march=x86-64 -mtune=generic'}),
+                ]
```
boegel: style nitpicking: indent list elements with 4 spaces (no more), single space after `,`, left-align the closing `]`:

```python
test_cases = [
    ('', ''),
    ...
]
```

also, maybe use `gcc_flags = "march=x86-64 -mtune=generic"` to avoid too long lines?

```python
    ('Intel:;GCC:%s' % gcc_flags, {'Intel': '', 'GCC': gcc_flags}),
```

test/framework/options.py
```diff
@@ -2968,6 +2968,37 @@ def test_list_software(self):
...
+
+        for optarch_string, optarch_parsed in test_cases:
+            options.options.optarch = optarch_string
+            options.postprocess()
+            self.assertTrue(options.options.optarch == optarch_parsed, "Parsing of optarch: True means %s == %s" %
```
boegel: use `self.assertEqual`? then you don't need to provide a custom error message either, the output will show both values in case there's a mismatch

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
         self.assertEqual(tc.options.options_map['optarch'], 'mcpu=cortex-a72.cortex-a53')
         self.assertTrue('-mcpu=cortex-a72.cortex-a53' in os.environ['CFLAGS'])
         self.modtool.purge()
+  
+    def test_compiler_dependent_optarch(self):
+        """Test whether specifying optarch on a per compiler basis works."""
+        flag_vars = ['CFLAGS', 'CXXFLAGS', 'FCFLAGS', 'FFLAGS', 'F90FLAGS']
+        intel_options = [('intelflag','intelflag'),('GENERIC','xSSE2'),('','')]
+        gcc_options = [('gccflag','gccflag'),('GENERIC','march=x86-64 -mtune=generic'),('','')]
+        toolchains = [('iccifort','2011.13.367'),('GCC','4.7.2'),('PGI','16.7-GCC-5.4.0-2.26')]
```
boegel: more style nitpicking: space after `,` please :)

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+        enabled = [True, False]
+
+        i = product(intel_options, gcc_options, toolchains, enabled)
```
boegel: don't use single-letter variables outside of list comprehensions please; in particular not `i` which has the connotation of being an index

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+
+        for (intelflags, intel_expanded_flags), (gccflags, gcc_expanded_flags), (toolchain, toolchain_version), enable in i:
```
boegel: you could use `(intel_flags, intel_flags_exp), (gcc_flags, gcc_flags_exp), (tc, tcver)` to shorten this a bit?

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+            optarch_var = {} 
+            optarch_var['Intel'] = '%s' % intelflags
```
boegel: `'%s' % intelflags` is equivalent to just `intelflags`?

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+            optarch_var['GCC'] = '%s' % gccflags
```
boegel: same here, just use `gccflags`

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+            if toolchain == 'iccifort':
+                flag = intel_expanded_flags
+            elif toolchain == 'GCC':
+                flag = gcc_expanded_flags 
+            else: # PGI as an example of compiler not set
+                # default optarch flag, should be the same as the one in
+                # tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(tc.arch,tc.cpu_family)]
+                flag = ''
+            
+            flags = tc.options.options_map['optarch']
```
boegel: hmm, rename to `optarch_flags`?

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+        for (intelflags, intel_expanded_flags), (gccflags, gcc_expanded_flags), (toolchain, toolchain_version), enable in i:
+            optarch_var = {} 
+            optarch_var['Intel'] = '%s' % intelflags
+            optarch_var['GCC'] = '%s' % gccflags
+            build_options = {'optarch': optarch_var}
+            init_config(build_options=build_options)
+            tc = self.get_toolchain(toolchain, version=toolchain_version)
+            tc.set_options({'optarch': enable})
+            tc.prepare()
+            flag = None
```
boegel: this should be named `flags`?

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+            elif toolchain == 'GCC':
+                flag = gcc_expanded_flags 
+            else: # PGI as an example of compiler not set
+                # default optarch flag, should be the same as the one in
+                # tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(tc.arch,tc.cpu_family)]
+                flag = ''
+            
+            flags = tc.options.options_map['optarch']
+
+            self.assertTrue(flag == flags, "optarch: True means '%s' == '%s'" % (flag, flags))
```
boegel: please use `self.assertEquals`

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+            # Also check that it is correctly passed to xFLAGS, honoring 'enable'
+            if flag == '':
+                blacklist = [
+                       intel_options[0][1],
+                       intel_options[1][1],
+                       gcc_options[0][1],
+                       gcc_options[1][1],
+                       'xHost', # default optimal for Intel
+                       'march=native', # default optimal for GCC
+                       ]
```
boegel: please fix indent + left-align `]`

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+
+            self.assertTrue(flag == flags, "optarch: True means '%s' == '%s'" % (flag, flags))
+            
+            # Also check that it is correctly passed to xFLAGS, honoring 'enable'
+            if flag == '':
+                blacklist = [
+                       intel_options[0][1],
+                       intel_options[1][1],
+                       gcc_options[0][1],
+                       gcc_options[1][1],
```
boegel: hmm, would it be clearer to just hardcode those strings here?

hardcoding is fine in tests if it makes things clearer

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+            else:
+                blacklist = [flag]
+
+            for var in flag_vars:
+                 flags = tc.get_variable(var)
+                
+                 # Check that the correct flags are there
+                 if enable and flag != '':
+                     self.assertTrue(flag in flags, "optarch: True means '%s' in '%s'" 
+                             % (flag, flags))
```
boegel: I'd prefer:

```python
error_msg = "optarch: True means '%s' in '%s'" % (flags, optarch_flags)
self.assertTrue(flags in optarch_flags, error_msg)
```

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+
+                 # Check that there aren't unexpected flags
+                 else:
+                     for blacklisted_flag in blacklist:
+                         self.assertFalse(blacklisted_flag in flags, "optarch: False means no '%s' in '%s'" 
+                                 % (blacklisted_flag, flags))
```
boegel: same here:

```python
error_msg = "optarch: False means no '%s' in '%s'" % (blacklisted_flag, flags)
self.assertFalse(blacklisted_flag in optarch_flags, error_msg)
```

test/framework/toolchain.py
```diff
@@ -381,6 +382,68 @@ def test_optarch_aarch64_heuristic(self):
...
+
+            for var in flag_vars:
+                 flags = tc.get_variable(var)
+                
+                 # Check that the correct flags are there
+                 if enable and flag != '':
+                     self.assertTrue(flag in flags, "optarch: True means '%s' in '%s'" 
+                             % (flag, flags))
+
+                 # Check that there aren't unexpected flags
```
boegel: `aren't any`

easybuild/tools/options.py
```diff
@@ -724,6 +730,31 @@ def postprocess(self):
...
+    def _postprocess_optarch(self):
+        """Postprocess --optarch option."""
+        optarch_parts = self.options.optarch.split(OPTARCH_SEP)
+        n_compilers = len(optarch_parts)
+        
+        # if the format is not adequate we refuse continuing
+        if n_compilers > 1 and n_compilers != self.options.optarch.count(OPTARCH_MAP_CHAR):
+            raise EasyBuildError("The optarch option has an incorrect syntax: %s" % self.options.optarch)
+        else:
+            logger = fancylogger.getLogger()
```
damianam: Sounds reasonable, but we should be consistent. There are more loggers being instantiated in that file.

easybuild/tools/options.py
```diff
@@ -724,6 +730,31 @@ def postprocess(self):
...
+            # if there are options for different compilers, we set up a dict
+            if OPTARCH_MAP_CHAR in optarch_parts[0]:
+                optarch_dict = {}
+                for compiler, compiler_opt in [x.split(OPTARCH_MAP_CHAR) for x in optarch_parts]:
+                    if compiler in optarch_dict:
+                        raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s"
```
damianam: Same comment as above. I am ok with the change, but we should be consistent (to be done in a different PR?)

test/framework/options.py
damianam: That's a bit weird to treat gcc flags and intel flags differently, but ok.

test/framework/options.py
damianam: Makes sense, but again, we are missing consistency. I coded it that way following the style of similar code in the same file.

test/framework/toolchain.py
damianam: Or iterator, which is what this thing was. But that's fine, I am happy with `test_cases`

test/framework/toolchain.py
damianam: That was silly, wasn't it?

test/framework/toolchain.py
damianam: The renaming of variables in that section might screw up the code, but that's fine, I renamed them using more sensible names.

test/framework/toolchain.py
damianam: `GENERIC_CONSISTENCY_COMMENT`

test/framework/toolchain.py
damianam: I think hardcoding will create more troubles down the road. If somebody needs to update this test they'll have to update that in various places........ which is plain annoying. Sure, that will be more readable, but less maintainable, and the only time somebody will read this test is to maintain it, so........ I vote for leaving it as it is ;-P

easybuild/tools/options.py
boegel: In a method (where `self` is an argument), you should use `self.log`.

In a function (no `self` in scope, like in `process_software_build_specs`), you may need to create a separate logger.

In this case, there's actually a global `_log` available, we should be using that consistently where a `self.log` is not available, e.g. in `process_software_build_specs`

easybuild/tools/options.py
boegel: Yeah, let's not clutter this PR with pure style changes in code we're not touching.

Feel free to tackle this in a separate PR, if you're up for it.

test/framework/options.py
boegel: It's more about a large set of flags that include spaces vs a single flag rather than about GCC vs Intel...

test/framework/options.py
boegel: Using `assertTrue` with `==` is a bit weird, so we should change that (but also in a separate PR).

test/framework/toolchain.py
boegel: It happens to the best of us. ;-)

test/framework/toolchain.py
boegel: Looks OK after review, worst thing that can happen is that the tests start failing?

test/framework/toolchain.py
boegel: Hmm, I'm getting `undefined fake constant` here...

test/framework/toolchain.py
boegel: I consider maintainability of tests less of an issue...

If you changes the values in `intel_options` and `gcc_options`, it's good to know where those are (also) used. Hardcoding the strings here forces you to also make the corresponding changes here, and makes you think about whether making the changes actually makes sense.

But fine, I don't feel too strongly about it, we're less strict in the tests.

easybuild/tools/options.py
```diff
@@ -724,6 +729,29 @@ def postprocess(self):
             self.show_config()
             cleanup_and_exit(self.tmpdir)
 
+    def _postprocess_optarch(self):
+        """Postprocess --optarch option."""
+        optarch_parts = self.options.optarch.split(OPTARCH_SEP)
+        
+        # we expect to find a ':' in every entry in optarch, in case optarch is specified on a per-compiler basis
+        if len(optarch_parts) > 1 and not all(OPTARCH_MAP_CHAR in p for p in optarch_parts):
```
boegel: @damianam I gave this some more thought, and I actually think we need a more complex check to *really* validate this... We should check that there *exactly* one `OPTARCH_MAP_CHAR` in each part, now we may have multiple, which will screw things up later where we use `.split(OPTARCH_MAP_CHAR)`...

easybuild/tools/options.py
```diff
@@ -724,6 +729,29 @@ def postprocess(self):
...
+            raise EasyBuildError("The optarch option has an incorrect syntax: %s", self.options.optarch)
+        else:
+            # if there are options for different compilers, we set up a dict
+            if OPTARCH_MAP_CHAR in optarch_parts[0]:
+                optarch_dict = {}
+                for compiler, compiler_opt in [x.split(OPTARCH_MAP_CHAR) for x in optarch_parts]:
+                    if compiler in optarch_dict:
+                        raise EasyBuildError("The optarch option contains duplicated entries for compiler %s: %s",
+                                compiler, self.options.optarch)
```
boegel: @damianam minor style issue: please indent further to align the `c` with the first `"` above...

easybuild/tools/toolchain/compiler.py
```diff
@@ -277,19 +281,39 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+                        use_generic = True
+                    else:
+                        optarch = optarch[current_compiler]
+                # no option for this compiler
+                else:
+                    optarch = None
+                    self.log.info("_set_optimal_architecture: no optarch found for compiler %s. Ignoring option.", 
+                            current_compiler)
+            else:
+                raise EasyBuildError("optarch is neither an string or a dict. This should never ever happen")
```
boegel: @damianam we should show what's in `optarch` to make the error clearer? (although it's unlikely this will ever happen)

test/framework/toolchain.py
damianam: You are missing one import:

`from damian.awesomeness import GENERIC_CONSISTENCY_COMMENT`

easybuild/tools/options.py
```diff
@@ -724,6 +729,30 @@ def postprocess(self):
             self.show_config()
             cleanup_and_exit(self.tmpdir)
 
+    def _postprocess_optarch(self):
+        """Postprocess --optarch option."""
+        optarch_parts = self.options.optarch.split(OPTARCH_SEP)
+        
+        # we expect to find a ':' in every entry in optarch, in case optarch is specified on a per-compiler basis
+        if (len(optarch_parts) > 1 and not all(p.count(OPTARCH_MAP_CHAR) == 1 for p in optarch_parts)) or \
+                (len(optarch_parts) == 1 and optarch_parts[0].count(OPTARCH_MAP_CHAR) > 1):
```
boegel: wait, this is just

```python
if not all(p.count(OPTARCH_MAP_CHAR) == 1 for p in optarch_parts)):
```

is it not?

damianam: No. If you pass a simple string you'll raise and exception and break backwards compatibility, and force everyone to specify compiler. The count of `OPTARCH_MAP_CHAR` can be 0 if no compiler is specified

boegel: OK, I missed that bit, but this is very long and messy...

How about:

```python
n_parts = len(optarch_parts)
map_char_cnts = [p.count(OPTARCH_MAP_CHAR) for p in optarch_parts]
if (n_parts > 1 and any(c != 1 for c in map_char_cnts)) or (n_parts == 1 and map_char_cnts[0] > 1):
```

damianam: Looks more messy to me..... The `if` is shorter, sure, but everything else seems unnecessarly complicated to me. But I don't really care....

boegel: Well, the variable names are key in my approach, maybe they could be improved.

In your version there's a lot of copy-pasting of code snippets, which makes it hard to parse imho...

damianam: Very well then. Done

boegel:
Going in, thanks @damianam!

Please also look into a documentation update @ http://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html

ocaisa:
This has got to be the longest conversation for a PR EasyBuild has ever had?

damianam:
We are social people, we like to talk. That has to be reason, right?

PR #2074, opened 2017-01-05
implement get_cpu_features function in systemtools

boegel:
This could be used to implement an FFTW easyblock that auto-detects which configure flags can be specified to optimise performance (e.g. `--enable-avx`, `--enable-avx2`, etc.), rather than hardcoding configure flags in the FFTW easyconfigs (which is an issue already, since `--enable-sse2` doesn't work on Linux/POWER for example).

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -418,7 +418,6 @@ def get_cpu_features():
             _log.debug("Trying to determine CPU features on Darwin via cmd '%s'", cmd)
             out, ec = run_cmd(cmd, force_in_dry_run=True)
             if ec == 0:
-                # returns clock frequency in cycles/sec, but we want MHz
                 cpu_feat.extend(out.strip().lower().split())
 
         cpu_feat.sort()
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -385,6 +385,50 @@ def get_cpu_speed():
     return cpu_freq
 
 
+def get_cpu_features():
+    """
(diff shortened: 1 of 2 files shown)


easybuild/tools/systemtools.py
```diff
@@ -385,6 +385,50 @@ def get_cpu_speed():
...
+        else:
+            _log.debug("%s not found to determine CPU features", PROC_CPUINFO_FP)
+
+    elif os_type == DARWIN:
+        for feature_set in ['extfeatures', 'features', 'leaf7_features']:
+            cmd = "sysctl -n machdep.cpu.%s" % feature_set
+            _log.debug("Trying to determine CPU features on Darwin via cmd '%s'", cmd)
+            out, ec = run_cmd(cmd, force_in_dry_run=True)
+            if ec == 0:
+                # returns clock frequency in cycles/sec, but we want MHz
```
JensTimmerman: this doesn't look right at all

easybuild/tools/systemtools.py
```diff
@@ -385,6 +385,50 @@ def get_cpu_speed():
...
+                cpu_feat.extend(out.strip().lower().split())
```
JensTimmerman: you probably need to strip out the `xxx:` ?

easybuild/tools/systemtools.py
boegel: woops, will fix, thanks

JensTimmerman:
lgtm

boegel:

PR #2075, opened 2017-01-05
Fix docstring in toolchain class

geimer:


boegel:
Nice catch @geimer. :)

PR #2078, opened 2017-01-07
also detect VSX support for POWER

boegel:
motivated by https://github.com/hpcugent/easybuild-easyblocks/pull/1083

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -406,7 +406,11 @@ def get_cpu_features():
                 # for Linux@POWER systems, no flags/features are listed, but we can check for Altivec
                 cpu_altivec_regex = re.compile("^cpu\s*:.*altivec supported", re.M)
                 if cpu_altivec_regex.search(proc_cpuinfo):
-                    cpu_feat = ['altivec']
+                    cpu_feat.append('altivec')
+                # VSX is supported since POWER7
+                cpu_power7_regex = re.compile("^cpu\s*:.*POWER(7|8|9)", re.M)
+                if cpu_power7_regex.search(proc_cpuinfo):
+                    cpu_feat.append('vsx')
             else:
                 _log.debug("Failed to determine CPU features from %s", PROC_CPUINFO_FP)
         else:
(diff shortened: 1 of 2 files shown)


boegel:
@JackPerdue, @geimer: please review?

JackPerdue:
Looks good here.

easybuild/tools/systemtools.py
```diff
@@ -406,7 +406,11 @@ def get_cpu_features():
                 # for Linux@POWER systems, no flags/features are listed, but we can check for Altivec
                 cpu_altivec_regex = re.compile("^cpu\s*:.*altivec supported", re.M)
                 if cpu_altivec_regex.search(proc_cpuinfo):
-                    cpu_feat = ['altivec']
+                    cpu_feat.append('altivec')
+                # VSX is supported since POWER7
+                cpu_power7_regex = re.compile("^cpu\s*:.*POWER(7|8|9)", re.M)
```
geimer: What about extracting the trailing number and then compare >= 7?  Seems more future-proof (although even POWER9 isn't available yet).

boegel: That's trickier... Can I assume there will always be a number after '`POWER`'?

geimer: For the "real" IBM POWER systems up until now, I believe yes. But I have no clue how things look like for other POWER-based designs (remember, there is the OpenPOWER initiative). Though other things in `systemtools` may break there as well...

boegel: @geimer I prefer sticking to what we have now. Who knows what they come up with after POWER9, it may not even be numbers anymore...

JackPerdue: I liked it.... made sense to me (for now)...  The following (which seems to build nicely though I haven't tested it., uses the same logic:

http://siliconslick.com/easybuild/easyconfigs/curie/curie/FFTW/FFTW-3.3.5-gompi-2016D.eb

boegel:
OK, going in then, thanks for the feedback @geimer and @JackPerdue!

PR #2083, opened 2017-01-13
Spoof HTTP request header with empty agent

vanzod:
A quick workaround to avoid 403 errors from HTTP requests to websites that do not allow connections from Python-urllib. This problem affects the installation of Anaconda when the installation shell script is not already present locally and needs to be downloaded.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -345,12 +345,11 @@ def download_file(filename, url, path, forced=False):
     max_attempts = 3
     attempt_cnt = 0
 
-    # use blank HTTP header
-    header={'User-Agent':'EasyBuild'}
+    # use custom HTTP header
+    url_req = urllib2.Request(url, headers={'User-Agent': 'EasyBuild'})
     
     while not downloaded and attempt_cnt < max_attempts:
         try:
-            url_req = urllib2.Request(url, headers=header)
             # urllib2 does the right thing for http proxy setups, urllib does not!
             url_fd = urllib2.urlopen(url_req, timeout=timeout)
             _log.debug('response code for given url %s: %s' % (url, url_fd.getcode()))
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -34,6 +34,7 @@ Set of file tools.
 :author: Ward Poelmans (Ghent University)
 :author: Fotis Georgatos (Uni.Lu, NTUA)
 :author: Sotiris Fragkiskos (NTUA, CERN)
+:author: Davide Vanzo (ACCRE, Vanderbilt University)
 """
 import fileinput
 import glob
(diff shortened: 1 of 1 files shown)


boegel:
@vanzod tests failed because of GitHub trouble it seems, just retriggered them

easybuild/tools/filetools.py
```diff
@@ -343,10 +344,15 @@ def download_file(filename, url, path, forced=False):
     downloaded = False
     max_attempts = 3
     attempt_cnt = 0
+
+    # use blank HTTP header
+    header={'User-Agent':''}
```
boegel: Why not be fair and use `easybuild` as user agent?

pescobar:
related: https://github.com/hpcugent/easybuild-framework/issues/2021

easybuild/tools/filetools.py
vanzod: No actual reason other than waiting for your reaction.

easybuild/tools/filetools.py
```diff
@@ -343,10 +344,15 @@ def download_file(filename, url, path, forced=False):
     downloaded = False
     max_attempts = 3
     attempt_cnt = 0
+
+    # use blank HTTP header
```
boegel: @vanzod this is not exactly blank anymore... :)

easybuild/tools/filetools.py
```diff
@@ -343,10 +344,15 @@ def download_file(filename, url, path, forced=False):
...
+    header={'User-Agent':'EasyBuild'}
+    
     while not downloaded and attempt_cnt < max_attempts:
         try:
+            url_req = urllib2.Request(url, headers=header)
```
boegel: @vanzod lets move this out of the while loop, only need to be constructed once, right? Also, you can merge this line with the one you addd above:

```python
url_req = urllib2.Request(url, headers={'User-Agent': 'EasyBuild'})

while ...
```

easybuild/tools/filetools.py
```diff
@@ -343,10 +344,15 @@ def download_file(filename, url, path, forced=False):
     downloaded = False
     max_attempts = 3
     attempt_cnt = 0
+
+    # use blank HTTP header
+    header={'User-Agent':'EasyBuild'}
```
boegel: style nitpicking: space after `:` (and spaces around `=`, but irrelevant with comment below)

pescobar:
it works for me:

```
[escobar@login18 ~]$ eb Anaconda2-4.0.0.eb
== temporary log file in case of crash /scratch/eb-IrfAkN/easybuild-AgIwYq.log
== processing EasyBuild easyconfig /scicore/home/scicore/escobar/.local/easybuild/software/EasyBuild/3.0.2/lib/python2.6/site-packages/easybuild_easyconfigs-3.0.2-py2.6.egg/easybuild/easyconfigs/a/Anaconda2/Anaconda2-4.0.0.eb
== building and installing Anaconda2/4.0.0...
== fetching files...
.
.
.
`````

boegel:
Looks good, thanks for the fix @vanzod and @pescobar for verifying the fix!

PR #2096, opened 2017-01-25
add support for using an 'intel' toolchain that only loads an 'intel-psxe' module (WIP)

boegel:
This change allows to use an `intel` module that was installed using a single `intel-psxe` module rather than using separate installations for `icc`, `ifort`, `impi` and `imkl`.

I implemented this by introducing a generic concept or 'alternate toolchain definitions', since the need for this may also pop up later for other toolchains.

We *may* be going forward with this approach for `intel/2017a`, that's not fully clear yet.

TODO:

* verify that this works as expected outside of the tests
* look into effect of using an `intel` toolchain that uses `intel-psxe` with `HierarchicalMNS` (an update will be required there too)
* maybe also verify that the `intel-psxe` module defines `$EBROOTICC` & co somewhere, since this is assumed in some places now when the `intel` toolchain is used...

### DIFF
--- a/easybuild/toolchains/intel.py
+++ b/easybuild/toolchains/intel.py
@@ -42,3 +42,20 @@ class Intel(Iimpi, IntelMKL, IntelFFTW):
     """
     NAME = 'intel'
     SUBTOOLCHAIN = Iimpi.NAME
+
+    def alt_definitions(self):
+        """
+        Return list of alternate definitions for the 'intel' toolchain.
+        """
+        # Intel compilers, MPI and MKL can also be installed as a whole as Intel Parallel Studio XE (PSXE)
+        intel_psxe_modname = 'intel-psxe'
+        intel_psxe = {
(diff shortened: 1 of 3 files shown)


boegel:
@ocaisa please review?

boegel:
This is more work than I figured it would be, so this isn't going to be ready for EB v3.1.0 or the `2017a` update of the common toolchains...

geimer:
@boegel Would this 'alternate toolchain definitions' concept also allow to install `iccifort` as a bundle with components rather than dependencies?  With this my dream of having just a single module for the Intel C/C++/Fortran compilers could become true...

boegel:
@geimer Yes, definitely... You could define an `iccifort` bundle, but you'd need to make sure that it defines both `$EBROOTICC` and `$EBROOTIFORT` (same for `$EBVERSION*`).

PR #2110, opened 2017-02-09
add support for marking installed module file as new default version

victorusu:
This commit adds support to define a default module version on command
line.

The command line is:
`--set-default-module `

And both Tcl and Lua modules are supported.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2081,10 +2081,11 @@ class EasyBlock(object):
 
         sets the default module version except if we are in dry run.
         """
-        mod_filepath = self.module_generator.get_module_filepath()
-        mod_folderpath = os.path.dirname(mod_filepath)
+        mod_folderpath = os.path.dirname(self.module_generator.get_module_filepath())
 
-        if not self.dry_run:
+        if self.dry_run:
+            dry_run_msg("Marked %s v%s as default version" % (self.name, self.version))
+        else:
             self.module_generator.set_as_default(mod_folderpath, self.version)
 
     def cleanup_step(self):
@@ -2150,8 +2151,6 @@ class EasyBlock(object):
                 self.dry_run_msg("Generating module file %s, with contents:\n", mod_filepath)
                 for line in txt.split('\n'):
                     self.dry_run_msg(' ' * 4 + line)
-
-            self.dry_run_msg("Generating file to set default module version to %s\n", self.version)
         else:
             write_file(mod_filepath, txt)
             self.log.info("Module file %s written: %s", mod_filepath, txt)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2075,45 +2075,17 @@ class EasyBlock(object):
         else:
             self.log.debug("Sanity check passed!")
 
-    def _set_module_as_default(self, fake=False):
+    def _set_module_as_default(self):
+        """
+        Defining default module Version
 
-        if build_option('set_as_default'):
-            # version_module_generator = module_generator(self, fake=True)
-            # modpath = version_module_generator.prepare(fake=fake)
+        sets the default module version except if we are in dry run.
+        """
+        mod_filepath = self.module_generator.get_module_filepath()
+        mod_folderpath = os.path.dirname(mod_filepath)
 
-            mod_filepath = self.module_generator.get_module_filepath(fake=fake)
-            mod_folderpath = os.path.dirname(mod_filepath)
-            mod_filepath = os.path.join(mod_folderpath, '.version')
-
-            if isinstance(self.module_generator, ModuleGeneratorLua):
-                mod_filepath = os.path.join(mod_folderpath, 'default')
-            elif isinstance(self.module_generator, ModuleGeneratorTcl):
-                mod_filepath = os.path.join(mod_folderpath, '.version')
-
-            if self.dry_run:
-                # only report generating actual module file during dry run, don't mention temporary module files
-                if not fake:
-                    self.dry_run_msg("Generating default module file %s\n", mod_filepath)
-            else:
-                self.module_generator.set_as_default(mod_folderpath, self.version)
-                self.log.info("Module default version file %s written", mod_filepath)
-
-                # invalidate relevant 'module avail'/'module show' cache entries
-                modpath = self.module_generator.get_modules_path(fake=fake)
-                # consider both paths: for short module name, and subdir indicated by long module name
-                paths = [modpath]
-                if self.mod_subdir:
-                    paths.append(os.path.join(modpath, self.mod_subdir))
-
-                for path in paths:
-                    invalidate_module_caches_for(path)
-
-                # only update after generating final module file
-                if not fake:
-                    self.modules_tool.update()
-
-                if not fake:
-                    self.make_devel_module()
+        if not self.dry_run:
+            self.module_generator.set_as_default(mod_folderpath, self.version)
 
     def cleanup_step(self):
         """
@@ -2172,8 +2144,6 @@ class EasyBlock(object):
         txt += self.make_module_extra()
         txt += self.make_module_footer()
 
-
-
         if self.dry_run:
             # only report generating actual module file during dry run, don't mention temporary module files
             if not fake:
@@ -2181,6 +2151,7 @@ class EasyBlock(object):
                 for line in txt.split('\n'):
                     self.dry_run_msg(' ' * 4 + line)
 
+            self.dry_run_msg("Generating file to set default module version to %s\n", self.version)
         else:
             write_file(mod_filepath, txt)
             self.log.info("Module file %s written: %s", mod_filepath, txt)
@@ -2205,7 +2176,8 @@ class EasyBlock(object):
             if not fake:
                 self.make_devel_module()
 
-        self._set_module_as_default(fake=fake)
+        if build_option('set_as_default'):
+            self._set_module_as_default()
 
         return modpath
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2075,6 +2075,46 @@ class EasyBlock(object):
         else:
             self.log.debug("Sanity check passed!")
 
+    def _set_module_as_default(self, fake=False):
+
+        if build_option('set_as_default'):
+            # version_module_generator = module_generator(self, fake=True)
+            # modpath = version_module_generator.prepare(fake=fake)
+
+            mod_filepath = self.module_generator.get_module_filepath(fake=fake)
+            mod_folderpath = os.path.dirname(mod_filepath)
+            mod_filepath = os.path.join(mod_folderpath, '.version')
+
+            if isinstance(self.module_generator, ModuleGeneratorLua):
+                mod_filepath = os.path.join(mod_folderpath, 'default')
+            elif isinstance(self.module_generator, ModuleGeneratorTcl):
+                mod_filepath = os.path.join(mod_folderpath, '.version')
+
+            if self.dry_run:
+                # only report generating actual module file during dry run, don't mention temporary module files
+                if not fake:
+                    self.dry_run_msg("Generating default module file %s\n", mod_filepath)
+            else:
+                self.module_generator.set_as_default(mod_folderpath, self.version)
+                self.log.info("Module default version file %s written", mod_filepath)
+
+                # invalidate relevant 'module avail'/'module show' cache entries
+                modpath = self.module_generator.get_modules_path(fake=fake)
+                # consider both paths: for short module name, and subdir indicated by long module name
+                paths = [modpath]
+                if self.mod_subdir:
+                    paths.append(os.path.join(modpath, self.mod_subdir))
+
+                for path in paths:
+                    invalidate_module_caches_for(path)
+
+                # only update after generating final module file
+                if not fake:
+                    self.modules_tool.update()
+
+                if not fake:
+                    self.make_devel_module()
+
     def cleanup_step(self):
         """
         Cleanup leftover mess: remove/clean build directory
@@ -2132,6 +2172,8 @@ class EasyBlock(object):
         txt += self.make_module_extra()
         txt += self.make_module_footer()
 
+
+
         if self.dry_run:
             # only report generating actual module file during dry run, don't mention temporary module files
             if not fake:
@@ -2162,6 +2204,8 @@ class EasyBlock(object):
 
             if not fake:
                 self.make_devel_module()
+
+        self._set_module_as_default(fake=fake)
 
         return modpath
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2175,7 +2175,7 @@ class EasyBlock(object):
             if not fake:
                 self.make_devel_module()
 
-        if build_option('set_as_default'):
+        if build_option('set_default_module'):
             self._set_module_as_default()
 
         return modpath
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -177,21 +177,21 @@ def readlink(symlink_path):
 
     :param symlink_path: symlink file path
     """
-
-    # note: we can't use try-except-finally, because Python 2.4 doesn't support it as a single block
     try:
         target_symlink_path = os.readlink(symlink_path)
-    except IOError, err:
+    except OSError as err:
         raise EasyBuildError("Failed to get target of symlink: %s", symlink_path)
-
-    if build_option('extended_dry_run'):
-        dry_run_msg("symlink file %s points to: %s", symlink_path, target_symlink_path)
 
     return target_symlink_path
 
 
 def symlink(source_path, symlink_path):
     """Create a symlink at the specified path to the given path."""
+
+    if build_option('extended_dry_run'):
+        dry_run_msg("Symlinked file %s to %s" % (source_path, symlink_path), silent=build_option('silent'))
+        return
+
     try:
         os.symlink(os.path.abspath(source_path), symlink_path)
         _log.info("Symlinked %s to %s", source_path, symlink_path)
@@ -1072,15 +1072,6 @@ def weld_paths(path1, path2):
             path2_head = None
 
     return os.path.join(path1, path2_tail)
-
-
-def symlink(source_path, symlink_path):
-    """Create a symlink at the specified path to the given path."""
-    try:
-        os.symlink(os.path.abspath(source_path), symlink_path)
-        _log.info("Symlinked %s to %s", source_path, symlink_path)
-    except OSError as err:
-        raise EasyBuildError("Symlinking %s to %s failed: %s", source_path, symlink_path, err)
 
 
 def path_matches(path, paths):
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -188,9 +188,9 @@ def readlink(symlink_path):
 def symlink(source_path, symlink_path):
     """Create a symlink at the specified path to the given path."""
 
-    if build_option('extended_dry_run', default=False):
-        dry_run_msg("Symlinked file %s to %s" % (source_path, symlink_path), silent=build_option('silent'))
-        return
+    # if build_option('extended_dry_run', default=False):
+    #     dry_run_msg("Symlinked file %s to %s" % (source_path, symlink_path), silent=build_option('silent'))
+    #     return
 
     try:
         os.symlink(os.path.abspath(source_path), symlink_path)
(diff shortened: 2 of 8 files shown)


pforai:
LGTM

easybuild/framework/easyblock.py
```diff
@@ -2075,6 +2075,18 @@ def _sanity_check_step(self, custom_paths=None, custom_commands=None, extension=
...
             self.log.debug("Sanity check passed!")
 
+    def _set_module_as_default(self):
+        """
+        Defining default module Version
+
+        sets the default module version except if we are in dry run.
+        """
+        mod_filepath = self.module_generator.get_module_filepath()
+        mod_folderpath = os.path.dirname(mod_filepath)
```
boegel: why not combine these lines together, `mod_filepath` is only used in one place?

easybuild/framework/easyblock.py
```diff
@@ -2075,6 +2075,18 @@ def _sanity_check_step(self, custom_paths=None, custom_commands=None, extension=
...
+
+        if not self.dry_run:
+            self.module_generator.set_as_default(mod_folderpath, self.version)
```
boegel: maybe emit a meaningful message in the case of dry run?

```python
if self.dry_run:
    dry_run_msg("Marked %s v%s as default version" % (self.name, self.version))
else:
    self.module_generator.set_as_default(mod_folderpath, self.version)

File: easybuild/framework/easyblock.py [hunk 3]
```diff
@@ -2139,6 +2151,7 @@ def make_module_step(self, fake=False):
                 for line in txt.split('\n'):
                     self.dry_run_msg(' ' * 4 + line)
 
+            self.dry_run_msg("Generating file to set default module version to %s\n", self.version)
```
boegel: hmm, drop this, see dry run msg in `_set_module_as_default`?

easybuild/tools/filetools.py
```diff
@@ -171,6 +171,34 @@ def write_file(path, txt, append=False, forced=False):
...
 
 
+def readlink(symlink_path):
+    """
+    Read symlink target at the specified path to the given path.
+
+    :param symlink_path: symlink file path
+    """
+
+    # note: we can't use try-except-finally, because Python 2.4 doesn't support it as a single block
```
boegel: minimal required version is Python 2.6 now, so you can use `try`/`except`/`finally`

but, not relevant here, so please just drop this comment

easybuild/tools/filetools.py
```diff
@@ -171,6 +171,34 @@ def write_file(path, txt, append=False, forced=False):
...
+    try:
+        target_symlink_path = os.readlink(symlink_path)
+    except IOError, err:
```
boegel: should be `OSError` rather than `IOError`?

also, use 

```python
except OSError as err:

File: easybuild/tools/filetools.py [hunk 6]
```diff
@@ -171,6 +171,34 @@ def write_file(path, txt, append=False, forced=False):
...
+        raise EasyBuildError("Failed to get target of symlink: %s", symlink_path)
+
+    if build_option('extended_dry_run'):
+        dry_run_msg("symlink file %s points to: %s", symlink_path, target_symlink_path)
```
boegel: not sure if a dry run message is useful here (no action being taken)

easybuild/tools/filetools.py
```diff
@@ -171,6 +171,34 @@ def write_file(path, txt, append=False, forced=False):
...
+
+    return target_symlink_path
+
+
+def symlink(source_path, symlink_path):
+    """Create a symlink at the specified path to the given path."""
+    try:
+        os.symlink(os.path.abspath(source_path), symlink_path)
```
boegel: here a dry run message *would* be useful?

easybuild/tools/filetools.py
```diff
@@ -171,6 +171,34 @@ def write_file(path, txt, append=False, forced=False):
...
+    except IOError, err:
+        raise EasyBuildError("Failed to get target of symlink: %s", symlink_path)
+
+    if build_option('extended_dry_run'):
+        dry_run_msg("symlink file %s points to: %s", symlink_path, target_symlink_path)
+
+    return target_symlink_path
+
+
+def symlink(source_path, symlink_path):
```
boegel: this already exists in the `filetools` module?

easybuild/tools/module_generator.py
```diff
@@ -453,6 +456,21 @@ def getenv_cmd(self, envvar):
         return '$env(%s)' % envvar
 
 
+    def set_as_default(self, module_folder_path, module_version):
+        """
+        Create .version file inside the package module folder in order to set the default version
+        for TMod
+        :param module_folder_path: module folder path, e.g.  /easybuild/modules/all/Bison
+        :param module_version: 3.0.4
```
boegel: ok to give an example, but also explain what it is?

easybuild/tools/module_generator.py
```diff
@@ -453,6 +456,21 @@ def getenv_cmd(self, envvar):
...
+        """
+        txt = self.MODULE_SHEBANG + '\n'
+        txt += 'set ModulesVersion %s\n' % module_version
+        mod_filepath = os.path.join(module_folder_path, '.version')
```
boegel: just collapse straight into `write_file`? only used in one place...

easybuild/tools/module_generator.py
```diff
@@ -693,6 +711,27 @@ def getenv_cmd(self, envvar):
         return 'os.getenv("%s")' % envvar
 
 
+    def set_as_default(self, module_folder_path, module_version):
+        """
+        Create file called "default" inside the package's module folder in order
+        to set the default module version
+        :param module_folder_path: module folder path, e.g.  $HOME/easybuild/modules/all/Bison
+        :param module_version: 3.0.4
```
boegel: include proper help message rather than (only) a random version string?

test/framework/filetools.py
```diff
@@ -355,6 +355,23 @@ def test_is_readable(self):
...
+        fp = os.path.join(self.test_prefix, 'test.txt')
+        txt = "test_my_link_file"
+        ft.write_file(fp, txt)
+
+        link = os.path.join(self.test_prefix, 'test.link')
+        # creating the link file
+        ft.symlink(fp, link)
+
+        # reading file contents and comparing to link
+        self.assertEqual(ft.read_file(fp), ft.read_file(link))
```
boegel: this doesn't mean at all that the files are correctly symlinked? could be a copy? just test below is enough imho, although maybe also check that `link` *is* a symbolic link (as opposed to a copied file)

test/framework/module_generator.py
```diff
@@ -40,13 +40,16 @@
 from easybuild.tools import config
 from easybuild.tools.module_generator import ModuleGeneratorLua, ModuleGeneratorTcl
 from easybuild.tools.module_naming_scheme.utilities import is_valid_module_name
+from easybuild.tools.filetools import mkdir, write_file
```
boegel: sort line alphabetically please?

test/framework/module_generator.py
```diff
@@ -40,13 +40,16 @@
...
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig.easyconfig import EasyConfig, ActiveMNS
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.modules import Lmod
 from easybuild.tools.utilities import quote_str
 from test.framework.utilities import EnhancedTestCase, TestLoaderFiltered, find_full_path, init_config
 
+from vsc.utils.missing import nub
+from easybuild.tools.modules import curr_module_paths, reset_module_caches
```
boegel: this should be moved up...

test/framework/module_generator.py
```diff
@@ -139,6 +142,62 @@ def test_descr(self):
         desc = self.modgen.get_description()
         self.assertEqual(desc, expected)
 
+    def test_set_default(self):
+        """
+        Test load part in generated module file.
+        """
+        # if not self.MODULE_GENERATOR_CLASS == ModuleGeneratorTcl:
+        #     return
```
boegel: please remove?

test/framework/module_generator.py
```diff
@@ -139,6 +142,62 @@ def test_descr(self):
...
+
+        # creating base path
+        base_path = os.path.join(self.test_prefix, 'all')
+        mkdir(base_path)
+
+        # creating package module
+        module_name = 'dummy'
```
boegel: `dummy` has a particular meaning, use just a random module name instead like `foobar` or whatever

test/framework/module_generator.py
```diff
@@ -139,6 +142,62 @@ def test_descr(self):
...
+        modules_base_path = os.path.join(base_path, module_name)
+        mkdir(modules_base_path)
+
+        # creating two empty modules
+        txt = self.modgen.MODULE_SHEBANG
+        if txt:
+            txt += '\n'
+
+        foo_version = 'foo_version'
```
boegel: why not make this 'proper' versions like `1.0` and `2.0`?

easybuild/tools/filetools.py
boegel: nevermind, this was moved up, OK

easybuild/tools/filetools.py
```diff
@@ -180,7 +204,8 @@ def remove_file(path):
         return
 
     try:
-        if os.path.exists(path):
+        # note: file may also be a broken symlink...
+        if os.path.exists(path) or os.path.islink(path):
```
boegel: @victorusu please add a dedicated test case for this fixed bug?

easybuild/tools/module_generator.py
```diff
@@ -452,6 +456,20 @@ def getenv_cmd(self, envvar):
         """
         return '$env(%s)' % envvar
 
+    # noinspection PyPackageRequirements,PyPackageRequirements
```
boegel: @victorusu what's this about (and why twice the same thing)?

easybuild/tools/module_generator.py
```diff
@@ -693,6 +711,29 @@ def getenv_cmd(self, envvar):
         return 'os.getenv("%s")' % envvar
 
 
+    def set_as_default(self, module_folder_path, module_version):
+        """
+        Create file called "default" inside the package's module folder in order
```
boegel: it's not a file, but a symlink?

easybuild/tools/module_generator.py
```diff
@@ -693,6 +711,29 @@ def getenv_cmd(self, envvar):
...
+        to set the default module version
+        :param module_folder_path: module folder path, e.g. $HOME/easybuild/modules/all/Bison
+        :param module_version: module version, e.g. 3.0.4
+        """
+        default_filepath = os.path.join(module_folder_path, 'default')
+
+        if os.path.islink(default_filepath):
+            points_to = readlink(default_filepath)
```
boegel: rename to `link_target`?

easybuild/tools/module_generator.py
```diff
@@ -693,6 +711,29 @@ def getenv_cmd(self, envvar):
...
+            remove_file(default_filepath)
+            self.log.info("Removed default version marking from %s.", points_to)
+        elif os.path.exists(default_filepath):
+            raise EasyBuildError('Found an unexpected file called default in dir %s' % module_folder_path)
+
+        cwd = os.getcwd()
+        #os.chdir(module_folder_path)
+        symlink(module_version + self.MODULE_FILE_EXTENSION, default_filepath)
+        #os.chdir(cwd)
```
boegel: please clean up the uncommented code, and you can drop the `cwd = ` line too (not used anymore)

test/framework/filetools.py
```diff
@@ -355,6 +355,23 @@ def test_is_readable(self):
...
+        ft.write_file(fp, txt)
+
+        link = os.path.join(self.test_prefix, 'test.link')
+        # creating the link file
+        ft.symlink(fp, link)
+
+        # checking if file is symlink
+        self.assertTrue(os.path.islink(link))
+        # reading link target and comparing to file name
+        self.assertEqual(fp, ft.readlink(link))
```
boegel: hmm, be careful here, what if `fp` is a file in a symlinked location?

maybe use `os.path.realpath` or `ft.readlink` when defining `fp`?

test/framework/module_generator.py
```diff
@@ -46,7 +48,7 @@
 from easybuild.tools.modules import Lmod
 from easybuild.tools.utilities import quote_str
 from test.framework.utilities import EnhancedTestCase, TestLoaderFiltered, find_full_path, init_config
-
+from vsc.utils.missing import nub
```
boegel: move this up, next to the other imports from `vsc.utils`?

test/framework/module_generator.py
```diff
@@ -139,6 +141,58 @@ def test_descr(self):
...
+        write_file(version_one_path, txt)
+
+        version_two = '2.0'
+        version_two_path = os.path.join(modules_base_path, version_two + self.modgen.MODULE_FILE_EXTENSION)
+        write_file(version_two_path, txt)
+
+        # removing all available modules from the environment
+        mod_paths = [p for p in nub(curr_module_paths()) if os.path.exists(p)]
+        for mod_path in nub(mod_paths):
+            self.modtool.unuse(mod_path)
```
boegel: do you really need to do this? if not, then don't?

easybuild/tools/filetools.py
victorusu: I have

boegel:
@victorusu if you consider this fit for rereview, please update PR title and ping me in a comment

victorusu:
@boegel, I changed the symlink function in order to allow non-full absolute source paths. Which is a required for older versions of LMod (< 7.1 at least). So now, all the LMod tests are passing. 
I am still experiencing a failure due to TMod. The test `test_set_default` is creating `.lua` module files, even though the env module system is TMod. So I wonder if this is the desired behaviour and I should add an if statement like the following in order to skip this case?

```
if self.MODULE_GENERATOR_CLASS == ModuleGeneratorLua and os.getenv("EASYBUILD_MODULE_SYNTAX") == "Tcl":
pass
```

Thanks!

boegel:
@victorusu How can the test still be creating Lua module file, we're using `self.modgen.MODULE_FILE_EXTENSION` to make sure we create the right type of module files?

Unless I'm missing something, the test should work unconditionally regardless of which modules tool & syntax is being used?

boegel:
OK, I think I see the problem now...

The tests in `test/framework/module_generator.py` are being run twice automagically, once for Tcl syntax, and once for Lua syntax.

However, running them with Lua syntax only works when Lmod is being used as a modules tool, of course.

So, yes @victorusu, you'll need to add a check like:

```python
if self.MODULE_GENERATOR_CLASS == ModuleGeneratorLua and not isinstance(self.modtool, Lmod):
    pass
```

victorusu:
@boegel, If think the issue is related to the fact that the `test.suite` calls
```
suite.addTests(TestLoaderFiltered().loadTestsFromTestCase(TclModuleGeneratorTest, sys.argv[1:]))
suite.addTests(TestLoaderFiltered().loadTestsFromTestCase(LuaModuleGeneratorTest, sys.argv[1:]))
```
So, the generator is `Lua` but the environment is `Tcl`. This issue was not observed before because no, previous, test generated any module. The current tests only `load` and `unload` modules.

Isn't it?

test/framework/module_generator.py
```diff
@@ -139,6 +141,57 @@ def test_descr(self):
         desc = self.modgen.get_description()
         self.assertEqual(desc, expected)
 
+    def test_set_default(self):
+        """
+        Test load part in generated module file.
+        """
+        if self.MODULE_GENERATOR_CLASS == ModuleGeneratorLua and not isinstance(self.modtool, Lmod):
+            pass
```
boegel: @victorusu: you'll need to make this a `return` rather than a `pass` to actually skip the test

victorusu: Yes, thanks!

victorusu:
@boegel, The unittests are finally passing! 👍 
I removed the WIP from the title. I am waiting for feedback. :-)
Thanks!

easybuild/tools/module_generator.py
victorusu: I removed. It was a mistake...

easybuild/tools/filetools.py
```diff
@@ -171,6 +171,38 @@ def write_file(path, txt, append=False, forced=False):
...
+
+    return target_symlink_path
+
+
+def symlink(source_path, symlink_path, use_abspath_source=True):
+    """
+    Create a symlink at the specified path to the given path.
+
+    :param source_path: source file path
+    :param symlink_path: symlink file path
```
boegel: `use_abspath_source` is not documented in docstring, please fix

easybuild/tools/filetools.py
```diff
@@ -171,6 +171,38 @@ def write_file(path, txt, append=False, forced=False):
...
+    """
+    try:
+        if use_abspath_source:
+            os.symlink(os.path.abspath(source_path), symlink_path)
+            _log.info("Symlinked %s to %s", os.path.abspath(source_path), symlink_path)
+        else:
+            os.symlink(source_path, symlink_path)
+            _log.info("Symlinked %s to %s", source_path, symlink_path)
```
boegel: you can shorten this a bit:

```python
if use_abspath_source:
    source_path = os.path.abspath(source_path)

 os.symlink(source_path, symlink_path)
_log.info("Symlinked %s to %s", source_path, symlink_path)

File: easybuild/tools/module_generator.py [hunk 29]
```diff
@@ -213,7 +213,11 @@ def swap_module(self, mod_name_out, mod_name_in, guarded=True):
         """
         raise NotImplementedError
 
+    def set_as_default(self, module_folder_path, module_version):
+        raise NotImplementedError
+
 
+# noinspection PyPackageRequirements
```
boegel: please remove?

easybuild/tools/module_generator.py
```diff
@@ -693,6 +710,26 @@ def getenv_cmd(self, envvar):
...
+        :param module_version: module version, e.g. 3.0.4
+        """
+        default_filepath = os.path.join(module_folder_path, 'default')
+
+        if os.path.islink(default_filepath):
+            link_target = readlink(default_filepath)
+            remove_file(default_filepath)
+            self.log.info("Removed default version marking from %s.", link_target)
+        elif os.path.exists(default_filepath):
+            raise EasyBuildError('Found an unexpected file called default in dir %s' % module_folder_path)
```
boegel: `s/called default/named 'default'/g`

easybuild/tools/options.py
```diff
@@ -383,6 +383,8 @@ def override_options(self):
                                      None, 'store_true', False),
             'zip-logs': ("Zip logs that are copied to install directory, using specified command",
                          None, 'store_or_None', 'gzip'),
+            'set-as-default': ("Set the generated module as default", None, 'store_true', False),
```
boegel: maybe rename to `--set-default-module`? leaves less room for misinterpretation...

test/framework/filetools.py
```diff
@@ -355,6 +355,49 @@ def test_is_readable(self):
...
+        ft.write_file(fp, txt)
+
+        link = os.path.join(self.test_prefix, 'test.link')
+        # creating the link file
+        ft.symlink(fp, link)
+
+        # checking if file is symlink
+        self.assertTrue(os.path.islink(link))
+        # reading link target and comparing to file name
+        self.assertEqual(os.path.realpath(fp), ft.readlink(link))
```
boegel: hmm, I think this may fail, e.g. if `self.test_prefix` is a symlink...

`os.path.realpath` will give you the fully resolved location, while `ft.readlink` will only resolve one symlink 'level', I think?

test/framework/module_generator.py
```diff
@@ -139,6 +141,57 @@ def test_descr(self):
         desc = self.modgen.get_description()
         self.assertEqual(desc, expected)
 
+    def test_set_default(self):
+        """
+        Test load part in generated module file.
+        """
+        if self.MODULE_GENERATOR_CLASS == ModuleGeneratorLua and not isinstance(self.modtool, Lmod):
+            return
```
boegel: include a short comment to explain why you're bailing out here, this looks suspicious, but we're not cheating here :)

test/framework/filetools.py
victorusu: You are right! Setting
```
self.assertEqual(os.path.realpath(fp), os.path.realpath(ft.readlink(link)))
```
worked for me.

boegel: Yes, ok, but then the test is worthless, no?

I'm starting to think we should change `ft.readlink` to `ft.resolve_symlink`, and basically make it call `os.path.realpath` in a `try-except`?

boegel:
@victorusu https://github.com/victorusu/easybuild-framework/pull/1 should make this good to go

boegel:
Going in, thanks @victorusu!

Issue #1877, opened 2016-08-17
--module-only skips over important steps and crashes


MalkinE:
Perhaps this should be split into several issues. I kept it together for an overview since they have the `--module-only` in common.

There are around 70 cases I know of where already installed software packages can not have their modules recreated with `--module-only`. In all cases I discuss here I first did a full installation and then tried to generate a second module in a new module directory. Mostly it comes down to either some path not being set up, or a build or an install directory being empty, because several steps were skipped. Up to now I have found the following sources of this problem:

1) in the easyblock "self.build_in_installdir = True"
   This causes `cleanupoldbuild` to be True and then the install directory to be cleared. After the installed software has been deleted (the old module file stays), the sanity check of course fails.
F.e.: OpenFOAM, FSL, QuantumESPRESSO

2) in the easyconfig `start_dir` is defined
  This usually points to some place in the build directory. With the default settings the build directory is cleaned after a successful installation, so EasyBuild can not find the start_dir and gives an error message. See also issue #1642.
F.e.: Tk, Tcl, Hypre

3) in the easyblock some custom_path is defined which uses a parameter set during the build or the installation step
   For example in the gromacs.py, in the install_step self.lib_subdir is set to include the lib64 directory. This is skipped during a module-only installation, so in the end the custom_paths used in the sanity_check_step is not correct. EasyBuild installs the files are in the right place, but during a module recreation it can not find them and gives an error message.
F.e. GROMACS

boegel:
@MalkinE Thanks for reporting these issues in detail!

These issues are probably also relevant for `--extended-dry-run`/`-x`... So we should update the documentation at http://easybuild.readthedocs.io/en/latest/Extended_dry_run.html#guidelines-for-easyblocks, and also refer to it from the section where `--module-only` is documented.

From the top of my head, I think problems 1 & 2 can be fixed in the framework in a general way.

For the GROMACS issue, I think this has to be fixed in the GROMACS easyblock itself, potentially combined with enhancing the test we have in the easyblocks repository that tries to ensure an easyblock is compatible with `--module-only`; it is clearly missing something that allowed these issues to go unnoticed until now...

MalkinE:
Thanks @boegel !

 I just realized the "dangerous" side of the first case: you are deleting a functioning installation without using `--force` or `-f`. And this, as I understand it, is against the EasyBuild philosophy.

riccardomurri:
[...]

PR #2112, opened 2017-02-10
ensure that $EBEXTSLIST* is also included in generated module under --module-only

geimer:
Rather than using the filtered list of extensions built up during installation, use the full list from the easyconfig.  Using the filtered list breaks when adding extensions and with `--module-only`.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1067,7 +1067,8 @@ class EasyBlock(object):
         footer = [self.module_generator.comment("Built with EasyBuild version %s" % VERBOSE_VERSION)]
 
         # add extra stuff for extensions (if any)
-        footer.append(self.make_module_extra_extensions())
+        if self.cfg['exts_list']:
+            footer.append(self.make_module_extra_extensions())
 
         # include modules footer if one is specified
         if self.modules_footer is not None:
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1046,7 +1046,7 @@ class EasyBlock(object):
 
         return txt
 
-    def make_module_extra_extensions(self, extensions):
+    def make_module_extra_extensions(self):
         """
         Sets optional variables for extensions.
         """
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1068,8 +1067,9 @@ def make_module_footer(self):
         footer = [self.module_generator.comment("Built with EasyBuild version %s" % VERBOSE_VERSION)]
 
         # add extra stuff for extensions (if any)
-        if self.cfg['exts_list']:
-            footer.append(self.make_module_extra_extensions())
+        extensions = self.cfg['exts_list']
+        if extensions:
+            footer.append(self.make_module_extra_extensions(extensions))
```
boegel: hmm, there's little point in passing down `self.cfg['exts_list']` here, since self.make_module_extra_extensions also has access to it?

easybuild/framework/easyblock.py
```diff
@@ -1068,8 +1067,9 @@ def make_module_footer(self):
         footer = [self.module_generator.comment("Built with EasyBuild version %s" % VERBOSE_VERSION)]
 
         # add extra stuff for extensions (if any)
-        if self.cfg['exts_list']:
-            footer.append(self.make_module_extra_extensions())
+        extensions = self.cfg['exts_list']
+        if extensions:
```
boegel: it should be fine to just drop this condition imho... with `self.exts_all` we had to be more careful since the default value was `None`, but `self.cfg['exts_list']` is an empty list by default so it's fine to always iterate over it

easybuild/framework/easyblock.py
```diff
@@ -1046,18 +1046,17 @@ def make_module_extra(self, altroot=None, altversion=None):
...
         """
         # add stuff specific to individual extensions
         lines = [self.module_extra_extensions]
 
         # set environment variable that specifies list of extensions
-        if self.exts_all:
-            exts_list = ','.join(['%s-%s' % (ext['name'], ext.get('version', '')) for ext in self.exts_all])
-            env_var_name = convert_name(self.name, upper=True)
-            lines.append(self.module_generator.set_environment('EBEXTSLIST%s' % env_var_name, exts_list))
+        exts_list = ','.join(['%s-%s' % (ext[0], ext[1]) for ext in extensions])
```
boegel: just use `self.cfg['exts_list']` here rather than passing it down `extensions`

easybuild/framework/easyblock.py
geimer: Nope.  We will generate an empty env var in this case...

easybuild/framework/easyblock.py
```diff
@@ -1054,10 +1054,9 @@ def make_module_extra_extensions(self):
         lines = [self.module_extra_extensions]
```
boegel: Yes, this is extra stuff that can go into the module file that is specific to a particular extension, which is returned by the `run` method.
This is still broken now, since every call to `run` is being skipped under `--module-only`...

It's possible to fix this, but that's orthogonal to this fix imho. I'll mention this in #1877, which is sort of a tracking issue for problems with `--module-only`

PR #2113, opened 2017-02-10
Additional easyconfig parameters for documentation (REVIEW)

geimer:
This PR implements support for additional easyconfig parameters for customizing the documentation as discussed here: https://www.mail-archive.com/easybuild@lists.ugent.be/msg03050.html
It also includes the list of installed extensions of a package in the `module help` output.


easybuild/framework/easyconfig/default.py
```diff
@@ -158,6 +158,11 @@
 
     # MODULES easyconfig parameters
     'whatis': [None, "List of brief (one line) package description entries", MODULES],
+    'usage': [None, "Usage instructions for the software", MODULES],
+    'docpaths': [None, "List of paths of installed documentation relative to package root dir", MODULES],
+    'docurls': [None, "List of URLs to package documentation", MODULES],
+    'support': [None, "Package-specific support/bug report address", MODULES],
+    'contact': [None, "Site contact for the package", MODULES],
```
boegel: hmm, this is getting a bit messy, is it not?

we should try and sort these alphabetically, although maybe also keep related ones together in groups?

easybuild/tools/module_generator.py
```diff
@@ -41,9 +41,10 @@
 
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option, get_module_syntax, install_path
-from easybuild.tools.filetools import mkdir, read_file
-from easybuild.tools.modules import modules_tool
+from easybuild.tools.filetools import convert_name, mkdir, read_file
+from easybuild.tools.modules import ROOT_ENV_VAR_NAME_PREFIX, modules_tool
 from easybuild.tools.utilities import quote_str
+from textwrap import wrap
```
boegel: this is a non-`easybuild` import, it should go up with the others?

easybuild/tools/module_generator.py
```diff
@@ -213,6 +214,79 @@ def swap_module(self, mod_name_out, mod_name_in, guarded=True):
...
+        Generate syntax-independent help text used for `module help`.
+        """
+        # General package description (mandatory)
+        description = self.app.cfg['description']
+        lines = [
+            '',
+            'Description',
+            '===========',
+            "%s" % description.strip(),
+        ]
```
boegel: this is a common pattern below, so maybe define a helper function inside of this method to ensure consistency of the format?

```python
def _generate_help_text(self):
    ...

    def generate_section(sec_name, sec_txt):
        """
        Generate section with given name and contents
        """"
        if sec_txt:
            return [
                '',
                sec_name,
                '=' * len(sec_name),
                sec_txt.strip(),
             ]
         else:
             return []

        # General package description (mandatory)
        lines = generate_section('Description', self.app.cfg['description'])

        # Package usage instructions (optional)
        lines.extend('Usage', self.app.cfg['usage'])

        # Additional information: homepage + (if available) doc paths/urls, support info, contacts, ...
        lines.extend('More information', " - Homepage: %s" % self.app.cfg['homepage'])
        ...

        extensions = ', '.join(sorted(['%s-%s' % (ext[0], ext[1]) for ext in exts_list], key=str.lower))
        lines.extend('Extensions', '\n'.join(wrap(extensions, 78)))
```

easybuild/tools/module_generator.py
```diff
@@ -213,6 +214,79 @@ def swap_module(self, mod_name_out, mod_name_in, guarded=True):
...
+        # Extensions (if any)
+        exts_list = self.app.cfg['exts_list']
+        if exts_list:
+            extensions = ', '.join(sorted(['%s-%s' % (ext[0], ext[1]) for ext in exts_list], key=str.lower))
+            lines.extend([
+                '',
+                '',
+                "Included extensions",
+                '===================',
+                "%s" % '\n'.join(wrap(extensions, 78)),
```
boegel: hmm, the 78 here looks very arbitrary/harcoded... why exactly 78?

easybuild/tools/module_generator.py
```diff
@@ -270,7 +344,8 @@ def get_description(self, conflict=True):
 
         lines = [
             "proc ModulesHelp { } {",
-            "    puts stderr { %(description)s",
+            "    puts stderr {"
+            "%s" % self._generate_help_text(),
```
boegel: `"%s" % x` is the same thing as just `x`

easybuild/tools/module_generator.py
```diff
@@ -516,7 +591,9 @@ def get_description(self, conflict=True):
             whatis = ["Description: %s" % description]
 
         lines = [
-            "help([[%(description)s]])",
+            'help([[',
+            "%s" % self._generate_help_text(),
```
boegel: here too, no need for the `"%s" %`

easybuild/framework/easyconfig/default.py
boegel: for `contact`, don't make it specific to `site contact`, could also be upstream contact for this software package (e.g. a mailing list)?

easybuild/tools/module_generator.py
```diff
@@ -213,6 +214,79 @@ def swap_module(self, mod_name_out, mod_name_in, guarded=True):
...
+            if isinstance(support, list):
+                lines.extend([" - Support/bug reports:"])
+                lines.extend(["    - %s" % address for address in support])
+            else:
+                lines.extend([" - Support/bug reports: %s" % support])
+
+        contact = self.app.cfg['contact']
+        if contact:
+            if isinstance(contact, list):
+                lines.extend([" - Site contacts:"])
```
boegel: drop the `Site `?

easybuild/framework/easyconfig/default.py
geimer: Well, the idea was to separate upstream contacts (`support`) from site-local contacts (`contact`).  The documentation of those two parameters already tries to make that clear.  But maybe the naming is sub-optimal.  I'm open for suggestions.

easybuild/tools/module_generator.py
geimer: To make it fit nicely into a 80-character-wide terminal window :grin:

When making this a configuration parameter, I'd also like to see all other sections to be wrapped at the given width. But this is (at least from my "I don't know much Python" perspective) by no means trivial to achieve with `textwrap.wrap()`.  Maybe someone knows a good rst-to-ascii-text renderer that we could leverage?

easybuild/framework/easyconfig/default.py
geimer: Hmm... I just noticed that there is already a `docurls` parameter defined in the `MANDATORY` section, though it is not (yet) enforced.  I guess it hasn't been used so far?

Regarding the "messiness": Would it make sense to add a `DOCUMENTATION` group and add all those variables to it?  Though `description` would then need to be in `MANDATORY` and `DOCUMENTATION`...

pforai:
I like the idea  👍

easybuild/framework/easyconfig/default.py
verdurin: I think the `support` documentation needs to indicate that this is for upstream, not site-local support. Otherwise it's rather ambiguous.

geimer: What about `upstream_contact` and `site_contact`?  This should make it clear.  And since the list values are free-form text fields, one could do something like
```
upstream_contact = [
    'Support e-mail: support@project.org',
    'Mailing list: discuss@project.org',
    'Bugtracker: https://www.project.org/bugzilla'
]
site_contact = ['Jane Admin']
```
Which would be rendered as
```
More information
================
 - Homepage: https://www.project.org
 - Upstream contact:
    - Support e-mail: support@project.org
    - Mailing list: discuss@project.org
    - Bugtracker: https://www.project.org/bugzilla
 - Site contact: Jane Admin
```

verdurin: Yes, exactly what I had in mind.

geimer:
@boegel There is no test yet for the whatis/help text in the presence of extensions, and I'm not quite sure where to put it.  Any suggestion?  Otherwise I think this is now ready for prime time.

boegel:
@geimer Can you take care of the conflict before I revisit this?

W.r.t. a test for whatis/help with extensions involved, we have `toy` tests that involve extensions, maybe there? Let me know if you're up for looking into that.

boegel:
@geimer @damianam did an effort to sort things alphabetically in `module_generator`, cfr. #2132

boegel:
@geimer why would using `test_toy_advanced` be better (or why would it create problems)?

damianam:
> Who f?$%ed up the `module_generator`

Ups ;-P. Sorting it wasn't really necessary, but working with it with methods put at random places in the file was kind of annoying, that's why I shuffled everything around.

geimer:
@boegel If I'm not mistaken, `test_toy_advanced` is the one using extensions. But it would have to be adjusted to include the additional ec parameters which may have undesired side effects.

easybuild/framework/easyconfig/default.py
```diff
@@ -171,6 +170,16 @@
     'include_modpath_extensions': [True, "Include $MODULEPATH extensions specified by module naming scheme.", MODULES],
     'recursive_module_unload': [False, 'Recursive unload of all dependencies when unloading module', MODULES],
 
+    # MODULES documentation easyconfig parameters
+    #    (docurls is part of MANDATORY)
+    'docpaths': [None, "List of paths of installed documentation relative to package root dir", MODULES],
```
boegel: `List of paths for documentation relative to installation directory`

easybuild/framework/easyconfig/default.py
```diff
@@ -171,6 +170,16 @@
...
+    'examples': [None, "Free-form text with examples", MODULES],
```
boegel: `... with examples on using the software`

easybuild/framework/easyconfig/default.py
```diff
@@ -171,6 +170,16 @@
...
+    'site_contact': [None, "String/list of strings with site contacts for the package", MODULES],
```
boegel: plural: `site_contacts`

also: `s/package/software/`

easybuild/framework/easyconfig/default.py
```diff
@@ -171,6 +170,16 @@
...
+    'upstream_contact': [None, ("String/list of strings with upstream contact addresses "
+                                "(e.g., support e-mail, mailing list, bugtracker)"), MODULES],
+    'usage': [None, "Usage instructions for the package", MODULES],
```
boegel: `s/package/software/`

easybuild/framework/easyconfig/default.py
```diff
@@ -171,6 +170,16 @@
...
+    'whatis': [None, "List of brief (one line) package description entries", MODULES],
```
boegel: `s/package/software/`

easybuild/tools/module_generator.py
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+
+                return [
+                    '',
+                    '',
+                    sec_name,
+                    '=' * len(sec_name),
+                    text,
+                ]
+            else:
+                return []
```
boegel: hmm, this is a bit too vertically verbose, no? also, we generally avoid inline `return`s

also, let's make this a private method like the others you're calling out to, rather than an inner function?

```python
def _generate_help_section(sec_name, sec_txt, strip=False):
    """
    Generate section with given name and contents.
    """
    res = []
    if sec_txt:
        if strip:
            sec_txt = sec_txt.strip()
        res = ['', '', sec_name, '=' * len(sec_name), sec_txt]
    return res
```

easybuild/tools/module_generator.py
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+        # Additional information: homepage + (if available) doc paths/urls, upstream/site contact
+        lines.extend(generate_section("More information", " - Homepage: %s" % self.app.cfg['homepage']))
+
+        docpaths = self.app.cfg['docpaths'] or []
+        docurls = self.app.cfg['docurls'] or []
+        if docpaths or docurls:
+            env_name = convert_name(self.app.name, upper=True)
+            root_envvar = ROOT_ENV_VAR_NAME_PREFIX + env_name
+            lines.extend([" - Documentation:"])
+            lines.extend(["    - $%s/%s" % (root_envvar,path) for path in docpaths])
```
boegel: style nitpicking: space after `,`

easybuild/tools/module_generator.py
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+        lines.extend(generate_section('Examples', self.app.cfg['examples'], strip=True))
+
+        # Additional information: homepage + (if available) doc paths/urls, upstream/site contact
+        lines.extend(generate_section("More information", " - Homepage: %s" % self.app.cfg['homepage']))
+
+        docpaths = self.app.cfg['docpaths'] or []
+        docurls = self.app.cfg['docurls'] or []
+        if docpaths or docurls:
+            env_name = convert_name(self.app.name, upper=True)
+            root_envvar = ROOT_ENV_VAR_NAME_PREFIX + env_name
```
boegel: collapse these into

```python
root_envvar = ROOT_ENV_VAR_NAME_PREFIX + convert_name(self.app.name, upper=True)
```

easybuild/tools/module_generator.py
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+            lines.extend([" - Documentation:"])
+            lines.extend(["    - $%s/%s" % (root_envvar,path) for path in docpaths])
+            lines.extend(["    - %s" % url for url in docurls])
+
+        upstream = self.app.cfg['upstream_contact']
+        if upstream:
```
boegel: `s/upstream/upstream_contacts/`

easybuild/tools/module_generator.py
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+            if isinstance(upstream, list):
+                lines.extend([" - Upstream contacts:"])
+                lines.extend(["    - %s" % address for address in upstream])
```
boegel: `s/address/contact/`

(may be just a name, or an URL)

easybuild/framework/easyconfig/default.py
```diff
@@ -171,6 +170,16 @@
     'include_modpath_extensions': [True, "Include $MODULEPATH extensions specified by module naming scheme.", MODULES],
     'recursive_module_unload': [False, 'Recursive unload of all dependencies when unloading module', MODULES],
 
+    # MODULES documentation easyconfig parameters
+    #    (docurls is part of MANDATORY)
+    'docpaths': [None, "List of paths of installed documentation relative to package root dir", MODULES],
+    'examples': [None, "Free-form text with examples", MODULES],
+    'site_contact': [None, "String/list of strings with site contacts for the package", MODULES],
+    'upstream_contact': [None, ("String/list of strings with upstream contact addresses "
```
boegel: I'd make this plural: `upstream_contacts`

easybuild/tools/module_generator.py
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+            else:
+                lines.extend([" - Upstream contact: %s" % upstream])
+
+        site_contact = self.app.cfg['site_contact']
+        if site_contact:
+            if isinstance(site_contact, list):
+                lines.extend([" - Site contacts:"])
+                lines.extend(["    - %s" % address for address in site_contact])
```
boegel: `s/address/contact/g`

easybuild/tools/module_generator.py
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+            else:
+                lines.extend([" - Site contact: %s" % site_contact])
```
boegel: since the code is copy-paste for `upstream_contacts` and `site_contacts`, how about:

```python
for contacts_type in ['upstream', 'site']:
    contacts = self.cfg['%s_contacts' % contacts_type]
    if contacts:
        if isinstance(contacts, list):
            lines.append(" - %s contacts" % contacts_type.capitalize())
            lines.extend(["    - %s" % c for c in contacts])
        else:
            lines.append(" - %s contact: %s" % (contacts_type.capitalize(), contacts))

File: easybuild/tools/module_generator.py [hunk 21]
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+                "Description: %s" % self.app.cfg['description'],
+                "Homepage: %s" % self.app.cfg['homepage']
+            ]
+            extensions = self._generate_extension_list()
+            if extensions:
+                whatis.append("Extensions: %s" % extensions)
+
+        return whatis
+
+    def _generate_extension_list(self):
```
boegel: sort these methods alphabetically, let's not let @damianam's efforts go to waste ;)

easybuild/tools/module_generator.py
```diff
@@ -298,6 +299,100 @@ def use(self, paths, prefix=None, guarded=False):
...
+        """
+        Generate a string with a comma-separated list of extensions.
+        """
+        exts_list = self.app.cfg['exts_list']
+        extensions = ', '.join(sorted(['%s-%s' % (ext[0], ext[1]) for ext in exts_list], key=str.lower))
```
boegel: `s/(ext[0], ext[1])/ext[:2]/`

easybuild/tools/module_generator.py
```diff
@@ -376,10 +464,11 @@ def get_description(self, conflict=True):
             # - 'conflict Compiler/GCC/4.8.2/OpenMPI' for 'Compiler/GCC/4.8.2/OpenMPI/1.6.4'
             lines.extend(['', "conflict %s" % os.path.dirname(self.app.short_mod_name)])
 
+        whatis = self._generate_whatis_lines()
```
boegel: just move this inline, only used in one place?

easybuild/tools/module_generator.py
```diff
@@ -624,10 +706,11 @@ def get_description(self, conflict=True):
             # conflict on 'name' part of module name (excluding version part at the end)
             lines.extend(['', 'conflict("%s")' % os.path.dirname(self.app.short_mod_name)])
 
+        whatis = self._generate_whatis_lines()
```
boegel: same here, move this inline?

test/framework/module_generator.py
```diff
@@ -74,17 +74,27 @@ def setUp(self):
     def test_descr(self):
         """Test generation of module description (which includes '#%Module' header)."""
 
-        gzip_txt = "gzip (GNU zip) is a popular data compression program as a replacement for compress "
-        gzip_txt += "- Homepage: http://www.gzip.org/"
+        gzip_txt = "gzip (GNU zip) is a popular data compression program as a replacement for compress"
```
boegel: rename `gzip_txt` to `descr`?

easybuild/tools/module_generator.py
geimer: `TypeError: not enough arguments for format string`

geimer: Now using `'-'.join(ext[:2]) for ext in ...` but I'm not sure it helps w.r.t. readablity...

boegel:
lgtm, going in, thanks for your hard work and patience on this @geimer!

PR #2122, opened 2017-02-15
WIP Fat yaml support

rjeschmi:
Don't spend too much time reviewing this right now. I know there are some problems.


ocaisa:
I started looking in to my part on the plane last night. As expected I ran into some unknown unknowns, the main problem being I can't dump an easyconfig that was read in one format into the other. I'll need to figure out a way around this before I can do anything. I think what I will do is write a generic method to dump an easyconfig dict as one (and eventually the other) format.

rjeschmi:
I think this might be more promising as a better config format language:

http://jsonnet.org/

Issue #420, opened 2013-01-10
Allow root to execute eb


dominikborkowski:
Currently easybuild refuses to run when executed under root account. While it's often prudent to have separate accounts for compile processes, it seems that this should be an option, rather than a forced requirement. 

What would be nice is to have `--allow-root`/`--allow-superuser`/`--allow-admin` or similar option, that would allow an override of this behavior. 

Cheers!

fedepad:
What's the status? Because I don't see this thing in the repo: downloaded the (latest?) bootstrap_eb.py as described in the website but fails as root. .
In my case I have a system that I use always as root (no multi-user....); I think "not being root" as a requirement it's not the best idea: probably make it as a default option that is configurable though in case someone who knows what is doing wants to just run and try.

boegel:
@fedepad Support for allowing to run as `root` is implemented in a straightforward way in https://github.com/hpcugent/easybuild-framework/pull/2123 .

That means that `--allow-use-as-root` will be supported in a future EasyBuild release (under the condition that the PR gets reviewed/approved/merged), so you should be able to bootstrap as `root` after running `export EASYBUILD_ALLOW_USE_AS_ROOT=1` soonish.

fedepad:
@boegel Thank you very much for the explanation! Appreciated!

fedepad:
I was able after a while (a lot...) to build, install Easybuild with your patch and load the "main" module. Now I execute:  

`eb mpiBLAST-1.6.0-goolf-1.4.10.eb --robot`
and I get:
> == temporary log file in case of crash /tmp/eb-2Trx7U/easybuild-he0M0p.log
> ERROR: You seem to be running EasyBuild with root privileges which is not wise, so let's end this here.

How can this be solved **QUICKLY**?

ocaisa:
The quickest thing to do would be to not use `root` and take the more standard approach that is well documented. 

Installing EB with `root` is one thing since @boegel know what the source does, installing all software through EB with `root` is another, how can you know that what you are doing is in any way secure? The EasyBuild recipes are community developed and not checked for security (they don't need to  be if they are installed in user space).

boegel:
The security concern is broader than just EasyBuild: you're running scripts, Makefiles, etc. from a random source on the worldwide web, and there have been (unintended) problems with stuff like that before (google "bumblebee github fail" to see a 'nice' example).

boegel:
[...]

PR #2123, opened 2017-02-15
add support for --allow-use-as-root-and-accept-consequences

boegel:
cfr. #420


boegel:
@rjeschmi please review?

test/framework/options.py
```diff
@@ -3048,6 +3051,26 @@ def test_check_style(self):
         for pattern in patterns:
             self.assertTrue(re.search(pattern, stdout, re.M), "Pattern '%s' found in: %s" % (pattern, stdout))
 
+    def test_allow_use_as_root(self):
+        """Test --allow-use-as-root"""
+
+        # pretend we're running as root by monkey patching os.getuid used in main
+        easybuild.main.os.getuid = lambda: 0
```
JensTimmerman: some linting tools don't like it when you assign lambda's, but this one is so trivial it doesn't require a def imho.

JensTimmerman:
lgtm

boegel:
@JensTimmerman Any concerns about running everything as root vs only the steps that would really require root access (cfr. your comments in #420)?

The latter would be a lot more difficult, and may not be worth the effort, since it wouldn't provide much protection against nasty things from happening during `make install` for example...

rjeschmi:
I think this is fine in general. I think another useful option is to demote from root to a particular user. I have a wrapper that does that for me so I can run easybuild as root more easily.

JensTimmerman:
@boegel I would add -i-know-this-is-a-bad-idea to the end of the option to make it clear this is an option that actually isn't supposed to be used.

pescobar:
hahahaha +1 for @JensTimmerman suggestion ;)

boegel:
@JensTimmerman That's what the warning is for. Making the option annoyingly long/silly is something we can do, but it won't make any difference if you set it via the environment or in a configuration file (which makes it do once and forget).

JensTimmerman:
@boegel sure you can once and forget, but the point is that a new user can immediately identify this option as something that shouldn't be just be used lightheartedly. 
The warning surely helps here, but once the warning is shown it's already to late...

boegel:
OK, fair point @JensTimmerman...

So, pick one:

* `--allow-use-as-root-even-though-it-is-a-bad-idea`
* `--allow-use-as-root-despite-being-not-recommended`
* `--allow-use-as-root-and-accept-consequences`

pescobar:
`--allow-use-as-root-even-though-it-is-a-bad-idea-but-I-am-a-shell-cowboy`

;)

JensTimmerman:
first one

boegel:
We briefly discussed this during today's EasyBuild conf call, where @pescobar mentioned that there may actually be good reasons to install something with root privileges, e.g. installing Singularity with EasyBuild (which requires enabling `setuid` on a binary).

It would be nicer to have support to only leverage admin privileges when really needed, i.e. only when actually enabled `setuid` on something, rather than doing the whole installation as root, but this would require quite a bit more work.

There also seemed to be a consensus that `--allow-use-as-root-and-accept-consequences` is a better option name, since it makes it clear that this configuration setting should be used with care, while not making claims that doing so is *always* wrong (as opposed to `--allow-use-as-root-even-though-it-is-a-bad-idea`). So, I renamed the option to `--allow-use-as-root-and-accept-consequences`.

The idea is to include this support for `--allow-use-as-root-and-accept-consequences` in the upcoming EasyBuild v3.2.0.

Issue #2014, opened 2016-11-22
Bash escape issues can cause fpm command to fail when --package is used


ajdecon:
The description in the gettext easyconfig ([example](https://github.com/hpcugent/easybuild-easyconfigs/blob/c849c699384d548feb18b72ab5b4f1c637bcf313/easybuild/easyconfigs/g/gettext/gettext-0.19.4.eb#L7)) includes a single unbalanced backtick in its description: "GNU \`gettext' is an important...". When I attempted to build this easyconfig with the `--package` flag, I saw the following error in the `fpm` step due to bash attempting to evaluate the backtick:

```
== 2016-11-22 17:42:05,402 build_log.py:149 ERROR EasyBuild crashed with an error (at ?:124 in __init__): cmd "fpm --workdir /tmp/eb-LJE_ig/eb-pkgs-JJ0Bek --name "gettext-0.19.8" --provides "gettext-0.19.8" -t rpm -s dir --version eb-3.0.0 --iteration 1 --description "GNU `gettext' is an important step for the GNU Translation Project, as it is an asset on which we may
build many other steps. This package offers to programmers, translators, and even users, a well integrated set of tools
and documentation" --url "http://www.gnu.org/software/gettext/" --exclude "opt/ajdecon/software/gettext/0.19.8/easybuild/*.log" --exclude "opt/ajdecon/software/gettext/0.19.8/easybuild/*.md"  --depends "ncurses-6.0" /opt/ajdecon/software/gettext/0.19.8 /opt/ajdecon/modules/all/gettext/0.19.8" exited with exitcode 1 and output:
/bin/bash: -c: line 0: unexpected EOF while looking for matching ``'
/bin/bash: -c: line 3: syntax error: unexpected end of file
 (at easybuild/tools/run.py:438 in parse_cmd_output)
```

I was able to proceed with the build by editing the easyconfig and replacing the opening backtick in the description with a single-quote.

However, rather than opening a bug against the easyconfig (and any others I find with unbalanced backticks ;-) ), I think this is probably an indication that the `package_with_fpm` [method](https://github.com/hpcugent/easybuild-framework/blob/ae1ddb4f08fa89aa2cb803dbcc996909f1fe4fec/easybuild/tools/package/utilities.py#L78) should do better escaping of content from easyconfigs before passing them to fpm.

Easybuild version info:
```
[ajdecon@buildbox Java]$ eb --version
[...]

PR #2124, opened 2017-02-16
Fix fpm description by wrapping in shell quote

rjeschmi:
Still need to write the test.


rjeschmi:
This will resolve #2014

easybuild/tools/package/utilities.py
```diff
@@ -134,7 +135,7 @@ def package_with_fpm(easyblock):
         '-s', 'dir',  # source
         '--version', pkgver,
         '--iteration', pkgrel,
-        '--description', quote_str(easyblock.cfg["description"]),
+        '--description', quote_str(eb_shell_quote(easyblock.cfg["description"])),
```
boegel: @rjeschmi do we need both `quote_str` and `eb_shell_quote`?

rjeschmi: I'll check. I thought the quote_str would handle the outside wrapping, but it probably doesn't work the way I think it does.

The tests should help me figure it out I'd think.

rjeschmi:
well the test is failing nicely, but it seems like both escapes are not enough :)

```
    raise EasyBuildError('cmd "%s" exited with exitcode %s and output:\n%s', cmd, ec, stdouterr)
EasyBuildError: 'cmd "fpm --workdir /tmp/eb-fveqiq/eb-OWKZv5/eb-fh53Jr/eb-CTbwdf/eb-pkgs-WRzb40 --name "toy-0.0-gompi-1.3.12-test" --provides "toy-0.0-gompi-1.3.12-test" -t rpm -s dir --version eb-3.1.0 --iteration 1 --description "\'Toy C program. Now with `backticks\\\'\'" --url "http://hpcugent.github.com/easybuild" --exclude "tmp/eb-fveqiq/eb-OWKZv5/tmpQlV2Qs/software/toy/0.0-gompi-1.3.12-test/easybuild/*.log" --exclude "tmp/eb-fveqiq/eb-OWKZv5/tmpQlV2Qs/software/toy/0.0-gompi-1.3.12-test/easybuild/*.md"  --depends "gompi-1.3.12" /tmp/eb-fveqiq/eb-OWKZv5/tmpQlV2Qs/software/toy/0.0-gompi-1.3.12-test /tmp/eb-fveqiq/eb-OWKZv5/tmpQlV2Qs/modules/all/toy/0.0-gompi-1.3.12-test.lua" exited with exitcode 1 and output:\n/bin/bash: -c: line 0: unexpected EOF while looking for matching ``\'\n/bin/bash: -c: line 1: syntax error: unexpected end of file\n'

```

easybuild/tools/package/utilities.py
rjeschmi: You can see now that I've removed both quotes. If you pass the commands as a list it will handle doing the escaping properly.

It is necessary to pass each separate part of the command line as its own list entry though (so I had to also change the way I was handling the excludes and such).

I think I'm missing a test case...

easybuild/tools/options.py
```diff
@@ -108,7 +108,7 @@ def eb_shell_quote(token):
     so it can be used in a shell command. This results in token that is not expanded/interpolated by the shell.
     """
     # escape any non-escaped single quotes, and wrap entire token in single quotes
-    return "'%s'" % re.sub(r"(?<!\\)'", r"\'", str(token))
+    return "'%s'" % re.sub(r"(?<!\\)'", r"'\''", str(token))
```
rjeschmi: I think this is "more correct" but I'm no longer using this command so I could remove this from this PR.

boegel:
@rjeschmi I plan to do a v3.1.1 bugfix release this week, any idea how close this is to being good to go?

rjeschmi:
I don't think it is too far. I'll maybe remove the changes to eb_shell_quote?

rjeschmi:
and you wanted  me to write command lists to tuples right?

easybuild/tools/options.py
boegel: if this is not relevant to this PR, please roll back (and maybe open a new PR for it)

easybuild/tools/package/utilities.py
```diff
@@ -79,6 +79,9 @@ def package_with_fpm(easyblock):
     """
     This function will build a package using fpm and return the directory where the packages are
     """
+    # avoid circular import, should move eb_shell_quote
+    from easybuild.tools.options import eb_shell_quote
```
boegel: no longer used, so please remove?

easybuild/tools/package/utilities.py
```diff
@@ -105,51 +108,53 @@ def package_with_fpm(easyblock):
...
-        for x in exclude_files_glob
-    ]
+    exclude_files_cmd = []
+    for x in exclude_files_glob:
+        exclude_files_cmd.extend(['--exclude',  os.path.join(easyblock.installdir.lstrip(os.sep), x)])
+
     _log.debug("The list of excluded files passed to fpm: %s", exclude_files_glob)
+    # use env to find FPM https://stackoverflow.com/questions/5658622/python-subprocess-popen-environment-path
     cmdlist = [
+        '/usr/bin/env',
```
boegel: this should be done by `run_cmd` in case `shell=False` is used?

easybuild/tools/package/utilities.py
```diff
@@ -105,51 +108,53 @@ def package_with_fpm(easyblock):
...
+        '--description', easyblock.cfg["description"],
+        '--url', easyblock.cfg["homepage"],
     ]
-    cmdlist.extend(exclude_files_glob)
+    cmdlist.extend(exclude_files_cmd)
 
     if build_option('debug'):
         cmdlist.append('--debug')
-
+    cmdlist.extend(deplist)
```
boegel: should we move down the definition of `deplist` here, to keep things together?

easybuild/tools/package/utilities.py
```diff
@@ -105,51 +108,53 @@ def package_with_fpm(easyblock):
...
     cmdlist.extend([
-        depstring,
         easyblock.installdir,
         easyblock.module_generator.get_module_filepath(),
     ])
     cmd = ' '.join(cmdlist)
+    print "cmd: %s" % cmdlist
```
boegel: please remove, or add to debug log statement below

easybuild/tools/package/utilities.py
```diff
@@ -105,51 +108,53 @@ def package_with_fpm(easyblock):
...
     _log.debug("The flattened cmdlist looks like: %s", cmd)
-    run_cmd(cmd, log_all=True, simple=True)
+    run_cmd(cmdlist, log_all=True, simple=True, cache=False, shell=False)
```
boegel: it's not immediately clear to me why you are disabling caching here?

test/framework/easyconfigs/test_ecs/t/toy/toy-0.0-gompi-1.3.12-test-description.eb
```diff
@@ -0,0 +1,34 @@
+name = 'toy'
+version = '0.0'
+versionsuffix = '-test'
+
+homepage = 'http://hpcugent.github.com/easybuild'
+description = "Toy C program. Now with `backticks'"
```
boegel: rather than adding a new easyconfig, can we just use `toy-0.0.eb`, read it, do:

```python
toytxt = re.sub('description = .*', 'description = "Toy C program. Now with `backticks'", toytxt)
```

and then write it to a temporary file to use for the test?

test/framework/package.py
```diff
@@ -224,6 +224,13 @@ def test_package(self):
             print "The Package File"
             print read_file(pkgfile)
 
+        ec_desc = EasyConfig(
+            os.path.join(test_easyconfigs, 't', 'toy', 'toy-0.0-gompi-1.3.12-test-description.eb'),
+            validate=False)
```
boegel: do a hard check on the `description` with the backtick in it, to make it clear why this is an important test case

test/framework/package.py
```diff
@@ -224,6 +224,13 @@ def test_package(self):
...
+        easyblock_desc = EB_toy(ec_desc)
+        easyblock_desc.run_all_steps(False)
+        pkgdir = package(easyblock_desc)
```
boegel: also check that the package is really there?

easybuild/tools/package/utilities.py
rjeschmi: possibly, but do you think shell=false will rely on path in all cases?

easybuild/tools/package/utilities.py
rjeschmi: the cache doesn't work (in that it tries to hash a list, I will try to fix that, but this command is also not cacheable).

rjeschmi:
I think this resolves the review and it passed package test locally.

easybuild/tools/package/utilities.py
```diff
@@ -96,6 +97,23 @@ def package_with_fpm(easyblock):
     pkgrel = package_naming_scheme.release(easyblock.cfg)
 
     _log.debug("Got the PNS values name: %s version: %s release: %s", pkgname, pkgver, pkgrel)
+    cmdlist = [
+        '/usr/bin/env',
```
boegel: @rjeschmi I really think this belongs in `run_cmd`...

easybuild/tools/package/utilities.py
```diff
@@ -105,51 +123,37 @@ def package_with_fpm(easyblock):
...
-    depstring = ''
+    deplist = []
     for dep in deps:
         if dep.get('external_module', False):
             _log.debug("Skipping dep marked as external module: %s", dep['name'])
         else:
             _log.debug("The dep added looks like %s ", dep)
             dep_pkgname = package_naming_scheme.name(dep)
-            depstring += " --depends %s" % quote_str(dep_pkgname)
+            deplist.extend(["--depends", dep_pkgname])
```
boegel: why not just extend `cmdlist`?

also, the `extend(deplist)` below should be right below this for loop, having something unrelated in between is confusing (although it doesn't matter anymore if you just extend to `cmdlist` straight away)

easybuild/tools/package/utilities.py
```diff
@@ -105,51 +123,37 @@ def package_with_fpm(easyblock):
...
-    ]
-    cmdlist.extend(exclude_files_glob)
+    exclude_files_cmd = []
+    for x in exclude_files_glob:
+        exclude_files_cmd.extend(['--exclude',  os.path.join(easyblock.installdir.lstrip(os.sep), x)])
 
-    if build_option('debug'):
-        cmdlist.append('--debug')
+    _log.debug("The list of excluded files passed to fpm: %s", exclude_files_glob)
+    # use env to find FPM https://stackoverflow.com/questions/5658622/python-subprocess-popen-environment-path
```
boegel: comment is out of place?

easybuild/tools/package/utilities.py
```diff
@@ -105,51 +123,37 @@ def package_with_fpm(easyblock):
...
-        '-s', 'dir',  # source
-        '--version', pkgver,
-        '--iteration', pkgrel,
-        '--description', quote_str(easyblock.cfg["description"]),
-        '--url', quote_str(easyblock.cfg["homepage"]),
-    ]
-    cmdlist.extend(exclude_files_glob)
+    exclude_files_cmd = []
+    for x in exclude_files_glob:
+        exclude_files_cmd.extend(['--exclude',  os.path.join(easyblock.installdir.lstrip(os.sep), x)])
```
boegel: same here, just use `cmdlist.extend`?

easybuild/tools/run.py
```diff
@@ -80,16 +80,18 @@ def run_cmd_cache(func):
     @functools.wraps(func)
     def cache_aware_func(cmd, *args, **kwargs):
         """Retrieve cached result of selected commands, or run specified and collect & cache result."""
-
-        # fetch from cache if available, cache it if it's not
-        if cmd in cache:
-            _log.debug("Using cached value for command '%s': %s", cmd, cache[cmd])
-            return cache[cmd]
+        # fetch from cache if available, cache it if it's not, but only on cmd strings
```
boegel: why only on strings?

if they're a list, you can just use `tuple(cmd)` as cache key?

easybuild/tools/run.py
```diff
@@ -148,9 +151,9 @@ def run_cmd(cmd, log_ok=True, log_all=False, simple=False, inp=None, regexp=True
         runLog = None
 
     readSize = 1024 * 8
-
+    _log.info('running cmd: %s ' % cmd)
     try:
-        p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+        p = subprocess.Popen(cmd, shell=shell, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
```
boegel: prefix `cmd` with `/usr/bin/env` if `shell` is `False` (and take into account that `cmd` may be either a string or a list/tuple)

test/framework/package.py
```diff
@@ -224,6 +224,28 @@ def test_package(self):
...
+            toy_file,
+            validate=False)
+        easyblock_desc = EB_toy(ec_desc)
+        easyblock_desc.run_all_steps(False)
+        pkgdir = package(easyblock_desc)
+        pkgfile = os.path.join(pkgdir, 'toy-0.0-gompi-1.3.12-test-eb-%s.1.rpm' % EASYBUILD_VERSION)
+        self.assertTrue(os.path.isfile(pkgfile))
+        pkgtxt = read_file(pkgfile)
+        regex_pkg_regex = re.compile(r"""DESCRIPTION:.*`backticks'.*""")
+        self.assertTrue(regex_pkg_regex.search(pkgtxt))
```
boegel:
```python
self.assertTrue(regex_pkg_regex.search(pkgtxt), "Pattern '%s' not found in: %s" % (regex.pattern, pkgtxt))
```

easybuild/tools/package/utilities.py
rjeschmi: It just kept it more like it was, but sure I can do it the other way now too since I have the cmdlist moved up.

easybuild/tools/run.py
rjeschmi: You only cache based on strings. It seems safer to harden the code for what is expected rather than trying to handle all sorts of things that aren't necessary (including lists since none of the cached commands are lists at the moment and it doesn't work).

easybuild/tools/run.py
rjeschmi: ok, I worry that it will have side effects, but it should be backwards compatible anyway.

easybuild/tools/run.py
rjeschmi: This one not done. Rest are done.

rjeschmi:
I've handled the rest of the comments save the one about checking for string in the cache code. All of the commands in CACHED_COMMANDS are strings.

If it is really a useful feature to allow it I think you should rethink how cached commands are handled. Right now it works and is better performing since it doesn't uselessly convert the command list to a tuple to check against a constant list of strings.

easybuild/tools/package/utilities.py
```diff
@@ -47,7 +47,7 @@
 from easybuild.tools.run import run_cmd
 from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME
 from easybuild.tools.utilities import import_available_modules, quote_str
-
+from subprocess import list2cmdline
```
boegel: @rjeschmi unused `import`?

easybuild/tools/run.py
boegel: Since we're only using strings in `CACHED_COMMANDS`, why do we need to check the type of `cmd` at all?

Since we never cache a non-string `cmd` right now, `cmd in cache` will always be `False`, and so we'll always call out to the actual `func`?

This looks like a case of premature optimisation here, especially for a command that's executed very rarely...

easybuild/tools/run.py
```diff
@@ -147,10 +150,15 @@ def run_cmd(cmd, log_ok=True, log_all=False, simple=False, inp=None, regexp=True
     else:
         runLog = None
 
-    readSize = 1024 * 8
+    if not shell and isinstance(cmd, list):
+        cmd.insert(0, '/usr/bin/env')
+    elif not shell and isinstance(cmd, str):
+        cmd = '/usr/bin/env %s' % cmd
```
boegel:
```python
# run command via 'env' if it is not run in a shell to trigger $PATH lookup of used command;
# cfr. <include-missing-reference>
if not shell:
    if isinstance(cmd, list):
        cmd.insert(0, '/usr/bin/env')
    elif isinstance(cmd, basestring):
        cmd = '/usr/bin/env ' + cmd
    else:
        raise EasyBuildError("Don't know how to prefix with /usr/bin/env for commands of type %s", type(cmd))
```

easybuild/tools/run.py
boegel: OK, ignore my rant, the `cmd in cache` will fail if `cmd` is of type `list`:

```
>>> [] in {}
TypeError: unhashable type: 'list'

File: easybuild/tools/run.py [hunk 20]
```diff
@@ -80,16 +80,18 @@ def run_cmd_cache(func):
...
     def cache_aware_func(cmd, *args, **kwargs):
         """Retrieve cached result of selected commands, or run specified and collect & cache result."""
-
-        # fetch from cache if available, cache it if it's not
-        if cmd in cache:
-            _log.debug("Using cached value for command '%s': %s", cmd, cache[cmd])
-            return cache[cmd]
+        # fetch from cache if available, cache it if it's not, but only on cmd strings
+        if isinstance(cmd, str):
+            if cmd in cache:
```
boegel: let's chain these together:

```python
if isinstance(cmd, basestring) and cmd in cache:
```
```

File: easybuild/tools/package/utilities.py (hunk 18 above)
Comment (rjeschmi) [inline easybuild/tools/package/utilities.py]: yeah sorry I was using it to debug, should have checked pylint.

File: easybuild/tools/run.py (hunk 20 above)
Comment (rjeschmi) [inline easybuild/tools/run.py]: you want me to prefix the next conditional with isinstance as well? Just seems harder to read...

Comment (boegel) [inline easybuild/tools/run.py]: it'll avoid the indent and the extra `else` at the end which is just a trimmed-down version of the inner `else`

Comment (rjeschmi) [inline easybuild/tools/run.py]: ok, thanks for the explanation on chat. I have it fixed now.

**Comment by rjeschmi** [comment]
I think I got them all this time.

File: easybuild/tools/package/utilities.py (hunk 18 above)
Comment (rjeschmi) [inline easybuild/tools/package/utilities.py]: I actually removed quote_str as well since I no longer use it with the cmd list version.

File: test/framework/package.py [hunk 21]
```diff
@@ -133,8 +133,7 @@ def mock_fpm(tmpdir):
     fpm = os.path.join(tmpdir, 'fpm')
     write_file(fpm, MOCKED_FPM % {
         "debug": ('', 'on')[DEBUG],
-        "debug_fpm_file": os.path.join(tmpdir, DEBUG_FPM_FILE)}
-    )
+        "debug_fpm_file": os.path.join(tmpdir, DEBUG_FPM_FILE)})
```
boegel: put `})` on new line instead, indenting like `)` was before

test/framework/package.py
```diff
@@ -224,6 +223,30 @@ def test_package(self):
             print "The Package File"
             print read_file(pkgfile)
 
+        toy_txt = read_file(os.path.join(test_easyconfigs, 't', 'toy', 'toy-0.0-gompi-1.3.12-test.eb'))
+        replace_str = '''description = """Toy C program. Now with `backticks' ''' "\n" '''and newlines"""'''
```
boegel: since you're using triple quotes, you can just insert a hard newline, but then it looks icky due to left indent

you can also:

```python
replace_str = '''description = """Toy C program. Now with `backticks'\n''''
replace_str += '''and newlines"""'''
```

test/framework/package.py
```diff
@@ -224,6 +223,30 @@ def test_package(self):
...
+        toy_file = os.path.join(self.test_prefix, 'toy-test-description.eb')
+        write_file(toy_file, toy_txt)
+
+        desc_backticks_regex = re.compile(r"""`backticks'""")
+        self.assertTrue(
+            desc_backticks_regex.search(toy_txt),
+            "Pattern '%s' found in: %s" % (desc_backticks_regex.pattern, toy_txt))
+        ec_desc = EasyConfig(
+            toy_file,
+            validate=False)
```
boegel: looks a bit funky here, not the usual code style

how about:

```python
regex = re.compile(r"""`backticks'""")
self.assertTrue(regex.search(toy_txt), "Pattern '%s' found in: %s" % (regex.pattern, toy_txt))
ec_desc = EasyConfig(toy_file, validate=False)
```

test/framework/package.py
rjeschmi: this is a syntax error. I assume you mean one less '

boegel: yes, first line has an extra trailing `'`... I didn't say I tested this ;)

boegel:
Going in, thanks for the effort @rjeschmi!

PR #2128, opened 2017-02-22
Enable extraction of patches from compressed files before applying them

mboisson:
In some cases, the patches which easybuild could download are available in gzip format. This PR will test if that is the case, and if so, will gunzip the patch prior to applying it.


rjeschmi:
I'm working on a test case for you. Is helpful since there seems to be some inconsistency with gzip commands lines (centos 7 ships with gzip 1.5)

    raise EasyBuildError('cmd "%s" exited with exitcode %s and output:\n%s', cmd, ec, stdouterr)
EasyBuildError: 'cmd "gunzip --force --keep /easybuild/easybuild-develop/easybuild-framework/test/framework/sandbox/sources/toy/toy-0.0_gzip.patch.gz" exited with exitcode 1 and output:\ngzip: unrecognized option \'--keep\'\nTry `gzip --help\' for more information.\n'

I'll push the test case to you. I think maybe you had to pipe it to stdout or something.

mboisson:
ok, @rjeschmi, I changed the gunzip command to use the built-in extract_file command, which does use -c and a pipe through stdout

rjeschmi:
I think we should put it into a tmp dir (even just in testing the extraction in place is a bit annoying). I'll PR it to you

test/framework/filetools.py
```diff
@@ -793,6 +793,12 @@ def test_apply_patch(self):
         pattern = "I'm a toy, and very proud of it"
         self.assertTrue(pattern in patched)
 
+        toy_patch_gz = os.path.join(testdir, 'sandbox', 'sources', 'toy', 'toy-0.0_gzip.patch.gz')
+        self.assertTrue(ft.apply_patch(toy_patch_gz, path))
```
boegel: hmm, `path` still contains the previous patched file, no?

we should clean out `self.test_prefix`, and call `extract_file` again on `toy-0.0.tar.gz` before applying the zipped patch?

easybuild/tools/filetools.py
```diff
@@ -763,6 +763,18 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
     apatch = os.path.abspath(patch_file)
     adest = os.path.abspath(dest)
 
+    if apatch.endswith(".gz"):
```
boegel: This will only work if the patch if gzipped, what about `.zip`, `.bz2`, etc.?

Can we think of a better/more generic way to check whether we're dealing with a zipped file?

Also, there's a hard check in `fetch_patches` in `easyblock.py` on `.patch` extension, to distinguish between actual patch files and files that listed in `patches` are simply being copied, that will need to be enhanced too (maybe by changing the check to:

```python
if '.patch' not in patch_spec[0]:
```

easybuild/tools/filetools.py
```diff
@@ -763,6 +763,18 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
+        workdir = tempfile.mkdtemp(prefix='eb-patch-')
+        _, patchfilename_gz = os.path.split(apatch)
+        patchfilename, _ = os.path.splitext(patchfilename_gz)
+        dest_tmp = os.path.join(workdir, patchfilename)
+        _log.debug("Ungzipping the patch to: %s", workdir)
```
boegel: `s/Ungzipping/Extracting/`

easybuild/tools/filetools.py
```diff
@@ -763,6 +763,18 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
+
+        # gunzipping the patch.
+        extract_file(apatch, workdir)
+        # remove the '.gz' extension
+        apatch = dest_tmp
```
boegel: `extract_file` returns the path to the extract file, no need to guess...

this simplifies this quite a bit:

```python
apatch = extract_file(apatch, tempfile.mkdtemp(prefix='eb-patch-'))
```

test/framework/filetools.py
rjeschmi: Yeah I patch the same file again as I figured that would be sufficient. We could do it the other way too. Mine also tests a dependent patch FWIW.

boegel: ah, I missed the `very very` bit, ok, fine by me, but let's add a comment then to make it clear that the zipped patch is an incremental one

easybuild/tools/filetools.py
rjeschmi: The extract_cmd dict might be helpful. Maybe we should store it somewhere accessible and use the keys?

boegel: well, you could just call `extract_cmd` to get the matching command?

mboisson:
I think I have addressed all comments.

easybuild/tools/filetools.py
mboisson: I changed this for checking whether apatch does *not* end with ".patch". Thoughts ?

boegel:
@mboisson you broke two tests, please try looking into fixing them?

easybuild/tools/filetools.py
```diff
@@ -763,6 +763,13 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
     apatch = os.path.abspath(patch_file)
     adest = os.path.abspath(dest)
 
+    # Attempt extracting the patch if it does not end in .patch (may end in .gz, .zip, etc.)
+    if not apatch.endswith(".patch"):
+        workdir = tempfile.mkdtemp(prefix='eb-patch-')
+        _log.debug("Extracting the patch to: %s", workdir)
+        # extracting the patch
+        apatch = extract_file(apatch, workdir)
```
boegel: @rjeschmi pointed out to me that `extract_file` actually doesn't return the path to the unpacked file (but to it's parent directory). This is probably also why one of the tests (i.e. `test_apply_patch `) now fails.

It probably *should* though, in case the unpacked file is again a file (rather than a directory, which is usually what we're dealing with)...

easybuild/tools/filetools.py
```diff
@@ -763,6 +763,13 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
     apatch = os.path.abspath(patch_file)
     adest = os.path.abspath(dest)
 
+    # Attempt extracting the patch if it does not end in .patch (may end in .gz, .zip, etc.)
+    if not apatch.endswith(".patch"):
```
boegel: this is a bit too generic, you should still require to have `.patch` in the filename, see my earlier remark about `patches` that can also contain files that are being copied rather than actual patches; see for example https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.2.19-gompi-2016.09-LAPACK-3.6.1.eb#L35

This seems to be the cause of the other failing test (`test_extended_dry_run`)

mboisson: Ok. In the latest commit, it has to end in ".patch.gz", ".patch.bz2" or ".patch.xz". 

I don't think we can assume other formats are not an archive with a bunch of directories in it.

easybuild/tools/filetools.py
mboisson: This should be fixed in the latest commit where I rebuild the patch name with the extract directory and the original name (minus the extension for compression)

rjeschmi:
I pushed a PR to @mboisson to fix the remaining test.

easybuild/tools/filetools.py
```diff
@@ -763,6 +763,21 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
     apatch = os.path.abspath(patch_file)
     adest = os.path.abspath(dest)
 
+    # Attempt extracting the patch if it ends in .patch.gz, .patch.bz2, .patch.xz
+    # split in name + extension
+    apatch_root, apatch_file = os.path.split(apatch)
+    apatch_name, apatch_extension = os.path.splitext(apatch_file)
+    if apatch_extension in ['.gz','.bz2','.xz']:
```
boegel: `.zip`?

easybuild/tools/filetools.py
```diff
@@ -763,6 +763,21 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
...
+        # split again to get the second extension
+        apatch_subname, apatch_subextension = os.path.splitext(apatch_name)
+        if apatch_subextension == ".patch":
+            workdir = tempfile.mkdtemp(prefix='eb-patch-')
+            _log.debug("Extracting the patch to: %s", workdir)
+            # extracting the patch
+            apatch_dir = extract_file(apatch, workdir)
+            apatch = os.path.join(apatch_dir, apatch_name)
+
```
boegel: avoid double blank line

easybuild/tools/filetools.py
boegel: also, sort these alphabetically while you're at it?

mboisson: I did not add zip because zip can be an archive. I'm only supporting ".patch.<compression extension>", which are not archive. 

I don't think we can make the script clever enough to support archive of multiple patches.

boegel: OK, then make that clear with a comment in the code, good enough I guess.

mboisson: Done

easybuild/tools/filetools.py
mboisson: and done

mboisson:
Is this getting merged ?

boegel:
@mboisson Thanks for the reminder, I was going to get to it eventually (I postponed merging this until after the EB v3.1.2 release).

Going in, thanks again!

PR #2129, opened 2017-02-23
clarify unstable/closed PR warning message

migueldiascosta:


easybuild/tools/github.py
```diff
@@ -382,7 +382,8 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
 
     if not(pr_data['merged']):
         if not stable or closed:
-            print "\n*** WARNING: Using easyconfigs from unstable/closed PR #%s ***\n" % pr
+            state_msg = "unstable (failed tests or merge conflict)" if not stable else "closed (but not merged)"
```
boegel: I think pending tests also result in 'unstable' state?

migueldiascosta: sorry, right (that was the original scenario...) 

can we distinguish between pending and failed?

boegel: Hmm, we may be able to, but the GitHub API documentation isn't very clear on that...

boegel:
lgtm

Issue #2055, opened 2016-12-10
problem building Python-2.7.5-ictce-6.2.5.eb and mpi4py extension  with --rpath


pescobar:
The crash appears when building the `mpi4py` extension.

The interesting messages start in line 56084

```
56084 icc: command line warning #10006: ignoring unknown option '-rpath'
56085 ipo: warning #11016: Warning unknown option -Xlinker
56086 /usr/bin/ld: cannot find -linker
56087 failure.
56088 removing: _configtest.c _configtest.o
56089 error: Cannot link MPI programs. Check your configuration!!!
```

Here is the full debug log:
https://gist.github.com/d5190e29433eab8e8544b8d6c98e1497

jdonners:
Hi,

this is most probably caused by the rpath_args.py script, which removes the flag --enable-new-dtags from the command line. The compiler passes this flag to the linker using '-Wl, --enable-new-dtags' or '-Xlinker --enable-new-dtags'. The script would then remove --enable-new-dtags and the compiler basically passes whatever random flag is next on the command-line on to the linker. I created a simple fix to the problem, which replaces --enable-new-dtags with --disable-new-dtags:

https://github.com/sara-nl/easybuild-framework/commit/3ec052c05b3b7e47c3922aaab023a02d4a1bab8d

I'll try to add a pull request to this issue..

jdonners:
hmm, I'm not authorized to do that. Should I have created a separate branch from develop?

boegel:
@jdonners Does the documentation at http://easybuild.readthedocs.io/en/latest/Contributing.html#pull-requests help?

jdonners:
pull request is here: https://github.com/hpcugent/easybuild-framework/pull/2131

boegel:
@pescobar @jdonners's PR #2131 with the fix for this issue is merged into `develop`, please check whether this solves the problem you were seeing?

jdonners:
I didn't check the exact version that @pescobar used, but I could build Python-2.7.12-intel-2016b.eb including mpi4py using --rpath

boegel:
OK, I'll close this as fixed, thanks for the fix & feedback @jdonners!

PR #2131, opened 2017-02-24
#2055: replace --enable-new-dtags with --disable-new-dtags instead of remove…

jdonners:
… from command-line

### DIFF
--- a/easybuild/scripts/rpath_args.py
+++ b/easybuild/scripts/rpath_args.py
@@ -91,11 +91,14 @@ while idx < len(args):
             # it doesn't make much sense, and it can also break the build because it may result in reordering lib paths
             cmd_args.append('-L%s' % lib_path)
 
-    # filter out --enable-new-dtags if it's used;
-    # this would result in copying rpath to runpath, meaning that $LD_LIBRARY_PATH is taken into account again
+    # replace --enable-new-dtags with --disable-new-dtags if it's used;
+    # --enable-new-dtags would result in copying rpath to runpath, meaning that $LD_LIBRARY_PATH is taken into account again
+    # --enable-new-dtags is not removed but replaced to prevent issues when the linker flag is forwarded from the compiler 
+    # to the linker with an extra prefixed flag (either -Xlinker or -Wl,). In that case, the compiler would erroneously pass
+    # the next random argument to the linker.
     elif arg == '--enable-new-dtags':
         cmd_args.append('--disable-new-dtags');
-    elif arg != '--enable-new-dtags':
+    else:
         cmd_args.append(arg)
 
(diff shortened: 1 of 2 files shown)


easybuild/scripts/rpath_args.py
```diff
@@ -93,6 +93,8 @@
 
     # filter out --enable-new-dtags if it's used;
     # this would result in copying rpath to runpath, meaning that $LD_LIBRARY_PATH is taken into account again
+    elif arg == '--enable-new-dtags':
+        cmd_args.append('--disable-new-dtags');
     elif arg != '--enable-new-dtags':
```
boegel: with introducing the `arg == '--enable-new-dtags'` above, we can just make this `else:` rather than `elif ...:`?

easybuild/scripts/rpath_args.py
```diff
@@ -93,6 +93,8 @@
 
     # filter out --enable-new-dtags if it's used;
```
boegel: @jdonners please also fix this comment, to something like `# replace --enable-new-dtags with --disable-new-dtags if it's used`

the last bit makes me wonder: should we just *always* inject `--disable-new-dtags` (maybe combined with filtering out `--enable-new-dtags`, to be sure)?

boegel: actually, we already inject `--disable-new-dtags`, see line 114... so how can this patch fix anything?

jdonners: it fixes the cases where the command-line consists of '... -Wl, --enable-new-dtags -fomit-frame-pointer'. In the current version, rpath_args.py would modify this into '-Wl, -fomit-frame-pointer' and the option -fomit-frame-pointer is passed to the linker, which fails because it is an unknown option. The fix just passes the same option --disable-new-dtags multiple times to the linker, which shouldn't be a problem.

boegel: Ah, ok, good point...

Maybe clarify that with a comment here, since I suspect otherwise people (e.g. me) will go wonder again why we don't just drop `--enable-new-dtags` or why we possible add `--disable-new-dtags` multiple times... :)

boegel:
@jdonners Are you up for making the requested changes so we can merge this in? See http://easybuild.readthedocs.io/en/latest/Contributing.html#updating-existing-pull-requests

jdonners:
The checks will fail, since it expects only once --disable-new-dtags, while it now includes the same flag twice.

boegel:
@jdonners Can you also change the test accordingly? see `test/framework/toolchain.py`

boegel:
Thanks for the fix @jdonners!

PR #2132, opened 2017-02-24
alphabetically sort functions and methods in module_generator module

damianam:
Just to have a cleaner file for another PR on top


easybuild/tools/module_generator.py
```diff
@@ -226,6 +302,10 @@ class ModuleGeneratorTcl(ModuleGenerator):
     LOAD_REGEX = r"^\s*module\s+load\s+(\S+)"
     LOAD_TEMPLATE = "module load %(mod_name)s"
 
+    def __init__(self, *args, **kwargs):
+        """ModuleGeneratorTcl constructor."""
+        super(ModuleGeneratorTcl, self).__init__(*args, **kwargs)
```
boegel: @damianam this is useless, if there's nothing to do in the constructor, then don't override it

boegel:
@damianam It's hard to tell from the diff, but this is just literally moving stuff around without any changes to the code in the methods, right?

One thing I'm slightly worried about is that this is going to clash with #2110 which adds a `set_as_default` method in `module_generator.py`. As far as I can tell, that's the only PR it would clash with.

I expect that resolving the conflict here rather than in #2110 is going to be easier, so I'll get #2110 merged first, and then help out if needed with resolving the conflict in here (if needed).

easybuild/tools/module_generator.py
damianam: Well, at least now it is **consistently** useless. Before one of the classes had an useless constructor and one didn't. I included it in both because I thought that maybe you had an useless constructor there for some obscure reason I didn't know about...

damianam:
Merging https://github.com/hpcugent/easybuild-framework/pull/2110 seems reasonable, but that means that this PR and the one that was coming after this one will have to wait do don't do things multiple times. Which is fine for me TBH, just a bit of pain to do things twice.

easybuild/tools/module_generator.py
boegel: No, there is no reason, so better clean up the empty constructor in `ModuleGeneratorLua` instead (and drop this one here).

boegel:
@damianam I hope we can get #2110 merged soon, it's up to @victorusu now to tackle the last remaining remarks there

boegel:
Did some checks to confirm that this is just shuffling things around without making changes, tests confirm, so going in, thanks for the effort @damianam!

PR #2134, opened 2017-02-27
add support for backing up modules with --module-only via --backup-modules

damianam:
With this PR easybuild will create a backup of existing modules when using `--modules-only`. 2 extra comments:

1. While I was at it, I also used the new `back_up_file` function to rotate logs
2. I think it makes sense to print the `diff` of modules in `stdout`. Thoughts?


boegel:
@damianam This needs more work (e.g. review & tests), and since neither of us will have time to work on this in the coming days, let's bump the milestone to `3.3.0` on this and revisit it later.

damianam:
@boegel take a look when you can. I'll see if tomorrow I can write some tests for it.

easybuild/framework/easyblock.py
```diff
@@ -2171,6 +2172,14 @@ def make_module_step(self, fake=False):
                 for line in txt.split('\n'):
                     self.dry_run_msg(' ' * 4 + line)
         else:
+            module_only = build_option('module_only')
+            if module_only and os.path.exists(mod_filepath):
+               warning_msg = "Old module file found. Backing it up in %s. Diff is:\n%s"
+               mod_bck_filepath = back_up_file(mod_filepath, backup_extension="bck")
+               (mod_diff, _) = run_cmd("diff %s %s" % (mod_bck_filepath, mod_filepath))
```
boegel: use `diff -u`?

easybuild/framework/easyblock.py
```diff
@@ -2171,6 +2172,14 @@ def make_module_step(self, fake=False):
                 for line in txt.split('\n'):
                     self.dry_run_msg(' ' * 4 + line)
         else:
+            module_only = build_option('module_only')
+            if module_only and os.path.exists(mod_filepath):
+               warning_msg = "Old module file found. Backing it up in %s. Diff is:\n%s"
+               mod_bck_filepath = back_up_file(mod_filepath, backup_extension="bck")
```
boegel: we should make this a hidden module, no?

also, this should be configurable, people should be able to disable the backup if they're confident enough (but I guess having it enabled by default makes sense)

easybuild/tools/filetools.py
```diff
@@ -1168,16 +1198,12 @@ def move_logs(src_logfile, target_logfile):
...
-                i = 0
-                oldlog_backup = "%s_%d" % (new_log_path, i)
-                while os.path.exists(oldlog_backup):
-                    i += 1
-                    oldlog_backup = "%s_%d" % (new_log_path, i)
-                shutil.move(new_log_path, oldlog_backup)
-                _log.info("Moved existing log file %s to %s" % (new_log_path, oldlog_backup))
+                back_up_file(new_log_path)
 
+            # remove first to ensure portability
```
boegel: clarify portability?

easybuild/tools/filetools.py
```diff
@@ -1150,6 +1151,35 @@ def rmtree2(path, n=3):
...
+        dst_file = "%s_%d" % (src_file, i)
+        while os.path.exists(dst_file):
+            i += 1
+            dst_file = "%s_%d" % (src_file, i)
+
+    return dst_file
+
+
+def back_up_file(src_file, backup_extension=""):
+    """Backs up a file appending a backup extension and a number to it. Returns the name of the backup"""
```
boegel: clarify: the number is optional (only when existing backups are already there)

use `:param:` to document arguments

also support `hide` named argument to backup to a hidden file

boegel:
@damianam For functions you touch/change, please check whether they have a dedicated unit test already (and maybe enhance it while you're at it).

When adding new functions, add a dedicated test for them.

easybuild/framework/easyblock.py
damianam: Making it a hidden module makes sense just for TCL modules. I had lua modules in mind, but since there are still people out there with TCL modules I agree with you.

Regarding making it configurable..... I am not so sure.

Pros:
1. Flexibility.

Cons:
1. Yet another option cluttering the `--help`
2. More code to maintain

However, if you are using TCL modules with Lmod, these will be recognized as modules that can be loaded, which is not what you want. Any other combination (TCL modules with Tmod, or lua modules with Lmod), will result in an invisible (to the modules tool) backup. So yes, I guess I should also prepare an option for it.....

damianam:
@boegel I addressed all your comments. The code hasn't been tested yet, and the tests are still missing, but most of the bits and pieces should be there already.

damianam:
Tests are complete and code seems to be working. @easybuilders/easybuild-framework-maintainers  please review.

easybuild/framework/easyblock.py
```diff
@@ -2202,8 +2202,30 @@ def make_module_step(self, fake=False):
...
+                # diff will return 1 if there are differences. We have to set log_ok and log_all to False to ignore the
+                # return code
+                (mod_diff, _) = run_cmd("diff -u %s %s" % (mod_bck_filepath, mod_filepath), log_ok=False, log_all=False)
+                if mod_diff:
+                    self.log.info(warning_msg + ' Diff is:\n%s', mod_bck_filepath, mod_diff)
+                    print_warning(warning_msg % mod_bck_filepath + ' Diff is:\n%s' % mod_diff)
+                else:
+                    self.log.info(warning_msg + ' No differences found.', mod_bck_filepath)
+                    print_warning(warning_msg % mod_bck_filepath + ' No differences found.')
+                self.log.info("Module file %s written: %s", mod_filepath, txt)
```
boegel: this whole block is rather big to just shove it into `make_module_step` like this, it's better to move it into a separate method named `back_up_existing_module_file` or something like that?

easybuild/framework/easyblock.py
```diff
@@ -2202,8 +2202,30 @@ def make_module_step(self, fake=False):
...
         else:
-            write_file(mod_filepath, txt)
-            self.log.info("Module file %s written: %s", mod_filepath, txt)
+            module_only = build_option('module_only')
+            backup_modules = build_option('backup_modules')
+            if module_only and os.path.exists(mod_filepath) and backup_modules and not fake:
+                warning_msg = "Old module file found. Backing it up in %s."
+                hidden = False
+                if isinstance(self.module_generator, ModuleGeneratorTcl):
+                     hidden = True
```
boegel: indent is off here?

Also, why only do hidden for Tcl modules?

The backup should always be a hidden module imho, you don't want to expose the backup module to users?

Maybe we should also back up outside of the module tree, e.g. in the current directory (or at a location specified via `--backup-modules`?

easybuild/framework/easyblock.py
```diff
@@ -2202,8 +2202,30 @@ def make_module_step(self, fake=False):
                 for line in txt.split('\n'):
                     self.dry_run_msg(' ' * 4 + line)
         else:
-            write_file(mod_filepath, txt)
-            self.log.info("Module file %s written: %s", mod_filepath, txt)
+            module_only = build_option('module_only')
+            backup_modules = build_option('backup_modules')
+            if module_only and os.path.exists(mod_filepath) and backup_modules and not fake:
+                warning_msg = "Old module file found. Backing it up in %s."
```
boegel: I would move the `os.path.exists` check down into the `back_up_existing_module_file` method, and just exit early with a clear log message to say there's nothing to back up?

easybuild/framework/easyblock.py
```diff
@@ -2202,8 +2202,30 @@ def make_module_step(self, fake=False):
...
+                hidden = False
+                if isinstance(self.module_generator, ModuleGeneratorTcl):
+                     hidden = True
+                else:
+                    hidden = False
```
boegel: `hidden` was set to `False` already...

easybuild/framework/easyblock.py
```diff
@@ -2202,8 +2202,30 @@ def make_module_step(self, fake=False):
...
+                if mod_diff:
+                    self.log.info(warning_msg + ' Diff is:\n%s', mod_bck_filepath, mod_diff)
+                    print_warning(warning_msg % mod_bck_filepath + ' Diff is:\n%s' % mod_diff)
+                else:
+                    self.log.info(warning_msg + ' No differences found.', mod_bck_filepath)
+                    print_warning(warning_msg % mod_bck_filepath + ' No differences found.')
+                self.log.info("Module file %s written: %s", mod_filepath, txt)
+            else:
+                write_file(mod_filepath, txt)
+                self.log.info("Module file %s written: %s", mod_filepath, txt)
```
boegel: This is there twice now, why not just do this after backing up?

You would need delay the diff'ing a bit...

easybuild/framework/easyblock.py
```diff
@@ -2202,8 +2202,30 @@ def make_module_step(self, fake=False):
...
+                else:
+                    hidden = False
+                mod_bck_filepath = back_up_file(mod_filepath, backup_extension="bck", hidden=hidden)
+                write_file(mod_filepath, txt)
+                # diff will return 1 if there are differences. We have to set log_ok and log_all to False to ignore the
+                # return code
+                (mod_diff, _) = run_cmd("diff -u %s %s" % (mod_bck_filepath, mod_filepath), log_ok=False, log_all=False)
+                if mod_diff:
+                    self.log.info(warning_msg + ' Diff is:\n%s', mod_bck_filepath, mod_diff)
+                    print_warning(warning_msg % mod_bck_filepath + ' Diff is:\n%s' % mod_diff)
```
boegel: The devil himself invented code duplication. Just construct the message once, and log/print afterwards:

```python
if mod_diff:
    diff_msg = (warning_msg + " Diff is:\n%s") % (mod_bck_filepath, mod_diff)
else:
    diff_msg = (warning_msg + "No differences found.") % mod_bck_filepath

self.log.info(diff_msg)
print_warning(diff_msg)

File: easybuild/tools/filetools.py [hunk 11]
```diff
@@ -1162,6 +1163,50 @@ def rmtree2(path, n=3):
     else:
         _log.info("Path %s successfully removed." % path)
 
+def find_backup_name_candidate(src_file):
+    """Returns a non-existing file to be used as destination for rotating backups"""
```
boegel: Not really rotating, I think, i.e. we don't recycle filenames after N backups?

easybuild/tools/filetools.py
```diff
@@ -1162,6 +1163,50 @@ def rmtree2(path, n=3):
...
+    dst_file = src_file
```
boegel: Hmm, shouldn't the default be `None`?

If the specified filepath doesn't exist, you can't back up that file either?

easybuild/tools/filetools.py
```diff
@@ -1162,6 +1163,50 @@ def rmtree2(path, n=3):
...
+    """
+    if hidden:
+        leading_string = "."
+    else:
+        leading_string = ""
+
+    if backup_extension:
+        trailing_string = ".%s" % backup_extension
+    else:
+        trailing_string = ""
```
boegel: Shorter/cleaner (imho):

```python
fn_pref, fn_suff = '', ''
if hidden:
    fn_pref = '.'
if backup_extension:
    fn_suff = '.%s' % backup_extension
```

Also, I prefer using `''` rather than `""` for strings without spaces (I find it easier on the eyes), but that's nitpicking ;)

easybuild/tools/filetools.py
```diff
@@ -1162,6 +1163,50 @@ def rmtree2(path, n=3):
...
+
+    backup_file = find_backup_name_candidate("%s/%s%s%s" % \
```
```python
src_dir, src_fn = os.path.split(src_file)
backup_filepath = find_backup_name_candidate(os.path.join(src_dir, fn_pref + src_fn + fn_suff))

File: easybuild/tools/filetools.py [hunk 15]
```diff
@@ -1180,16 +1225,12 @@ def move_logs(src_logfile, target_logfile):
...
-                oldlog_backup = "%s_%d" % (new_log_path, i)
-                while os.path.exists(oldlog_backup):
-                    i += 1
-                    oldlog_backup = "%s_%d" % (new_log_path, i)
-                shutil.move(new_log_path, oldlog_backup)
-                _log.info("Moved existing log file %s to %s" % (new_log_path, oldlog_backup))
+                back_up_file(new_log_path)
 
+            # remove first to ensure portability (shutil.move might fail when overwriting files in some systems)
+            remove_file(new_log_path)
```
boegel: this should be part of `move_file` (under the condition that the target location already exists)?

easybuild/tools/module_generator.py
```diff
@@ -185,7 +186,8 @@ def prepare(self, fake=False):
         mkdir(os.path.dirname(mod_filepath), parents=True)
 
         # remove module file if it's there (it'll be recreated), see EasyBlock.make_module
-        if os.path.exists(mod_filepath) and not build_option('extended_dry_run'):
+        if os.path.exists(mod_filepath) and not build_option('extended_dry_run') and \
+                (not build_option('backup_modules') or fake):
             self.log.debug("Removing existing module file %s", mod_filepath)
             os.remove(mod_filepath)
```
boegel: Hmm, isn't this duplicate code since we're already removing the module file in `check_readiness_step` as well?

Also here: we should rename here already rather than not remove?

easybuild/tools/options.py
```diff
@@ -324,6 +324,8 @@ def override_options(self):
                                             None, 'store_true', False),
             'allow-use-as-root-and-accept-consequences': ("Allow using of EasyBuild as root (NOT RECOMMENDED!)",
                                                           None, 'store_true', False),
+            'backup-modules': ("Back up an existing module file, if any. Only works when using --module-only",
+                            None, 'store_true', False),
```
boegel: align the `None` with the first `"` above

test/framework/filetools.py
```diff
@@ -494,6 +494,53 @@ def test_guess_patch_level(self):
         ]:
             self.assertEqual(ft.guess_patch_level([patched_file], self.test_buildpath), correct_patch_level)
 
+    def test_back_up_file(self):
+        """Test back_up_file function."""
+        fh, fp = tempfile.mkstemp()
+        os.close(fh)
```
boegel: or just use `fp = os.path.join(self.test_prefix, 'test.txt')`

test/framework/filetools.py
```diff
@@ -494,6 +494,53 @@ def test_guess_patch_level(self):
...
+
+        ft.write_file(fp, 'foobar')
+
+        # Test simple file backup
+        ft.back_up_file(fp)
+        self.assertEqual(ft.read_file(fp+'_0'), 'foobar')
```
boegel: spaces around `+`

test/framework/filetools.py
```diff
@@ -494,6 +494,53 @@ def test_guess_patch_level(self):
         ]:
             self.assertEqual(ft.guess_patch_level([patched_file], self.test_buildpath), correct_patch_level)
 
+    def test_back_up_file(self):
+        """Test back_up_file function."""
+        fh, fp = tempfile.mkstemp()
+        os.close(fh)
+
+        ft.write_file(fp, 'foobar')
```
boegel: define a variable for the file contents rather than copy-pasting `'foobar'` everywhere?

test/framework/filetools.py
```diff
@@ -494,6 +494,53 @@ def test_guess_patch_level(self):
...
+
+        # Test simple file backup
+        ft.back_up_file(fp)
+        self.assertEqual(ft.read_file(fp+'_0'), 'foobar')
+        self.assertEqual(ft.read_file(fp), 'foobar')
+        # Test hidden simple file backup
+        ft.back_up_file(fp, hidden=True)
+        self.assertEqual(ft.read_file(os.path.dirname(fp)+'/.'+os.path.basename(fp)), 'foobar')
```
boegel:
```python
dir, fn = os.path.split(fp)
self.assertEqual(ft.read_file(os.path.join(dir, '.' + fn)), test_txt)

File: test/framework/filetools.py [hunk 22]
```diff
@@ -494,6 +494,53 @@ def test_guess_patch_level(self):
...
 
+    def test_back_up_file(self):
+        """Test back_up_file function."""
+        fh, fp = tempfile.mkstemp()
+        os.close(fh)
+
+        ft.write_file(fp, 'foobar')
+
+        # Test simple file backup
+        ft.back_up_file(fp)
```
boegel: maybe check with `os.listdir` that you now only have two files, the original and the backup, nothing else?

test/framework/filetools.py
```diff
@@ -494,6 +494,53 @@ def test_guess_patch_level(self):
...
+        self.assertEqual(ft.read_file(fp+'_0'), 'foobar')
+        self.assertEqual(ft.read_file(fp), 'foobar')
+        # Test hidden simple file backup
+        ft.back_up_file(fp, hidden=True)
+        self.assertEqual(ft.read_file(os.path.dirname(fp)+'/.'+os.path.basename(fp)), 'foobar')
+        self.assertEqual(ft.read_file(fp), 'foobar')
+        # Test simple file backup with extension
+        ft.back_up_file(fp, backup_extension="bck")
+        self.assertEqual(ft.read_file(fp+'.bck'), 'foobar')
```
boegel: spaces around `+`

test/framework/filetools.py
```diff
@@ -494,6 +494,53 @@ def test_guess_patch_level(self):
...
+        self.assertEqual(ft.read_file(fp), 'foobar')
+        # Test hidden simple file backup with extension
+        ft.back_up_file(fp, backup_extension="bck", hidden=True)
+        self.assertEqual(ft.read_file(os.path.dirname(fp)+'/.'+os.path.basename(fp)+'.bck'), 'foobar')
```
boegel: use `os.path.split` & `os.path.join`, spaces around `+`

same remarks below

test/framework/toy_build.py
```diff
@@ -1223,6 +1224,101 @@ def test_module_only(self):
...
+        self.mock_stderr(False)
+        self.assertFalse(re.search(warning, stderr, re.M))
+
+        # Prepare for the real test
+        ec_files_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'test_ecs')
+        ec_file = os.path.join(ec_files_path, 't', 'toy', 'toy-0.0-deps.eb')
+        toy_mod = os.path.join(self.test_installpath, 'modules', 'all', 'toy', '0.0-deps')
+
+        # only consider provided test modules
+        self.reset_modulepath([os.path.join(os.path.dirname(os.path.abspath(__file__)), 'modules')])
```
boegel: why is this needed?

test/framework/toy_build.py
```diff
@@ -1223,6 +1224,101 @@ def test_module_only(self):
...
+        warning = "WARNING: Old module file found. Backing it up in .* No differences found"
+
+        self.eb_main(args, do_build=True, raise_error=True)
+        self.assertTrue(os.path.exists(toy_mod))
+        self.mock_stderr(True)
+        self.eb_main(args + ['--backup-modules'], do_build=True, raise_error=True)
+        stderr = self.get_stderr()
+        self.mock_stderr(False)
+        self.assertTrue(os.path.exists(toy_mod))
+        self.assertTrue(os.path.exists(os.path.join(os.path.dirname(toy_mod), '.'+os.path.basename(toy_mod)+'.bck')))
```
boegel: use `os.path.split` & `os.path.join`, spaces around `+`

test/framework/toy_build.py
```diff
@@ -1223,6 +1224,101 @@ def test_module_only(self):
...
+
+        warning = "WARNING: Old module file found. Backing it up in .* Diff is:"
+        diff = "some difference\n"
+        with open(toy_mod, "a") as fh:
+            fh.write(diff)
+        self.mock_stderr(True)
+        self.eb_main(args + ['--backup-modules'], do_build=True, raise_error=True, verbose=True)
+        stderr = self.get_stderr()
+        self.mock_stderr(False)
+        self.assertTrue(os.path.exists(os.path.join(os.path.dirname(toy_mod), '.'+os.path.basename(toy_mod)+'.bck_0')))
```
boegel: use `os.path.split` & `os.path.join`, spaces around `+`

test/framework/toy_build.py
```diff
@@ -1223,6 +1224,101 @@ def test_module_only(self):
...
+        stderr = self.get_stderr()
+        self.mock_stderr(False)
+        self.assertTrue(os.path.exists(toy_mod))
+        self.assertTrue(os.path.exists(os.path.join(os.path.dirname(toy_mod), '.'+os.path.basename(toy_mod)+'.bck')))
+        self.assertTrue(re.search(warning, stderr, re.M))
+
+        warning = "WARNING: Old module file found. Backing it up in .* Diff is:"
+        diff = "some difference\n"
+        with open(toy_mod, "a") as fh:
+            fh.write(diff)
```
boegel: use `write_file` (it has an append mode)

test/framework/toy_build.py
```diff
@@ -1223,6 +1224,101 @@ def test_module_only(self):
...
+            '--debug',
+            '--unittest-file=%s' % self.logfile,
+            '--robot=%s' % ec_files_path,
+            '--module-only',
+            '--force',
+            '--disable-cleanup-tmpdir'
+        ]
+        args = common_args + ['--module-syntax=Tcl']
+
+        warning = "WARNING: Old module file found. Backing it up in .* No differences found"
```
boegel: Hmm, should this be a *warning*? It's just an informative message, no?

test/framework/toy_build.py
```diff
@@ -1223,6 +1224,101 @@ def test_module_only(self):
...
+        self.mock_stderr(False)
+        self.assertTrue(os.path.exists(os.path.join(os.path.dirname(toy_mod), '.'+os.path.basename(toy_mod)+'.bck_0')))
+        self.assertTrue(re.search(warning, stderr, re.M))
+        self.assertTrue(re.search(diff, stderr, re.M))
+
+        # Test also with lua syntax if Lmod is available. In particular, that the backup is not hidden
+        if isinstance(self.modtool, Lmod):
+            args = common_args + ['--module-syntax=Lua']
+
+            toy_mod = os.path.join(self.test_installpath, 'modules', 'all', 'toy', '0.0-deps.lua')
```
boegel: rather than copy-pasting all the checks, we should hoist them in a helper function, and simplify the test to

```python
def test_backup_modules(self):
    """Test use of --backup-modules."""
    def run_checks():
        ...

    ...

    toy_mod_fn = '0.0-deps'
    run_checks()

    if isinstance(self.modtool, Lmod):
      toy_mod_fn = '0.0-deps.lua'
      run_checks()

easybuild/framework/easyblock.py
damianam: Because lua modules have to end in `.lua`, otherwise they aren't treated as modules by Lmod, and don't show up with `ml av`. So there is no need to keep them hidden since the backup extension avoids that file ends in `.lua`.

Tcl modules on the other need to be hidden. Which is not that nice, since Lmod can still show them with `--show-hidden`. But that's the best compromise IMO. The alternative would be to remove the module header.

Backing up outside of the module tree kind of defeats the purpose of the backup. In my opinion the backup should be used for quick verification of the changes and for doing a "rollback" if necessary. If I have the current module and the backup in different files it slightly makes it more complicated and cumbersome, and more difficult to directly see if there was a previous version at any point with a simple `ls`

boegel:
Going in, thanks @damianam!

PR #2135, opened 2017-02-28
avoid reloading modules that open up a hierarchical module tree

boegel:
Fix for bug introduced in #2091 by not purging loaded modules before loading modules that extend `$MODULEPATH` in `make_module_dep`.

We discussed this in #2091 and purging was not deemed necessary, but it is because `make_module_step` (and hence `make_module_dep`) is being called twice in a row, once for the actual module file, and again for the 'devel' module.

This bug was reported by @akesandgren manifested itself by the `binutils/2.27` not being found anymore when loading the `OpenMPI` module when creating the 'devel' module.

cc @bartoldeman


boegel:
(close/open to re-trigger Travis, can't seem to access the build logs...)

akesandgren:
@boegel want me to test that new commit?

akesandgren:
I tried the latest commit and it doesn't help.

boegel:
@akesandgren The previous commit introduced a regression, it not only unloaded all modules before loading the modules that open up the hierarchy, it was also resetting all other environment variables that were defined during the build & install procedure.
That's a problem, since the other parts of `make_module_step` may be relying on those environment variables to still be around...

Are you saying that avoiding to reload an already loaded module is not fixing the original problem you were seeing? If so, can you provide a new debug log that sits on top of the last commit?

akesandgren:
I added the changes from "files changed" at its current state (first reverting the changes you gave me yesterday), and with that in place it fails the same way as without any changes.

So you want a --debug-lmod --debug log?

boegel:
@akesandgren that would be great, yes

boegel:
@akesandgren problem should be fixed with aa6c4c2, although I am a bit puzzled about all this...

I have a suspicion that we're dancing around a bug in Lmod here, although the changes do make sense regardless (it'll speed things up a bit).

For the record, which Lmod version are you using, and what's your exact EasyBuild configuration?

akesandgren:
Lmod v6.5

'''
eb --show-config
#
# Current EasyBuild configuration
# (C: command line argument, D: default value, E: environment variable, F: configuration file)
#
buildpath               (F) = /scratch
filter-deps             (F) = pkg-config
git-working-dirs-path   (F) = /afs/hpc2n.umu.se/home/a/ake/Projects/git-repos/easybuild
github-user             (F) = akesandgren
include-easyblocks      (F) = /afs/hpc2n.umu.se/home/a/ake/Public/easybuild/easyblocks/*.py
installpath             (F) = /hpc2n/eb
job-backend-config      (F) = /hpc2n/eb/gc3pie.slurm.cfg
minimal-toolchains      (F) = True
module-naming-scheme    (F) = HierarchicalMNS
packagepath             (F) = /hpc2n/eb/packages
prefix                  (F) = /hpc2n/eb
recursive-module-unload (F) = True
repositorypath          (F) = /hpc2n/eb/ebfiles_repo
robot-paths             (F) = /afs/hpc2n.umu.se/home/a/ake/Public/easybuild/easyconfigs, /hpc2n/eb/ebfiles_repo, /hpc2n/eb/software/Core/EasyBuild/3.1.0-hpc2n.00/lib/python2.7/site-packages/easybuild_easyconfigs-3.1.0-py2.7.egg/easybuild/easyconfigs
sourcepath              (F) = /hpc2n/eb/sources:/pfs/software/sources
tmpdir                  (F) = /scratch
'''

boegel:
@geimer, @ocaisa, @damianam: thoughts on this?

ocaisa:
That's a pretty old version of Lmod. I wonder what behaviour you would see with the latest version?

geimer:
I'm still trying to make sense out of this (and #2091).  Thinking about it, I was wondering what these changesets do to packages that depend on `LLVM` (e.g., `Mesa`)?  `LLVM` is marked as `moduleclass = 'compiler'` and thus extends `$MODULEPATH`.  However, the depending package is not located in the path added by `LLVM`...

wpoely86:
LLVM should not be `compiler` but `lib`. Clang is `compiler`.

boegel:
@geimer The change in #2091 is needed when the MNS being used enables `expand_toolchain_load`, and you want to avoid having needless `load` statements included in the module files, e.g. for `icc` & `ifort` when `iccifort` is the toolchain.

In order to check which dependencies of the excluded dependencies also extend `$MODULEPATH` (and hence should be excluded too), we need to open up that part of the hierarchy where the modules for those deps-of-excl-deps, so we need to load the excluded dependencies first.

Apparently care has to be taken when doing that, since loading for example the `OpenMPI` module when it is already loaded can cause problems.
It has something to do with Lmod actually unloading and reloading the specified module. Unloading OpenMPI means unloading it's dependencies incl. the toolchain (when `--recursive-module-unload` is used), which rolls back `$MODULEPATH` to only including `<prefix>/Core/`.

For some reason in @akesandgren's setup, this leads to the `binutils/2.27` module that is housed in `Compiler/GCCcore/6.3.0` (which is a part of the `GCC/6.3.0-2.27` toolchain that was used to install `OpenMPI`) to become unavailable when the `OpenMPI` module is reloaded.

I'm not entirely sure why this happens, since it *should* work. This may be a bug in the Lmod version (6.5) that @akesandgren is using.

To dance around that, we can simply avoiding reloading modules if they are already loaded, in certain scenarios (basically when we're loading modules that extend $MODULEPATH).

This should be safe to do, and it also results in a little bit of speedup since reloading a module that extends `$MODULEPATH` triggers a (re)load storm and hence can be fairly costly.

boegel:
@geimer, @ocaisa: Thoughts on the above? Regardless of what the cause of @akesandgren's problem is, do the changes being made here make sense according to you?

I would really like to include this into EasyBuild v3.1.1 which I'd like to release ASAP...

ocaisa:
I can't think why there would be a problem with it but I'm no expert on this area of the code

boegel:
With some help from @rtmclay, I have found the underlying issue, see #2140.

I consider this PR to be obsolete now, but I'd like to get some feedback on #2140 first before I close this one.

@akesandgren If you test #2140, you should regenerate the modules in `modules/all/Compiler/GCC/6.3.0-2.27` since they should not include `load("GCCcore/6.3.0")` (nor `load("binutils/2.27")`)

akesandgren:
PR #2140  works like a charm :-)

boegel:
closing this in favor of #2140, thanks for the feedback @akesandgren!

boegel:
I did include the enhanced test that was added in 9d49d472e575081a39409ef066b107f385ad5cdb into #2140, since that is still useful.

PR #2140, opened 2017-03-05
fix logic in make_module_dep w.r.t. excluding loads for toolchain & toolchain components

boegel:
This provides a proper fix for the problem that #2135 dances around, and basically undoes @bartoldeman's PR #2091 since it implements a better solution that also does the right thing when `GCC` (i.e. the bundle of `GCCcore` and `binutils`) is used as a toolchain.

Thanks to @rtmclay for helping out with figuring out the actual problem by looking at the Lmod debug output provided by @akesandgren where reloading the `OpenMPI` module failed because the `binutils` module couldn't be found anymore.

The problem was that `load` statements were still being inserted for `GCCcore` and `binutils` when something was being built with the `GCC` toolchain (e.g. `OpenMPI`) when a module naming scheme that enables `expand_toolchain_load` was being used (for example `HierarchicalMNS`).

The existing code that excludes dependencies that extend `$MODULEPATH` failed because `GCC` (i.e. the toolchain being used) was expanded into `GCCcore` + `binutils` before the list of dependencies that extend `$MODULEPATH` are determined via `path_to_top_of_module_tree`, and hence `excluded_deps` was an empty list while it should have included `GCC`.

This is only a problem when the `GCC` toolchain was being used, since both `GCC` and its `GCCcore` component extend `$MODULEPATH`.

By now passing both the toolchain (e.g. `GCC`) and the toolchain components (`GCCcore` + `binutils`) to `path_to_top_of_module_tree`, the `GCC` toolchain itself is correctly excluded as well (and hence so are the toolchain components).

The additional loop that excludes dependencies of dependencies that was added in #2091 by @bartoldeman is no longer needed now.

cc @geimer, @ocaisa, @damianam

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -912,21 +912,23 @@ class EasyBlock(object):
 
         :param unload_info: dictionary with full module names as keys and module name to unload first as corr. value
         """
-        deps = []
         mns = ActiveMNS()
         unload_info = unload_info or {}
 
-        # include load statements for toolchain, either directly or for toolchain dependencies
+        # include toolchain as first dependency to load
+        tc_mod = None
         if self.toolchain.name != DUMMY_TOOLCHAIN_NAME:
-            if mns.expand_toolchain_load(ec=self.cfg):
-                mod_names = self.toolchain.toolchain_dep_mods
-                deps.extend(mod_names)
-                self.log.debug("Adding toolchain components as module dependencies: %s" % mod_names)
-            else:
-                deps.append(self.toolchain.det_short_module_name())
-                self.log.debug("Adding toolchain %s as a module dependency" % deps[-1])
+            tc_mod = self.toolchain.det_short_module_name()
+            self.log.debug("Toolchain to load in generated module (before excluding any deps): %s", tc_mod)
+
+        # expand toolchain into toolchain components if desired (and if the toolchain was retained as a dep)
+        tc_dep_mods = None
+        if mns.expand_toolchain_load(ec=self.cfg):
+            tc_dep_mods = self.toolchain.toolchain_dep_mods
+            self.log.debug("Toolchain components to load in generated module (before excluding any): %s", tc_dep_mods)
 
         # include load/unload statements for dependencies
+        deps = []
         self.log.debug("List of deps considered to load in generated module: %s", self.toolchain.dependencies)
         for dep in self.toolchain.dependencies:
             if dep['build_only']:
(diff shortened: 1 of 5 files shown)


akesandgren:
This works like a charm :-)

ocaisa:
lgtm

bartoldeman:
This certainly a more elegant solution! lgtm2

boegel:
Thanks for the feedback everyone, going in to be included in EasyBuild v3.1.1.

bartoldeman:
I did not check properly... This fix is good for the compiler-only toolchain case where the toolchain used equals the MODULEPATH extender (in fact my hook for ec in expand_toolchain_load(ec=self.cfg) was another way to fix this because our module naming scheme uses that to check for compiler-only toolchains, but your fix is much better than this).

However the original problem fixed in #2091 is still there.
I can reproduce in the same way as in #2091 with HPL-2.1-intel-2016b.eb. It now includes again these two superfluous loads:
```
load("GCCcore/5.4.0")
 
load("binutils/2.26")
```

bartoldeman:
How those that work: HPL using intel-2016b expands the dependencies from intel-2016b.eb:
```
dependencies = [
    ('GCCcore', gccver),
    ('binutils', binutilsver, '-GCCcore-%s' % gccver),
    ('icc', compver, gccsuff),
    ('ifort', compver, gccsuff),
    ('impi', '5.1.3.181', '', ('iccifort', '%s%s' % (compver, gccsuff))),
    ('imkl', '11.3.3.210', '', ('iimpi', version)),
]
```
The modules icc, ifort, and impi are eliminated since they expand $MODULEPATH.
However, GCCcore and binutils, which are dependencies of icc and ifort are no longer eliminated.

boegel:
@bartoldeman Hmm, ok, I indeed overlooked that... And since I also changed to test case, the tests didn't complain either.

Can you open a new PR to current `develop`, basically re-adding the changes you had in #2091?

Do also include test case from #2091 in the existing `test_make_module_dep_hmns` test, so we retain both test cases (i.e. the one from #2091 and the one I changed it to in this PR).

Do make sure that the original test from #2091 highlights the reintroduced bug; if not, we need a better test case.

boegel:
> I am a little worried about reintroducing #2091 because that also means adding #2135 again.

No, it doesn't, #2135 was a workaround for modules that include `load GCCcore` while they shouldn't, while #2091 *avoids* inserting the `load GCCcore` statements in the first place.

So, as long as your modules have been regenerating with the fix included in #2091, you don't need to avoid reloading modules anymore like we do in #2135 (other than maybe to optimise things, but you'll gain very little from it).

boegel:
fix was (re)implemented by @bartoldeman in #2152

PR #2141, opened 2017-03-05
Architecture agnostic unit tests (WIP)

geimer:
Various framework unit tests fail on POWER and ARM, mostly due to hard-coded comparisons with compiler flags that are only valid on x86_64.  This PR tries to address this by
 - first checking whether the correct compiler flags end up in `toolchain.options.option_map['optarch']` and
 - using `toolchain.options.option_map['optarch']` in the other tests.

There are also some other issues that I haven't looked into yet, in particular exceptions thrown due to ~~#1950~~.  Here is the full list of failures on AArch64:
 - [ ] `test.framework.toolchain.test_blas_lapack_family` (exception)
 - [ ] `test.framework.toolchain.test_ictce_toolchain` (exception)
 - [ ] `test.framework.toolchain.test_mpi_cmd_for` (exception)
 - [ ] `test.framework.toolchain.test_mpi_family` (exception)
 - [ ] `test.framework.toolchain.test_old_new_iccifort` (exception)
 - [ ] `test.framework.toolchain.test_pgi_imkl` (exception)
 - [ ] `test.framework.toolchain.test_pgi_toolchain` (exception)
 - [ ] `test.framework.toolchain.test_compiler_dependent_optarch` (hard-coded value)
 - [x] `test.framework.toolchain.test_goolfc` (hard-coded value)
 - [x] `test.framework.toolchain.test_misc_flags_unique` (hard-coded value)
 - [ ] `test.framework.toolchain.test_optarch_generic` (hard-coded value)
 - [x] `test.framework.toolchain.test_override_optarch` (hard-coded value)
 - [x] `test.framework.toolchain.test_precision_flags` (hard-coded value)
 - [ ] `test.framework.options.test_dump_env_config` (hard-coded value)

### DIFF
--- a/easybuild/toolchains/compiler/ibmxl.py
+++ b/easybuild/toolchains/compiler/ibmxl.py
@@ -41,8 +41,8 @@ class IBMXL(Compiler):
     }
 
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
-        (systemtools.POWER, systemtools.POWER): ['qtune=auto', 'qmaxmem=-1'],
-        (systemtools.POWER, systemtools.POWER_LE): ['qtune=auto', 'qmaxmem=-1'],
+        (systemtools.POWER, systemtools.POWER): 'qtune=auto -qmaxmem=-1',
+        (systemtools.POWER, systemtools.POWER_LE): 'qtune=auto -qmaxmem=-1',
     }
 
     COMPILER_CC = 'xlc'
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -319,8 +319,8 @@ class ToolchainTest(EnhancedTestCase):
                 if optarch_var is not None:
                     flag = '-%s' % optarch_var
                 else:
-                    # default optarch flag
-                    flag = tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(tc.arch, tc.cpu_family)]
+                    # default optarch flag, contents of tc.options.options_map['optarch'] checked in test_optarch_flags
+                    flag = tc.options.options_map['optarch']
 
                 for var in flag_vars:
                     flags = tc.get_variable(var)
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -632,8 +632,10 @@ class ToolchainTest(EnhancedTestCase):
         tc = self.get_toolchain("goalf", version="1.1.0-no-OFED")
         tc.set_options({})
         tc.prepare()
+        # Contents of tc.options.options_map['optarch'] checked in test_optarch_flags
+        optarch_flags = tc.options.options_map['optarch']
         for var in flag_vars:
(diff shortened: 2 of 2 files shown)


boegel:
@geimer I ran into 5 failing tests on my Raspberry Pi Zero W, see https://gist.github.com/boegel/6ea2ec90a78bf3ac1549266b09dd2517, one of which (`test_check_style )` only because `pep8` was not installed, I'll add a *skip* for that.

The others are: `test_dump_env_config`, `test_compiler_dependent_optarch`, `test_optarch_generic`, `test_precision_flags`.

geimer:
@boegel Wow, a bit more than 2,5 hours.  That's impressively slow :wink:

Do these tests fail with or without this PR?  If it's with the PR, then I'm puzzled why `test_precision_flags` is still failing.

geimer:
@boegel OK, just saw that it was with the 3.1.0 release.  But then it's weird that for example `test_goolfc` didn't fail.  But this may be due to the platform not being detected correctly?

boegel:
@geimer This was without this PR, indeed.

What do you mean exactly with "platform not being detected correctly"?

All the `systemtools` tests passed?

geimer:
@boegel Well, someone :wink: put this line in the `systemtools` unit tests: https://github.com/hpcugent/easybuild-framework/blob/master/test/framework/systemtools.py#L571 So the family could be `UNKNOWN`.  Although that shouldn't happen on ARM.  Also, the `_native` versions of the various unit tests only verify that the CPU architecture/family/whatever is one of the known ones, but not that it is the correct one. That is, they wouldn't fail if your ARM CPU would be detected as Intel for whatever reason.

In any case, `test_goolfc` should have failed too, as it includes a hard-coded check for `-march=native` which you should never ever get on ARM.  So it would be good if you could check this again.

JackPerdue:
@geimer anything I might have of use in terms of XL is at http://www.siliconslick.com/easybuild/easyconfigs/curie/
Since we terminated the lease on the BG/Q there hasn't been much interest in Power so I've basically dropped it.   It's much more fun to discus the intricacies of GCCcore naming with Kenneth. :)   Perhaps once we can agree on that I can play with IBM XL again but we aren't renewing the license and don't have future plans for it.

geimer:
@JackPerdue Thanks. This looks pretty much like the `iccifort` approach. I think I can make up some dummy modulefiles for this for the unit tests. They don't have to be from a real install (though it would have been nice).

akesandgren:
@geimer Conflict needs fixing. And what is the status of this?

geimer:
@akesandgren  I haven't looked into it for a while.  IIRC, I got stuck with some test where I couldn't really figure out how it was working due to some really dense Python-foo I didn't understand, and then got distracted by other work.  I would really like to see the tests being platform-agnostic, but cannot make and promises when I will find the time to look into this again.  (This also depends on the state of my private AArch64 toy, which is a bit messed up right now.)

Issue #2148, opened 2017-03-08
Test failure: KeyError: 'PYTHONPATH'


opoplawski:
Running tests with "/usr/bin/python2 setup.py test" I get lots of these errors:
```
Traceback (most recent call last):
  File "/export/home/orion/redhat/python-easybuild-framework/easybuild-framework-easybuild-framework-v3.1.1/test/framework/utilities.py", line 185, in setUp
    self.env_pythonpath = os.environ['PYTHONPATH']
  File "/usr/lib64/python2.7/UserDict.py", line 23, in __getitem__
    raise KeyError(key)
KeyError: 'PYTHONPATH'
```
Looks like utilities.py needs to handle the case where PYTHONPATH is not set.

boegel:
Should be fixed by the changes made in #2149, thanks for reporting this @opoplawski!

PR #2149, opened 2017-03-09
take into account that $PATH or $PYTHONPATH may be empty when running tests

boegel:
fix for #2148


boegel:

Issue #2150, opened 2017-03-09
multiple paths in include-easyblocks doesn't always work


akesandgren:
Having
```
include-easyblocks=/home/a/ake/Public/easybuild/easyblocks/*.py
,/hpc2n/eb/custom/easyblocks/*.py
```
in my config file fails if the same easyblock is present in both directories.
```
eb --show-config
ERROR: Failed to parse configuration options: 'Symlinking /hpc2n/eb/custom/easyblocks/siesta.py to /scratch/eb-yAoV90/included-easyblocks/easybuild/easyblocks/siesta.py failed: [Errno 17] File exists'
```

This makes it harder to keep on developing an easyblock that haven't been accepted upstreams yet but is good enough for a site to move forward. The framework should handle this in my opinion. First seen should take precedence.

boegel:
fixed in #2151, let's follow up there

PR #2151, opened 2017-03-09
handle duplicates in --include-*

boegel:
fix for #2150

I verified that the enhanced tests indeed replicate the bug reported in #2150.

cc @akesandgren

### DIFF
--- a/easybuild/tools/include.py
+++ b/easybuild/tools/include.py
@@ -162,7 +162,8 @@ def include_easyblocks(tmpdir, paths):
         else:
             target_path = os.path.join(easyblocks_dir, filename)
 
-        symlink(easyblock_module, target_path)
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86 please review?

easybuild/tools/include.py
```diff
@@ -162,7 +162,8 @@ def include_easyblocks(tmpdir, paths):
         else:
             target_path = os.path.join(easyblocks_dir, filename)
 
-        symlink(easyblock_module, target_path)
+        if not os.path.exists(target_path):
```
wpoely86: what if it exists and points to something else? Can that happen?

boegel: no, this is in a dedicated tmpdir

boegel:

PR #2153, opened 2017-03-09
FIX run_cmd bug with shell=False

rjeschmi:
executable needs to be set to None if shell=False or interpretor won't be run (and other weirdness will happen)


boegel:
@rjeschmi Can we enhance the test for `run_cmd` to avoid reintroducing this bug?

boegel:

boegel:
Going in, thanks @rjeschmi!

PR #2155, opened 2017-03-09
implement change_dir function in filetools module

boegel:
`change_dir` is currently not used in `easybuild/tools/run.py`, since that would introduce a cyclic dependency between `run.py` and `filetools.py`...


boegel:
@wpoely86 please review?

boegel:

PR #2156, opened 2017-03-10
Added ld.bfd to wrappers for rpath

jdonners:
e.g. the OpenFOAM build uses ld.bfd


boegel:
lgtm, thanks @jdonners!

PR #2157, opened 2017-03-10
Introduce function ensure_iterable_license_specs

geimer:
...as a result of refactoring https://github.com/hpcugent/easybuild-easyblocks/pull/1129

### DIFF
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -433,6 +433,26 @@ def to_checksums(checksums):
 
     return res
 
+def ensure_iterable_license_specs(specs):
+    """
+    Ensures that the provided license file/server specifications are of correct type and converts
+    them to either a list or a tuple.  The input can either be None, a string, or a list/tuple of
+    strings.
+
+    :param specs: License file/server specifications as provided via `license_file` easyconfig parameter
+    """
+    if specs is None:
+        license_specs = [None]
+    elif isinstance(specs, basestring):
+        license_specs = [specs]
+    elif isinstance(specs, (list, tuple)) and all(isinstance(x, basestring) for x in specs):
+        license_specs = specs
+    else:
+        msg = "Unsupported type %s for easyconfig parameter 'license_file'! " % type(specs)
+        msg += "Can either be None, a string, or a tuple/list of strings."
+        raise EasyBuildError(msg)
+
+    return license_specs
 
 # these constants use functions defined in this module, so they needs to be at the bottom of the module
 # specific type: dict with only name/version as keys with string values, and optionally a hidden key with bool value
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -436,8 +436,7 @@ def to_checksums(checksums):
 def ensure_iterable_license_specs(specs):
     """
     Ensures that the provided license file/server specifications are of correct type and converts
-    them to either a list or a tuple.  The input can either be None, a string, or a list/tuple of
-    strings.
+    them to a list.  The input can either be None, a string, or a list/tuple of strings.
 
     :param specs: License file/server specifications as provided via `license_file` easyconfig parameter
     """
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/types.py
```diff
@@ -433,6 +433,26 @@ def to_checksums(checksums):
...
+    strings.
+
+    :param specs: License file/server specifications as provided via `license_file` easyconfig parameter
+    """
+    if specs is None:
+        license_specs = [None]
+    elif isinstance(specs, basestring):
+        license_specs = [specs]
+    elif isinstance(specs, (list, tuple)) and all(isinstance(x, basestring) for x in specs):
+        license_specs = specs
```
boegel: I would ensure a consistent return type here, either a list of a tuple...

To ensure a list, use `list(specs)` here.

To ensure a tuple, use `tuple(specs)` here and `(None,)` and `(specs,)` above.

boegel:
@geimer You forgot the commit for the unit test for the new function. ;-)

geimer:
@boegel I only have two hands and a single keyboard, so my multi-threading capabilities are limited ;-)
Will follow-up after lunch...

test/framework/type_checking.py
```diff
@@ -565,6 +565,27 @@ def test_to_checksums(self):
...
             self.assertEqual(to_checksums(checksums), checksums)
 
+    def test_ensure_iterable_license_specs(self):
+        """Test ensure_iterable_license_specs function."""
+        # Test acceptable inputs
+        self.assertEqual(ensure_iterable_license_specs(None), [None])
+        self.assertEqual(ensure_iterable_license_specs('foo'), ['foo'])
+        self.assertEqual(ensure_iterable_license_specs(['foo']), ['foo'])
+        self.assertEqual(ensure_iterable_license_specs(['foo', 'bar']), ['foo', 'bar'])
+        self.assertEqual(ensure_iterable_license_specs(('foo')), ['foo'])
```
boegel: if you want to test with a single-element tuple, use `('foo',)` rather than `('foo')` (which is identical to `'foo'`

test/framework/type_checking.py
```diff
@@ -565,6 +565,27 @@ def test_to_checksums(self):
...
+        # Test unacceptable inputs
+        error_msg = "Unsupported type .* for easyconfig parameter 'license_file'!"
+        self.assertErrorRegex(EasyBuildError, error_msg, ensure_iterable_license_specs, 42)
+        self.assertErrorRegex(EasyBuildError, error_msg, ensure_iterable_license_specs, {'1': 'foo'})
+        self.assertErrorRegex(EasyBuildError, error_msg, ensure_iterable_license_specs, [None])
+        self.assertErrorRegex(EasyBuildError, error_msg, ensure_iterable_license_specs, [42])
+        self.assertErrorRegex(EasyBuildError, error_msg, ensure_iterable_license_specs, [42, 'foo'])
+        self.assertErrorRegex(EasyBuildError, error_msg, ensure_iterable_license_specs, [['foo']])
+        self.assertErrorRegex(EasyBuildError, error_msg, ensure_iterable_license_specs, [(42, 'foo')])
+        self.assertErrorRegex(EasyBuildError, error_msg, ensure_iterable_license_specs, (42))
```
boegel: `(42,)`

geimer:
@boegel Travis doesn't seem to pick up the last commit.  What was the trick again?

boegel:
@geimer close/open the PR, but other (more recent) PRs (cfr. #2159) are also having the same issue, seems like Travis (or GitHub) are experiencing some problems, so we'll have to give it some time, I think...

boegel:

boegel:
@geimer This is good to go, but I'll hold off merging it into `develop` until EasyBuild v3.1.2 is release (which is a bugfix only release), cfr. https://github.com/hpcugent/easybuild-framework/milestone/45

boegel:
Going in, thanks @geimer!

Issue #2154, opened 2017-03-09
test_vsc_location (test.framework.general.GeneralTest) fails


opoplawski:
I'm working on building Fedora rpms for easybuild packages.  I'm getting the following test error:
```
ERROR [1.860s]: test_vsc_location (test.framework.general.GeneralTest)
Make sure location of imported vsc module is not the framework itself.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/export/home/orion/redhat/python-easybuild-framework/easybuild-framework-easybuild-framework-v3.1.1/test/framework/general.py", line 53, in test_vsc_location
    vsc_loc = os.path.dirname(os.path.dirname(os.path.abspath(vsc.__file__)))
AttributeError: 'module' object has no attribute '__file__'
```
The vsc-base package does not install a __init__.py file in site-packages/vsc, I believe because it is a namespace package, so it doesn't not have a __file__ attribute.

boegel:
@opoplawski Hmm, this is new to me... I can probably come up with a patch that dances around this, for example by checking `vsc.utils.__file__` or `vsc.utils.fancylogger.__file__` instead, but I'd like to understand what is causing this exactly.

You may be right that this may be due to `vsc` being registered as a namespace package, but it may also indicate a mistake in the packaging of `vsc-base` and/or `vsc-install`.

Is that the only test failing in the whole suite?

Can we inspect the `vsc-base` and `vsc-install` packages that you are using as dependencies somewhere?

BTW, you should try and run the tests in Python 'optimised' mode, since the main `eb` command also uses `python -O`; one way is to define `$PYTHONOPTIMIZE` as `1`, see https://docs.python.org/2/using/cmdline.html#cmdoption-O .

opoplawski:
One difference may be how you are installing the vsc packages when running the tests.  When you build the vsc packages locally, you get an __init__,py.  e.g.:
```
$ ls build/lib/vsc/
__init__.py  utils
$ PYTHONPATH=build/lib python -c 'import vsc;print vsc.__file__'
/export/home/orion/redhat/python-vsc-base/vsc-base-vsc-base-2.5.7/build/lib/vsc/__init__.pyc
```
But that doesn't get installed into site-packages.  Instead a .pth file is installed:
```
/usr/lib/python2.7/site-packages/vsc_base-2.5.7-py2.7-nspkg.pth
# ls /usr/lib/python2.7/site-packages/vsc
[...]

PR #2159, opened 2017-03-10
fix test_vsc_location, vsc.__file__ may not be available when vsc is installed as a namespace package

boegel:
fix for #2154


boegel:

PR #2160, opened 2017-03-10
use checker_state in trailing whitespace style check + add dedicated test

boegel:
@wpoely86 While working on additional style check for the order of easyconfig parameters, I noticed that we're not using the `checker_state` that `pep8` supports, which comes in useful in the  check for trailing whitespace where we want to ignore trailing whitespace in the `description`...

### DIFF
--- a/easybuild/framework/easyconfig/style.py
+++ b/easybuild/framework/easyconfig/style.py
@@ -43,9 +43,16 @@ _log = fancylogger.getLogger('easyconfig.style', fname=False)
 
 EB_CHECK = '_eb_check_'
 
+COMMENT_REGEX = re.compile(r'^\s*#')
+PARAM_DEF_REGEX = re.compile(r"^(?P<key>[a-z_]+)\s*=\s*")
 
-# any function starting with _eb_check_ (see EB_CHECK variable) will be
+
+# Any function starting with _eb_check_ (see EB_CHECK variable) will be
 # added to the tests if the test number is added to the select list.
+#
+# Note: only functions that have a first argument named 'physical_line' or 'logical_line'
+# will actually be used!
+#
 # The test number is definied as WXXX and EXXX (for warnings and errors)
 # where XXX is a 3 digit number.
 #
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/style.py
```diff
@@ -62,27 +70,25 @@ def _eb_check_trailing_whitespace(physical_line, lines, line_number, total_lines
...
-    result = pep8.trailing_whitespace(physical_line)
+    result = pep8.trailing_whitespace(line)
     if result:
-        result = (result[0], result[1].replace("W291", "W299"))
+        result = (result[0], result[1].replace('W291', 'W299'))
+
+    # keep track of name of last parameter that was defined
+    param_def = PARAM_DEF_REGEX.search(line)
+    if param_def:
+        checker_state['eb_last_key'] = param_def.group('key')
```
wpoely86: this is nice!

easybuild/framework/easyconfig/style.py
```diff
@@ -62,27 +70,25 @@ def _eb_check_trailing_whitespace(physical_line, lines, line_number, total_lines
     The arguments are explained at
     https://pep8.readthedocs.org/en/latest/developer.html#contribute
     """
-    comment_re = re.compile(r'^\s*#')
-    if comment_re.match(physical_line):
+    # apparently this is not the same as physical_line line?!
+    line = lines[line_number-1]
```
wpoely86: why switch?

boegel: Because apparently this is not the same as `physical_line`... I noticed that for the `description =` line, only the part *after* the `=` was included in `physical_line`...

test/framework/style.py
```diff
@@ -60,6 +60,33 @@ def test_style_conformance(self):
...
+        ]
+        line_numbers = range(1, len(lines) + 1)
+        test_cases = [
+            ({}, None),
+            ({}, (17, "W299 trailing whitespace")),
+            ({}, (0, "W293 blank line contains whitespace")),
+            ({}, None),
+            ({'eb_last_key': 'description'}, None),
+            ({'eb_last_key': 'description'}, None),
+            ({'eb_last_key': 'description'}, (21, "W299 trailing whitespace")),
```
boegel: I should change this, I should be specifying the state manually, but just use `state = {}` before the loop and then pass in `state` without touching it in the test.

boegel:

PR #2161, opened 2017-03-11
consider both pycodestyle and pep8 Python modules in style checks

boegel:
`pep8` was renamed to `pycodestyle` a while ago, so we should give preference to `pycodestyle`...


boegel:
@wpoely86 please review?

boegel:

PR #2165, opened 2017-03-14
enhance easyconfigs style check to also check grouping & order of easyconfig parameters (WIP)

boegel:
not finished yet, but interested in getting some early feedback on this (@wpoely86?)

### DIFF
--- a/easybuild/framework/easyconfig/style.py
+++ b/easybuild/framework/easyconfig/style.py
@@ -33,7 +33,7 @@ import sys
 from vsc.utils import fancylogger
 
 from easybuild.framework.easyconfig.default import DEFAULT_CONFIG
-from easybuild.tools.build_log import print_msg
+from easybuild.tools.build_log import EasyBuildError, print_msg
 from easybuild.tools.utilities import only_if_module_is_available
 
 try:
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/style.py
```diff
@@ -68,6 +79,104 @@
...
+def _check_param_group(params, done_params):
+    """
+    Check whether the specified group of parameters conforms to the style guide w.r.t. order & grouping of parameters
+    """
+    result, fail_msgs = None, []
+
+    fail_msgs.extend(_check_param_group_head(params, done_params))
+
+    if fail_msgs:
+        result = (0, "W001 %s" % ', '.join(fail_msgs))
```
wpoely86: `W001` is still free?

boegel: Yes, see https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes, but I'm open to suggestions for a better choice...

wpoely86:
Seems fine but more complex then what I had in mind.
I was just going to collect all keys, check if the mandatory ones are there, check if they all are known, check the order by weighting them by line number.

akesandgren:
@boegel Care to fix conflict?

Issue #2163, opened 2017-03-13
--try-software-version leads to syntax error in Infernal EasyConfig


SethosII:
EasyBuild generates a wrong EasyConfig when I try to use `--try-software-version` on the [Infernal easyconfig](https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/i/Infernal/Infernal-1.1-goolf-1.4.10.eb). See this example:

```
$ eb --buildpath=/dev/shm --optarch=march=corei7 --prefix=/mnt/easybuild/test --try-software-version=1.0 Infernal-1.1-goolf-1.4.10.eb -dr # same problem on other values for software version (e. g. `1.1.2`) as well
== temporary log file in case of crash /tmp/eb-lmDJNO/easybuild-LW76nC.log
ERROR: SyntaxError in easyconfig pyheader easyblock = 'ConfigureMake'

name = 'Infernal'
version = "1.0"

homepage = 'http://eddylab.org/infernal/'
description = """Infernal ("INFERence of RNA ALignment") is for searching DNA sequence databases
 for RNA structure and sequence similarities."""

toolchain = {'name': 'goolf', 'version': '1.4.10'}
toolchainopts = {'optarch': True, 'pic': True}

sources = [SOURCELOWER_TAR_GZ]
source_urls = ['http://eddylab.org/%(namelower)s']

checksums = []
}

moduleclass = 'bio'
: invalid syntax (<string>, line 17)
```

The original EasyConfig contains this:

```
##
# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
#
# Copyright:: Copyright 2012-2014 Uni.Lu/LCSB, NTUA
# Authors::   Cedric Laczny <cedric.laczny@uni.lu>, Fotis Georgatos <fotis@cern.ch>
# License::   MIT/GPL
# $Id$
#
# This work implements a part of the HPCBIOS project and is a component of the policy:
# http://hpcbios.readthedocs.org/en/latest/HPCBIOS_2012-94.html
##

easyblock = 'ConfigureMake'

name = 'Infernal'
version = "1.1"

homepage = 'http://eddylab.org/infernal/'
description = """Infernal ("INFERence of RNA ALignment") is for searching DNA sequence databases
 for RNA structure and sequence similarities."""

toolchain = {'name': 'foss', 'version': '2016b'}
toolchainopts = {'optarch': True, 'pic': True}

[...]

PR #2166, opened 2017-03-14
fix --try-software-version using non-greedy matching + lookahead assertion, add test for tweak_one

boegel:
fix for #2163; https://github.com/hpcugent/easybuild-framework/issues/2163#issuecomment-286532957 is probably useful to get a good idea of why this fix is needed

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -179,7 +179,9 @@ def tweak_one(src_fn, target_fn, tweaks, targetdir=None):
     for (key, val) in tweaks.items():
 
         if isinstance(val, list):
-            regexp = re.compile(r"^(?P<key>\s*%s)\s*=\s*(?P<val>\[(.|\n)*\])\s*$" % key, re.M)
+            # use non-greedy matching for list value using '*?' to avoid including other parameters in match,
+            # and a lookahead assertion (?=...) so next line is either another parameter definition or a blank line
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/tweak.py
```diff
@@ -179,7 +179,9 @@ def __repr__(self):
     for (key, val) in tweaks.items():
 
         if isinstance(val, list):
-            regexp = re.compile(r"^(?P<key>\s*%s)\s*=\s*(?P<val>\[(.|\n)*\])\s*$" % key, re.M)
+            # use non-greedy matching for list value using '*?' to avoid including other parameters in match,
+            # and a lookahead assertion (?=...) so next line is either another parameter definition or a blank line
+            regexp = re.compile(r"^(?P<key>\s*%s)\s*=\s*(?P<val>\[(.|\n)*?\])\s*$(?=\n^(\w+\s*=.*|\s*)$)" % key, re.M)
```
wpoely86: why the last `|`?

wpoely86: Doesn't `\s` also match `\n` in multiline mode?

boegel: The last `|` in the lookahead assertion is there to check that the next line is either a parameter definition or a blank line (or the end of the file, see last commit).

Where is it relevant that `\s` also matches `\n`?

boegel:

PR #2168, opened 2017-03-16
Avoid creating empty modulefile directories when using `modaltsoftname`

geimer:
Use module name from naming scheme rather than deriving path from install directory. Fixes https://github.com/hpcugent/easybuild-framework/issues/2167 for my use cases.

@boegel I'm not sure about the module symlinking at https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/framework/easyblock.py#L1527, but my gut feeling is that it was broken as well and will be implicitly fixed, too.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1519,7 +1519,7 @@ class EasyBlock(object):
         # create parent dirs in install and modules path already
         # this is required when building in parallel
         mod_symlink_paths = ActiveMNS().det_module_symlink_paths(self.cfg)
-        parent_subdir = os.path.dirname(self.install_subdir)
+        parent_subdir = os.path.dirname(ActiveMNS().det_full_module_name(self.cfg))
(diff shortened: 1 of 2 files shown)


boegel:
lgtm, but we've learned the hard way not to fix bugs without making sure we have a test that triggers it (to avoid reintroducing the same problem later)

done in https://github.com/geimer/easybuild-framework/pull/16

geimer:
@boegel I had to fix the unit test you provided in https://github.com/geimer/easybuild-framework/pull/16 as it was assuming Lua module syntax (hard-coded).  Please have quick look at the fix and let me know whether it's OK or whether you have any suggestions for improvement.

boegel:
@geimer Oops, I indeed overlooked that even though I made a mental note to fix that before passing it to you, sorry. Fix looks great, thanks!

boegel:
Going in, thanks for the bugfix @geimer!

Issue #2162, opened 2017-03-13
Incorrect module loads in modules built by users with HMNS, minimal-toolchain, recursive-unload


akesandgren:
We have the following system setup:
```
eb --show-config
#
# Current EasyBuild configuration
# (C: command line argument, D: default value, E: environment variable, F: configuration file)
#
buildpath               (F) = /scratch
filter-deps             (F) = pkg-config
git-working-dirs-path   (F) = /afs/hpc2n.umu.se/home/a/ake/Projects/git-repos/easybuild
github-user             (F) = akesandgren
include-easyblocks      (F) = /hpc2n/eb/custom/easyblocks/*.py
installpath             (F) = /hpc2n/eb
job-backend-config      (F) = /hpc2n/eb/gc3pie.slurm.cfg
minimal-toolchains      (F) = True
module-naming-scheme    (F) = HierarchicalMNS
packagepath             (F) = /hpc2n/eb/packages
prefix                  (F) = /hpc2n/eb
recursive-module-unload (F) = True
repositorypath          (F) = /hpc2n/eb/ebfiles_repo
robot-paths             (F) = /afs/hpc2n.umu.se/home/a/ake/Public/easybuild/easyconfigs, /hpc2n/eb/custom/easyconfigs, /hpc2n/eb/ebfiles_repo, /hpc2n/eb/software/Core/EasyBuild/3.1.1/lib/python2.7/site-packages/easybuild_easyconfigs-3.1.1-py2.7.egg/easybuild/easyconfigs
sourcepath              (F) = /hpc2n/eb/sources:/pfs/software/sources
subdir-user-modules     (F) = Public/easybuild/modules
tmpdir                  (F) = /scratch
```

Things work as expected and modules get correctly created.
But when adding the SoftInst module from @geimer on top of this to allow for users to build their own softwares we get incorrectly built modules.

```
ml SoftInst
eb --show-config
#
# Current EasyBuild configuration
# (C: command line argument, D: default value, E: environment variable, F: configuration file)
#
buildpath               (E) = /scratch/eb-ake
experimental            (E) = True
filter-deps             (F) = pkg-config
git-working-dirs-path   (F) = /afs/hpc2n.umu.se/home/a/ake/Projects/git-repos/easybuild
github-user             (F) = akesandgren
include-easyblocks      (F) = /hpc2n/eb/custom/easyblocks/*.py
installpath             (E) = /home/a/ake/Public/easybuild
[...]

PR #2172, opened 2017-03-17
add test to verify that modules installed in user HMNS module tree don't have superfluous load statements (cfr. #2162)

boegel:
This test reproduces the bug reported in #2162 by @akesandgren and confirmed by @geimer.

With current `develop`, the test will fail (confirming that it reproduces #2162), but once #2104 is merged, the test will pass, which confirms that #2104 indeed fixes the problem as @bartoldeman suggested in #2162.

I will sync my branch used for this PR with `develop` once #2104 is merged.


boegel:
Since this is 'just' a test that is verified to cover the bug reported in #2162, I'm going to merge this in now Travis has confirmed that the changes in #2014 fixes the bug.

PR #2178, opened 2017-03-22
clarify error message when no software-specific easyblock was found

boegel:
This changes a very cryptic error message like:

```
NO LONGER SUPPORTED since v2.0: Fallback to default easyblock ConfigureMake
(from easybuild.easyblocks.generic.configuremake); use "easyblock = \'ConfigureMake\'" in
easyconfig file?;
see http://easybuild.readthedocs.org/en/latest/Deprecated-functionality.html for more information
```

to:

```
No software-specific easyblock 'EB_MXNet' found for MXNet
```

I made a breaking change to the signature of `get_easyblock_class` for this, basically replacing the `default_fallback` optional argument with `error_on_missing_easyblock`.

Should I include a deprecation warning for that, which detects the use of `default_fallback` and assigns `error_on_missing_easyblock` instead? Just in case this breaks a script that talks with the framework API?

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1727,7 +1727,8 @@ class EasyBlock(object):
             try:
                 # no error when importing class fails, in case we run into an existing easyblock
                 # with a similar name (e.g., Perl Extension 'GO' vs 'Go' for which 'EB_Go' is available)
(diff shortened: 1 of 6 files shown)


wpoely86:
We should probably not break api during minor releases?

boegel:
@wpoely86 updated, please rereview?

boegel:

PR #2179, opened 2017-03-22
make log.deprecated more verbose by also printing deprecation warnings to stderr

boegel:
`log.deprecated` now only results in logging warnings in the log file, where they (too) easily get ignored.

This changes that by also printing deprecation messages to `stderr`, where they should pop up more easily.

Unless you are using deprecated functionality, you should not be seeing any warnings like this.

(note: requires https://github.com/hpcugent/vsc-base/pull/250)


boegel:

PR #2180, opened 2017-03-23
add 'copy' function to filetools for easy copying of lists of files/directories

boegel:


easybuild/tools/filetools.py
```diff
@@ -1455,3 +1457,29 @@ def copy_dir(path, target_path, force_in_dry_run=False):
...
+    :param force_in_dry_run: force running the command during dry run
+    """
+    if isinstance(paths, basestring):
+        paths = [paths]
+
+    _log.info("Copying %d files & directories to %s", len(paths), target_path)
+
+    for path in paths:
+
+        full_target_path = os.path.join(target_path, os.path.basename(path))
```
wpoely86: `target_path` (in the last part)?

easybuild/tools/filetools.py
```diff
@@ -1455,3 +1457,29 @@ def copy_dir(path, target_path, force_in_dry_run=False):
...
+    :param target_path: target location
+    :param force_in_dry_run: force running the command during dry run
+    """
+    if isinstance(paths, basestring):
+        paths = [paths]
+
+    _log.info("Copying %d files & directories to %s", len(paths), target_path)
+
+    for path in paths:
+
```
wpoely86: less whitespace?

easybuild/tools/filetools.py
boegel: No, `target_path` is the directory we're copying to, `full_target_path` is the path that also includes the name of the file/directory we're copying to...

boegel:

PR #2181, opened 2017-03-23
Convert yeb to eb

ocaisa:
Add a function to dump a parsed yeb to an equivalent .eb easyconfig (without any comments)


ocaisa:
Should add a check to make sure it is a yeb easyconfig instance before allowing the conversion

ocaisa:
@boegel Ok, that's the first step down from my side for #2122

ocaisa:
@boegel I still believe yeb files have a future, kicking the can on this :)

ocaisa:
`.yeb` support is now deprecated so this no longer makes sense

Issue #2182, opened 2017-03-27
optarch for different compilers breaks --job


mstud:
I was trying to use the GC3Pie job backend but keep getting this error message in the job log file:

ERROR: Failed to parse configuration options: "The optarch option has an incorrect syntax: '{\\'GCC\\': \\'march=westmere -mtune=haswell\\', \\'Intel\\': \\'xSSE4.2 -axCORE-AVX2\\'}'"

My optarch line in the EasyBuild config file:
optarch=Intel:xSSE4.2 -axCORE-AVX2;GCC:march=westmere -mtune=haswell

It works fine without --job. Looks like the rather new feature that lets us set optarch per compiler somehow breaks when being used in conjunction with the --job feature. Or am I doing something wrong?

boegel:
@damianam Any ideas on this one?

damianam:
I haven't tested the option using a config file, but in a terminal I'd put the options between single quotes. However, the error seems to spit a valid dictionary.......

The troubled lines are these: https://github.com/hpcugent/easybuild-framework/blob/261861679742a8cc69cca232bb878bd698cd93f7/easybuild/tools/options.py#L745 but I don't see how that can happen...... I'll try to take a look at it tomorrow, sadly I can't do it now.

Sneeg:
The problem seems to be that _postprocess_optarch is called twice when using --job, once immediately when running eb and then later in the job again. The second time, self.options.optarch has already been overwritten by the string representation of the dictionary with the parsed data. Then,  of course, there is no OPTARCH_SEP in the string, but two OPTARCH_MAP_CHARs, so the exception is raised.

damianam:
I'd say you are right, but there is more to it. `self.options.optarch` is a string that looks like an already processed `dict`, and that's why it doesn't fail during `self.options.optarch.split(OPTARCH_SEP)`

I think the issue is more complex than it looks like because somewhere, when using `--job`, the `dict` gets serialized to a string, and then things go down. We could apply a quick and dirty fix here to check if the `optarch` string looks like a `dict`, but in my opinion we should treat the "illness", not the "symptom"

damianam:
> but in my opinion we should treat the "illness", not the "symptom"

On the other hand, this can be seen as a side effect of doing something nasty, like shoving a dict into what it used to be a string. I remember saying something like "I am not convinced that's a great idea...." ;-P. Maybe it is better to simply check if `--job` is part of the options, and avoid the processing altogether when that happens?

[...]

PR #2183, opened 2017-03-28
fix use of compiler-specific --optarch value in combination with --job

damianam:
Fixes https://github.com/hpcugent/easybuild-framework/issues/2182

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -107,8 +107,11 @@ def eb_shell_quote(token):
     Wrap provided token in single quotes (to escape space and characters with special meaning in a shell),
     so it can be used in a shell command. This results in token that is not expanded/interpolated by the shell.
     """
+    # first, strip off double quotes that may wrap the entire value,
+    # we don't want to wrap single quotes around a double-quoted value
+    token = str(token).strip('"')
     # escape any non-escaped single quotes, and wrap entire token in single quotes
-    return "'%s'" % re.sub(r"(?<!\\)'", r"\'", str(token))
+    return "'%s'" % re.sub(r"(?<!\\)'", r"\'", token)
 
 vsc.utils.generaloption.shell_quote = eb_shell_quote
 
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -727,8 +727,9 @@ class EasyBuildOptions(GeneralOption):
         if self.options.last_log:
             self.options.terse = True
 
-        # make sure --optarch has a valid format
-        if self.options.optarch:
+        # make sure --optarch has a valid format, but do it only if we are not going to submit jobs. Otherwise it gets
+        # processed twice and fails when trying to parse a dictionary as if it was a string
+        if self.options.optarch and not self.options.job:
             self._postprocess_optarch()
 
         # handle configuration options that affect other configuration options
--- a/easybuild/tools/parallelbuild.py
+++ b/easybuild/tools/parallelbuild.py
@@ -35,7 +35,6 @@ Support for PBS is provided via the PbsJob class. If you want you could create o
 import math
 import os
 import re
-import subprocess
 
 from easybuild.framework.easyblock import get_easyblock_instance
 from easybuild.framework.easyconfig.easyconfig import ActiveMNS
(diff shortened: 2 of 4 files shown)


boegel:
lgtm, let's wait for confirmation from @mstud?

mstud:
As already mentioned in #2182, it is unfortunately not sufficient to fix the issue. While using `--job` does not fail anymore, it now ignores `optarch` completely inside of a job, because the same mechanism that lead to the `dict` getting converted to a string before the second pass apparently now leads to the string getting converted to a string encapsulated in single quotes, which messes with the parsing.
If I also change 
`optarch_parts = self.options.optarch.split(OPTARCH_SEP)`
to
`optarch_parts = self.options.optarch[1:-1].split(OPTARCH_SEP)`
stripping off the quotes, it works just fine, but this seems a bit like a hack.

damianam:
Argh. @boegel why/where is it that happening? Shouldn't the options be keep verbatim, if they are going to be "reprocessed"?

boegel:
@damianam https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/tools/parallelbuild.py#L136

It quotes the arguments to make sure the shell doesn't trip over them, but it seems like something is going wrong in this case, `eb` should never actually see the quotes...

@mstud Can you look up the log line that says `Command template for jobs` in the log where you use `--job`? You can collect it either via `eb -ld` or using `--disable-cleanup-tmpdir` (see log file mentioned at very beginning).

damianam:
```
In [8]: subprocess.list2cmdline(['mystuff', 'more stuff', 'something is off', "--optarch='something:is weird,here:I believe'"]).replace('%', '%%')
Out[8]: 'mystuff "more stuff" "something is off" "--optarch=\'something:is weird,here:I believe\'"'
```

It seems like `list2cmdline` doesn't do what you expected it to do. I guess the same issue would appear in any option that accepts spaces.

boegel:
@damianam We could probably fix this by using the (monkey-patched) `shell_quote`, see `tools/options.py`

mstud:
I guess it's not relevant anymore after damianam's discovery, but in the `Command template for jobs` line the parameter shows up as:
`eb [...] "--optarch='Intel:xSSE4.2 -axCORE-AVX2;GCC:march=westmere -mtune=haswell'" [...]`

damianam:
@boegel that will result in a token like this: `'--optarch=\'Intel:xSSE4.2 -axCORE-AVX2;GCC:march=westmere -mtune=haswell\''`. That's precisely what we don't want, right?

Looks to me like we need to write a method to append options to a string, and wrap them around in quotes just if there are spaces outside of any quoted area. Something like this maybe?

```
import re

tokens=['software.eb', '-r path', "--optarch='Intel:xSSE4.2 -axCORE-AVX2;GCC:march=westmere -mtune=haswell'"]

blank_checker = re.compile(r".*('.*\s.*')|(\".*\s.*\").*")

tokens_str=""

for token in tokens:
    if " " in token:
        # Wrap in double quotes
        if not blank_checker.match(token) and "'" in token:
            tokens_str+=' '+'"'+token+'"'
        # Wrap in single quotes
        elif not blank_checker.match(token) and '"' in token:
            tokens_str+=" "+"'"+token+"'"
        # Wrap in single quotes
        elif not blank_checker.match(token):
            tokens_str+=" "+"'"+token+"'"
        else:
            tokens_str+=" "+token
    else:
        tokens_str+=" "+token

print(tokens_str)
```

It is ugly, I know, I don't like it either.......

boegel:
The problem is in `generate_cmd_line`, not in the use of `subprocess.list2cmdline`.

Here's what `subprocess.list2cmdline` spits out for options that include `--optarch="GCC:O3 -mtune=generic;Intel:O3 -xHost"`:

```python
[..., '--optarch=\'"GCC:O3 -mtune=generic;Intel:O3 -xHost"\'', ...]
```

So, we'll need to fix `generate_cmd_line`, probably via `eb_shell_quote` in `options.py`, I'm looking into it.

Options values that include spaces are quite rare, so this hasn't popped up before.

boegel:
@damianam problems should be fixed with additional changes in https://github.com/damianam/easybuild-framework/pull/2, confirmed by enhanced tests

boegel:
@mstud Please verify whether the additional changes fix your problem?

mstud:
@boegel Yes, it works properly now both with and without --job. Thank you very much.

boegel:
Good to go to be included in EasyBuild v3.2.0, thanks for the (partial ;)) fix @damianam and the feedback @mstud!

PR #2185, opened 2017-03-29
add support for --verify-easyconfig-filenames

boegel:
by popular request...

cc @shahzebsiddiqui, @damianam, @pneerincx

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1379,7 +1379,7 @@ def robot_find_easyconfig(name, version):
     return res
 
 
-def verify_easyconfig_filename(path, specs=None, parsed_ec=None):
+def verify_easyconfig_filename(path, specs, parsed_ec=None):
     """
     Check whether parsed easyconfig at specified path matches expected specs;
     this basically verifies whether the easyconfig filename corresponds to its contents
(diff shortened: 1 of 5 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1389,7 +1389,7 @@ def verify_easyconfig_filename(path, specs, parsed_ec=None):
     :param parsed_ec: (list of) EasyConfig instance(s) corresponding to easyconfig file
     """
     if isinstance(parsed_ec, EasyConfig):
-        ecs = [parsed_ec]
+        ecs = [{'ec': parsed_ec}]
```
wpoely86: is this a bug fix?

easybuild/tools/options.py
```diff
@@ -383,9 +384,10 @@ def override_options(self):
                              str, 'store', False, {'metavar': "PATH"}),
             'use-existing-modules': ("Use existing modules when resolving dependencies with minimal toolchains",
                                      None, 'store_true', False),
+            'verify-easyconfig-filenames': ("Verify whether filename of specified easyconfigs matches with contents",
```
wpoely86: put it to `True` by default?

easybuild/framework/easyconfig/easyconfig.py
boegel: yes, and the enhanced test covers it now

easybuild/tools/options.py
boegel: No, that would change the behaviour, sites should enable this if they want to stricter behaviour.
I don't consider the flexibility of filenames for specified easyconfigs a bug, but a feature. I'd get annoyed fi EasyBuild gets picky about filenames when it doesn't really matter; it only really matters for the dependency resolver, which already complains if an easyconfig file that it picks up by filename doesn't contain what is expected.

We can reconsider this in EasyBuild v4.0, but I'm in favour of leaving the default behaviour as is.

boegel:
Going in, thanks for the review @wpoely86 and the thumbs up @pneerincx

PR #2193, opened 2017-04-18
add support for toolchain options like 'extra_cxxflags' to specify extra compiler options

mboisson:
Tested with a NAMD recipe.

### DIFF
--- a/easybuild/tools/toolchain/compiler.py
+++ b/easybuild/tools/toolchain/compiler.py
@@ -93,8 +93,11 @@ class Compiler(Toolchain):
         'vectorize': (None, "Enable compiler auto-vectorization, default except for noopt and lowopt"),
         'packed-linker-options': (False, "Pack the linker options as comma separated list"),  # ScaLAPACK mainly
         'rpath': (True, "Use RPATH wrappers when --rpath is enabled in EasyBuild configuration"),
-        'extra_cflags': (None, "Specify extra CFLAGS options. Do not specify a leading dash, one is prepended already."),
-        'extra_fflags': (None, "Specify extra FFLAGS options. Do not specify a leading dash, one is prepended already."),
+        'extra_cflags': (None, "Specify extra CFLAGS options."),
+        'extra_cxxflags': (None, "Specify extra CXXFLAGS options."),
+        'extra_fflags': (None, "Specify extra FFLAGS options."),
+        'extra_fcflags': (None, "Specify extra FCFLAGS options."),
+        'extra_f90flags': (None, "Specify extra F90FLAGS options."),
     }
 
     COMPILER_UNIQUE_OPTION_MAP = None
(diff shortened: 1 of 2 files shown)


wpoely86 (dismissed):

easybuild/tools/toolchain/compiler.py
```diff
@@ -90,6 +90,8 @@ class Compiler(Toolchain):
         'openmp': (False, "Enable OpenMP"),
         'packed-linker-options': (False, "Pack the linker options as comma separated list"),  # ScaLAPACK mainly
         'rpath': (True, "Use RPATH wrappers when --rpath is enabled in EasyBuild configuration"),
+        'extra_cflags':(None,"Specify extra CFLAGS options. Do not specify a leading dash as one is prepended already."),
+        'extra_fflags':(None,"Specify extra FFLAGS options. Do not specify a leading dash as one is prepended already."),
```
boegel: style nitpicking: space after `:`

easybuild/tools/toolchain/compiler.py
```diff
@@ -107,6 +109,8 @@ class Compiler(Toolchain):
         'static': 'static',
         'unroll': 'unroll',
         'verbose': 'v',
+        'extra_cflags':'%(value)s',
+        'extra_fflags':'%(value)s',
```
boegel: style nitpicking: space after `:`

easybuild/tools/toolchain/compiler.py
```diff
@@ -118,13 +122,13 @@ class Compiler(Toolchain):
 
     COMPILER_CC = None
     COMPILER_CXX = None
-    COMPILER_C_FLAGS = ['cstd']
+    COMPILER_C_FLAGS = ['cstd','extra_cflags']
```
boegel: style nitpicking: space after `,`

easybuild/tools/toolchain/compiler.py
```diff
@@ -118,13 +122,13 @@ class Compiler(Toolchain):
...
     COMPILER_C_UNIQUE_FLAGS = []
 
     COMPILER_F77 = None
     COMPILER_F90 = None
     COMPILER_FC = None
-    COMPILER_F_FLAGS = ['i8', 'r8']
+    COMPILER_F_FLAGS = ['i8', 'r8','extra_fflags']
```
boegel: style nitpicking: space after `:`

boegel:
@mboisson do we want to be able to specify separate/different flags for C and C++ compilers (i.e. in `$CFLAGS` and `$CXXFLAGS`)?

boegel:
also, we're missing an enhanced test for this...

mboisson:
> @mboisso do we want to be able to specify separate/different flags for C and C++ compilers (i.e. in $CFLAGS and $CXXFLAGS)?


It would be nice, but there is no current distinction between CFLAGS and CXXFLAGS in the code for any other parameters. I am not sure why this is the case, but the purpose of this PR is not to introduce such as split (even though I think there should be one).

mboisson:
> also, we're missing an enhanced test for this...

I have to say I'm lost when it comes to tests. Any pointers as to what I should do ? Extra arguments can be anything. There is no sample code that I can compile for anything.

mboisson:
Any guidance as to how to create the missing tests ?

boegel:
@mboisson Examples of tests that check what goes into `$CFLAGS` & co and fiddle with toolchain options are in `test/framework/toolchain.py`, see for example `test_misc_flags_shared`

We should really extend this to cover `$CXXFLAGS` separately though, I can help with that, hopefully sometime next week.

Are you OK with not including this in EasyBuild v3.2.0 (which I hope to release next week)?

mboisson:
Sure. We have merged it in our own fork anyway, so we are using it.

mboisson:
Did you have time to extend on this ?

boegel:
@mboisson I'm afraid not, not sure when I'll find the time to flesh this out...

akesandgren:
@mboisson What is the status of this? Looks like we need to kick travis if we want to move forward on it...

mboisson:
Mmm, the status is issue-necropsy :P  I haven't touched this for two years. @boegel wanted some changes

bartoldeman:
I discussed this with @boegel briefly last week, had forgotten you filed a PR... basically this is one of only two changes we have left in our fork (the other change pertains to the permissions of the temporary file folders `/tmp/eb-*` so we can look at each others' logs on the build node, but that is still to be PR'ed.

He mentioned that a generic `--optflags` CLI option is another way to do this and we could have a good look at that.

Neither of us is working this week though :)

boegel:
Tested with Boost easyconfig from https://github.com/easybuilders/easybuild-easyconfigs/pull/11642 that uses `extra_cxxflags` in `toolchainopts`, works like a charm!

Thanks a lot both @mboisson and @bartoldeman!

boegel (approved):

Issue #991, opened 2014-08-01
Extensions using the same buildopts than main program


hajgato:
The extensions using the same buildopts than the main program. For extensions, we have to have the possibility to define separate buildopts, prebuilopts, and so on.

pescobar:
maybe related: https://github.com/hpcugent/easybuild-easyblocks/issues/587

verdurin:
Relevant mentions on IRC:

http://pastebin.com/EdYhRQQg

https://gist.github.com/chrisacbr/eed6b26bc49576c3b2707cf61a08c91d

gatagat:
I am new to eb, but I was surprised to see that the extensions are not handled just as normal main programs. I mean having a separate eb file for each extension and being able to include any eb as extension. That would provide all the options available for normal eb files out of the box.

boegel:
Support for specifying custom easyconfig parameters per extension is added with #2194, which should close this issue.

@gatagat The concept of extensions was introduced exactly to avoid having a separate easyconfig file per extension. Later on, we also added support for installing extensions like `numpy` as stand-alone modules, where you indeed need a separate easyconfig file.

boegel:
#2194 is merged so support for specifying `buildopts` and other easyconfig parameters will be included in the upcoming EasyBuild v3.2.0

PR #2194, opened 2017-04-20
add support for customising easyconfig parameters on a per-extension basis

boegel:
With this in place, custom values for easyconfig parameters like `buildopts` & co can be specified on a per-extension basis, just like is the case already for `source_urls`.

This fixes a long-standing issue (#991) that has been popping up more frequently recently, for example when certain Python packages must be installed with `pip`, which requires setting the `use_pip` parameter to `True`.

### DIFF
--- a/easybuild/framework/extension.py
+++ b/easybuild/framework/extension.py
@@ -48,7 +48,11 @@ class Extension(object):
     """
     def __init__(self, mself, ext, extra_params=None):
         """
-        mself has the logger
+        Constructor for Extension class
+
+        :param mself: parent Easyblock instance
+        :param ext: dictionary with extension metadata (name, version, src, patches, options, ...)
+        :param extra_params: extra custom easyconfig parameters to take into account for this extension
         """
         self.master = mself
         self.log = self.master.log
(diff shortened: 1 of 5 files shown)


boegel:
@wpoely86 please review?

easybuild/framework/extension.py
```diff
@@ -46,7 +46,7 @@ class Extension(object):
     """
     Support for installing extensions.
     """
-    def __init__(self, mself, ext):
+    def __init__(self, mself, ext, extra_params=None):
```
wpoely86: doc the arguments 😉

easybuild/framework/extension.py
```diff
@@ -68,6 +68,22 @@ def __init__(self, mself, ext):
         self.patches = self.ext.get('patches', [])
         self.options = copy.deepcopy(self.ext.get('options', {}))
 
+        if extra_params:
+            self.cfg.extend_params(extra_params, overwrite=False)
```
wpoely86: why not overwrite?

easybuild/framework/extension.py
```diff
@@ -68,6 +68,22 @@ def __init__(self, mself, ext):
...
+
+        # custom easyconfig parameters for extension are included in self.options
+        # make sure they are merged into self.cfg so they can be queried;
+        # unknown easyconfig parameters are ignored since self.options may include keys only there for extensions;
+        # this allows to specify custom easyconfig parameters on a per-extension basis
+        for key in self.options:
+            if key in self.cfg:
+                self.cfg[key] = self.options[key]
+                self.log.debug("Customising known easyconfig parameter '%s' for extension %s v%s: %s",
```
wpoely86: `%(name)s/%(version)s` ?

easybuild/framework/extension.py
boegel: To be honest, I'm not sure. It was added like this in #1498, and even though I can't think of a reason for it now, there must have been a good reason since support for not overwriting was added to `extend_params` in the same go. That reason may no longer be valid now though...

I'll leave this as is for now, but this is probably a bug since this means that the `runtest` easyconfig parameter that is hard set to `True` in `PythonPackage` (while the default is `False`) is not being set when Python packages are installed as extensions.
It's not a huge problem, but it does mean that the `scipy` tests are not being run when `scipy` is installed as an extension...

The only case where overwriting matters or not is when the easyblock that is used for installing the extension redefines a general easyconfig parameter, which only occurs for `runtest` for stuff like Python packages, at first sight, and probably only affects `scipy` (the `numpy` tests are always run, regardless of `runtest`, which is also a bug I guess).

Follow-up in #2195.

boegel:

PR #2197, opened 2017-04-29
clean up easyconfigs that are copied for inclusion in pull request

boegel:
cfr. #1674

enabled by default, can be disabled via `--disable-cleanup-easyconfigs`

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1508,6 +1508,27 @@ def det_location_for(path, target_dir, soft_name, target_file):
     return target_path
 
 
+def cleanup_easyconfigs(paths):
+    """
+    Clean up easyconfigs (in place) by filtering out comments/buildstats included by EasyBuild in archived easyconfigs
+    (cfr. FileRepository.add_easyconfig in easybuild.tools.repository.filerepo)
+
+    :param paths: list of paths to easyconfigs to clean up
+    """
+    regexs = [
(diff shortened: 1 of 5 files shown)


boegel:
@wpoely86 please review?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1508,6 +1508,27 @@ def det_location_for(path, target_dir, soft_name, target_file):
...
 
+def cleanup_easyconfigs(paths):
+    """
+    Clean up easyconfigs (in place) by filtering out comments/buildstats included by EasyBuild in archived easyconfigs
+    (cfr. FileRepository.add_easyconfig in easybuild.tools.repository.filerepo)
+
+    :param paths: list of paths to easyconfigs to clean up
+    """
+    regexs = [
+        re.compile(r"^# Built with EasyBuild.*\n", re.M),
```
wpoely86: why `re.M`? Do actually deleted the `\n` too?

boegel: In this case `re.M` isn't required, I think you're right.

Can you clarify the question on the `\n`? If you're asking if that's deleted too, yes, it is, and it should be, otherwise you'll end up with a blank line?

boegel: On 2nd thought, the `re.M` is required, to ensure that also a comment like `# Built with EasyBuild...` is filtered out if it doesn't occur on the 1st line (maybe the easyconfig was manually edited afterwards...).

boegel:

PR #2198, opened 2017-04-29
use devel logging where relevant in easybuild.tools.toolchain

boegel:
This will significantly clean up the log file produced under `--debug` or `-d`, while still allowing to get all the (internal) details via `--devel`


boegel:
@wpoely86 please review?

boegel:

PR #2199, opened 2017-04-29
perform sanity check after stage 2 of EasyBuild bootstrap script, module file should be in place

boegel:


boegel:
@wpoely86 please review?

boegel:
Tested extensively under a variety of configuration options that influence the location/name of the generated module, works as intended.

Issue #502, opened 2013-02-25
regtest should also use run_step from easyblock module


boegel:
The regtest functionality now uses it's own `perform_step` function for each of the build steps, instead of the `run_step` function used in non-regtest builds.

This has caused confusing problems like #498, since this code basically does the same.

The regtest functionality should be changed to also use the `run_step` function.

boegel:
fixed in #2203

PR #2203, opened 2017-05-04
call run_all_steps in regtest mode rather than running steps individually

boegel:
Installing FFTW was failing (only) under `--regtest` because `run_all_steps` was not being used in that context.

This is a long-standing issue, cfr. #502.


boegel:

PR #2208, opened 2017-05-05
bump version bootstrap script to sync with latest update

boegel:
This should have been part of #2199.

We should look into a way of verifying the version of the bootstrap script so updating it is not overlooked...


boegel:
@wpoely86 please review?

boegel:

Issue #2012, opened 2016-11-20
gc3pie fails on cleanup, directory already gone


rjeschmi:
I noticed that #1441 was still an issue, but in my case it is a "feature" since my jobs are succeeding except for:

```
== 2016-11-20 11:14:14,022 pbs.py:170 INFO Output line:
== 2016-11-20 11:14:14,022 pbs.py:356 INFO Updated resource 'pbs' status: free slots: -1, total running: 0, own running jobs: 0, own queued jobs: 0, total
 queued jobs: 0
== 2016-11-20 11:14:14,353 batch.py:740 WARNING Failed removing remote folder '/home/rjeschmi/.gc3pie_jobs/lrms_job.16ioTk2jpk': <class 'gc3libs.exception
s.TransportError'>: Could not remove directory tree '/home/rjeschmi/.gc3pie_jobs/lrms_job.16ioTk2jpk': OSError: [Errno 2] No such file or directory: '/hom
e/rjeschmi/.gc3pie_jobs/lrms_job.16ioTk2jpk'
== 2016-11-20 11:14:14,353 build_log.py:220 INFO GC3Pie job overview: 1 terminated, 1 failed (total: 19)
== 2016-11-20 11:14:44,871 pbs.py:170 INFO Output line:
```

It seems like the job directory is already cleaned up for some reason before this line. This might be a bug in gc3pie instead, but I'll put it here first.

wpoely86:
I'm hitting the same issue. Any clues @boegel or @riccardomurri ?
It makes that GC3Pie thinks that jobs have failed while the succeed.

boegel:
Is the home directory on a shared filesystem (NFS)? Not sure if that was the case for @rjeschmi...

riccardomurri:
Re: the log lines reported by @rjeschmi , this looks like a bug in GC3Pie (directory cleaned up twice). So far, I had only seen it happening with the "localhost" execution backend, never with batch systems, though...

Re: the general issue of GC3Pie mis-detecting the fate of jobs: it's hard to tell what's happening without some more data printed to the logs.  I can cook up a PR to make DEBUG logs more verbose but likely not before end of the week (travelling to HEPiX and I have yet my talk to prepare...)

wpoely86:
It's on NFS.

@riccardomurri Sure, I'm happy to provide any additional info you want. 

[...]

PR #2211, opened 2017-05-08
Log detailed dump of GC3Pie task objects

riccardomurri:
Meant mainly to aid debugging in issue #2012 but could be useful elsewhere.

### DIFF
--- a/easybuild/tools/job/gc3pie.py
+++ b/easybuild/tools/job/gc3pie.py
@@ -271,11 +271,15 @@ class GC3Pie(JobBackend):
         print_msg("GC3Pie job overview: %s (total: %s)" % (states, self.job_cnt),
                   log=self.log, silent=build_option('silent'))
         if build_option('debug'):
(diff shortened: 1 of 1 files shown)


easybuild/tools/job/gc3pie.py
```diff
@@ -260,3 +270,12 @@ def _print_status_report(self):
         states = ', '.join(["%d %s" % (stats[s], s.lower()) for s in stats if s != 'total' and stats[s]])
         print_msg("GC3Pie job overview: %s (total: %s)" % (states, self.job_cnt),
                   log=self.log, silent=build_option('silent'))
+        if __debug__:
```
boegel: Since `eb` runs `python -O -m easybuild.main`, this is never going to pass?

Maybe you want to use `if build_option('debug'):`?

boegel:
I don't think this is still relevant, and it has bitrotted (merge conflict), so closing...

PR #2214, opened 2017-05-11
Allow https on direct download in sources

wpoely86:


boegel:
Thanks for the fix @wpoely86, and the review @ocaisa!

PR #2215, opened 2017-05-11
work towards enforcing availability of SHA256 checksums for all sources/patches

boegel:
This PR makes a couple of changes:

* also support `sha256` and `sha512` as checksum types
* auto-detect MD5 vs SHA256 checksums when no checksum type is specified (based on length of provided checksum)
* report both MD5 and SHA256 checksums in log for all sources/patches (incl. extensions)
* add `--enforce-checksums` option to enforce presence of checksums for all sources/patches

It's a first step towards having SHA256 checksums for all sources/patches in all centrally provided easyconfigs.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -64,7 +64,7 @@ from easybuild.tools.build_log import print_error, print_msg
 from easybuild.tools.config import build_option, build_path, get_log_filename, get_repository, get_repositorypath
 from easybuild.tools.config import install_path, log_path, package_path, source_paths
 from easybuild.tools.environment import restore_env, sanitize_env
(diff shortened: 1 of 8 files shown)


ocaisa:
We could also add the checksums automatically to the archived easyconfig

boegel:
@ocaisa Sure, but I think that should be done in a separate PR, since that's a bit fiddly (you need to inject `checksums =`, preferably under `sources = `...)

Also, there's a bit of chicken-egg situation there, I think, especially if we want to add SHA256 checksums automatically, i.e. they could only be consumed with an EasyBuild version that includes support for (auto-detection of) SHA256 checksums...

vanzod:
This PR seems good to me. Nice work

boegel:
Thanks for the review @wpoely86 and @vanzod!

PR #2220, opened 2017-05-12
make sure test account & accompanying token is used in tests for github.py

boegel:
some tests occasionally fail in my fork because the available GitHub token isn't always being used, these (trivial) changes fix that problem


boegel:
@wpoely86 Please review?

PR #2221, opened 2017-05-12
only use MPD for old versions of Intel MPI (<4.1) in mpi_cmd_for

boegel:


boegel:
@wpoely86 please review

boegel:

PR #2225, opened 2017-05-16
escape dots in package filename to ensure correct match in derive_alt_pypi_url

boegel:


boegel:

Issue #153, opened 2012-08-24
make sure no modules are loaded when EasyBuild is started


boegel:
(old internal ticket 346)

EasyBuild should check whether (software) modules (built with EasyBuild) were loaded when it stars, and refuse to continue if it detects it.

Having modules loaded outside of EasyBuild could make or break build, and the result may be different with an environment in which no modules were loaded.

Currently, it only checks whether the software it is trying to built is loaded, but it should do the same for dependencies and non-dependencies.

damianam:
Update on this ancient issue:

I think it makes sense to have an option to unload everything except a set of modules listed in a whitelist. At least in our setup that will be a nice way to sanitize the environment without breaking other things.

wpoely86:
@damianam use sticky modules?

damianam:
We have a module for setting up EB related variables. And another one for GC3Pie. And another one for EB itself. I don't want to make them sticky and force me to use `--purge` in my shell just to allow EB to have an easier way to sanitize its environment prior to running. That is for our case, but other people might have other reasons to keep other modules loaded, and not make them sticky.  It has implications in the environment outside of EB, doesn't seem like a clean solution to me.

ocaisa:
You could just make sure there are no modules loaded that define `EBROOT*`, that's pretty clean I think

damianam:
It is clean in our case, but if somebody mixes EB modules with other modules that shouldn't be loaded when installing software it wouldn't be enough. Besides that, it would unload the EasyBuild module, so you would have to make at least one exception already

ocaisa:
I think you'd be safe if you only unloaded the modules after the EB process has finished grabbing its configuration

migueldiascosta:
I think I just hit this problem with a `PythonPackage` (e.g. GPAW) that depends on another one (e.g. ASE), in a counterintuitive way: if ASE is already loaded, the sanity check fails saying ASE cannot be found, but if ASE is not loaded, the sanity check passes

Because it is counterintuive, it would be helpful that at least some warning was issued at the beginning saying a dependency shouldn't be loaded

(in any case, why doesn't the sanity check find/reload the dependency when it was already loaded?)

vanzod:
I also hit the same problem when trying to build `Python` with `foss` toolchain when the `ncurses` library built with the `dummy` toolchain has been built with separate terminfo symbols (see PR #3545) as described in #4049.

[...]

PR #2228, opened 2017-05-19
add support for --detect-loaded-modules

boegel:
This makes EasyBuild check early on whether any EasyBuild-generated modules are loaded in the current environment (fix for #153).

Via `--detect-loaded-modules`, you can configure EasyBuild to either print a big fat warning (`warn`, done by default), raise an error (`fail`) or don't complain (`ignore`, equivalent to current behaviour).

The detection is done by checking whether any `$EBROOT*` environment variables are defined, which are then mapped to the loaded modules that define them.

Through `--allow-loaded-modules`, a list of software names for which modules are allowed to be loaded can be specified.

Still WIP because I'd like to add support for trying to automagically fix the environment by either running `module purge` or trying to unload all the loaded non-ignored EasyBuild modules...

Thoughts on this @wpoely86, @vanzod, @migueldiascosta, @ocaisa, @damianam?


easybuild/tools/modules.py
```diff
@@ -718,6 +719,53 @@ def loaded_modules(self):
...
+                out = self.show(loaded_module)
+                for key in eb_module_keys[:]:
+                    if key in out:
+                        loaded_eb_modules.append(loaded_module)
+                        eb_module_keys.remove(key)
+
+            # warn about $EBROOT* environment variables without matching loaded module
+            if eb_module_keys:
+                tup = (len(eb_module_keys), ROOT_ENV_VAR_NAME_PREFIX, '$' + ', $'.join(eb_module_keys))
+                print_warning("Found %d defined $%s* environment variables without matching loaded module: %s" % tup)
```
wpoely86: Shouldn't this be fatal?

boegel: I'm not sure, who knows what's in the environment...
It's definitely worth warning about it though.

Also, I wouldn't be surprised if people already trick EB into doing this by defining `$EBROOT*` variables manually.

Which makes me really, I need to check how this plays together with the support for using external modules... I know we define `$EBROOT*` variables for external modules that get loaded (cfr. http://easybuild.readthedocs.io/en/latest/Using_external_modules.html#handling-of-provided-metadata), but I need to check where we do that exactly.

boegel:
With the last update, I consider this ready. As opposed to what I proposed earlier, I think it makes sense to make the possible values for `--detect-loaded-modules` mutually exclusive.

In the current implementation, when using `purge` or `unload`, a warning will be printed to inform the user that `purge` is run or that the detected loaded modules are unloaded, so it doesn't make much sense to combine this with `warn`. Combining it with `fail` doesn't make sense either, what's the point in purging or unloading if we're going to error out anyway...

Thoughts @wpoely86, @vanzod, @damianam?

easybuild/tools/modules.py
damianam: I agree with @wpoely86. Warnings tend to be ignored, but more importantly, a warning allows to proceed. So the door is open to break something unintentionally.

damianam:
@boegel, like we discussed in the last telco, I think it would be a good idea to add to the warning or error messages a tip on how to move forward.

Also, I think at the moment you are simply checking for EB generated modules. But taking a look at the external modules being used might be also desirable. Wouldn't you agree? I think simply appending to `eb_module_keys` the list of external modules included in the `.cfg` file would be enough (I am thinking about Cray here, I am not sure about other use cases).

easybuild/tools/modules.py
boegel: Yes, but it may also be totally harmless, so making it fatal seems overly harsh to me...

I guess we could a configuration option for it:

```
--check-ebroot-env-vars={ignore,warn,fail}
```

damianam: I guess it simply depends on how you want to treat "may be harmless" vs "won't be harmless" vs "will be harmless". It looks safer to me to err on the safe side and decline to proceed it is is not sure that it will be harmless. But in any case, I have no strong objection against keeping it as warning.

boegel:
@damianam I'd like to get feedback from @gppezzi on the latter, especially w.r.t. Cray where you typically have a bunch of modules loaded by default, which are likely going to include ones that EasyBuild leverages as external modules.

easybuild/tools/modules.py
boegel: Yeah, I guess you're right, how this should be treated is site-specific.

I've added an option for this now, default is to issue a warning, but you can also get an error, totally ignore it, or just unset them and continue (with a warning).

test/framework/modules.py
```diff
@@ -750,6 +750,110 @@ def test_exit_code_check(self):
...
         self.assertErrorRegex(EasyBuildError, error_pattern, self.modtool.load, ['nosuchmoduleavailableanywhere'])
 
+    def test_check_loaded_modules(self):
+        """Test check_loaded_modules method."""
+        # try and make sure we start with a clean slate
+        self.modtool.purge()
+
+        def check_loaded_modules():
+            "Helper function to run check_loaded_modules and check on stdout/stderr."
+            # there should be no errors/warnings by default is no (EasyBuild-generated) modules are loaded
```
damianam: `if no` instead of `is no`?

test/framework/modules.py
```diff
@@ -750,6 +750,110 @@ def test_exit_code_check(self):
...
+        self.assertEqual(check_loaded_modules(), '')
+
+        self.modtool.unload(['EasyBuild/fake'])
+
+        # load OpenMPI module, which also loads GCC & hwloc
+        self.modtool.load(['OpenMPI/1.6.4-GCC-4.6.4'])
+
+        # default action is to print a clear warning message
+        stderr = check_loaded_modules()
+        patterns = [
```
damianam: Put the message in a variable in `modules.py` and read it from there? Otherwise you have to rewrite the message twice for any change on it.

test/framework/modules.py
```diff
@@ -750,6 +750,110 @@ def test_exit_code_check(self):
...
+            r"^WARNING: Found one or more non-ignored loaded \(EasyBuild-generated\) modules in current environment:",
+            r"^\* GCC/4.6.4",
+            r"^\* hwloc/1.6.2-GCC-4.6.4",
+            r"^\* OpenMPI/1.6.4-GCC-4.6.4",
+            "This is not recommended since it may affect the installation procedure\(s\) performed by EasyBuild.",
+            "To make EasyBuild allow particular loaded modules, use the --allow-loaded-modules configuration option.",
+            "To specify action to take when loaded modules are detected, use "
+                "--detect-loaded-modules={error,ignore,purge,unload,warn}",
+        ]
+        for pattern in patterns:
```
damianam: Change this to a single check? Slicing the pattern in subpatterns is kind of weird and will produce also a weird output with too many "Pattern found" messages. Alternatively, maybe check for patterns, but print just the ones that do not match?

easybuild/tools/modules.py
```diff
@@ -718,6 +720,86 @@ def loaded_modules(self):
...
+                action = build_option('detect_loaded_modules')
+
+                if action == ERROR:
+                    raise EasyBuildError(verbose_msg)
+
+                elif action == IGNORE:
+                    self.log.info("Found non-ignored loaded (EasyBuild-generated) modules, but ignoring it as configured")
+
+                elif action == PURGE:
+                    msg = "Found non-ignored loaded (EasyBuild-generated) modules, running 'module purge': %s"
```
damianam: ` msg = "Found non-ignored loaded (EasyBuild-generated) modules (%s), running 'module purge'"`?

test/framework/modules.py
```diff
@@ -750,6 +750,110 @@ def test_exit_code_check(self):
...
+        # reconfigure EasyBuild to ignore loaded modules for GCC & hwloc & error out when loaded modules are detected
+        options = init_config(args=['--allow-loaded-modules=GCC,hwloc', '--detect-loaded-modules=error'])
+        build_options = {
+            'allow_loaded_modules': options.allow_loaded_modules,
+            'detect_loaded_modules': options.detect_loaded_modules,
+        }
+        init_config(build_options=build_options)
+
+        # error mentioning 1 non-ignored module (OpenMPI), both GCC and hwloc loaded modules are ignored
+        error_pattern = "Found one or more non-ignored loaded .* module.*\n\* OpenMPI/1.6.4-GCC-4.6.4"
```
damianam: Same as before? Sync this automatically with the message in `modules.py` instead of rewriting it here?

test/framework/modules.py
```diff
@@ -750,6 +750,110 @@ def test_exit_code_check(self):
...
+        expected = "WARNING: Unloading non-ignored loaded (EasyBuild-generated) modules: OpenMPI/1.6.4-GCC-4.6.4"
+        self.assertEqual(check_loaded_modules(), expected)
+
+        # when loaded modules are ignored there are no warnings/errors
+        self.modtool.load(['OpenMPI/1.6.4-GCC-4.6.4'])
+        build_options.update({'detect_loaded_modules': 'ignore'})
+        init_config(build_options=build_options)
+        self.assertEqual(check_loaded_modules(), '')
+
+        # error if any $EBROOT* environment variables are defined that don't match a loaded module
```
damianam: Make this a separate test?

easybuild/tools/modules.py
```diff
@@ -718,6 +720,86 @@ def loaded_modules(self):
...
+            loaded_eb_modules = []
+            for loaded_module in loaded_modules:
+                out = self.show(loaded_module)
+                for key in eb_module_keys[:]:
+                    if key in out:
+                        loaded_eb_modules.append(loaded_module)
+                        eb_module_keys.remove(key)
+
+            # warn about $EBROOT* environment variables without matching loaded module
+            if eb_module_keys:
```
damianam: Flush this out to a different method?

test/framework/modules.py
damianam: Couldn't this pattern also match if GCC or hwloc are incorrectly not ignored?

boegel:
docs update @ https://github.com/hpcugent/easybuild/pull/344

test/framework/modules.py
boegel: Part of the reason for copy-pasting the message in the tests is so that we know when the message gets changed (potentially by accident), so this is intentional.

test/framework/modules.py
boegel: Only when one of the patterns isn't found, this will produce output (as a failing test).

The check is an assertion, so you'll see something like:

```
AssertionError: Pattern ... found in ...
```

So, the assertion is that a certain pattern should be found, and if not, the assertion fails and you get an assertion error.

This is done consistently across all tests. It looks backwards, but it's not if you interpret it right.

I could make this one big pattern, sure, but that makes spotting a mismatch visually a lot harder, so I prefer to break it down in parts. Speedup of the tests is going to be negligible, we should optimise for human effort here.

test/framework/modules.py
boegel: Hardcoding of the message is deliberate.

I'll make sure that also checking for just a single module name (`OpenMPI`) being mentioned is checked (by injecting a `$` to enforce that the `OpenMPI` line is the last and only module being mentioned).

test/framework/modules.py
boegel: The test is for the `check_loaded_modules` method in `easybuild/tools/modules.py`, which also checks the orphaned `$EBROOT*` env vars, so it makes sense to keep it as a single test?

easybuild/tools/modules.py
boegel: The reason for checking for orphaned `$EBROOT*` env vars is because that's we rely on for checking for loaded (EasyBuild-generated) modules, so I prefer keeping this together in the same method.

If I flesh it out, we would need to call boths methods in sequence anyway, and we'd have to duplicate some of the code, i.e. collecting defined `$EBROOT*` env vars, and tying the knot with the loaded modules...

easybuild/tools/modules.py
boegel: OK, I'll change that if you feel it's clearer, thanks for the suggestion!

boegel:
Thanks for the review @vanzod!

@damianam I'll go ahead and merge this in to be included with EasyBuild v3.3.0, but do let me know if there's anything you want to get back to w.r.t. my replies to your review.

PR #2230, opened 2017-05-22
pass down additional arguments to copy_dir down to shutil.copytree

boegel:
`shutil.copytree` has optional arguments like `ignore` and `symlinks`, `copy_dir` should support those too


easybuild/tools/filetools.py
```diff
@@ -1490,7 +1490,7 @@ def copy_file(path, target_path, force_in_dry_run=False):
             raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
 
 
-def copy_dir(path, target_path, force_in_dry_run=False):
+def copy_dir(path, target_path, force_in_dry_run=False, **kwargs):
```
wpoely86: add to the doc string

boegel:

PR #2232, opened 2017-05-31
avoid reloading already loaded modules that extend $MODULEPATH

boegel:
Fix for an issue reported by @akesandgren where reloading modules for the `icc`,`ifort`,`CUDA` toolchain components of the `intelcuda` caused problems with picking up wrong dependencies from the wrong locations because of order changes in `$MODULEPATH`.

This should also help avoiding load storms since Lmod does an unload/reload of all modules every time something changes to `$MODULEPATH`.

@ocaisa, @damianam, @geimer: please review?

### DIFF
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -192,12 +192,16 @@ class ModulesTest(EnhancedTestCase):
         # if GCC is loaded again, $EBROOTGCC should be set again, and GCC should be listed last
         self.modtool.load(['GCC/4.6.4'])
         self.assertTrue(os.environ.get('EBROOTGCC'))
-        self.assertTrue(self.modtool.loaded_modules()[-1] == 'GCC/4.6.4')
+        if isinstance(self.modtool, Lmod):
+            # order of loaded modules only changes with Lmod
+            self.assertTrue(self.modtool.loaded_modules()[-1] == 'GCC/4.6.4')
 
         # set things up for checking that GCC does *not* get reloaded when requested
         del os.environ['EBROOTGCC']
         self.modtool.load(['OpenMPI/1.6.4-GCC-4.6.4'])
-        self.assertTrue(self.modtool.loaded_modules()[-1] == 'OpenMPI/1.6.4-GCC-4.6.4')
+        if isinstance(self.modtool, Lmod):
+            # order of loaded modules only changes with Lmod
+            self.assertTrue(self.modtool.loaded_modules()[-1] == 'OpenMPI/1.6.4-GCC-4.6.4')
 
         # reloading can be disabled using allow_reload=False
         self.modtool.load(['GCC/4.6.4'], allow_reload=False)
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -178,6 +178,18 @@ class ModulesTest(EnhancedTestCase):
         for mod in mods:
             self.assertErrorRegex(EasyBuildError, '.*', self.modtool.load, [mod])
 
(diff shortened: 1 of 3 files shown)


test/framework/modules.py
```diff
@@ -178,6 +178,18 @@ def test_load(self):
...
+        del os.environ['EBROOTGCC']
+        self.assertTrue('GCC/4.7.2' in self.modtool.loaded_modules())
+        self.modtool.load(['GCC/4.7.2'])
+        self.assertTrue(os.environ.get('EBROOTGCC'))
+
+        # reloading can be disabled using allow_reload=False
+        del os.environ['EBROOTGCC']
+        self.modtool.load(['GCC/4.7.2'], allow_reload=False)
+        self.assertEqual(os.environ.get('EBROOTGCC'), None)
+
```
damianam: I am not sure I like the test. It does the job but if the target is to make sure the `$MODULEPATH` stays the same, doesn't it make more sense to check if it changed in anyway, rather than rely in a variable `$EBROOTXXX` that might or might not be set in the future (due to a bug, or changes in EB)?

Also, checking for changes in the `$MODULEPATH` won't be enough in flat module naming schemes, I guess the order of the loaded modules should be checked as well. But it seems pretty doable.

akesandgren:
@damianam any thoughts on the rest of the changes?

test/framework/modules.py
boegel: It's not like this will go unnoticed if `$EBROOTGCC` would no longer be set by the GCC module (which is unlikely to happen imho, I see no reason for it?), since we also assert that it *is* set higher up.
We can't check for the order in `$MODULEPATH` with this particular module, since it lives in flat MNS.

I can't think of another way to check whether a module actually gets reloaded... Maybe I can (also) check the `$PATH` entries (we should see the entry for `GCC` added again when it's really reloaded).

When reloading a module, I don't think anything changes in the order in `module list`, even in a module hierarchy?

That said, I can enhance the test if you feel strongly about it, or write a separate test to explicitly check that the order in `$MODULEPATH` in unaffected when modules are being reloaded with `allow_reload=False`, since this test is purely checking what the `load` method does...

damianam: Well, the order does change, even ignoring hierarchies (at least with lmod, I am not sure with tmod):

```
~]$ ml

Currently Loaded Modules:
  1) GCCcore/.5.4.0 (H)   2) binutils/.2.28 (H)   3) StdEnv (H)   4) icc/.2017.2.174-GCC-5.4.0 (H)   5) ifort/.2017.2.174-GCC-5.4.0 (H)   6) Intel/2017.2.174-GCC-5.4.0

  Where:
   H:  Hidden Module



~]$ ml StdEnv
~]$ ml

Currently Loaded Modules:
  1) icc/.2017.2.174-GCC-5.4.0 (H)   2) ifort/.2017.2.174-GCC-5.4.0 (H)   3) Intel/2017.2.174-GCC-5.4.0   4) GCCcore/.5.4.0 (H)   5) binutils/.2.28 (H)   6) StdEnv (H)
```

I agree with you that we have no reason at all for getting rid of `$EBROOTXXX`. It is simply that instead of really checking if the module has been reloaded directly, you are checking for the byproducts of loading modules, which works fine, but is not the same. But if lmod and tmod behave differently here I think your check is good enough, no need to over-engineer it.

damianam:
@akesandgren not really, they lgtm.

boegel:
Going in, thanks for the review @akesandgren and @damianam!

PR #2234, opened 2017-05-31
[WIP/RFP] Secondary subtoolchains

bartoldeman:
With the introduction of the iimkl (Intel + MKL without MPI) toolchain in our software stack it could not be a subtoolchain with the current infrastructure. In fact the subtoolchain structure becomes a DAG:
```
   iomkl
   /    \
iompi iimkl
   \     /
   iccifort
      |
   GCCcore
      |
   dummy
```
This patchset changes iomkl to have two subtoolchains. However it does not implement a DAG, as it assumes that the sub-subtoolchains of iomkl (the subtoolchains of iimkl and iompi, that is, iccifort) are identical.

This is sufficient for our setup but feels a bit quick and dirty. Please let me know what you think.


easybuild/toolchains/iomkl.py
```diff
@@ -31,14 +31,15 @@
 """
 
 from easybuild.toolchains.iompi import Iompi
+from easybuild.toolchains.iimkl import Iimkl
 from easybuild.toolchains.fft.intelfftw import IntelFFTW
 from easybuild.toolchains.linalg.intelmkl import IntelMKL
 
 
-class Iomkl(Iompi, IntelMKL, IntelFFTW):
+class Iomkl(Iompi, Iimkl, IntelMKL, IntelFFTW):
```
ocaisa: Since your inheriting from `Iimkl` you don't need to also inherit from `IntelMKL, IntelFFTW` (they're already there)

ocaisa:
What does the toolchain hierarchy output look like after this?

ocaisa:
For me, I'm ok with the concept but I do think you should be verifying that the underlying toolchains of `iimkl` and `iimpi` are consistent...but I think that may already be happening because I believe there is already a uniqueness check in the toolchain hierarchy.

bartoldeman:
For toolchain output you mean this using `--info`? This is using minimal toolchains including dummy:
```
INFO Found toolchain hierarchy for toolchain {'version': '2016.4', 'name': 'iomkl'}: [{'version': '', 'name': 'dummy'}, {'version': '5.4.0', 'name': 'GCCcore'}, {'version': '2016.4', 'name': 'iccifort'}, {'version': '2016.4', 'name': 'iompi'}, {'version': '2016.4', 'name': 'iimkl'}, {'version': '2016.4', 'name': 'iomkl'}]
```

easybuild/toolchains/iomkl.py
boegel: +1, but this may have impact on how toolchains are verified by EasyBuild (not sure though)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -134,6 +134,71 @@ def cache_aware_func(toolchain):
 
     return cache_aware_func
 
+def get_subtoolchain_in_deps(parsed_ec, subtoolchain_name, current_tc_name):
```
boegel: @bartoldeman This is begging for a dedicated unit test, to verify that it works as intended (and doesn't get broken in the future somehow)...

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -166,73 +231,22 @@ def get_toolchain_hierarchy(parent_toolchain):
...
-                if depdep['external_module']:
-                    continue
-
-                cands.append({'name': depdep['name'], 'version': depdep['version'] + depdep['versionsuffix']})
-                cands.append(depdep['toolchain'])
-
-        # only retain candidates that match subtoolchain name
-        cands = [c for c in cands if c['name'] == subtoolchain_name]
+        if isinstance(subtoolchain_name, list):
+            for name in subtoolchain_name[1:]:
```
boegel: please clarify with a comment why you are skipping the 1st subtoolchain?

boegel:
@bartoldeman This gives you a flat list, i.e. not a DAG, which basically means there's an order implied between `iompi` and `iimkl`, while there is none. I.e., either of both will be considered first, while in a DAG you would treat both equally, and act accordingly when there's a hit for both of them.

That's not necessarily a big issue though, since which one you pick may not matter too much, and having a hit for both may actually not happen in practice: software that doesn't depend on MPI would only have an easyconfig with `iimkl`, while something that doesn't depend on MPI would only be there with `iompi`...

So, I think this is OK. I assume this work as intended for you? Do you have some practical examples?

bartoldeman:
Hello Kenneth, Alan,

thanks for the feedback! Yes this is working as intended for us. One example I have is mpi4py (iomkl), which depends on numpy or in our case the scipy stack module (iimkl).

I won't have time to work on this this week though. It will have to wait until next week or the week there after. Our nix+easybuild stack is moving into production so these are exciting but busy times.

Bart

PR #2235, opened 2017-05-31
fix GNU_SOURCE template by adding missing /gnu/

boegel:
fix for bug reported by @rcespa (cfr. https://github.com/hpcugent/easybuild-framework/issues/2233)


boegel:
Thanks for reporting this @rcespa!

Issue #2237, opened 2017-06-01
Copy patch file fails if easyconfig archive path is in robot path and file already archived


vanzod:
After #2202 got merged, if the easyconfig files archive path is present in the robot path and the patch file has already been archived during a previous build, EB fails to copy the file because source and destination are the same.

Here is the traceback:
```Traceback (most recent call last):
  File "/home/davide/Git/EasyBuild-develop/easybuild-framework/easybuild/main.py", line 117, in build_and_install_software
    (ec_res['success'], app_log, err) = build_and_install_one(ec, init_env)
  File "/home/davide/Git/EasyBuild-develop/easybuild-framework/easybuild/framework/easyblock.py", line 2597, in build_and_install_one
    repo.add_patch(patch['path'], app.name)
  File "/home/davide/Git/EasyBuild-develop/easybuild-framework/easybuild/tools/repository/filerepo.py", line 129, in add_patch
    copy_file(patch, full_path)
  File "/home/davide/Git/EasyBuild-develop/easybuild-framework/easybuild/tools/filetools.py", line 1487, in copy_file
    shutil.copy2(path, target_path)
  File "/usr/lib/python2.7/shutil.py", line 130, in copy2
    copyfile(src, dst)
  File "/usr/lib/python2.7/shutil.py", line 69, in copyfile
    raise Error("`%s` and `%s` are the same file" % (src, dst))
Error: `/opt/easybuild/ebfiles_repo/XZ/XZ-5.2.2_compat-libs.patch` and `/opt/easybuild/ebfiles_repo/XZ/XZ-5.2.2_compat-libs.patch` are the same file```

boegel:
fixed in #2241

PR #2239, opened 2017-06-01
Catch exception shutil.Error

vanzod:
If the `shutil.copy2` operation fails for any reason (e.g. source and destination are the same), catch the `shutil.Error` exception.
This is a way to solve issues like #2237, where the test report shows `FAIL` even if the build ends successfully and no relevant messages are added to the debug log.


boegel:
lgtm, but let's enhance the test for `copy_file`, done in https://github.com/vanzod/easybuild-framework/pull/1

boegel:
Thanks @vanzod!

PR #2243, opened 2017-06-21
Fix --set-default-module flag

victorusu:
Now, modules with prefix and suffix versions can be set as default


boegel:
@victorusu Can we enhance the tests to make sure the bug you fixed here doesn't get reintroduced?

boegel:
@victorusu psst, always target `develop` for PRs (I know, it's easy to overlook); I've changed the target branch just now

boegel:
@victorusu Test added in https://github.com/victorusu/easybuild-framework/pull/2, please review/merge that PR to make this PR good to go?

victorusu:
I merged the test. And I think it is good to go! Thanks @boegel!

boegel:
Going in, thanks for the fix @victorusu!

PR #2248, opened 2017-06-27
resolve symlinks to location of 'eb' in get_paths_for

boegel:
Fix for bug reported by @pescobar 

The easyconfig files included with EasyBuild were not being found because the `eb` command was found in `/bin/eb` (which is a symlink to `/usr/bin/eb`), while the easyconfig files were located in `/usr/easybuild/easyconfigs`.


boegel:

PR #2253, opened 2017-06-29
Add fosscuda toolchain

vanzod:
Add a `fosscuda` toolchain in sync with `foss` releases with the addition of `CUDA` support.

### DIFF
--- a/easybuild/toolchains/fosscuda.py
+++ b/easybuild/toolchains/fosscuda.py
@@ -1,0 +1,41 @@
+##
+# Copyright 2013-2017 Ghent University
+#
+# This file is part of EasyBuild,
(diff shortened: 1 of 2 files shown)


vanzod:
The main issue now is the fact that there is no module that prepends `<prefix>/modules/all/Compiler/GCC-CUDA` to `MODULEPATH`. Ideally this should be done in `gcccuda` toolchain module.

boegel:
@vanzod Did you mean `Compiler/GCCcore-CUDA`? That should get taken care of by `det_modpath_extensions` in `hierarchical_mns.py`...

boegel:
replaced (partially at least) with #2507

PR #2254, opened 2017-06-29
add support for 'allow_prepend_abs_path' easyconfig parameter

boegel:
Relative paths are currently hard enforced in `modextrapaths`, while there are situations where passing an absolute path actually makes sense.

We should keep discouraging the use of absolute paths by default though, to avoid mistakes like specifying `/bin` while really the `bin` subdirectory in the install directory was meant...


boegel:
@wpoely86 please review?

boegel:

Issue #2252, opened 2017-06-29
Keeping an open search but restricting the robot


ocaisa:
To my understanding it's not possible to have an open search but restrict the robot. At the moment you can restrict the robot through `EASYBUILD_ROBOT` and `EASYBUILD_ROBOT_PATHS`, but once you set `EASYBUILD_ROBOT_PATHS` your search space is also restricted. It would be nice if these could be tweaked independently (say with a new setting `EASYBUILD_SEARCH_PATHS`).
This makes sense when you start thinking about looking into other people's repos.

boegel:
@ocaisa Makes sense to me, and it should be fairly easy to do.

The `search_easyconfigs` function in `easybuild.tools.robot` now only considers `robot_path` (which is a combination of the paths specified in `--robot` and `--robot-paths`, according to the rules outlines at http://easybuild.readthedocs.io/en/latest/Using_the_EasyBuild_command_line.html#robot-search-path), but it can be easily extended to also consider (the new) `search_paths` if one or more paths were specified for it.

Are you up for taking a stab at this yourself?

boegel:
follow up in #2255

PR #2255, opened 2017-06-29
Add support for --search-paths to extend list of locations considered by --search/-S

ocaisa:
Initial attempt at supporting `search_path`, fix for #2252


ocaisa:
@boegel Doesn't work...says `search_paths` is not a build option

ocaisa:
Fixed that part but don't know where to actually set that value

easybuild/tools/robot.py
```diff
@@ -356,7 +356,7 @@ def resolve_dependencies(easyconfigs, modtool, retain_all_deps=False):
 
 def search_easyconfigs(query, short=False, filename_only=False, terse=False):
     """Search for easyconfigs, if a query is provided."""
-    search_path = build_option('robot_path')
+    search_path = build_option('search_paths')
```
boegel: @ocaisa This is not what I had in mind, I was thinking about letting `--search-paths` empty by default, and doing this:

```python
search_path = build_option('robot_path')
if not search_path:
    search_path = [os.getcwd()]
extra_search_paths = build_option('search_paths')
if extra_search_paths:
    search_path.extend(extra_search_paths)

File: easybuild/tools/options.py [hunk 2]
```diff
@@ -957,7 +969,7 @@ def show_config(self):
 
         # options that should never/always be printed
         ignore_opts = ['show_config', 'show_full_config']
-        include_opts = ['buildpath', 'installpath', 'repositorypath', 'robot_paths', 'sourcepath']
+        include_opts = ['buildpath', 'installpath', 'repositorypath', 'robot_paths', 'search_paths', 'sourcepath']
```
boegel: I wouldn't include it here to make it always printed, just let it be printed if it's different from the default (works automagically)?

easybuild/tools/options.py
```diff
@@ -823,12 +829,18 @@ def _postprocess_config(self):
...
+
         if self.options.robot is not None:
             # paths specified to --robot have preference over --robot-paths
             # keep both values in sync if robot is enabled, which implies enabling dependency resolver
             self.options.robot_paths = [os.path.abspath(path) for path in self.options.robot + self.options.robot_paths]
             self.options.robot = self.options.robot_paths
 
+        # Update the search_paths keeping it in sync with robot_paths
+        self.options.search_paths = [os.path.abspath(path) for path in
+                                     self.options.robot_paths + self.options.search_paths]
```
boegel: I would drop the `robot_paths` here, and just do `abspath` on the entries in `search_paths` (if any)

easybuild/tools/options.py
```diff
@@ -265,6 +268,9 @@ def basic_options(self):
                       'pathlist', 'store_or_None', [], 'r', {'metavar': 'PATH[%sPATH]' % os.pathsep}),
             'robot-paths': ("Additional paths to consider by robot for easyconfigs (--robot paths get priority)",
                             'pathlist', 'add_flex', self.default_robot_paths, {'metavar': 'PATH[%sPATH]' % os.pathsep}),
+            'search-paths': ("Additional paths to consider in search for easyconfigs (--robot and --robot-paths paths "
+                             "get priority)",
+                            'pathlist', 'add_flex', self.default_search_paths, {'metavar': 'PATH[%sPATH]' % os.pathsep}),
```
boegel: I would use an empty list of default value

easybuild/tools/options.py
```diff
@@ -265,6 +268,9 @@ def basic_options(self):
                       'pathlist', 'store_or_None', [], 'r', {'metavar': 'PATH[%sPATH]' % os.pathsep}),
             'robot-paths': ("Additional paths to consider by robot for easyconfigs (--robot paths get priority)",
                             'pathlist', 'add_flex', self.default_robot_paths, {'metavar': 'PATH[%sPATH]' % os.pathsep}),
+            'search-paths': ("Additional paths to consider in search for easyconfigs (--robot and --robot-paths paths "
+                             "get priority)",
```
boegel: `Additional locations to consider in --search (next to --robot and --robot-paths paths)`

easybuild/tools/options.py
ocaisa: You may have guess that I was blindly following the behaviour of `robot_paths`

ocaisa:
I'm missing something it is still not picking the path from the command line

easybuild/tools/options.py
```diff
@@ -829,6 +831,10 @@ def _postprocess_config(self):
             self.options.robot_paths = [os.path.abspath(path) for path in self.options.robot + self.options.robot_paths]
             self.options.robot = self.options.robot_paths
 
+        # Update the search_paths (if any) to absolute paths
+        if self.options.search_paths is not None:
+            self.options.search_paths = [os.path.abspath(path) for path in self.options.search_paths]
+
```
ocaisa: Everything is fine at this point

easybuild/tools/robot.py
```diff
@@ -359,6 +359,9 @@ def search_easyconfigs(query, short=False, filename_only=False, terse=False):
     search_path = build_option('robot_path')
     if not search_path:
         search_path = [os.getcwd()]
+    extra_search_paths = build_option('search_paths')
```
ocaisa: ...but this value is empy.
Where do I do the update on the build option?

boegel: If `--search-paths` is specified, and `search_paths` is registered as a build option, you should be getting the corresponding list of paths here.

Are you saying it's always empty, regardless of the value of `--search-paths`?

test/framework/robot.py
```diff
@@ -595,6 +595,31 @@ def test_det_easyconfig_paths(self):
...
+        shutil.copy2(os.path.join(test_ecs_path, 't', 'toy', test_ec), self.test_prefix)
+        self.assertFalse(os.path.exists(test_ec))
+
+        args = [
+            '--search-paths=%s' % self.test_prefix,  # add to search path
+            '--tmpdir=%s' % self.test_prefix,
+            '--search',
+            'toy',
+        ]
+        outtxt = self.eb_main(args, logfile=dummylogfn, raise_error=True) # How do I get the regular output into outtxt?
```
ocaisa: @boegel The output is correct but it is not being captured by outtxt, only the logging output seems to go there, the search results get printed to the screen

ocaisa:
@easybuilders/easybuild-framework-maintainers This is ready for review

easybuild/tools/robot.py
ocaisa: Resolved in https://github.com/easybuilders/easybuild-framework/pull/2255/commits/58262d770d8c3ce0887f128e3316688613e1d203

boegel:
lgtm, thanks @ocaisa!

PR #2259, opened 2017-07-06
fix checking for new easyconfigs in copy_easyconfigs, pick up specified commit message as PR title if none was provided

boegel:


boegel:
Some broken tests went unnoticed because not all the tests for the GitHub integration functionality are run for pull requests, due to the lack of an available GitHub token (it's not provided for PRs to avoid it gets leaked via a rogue PR).

easybuild/tools/github.py
```diff
@@ -888,6 +888,8 @@ def new_pr(paths, ecs, title=None, descr=None, commit_msg=None):
                 main_title = ', '.join(names_and_versions[:3] + ['...'])
 
             title = "{%s}[%s] %s" % (class_label, toolchain_label, main_title)
+        elif commit_msg:
```
damianam: Are `file_info['ecs'] and all(file_info['new']) and not deleted_paths` and `commit_msg` mutually exclusive? I guess they aren't, but if they are, I'd suggest to swap the order of the ifs, since evaluating `file_info['ecs'] and all(file_info['new']) and not deleted_paths` is way more expensive than `commit_msg`. Nitpicking, I know.

boegel: Actually, you may have a good point there, although efficiency isn't a real concern here.

If a custom commit message is specified, that should determine the PR title, rather than an auto-guessed title.

I'll flip them around, thanks for the suggestion!

boegel:
@damianam all good now?

test/framework/easyconfig.py
```diff
@@ -1833,8 +1840,19 @@ def test_copy_easyconfigs(self):
         write_file(toy_ec, toy_ec_txt)
 
         # verify whether copied easyconfig gets cleaned up
```
damianam: Is this comment accurate? I don't think so

test/framework/easyconfig.py
```diff
@@ -1833,8 +1840,19 @@ def test_copy_easyconfigs(self):
...
-        copy_easyconfigs([toy_ec], target_dir)
+        res = copy_easyconfigs([toy_ec], target_dir)
+
+        self.assertEqual([len(x) for x in res.values()], [1, 1, 1])
+
+        self.assertTrue(isinstance(res['ecs'][0], EasyConfig))
```
damianam: This is checked for in line 1815 already. Why do the same kind of check again?

test/framework/easyconfig.py
boegel: it sort of is, but the actual check is deeper down...

test/framework/easyconfig.py
boegel: ok, sure, I'll drop it

damianam:
lgtm now. Going in after travis is done.

Issue #2260, opened 2017-07-06
--new-pr issue, need to ignore os dependencies 


shahzebsiddiqui:
[siddis14@amrndhl1315 Perl]$ eb --new-pr Perl-5.22.1-foss-2016b.eb 
== temporary log file in case of crash /dev/shm/tmp/eb-C65okM/easybuild-WaHBf3.log
ERROR: Failed to process easyconfig /home/siddis14/easybuild-easyconfigs/easybuild/easyconfigs/p/Perl/Perl-5.22.1-foss-2016b.eb: Failed to process easyconfig /home/siddis14/easybuild-easyconfigs/easybuild/easyconfigs/f/foss/foss-2016b.eb: Failed to process easyconfig /home/siddis14/easybuild-easyconfigs/easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.10.3-GCC-5.4.0-2.26.eb: One or more OS dependencies were not found: [('libibverbs-dev', 'libibverbs-devel')]

boegel:
@shahzebsiddiqui Please check whether the change in https://github.com/easybuilders/easybuild-framework/pull/2262/files fixes this for you?

PR #2262, opened 2017-07-06
automatically enable --ignore-osdeps under --new-pr and --update-pr

boegel:
fix for #2260

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -360,9 +360,10 @@ def init_build_options(build_options=None, cmdline_options=None):
 
         auto_ignore_osdeps_options = [cmdline_options.check_conflicts, cmdline_options.dep_graph,
                                       cmdline_options.dry_run, cmdline_options.dry_run_short,
-                                      cmdline_options.extended_dry_run, cmdline_options.dump_env_script]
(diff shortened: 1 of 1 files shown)


boegel:
@easybuilders/easybuild-framework-maintainers fix is confirmed in #2260, please review & merge?

easybuild/tools/config.py
```diff
@@ -360,9 +360,10 @@ def init_build_options(build_options=None, cmdline_options=None):
 
         auto_ignore_osdeps_options = [cmdline_options.check_conflicts, cmdline_options.dep_graph,
                                       cmdline_options.dry_run, cmdline_options.dry_run_short,
-                                      cmdline_options.extended_dry_run, cmdline_options.dump_env_script]
+                                      cmdline_options.extended_dry_run, cmdline_options.dump_env_script,
+                                      cmdline_options.new_pr, cmdline_options.update_pr]
```
damianam: Doesn't it make sense to add `merge_pr` too?

boegel: There's no point, since `--merge-pr` doesn't involve any parsing of easyconfig files.

damianam:
lgtm, going in

PR #2266, opened 2017-07-07
add support for --merge-pr

boegel:
support for `eb --merge-pr`, meant to help EasyBuild maintainers merging PRs (easyconfigs only for now)

checks the requirements documented at http://easybuild.readthedocs.io/en/latest/Contributing.html#requirements-for-pull-requests as best as possible, see implementation of the `check_pr_eligible_to_merge` function

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -225,7 +225,7 @@ def print_msg(msg, log=None, silent=False, prefix=True, newline=True, stderr=Fal
     :param silent: be silent (only log, don't print)
     :param prefix: include message prefix characters ('== ')
     :param newline: end message with newline
-    :param stderr: print to stderr rather than stderr
+    :param stderr: print to stderr rather than stdout
     """
     if log:
         log.info(msg)
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -217,9 +217,15 @@ def get_log(name=None):
     log.nosupport("Use of get_log function", '2.0')
 
 
-def print_msg(msg, log=None, silent=False, prefix=True, newline=True):
+def print_msg(msg, log=None, silent=False, prefix=True, newline=True, stderr=False):
     """
-    Print a message to stdout.
+    Print a message.
+
+    :param log: logger instance to also message to
+    :param silent: be silent (only log, don't print)
+    :param prefix: include message prefix characters ('== ')
+    :param newline: end message with newline
+    :param stderr: print to stderr rather than stderr
     """
     if log:
         log.info(msg)
@@ -228,9 +234,12 @@ def print_msg(msg, log=None, silent=False, prefix=True, newline=True):
             msg = ' '.join([EB_MSG_PREFIX, msg])
 
         if newline:
-            print msg
+            msg += '\n'
+
+        if stderr:
+            sys.stderr.write(msg)
         else:
-            print msg,
+            sys.stdout.write(msg)
 
 
 def dry_run_set_dirs(prefix, builddir, software_installdir, module_installdir):
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -233,7 +233,7 @@ def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
     try:
         status, data = url.get(**kwargs)
     except socket.gaierror, err:
-        _log.warning("Error occured while performing get request: %s", err)
+        _log.warning("Error occurred while performing get request: %s", err)
         status, data = 0, None
 
     _log.debug("get request result for %s: status: %d, data: %s", url, status, data)
@@ -259,7 +259,7 @@ def github_api_put_request(request_f, github_user=None, token=None, **kwargs):
     try:
         status, data = url.put(**kwargs)
     except socket.gaierror, err:
-        _log.warning("Error occured while performing put request: %s", err)
+        _log.warning("Error occurred while performing put request: %s", err)
         status, data = 0, {'message': err}
 
     if status == 200:
@@ -1260,8 +1260,8 @@ def check_github():
 
     if res:
         if res[0].flags & res[0].ERROR:
-            _log.warning("Error occured when pushing test branch to GitHub: %s", res[0].summary)
-            check_res = "FAIL (error occured)"
+            _log.warning("Error occurred when pushing test branch to GitHub: %s", res[0].summary)
+            check_res = "FAIL (error occurred)"
         else:
             check_res = "OK"
     elif github_user:
@@ -1295,7 +1295,7 @@ def check_github():
     try:
         res = create_gist("This is just a test", 'test.txt', descr='test123', github_user=github_user)
     except Exception as err:
-        _log.warning("Exception occured when trying to create gist: %s", err)
+        _log.warning("Exception occurred when trying to create gist: %s", err)
 
     if res and re.match('https://gist.github.com/[0-9a-f]+$', res):
         check_res = "OK"
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -915,25 +915,26 @@ def check_pr_eligible_to_merge(pr_data):
     else:
         res = not_eligible(msg_tmpl % "(result unknown)")
 
-    # check for successful test report (checked in reverse order)
-    msg_tmpl = "* last test report is successful: %s"
-    test_report_regex = re.compile(r"^Test report by @\S+")
-    test_report_found = False
-    for comment in pr_data['issue_comments'][::-1]:
-        comment = comment['body']
-        if test_report_regex.search(comment):
-            if 'SUCCESS' in comment:
-                print_msg(msg_tmpl % 'OK', prefix=False)
-                test_report_found = True
-                break
-            elif 'FAILED' in comment:
-                res = not_eligible(msg_tmpl % 'FAILED')
-                test_report_found = True
-            else:
-                print_warning("Failed to determine outcome of test report for comment:\n%s" % comment)
+    if pr_data['base']['repo']['name'] == GITHUB_EASYCONFIGS_REPO:
+        # check for successful test report (checked in reverse order)
+        msg_tmpl = "* last test report is successful: %s"
+        test_report_regex = re.compile(r"^Test report by @\S+")
+        test_report_found = False
+        for comment in pr_data['issue_comments'][::-1]:
+            comment = comment['body']
+            if test_report_regex.search(comment):
+                if 'SUCCESS' in comment:
+                    print_msg(msg_tmpl % 'OK', prefix=False)
+                    test_report_found = True
+                    break
+                elif 'FAILED' in comment:
+                    res = not_eligible(msg_tmpl % 'FAILED')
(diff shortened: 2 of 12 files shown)


easybuild/tools/build_log.py
```diff
@@ -217,9 +217,15 @@ def get_log(name=None):
...
+def print_msg(msg, log=None, silent=False, prefix=True, newline=True, stderr=False):
     """
-    Print a message to stdout.
+    Print a message.
+
+    :param log: logger instance to also message to
+    :param silent: be silent (only log, don't print)
+    :param prefix: include message prefix characters ('== ')
+    :param newline: end message with newline
+    :param stderr: print to stderr rather than stderr
```
verdurin: Is this a typo?

verdurin: Should be 'print to stdout rather than stderr'?

easybuild/tools/github.py
```diff
@@ -233,10 +233,43 @@ def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
     try:
         status, data = url.get(**kwargs)
     except socket.gaierror, err:
-        _log.warning("Error occured while performing get request: %s" % err)
+        _log.warning("Error occured while performing get request: %s", err)
```
verdurin: occurred

easybuild/tools/github.py
```diff
@@ -233,10 +233,43 @@ def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
...
+
+    if token is None:
+        token = fetch_github_token(github_user)
+
+    url = request_f(RestClient(GITHUB_API_URL, username=github_user, token=token))
+
+    try:
+        status, data = url.put(**kwargs)
+    except socket.gaierror, err:
+        _log.warning("Error occured while performing put request: %s", err)
```
verdurin: occurred

easybuild/tools/github.py
```diff
@@ -835,6 +877,136 @@ def is_patch_for(patch_name, ec):
...
+            found_lgtm = True
+
+    if not test_report_found:
+        res = not_eligible(msg_tmpl % "(no test reports found)")
+
+    # check for style review by a human
+    found_lgtm = False
+    for comment in pr_data['issue_comments']['bodies']:
+        if 'lgtm' in comment:
+            found_lgtm = True
```
verdurin: According to this all maintainers *have* to use the formulation 'lgtm' then, for this test to work? Worth mentioning that in the contributors' guide, if it isn't there already.

easybuild/tools/build_log.py
boegel: The other way around `print to stderr rather than stdout`

easybuild/tools/github.py
boegel: I want to relax this a bit and also make this check pass when an approved review is submitted, especially since that's required now to merge (via the web interface).

boegel:
@verdurin Please re-review, and maybe also give it a try yourself?

Example usage (I'll write up documentation too) in dry run mode (only difference is that submitted the comment and actually merging are not done if PR is eligible for merging):

```
$ eb --merge-pr 4827 -x
== temporary log file in case of crash /tmp/eb-Gz20kc/easybuild-G3wn6E.log

easybuilders/easybuild-easyconfigs PR #4827 was submitted by nathanhaigh, you are using GitHub account 'boegel'

Checking eligibility of easybuilders/easybuild-easyconfigs PR #4827 for merging...
* targets develop branch: OK
* test suite passes: OK
* last test report is successful: (no test reports found) => not eligible for merging!
* approved review: MISSING => not eligible for merging!
* milestone is set: no milestone found => not eligible for merging!

WARNING: Review indicates this PR should not be merged (use -f/--force to do so anyway)
```

boegel:
documentation update @ https://github.com/easybuilders/easybuild/pull/356

easybuild/tools/github.py
```diff
@@ -835,6 +877,147 @@ def is_patch_for(patch_name, ec):
...
+    msg = "\n%s/%s PR #%s was submitted by %s, " % (pr_target_account, pr_target_repo, pr, pr_data['user']['login'])
+    msg += "you are using GitHub account '%s'\n" % github_user
+    print_msg(msg, prefix=False)
+    if pr_data['user']['login'] == github_user:
+        raise EasyBuildError("Please do not merge your own PRs!")
+
+    # also fetch status of last commit
+    pr_head_sha = pr_data['head']['sha']
+    status_url = lambda g: g.repos[pr_target_account][pr_target_repo].commits[pr_head_sha].status
+    status, status_data = github_api_get_request(status_url, github_user)
```
damianam: If `status` is not used, maybe it makes sense to replace it with `_` to make it clear? Even though I'd suggest to check the status here too, as individual consecutive requests might return different status.

easybuild/tools/github.py
```diff
@@ -835,6 +877,147 @@ def is_patch_for(patch_name, ec):
...
+    pr_data['status_last_commit'] = status_data['state']
+
+    # also fetch comments
+    comments_url = lambda g: g.repos[pr_target_account][pr_target_repo].issues[pr].comments
+    status, comments_data = github_api_get_request(comments_url, github_user)
```
damianam: Same

easybuild/tools/github.py
```diff
@@ -835,6 +877,147 @@ def is_patch_for(patch_name, ec):
...
+    pr_data['issue_comments'] = comments_data
+
+    # also fetch reviews
+    reviews_url = lambda g: g.repos[pr_target_account][pr_target_repo].pulls[pr].reviews
+    status, reviews_data = github_api_get_request(reviews_url, github_user)
```
damianam: Same

easybuild/tools/github.py
```diff
@@ -835,6 +877,147 @@ def is_patch_for(patch_name, ec):
...
+    pr_data['reviews'] = reviews_data
+
+    force = build_option('force')
```
damianam: I wouldn't give this possibility. If it is not elegible for merging it needs more careful evaluation. Slapping `--force` is an easy way out for potentially lazy maintainers.

easybuild/tools/github.py
```diff
@@ -835,6 +877,147 @@ def is_patch_for(patch_name, ec):
     return soft_name
 
 
+def check_pr_eligible_to_merge(pr_data):
+    """
+    Check whether PR is eligible for merging.
+
```
boegel: We should stick to one style, and I generally prefer having a blank line there as well, but the only way to make this consistent is to have an automated style check for it...

easybuild/tools/github.py
```diff
@@ -835,6 +877,147 @@ def is_patch_for(patch_name, ec):
...
+                test_report_found = True
+                break
+            elif 'FAILED' in comment:
+                res = not_eligible(msg_tmpl % 'FAILED')
+                test_report_found = True
+            else:
+                print_warning("Failed to determine outcome of test report for comment:\n%s" % comment)
+
+        if 'lgtm' in comment:
+            found_lgtm = True
```
boegel: This didn't get cleaned up, indeed, thanks.

easybuild/tools/github.py
boegel: I disagree here, in rare cases it makes sense to merge a PR that doesn't strictly meet all requirements.

For example, when only trivial changes are made that do not add new easyconfigs and do not alter the semantics of existing easyconfigs, we can merge it without requiring a test report. One recent example is https://github.com/easybuilders/easybuild-easyconfigs/pull/4815...

I would rather have maintainers use `--merge-pr X --force` after running it without `--force` first, than favouring manual merges where it's easier to overlook one of the other requirements. Of course, people shouldn't be using `--force` all the time...

easybuild/tools/github.py
boegel: fixed by checking `status`, thanks for bringing this up

easybuild/tools/github.py
damianam: That's a good point and a good example. I was thinking more about new additions or version changes, than that kind of small tuning. I guess it makes sense then. But I wonder if a middle ground is possible. In other words, check what is the nature of the changes, and reject to merge, even with `--force`, if new files are added, or there are changes in `version`, `toolchain`, `configopts`, etc, but allow it if the changes are just comments, or `description`, `whatis` and maybe one or two more parameters. That's probably a PITA to implement and possibly not worth the effort though, as long as all behave responsibly ;-)

boegel: That would be nice, but that's definitely tricky to implement, and I'm not sure it'll be worth the trouble... 
We can always look into that enhancement later though, if we find it's really necessary, but I don't expect that maintainers will go and 'abuse' `--merge-pr X --force`, to avoid tarring & feathering by the other maintainers. ;-)

damianam:
I haven't have time to take a look at the tests, but otherwise lgtm. I'll try to take a look at the tests tomorrow. Anyway, maybe  @vanzod or @migueldiascosta can give their feedback too before merging?

easybuild/tools/github.py
vanzod: I agree on checking against approved review. Cleaner and less prone to personal quirks.

easybuild/tools/github.py
```diff
@@ -835,6 +880,153 @@ def is_patch_for(patch_name, ec):
...
+    """
+    Merge specified pull request
+    """
+    github_user = build_option('github_user')
+    if github_user is None:
+        raise EasyBuildError("GitHub user must be specified to use --merge-pr")
+
+    pr_target_account = build_option('pr_target_account')
+    pr_target_repo = build_option('pr_target_repo')
+
```
migueldiascosta: make sure `pr_target_repo == GITHUB_EASYCONFIGS_REPO`?

boegel: Well, we could be a bit smarter, and skip the check for a successful test report for other repositories, since the other checks are just as important for framework & easyblocks?

damianam: I'd even argue that for easyblocks a successful test report might be just as important as for the easyconfigs repo, given that both decide how the software is built. But we would have to make sure that the easyblock is used in the easyconfigs being built. In any case, I thought that at this moment the github integration worked just with the easyconfigs repo, so I wouldn't delay this PR because of this.

boegel: We don't have a way to submit test reports for easyblock PRs yet.

It's true that currently most GitHub-related options only work with the easyconfigs repo for now, but the intention is to extend that to also easyblocks/framework, certainly for `--new-pr` and `--update-pr`.

The extra effort here to make `--merge-pr` also work well for the easyblocks & framework repos is minimal, see 17789d55.

verdurin:
Looks good, just wanted to reiterate my earlier comment that if we're expecting 'lgtm' and 'Going in, thanks...' for each PR, then that should be documented.

damianam:
Seems like it lgte (looks good to everyone ;-)). Going in!

boegel:
@verdurin the `lgtm` is no longer needed, `--merge-pr` will only look for an approved review (which basically is a `lgtm` statement)

the `Going in, thanks` is just being friendly, but it's not a strict requirement ;)

PR #2267, opened 2017-07-09
notify EasyBuild maintainers of failing tests

boegel:


boegel:
Hmm, this also results in mails to the maintainers list for new branches that are created, which is definitely what we want... We'll need to figure out a better way.

PR #2269, opened 2017-07-10
Lower case name scheme

victorusu:
We got a request from one of our customers to have lower case name
schemes using the TMod.

This is how we achieved it, so we would like to contribute back.

### DIFF
--- a/easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
+++ b/easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
@@ -1,0 +1,62 @@
+##
+# Copyright 2013-2017 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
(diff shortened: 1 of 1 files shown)


boegel:
(close/open to retrigger tests after changing target branch to `develop`)

easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
```diff
@@ -0,0 +1,62 @@
...
+
+    REQUIRED_KEYS = ['name', 'version', 'versionsuffix', 'toolchain']
+
+    def det_full_module_name(self, ec):
+        """
+        Determine full module name from given easyconfig, according to the EasyBuild module naming scheme.
+        :param ec: dict-like object with easyconfig parameter values (e.g. 'name', 'version', etc.)
+        :return: string with full module name <name>/<installversion>, e.g.: 'gzip/1.5-goolf-1.4.10'
+        """
+        return os.path.join(ec['name'], det_full_ec_version(ec)).lower()
```
boegel: @victorusu Since you're basically creating a lower-case version of `EasyBuildMNS` here, wouldn't it make more sense to derive from the `EasyBuildMNS` class, change the name to `LowercaseEasyBuildMNS`, and change this to:

```python
return super(LowercaseEasyBuildMNS, self). det_full_module_name(ec).lower()
```

(and then of course analogous in `is_short_modname_for`)?

boegel:
@victorusu A dedicated test for this module naming scheme should be added, probably to `test/framework/module_generator.py` since it already has similar tests.

Issue #2127, opened 2017-02-20
Yet another lmod integration feature (userInGroup)


damianam:
In lmod it is possible to use the `userInGroup` function to restrict the loading of modules to certain user groups:

```
if (userInGroup("namd_group")) then
   something()
else
   LmodError(err_message,"\n")
end
```

I think this would be a nice thing to integrate together with the `group` key in easyconfigs.

I'll see if I can get to it soon[ish]

mboisson:
Note that the error message and the name of the group must be site-customizable, i.e. in the easyconfig. 

Also, for LmodError, if we want to use localization, we'ld have to wrap the error message to get the right language.

damianam:
The name of the group is already customizable.

Regarding the error message, I am not superconvinced about being it site customizable through another easyconfig parameter. As I see it there are three ways:

1. Set a standard error message in Lmod, that is localizable and that can be leveraged from EB.
2. Generate from EB a generic error message ("You are not part of the `group` group. Please consult with user support how to become a member of that group.").
3. Generate from EB a customizable error message.

1 has to be tackled in Lmod which makes it Lmod version dependent, which means we shouldn't probably consider it. 2 is the easiest. 3 will need a fallback if no customized error is provided. I am afraid that at the end nobody will use the custom message, since the fallback would be good enough for almost everyone. I am hesitant to spend time implementing a feature that I don't need and maybe nobody else will need.

boegel:
I agree with @damianam here, 2 is exactly what I had in mind. 1 is fine too, depending on a particular version of Lmod for an optional feature isn't a huge problem imho.

I'm not planning to spend time on adding support on localisation in EasyBuild anytime soon, but if there are pull requests for it I'll certainly try and take a look at them if time allows for it.

mboisson:
At the very least, if I were to enable this feature for Compute Canada, I would say something like "Please write to us at <email>@computecanada.ca with your NAMD username to become a member of that group", rather than "Please consult with user support...". I would probably make the message bilingual as well.

boegel:
We don't have anything like that in EasyBuild yet, i.e. allow for customising EasyBuild error messages or statements that are included in module files. If we support this, it should be done as a part of a bigger effort, not as something for one specific feature...

mboisson:
That's fine. I'll probably disable this feature and have it module-specific through the module footer in the EB recipe. Would there be a good place to document this alternative solution ?

boegel:
[...]

PR #2274, opened 2017-07-12
include userInGroup check in Lua modules when installation is group-restricted

damianam:
This enables the use of `userInGroup` function in lua modules (see https://github.com/easybuilders/easybuild-framework/issues/2127).

The test is missing because it is not straight forward to implement. We need to create a module with an associated group, but when loading it it needs to fail (so the user can't belong to the group). If the user can't belong to the group, he/she can't install it using that group. Moreover, the group has to exist. This makes it very difficult to implement a reliable test.

Any idea to implement a meaningful test? Should this feature be optional (which would require adding yet another option to enable/disable it)?


damianam:
As commented in our last telco (https://github.com/easybuilders/easybuild/wiki/Conference-call-notes-20170719), we could extend the `group` easyconfig parameter, and make it (optionally) a tuple, with a load error message, allowing customization that way.

Thoughts?

damianam:
@boegel currently we can't check properly that the module is correctly generated, since for it we need to have the proper OS setup (user running the tests has to belong the a group used for these tests, and that obviously changes with every computer). What would you suggest?

easybuild/framework/easyblock.py
```diff
@@ -227,9 +227,15 @@ def __init__(self, ec):
...
-        if self.cfg['group'] is not None:
-            self.log.warning("Group spec '%s' is overriding config group '%s'." % (self.cfg['group'], group_name))
-            group_name = self.cfg['group']
+        group_spec = self.cfg['group']
+        if group_spec is not None:
+            if isinstance(group_spec, tuple):
+                if len(group_spec) == 2:
+                    group_spec = group_spec[0]
+                else:
+                    raise EasyBuildError("Found group spec in tuple format that is not a 2-tuple: %s", ec_group)
```
boegel: this is not correct:

* uses non-existing variable `ec_group`
* string templating with a tuple value (`group_spec`) doesn't work for a single `%s`, should be `str(group_spec)`

fixed in https://github.com/damianam/easybuild-framework/pull/11, which includes an enhanced test

boegel:
Going in, thanks @damianam!

mboisson:
Can we make this optional through a config item ?

boegel:
@mboisson Sure we can. Any particular reason you don't like to have this enabled by default? And are you up for looking into the PR to make this optional yourself? ;)

mboisson:
We use other means of managing user group checks, and the code added by this PR is broken on our system.

mboisson:
Specifically, we do want easybuild to set the correct group-owner for the files, but we do not want to have the check in the module.

boegel:
@bartoldeman Thanks, but, your idea is not going to work out well, since if `None` is specified as error message, EasyBuild falls back to a standard error message for users not in the group, see https://github.com/easybuilders/easybuild-framework/pull/2274/files#diff-8cdd2ad54096882655bd836265e8e26bR677.

A configuration option like `--add-user-in-group-check` (that is enabled by default, so you could use `--disable-add-user-in-group-check`) seems like a better option?

damianam:
> We use other means of managing user group checks, and the code added by this PR is broken on our system.

Broken how exactly? We definitely don't want to have broken code anywhere, if it breaks in your system I guess it is a bug on the current implementation.

> Specifically, we do want easybuild to set the correct group-owner for the files, but we do not want to have the check in the module.

Can I ask why? I have no problem with adding options to remove this check (even though I can't spend time on it until November the earliest), but I wonder what is the rationale behind. Understanding the issue might help to have a better solution.

mboisson:
Broken how: For some reason, on our system, the "userInGroup" function from Lmod fails to resolve the group name, and barfs on numeric IDs. We have a "localUserInGroup" function which works. It does not

```
local function localUserInGroup(group)
        local handle = io.popen("groups 2>/dev/null")
        local grps = handle:read()
        handle:close()
        local found  = false
        for g in string.gmatch(grps, '([^ ]+)') do
                if (g == group)  then
                        found = true
                        break
                end
        end
        return found
end``` 


In comparison, Lmod's userInGroup does: 
``` 
function userInGroup(group)
   local grps   = capture("groups")
   local found  = false
   local userId = capture("id -u")
   local isRoot = tonumber(userId) == 0
   for g in grps:split("[ \n]") do
      if (g == group or isRoot)  then
         found = true
         break
      end
   end
   return found
end
``` 

We also tend to centralize any such check in the SitePackage.lua, through a hook. This prevents having one version of a software which has the check, and a second version which does not.

**Comment by bartoldeman** [comment]
Not a PR yet but at least I solved the mystery on why this happened:
according to http://lmod.readthedocs.io/en/latest/050_lua_modulefiles.html 

> capture (“string”):
> Run the “string” as a command and capture the output. This function uses the value of LD_PRELOAD and LD_LIBRARY_PATH found when Lmod is configured. Use subprocess if you wish to use the current values.

A problem with our Nix setup is that we rely on LD_LIBRARY_PATH to find libnss_sss.so.2 to authenticate successfully. Since the LD_LIBRARY_PATH was not set when Lmod was configured, the  capture command did not get the group names. I have now fixed our Lmod to avoid this, but in any case Maxime's comments still hold (the message overrides our logic in SitePackage.lua which is as follows):
```
                                if (not localUserInGroup(groupT[name])) then
                                        log_module_load(t,false)
                                        local message_found = false
                                        for k2,v2 in pairs(posix_group_messageT) do
                                                if (has_value(k2,name)) then
                                                        LmodError(v2)
                                                        message_found = true
                                                end
                                        end
                                        if (not message_found) then
                                                LmodError(posix_group_message)
                                        end
                                end
```
if the module errors out on its own, `log_module_load(t,false)` fails and the LmodError in the above chunk is never invoked.

**Comment by mboisson** [comment]
How about
```
group = ("supergroup", "")
``` 
(rather than the above suggestion of "None") can do what we want ?

PR #2275, opened 2017-07-12
remove 'provides' line from setup.py

boegel:
I had to remove the `provides` line from `setup.py` in order to be able to release EasyBuild v3.3.1 with a recent version of `setuptools`, to work around this problem:

```
Upload failed (400): provides: Must be a valid Python identifier.
error: Upload failed (400): provides: Must be a valid Python identifier.
```

As far as I can tell, no information is lost by dropping this line.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -116,7 +116,6 @@ implement support for installing particular (groups of) software packages.""",
         "Topic :: Software Development :: Build Tools",
     ],
     platforms="Linux",
-    provides=["eb"] + easybuild_packages,
     test_suite="test.framework.suite",
     zip_safe=False,
     install_requires=[


damianam:
@boegel I've been doing some tests to find out why that happens. Turns out that `pypi` became more picky, and now it actually does the right[?] thing and just accepts valid python identifiers passed in the `provides` list.  So all the strings that include a `.` are not correct (`easybuild.framework`, `easybuild.framework.easyconfig`, etc), and therefore are rejected.

I think there was a misuse of `provides` until you remove it, but it doesn't necessarily mean we have to get rid of it. Just get rid of the faulty bits (`easybuild_packages`). After all, these packages are already listed in `packages`, where they belong.

On a related note, why not use `find_packages()` instead of listing them manually?

Also, I think this issue[?] lies purely in `pypi`, as it happens once the upload is completed. I don't think the `setuptools` version is relevant.

boegel:
@damianam So, what should be in the list? Just `easybuild`, or both `eb` and `easybuild`?

Also, what's the value of having the `provides` line? Any fallout from removing it?

The 3.3.1 release was done without this line, see https://pypi.python.org/pypi/easybuild-framework/3.3.1

damianam:
@boegel as per https://docs.python.org/3/distutils/setupscript.html#relationships-between-distributions-and-packages:

> [...]we also need to be able to specify what we provide that other distributions can require. This is done using the provides keyword argument to setup(). The value for this keyword is a list of strings, each of which names a Python module or package, and optionally identifies the version. If the version is not specified, it is assumed to match that of the distribution.

So it looks to me like we should simply include `easybuild`. Also, in practice, it seems that this is rarely used, that's why it didn't do any harm to remove it. But IMO it doesn't hurt to have the correct content in place either

boegel:
@damianam I'm OK with specifying something else rather than removing it, but I'm not sure `'easybuild'` makes sense... Maybe it does here (although it's up for debate imho), but then what do we specify for the `easybuild-easyblocks` package (cfr. https://github.com/easybuilders/easybuild-easyblocks/pull/1217).

Since the added value of using `provides` is questionable, I'd rather just drop it and be done with it...

damianam:
I'd remove it in the `easybuild-*` packages, since those are incomplete by themselves and thus it can't be considered that they provide `easybuild`. But for uploading the whole thing to pypi I still think it makes sense to keep it. Of course, if it complicates things, just drop it, since it has no real meaning as of today.

boegel:
So, only use `provides` in the `setup.py` in https://github.com/easybuilders/easybuild?

Separate EasyBuild packages are uploaded to PyPI for each repository, see e.g. https://pypi.python.org/pypi/easybuild-framework, etc.

So, this PR is good to go then?

damianam:
Yes, good to go. Finally, eh? ;-)

JensTimmerman:
`provides` is actually used by bdist_rpm to put this line into the rpm 'provides', so dependencies that needed easybuild would know this rpm provided this dependency. Once needed because a naming scheme changed I think (when we split easybuild in 3 repo's long long ago)
So any new rpm's that need EasyBuild should long be using the correct name as a dependency and this provides should not be needed anymore. 
So as a heads up, if someone is running into rpm issues, this could be the culprit.

boegel:
Hah, ok, thanks for clarifying @JensTimmerman!

PR #2276, opened 2017-07-13
pass down stdin in 'import' check for extensions during sanity check

boegel:
By not passing down `stdin` into `run_cmd` in the `sanity_check` implementation for extensions, the 'import' check for R libraries is basically being skipped...

In addition, the caching mechanism for `run_cmd` should also take into account the `inp` argument.

### DIFF
--- a/test/framework/run.py
+++ b/test/framework/run.py
@@ -132,6 +132,17 @@ class RunTest(EnhancedTestCase):
         self.assertEqual(ec, 123)
         self.assertEqual(cached_out, "123456")
 
+        # also test with command that uses stdin
+        (out, ec) = run_cmd("cat", inp='foo')
+        self.assertEqual(ec, 0)
+        self.assertEqual(out, 'foo')
+
+        # inject different output for cat with 'foo' as stdin to check whether cached value is used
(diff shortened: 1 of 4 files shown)


boegel:
This probably needs enhanced tests to verify the changes (and make sure the bug being fixed here isn't re-introduced).

boegel:
Without the changes to the Seurat easyconfig in https://github.com/easybuilders/easybuild-easyconfigs/pull/4889, the sanity check fails as it should on top of this...

boegel:
tests are enhanced to catch the bug

@easybuilders/easybuild-framework-maintainers ready for review

test/framework/run.py
```diff
@@ -118,6 +118,22 @@ def test_run_cmd_simple(self):
         self.assertEqual(True, run_cmd("echo hello", simple=True))
         self.assertEqual(False, run_cmd("exit 1", simple=True, log_all=False, log_ok=False))
 
+    def test_run_cmd_cache(self):
```
ocaisa: What about a test for the new `inp` argument to `run_cmd`?

boegel: good point, added in #bc4be4154

Issue #2279, opened 2017-08-01
bootstrap install error for RHEL6


steevmi1:
I'm trying to install using the bootstrap method on a RHEL6 system with the stock python (2.6.6), and get the following error:
```
[[INFO]] +++ STAGE 2: installing EasyBuild in /opt/software/RHEL6-x86_64/easybuild with EasyBuild from stage 1...

Traceback (most recent call last):
  File "bootstrap_eb.py", line 1014, in <module>
    main()
  File "bootstrap_eb.py", line 817, in main
    stage2(tmpdir, templates, install_path, distribute_egg_dir, sourcepath)
  File "bootstrap_eb.py", line 687, in stage2
    easybuild_main()
  File "/tmp/tmpDDGLJh/eb_stage1/lib/python2.6/site-packages/easybuild_framework-3.3.1-py2.6.egg/easybuild/main.py", line 189, in main
    eb_go = eboptions.parse_options(args=args)
  File "/tmp/tmpDDGLJh/eb_stage1/lib/python2.6/site-packages/easybuild_framework-3.3.1-py2.6.egg/easybuild/tools/options.py", line 1061, in parse_options
    raise EasyBuildError("Failed to parse configuration options: %s" % err)
easybuild.tools.build_log.EasyBuildError: "Failed to parse configuration options: 'NoneType' object has no attribute 'startswith'"

-bash-4.1$ python -V
Python 2.6.6
```

boegel:
@steevmi1 Hmm, this is a curious error, I suspect it is something in your environment that is making the bootstrap procedure fail.

Do you have any `$EASYBUILD_*` environment variables defined (check with `env | grep ^EASYBUILD`)?

Can you check whether you get some more information by defining `$EASYBUILD_BOOTSTRAP_DEBUG` (just run `export EASYBUILD_BOOTSTRAP_DEBUG=1` before bootstrapping EasyBuild)?

You may also need to define `$DEBUG_EASYBUILD_OPTIONS` to get any additional useful information, since the problem occurs very early on, i.e. when EasyBuild tries to figure out how it is configured.

OndrejMarsalek:
I have the same problem on Fedora 20. Here is the debug output and my environment:

https://gist.github.com/OndrejMarsalek/454e4866bf187f90b6400e4dc3d30cb0

OndrejMarsalek:
[...]

PR #2282, opened 2017-08-08
make resolve_path robust against None path being provided

boegel:
companion fix for #2279 (alongside updated bootstrap script in #2281)

If this had been part of #2248, bootstrapping EasyBuild v3.3.1 would not have been broken as reported in #2279

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -180,10 +180,13 @@ def resolve_path(path):
 
     :param path: path that (maybe) contains symlinks
     """
-    try:
-        resolved_path = os.path.realpath(path)
-    except OSError as err:
-        raise EasyBuildError("Resolving path %s failed: %s", path, err)
+    if path is None:
+        resolved_path = None
+    else:
+        try:
+            resolved_path = os.path.realpath(path)
+        except OSError as err:
+            raise EasyBuildError("Resolving path %s failed: %s", path, err)
 
     return resolved_path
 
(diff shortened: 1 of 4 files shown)


easybuild/tools/filetools.py
```diff
@@ -180,10 +180,13 @@ def resolve_path(path):
 
     :param path: path that (maybe) contains symlinks
     """
-    try:
-        resolved_path = os.path.realpath(path)
-    except OSError as err:
-        raise EasyBuildError("Resolving path %s failed: %s", path, err)
+    if path is None:
```
wpoely86: Shouldn't we just avoid calling this with `None`?

boegel: @wpoely86 Well, sure, that's another option, but then we're bound to make the same mistake again later?

It would boil down to doing this in `get_paths_for`:

```python
    # figure out installation prefix, e.g. distutils install path for easyconfigs
    eb_path = which('eb')
    if eb_path is None:
        _log.warning("'eb' not found in $PATH, failed to determine installation prefix")
    else:
        # real location of eb should reside in <install_prefix>/bin/eb
        eb_path = resolve_path(eb_path)
        install_prefix = os.path.dirname(os.path.dirname(eb_path))
```

whereas now we have:

```python
    # figure out installation prefix, e.g. distutils install path for easyconfigs
    eb_path = resolve_path(which('eb'))
    if eb_path is None:
        ...
```

I'm not sure which one is better, I'm happy to change it as you suggest if you feel strongly about it.

boegel:
@wpoely86 Updated, this is indeed a bit cleaner, I don't think we want to go towards making every possible function robust against passing in None...

PR #2283, opened 2017-08-09
add 'rpath' easyconfig parameter to enable use of RPATH regardless of EasyBuild configuration

boegel:
We already have an `rpath` toolchain option that can be used to selectively *disable* the use of RPATH even if EasyBuild is configured to use RPATH linking via `--rpath`.

In some cases, it may also be interesting to be able to selectively enforce the use of RPATH, regardless of how EasyBuild was configured; see for example https://github.com/easybuilders/easybuild-easyconfigs/pull/4962#pullrequestreview-54647218

If the `rpath` easyconfig parameter is defined as `True`, RPATH linking is always performed, regardless of the EasyBuild configuration option `--rpath`.
If `--rpath` is used, RPATH linking is enabled, unless it's specifically disabled by setting the `rpath` toolchain option to `False` (`True` being the default value).

After implementing this, which was pretty straightforward, I'm starting to have 2nd thoughts about this approach, since having both an `rpath` toolchain option and easyconfig parameter (on top of the `--rpath` configuration option...) may lead to confusion...

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1654,7 +1654,8 @@ class EasyBlock(object):
             self.rpath_filter_dirs.append(self.builddir)
 
         # prepare toolchain: load toolchain module and dependencies, set up build environment
-        self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs)
+        self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath=self.cfg['rpath'],
+                               rpath_filter_dirs=self.rpath_filter_dirs)
 
         # handle allowed system dependencies
         for (name, version) in self.cfg['allow_system_deps']:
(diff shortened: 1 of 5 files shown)


boegel:
An alternate approach could be:

* use `None` as default value for the `rpath` toolchain option, meaning honor the `--rpath` configuration setting
* allow disabling using `'rpath': False` in `toolchainopts` (already the case)
* allow hard enabling using `'rpath': True` in `toolchainopts` (regardless of `--rpath`)

That's probably a better approach... Thoughts @wpoely86, @pescobar, @JackPerdue, @jdonners?

akesandgren:
"Default" instead of "None" would be less confusing in my opinion but None is easier to code for.
and Enforce (or Enable) vs Disable is probably easier to understand compared to True/False, and can be turned into true/false in the actual code.

akesandgren:
@boegel Care to fix the conflicts?

boegel:
bitrotted, no longer relevant, so closing...

PR #2286, opened 2017-08-16
add support for 'eb --inject-checksums'

boegel:
* [x] collect checksums for all files/patches, incl. for extensions
* [x] construct lines for `checksums` and `exts_list` easyconfig parameter to be injected into easyconfig file
* [x] replace existing value for `checksums` under `--force`
* [x] replace existing value for `exts_list` with value that includes checksums
* [x] inject `checksums` in easyconfig file where no checksums are provided yet
* [x] ~~make checksum type required parameter~~
   * (not needed, proper error message is printed when `--inject-checksums` accidentally tries to consume an easyconfig filename thanks to `choice` option type)
* [x] add test for `--inject-checksums`
* [x] update docs: https://github.com/easybuilders/easybuild/pull/369


easybuild/tools/options.py
```diff
@@ -607,9 +607,12 @@ def easyconfig_options(self):
         # easyconfig options (to be passed to easyconfig instance)
         descr = ("Options for Easyconfigs", "Options to be passed to all Easyconfig.")
 
-        opts = None
+        opts = OrderedDict({
+            'inject-checksums': ("Inject checksums of specified type for sources/patches into easyconfig file(s)",
+                                 'choice', 'store_or_None', CHECKSUM_TYPE_SHA256, CHECKSUM_TYPES),
+        })
         self.log.debug("easyconfig_options: descr %s opts %s" % (descr, opts))
-        self.add_group_parser(opts, descr, prefix='easyconfig')
```
wpoely86: why change the prefix?

easybuild/framework/easyblock.py
```diff
@@ -2795,3 +2793,127 @@ class StopException(Exception):
...
+        # compute & inject checksums for extension sources/patches
+        if app.exts:
+            print_msg("computing & injecting %s checksums for extensions in %s..." % (checksum_type, ec_fn))
+
+            exts_list_lines = ['exts_list = [']
+            for ext in app.exts:
+
+                if ext.keys() == ['name']:
+                    exts_list_lines.append("%s'%s'," % (' ' * 4, ext['name']))
+
```
wpoely86: less empty lines?

easybuild/framework/easyblock.py
```diff
@@ -2795,3 +2793,127 @@ class StopException(Exception):
...
+            found_checksums |= bool(ext.get('checksums'))
+        if found_checksums:
+            if build_option('force'):
+                print_warning("Found existing checksums in %s, overwriting them (due to use of --force)..." % ec_fn)
+            else:
+                raise EasyBuildError("Found existing checksums, use --force to overwrite them")
+
+        # back up easyconfig file before injecting checksums
+        ectxt = read_file(ec['spec'])
+        ec_backup = os.path.join(ec['spec'] + '.bck')
```
boegel: I'm not sure there's a generally accepted standard... See https://fileinfo.com/filetypes/backup

easybuild/framework/easyblock.py
```diff
@@ -2795,3 +2793,127 @@ class StopException(Exception):
...
+            checksum = compute_checksum(entry['path'], checksum_type)
+            print_msg("* %s: %s" % (os.path.basename(entry['path']), checksum))
+            checksums.append((os.path.basename(entry['path']), checksum))
+
+        if len(checksums) == 1:
+            checksum_lines = ["checksums = ['%s']\n" % checksums[0][1]]
+        else:
+            checksum_lines = ['checksums = [']
+            for fn, checksum in checksums:
+                checksum_lines.append("    '%s',  # %s" % (checksum, fn))
```
boegel: I'll use `INDENT_4SPACES` consistently instead...

easybuild/framework/easyblock.py
```diff
@@ -2795,3 +2793,127 @@ class StopException(Exception):
...
+            checksum_lines.append(']\n')
+
+        checksums_txt = '\n'.join(checksum_lines)
+
+        if app.cfg['checksums']:
+            regex = re.compile(r'^checksums(?:.|\n)*?\]\s*$', re.M)
```
boegel: good points, fixed

easybuild/framework/easyblock.py
```diff
@@ -2795,3 +2793,127 @@ class StopException(Exception):
...
+            ectxt = regex.sub(checksums_txt, ectxt)
+        elif app.src:
```
boegel: no, it's possible there are no sources, and hence no patches/checksums

for example: bundles, like `intel-2017a.eb` easyconfig

easybuild/tools/options.py
boegel: Otherwise the option would be `--easyconfig-inject-checksums`, which is overkill imho.

There were no options in this category yet, so this change is fine.

easybuild/framework/easyblock.py
boegel: Actually, there is a technical reason for splitting this up with `re.compile`: using `re.M` in `re.sub` requires Python 2.7... (see also 4f413282e)

boegel:
@wpoely86 remarks fixed, please rereview?

easybuild/framework/easyblock.py
wpoely86: Well, I would take `bak` 😉

easybuild/framework/easyblock.py
wpoely86: typo: `it is`

easybuild/framework/easyblock.py
boegel: changed to `.bak` after checking with @damianam, since he chose `.bck` initially via #2134

PR #2290, opened 2017-08-25
ensure clean error message on easyconfig file parse failure

boegel:
Without this, you sometimes get a nasty traceback when there are parsing problems with the easyconfig file, since not all mistakes lead to a `SyntaxError`.

For example, with something like this included:

```python
foo = '%(name)s %s' % version
```

You'd get a traceback for a `TypeError: format requires mapping`.

With this change, you'll now get a much cleaner error, that clearly points out the line in the easyconfig where the mistake is (hopefully).

### DIFF
--- a/easybuild/framework/easyconfig/format/pyheaderconfigobj.py
+++ b/easybuild/framework/easyconfig/format/pyheaderconfigobj.py
@@ -186,7 +186,7 @@ class EasyConfigFormatConfigObj(EasyConfigFormat):
 
         try:
             exec(pyheader, global_vars, local_vars)
-        except Exception as err:
+        except Exception as err:  # pylint: disable=broad-except
             err_msg = str(err)
             exc_tb = sys.exc_info()[2]
             if exc_tb.tb_next is not None:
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/format/pyheaderconfigobj.py
```diff
@@ -184,8 +186,12 @@ def parse_pyheader(self, pyheader):
 
         try:
             exec(pyheader, global_vars, local_vars)
-        except SyntaxError, err:
-            raise EasyBuildError("SyntaxError in easyconfig pyheader %s: %s", pyheader, err)
+        except Exception as err:
```
wpoely86: doesn't pep8/pylint complain about this? Add an exception for this exception line 😉

easybuild/framework/easyconfig/format/pyheaderconfigobj.py
```diff
@@ -184,8 +186,12 @@ def parse_pyheader(self, pyheader):
...
+            err_msg = str(err)
+            exc_tb = line_idx = sys.exc_info()[2]
```
wpoely86: what does this `line_idx` do?

easybuild/framework/easyconfig/format/pyheaderconfigobj.py
boegel: It probably does, but we're not consistently adding pylint `disable` comments anywhere

I agree it makes sense to do so, but then we need to: i) do it consistently, ii) let Travis complain when `pylint` doesn't agree on something, otherwise it's not worth the effort?

wpoely86: So, what reason do we have for not adding it from now on? We need to start somewhere...

PR #2291, opened 2017-08-25
fix regex to avoid sucking up additional lines prior to module file path in modulefile_path

boegel:
fix for problem reported in https://github.com/easybuilders/easybuild/issues/368

cc @paulmelis


boegel:
I'm fairly sure this fixes the problem reported in https://github.com/easybuilders/easybuild/issues/368, so I suggest we go ahead and merge this... @easybuilders/easybuild-framework-maintainers

PR #2293, opened 2017-08-29
avoid introducing duplicate entry for 'checksums' in exts_list with --inject-checksums

boegel:
Bug fix for `--inject-checksums` that was merged in #2286.

If checksums are already specified for extensions, you may end up with a result like this without this fix (note that `checksums` is specified twice).

```
exts_list = [
    ...
    ('Bundle::BioPerl', '2.1.9', {
        'checksums': ['<previous, possible wrong, checksum>'],
        'source_tmpl': 'Bundle-BioPerl-%(version)s.tar.gz',
        'source_urls': ['https://cpan.metacpan.org/authors/id/C/CJ/CJFIELDS'],
        'checksums': [
            '<new checksum>',  # Bundle-BioPerl-2.1.9.tar.gz
        ],
    }),
    ...
```

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2955,8 +2955,9 @@ def inject_checksums(ecs, checksum_type):
                         exts_list_lines[-1] += ' {'
 
                     for key, val in sorted(ext_options.items()):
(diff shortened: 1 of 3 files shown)


boegel:
@wpoely86 please review?

PR #2294, opened 2017-09-03
add support for append_paths in module generator

boegel:


easybuild/tools/module_generator.py
```diff
@@ -132,6 +132,23 @@ def __init__(self, application, fake=False):
...
+        :param allow_abs: allow providing of absolute paths
+        :param expand_relpaths: expand relative paths into absolute paths (by prefixing install dir)
+        """
+        try:
+            txt = self.prepend_paths(key, paths, allow_abs=allow_abs, expand_relpaths=expand_relpaths)
+        except EasyBuildError as err:
+            raise EasyBuildError(str(err).replace('prepend', 'append'))
+
+        regex = re.compile('^prepend([-_]path)', re.M)
+        return regex.sub(r'append\1', txt)
```
wpoely86: I don't really like this.

boegel: The alternative is to copy-paste the code for `prepend_paths`, and replace `prepend` with `append`. I'm not a big fan of that approach...

akesandgren: It would be a less confusing approach though. This one is slightly ugly in my opinion.

boegel:
@akesandgren @wpoely86 better now?

PR #2311, opened 2017-09-19
ensure $TMPDIR is set to a short path for OpenMPI v2.x

boegel:
If `$TMPDIR` is set to a long path, OpenMPI 2.x fails with an error like:

```
PMIx has detected a temporary directory name that results
in a path that is too long for the Unix domain socket:

    Temp dir: /local/305822X5092X.master21.swalot.gent.vsc/eb-riPl9d/openmpi-sessions-2540023@node2643_0/64817

Try setting your TMPDIR environmental variable to point to
something shorter in length
```

See also https://www.open-mpi.org/faq/?category=osx#startup-errors-with-open-mpi-2.0.x.

This can occur on macOS, where the default temporary directory is something weird like `/var/folders/mg/q0_5yv791yz65cdnbglcqjvc0000gp/T` rather than the more usual `/tmp`, but also in a job environment where `$TMPDIR` is set to a pre-created temporary directory that is specific to the job, e.g. `/local/305822[5092].master21.swalot.gent.vsc` for an array job in Torque (which EasyBuild garbles into `local/305822X5092X.master21.swalot.gent.vsc` to avoid using square brackets in the path name.

To ensure that OpenMPI 2.x doesn't go down on its knees because of this when installing software with EasyBuild (usually because some tests are run using `mpirun`), we can ensure that the `$TMPDIR` value is something short.

### DIFF
--- a/easybuild/toolchains/mpi/openmpi.py
+++ b/easybuild/toolchains/mpi/openmpi.py
@@ -79,8 +79,12 @@ class OpenMPI(Mpi):
         self.orig_tmpdir = os.environ.get('TMPDIR')
         ompi_ver = self.get_software_version(self.MPI_MODULE_NAME)[0]
         if LooseVersion(ompi_ver) >= LooseVersion('2.0') and LooseVersion(ompi_ver) < LooseVersion('3.0'):
-            if len(self.orig_tmpdir) > 50:
-                env.setvar('TMPDIR', tempfile.mkdtemp(prefix='/tmp/'))
+            if len(self.orig_tmpdir) > 40:
+                tmpdir = tempfile.mkdtemp(prefix='/tmp/')
+                env.setvar('TMPDIR', tmpdir)
+                warn_msg = "Long $TMPDIR path may cause problems with OpenMPI 2.x, using shorter path: %s" % tmpdir
+                self.log.warning(warn_msg)
+                print_warning(warn_msg)
 
     def _set_mpi_compiler_variables(self):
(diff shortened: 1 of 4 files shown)


easybuild/toolchains/mpi/openmpi.py
```diff
@@ -57,6 +62,26 @@ class OpenMPI(Mpi):
...
+        """
+        super(OpenMPI, self).prepare(*args, **kwargs)
+
+        # OpenMPI 2.x trips if path specified in $TMPDIR is too long
+        # see https://www.open-mpi.org/faq/?category=osx#startup-errors-with-open-mpi-2.0.x
+        self.orig_tmpdir = os.environ.get('TMPDIR')
+        ompi_ver = self.get_software_version(self.MPI_MODULE_NAME)[0]
+        if LooseVersion(ompi_ver) >= LooseVersion('2.0') and LooseVersion(ompi_ver) < LooseVersion('3.0'):
+            if len(self.orig_tmpdir) > 50:
+                env.setvar('TMPDIR', tempfile.mkdtemp(prefix='/tmp/'))
```
wpoely86: are you really sure that's a good idea? I would alert the user and let them handle it?

boegel: I don't see why this wouldn't be a good idea, since this is exactly the recommendation OpenMPI makes (set `$TMPDIR` to `/tmp`).
At least we're setting it to a unique subdirectory in `/tmp`...

Are you worried about the hardcoding of `/tmp`?

Issue #2317, opened 2017-09-28
Lua generator emits description verbatim into module source


zao:
A description like `description = "hi [there]"` will be emitted into a Lmod module as `whatis([[Description: hi [there]]])`.

There is no escaping inside of such "long brackets" and as such, the string will prematurely terminate on the first `]]` found.

The mechanism that Lua offers to have long brackets occur inside of long brackets is to add one or more `=` inside the left and right bracket pairs. This particular example would be mitigated if it was generated as `[=[hi [there]]=]`.

Other solutions would be to reformulate the output or to just not use obscene amounts of brackets in module descriptions, but this has occurred at least once in the wild in the `OTF` easyconfig.

boegel:
PR to fix this is at #2320, let's follow up there...

PR #2320, opened 2017-09-29
correctly deal with use of special characters in description & co

boegel:
fix for #2317

* curly braces (`{`, `}`) need special handling in module files in Tcl syntax
* square brackets (`[`, `]`) need special handling in module files in Lua syntax in some cases (e.g. `]` at end of string); easiest way is to wrap strings between `[==[` and `]==]` (and make sure that these character combos are not used in the text that gets enclosed in them)

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -454,7 +454,7 @@ class ModuleGeneratorTcl(ModuleGenerator):
         """
         txt = '\n'.join([
             "proc ModulesHelp { } {",
-            "    puts stderr {%s" % self._generate_help_text(),
+            "    puts stderr {%s" % re.sub('([{}])', r'\\\1', self._generate_help_text()),
             "    }",
             '}',
             '',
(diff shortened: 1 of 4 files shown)


easybuild/tools/module_generator.py
```diff
@@ -673,6 +674,9 @@ class ModuleGeneratorLua(ModuleGenerator):
     PATH_JOIN_TEMPLATE = 'pathJoin(root, "%s")'
     UPDATE_PATH_TEMPLATE = '%s_path("%s", %s)'
 
+    START_STR = '[==['
```
wpoely86: make it a list instead of two items?

boegel: Well, a tuple maybe, like this?

```python
START_END_STR = ('[==[', ']==]')
```

Not sure that's a lot better though, since that also means using `START_END_STR[0]` and `START_END_STR[1]` rather than the more straightforward `START_STR` and `END_STR`...

PR #2329, opened 2017-10-19
Add framework support for foss-like toolchain with Spectrum MPI: gsolf

ocaisa:
Not exactly sure what benefits Spectrum MPI from IBM brings to the table but it is based on OpenMPI so it is easy to incorporate.


boegel:
Good to go, in the light of the (manual) test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/5242...

PR #2331, opened 2017-10-23
add support for --preview-pr

migueldiascosta:
Is something like this useful to anyone else?

(I'm calling it when --new-pr is run with -x, but of course, a new option could be created)

### DIFF
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -499,6 +499,31 @@ def review_pr(pr, colored=True, branch='develop'):
 
     return '\n'.join(lines)
 
+def review_new_pr(paths, colored=True, branch='develop'):
+    """
+    Print multi-diff overview between easyconfigs in new PR and specified branch.
+    :param paths: paths to categorized lists of files (easyconfigs, files to delete, patches)
+    :param colored: boolean indicating whether a colored multi-diff should be generated
+    :param branch: easybuild-easyconfigs branch to compare with
+    """
+    tmpdir = tempfile.mkdtemp()
+
+    download_repo_path = download_repo(branch=branch, path=tmpdir)
+    repo_path = os.path.join(download_repo_path, 'easybuild', 'easyconfigs')
+    pr_files = [path for path in paths['easyconfigs']]
+
+    lines = []
+    ecs, _ = parse_easyconfigs([(fp, False) for fp in pr_files], validate=False)
+    for ec in ecs:
+        files = find_related_easyconfigs(repo_path, ec['ec'])
+        _log.debug("File in new PR %s has these related easyconfigs: %s" % (ec['spec'], files))
+        if files:
+            lines.append(multidiff(ec['spec'], files, colored=colored))
+        else:
+            lines.extend(['', "(no related easyconfigs found for %s)\n" % os.path.basename(ec['spec'])])
+
+    return '\n'.join(lines)
+
 
 def dump_env_script(easyconfigs):
     """
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -498,9 +498,10 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
     for ec in ecs:
         files = find_related_easyconfigs(repo_path, ec['ec'])
         if pr:
-            _log.debug("File in PR#%s %s has these related easyconfigs: %s" % (pr, ec['spec'], files))
+            pr_msg = "PR#%s" % pr
         else:
-            _log.debug("File in new PR %s has these related easyconfigs: %s" % (ec['spec'], files))
+            pr_msg = "new PR"
+            _log.debug("File in %s %s has these related easyconfigs: %s" % (pr_msg, ec['spec'], files))
         if files:
             lines.append(multidiff(ec['spec'], files, colored=colored))
         else:
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -491,7 +491,7 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
     elif paths:
         pr_files = [path[0] for path in paths]
     else:
-        return "no easyconfigs or pr specified"
+        raise EasyBuildError("No PR # or easyconfig path specified")
         
     lines = []
     ecs, _ = parse_easyconfigs([(fp, False) for fp in pr_files], validate=False)
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -502,7 +502,7 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
             pr_msg = "PR#%s" % pr
         else:
             pr_msg = "new PR"
-            _log.debug("File in %s %s has these related easyconfigs: %s" % (pr_msg, ec['spec'], files))
+        _log.debug("File in %s %s has these related easyconfigs: %s" % (pr_msg, ec['spec'], files))
         if files:
             lines.append(multidiff(ec['spec'], files, colored=colored))
         else:
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -502,7 +502,7 @@ def review_pr(pr, colored=True, branch='develop'):
 def review_new_pr(paths, colored=True, branch='develop'):
     """
     Print multi-diff overview between easyconfigs in new PR and specified branch.
-    :param paths: paths to categorized lists of files (easyconfigs, files to delete, patches)
+    :param paths: path tuples (path, generated)
     :param colored: boolean indicating whether a colored multi-diff should be generated
     :param branch: easybuild-easyconfigs branch to compare with
     """
@@ -510,7 +510,7 @@ def review_new_pr(paths, colored=True, branch='develop'):
 
     download_repo_path = download_repo(branch=branch, path=tmpdir)
     repo_path = os.path.join(download_repo_path, 'easybuild', 'easyconfigs')
-    pr_files = [path for path in paths['easyconfigs']]
+    pr_files = [path[0] for path in paths]
 
     lines = []
     ecs, _ = parse_easyconfigs([(fp, False) for fp in pr_files], validate=False)
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -499,7 +499,7 @@ def review_pr(pr, colored=True, branch='develop'):
 
     return '\n'.join(lines)
 
-def review_new_pr(paths, colored=True, branch='develop'):
+def preview_pr(paths, colored=True, branch='develop'):
     """
     Print multi-diff overview between easyconfigs in new PR and specified branch.
     :param paths: path tuples (path, generated)
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -490,7 +490,7 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
     if pr:
         pr_files = [path for path in fetch_easyconfigs_from_pr(pr) if path.endswith('.eb')]
     elif paths:
-        pr_files = [path[0] for path in paths]
+        pr_files = paths
     else:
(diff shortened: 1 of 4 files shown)


boegel:
Hmm, just gave this a quick try, doesn't seem to work for me, possibly because of the filename I'm using?

```
$ ls -l CP2K.eb
-rw-r--r--  1 kehoste  staff  1860 Oct 20 14:25 CP2K.eb
$ eb CP2K.eb --new-pr -x
== temporary log file in case of crash /tmp/eb-Ebn_P5/easybuild-_tfsfo.log
ERROR: Can't find path /private/tmp/eb-Ebn_P5/tmp0F3jsT/easybuilders/CP2K.eb
```

Also, I think it makes more sense to park this under a dedicated option like `--preview-pr`?

migueldiascosta:
- it doesn't seem to be the filename, it does work for me if I use it.. strange

- regarding the dedicated option, yes, I suppose so, if other people find it useful (and we get it to actually work :) )

boegel:
```
== 2017-10-23 10:40:48,937 github.py:353 DEBUG Repo easybuild-easyconfigs at branch develop extracted into /private/tmp/eb-81EYLU/tmpDcQKjl/easybuilders/easybuild-easyconfigs-develop
== 2017-10-23 10:40:48,989 build_log.py:158 ERROR EasyBuild crashed with an error (at ?:124 in __init__): Can't find path /private/tmp/eb-81EYLU/tmpDcQKjl/easybuilders/CP2K.eb (at easybuild/framework/easyconfig/tools.py:375 in parse_easyconfigs)
ERROR: Can't find path /private/tmp/eb-81EYLU/tmpDcQKjl/easybuilders/CP2K.eb
```

Does this help at all?

I also see it or other easyconfigs that are named `<name>.eb`...

migueldiascosta:
it's looking for it under `/private/tmp/..`. instead of `/tmp/...`, I guess I'll need to make sure it is copied to the correct path (where does the `private` come from in your environment?) or the correct path is used...

boegel:
@migueldiascosta That's not the problem, `/tmp` and `/private/tmp` are the same thing...

```
$ ls -l /tmp
lrwxr-xr-x@ 1 root  wheel  11 Dec 26  2016 /tmp -> private/tmp
```

boegel:
Hmm, doesn't seem to be related to filename at all:

```
$ eb --new-pr netCDF-Fortran-4.4.4-foss-2017b.eb netCDF-4.4.1.1-foss-2017b.eb HDF5-1.10.1-foss-2017b.eb -x
== temporary log file in case of crash /tmp/eb-tn6hu6/easybuild-KJzqnn.log
ERROR: Can't find path /private/tmp/eb-tn6hu6/tmp_AQfBZ/easybuilders/netCDF-Fortran-4.4.4-foss-2017b.eb
```

migueldiascosta:
ok, I can reproduce if I pass the eb filename without a path, I was always using one. I suppose I need to explicitly copy the file to the tmp folder.

Can you just check with an absolute path to make sure that's the problem?

easybuild/framework/easyconfig/tools.py
```diff
@@ -499,6 +499,31 @@ def review_pr(pr, colored=True, branch='develop'):
...
+    ecs, _ = parse_easyconfigs([(fp, False) for fp in pr_files], validate=False)
+    for ec in ecs:
+        files = find_related_easyconfigs(repo_path, ec['ec'])
+        _log.debug("File in new PR %s has these related easyconfigs: %s" % (ec['spec'], files))
+        if files:
+            lines.append(multidiff(ec['spec'], files, colored=colored))
+        else:
+            lines.extend(['', "(no related easyconfigs found for %s)\n" % os.path.basename(ec['spec'])])
+
+    return '\n'.join(lines)
```
boegel: @migueldiascosta This is working as expected now, but there's way too much copy-pasting going on here when comparing `preview_pr` with `review_pr`.
The only thing that differs is the list of files (`pr_files`) and a part of the log message.

Can we create a private method (`_pr_diff_common` or something like that) which accepts these two as additional arguments?

easybuild/main.py
```diff
@@ -402,7 +402,11 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
 
     # creating/updating PRs
     if new_update_pr:
-        if options.new_pr:
+        if options.preview_pr:
+            print preview_pr(paths, colored=use_color(options.color))
+            cleanup(logfile, eb_tmpdir, testing, silent=True)
+            sys.exit(0)
```
boegel: Hmm, can we avoid doing cleanup/exit here? How is this handled for `--new-pr`?

easybuild/framework/easyconfig/tools.py
migueldiascosta: definitely, will do

easybuild/main.py
migueldiascosta: I think it's because `options.new_pr` is in `no_ec_opts`, there's a cleanup later for those, I suppose `options.preview_pr` can also be added to them (although it's a bit counterintuitive that these don't require any easyconfigs to be specified?)

easybuild/framework/easyconfig/tools.py
```diff
@@ -473,33 +473,41 @@ def find_related_easyconfigs(path, ec):
...
     repo_path = os.path.join(download_repo_path, 'easybuild', 'easyconfigs')
-    pr_files = [path for path in fetch_easyconfigs_from_pr(pr) if path.endswith('.eb')]
-
+    
+    if pr:
+        pr_files = [path for path in fetch_easyconfigs_from_pr(pr) if path.endswith('.eb')]
+    elif paths:
+        pr_files = [path[0] for path in paths]
+    else:
+        return "no easyconfigs or pr specified"
```
boegel: I would `raise` an error instead here, one of both *mus* be provided or something is wrong in the way `review_pr` is used...

easybuild/framework/easyconfig/tools.py
```diff
@@ -473,33 +473,41 @@ def find_related_easyconfigs(path, ec):
...
 
     download_repo_path = download_repo(branch=branch, path=tmpdir)
     repo_path = os.path.join(download_repo_path, 'easybuild', 'easyconfigs')
-    pr_files = [path for path in fetch_easyconfigs_from_pr(pr) if path.endswith('.eb')]
-
+    
+    if pr:
+        pr_files = [path for path in fetch_easyconfigs_from_pr(pr) if path.endswith('.eb')]
+    elif paths:
+        pr_files = [path[0] for path in paths]
```
boegel: Hmm, shouldn't we pass the paths without the `generated` part down to `review_pr` instead?

easybuild/framework/easyconfig/tools.py
```diff
@@ -473,33 +473,41 @@ def find_related_easyconfigs(path, ec):
...
+        
     lines = []
     ecs, _ = parse_easyconfigs([(fp, False) for fp in pr_files], validate=False)
     for ec in ecs:
         files = find_related_easyconfigs(repo_path, ec['ec'])
-        _log.debug("File in PR#%s %s has these related easyconfigs: %s" % (pr, ec['spec'], files))
+        if pr:
+            _log.debug("File in PR#%s %s has these related easyconfigs: %s" % (pr, ec['spec'], files))
+        else:
+            _log.debug("File in new PR %s has these related easyconfigs: %s" % (ec['spec'], files))
```
boegel: Avoid repeating the whole thing:

```python
if pr:
    pr_msg = "PR#%s" % pr
else:
    pr_msg = "new PR"
_log.debug("File in %s %s has these related easyconfigs: %s" % (pr_msg, ec['spec'], files))

File: easybuild/main.py [hunk 6]
```diff
@@ -323,7 +323,7 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     categorized_paths = categorize_files_by_type(orig_paths)
 
     # command line options that do not require any easyconfigs to be specified
-    no_ec_opts = [options.aggregate_regtest, options.new_pr, options.regtest, options.update_pr, search_query]
+    no_ec_opts = [options.aggregate_regtest, options.new_pr, options.preview_pr, options.regtest, options.update_pr, search_query]
```
boegel: line too long, needs wrapping...

boegel: Also, can't we use `no_ec_opts or new_update_pr` below, and avoid including `options.new_pr`, `options.update_pr` and `options.preview_pr` in `no_ec_opts` (since it's wrong, sematically)

easybuild/main.py
```diff
@@ -375,7 +375,7 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
 
     forced = options.force or options.rebuild
     dry_run_mode = options.dry_run or options.dry_run_short
-    new_update_pr = options.new_pr or options.update_pr
+    new_update_pr = options.new_pr or options.update_pr or options.preview_pr
```
boegel: maybe this should be renamed to `new_update_preview_pr`?

easybuild/main.py
```diff
@@ -402,7 +402,9 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
 
     # creating/updating PRs
     if new_update_pr:
-        if options.new_pr:
+        if options.preview_pr:
+            print review_pr(paths=paths, colored=use_color(options.color))
+        elif options.new_pr:
```
boegel: keep these alphabetically sorted, so `new_pr` before `preview_pr`

easybuild/framework/easyconfig/tools.py
migueldiascosta: in main the paths from `det_easyconfig_paths` are transformed to these tuples and I used those, but I suppose I can call `det_easyconfig_paths` in main again, or store the original paths when it is called the first time

easybuild/framework/easyconfig/tools.py
migueldiascosta: done

easybuild/main.py
migueldiascosta: actually, an update_pr could just add or remove a patch, without any easyconfig in the command line...?

easybuild/framework/easyconfig/tools.py
```diff
@@ -474,24 +474,35 @@ def find_related_easyconfigs(path, ec):
...
     lines = []
     ecs, _ = parse_easyconfigs([(fp, False) for fp in pr_files], validate=False)
     for ec in ecs:
         files = find_related_easyconfigs(repo_path, ec['ec'])
-        _log.debug("File in PR#%s %s has these related easyconfigs: %s" % (pr, ec['spec'], files))
+        if pr:
+            pr_msg = "PR#%s" % pr
+        else:
+            pr_msg = "new PR"
+            _log.debug("File in %s %s has these related easyconfigs: %s" % (pr_msg, ec['spec'], files))
```
boegel: the `_log.debug` should not be under the `else:`, should be intended left one level so a log message is also emitted for `--review-pr`

easybuild/main.py
```diff
@@ -428,7 +430,7 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         inject_checksums(ordered_ecs, options.inject_checksums)
 
     # cleanup and exit after dry run, searching easyconfigs or submitting regression test
-    if any(no_ec_opts + [options.check_conflicts, dry_run_mode, options.dump_env_script, options.inject_checksums]):
+    if any(no_ec_opts + [new_update_preview_pr, options.check_conflicts, dry_run_mode, options.dump_env_script, options.inject_checksums]):
```
boegel: now this line is too long, should be shortened somehow...

maybe like this:

```python
stop_options = [options.check_conflicts, dry_run_mode, options.dump_env_script, options.inject_checksums]
if no_ec_opts or new_update_preview_pr or any(stop_options):
    cleanup(logfile, eb_tmpdir, testing)
    sys.exit(0)

easybuild/main.py
boegel: OK, sure, but it doesn't make much sense for `--new-pr` and `--preview-pr` (although in theory you could also only use a patch file there)

boegel:
lgtm, just missing a dedicated test for `--preview-pr`, and an accompanying PR to update the docs?

boegel:
@migueldiascosta Good to go, thanks!

Can you look into an update for the documentation too?

boegel:
documentation added in https://github.com/easybuilders/easybuild/pull/388, thanks again @migueldiascosta!

PR #2337, opened 2017-10-30
fix bug that slipped in with #2331

boegel:
Deleting easyconfigs via `--new-pr` was broken...


easybuild/main.py
```diff
@@ -334,7 +335,9 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         if 'name' in build_specs:
             # try to obtain or generate an easyconfig file via build specifications if a software name is provided
             paths = find_easyconfigs_by_specs(build_specs, robot_path, try_to_generate, testing=testing)
-        elif not any(no_ec_opts):
+        elif no_ec_opts:
```
wpoely86: I'm not understand with you do here? This will always be `True`

boegel: Sorry, missed the any while cleaning up...

PR #2345, opened 2017-11-16
add support for 'exts_default_options' easyconfig parameter

boegel:
We currently use something like this in some easyconfig that have a long list of extensions (e.g. `R`):

```python
ext_options = { ... }

exts_list = [
    ('<name1>', '<version1>', ext_options),
    ...,
]
```

(where `ext_options` is just a local variable in the easyconfig file, EasyBuild itself ignores it when parsing)

In places where we need to list additional options, we do ugly things like:

```python
    ('<name2>', '<version2>', dict(ext_options.items() + [('patches', ['example.patch'])])),
```

This change adds support for `exts_default_options`, which cleans up this mess thoroughly, as follows:

```python
exts_default_options = { ... }

exts_lists = [
    ('<name1>', '<version1>'),
    ('<name2>', '<version2>', {
        'patches': ['example.patch'],
    }),
```

This is particularly useful when we want to enforce having checksums for all extensions...

### DIFF
--- a/test/framework/easyconfig.py
+++ b/test/framework/easyconfig.py
@@ -310,13 +310,13 @@ class EasyConfigTest(EnhancedTestCase):
             'homepage = "http://example.com"',
             'description = "test easyconfig"',
             'toolchain = {"name": "dummy", "version": "dummy"}',
+            'exts_default_options = {',
+            '    "source_tmpl": "gzip-1.4.eb",',  # dummy source template to avoid downloading fail
+            '    "source_urls": ["http://example.com/"]',
+            '}',
             'exts_list = [',
-            '   ("ext1", "ext_ver1", {',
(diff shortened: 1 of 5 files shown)


test/framework/easyconfig.py
```diff
@@ -310,13 +310,13 @@ def test_exts_list(self):
             'homepage = "http://example.com"',
             'description = "test easyconfig"',
             'toolchain = {"name": "dummy", "version": "dummy"}',
+            'exts_default_options = {',
+            '    "source_tmpl": "gzip-1.4.eb",',  # dummy source template to avoid downloading fail
+            '    "source_urls": ["http://example.com/"]',
```
wpoely86: add something with a `%(...)s` too?

PR #2353, opened 2017-11-27
Allow use of `start_dir` in extensions

wpoely86:
It's really annoying that symmetric between an `EasyBlock` and `ExtensionEasyBlock` is so broken. Why does an extension use a different extractor method than the general `extractor_step`? It should use the same function so we can use the same parameters (like `start_dir`).

This PR is a dirty hack to fix it.

### DIFF
--- a/easybuild/framework/extensioneasyblock.py
+++ b/easybuild/framework/extensioneasyblock.py
@@ -32,7 +32,7 @@ from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extension import Extension
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import apply_patch, extract_file
+from easybuild.tools.filetools import apply_patch, change_dir, extract_file
 from easybuild.tools.utilities import remove_unwanted_chars
 
 
@@ -100,6 +100,10 @@ class ExtensionEasyBlock(EasyBlock, Extension):
         if unpack_src:
             targetdir = os.path.join(self.master.builddir, remove_unwanted_chars(self.name))
             self.ext_dir = extract_file("%s" % self.src, targetdir, extra_options=self.unpack_options)
+
+            if self.cfg['start_dir'] and os.path.isdir(self.cfg['start_dir']):
+                self.log.debug("Using start_dir: %s", self.cfg['start_dir'])
+                change_dir(self.cfg['start_dir'])
 
         # patch if needed
         if self.patches:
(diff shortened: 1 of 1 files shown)


easybuild/framework/extensioneasyblock.py
```diff
@@ -81,6 +81,7 @@ def __init__(self, *args, **kwargs):
             # name and version properties of EasyBlock are used, so make sure name and version are correct
             self.cfg['name'] = self.ext.get('name', None)
             self.cfg['version'] = self.ext.get('version', None)
+            self.cfg['start_dir'] = self.ext.get('options', {}).get('start_dir', None)
```
boegel: Do you have an example easyconfig where you'd like to use this?

Since #2194, I don't think this is actually needed?

easybuild/framework/extensioneasyblock.py
```diff
@@ -101,6 +102,10 @@ def run(self, unpack_src=False):
             targetdir = os.path.join(self.master.builddir, remove_unwanted_chars(self.name))
             self.ext_dir = extract_file("%s" % self.src, targetdir, extra_options=self.unpack_options)
 
+            if self.cfg['start_dir'] and os.path.isdir(self.cfg['start_dir']):
```
boegel: use `self.start_dir` rather than `self.cfg['start_dir']` when just reading the value, cfr. #2339 (we should do a cleanup sweep in framework & easyblocks for this)

easybuild/framework/extensioneasyblock.py
wpoely86: Why would it work? The `prepare_step` is never called if you install an extension under the `extensions_step`? 

Current use case: arrow. It has the main C++ library in a subdirectory that needs to be build and installed first. Afterwards you go to the python subdirectory to build the python extensions.
See: easybuilders/easybuild-easyconfigs#5416

boegel: OK, now I understand better what the issue is here, you need to prevent that the `start_dir` value is inherited from the parent, where it's usually defined...

Can you clarify that with a comment above this line?

easybuild/framework/extensioneasyblock.py
```diff
@@ -81,6 +81,8 @@ def __init__(self, *args, **kwargs):
             # name and version properties of EasyBlock are used, so make sure name and version are correct
             self.cfg['name'] = self.ext.get('name', None)
             self.cfg['version'] = self.ext.get('version', None)
+            # We cannot inherit the start_dir from the easyconfig. It should be specified in the extension
+            # itself or be empty.
```
boegel: Do you mind rephrasing this to:

```python
# We can't inherit the 'start_dir' value from the parent (which will be set, and will most likely be wrong).
# It should be specified for the extension specifically, or be empty (so it is auto-derived).

Issue #2351, opened 2017-11-22
Way to declare a dependency as unfiltered


mboisson:
Is it possible to add a way to declare a dependency in an easyconfig file to be unfiltered despite the easybuild config.cfg ?

boegel:
We discussed this a bit more, and @mboisson came up with another idea that would cover his use case: being able to filter out only certain version of particular dependencies, e.g. filter out `flex` `2.6.*` (and up?), but keep older versions, so support something like `--filter-deps='flex>=2.6,foo,bar'`

mboisson:
The new proposal is this : 
1) `name` only means any version
2) `name=version` means this exact version
3) `name=[low,high[` means any version between low (inclusively) and high (exclusively). We could also support `name=[low,high)`, and similarly `name=]low,high]` or `name=(low,high]` to exclude the low version and include the high version.

PR #2355, opened 2017-11-28
added code to support complex specification (with version range) for …

mboisson:
…filter-deps argument

This is to address issue #2351

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -45,6 +45,7 @@ import shutil
 from vsc.utils import fancylogger
 from vsc.utils.missing import get_class_for, nub
 from vsc.utils.patterns import Singleton
+from distutils.version import LooseVersion
 
 import easybuild.tools.environment as env
 from easybuild.framework.easyconfig import MANDATORY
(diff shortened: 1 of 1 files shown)


mboisson:
@boegel, I don't think that the latest errors from travis have anything to do with this PR.

boegel:
@mboisson You're missing #2344 because you're not targeting the `develop` branch with your PR.

I'll change the target branch and retrigger the tests in Travis, let's see if that helps.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -607,11 +608,56 @@ def dependencies(self):
...
         filter_deps = build_option('filter_deps')
+
         if filter_deps:
+            # find if any of the filter_deps are in the form name=version or name=[low:high[
+            complex_fdeps_list = [x for x in filter_deps if "=" in x]
+            complex_fdeps = { }
+            # convert list of k=v into dictionary
+            if complex_fdeps_list:
+                for s in complex_fdeps_list:
+                    k,v = s.split('=')
```
boegel: don't use single-letter variables (outside of list comprehensions), it makes the code quite hard to digest

I'd go with `dep_name` & `dep_version` here, and `dep_spec` rather than `s`?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -607,11 +608,56 @@ def dependencies(self):
...
         self.log.debug("Dependencies BEFORE filtering: %s" % deps)
         filter_deps = build_option('filter_deps')
+
         if filter_deps:
+            # find if any of the filter_deps are in the form name=version or name=[low:high[
+            complex_fdeps_list = [x for x in filter_deps if "=" in x]
+            complex_fdeps = { }
+            # convert list of k=v into dictionary
+            if complex_fdeps_list:
+                for s in complex_fdeps_list:
```
boegel: this can be simplified to:

```python
for dep_spec in [x for x in filter_deps if "=" in x]:
```

No need to make it conditional, for loop over empty list won't do anything, and if you only use the `complex_fdeps_list` value in one place using a variable is overkill imho.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -607,11 +608,56 @@ def dependencies(self):
...
+                    k,v = s.split('=')
+                    
+                    # test whether this is a range
+                    if ":" in v:
+                        # remove range characters
+                        values = v.translate(None, '][()')
```
boegel: The range characters are actually overkill imho, especially since the Pythonic notation `low:high` (with `high` excluded) is more natural for version ranges (e.g., `some_list[2:6]`)

In addition, this will greatly simplify the code below too.

Are you also allowing open ended like `low:` and `:high`?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -607,11 +608,56 @@ def dependencies(self):
...
+                        low, high = values.split(":")
+                        excl_low = v[0] in [']', '(']
+                        excl_high = v[-1] in ['[', ')']
+                    else:
+                        low = v
+                        high = v
```
boegel: use same syntax as above, easier to spot the differences

```python
low, high = dep_version

boegel:
@mboisson Looks like changing the target branch is not sufficient.

Easiest thing to do is to sync your branch `complex_filter_deps` with current `develop` to update this PR.

mboisson:
Apparently, I was not working on the up-to-date develop branch when I originally created this PR. I pulled the most recent changes and created a new PR instead. Closing this one.

PR #2357, opened 2017-11-30
support use of version ranges in --filter-deps

mboisson:
Replacing PR #2355 to address issue #2351


mboisson:
Hi @boegel. I addressed your formatting comments from the other PR in my last commit. All comments except this one : 
```The range characters are actually overkill imho, especially since the Pythonic notation low:high (with high excluded) is more natural for version ranges (e.g., some_list[2:6])

In addition, this will greatly simplify the code below too.

Are you also allowing open ended like low: and :high?``` 


Yes, I am allowing open-ended likw low: and :high. I disagree that it is over-kill. For example, if I want to filter all flex versions except those lower or equal to 2.5.39, I'ld want to filter `]2.5.39:]`

Issue #2076, opened 2017-01-05
RPATH with Python and virtualenv


rjeschmi:
One use case I'd like to see work is to have a version of Python that works properly with virtualenv without 

after compiling Python with --rpath I can see that RPATH is set with readelf:

```
[...]

PR #2358, opened 2017-11-30
Use absolute paths to lib and lib64 icw $ORIGIN for the RPATH

jdonners:
Add %installdir/lib, %installdir/lib64 and $ORIGIN to RPATH. This works much better with shared libraries in non-standard locations, esp. python packages. See the discussion in https://github.com/easybuilders/easybuild-framework/issues/2076. This patch has been tested for building Tensorflow+python (using foss-2016b) and VTK (for intel-2016b) from scratch. Works without hacks in the easyconfig.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -195,6 +195,9 @@ class EasyBlock(object):
         # list of locations to include in RPATH filter used by toolchain
         self.rpath_filter_dirs = []
 
+        # list of locations to include in RPATH used by toolchain
+        self.rpath_include_dirs = []
+
         # logging
         self.log = None
         self.logfile = None
@@ -1734,8 +1737,15 @@ class EasyBlock(object):
         if not self.build_in_installdir:
             self.rpath_filter_dirs.append(self.builddir)
 
+        # always include self.installdir+'/lib' and self.installdir+'/lib64' and $ORIGIN
+        # $ORIGIN will be resolved by the loader to be the full path to the executable or shared object
+        # see also https://linux.die.net/man/8/ld-linux;
+        self.rpath_include_dirs.append(self.installdir+'/lib')
+        self.rpath_include_dirs.append(self.installdir+'/lib64')
+        self.rpath_include_dirs.append('$ORIGIN')
+
         # prepare toolchain: load toolchain module and dependencies, set up build environment
-        self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs)
+        self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs, rpath_include_dirs=self.rpath_include_dirs)
 
         # handle allowed system dependencies
         for (name, version) in self.cfg['allow_system_deps']:
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1745,7 +1745,8 @@ class EasyBlock(object):
         self.rpath_include_dirs.append('$ORIGIN')
 
         # prepare toolchain: load toolchain module and dependencies, set up build environment
-        self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs, rpath_include_dirs=self.rpath_include_dirs)
+        self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs, 
+                               rpath_include_dirs=self.rpath_include_dirs)
 
         # handle allowed system dependencies
         for (name, version) in self.cfg['allow_system_deps']:
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1737,12 +1737,14 @@ class EasyBlock(object):
         if not self.build_in_installdir:
             self.rpath_filter_dirs.append(self.builddir)
 
-        # always include self.installdir+'/lib' and self.installdir+'/lib64' and $ORIGIN
+        # always include '<installdir>/lib', '<installdir>/lib64', $ORIGIN, $ORIGIN/../lib and $ORIGIN/../lib64
         # $ORIGIN will be resolved by the loader to be the full path to the executable or shared object
         # see also https://linux.die.net/man/8/ld-linux;
-        self.rpath_include_dirs.append(self.installdir+'/lib')
-        self.rpath_include_dirs.append(self.installdir+'/lib64')
+        self.rpath_include_dirs.append(os.path.join(self.installdir, 'lib'))
+        self.rpath_include_dirs.append(os.path.join(self.installdir, 'lib64'))
         self.rpath_include_dirs.append('$ORIGIN')
+        self.rpath_include_dirs.append('$ORIGIN/../lib')
+        self.rpath_include_dirs.append('$ORIGIN/../lib64')
 
         # prepare toolchain: load toolchain module and dependencies, set up build environment
         self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs, 
--- a/easybuild/scripts/rpath_args.py
+++ b/easybuild/scripts/rpath_args.py
@@ -53,7 +53,10 @@ if rpath_filter:
 else:
     rpath_filter = None
 
-rpath_include = rpath_include.split(',')
+if not rpath_include:
+    rpath_include = []
+else:
+    rpath_include = rpath_include.split(',')
 
 version_mode = False
 cmd_args, cmd_args_rpath = [], []
@@ -110,8 +113,9 @@ while idx < len(args):
 cmd_args = cmd_args_rpath + cmd_args
 
 cmd_args_rpath = []
-for path in rpath_include:
-    cmd_args_rpath.append(flag_prefix + '-rpath=%s' % path)
+if rpath_include:
+    for path in rpath_include:
+        cmd_args_rpath.append(flag_prefix + '-rpath=%s' % path)
 
 
 if not version_mode:
--- a/easybuild/scripts/rpath_args.py
+++ b/easybuild/scripts/rpath_args.py
@@ -38,7 +38,8 @@ import sys
 
 cmd = sys.argv[1]
 rpath_filter = sys.argv[2]
-args = sys.argv[3:]
+rpath_include = sys.argv[3]
+args = sys.argv[4:]
 
 # wheter or not to use -Wl to pass options to the linker
 if cmd in ['ld', 'ld.gold', 'ld.bfd']:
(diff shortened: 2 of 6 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1734,8 +1737,15 @@ def prepare_step(self, start_dir=True):
...
+        # always include self.installdir+'/lib' and self.installdir+'/lib64' and $ORIGIN
+        # $ORIGIN will be resolved by the loader to be the full path to the executable or shared object
+        # see also https://linux.die.net/man/8/ld-linux;
+        self.rpath_include_dirs.append(self.installdir+'/lib')
+        self.rpath_include_dirs.append(self.installdir+'/lib64')
+        self.rpath_include_dirs.append('$ORIGIN')
+
         # prepare toolchain: load toolchain module and dependencies, set up build environment
-        self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs)
+        self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs, rpath_include_dirs=self.rpath_include_dirs)
```
boegel: line too long (max 120 chars), so wrap it, like so:

```python
self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent, rpath_filter_dirs=self.rpath_filter_dirs,
                       rpath_include_dirs=self.rpath_include_dirs)

File: easybuild/scripts/rpath_args.py [hunk 2]
```diff
@@ -106,13 +112,14 @@
 # add -rpath flags in front
 cmd_args = cmd_args_rpath + cmd_args
 
+cmd_args_rpath = []
+if rpath_include:
+    for path in rpath_include:
+        cmd_args_rpath.append(flag_prefix + '-rpath=%s' % path)
```
boegel: This can be collapsed to a single line, since we're sure `rpath_include` is a list (possibly empty, but that's fine):

```python
cmd_args_rpath = [flag_prefix + '-rpath=%s' % inc for inc in rpath_include]

File: easybuild/scripts/rpath_args.py [hunk 3]
```diff
@@ -52,6 +53,11 @@
 else:
     rpath_filter = None
 
+if not rpath_include:
+    rpath_include = []
+else:
+    rpath_include = rpath_include.split(',')
```
boegel: positive logic is easier to follow (and maybe worth a comment to clarify possible `None` value):

```python
if rpath_include:
    rpath_include = rpath_include.split(',')
else:  # take into account that rpath_include may be None
    rpath_include = []

File: easybuild/tools/toolchain/toolchain.py [hunk 4]
```diff
@@ -796,6 +797,12 @@ def prepare_rpath_wrappers(self, rpath_filter_dirs=None):
         rpath_filter = ','.join(rpath_filter + ['%s.*' % d for d in rpath_filter_dirs or []])
         self.log.debug("Combined RPATH filter: '%s'", rpath_filter)
 
+        # figure out list of patterns to use in rpath include
+        # rpath_include = build_option('rpath_include')
+        rpath_include = ','.join(['%s' % d for d in rpath_include_dirs or []])
+        self.log.debug("Combined RPATH includes: '%s'", rpath_include)
+
+
```
boegel: avoid duplicate blank line, just one is enough

easybuild/scripts/rpath_args.py
```diff
@@ -106,13 +112,14 @@
 # add -rpath flags in front
 cmd_args = cmd_args_rpath + cmd_args
 
+cmd_args_rpath = []
+if rpath_include:
+    for path in rpath_include:
+        cmd_args_rpath.append(flag_prefix + '-rpath=%s' % path)
+
+
```
boegel: avoid duplicate blank line, just one is enough

easybuild/tools/toolchain/toolchain.py
```diff
@@ -702,7 +703,7 @@ def prepare(self, onlymod=None, silent=False, loadmod=True, rpath_filter_dirs=No
 
         if build_option('rpath'):
             if self.options.get('rpath', True):
-                self.prepare_rpath_wrappers()
+                self.prepare_rpath_wrappers(rpath_filter_dirs, rpath_include_dirs)
```
boegel: Wait, we were not passing down `rpath_filter_dirs` here?

That's a bug, right, which wasn't captured by the tests?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -796,6 +797,12 @@ def prepare_rpath_wrappers(self, rpath_filter_dirs=None):
         rpath_filter = ','.join(rpath_filter + ['%s.*' % d for d in rpath_filter_dirs or []])
         self.log.debug("Combined RPATH filter: '%s'", rpath_filter)
 
+        # figure out list of patterns to use in rpath include
+        # rpath_include = build_option('rpath_include')
```
boegel: remove this commented out line?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -796,6 +797,12 @@ def prepare_rpath_wrappers(self, rpath_filter_dirs=None):
...
+        rpath_include = ','.join(['%s' % d for d in rpath_include_dirs or []])
```
boegel: `'%s' % d` doesn't do much, since the directories are already strings, no?

So this is equivalent with:

```python
rpath_include = ','.join(rpath_include_dirs or [])
```

easybuild/tools/toolchain/toolchain.py
```diff
@@ -824,6 +831,7 @@ def prepare_rpath_wrappers(self, rpath_filter_dirs=None):
                     'orig_cmd': orig_cmd,
                     'python': sys.executable,
                     'rpath_args_py': rpath_args_py,
+                    'rpath_include': rpath_include,
```
boegel: Please keep this sorted alphabetically (`i` comes after `f` ;-)).

easybuild/tools/toolchain/toolchain.py
boegel: remove this commented out line?

test/framework/toolchain.py
```diff
@@ -1066,61 +1066,66 @@ def test_rpath_args_script(self):
...
         self.assertEqual(out.strip(), "CMD_ARGS=(%s)" % ' '.join(cmd_args))
 
         # linker command, --enable-new-dtags should be replaced with --disable-new-dtags
-        out, ec = run_cmd("%s ld '' --enable-new-dtags foo.o" % script, simple=False)
+        out, ec = run_cmd("%s ld '' '%s/lib,%s/lib64,$ORIGIN' --enable-new-dtags foo.o" % (script, self.test_prefix, self.test_prefix), simple=False)
         self.assertEqual(ec, 0)
         expected = '\n'.join([
             "CMD_ARGS=('foo.o')",
-            "RPATH_ARGS='--disable-new-dtags -rpath=$ORIGIN/../lib -rpath=$ORIGIN/../lib64'",
+            "RPATH_ARGS='--disable-new-dtags -rpath=%s/lib -rpath=%s/lib64 -rpath=$ORIGIN'" % (self.test_prefix, self.test_prefix),
```
boegel:
```python
"RPATH_ARGS='--disable-new-dtags -rpath=%(p)s/lib -rpath=%(p)s/lib64 -rpath=$ORIGIN'" % {'p': self.test_prefix},

File: test/framework/toolchain.py [hunk 11]
```diff
@@ -1066,61 +1066,66 @@ def test_rpath_args_script(self):
         script = find_eb_script('rpath_args.py')
 
         # simplest possible compiler command
-        out, ec = run_cmd("%s gcc '' -c foo.c" % script, simple=False)
+        out, ec = run_cmd("%s gcc '' '%s/lib,%s/lib64,$ORIGIN' -c foo.c" % (script, self.test_prefix, self.test_prefix), simple=False)
```
boegel: Construct the argument separately, since you're using it in multiple tests.
This also avoids making the line too long.

```python
rpath_inc = '%(prefix)s/lib,%(prefix)s/lib64,$ORIGIN' % {'prefix': self.test_prefix}
out, ec = run_cmd("%s gcc '' '%s' -c foo.c" % (script, rpath_inc), simple=False)
```

And then reuse `rpath_inc` below too.

easybuild/tools/toolchain/toolchain.py
jdonners: Uhm, it seems so. I just passed it as well, because I wanted to pass `rpath_include_dirs` (which I guess I could've passed as keyword argument, still learning python..). The test in `test_toolchain_prepare_rpath` sets the build_option `rpath_filter`, while `prepare_rpath_wrappers` combines the build_option `rpath_filter` AND the optional argument `rpath_filter_dirs`. The last one is not tested (and apparently not yet used).

boegel: OK, we should make sure the tests check the usage... I'll take a look at that.

easybuild/framework/easyblock.py
```diff
@@ -1734,8 +1737,16 @@ def prepare_step(self, start_dir=True):
         if not self.build_in_installdir:
             self.rpath_filter_dirs.append(self.builddir)
 
+        # always include self.installdir+'/lib' and self.installdir+'/lib64' and $ORIGIN
+        # $ORIGIN will be resolved by the loader to be the full path to the executable or shared object
+        # see also https://linux.die.net/man/8/ld-linux;
+        self.rpath_include_dirs.append(self.installdir+'/lib')
+        self.rpath_include_dirs.append(self.installdir+'/lib64')
+        self.rpath_include_dirs.append('$ORIGIN')
```
boegel: @jdonners Shouldn't we also retain `$ORIGIN/../lib` and `$ORIGIN/../lib64` in here? That's not necessarily the same as `<installdir>/lib` and `<installdir>/lib64`, is it?

jdonners: @boegel indeed they're not necessarily the same. Conservatively, I thought it's better to add fewer paths to the `RPATH`. I'd expect that packages that install shared objects in non-standard locations, also somehow make these directories findable in some wrapper scripts. There's no guarantee that that is indeed the case. On the other side, it still is an experimental feature..

boegel: For now, I'd stick to also included `$ORIGIIN/../lib` and `$ORIGIN/../lib64`.

It indeed makes sense to limit the amount of paths we inject, but 3 or 5 isn't going to make a big difference I think.

The order may be important though (w.r.t. speeding up the loader): we should probably list the most common locations first?

boegel: @jdonners Fixed in https://github.com/sara-nl/easybuild-framework/pull/1

easybuild/tools/toolchain/toolchain.py
boegel: Enhanced test to catch this (now fixed) bug is included in updates in https://github.com/sara-nl/easybuild-framework/pull/1

boegel:
@jdonners Please review/merge https://github.com/sara-nl/easybuild-framework/pull/1, with that included this is good to go imho.

boegel:
Going in, thanks @jdonners!

PR #2365, opened 2017-12-20
add support for Environment Modules 4

ebagrenrut:
Add a new class `EnvironmentModules4`. Introduces a `MAX_VERSION` to `ModulesTool` that can be used to set a maximum version for a particular module tool. Limits `EnvironmentModulesC` to a maximum version of `3.2.10` (the last release of the C implementation of Environment Modules). Limits `EnvironmentModulesTcl` to a maximum version of `1.923` (the last version of modules-tcl).


easybuild/tools/modules.py
```diff
@@ -237,14 +239,30 @@ def set_and_check_version(self):
         except (OSError), err:
             raise EasyBuildError("Failed to check version: %s", err)
 
-        if self.REQ_VERSION is None:
-            self.log.debug("No version requirement defined.")
+        if self.REQ_VERSION is None and self.MAX_VERSION is None:
+                self.log.debug("No version requirement defined.")
```
boegel: No need for the extra indent here?

easybuild/tools/modules.py
```diff
@@ -237,14 +239,30 @@ def set_and_check_version(self):
...
         else:
-            if StrictVersion(self.version) < StrictVersion(self.REQ_VERSION):
-                raise EasyBuildError("EasyBuild requires v%s >= v%s (no rc), found v%s",
-                                     self.__class__.__name__, self.REQ_VERSION, self.version)
+            if self.REQ_VERSION is None: # MAX_VERSION is not None
```
boegel: Why not check directly whether `self.MAX_VERSION` is different from `None`?

```python
elif self. MAX_VERSION is not None:

File: easybuild/tools/modules.py [hunk 3]
```diff
@@ -237,14 +239,30 @@ def set_and_check_version(self):
...
+                self.log.debug("No required minimum version defined.")
+                if StrictVersion(self.version) > StrictVersion(self.MAX_VERSION):
+                    raise EasyBuildError("EasyBuild requires v%s <= v%s (no rc), found v%s",
+                                         self.__class__.__name__, self.version, self.MAX_VERSION)
+                else:
+                    self.log.debug('Version %s matches requirement <= %s' % (self.version, self.MAX_VERSION))
+            elif self.MAX_VERSION is None: # REQ_VERSION is not None
```
boegel: You should change this to an `if`, to make sure that if both `self.REQ_VERSION` and `self.MAX_VERSION` are specified, that both are checked?

```python
if self. REQ_VERSION is not None:

File: easybuild/tools/modules.py [hunk 4]
```diff
@@ -237,14 +239,30 @@ def set_and_check_version(self):
...
+        if self.REQ_VERSION is None and self.MAX_VERSION is None:
+                self.log.debug("No version requirement defined.")
         else:
-            if StrictVersion(self.version) < StrictVersion(self.REQ_VERSION):
-                raise EasyBuildError("EasyBuild requires v%s >= v%s (no rc), found v%s",
-                                     self.__class__.__name__, self.REQ_VERSION, self.version)
+            if self.REQ_VERSION is None: # MAX_VERSION is not None
+                self.log.debug("No required minimum version defined.")
+                if StrictVersion(self.version) > StrictVersion(self.MAX_VERSION):
+                    raise EasyBuildError("EasyBuild requires v%s <= v%s (no rc), found v%s",
```
boegel: You should probably drop the `no rc` part here, since using a release candidate of the highest allowed version is  older than that version, so it's OK to use it.

easybuild/tools/modules.py
```diff
@@ -237,14 +239,30 @@ def set_and_check_version(self):
...
+                                         self.__class__.__name__, self.version, self.MAX_VERSION)
+                else:
+                    self.log.debug('Version %s matches requirement <= %s' % (self.version, self.MAX_VERSION))
```
boegel: Style nitpicking: Please use `,` rather than `%`:

```python
self.log.debug('Version %s matches requirement <= %s', self.version, self.MAX_VERSION)

File: easybuild/tools/modules.py [hunk 6]
```diff
@@ -237,14 +239,30 @@ def set_and_check_version(self):
...
+            elif self.MAX_VERSION is None: # REQ_VERSION is not None
+                self.log.debug("No allowed maximum version defined.")
+                if StrictVersion(self.version) < StrictVersion(self.REQ_VERSION):
+                    raise EasyBuildError("EasyBuild requires v%s >= v%s (no rc), found v%s",
+                                         self.__class__.__name__, self.REQ_VERSION, self.version)
+                else:
+                    self.log.debug('Version %s matches requirement >= %s' % (self.version, self.REQ_VERSION))
```
boegel: Use `,` rather than `%`:

```python
self.log.debug('Version %s matches requirement >= %s', self.version, self.REQ_VERSION)

File: easybuild/tools/modules.py [hunk 7]
```diff
@@ -237,14 +239,30 @@ def set_and_check_version(self):
...
             else:
-                self.log.debug('Version %s matches requirement %s' % (self.version, self.REQ_VERSION))
+                self.log.debug("Required minimum version and maximum allowed version defined.")
+                # Check that we have not made a mistake and set REQ_VERSION > MAX_VERSION
+                if StrictVersion(self.REQ_VERSION) > StrictVersion(self.MAX_VERSION):
+                    raise EasyBuildError("v%s, REQ_VERSION must be <= MAX_VERSION", self.__class__)
+                else:
+                    self.log.debug('Version %s matches requirements >= %s, <= %s' % (self.version, self.REQ_VERSION, self.MAX_VERSION))
```
boegel: I would drop this whole block, and use two `if` conditions above.

easybuild/tools/modules.py
```diff
@@ -997,12 +1015,18 @@ class EnvironmentModulesC(ModulesTool):
     """Interface to (C) environment modules (modulecmd)."""
     COMMAND = "modulecmd"
     REQ_VERSION = '3.2.10'
+    MAX_VERSION = '3.2.10'
```
boegel: Make this a bit more liberal, in case people have local made local modifications and bumped this version.

```python
MAX_VERSION = '3.99'  # fictional version, just to indicate that version 4.0.0 and up are not compatible

File: easybuild/tools/modules.py [hunk 9]
```diff
@@ -997,12 +1015,18 @@ class EnvironmentModulesC(ModulesTool):
...
     VERSION_REGEXP = r'^\s*(VERSION\s*=\s*)?(?P<version>\d\S*)\s*'
 
     def update(self):
         """Update after new modules were added."""
         pass
 
+class EnvironmentModules4(EnvironmentModulesC):
```
boegel: Maybe `EnvironmentModulesTcl4` is a better name?

easybuild/tools/modules.py
```diff
@@ -1014,6 +1038,7 @@ class EnvironmentModulesTcl(EnvironmentModulesC):
     COMMAND_SHELL = ['tclsh']
     VERSION_OPTION = ''
     REQ_VERSION = None
+    MAX_VERSION = '1.923'
```
boegel: I would leave this unspecified, there's little point to it?

If somebody uses `EnvironmentModulesTcl` with Modules v4.0, it will fail anyway because `eb` can't determine the version?

easybuild/tools/modules.py
ebagrenrut: You're right, my logic is broken. Will fix.

easybuild/tools/modules.py
```diff
@@ -1076,6 +1088,14 @@ def remove_module_path(self, path, set_mod_paths=True):
             self.set_mod_paths()
 
 
+class EnvironmentModules(EnvironmentModulesTcl):
+    """Interface to environment modules 4.0+"""
+    COMMAND = os.path.join(os.environ['MODULESHOME'], 'libexec', 'modulecmd.tcl')
```
boegel: This should be done more carefully, if `$MODULESHOME` is not defined, this will lead to a nasty crash/traceback.

My suggestion is to use a fallback value, like this:

```python
COMMAND = os.path.join(os.getenv('MODULESHOME', 'MODULESHOME_NOT_DEFINED'), 'libexec', 'modulecmd.tcl')

Comment (ebagrenrut) [reply easybuild/tools/modules.py]: DO we need both MODULESHOME and MODULESHOME_NOT_DEFINED? If using os.getenv and MODULESHOME is undefined, None will be returned and the traceback won't occur. In that case, eb would look for None/libexec/modulecmd.tcl and throw an exception.

Comment (boegel) [reply easybuild/tools/modules.py]: @ebagrenrut Yes, we need the fallback value, since `os.path.join` won't be happy with the `None` value:

```
>>> os.path.join(None, 'libexec', 'modulecmd.tcl')
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/posixpath.py", line 70, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

**Comment by ebagrenrut** [comment]
@boegel I think there is an assumption in the test_load method of the ModulesTest class that is probably not true for Modules 4. The test loads the GCC/4.6.4 modules, then unsets $EBROOTGCC. A comment indicates that EBROOTGCC will be reset if the GCC/4.6.4 module is loaded again. Modules 4 doesn't reset EBROOTGCC (I've confirmed by simply trying it in my environment), and I think that may be by design.

**Comment by boegel** [comment]
@ebagrenrut The last failing test is because the bootstrap script is not aware of the new `EnvironmentModules` class yet.

That's not hard to fix, but when that is fixed there's another problem: the test for the bootstrap script will still fail because it will try to install the latest EasyBuild release (v3.5.0), which doesn't know about `EnvironmentModules` yet. This leads to:

```
eb: error: option --modules-tool: invalid choice: 'EnvironmentModules' (choose from 'EnvironmentModulesC', 'EnvironmentModulesTcl', 'Lmod')
```

In other words, a chicken-or-egg problem.

So, I think we should just bypass the testing of the bootstrap script for now under `EnvironmentModules`, and open another PR to fix the bootstrap script once EasyBuild v3.5.1 (which includes this PR that adds support for Modules v4).

boegel:
@ebagrenrut Change in Travis config file to bypass testing of bootstrap script done in https://github.com/ebagrenrut/easybuild-framework/pull/4.

boegel:
This looks good to go now, thanks for you help with this @ebagrenrut and @xdelaruelle!

I'll follow up with another PR to re-enable testing of the bootstrap script with Modules v4, which can only be merged once EasyBuild v3.5.1 is released...

PR #2368, opened 2018-01-04
support for injecting additional toolchainopts

migueldiascosta:
this is mostly because --try-amend does not support dictionaries, and supporting it for arbitrary easyconfig parameters would require defining a new (too convoluted?) argument syntax

improvements/alternatives?

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -321,6 +321,7 @@ class EasyBuildOptions(GeneralOption):
 
         opts = OrderedDict({
             'add-dummy-to-minimal-toolchains': ("Include dummy in minimal toolchain searches", None, 'store_true', False),
+            'additional-toolchain-opts': ("List of toolchain options to add", 'strlist', 'store', []),
             'allow-loaded-modules': ("List of software names for which to allow loaded modules in initial environment",
                                      'strlist', 'store', DEFAULT_ALLOW_LOADED_MODULES),
             'allow-modules-tool-mismatch': ("Allow mismatch of modules tool and definition of 'module' function",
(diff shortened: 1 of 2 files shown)


easybuild/tools/options.py
```diff
@@ -1178,6 +1179,16 @@ def process_software_build_specs(options):
...
+    # process --extra-toolchainopts
+    if options.extra_toolchainopts:
+        try_to_generate = True
+
+        toolchainopts = {}
+        for spec in options.extra_toolchainopts:
+            param, value = spec.split('=')
+            toolchainopts.update({param: value})
+        build_specs.update({'toolchainopts': toolchainopts})
+
```
akesandgren: In case someone uses --extra-toolchains without arguments, is this caught somewhere?
If not, only set try_to_generate if toolchainopts is actually non-empty, and don't call build_specs.update if it is empty.

Or perhaps error out if it is empty?

migueldiascosta: thanks for looking into this

if it is empty the options parser will error out saying that option it requires an argument

but if it is malformed, that's another issue, for instance if there is no '=' the split above will fail with no more values to unpack...

the correct usage allows something like `--extra-toolchainopts=usempi=True,openmp=True,optarch="Intel:-O2;GCC=-O3"`, maybe this should be made clearer somewhere?

boegel:
@migueldiascosta Needs a sync with `develop` to fix merge conflict.

I think it makes sense to rework this a bit to make it use the `parse_param_value` function that is being added in #2665 (syntax is a little bit different, but there's little point in supporting two different ways of specifying a `dict`-value?

migueldiascosta:
@boegel will get back to this after #2665 is merged, then

PR #2369, opened 2018-01-05
bump copyright to 2018 in license headers

boegel:


easybuild/tools/hooks.py
```diff
@@ -1,5 +1,5 @@
 # #
-# Copyright 2017-2017 Ghent University
+# Copyright 2018-2018 Ghent University
```
ocaisa: 2017-2018?

test/framework/hooks.py
```diff
@@ -1,5 +1,5 @@
 # #
-# Copyright 2017-2017 Ghent University
+# Copyright 2018-2018 Ghent University
```
ocaisa: Same?

easybuild/tools/hooks.py
boegel: Nice catch! Now fixed...

PR #2370, opened 2018-01-05
update bootstrap script to be compatible with Modules v4

boegel:
follow-up for ~~#2365~~

This requires that EasyBuild v3.5.1 is released, since the bootstrap script can only work with Modules v4 if the latest EasyBuild releases actually supports using Modules v4 as modules tool

### DIFF
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -705,6 +705,9 @@ def stage2(tmpdir, templates, install_path, distribute_egg_dir, sourcepath):
     from easybuild.main import main as easybuild_main
     easybuild_main()
 
+    if print_debug:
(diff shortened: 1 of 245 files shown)


easybuild/scripts/bootstrap_eb.py
```diff
@@ -240,7 +240,7 @@ def check_module_command(tmpdir):
 
     def check_cmd_help(modcmd):
         """Check 'help' output for specified command."""
-        modcmd_re = re.compile(r'module\s.*command\s')
+        modcmd_re = re.compile(r'module\s.*command')
```
boegel: Note to self: the previous regex was slightly too strict for the output of `modulecmd.tcl python help` produced by Modules v4, but still good enough to check that the command works in general imho.

boegel:
This was fixed via #2390 by @xdelaruelle on top of this, thanks again!

Issue #2342, opened 2017-11-13
Error with --try-toolchain


mvsurfsara:
Getting the following error when trying the option --try-toolchain=intel,2016b for a intel/2017a easyconfig file:

```
~> eb --dry-run --try-toolchain=intel,2016b NCO-4.6.6-intel-2017a.eb --robot
== temporary log file in case of crash /tmp/marcov/eb-mT2Kk1/easybuild-o4bE7L.log
ERROR: Irresolvable dependencies encountered: cURL/7.53.1-intel-2016b, flex/2.6.4-intel-2016b, libreadline/7.0-intel-2016b, binutils/2.27-intel-2016b
```

The Irresolvable dependencies listed here are all available for the GCCcore toolchain.

ocaisa:
Unfortunately this shortcoming is well known but still a very early work in progress, see https://github.com/easybuilders/easybuild-framework/issues/1627

boegel:
@ocaisa I was sitting next to @mvsurfsara when he ran into this, and asked to open an issue since I think we can do better here with limited extra effort.

Since we're sticking to the same toolchain, just a different version, so we don't have the problem of needing to map subtoolchains between different toolchains.

For some reason, there is no easyconfig generated for `cURL`, not even with the target toolchain `intel/2016`. That's definitely a bug, I'm not sure why that happens at all...

Even better would be to map the `GCCcore` used in `intel/2017a` to the one used in `intel/2016b`, and then just replace the version in the `cURL` easyconfig that's available for `GCCcore/6.3.0` (subtoolchain of `intel/2017a`.)

Also supporting going from `foss` to `intel` is more difficult here, since we then (potentially) need to map `gompi` to `iimpi` for example (but that almost never happens, it's usually either `GCCcore` or `foss`/`intel` in practice).

boegel:
I found the problem causing this issue, and I'm working on fix a fix (which is less trivial then I initially thought it was going to be, of course it is...).

The problem is that the `while` loop in the `tweak` that is intended to filter out the toolchain & its deps from the list of easyconfigs to tweak is too aggressive: it also filters out dependencies that use a subtoolchain like `GCCcore`.

Fixing this leads to another problem: the code in `tweak_one` blindly replacing the toolchain, not taking into account whether dependencies are using a subtoolchain (incl. `dummy`) or not.

[...]

PR #2371, opened 2018-01-05
correctly filter out toolchain & dependencies under --try-toolchain --robot

boegel:
(unfinished) fix for issue reported in #2342

cc @ocaisa

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -107,6 +107,8 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
             else:
                 raise EasyBuildError("Failed to isolate toolchain '%s' in dependency graph: %s", tcname, orig_ecs)
 
+            # FIXME: need to actually construct dep graph without toolchain,
+            #        since some dependencies of toolchain may also be needed for dependencies of software...
             # determine full dep graph for toolchain itself
             tc_dep_mod_names = set()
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyconfig/tweak.py
```diff
@@ -93,18 +93,46 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
...
+        if tcname != DUMMY_TOOLCHAIN_NAME:
+
+            # locate toolchain itself, there should be exactly one match
+            tc_ecs = [ec for ec in orig_ecs if ec['ec']['name'] == tcname]
+            if len(tc_ecs) == 1:
+                tc_ec = tc_ecs[0]
+            else:
+                raise EasyBuildError("Failed to isolate toolchain '%s' in dependency graph: %s", tcname, orig_ecs)
+
+            # FIXME: need to actually construct dep graph without toolchain,
```
ocaisa: I don't think this is necessary. Only explicitly listed easyconfigs are tweaked and put first in the search, everything else is tweaked and placed last in the search (i.e., they're only used if they don't already exist). When it comes to the toolchain, that will have been replaced and so once the deps of the tweaked easyconfigs (that appeared on the command line) are resolved the tweaked toolchain easyconfigs will never get used.

akesandgren:
@boegel  Care to fix the conflict and what is the state of this?

PR #2373, opened 2018-01-12
Make code in `tools/job/gc3pie.py` forward-compatible with GC3Pie 2.5

riccardomurri:


riccardomurri:
This is needed to allow people to run code that has already fixed issues like https://github.com/uzh/gc3pie/issues/622

boegel:
I see no reason to hold this back, thanks @riccardomurri!

PR #2374, opened 2018-01-12
Added functionality to skip devel module with naming scheme

jdonners:
So far, we have installed all modules with the default naming scheme. I would like to install a separate set of modules with a lowercase naming scheme (I know about the caveats of using such a scheme). This change allows to install an alternative set of modules that points to the same installation directory, without overwriting the devel module. I also attached the lowercase naming scheme that I used (I had to rename it from `.py` to `.txt`).
[lowercase_easybuild_mns.txt](https://github.com/easybuilders/easybuild-framework/files/1626955/lowercase_easybuild_mns.txt)

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -923,10 +923,6 @@ class EasyBlock(object):
         WARNING: you cannot unload using $EBDEVELNAME (for now: use module unload `basename $EBDEVELNAME`)
         """
 
-        if not ActiveMNS().mns.det_generate_devel_module():
-            self.log.info("Skipping devel module...")
-            return
-
         self.log.info("Making devel module...")
 
         # load fake module
@@ -2328,8 +2324,11 @@ class EasyBlock(object):
             mod_symlink_paths = ActiveMNS().det_module_symlink_paths(self.cfg)
             self.module_generator.create_symlinks(mod_symlink_paths, fake=fake)
 
-            if not fake:
+            if ActiveMNS().mns.det_make_devel_module() and not fake:
                 self.make_devel_module()
+            else:
+                self.log.info("Skipping devel module...")
+
 
         if build_option('set_default_module'):
             self._set_module_as_default()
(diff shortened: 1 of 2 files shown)


jdonners:
the failed test looks like a network hiccup..

boegel:
@jdonners It most likely is, I've re-triggered the test.

easybuild/framework/easyblock.py
```diff
@@ -923,6 +923,10 @@ def make_devel_module(self, create_in_builddir=False):
         WARNING: you cannot unload using $EBDEVELNAME (for now: use module unload `basename $EBDEVELNAME`)
         """
 
+        if not ActiveMNS().mns.det_generate_devel_module():
+            self.log.info("Skipping devel module...")
+            return
```
boegel: Hmm, we try to avoid inline `return` statements across the codebase (doesn't help with readability).

To avoid changing indentation in this whole method, why not change the condition that guards the (only) call to `make_devel_module` instead to include `and ActiveMNS().mns.det_generate_devel_module()`?

easybuild/tools/module_naming_scheme/mns.py
```diff
@@ -164,3 +164,11 @@ def is_short_modname_for(self, short_modname, name):
                        short_modname, name, modname_regex.pattern, res)
 
         return res
+
+    def det_generate_devel_module(self):
+        """
+        Determine if a devel module should be generated. Can be used to create a separate set of modules with a different naming scheme.
```
boegel: line too long (> 120 chars), easy/sensible fix is to move 2nd sentence to new line

jdonners:
should I create another wrapper function for det_generate_devel_module in ActiveMNS()? I'm not sure what it's meant for..

jdonners:
and I could just rename the function to `det_make_devel_module`, so it matches the routine it is meant for.

jdonners:
Hi @boegel, is there something else you'd like to see changed about this pull request?

boegel:
@jdonners Tests in Travis fail on version check for bootstrap script, which is clearly not related to the changes you made.

Can you sync your branch with current `develop`? That should fix that issue...

boegel:
Looks good, thanks @jdonners!

PR #2378, opened 2018-01-16
New command-line option `--job-max-jobs` to cap nr of submitted build jobs

riccardomurri:


riccardomurri:
(Well, maybe the option name was not so well-chosen...)

easybuild/tools/options.py
```diff
@@ -634,6 +634,7 @@ def job_options(self):
         opts = OrderedDict({
             'backend-config': ("Configuration file for job backend", None, 'store', None),
             'cores': ("Number of cores to request per job", 'int', 'store', None),
+            'max-jobs': ("Maximum number of concurrent jobs (queued or running)", 'int', 'store', 50),
```
boegel: @riccardomurri Shouldn't we set to default to `0` (which I'm guessing means no limit), to maintain current behaviour?

In any case, `50` seems so arbitrary...

Thoughts @victorusu?

riccardomurri:
Good catch. Thanks @boegel !

(50 is the default that we have in GC3Pie's `SessionBasedScript` so I just assumed we had the same in EasyBuild...)

boegel:
Thanks @riccardomurri!

PR #2379, opened 2018-01-16
Add pagination support in clean_gists.py

wpoely86:


boegel:
@wpoely86 Hmm, one of the checks in Travis keeps failing on the version check for the bootstrap script... That doesn't make any sense to me, since you're not touching that at all, but re-triggering the test isn't helping.

Can you sync this up with current `develop`, maybe that will do the trick...?

PR #2382, opened 2018-01-17
adding env var to force EasyBuild version during bootstrap

bmcgough:
We are automating EasyBuild within Docker and need to get a predictable EasyBuild version when bootstrapping.

### DIFF
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -86,6 +86,7 @@ os.environ['PYTHONPATH'] = ''
 
 EASYBUILD_BOOTSTRAP_SOURCEPATH = os.environ.pop('EASYBUILD_BOOTSTRAP_SOURCEPATH', None)
 EASYBUILD_BOOTSTRAP_SKIP_STAGE0 = os.environ.pop('EASYBUILD_BOOTSTRAP_SKIP_STAGE0', False)
+EASYBUILD_BOOTSTRAP_FORCE_VERSION = os.environ.pop('EASYBUILD_BOOTSTRAP_FORCE_VERSION', None)
 
 # keep track of original environment (after clearing PYTHONPATH)
 orig_os_environ = copy.deepcopy(os.environ)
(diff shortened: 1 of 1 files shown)


easybuild/scripts/bootstrap_eb.py
```diff
@@ -502,7 +503,10 @@ def stage1(tmpdir, sourcepath, distribute_egg_dir):
             cmd.append(source_tarballs[VSC_BASE])
     else:
         # install meta-package easybuild from PyPI
-        cmd.append('easybuild')
+        if forcedversion is not None:
```
akesandgren: if forcedversion: should be enough here and is easier to read.

easybuild/scripts/bootstrap_eb.py
```diff
@@ -767,6 +771,10 @@ def main():
     if sourcepath is not None:
         info("Fetching sources from %s..." % sourcepath)
 
+    forcedversion = EASYBUILD_BOOTSTRAP_FORCE_VERSION
+    if EASYBUILD_BOOTSTRAP_FORCE_VERSION is not None:
+        info("Forcing specified version %s..." % forcedversion)
```
akesandgren: If forcedversion isn't use further down, then drop it and use EASYBUILD_BOOTSTRAP_FORCE_VERSION in the info(...).
If it is used then changing the if EASYBUILD.... to
if forcedversion:
makes it easier to read.

boegel: @akesandgren Well, it *is* used further down, since it's being passed to `stage1`.

So, your second suggestion is right on the spot, please use `if forcedversion:` above

boegel:
@bmcgough You'll need to update the version of the bootstrap script in `bootstrap_eb.py` and then also in `.travis.yml` where you'll also need to update the `SHA256` checksum of `bootstrap_eb.py` to make Travis happy (we do this to ensure the version of the bootstrap script is bumped whenever any changes are made to it).

boegel:
Wow, just noticed how long this has been open...

I'm very sorry for not getting back to you on this @bmcgough, since it's a useful contribution, it somehow slipped through the cracks... :(

PR #2384, opened 2018-01-29
automatically add 'new' and/or 'update' labels in --new-pr

migueldiascosta:
tested at https://github.com/migueldiascosta/easybuild-easyconfigs/pulls

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1555,11 +1555,11 @@ def copy_easyconfigs(paths, target_dir):
 
             target_path = det_location_for(path, target_dir, soft_name, ec_filename)
 
-            file_info['new'].append(not os.path.exists(target_path))
-
-            file_info['new_folder'].append(not os.path.exists(os.path.dirname(target_path)))
-
-            file_info['new_file_in_existing_folder'].append(file_info['new'][-1] and not file_info['new_folder'][-1])
+            new_file = not os.path.exists(target_path)
+            new_folder = not os.path.exists(os.path.dirname(target_path))
+            file_info['new'].append(new_file)
+            file_info['new_folder'].append(new_folder)
+            file_info['new_file_in_existing_folder'].append(new_file and not new_folder)
 
             copy_file(path, target_path, force_in_dry_run=True)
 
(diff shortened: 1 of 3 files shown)


migueldiascosta:
regarding the test errors, `test_copy_easyconfigs` would need to take into account the new `file_info` keys if this approach is used

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1554,6 +1556,11 @@ def copy_easyconfigs(paths, target_dir):
             target_path = det_location_for(path, target_dir, soft_name, ec_filename)
 
             file_info['new'].append(not os.path.exists(target_path))
+
+            file_info['new_folder'].append(not os.path.exists(os.path.dirname(target_path)))
+
+            file_info['new_file_in_existing_folder'].append(file_info['new'][-1] and not file_info['new_folder'][-1])
```
boegel: I would clean this up a bit:

```python
new_file = not os.path.exists(target_path)
new_folder = not os.path.exists(os.path.dirname(target_path))
file_info['new'].append(new_file)
file_info['new_folder'].append(new_folder)
file_info['new_file_in_existing_folder'].append(new_file and not new_folder)
```

easybuild/tools/github.py
```diff
@@ -1100,6 +1107,7 @@ def new_pr(paths, ecs, title=None, descr=None, commit_msg=None):
         "* target: %s/%s:%s" % (pr_target_account, pr_target_repo, pr_target_branch),
         "* from: %s/%s:%s" % (github_account, pr_target_repo, branch),
         "* title: \"%s\"" % title,
+        "* labels: %s" % labels,
```
boegel: I would use:

```python
"* labels: %s" % (', '.join(labels) or '(none)'),
```

easybuild/tools/github.py
```diff
@@ -1124,6 +1132,15 @@ def new_pr(paths, ecs, title=None, descr=None, commit_msg=None):
...
         print_msg("Opened pull request: %s" % data['html_url'], log=_log, prefix=False)
 
+        # post labels
+        pr = data['html_url'].split('/')[-1]
+        pr_url = g.repos[pr_target_account][pr_target_repo].issues[pr]
+        status, data = pr_url.labels.post(body=labels)
+        if not status == HTTP_STATUS_OK:
+            raise EasyBuildError("Failed to add labels to PR# %s; status %s, data: %s", pr, status, data)
+
+        print_msg("Added labels %s to PR#%s" % (labels, pr), log=_log, prefix=False)
```
boegel: Whole block should be guarded with `if labels:`?

PR #2385, opened 2018-01-29
experimental support for --continue

migueldiascosta:
in case anyone else is interested

my use case is for users to be able to debug/devel long compilations without rebuilding everything from scratch (often they are changing only one file in a large package)

they would run e.g. `eb --disable-cleanup-builddir`, change something in the build dir and then run `eb --continue=build --experimental`, eventually multiple times

there may be other use cases

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -70,8 +70,8 @@ from easybuild.tools.environment import restore_env, sanitize_env
 from easybuild.tools.filetools import CHECKSUM_TYPE_MD5, CHECKSUM_TYPE_SHA256
 from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file, change_dir, convert_name
 from easybuild.tools.filetools import compute_checksum, copy_file, derive_alt_pypi_url, diff_files, download_file
-from easybuild.tools.filetools import encode_class_name, extract_file, is_alt_pypi_url, mkdir, move_logs, read_file
-from easybuild.tools.filetools import remove_file, rmtree2, verify_checksum, weld_paths, write_file
+from easybuild.tools.filetools import encode_class_name, extract_file, find_base_dir, is_alt_pypi_url, mkdir, move_logs
+from easybuild.tools.filetools import read_file, remove_file, rmtree2, verify_checksum, weld_paths, write_file
 from easybuild.tools.hooks import BUILD_STEP, CLEANUP_STEP, CONFIGURE_STEP, EXTENSIONS_STEP, FETCH_STEP, INSTALL_STEP
 from easybuild.tools.hooks import MODULE_STEP, PACKAGE_STEP, PATCH_STEP, PERMISSIONS_STEP, POSTPROC_STEP, PREPARE_STEP
 from easybuild.tools.hooks import READY_STEP, SANITYCHECK_STEP, SOURCE_STEP, TEST_STEP, TESTCASES_STEP, run_hook
@@ -802,8 +802,12 @@ class EasyBlock(object):
 
         builddir = os.path.join(os.path.abspath(build_path()), clean_name, self.version, lastdir)
 
+        if build_option('continue'):
+            self.log.debug("Setting cleanupoldbuild to False in continue mode")
+            self.cfg['cleanupoldbuild'] = False
+
         # make sure build dir is unique if cleanupoldbuild is False or not set
-        if not self.cfg.get('cleanupoldbuild', False):
+        if not self.cfg.get('cleanupoldbuild', False) and not build_option('continue'):
             uniq_builddir = builddir
             suff = 0
             while(os.path.isdir(uniq_builddir)):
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2667,6 +2678,18 @@ def build_and_install_one(ecdict, init_env, hooks=None):
         _log.debug("Stop set to %s" % stop)
         app.cfg['stop'] = stop
 
+    continue_from = build_option('continue')
+    if continue_from is not None:
```
akesandgren: if continue_from: should be enough and is easier to read.

easybuild/framework/easyblock.py
```diff
@@ -2667,6 +2678,18 @@ def build_and_install_one(ecdict, init_env, hooks=None):
...
+        _log.experimental("Continuing from step %s" % continue_from)
+        skipsteps = []
+        for step in build_option('valid_stops'):
```
akesandgren: I haven't checked, but is build_option('valid_stops') guaranteed to be in the correct order?

easybuild/framework/easyblock.py
```diff
@@ -802,8 +802,12 @@ def gen_builddir(self):
 
         builddir = os.path.join(os.path.abspath(build_path()), clean_name, self.version, lastdir)
 
+        if build_option('continue'):
+            self.log.debug("Setting cleanupoldbuild to False in continue mode")
+            self.cfg['cleanupoldbuild'] = False
+
         # make sure build dir is unique if cleanupoldbuild is False or not set
```
boegel: update comment, also mention `or if --continue is used` (or just drop the comment, maybe...)

easybuild/tools/options.py
```diff
@@ -252,6 +252,8 @@ def basic_options(self):
         opts = OrderedDict({
             'dry-run': ("Print build overview incl. dependencies (full paths)", None, 'store_true', False),
             'dry-run-short': ("Print build overview incl. dependencies (short paths)", None, 'store_true', False, 'D'),
+            'continue': ("Continue the installation from certain step",
```
boegel: reword to `from specified step`?

easybuild/tools/options.py
```diff
@@ -252,6 +252,8 @@ def basic_options(self):
...
+                     'choice', 'store_or_None', SOURCE_STEP, 'c', all_stops),
```
boegel: Not sure if we want `eb -c` to be equivalent with `eb --continue`, maybe we should save up `-c` for something else?

It seems to me that `--continue` will always be a bit 'risky' (it may fail for all kinds of reasons that are going to be hard to fix)...

easybuild/framework/easyblock.py
```diff
@@ -2667,6 +2678,18 @@ def build_and_install_one(ecdict, init_env, hooks=None):
...
+    if continue_from is not None:
+        _log.experimental("Continuing from step %s" % continue_from)
+        skipsteps = []
+        for step in build_option('valid_stops'):
+            if step != continue_from:
+                _log.debug("Adding %s to skipsteps in continue mode" % step)
+                skipsteps.append(step)
+            else:
+                break
+        app.cfg['skipsteps'] = skipsteps
```
boegel: This may result in the `skipsteps` being added to the saved easyconfig file, which doesn't seem like a good idea to me...

Maybe introduce a `self.steps_to_skip`, which can be used both for `skipsteps` and `--continue`?

easybuild/framework/easyblock.py
boegel: Yes, it will be in the correct execution order. The list in `valid_stops` is in the same order as `get_steps`.

Maybe it's worth the trouble to add an explicit test for this if we really on that, doesn't hurt...

easybuild/framework/easyblock.py
```diff
@@ -2231,7 +2242,7 @@ def cleanup_step(self):
         except when we're building in the installation directory or
         cleanup_builddir is False, otherwise we remove the installation
         """
-        if not self.build_in_installdir and build_option('cleanup_builddir'):
+        if not self.build_in_installdir and build_option('cleanup_builddir') and not build_option('continue'):
```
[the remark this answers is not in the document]
migueldiascosta: well, if I properly interpret the intentions of my past self, the idea of `--continue` was in the context of a debug workflow, so it is assumed that one may need to continue to use `--continue` (pun intended)

once final, the changes should always be put in a patch and everything built from scratch.

but I have to say that my users have simply been debugging using `buildenv` or `--dump-env-script` on top of a manually prepared build dir, it's not exactly the same as this would allow but it fits most cases, so not sure if this is worth the trouble...

bryank-cs:
While it's nice to be able to start from a particular step, I don't see that the patch validates the previous steps are completed successfully. Gentoo's Portage systems creates files like .unpacked, .configured, etc to mark that a step has completed correctly. In this case, you could even just specify --continue to have it try again from wherever it left off, although I do like the idea of having an optional argument in case you want to back up a step (like reapply a patch).

PR #2387, opened 2018-01-31
Allow basic compiler modulenames to be specified as keys in optarch.

bartoldeman:
Examples: 'GCCcore', 'icc', 'ifort'. They will be considered before
the compiler family ('Intel', etc).

### DIFF
--- a/easybuild/tools/toolchain/compiler.py
+++ b/easybuild/tools/toolchain/compiler.py
@@ -280,31 +280,29 @@ class Compiler(Toolchain):
         :param default_optarch: default value to use for optarch, rather than using default value based on architecture
                                 (--optarch and --optarch=GENERIC still override this value)
         """
-        use_generic = False
         optarch = build_option('optarch') 
         # --optarch is specified with flags to use
+        if optarch is not None and isinstance(optarch, dict):
+            # optarch has been validated as complex string with multiple compilers and converted to a dictionary
+            # first try module names, then the family in optarch
+            current_compiler_names = (getattr(self, 'COMPILER_MODULE_NAME', []) +
+                                      [getattr(self, 'COMPILER_FAMILY', None)])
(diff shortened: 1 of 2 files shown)


bartoldeman:
@damianam this is pretty much doing what we discussed, considering the compiler module names as keys before the family.

easybuild/tools/toolchain/compiler.py
```diff
@@ -291,14 +291,17 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+                                          [getattr(self, 'COMPILER_FAMILY', None)])
+                for current_compiler in current_compiler_names:
+                    if current_compiler in optarch:
                         optarch = optarch[current_compiler]
+                        break
+                if optarch == OPTARCH_GENERIC:
+                    use_generic = True
                 # no option for this compiler
-                else:
+                if isinstance(optarch, dict):
```
damianam: I find it odd the recheck for `optarch` being a `dict`. It makes sense, but I'd clarify it in the comment above. Something like:

`# optarch hasn't changed, it is still a dict, which means that there is no option for this compiler`

easybuild/tools/toolchain/compiler.py
```diff
@@ -291,14 +291,17 @@ def _set_optimal_architecture(self, default_optarch=None):
 
             # optarch has been validated as complex string with multiple compilers and converted to a dictionary
             elif isinstance(optarch, dict):
-                current_compiler = getattr(self, 'COMPILER_FAMILY', None)
-                if current_compiler in optarch:
-                    if optarch[current_compiler] == OPTARCH_GENERIC:
-                        use_generic = True
-                    else:
+                # first try module names, than the family in optarch
```
damianam: then, not than.

damianam:
lgtm!

PR #2391, opened 2018-02-01
support depends_on "load" statements in generated modules via --module-depends-on and module_depends_on easyconfig parameter

bartoldeman:
edit: see http://lmod.readthedocs.io/en/latest/098_dependent_modules.html


bartoldeman:
Ah Travis can't download env modules 4.0:
Resolving prdownloads.sourceforge.net (prdownloads.sourceforge.net)... 52.85.131.153, 52.85.131.22, 52.85.131.98, ...
Connecting to prdownloads.sourceforge.net (prdownloads.sourceforge.net)|52.85.131.153|:80... connected.
HTTP request sent, awaiting response... 403 Forbidden
2018-03-01 02:04:34 ERROR 403: Forbidden.

we'll try again later then...

boegel:
@bartoldeman Seems like SourceForge was having trouble for a while, seems to be fixed now, I've re-triggered the tests.

boegel:
@bartoldeman Please sync with current `develop` and fix the merge conflict on `.travis.yml` (should be trivial to resolve)?

easybuild/framework/easyconfig/default.py
```diff
@@ -172,6 +172,7 @@
     'moduleforceunload': [False, 'Force unload of all modules when loading the extension', MODULES],
     'moduleloadnoconflict': [False, "Don't check for conflicts, unload other versions instead ", MODULES],
     'recursive_module_unload': [False, 'Recursive unload of all dependencies when unloading module', MODULES],
+    'recursive_module_unload_depends_on' : [False, 'Use depends_on (Lmod 7.6.1+) for dependencies in generated module', MODULES],
```
boegel: @bartoldeman Do you think we need this both as configuration option (`--recursive-module-unload-depends-on`) as well as an easyconfig parameter (`recursive_module_unload_depends_on`)?

What use case do you see for latter, i.e. selectively enabling the use of `depends_on` rather than setting it in the global configuration?

easybuild/tools/options.py
```diff
@@ -474,6 +474,8 @@ def config_options(self):
                        None, 'store', None),
             'recursive-module-unload': ("Enable generating of modules that unload recursively.",
                                         None, 'store_true', False),
+            'recursive-module-unload-depends-on': ("Use depends_on (Lmod 7.6.1+) for dependencies in all generated modules.",
```
boegel: Should this just be `use-depends-on`, and then have the help message clarify that this implies that it makes unloading of modules recursive?

easybuild/tools/modules.py
```diff
@@ -209,8 +209,11 @@ def modules(self):
 
     def set_and_check_version(self):
         """Get the module version, and check any requirements"""
+        lmod_depends_on_min = '7.6.1'
         if self.COMMAND in MODULE_VERSION_CACHE:
             self.version = MODULE_VERSION_CACHE[self.COMMAND]
+            self.has_depends_on = (isinstance(self, Lmod) and
```
boegel: I'd go with `supports_depends_on` rather than `has_depends_on`

`supports_depends_on` should be set to `False` in `ModulesTool.__init__`

and then in `Lmod.__init__` it should be set based on the Lmod version after the `super` call (don't use `isinstance(self, Lmod)`)

easybuild/tools/modules.py
```diff
@@ -259,6 +262,8 @@ def set_and_check_version(self):
                 self.log.debug('Version %s matches requirement <= %s', self.version, self.MAX_VERSION)
 
         MODULE_VERSION_CACHE[self.COMMAND] = self.version
+        self.has_depends_on = (isinstance(self, Lmod) and
+                               StrictVersion(self.version) >= StrictVersion(lmod_depends_on_min))
```
boegel: You're setting this twice (but it should be moved to the `Lmod` class anyway)

easybuild/framework/easyblock.py
```diff
@@ -1042,12 +1044,15 @@ def make_module_dep(self, unload_info=None):
         self.log.debug("List of retained deps to load in generated module: %s", deps)
 
         # include load statements for retained dependencies
+        recursive_unload = self.cfg['recursive_module_unload']
+        if self.cfg['recursive_module_unload_depends_on']:
+            recursive_unload = 'depends_on'
```
boegel: rather than introducing a magic value for the `recursive_unload` named argument of `load_module`, why not add a new named argument `use_depends_on`?

easybuild/tools/module_generator.py
```diff
@@ -505,9 +506,15 @@ def load_module(self, mod_name, recursive_unload=False, unload_modules=None):
         body = []
         if unload_modules:
             body.extend([self.unload_module(m).strip() for m in unload_modules])
-        body.append(self.LOAD_TEMPLATE)
-
-        if build_option('recursive_mod_unload') or recursive_unload:
+        load_template = self.LOAD_TEMPLATE
+        # Lmod 7.6.1+ supports depends-on which does this most nicely:
+        if (build_option('recursive_mod_unload_depends_on') or
+            recursive_unload == 'depends_on') and modules_tool().has_depends_on:
```
boegel: if `depends_on` is requested but not supported (`supports_depends_on` is `False`), then we should get an error rather than silently ignoring that request?

easybuild/framework/easyblock.py
boegel: Also, you should take the `build_option` into account here (rather than in `load_module`)?

easybuild/tools/module_generator.py
boegel: Taking into account the above, this should become something like:

```python
load_template = self.LOAD_TEMPLATE
if use_depends_on:
    if modules_tool().supports_depends_on:
        load_template = self.LOAD_TEMPLATE_DEPENDS_ON
    else:
        raise EasyBuildError("depends_on statements in generated module are not supported by modules tool")

File: test/framework/options.py [hunk 7]
```diff
@@ -1156,23 +1156,27 @@ def test_recursive_module_unload(self):
...
-
-        toy_module = os.path.join(self.test_installpath, 'modules', 'all', 'toy', '0.0-deps')
-        if get_module_syntax() == 'Lua':
-            toy_module += '.lua'
-        toy_module_txt = read_file(toy_module)
-        is_loaded_regex = re.compile(r"if { !\[is-loaded gompi/1.3.12\] }", re.M)
-        self.assertFalse(is_loaded_regex.search(toy_module_txt), "Recursive unloading is used: %s" % toy_module_txt)
+        lastargs = ['--recursive-module-unload']
+        if self.modtool.has_depends_on:
+            lastargs.append(lastargs[0]+'-depends-on')
```
boegel: We should not stop testing `--recursive-module-unload` with recent Lmod versions?

bartoldeman: With this change it tests both, but I renamed the option now.

easybuild/framework/easyconfig/default.py
bartoldeman: i will check... this simply mirrors recursive_module_unload.
If I check the history it looks like the people in Juelich wanted to only have recursive unloading enabled for toolchain modules, and not for other modules, to avoid loadstorms. I will double check how depends_on behaves for us on toolchain modules (we hide them but obviously easybuild uses them itself).

bartoldeman:
I addressed all comments I believe, except for the easyconfig parameter. I'll double-check depends_on behaviour in toolchain modules before I remove that (see my comment in the review).

bartoldeman:
One could argue that since #2316 the `recursive_module_unload` easyconfig parameter is no longer needed since the global config no longer causes load storms.

depends_on() seems to work ok. I get funny module lists if I load foss after iomkl but that could be fixed by using e.g. `family('linalg')` in  the OpenBLAS and imkl modules, and `family('toolchain')` in the toolchain modules. This is not a very common use case though since EB itself doesn't do it, and `module swap` works ok.

easybuild/framework/easyconfig/default.py
```diff
@@ -171,6 +171,8 @@
     'moduleclass': ['base', 'Module class to be used for this software', MODULES],
     'moduleforceunload': [False, 'Force unload of all modules when loading the extension', MODULES],
     'moduleloadnoconflict': [False, "Don't check for conflicts, unload other versions instead ", MODULES],
+    'module_use_depends_on' : [False, 'Use depends_on (Lmod 7.6.1+) for dependencies in generated module '
```
boegel: @bartoldeman I would avoid `module_use`, since that has a particular meaning (not at all related to `depends_on`).

How about `module_depends_on` or `use_depends_on`?

easybuild/tools/config.py
```diff
@@ -386,6 +387,7 @@ def init_build_options(build_options=None, cmdline_options=None):
             'check_osdeps': not cmdline_options.ignore_osdeps,
             'dry_run': cmdline_options.dry_run or cmdline_options.dry_run_short,
             'recursive_mod_unload': cmdline_options.recursive_module_unload,
+            'mod_use_depends_on': cmdline_options.module_use_depends_on,
```
boegel: same here, avoid `mod(ule)_use`, since it has a very particular meaning

easybuild/tools/module_generator.py
```diff
@@ -505,9 +506,16 @@ def load_module(self, mod_name, recursive_unload=False, unload_modules=None):
...
+        load_template = self.LOAD_TEMPLATE
+        # Lmod 7.6.1+ supports depends-on which does this most nicely:
+        if build_option('mod_use_depends_on') or use_depends_on:
+            if not modules_tool().supports_depends_on:
+                raise EasyBuildError("depends_on statements in generated module are not supported by modules tool")
+            load_template = self.LOAD_TEMPLATE_DEPENDS_ON
+        body.append(load_template)
+
+        if (build_option('recursive_mod_unload') or recursive_unload or
+            load_template == self.LOAD_TEMPLATE_DEPENDS_ON):
```
boegel: no need to wrap this across two lines? looks confusing imho...

easybuild/tools/module_generator.py
```diff
@@ -795,19 +804,29 @@ def load_module(self, mod_name, recursive_unload=False, unload_modules=None):
...
+                raise EasyBuildError("depends_on statements in generated module are not supported by modules tool")
+            load_template = self.LOAD_TEMPLATE_DEPENDS_ON
+
+        body.append(load_template)
+        if load_template == self.LOAD_TEMPLATE_DEPENDS_ON:
+            load_statement = body + ['']
         else:
-            load_guard = 'isloaded("%(mod_name)s")'
-        load_statement = [self.conditional_statement(load_guard, '\n'.join(body), negative=True)]
+            if build_option('recursive_mod_unload') or recursive_unload:
```
boegel: make this `elif`?

easybuild/tools/options.py
```diff
@@ -456,6 +456,9 @@ def config_options(self):
             'module-naming-scheme': ("Module naming scheme to use", None, 'store', DEFAULT_MNS),
             'module-syntax': ("Syntax to be used for module files", 'choice', 'store', DEFAULT_MODULE_SYNTAX,
                               sorted(avail_module_generators().keys())),
+            'module-use-depends-on': ("Use depends_on (Lmod 7.6.1+) for dependencies in all generated modules "
```
boegel: same here, I'd go with either `--module-depends-on` or `--use-depends-on`

test/framework/module_generator.py
```diff
@@ -258,6 +258,17 @@ def test_load(self):
...
+            # Lmod 7.6+ depends-on
+            if self.modtool.supports_depends_on:
+                expected = '\n'.join([
+                    '',
+                    "depends-on mod_name",
+                    '',
+                ])
+                self.assertEqual(expected, self.modgen.load_module("mod_name", use_depends_on=True))
+                init_config(build_options={'mod_use_depends_on': 'True'})
+                self.assertEqual(expected, self.modgen.load_module("mod_name"))
```
boegel: also check the other case, with `self.assertErrorRegex`?

easybuild/tools/module_generator.py
bartoldeman: No, it really needs to be `else:` because of the common `load_statement =` statement at the bottom.

boegel: OK, I overlooked that indeed, thanks or clarifying

boegel:
lgtm, thanks for the effort @bartoldeman!

PR #2392, opened 2018-02-01
Initial set of OHPC module meta data for EasyBuild

crbaird:
Signed-off-by: Reese Baird <reese.baird@intel.com>

### DIFF
--- a/etc/ohpc_external_modules_metadata.cfg
+++ b/etc/ohpc_external_modules_metadata.cfg
@@ -1,0 +1,225 @@
+# metadata useful to EasyBuild (easybuilders.github.io/easybuild) for modules provided by OpenHPC (github.com/openhpc/ohpc)
+
+[R/3.4.2]
+name = R
+version = 3.4.2
+prefix = R_DIR
+
+[adios/1.12.0]
+name = adios
+version = 1.12.0
+prefix = ADIOS_DIR
+
+[autotools]
+name = Autoconf, Automake, Libtool
+version = 2.69, 1.15, 2.4.6
+prefix = /opt/ohpc/pub/utils/autotools/bin
+
+[boost/1.65.1]
+name = Boost
+version = 1.65.1
+prefix = BOOST_DIR
+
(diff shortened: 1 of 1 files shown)


boegel:
@crbaird This looks great, thanks a look for looking into this!
(upstream ref: https://github.com/openhpc/ohpc/issues/110)

I'm happy to merge this as is, but I'd like to get a better idea on how we move forward from here first...

Do we consider this to be sufficient, or do we want to take this a step further, e.g. by defining custom `ohpc` toolchains in EasyBuild that could leverage some of these external modules?
For example, `gnu` +  `openmpi` + `openblas` + `fftw` could form `foss-ohpc` with it's own versioning scheme, probably tied to the OpenHPC release version.

And that could be taken even a step further, since EasyBuild also supports creating RPM packages by leveraging FPM (see http://easybuild.readthedocs.io/en/latest/Packaging_support.html).

Also, how was this metadata file created, how can we ensure this is easy to update in the future for upcoming `OpenHPC` releases?

One thing to keep in mind is that since we don't know which OpenHPC version EasyBuild will be used on, we should avoid blowing away old modules in this metadata file...

crbaird:
I think OHPC toolchains would be great, it just seemed like the module metadata was the lowest hanging fruit. I can take another look at what would be required for that.

There was some human intervention in creation of this version. The OHPC module/version pairs are easy to generate programmatically, but I didn't see a 1-1 mapping to the EasyBuild software names. Knowing that this file might be useful I can put some more effort in to automation, as well as adding in older modules.

boegel:
@crbaird How shall we proceed here? Shall I include this as is, and revisit this later with upcoming OpenHPC releases?

I think it's important to try and come up with an automated process to update this file, otherwise it's going to be a maintenance nightmare, which will boil down to the file going stale I think...

boegel:
@crbaird I see no reason not to include this, it's a good starting point. I'm merging this in, will be included with the next EasyBuild release (v3.5.2).

PR #2397, opened 2018-02-09
Avoid EasyBuild crashing hard when it can't find glibc on Linux

alainvanhoof:
Not all Linux system use glibc.

edit: see also https://github.com/easybuilders/easybuild/issues/410

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -641,8 +641,10 @@ def get_glibc_version():
             _log.debug("Found glibc version %s" % glibc_version)
             return glibc_version
         else:
-            raise EasyBuildError("Failed to determine glibc version from '%s' using pattern '%s'.",
-                                 glibc_ver_str, glibc_ver_regex.pattern)
+            # no glibc on this Linux
+            _log.warning("No glibc on this Linux system, so can't determine version.")
+            return UNKNOWN
+        
     else:
         # no glibc on OS X standard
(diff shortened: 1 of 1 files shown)


boegel:
@alainvanhoof All pull requests should target the `develop` branch, I changed this for this particular PR. See also http://easybuild.readthedocs.io/en/latest/Contributing.html#pull-requests for more information.

easybuild/tools/systemtools.py
```diff
@@ -641,8 +641,10 @@ def get_glibc_version():
             _log.debug("Found glibc version %s" % glibc_version)
             return glibc_version
         else:
-            raise EasyBuildError("Failed to determine glibc version from '%s' using pattern '%s'.",
-                                 glibc_ver_str, glibc_ver_regex.pattern)
+            # no glibc on this Linux
+            _log.warning("No glibc on this Linux system, so can't determine version.")
```
boegel: Changing from error to warning makes sense, there's little point in making this fatal.

We should mention the regex pattern that was used to try and find the `glibc` version though, no need to drop that info...

boegel:
A more thorough alternative for this: #2398

boegel:
Fixed in #2398, thanks again for reporting this issue @alainvanhoof!

alainvanhoof:
Thank you for allowing me to play a part, and sorry I added the pull request to master and not development.

Cheers,
Alain van Hoof

PR #2400, opened 2018-02-12
add support for --list-prs (REVIEW)

migueldiascosta:


easybuild/main.py
```diff
@@ -284,6 +284,11 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     elif options.install_github_token:
         install_github_token(options.github_user, silent=build_option('silent'))
 
+    elif options.list_prs:
+
```
boegel: drop empty line?

easybuild/main.py
```diff
@@ -284,6 +284,11 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+        parameters = {'state': options.list_prs_state, 'sort': options.list_prs_sort, 'direction': options.list_prs_direction}
```
boegel:
```python
parameters = {
    'direction': options.list_prs_direction,
    'sort': options.list_prs_sort,
    'state': options.list_prs_state,
}

File: easybuild/tools/options.py [hunk 3]
```diff
@@ -562,6 +562,10 @@ def github_options(self):
             'github-user': ("GitHub username", str, 'store', None),
             'github-org': ("GitHub organization", str, 'store', None),
             'install-github-token': ("Install GitHub token (requires --github-user)", None, 'store_true', False),
+            'list-prs': ("List pull requests", None, 'store_true', None),
+            'list-prs-state': ("List pull requests in state open/closed", str, 'store', 'open'),
+            'list-prs-sort': ("Sort pull requests by created/updated/popularity", str, 'store', 'created'),
+            'list-prs-direction': ("Sorting pull requests in asc/desc order", str, 'store', 'desc'),
```
boegel: Also `choice` option? Maybe `list-prs-order` is better?

easybuild/tools/options.py
```diff
@@ -562,6 +562,10 @@ def github_options(self):
             'github-user': ("GitHub username", str, 'store', None),
             'github-org': ("GitHub organization", str, 'store', None),
             'install-github-token': ("Install GitHub token (requires --github-user)", None, 'store_true', False),
+            'list-prs': ("List pull requests", None, 'store_true', None),
+            'list-prs-state': ("List pull requests in state open/closed", str, 'store', 'open'),
```
boegel: Make this a `choice` option, anything other than `open` or `close` doesn't make sense?

easybuild/tools/options.py
```diff
@@ -562,6 +562,10 @@ def github_options(self):
...
+            'list-prs-sort': ("Sort pull requests by created/updated/popularity", str, 'store', 'created'),
```
boegel: Also `choice` option?

easybuild/tools/github.py
```diff
@@ -958,6 +958,28 @@ def not_eligible(msg):
     return res
 
 
+def list_prs(parameters):
+    """
+    List PRs
+    
+    :param parameters: https://developer.github.com/v3/pulls/#parameters
```
boegel: `Selection parameters for PRs, see  https://developer.github.com/v3/pulls/#parameters`

easybuild/tools/github.py
```diff
@@ -958,6 +958,28 @@ def not_eligible(msg):
...
+    parameters.update({'per_page': GITHUB_MAX_PER_PAGE})
+
+    pr_target_account = build_option('pr_target_account')
+    pr_target_repo = build_option('pr_target_repo')
+
+    pr_url = lambda g: g.repos[pr_target_account][pr_target_repo].pulls
+    status, pr_data = github_api_get_request(pr_url, None, **parameters)
+    if status != HTTP_STATUS_OK:
+        raise EasyBuildError("Failed to get data for PR #%d from %s/%s (status: %d %s)",
+                             pr, pr_target_account, pr_target_repo, status, pr_data)
```
boegel: `pr` is undefined here? You should include `parameters` in the error message?

boegel:
@migueldiascosta Can you look into adding a test too?

migueldiascosta:
@boegel using the actual PRs as expected output? Hm, I suppose the output of
``` python
list_prs({'direction': 'asc',  'sort': 'created', 'state': 'closed'})
```
should be constant...

migueldiascosta:
ah, the tests use `hpcugent/testrepository`, ok. 

I'll submit and close a PR to it so it can be tested.

easybuild/tools/options.py
```diff
@@ -562,6 +566,10 @@ def github_options(self):
             'github-user': ("GitHub username", str, 'store', None),
             'github-org': ("GitHub organization", str, 'store', None),
             'install-github-token': ("Install GitHub token (requires --github-user)", None, 'store_true', False),
+            'list-prs': ("List pull requests", None, 'store_true', None),
+            'list-prs-state': ("State for listing pull requests", 'choice', 'store', 'open', github_list_pr_states),
+            'list-prs-order': ("Order for sorting pull requests", 'choice', 'store', 'created', github_list_pr_orders),
+            'list-prs-direction': ("Direction for sorting pull requests", 'choice', 'store', 'desc', github_list_pr_directions),
```
boegel: @migueldiascosta Do you think it makes sense to make this a single option, that accepts three values (optionally) for state/order/direction?

Something like:

```
eb --list-prs  # list all open PRs, order by created (desc.)
eb --list-prs closed  # list all closed PRs, default order
eb --list-prs all:updated  # list all PRs, order by updated (default ordering)
eb --list-prs open:updated:asc

File: easybuild/tools/github.py [hunk 9]
```diff
@@ -958,6 +958,32 @@ def not_eligible(msg):
     return res
 
 
+def list_prs(parameters):
+    """
+    List PRs
```
boegel: I'd reword this to `List pull requests according to specified selection/order parameters`

easybuild/tools/options.py
```diff
@@ -550,6 +550,10 @@ def informative_options(self):
 
     def github_options(self):
         """GitHub integration configuration options."""
+        github_list_pr_states = ['open', 'closed', 'all']
+        github_list_pr_orders = ['created', 'updated', 'popularity', 'long-running']
+        github_list_pr_directions = ['asc', 'desc']
```
boegel: Should we use constants for these (especially since we reuse some of these strings for default settings)?

If so, they should be defined in `tools/github.py` and imported here (since the values are determined by the GitHub API, not by us).

easybuild/tools/options.py
```diff
@@ -562,6 +563,9 @@ def github_options(self):
             'github-user': ("GitHub username", str, 'store', None),
             'github-org': ("GitHub organization", str, 'store', None),
             'install-github-token': ("Install GitHub token (requires --github-user)", None, 'store_true', False),
+            'list-prs': ("List pull requests", str, 'store_or_None', 
+                         ":".join([GITHUB_LIST_PR_STATES[0], GITHUB_LIST_PR_ORDERS[0], GITHUB_LIST_PR_DIRECTIONS[1]]),
```
boegel: @migueldiascosta I would avoid using indices in a list of constants... Instead, use a dedicated constant for the default value (and use that constant when defining the list of acceptable values)?

I would say `,` makes more sense as a separator (`:` is used for `PATH`-like variables).

easybuild/tools/options.py
```diff
@@ -813,6 +821,29 @@ def _postprocess_optarch(self):
...
 
+    def _postprocess_list_prs(self):
+        """Postprocess --list-prs options"""
+        list_prs_parts = self.options.list_prs.split(':')
+        nparts = len(list_prs_parts)
+
+        if nparts == 0:
+            list_prs_parts = [GITHUB_LIST_PR_STATES[0], GITHUB_LIST_PR_ORDERS[0], GITHUB_LIST_PR_DIRECTIONS[1]]
+        elif nparts == 1:
+            list_prs_parts.extend([GITHUB_LIST_PR_ORDERS[0], GITHUB_LIST_PR_DIRECTIONS[1]])
```
boegel: another good reason to use dedicated constants for the default values imho, that'll make this more readable :)

easybuild/tools/options.py
```diff
@@ -813,6 +821,29 @@ def _postprocess_optarch(self):
...
+            raise EasyBuildError("Argument to --list-prs must be in the format state:order:direction")
+
+        if list_prs_parts[0] not in GITHUB_LIST_PR_STATES:
+            raise EasyBuildError("First element of --list-prs must be on of %s" % GITHUB_LIST_PR_STATES)
+        if list_prs_parts[1] not in GITHUB_LIST_PR_ORDERS:
+            raise EasyBuildError("Second element of --list-prs must be on of %s" % GITHUB_LIST_PR_ORDERS)
+        if list_prs_parts[2] not in GITHUB_LIST_PR_DIRECTIONS:
+            raise EasyBuildError("Third element of --list-prs must be on of %s" % GITHUB_LIST_PR_DIRECTIONS)
+
+        self.options.list_prs = list_prs_parts
```
boegel: @migueldiascosta Rather than using `list_prs_parts`, I would create dedicated local constants, and then compose those at the end? And I would use a 3-tuple rather than a list (since it doesn't make sense to append more values to it, should always be exactly three).

```
self.options.list_prs = (list_pr_state, list_pr_order, list_pr_direc)

File: test/framework/github.py [hunk 14]
```diff
@@ -114,6 +114,25 @@ def test_read(self):
...
+
+        parameters = {'state': 'closed', 'sort': 'created', 'direction': 'asc', 'per_page': 1}
+
+        init_config(build_options={'pr_target_account': GITHUB_USER,
+                                   'pr_target_repo': GITHUB_REPO})
+
+        expected = "PR #1: a pr"
+
+        try:
+            output = gh.list_prs(parameters)
```
boegel: Given all the parsing magic we have, I would test `--list-prs` directly via a test in `test/framework/options.py`?

easybuild/tools/options.py
```diff
@@ -780,6 +784,10 @@ def postprocess(self):
         if self.options.optarch and not self.options.job:
             self._postprocess_optarch()
 
+        # make sure --list-format has a valid format
```
boegel: what's `--list-format`?

easybuild/main.py
```diff
@@ -284,6 +284,15 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
         install_github_token(options.github_user, silent=build_option('silent'))
 
+    elif options.list_prs:
+        parameters = {
+            'direction': options.list_prs[2],
+            'sort': options.list_prs[1],
+            'state': options.list_prs[0],
+            }
+        print_msg("Listing PRs with parameters %s" % parameters)
+        print list_prs(parameters)
```
boegel: Just pass down `options.list_prs` to `list_prs`, and unpack there?

`main.py` is big enough as is, let's not make it bigger without good reason :)

boegel:
Thanks @migueldiascosta!

PR #2404, opened 2018-02-14
Keep the current timestamp when extracting files

jdonners:
The code adds an option `--keep-current-timestamp` to extract files without restoring the original timestamp. This is useful on a scratch filesystem that automatically cleans files older than a few weeks.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -122,28 +122,28 @@ CHECKSUM_TYPES = sorted(CHECKSUM_FUNCTIONS.keys())
 
 EXTRACT_CMDS = {
     # gzipped or gzipped tarball
-    '.gtgz':    "tar xzf %(filepath)s",
+    '.gtgz':    "tar x%(timestamp_tar)szf %(filepath)s",
     '.gz':      "gunzip -c %(filepath)s > %(target)s",
-    '.tar.gz':  "tar xzf %(filepath)s",
-    '.tgz':     "tar xzf %(filepath)s",
+    '.tar.gz':  "tar x%(timestamp_tar)szf %(filepath)s",
+    '.tgz':     "tar x%(timestamp_tar)szf %(filepath)s",
     # bzipped or bzipped tarball
     '.bz2':     "bunzip2 -c %(filepath)s > %(target)s",
-    '.tar.bz2': "tar xjf %(filepath)s",
-    '.tb2':     "tar xjf %(filepath)s",
-    '.tbz':     "tar xjf %(filepath)s",
-    '.tbz2':    "tar xjf %(filepath)s",
+    '.tar.bz2': "tar x%(timestamp_tar)sjf %(filepath)s",
+    '.tb2':     "tar x%(timestamp_tar)sjf %(filepath)s",
+    '.tbz':     "tar x%(timestamp_tar)sjf %(filepath)s",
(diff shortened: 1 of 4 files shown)


easybuild/tools/filetools.py
```diff
@@ -768,7 +768,14 @@ def extract_cmd(filepath, overwrite=False):
         if 'unzip -qq' in cmd_tmpl:
             cmd_tmpl = cmd_tmpl.replace('unzip -qq', 'unzip -qq -o')
 
-    return cmd_tmpl % {'filepath': filepath, 'target': target}
+    if build_option('keep_current_timestamp'):
+         ts_tar='m'
```
boegel: @jdonners Please use the long option `--touch` rather than `(-)m`, since it's an uncommon option to use with `tar`.

easybuild/tools/filetools.py
```diff
@@ -122,28 +122,28 @@
 
 EXTRACT_CMDS = {
     # gzipped or gzipped tarball
-    '.gtgz':    "tar xzf %(filepath)s",
+    '.gtgz':    "tar x%(timestamp_tar)szf %(filepath)s",
```
boegel: Hmm, this makes it a bit hard to read (and it won't work if you use the long option `--touch` instead, see below), so this should become `tar %(timestamp_opt_tar)sxzf %(filepath)s` (and make sure to use `--touch `, i.e. add a space at the end of the `ts_tar` value)

easybuild/tools/filetools.py
boegel: Also: include spaces around `=`, so `ts_tar = '--touch '`

easybuild/tools/filetools.py
```diff
@@ -122,28 +122,28 @@
...
-    '.txz':     "unxz %(filepath)s --stdout | tar x",
+    '.tar.xz':  "unxz %(filepath)s --stdout | tar x%(timestamp_tar)s",
+    '.txz':     "unxz %(filepath)s --stdout | tar x%(timestamp_tar)s",
     '.xz':      "unxz %(filepath)s",
     # tarball
-    '.tar':     "tar xf %(filepath)s",
+    '.tar':     "tar x%(timestamp_tar)sf %(filepath)s",
     # zip file
-    '.zip':     "unzip -qq %(filepath)s",
+    '.zip':     "unzip -qq %(timestamp_zip)s %(filepath)s",
```
boegel: Avoid double space by using `unzip -qq %(timestamp_zip)s%(filepath)s` and including the space in `ts_zip`

easybuild/tools/filetools.py
jdonners: well, the `--touch` option is specific to GNU tar, but wouldn't work for *BSDtar. So the long option would not work on OS/X, for example.

easybuild/tools/filetools.py
jdonners: see my remark about the long option below

jdonners:
hmm, probably this is a bad idea after all, judgin from the issues that my colleague Paul had (see https://github.com/easybuilders/easybuild-easyconfigs/issues/5644). So don't merge this one for now.

easybuild/tools/filetools.py
boegel: Ugh, ok, then it indeed makes sense to stick to using `-m`.

Issue #2402, opened 2018-02-12
combining multiple compiler wrappers may result in 'fork bomb'


boegel:
When multiple compiler wrappers are being used together, e.g. when `--rpath` or `--use-ccache` are used when building `TensorFlow` with the `intel` toolchain (which involves a wrapper to ensure `icc` still works when being called by Bazel, cfr. https://github.com/easybuilders/easybuild-easyblocks/pull/1287/files#diff-925e7942309e22e4c64cb2d99b50c206R77, one wrapper ends up calling the other wrapper, with neither calling out to the actual compiler, basically resulting in a fork bomb.

This is largely due to too naive wrappers script, which simply assume that the next compiler command available through `$PATH` is the correct one.

One possible solution for this would be to let the wrappers created by EasyBuild check whether the command being called is the actual compiler, i.e. a binary, rather than a wrapper script.

I know that SURFsara has run into similar issues when combining their 'manual' wrappers with `eb --rpath`, so thoughts on this @jdonners, @casparvl?

boegel:
Another option would be that the wrappers we generate filter their own location from `$PATH` before calling out to whichever compiler command is next in `$PATH`...

jdonners:
the second option is the right one, IMHO. That is what our compiler wrappers do as well. One issue with this option is that the compiler and linker wrappers would have to be in separate paths, because the compiler usually calls the linker, which should call the linker wrapper first to add the RPATH.

boegel:
Fix for wrapper used in `TensorFlow` easyblock when building with `icc` is in https://github.com/easybuilders/easybuild-easyblocks/pull/1361, fix for RPATH wrapper is in https://github.com/easybuilders/easybuild-framework/pull/2410

fgeorgatos:
Apologies for arriving late for the show (but you see, reading your email backlog during a flight has its advantages!)

Couldn’t this need by treated somewhat by hacking together a solution around Lmod’s StackedVariables concept?!

Don’t get me wrong, that might impose a single modules back-end but it comes with the benefit of ability to control the ordering in the stack of wrappers, which might in turn have its own semantic, and therefor use-cases!

F.

Sent from my iPhone

boegel:
@fgeorgatos The RPATH wrappers are put in place dynamically, as a part of setting up the build environment for the toolchain & dependencies. I don't think Lmod is of much help in that context.

Issue #1992, opened 2016-11-13
known issues with --rpath support


boegel:
* [ ] RPATH sanity check with default paths fails on binary installations (?)
* [ ] some easyblocks need to specialise the locations for the RPATH sanity check (e.g. OpenFOAM)
* [ ] locations in build directory and temporary directory should not get RPATH'ed

jdonners:
Are there more details about the remark 'some easyblocks need to specialise the locations for the RPATH sanity check (e.g. OpenFOAM)'? I have been able to install OpenFOAM 2.3.1 using the --rpath support without any tweaks. I could install OpenFOAM-4.0 using the --rpath support with an additional line to the easyconfig for the ParaView installation.

boegel:
@jdonners I created a new issue for the OpenFOAM issues at https://github.com/hpcugent/easybuild-easyblocks/issues/1193

PR #2409, opened 2018-02-20
declare support for RPATH linking stable

boegel:
Based on the feedback from @jdonners and @pescobar and the discussion we had during the EasyBuild User Meeting in Amsterdam (cfr. https://github.com/easybuilders/easybuild/wiki/3rd-EasyBuild-User-Meeting), the conclusion was that we can declare the RPATH support as stable.

There are a couple of outstanding issues (cfr. #1992), but none of these seem blockers to me; either they are nice-to-have features, or problems specific to a particular software package that can be handled case-by-case.

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -771,8 +771,6 @@ class Toolchain(object):
 
         :param rpath_filter_dirs: extra directories to include in RPATH filter (e.g. build dir, tmpdir, ...)
         """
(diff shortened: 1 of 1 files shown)


boegel:
I marked this as WIP since I'd like to fix #2402 first before declaring this stable (working on it)

boegel:
With ~~#2402~~ merged, OK to declare RPATH support stable? @jdonners @pescobar @wpoely86

wpoely86:
:+1:

jdonners:
:+1: :1st_place_medal:

boegel:
:shipit:

PR #2410, opened 2018-02-20
exclude location of RPATH wrappers from $PATH to avoid fork bomb

boegel:


boegel:
I tested this with a couple of builds under `--rpath --experimental`, didn't run into any surprises, so good to go @wpoely86?

Issue #2418, opened 2018-03-01
--update-pr issue with multilevel Fork structure


shahzebsiddiqui:
We have a fork of easybuild-easyconfig in organization PfizerRD and I have my fork from PfizerRD.

See https://github.com/PfizerRD/easybuild-easyconfigs and https://github.com/shahzebsiddiqui/easybuild-easyconfigs 

My git remotes are the following, and I am cloning from origin

```
[siddis14@amrndhl1157 DAAL]$ git remote -v
easybuilders    git@github.com:easybuilders/easybuild-easyconfigs.git (fetch)
easybuilders    git@github.com:easybuilders/easybuild-easyconfigs.git (push)
origin  git@github.com:shahzebsiddiqui/easybuild-easyconfigs.git (fetch)
origin  git@github.com:shahzebsiddiqui/easybuild-easyconfigs.git (push)
upstream        git@github.com:PfizerRD/easybuild-easyconfigs.git (fetch)
upstream        git@github.com:PfizerRD/easybuild-easyconfigs.git (push)
```

My github easybuild is configured to push to origin 

```
[siddis14@amrndhl1157 DAAL]$ eb --check-github
== temporary log file in case of crash /tmp/eb-h_dQfs/easybuild-UnX_RQ.log

Checking status of GitHub integration...

Making sure we're online...OK

* GitHub user...shahzebsiddiqui => OK
* GitHub token...625..998 (len: 40) => OK (validated)
* git command...OK ("git version 1.8.3.1; ")
* GitPython module...OK (GitPython version 2.1.5)
* push access to shahzebsiddiqui/easybuild-easyconfigs repo @ GitHub...OK
* creating gists...OK
* location to Git working dirs... not found (suboptimal)

All checks PASSed!

Status of GitHub integration:
* --from-pr: OK
* --new-pr: OK
* --review-pr: OK
* --update-pr: OK
* --upload-test-report: OK

```

I have a PR https://github.com/easybuilders/easybuild-easyconfigs/pull/4825 that was pushed from PfizerRD:eb-daal branch.

When I run 

```
eb --update-pr 4825 --pr-commit-msg "adding sha256 and fix few issues requested by @boegel" DAAL-2017.1.132.eb

```

[...]

PR #2419, opened 2018-03-01
fix target account for --update-pr in case it's different from GitHub account being used to push branch

boegel:
Sometimes `--github-user` is not the same as the account/organisation that was used to open a PR, cfr. bug report by @shahzebsiddiqui in #2418...

Given it's limited scope & impact I consider this a candidate for a last-minute inclusion in EasyBuild v3.5.2...


boegel:
tests fail because of SourceForge downtime, I propose we work around this by temporarily disabling those tests, cfr. #2420

easybuild/tools/github.py
```diff
@@ -769,15 +769,19 @@ def _easyconfigs_pr_common(paths, ecs, start_branch=None, pr_branch=None, target
...
     dry_run = build_option('dry_run') or build_option('extended_dry_run')
 
+    push_branch_msg = "pushing branch '%s' to remote '%s' (%s)" % (pr_branch, remote_name, github_url)
+    if dry_run:
+        push_branch_msg += ' [DRY RUN]'
+    print_msg(push_branch_msg)
+
     if not dry_run:
-        _log.debug("Pushing branch '%s' to remote '%s' (%s)", pr_branch, remote_name, github_url)
+        _log.debug(push_log_msg)
```
wpoely86: where is the definied? Don't you mean `push_branch_mg`?

boegel: Ugh, nice catch, thanks.

I fixed this by logging via `print_msg` (which is covered by the unit tests). The non-dry-run `--update-pr` is not covered by the tests currently (maybe we should look into mocking the actual push rather than only checking the dry-run...)

easybuild/tools/github.py
```diff
@@ -777,11 +777,9 @@ def _easyconfigs_pr_common(paths, ecs, start_branch=None, pr_branch=None, target
 
     push_branch_msg = "pushing branch '%s' to remote '%s' (%s)" % (pr_branch, remote_name, github_url)
     if dry_run:
-        push_branch_msg += ' [DRY RUN]'
-    print_msg(push_branch_msg)
-
-    if not dry_run:
-        _log.debug(push_log_msg)
+        print_msg(push_branch_msg + '[DRY RUN]', log=_log)
```
wpoely86: space between `push_branch_msg` and `[DRY RUN]` ?

PR #2421, opened 2018-03-02
fix order in result of gen_list_easyblocks and gen_easyblocks_overview_rst

boegel:


easybuild/tools/docs.py
```diff
@@ -490,7 +490,7 @@ def add_class(classes, cls):
         children = cls.__subclasses__()
         classes.update({cls.__name__: {
             'module': cls.__module__,
-            'children': [x.__name__ for x in children]
+            'children': sorted([x.__name__ for x in children], key=lambda x: x.lower())
```
wpoely86: it's minor but to avoid confusion I would not use `x` twice here? Make it `lambda y: y.lower()` or so?

Issue #2407, opened 2018-02-16
Add hooks executed to the output


damianam:
Wouldn't it make sense to add something like `== running whatever_hook` to the output, so at least it's clear that a hook is being executed?

fgeorgatos:
+1
-- 
echo "sysadmin know better bash than english"|sed s/min/mins/ \
  | sed 's/better bash/bash better/' # signal detected in a CERN forum

boegel:
It's already mentioned in the log, but I guess it doesn't hurt to have it more prominent...

Easy to change in `run_hook` (https://github.com/easybuilders/easybuild-framework/blob/master/easybuild/tools/hooks.py#L160), just add a call to `print_msg` next to the `log.info` call...

boegel:
fixed in #2427

PR #2427, opened 2018-03-06
Print which hook is being executed in the command line output

damianam:
See https://github.com/easybuilders/easybuild-framework/issues/2407


boegel:
@damianam fixed the broken test and made the output a bit prettier in https://github.com/damianam/easybuild-framework/pull/12

boegel:
Thanks @damianam!

PR #2432, opened 2018-03-08
determine whether included easyblocks are generic or not based on class they define

boegel:
fix or #1964

cc @damianam, @pescobar

### DIFF
--- a/easybuild/tools/include.py
+++ b/easybuild/tools/include.py
@@ -145,7 +145,7 @@ def verify_imports(pymods, pypkg, from_path):
 
 def is_software_specific_easyblock(module):
     """Determine whether Python module at specified location is a software-specific easyblock."""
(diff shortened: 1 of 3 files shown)


easybuild/tools/include.py
```diff
@@ -142,6 +143,11 @@ def verify_imports(pymods, pypkg, from_path):
         _log.debug("Import of %s from %s verified", pymod_spec, from_path)
 
 
+def is_software_specific_easyblock(module):
+    """Determine whether Python module at specified location is a software-specific easyblock."""
+    return bool(re.search('^class EB_.*\(.*\):$', read_file(module), re.M))
```
wpoely86: keep accidental trailing whitespace in mind?

boegel: Makes sense, done

PR #2434, opened 2018-03-10
avoid that '--inject-checksums' adds lines longer than 120 characters

smoors:
if the line is too long, put the comment in a separate line before the checksum.


boegel:
@smoors We should do the same for checksums for extensions, and also enhance the existing test for `--inject-checksums` to cover the new code. I looked into this, see https://github.com/smoors/easybuild-framework/pull/1, please review/merge to update this PR.

easybuild/framework/easyblock.py
```diff
@@ -3045,7 +3053,15 @@ def inject_checksums(ecs, checksum_type):
...
-                                exts_list_lines.append("%s'%s',  # %s" % (INDENT_4SPACES * 3, checksum, fn))
+                                line_indent = INDENT_4SPACES * 3
+                                ext_checksum_line = "%s'%s',  # %s" % (line_indent, checksum, fn)
+                                if len(ext_checksum_line) > MAX_LINE_LENGTH:
+                                    exts_list_lines.extend([
+                                        "%s# %s" % (line_indent, fn),
+                                        "%s'%s'," % (line_indent, checksum),
+                                    ])
+                                else:
+                                    exts_list_lines.append(ext_checksum_line)
```
boegel: @smoors This diff makes it really stand out that this is a whole bunch of duplicate code, which we should avoid...

We should look into adding an (inner?) function that takes care of this once, for a specified indent level, so we can use it in both situations...

Let me know if you're up for looking into that or not.

test/framework/easyblock.py
```diff
@@ -847,27 +847,28 @@ def test_fetch_patches(self):
...
         eb.fetch_patches()
         self.assertEqual(len(eb.patches), 2)
-        self.assertEqual(eb.patches[0]['name'], 'toy-0.0_typo.patch')
+        self.assertEqual(eb.patches[0]['name'], 'toy-0.0_fix-silly-typo-in-printf-statement.patch')
         self.assertFalse('level' in eb.patches[0])
 
         # reset
         eb.patches = []
 
+        toy_patch = 'toy-0.0_fix-silly-typo-in-printf-statement.patch'
```
boegel: This one should be moved up I guess so `toy_patch` can also be used above in this test, I overlooked that.

easybuild/framework/easyblock.py
smoors: ok, I will look into it.

boegel:
Thanks a lot for looking into this @smoors!

PR #2437, opened 2018-03-13
add Accept header when downloading file

oliviermattelaer:
when trying to download file with urllib2 adding to the header the field Accept since the absence of such field can lead to some security module to raise a 403 response:
https://serverfault.com/questions/377656/why-does-mod-security-require-an-accept-http-header-field

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -459,7 +459,7 @@ def download_file(filename, url, path, forced=False):
     attempt_cnt = 0
 
     # use custom HTTP header
-    url_req = urllib2.Request(url, headers={'User-Agent': 'EasyBuild',  "Accept" : "text/html"})
+    url_req = urllib2.Request(url, headers={'User-Agent': 'EasyBuild',  "Accept" : "*/*"})
 
     while not downloaded and attempt_cnt < max_attempts:
         try:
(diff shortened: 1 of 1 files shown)


easybuild/tools/filetools.py
```diff
@@ -459,7 +459,7 @@ def download_file(filename, url, path, forced=False):
     attempt_cnt = 0
 
     # use custom HTTP header
-    url_req = urllib2.Request(url, headers={'User-Agent': 'EasyBuild'})
+    url_req = urllib2.Request(url, headers={'User-Agent': 'EasyBuild',  "Accept" : "text/html"})
```
boegel: @oliviermattelaer What does the `text/html` part mean here exactly?

boegel: I think we should use `*/*` rather than `text/html`, to accept all types... Cfr. https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

Please make that change, then it should be OK to merge this.

oliviermattelaer: Sure. Done.

Thanks,

Olivier

boegel:
@oliviermattelaer For future pull requests, please target the `develop` branch (I've changed this myself for this PR).

Going in, thanks for your contribution!

PR #2438, opened 2018-03-14
loosen commit message requirements w.r.t. to patches as long as all easyconfigs are new

migueldiascosta:
It seems to me that as long as all easyconfigs are new, the commit message could still be automatically generated

If a new patch being added applies to an existing easyconfig, that easyconfig would need to be modified so a meaningful commit message is still needed, this change only affects the cases where all easyconfig files are new

do close if I'm missing something

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -689,9 +689,9 @@ def _easyconfigs_pr_common(paths, ecs, start_branch=None, pr_branch=None, start_
     if commit_msg:
         cnt = len(file_info['paths_in_repo'])
         _log.debug("Using specified commit message for all %d new/modified easyconfigs at once: %s", cnt, commit_msg)
-    elif all(file_info['new']) and not paths['patch_files'] and not paths['files_to_delete']:
+    elif all(file_info['new']) and not paths['files_to_delete']:
         # automagically derive meaningful commit message if all easyconfig files are new
-        commit_msg = "adding easyconfigs: %s" % ', '.join(os.path.basename(p) for p in file_info['paths_in_repo'])
+        commit_msg = "adding easyconfigs: %s" % ', '.join(os.path.basename(p) for p in file_info['paths_in_repo']+paths['patch_files'])
     else:
         raise EasyBuildError("A meaningful commit message must be specified via --pr-commit-msg when "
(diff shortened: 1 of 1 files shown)


boegel:
@migueldiascosta This makes sense to me, I've found the current behaviour rather annoying and unnecessary myself... :)

We need to be a bit careful when changing `tools/github.py`, since most of the relevant tests for that module are effectively being skipped for PRs (because the GitHub token is not available to prevent that  is being leaked).

I'll run the `github` subsuite myself with this change, and maybe try to enhance the tests to check that this behaviour is retained.

boegel:
No problems with tests, so this is good to go imho. Thanks @migueldiascosta!

boegel:
Ugh, seems like I wasn't careful enough, I didn't check the GitHub-related tests in `test/options.py`...

@migueldiascosta Can you look into fixing this broken test?

```
FAIL: Test use of --new-pr (dry run only).
  File "/tmp/353536935/lib/python2.6/site-packages/easybuild_framework-3.6.0.dev0-py2.6.egg/test/framework/options.py", line 2508, in test_new_update_pr
    self.assertErrorRegex(EasyBuildError, error_msg, self.eb_main, args, raise_error=True, testing=False)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/vsc_install-0.10.32-py2.6.egg/vsc/install/testing.py", line 122, in assertErrorRegex
    self.assertTrue(False, "Expected errors with %s(%s) call should occur" % (call.__name__, str_args))
AssertionError: Expected errors with eb_main(['--new-pr', '--github-user=easybuild_test', '/tmp/eb-OQQvWe/eb-qGd7yf/eb-QxtCZq/toy.eb', '-D', '--disable-cleanup-tmpdir', '--git-working-dirs-path=/tmp/eb-OQQvWe/eb-qGd7yf/eb-QxtCZq/eb-k2lenT/eb-uNqV__/git-working-dir8H5nMe', '/tmp/353536935/lib/python2.6/site-packages/easybuild_framework-3.6.0.dev0-py2.6.egg/test/framework/sandbox/sources/toy/toy-0.0_fix-silly-typo-in-printf-statement.patch'], raise_error=True, testing=False) call should occur
```

boegel:
broken test fixed in #2443

PR #2442, opened 2018-03-15
add force_in_dry_run=True to copy_file in copy_patch_files

### LINKED ISSUE #2334: --new/update-pr fails in extended dry run mode when adding a patch file
When running `--update-pr -x` and adding a new patch file, it fails saying it cannot find the patch file, which should have been copied to the temp dir. But the actual `--update-pr` works, so `-x` must be disabling the copy?

Additionally, doing the same to submit changes to a patch, dry run mode doesn't show the changes to the patch file, although the actual `--update-pr` does submit them.

migueldiascosta:
fixes #2334


boegel:
@migueldiascosta We should look into covering this bug (fix) in the tests so it's not re-introduced?

PR #2443, opened 2018-03-15
 test for custom commit message when deleting a file instead of when adding a patch

migueldiascosta:
fixes broken test introduced by #2438


boegel:
fix confirmed, see https://travis-ci.org/easybuilders/easybuild-framework/builds/353717946

PR #2446, opened 2018-03-16
Replace raw strings with bytes literal as iterator sentinels in checksum calculation

vanzod:
When porting the same construct in Python 3, the use of `r''` as iterator sentinel result in an infinite loop.
This can be solved by replacing it with a bytes literal.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -154,7 +154,7 @@ class ZlibChecksum(object):
     """
     def __init__(self, algorithm):
         self.algorithm = algorithm
-        self.checksum = algorithm(r'')  # use the same starting point as the module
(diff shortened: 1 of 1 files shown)


boegel:
tests fail, but for a reason unrelated to this PR, follow up in #2447 ...

boegel:
Based on https://docs.python.org/2/whatsnew/2.6.html#pep-3112-byte-literals and this:

```
$ python
Python 2.7.5 (default, Aug  4 2017, 00:39:18)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux2
>>> r'' == b''
True
```

+ the passing tests, this is good to go. One tiny step towards supporting Python 3... Thanks @vanzod!

PR #2448, opened 2018-03-16
(just another test for houndci)

boegel:


boegel:
oops, wrong target branch...

PR #2452, opened 2018-03-19
flesh out common code blocks in test/framework/options.py

migueldiascosta:


test/framework/options.py
```diff
@@ -2433,6 +2433,27 @@ def test_fixed_installdir_naming_scheme(self):
         app.gen_installdir()
         self.assertTrue(app.installdir.endswith('software/toy/0.0'))
 
+    def _assert_regexs(self, regexs, txt, assert_true=True):
+        for regex in regexs:
```
boegel: @migueldiascosta Please include a docstring when adding methods

test/framework/options.py
```diff
@@ -2433,6 +2433,27 @@ def test_fixed_installdir_naming_scheme(self):
...
+            regex = re.compile(regex, re.M)
+            if assert_true:
+                self.assertTrue(regex.search(txt), "Pattern '%s' found in: %s" % (regex.pattern, txt))
+            else:
+                self.assertFalse(regex.search(txt), "Pattern '%s' NOT found in: %s" % (regex.pattern, txt))
+
+    def _run_mock_eb(self, args, stdout=True, stderr=False, do_build=False, raise_error=False, verbose=False, testing=False, strip=False):
```
boegel: @migueldiascosta Please include a docstring when adding methods

boegel: Also, why not always mock both `stdout` and `stderr`?

I would remove the `stdout` & `stderr` named arguments in this method, and always mock & return both?

test/framework/options.py
```diff
@@ -3570,13 +3479,7 @@ def test_force_download(self):
...
         ]
-        self.mock_stdout(True)
-        self.mock_stderr(True)
-        self.eb_main(args, raise_error=True, verbose=True, do_build=True)
-        stdout = self.get_stdout().strip()
-        stderr = self.get_stderr().strip()
-        self.mock_stdout(False)
-        self.mock_stderr(False)
+        stdout, stderr = self._run_mock_eb(args, do_build=True, raise_error=True, verbose=True, strip=True)
         self.assertEqual(stdout, '')
```
migueldiascosta: @boegel this is failing, but I don't understand, why should `stdout` be empty?

It doesn't fail without the new helper method, so I must be missing something...

boegel:
@migueldiascosta `testing` is `True` by default in `eb_main`, but in your utility function `_run_mock_eb` it's set to `False` by defaut.

With `testing=True`, you don't get the "step" `==` lines printed to `stdout`.

So, you can fix this by using `testing=True` in your call to `_run_mcok_` in `test_force_download`, I think.

migueldiascosta:
@boegel yes, I had noticed that just now, sorry about that :)

migueldiascosta:
inverting the default behaviour was not a brilliant idea...

boegel:
@migueldiascosta I agree, it's probably better to use `testing=False` as default in `_run_mock_eb` just like in `eb_main` to avoid confusion in the future as well?

migueldiascosta:
@boegel you mean `testing=True` by default, right? (it's already confusing :) ) - will do

boegel:
@migueldiascosta Yes, I just confirmed my own statement. I indeed mean `testing=False` in `_run_mock_eb`.

No `True`, I mean `testing=True`! ;-)

boegel:
Thanks for the effort @migueldiascosta!

Issue #2456, opened 2018-03-27
Add a "fetch only" option


MisterFruits:
There is currently a way to easily get tarballs for an easyconfig (and dependencies). This is very convenient when the WWW is not accessible from where installation takes place.

It can be done this way:

```` shell
eb --stop fetch --ignore-osdeps --robot <easyconfig.eb>
````

Nevertheless it still requires a module tool on the machine. While it could be easy to install, it also could no be wanted.

I which I could work on this issue if implementation is not too tricky. I had some experimentation and there is an easy way to do that: create a  modulestool that mocks a modules tools (eg the one in the regression test echoing instead of passing modules command).

I ended up with the following command downloading all the tarball I needed without have any modulestool installed:

``` shell
eb --stop fetch --ignore-osdeps --robot --modules-tool=MockModulesTool --module-syntax=Tcl <easyconfig.eb>
````

There is in my opinion 2 big drawbacks to this implementation:
 - when used with something else than ``--stop fetch``, ``--modules-tool=MockModulesTool`` will make the build fail in a non obvious/explicit way
 - people knowing easybuild would eventually figure out all this option while a beginner will have to ask. Also the ``--module-syntax=Tcl`` has no sense.

What do you think ?

Maybe creating a ``--fetch`` option implying ``--stop fetch --ignore-osdeps --modules-tool=MockModulesTool --module-syntax=Tcl`` and making ``MockModulesTool`` not eventually 'callable' would be the solution ?

MisterFruits:
I started implementing this feature. Few interrogation:
 - would ``--fetch`` belongs to "Basic Options" ?
 - where should I document the feature in the documentation ? (beside the option page)

Flamefire:
Closed by #2457

BTW: Can this be used to fetch ALL given sources no matter if their dependencies have been installed already?

ocaisa:
@Flamefire Yes, if you couple it with `--robot`.

PR #2457, opened 2018-03-27
add support for "eb --fetch" to only download sources (even without having a modules tool installed)

MisterFruits:
This pull request addresses #2456
It is working at home. Please review/help me with the following points
 - [ ] would --fetch belongs to "Basic Options" ?
 - [ ] where should I document the feature in the documentation ? (beside the option page)
 - [ ] where are options tested ? should I care about testing anything else ?

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -62,7 +62,7 @@ from easybuild.tools.filetools import adjust_permissions, cleanup, write_file
 from easybuild.tools.github import check_github, find_easybuild_easyconfig, install_github_token
 from easybuild.tools.github import new_pr, merge_pr, update_pr
 from easybuild.tools.hooks import START, END, load_hooks, run_hook
-from easybuild.tools.modules import modules_tool, MockModulesTool
+from easybuild.tools.modules import modules_tool, NoModulesTool
 from easybuild.tools.options import parse_external_modules_metadata, process_software_build_specs, use_color
 from easybuild.tools.robot import check_conflicts, det_robot_path, dry_run, resolve_dependencies, search_easyconfigs
 from easybuild.tools.package.utilities import check_pkg_support
@@ -256,8 +256,8 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     config.init_build_options(build_options=build_options, cmdline_options=options)
 
     if options.fetch:
-        _log.debug("Fetching sources only, implies stop=fetch, moduletool=MockModulesTool, ignore-osdep.")
-        modtool = MockModulesTool(testing=testing)
+        _log.debug("Fetching sources only, implies stop=fetch, moduletool=NoModulesTool, ignore-osdep.")
+        modtool = NoModulesTool(testing=testing)
 
     if modtool is None:
         modtool = modules_tool(testing=testing)
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -62,7 +62,7 @@ from easybuild.tools.filetools import adjust_permissions, cleanup, write_file
 from easybuild.tools.github import check_github, find_easybuild_easyconfig, install_github_token
 from easybuild.tools.github import new_pr, merge_pr, update_pr
 from easybuild.tools.hooks import START, END, load_hooks, run_hook
-from easybuild.tools.modules import modules_tool, NoModulesTool
+from easybuild.tools.modules import modules_tool
 from easybuild.tools.options import parse_external_modules_metadata, process_software_build_specs, use_color
 from easybuild.tools.robot import check_conflicts, det_robot_path, dry_run, resolve_dependencies, search_easyconfigs
 from easybuild.tools.package.utilities import check_pkg_support
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -255,10 +255,6 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     config.init(options, config_options_dict)
     config.init_build_options(build_options=build_options, cmdline_options=options)
 
-    if options.fetch:
-        _log.debug("Fetching sources only, implies stop=fetch, moduletool=NoModulesTool, ignore-osdep.")
-        modtool = NoModulesTool(testing=testing)
-
     if modtool is None:
         modtool = modules_tool(testing=testing)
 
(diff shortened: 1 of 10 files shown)


MisterFruits:
It seems that I added some erronous code in the modulestool options behaviour. Let me check that before reviewing.

MF

boegel:
@MisterFruits Please sync your branch with current `develop` to pick up the fix for the Travis configuration from #2462

MisterFruits:
Hi !!
I think this PR can be reviewed now. It took me some tries with tests because I can not launch them on my local machine so I have to rely on travis.

What's missing:

- [ ] Re-trigger tests that have been failed to to autopep8 update (see #2462)
- [ ] confirm that ``--fetch`` option is in the good option group
- [ ] indicate the place where I should document the option (if needed)

easybuild/main.py
```diff
@@ -255,6 +255,10 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     config.init(options, config_options_dict)
     config.init_build_options(build_options=build_options, cmdline_options=options)
 
+    if options.fetch:
+        _log.debug("Fetching sources only, implies stop=fetch, moduletool=NoModulesTool, ignore-osdep.")
+        modtool = NoModulesTool(testing=testing)
```
boegel: @MisterFruits Why not enhance the `modules_tool` function to handle this, that way you don't need to touch `main.py` at all?

Seems like you just need to return a `NoModulesTool` instance when `get_modules_tool` returns `None` in `modules_tool`?

That's actually better than what it does now, since now it will return `None` when `get_modules_tool` returns `None` somehow, which would make `main.py` go down hard as soon as `modtool` is used for anything...

easybuild/tools/modules.py
```diff
@@ -1340,6 +1340,8 @@ def avail_modules_tools():
     # filter out legacy Modules class
     if 'Modules' in class_dict:
         del class_dict['Modules']
+    if 'NoModulesTool' in class_dict:
```
boegel: maybe add a comment above this line to explain why: `# NoModulesTool should never be used deliberately, so remove it from the list of available module tools`

boegel:
@MisterFruits To answer your questions:

* maybe putting the `--fetch` option in the `override` section makes more sense
* I'd say http://easybuild.readthedocs.io/en/latest/Partial_installations.html is a good place to document `--fetch`, since that's also where `--stop` is documented?
* it seems like you've already figured out where to test the `--fetch` option?

test/framework/options.py
```diff
@@ -2970,6 +2970,28 @@ def test_stop(self):
...
+    def test_fetch(self):
+        options = EasyBuildOptions(go_args=['--fetch'])
+
+        self.assertTrue(options.options.fetch)
+        self.assertEqual(options.options.stop, 'fetch')
+        self.assertEqual(options.options.modules_tool, None)
+        self.assertTrue(options.options.ignore_osdeps)
+
+        args = ['toy-0.0.eb', '--fetch']
+        stdout, stderr = self._run_mock_eb(args, raise_error=True, strip=True, testing=False)
```
boegel: @MisterFruits You should add `self.modtool = None` above this line to avoid that the prepared modules tool is passed down (that's also why the test fails)

easybuild/tools/modules.py
```diff
@@ -1353,7 +1356,7 @@ def modules_tool(mod_paths=None, testing=False):
         modules_tool_class = avail_modules_tools().get(modules_tool)
         return modules_tool_class(mod_paths=mod_paths, testing=testing)
     else:
-        return None
+        return NoModulesTool(mod_paths=mod_paths, testing=testing)
```
boegel: @MisterFruits Please change this to use positive logic, and avoid the duplicate stuff:

```python
if modules_tool is None:
    modules_tool_class = NoModulesTool
else:
    modules_tool_class = avail_modules_tools().get(modules_tool)
return modules_tool_class(mod_paths=mod_paths, testing=testing)
```

Another option is to simplify this a bit and use `NoModulesTool` as a fallback:

```python
modules_tool = get_modules_tool()
modules_tool_class = avail_modules_tools().get(modules_tool, NoModulesTool)
return modules_tool_class(mod_paths=mod_paths, testing=testing)

File: test/framework/options.py (hunk 3 above)
Comment (MisterFruits) [reply test/framework/options.py]: I didn't notice your comment, I took me some time to understand the trick here x)
In 12498c12b509fb645f33b7e7e10fd22f80bb111f I restored back the original ``self.modtool`` after ``self._run_mock_eb(...)``. Not sure it is required, I was afraid I could brake other tests.

File: easybuild/tools/modules.py (hunk 4 above)
Comment (MisterFruits) [reply easybuild/tools/modules.py]: I like the second solution =)
I named the argument in 690523d so it's more explicit.

Comment (MisterFruits) [reply easybuild/tools/modules.py]: Okay, there is no named argument here... rolling back

**Comment by MisterFruits** [comment]
@boegel, looks good to review again.
Documentation eventualy goes with easybuilders/easybuild#420

File: test/framework/options.py [hunk 5]
```diff
@@ -2970,6 +2970,33 @@ def test_stop(self):
...
+        self.assertEqual(options.options.modules_tool, None)
+        self.assertTrue(options.options.ignore_osdeps)
+
+        # in this test we want to fake the case were no modules tool are in the system so teak it
+        original_modtool = self.modtool
+        self.modtool = None
+        args = ['toy-0.0.eb', '--fetch']
+        stdout, stderr = self._run_mock_eb(args, raise_error=True, strip=True, testing=False)
+        # back to orinal modules tool so further tests can continue
+        self.modtool = original_modtool
```
boegel: @MisterFruits There's no need to restore `self.modtool`, it gets reset for every test automatically.

The tests are pretty well isolated from each other.

easybuild/main.py
```diff
@@ -62,7 +62,7 @@
 from easybuild.tools.github import check_github, find_easybuild_easyconfig, install_github_token
 from easybuild.tools.github import new_pr, merge_pr, update_pr
 from easybuild.tools.hooks import START, END, load_hooks, run_hook
-from easybuild.tools.modules import modules_tool
+from easybuild.tools.modules import modules_tool, NoModulesTool
```
boegel: No longer needed to import `NoModulesTool` here

MisterFruits:
@boegel review addressed ! ;)

PR #2458, opened 2018-03-27
Introduce golf: compiler toolchain with GCC, OpenBLAS, and FFTW.

bartoldeman:
Since BLACS_MODULE_NAME and SCALAPACK_MODULE_NAME are set to None in
this case we need to filter out None for the toolchain definition.


bartoldeman:
And so we are easily building a golf course...

I tested this locally with numpy and golf easyconfigs. Will submit PRs for those too.

PR #2464, opened 2018-04-04
Add goolfc-2.6.10 toolchain for test_get_toolchain_hierarchy robot test.

bartoldeman:
goolfc has a more complex subtoolchain hierarchy than goolf/foss so it's useful to test.

I added 2.6.10 because of the existing CUDA test easyconfig, it's all old software but that does not matter really.

### DIFF
--- a/test/framework/easyconfigs/test_ecs/o/OpenMPI/OpenMPI-1.7.3-gcccuda-2.6.10.eb
+++ b/test/framework/easyconfigs/test_ecs/o/OpenMPI/OpenMPI-1.7.3-gcccuda-2.6.10.eb
@@ -1,0 +1,42 @@
+easyblock = 'ConfigureMake'
+
+name = 'OpenMPI'
+version = "1.7.3"
+
+homepage = 'http://www.open-mpi.org/'
+description = """The Open MPI Project is an open source MPI-2 implementation."""
+
+toolchain = {'name': 'gcccuda', 'version': '2.6.10'}
+
+sources = [SOURCELOWER_TAR_GZ]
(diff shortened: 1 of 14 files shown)


test/framework/easyconfigs/test_ecs/o/OpenMPI/OpenMPI-1.7.3-gcccuda-2.6.10.eb
```diff
@@ -0,0 +1,37 @@
...
+
+dependencies = [('hwloc', '1.8')]
+
+configopts = '--with-threads=posix --enable-shared --enable-mpi-thread-multiple --with-verbs '
+configopts += '--enable-mpirun-prefix-by-default '  # suppress failure modes in relation to mpirun path
+configopts += '--with-hwloc=$EBROOTHWLOC '  # hwloc support
+configopts += '--with-cuda=$CUDA_HOME '             # CUDA-aware build; N.B. --disable-dlopen is incompatible
+
+# needed for --with-verbs
+osdependencies = [('libibverbs-dev', 'libibverbs-devel')]
```
akesandgren: Should we add rdma-core-devel to be consistent with how OpenMPI easyconfigs look now adays?

boegel: @akesandgren We could, but it doesn't really matter for these tests, there's little point in keeping the test easyconfigs in sync with the central easyconfigs repo, unless there's a reason for it (e.g. to reproduce a bug or something).

I don't see how having two or three options here matters in any way.

PR #2465, opened 2018-04-04
Use GCC instead of gompi for OpenBLAS in goolf-1.4.10 test toolchain.

bartoldeman:
This allows to use the golf toolchain within goolf-1.4.10, also added.

Issue #2204, opened 2017-05-04
Crashes while using --job (GC3Pie/2.4.2 + Slurm)


nortex:
Hi all,

Everytime i try to install software with the --job option i get this message after few minutes:

`== temporary log file in case of crash /tmp/eb-DqqWHV/easybuild-uUo01C.log
== resolving dependencies ...
== GC3Pie job overview: 3 submitted (total: 7)
Traceback (most recent call last):
  File "/usr/lib64/python2.6/runpy.py", line 122, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.6/runpy.py", line 34, in _run_code
    exec code in run_globals
  File "/Easybuild/software/EasyBuild/3.1.2/lib/python2.6/site-packages/easybuild_framework-3.1.2-py2.6.egg/easybuild/main.py", line 453, in <module>
    main()
  File "/Easybuild/software/EasyBuild/3.1.2/lib/python2.6/site-packages/easybuild_framework-3.1.2-py2.6.egg/easybuild/main.py", line 413, in main
    submit_jobs(ordered_ecs, eb_go.generate_cmd_line(), testing=testing)
  File "/Easybuild/software/EasyBuild/3.1.2/lib/python2.6/site-packages/easybuild_framework-3.1.2-py2.6.egg/easybuild/tools/parallelbuild.py", line 144, in submit_jobs
    return build_easyconfigs_in_parallel(command, ordered_ecs, prepare_first=prepare_first)
  File "/Easybuild/software/EasyBuild/3.1.2/lib/python2.6/site-packages/easybuild_framework-3.1.2-py2.6.egg/easybuild/tools/parallelbuild.py", line 115, in build_easyconfigs_in_parallel
    active_job_backend.complete()
  File "/Easybuild/software/EasyBuild/3.1.2/lib/python2.6/site-packages/easybuild_framework-3.1.2-py2.6.egg/easybuild/tools/job/gc3pie.py", line 239, in complete
    self._engine.progress()
  File "build/bdist.linux-x86_64/egg/gc3libs/core.py", line 1305, in progress
  File "build/bdist.linux-x86_64/egg/gc3libs/core.py", line 416, in update_job_state
  File "build/bdist.linux-x86_64/egg/gc3libs/core.py", line 452, in __update_application
IndexError: string index out of range
`

At log file i can see the same error as above, and i do not know if some software was installed successfully or not.

[...]

PR #2466, opened 2018-04-04
consider potential of multiple subtoolchains in get_toolchain_hierarchy 

bartoldeman:
This PR replaces #2234, and depends on ~~#2464~~ and ~~#2465~~.
subtoolchains are now searched using a breadth-first-search, which
allows multiple subtoolchains per toolchain.

Some subtoolchains such as golf and golfc are now marked OPTIONAL
since they do not have to exist as modules, and can be skipped.

Since those optional subtoolchains cannot usually be found via a
dependency search (as toolchain easyconfigs do not typically have
subtoolchains as dependencies)
try to search for subtoolchain/version with the same version as
the parent. E.g. goolfc/2.6.10 searches for golfc/2.6.10, and
golfc/2.6.10 searches for golf/2.6.10 and gcccuda/2.6.10.

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -149,14 +149,22 @@ def get_toolchain_hierarchy(parent_toolchain):
     # obtain list of all possible subtoolchains
     _, all_tc_classes = search_toolchain('')
     subtoolchains = dict((tc_class.NAME, getattr(tc_class, 'SUBTOOLCHAIN', None)) for tc_class in all_tc_classes)
-
-    current_tc_name, current_tc_version = parent_toolchain['name'], parent_toolchain['version']
-    subtoolchain_name, subtoolchain_version = subtoolchains[current_tc_name], None
+    optional_toolchains = set(tc_class.NAME for tc_class in all_tc_classes if getattr(tc_class, 'OPTIONAL', False))
 
     # the parent toolchain is at the top of the hierarchy
     toolchain_hierarchy = [parent_toolchain]
+    # use a queue to handle a breadth-first-search of the hierarchy
+    bfs_queue = [parent_toolchain]
+    visited = set()
 
-    while subtoolchain_name:
+    while bfs_queue:
+        current_tc = bfs_queue.pop()
+        current_tc_name, current_tc_version = current_tc['name'], current_tc['version']
+        subtoolchain_names = subtoolchains[current_tc_name]
+        if subtoolchain_names is None:
+            continue
+        if not isinstance(subtoolchain_names, list):
+            subtoolchain_names = [subtoolchain_names]
         # grab the easyconfig of the current toolchain and search the dependencies for a version of the subtoolchain
         path = robot_find_easyconfig(current_tc_name, current_tc_version)
         if path is None:
@@ -196,44 +204,54 @@ def get_toolchain_hierarchy(parent_toolchain):
                 cands.append({'name': depdep['name'], 'version': depdep['version'] + depdep['versionsuffix']})
                 cands.append(depdep['toolchain'])
 
-        # only retain candidates that match subtoolchain name
-        cands = [c for c in cands if c['name'] == subtoolchain_name]
+        # only retain candidates that match subtoolchain names
+        cands = [c for c in cands if c['name'] in subtoolchain_names]
 
-        uniq_subtc_versions = set([subtc['version'] for subtc in cands])
+        for subtoolchain_name in subtoolchain_names:
+            uniq_subtc_versions = set([subtc['version'] for subtc in cands if subtc['name'] == subtoolchain_name])
 
-        if len(uniq_subtc_versions) == 1:
-            subtoolchain_version = cands[0]['version']
+            if len(uniq_subtc_versions) == 1:
+                subtoolchain_version = list(uniq_subtc_versions)[0]
 
-        elif len(uniq_subtc_versions) == 0:
-            # only retain GCCcore as subtoolchain if version was found
-            if subtoolchain_name == GCCcore.NAME:
-                _log.info("No version found for %s; assuming legacy toolchain and skipping it as subtoolchain.",
-                          subtoolchain_name)
-                subtoolchain_name = GCCcore.SUBTOOLCHAIN
-                subtoolchain_version = ''
-            # dummy toolchain: end of the line
-            elif subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
-                subtoolchain_version = ''
+            elif len(uniq_subtc_versions) == 0:
+                # only retain GCCcore as subtoolchain if version was found
+                if subtoolchain_name == GCCcore.NAME:
+                    _log.info("No version found for %s; assuming legacy toolchain and skipping it as subtoolchain.",
+                              subtoolchain_name)
+                    subtoolchain_name = GCCcore.SUBTOOLCHAIN
+                    subtoolchain_version = ''
+                # dummy toolchain: end of the line
+                elif subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
+                    subtoolchain_version = ''
+                elif ((subtoolchain_name in optional_toolchains or current_tc_name in optional_toolchains) and
+                      robot_find_easyconfig(subtoolchain_name, current_tc_version)):
+                    # special case: optionally find e.g. golf/1.4.10 for goolf/1.4.10 even if it is not in
+                    # the module dependencies. This is only allowed for and inside optional subtoolchains.
+                    subtoolchain_version = current_tc_version
+                elif subtoolchain_name in optional_toolchains:
+                    continue
+                else:
+                    raise EasyBuildError("No version found for subtoolchain %s in dependencies of %s",
+                                         subtoolchain_name, current_tc_name)
             else:
-                raise EasyBuildError("No version found for subtoolchain %s in dependencies of %s",
-                                     subtoolchain_name, current_tc_name)
-        else:
-            if subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
-                # Don't care about multiple versions of dummy
-                _log.info("Ignoring multiple versions of %s in toolchain hierarchy", DUMMY_TOOLCHAIN_NAME)
-                subtoolchain_version = ''
-            else:
-                raise EasyBuildError("Multiple versions of %s found in dependencies of toolchain %s: %s",
-                                     subtoolchain_name, current_tc_name, ', '.join(sorted(uniq_subtc_versions)))
+                if subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
+                    # Don't care about multiple versions of dummy
+                    _log.info("Ignoring multiple versions of %s in toolchain hierarchy", DUMMY_TOOLCHAIN_NAME)
+                    subtoolchain_version = ''
+                else:
+                    raise EasyBuildError("Multiple versions of %s found in dependencies of toolchain %s: %s",
+                                         subtoolchain_name, current_tc_name, ', '.join(sorted(uniq_subtc_versions)))
 
-        if subtoolchain_name == DUMMY_TOOLCHAIN_NAME and not build_option('add_dummy_to_minimal_toolchains'):
-            # we're done
-            break
+            if subtoolchain_name == DUMMY_TOOLCHAIN_NAME and not build_option('add_dummy_to_minimal_toolchains'):
+                # we're done
+                continue
 
-        # add to hierarchy and move to next
-        current_tc_name, current_tc_version = subtoolchain_name, subtoolchain_version
-        subtoolchain_name, subtoolchain_version = subtoolchains[current_tc_name], None
-        toolchain_hierarchy.insert(0, {'name': current_tc_name, 'version': current_tc_version})
+            # add to hierarchy and move to next
+            if subtoolchain_name not in visited:
+                tc = {'name': subtoolchain_name, 'version': subtoolchain_version}
+                toolchain_hierarchy.insert(0, tc)
+                bfs_queue.insert(0, tc)
+                visited.add(tc['name'])
 
     _log.info("Found toolchain hierarchy for toolchain %s: %s", parent_toolchain, toolchain_hierarchy)
     return toolchain_hierarchy
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -134,6 +134,60 @@ def toolchain_hierarchy_cache(func):
     return cache_aware_func
 
 
+def det_subtoolchain_version(current_tc, subtoolchain_name, optional_toolchains, cands):
+    """
+    Returns unique version for subtoolchain, in tc dict.
+    If there is no unique version:
+    * use '' for dummy, if dummy is not skipped.
+    * GCCcore can be downgraded to dummy.
+    * return None for skipped subtoolchains, that is,
+      optional toolchains or dummy without add_dummy_to_minimal_toolchains.
+    * in all other cases, raises an exception.
+    """
+    current_tc_name, current_tc_version = current_tc['name'], current_tc['version']
+
+    uniq_subtc_versions = set([subtc['version'] for subtc in cands if subtc['name'] == subtoolchain_name])
+
+    if len(uniq_subtc_versions) == 1:
+        subtoolchain_version = list(uniq_subtc_versions)[0]
+    elif len(uniq_subtc_versions) == 0:
+        # only retain GCCcore as subtoolchain if version was found
+        if subtoolchain_name == GCCcore.NAME:
+            _log.info("No version found for %s; assuming legacy toolchain and skipping it as subtoolchain.",
+                      subtoolchain_name)
+            # downgrade to dummy
+            subtoolchain_name = GCCcore.SUBTOOLCHAIN
+            subtoolchain_version = ''
+        # dummy toolchain: bottom of the hierarchy
+        elif subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
+            subtoolchain_version = ''
+        else:
+            is_optional_tc = subtoolchain_name in optional_toolchains or current_tc_name in optional_toolchains
+            if is_optional_tc and robot_find_easyconfig(subtoolchain_name, current_tc_version):
+                # special case: optionally find e.g. golf/1.4.10 for goolf/1.4.10 even if it is not in
+                # the module dependencies. This is only allowed for and inside optional subtoolchains.
+                subtoolchain_version = current_tc_version
+            elif subtoolchain_name in optional_toolchains:
+                # skip if the subtoolchain considered now is optional
+                return None
+            else:
+                raise EasyBuildError("No version found for subtoolchain %s in dependencies of %s",
+                                     subtoolchain_name, current_tc_name)
+    elif subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
+        # Don't care about multiple versions of dummy
+        _log.info("Ignoring multiple versions of %s in toolchain hierarchy", DUMMY_TOOLCHAIN_NAME)
+        subtoolchain_version = ''
+    else:
+        raise EasyBuildError("Multiple versions of %s found in dependencies of toolchain %s: %s",
+                             subtoolchain_name, current_tc_name, ', '.join(sorted(uniq_subtc_versions)))
+
+    if subtoolchain_name == DUMMY_TOOLCHAIN_NAME and not build_option('add_dummy_to_minimal_toolchains'):
+        # skip dummy if add_dummy_to_minimal_toolchains is not set
+        return None
+
+    return {'name': subtoolchain_name, 'version': subtoolchain_version}
+
+
 @toolchain_hierarchy_cache
 def get_toolchain_hierarchy(parent_toolchain):
     """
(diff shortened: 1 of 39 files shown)


akesandgren:
Now a bit of merging of conflicts is in order here.

bartoldeman:
Just a merge, no conflicts.

bartoldeman:
Reopening since travis still has not reported after 19 hours.

akesandgren:
This looks sane to me. How much testing have you done?

bartoldeman:
I've had a simpler variant (from #2234) of this in Compute Canada for ages (since May 2017), but not this exact code yet. Will do a few tests here.

bartoldeman:
I did a test with this code on CP2K-iomkl now which in our setup depends on libxsmm-iimkl using this subtoolchain mechanism. It works.

I did promise in last week's conf call however to add a comment with the dependency tree for the most complex case (goolfc). So I will do that.

bartoldeman:
One more thing, this approach should be applied to many other toolchains to be complete, at least the following:
foss.py
gmkl.py
gmklc.py
iimkl.py
iimklc.py
intel.py
iomkl.py
pmkl.py

should I submit a seperate PR for this fairly boring but noisy diff?

akesandgren:
I think a separate PR for that would be easier to handle? @boegel ??

bartoldeman:
travis now fails with ```
 File "/tmp/422121558/lib/python2.6/site-packages/easybuild_framework-3.6.3.dev0-py2.6.egg/easybuild/tools/job/gc3pie.py", line 94, in _check_version
    version_str = gc3libs.core.__version__
AttributeError: 'module' object has no attribute '__version__'```
which looks unrelated.

bartoldeman:
Related to Travis issue:
https://github.com/easybuilders/easybuild-framework/issues/2204

boegel:
> travis now fails with ```
 File "/tmp/422121558/lib/python2.6/site-packages/easybuild_framework-3.6.3.dev0-py2.6.egg/easybuild/tools/job/gc3pie.py", line 94, in _check_version
    version_str = gc3libs.core.__version__
AttributeError: 'module' object has no attribute '__version__'```
> which looks unrelated.

@riccardomurri Can this somehow be caused by the GC3Pie release? The latest version of GC3Pie gets installed automagically, it seems like 2.5.0 includes a backwards incompatible change?!

riccardomurri:
The `__version__` fields were removed from the GC3Pie source files, to keep the release version in one place only (`setup.py`). 

The `pkg_resources` API should be used to check the package version -- I'll add to the GC3Pie 2.5.0 compatibility PR.

boegel:
@riccardomurri Well, ok, I can see the motivation, but this is a backward-incompatible change, so anyone who will be trying to use an existing EasyBuild release with GC3Pie 2.5.0 will run into this... :)

Very good reason to get #2474 ready & merged to be included in EasyBuild v3.7.0 ASAP... :)

riccardomurri:
Fix for this "version" issue is in: https://github.com/easybuilders/easybuild-framework/pull/2554

akesandgren:
Close/reopen to restart travis checks

akesandgren:
lgtm,  @damianam any other comments?

boegel:
I'll close/re-open so Travis picks up the changes from the now merged #2554

akesandgren:
actually the most complicated one is likely to be intelcuda due to the split of icc/ifort :-)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -143,19 +143,41 @@ def get_toolchain_hierarchy(parent_toolchain):
...
-
-    current_tc_name, current_tc_version = parent_toolchain['name'], parent_toolchain['version']
-    subtoolchain_name, subtoolchain_version = subtoolchains[current_tc_name], None
+    optional_toolchains = set(tc_class.NAME for tc_class in all_tc_classes if getattr(tc_class, 'OPTIONAL', False))
 
     # the parent toolchain is at the top of the hierarchy
     toolchain_hierarchy = [parent_toolchain]
-
-    while subtoolchain_name:
+    # use a queue to handle a breadth-first-search of the hierarchy
```
boegel: Maybe extend the comment to mention why we need breadth-first search?

```python
# use a queue to handle a breadth-first-search of the hierarchy,
# which is required to take into account the potential for multiple subtoolchains

File: easybuild/framework/easyconfig/easyconfig.py [hunk 2]
```diff
@@ -143,19 +143,41 @@ def get_toolchain_hierarchy(parent_toolchain):
...
+    bfs_queue = [parent_toolchain]
+    visited = set()
+
+    while bfs_queue:
+        current_tc = bfs_queue.pop()
+        current_tc_name, current_tc_version = current_tc['name'], current_tc['version']
+        subtoolchain_names = subtoolchains[current_tc_name]
+        if subtoolchain_names is None:
+            continue
```
boegel: please add a clarifying comment above

```python
# if current toolchain has no subtoolchains, consider next toolchain in queue
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -143,19 +143,41 @@ def get_toolchain_hierarchy(parent_toolchain):
...
+        if not isinstance(subtoolchain_names, list):
+            subtoolchain_names = [subtoolchain_names]
```
boegel: comment:

```python
# make sure we always have a list of subtoolchains, even if there's only one

File: easybuild/framework/easyconfig/easyconfig.py [hunk 4]
```diff
@@ -195,44 +217,54 @@ def get_toolchain_hierarchy(parent_toolchain):
...
+                subtoolchain_version = list(uniq_subtc_versions)[0]
+
+            elif len(uniq_subtc_versions) == 0:
+                # only retain GCCcore as subtoolchain if version was found
+                if subtoolchain_name == GCCcore.NAME:
+                    _log.info("No version found for %s; assuming legacy toolchain and skipping it as subtoolchain.",
+                              subtoolchain_name)
+                    subtoolchain_name = GCCcore.SUBTOOLCHAIN
+                    subtoolchain_version = ''
+                # dummy toolchain: end of the line
```
boegel: comment is no longer valid? clearly not "end of the line", there's 2 more `elif`s below...

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -195,44 +217,54 @@ def get_toolchain_hierarchy(parent_toolchain):
...
+                elif subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
+                    subtoolchain_version = ''
+                elif ((subtoolchain_name in optional_toolchains or current_tc_name in optional_toolchains) and
+                      robot_find_easyconfig(subtoolchain_name, current_tc_version)):
```
```python
is_optional_tc = subtoolchain_name in optional_toolchains or current_tc_name in optional_toolchains
...
elif is_optional_tc and robot_find_easyconfig(subtoolchain_name, current_tc_version):

File: easybuild/framework/easyconfig/easyconfig.py [hunk 6]
```diff
@@ -195,44 +217,54 @@ def get_toolchain_hierarchy(parent_toolchain):
...
+                    # special case: optionally find e.g. golf/1.4.10 for goolf/1.4.10 even if it is not in
+                    # the module dependencies. This is only allowed for and inside optional subtoolchains.
+                    subtoolchain_version = current_tc_version
+                elif subtoolchain_name in optional_toolchains:
+                    continue
```
boegel: @bartoldeman This needs a clarifying comment for me (although I guess the term "optional" is a good hint here...)
Probably something like:

```python
# consider next subtoolchain in case the current want was optional

File: easybuild/framework/easyconfig/easyconfig.py [hunk 7]
```diff
@@ -195,44 +217,54 @@ def get_toolchain_hierarchy(parent_toolchain):
...
+                else:
+                    raise EasyBuildError("No version found for subtoolchain %s in dependencies of %s",
+                                         subtoolchain_name, current_tc_name)
             else:
-                raise EasyBuildError("Multiple versions of %s found in dependencies of toolchain %s: %s",
-                                     subtoolchain_name, current_tc_name, ', '.join(sorted(uniq_subtc_versions)))
+                if subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
```
boegel: this should be an `elif` (one indent level to the left?)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -195,44 +217,54 @@ def get_toolchain_hierarchy(parent_toolchain):
...
+                else:
+                    raise EasyBuildError("Multiple versions of %s found in dependencies of toolchain %s: %s",
+                                         subtoolchain_name, current_tc_name, ', '.join(sorted(uniq_subtc_versions)))
 
-        if subtoolchain_name == DUMMY_TOOLCHAIN_NAME and not build_option('add_dummy_to_minimal_toolchains'):
-            # we're done
-            break
+            if subtoolchain_name == DUMMY_TOOLCHAIN_NAME and not build_option('add_dummy_to_minimal_toolchains'):
+                # we're done
+                continue
```
boegel: the combination of `we're done` and `continue` here is funny, please change the comment to something more sensible :)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -195,44 +217,54 @@ def get_toolchain_hierarchy(parent_toolchain):
...
-                                     subtoolchain_name, current_tc_name)
-        else:
-            if subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
-                # Don't care about multiple versions of dummy
-                _log.info("Ignoring multiple versions of %s in toolchain hierarchy", DUMMY_TOOLCHAIN_NAME)
-                subtoolchain_version = ''
+        # only retain candidates that match subtoolchain names
+        cands = [c for c in cands if c['name'] in subtoolchain_names]
+
+        for subtoolchain_name in subtoolchain_names:
```
boegel: @bartoldeman Shouldn't we make this whole `for` loop a separate function, named something like `det_subtoolchain_version`, which we can test separately?
It could help a lot in making this code easier to understand I think...

Also, shouldn't we `break` out of this `for` loop every time we have found a subtoolchain version?

boegel:
@bartoldeman For the other toolchains, it's just a one-line diff per toolchain to make `SUBTOOLCHAINS` a list, right?

That should be included in this PR imho, to avoid it being pulled out of context...

easybuild/framework/easyconfig/easyconfig.py
bartoldeman: No, don't `break`! It's a loop over the names. Once the name has been found it is added to the toolchain_hierarchy list. The version check just needs to see if there is one version only, and else raise an error. I am moving it to a separate function now and will push that.

boegel: @bartoldeman OK, thanks for clarifying :)

Maybe mention that as a comment above the `for`:

```python
# make sure exactly one version if found for each subtoolchain

File: easybuild/framework/easyconfig/easyconfig.py [hunk 10]
```diff
@@ -134,6 +134,60 @@ def cache_aware_func(toolchain):
...
+        subtoolchain_version = ''
+    else:
+        raise EasyBuildError("Multiple versions of %s found in dependencies of toolchain %s: %s",
+                             subtoolchain_name, current_tc_name, ', '.join(sorted(uniq_subtc_versions)))
+
+    if subtoolchain_name == DUMMY_TOOLCHAIN_NAME and not build_option('add_dummy_to_minimal_toolchains'):
+        # skip dummy if add_dummy_to_minimal_toolchains is not set
+        return None
+
+    return {'name': subtoolchain_name, 'version': subtoolchain_version}
```
boegel: @bartoldeman I would just return `subtoolchain_version` here? I don't really see the point in returning a dict value here...

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -195,44 +274,16 @@ def get_toolchain_hierarchy(parent_toolchain):
...
-            break
+        # only retain candidates that match subtoolchain names
+        cands = [c for c in cands if c['name'] in subtoolchain_names]
 
-        # add to hierarchy and move to next
-        current_tc_name, current_tc_version = subtoolchain_name, subtoolchain_version
-        subtoolchain_name, subtoolchain_version = subtoolchains[current_tc_name], None
-        toolchain_hierarchy.insert(0, {'name': current_tc_name, 'version': current_tc_version})
+        for subtoolchain_name in subtoolchain_names:
+            tc = det_subtoolchain_version(current_tc, subtoolchain_name, optional_toolchains, cands)
```
boegel: If you change the return type of `det_subtoolchain_version`, I'd go with `subtc_ver` (but I guess `subtoolchain_version` is better for consistency with `subtoolchain_name`)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -195,44 +274,16 @@ def get_toolchain_hierarchy(parent_toolchain):
...
+            # add to hierarchy and move to next
+            if tc is not None and tc['name'] not in visited:
```
boegel: It's the `tc['name']` that makes this a bit awkward, just use `subtoolchain_name`?

easybuild/framework/easyconfig/easyconfig.py
bartoldeman: I can't because of the change of name from "GCCcore" to "dummy" for legacy toolchains. I did not want to remove that code.

I suspect the idea is that older toolchains had GCC directly above dummy, but the gcc toolchain has GCCcore as the subtoolchain which does not exist in those situations.

boegel: Ah, OK, I figured there must have been a good reason.

Then: i) mention that in a comment, explain why you need to return the subtoolchain name too, ii) just make this a tuple `(subtoolchain_name, subtoolchain_version)`.

easybuild/framework/easyconfig/easyconfig.py
boegel: With the above in mind:

```python
subtc_name, subtc_ver = det_subtoolchain_version(current_tc, subtoolchain_name, optional_toolchains, cands)

File: easybuild/framework/easyconfig/easyconfig.py (hunk 10 above)
Comment (bartoldeman) [reply easybuild/framework/easyconfig/easyconfig.py]: There are some ways around this of course, but e.g. moving the GCCcore logic back to the main function needs another check for dummy there too then, which is awkward...

Comment (boegel) [reply easybuild/framework/easyconfig/easyconfig.py]: No, it's fine like this, just needs a comment since it doesn't make sense unless you take the whole function into account.

Comment (bartoldeman) [reply easybuild/framework/easyconfig/easyconfig.py]: In the end I found a more elegant way by declaring the gcccore toolchain as optional. It still needs some special logic in the main loop, but det_subtoolchain_version does exactly that now and nothing more.

Next will be tests for that but probably on Monday :)

**Comment by akesandgren** [comment]
The OPTIONAL flag in the toolchain definitions, that's used in a toolchain that is optional, not in toolchains that have optional sub-toolchains?
Was trying to look at the toolchain changes and want to make sure...

**Comment by bartoldeman** [comment]
It is the toolchain itself that is optional. However I hit an issue by marking gcccore optional it marked almost all others optional too through subclassing...will correct tomorrow. Today we are all off in Canada (labour day)

**Comment by bartoldeman** [comment]
ok, I have been able to simplify the code quite a bit by giving the gcc, iccifort and pgi toolchains multiple subtoolchains (GCCcore and dummy). This way the data can do the talking which looks much more elegant than special case code.

Test case added for det_subtoolchain_version too.

File: easybuild/toolchains/gcc.py [hunk 13]
```diff
@@ -29,9 +29,11 @@
...
 
 from easybuild.toolchains.gcccore import GCCcore
+from easybuild.tools.toolchain import DUMMY_TOOLCHAIN_NAME
 
 class GccToolchain(GCCcore):
     """Simple toolchain with just the GCC compilers."""
     NAME = 'GCC'
     COMPILER_MODULE_NAME = [NAME]
-    SUBTOOLCHAIN = GCCcore.NAME
+    SUBTOOLCHAIN = [GCCcore.NAME, DUMMY_TOOLCHAIN_NAME]
```
boegel: @bartoldeman How does this relate to `--add-dummy-to-minimal-toolchains`?

`dummy` should only be considered a subtoolchain when `--add-dummy-to-minimal-toolchains` is enabled?

boegel: Actually, it's probably OK, since `dummy` was already a subtoolchain for `GCCcore` (see below)..

bartoldeman: exactly. the bfs tree has dummy twice here:
```
       gcc
     /     \
gcccore    dummy
    |
  dummy
```
which actually corresponds to pre-existing behaviour since the legacy gcc toolchains do not have gcccore. `det_toolchain_version` then explicitly filters out dummy if `--add-dummy-to-minimal-toolchains` is not set.

bartoldeman: hmm, should adjust the ascii-art diagram comment then!

easybuild/toolchains/gcccore.py
```diff
@@ -37,3 +37,4 @@ class GCCcore(Gcc):
     # Replace the default compiler module name with our own
     COMPILER_MODULE_NAME = [NAME]
     SUBTOOLCHAIN = DUMMY_TOOLCHAIN_NAME
+    OPTIONAL = True
```
boegel: @bartoldeman Can you clarify why `GCCcore` is optional?

It's the base of every toolchain, or am I overlooking something?

bartoldeman: it's not the base of the older toolchains with GCC 4.8 / intel 2015.2 and older.

easybuild/toolchains/gcc.py
boegel: Yes, indeed :)

It's basically this:

```
         gcc
       /     \
     /        \
gcccore       |
        \    /
        dummy
```

easybuild/toolchains/gcccore.py
boegel: Right, I overlooked that...

Please clarify with a comment above:

```python
# GCCcore is only guaranteed to be present in recent toolchains
# for old versions of some toolchains (GCC, intel) it is not part of the hierarchy and hence optional

**Comment by akesandgren** [comment]
Apart from the missing toolchains (esp intelcuda and friends), this looks good.

**Comment by boegel** [comment]
@akesandgren Isn't `intelcuda` handled via `iimklc.py` which is updated here?

**Comment by boegel** [comment]
nvm, changes in 345c515 make sense...

Anything else @akesandgren?

**Comment by bartoldeman** [comment]
No @boegel, @akesandgren is right. I missed intelcuda and intel-para as well.

Some others could have multiple subtoolchains except they do not exist yet. I skipped them for now:
cgmpolf/cgmvolf/cgoolf => cgolf (all Clang based toolchain easyconfigs are archived)
gmacml => gacml (not done yet, also ACML looks no longer current)
goalf => galf (not done yet, atlas is also mostly archived)
goblf => gblf (not done yet, on my plate for later)
iqacml => iacml (see above; QLogic MPI is also pretty dead)

**Comment by akesandgren** [comment]
I'm a bit tired so this is probably wrong^3, but why doesn't iccifort have both icc and ifort listed as subtoolchains

**Comment by boegel** [comment]
@akesandgren Neither `icc` nor `ifort` are toolchains by themselves currently. We can consider refactoring that, but that should be done in a separate PR.

**Comment by akesandgren** [comment]
Going in, thanks @bartoldeman

File: easybuild/framework/easyconfig/easyconfig.py [hunk 15]
```diff
@@ -134,6 +134,39 @@ def cache_aware_func(toolchain):
...
+    # dummy toolchain: bottom of the hierarchy
+    if subtoolchain_name == DUMMY_TOOLCHAIN_NAME:
+        if build_option('add_dummy_to_minimal_toolchains'):
+            subtoolchain_version = ''
+    elif len(uniq_subtc_versions) == 1:
+        subtoolchain_version = list(uniq_subtc_versions)[0]
+    elif len(uniq_subtc_versions) == 0:
+        if subtoolchain_name not in optional_toolchains:
+            # raise error if the subtoolchain considered now is not optional
+            raise EasyBuildError("No version found for subtoolchain %s in dependencies of %s",
```
ocaisa: @bartoldeman I'm hitting this error with our `intel-para` toolchain but I don't understand why since `Iimkl` is an optional toolchain
```
== 2018-09-21 14:42:03,912 build_log.py:158 ERROR EasyBuild crashed with an error (at ?:124 in __init__): No version found for subtoolchain <class 'easybuild.toolchains.iimkl.Iimkl'> in dependencies of intel-para (at easybuild/framework/easyconfig/easyconfig.py:162 in det_subtoolchain_version)
```

ocaisa: Never mind PR on the way

easybuild/toolchains/intel-para.py
```diff
@@ -39,4 +40,4 @@ class IntelPara(Ipsmpi, IntelMKL, IntelFFTW):
     Intel Math Kernel Library (MKL) and Intel FFTW wrappers.
     """
     NAME = 'intel-para'
-    SUBTOOLCHAIN = Ipsmpi.NAME 
+    SUBTOOLCHAIN = [Ipsmpi.NAME, Iimkl]
```
ocaisa: `Iimkl.NAME` slipped through the cracks

boegel: fixed in #2585

PR #2468, opened 2018-04-08
check for modulecmd.tcl before modulecmd in bootstrap script to discriminate between Modules 4.1.x vs 3.2.10

boegel:
Fix for issue reported on EasyBuild mailing list, see https://lists.ugent.be/wws/arc/easybuild/2018-04/msg00001.html

We need to check for `modulecmd.tcl` before `modulecmd`, since Modules v4.1.0 also includes a `modulecmd` wrapper.


boegel:
fix confirmed (off-list) by person who posted the issue

PR #2470, opened 2018-04-12
fix compatibility with Modules v4.1.x

boegel:
These changes are required to make EasyBuild fully compatible with Modules v4.1.x (`--modules-tool=EnvironmentModules`):

* when loading non-existing modules, the output of `modulecmd.tcl python load` used to include a `raise RuntimeError` statement which made the `load` command fail; this is no longer the case in Modules v4.1.x, so we need to customise `check_module_output` to detect failing `module` commands based on the presence of `_mlstatus = False` in the output

* we need to disable the output check for `module show` because of the above, i.e. `_mlstatus = False` is also included in the output of `module show` on a non-existent module

@xdelaruelle Please review?


xdelaruelle:
Seems ok. ``_mlstatus = False`` is now produced on stdout whether an issue occur during module command or this command needs to return a False boolean result.

PR #2474, opened 2018-04-17
make GC3Pie stop build process if a dependency failed (fix for #1441)

### LINKED ISSUE #1441: GC3Pie backend submits jobs even if dependency build failed
When robot-building a bunch of software via the GC3Pie job backend with SLURM, GC3Pie happily submits follow-up jobs even if building of one (or more) of the dependencies failed.  I am not entirely sure whether this is an issue in EasyBuild, in GC3Pie, or in the way the two are working together.  I also don't know whether this also happens with other job schedulers.

Steps to reproduce:
- Start with an empty installdir
- Edit `M4-1.4.17.eb` and add `preconfigopts='false && '`
- Run `eb GCC-5.1.0.eb -r --job` using GC3Pie with SLURM
- Both jobs will be submitted, both fail (the first for obvious reasons, the second because the M4 module can't be found)

riccardomurri:
This should fix #1441 

I have not yet tested it -- please do if you have time.

### DIFF
--- a/easybuild/tools/job/gc3pie.py
+++ b/easybuild/tools/job/gc3pie.py
@@ -51,7 +51,7 @@ try:
     from gc3libs import Application, Run, create_engine
     from gc3libs.core import Engine
     from gc3libs.quantity import hours as hr
-    from gc3libs.workflow import DependentTaskCollection
+    from gc3libs.workflow import DependentTaskCollection, AbortOnError
 
     # inject EasyBuild logger into GC3Pie
     gc3libs.log = fancylogger.getLogger('gc3pie', fname=False)
@@ -60,6 +60,16 @@ try:
 
     # instruct GC3Pie to not ignore errors, but raise exceptions instead
     gc3libs.UNIGNORE_ALL_ERRORS = True
+
+    # note: order of class inheritance is important!
+    class _BuildTaskCollection(AbortOnError, DependentTaskCollection)
+        """
+        A `DependentTaskCollection`:class: that aborts execution upon error.
+
+        This is used to stop the build process in case some dependency
+        fails.  See also `<https://github.com/easybuilders/easybuild-framework/issues/1441>`_
+        """
+        pass
 
 except ImportError as err:
     _log.debug("Failed to import gc3libs from GC3Pie."
@@ -124,7 +134,7 @@ class GC3Pie(JobBackend):
             self.config_files.append(cfgfile)
 
         self.output_dir = build_option('job_output_dir')
-        self.jobs = DependentTaskCollection(output_dir=self.output_dir)
+        self.jobs = _BuildTaskCollection(output_dir=self.output_dir)
         self.job_cnt = 0
 
         # after polling for job status, sleep for this time duration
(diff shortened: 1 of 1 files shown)


riccardomurri:
Also note: requires GC3Pie "master", I think the `AbortOnError` class was not there in 2.4.2. (But I can backport it if really needed.)

boegel:
@riccardomurri Any ETA for a GC3Pie release that is compatible with this?

I'm not sure merging this into an EasyBuild release is wise if it'll require people to run GC3Pie from `master`.
Requiring that people update to the latest GC3Pie release is another matter though...

riccardomurri:
> Any ETA for a GC3Pie release that is compatible with this?

Well, I think I could release 2.5.0 any time -- the planned features are all there.  Do you have any deadlines or wishes in this respect?

boegel:
@riccardomurri EasyBuild v3.6.0 will be released this week, that may be a good time to squeeze this in (although it's *very* last minute at this point), but then we'd need a GC3Pie release that has what is needed...

We should also bump the `REQ_VERSION` here btw...

riccardomurri:
Unfortunately I just got a pretty important bug report for GC3Pie "master", so I won't be able to release until I get that fixed.  I guess this rules out that EB v3.6.0 comes with GC3Pie 2.5.0...

boegel:
@riccardomurri OK, we'll postpone this until a future EasyBuild release then, there's no point in including this if there's no GC3Pie release that supports it imho.

boegel:
@riccardomurri Any updates on a new GC3Pie release which includes what is needed by the changes in this PR?

riccardomurri:
@boegel I'm testing the fixes -- what's the time line for EB release which should include the new version?

boegel:
@riccardomurri I'm hoping to release EasyBuild v3.6.1 next week.

akesandgren:
@riccardomurri Ping? Progress on 2.5.0?

boegel:
@riccardomurri I would really prefer having a GC3Pie release available before merging this, so we can properly check the GC3Pie version & spit out a meaningful error.

With the upcoming EasyBuild 3.7.0 release we have another window of opportunity for this (bumping the GC3Pie requirement in a bugfix release like 3.6.3 or 3.7.1 is a bit awkward).

riccardomurri:
GC3Pie 2.5.0 was just released: https://pypi.org/project/gc3pie/2.5.0/

Note that a few dependencies  were added since the 2.4.2 days:
- blinker
- dictproxyhack
- python-daemon

All of them are `pip`-installable.

riccardomurri:
Might also be needed: https://github.com/easybuilders/easybuild-framework/pull/2554

easybuild/tools/job/gc3pie.py
```diff
@@ -51,7 +51,7 @@
     from gc3libs import Application, Run, create_engine
     from gc3libs.core import Engine
     from gc3libs.quantity import hours as hr
-    from gc3libs.workflow import DependentTaskCollection
+    from gc3libs.workflow import DependentTaskCollection, AbortOnError
```
akesandgren: Please list them in alphabetical order

PR #2476, opened 2018-04-18
add support to run_cmd to enable streaming output

boegel:


boegel:
We can take this a bit further:

* enable streaming output of commands under `eb -l` (unless streaming is hard disabled)
* provide an option to enable streaming output of commands: `eb --stream-cmd-output`

Thoughts @wpoely86?

boegel:
Auto-enabled streaming output under `-l` (which was more complex than I anticipated, of course it was).

I'm having second thoughts on the `--stream-cmd-output` option, since we have a lot of "mickey mouse" commands, so I'm not sure it would make much sense to use that.

easybuild/tools/run.py
```diff
@@ -134,6 +136,14 @@ def run_cmd(cmd, log_ok=True, log_all=False, simple=False, inp=None, regexp=True
     else:
         cmd_log_fn, cmd_log = None, None
 
+    # auto-enable streaming of command output under --logtostdout/-l, unless it was disabled explicitely
+    if stream_output is None and build_option('logtostdout'):
+        _log.info("Auto-enabling streaming output of '%s' command because loggin to stdout is enabled", cmd_msg)
```
wpoely86: logging

PR #2477, opened 2018-04-18
add support for enabling fallback in sanity check to consider lib64 equivalent for seemingly missing libraries

boegel:
This is an alternate implementation of the proposed change by @torbjoernk in #1405, as a fix for #1404.

I prefer keeping this optional, since having the fallback enabled by default may result in the sanity check passing where it perhaps shouldn't (although maybe I'm being too careful here, I'm not sure...).

This is fueled by the issues that @pmatousu has been reporting for failing sanity checks on SUSE systems, cfr. https://github.com/easybuilders/easybuild-easyblocks/issues/1400 and https://github.com/easybuilders/easybuild-easyconfigs/issues/6165 .

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2147,6 +2147,21 @@ class EasyBlock(object):
         """Real version of sanity_check_step method."""
         paths, path_keys_and_check, commands = self._sanity_check_step_common(custom_paths, custom_commands)
 
+        # helper function to sanity check (alternatives for) one particular path
+        def check_path(xs, typ, check_fn):
+            """Sanity check for one particular path."""
+            found = False
+            for name in xs:
+                path = os.path.join(self.installdir, name)
+                if check_fn(path):
+                    self.log.debug("Sanity check: found %s %s in %s" % (typ, name, self.installdir))
+                    found = True
+                    break
+                else:
+                    self.log.debug("Could not find %s %s in %s" % (typ, name, self.installdir))
+
+            return found
+
         # check sanity check paths
         for key, (typ, check_fn) in path_keys_and_check.items():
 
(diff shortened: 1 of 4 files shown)


vanzod:
Very useful addition. I would even push it forward an make it two-ways, so that `lib <--> lib64`. This would simplify the sanity check in easyconfig files like the ones in [1] and [2].
Ideally I would like having it as a default behavior, despite the risks of breaking installations that rely on library directory names that are hardcoded in the easyblocks. In such cases it is a "bug" that should be addressed anyway.

[1] https://github.com/easybuilders/easybuild-easyconfigs/pull/4722
[2] https://github.com/easybuilders/easybuild-easyconfigs/pull/5897

migueldiascosta:
As a more general solution, perhaps for later, wondering if wouldn't be better to put symlinks to all libs found inside `installdir` in a single folder and only check/expose that one...

easybuild/framework/easyblock.py
```diff
@@ -2156,15 +2171,15 @@ def _sanity_check_step(self, custom_paths=None, custom_commands=None, extension=
...
-                        found = True
-                        break
-                    else:
-                        self.log.debug("Could not find %s %s in %s" % (typ, name, self.installdir))
+
+                found = check_path(xs, typ, check_fn)
+
+                # for library files in lib/, also consider fallback to lib64/ equivalent
+                if not found and build_option('lib64_fallback_sanity_check'):
+                    if all(x.startswith('lib/') for x in xs):
```
wpoely86: I don't understand this `all`? why do all paths have to start with `lib/`?

wpoely86: a wait, xs is a tuple which we do OR?

boegel: Yes, `xs` is one entry from e.g. the `files` list from `sanity_check_paths` can be a tuple (and if it's just a string, we make it a tuple with a single element higher up).

boegel:
@vanzod Reverse (also considering `lib/libfoo.a` when `lib64/libfoo.a` is not found) is now also implemented.

In addition, I enabled the lib64 fallback in the sanity check by default; it can be disabled if desired via `--disable-lib64-fallback-sanity-check`.

PR #2478, opened 2018-04-20
remove duplicates from --new-pr title

migueldiascosta:


easybuild/tools/github.py
```diff
@@ -1096,6 +1096,7 @@ def new_pr(paths, ecs, title=None, descr=None, commit_msg=None):
         elif file_info['ecs'] and all(file_info['new']) and not deleted_paths:
             # mention software name/version in PR title (only first 3)
             names_and_versions = ["%s v%s" % (ec.name, ec.version) for ec in file_info['ecs']]
+            names_and_versions = list(set(names_and_versions))  # remove duplicates
```
boegel: you can use `nub` for this, from `vsc.utils.missing` (and then you don't need two lines)

migueldiascosta: cool (it is indeed `missing` from Python :) )

boegel:
Going in, thanks @migueldiascosta!

PR #2479, opened 2018-04-20
initial work on docker containerization support

FooBarQuaxx:
This is the initial implementation of packaging easybuild buids into containers.


setup.cfg
```diff
@@ -7,3 +7,12 @@ requires = environment-modules, bash, python >= 2.6, python < 3
...
 
+[flake8]
+max-line-length = 120
+
+# Hound CI runs with Python 3 (no way around it),
+# so we need to specify some Python 2 builtins to avoid that it complains about them
+# cfr. https://stackoverflow.com/questions/47427916/how-to-config-hound-ci-to-support-python2-7
+builtins =
+    basestring,
+    reduce
```
boegel: @FooBarQuaxx Why add this here too, only belongs in `.flake8.ini`?

boegel:
@FooBarQuaxx This needs another sync with `develop`, after merge of #2483

setup.cfg
FooBarQuaxx: Sorry for the late response. `setup.cfg` according to the [documentation](http://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations) is one the default locations that `flake8` checks for configuration. I use `flake8` as part of the development workflow ( without specifying the exact location of the config file) so I've replicated that section from `.flake8.ini` into the `setup.cfg`. Since duplication is obviously not a good thing, I can (1) keep this change (since it's more standard) and remove the reference to `.flake8.ini` from the houndci config files or (2) keep using the current way.

boegel: What happens if we change `.hound.yml` to use `setup.cfg` as `config_file` rather than using `.flake8.ini`?

If that doesn't cause trouble, let's make that change (maybe in a separate PR) and drop `.flake8.ini`?

boegel:
I gave this a quick look, and I think it needs a bit of work before it can go into `develop`. My biggest suggestions would be to:

* add missing docstrings & license headers
* try and avoid the hardcoding being done (e.g. Ubuntu/CentOS versions)
* refactor the code to get rid of code duplication, both:
  * within `docker.py` (the template constants for example)
  * between `docker.py` and `singularity.py`; we should probably implement more general functions in `containers/common.py` which can then be leveraged from `docker.py` and `singularity.py`

After (finally) taking a closer look at this, I really feel this should go in first before we start tackling other issues to enhance the current containers support we have (see https://github.com/easybuilders/easybuild-framework/labels/containers).

@FooBarQuaxx Any idea if and when you would have time to work with us on that?
If you won't have time in the coming weeks that's fine, just let us know so we can build on top of this PR to get things in a state that we're happy with it and get it merged into `develop`.

FooBarQuaxx:
It seems some jobs are failing because of an unavailable python module.

boegel:
@FooBarQuaxx I've retriggered the tests, they were failing because some kind of caching issue (we released a new version of `vsc-install` very recently, which seemed to be the cause of the problems).

setup.cfg
boegel: I've made that change in #2511, together with the pure stylistic changes in `easybuild/tools/options.py`

Once that's merged, please re-sync this with `develop`

boegel:
@FooBarQuaxx I'd say we're not entirely sure yet how the integration with HPCCM is going work (cfr. #2508), so we should definitely go forward with this PR first, and then potentially rework some of that with HPCCM.

You're right that knowing that we'll refactor anyway does mean we don't need to spend a huge amount of time getting things "perfect" in this PR though.

I'll try to find time to deep dive into this ASAP and make concrete suggestions (or maybe just sned a PR to your branch), hopefully next week...

boegel:
@FooBarQuaxx Please take a look at https://github.com/FooBarQuaxx/easybuild-framework/pull/1, I'll do some more work on top of this once that PR is merged into your branch (which will update this PR).

shahzebsiddiqui:
@FooBarQuaxx @boegel I would recommend you remove the ``singularity.py`` from this PR so I can work on rewrite for singularity code based on some of the work done in this PR.

boegel:
@shahzebsiddiqui We can't remove `singularity.py` without breaking stuff (which we shouldn't)...

I hope to find time to get this merged soon, but that's proving to be difficult.

One main thing I'd like to change here is to introduce `containers/common.py` and flesh out some duplicate code in there.

FooBarQuaxx:
@boegel It will help if you give concrete direction on what pieces of code to move onto a `containers/common.py`.

boegel:
@FooBarQuaxx 

* the `containerize` function should be moved into `containers/common.py` (the `__init__.py` should basically be empty, except for the license header)

* the `singularity` and `docker_containerize` functions could probably be collapsed into a single more general function in `containers/common.py` (or even into the `containerize` function directly), which calls out to the `docker_*` or `singularity_*` specific functions

* `check_singularity` and `check_docker_containerize` could maybe also be made more generic, at least partially, e.g. into a function named `check_container_tool` in `containers/common.py` that takes the name of the tool (`docker`/`singularity`) as an argument; right now, we're checking for `sudo` in `check_docker_containerize`, but not in `check_singularity`, we should avoid that sort of inconsistency by moving things 'up' in a single function that is leveraged by both; some parts will still be specific to the tool, like the version checking, so we'll need something like:

    ```python
    def check_container_tool(cmd):
        # check for availability of <cmd>, but only if we're actually going to build an image
        ...

        # grab version and return it
        ...
    ```

    and 

    ```python
    def check_singularity():
        docker_version = check_container_tool('singularity')

        # check version
        ...
    ```

* `_det_os_deps` is not specific to Docker at all, so it belongs in `containers/common.py` and leveraged in both `docker.py` and `singularity.py` (or maybe even in another general function in `common.py` that is leveraged by both)

* `build_docker_image` and `build_singularity_image` should only compose the command to be called, and then call out to a general function `build_container_image` that lives in `containers/common.py` , determine the actual name/location of the image file, prefixes the command with `sudo` and also checks whether the expected container image file was actually created (this is also where we could later `chown` the generated container image, cfr. #2484)

* likewise for `generate_singularity_recipe` and `generate_dockerfile`: those functions should only puzzle together a dictionary with the values for the template, and then call out to a general `generate_container_recipe` function that takes a template and the dict, and actually creates the recipe file; this is where we could leverage HPCCM in the future (the template values should container what we need to pass to `hpccm.recipe()`?)

* `parse_container_base` could probably be moved into `containers/common.py`; we should strive towards a common format for `--container-base`; of course which values are valid will depend on the container tool being used, but parsing the base spec should be done with a general function

One potential issue that comes to mind is an `import` loop: if `common.py` needs stuff from `docker.py` and `singularity.py` and also the other way around, that's not going to work. So maybe we need a `containers/utils.py` with generic functions that can be used in both `docker.py` and `singularity.py`, and a `containers/common.py` where `containerize` lives?

shahzebsiddiqui:
@FooBarQuaxx can you please join easybuild slack so we can discuss further. I wanted to get your attention on these issues relating to converting recipe between singularity and docker format. This is related to the PR if we can simplify this by using existing tools to get our job done. 

https://github.com/singularityhub/singularity-cli/issues/38
https://github.com/easybuilders/easybuild-framework/issues/2509

Also I had a dockerfile for easybuild base container https://github.com/shahzebsiddiqui/eb-singularity/blob/eb_images/bootstrap/Docker/centos/7.3.1611/Dockerfile that can be built in docker cloud that way you can use ``--container-base docker`` to help bootstrap container from docker

boegel:
@shahzebsiddiqui We should not expand the scope of this PR further to what it's doing already (it will only getting this merged). Switching to `spython` should be considered later imho.

One significant limitation there is that you need to use a Docker image as a base, either the automatic conversion to/from a Docker file won't work, I'm not sure that's the way to go...

FooBarQuaxx:
@boegel I've done some refactoring according to your review on a branch other that this PR's branch. Please give feedback on whether the layout of the code is any better. Note that I intentional left out tests and docstring/headers. I didn't wanted to bother creating those before getting your opinion about the shape of new code. The commit is [here].(https://github.com/FooBarQuaxx/easybuild-framework/commit/68494633b9fb1099125cb45d885681c65c2a35f9) and the branch name on my fork is named [docker-packaging-refactoring](https://github.com/FooBarQuaxx/easybuild-framework/tree/docker-packaging-refactoring).

boegel:
@FooBarQuaxx Any updates?

FooBarQuaxx:
@boegel The tests are green now. It would be time to give any review for the structure. Afterward, I will try to add some docstring.

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __init__(self, easyconfigs):
+        self._easyconfigs = easyconfigs
+        self._img_name = build_option('container_image_name')
+        self._force = build_option('force')
+        self._image_format = build_option('container_image_format')
+        self._tmpdir = build_option('container_tmpdir')
+        self._container_build_image = build_option('container_build_image')
+        self._container_base = build_option('container_base')
+        self._container_path = container_path()
```
boegel: @FooBarQuaxx Any specific reason you're making these 'hidden' via `_`?

I see no need for it, and dropping the `_` helps with readability imho...

Style nitpicking: please sort these initializations alphabetically

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def generate(self):
+        self.validate()
+        recipe_path = self.generate_recipe()
+        if self._container_build_image:
+            self.build_image(recipe_path)
+
+    def validate(self):
+        if not self._container_build_image:
+            return
```
boegel: Hmm, wrong place, only call `validate` when `container_build_image` is `True` in `generate` instead?

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,103 @@
...
+"""
+:author: Mohamed Abidi (Bright Computing)
+"""
+from vsc.utils import fancylogger
+
+import os
+from easybuild.tools.filetools import write_file
+from easybuild.tools.config import build_option, container_path
+from easybuild.tools.build_log import EasyBuildError, print_msg
+from .utils import check_tool
```
boegel: @FooBarQuaxx We don't use relative imports anywhere, so make this `from easybuild.tools.containers.utils` for consistency?

Also, please sort alphabetically (`build_log`, `config`, `containers`, `filetools`)

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+:author: Mohamed Abidi (Bright Computing)
+"""
+from vsc.utils import fancylogger
+
+import os
```
boegel: this should go higher up, above the `from vsc...` onw

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,103 @@
...
+            if not check_tool(tool_name, tool_version):
+                raise EasyBuildError("{0} not found on your system.".format(tool_name,))
+
+    def resolve_template(self):
+        raise NotImplementedError
+
+    def resolve_template_data(self):
+        return {}
+
+    def _write_recipe(self, template, data):
```
boegel: same here, no reason to make this hidden?

any reason this is not collapsed under `generate_recipe` (which is basically the same as `_write_recipe` but just two extra lines)?

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,103 @@
...
+class ContainerGenerator(object):
+
+    TOOLS = {}
+
+    RECIPE_FILE_NAME = None
+
+    def __init__(self, easyconfigs):
+        self._easyconfigs = easyconfigs
+        self._img_name = build_option('container_image_name')
+        self._force = build_option('force')
```
boegel: not sure if it's worth creating a class variable for this, just use `if build_option('force')` below?

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,103 @@
...
+        if os.path.exists(recipe_path):
+            if self._force:
+                print_msg("WARNING: overwriting existing container recipe at %s due to --force" % recipe_path)
+            else:
+                raise EasyBuildError("Container recipe at %s already exists, not overwriting it without --force",
+                                     recipe_path)
+
+        recipe_content = template % data
+        write_file(recipe_path, recipe_content)
+        print_msg("%s definition file created at %s" % (self.RECIPE_FILE_NAME, recipe_path), log=_log)
```
boegel: use `self.log` here instead

easybuild/tools/containers/common.py
```diff
@@ -0,0 +1,57 @@
...
+:author: Shahzeb Siddiqui (Pfizer)
+:author: Kenneth Hoste (HPC-UGent)
+:author: Mohamed Abidi (Bright Computing)
+"""
+from vsc.utils import fancylogger
+
+from easybuild.tools.config import build_option
+from easybuild.tools.build_log import EasyBuildError
+from .docker import DockerContainer  # noqa
+from .singularity import SingularityContainer  # noqa
```
boegel: please use absolute imports, just for consistency with existing `import` statements, and sort alphabetically

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+
+from easybuild.framework.easyconfig.easyconfig import ActiveMNS
+from easybuild.tools.build_log import print_msg, EasyBuildError
+from easybuild.tools.config import build_option
+from easybuild.tools.config import DOCKER_BASE_IMAGE_UBUNTU, DOCKER_BASE_IMAGE_CENTOS, DEFAULT_DOCKER_BASE_IMAGE
+from easybuild.tools.run import run_cmd
+from .base import ContainerGenerator
+from .utils import det_os_deps
+
+_log = fancylogger.getLogger('tools.containers.singularity')  # pylint: disable=C0103
```
boegel: `tools.containers.docker`

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+
+from vsc.utils import fancylogger
+
+from easybuild.framework.easyconfig.easyconfig import ActiveMNS
+from easybuild.tools.build_log import print_msg, EasyBuildError
+from easybuild.tools.config import build_option
+from easybuild.tools.config import DOCKER_BASE_IMAGE_UBUNTU, DOCKER_BASE_IMAGE_CENTOS, DEFAULT_DOCKER_BASE_IMAGE
+from easybuild.tools.run import run_cmd
+from .base import ContainerGenerator
+from .utils import det_os_deps
```
boegel: no relative imports please (& sort alphabetically)

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+
+_log = fancylogger.getLogger('tools.containers.singularity')  # pylint: disable=C0103
+
+DOCKER_UBUNTU1604_TMPL = """\
+FROM ubuntu:16.04
+LABEL maintainer=mohamed.abidi@brightcomputing.com
```
boegel: We shouldn't hardcode your email address here?

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+from easybuild.tools.config import build_option
+from easybuild.tools.config import DOCKER_BASE_IMAGE_UBUNTU, DOCKER_BASE_IMAGE_CENTOS, DEFAULT_DOCKER_BASE_IMAGE
+from easybuild.tools.run import run_cmd
+from .base import ContainerGenerator
+from .utils import det_os_deps
+
+_log = fancylogger.getLogger('tools.containers.singularity')  # pylint: disable=C0103
+
+DOCKER_UBUNTU1604_TMPL = """\
+FROM ubuntu:16.04
```
boegel: I don't like this hardcoding (especially since we have constants for this too), can we avoid this easily, for example by introducing another configuration option?

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+    echo '\\n# Added by easybuild docker packaging' >> ${HOME}/.profile && \\
+    echo 'source /usr/share/lmod/lmod/init/bash' >> ${HOME}/.profile && \\
+    echo 'module use %(init_modulepath)s' >> ${HOME}/.profile && \\
+    echo 'module load %(mod_names)s' >> ${HOME}/.profile
+
+CMD ["/bin/bash", "-l"]
+"""
+
+DOCKER_CENTOS7_TMPL = """\
+FROM centos:7
```
boegel: Again hardcoding even though we have a constant for this...

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+LABEL maintainer=mohamed.abidi@brightcomputing.com
```
boegel: Don't hardcode your e-mail address please...

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+
+RUN pip install -U pip setuptools && \\
+    hash -r pip && \\
+    pip install -U easybuild
+
+RUN OS_DEPS='%(os_deps)s' && \\
+    test -n "${OS_DEPS}" && \\
+    yum --skip-broken install -y "${OS_DEPS}" || true
+
+RUN mkdir /app && \\
```
boegel: This last part is entirely duplicate, can we flesh this out as `DOCKER_COMMON_TMPL` and then define `DOCKER_CENTOS7_TMPL` as `"""...""" + DOCKER_COMMON_TMPL` (and the same for `DOCKER_UBUNTU1604_TMPL`)?

Or maybe use `DOCKER_CENTOS7_TMPL_HEADER` together with `DOCKER_COMMON_TMPL` in `resolve_template` below?

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+
+class DockerContainer(ContainerGenerator):
+
+    TOOLS = {'docker': None, 'sudo': None}
+
+    RECIPE_FILE_NAME = 'Dockerfile'
+
+    def __init__(self, *args, **kwargs):
+        super(DockerContainer, self).__init__(*args, **kwargs)
+        # NOTE (med): set default value for _container_base
```
boegel: Is this fixed yet?

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+        self._container_base = self._container_base or DEFAULT_DOCKER_BASE_IMAGE
+
+    def resolve_template(self):
+        return _DOCKER_TMPLS[self._container_base]
```
boegel: No `try`/`except` to avoid nasty crash in case of `KeyError`?

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def resolve_template_data(self):
+        os_deps = det_os_deps(self._easyconfigs)
+
+        module_naming_scheme = ActiveMNS()
+
+        ec = self._easyconfigs[-1]['ec']
+
+        init_modulepath = os.path.join("/app/modules/all", *module_naming_scheme.det_init_modulepaths(ec))
```
boegel: We'll need to flesh out the hardcoded `/app/...` here, but I guess that can be done later...

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+
+        mod_names = [e['ec'].full_mod_name for e in self._easyconfigs]
+
+        eb_opts = [os.path.basename(ec['spec']) for ec in self._easyconfigs]
```
boegel: please be consistent (either `for ec in` or `for e in`)

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+
+        tempdir = tempfile.mkdtemp(prefix='easybuild-docker')
+        container_name = build_option('container_image_name') or "%s:latest" % module_name.replace('/', '-')
+        docker_cmd = ' '.join(['sudo', 'docker', 'build', '-f', dockerfile, '-t', container_name, '.'])
+
+        print_msg("Running '%s', you may need to enter your 'sudo' password..." % docker_cmd)
+        run_cmd(docker_cmd, path=tempdir, stream_output=True)
+        print_msg("Docker image created at %s" % container_name, log=_log)
+
+        shutil.rmtree(tempdir)
```
boegel: please use `rmtree2` from `easybuild.tools.filetools` (the function name is silly, but it does more than just plain `rmtree`)

easybuild/tools/containers/utils.py
```diff
@@ -0,0 +1,73 @@
...
+from distutils.version import LooseVersion
+from easybuild.tools.filetools import which
+from easybuild.tools.build_log import print_msg
+from easybuild.tools.run import run_cmd
+from easybuild.tools.build_log import EasyBuildError
+
+
+def det_os_deps(easyconfigs):
+    res = set()
+    _os_deps = reduce(operator.add, [obj['ec']['osdependencies'] for obj in easyconfigs], [])
```
boegel: no need to name it `_os_deps`, since it's a local variable, `os_deps` is fine?

easybuild/tools/containers/utils.py
```diff
@@ -0,0 +1,73 @@
...
+            res.update(os_dep)
+    return res
+
+
+def check_tool(tool_name, min_tool_version=None):
+    tool_path = which(tool_name)
+    if not tool_path:
+        return False
+
+    print_msg("{0} tool found at {1}".format(tool_name.capitalize(), tool_path))
```
boegel: @FooBarQuaxx I wouldn't use capitalize here, don't see why?

easybuild/tools/containers/utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+    if not min_tool_version:
+        return True
+
+    out, ec = run_cmd("{0} --version".format(tool_name), simple=False, trace=False, force_in_dry_run=True)
+    if ec:
+        return False
```
boegel: some logging would be nice here (show the error message that occurred when running `--version`), or even raise an error?

test/framework/containers.py
```diff
@@ -89,11 +99,11 @@ def test_parse_container_base(self):
             expected.update({'arg2': 'bar'})
             self.assertEqual(parse_container_base('%s:foo:bar' % agent), expected)
 
-    def run_main(self, args):
+    def run_main(self, args, raise_error=False):
```
boegel: @FooBarQuaxx Leave the default to `True` for `raise_error`?

test/framework/containers.py
```diff
@@ -120,6 +130,7 @@ def test_end2end_singularity_recipe(self):
         args = [
             toy_ec,
             '--containerize',
+            '--container-type=singularity',
```
boegel: This shouldn't be needed (since default type is Singularity)?

easybuild/tools/containers/docker.py
FooBarQuaxx: Any suggestion about what to put for that.

easybuild/tools/containers/docker.py
FooBarQuaxx: Same as above, is there a way to get this right without hardcoding the email value?

easybuild/tools/containers/docker.py
FooBarQuaxx: It's an explanation comment. Nothing should be fixed.

easybuild/tools/containers/docker.py
FooBarQuaxx: I'm assuming that the validation will shield the access from any invalid key access.

easybuild/tools/containers/utils.py
FooBarQuaxx: I see that's not needed, but the testing code was using a capitalized string for the docker format. I can get rid of that.

test/framework/containers.py
FooBarQuaxx: I thought that will be explicit and clearer.

easybuild/tools/containers/docker.py
boegel: Not sure, to be honest... Does this have to be an e-mail address? If not, how about `EasyBuild` or `easybuild@lists.ugent.be` (the EasyBuild mailing list)?

easybuild/tools/containers/utils.py
boegel: Not capitalizing makes more sense to me

easybuild/tools/containers/docker.py
boegel: The only way I see to hardcode anything is to provide a configuration setting for this... `--container-maintainer`?

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,179 @@
...
+class DockerContainer(ContainerGenerator):
+
+    TOOLS = {'docker': None, 'sudo': None}
+
+    RECIPE_FILE_NAME = 'Dockerfile'
+
+    def __init__(self, *args, **kwargs):
+        super(DockerContainer, self).__init__(*args, **kwargs)
+        # NOTE (med): set default value for _container_base
+        self._container_base = self._container_base or DEFAULT_DOCKER_BASE_IMAGE
```
boegel: @FooBarQuaxx Isn't this something that should be covered in `tools/config.py`, where we define defaults?

easybuild/tools/containers/docker.py
boegel: That's probably sufficient, yes, and I guess if this is a problem it will surface quite easily in the tests, so OK.

test/framework/containers.py
boegel: Sure, but this way you're not testing what the default container type is, since it's always specified.

Leaving it unspecified is intentional here, it serves as a check that Singularity is the default.

easybuild/tools/containers/docker.py
FooBarQuaxx: The current state of the `--container-base` option is kind of non consistent. For singularity, the option does not allow for an empty value so it does not default to anything. But for docker, we're defaulting to ubuntu:1604. I'll change it to also be require for docker too.

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,100 @@
...
+    RECIPE_FILE_NAME = None
+
+    def __init__(self, easyconfigs):
+        self.container_base = build_option('container_base')
+        self.container_build_image = build_option('container_build_image')
+        self.container_path = container_path()
+        self.easyconfigs = easyconfigs
+        self.image_format = build_option('container_image_format')
+        self.img_name = build_option('container_image_name')
+        self.log = fancylogger.getLogger(self.__class__.__module__.replace('easybuild.', ''))
```
boegel: please initialize the logger with `fancylogger.getLogger(self.__class__.__name__, fname=False)` like we do elsewhere

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,100 @@
...
+from easybuild.tools.filetools import write_file
+
+
+class ContainerGenerator(object):
+
+    TOOLS = {}
+
+    RECIPE_FILE_NAME = None
+
+    def __init__(self, easyconfigs):
```
boegel: docstrings are still missing?

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class DockerContainer(ContainerGenerator):
+
+    TOOLS = {'docker': None, 'sudo': None}
+
+    RECIPE_FILE_NAME = 'Dockerfile'
+
+    def resolve_template(self):
+        return (2 * "\n").join([
```
boegel: style nitpicking: I'd go with `'\n\n'`, looks a bit weird otherwise...

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,153 @@
...
+
+        return {
+            'os_deps': ' '.join(os_deps),
+            'eb_opts': ' '.join(eb_opts),
+            'init_modulepath': init_modulepath,
+            'mod_names': ' '.join(mod_names),
+        }
+
+    def validate(self):
+        if self.container_base not in [DOCKER_BASE_IMAGE_UBUNTU, DOCKER_BASE_IMAGE_CENTOS]:
```
boegel: @FooBarQuaxx does it make sense to use `DOCKER_OS_INSTALL_DEPS_TMPLS.keys()` here?

easybuild/tools/containers/singularity.py
```diff
@@ -0,0 +1,233 @@
...
+
+:author: Shahzeb Siddiqui (Pfizer)
+:author: Kenneth Hoste (HPC-UGent)
+:author: Mohamed Abidi (Bright Computing)
+"""
+import os
+
+from easybuild.tools.build_log import EasyBuildError, print_msg
+from easybuild.tools.config import CONT_IMAGE_FORMAT_EXT3, CONT_IMAGE_FORMAT_SANDBOX, CONT_IMAGE_FORMAT_SQUASHFS
+from easybuild.tools.config import container_path, build_option
```
boegel: style nitpicking: fix sorting here, `build_option` goes first

easybuild/tools/containers/utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+    print_msg("{0} tool found at {1}".format(tool_name, tool_path))
+
+    if not min_tool_version:
+        return True
+
+    out, ec = run_cmd("{0} --version".format(tool_name), simple=False, trace=False, force_in_dry_run=True)
+    if ec:
+        print_error("command '{0} --version' failed with output: {1}".format(tool_name, out))
```
boegel: should be `raise EasyBuildError`, since `print_error` is not fatal?

cfr. original behaviour in `check_singularity`

easybuild/tools/containers/utils.py
```diff
@@ -0,0 +1,73 @@
...
+        return False
+    res = re.search("\d+\.\d+(\.\d+)?", out.strip())
+    if not res:
+        raise EasyBuildError("Error parsing version for tool %s" % tool_name)
+    tool_version = res.group(0)
+    version_ok = LooseVersion(str(min_tool_version)) <= LooseVersion(tool_version)
+    if version_ok:
+        print_msg("{0} version '{1}' is {2} or higher ... OK".format(tool_name,
+                                                                     tool_version,
+                                                                     min_tool_version))
```
boegel: no more need to spread this across multiple lines?

test/framework/containers.py
```diff
@@ -138,7 +148,7 @@ def test_end2end_singularity_recipe(self):
             remove_file(os.path.join(self.test_prefix, 'containers', 'Singularity.toy-0.0'))
 
         args.append("--container-base=shub:test123")
-        self.run_main(args)
+        self.run_main(args, raise_error=True)
```
boegel: default is `True` for `raise_error`, so this change is not needed

test/framework/containers.py
```diff
@@ -282,6 +292,107 @@ def test_end2end_singularity_image(self):
...
+                              raise_error=True)
+
+        for cont_base in ['ubuntu:16.04', 'centos:7']:
+            stdout, stderr = self.run_main(base_args + ['--container-base=%s' % cont_base])
+            self.assertFalse(stderr)
+            regexs = ["^== Dockerfile definition file created at %s/containers/Dockerfile.toy-0.0" % self.test_prefix]
+            self.check_regexs(regexs, stdout)
+            remove_file(os.path.join(self.test_prefix, 'containers', 'Dockerfile.toy-0.0'))
+
+        self.run_main(base_args + ['--container-base=centos:7'], raise_error=True)
```
boegel: default is `True` for `raise_error`, no need to specify it

test/framework/containers.py
```diff
@@ -282,6 +292,107 @@ def test_end2end_singularity_image(self):
...
+        self.assertErrorRegex(EasyBuildError,
+                              error_pattern,
+                              self.eb_main,
+                              base_args + ['--container-base=centos:7'],
+                              raise_error=True)
+
+        remove_file(os.path.join(self.test_prefix, 'containers', 'Dockerfile.toy-0.0'))
+
+        base_args.insert(1, os.path.join(test_ecs, 'g', 'GCC', 'GCC-4.9.2.eb'))
+        self.run_main(base_args + ['--container-base=ubuntu:16.04'], raise_error=True)
```
boegel: default is `True` for `raise_error`, no need to specify it

test/framework/containers.py
```diff
@@ -282,6 +292,107 @@ def test_end2end_singularity_image(self):
         stdout, stderr = self.run_main(args)
         self.assertFalse(stderr)
         regexs[-3] = "^== Running 'sudo\s*SINGULARITY_TMPDIR=%s \S*/singularity build .*" % self.test_prefix
```
boegel: missing `self.check_regexs(regexs, stdout)`?

test/framework/containers.py
```diff
@@ -282,6 +292,107 @@ def test_end2end_singularity_image(self):
...
+            self.check_regexs(regexs, stdout)
+            remove_file(os.path.join(self.test_prefix, 'containers', 'Dockerfile.toy-0.0'))
+
+        self.run_main(base_args + ['--container-base=centos:7'], raise_error=True)
+
+        error_pattern = "Container recipe at %s/containers/Dockerfile.toy-0.0 already exists, " \
+                        "not overwriting it without --force" % self.test_prefix
+        self.assertErrorRegex(EasyBuildError,
+                              error_pattern,
+                              self.eb_main,
```
boegel: wouldn't it be better to use `self.run_main` here?

test/framework/containers.py
```diff
@@ -282,6 +292,107 @@ def test_end2end_singularity_image(self):
...
+        base_args.insert(1, os.path.join(test_ecs, 'g', 'GCC', 'GCC-4.9.2.eb'))
+        self.run_main(base_args + ['--container-base=ubuntu:16.04'], raise_error=True)
+        def_file = read_file(os.path.join(self.test_prefix, 'containers', 'Dockerfile.toy-0.0'))
+        regexs = [
+            "FROM ubuntu:16.04",
+            "eb toy-0.0.eb GCC-4.9.2.eb",
+            "module load toy/0.0 GCC/4.9.2",
+        ]
+        self.check_regexs(regexs, def_file)
+        remove_file(os.path.join(self.test_prefix, 'containers', 'Dockerfile.toy-0.0'))
```
boegel: not needed strictly speaking, the test-specific temporary directory at `self.test_prefix` is automatically cleaned up for each test

easybuild/tools/containers/singularity.py
```diff
@@ -0,0 +1,233 @@
...
+module load %(mod_names)s
+
+%%labels
+
+"""
+
+
+class SingularityContainer(ContainerGenerator):
+
+    TOOLS = {'singularity': '2.4'}
```
boegel: @FooBarQuaxx `sudo` is also required here?

easybuild/tools/containers/base.py
```diff
@@ -0,0 +1,100 @@
...
+    def generate(self):
+        self.validate()
+        recipe_path = self.generate_recipe()
+        if self.container_build_image:
+            self.build_image(recipe_path)
+
+    def validate_tools(self):
+        for tool_name, tool_version in self.TOOLS.items():
+            if not check_tool(tool_name, tool_version):
+                raise EasyBuildError("{0} not found on your system.".format(tool_name,))
```
boegel: Maybe the tool is there, but it's not the correct version?

easybuild/tools/containers/docker.py
```diff
@@ -0,0 +1,153 @@
...
+            DOCKER_TMPL_HEADER % {'container_base': self.container_base},
+            DOCKER_OS_INSTALL_DEPS_TMPLS[self.container_base],
+            DOCKER_INSTALL_EASYBUILD,
+            DOCKER_TMPL_FOOTER,
+        ])
+
+    def resolve_template_data(self):
+        os_deps = det_os_deps(self.easyconfigs)
+
+        module_naming_scheme = ActiveMNS()
```
boegel: since we need this in two places, should we define a `self.mns` in `__init__`?

easybuild/tools/containers/utils.py
```diff
@@ -0,0 +1,73 @@
...
+import operator
+import re
+
+from distutils.version import LooseVersion
+from easybuild.tools.build_log import EasyBuildError, print_error, print_msg
+from easybuild.tools.filetools import which
+from easybuild.tools.run import run_cmd
+
+
+def det_os_deps(easyconfigs):
```
boegel: more missing docstrings here

easybuild/tools/containers/utils.py
FooBarQuaxx: `check_tool` is a predicate (returns a boolean) so raising an exception is IMO the responsibility of the caller.

test/framework/containers.py
FooBarQuaxx: I was just using the value that's used elsewhere on the code.

easybuild/tools/containers/base.py
FooBarQuaxx: Do you suggest changing the message or being more precise about the non valid condition?

boegel: I would change the message to also include the version requirement (if there is one)

easybuild/tools/containers/utils.py
boegel: @FooBarQuaxx We also raise an error when we fail to parse the version...

If running `--version` doesn't work, there's something seriously wrong, so the right thing to do is to exit with an error straight away I think...

test/framework/containers.py
boegel: I understand, but we introduced `run_main` with a reason, so we should try and be consistent to use it...

Maybe also change it in other places where `self.eb_main` is used (but make sure the test is still the same)

easybuild/tools/containers/utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+    print_msg("{0} tool found at {1}".format(tool_name, tool_path))
+
+    if not min_tool_version:
+        return True
+
+    version_cmd = "{0} --version".format(tool_name)
+    out, ec = run_cmd(version_cmd, simple=False, trace=False, force_in_dry_run=True)
+    if ec:
+        print_error("command '{0}' failed with output: {1}".format(version_cmd, out))
```
boegel: no more need for `print_error` if an error is being raised?

boegel:
This looks good to go, thanks a lot for all your efforts here @FooBarQuaxx!

There's certainly room for more improvement, but I don't see a need to hold this PR back any further, so I'll merge it so these changes are included in EasyBuild v3.6.2 .

Further enhancements can be done step by step, following the relevant issues in https://github.com/easybuilders/easybuild-framework/labels/containers.

An update to the documentation (http://easybuild.readthedocs.io/en/latest/Containers.html, via https://github.com/easybuilders/easybuild/tree/master/docs) would make sense too, if you're up for it @FooBarQuaxx ...

cc @shahzebsiddiqui

FooBarQuaxx:
@boegel Thanks for assisting and accepting my contribution to this awesome project.

PR #2483, opened 2018-04-25
add option to define SINGULARITY_TMPDIR

migueldiascosta:
Singularity defaults to building the images in `/tmp` (http://singularity.lbl.gov/build-environment#temporary-folders), which will fail if it is not large enough

Additionally, this environment variable must be defined in the root shell where singularity is run, via sudo.

This adds options to define the path as passes it to singularity.

cc @shahzebsiddiqui @boegel

### DIFF
--- a/easybuild/tools/containers.py
+++ b/easybuild/tools/containers.py
@@ -245,7 +245,9 @@ def build_singularity_image(def_path):
     # resolve full path to 'singularity' binary, since it may not be available via $PATH under sudo...
     singularity = which('singularity')
 
-    cmd = "sudo %s build %s %s %s" % (singularity, cmd_opts, img_path, def_path)
+    singularity_tmp = build_option('container_tmp_path')
+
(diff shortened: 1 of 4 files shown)


easybuild/tools/containers.py
```diff
@@ -245,7 +245,9 @@ def build_singularity_image(def_path):
     # resolve full path to 'singularity' binary, since it may not be available via $PATH under sudo...
     singularity = which('singularity')
 
-    cmd = "sudo %s build %s %s %s" % (singularity, cmd_opts, img_path, def_path)
+    singularity_tmp = build_option('container_tmp_path')
```
boegel: @migueldiascosta What if this is not specified, you'll inject `SINGULARITY_TMPDIR=None` into the command?

Fixes in https://github.com/migueldiascosta/easybuild-framework/pull/1

migueldiascosta: @boegel you're right, I assumed it would be empty and singularity would default to `/tmp` (and thought I had tested it but apparently hadn't)

boegel:
Thanks @migueldiascosta!

PR #2490, opened 2018-04-27
Fix git remote url in CONTRIBUTING.md

SethosII:
Otherwise `git` fails  when using the remote:

```
$ git pull github_easybuilders develop
fatal: 'git@github.com/easybuilders/easybuild-framework.git' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

The url is correct in the other EasyBuild repositories.


boegel:
:+1: thanks @SethosII!

PR #2493, opened 2018-04-29
add 'exts_download_dep_fail' as known easyconfig parameter

boegel:
For https://github.com/easybuilders/easybuild-easyblocks/pull/1377, so it leveraged for Python packages that are installed as extensions (e.g. via `Bundle` easyblock).

This is required because variables defined in an easyconfig file that are not known easyconfig parameters are thrown away, and thus they can't be inherited by extensions...

### DIFF
--- a/easybuild/framework/easyconfig/default.py
+++ b/easybuild/framework/easyconfig/default.py
@@ -151,6 +151,7 @@ DEFAULT_CONFIG = {
     'license_server_port': [None, 'Port for license server', LICENSE],
 
     # EXTENSIONS easyconfig parameters
+    'exts_download_dep_fail': [False, "Fail if downloaded dependencies are detected for extensions", EXTENSIONS],
     'exts_classmap': [{}, "Map of extension name to class for handling build and installation.", EXTENSIONS],
     'exts_defaultclass': [None, "List of module for and name of the default extension class", EXTENSIONS],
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyconfig/default.py
```diff
@@ -40,25 +40,25 @@
...
-    'MODULES': (8, 'modules'),
-    'OTHER': (9, 'other'),
-}
-# define constants so they can be used below
-# avoid that pylint complains about unknown variables in this file
-# pylint: disable=E0602
-globals().update(ALL_CATEGORIES)
+CATEGORY_NAMES = ['BUILD', 'CUSTOM', 'DEPENDENCIES', 'EXTENSIONS', 'FILEMANAGEMENT', 'HIDDEN',
+                  'LICENSE', 'MANDATORY', 'MODULES', 'OTHER', 'TOOLCHAIN']
+ALL_CATEGORIES = dict((name, eval(name)) for name in CATEGORY_NAMES)
```
wpoely86: why change this?

boegel: Because `flake8` (rightfully) doesn't like us defining constants at runtime by injecting stuff into `globals()`.

If you want me to, I can flesh this out in a separate PR.

PR #2494, opened 2018-04-29
add support for passing custom messages on failing sanity check for extensions

boegel:
This is motivated by a suggestion by @akesandgren in the context of https://github.com/easybuilders/easybuild-easyblocks/pull/1377, to postpone raising of errors at the end, which helps with limiting a lot of back & forth when the same type of problems arise for multiple extensions.

One issue with moving the download check in https://github.com/easybuilders/easybuild-easyblocks/pull/1377 to the sanity check step is error reporting: the actual cause of the failing sanity check for an extension is hidden higher up in the log.
This is fixed by the (backwards-compatible) changes made in this PR.

For example, with current `develop` you would get an error message like:

```
Sanity check failed: sanity checks for ['ipython', 'pyzmq'] extensions failed!
```

(and then you had to go back in the log to figure out the actual problem)

With these changes, you'll see:

```
Sanity check failed: extensions sanity check failed for 2 extensions: ipython, pyzmq
failing sanity check for 'ipython' extension: command "python -c 'import ipython'" failed; output:
ModuleNotFoundError: No module named 'ipython'
failing sanity check for 'pyzmq' extension: command "python -c 'import pyzmq'" failed; output:
ModuleNotFoundError: No module named 'pyzmq'
```

cc @vanzod, @wpoely86

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2202,6 +2202,7 @@ class EasyBlock(object):
 
         # sanity check for extensions
         if not extension:
+            failed_exts = []
             for ext in self.ext_instances:
                 fail_msg = None
                 res = ext.sanity_check_step()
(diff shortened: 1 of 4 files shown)


boegel:
@wpoely86 Good to go?

Issue #2495, opened 2018-05-02
error when using eb --new-pr


smoors:
when I create a PR with `eb --new-pr`, it works, but then I get this error:
```
== temporary log file in case of crash /tmp/smoors/eb-lwSRJY/easybuild-DTwRYo.log
== fetching branch 'develop' from https://github.com/easybuilders/easybuild-easyconfigs.git...
== copying easyconfigs to /tmp/smoors/eb-lwSRJY/git-working-dira93jO_/easybuild-easyconfigs...
== pushing branch '20180502115945_new_pr_SLEPc383' to remote 'github_smoors_JmkkH' (git@github.com:smoors/easybuild-easyconfigs.git)

Opening pull request
* target: easybuilders/easybuild-easyconfigs:develop
* from: smoors/easybuild-easyconfigs:20180502115945_new_pr_SLEPc383
* title: "{numlib}[foss/2017b] SLEPc v3.8.3"
* labels: update
* description:
"""
(created using `eb --new-pr`)

"""
* overview of changes:
 easybuild/easyconfigs/s/SLEPc/SLEPc-3.8.3-foss-2017b.eb | 21 +++++++++++++++++++++
 1 file changed, 21 insertions(+)

Opened pull request: https://github.com/easybuilders/easybuild-easyconfigs/pull/6260
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/gpfs/work/smoors/hpc/easybuild-dev/easybuild-framework/easybuild/main.py", line 510, in <module>
    main()
  File "/gpfs/work/smoors/hpc/easybuild-dev/easybuild-framework/easybuild/main.py", line 427, in main
    commit_msg=options.pr_commit_msg)
  File "/gpfs/work/smoors/hpc/easybuild-dev/easybuild-framework/easybuild/tools/github.py", line 1151, in new_pr
    status, data = pr_url.labels.post(body=labels)
  File "/work/smoors/hpc/easybuild-dev/vsc-base/lib/vsc/utils/rest.py", line 144, in post
    return self.request(self.POST, url, json.dumps(body), headers, content_type='application/json')
  File "/work/smoors/hpc/easybuild-dev/vsc-base/lib/vsc/utils/rest.py", line 175, in request
[...]

PR #2497, opened 2018-05-03
handle missing permissions for adding labels in --new-pr

migueldiascosta:
mitigates #2495

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1148,11 +1148,13 @@ def new_pr(paths, ecs, title=None, descr=None, commit_msg=None):
             # post labels
             pr = data['html_url'].split('/')[-1]
             pr_url = g.repos[pr_target_account][pr_target_repo].issues[pr]
-            status, data = pr_url.labels.post(body=labels)
-            if not status == HTTP_STATUS_OK:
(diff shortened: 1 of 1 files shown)


easybuild/tools/github.py
```diff
@@ -1148,11 +1148,13 @@ def new_pr(paths, ecs, title=None, descr=None, commit_msg=None):
...
-            if not status == HTTP_STATUS_OK:
-                raise EasyBuildError("Failed to add labels to PR# %s; status %s, data: %s", pr, status, data)
-
-            print_msg("Added labels %s to PR#%s" % (', '.join(labels), pr), log=_log, prefix=False)
+            try:
+                status, data = pr_url.labels.post(body=labels)
+                if status == HTTP_STATUS_OK:
+                    print_msg("Added labels %s to PR#%s" % (', '.join(labels), pr), log=_log, prefix=False)
+            except urllib2.HTTPError as err:
+                print_msg("Failed to add labels to PR# %s: %s. A maintainer will do it later." % (pr, err),
```
boegel: @migueldiascosta I'm not sure whether it's worth printing the warning message, since there's no way the user can fix this if they're not a maintainer... Maybe just logging with `_log.info` is enough?

boegel:
Thanks for the fix @migueldiascosta!

Issue #2503, opened 2018-05-17
Autogenerated EasyConfig overwrites existing file


Flamefire:
The autogenerated EasyConfig (by tweak.py) does not seem to check for existence of the file to generate. So if there is a (maybe not exactly matching) EasyConfig, then this gets overwritten.

This just happened to me:

- Create GPI2-1.3.0-GCCcore-6.4.0.eb in current directory
- Run eb --software=GPI2,1.3.0 --toolchain=GCCcore,6.4.0 -r
- EB fails with `Unable to find an easyconfig for the given specifications`
- GPI2-1.3.0-GCCcore-6.4.0.eb is missing !!!

Log output:

> 
> == 2018-05-17 16:38:53,732 tweak.py:410 INFO No easyconfig file for specified toolchain, but template is available.
> == 2018-05-17 16:38:53,735 tweak.py:176 WARNING Tweaking version: checksums cleared, verification disabled.
> == 2018-05-17 16:38:53,737 tweak.py:231 INFO Tweaked 'version' to '"1.3.0"'
> == 2018-05-17 16:38:53,738 tweak.py:231 INFO Tweaked 'toolchain' to '{'version': '6.4.0', 'name': 'GCCcore'}'
> == 2018-05-17 16:38:53,739 tweak.py:231 INFO Tweaked 'name' to '"GPI2"'
> == 2018-05-17 16:38:53,740 tweak.py:267 INFO Tweaked easyconfig file written to GPI2-1.3.0-GCCcore-6.4.0.eb
> == 2018-05-17 16:38:53,741 tweak.py:571 INFO Generated easyconfig file GPI2-1.3.0-GCCcore-6.4.0.eb, and using it to build the requested software.
> == 2018-05-17 16:38:53,798 build_log.py:158 ERROR EasyBuild crashed with an error (at ?:124 in __init__): Unable to find an easyconfig for the given specifications: {'toolchain': {'version': '6.4.0', 'nam    e': 'GCCcore'}, 'version': '1.3.0', 'toolchain_name': 'GCCcore', 'name': 'GPI2', 'toolchain_version': '6.4.0'}; to make EasyBuild try to generate a matching easyconfig, use the --try-X options  (at easybu    ild/main.py:104 in find_easyconfigs_by_specs)

I'd consider this bug as critical as it effectively deletes files from the user.

[...]

PR #2504, opened 2018-05-18
don't blindly overwrite an existing easyconfig in tweak_one

boegel:
fix for bug reported by @Flamefire in #2503

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -263,8 +263,11 @@ def tweak_one(src_fn, target_fn, tweaks, targetdir=None):
         _log.debug("Generated file name for tweaked easyconfig file: %s" % target_fn)
 
     # write out tweaked easyconfig file
-    write_file(target_fn, ectxt)
-    _log.info("Tweaked easyconfig file written to %s" % target_fn)
+    if os.path.exists(target_fn):
+        raise EasyBuildError("A file already exists at %s where tweaked easyconfig file would be written", target_fn)
+    else:
+        write_file(target_fn, ectxt)
+        _log.info("Tweaked easyconfig file written to %s", target_fn)
 
     return target_fn
 
--- a/test/framework/tweak.py
+++ b/test/framework/tweak.py
@@ -34,6 +34,8 @@ from unittest import TextTestRunner
 
 from easybuild.framework.easyconfig.parser import EasyConfigParser
 from easybuild.framework.easyconfig.tweak import find_matching_easyconfigs, obtain_ec_for, pick_version, tweak_one
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import remove_file, write_file
 
 
 class TweakTest(EnhancedTestCase):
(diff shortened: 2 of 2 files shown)


Flamefire:
What about the non-symetric handling of `target_dir`? In the case of no filename, `target_dir` is used (set to tmp-path if None) but if a filename is given it is not used.
Maybe the indentation is wrong and `target_dir` should always be used? (At least unless `target_fn`, despite its name` is absolute) This would also solve this.

It also doesn't help that `target_dir` is not documented. So what is its intended behaviour? The current usages seem to be with `target_fn=None` a `target_dir` set and `target_fn` set and `target_dir=None`. So I would guess that `target_dir` was intended to be used always. That makes the most sense (pass filename and path and auto-generate each if not set)

boegel:
@Flamefire The correct implementation is what is intended: if `target_fn` is specified, the generated easyconfig file is created at the location it specifies (the `_fn` is indeed a misnomer, I'll fix that).

This is clearly mentioned in docstring for `tweak_one`:

```
Reads easyconfig file at path <src_fn>, and writes the tweaked easyconfig file to <target_fn>.
```

I'll clarify this by better documenting the arguments of `tweak_one` (especially `targetdir`) and by renaming `src_fn` and `target_fn` to better reflect they specify the (possibly relative) path to the location of the filename to read and the one to generate.

Thanks a lot for pointing this out, I appreciate thoughtful feedback like this a lot!

Flamefire:
Much better. However at line 140 there is already a description of what happens if the filename is None. Maybe leave that paragraph as it was and just add the mentioning of target_dir to the paragraph at 140.

Although it is clear now I don't fully understand why it is written to the current working directory. One might not even have write-access to it (e.g. clusters). Writing it to a temporary directory makes more sense IMO.

boegel:
@Flamefire Thanks for the suggestion, I updated the docstring for `tweak_one` accordingly. I've also enabled using `--force` to let `tweak_one` overwrite an existing file.

W.r.t. writing to the current directory: that's a design choice I guess, and it still makes sense to me... You usually run `eb --try-*` in a directory where you have write permissions, and letting `eb` generate the tweaked easyconfig directly where the user wants it is probably too much since there are different use cases there.

Flamefire:
I find the following easier to understand:

    Reads easyconfig file at path <orig_ec>, and writes the tweaked easyconfig file to <tweaked_ec>.
    
    If <tweaked_ec> is not provided, a target filepath is generated based on <targetdir> and the
    contents of the tweaked easyconfig file.

And:

    :param targetdir: target directory for tweaked easyconfig file, defaults to temporary directory
                              Only used if tweaked_ec is None

Wrt current dir: Ok I kinda understand that in the context of `try-xxx` but for me this happened while just using `eb --software xxx --toolchain yyy`. So the generated EC was just a purely heuristic guess by EB and this kind of command may not be what you run in a writeable directory. EB simply failed to find the EC I intended to use. So here a permission error might confuse even more.

However if this logic is mangled with the `try-xxx`-Creation then its not worth to fix it. Just wanted to provide the context where I encountered this problem.

Issue #2499, opened 2018-05-04
sanity_check does not load the temporary module if there is a system one that exists already when using HMNS


mboisson:
Ran into that problem testing
https://github.com/easybuilders/easybuild-easyblocks/pull/594

You can see in that gist that `/cvmfs/soft.computecanada.ca/easybuild/modules/2017/avx2/MPI/intel2016.4/openmpi2.1/lammps-omp/20170811.lua` 
is loaded instead of loading the temporary one 
`/tmp/eb-oNdpYJ/tmpRugaDe/avx2/MPI/intel2016.4/openmpi2.1/lammps-omp/20170811.lua` 

https://gist.github.com/mboisson/408d3df6335276f87fb8428bc93b386f

mboisson:
Tagging @boegel

boegel:
I need to look into this in more detail, but I think this is a bug in framework...

Have you ever seen this popping up @damianam or @ocaisa?

ocaisa:
What version of Lmod are you using? The temporary path should be the last `module use` statement. Do you have a module that adds modulepaths? A (somehow triggered) unload/load of such a module could put the system one back at the front of the path.

mboisson:
We're using Lmod 7.7.4. 

We do use a hierarchical module schema, so yes, the compiler, MPI and Cuda modules all add to the modulepath. 

Lmod offers a way to deal with module path priority however. 
Something like : 
module use --priority=1000000 /tmp/...

should do it to ensure that this module is the one that is being used.

ocaisa:
This problem doesn't really ring any bells but there quite a few configuration possibilities for the hierarchies that might trigger a reload, the temporary `priority` option to Lmod  seems like a straightforward way to sidestep this issue.

boegel:
After taking a closer look at this, I'm wondering whether this is a bug in Lmod, I'm not sure...

Here's what's going on with a hierarchical module naming scheme, assuming that we're (re)installing `foo/1.2.3` with `foss/2018a` (via `load_fake_module` in `easyblock.py`):

* `eb` generates a temporary module in `/tmp`, say `/tmp/ebtmp/modules/all/foo/1.2.3.lua`
* `eb` does a `ml use /tmp/ebtmp/modules/all` so `foo/1.2.3` can be loaded
* `eb` runs `module load foo/1.2.3` to load that temporary module file
[...]

PR #2506, opened 2018-05-22
prepend location for temporary module file to $MODULEPATH with high priority + mark it as default in load_fake_module method

boegel:
fix for #2499 reported by @mboisson based on suggestion by @ocaisa

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -45,7 +45,7 @@ from vsc.utils import fancylogger
 from vsc.utils.missing import get_subclasses
 
 from easybuild.tools.build_log import EasyBuildError, print_warning
-from easybuild.tools.config import ERROR, IGNORE, PURGE, UNLOAD, UNSET, WARN
+from easybuild.tools.config import ERROR, IGNORE, PURGE, UNLOAD, UNSET
 from easybuild.tools.config import EBROOT_ENV_VAR_ACTIONS, LOADED_MODULES_ACTIONS
 from easybuild.tools.config import build_option, get_modules_tool, install_path
 from easybuild.tools.environment import ORIG_OS_ENVIRON, restore_env, setvar, unset_env_vars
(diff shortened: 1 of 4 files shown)


easybuild/tools/modules.py
```diff
@@ -330,8 +325,16 @@ def set_mod_paths(self, mod_paths=None):
...
-        """Add module path via 'module use'."""
+    def use(self, path, priority=None):
+        """
+        Add path to $MODULEPATH via 'module use'.
+
+        :param path: path to add to $MODULEPATH
+        :param priority: priority for this path in $MODULEPATH (Lmod-specific)
+        """
+        if priority:
+            self.log.info("Ignoring specified priority '%s' when prepending %s to $MODULEPATH", priority, path)
```
wpoely86: I would add something here that it's called in `use`, not a `prepend_path`?

easybuild/tools/modules.py
```diff
@@ -1212,17 +1208,33 @@ def update(self):
...
+        Add path to $MODULEPATH via 'module use'.
+
+        :param path: path to add to $MODULEPATH
+        :param priority: priority for this path in $MODULEPATH (Lmod-specific)
+        """
+        # make sure path exists before we add it
+        mkdir(path, parents=True)
+
+        if priority:
+            self.run_module(['use', '--priority', str(priority), path])
```
wpoely86: I thought it was Lmod specific?

easybuild/tools/modules.py
boegel: fixed

easybuild/tools/modules.py
boegel: It is, this `use` implementation is custom to the `Lmod` class.

easybuild/tools/modules.py
```diff
@@ -330,8 +325,16 @@ def set_mod_paths(self, mod_paths=None):
...
-        """Add module path via 'module use'."""
+    def use(self, path, priority=None):
+        """
+        Add path to $MODULEPATH via 'module use'.
+
+        :param path: path to add to $MODULEPATH
+        :param priority: priority for this path in $MODULEPATH (Lmod-specific)
+        """
+        if priority:
+            self.log.info("Ignoring specified priority '%s' when running 'module use %s'", priority, path)
```
migueldiascosta: while in the code it's clear, I wonder if the log messages should also mention why the priority is ignored, for those not using `Lmod`

PR #2515, opened 2018-05-29
Do not load modules for filtered build dependencies

zao:
When filtering out build dependencies like `pkg-config`, the `short_mod_name` appears to be None instead of a string, which blows the whole thing up.
Is this the proper place to mitigate it, or should the list of dependencies be curated earlier?

PR #2516, opened 2018-05-30
take into account --filter-deps when re-loading build dependencies in extensions_step

boegel:
alternate/proper fix for bug reported in #2515 (incl. a test that triggers the bug)

cc @zao


zao:
LGTM, works on my module.

PR #2521, opened 2018-06-19
make EB logs more syntax highlighting friendly

smoors:
i am trying to make custom syntax highlighting for EB log files to make debugging a bit more eye-friendly.
however, getting syntax highlighting right for strings with multiple nested quotes is quite hard.
removing the outer quotes from `key`, `cmd` and `path` improves things a lot.

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1385,7 +1385,7 @@ def invalidate_module_caches_for(path):
     if not os.path.exists(path):
         raise EasyBuildError("Non-existing path specified to invalidate module caches: %s", path)
 
-    _log.debug("Invallidating module cache entries for path '%s'", path)
+    _log.debug("Invalidating module cache entries for path: %s", path)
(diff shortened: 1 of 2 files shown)


easybuild/tools/modules.py
```diff
@@ -1385,14 +1385,14 @@ def invalidate_module_caches_for(path):
     if not os.path.exists(path):
         raise EasyBuildError("Non-existing path specified to invalidate module caches: %s", path)
 
-    _log.debug("Invallidating module cache entries for path '%s'", path)
+    _log.debug("Invalidating module cache entries for path: %s", path)
```
damianam: Parenthesis are missing here

smoors: parentheses are not needed here, because the path is at the end of the string.

smoors:
@damianam this is not meant for the shell, but for text editors such as `vim`.

damianam:
Oh, got you! I misunderstood what you meant.

boegel:
I don't strongly oppose this change, but to avoid that we re-introduce quotes the logging methods should check for it or something like this?

PR #2525, opened 2018-06-29
add 'bitbucket_account' easyconfig parameter and template and let BITBUCKET* templates use it

SethosII:
This allows the bitbucket project owner to be specified similar to the github templates. Currently the owner is the same as the project name which is not always the case.

This will need changes in the easyconfigs currently using this template and would allow to use the bitbucket templates for more projects.

Two questions:
- Should there be a separate `bitbucket_account` variable? I would prefer one general `account` variable, which should also be used for github templates and this change would also require to change the existing easyconfigs with github templates.
- Should the URL be changed to use https? Bitbucket supports this. Also other template URLs could be changed to use https. I would open a new PR for this.


SethosII:
@boegel It is possible to add a variable as a default in `easybuild/framework/easyconfig/default.py`. This change could also be added to the `github_account`?

I will open a new PR for the https change.

boegel:
@SethosII I'm not sure I understand your first question...

What you did now is OK I think, i.e. use `%(namelower)s` as default value for `bitbucket_account`.

Making the same change for the `GITHUB*` template doesn't make much sense I think, since generally there the project name and owner are different?

In it's current state, this looks good to me, let me know if you're planning on making any further changes or not...

SethosII:
@boegel Yes, you got it right. Some projects on Github are part of an organization with the same name, examples are `samtools`, `vcftools`, `lz4` and others (I created a oneliner for this: `grep -ri source.*github /path/to/easybuild-easyconfigs/ | sed "s/.*github.com//" | sort | uniq | tr '/' ':' | awk -F':' '{ print $2 " " $3 }'`). I think this is useful for these tools and since the `bitbucket_account` and `github_account` variables are similar they should behave identical. For Bitucket it's also not the general case for the owner and name to be identical.

SethosII:
@boegel Yes, from my side this is good to go.

PR #2528, opened 2018-07-09
Use namelower as default for github_account

SethosII:
This will change the default for the variable `github_account` from `None` to `namelower` so that the variable behaves identical to `bitbucket_account` as proposed in https://github.com/easybuilders/easybuild-framework/pull/2525.


boegel:
@SethosII As reported by the tests, this change has an unintended side-effect when easyconfig files are dumped: `%(github_account)s-%(version)s.tar.gz` is now used rather than the expected `SOURCELOWER_TAR_GZ`. Similar issue in `sanity_check_paths`.

I'm not sure how we will be able to avoid this, but we should, since using `%(github_account)s` in an easyconfig file that doesn't involve GitHub source URLs is just wrong...

SethosII:
@boegel I don't get why the `name` variable gets replaced with `github_account` there.

damianam:
It looks to me like EB is trying to map `foo-0.0.1.tar.gz` to a string that is equivalent to `%(some_dictionary_key)s-%(some_other_key)s.%(some_extension)s` and finds that `foo-0.0.1.tar.gz` is equal to `%(github_account)s-%(version)s.tar.gz`, so it settles for it. Then it tries to template the easyconfig, but `SOURCELOWER_TAR_GZ != %(github_account)s-%(version)s.tar.gz`, so it stays untemplated.

Where exactly the first step happens...... I have no idea. I took a look at the parser but I see nothing there. I don't think it happens during the dump either, since there the templates and values are already defined.

boegel:
@SethosII After looking into this, I noticed that the broken test actually reveals a problem that was already there, it's just being exposes now because `%(namelower)s` is used as default value for both the `bitbucket_account` and `github_account` templates...

There was no order being enforced when duplicate template values were involved, so whether `namelower`, `github_account` or `bitbucket_account` were going to be used was a matter of chance, which is obviously not what we want.

By preferring template keys that are shorter, we enforce a strict order, and we also prefer `namelower` over `github_account` and `bitbucket_account` which makes perfect sense...

This is fixed with https://github.com/SethosII/easybuild-framework/pull/1, please take a look at it and merge that PR if it makes sense, to make this PR good to go.
The tests pass with this change included, see https://travis-ci.org/boegel/easybuild-framework/builds/429187026 .

SethosII:
@boegel Thank you for looking into this! I merged your pull request.

boegel:
Going in, thanks @SethosII!

Issue #2543, opened 2018-08-01
Feature request: --hide-all-deps


casparvl:
I'd like to hide ALL dependencies (I'll probably come up with a pull request for a '--hide-all-deps' option), as I only want my users to see the software I have in a build list (all of this software is built one by one with eb [software] -r). 

Providing --hide-deps with an exhaustive list of all possible software that might be in the dependency lists is both unfeasible and ugly. I already implemented a --hide-all-deps in a local copy, it only requires 3 or 4 lines of code... I'll come up with a pull request, this issue is just for 'documenting' the feature request :)

akesandgren:
You are likely going to shoot yourself in the foot by using this.

Take for example TensorFlow and Keras (there are multiple other examples too, I just picked one).
You want to install TensorFlow as a user visible module since users are going to use it directly.
Then installing Keras, a top level user visible module, which depends on TensorFlow, Theano (also user visible) and a couple of other things that you, as per your description above, would have installed as hidden deps. Here you can't use --hide-all-deps.
There are likely more complex cases too, and you would probably end up being able to use it only very infrequently except, perhaps, at the very initial stages of software installation.

I can see the allure of having it, but I strongly think it will be mostly unused even for you.

casparvl:
> Here you can't use --hide-all-deps.
You mean because, when installing Keras, EasyBuild will fail to find that TensorFlow is already present as non-hidden modules? https://github.com/easybuilders/easybuild-framework/issues/2542

I think we may first need to agree on what EasyBuild should do in that case. In my view, it should just pick up on the fact that TensorFlow is already present as a non-hidden module, and simply pick up that non-hidden module as a dependency. In that case, there is no problem installing everything with --hide-all-deps. The only (minor) thing is that if you then build TensorFlow explicitly (i.e. eb Tensorflow-*.eb), EasyBuild will reinstall it and create a non-hidden module, next to the hidden module. A bit of a waste of time to reinstall, but no real issue.

Current behaviour is that EasyBuild will say "A hidden module for Tensorflow is missing, hence I will install Tensorflow, and create a hidden module". It will install TensorFlow in the same software directory it was already present in when it was installed as non-hidden. In the end, you will end up with 1 software installation, with 2 modules (a hidden & a non-hidden) referring to it. Again, no real issue (it will work), but this seems a bit silly.

casparvl:
Maybe I should take the argument one step back: what is the intended usage of hide-deps?

[...]

PR #2544, opened 2018-08-01
2543 hide all deps

casparvl:
Change proposed in this feature request: https://github.com/easybuilders/easybuild-framework/issues/2543

This is actually my first contribution to the framework. I'm not sure how to properly test the framework just from a clone of its repository (I guess I need to 'build' or 'install' the framework in soem way?). I tested these changes by performing them in an actual installation (installed through eb EasyBuild/3.6.0). I then simply performed the exact same changes to my own fork of the framework to create this pull request...


ocaisa:
This is passing the tests because you would need explicit tests for this build option.

It looks like the best template/place to do that would be near https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/easyconfig.py#L1362 on, for example, the test easyconfig https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/easyconfigs/test_ecs/b/bzip2/bzip2-1.0.6-GCC-4.9.2.eb where you can check that the build dependency on `gzip` is returned as a hidden module.

damianam:
As @ocaisa said, tests are missing. But besides that, I think we need to make this PR a bit smarter. This PR will force to rebuild all the dependencies, even if they are already installed (but not hidden). This means risking changes in third party software. Let's say software A has dependencies X and Y already installed with a given option. Then software B has also X and Y as dependencies, but hides all of them. So they are reinstalled potentially with different options, overwriting X and Y, and changing the behaviour of software A for no reason.

ocaisa:
I wonder if there is a role here for the new `.modulerc` features? You could add an additional option like `--check-for-and-expose-hidden-modules`  which would just create the `.modulerc` to expose hidden module(s)

boegel:
@casparvl Any plans for working on this further?

casparvl:
@boegel Hm, not really.

I must admit, it seemed very useful since with older toolchains, I'd sometimes get a myriad of versions for the same software as dependencies. However, it seems we (as a community) are now more consistent in picking (for dependencies) one version per software per toolchain, limiting the usefulness of this 'hide-all-deps' kind of thing. 

Out of curiosity, I guess we are now more selective because this also avoids conflicts higher up in the hierarchy? E.g. problems that occur if there are deps like 
```
lib C -> lib A 1.1
lib B -> lib A 1.0
```
and later, some lib
```
lib D -> lib B + lib C
```
is added. Is the one version per software per toolchain (for dependencies) a policy that you actively pursue now?

boegel:
@casparvl We have a test in place in the easyconfigs repository that ensures we only have one "dependency variant" (basically `version` + `versionsuffix`) per toolchain generation (current is `2018b`). The reason for this is indeed to avoid creating conflicts when suddenly two things come together in the same dependency graph, but also to allow for more easily combining multiple tools that are disjoint w.r.t. dependencies in the same workflow.

There are a couple of exceptions to the rule though, the most obvious one being `Python` where we do allow a single 2.x and 3.x version. See the `check_dep_vars` method in https://github.com/easybuilders/easybuild-easyconfigs/blob/master/test/easyconfigs/easyconfigs.py#L172

PR #2547, opened 2018-08-11
filter out patched files in test/ in fetch_easyconfigs_from_pr

boegel:
Without this fix, `--from-pr` fails on PRs that also change files in `test/`, for example:

```
eb --from-pr 6587 -D
== temporary log file in case of crash /tmp/eb-QVxdiy/easybuild-6UKT1p.log
ERROR: Couldn't find path to patched file /tmp/eb-QVxdiy/files_pr6587/test/easyconfigs/easyconfigs.py
```

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -438,7 +438,7 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
 
     # sanity check: make sure all patched files are downloaded
     ec_files = []
-    for patched_file in patched_files:
+    for patched_file in [f for f in patched_files if not f.startswith('test/')]:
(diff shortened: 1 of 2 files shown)


akesandgren:
There are now two occurrences of ec_files = gh.fetch_easyconfigs_from_pr( in test_fetch_easyconfigs_from_pr

Perhaps it should be rewritten to have a list (hash?) of tuples with PR and list-of-files outside "try" and then a for loop over the PRs. That would make it easier to add a new PR that need special testing in the future.
And it would probably be better to call sorted even on the manually generated list in case of clumsy patching.

akesandgren:
Not exactly what i meant by my comment but easy enough to change if there are more of them in the future.

Issue #1626, opened 2016-02-19
Skip sanity check step


lucamar:
Hi,
I've tried to skip the sanity check step of a python package (GPAW) that uses the easyblock pythonpackage.py, but I could not manage to let EasyBuild skip that step.
I've tried using the following in the easyconfig:
skipsteps = ['install', 'sanitycheck']
It skips the install step, but not the sanity check. Changing the spelling to "sanity_check" did not help either, as it seems that the sanity_check step is treated differently with respect to the configuration, build, etc...
Thanks for any advice.
Luca

boegel:
@lucamar: some steps are not skippable, including the sanity check step...

Why are you trying to skip the sanity check step? Because it's failing? There's probably a good reason for it...

You can try to redefine `sanity_check_paths` if you're sure the installation is what it should be; you may need to redefine other things to make the sanity check happy.

lucamar:
It fails due to the issue with Python provided by Easybuild and MPI on the Cray systems. 
The issue will be solved in a future release, but it would be convenient in general to be able to skip the sanity check as for other steps.
Otherwise one has to manually change/redefine other elements, as you suggest, which is not as straightforward as defining "skipsteps = ['sanitycheck']" in the easyconfig file (provided that one knows that the installation is fine).

gppezzi:
Does this mean that the sanity check in this case assumes that you are on a compute node (or have a working MPI environment)?

How about doing the sanity check through the batch system? would it be hard to implement?

pforai:
@gppezzi Yeah, it basically expects that your system looks like the one EB was designed on, ie wherever you build things on allows you to run whatever you built before, a separation of building and running tests/executing as is standard on very large installations (Cray, BlueGene, etc.) is not supported right now.

One would need to implement the run_cmd() function to make use of GC3Pie and execute the sanity check through a DRM as well a specify maximum time outs (or use sane defaults).

boegel:
@lucamar does you closing the issue mean you no longer require the ability to skip the sanity check?

lucamar:
@boegel I've seen that you had added this issue to v2.x and we have now 3.0.2

boegel:
@lucamar the milestone was later bumped to v3.x, i.e. being able to skip the sanity check hasn't been implemented yet...

lucamar:
@boegel I see: to be honest haven't tried that any longer. If you think that it is feasible to have it as a feature, then I'll reopen the issue

[...]

PR #2549, opened 2018-08-18
allow skipping sanity check

boegel:
This change makes the sanity check step skippable, by including `'sanitycheck'` in `skipsteps` in the easyconfig file, as is requested in #1626.
Additional care was taken to make sure that the sanity check step is *not* skipped when using the `--skip` command line option (as is currently the case already).

I should mention that I very much advise against skipping the sanity check step, and that we will not accept easyconfigs in the central repository (https://github.com/easybuilders/easybuild-easyconfigs) that have `'sanitycheck'` included in `skipsteps`.

I may enhance this PR to spit out a big fat warning when the sanity check is being skipped, to make it clear that this is consider bad practice...

@easybuilders/easybuild-framework-maintainers Please provide some feedback on this change.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2498,8 +2498,11 @@ class EasyBlock(object):
         force = build_option('force') or build_option('rebuild')
         skip = False
 
-        # skip step if specified as individual (skippable) step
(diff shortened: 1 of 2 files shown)


damianam:
lgtm!

PR #2553, opened 2018-08-23
grab maximum amount of comments to fix issue with --merge-pr if there any many comments in a PR

boegel:
This fixes a problem I ran into when merging https://github.com/easybuilders/easybuild-easyconfigs/pull/6677 using `eb --merge-pr`:

```
$ eb --merge-pr 6677

== temporary log file in case of crash /Users/kehoste/work/TMP/eb-OIjkhI/easybuild-VQoHGP.log

easybuilders/easybuild-easyconfigs PR #6677 was submitted by akesandgren, you are using GitHub account 'boegel'

Checking eligibility of easybuilders/easybuild-easyconfigs PR #6677 for merging...
* targets develop branch: OK
* test suite passes: OK
* last test report is successful: FAILED => not eligible for merging!
* last test report is successful: FAILED => not eligible for merging!
* last test report is successful: OK
* approved review: OK (by boegel)
* milestone is set: OK (3.7.0)

WARNING: Review indicates this PR should not be merged (use -f/--force to do so anyway)
```

The problem is that this PR has many comments, exceeded the default number of comments that are returned by the GitHub API.

There's still a problem when a PR has more than 100 comments, but to fix that we'll need to make the fetching of comments page-aware, cfr. https://developer.github.com/v3/#pagination + https://developer.github.com/v3/guides/traversing-with-pagination/

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -942,6 +942,7 @@ def check_pr_eligible_to_merge(pr_data):
                 elif 'FAILED' in comment:
                     res = not_eligible(msg_tmpl % 'FAILED')
                     test_report_found = True
+                    break
                 else:
                     print_warning("Failed to determine outcome of test report for comment:\n%s" % comment)
 
@@ -1004,7 +1005,7 @@ def merge_pr(pr):
 
     # also fetch comments
     comments_url = lambda g: g.repos[pr_target_account][pr_target_repo].issues[pr].comments
-    status, comments_data = github_api_get_request(comments_url, github_user)
+    status, comments_data = github_api_get_request(comments_url, github_user, per_page=GITHUB_MAX_PER_PAGE)
     if status != HTTP_STATUS_OK:
         raise EasyBuildError("Failed to get comments for PR #%d from %s/%s (status: %d %s)",
                              pr, pr_target_account, pr_target_repo, status, comments_data)


damianam:
@boegel I think it is possible to disable pagination altogether: https://github.com/api-platform/docs/blob/master/core/pagination.md

Whether that makes sense or not..... I am not sure. It could potentially be a problem, but I doubt in reality it will happen, as it would need a very large number of comments to be an issue.

test/framework/options.py
```diff
@@ -2822,6 +2822,28 @@ def test_merge_pr(self):
...
 
+        # test --merge-pr on a PR that has a lot of comments
+        # cfr. https://github.com/easybuilders/easybuild-framework/pull/2553
+        args = [
+            '--merge-pr',
+            '6677',
+            '-D',
+            '--github-user=%s' % GITHUB_TEST_ACCOUNT,
+        ]
+        stdout, stderr = self._run_mock_eb(args, do_build=True, raise_error=True, testing=False)
```
damianam: Just a silly comment: Doesn't it make sense to set `do_build=False` and `testing=True`? I took a brief look at the `_run_mock_eb` code (and `eb_main`, and `main`), and I don't think it makes any difference, but still.....

easybuild/tools/github.py
```diff
@@ -1004,7 +1005,7 @@ def merge_pr(pr):
 
     # also fetch comments
     comments_url = lambda g: g.repos[pr_target_account][pr_target_repo].issues[pr].comments
-    status, comments_data = github_api_get_request(comments_url, github_user)
+    status, comments_data = github_api_get_request(comments_url, github_user, per_page=GITHUB_MAX_PER_PAGE)
```
damianam: Add a comment here stating that a PR with lots of comments (`> GITHUB_MAX_PER_PAGE `) might fail?

migueldiascosta:
more generically (since other types of API calls can also be paginated and the defaults can be different), we could traverse pages whenever they are present (according to the Link header)

I had to slightly change `vsc.utils.rest` in order for `head()` to actually return the headers, e.g.

```diff
--- a/lib/vsc/utils/rest.py
+++ b/lib/vsc/utils/rest.py
@@ -174,11 +174,14 @@ class Client(object):
         # TODO: in recent python: Context manager
         conn = self.get_connection(method, url, body, headers)
         status = conn.code
-        body = conn.read()
-        try:
-            pybody = json.loads(body)
-        except ValueError:
-            pybody = body
+        if method == self.HEAD:
+            pybody = conn.headers
+        else:
+            body = conn.read()
+            try:
+                pybody = json.loads(body)
+            except ValueError:
+                pybody = body
         fancylogger.getLogger().debug('reponse len: %s ', len(pybody))
         conn.close()
         return status, pybody
```

but once we have the headers, `githup_api_get_request` could generically handle pagination with something like

```diff
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -243,10 +243,36 @@ def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
     url = request_f(RestClient(GITHUB_API_URL, username=github_user, token=token))
 
     try:
-        status, data = url.get(**kwargs)
+        status, headers = url.head(**kwargs)
     except socket.gaierror as err:
-        _log.warning("Error occurred while performing get request: %s", err)
-        status, data = 0, None
+        _log.warning("Error occurred while getting headers from request: %s", err)
+        status, headers = 0, None
+
+    npages = 1
+    if headers:
+        links = headers.getheader('Link')
+        if links:
+            match = re.search(r"page=(\d+)>; rel=\"last\"", links)
+            if match:
+                npages = int(match.groups()[0])
+
+    if npages > 1:
+        data = []
+        for page in range(1, npages+1):
+            kwargs['page'] = page
+            try:
+                status, page_data = url.get(**kwargs)
+            except socket.gaierror as err:
+                _log.warning("Error occurred while performing get request: %s", err)
+                status, page_data = 0, None
+                break
+            data.extend(page_data)
+    else:
+        try:
+            status, data = url.get(**kwargs)
+        except socket.gaierror as err:
+            _log.warning("Error occurred while performing get request: %s", err)
+            status, data = 0, None
 
     _log.debug("get request result for %s: status: %d, data: %s", url, status, data)
     return (status, data)
```

?

PR #2554, opened 2018-08-29
Check gc3pie version using the `pkg_resources` API.

riccardomurri:
Do not rely on `__version__` attributes, which were removed during the development of the 2.5.0 release.

Also, increase minimal required GC3Pie version to 2.5.0


boegel:
@riccardomurri Does this approach also work with `GC3Pie` 2.4.0?

boegel:
OK, thanks for clarifying!

PR #2555, opened 2018-08-30
added support to 'download' sources from git

mboisson:
This pull requests allows one to specify a dictionary of the following form to "download" sources from a git repository

```
sources = [{
        'filename':'%(namelower)s-%(version)s.tar.gz',
        'git_config': {
                'tag': 'v%(version)s',
                'repo_name': 'speedseq',
                'url': 'https://github.com/hall-lab',
                'recursive': True,
                }
        }]
```


easybuild/framework/easyblock.py
```diff
@@ -663,6 +667,52 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
...
+                    (cmdstdouterr, ec) = run_cmd(cmd, log_all=True, log_ok=False, simple=False, regexp=False)
+                    if commit:
+                        change_dir(os.path.join(targetdir, repo_name))
+                        recursive = " && git submodule update " if recursive else ""
+                        cmd = "git checkout %s %s " % (commit, recursive)
+                        (cmdstdouterr, ec) = run_cmd(cmd, log_all=True, log_ok=False, simple=False, regexp=False)
+                        change_dir(targetdir)
+                    cmd = "tar cfvz %s --exclude-vcs %s && rm -rf %s" % (targetpath, repo_name, repo_name)
+                    (cmdstdouterr, ec) = run_cmd(cmd, log_all=True, log_ok=False, simple=False, regexp=False)
+                    return targetpath
```
boegel: @mboisson Quick review: please move this to a dedicated function, the `obtain_file` method is already too big as is... Throwing that in `filetools` probably makes most sense? And then we also should have a dedicated test for it...

easybuild/framework/easyblock.py
```diff
@@ -559,7 +561,8 @@ def fetch_extension_sources(self, skip_checksums=False):
 
         return exts_sources
 
-    def obtain_file(self, filename, extension=False, urls=None, download_filename=None, force_download=False):
+    def obtain_file(self, filename, extension=False, urls=None, download_filename=None, force_download=False,
+                    git_config={}):
```
boegel: @mboisson Using a mutable value like `{}` as default value is evil (see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

You have a check `if gitconfig`, so just use `None` as default value?

easybuild/framework/easyblock.py
```diff
@@ -663,6 +667,52 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
                 if self.dry_run:
                     self.dry_run_msg("  * %s found at %s", filename, foundfile)
                 return foundfile
+            elif git_config:
+                # if a non-empty dictionary was provided, download from git and archive
+                if not isinstance(git_config, dict):
+                    raise EasyBuildError("Found a non null git_config in 'sources', but value is not a dictionary")
+                else:
+                    git_config = git_config.copy()
```
boegel: Why the `copy()`? I don't see the point of that...

easybuild/framework/easyblock.py
```diff
@@ -663,6 +667,52 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
                 if self.dry_run:
                     self.dry_run_msg("  * %s found at %s", filename, foundfile)
                 return foundfile
+            elif git_config:
+                # if a non-empty dictionary was provided, download from git and archive
+                if not isinstance(git_config, dict):
+                    raise EasyBuildError("Found a non null git_config in 'sources', but value is not a dictionary")
+                else:
```
boegel: No need for the else, the `raise` makes sure you'll never reach the below...

easybuild/framework/easyblock.py
mboisson: I took this strategy from similar code above (line 347).

boegel: Ah, now it clicks... It's done to avoid modifying the value that is passed into `obtain_file` with the `.pop`s...

Better add a comment to explain that (on line 347 too, while you're at it). ;)

easybuild/framework/easyblock.py
mboisson: Just this or the whole logic to download from git ?

boegel: I would create a function that takes `git_config` as an argument, and trim this down to something like:

```python
elif git_config:
    return new_function(git_config)
```

(but find a better name than `new_function`, of course...)

mboisson: Ok. Moving it to filetools would require quite a few more parameters, because this bit of code uses a couple of variables internal to easyblock.py : 
- srcpaths[0]
- self.name
it also calls change_dir, I'll try to see how I can rework those.

boegel: OK, but that's only 2 parameters to add, that's OK.

`change_dir` returns the original dir you were in, you could make sure you restore where you were in the function:

```python
cwd = change_dir(new_dir)
...
change_dir(cwd)

File: easybuild/framework/easyblock.py (hunk 3 above)
Comment (mboisson) [reply easybuild/framework/easyblock.py]: done

File: easybuild/framework/easyblock.py (hunk 2 above)
Comment (mboisson) [reply easybuild/framework/easyblock.py]: done

File: easybuild/framework/easyblock.py (hunk 1 above)
Comment (mboisson) [reply easybuild/framework/easyblock.py]: done

File: easybuild/framework/easyblock.py [hunk 5]
```diff
@@ -568,6 +573,7 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
         :param extension: indicates whether locations for extension sources should also be considered
         :param urls: list of source URLs where this file may be available
         :param download_filename: filename with which the file should be downloaded, and then renamed to <filename>
+        :param git_config: dictionary to parametrize how to download the repository
```
boegel: @mboisson typo: `parameterize`

boegel: @mboisson typo: `parameterize`

easybuild/tools/filetools.py
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
...
+    """
+    Downloads a git repository, at a specific tag or commit, recursively or not, and make an archive with it
+
+    :param filename: name of the archive to save the code to (must be .tar.gz)
+    :param targetdir: target directory where to save the archive to
+    :param git_config: dictionary containing url, repo_name, recursive, and one of tag or commit
+    """
+    # if a non-empty dictionary was provided, download from git and archive
+    if not isinstance(git_config, dict):
+        raise EasyBuildError("Found a non null git_config in 'sources', but value is not a dictionary")
```
boegel: @mboisson There's no link to `sources` here, `get_source_from_git` can also be used in other places in the future, so I would rephrase this to: `"Found unexpected type of value for 'git_config' argument: %s" % type(git_config)`

easybuild/tools/filetools.py
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
...
+        raise EasyBuildError("Found one or more unexpected keys in 'git_config' specification: %s", git_config)
+    if not repo_name:
+        raise EasyBuildError("repo_name not specified in git_config parameter")
+    if not tag and not commit:
+        raise EasyBuildError("Neither tag nor commit found in git_config parameter")
+    if tag and commit:
+        raise EasyBuildError("Tag and commit are mutually exclusive in git_config parameter")
+    if not url:
+        raise EasyBuildError("url not specified in git_config parameter")
+    if '.tar.gz' not in filename:
```
boegel: `if not filename.endswith('.tar.gz'):`

easybuild/tools/filetools.py
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
...
+        raise EasyBuildError("git_config only supports filename ending in .tar.gz")
```
boegel: `currently only supports`

easybuild/tools/filetools.py
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
...
+    # Making a copy to avoid modifying the object with pops
+    git_config = git_config.copy()
+    tag = git_config.pop('tag', None)
+    url = git_config.pop('url', None)
+    repo_name = git_config.pop('repo_name', None)
+    commit = git_config.pop('commit', None)
+    recursive = git_config.pop('recursive', False)
+    if git_config:
+        raise EasyBuildError("Found one or more unexpected keys in 'git_config' specification: %s", git_config)
+    if not repo_name:
```
boegel: add blank line + comment above this line:

```python
# input validation of git_config dict

File: easybuild/tools/filetools.py [hunk 10]
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
...
+    mkdir(targetdir, parents=True)
+    targetpath = os.path.join(targetdir, filename)
+
+    cwd = change_dir(targetdir)
+    recursive = " --recursive " if recursive else ""
+    if tag:
+        cmd = "git clone --branch %s %s %s/%s.git " % (tag, recursive, url, repo_name)
+    else:
+        cmd = "git clone %s %s/%s.git" % (recursive, url, repo_name)
+    (cmdstdouterr, ec) = run.run_cmd(cmd, log_all=True, log_ok=False, simple=False, regexp=False)
```
boegel: @mboisson You're not doing anything with `cmdstdouterr` or `ec` here, so I would let `run_cmd` check the exit code instead?

```python
run.run_cmd(cmd, log_all=True, log_ok=False, regexp=False)
```

easybuild/tools/filetools.py
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
...
+
+    # if a specific commit is asked for, check it out
+    if commit:
+        change_dir(os.path.join(targetdir, repo_name))
+        recursive = " && git submodule update " if recursive else ""
+        cmd = "git checkout %s %s " % (commit, recursive)
+        (cmdstdouterr, ec) = run.run_cmd(cmd, log_all=True, log_ok=False, simple=False, regexp=False)
```
boegel: same here, no point in capturing `cmdstdouterr` and `ec` if they're not used for anything:

```python
run.run_cmd(cmd, log_all=True, log_ok=False, regexp=False)
```

easybuild/tools/filetools.py
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
...
+        change_dir(targetdir)
+
+    # create an archive and delete the git repo
+    cmd = "tar cfvz %s --exclude-vcs %s && rm -rf %s" % (targetpath, repo_name, repo_name)
+    (cmdstdouterr, ec) = run.run_cmd(cmd, log_all=True, log_ok=False, simple=False, regexp=False)
```
boegel: same here :)

```python
run.run_cmd(cmd, log_all=True, log_ok=False, regexp=False)
```

easybuild/tools/filetools.py
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
...
+    # if a specific commit is asked for, check it out
+    if commit:
+        change_dir(os.path.join(targetdir, repo_name))
+        recursive = " && git submodule update " if recursive else ""
+        cmd = "git checkout %s %s " % (commit, recursive)
+        (cmdstdouterr, ec) = run.run_cmd(cmd, log_all=True, log_ok=False, simple=False, regexp=False)
+        change_dir(targetdir)
+
+    # create an archive and delete the git repo
+    cmd = "tar cfvz %s --exclude-vcs %s && rm -rf %s" % (targetpath, repo_name, repo_name)
```
boegel: @mboisson Please use the `remove_file` function rather than `rm -rf`... :)

easybuild/tools/filetools.py
```diff
@@ -1628,6 +1629,65 @@ def copy(paths, target_path, force_in_dry_run=False):
             raise EasyBuildError("Specified path to copy is not an existing file or directory: %s", path)
 
 
+def get_source_from_git(filename, targetdir, git_config):
```
boegel: @mboisson Maybe rename to `get_source_tarball_from_git`?

easybuild/framework/easyblock.py
mboisson: Mmm ? https://en.wiktionary.org/wiki/parametrize

mboisson: duplicate conversation

easybuild/tools/filetools.py
mboisson: fixing in next commit

easybuild/tools/filetools.py
mboisson: fixed in upcoming commit

easybuild/tools/filetools.py
mboisson: fixed in upcoming commit

easybuild/tools/filetools.py
mboisson: fixed in upcoming commit

easybuild/tools/filetools.py
mboisson: fixed in upcoming commit

easybuild/tools/filetools.py
mboisson: fixed in upcoming commit

easybuild/tools/filetools.py
mboisson: remove_file does not remove a directory recursively, unless I am missing something ?

mboisson:
Regarding writing a test for the new function, I am quite puzzled how to test this. I guess I could try to download the EasyBuild repo, you are not using submodules. I probably should not test with a real application repository either.

Ideas about how/where to start ?

easybuild/framework/easyblock.py
mboisson: "fixed" by changing verb ;)

easybuild/tools/filetools.py
boegel: You're right, we need a `remove_dir` as well...

mboisson:
I merged your PR @boegel even though I don't like merging develop, I don't have enough time to figure out this ninja-level git... I'll just close and remove this branch as soon as it's merged into upstream develop and not merge it with our master branch.

boegel:
@mboisson tests fail because of problem in the Travis config which is being fixed in #2584, but also because of using a `git` URL for the testrepository in the tests, which is fixed in https://github.com/ComputeCanada/easybuild-framework/pull/6

boegel:
Going in, right in time to be included in EasyBuild v3.7.0, thanks a lot @mboisson!

boegel:
Going in, thanks @mboisson!

Issue #2557, opened 2018-08-31
Adding dependencies through the hook framework is a bit overly complicated and undocumented


akesandgren:
The implemented hook framework has no easy way to add dependencies (or remove them).
There are two problems:
- adding/removing dependencies must be done very early
- to complete the adding of dependencies one has to call EasyConfig.generate_template_values() and EasyConfig._finalize_dependencies(). Doing that in a hook is a bit awkward.

Adding a dependency_hook in EasyConfig.parse() before the call to self.generate_template_values() would make this much cleaner. One would still have to call self.cfg._parse_dependency() but one could add a helper function in the framework to handle the actual add/remove of a list of dependencies.

The other approach would be to write good documentation on this.

A piece of hook code that works is the following:
```
from distutils.version import LooseVersion
from easybuild.framework.easyconfig.format.format import DEPENDENCY_PARAMETERS
from easybuild.tools.build_log import EasyBuildError

# Changing dependencies can't be done in start_hook since it doesn't get
# any "self" argument. Use pre_fetch_hook instead since it is as early
# as it gets except start_hook
def pre_fetch_hook(self, *args, **kwargs):

    # Internal helper function
    def add_extra_dependencies(self, dep_type, extra_deps):
        """dep_type: must be in DEPENDENCY_PARAMETERS or 'osdependencies'"""
        self.log.info("[pre-fetch hook] Adding %s: %s" % (dep_type, extra_deps))

        self.cfg.enable_templating = False

        if dep_type in DEPENDENCY_PARAMETERS:
            for dep in extra_deps:
                self.cfg[dep_type].append(self.cfg._parse_dependency(dep))
        elif dep_type == 'osdependencies':
                # This currently don't work. Need to figure out how to do this.
                self.cfg[dep_type].append(extra_deps)
        else:
            raise EasyBuildError("pre_fetch_hook: Incorrect dependency type in add_extra_dependencies: %s" % dep_type)

        self.cfg.enable_templating = True
        self.cfg.generate_template_values()
        self.cfg._finalize_dependencies()

    extra_deps = []
    if self.name == 'OpenMPI':
        if LooseVersion(self.version) >= LooseVersion('2.1'):
            pmix_version = '1.2.5'
            if LooseVersion(self.version) >= LooseVersion('3'):
                pmix_version = '2.1.3'

            ucx_version = '1.3.1'

            extra_deps.append(('PMIx', pmix_version))
            # Use of external PMIx requires external libevent
[...]

PR #2562, opened 2018-09-04
add 'parse' hook to add support for tweaking 'raw' easyconfig (REVIEW)

boegel:
see #2557 

cc @akesandgren

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -446,8 +446,12 @@ class EasyConfig(object):
                 self.log.debug("Ignoring unknown easyconfig parameter %s (value: %s)" % (key, local_vars[key]))
 
         # trigger parse hook
+        # templating is disabled when parse_hook is called to allow for easy updating of mutable easyconfig parameters
+        # (see also comment in resolve_template)
         hooks = load_hooks(build_option('hooks'))
+        self.enable_templating = False
         run_hook(PARSE, hooks, args=[self])
+        self.enable_templating = True
 
         # parse dependency specifications
         self.log.info("Parsing dependency specifications...")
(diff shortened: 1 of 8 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -446,8 +446,12 @@ def parse(self):
                 self.log.debug("Ignoring unknown easyconfig parameter %s (value: %s)" % (key, local_vars[key]))
 
         # trigger parse hook
+        # templating is disabled when parse_hook is called to allow for easy updating of mutable easyconfig parameters
+        # (see also comment in resolve_template)
         hooks = load_hooks(build_option('hooks'))
+        self.enable_templating = False
```
akesandgren: old_templating_state = self.enable_templating
and reset to old value after

boegel:
@akesandgren Good to go?

akesandgren:
Yeah i think so, can't find any problems with it.

akesandgren:
@vanzod Any thoughts on this?

boegel:
@ocaisa To answer the issue you raised during today's conf call: any changes made to the `EasyConfig` object in `parse_hook` are (currently) *not* included in the dumped easyconfig file in the installation directory or the easyconfigs archive (i.e. `--repositorypath`).

So, unless you make sure you have the hooks active every time, picking up a dumped easyconfig file later may result in surprises...

We can look into fixing that, but maybe some people wouldn't expect things to trickle down, I'm not quite sure (I can see problems with both approaches), and I'm not sure how easy it would be either to make things trickle down (the classic "there be dragons here" probably applies...).

akesandgren:
I for one would expect it NOT to end up in the dumped one, but it might be good to have a comment added that hooks where changing stuff.

But if someone wants the changes to end up in the dump, this should probably be configurable.

ocaisa:
@boegel I wouldn't have a problem with the archived easyconfig not including the changes but I think I would expect to see them in the `reprod` directory

akesandgren:
Yes, that would be good, didn't think of that one.

boegel:
follow-up on what @ocaisa mentioned in #2565

PR #2566, opened 2018-09-06
templating must be disabled when parsing dependencies in EasyConfig.parse

boegel:
fix for problem introduced in #2562 (cc @akesandgren)

Before the changes in #2562, it was OK that templating was enabled when `_parse_dependency` was called, because the templating only occurs when easyconfig parameters are fetched via `self` (using `self[key`), i.e. the `EasyConfig` instance. When the easyconfig parameters are fetching via `local_vars[key]` there is no templating since `local_vars` is a regular `dict` value...

also fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/6818


akesandgren:
Ah, should have thought of that before...

akesandgren:
Going in, thanks @boegel

PR #2571, opened 2018-09-07
lift invalidating of module caches into helper method that can be used by easyblocks

boegel:
While testing the `ModulesAlias` easyblock (cfr. https://github.com/easybuilders/easybuild-easyblocks/pull/1503), I noticed that the caching we do for `module show` commands is a bit too aggressive...

Module aliases show up in the output of `module --terse avail` in such a way that a fallback is needed to `module show` to recognize that a module alias is already in place.

The result of a `module show` command is cached the first time if it is run, but if multiple builds are being performed in the same session, this is a problem since the later build will see the outdated/incorrect result in a module alias was install in an early build.

For example, if `Bazel` has `Java/1.8` specified as a dependency to be resolved by a module alias that still needs to be installed, the `Java/1.8` alias that is in place by the time `Bazel` is installed will not be observed because the original `module show Java/1.8` that came back empty is cached.

Clearing the `module avail` and `module show` caches before every new build is started deals with this issue.

### DIFF
--- a/test/framework/easyblock.py
+++ b/test/framework/easyblock.py
@@ -45,6 +45,7 @@ from easybuild.tools import config
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import get_module_syntax
 from easybuild.tools.filetools import copy_dir, copy_file, mkdir, read_file, remove_file, write_file
+from easybuild.tools.module_generator import module_generator
 from easybuild.tools.version import get_git_revision, this_is_easybuild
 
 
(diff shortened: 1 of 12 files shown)


boegel:
After trying to come up with a test to verify the fix done in 3354ade, I noticed there wasn't actually a problem at all in framework...

The actual bug was that the `ModuleAlias` easyblock in https://github.com/easybuilders/easybuild-easyblocks/pull/1503 was overwriting `make_module_step` but not calling `invalidate_module_caches_for` to ensure that the module caches were reset where needed.

So, I've rolled back the call to `reset_module_caches` between builds, and just lifted the calls to `invalidate_module_caches_for` into a helper method that can be leveraged by the `ModuleAlias` easyblock (see https://github.com/easybuilders/easybuild-easyblocks/pull/1503/commits/f30e46c3b0dfb8f27b90d1f24c0f02d787fb11c3).

I've added the test I was working on as well, since it doesn't hurt; as far as I can tell there was no test yet that accurately verifies that the module caches are reset in between builds...

test/framework/easyblock.py
```diff
@@ -1411,6 +1412,51 @@ def test_this_is_easybuild(self):
...
+        self.assertFalse(self.modtool.exist(['one/1.0'])[0])
+
+        # add .modulerc to install version alias one/1.0 for one/1.0.2
+        # this makes cached result for "show one/1.0" incorrect as soon as one/1.0.2 is installed via one.eb
+        modgen = module_generator(None)
+        module_version_spec = {'modname': 'one/1.0.2', 'sym_version': '1.0', 'version': '1.0.2'}
+        modulerc_txt = modgen.modulerc(module_version=module_version_spec)
+        one_moddir = os.path.join(self.test_installpath, 'modules', 'all', 'one')
+        write_file(os.path.join(one_moddir, '.modulerc'), modulerc_txt)
+
```
ocaisa: What happens if you repeat 1446 at this point (1.0.2 doesn't exist yet but the rc file does)?

boegel: Well, it won't make a difference because of the modules cache (the result for `avail one/1.0` & `show one/1.0` is cached).
And even if we clear the cache, it will still fail, since the underlying module doesn't exist yet (I'll add that to the test).

PR #2573, opened 2018-09-09
use .counts() rather than deprecated .stats() for GC3Pie

boegel:
@riccardomurri Please review?

### DIFF
--- a/easybuild/tools/job/gc3pie.py
+++ b/easybuild/tools/job/gc3pie.py
@@ -220,9 +220,6 @@ class GC3Pie(JobBackend):
         # some sites may not be happy with flooding the cluster with build jobs...
         self._engine.max_in_flight = build_option('job_max_jobs')
 
-        # prepare for .counts() in _print_status_report
-        self._engine.init_counts_for(Application)
-
         # Add your application to the engine. This will NOT submit
         # your application yet, but will make the engine *aware* of
         # the application.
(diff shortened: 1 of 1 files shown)


easybuild/tools/job/gc3pie.py
```diff
@@ -220,14 +220,8 @@ def complete(self):
...
-        # `Engine.stats()` (which is used later on in `_print_status_report()`)
-        # changed between 2.4.2 and 2.5.0.dev -- make sure we stay compatible
-        # with both
-        try:
-            self._engine.init_stats_for(Application)
-        except AttributeError:
-            _log.debug("No `init_stats_for` method in the Engine class;"
-                       " assuming pre-2.5.0 GC3Pie and ignoring error.")
+        # prepare for .counts() in _print_status_report
+        self._engine.init_counts_for(Application)
```
riccardomurri: This is actually no longer required: counts for `Application` are init'ed by default in the `Engine` class. (This changed during the development of 2.5)

riccardomurri: (Does no harm to keep -- it's just redundant now.)

boegel: Probably better to just drop it then?

riccardomurri: If the minimally required version of GC3Pie is 2.5.0, then yes.

boegel: That's indeed the case since #2554, so I'll drop it. Thanks for the feedback!

PR #2574, opened 2018-09-11
Dump a fully parsed easyconfig to the reproducibility dir in all cases 

### LINKED ISSUE #2565: Dumping parsed easyconfig files
With #2562 we introduce the capability of making modifications to the easyconfig file that is read from disk in a (potentially) non-reproducible way.

At https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/framework/easyblock.py#L2869 there is already a feature that dumps the final parsed easyconfig to the installation directory (in a `reprod` subdir).  This was originally introduced for minimal toolchains but since (some level of) subtoolchain resolution is now the default, and because of the new hook, this should be done more widely.

ocaisa:
Fixes #2565

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2871,9 +2871,9 @@ def build_and_install_one(ecdict, init_env):
             reprod_spec = os.path.join(new_log_dir, 'reprod', ec_filename)
             try:
                 app.cfg.dump(reprod_spec)
-                _log.debug("Dumped fully parsed easyconfig to %s", reprod_spec)
-            except NotImplementedError, err:
-                _log.warn("Unable to dumped fully parsed easyconfig to %s: %s", reprod_spec, err)
+                _log.info("Dumped fully processed easyconfig to %s", reprod_spec)
+            except NotImplementedError as err:
+                _log.warn("Unable to dumped fully processed easyconfig to %s: %s", reprod_spec, err)
 
             try:
                 # upload easyconfig (and patch files) to central repository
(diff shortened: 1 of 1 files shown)


ocaisa:
Hmm, not so simple, this is kicking a ton of errors with the tests.

boegel:
@ocaisa The only 3 failing tests are the ones that involve `.yeb` easyconfigs, so it's more an issue with those tests imho...

ocaisa:
@akesandgren Can you review this, it's a minor change

easybuild/framework/easyblock.py
```diff
@@ -2866,13 +2866,14 @@ def build_and_install_one(ecdict, init_env):
...
-                # --minimal-toolchains (and --use-existing-modules) is used
-                # TODO --try-toolchain needs to be fixed so this doesn't play havoc with it's usability
-                reprod_spec = os.path.join(new_log_dir, 'reprod', ec_filename)
+            # for reproducability we dump out the fully parsed easyconfig since the contents can be affected by
+            # subtoolchain resolution (and related options) and/or hooks
+            reprod_spec = os.path.join(new_log_dir, 'reprod', ec_filename)
+            try:
                 app.cfg.dump(reprod_spec)
-                _log.debug("Dumped easyconfig tweaked via --minimal-toolchains to %s", reprod_spec)
+                _log.debug("Dumped fully parsed easyconfig to %s", reprod_spec)
```
boegel: @ocaisa I'd promote this to `_log.info`?

I also wonder if we can come up with a better term than "fully parsed easyconfig", that doesn't mean much to me to be honest... How about "fully processed easyconfig"?

easybuild/framework/easyblock.py
```diff
@@ -2866,13 +2866,14 @@ def build_and_install_one(ecdict, init_env):
...
+            except NotImplementedError, err:
```
boegel: please use `as err` rather than `, err` (the latter is no longer supported in Python 3 iirc)

PR #2575, opened 2018-09-11
add modulerc method to ModuleGenerator class 

boegel:


easybuild/tools/module_generator.py
```diff
@@ -197,6 +198,51 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
+                        # determine module name with symbolic version
+                        if version in modname:
+                            # take a copy so we don't modify original value
+                            module_version = copy.copy(module_version)
+                            module_version['sym_modname'] = modname.replace(version, sym_version)
+                        else:
+                            raise EasyBuildError("Version '%s' does not appear in module name '%s'", version, modname)
+
+                        module_version_statement = '\n'.join([
+                            'if {"%(sym_modname)s" eq [module-info version %(sym_modname)s]} {',
```
ocaisa: I know the explanation is referenced above  in the URL but this behaviour is very non-intuitive, a comment could really help to understand this in 6 months: 
```
If you pass a string version to "module-info version" for a version that doesn't exist you get the same string back as
output. However after you've used module-version to define that as a "valid" version you get back the real modulefile name
```

easybuild/tools/modules.py
```diff
@@ -1106,7 +1107,7 @@ class Lmod(ModulesTool):
     """Interface to Lmod."""
     COMMAND = 'lmod'
     COMMAND_ENVIRONMENT = 'LMOD_CMD'
-    REQ_VERSION = '5.8'
+    REQ_VERSION = '6.6.3'
```
boegel: note: Lmod 6.0 or newer is needed to pick up on `.modulerc` files in module file subdirectory.

Older Lmod 6.x versions have bugs that make it incompatible with EasyBuild in other (unrelated ways), hence the more recent 6.x being required.

Lmod 6.6.3 is the latest 6.x release, which was released Nov 13th 2016 (close to 2 years old), so I don't expect this to be a big problem.

Also, this way EasyBuild is still compatible with both Lmod 6.x & 7.x...

PR #2576, opened 2018-09-13
Correct error statements in module checks

ocaisa:
Without this you get an error like:
```
ERROR: EasyBuild requires vLmod >= v6.6, found v6.6.3
```


boegel:
Hah, that must have confused you quite a bit at first... ;)

PR #2585, opened 2018-09-21
fix SUBTOOLCHAINS spec for intel-para toolchain

ocaisa:
slipped through the cracks in #2466


bartoldeman:
ah yes, what was I thinking then, thanks for spotting it.

Issue #2581, opened 2018-09-17
bootstrap process breaks with python setuptool > 39.0.1-2 on Ubuntu 18.04


chossimbaone:
Bootstrap installer process breaks on Ubuntu 18.04.
Problem is that python-setuptool has removed easy_install from version [39.0.1-2](https://launchpad.net/ubuntu/+source/python-setuptools/39.0.1-2), which is need for stage 1 of the bootstrap process.

Workaround for now, was to install version 39.0.0
`sudo pip install setuptools==39.0.0`

boegel:
Thanks for reporting this @chossimbaone.

It shouldn't be too hard to get this fixed in the short term, since EasyBuild doesn't actually use the `easy_install` command directly (it talks to `setuptools.command.easy_install` instead).

So, as long as the Python module still works, it's mostly a matter of stop checking for the `easy_install` command in the bootstrap script.

@chossimbaone @mcakircali: Can you check what these commands spit out on Ubuntu 18.04?

```
$ python -m easy_install --version
$ python -c "from setuptools.command import easy_install; print easy_install.__file__"
```

In the longer term, we should make the bootstrap script try to use `pip` first, and only fall back to `setuptools` for installing stage 1.

I'll take a stab at fixing this soon...

mcakircali:
Here the outputs:
$ python -m easy_install --version
setuptools 39.0.1 from /usr/lib/python2.7/dist-packages (Python 2.7)
$ python -c "from setuptools.command import easy_install; print easy_install.__file__"
/usr/lib/python2.7/dist-packages/setuptools/command/easy_install.pyc

boegel:
Problem should be fixed with https://github.com/easybuilders/easybuild-framework/pull/2590, bootstrapping on Ubuntu 18.04 should no longer fail because of a missing `easy_install` command in the latest version of the bootstrap script.

PR #2590, opened 2018-09-25
stop relying on 'easy_install' in bootstrap script, use 'python -m easy_install' instead

boegel:
This should fix the problem reported in https://github.com/easybuilders/easybuild-framework/issues/2581

@chossimbaone @mcakircali: Can you test this and let us know whether this fixes the reported problem on Ubuntu 18.04?

You can download the updated bootstrap script (version `20180925.01`) using the following command while this PR is open:

```
curl -OL https://raw.githubusercontent.com/boegel/easybuild-framework/fix_bootstrap_ubuntu1804/easybuild/scripts/bootstrap_eb.py
```


mcakircali:
with 3.6.2, it works as expected..

boegel:
Thanks for the feedback @mcakircali!

@akesandgren Were you planning to test this too? Should be good to go...

PR #2591, opened 2018-09-25
fix templating of values in list_software function

boegel:
fix for problem reported in https://github.com/easybuilders/easybuild/issues/446


boegel:
There's a test for `list_software` already, but we didn't have any templates in the easyconfigs used for the test. That is the case now, and the `test_list_software` fails without the fix with the modified test easyconfigs.

vanzod:
Tested and ok

PR #2595, opened 2018-09-27
allow setting optarch compiler flags in the easyconfig

migueldiascosta:
If `toolchainopts[optarch]` from the easyconfig is a string, use that as architecture optimization flags instead of `build_options('optarch')`

One problem with this approach is that the per-compiler syntax is not supported, unlike `build_option('optarch')` (which has been parsed earlier, in `tools.options._postprocess_optarch`)

Using only compiler-specific flags in the easyconfig optarch would not usually be an issue, since the compiler is already set by the easyconfig and this is meant for very specific cases, but I suppose it will break `--try-toolchain`...

Should we also postprocess optarch here? (or is there a better way?)


boegel:
@migueldiascosta That's a good point, but on the other hand it would be weird to specify flags specific to a compiler different from the toolchain compiler in the easyconfig file...

It's nice to support this for flexibility, but this should really only be used by exception I think, of for site customization reasons.

And `--try-toolchain` is named `'try'` on purpose, there's lots of other things that can go wrong there already (e.g. compiler-specific patches or `configopts`, etc.).

@damianam Thoughts on this?
Maybe also @bartoldeman who frequently uses `--try-toolchain` I think...

damianam:
IMO if this doesn't break anything it can go in. I agree that supporting the per-compiler syntax is weird in easyconfigs. It will break `--try-toolchain`, but so do other things (per compiler patches, environment variables that depend on the particular MPI used, etc)

easybuild/tools/toolchain/compiler.py
```diff
@@ -292,7 +292,17 @@ def _set_optimal_architecture(self, default_optarch=None):
         :param default_optarch: default value to use for optarch, rather than using default value based on architecture
                                 (--optarch and --optarch=GENERIC still override this value)
         """
-        optarch = build_option('optarch') 
+        ec_optarch = self.options.get('optarch', False)
+        if isinstance(ec_optarch, basestring):
+            if ':' in ec_optarch:
```
damianam: Use `OPTARCH_MAP_CHAR`

easybuild/tools/toolchain/compiler.py
```diff
@@ -292,7 +292,17 @@ def _set_optimal_architecture(self, default_optarch=None):
...
+                error_msg = "When setting optarch in the easyconfig (found %s), " % ec_optarch
+                error_msg += "the <compiler:flags> syntax is not allowed. "
```
damianam: Use `OPTARCH_MAP_CHAR` here as well

boegel:
I'd love to see a test for this before merging it in... Up for that @migueldiascosta? See `test/framework/toolchain.py`

migueldiascosta:
but I probably should have used a different toolchain for the new test easyconfigs...

boegel:
Thanks @migueldiascosta!

PR #2597, opened 2018-09-28
generate .modulerc.lua when Lua syntax and Lmod >= 7.8 is used

boegel:
Strictly speaking, there's nothing wrong with sticking to a `.modulerc` in Tcl syntax, but it has the indirect implication that `tcslh` must be available , causing problems like (reported by @vanzod):

```
$ module use /tmp/eb-93S43S/tmpR_HaVA/all/Core
sh: 1: /usr/bin/tclsh: not found
Lmod has detected the following error:  Unable to parse: "/tmp/eb-93S43S/tmpR_HaVA/all/Core/Java/.modulerc". Aborting!
```

These changes spit out the `.modulerc` in Lua syntax when Lua is used as module syntax *and* if the Lmod version is recent enough (7.8 or newer).

The `ModuleRC` easyblock that was added in https://github.com/easybuilders/easybuild-easyblocks/pull/1503 will need to made aware of this too, since when Lua syntax is used for the `.modulerc` file the filename should be `.modulerc.lua`...


vanzod:
I have been testing this together with https://github.com/easybuilders/easybuild-easyblocks/pull/1533 and they definitely solve the `tclsh` problem by creating a `.modulerc.lua`. However I am still having troubles installing `Java/1.8`, which returns a module not found error. I cannot understand if it is in any way related.
[easybuild-Java-1.8-20180928.172250.uEVRe.log](https://github.com/easybuilders/easybuild-framework/files/2429941/easybuild-Java-1.8-20180928.172250.uEVRe.log)

zao:
@vanzod I'm experiencing a similar Java problem with 3.7.0 on 6.6 series of Lmod, but not with the current 7 version - easybuilders/easybuild#462

boegel:
@vanzod You're hitting the same problem as @guest1604 is reporting in https://github.com/easybuilders/easybuild-easyblocks/issues/1534, based on this line from the log:

```
== 2018-09-28 17:22:50,885 modulerc.py:78 INFO Adding module version alias for Core/Java/1.8.0_181 to /tmp/eb-Yr5Ae5/tmpmDy216/all/Core/Java/.modulerc.lua
```

migueldiascosta:
@vanzod can you check again now that https://github.com/easybuilders/easybuild-easyblocks/pull/1535 was merged?

vanzod:
@migueldiascosta Test passed. Now it works like a charm.

PR #2599, opened 2018-09-29
Add ability to upgrade dependencies based on available easyconfigs

ocaisa:
An example of how this can be used:
```
eb --dry-run --robot --try-toolchain=foss,2018b --minimal-toolchains --try-update-deps Keras-2.1.3-intel-2017b-Python-3.6.3.eb
```
It currently does the less invasive update of the dependencies, i.e., if we upgrade software with version `1.2.3`, it would chose `1.2.4` over `1.3.0` (if both existed in the target toolchain).


ocaisa:
@boegel This is ready for a look and initial comments.

I already see a case where it won't work as intended, `Java/1.8` with the new approach there, since it relies on `LooseVersion` to do the comparison. We may need to make this expand the logic there to deal with this.

ocaisa:
This works but is quite aggressive

easybuild/framework/easyconfig/tweak.py
```diff
@@ -833,6 +839,54 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                                                 candidate_ver, version_suffix] if x])
+                    cand_paths = search_easyconfigs(dep['name'] + '-' + ecver, return_robot_list=True)
+                    # Add them to the possibilities
+                    for path in cand_paths:
+                        # Get the version from the path
+                        filename = os.path.basename(path)
+                        # Find the version sandwiched between our known values
+                        try:
+                            version_prefix = ''.join([x for x in [dep['name'] + '-',
+                                                                  parsed_ec.get('versionprefix', '')] if x])
```
ocaisa: Whoops this is incorrect, it should be the dep `versionprefix`

easybuild/framework/easyconfig/tweak.py
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                if 'versionsuffix' in dep:
+                    if dep['versionsuffix'] in versonsuffix_mapping:
+                        versionsuffix = versonsuffix_mapping[dep['versionsuffix']]
+                    else:
+                        versionsuffix = dep['versionsuffix']
+                else:
+                    versionsuffix = ''
+
+                for toolchain in toolchain_hierarchy:
+                    candidate_ver = '.*'  # using regex for *
```
ocaisa: Be careful about Python 2 -> 3 here...and maybe handle Java (with modulerc) here too

easybuild/tools/robot.py
```diff
@@ -379,13 +379,24 @@ def resolve_dependencies(easyconfigs, modtool, retain_all_deps=False):
     return ordered_ecs
 
 
-def search_easyconfigs(query, short=False, filename_only=False, terse=False):
-    """Search for easyconfigs, if a query is provided."""
+def search_easyconfigs(query, short=False, filename_only=False, terse=False, return_robot_list=False):
```
boegel: @ocaisa Wouldn't two separate arguments make more sense, for better control?

* `consider_extra_paths=True` to control whether or not to include the paths listed in `--search-paths`
* `return_list=False` to control whether or not a list is returned rather than being printed
   (or maybe `print_result=True` is better, and always return the list regardless of whether or not printing is being done)

easybuild/framework/easyconfig/tweak.py
```diff
@@ -88,17 +90,21 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
...
     source_toolchain = easyconfigs[-1]['ec']['toolchain']
-    modifying_toolchains = False
+    modifying_toolchains_or_deps = False
     target_toolchain = {}
     src_to_dst_tc_mapping = {}
     revert_to_regex = False
+    if 'upgrade_deps' in build_specs:
+        upgrade_dependencies = build_specs['upgrade_deps']
+    else:
+        upgrade_dependencies = False
```
boegel: This can be a one-liner, and `None` is a better default value?

```python
upgrade_dependencies = build_specs.get('upgrade_deps', None)
```

easybuild/framework/easyconfig/tweak.py
```diff
@@ -88,17 +90,21 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
...
 
-    if 'toolchain_name' in build_specs or 'toolchain_version' in build_specs:
+    if 'toolchain_name' in build_specs or 'toolchain_version' in build_specs or upgrade_dependencies:
         keys = build_specs.keys()
 
         # Make sure there are no more build_specs, as combining --try-toolchain* with other options is currently not
         # supported
-        if any(key not in ['toolchain_name', 'toolchain_version', 'toolchain'] for key in keys):
+        if any(key not in ['toolchain_name', 'toolchain_version', 'toolchain', 'upgrade_deps'] for key in keys):
             print_warning("Combining --try-toolchain* with other build options is not fully supported: using regex")
```
boegel: @ocaisa Warning message should be update to mention `--try-toolchain* or --try-upgrade-deps`?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -161,10 +167,11 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
         new_ec_file = None
         verification_build_specs = copy.copy(build_specs)
         if orig_ec['spec'] in listed_ec_paths:
-            if modifying_toolchains:
+            if modifying_toolchains_or_deps:
                 if tc_name in src_to_dst_tc_mapping:
                     new_ec_file = map_easyconfig_to_target_tc_hierarchy(orig_ec['spec'], src_to_dst_tc_mapping,
-                                                                        tweaked_ecs_path)
+                                                                        tweaked_ecs_path,
+                                                                        update_dep_versions=upgrade_dependencies)
```
boegel: keep things consistent: `upgrade` or `update`, pick one

easybuild/framework/easyconfig/tweak.py
```diff
@@ -807,6 +914,16 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
     """
     # Fully parse the original easyconfig
     parsed_ec = process_easyconfig(ec_spec, validate=False)[0]
+
+    # There are some common versionsuffixes (like '-Python-(%pyver)s') that also need dynamic searching/updating
+    versonsuffix_mapping = map_common_versionsuffixes('Python', '2', parsed_ec['ec']['toolchain'], toolchain_mapping)
+    versonsuffix_mapping.update(map_common_versionsuffixes('Python', '3', parsed_ec['ec']['toolchain'],
```
boegel: Hmm, can we avoid hardcoding versions here somehow? What if there's a Python 4.x at some point?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
                 # set module names
                 orig_dep['short_mod_name'] = ActiveMNS().det_short_module_name(dep)
                 orig_dep['full_mod_name'] = ActiveMNS().det_full_module_name(dep)
+            elif update_dep_versions:
+                # Search for available updates for this dependency:
+                # First get all candidate paths for this (include search through subtoolchains)
+                toolchain_hierarchy = get_toolchain_hierarchy(orig_dep['toolchain'])
+                potential_version_matches = []
```
boegel: @ocaisa This big chunk of code should go into a dedicated function imho (which can help a lot with writing tests for this stuff)

test/framework/tweak.py
```diff
@@ -276,12 +276,36 @@ def test_map_toolchain_hierarchies(self):
...
+        })
+        get_toolchain_hierarchy.clear()
+        gcc_binutils_tc = {'name': 'GCC', 'version': '4.9.3-2.26'}
+        iccifort_binutils_tc = {'name': 'iccifort', 'version': '2016.1.150-GCC-4.9.3-2.25'}
+        toolchain_mapping = map_toolchain_hierarchies(iccifort_binutils_tc, gcc_binutils_tc, self.modtool)
+        possible_mappings = map_common_versionsuffixes('binutils', '2', iccifort_binutils_tc, toolchain_mapping)
+        expected_mappings = {'-binutils-2.25': '-binutils-2.26'}
+        self.assertEqual(possible_mappings, expected_mappings)
+
+        # Make sure we only map upwards, here it's gzip 1.4 in gcc and 1.6 in iccifort
```
boegel: if this is only mapping upwards, the function should be renamed to `update`?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -795,7 +802,107 @@ def map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool):
...
+
+        # fetch from cache if available, cache it if it's not
+        if cache_key in cache:
+            _log.debug("Using cache to return version suffix mapping for toolchain %s: %s", str(cache_key),
+                       cache[cache_key])
+            return cache[cache_key]
+        else:
+            versionsuffix_mappings = func(software_name, software_version_stub, original_toolchain, toolchain_mapping)
+            cache[cache_key] = versionsuffix_mappings
+            return cache[cache_key]
```
boegel: do the return out of the `if/else`, since it's common

easybuild/framework/easyconfig/tweak.py
```diff
@@ -795,7 +802,107 @@ def map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool):
...
+    # Expose clear method of cache to wrapped function
+    cache_aware_func.clear = cache.clear
+
+    return cache_aware_func
+
+
+@map_versionsuffixes_cache
+def map_common_versionsuffixes(software_name, software_version_stub, original_toolchain, toolchain_mapping):
+    """
+    Create a mapping of common versionssuffixes (like `-Python-%(pyvers)`) between toolchains
```
boegel: If the `versionsuffix` is well written, there should be no need for this actually... If `%(pyver)s` is used, that should just follow the version of the `Python` in `dependencies`. What am I missing here?

How about just failing hard if there's a hardcoded version found in `versionsuffix`?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -795,7 +802,107 @@ def map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool):
...
+            cand_paths = search_easyconfigs(regex_search_query, return_robot_list=True)
+            for path in cand_paths:
+                # Get the version from the path
+                filename = os.path.basename(path)
+                # Find the version sandwiched between our known values
+                try:
+                    regex = '%s(.*)%s' % (versionprefix_name, toolchain_suffix)
+                    version = re.search(regex, filename).group(1)
+                    if LooseVersion(version) < LooseVersion(target_version):
+                        source_versions.add(version)
```
boegel: A *very* large part of this is duplicate code, this should definitely go into a (sub-) function rather than copy-pasting it...

easybuild/framework/easyconfig/tweak.py
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                # Search for available updates for this dependency:
+                # First get all candidate paths for this (include search through subtoolchains)
+                toolchain_hierarchy = get_toolchain_hierarchy(orig_dep['toolchain'])
+                potential_version_matches = []
+
+                # Figure out what precedes the version
+                if 'versionprefix' in dep:
+                    versionprefix = dep['versionprefix']
+                else:
+                    versionprefix = ''
```
boegel: This can be collapsed to:

```python
versionprefix = dep.get('versionprefix', '')

File: easybuild/framework/easyconfig/tweak.py [hunk 14]
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                versionprefix = ''.join([x for x in [dep['name'] + '-', versionprefix] if x])
```
boegel: Hmm, shouldn't be called `versionprefix` anymore if it includes `name`?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                # Figure out the main versionsuffix (altered depending on toolchain in the loop below)
+                if 'versionsuffix' in dep:
+                    if dep['versionsuffix'] in versonsuffix_mapping:
+                        versionsuffix = versonsuffix_mapping[dep['versionsuffix']]
+                    else:
+                        versionsuffix = dep['versionsuffix']
+                else:
+                    versionsuffix = ''
```
boegel: This whole block can be collapsed to a single line as well:

```python
versionsuffix = versonsuffix_mapping.get(dep.get('versionsuffix', None), '')
```

If there is no `versionsuffix`, `dep.get('versionsuffix', None)` will return `None`, which won't be included in `versonsuffix_mapping`, so you'll get the fallback value `''`.

You can even just use `dep.get('versionsuffix')` rather than `dep.get('versionsuffix', None)`, since `None` is the default for `.get()` on a dict value:

```python
>>> print {}.get('foo')
None

File: easybuild/framework/easyconfig/tweak.py (hunk 2 above)
Comment (boegel) [reply easybuild/framework/easyconfig/tweak.py]: Should this be a comment in the code instead? Should also be clarified a bit imho...

File: easybuild/framework/easyconfig/tweak.py [hunk 16]
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+
+                for toolchain in toolchain_hierarchy:
+                    candidate_ver = '.*'  # using regex for *
+
+                    # determine main install version based on toolchain
+                    if toolchain['name'] != DUMMY_TOOLCHAIN_NAME:
+                        toolchain_suffix = "-%s-%s" % (toolchain['name'], toolchain['version'])
+                    else:
+                        toolchain_suffix = ''
+                    full_versionsuffix = ''.join([x for x in [toolchain_suffix, versionsuffix, EB_FORMAT_EXTENSION]
```
boegel: Hmm, is `EB_FORMAT_EXTENSION` really needed? If so, it should be handled somewhere else, I think...

easybuild/framework/easyconfig/tweak.py
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                    for path in cand_paths:
+                        # Get the version from the path
+                        filename = os.path.basename(path)
+                        # Find the version sandwiched between our known values
+                        try:
+                            regex = '^%s(.+?)%s' % (versionprefix, full_versionsuffix)
+                            version = re.search(regex, filename).group(1)
+                        except AttributeError:
+                            raise EasyBuildError("Somethings wrong, could not extract version from %s using %s",
+                                                 filename, regex)
```
boegel: This shouldn't be a `try`/`except`, you should check the result or `re.search`:

```python
regex = re.compile('^%s(.+?)%s' % (versionprefix, full_versionsuffix))
res = regex.search(filename)
if res:
    version = res.group(1)
else:
    raise EasyBuildError("Failed to determine version from '%s' using regex pattern '%s'", filename, regex.pattern)

File: easybuild/framework/easyconfig/tweak.py [hunk 18]
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                        potential_version_matches.append({'version': version, 'path': path, 'toolchain': toolchain})
+                _log.debug("Found possible dependency upgrades: %s\n", potential_version_matches)
```
boegel: @ocaisa No need for the `\n'`?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+
+                # Compare this version to the original versions and replace if appropriate (upgrades only)
+                highest_version = dep['version']
+                for candidate in potential_version_matches:
+                    if LooseVersion(candidate['version']) >= LooseVersion(highest_version):
+                        highest_version = candidate['version']
+                if highest_version != dep['version']:
+                    _log.info("Increasing version to %s for dependency %s.", highest_version, orig_dep['name'])
```
boegel: Also mention current dep version, and keep consistent w.r.t. terminology:

```python
_log.info("Upgrading version of %s dependency from %s to %s", dep['name'], dep['version'], highest_version)

File: easybuild/framework/easyconfig/tweak.py [hunk 20]
```diff
@@ -833,6 +950,71 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                    _log.info("Depending on your configuration, this will be resolved with one of the following "
+                              "easyconfigs: %s", '\n'.join(cand['path'] for cand in potential_version_matches
+                                                           if cand['version'] == highest_version))
+                    orig_dep['version'] = highest_version
+                    if orig_dep['versionsuffix'] in versonsuffix_mapping:
+                        orig_dep['versionsuffix'] = versonsuffix_mapping[orig_dep['versionsuffix']]
+                    orig_dep['short_mod_name'] = ActiveMNS().det_short_module_name(dep)
+                    orig_dep['full_mod_name'] = ActiveMNS().det_full_module_name(dep)
```
boegel: @ocaisa We're also doing this above, can we avoid this code duplication?

Something like:

```python
if 'binutils' in ...:
    ...
    dep_changed = True
elif update_dep_versions:
    ...
    dep_changed = True

if dep_changed:
    orig_dep['short_mod_name'] = ActiveMNS().det_short_module_name(dep)
    orig_dep['full_mod_name'] = ActiveMNS().det_full_module_name(dep)
    _log.info("Updated module names for dep %s v%s: %s (short), %s (long)",
              orig_dep['name'], orig_dep['version'], orig_dep['short_mod_name'], orig_dep['full_mod_name'])

File: test/framework/easyconfigs/test_ecs/g/gzip/gzip-1.6-iccifort-2016.1.150-GCC-4.9.3-2.25.eb [hunk 21]
```diff
@@ -0,0 +1,40 @@
...
+# License::   MIT/GPL
+# $Id$
+#
+# This work implements a part of the HPCBIOS project and is a component of the policy:
+# http://hpcbios.readthedocs.org/en/latest/HPCBIOS_06-19.html
+##
+easyblock = 'ConfigureMake'
+
+name = 'gzip'
+version = '1.4'
```
boegel: @ocaisa Should be `1.6` here?

easybuild/framework/easyconfig/tweak.py
ocaisa: We are (I think) working with a fully parsed which means the template has been resolved (I might be wrong here). In the dump method if the python version doesn't match that in the prefix (which happens if we update the python version) I imagine that the template isn't put back in place...am I wrong?

easybuild/framework/easyconfig/tweak.py
ocaisa: Well, I wanted to know that there's nothing else coming after when I do the search below, I can move it down there

easybuild/tools/robot.py
```diff
@@ -394,40 +407,49 @@ def search_easyconfigs(query, short=False, filename_only=False, terse=False):
...
+                    lines.extend(['', "Matching archived easyconfigs:", ''])
+                lines.extend(tmpl % hit for hit in archived_hits)
+            elif not terse:
+                cnt = len(archived_hits)
+                lines.extend([
+                    '',
+                    "Note: %d matching archived easyconfig(s) found, use --consider-archived-easyconfigs to see them" % cnt,
+                ])
+
+        print '\n'.join(lines)
```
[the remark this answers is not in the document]
ocaisa: Not sure what the problem is here

easybuild/framework/easyconfig/tweak.py
ocaisa: @boegel I could change this so that we only never do major version updates for anything (or make that an option)?

ocaisa:
@boegel This now works even for complex use cases such as
```
eb -D --robot --try-toolchain=foss,2018b --minimal-toolchains --disable-cleanup-tmpdir --try-update-deps Keras-2.1.3-intel-2017b-Python-3.6.3.eb
```
I don't know how to represent such a complex case in the tests though. What I was trying to cover with that was a case where we have python bundle that depends on another python bundle where there is an update in Python between the two toolchains!

easybuild/framework/easyconfig/tweak.py
ocaisa: Actually this is indeed required. At the very least I need the mapping to know how to search correctly for possible matches in `search_easyconfigs` there I need to know the actual version of Python in the version suffix, i.e., if I am updating Python to 3.7.0 I can only look for possible matches in the target toolchain with version suffix `-Python-3.7.0`

easybuild/framework/easyconfig/tweak.py
```diff
@@ -842,16 +973,103 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+    # Figure out what precedes the version
+    versionprefix = dep.get('versionprefix', '')
+    prefix_to_version = ''.join([x for x in [dep['name'] + '-', versionprefix] if x])
+    # Figure out the main versionsuffix (altered depending on toolchain in the loop below)
+    versionsuffix = dep.get('versionsuffix', '')
+    # If versionsuffix is in our mapping then we expect it to be updated
+    if versionsuffix in versonsuffix_mapping:
+        versionsuffix = versonsuffix_mapping[versionsuffix]
+    for toolchain in toolchain_hierarchy:
+        candidate_ver = '.*'  # using regex for *
```
ocaisa: @boegel This is probably the biggest remaining issue. This needs to be configurable somehow as at present it will, for example, upgrade from Python2 to Python3. Do I restrict it to minor versions only? I could do 2 searches, one global and one with only a minor version update, if the global returns a value I use that, otherwise I use the minor version update

ocaisa: Or we could maintain a blacklist for major updates?

ocaisa: I've gone with prioritising minor version updates and only if that fails trying major versions, works fine for a Python dep

ocaisa:
@boegel Monster PR now but all comments addressed and tests passing.

ocaisa:
@boegel I believe this is now ready for showtime. I simplified the versionsuffix mapping and have made the least intrusive thing the default (it searches first if it can update the patch version, if that fails the minor version and finally the major version). It will work very well with "standard" version numbers and, given that we typically only allow one software version per toolchain, should still work well in all cases.

I've used a very complex test case to verify this:
```
eb -D --robot --try-toolchain=intel,2018b --minimal-toolchains --try-update-deps Keras-2.1.3-intel-2017b-Python-3.6.3.eb
```
In this case we have two Python 3 versions in the target toolchain, 3.7.0 and 3.6.6. It correctly uses 3.6.6 for Python and manages to resolve a dep upgrade to `pkgconfig-1.3.1-intel-2018b-Python-3.6.6.eb` (from version 1.2.2) in the target toolchain

ocaisa:
This is working fine, there is one niggle that is actually a bug with the `dump` method that I can't figure out: in the tweaked easyconfigs `versionsuffix` is not being templated, it has the literal value (i.e., `-Python-3.6.6` rather than `'-Python-%(pyver)s'`), see the tweaked Keras easyconfig of 
```
eb -D --robot --try-toolchain=foss,2018b --minimal-toolchains --disable-cleanup-tmpdir --try-update-deps Keras-2.1.3-intel-2017b-Python-3.6.3.eb
```

victorusu:
@ocaisa, is it possible to extend this PR to have the possibility to state the version of the dependency?

The reason for my question is because different software version may have different APIs. So, always using the latest version of the package may not be very useful. One example is [NCL](https://github.com/easybuilders/easybuild-easyconfigs/blob/96d4f7857072ade7e058ed0f83b53791ef6b6b12/easybuild/easyconfigs/n/NCL/NCL-6.4.0-intel-2017a.eb).

Among its very long list of decencies, NCL 6.4.0 depends on  
* HDF5 version 1.8; and 
* JasPer 1.900.1. 

Since JasPer 2.0.12 is available in the robot path and HDF5 1.10.1 is available in the robot path, 
`eb --try-update-deps NCL-6.4.0-intel-2017b.eb` will not work because NCL 6.4.0 cannot be used with JasPer 2.0.x and HDF5 1.10.1.

But if we could say
```
eb --try-update-deps=JasPer,1.900.1:HDF5,1.8.6 NCL-6.4.0-intel-2017b.eb
```
Then, all other NCL 6.4.0 dependencies would be upgraded and the code would work.

ocaisa:
@victorusu I'm very open to this idea, an override mechanism like this is very necessary. This PR is already very massive though, and I'm pretty sure @boegel doesn't want to have to keep re-reviewing it, I'd prefer to get this merged first and then think about improving it.

boegel:
+1 on getting this merged as is first, which will already be a significant effort (I've been trying to get to this, haven't found enough time yet...)

@ocaisa Can you first look into fixing the merge conflicts with current `develop`?

ocaisa:
@boegel Done...and tests are passing

ocaisa:
@boegel Yeah, it would be easy to make it work without `--try-toolchain` (if that were true, you just make the trivial mapping from the toolchain to itself). I don't fully understand the use case for that though, I guess you are thinking of this for old toolchains that people still use?

easybuild/tools/options.py
```diff
@@ -318,6 +318,9 @@ def software_options(self):
                                None, 'store', None, {'metavar': 'NAME'}),
             'toolchain-version': ("Search and build with given toolchain version",
                                   None, 'store', None, {'metavar': 'VERSION'}),
+            'update-deps': ("Try to update versions of the dependencies of an easyconfig based on what is available in "
```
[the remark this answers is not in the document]
ocaisa: I thought all the `try` options were supposed to be listed here, when you say "below" do you mean together with `map-toolchains`? Reading it, I understand what you mean, will change

easybuild/framework/easyconfig/tweak.py
```diff
@@ -843,16 +963,126 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                                                                            versonsuffix_mapping=versonsuffix_mapping)
+                # Only highest version match is retained by default in potential_version_matches, compare that version
+                # to the original version and replace if appropriate (upgrades only).
+                if potential_version_matches:
+                    highest_version_match = potential_version_matches[0]['version']
+                    if LooseVersion(highest_version_match) > LooseVersion(dep['version']):
+                        _log.info("Updating version of %s dependency from %s to %s", dep['name'], dep['version'],
+                                  highest_version_match)
+                        _log.info("Depending on your configuration, this will be resolved with one of the following "
+                                  "easyconfigs: \n%s", '\n'.join(cand['path'] for cand in potential_version_matches))
```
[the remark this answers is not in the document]
ocaisa: No, not by default (`highest_versions_only=True`), what are listed are all the files in the search path that provide the highest version. The distinction is that what is in the search path may not be in the same as what is in your robot path

easybuild/framework/easyconfig/tweak.py
```diff
@@ -843,16 +963,126 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                full_versionsuffix = ''.join([x for x in [toolchain_suffix, versionsuffix, EB_FORMAT_EXTENSION]
+                                              if x])
+
+                # prepend/append version prefix/suffix
+                depver = ''.join([x for x in ['^', prefix_to_version, candidate_ver, full_versionsuffix] if x])
+                cand_paths = search_easyconfigs(depver, consider_extra_paths=False, print_result=False,
+                                                case_sensitive=True)
+                # Filter out easyconfigs that have been tweaked in this instance (they sit in the tempdir in a subdir
+                # that starts with 'tweaked_*')
+                tweaked_ec_stub = os.path.join(tempfile.gettempdir(), 'tweaked_')
```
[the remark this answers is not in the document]
ocaisa: For the first point this is true.
For the second, the specific directory where the tweaked files are is specific to this instance (I thought, but am no longer sure), i.e., it doesn't exist until we create it.
I don't know if the tweaked path is stored, if not I guess I can find a way to pass it in as an arg

easybuild/framework/easyconfig/tweak.py
```diff
@@ -843,16 +963,126 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+                    highest_version_match = potential_version_matches[0]['version']
+                    if LooseVersion(highest_version_match) > LooseVersion(dep['version']):
+                        _log.info("Updating version of %s dependency from %s to %s", dep['name'], dep['version'],
+                                  highest_version_match)
+                        _log.info("Depending on your configuration, this will be resolved with one of the following "
+                                  "easyconfigs: \n%s", '\n'.join(cand['path'] for cand in potential_version_matches))
+                        orig_dep['version'] = highest_version_match
+                        if orig_dep['versionsuffix'] in versonsuffix_mapping:
+                            dep['versionsuffix'] = versonsuffix_mapping[orig_dep['versionsuffix']]
+                            orig_dep['versionsuffix'] = versonsuffix_mapping[orig_dep['versionsuffix']]
```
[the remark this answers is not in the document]
ocaisa: I am struggling to find any reason why we do :)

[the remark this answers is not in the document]
ocaisa: Ah, I took this approach from somewhere else in the code base. I don't understand what this does
```
        for idx, dep in enumerate(parsed_ec['ec'][key]):

            # reference to original dep dict, this is the one we should be updating
            orig_dep = parsed_ec['ec']._config[key][0][idx]
```
so I assumed it is somehow necessary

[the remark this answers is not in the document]
ocaisa: I think it is necessary because one of them is parsed and the other isn't but you know better than me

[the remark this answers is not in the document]
ocaisa: I tried removing `orig_dep` and you do need both, you need the parsed deps for the MNS

ocaisa:
@boegel Actually, you can already use `try-update-deps` without `try-toolchain`, the relevant logic starts at 
```
    update_dependencies = build_specs.get('update_deps', None)
    if 'toolchain_name' in build_specs or 'toolchain_version' in build_specs or update_dependencies:
```

easybuild/tools/options.py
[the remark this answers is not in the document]
boegel: For most `--try-*` options, the "non-try" version is also meaning (although not used very often), it basically determines specifics for the easyconfigs to consider.

easybuild/framework/easyconfig/tweak.py
```diff
@@ -79,6 +83,9 @@ def ec_filename_for(path):
 
 def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
     """Tweak list of easyconfigs according to provided build specifications."""
+    # keep track of originally listed easyconfigs (via their path)
```
boegel: not path, filename

easybuild/framework/easyconfig/tweak.py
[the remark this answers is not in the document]
boegel: OK

easybuild/framework/easyconfig/tweak.py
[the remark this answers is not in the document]
boegel: What do you mean with "parsed deps" exactly? Where template values are resolved?
Where is the MNS relevant here?

boegel:
@ocaisa When I tried `--try-update-deps` without `--try-toolchain*`, the deps were not being updated.

My use case was having an easyconfig where only the toolchain was (manually) changed, but the dependencies were not in sync yet with other easyconfigs using the same toolchain.
I was expecting that `--try-update-deps` would fix that for me...

easybuild/framework/easyconfig/tweak.py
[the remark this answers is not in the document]
ocaisa: Maybe I should edit that comment :) My understanding is that `dep` has been parsed but `orig_dep` has not and `orig_dep` is the one that is relevant to the `dump` method which is why we need that target specifically. I assumed it must be related to the MNS because below the MNS is called to determine the module name using `dep` and not `orig_dep`.

The truth is, I don't know what is wrong but it doesn't work correctly if you replace `dep` by `orig_dep` (and modify the name to just `dep`), the only places where this can have an impact is
```
dep_tc_name = dep['toolchain']['name']  # unlikely
potential_version_matches = find_potential_version_mappings(dep, toolchain_mapping, versonsuffix_mapping=versonsuffix_mapping)
LooseVersion(dep['version'])  # unlikely
orig_dep['short_mod_name'] = ActiveMNS().det_short_module_name(dep)
orig_dep['full_mod_name'] = ActiveMNS().det_full_module_name(dep)
```

There is a material difference in the output with the following case if you try to make the switch (at least when I tried it):
```
eb -D --robot --try-toolchain=foss,2018b --minimal-toolchains --disable-cleanup-tmpdir --try-update-deps Keras-2.1.3-intel-2017b-Python-3.6.3.eb
```

ocaisa:
@boegel Urgh, why would you do that? So you want to start from an easyconfig with dependencies that are potentially not resolvable and fix them? That is really a different feature that what is implemented here.

easybuild/framework/easyconfig/tweak.py
```diff
@@ -79,6 +83,9 @@ def ec_filename_for(path):
 
 def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
     """Tweak list of easyconfigs according to provided build specifications."""
+    # keep track of originally listed easyconfigs (via their filename)
+    listed_ec_filenames = [os.path.basename(ec['spec']) for ec in easyconfigs]
```
ocaisa: @boegel Is this really wise actually? You just mentioned a use case where the file name does not correspond to what is in the file, if we don't keep the entire path then we have lost uniqueness

ocaisa:
@boegel Not sure what you expect to happen here, without knowing the original

> @boegel Urgh, why would you do that? So you want to start from an easyconfig with dependencies that are potentially not resolvable and fix them? That is really a different feature that what is implemented here.

My guess would be that https://github.com/easybuilders/easybuild-framework/pull/2599/files#diff-a902da109857bb737803736816a8f586R852 is failing in this case

easybuild/framework/easyconfig/tweak.py
boegel: Right, I overlooked that...

But with the changes you made, full path is also not a good idea, since you use the location of the files and make assumptions.

boegel:
@ocaisa Imagine if I first used `--try-toolchain`, and then realized I also have to update the dependencies. The use case is not that strange imho.

If that is beyond the scope of this PR, that's OK, but my gut feeling tells me it shouldn't be that hard to achieve...

W.r.t. `map_common_versionsuffixes`, can't we circumvent that entirely by grabbing the "raw" `versionsuffix` value, which usually has a template like `%(pyver)s` rather than a hardcoded value?

ocaisa:
@boegel Which `try-toolchain` are you using? Sounds like the old regex case.

In the 2 step case you would need to ensure that the tweaked deps from the first command are in your robot path when calling the second command. If that was true then I think this should still work.

easybuild/framework/easyconfig/tweak.py
ocaisa: I don't think I do any more (I use `alt_easyconfig_paths()` now) so I believe I can switch this back

boegel:
Hmm, can't seem to reproduce what I was seeing earlier easily now...

It was something like this (but then without the missing deps error):

```
$ cp easybuild/easyconfigs/p/Python/Python-3.6.1-foss-2017a.eb Python-test.eb

$ vim Python-test.eb.  # change toolchain to foss/2018b

$ eb Python-test.eb --try-update-deps --ignore-osdeps
== temporary log file in case of crash /tmp/eb-pBLTDf/easybuild-yMHiUh.log
ERROR: Missing dependencies: ncurses/6.0-foss-2018b, SQLite/3.17.0-foss-2018b, XZ/5.2.3-foss-2018b (no easyconfig file or existing module found)

**Comment by ocaisa** [comment]
I'm still of the opinion that you're really asking to fix a broken easyconfig so the error seems reasonable to me. While I'd say there is plenty in this PR to help you do that, I do think it is out of scope to try include that here.

Thinking about it, you're way more likely to run into this error than a success with such a manual approach (since this is exactly the reason I implemented the new `--try-toolchain` method).

**Comment by ocaisa** [comment]
To be more specific, if we had `intel/2019b` which is just released (i.e. not much software available) and I take my `Python` for `foss/2018a` and change the name, most deps are not probably not resolvable at all. You're relying on the fact that every dep in the hierarchy has a version `>=` to the dep version. That is quite a big prereq.

If you use `--try-toolchain=intel,2019b --try-update-deps` things will not fail.

**Comment by boegel** [comment]
@ocaisa Naive question (playing devil's advocate here, since I think we'll need to document this properly): why do the dependencies need to be *resolved*?

Can you just take the dep specs as is in the easyconfig, and then look for the closest match in the current toolchain (whether that is specified via `toolchain =` or `--try-toolchain`)?

Is there a strict requirement for easyconfigs being available for the dependencies just for updating them?

**Comment by ocaisa** [comment]
> @ocaisa Naive question (playing devil's advocate here, since I think we'll need to document this properly): why do the dependencies need to be _resolved_?
 
Because at the moment this is all happening in the same logical step as `--try-toolchain` where this is required. If the source and destination toolchains match we could avoid the resolve step I guess but this opens the whole thing up to failing miserably when the dep does not appear at all in the target toolchain (so there is nothing to map to).

> Can you just take the dep specs as is in the easyconfig, and then look for the closest match in the current toolchain (whether that is specified via `toolchain =` or `--try-toolchain`)?

These are two different things. If you use `--try-toolchain` you are tweaking the entire dependency graph, if the dep doesn't already exist in the target we pick up the tweaked version. With a modified easyconfig this avenue is not open and things can fail hard.

> Is there a strict requirement for easyconfigs being available for the dependencies just for updating them?

In the manual approach you describe there would be a strict requirement that they exist in the target toolchain with a version `>=` the version in the easyconfig, otherwise things fail.

I could see a way forward but I would really prefer to see this in another PR:

- Add conditional code for the case where the target and the source toolchain are identical
- Don't resolve deps in this case
- Try to find updates (downgrades should be possible too I guess in this scenario)
- Ensure that each resulting easyconfig is resolvable

File: easybuild/framework/easyconfig/tweak.py (hunk 27 above)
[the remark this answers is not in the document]
Comment (ocaisa) [reply easybuild/framework/easyconfig/tweak.py]: Just to confirm, I redid this change and there are many dependencies that are not updated in the case where `orig_dep` and `dep` are one and the same. Output difference from command
```
eb -D --robot --try-toolchain=foss,2018b --minimal-toolchains --disable-cleanup-tmpdir --try-update-deps Keras-2.1.3-intel-2017b-Python-3.6.3.eb
```
is
```
20c21
<  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/l/libyaml/libyaml-0.1.7-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | libyaml/0.1.7)
>  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/l/libyaml/libyaml-0.2.1-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | libyaml/0.2.1)
40a40,41
>  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/p/protobuf/protobuf-3.6.1-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | protobuf/3.6.1)
50,56c51,57
<  * [ ] /tmp/eb-2afs_g/tweaked_dep_easyconfigs/Theano-1.0.1-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | Theano/1.0.1-Python-3.6.6)
<  * [ ] /tmp/eb-2afs_g/tweaked_dep_easyconfigs/wheel-0.30.0-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | wheel/0.30.0-Python-3.6.6)
<  * [ ] /tmp/eb-2afs_g/tweaked_dep_easyconfigs/PyYAML-3.12-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | PyYAML/3.12-Python-3.6.6)
<  * [ ] /tmp/eb-2afs_g/tweaked_dep_easyconfigs/pkgconfig-1.2.2-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | pkgconfig/1.2.2-Python-3.6.6)
<  * [ ] /tmp/eb-2afs_g/tweaked_dep_easyconfigs/TensorFlow-1.5.0-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | TensorFlow/1.5.0-Python-3.6.6)
<  * [ ] /tmp/eb-2afs_g/tweaked_dep_easyconfigs/h5py-2.7.1-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | h5py/2.7.1-Python-3.6.6)
<  * [ ] /tmp/eb-2afs_g/tweaked_easyconfigs/Keras-2.1.3-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | Keras/2.1.3-Python-3.6.6)
---
>  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/t/Theano/Theano-1.0.3-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | Theano/1.0.3-Python-3.6.6)
>  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/w/wheel/wheel-0.31.1-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | wheel/0.31.1-Python-3.6.6)
>  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/p/PyYAML/PyYAML-3.13-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | PyYAML/3.13-Python-3.6.6)
>  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/p/pkgconfig/pkgconfig-1.3.1-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | pkgconfig/1.3.1-Python-3.6.6)
>  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/h/h5py/h5py-2.8.0-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | h5py/2.8.0-Python-3.6.6)
>  * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/t/TensorFlow/TensorFlow-1.12.0-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | TensorFlow/1.12.0-Python-3.6.6)
>  * [ ] /tmp/eb-neWxA8/tweaked_easyconfigs/Keras-2.1.3-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | Keras/2.1.3-Python-3.6.6)
```

File: test/framework/tweak.py [hunk 30]
```diff
@@ -331,6 +404,24 @@ def test_map_easyconfig_to_target_tc_hierarchy(self):
...
+        key, value = 'toolchain', iccifort_binutils_tc
+        self.assertTrue(key in tweaked_dict and value == tweaked_dict[key])
+        # Also check that binutils has been mapped
+        for key, value in {'name': 'binutils', 'version': '2.25', 'versionsuffix': ''}.items():
+            self.assertTrue(key in tweaked_dict['builddependencies'][0]
+                            and value == tweaked_dict['builddependencies'][0][key])
+        # Also check that the gzip dependency was upgraded
+        for key, value in {'name': 'gzip', 'version': '1.6', 'versionsuffix': ''}.items():
+            self.assertTrue(key in tweaked_dict['dependencies'][0]
+                            and value == tweaked_dict['dependencies'][0][key])
```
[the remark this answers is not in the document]
ocaisa: I'm caught in a trap, it doesn't like it before and it doesn't like it after

ocaisa:
For those interested, here is some sample output from the PR to get an idea of what happens. Note that the only thing a tweaked dep is used for is the specific easyconfig I asked for.
```
alanc@~/EasyBuild_Git/easybuild-framework(allow_dep_upgrades)$ eb --dry-run --robot --try-toolchain=foss,2018b --minimal-toolchains --try-update-deps --disable-cleanup-tmpdir TensorFlow-1.8.0-foss-2017b-Python-3.6.3.eb
== temporary log file in case of crash /tmp/eb-Hgmsx9/easybuild-NBsH11.log
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/j/Java/Java-1.8.0_231.eb (module: Core | Java/1.8.0_231)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/j/Java/Java-1.8.eb (module: Core | Java/1.8)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.18.eb (module: Core | M4/1.4.18)
 * [x] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/z/zlib/zlib-1.2.11.eb (module: Core | zlib/1.2.11)
 * [x] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/h/help2man/help2man-1.47.4.eb (module: Core | help2man/1.47.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.17.eb (module: Core | M4/1.4.17)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/Bison/Bison-3.0.4.eb (module: Core | Bison/3.0.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/f/flex/flex-2.6.4.eb (module: Core | flex/2.6.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/binutils/binutils-2.30.eb (module: Core | binutils/2.30)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/g/GCCcore/GCCcore-7.3.0.eb (module: Core | GCCcore/7.3.0)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/z/zlib/zlib-1.2.11-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | zlib/1.2.11)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/h/help2man/help2man-1.47.4-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | help2man/1.47.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.18-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | M4/1.4.18)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/Bison/Bison-3.0.4-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | Bison/3.0.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/Bison/Bison-3.0.5-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | Bison/3.0.5)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/f/flex/flex-2.6.4-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | flex/2.6.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/binutils/binutils-2.30-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | binutils/2.30)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | bzip2/1.0.6)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/g/GCC/GCC-7.3.0-2.30.eb (module: Core | GCC/7.3.0-2.30)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/n/ncurses/ncurses-6.1-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | ncurses/6.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.3.1-GCC-7.3.0-2.30.eb (module: Compiler/GCC/7.3.0-2.30 | OpenBLAS/0.3.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/Bazel/Bazel-0.20.0-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | Bazel/0.20.0)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | libreadline/7.0)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/t/Tcl/Tcl-8.6.8-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | Tcl/8.6.8)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/s/SQLite/SQLite-3.24.0-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | SQLite/3.24.0)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | Autoconf/2.69)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/l/libffi/libffi-3.2.1-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | libffi/3.2.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/a/Automake/Automake-1.16.1-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | Automake/1.16.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb (module: Core | ncurses/6.0)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/g/gettext/gettext-0.19.8.1.eb (module: Core | gettext/0.19.8.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/x/XZ/XZ-5.2.4-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | XZ/5.2.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/l/libxml2/libxml2-2.9.8-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | libxml2/2.9.8)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/l/libtool/libtool-2.4.6-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | libtool/2.4.6)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/a/Autotools/Autotools-20180311-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | Autotools/20180311)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/g/GMP/GMP-6.1.2-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | GMP/6.1.2)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/x/xorg-macros/xorg-macros-1.19.2-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | xorg-macros/1.19.2)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/n/numactl/numactl-2.0.11-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | numactl/2.0.11)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/l/libpciaccess/libpciaccess-0.14-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | libpciaccess/0.14)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/h/hwloc/hwloc-1.11.10-GCCcore-7.3.0.eb (module: Compiler/GCCcore/7.3.0 | hwloc/1.11.10)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/o/OpenMPI/OpenMPI-3.1.1-GCC-7.3.0-2.30.eb (module: Compiler/GCC/7.3.0-2.30 | OpenMPI/3.1.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/g/gompi/gompi-2018b.eb (module: Core | gompi/2018b)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/f/FFTW/FFTW-3.3.8-gompi-2018b.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | FFTW/3.3.8)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/s/ScaLAPACK/ScaLAPACK-2.0.2-gompi-2018b-OpenBLAS-0.3.1.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | ScaLAPACK/2.0.2-OpenBLAS-0.3.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/f/foss/foss-2018b.eb (module: Core | foss/2018b)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/p/Python/Python-3.6.6-foss-2018b.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | Python/3.6.6)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/w/wheel/wheel-0.31.1-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | wheel/0.31.1-Python-3.6.6)
 * [ ] /tmp/eb-Hgmsx9/tweaked_easyconfigs/TensorFlow-1.8.0-foss-2018b-Python-3.6.6.eb (module: MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1 | TensorFlow/1.8.0-Python-3.6.6)
== Keeping temporary log file(s) /tmp/eb-Hgmsx9/easybuild-NBsH11.log* and directory /tmp/eb-Hgmsx9.
```

zao:
How would this interact with software that has dependencies with frozen versions because newer versions of the dependency is incompatible or buggy? Bad experiences with `libxc` comes to mind.

Can we maybe somehow annotate a dependency to indicate that it needs special consideration?

ocaisa:
@zao If the older version appears in the other toolchain it would prefer it, if only a newer version exists then it would update it. Annotating would be easy to consider but needs implementing.

boegel:
@ocaisa Playing around with this a bit, hit this issue, looks like a bug in the regex mode?

Why is it trying to use a non-existing easyconfig parameter `update_deps`?

```
$ eb Python-3.7.4-GCCcore-8.3.0.eb --try-software-version 3.8.1 --try-toolchain-version 9.2.0 --try-update-deps --ignore-osdeps
== temporary log file in case of crash /Users/kehoste/work/TMP/eb-kQUh5Z/easybuild-dWl0Xn.log

WARNING: Combining --try-toolchain* or --try-update-deps with other build options is not fully supported: using regex

ERROR: Unknown easyconfig parameter: update_deps (known: ['allow_prepend_abs_path', 'allow_system_deps', 'bitbucket_account', 'block', 'builddependencies', 'buildininstalldir', 'buildopts', 'buildstats', 'checksums', 'cleanupoldbuild', 'cleanupoldinstall', 'configopts', 'dependencies', 'deprecated', 'description', 'docpaths', 'docurls', 'dontcreateinstalldir', 'easyblock', 'easybuild_version', 'examples', 'exts_classmap', 'exts_default_options', 'exts_defaultclass', 'exts_download_dep_fail', 'exts_filter', 'exts_list', 'fix_perl_shebang_for', 'fix_python_shebang_for', 'github_account', 'group', 'hidden', 'hiddendependencies', 'homepage', 'include_modpath_extensions', 'installopts', 'keeppreviousinstall', 'keepsymlinks', 'key', 'license_file', 'license_server', 'license_server_port', 'maxparallel', 'modaliases', 'modaltsoftname', 'modextrapaths', 'modextravars', 'modloadmsg', 'modluafooter', 'modtclfooter', 'module_depends_on', 'moduleclass', 'moduleforceunload', 'moduleloadnoconflict', 'multi_deps', 'multi_deps_load_default', 'name', 'onlytcmod', 'osdependencies', 'parallel', 'patches', 'postinstallcmds', 'prebuildopts', 'preconfigopts', 'preinstallopts', 'pretestopts', 'recursive_module_unload', 'runtest', 'sanity_check_commands', 'sanity_check_paths', 'site_contacts', 'skip', 'skipsteps', 'software_license', 'software_license_urls', 'source_urls', 'sources', 'start_dir', 'stop', 'testopts', 'tests', 'toolchain', 'toolchainopts', 'unpack_options', 'unwanted_env_vars', 'upstream_contacts', 'usage', 'version', 'versionprefix', 'versionsuffix', 'whatis'])
```


Also, can we avoid that using `--try-software-version` results in switching to regex mode?
Just wondering, that doesn't need to block this PR...

Also, wasn't the plan to push `--try-update-deps` under `--experimental` for now?

ocaisa:
@boegel 

> Why is it trying to use a non-existing easyconfig parameter `update_deps`?

That's (I guess) how the regex mode works. I will modify the behaviour so it just exits in the scenario you tried. As things stand, this PR is not intended to to support changing software version and toolchain at the same time. I'll see how hard that would be to support exactly that, but if it's hard I would leave it as a feature request.

ocaisa:
@boegel I'd like your opinion on https://github.com/easybuilders/easybuild-framework/pull/2599/commits/d3942240b5f2bdf56ec8f8b30691dc5854edd210 I was pretty naive in my implementation there but maybe it is ok?

ocaisa:
@boegel In the last commit I introduced a restriction on what this can handle when you are doing a software version upgrade, if the exts_lists is not empty it just refuses to try. The reasoning here is that I can't gracefully do this, for example for Keras, I had the case where I tried to update the version but I need to update the Keras version inside the exts_list as well (because the package was actually a PythonBundle), I don't know a good way to do this.

ocaisa:
@boegel Ok, this can now also update software versions as well as dependencies  even in the case where the software version appears in `exts_list` (like it might for a `PythonBundle`), so your (updated) command line
```
eb Python-3.7.4-GCCcore-8.3.0.eb --try-software-version 3.8.1 --try-toolchain-version 9.2.0 --try-update-deps --ignore-osdeps -D --experimental --disable-cleanup-tmpdir
```
returns
```
== temporary log file in case of crash /tmp/eb-RraSlk/easybuild-FFh1z5.log
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.18.eb (module: Core | M4/1.4.18)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/Bison/Bison-3.3.2.eb (module: Core | Bison/3.3.2)
 * [x] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/z/zlib/zlib-1.2.11.eb (module: Core | zlib/1.2.11)
 * [x] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/h/help2man/help2man-1.47.4.eb (module: Core | help2man/1.47.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.17.eb (module: Core | M4/1.4.17)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/Bison/Bison-3.0.4.eb (module: Core | Bison/3.0.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/f/flex/flex-2.6.4.eb (module: Core | flex/2.6.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/binutils/binutils-2.32.eb (module: Core | binutils/2.32)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/g/GCCcore/GCCcore-9.2.0.eb (module: Core | GCCcore/9.2.0)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/h/help2man/help2man-1.47.10-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | help2man/1.47.10)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.18-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | M4/1.4.18)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/z/zlib/zlib-1.2.11-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | zlib/1.2.11)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/Bison/Bison-3.3.2-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | Bison/3.3.2)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/f/flex/flex-2.6.4-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | flex/2.6.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/b/binutils/binutils-2.32-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | binutils/2.32)
 * [ ] /tmp/eb-RraSlk/tweaked_dep_easyconfigs/bzip2-1.0.8-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | bzip2/1.0.8)
 * [ ] /tmp/eb-RraSlk/tweaked_dep_easyconfigs/ncurses-6.1-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | ncurses/6.1)
 * [ ] /tmp/eb-RraSlk/tweaked_dep_easyconfigs/Tcl-8.6.9-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | Tcl/8.6.9)
 * [ ] /tmp/eb-RraSlk/tweaked_dep_easyconfigs/libreadline-8.0-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | libreadline/8.0)
 * [ ] /tmp/eb-RraSlk/tweaked_dep_easyconfigs/SQLite-3.29.0-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | SQLite/3.29.0)
 * [ ] /tmp/eb-RraSlk/tweaked_dep_easyconfigs/libffi-3.2.1-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | libffi/3.2.1)
 * [x] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/n/ncurses/ncurses-6.1.eb (module: Core | ncurses/6.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/g/gettext/gettext-0.20.1.eb (module: Core | gettext/0.20.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/x/XZ/XZ-5.2.4-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | XZ/5.2.4)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | Autoconf/2.69)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/a/Automake/Automake-1.16.1-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | Automake/1.16.1)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/l/libtool/libtool-2.4.6-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | libtool/2.4.6)
 * [ ] /home/alanc/EasyBuild_Git/easybuild-easyconfigs/easybuild/easyconfigs/a/Autotools/Autotools-20180311-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | Autotools/20180311)
 * [ ] /tmp/eb-RraSlk/tweaked_dep_easyconfigs/GMP-6.1.2-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | GMP/6.1.2)
 * [ ] /tmp/eb-RraSlk/tweaked_easyconfigs/Python-3.8.1-GCCcore-9.2.0.eb (module: Compiler/GCCcore/9.2.0 | Python/3.8.1)
== Keeping temporary log file(s) /tmp/eb-RraSlk/easybuild-FFh1z5.log* and directory /tmp/eb-RraSlk.
```

ocaisa:
@Micket made a valid comment that highlights the naivity of the software version upgrading in this approach. The problem is that the `%(version)s` template could be used anywhere in the easyconfig, but since this approach only works on the fully parsed easyconfig, that templating is lost...which could be really bad in the case where  `%(version)s` is used in `source_urls` (for example).

To do this properly I would need to know the places where `%(version)s` appears, and update these accordingly. Another problem would be that for `patches` you can't do this (since the patch wouldn't exist).

Micket:
@ocaisa What about all the other fields, like `%(pyver)s`; does it all get expanded in the generated easyconfigs with the current approach?

ocaisa:
@Micket I just tested that with
```
eb Boost-1.55.0-foss-2016a-Python-2.7.11.eb --experimental --try-software-version 3.8.1 --try-toolchain-version 2018b --try-update-deps --ignore-osdeps -D --disable-cleanup-tmpdir
```
and the `pyver` template survived.

For "other fields", I don't know, I think we just have to try it out. I also checked what happens when the templates like `%(version)s` are used in used `source_urls` and this also survives:
```
eb Perl-5.28.1-GCCcore-8.2.0.eb --experimental --try-software-version 6.30.2 --try-toolchain-version 9.2.0 --try-update-deps --ignore-osdeps -D --disable-cleanup-tmpdir
``` 

So, as far as I can tell, there are still no issues to fix in this PR.

boegel:
> @Micket made a valid comment that highlights the naivity of the software version upgrading in this approach. The problem is that the `%(version)s` template could be used anywhere in the easyconfig, but since this approach only works on the fully parsed easyconfig, that templating is lost...which could be really bad in the case where `%(version)s` is used in `source_urls` (for example).
> 
> To do this properly I would need to know the places where `%(version)s` appears, and update these accordingly. Another problem would be that for `patches` you can't do this (since the patch wouldn't exist).

This goes way beyond the `%(version)s` template too. If there's a list comprehension in `sanity_check_paths` for example, it will get expanded in the resulting easyconfig, which is not what you want either.

The only way around that is to work on the raw easyconfig.

I'm not considering that a block for this PR though, since that problem already exists with the `--try-*` options I think, but it's something to keep in mind when we start working on the `eb --tweak-ec` (or `eb --change-ec` idea I mentioned during the user meeting).

The main difference with `--try-*` would be that it only makes minimal changes on the raw easyconfig contents, and then dumps the tweaked (raw) easyconfig file (rather than starting to build it), perhaps with an additional option to opt-in to also building the resulting easyconfigs.

We can probably learn a great deal from the current support we have for `--try-*`, but the current implementation is complicated, and I think we can do a better job starting afresh.

With that being said, I'll take another detailed look at this, and go ahead on merge it unless I run into any blockers (hopefully not).

ocaisa:
@boegel I should say that it doesn't actually seem to be the case that the templating is lost, there is a little bit of voodoo in here that I don't fully understand (see https://github.com/easybuilders/easybuild-framework/pull/2599#discussion_r260827759) and I think that might be why things are more robust than one might expect.

ocaisa:
@boegel I tested this with
```
eb fastq-tools-0.8-foss-2018b.eb --experimental --try-software-version 6.30.2 --try-toolchain-version 2019b --try-update-deps --ignore-osdeps -D --disable-cleanup-tmpdir
```
and indeed the list comprehensions are being expanded, as well as the subtoolchains for dependencies being explicitly inserted. Functionally this is no different but of course we can perhaps do better in the `dump()` method in maintaining the original formatting.

ocaisa:
#3261 solves the problem of the dumped easyconfig not being "toolchain hierarchy aware", for list comprehensions this is not really relevant to this PR, it is more an issue with the `dump` method in general.

easybuild/framework/easyconfig/tweak.py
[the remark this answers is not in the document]
boegel: It's not so much parsed vs not parsed, it has to do with the templating.

`dep` is a *copy* of the original dependency, with templates resolved; `orig_dep` is a ref to the original dependency

We may need to update both in this context to make things work correctly. Perhaps updating `dep` isn't strictly needed, but it's cheap, and we've spent enough time on this... ;)

boegel:
@ocaisa I've done a thorough last review of this PR this morning, and just pushed in a couple of minor changes, mostly:

* enhanced test for `search_easyconfigs` + minor bug fix
* simple end-to-end test for `--try-upgrade-deps`
* fixed broken `test_index_functions` (due to 2 additional easyconfigs)
* minor changes in comments & docstrings

With this, I consider this PR good to go, I'll merge it as soon as the tests go green...

Thanks *a lot* for all your efforts on this, this clearly took a lot of effort, and was dragged along for way too long, I'm happy we can finally get this merged...

PR #2604, opened 2018-10-02
determine file_info for all easyconfigs before any actual copying

migueldiascosta:
fixes an issue with `--new-pr` where both `new` and `update` labels were added (if submitted by a maintainer) when the PR includes multiple versions of the same new software


boegel:
Looks great, thanks for the fix @migueldiascosta!

PR #2605, opened 2018-10-02
Allow --force to use regex if --try-toolchain cannot map intelligently.

### LINKED ISSUE #2592: Improved `--try-toolchain` overly picky
The improved `--try-toolchain` now rejects cases that used to work perfectly, where the toolchain can be "downgraded" to a subtoolchain.

Example:
```
eb libxc-4.2.3-intel-2018a.eb --try-toolchain=iccifort,2018.3.222-GCC-7.3.0-2.30
== temporary log file in case of crash /localscratch/eb-oe4GFb/easybuild-77BHp6.log
ERROR: No possible mapping from source toolchain spec {'lapack_family': None, 'name': 'iimpi', 'blas_family': None, 'version': '2018a', 'cuda': None, 'comp_family': 'Intel', 'mpi_family': 'IntelMPI'} to target toolchain hierarchy specs [{'lapack_family': None, 'name': 'GCCcore', 'blas_family': None, 'version': '7.3.0', 'cuda': None, 'comp_family': 'GCC', 'mpi_family': None}, {'lapack_family': None, 'name': 'iccifort', 'blas_family': None, 'version': '2018.3.222-GCC-7.3.0-2.30', 'cuda': None, 'comp_family': 'Intel', 'mpi_family': None}]
```

@ocaisa

bartoldeman:
Fixes #2592

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -119,12 +119,13 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
             else:
                 target_toolchain['version'] = source_toolchain['version']
 
-            src_to_dst_tc_mapping = map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool)
-            if src_to_dst_tc_mapping is None:
+            try:
+                src_to_dst_tc_mapping = map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool)
+            except EasyBuildError as err:
                 if not build_option('force'):
-                    raise EasyBuildError("Toolchain %s is not equivalent to toolchain %s in terms of capabilities. "
+                    raise EasyBuildError("%s:\nToolchain %s is not equivalent to toolchain %s in terms of capabilities. "
                                          "Use --force if you are sure.",
-                                         target_toolchain['name'], source_toolchain['name'])
+                                         err.msg, target_toolchain['name'], source_toolchain['name'])
                 print_warning("Combining --try-toolchain with --force for toolchains with unequal capabilities: "
                               "using regex.")
                 revert_to_regex = True
@@ -732,8 +733,12 @@ def match_minimum_tc_specs(source_tc_spec, target_tc_hierarchy):
                 target_compiler_family = target_tc_spec['comp_family']
                 break
 
+    if not minimal_matching_toolchain:
+        raise EasyBuildError("No possible mapping from source toolchain spec %s to target toolchain hierarchy specs %s",
+                             source_tc_spec, target_tc_hierarchy)
+
     # Warn if we are changing compiler families, this is very likely to cause problems
-    if minimal_matching_toolchain and target_compiler_family != source_tc_spec['comp_family']:
+    if target_compiler_family != source_tc_spec['comp_family']:
         print_warning("Your request will result in a compiler family switch (%s to %s). Here be dragons!" %
                       (source_tc_spec['comp_family'], target_compiler_family))
 
@@ -773,10 +778,7 @@ def map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool):
     target_tc_hierarchy = get_toolchain_hierarchy(target_toolchain, incl_capabilities=True)
 
     for toolchain_spec in source_tc_hierarchy:
-        minimal_matching_toolchain = match_minimum_tc_specs(toolchain_spec, target_tc_hierarchy)
-        if not minimal_matching_toolchain:
-            return None
-        tc_mapping[toolchain_spec['name']] = minimal_matching_toolchain
+        tc_mapping[toolchain_spec['name']] = match_minimum_tc_specs(toolchain_spec, target_tc_hierarchy)
 
     # Check for presence of binutils in source and target toolchain dependency trees
     # (only do this when GCCcore is present in both and GCCcore is not the top of the tree)
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -120,7 +120,16 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
                 target_toolchain['version'] = source_toolchain['version']
 
             src_to_dst_tc_mapping = map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool)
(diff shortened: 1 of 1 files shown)


bartoldeman:
This just moves the exception two levels up in the call stack and adjusts for --force. Maybe there is a better way @ocaisa?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -120,7 +120,17 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
...
             src_to_dst_tc_mapping = map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool)
+            if src_to_dst_tc_mapping is None:
+                if not build_option('force'):
+                    raise EasyBuildError("Toolchain %s is not equivalent to toolchain %s in terms of capabilities. "
+                                         "Use --force if you are sure.",
+                                         target_toolchain['name'], source_toolchain['name'])
+                print_warning("Combining --try-toolchain with --force for toolchains with unequal capabilities: "
+                              "using regex.")
+                revert_to_regex = True
+                modifying_toolchains = False
```
boegel: @bartoldeman Please refactor this a bit to i) use positive logic, ii) use `if/else` (it's a bit confusing now imho).

I would also clarify the "regex" part a bit in the warning, and also the error message(see below).

```python
if build_option('force'):
    print_warning("Combining --try-toolchain with --force for toolchains with unequal capabilities: "
                  "disabling recursion & not changing (sub)toolchains.")
    revert_to_regex = True
    modifying_toolchains = False
else:
    raise EasyBuildError("Toolchain %s is not equivalent to toolchain %s in terms of capabilities, aborting "
                         "(if you know what you are doing, you can use --force to proceed anyway)",
                         target_toolchain['name'], source_toolchain['name'])

File: easybuild/framework/easyconfig/tweak.py [hunk 2]
```diff
@@ -767,7 +773,10 @@ def map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool):
     target_tc_hierarchy = get_toolchain_hierarchy(target_toolchain, incl_capabilities=True)
 
     for toolchain_spec in source_tc_hierarchy:
-        tc_mapping[toolchain_spec['name']] = match_minimum_tc_specs(toolchain_spec, target_tc_hierarchy)
+        minimal_matching_toolchain = match_minimum_tc_specs(toolchain_spec, target_tc_hierarchy)
+        if not minimal_matching_toolchain:
+            return None
+        tc_mapping[toolchain_spec['name']] = minimal_matching_toolchain
```
boegel: Same here, please use positive logic + `if`/`else`, and maybe also add a debug log message:

```python
if minimal_matching_toolchain:
    tc_mapping[toolchain_spec['name']] = minimal_matching_toolchain
else:
    _log.debug("No minimal matching toolchain found for %s using %s", toolchain_spec, target_tc_hierarchy)
    return None
```

I'm also not a big fan of the inline `return` statement. Can we avoid this?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -722,12 +732,8 @@ def match_minimum_tc_specs(source_tc_spec, target_tc_hierarchy):
                 target_compiler_family = target_tc_spec['comp_family']
                 break
 
-    if not minimal_matching_toolchain:
-        raise EasyBuildError("No possible mapping from source toolchain spec %s to target toolchain hierarchy specs %s",
-                             source_tc_spec, target_tc_hierarchy)
-
     # Warn if we are changing compiler families, this is very likely to cause problems
-    if target_compiler_family != source_tc_spec['comp_family']:
+    if minimal_matching_toolchain and target_compiler_family != source_tc_spec['comp_family']:
```
boegel: @bartoldeman Wouldn't this make more sense?

```python
if target_compiler_family and target_compiler_family != source_tc_spec['comp_family']:
```

boegel:
Why is the test failing anyway? The error should still be thrown unless `--force` is used?

It seems to signal that the default behavior is being changed?

easybuild/framework/easyconfig/tweak.py
ocaisa: This is not quite the same thing. Here you are saying if we have `force` then asssume we want regex, whereas the other approach is to only go looking for force if the mapping was not possible.

bartoldeman:
I simplified the change by simply catching the deeper down exception into a new higher-level exception. This way I no longer need to distinguish between {} and None which is always a bit icky IMHO

easybuild/framework/easyconfig/tweak.py
```diff
@@ -119,8 +119,22 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
             else:
                 target_toolchain['version'] = source_toolchain['version']
 
-            src_to_dst_tc_mapping = map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool)
+            try:
+                src_to_dst_tc_mapping = map_toolchain_hierarchies(source_toolchain, target_toolchain, modtool)
+            except EasyBuildError as err:
+                if build_option('force'):
```
boegel: @bartoldeman We should probably check whether the raised exception is indeed the one raised by `match_minimum_tc_specs`, since (at least in theory) an `EasyBuildError` could be raised for a variety of reasons....

That could look something like this (untested):

```python
except EasyBuildError as err:
    # make sure exception was raised by match_minimum_tc_specs because toolchain mapping could not be done
    if "No possible mapping from source toolchain" in err.msg:

        if build_option('force'):
            warning_msg = "Combining --try-toolchain with --force for toolchains with unequal capabilities: "
            warning_msg += "disabling recursion and not changing (sub)toolchains for dependencies."
            print_warning(warning_msg)
            revert_to_regex = True
            modifying_toolchains = False
        else:
            error_msg = err.msg + '\n'
            error_msg += "Toolchain %s is not equivalent to toolchain %s in terms of capabilities. "
            error_msg += "(if you know what you are doing, you can use --force to proceed anyway)"
            raise EasyBuildError(error_msg, target_toolchain['name'], source_toolchain['name'])
    else:
        # simply re-raise the exception if something else went wrong
        raise err

bartoldeman:
Reopening for @boegelbot  and his master @boegel to see if this was a Travis glitch with Ubuntu updates...

Issue #2601, opened 2018-10-01
Robot does not recognize installed modulerc aliases with hierarchical MNS 


vanzod:
As reported in https://github.com/easybuilders/easybuild-easyblocks/issues/1534, the use of short names in a `.modulerc` wrapper file (introduced with https://github.com/easybuilders/easybuild-easyblocks/pull/1535) prevents robot to correctly identify installed module wrappers when a hierarchical MNS is used. This forces EB to re-install the wrapper every time it is needed as a dependency. The installation will then fail since the `.modulerc` file cannot be replaced without an explicit `--force` flag:
```
 == FAILED: Installation ended unsuccessfully (build directory: /mnt/ramdisk/Java/1.8/dummy-dummy): build failed (first 300 chars): Found existing file at /opt/easybuild/modules/all/Core/Java/.modulerc.lua, not overwriting without --force
```

boegel:
The actual problem here is that we check whether modules already exist based on the full module name, which is for example `Core/Java/1.8` rather than the short/user-facing module name `Java/1.8`.

The reason for this is that for non-Core modules may not be visible yet, unless one or more modules that open the part of the module tree where that module lives is loaded.

Using the full module name, we can always check whether a module file already exists by simply adding the top of the module tree to `$MODULEPATH`.
But this only works for module files that actually exist, not for wrappers defined in a `.modulerc`... That's the issue we're hitting now.

The fix that was done in https://github.com/easybuilders/easybuild-easyblocks/pull/1535 is correct, the wrapper should be defined for the short module name, i.e. `Java/1.8` is a wrapper/alias/symbolic version for `Java/1.8.0_181` (not for `Core/Java/1.8.0_181`).

I've been thinking what the proper fix for this would be...

This probably involves enhancing/fixing the `exist` function in the `Modules` implementation, which now already checks with `ml avail` first, and falls back to using `ml show` (which works for hidden modules for example).
I see two options:
* let `exist` also explicitly check whether the specified module exists as a wrapper, by looking for a `.modulerc` where the wrapper is defined; one problem here is that both `Tcl` & `Lua` syntax need to be taken into account
* also check with `ml spider` as a fallback, but that's also not trivial since we may be picking up modules (& wrappers) in other locations...

cc @zao

vanzod:
[...]

PR #2606, opened 2018-10-02
also check for module wrappers in 'ModulesTool.exist' method

boegel:
fix for #2601 ~~(clearly still WIP, enhanced test will fail in current status)~~

### DIFF
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -42,7 +42,7 @@ from unittest import TextTestRunner
 import easybuild.tools.modules as mod
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import copy_file, copy_dir, mkdir, read_file, write_file
+from easybuild.tools.filetools import copy_file, copy_dir, mkdir, read_file, remove_file, write_file
 from easybuild.tools.modules import EnvironmentModules, EnvironmentModulesTcl, Lmod, NoModulesTool
 from easybuild.tools.modules import curr_module_paths, get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.modules import invalidate_module_caches_for, modules_tool, reset_module_caches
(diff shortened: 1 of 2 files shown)


vanzod:
Tested with `R-3.5.1-foss-2018b.eb`. `Java/1.8` is correctly recognized as installed.

boegel:
Some tests are failing because of problems with the tests themselves, which I'll fix, but one failing test (`test_stale_module_caches`) highlights a problem that should be fixed: we're just checking for an existing `.modulerc` that defines a wrapper, but we're not checking whether the underlying module being wrapped actually exists.

Fixing that is fairly straightforward: check which module file is being wrapped, and then check whether that exists before spitting out a `True` for the existence check of the wrapper. This situation is probably mostly academic in the sense that it's unlikely to happen in practice (unless the module file that the wrapper refers to is removed, of course).

There's a complication though... The wrapped module is specified with the 'short' module name (e.g. `Java/1.8.0_181`) rather than the full module name (e.g. `Core/Java/1.8.0_181`), so we need to "translate" the short name to the full one based on the provided full name of the wrapper module (e.g. `Core/Java/1.8`).

Fixes for both issues coming up in next push (cfr. commits 06621bffc2 + 525d92e03)...

boegel:
@vanzod This should be good to go now...

test/framework/modules.py
```diff
@@ -178,7 +190,7 @@ def test_exists(self):
         self.assertTrue('Core/Java/1.8.0_181' in self.modtool.available())
         self.assertEqual(self.modtool.exist(['Core/Java/1.8.0_181']), [True])
         self.assertEqual(self.modtool.exist(['Core/Java/1.8']), [True])
-        self.assertTrue(self.modtool.module_wrapper_exists('Core/Java/1.8'))
+        self.assertEqual(self.modtool.module_wrapper_exists('Core/Java/1.8'), 'Core/Java/1.8.0_181')
```
vanzod: We have to be careful here if the `Java/1.8` wrapper will point to a different module in the future.

test/framework/modules.py
```diff
@@ -200,7 +213,7 @@ def test_exists(self):
             self.assertTrue('Core/Java/1.8.0_181' in self.modtool.available())
             self.assertEqual(self.modtool.exist(['Core/Java/1.8.0_181']), [True])
             self.assertEqual(self.modtool.exist(['Core/Java/1.8']), [True])
-            self.assertTrue(self.modtool.module_wrapper_exists('Core/Java/1.8'))
+            self.assertEqual(self.modtool.module_wrapper_exists('Core/Java/1.8'), 'Core/Java/1.8.0_181')
```
vanzod: Same as above

test/framework/modules.py
boegel: Doesn't matter, this is purely in the unit tests for framework, it has no direct relation to the easyconfigs repository.

A bit higher up in the test I'm making a dummy `Java/1.8.0_181` module file (in a temporary directoty) just for the sake of the test.

So, this won't break if we update the `Java-1.8.eb` easyconfig file in the easyconfigs repo. :)

PR #2607, opened 2018-10-03
support checking of PR dependencies in --merge-pr (WIP)

migueldiascosta:
until github supports PR dependencies, something like this could help maintainers using `--merge-pr` (?)

not very comprehensive, but it should stop obvious cases

can (should?) be extended to support dependencies in the other repositories

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -969,6 +969,14 @@ def check_pr_eligible_to_merge(pr_data):
     else:
         res = not_eligible(msg_tmpl % 'no milestone found')
 
+    # check whether all mentioned easyconfig PR dependencies have been merged
+    msg_tmpl = "* PR dependencies: %s"
+    if not pr_data.get('unmerged_pr_deps', False):
+        print_msg(msg_tmpl % 'OK (no unmerged easyconfig PR dependencies found)', prefix=False)
+    else:
+        unmerged_pr_deps = ', '.join(['#' + pr_dep for pr_dep in pr_data['unmerged_pr_deps']])
+        res = not_eligible(msg_tmpl % 'FAILED (%s not yet merged)' % unmerged_pr_deps)
+
     return res
 
 
@@ -1019,6 +1027,21 @@ def merge_pr(pr):
         raise EasyBuildError("Failed to get reviews for PR #%d from %s/%s (status: %d %s)",
                              pr, pr_target_account, pr_target_repo, status, reviews_data)
     pr_data['reviews'] = reviews_data
+
+    # check for required easyconfig PRs mentioned in the body 
+    match = re.search('[Rr]equires ((?:~~)?#.*)', pr_data['body'])
+    if match:
+        pr_deps = re.findall('(?:#)(\d+)', match.group())
+        unmerged_pr_deps = []
+        for pr_dep in pr_deps:
+            pr_dep_url = lambda g: g.repos[pr_target_account][pr_target_repo].pulls[pr_dep]
+            status, pr_dep_data = github_api_get_request(pr_dep_url, github_user)
+            if status != HTTP_STATUS_OK:
+                print_msg("Failed to get data for PR #%d from %s/%s (status: %d %s)",
+                          pr_dep, pr_target_account, pr_target_repo, status, pr_dep_data)
+            if not pr_dep_data.get('merged', False):
+                unmerged_pr_deps.append(pr_dep)
+        pr_data['unmerged_pr_deps'] = unmerged_pr_deps
 
     force = build_option('force')
     dry_run = build_option('dry_run') or build_option('extended_dry_run')


boegel:
@migueldiascosta This should indeed also consider cross-repo dependencies, since this is actually the most commonly overlooked case where an easyconfig PR is merged before the PR that updates an easyblock gets merged.

Also, please define a new function to go and check whether a dependency PR is merged (as opposed to adding a chunk of code to `merge_pr`.

easybuild/tools/github.py
```diff
@@ -1020,6 +1048,20 @@ def merge_pr(pr):
                              pr, pr_target_account, pr_target_repo, status, reviews_data)
     pr_data['reviews'] = reviews_data
 
+    # check for required PRs mentioned in the body
+    match = re.search('[Rr]equires .*', pr_data['body'])
+    if match:
+        pr_deps = {}
+        pr_deps['framework'] = re.findall('(?:framework/pull/)(\d+)', match.group())
+        pr_deps['easyblocks'] = re.findall('(?:blocks/pull/)(\d+)', match.group())
+        pr_deps['easyconfigs'] = re.findall('(?: #|~~#)(\d+)', match.group())
```
boegel: @migueldiascosta One thing you seem to be overlooking here is that `--merge-pr` also works on framework and easyblock PRs (it follows `--pr-target-repo`).

So, if framework PR is being merged, then we should also take `#[0-9]+` into account for `framework`, etc...

Also, are you taking into account that there may be multiple `requires` entries? That's not entirely clear to me at first glance...

easybuild/tools/github.py
```diff
@@ -1020,6 +1048,20 @@ def merge_pr(pr):
                              pr, pr_target_account, pr_target_repo, status, reviews_data)
     pr_data['reviews'] = reviews_data
 
+    # check for required PRs mentioned in the body
+    match = re.search('[Rr]equires .*', pr_data['body'])
```
boegel: I would rename `match` to something more meaningful like `pr_requirements`?

easybuild/tools/github.py
```diff
@@ -969,9 +969,37 @@ def not_eligible(msg):
     else:
         res = not_eligible(msg_tmpl % 'no milestone found')
 
+    # check whether all mentioned easyconfig PR dependencies have been merged
+    msg_tmpl = "* PR dependencies: %s"
+    if not pr_data.get('unmerged_pr_deps', False):
+        print_msg(msg_tmpl % 'OK (no unmerged PR dependencies found)', prefix=False)
+    else:
+        unmerged_pr_deps = ', '.join(pr_data['unmerged_pr_deps'])
+        res = not_eligible(msg_tmpl % 'FAILED (%s not yet merged)' % unmerged_pr_deps)
```
boegel: @migueldiascosta I would reverse the logic here, seems easier to understand. Also, we should always include `'unmerged_pr_deps'` in `pr_data` (see also below)?

```python
if pr_data['unmerged_pr_deps']:
    unmerged_pr_deps = ', '.join(pr_data['unmerged_pr_deps'])
    res = not_eligible(msg_tmpl % 'FAILED (%s not yet merged)' % unmerged_pr_deps)
else:
    print_msg(msg_tmpl % 'OK (no unmerged PR dependencies found)', prefix=False)

File: easybuild/tools/github.py [hunk 4]
```diff
@@ -1020,6 +1048,20 @@ def merge_pr(pr):
...
+        pr_deps = {}
+        pr_deps['framework'] = re.findall('(?:framework/pull/)(\d+)', match.group())
+        pr_deps['easyblocks'] = re.findall('(?:blocks/pull/)(\d+)', match.group())
+        pr_deps['easyconfigs'] = re.findall('(?: #|~~#)(\d+)', match.group())
+        unmerged_pr_deps = []
+        for repo in pr_deps.keys():
+            for pr_dep in pr_deps[repo]:
+                if not is_pr_merged(pr_dep, pr_target_account, 'easybuild-%s' % repo, github_user):
+                    unmerged_pr_deps.append('easybuild-%s#%s' % (repo, pr_dep))
+        pr_data['unmerged_pr_deps'] = unmerged_pr_deps
```
boegel: @migueldiascosta Maybe it's better to flesh out this code block in a dedicated function that we can easily test in isolation (w.r.t. having multiple requirements, different repos, etc.)?

```python
def check_unmerged_pr_deps(pr_body):
    """Check for unmerged PR dependencies in specified PR body."""
    unmerged_pr_deps = []

    pr_requirements = re.search('[Rr]equires .*', pr_data['body'])
    if pr_requirements:
        ...

    return unmerged_pr_deps

def merge_pr(pr):
    ...
    pr_data['unmerged_pr_deps'] = check_unmerged_pr_deps(pr_data['body'])

easybuild/tools/github.py
migueldiascosta: @boegel indeed, I was only thinking of easyconfig PRs, will need to check the `pr-target-repo` and update de regexs accordingly

regarding multiple entries, it should find all (in the body) following the first `requires`, but I'll check

Issue #2609, opened 2018-10-03
Lmod version check: provide option/variable to override


fgeorgatos:
...this is mostly to avoid the need for patching, when Lmod version control is not that readily available. 
(put your +1, if this helps you)

boegel:
support for `--disable-modules-tool-version-check` is added in #2610

fgeorgatos:
@boegel : ty! reviewing now!

PR #2610, opened 2018-10-03
add support for disabling modules tool version check

boegel:
fix for #2609 

This add support for disabling the version check on the modules tool (e.g. `Lmod`) via `--disable-modules-tool-version-check` or `export EASYBUILD_DISABLE_MODULES_TOOL_VERSION_CHECK=1` (or similar via an EasyBuild configuration file)

cc @fgeorgatos

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -217,6 +217,7 @@ BUILD_OPTIONS_CMDLINE = {
         'cleanup_tmpdir',
         'extended_dry_run_ignore_errors',
         'mpi_tests',
+        'modules_tool_version_check',
     ],
     WARN: [
         'check_ebroot_env_vars',
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -239,21 +239,26 @@ class ModulesTool(object):
         if self.REQ_VERSION is None and self.MAX_VERSION is None:
             self.log.debug("No version requirement defined.")
 
-        if self.REQ_VERSION is not None:
-            self.log.debug("Required minimum version defined.")
-            if StrictVersion(self.version) < StrictVersion(self.REQ_VERSION):
-                raise EasyBuildError("EasyBuild requires %s >= v%s, found v%s",
-                                     self.__class__.__name__, self.REQ_VERSION, self.version)
-            else:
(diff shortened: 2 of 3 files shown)


easybuild/tools/modules.py
```diff
@@ -239,21 +239,26 @@ def set_and_check_version(self):
...
-                self.log.debug('Version %s matches requirement <= %s', self.version, self.MAX_VERSION)
+            if self.MAX_VERSION is not None:
+                self.log.debug("Maximum allowed version defined.")
+                if StrictVersion(self.version) > StrictVersion(self.MAX_VERSION):
+                    raise EasyBuildError("EasyBuild requires %s <= v%s, found v%s",
+                                         self.__class__.__name__, self.MAX_VERSION, self.version)
+                else:
+                    self.log.debug('Version %s matches requirement <= %s', self.version, self.MAX_VERSION)
+        else:
+            self.log.debug("Skipping check to see whether modules tool version '%s' meets requirements", self.version)
```
fgeorgatos: 'Skipping the check which sees...' ?

easybuild/tools/config.py
```diff
@@ -217,6 +217,7 @@ def mk_full_default_path(name, prefix=DEFAULT_PREFIX):
         'cleanup_tmpdir',
         'extended_dry_run_ignore_errors',
         'mpi_tests',
+        'modules_tool_version_check',
```
fgeorgatos: This is conforming naming btw!

I am saying it, because it would allow to configure things on my end, like:
```
EASYBUILD:
  BUILDPATH:                    /dev/shm/$USER
  MODULES_TOOL:                 Lmod
  MODULES_TOOL_VERSION_CHECK:  "Yes" ## If you dare not quote it, it becomes "True" via yaml parser!
  MODULE_SYNTAX:                Tcl
```

ref. https://github.com/Bright-Computing/bic/blob/master/test/etc/profile.definitions/site/EasyBuild.yml
(that gets picked up on node initialisation - working logic, but can be improved)

fgeorgatos:
lgtm, btw! 

i'd be glad to review/approve, but it seems i am out of rights about it.

fgeorgatos:
@akesandgren : (on behalf of more) thanks

PR #2611, opened 2018-10-04
add support for ModuleGenerator.modulerc method to also write .modulerc file + Lmod 6.x specific check

boegel:


easybuild/tools/module_generator.py
```diff
@@ -221,46 +221,87 @@ def _modulerc_check_module_version(self, module_version):
...
+    def _write_modulerc_file(self, modulerc_path, modulerc_txt, wrapped_mod_name=None):
         """
-        Generate contents of .modulerc file, in Tcl syntax (compatible with all module tools, incl. Lmod)
+        Write modulerc file with specified contents.
+
+        :param modulerc_path: location of .modulerc file to write
+        :param modulerc_txt: contents of .modulerc file
+        :param wrapped_mod_name: name of module file for which a wrapper is defined in the .modulerc file (if any)
+        """
+        if os.path.exists(modulerc_path) and not build_option('force'):
```
vanzod: Should we also allow overwriting the `.modulerc` in the case the user passes `--rebuild`?

boegel: Good point... Fixed in e1161b525!

boegel:
@damianam Exactly that was implemented in #2597, which is merged for including in EasyBuild v3.7.1 already... :)

damianam:
Minimum Lmod version `7.7.38` BTW

boegel:
@damianam Well, OK, if you feel strongly about getting the Lmod required version 100% right, please open a PR to fix that.

Issue #2615, opened 2018-10-12
problem with new toolchain `foss/2018.08` and `--try-toolchain` option


pescobar:
Hi 

I have installed the new toolchain `foss/2018.08` (still in the devel branch)
https://github.com/easybuilders/easybuild-easyconfigs/pull/6992

After installing the new toolchain the `try-toolchain` option doesn't work in combitnation with it e.g

This works fine using the common toolchains provided by easybuild:
`eb Python-2.7.14-foss-2018a.eb --try-toolchain-version=2018b --robot`

This crashes (the only change is version=2018b to version=2018.08):

```
eb Python-2.7.14-foss-2018a.eb --try-toolchain-version=2018.08 --robot
== temporary log file in case of crash /scratch/eb-aZ31sN/easybuild-a9TnOl.log
ERROR: Could not find easyconfig for foss toolchain version 2018.08
```

If I try to pass as argument to `--robot` the path of my local clone that includes the easyconfig for `foss/2018.08` then:

```
eb Python-2.7.14-foss-2018a.eb --try-toolchain-version=2018.08 --robot -r easyconfigs/easybuild/easyconfigs/
== temporary log file in case of crash /scratch/eb-I3rh9M/easybuild-eIbvaZ.log
ERROR: Target hierarchy {'lapack_family': 'OpenBLAS', 'name': 'foss', 'blas_family': 'OpenBLAS', 'version': '2018.08', 'cuda': None, 'comp_family': 'GCC', 'mpi_family': 'OpenMPI'} should have binutils using GCCcore, can't determine mapping!
```

@boegel suggested that this could be a problem with the new toolchain

@ocaisa any idea what could be the problem ?

I used latest easybuild 3.7.0 for the tests.

ocaisa:
I can't reproduce this with the current devel branches, works fine. I suspect it might be something in your local configuration that hasn't been taken into account, anything helpful in the debug logs?

Relevant error comes from [here](https://github.com/easybuilders/easybuild-framework/blob/26d6b8789049d25df93f95390fb6a049b7ad41be/easybuild/framework/easyconfig/tweak.py#L792). If you do a dry run on the toolchain easyconfig for `2018.08` does binutils appear in the result (we currently rely on this)?

ocaisa:
Does `easyconfigs/easybuild/easyconfigs/` include all of the easyconfigs required to build the toolchain (and in particular the required `binutils/2.31.1`)?

[...]

PR #2617, opened 2018-10-12
retain all dependencies when determining dependency tree of a toolchain

boegel:
fix for problem reported in #2615


ocaisa:
I admit a bit of laziness when creating the tests

Issue #2618, opened 2018-10-12
%(namelower)s fails to expand in Python setuptools download with --try-toolchain


zao:
Using the patch from #2617 to handle the problem in #2615, the following failure to download setuptools occurred when running `eb Python-2.7.15-foss-2018b.eb -r --try-toolchain=foss,2018.08 --debug`.

As this is the first time I ever run develop, I have no idea if it affects regular un-try builds, and if the patch had this effect.
```
ERROR: Build of /tmp/eb-xeEGWZ/tweaked_easyconfigs/Python-2.7.15-foss-2018.08.eb failed (err: "build failed (first 300 chars): Couldn't find file setuptools-39.2.0.zip anywhere, and downloading it didn't work either... Paths attempted (in order): /tmp/eb-xeEGWZ/tweaked_easyconfigs/p/Python/extensions/setuptools-39.2.0.zip, /tmp/eb-xeEGWZ/tweaked_easyconfigs/p/Python/packages/setuptools-39.2.0.zip, /tmp/eb-xeEGWZ/tweaked_eas")
```
```
== 2018-10-12 20:58:17,199 filetools.py:1212 DEBUG Not creating existing path /home/zao/.local/eb-develop/sources/p/Python
== 2018-10-12 20:58:17,199 easyblock.py:728 DEBUG Trying to download file setuptools-39.2.0.zip from https://pypi.%(namelower)s.org/packages/source/s/setuptools/setuptools-39.2.0.zip to /home/zao/.local/eb-develop/sources/p/Python/extensions/setuptools-39.2.0.zip ...
== 2018-10-12 20:58:17,199 filetools.py:484 DEBUG Trying to download setuptools-39.2.0.zip from https://pypi.%(namelower)s.org/packages/source/s/setuptools/setuptools-39.2.0.zip to /home/zao/.local/eb-develop/sources/p/Python/extensions/setuptools-39.2.0.zip
== 2018-10-12 20:58:17,199 filetools.py:491 DEBUG Using timeout of 10 seconds for initiating download
== 2018-10-12 20:58:17,199 filetools.py:1184 INFO Creating directory /home/zao/.local/eb-develop/sources/p/Python/extensions (parents: True, set_gid: False, sticky: False)
[...]

PR #2619, opened 2018-10-15
Protect exts_lists from templating in dump method...

ocaisa:
(since they are templated internally themselves)

### DIFF
--- a/easybuild/framework/easyconfig/format/format.py
+++ b/easybuild/framework/easyconfig/format/format.py
@@ -53,8 +53,8 @@ FORMAT_DEFAULT_VERSION = EasyVersion('1.0')
 DEPENDENCY_PARAMETERS = ['builddependencies', 'dependencies', 'hiddendependencies']
 
 # values for these keys will not be templated in dump()
-EXCLUDED_KEYS_REPLACE_TEMPLATES = ['description', 'easyblock', 'homepage', 'name', 'toolchain', 'version', 'exts_list'] \
-                                  + DEPENDENCY_PARAMETERS
+EXCLUDED_KEYS_REPLACE_TEMPLATES = ['description', 'easyblock', 'homepage', 'name', 'toolchain', 'version'] + \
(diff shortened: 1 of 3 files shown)


ocaisa:
Fixes #2618

ocaisa:
Hmm, the new test isn't broken before I add the fix, not sure what's going on there...but I confirm that it does fix the error described in #2618

ocaisa:
@boegel I cannot get this additional test to break, I can see the difference in the outputs with `--try-toolchain` but the templating is not as aggressive for some reason in the Python case I included

ocaisa:
Ok, if I look directly at ectxt in the test I can see the fix works, but this is not causing problems when parsing. I tried a few things to get the test to break but I am out of ideas.

ocaisa:
Ok, the test now fails without the fix so this is good to go. I don't understand why templating needs to be switched off in the parsed easyconfig but it at least catches the error and seems to be what happens in the main trunk. I would suspect it is because that there is nested templating in `exts_list` but maybe @boegel can explain.

easybuild/framework/easyconfig/format/format.py
```diff
@@ -54,7 +54,7 @@
 
 # values for these keys will not be templated in dump()
 EXCLUDED_KEYS_REPLACE_TEMPLATES = ['description', 'easyblock', 'homepage', 'name', 'toolchain', 'version'] \
-                                  + DEPENDENCY_PARAMETERS
+                                  + ['exts_list'] + DEPENDENCY_PARAMETERS
```
boegel: @ocaisa Let's keep this sorted alphabetically, it'll be way easier to tell whether a particular parameter is included or not that way upon visual inspection...

Done in https://github.com/ocaisa/easybuild-framework/pull/38

boegel:
@ocaisa To answer your question: I think we deliberately don't resolve things like `%(namelower)s` for extensions because of the ambiguity w.r.t. main `name` or extension name...

boegel:
Going in, thanks @ocaisa!

Issue #2620, opened 2018-10-15
CUDA capability not detected correctly with --try-toolchain


mboisson:
We noticed that the new `--try-toolchain` behavior breaks our MNS. More specifically, when specifying a cuda toolchain, it no longer installs under the CUDA directory. 

For example : 
```
$ eb OpenMPI-2.1.1-iccifortcuda-2016.4.eb --try-toolchain=iccifortcuda,2016.4.100
== temporary log file in case of crash /tmp/eb-_nmzG4/easybuild-_8drJC.log
== avx2/Compiler/intel2016.4/openmpi/2.1.1 is already installed (module found), skipping
== No easyconfigs left to be built.
== Build succeeded for 0 out of 0
== Temporary log file(s) /tmp/eb-_nmzG4/easybuild-_8drJC.log* have been removed.
== Temporary directory /tmp/eb-_nmzG4 has been removed.
``` 

Previously, it would have installed under `avx2/CUDA/intel2016.4/cuda10.0/openmpi/2.1.1`. Adding some debugging code into our MNS (namely, I added `print("ec: %s" % str(ec.toolchain.definition()))` in `det_module_subdir`, I see that `det_module_subdir` is not being called with consistent toolchain descriptions throughout : 
```
[mboisson@build-node OpenMPI]$ ~/myeb --debug OpenMPI-2.1.1-iccifortcuda-2016.4.eb --try-toolchain=iccifortcuda,2016.4.100
== temporary log file in case of crash /tmp/eb-gPWJ9O/easybuild-519I3R.log
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER_CUDA': ['CUDA'], 'COMPILER': ['icc', 'ifort', 'CUDA']}
ec: {'COMPILER_CUDA': ['CUDA'], 'COMPILER': ['icc', 'ifort', 'CUDA']}
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER': ['icc', 'ifort']}
ec: {'COMPILER': ['icc', 'ifort']}
[...]

PR #2621, opened 2018-10-15
making cuda capability detection more robust

mboisson:


mboisson:
This is a fix to issue https://github.com/easybuilders/easybuild-framework/issues/2620

ocaisa:
@mboisson Thanks for this fix, I added a specific test for this in https://github.com/ComputeCanada/easybuild-framework/pull/7 , once that is included then this is good to go.

mboisson:
@ocaisa done.

PR #2635, opened 2018-10-23
fixing bug that could cause to silently overwrite an existing easyconfig

mboisson:


mboisson:
Tagging @ocaisa

easybuild/framework/easyconfig/tweak.py
```diff
@@ -836,6 +837,12 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
     # Determine the name of the modified easyconfig and dump it to target_dir
     ec_filename = '%s-%s.eb' % (parsed_ec['ec']['name'], det_full_ec_version(parsed_ec['ec']))
     tweaked_spec = os.path.join(targetdir or tempfile.gettempdir(), ec_filename)
+    if os.path.exists(tweaked_spec):
+        if build_option('force'):
+            print_warning("Overwriting existing file at %s with tweaked easyconfig file (due to --force)", tweaked_spec)
+        else:
+            raise EasyBuildError("A file already exists at %s where tweaked easyconfig file would be written",
+                                 tweaked_spec)
```
boegel: @mboisson Shouldn't we do this in `dump` itself instead? We may be using `dump` is other places in the future...

boegel:
@mboisson Also, we should get this covered in the tests, i.e. have a test that verifies that `dump` will not blindly overwrite an existing easyconfig file, and that you can use `--force` to convince it to do so...

easybuild/framework/easyconfig/tweak.py
boegel: In this case I really think doing it in `dump` itself makes more sense...

I'm not sure why there are two code paths where one uses `write_file` and another uses `dump`, maybe @ocaisa knows?

ocaisa: There are two paths because one uses regex substitution and the other modifies the EasyConfig object and dumps the modified object.

Ultimately the dump method does use `write_file` so perhaps the question is whether we want to pass down options, the choices currently are:
```
def write_file(path, txt, append=False, forced=False, backup=False):
    """
    Write given contents to file at given path;
    overwrites current file contents without backup by default!

    :param path: location of file
    :param txt: contents to write to file
    :param append: append to existing file rather than overwrite
    :param forced: force actually writing file in (extended) dry run mode
    :param backup: back up existing file before overwriting or modifying it
    """
```

ocaisa: Perhaps doing a backup is sufficient? That way nothing ever gets removed and no need for the force option

ocaisa: Would just need to add `backup=True` to https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/framework/easyconfig/easyconfig.py#L800 to switch this on by default (and can't think of a good reason you wouldn't want to)

mboisson: I think write_file should at the very least issue a warning if not fail if it is going to be over-writing an existing file.

boegel: If a backup is created, it will print a warning message for that.

But I agree with @mboisson here, it shouldn't overwrite existing files without `--force` (and it should still create a backup if `--force` is used).

boegel: @mboisson W.r.t. making `write_file` fail hard rather than overwriting an existing file: we can't do that without breaking backward compatibility, changing that would likely break a lot of easyblocks. Also, always having to use force to make `write_file` actually overwrite existing file would be a bit much, that's not standard semantics in Linux either...

ocaisa: So, how do we proceed? Are you saying that the original check that @mboisson had is appropriate then and maybe couple that with adding the backup kw in the dump?

boegel: I took a look at it, the best way is to enhance `write_file` a bit to optionally not blindly overwrite existing files, and then use `write_file` correctly in both places.

Done in https://github.com/ComputeCanada/easybuild-framework/pull/8

boegel:
@ocaisa Thoughts on current approach?

easybuild/framework/easyconfig/tweak.py
```diff
@@ -357,14 +358,7 @@ def __repr__(self):
...
     # write out tweaked easyconfig file
-    if os.path.exists(tweaked_ec):
-        if build_option('force'):
-            print_warning("Overwriting existing file at %s with tweaked easyconfig file (due to --force)", tweaked_ec)
-        else:
-            raise EasyBuildError("A file already exists at %s where tweaked easyconfig file would be written",
-                                 tweaked_ec)
-
-    write_file(tweaked_ec, ectxt)
+    write_file(tweaked_ec, ectxt, backup=True, overwrite=False, verbose=True)
```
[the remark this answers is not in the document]
ocaisa: Ah ok, I understand it now but the kw name is not very helpful, how about `unforced_overwrite`?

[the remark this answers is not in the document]
boegel: `overwrite=False` means "don't overwrite unless `--force` is used"

Without also adding `backup=True`, you may overwrite when using `--force` without backing up the file.

Whether you want a backup or not depends on in what context you're calling `write_file`.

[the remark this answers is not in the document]
ocaisa: Ok, but the kw name needs tweaking, it's not unreasonable to expect `overwrite=False` not to overwrite.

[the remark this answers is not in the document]
boegel: I had `overwrite_without_force` before, which I found too long, but would that be better?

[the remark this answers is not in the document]
mboisson: how about `always_overwrite` ?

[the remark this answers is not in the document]
ocaisa: `overwrite_without_force` is long but since the option is so connected to `--force` I think it is worth the keystrokes, it will save people having to read the docstring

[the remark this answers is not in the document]
boegel: @mboisson `always_overwrite` makes sense to me. Can you make that change?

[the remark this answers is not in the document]
mboisson: I'm kind of lost in all of the changes you PR'd to us now :/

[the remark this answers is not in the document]
mboisson: I *think* I did it.

test/framework/filetools.py
```diff
@@ -547,6 +547,34 @@ def test_read_write_file(self):
...
+
+        # by default, write_file will just blindly overwrite an already existing file
+        self.assertTrue(os.path.exists(fp))
+        ft.write_file(fp, 'blah')
+        self.assertEqual(ft.read_file(fp), 'blah')
+
+        # blind overwriting can be disabled via 'overwrite'
+        error_pattern = "File exists, not overwriting it without --force: %s" % fp
+        self.assertErrorRegex(EasyBuildError, error_pattern, ft.write_file, fp, 'blah', overwrite=False)
+        self.assertErrorRegex(EasyBuildError, error_pattern, ft.write_file, fp, 'blah', overwrite=False, backup=True)
```
boegel: @mboisson You overlooked renaming `overwrite` here too... Fixed in https://github.com/ComputeCanada/easybuild-framework/pull/10

mboisson: merged

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -790,9 +790,12 @@ def all_dependencies(self):
 
         return self._all_dependencies
 
-    def dump(self, fp):
+    def dump(self, fp, always_overwrite=True, backup=False):
         """
         Dump this easyconfig to file, with the given filename.
+
+        :param overwrite: overwrite existing file at specified location without use of --force
```
boegel: @mboisson Please also fix the '`overwrite`' in the docstring to `'always_overwrite'`

mboisson: done

boegel:
Good to go, thanks a lot for looking into this @mboisson, and to @ocaisa for the feedback!

PR #2638, opened 2018-10-24
flesh out setting up of configuration into dedicated 'set_up_configuration' function

boegel:
This is basically just shuffling code around, no functional changes are included.

Having a dedicated function for setting up the EasyBuild configuration is useful to support using the EasyBuild framework as a library (see also the tests that have been added).

Before these changes, you'd run into problems like:

```
$ python -c "from easybuild.tools.run import run_cmd; run_cmd('echo hello')"
  File "/Volumes/work/easybuild-framework/easybuild/tools/run.py", line 87, in cache_aware_func
  File "/Volumes/work/easybuild-framework/easybuild/tools/run.py", line 127, in run_cmd
    if log_output or (trace and build_option('trace')):
  File "/Volumes/work/easybuild-framework/easybuild/tools/config.py", line 448, in build_option
    raise EasyBuildError("Undefined build option: %s", key)
easybuild.tools.build_log.EasyBuildError: 'Undefined build option: trace'
```

When `set_up_configuration` is used, things work as expected:

```
$ python -c "from easybuild.tools.options import set_up_configuration; 
set_up_configuration(silent=True); from easybuild.tools.run import run_cmd;
print run_cmd('echo hello')"
('hello\n', 0)
```

Also, a clearer error is now produced in case `set_up_configuration` was not called first:

```
easybuild.tools.build_log.EasyBuildError: "Undefined build option: 'trace'.
Make sure you have set up the EasyBuild configuration using set_up_configuration()
(from easybuild.tools.options) in case you're not using EasyBuild via the 'eb' CLI."
```

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -190,7 +190,7 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     # purposely session state very early, to avoid modules loaded by EasyBuild meddling in
     init_session_state = session_state()
 
-    eb_go, cfg_settings = set_up_configuration(args)
+    eb_go, cfg_settings = set_up_configuration(args=args, logfile=logfile, testing=testing)
     options, orig_paths = eb_go.options, eb_go.args
 
     global _log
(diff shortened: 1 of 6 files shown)


bartoldeman:
Going in, thanks @boegel !

boegel:
@bartoldeman My local linter should have noticed if I did miss any obsolete imports ;)

Issue #2624, opened 2018-10-17
Modules: Set CMAKE_PREFIX_PATH


ax3l:
Hi,

I am a user of several HPC systems that use Easybuild as a package manager (CSCS, TU Dresden). As a power-user and heavy developer, we use package software from modules on HPC systems to develop and build further software - outside of those package managers.

In the case of our software stack, we heavily rely on CMake and for such we would like to set the `CMAKE_PREFIX_PATH` when loading modules automatically to the prefix of installed software. The CMake build system heavily relies on those to find software of any kind (and prefer it over e.g. system paths).

How can one generally *configure* Easybuild to *always* prepend `CMAKE_PREFIX_PATH` in modules?
Should one introduce a more general templating mechanism for modules outside of [`--modules-footer`](https://easybuild.readthedocs.io/en/latest/version-specific/help.html)?

As a heavy [Spack](http://spack.io/) user myself, I am quite in love with the module configuration I can use there and therefore I am quite biased :) Maybe that's useful for Easybuild as well?
https://spack.readthedocs.io/en/latest/module_file_support.html#write-a-configuration-file

ocaisa:
Hmm, I'd say people probably have some opinions on this. If you want a quick fix for now, you can just leverage the `$EBROOT*` variables that are set for every EB module:
```
export CMAKE_PREFIX_PATH=$(env | grep EBROOT | sed  s/=/\ /|awk '{print $2}'|tr '\n' ':')$CMAKE_PREFIX_PATH
```

ocaisa:
Having configurable modules has been discussed before but noone has gotten around to an implementation.

boegel:
The idea of having a default module file template that you can customize has indeed been discussed before.

It shouldn't be terribly hard to implement either, I think it's mostly because of a lack of time and the fact that nobody has had a strong use case for it yet...

To make it concrete: someone implementing this would have to dive into `easybuild/framework/easyblock.py` and make `make_module_step` use a template of a module file rather than concatenating a set of strings obtained by calling the different `make_module*` methods.

boegel:
First stab at adding support for using a custom module file template in #2639 .

For @ax3l's use case, `--modules-footer` is still a good option though:

```
$ cat $HOME/my-modules-footer.lua
prepend_path("CMAKE_PREFIX_PATH", "%(installdir)s")
```

```
$ eb --modules-footer $HOME/my-modules-footer.lua ...

# or

[...]

PR #2639, opened 2018-10-24
add support for using custom module file template (WIP)

boegel:
Adds support for specifying a custom module file template via `--modules-template`, as well as for dumping the default modules template:

```
$ eb --show-default-modules-template
$shebang$header$descr$group_check$deps$modpath$env_vars$footer
```

This should meet the use case of @ax3l discussed in #2624, since it allows using a custom template like:

```
$shebang$header$descr$group_check$deps$modpath$env_vars$footer
prepend_path("CMAKE_PREFIX_PATH", "%(installdir)s")
```

In some sense this is already supported via `--modules-footer` though, so the added benefit is limited for that particular use case, but it does provide more control over the generated module files...

(WIP since tests are missing)

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2438,20 +2438,25 @@ class EasyBlock(object):
         else:
             trace_msg("generating module file @ %s" % self.mod_filepath)
 
-        txt = self.module_generator.MODULE_SHEBANG
-        if txt:
-            txt += '\n'
-
+        mod_header = ''
         if self.modules_header:
-            txt += self.modules_header + '\n'
+            mod_header = self.modules_header + '\n'
 
-        txt += self.make_module_description()
-        txt += self.make_module_group_check()
-        txt += self.make_module_dep()
-        txt += self.make_module_extend_modpath()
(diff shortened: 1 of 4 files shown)


ax3l:
Thank you for the great feedback!
I can imagine several cases where one wants to add vars such as `MANPATH`, additional [program-specific helper vars](https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.4.0/INSTALL.rst#picongpu-source-code), etc.

As a background and comparison to [spack](https://spack.readthedocs.io/en/latest/module_file_support.html), one can export them on a per-package basis and a general filter (all, all depending on X, etc.)

boegel:
@ax3l So, rather than being able to provide a single fixed module file template that is used for everything, you'd like this be (potentially) be conditional, such that a different template is used in different situations?

That's way beyond what I had in mind for this... I guess this can be a first step.

This could be extended later to support that the `--modules-template` value points to a Python module/script that *generates* a template based on the availabe inputs... Is that what you're looking for?

akesandgren:
Considering that you can do basically anything you want with the easyhooks, is this still relevant?

Issue #2632, opened 2018-10-22
--job fails with 3.7.1 & gc3pie 2.5.0


jpecar:
We had our CI pipeline running fine all the way & including EasyBuild 3.6.2 & GC3Pie 2.4.2. Now I've upgraded to 3.7.1 and GC3Pie 2.5.0 and jobs are created by GC3Pie but fail immediately with "eb: command not found", causing gc3pie to fail with " gc3libs.exceptions.LRMSError: Could not retrieve status information for task Application@3aa5250". 

It appears that something changed in the way how environment is propagated down from eb that generates a job to a job itself. There are no changes that would mention anything like this in either easybuild or gc3pie changelogs. 

For now I've hardcoded our CI to use previous versions of EasyBuild and GC3Pie but would eventually like to see this resolved.

boegel:
Instinct tells me this is most likely a problem with GC3Pie, but I'll let @riccardomurri be the judge on this... Is GC3Pie somehow resetting the environment?

@jpecar Can you provide some more information on how EasyBuild was installed? Are you loading a module file for EasyBuild, or was it installed via `easy_install` or `pip`, etc.?

jpecar:
It's a module file.

riccardomurri:
GC3Pie does not reset the environment; all processes are run in the environment inherited from the parent process.  What has changed in rel 2.5.0 compared to rel 2.4.2 is that sub-processes are now executed directly without calling a shell (i.e., we use Python's `subprocess.Popen(..., shell=False)` instead of `subprocess.Popen(..., shell=True)`.  This might have consequences if your shell's startup files were used to set the environment/PATH where to find the `eb` executable.

Environmental variables can be overridden, but that has to be specified when you create the GC3Pie `Task` object; in EB's code, this is the argument `env_vars` to EB's `make_job()` method. As far as I can see, the only place where `make_job()` is called is in EB's `parallelbuild.py` where the `env_vars` parameter is passed a dictionary comprising all `EASYBUILD_*` variables plus `PYTHONPATH` and `MODULEPATH`.  Maybe what happens is that the previous version of EB included more stuff in the environment?  Or did *not* reset the environment *before* calling into GC3Pie?

I guess it would help in your case if https://github.com/uzh/gc3pie/issues/609 was solved?

boegel:
@riccardomurri: It seems like we should also be explicitly passing `PATH` via `env_vars`? That would probably fix the issue @jpecar is seeing?

[...]

PR #2642, opened 2018-10-29
add support for using Slurm as backend for --job

boegel:
With this in place, you configure EasyBuild to submit jobs to a SLURM cluster (using `--job-backend=PySlurm`), as long as a sufficiently recent version of `PySlurm` is installed (https://pypi.org/project/pyslurm/).


easybuild/tools/job/pyslurm_backend.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @only_if_module_is_available('pyslurm', pkgname='pyslurm')
+    def __init__(self, *args, **kwargs):
+        """Constructor."""
+        super(PySlurm, self).__init__(*args, **kwargs)
+
+    # _check_version is called by __init__, so guard it (too) with the decorator
+    @only_if_module_is_available('pyslurm', pkgname='pyslurm')
+    def _check_version(self):
+        """Check whether pbs_python version complies with required version."""
```
akesandgren: pbs_python??

easybuild/tools/job/pyslurm_backend.py
```diff
@@ -0,0 +1,146 @@
...
+        version = pyslurm_version()
+        if LooseVersion(version) < LooseVersion(self.REQ_VERSION):
+            raise EasyBuildError("Found pyslurm version %s, but version %s or more recent is required",
+                                 version, self.REQ_VERSION)
+
+    def init(self):
+        """
+        Initialise the job backend.
+
+        Connect to the PBS server & reset list of submitted jobs.
```
akesandgren: Not PBS, Slurm

easybuild/tools/job/pyslurm_backend.py
```diff
@@ -0,0 +1,146 @@
...
+        self._submitted = []
+
+    def queue(self, job, dependencies=frozenset()):
+        """
+        Add a job to the queue.
+
+        :param dependencies: jobs on which this job depends.
+        """
+        if dependencies:
+            job.job_specs['dependency'] = 'afterany:%s' % ':'.join(str(d.jobid) for d in dependencies)
```
akesandgren: Should probably be "afterok", since it's likely useless to continue if the dependency failed.

easybuild/tools/job/pyslurm_backend.py
```diff
@@ -0,0 +1,146 @@
...
+            hours = max_walltime
+        if hours > max_walltime:
+            self.log.warn("Specified %s hours, but this is impossible. (resetting to %s hours)" % (hours, max_walltime))
+            hours = max_walltime
+        self.job_specs['time'] = '%s:0:0' % hours
+
+        if cores:
+            self.job_specs['nodes'] = '1'
+            # value passed here must be an integer!
+            self.job_specs['ntasks_per_node'] = cores
```
akesandgren: And it must also be >0, should perhaps check user input here and fail nicely if it's bogus.

easybuild/tools/job/pyslurm_backend.py
boegel: We also use `afterany` with `pbs_pyton`, so I prefer keeping this consistent...

It's true that the dependent jobs will fail, but at least it'll be clear why (i.e. a missing module), while if they just disappear into thin air, it may not be clear at all...

easybuild/tools/job/pyslurm_backend.py
boegel: This is already done in `create_job` in `tools/parallelbuild.py`, where we only pass down `cores` if it's non-zero.

easybuild/tools/job/pyslurm_backend.py
akesandgren: Maybe so, but using afterany consumes resources when it shouldn't. And if the preceding dependency failed you wouldn't care about the next step.

easybuild/tools/job/pyslurm_backend.py
akesandgren: Ok.

easybuild/tools/job/pyslurm_backend.py
boegel: What about adding a configuration option for this? Something like `--job-deps-type=afterok`?

akesandgren: Yeah, that would work. And it would be backend agnostic if done correctly. (whatever that is in this case)

boegel:
@akesandgren Via `--job-deps-type`, you can now specify what type of dependency must be used.

The default depends on the job backend (to retain backward compatibility): `--job-deps-type=always_run` is the default for `PbsPython` (equivalent with using `afterany` deps), while `--jobs-deps-type=abort_on_error` is the default for both `GC3Pie` and the new `PySlurm` job backend.

boegel:
@akesandgren I hope to add some tests for the `PySlurm` backend as well...

Any chance that you can give this a spin yourself to see if `--job-backend --job` works for you?

akesandgren:
create_job in parallelbuild sets up easybuild_vars to contain only
PYTHONPATH, MODULEPATH and all variables starting with EASYBUILD
That won't fly... it should not touch the environment at all really (in my opinion at least).
One needs to be able to give it all sorts of env, like SBATCH_ACCOUNT, and a bunch of other slurm realted things, and other EB unrelated things.

boegel:
@akesandgren That's orthogonal to this PR I think, see the discussion in #2632

akesandgren:
There is something wrong.
--job-max-walltime=3 doesn't get passed down into pyslurm in the way it wants.
I get this back (when bypassing the account problem)
python2: error: must supply a time (-t ..)

easybuild/tools/job/slurm.py
```diff
@@ -0,0 +1,166 @@
...
+        if job_deps_type == JOB_DEPS_TYPE_ABORT_ON_ERROR:
+            self.job_deps_type = 'afterok'
+        elif job_deps_type == JOB_DEPS_TYPE_ALWAYS_RUN:
+            self.job_deps_type = 'afterany'
+        else:
+            raise EasyBuildError("Unknown job dependency type specified: %s", job_deps_type)
+
+    def _check_version(self):
+        """Check whether version of Slurm complies with required version."""
+        (out, _) = run_cmd("sbatch --version", trace=False)
```
akesandgren: This need a check on the return value too, in case sbatch doesn't exist in the PATH with a nice user visible error?

boegel: `run_cmd` does that by default, if the exit code is non-zero it'll result is a clear error message

I can also check with `which('sbatch')` too though, to make sure the required commands are available...

akesandgren:
There needs to be a config option for choosing what to put in "#!/bin/sh" of the submit script. We require /bin/bash for things to work.

Not sure that is the problem i ran into now or not though.

easybuild/tools/job/slurm.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def queue(self, job, dependencies=frozenset()):
+        """
+        Add a job to the queue.
+
+        :param dependencies: jobs on which this job depends.
+        """
+        if dependencies:
+            job.job_specs['dependency'] = self.job_deps_type + ':' + ':'.join(str(d.jobid) for d in dependencies)
+
```
akesandgren: For dependency jobs I think one should enforce --kill-on-invalid-dep=yes
This to avoid jobs with a dependency staying in the queue with a DependencyNeverSatisfied state.
You won't get a return result from the thus canceled jobs but they do vanish from the squeue list.

akesandgren:
With my latest PR to this one I can get stuff running.
(Although it failed since i was doing --from-pr and the submitted jobs of course didn't have my OAuth token ...)

Will try something local later this evening.

easybuild/tools/job/slurm.py
boegel: done in 0daf74b

easybuild/tools/job/slurm.py
```diff
@@ -0,0 +1,177 @@
...
+    def complete(self):
+        """
+        Complete a bulk job submission.
+
+        Release all user holds on submitted jobs, and disconnect from server.
+        """
+        for job in self._submitted:
+            if job.job_specs['hold']:
+                self.log.info("releasing user hold on job %s" % job.jobid)
+                run_cmd("scontrol release %s" % job.jobid, trace=False)
```
akesandgren: scontrol release can take multiple jobids, line them up and call run_cmd once instead.

test/framework/parallelbuild.py
```diff
@@ -63,6 +63,21 @@
...
 
+
+MOCKED_SBATCH = """#!/bin/bash
+if [[ $1 == '--version' ]]; then
+    echo "slurm 17.0"
+else
+    echo "Submitted batch job $RANDOM"
+    echo "(submission args: $@)"
+fi
+"""
```
akesandgren: It would probably be safer to use bourne-shell syntax and /bin/sh here. The code is simple enough not to need bash.

And same goes for the mocked scontrol

test/framework/parallelbuild.py
```diff
@@ -196,7 +211,7 @@ def test_build_easyconfigs_in_parallel_pbs_python(self):
     def test_build_easyconfigs_in_parallel_gc3pie(self):
         """Test build_easyconfigs_in_parallel(), using GC3Pie with local config as backend for --job."""
         try:
-            import gc3libs
+            import gc3libs  # noqa (ignore unused import)
         except ImportError:
             print "GC3Pie not available, skipping test"
             return
```
akesandgren: Is GC3Pie version verified in here? If not, should it be done?

test/framework/parallelbuild.py
boegel: Doesn't matter much in the tests? As long as Travis has `/bin/bash`, we're good?

test/framework/parallelbuild.py
boegel: That's done in the `gc3pie.py` implementation.

easybuild/tools/job/slurm.py
boegel: done

test/framework/parallelbuild.py
akesandgren: Yeah, was just a thought.

test/framework/parallelbuild.py
akesandgren: okidoki.

akesandgren:
@boegel Anything else you plan on doing with this?

boegel:
@akesandgren @migueldiascosta is testing this too, doesn't hurt to get some more feedback on it.

But it's good to go imho.

boegel:
@akesandgren docs update @ https://github.com/easybuilders/easybuild/pull/467

PR #2644, opened 2018-11-01
always skip symlinks in adjust_permissions

boegel:
This was pointed out by @akesandgren.

Using `os.lchmod` & `os.lchown` prevents EasyBuild from trying to change permissions on files it has no business in changing, in case a symlink points to a file outside of the build/install directory...

It probably also helps in avoiding problems like https://github.com/easybuilders/easybuild-easyblocks/issues/1564

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -956,7 +956,7 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None):
     apatch_root, apatch_file = os.path.split(apatch)
     apatch_name, apatch_extension = os.path.splitext(apatch_file)
     # Supports only bz2, gz and xz. zip can be archives which are not supported.
-    if apatch_extension in ['.gz','.bz2','.xz']:
+    if apatch_extension in ['.gz', '.bz2', '.xz']:
         # split again to get the second extension
         apatch_subname, apatch_subextension = os.path.splitext(apatch_name)
         if apatch_subextension == ".patch":
@@ -1095,7 +1095,7 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
             if relative:
 
                 # relative permissions (add or remove)
-                perms = os.stat(path)[stat.ST_MODE]
+                perms = os.lstat(path)[stat.ST_MODE]
 
                 if add:
                     os.lchmod(path, perms | permissionBits)
@@ -1108,7 +1108,7 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
 
             if group_id:
                 # only change the group id if it the current gid is different from what we want
-                cur_gid = os.stat(path).st_gid
+                cur_gid = os.lstat(path).st_gid
                 if not cur_gid == group_id:
                     _log.debug("Changing group id of %s to %s" % (path, group_id))
                     os.lchown(path, -1, group_id)
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1062,7 +1062,7 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
 
     if skip_symlinks is not None:
         depr_msg = "Use of 'skip_symlinks' argument for 'adjust_permissions' is deprecated "
-        depr_msg += "(symlinks are always skipped now)"
+        depr_msg += "(symlinks are never followed anymore)"
         _log.deprecated(depr_msg, '4.0')
 
     name = os.path.abspath(name)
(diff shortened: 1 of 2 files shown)


easybuild/tools/filetools.py
```diff
@@ -1098,20 +1098,20 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
                 perms = os.stat(path)[stat.ST_MODE]
```
akesandgren: This needs to be os.lstat to match the os.lchmod below

akesandgren:
Have you had time to test this besides the test in the framework/filetools ?

akesandgren:
Hmmm, hasattr(os, 'lchmod') returns False on my Ubuntu 16.04 Python 2.7.12 system.
In fact it seems to be a BSD only function.
And thus chmod should not be performed on symlinks, on Linux it is useless and wrong.

lchown, lstat however should still be used.

symlinks should just be ignored for chmod.

boegel:
@akesandgren Symbolic links are already skipped by default by `adjust_permissions`...

So should we be changing anything?

akesandgren:
Would have to look at the code properly again.... I'll get back on it.

easybuild/tools/filetools.py
```diff
@@ -1095,7 +1095,7 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
             if relative:
```
akesandgren: I would suggest a
if not os.path.islink(path):
before the if relative:
covering the whole if-else of the relative:
Since if skip_symlinks is False the chmod should not be done at all.

boegel: But that would make `skip_symlinks` effectively meaningless, since `adjust_permissions` would then always ignore symlinks?

akesandgren: No, it would still change group, but never touch symlinks, and for security reason one should never do chmod on a symlink, unless it's been verified that it stays within the intended tree.
(And that it haven't moved between checking that and the actual chmod, which is basically impossible)

akesandgren: But yes, ignoring symlinks is what you should always do when it comes to chmod, and for owner/group it really doesn't matter that much...

boegel:
@akesandgren deprecation of the now useless `skip_symlinks` is documented in https://github.com/easybuilders/easybuild/pull/466

easybuild/tools/filetools.py
```diff
@@ -1054,12 +1054,17 @@ def convert_name(name, upper=False):
...
-                       group_id=None, relative=True, ignore_errors=False, skip_symlinks=True):
+                       group_id=None, relative=True, ignore_errors=False, skip_symlinks=None):
     """
     Add or remove (if add is False) permissionBits from all files (if onlydirs is False)
     and directories (if onlyfiles is False) in path
     """
 
+    if skip_symlinks is not None:
+        depr_msg = "Use of 'skip_symlinks' argument for 'adjust_permissions' is deprecated "
+        depr_msg += "(symlinks are always skipped now)"
```
akesandgren: They aren't as much skipped as never followed, skipping them would involve not doing chgrp on the actual symlink, which is done through the lstat, lchown combo.

test/framework/filetools.py
```diff
@@ -961,20 +957,19 @@ def test_adjust_permissions_max_fail_ratio(self):
...
-        # one file remove, 1 dir + 2 files + 3 symlinks (of which 1 broken) left => 1/6 (16%) fail ratio is OK
+        ft.adjust_permissions(testdir, perms, recursive=True, ignore_errors=True)
+
+        # introducing a broken symlinks doesn't cause problems
         ft.remove_file(test_files[0])
-        ft.adjust_permissions(testdir, perms, recursive=True, skip_symlinks=False, ignore_errors=True)
-        # 2 files removed, 1 dir + 1 file + 3 symlinks (of which 2 broken) left => 2/5 (40%) fail ratio is OK
+        ft.adjust_permissions(testdir, perms, recursive=True, ignore_errors=True)
+
+        # multiple/all broken symlinks is no problem either, since symlinks are always skipped
```
akesandgren: "always skipped" -> "never followed" or something similar.

easybuild/tools/filetools.py
boegel: OK, reworded in 9f5c584

akesandgren:
should we test this on all easyblocks that uses it? Worst case is that some block uses it with "follow symlinks" to use that sideeffect. And would then need two invocations instead.

boegel:
@akesandgren There are no easyblocks that use `skip_symlinks=False`, and the current behavior matches the default behavior of `skip_symlinks=True`, so this can't affect easyblocks that use `adjust_permissions`?

In any case, this will be tested thoroughly during the next regression test for EasyBuild v3.8.0...

akesandgren:
Ah yes, monday morning...

PR #2648, opened 2018-11-02
update template constants to use https

orbsmiv:
I've updated all of the URLs in the template constants, hardcoding https. Whilst the majority of the sites included will automatically redirect to https there are a couple which don't (e.g. `GNU_SOURCE `) so for consistency, clarity and security I've modified all of them.
Additionally I've removed the `GOOGLECODE_SOURCE` template – there are only a handful of (older) EB configs which still use this and these will fail anyway as they'll result in a 404 error.


boegel:
@orbsmiv Please don't remove the `GOOGLECODE_SOURCE` template for now. I understand your motivation for doing so, but removing it will break any easyconfig file using it (i.e. they won't even parse anymore).

We can deprecate the use of `GOOGLECODE_SOURCE` though, and then drop this template in EasyBuild v4.0, but of course not before cleaning up the easyconfig files that still use it first.

zao:
Will this impose any additional constraint on the OS and Python used to provide up-to-date CA certs and having a working Python SSL impl, or is that already sufficiently required implicitly?

boegel:
@zao That's already required in other ways imho, most `github.com` & PyPI URLs we use are already `https` for example...

orbsmiv:
@boegel understood re the `GOOGLECODE_SOURCE` template (I suspected there would be a good reason to keep it in) so I've put it back.

boegel:
@orbsmiv Broken test is fixed with https://github.com/bear-software/easybuild-framework/pull/1

boegel:
other broken test also fixed in https://github.com/bear-software/easybuild-framework/pull/2 (which makes all tests pass, see https://travis-ci.org/boegel/easybuild-framework/builds/451277040)

boegel:
Good to go in, thanks a lot for your contribution @orbsmiv!

PR #2649, opened 2018-11-03
fix lib64 fallback for 'lib'/'lib64' dirs entry in sanity_check_paths

boegel:
fix for bug that popped up in https://github.com/easybuilders/easybuild-easyconfigs/pull/7069

For `'lib'` or `'lib64'` entries in `dirs` in `sanity_check_paths`, the fallback mechanism failed because it checks for entries starting with `lib/` or `lib64/` (to avoid that partial matches on something like `'library'` would trigger the fallback too).

### DIFF
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -1667,7 +1667,7 @@ class ToyBuildTest(EnhancedTestCase):
         # sanity check passes when lib64 fallback is enabled (by default), since lib/libtoy.a is also considered
         self.test_toy_build(ec_file=test_ec, raise_error=True)
 
-        # also check with 'lib' in sanity check dirs (special case)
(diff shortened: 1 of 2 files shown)


test/framework/toy_build.py
```diff
@@ -1656,6 +1667,17 @@ def test_sanity_check_paths_lib64(self):
         # sanity check passes when lib64 fallback is enabled (by default), since lib/libtoy.a is also considered
         self.test_toy_build(ec_file=test_ec, raise_error=True)
 
+        # also check with 'lib' in sanity check dirs (special case)
```
akesandgren: lib -> lib64 in the comment

boegel: fixed

PR #2653, opened 2018-11-05
Archive easyblocks along with easyconfig

### LINKED ISSUE #2633: Add ability to archive easyblocks
Based on a question that came up at the workshop in Birmingham 2018-10-22.

ocaisa:
Fixes #2633

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2900,7 +2900,7 @@ def build_and_install_one(ecdict, init_env):
             except NotImplementedError as err:
                 _log.warn("Unable to dumped fully processed easyconfig to %s: %s", reprod_spec, err)
             # also archive the relevant easyblocks
-            reprod_spec = os.path.join(reprod_spec, 'easyblocks')
+            reprod_spec = os.path.join(new_log_dir, 'reprod', 'easyblocks')
             for easyblock_class in inspect.getmro(app_class):
                 easyblock_path = inspect.getsourcefile(easyblock_class)
                 easyblock_splitpath = os.path.split(easyblock_path)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -72,10 +72,10 @@ from easybuild.tools.config import install_path, log_path, package_path, source_
 from easybuild.tools.environment import restore_env, sanitize_env
 from easybuild.tools.filetools import CHECKSUM_TYPE_MD5, CHECKSUM_TYPE_SHA256
 from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file, change_dir, convert_name
-from easybuild.tools.filetools import compute_checksum, copy_dir, copy_file, derive_alt_pypi_url, diff_files
-from easybuild.tools.filetools import download_file, encode_class_name, extract_file, get_source_tarball_from_git
-from easybuild.tools.filetools import is_alt_pypi_url, is_sha256_checksum, mkdir, move_logs, read_file, remove_file
-from easybuild.tools.filetools import rmtree2, verify_checksum, weld_paths, write_file
+from easybuild.tools.filetools import compute_checksum, copy_file, derive_alt_pypi_url, diff_files, download_file
+from easybuild.tools.filetools import encode_class_name, extract_file, get_source_tarball_from_git, is_alt_pypi_url
+from easybuild.tools.filetools import is_sha256_checksum, mkdir, move_logs, read_file, remove_file, rmtree2
+from easybuild.tools.filetools import verify_checksum, weld_paths, write_file
 from easybuild.tools.hooks import BUILD_STEP, CLEANUP_STEP, CONFIGURE_STEP, EXTENSIONS_STEP, FETCH_STEP, INSTALL_STEP
 from easybuild.tools.hooks import MODULE_STEP, PACKAGE_STEP, PATCH_STEP, PERMISSIONS_STEP, POSTPROC_STEP, PREPARE_STEP
 from easybuild.tools.hooks import READY_STEP, SANITYCHECK_STEP, SOURCE_STEP, TEST_STEP, TESTCASES_STEP
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2896,7 +2896,11 @@ def build_and_install_one(ecdict, init_env):
             _log.info("Build stats: %s" % buildstats)
 
             # move the reproducability files to the final log directory
-            copy_dir(reprod_dir, os.path.join(new_log_dir, os.path.basename(reprod_dir)))
+            archive_reprod_dir = os.path.join(new_log_dir, os.path.basename(reprod_dir))
+            if os.path.exists(archive_reprod_dir):
+                print_warning("Reproducability directory %s already exists, cannot overwrite", archive_reprod_dir)
+            else:
+                copy_dir(reprod_dir, archive_reprod_dir)
 
             try:
                 # upload easyconfig (and patch files) to central repository
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2866,6 +2866,8 @@ def build_and_install_one(ecdict, init_env):
     # make sure we're back in original directory before we finish up
     change_dir(cwd)
 
+    application_log = None
+
     # successful (non-dry-run) build
     if result and not dry_run:
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -72,10 +72,10 @@ from easybuild.tools.config import install_path, log_path, package_path, source_
 from easybuild.tools.environment import restore_env, sanitize_env
 from easybuild.tools.filetools import CHECKSUM_TYPE_MD5, CHECKSUM_TYPE_SHA256
 from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file, change_dir, convert_name
-from easybuild.tools.filetools import compute_checksum, copy_file, derive_alt_pypi_url, diff_files, download_file
-from easybuild.tools.filetools import encode_class_name, extract_file, get_source_tarball_from_git, is_alt_pypi_url
-from easybuild.tools.filetools import is_sha256_checksum, mkdir, move_logs, read_file, remove_file, rmtree2
-from easybuild.tools.filetools import verify_checksum, weld_paths, write_file
+from easybuild.tools.filetools import compute_checksum, copy_dir, copy_file, derive_alt_pypi_url, diff_files
+from easybuild.tools.filetools import download_file, encode_class_name, extract_file, get_source_tarball_from_git
+from easybuild.tools.filetools import is_alt_pypi_url, is_sha256_checksum, mkdir, move_logs, read_file, remove_file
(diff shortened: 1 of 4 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2899,6 +2899,22 @@ def build_and_install_one(ecdict, init_env):
                 _log.info("Dumped fully processed easyconfig to %s", reprod_spec)
             except NotImplementedError as err:
                 _log.warn("Unable to dumped fully processed easyconfig to %s: %s", reprod_spec, err)
+            # also archive the relevant easyblocks
+            reprod_spec = os.path.join(new_log_dir, 'reprod', 'easyblocks')
+            for easyblock_class in inspect.getmro(app_class):
+                easyblock_path = inspect.getsourcefile(easyblock_class)
+                easyblock_splitpath = os.path.split(easyblock_path)
+                easyblock_basedir = easyblock_splitpath[0]
+                easyblock_filename = easyblock_splitpath[1]
```
```python
easyblock_basedir, easyblock_filename = os.path.split(easyblock_path)

File: easybuild/framework/easyblock.py [hunk 2]
```diff
@@ -2899,6 +2899,22 @@ def build_and_install_one(ecdict, init_env):
...
+                framework_class_dir = os.path.join('easybuild', 'framework')
+                if easyblock_basedir.endswith(framework_class_dir):
+                    break
```
boegel: @ocaisa This is basically where you hit `easyblock.py`? Maybe add a comment to clarify...

test/framework/toy_build.py
```diff
@@ -1536,6 +1536,10 @@ def test_minimal_toolchains(self):
         reprod_ec = os.path.join(self.test_installpath, 'software', 'toy', '0.0', 'easybuild', 'reprod', 'toy-0.0.eb')
         self.assertTrue(os.path.exists(reprod_ec))
 
+        reprod_easyblock = os.path.join(self.test_installpath, 'software', 'toy', '0.0', 'easybuild', 'reprod',
+                                        'easyblocks', 'toy.py')
+        self.assertTrue(os.path.exists(reprod_easyblock))
```
boegel: @ocaisa Can we come up with an additional test to make sure inheritance is picked up correctly?

Also, make sure `easyblock.py` is *not* being copied, i.e. there's only one file being copied into `reprod/easyblocks` in this case?

ocaisa: 2nd comment is done, first is a bit more involved.

ocaisa: Ok done

test/framework/toy_build.py
```diff
@@ -1532,10 +1533,27 @@ def test_minimal_toolchains(self):
         # this test doesn't check for anything specific to using minimal toolchains, only side-effects
         self.test_toy_build(extra_args=['--minimal-toolchains'])
 
+    def test_reproducability(self):
+        """Test toy build produces expected reproducability files"""
+        # use the easyblock with inheritance to fully test
+        self.test_toy_build(extra_args=['--minimal-toolchains', '--easyblock=EB_toytoy'])
```
ocaisa: I happened to notice that the `--easyblock` option is not recorded in the reproducible easyconfig. Not sure where I would do this...

ocaisa: Fixed in this PR

easybuild/framework/easyblock.py
```diff
@@ -2859,7 +2862,6 @@ def build_and_install_one(ecdict, init_env):
         errormsg = "build failed (first %d chars): %s" % (first_n, err.msg[:first_n])
         _log.warning(errormsg)
         result = False
-    app.close_log()
```
ocaisa: Logging was being shut down too early, meaning all logging below this was lost

easybuild/framework/easyblock.py
```diff
@@ -2898,7 +2901,21 @@ def build_and_install_one(ecdict, init_env):
...
+            reprod_spec = os.path.join(new_log_dir, 'reprod', 'easyblocks')
+            for easyblock_class in inspect.getmro(app_class):
+                easyblock_path = inspect.getsourcefile(easyblock_class)
+                easyblock_basedir, easyblock_filename = os.path.split(easyblock_path)
+                # if we reach the framework modules we are done
+                if easyblock_filename in FRAMEWORK_EASYBLOCK_MODULES:
+                    break
+                else:
+                    # Check if the easyblock should be in the generic subdir
+                    if os.path.basename(easyblock_basedir) == 'generic':
```
boegel: @ocaisa We have `'generic'` hardcoded in a number of places, it's time we introduce a constant `GENERIC` for this to avoid silly bugs due to typos?

easybuild/framework/easyblock.py
```diff
@@ -2898,7 +2901,21 @@ def build_and_install_one(ecdict, init_env):
...
             except NotImplementedError as err:
-                _log.warn("Unable to dumped fully processed easyconfig to %s: %s", reprod_spec, err)
+                _log.warn("Unable to dump fully processed easyconfig to %s: %s", reprod_spec, err)
+            # also archive the relevant easyblocks
+            reprod_spec = os.path.join(new_log_dir, 'reprod', 'easyblocks')
+            for easyblock_class in inspect.getmro(app_class):
+                easyblock_path = inspect.getsourcefile(easyblock_class)
+                easyblock_basedir, easyblock_filename = os.path.split(easyblock_path)
+                # if we reach the framework modules we are done
+                if easyblock_filename in FRAMEWORK_EASYBLOCK_MODULES:
```
boegel: @ocaisa I think it's better to check this via `easyblock_class`, against `EasyBlock` and `ExtensionEasyblock`, rather than relying on filenames of the Python modules.

For `ExtensionEasyblock` that as a bit tricky though, since we can't import that in here (it would lead to a cyclic import). So, how about this:

```python
for easyblock_class in inspect.getmro(app_class):
    # if we reach EasyBlock or ExtensionEasyblock class, we are done
    if easyblock_class.__name__ in [EasyBlock.__name__, 'ExtensionEasyblock']:
        break
    ...
```

And then chuck `FRAMEWORK_EASYBLOCK_MODULES`, and hardcode `easyblock.py` and `extensioneasyblock.py` in the tests (it's OK there imho, and then we're checking indirectly compared to how this is actually implemented).

easybuild/framework/easyblock.py
```diff
@@ -2898,7 +2901,21 @@ def build_and_install_one(ecdict, init_env):
                 app.cfg.dump(reprod_spec)
                 _log.info("Dumped fully processed easyconfig to %s", reprod_spec)
             except NotImplementedError as err:
-                _log.warn("Unable to dumped fully processed easyconfig to %s: %s", reprod_spec, err)
+                _log.warn("Unable to dump fully processed easyconfig to %s: %s", reprod_spec, err)
+            # also archive the relevant easyblocks
+            reprod_spec = os.path.join(new_log_dir, 'reprod', 'easyblocks')
+            for easyblock_class in inspect.getmro(app_class):
```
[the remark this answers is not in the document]
ocaisa: Just for the easyblocks or the entire `reprod` dir?

easybuild/framework/easyblock.py
ocaisa: Actually the behaviour of including easyblocks has changed I think, I don't think this check is actually necessary (EB automatically puts an included easyblock in the appropriate directory)

easybuild/framework/easyblock.py
```diff
@@ -2982,6 +2978,39 @@ def build_and_install_one(ecdict, init_env):
     return (success, application_log, errormsg)
 
 
+def reproduce_build(app, reprod_dir_root):
```
ocaisa: @boegel I need some help with how to unit test this, not sure what's required to get a proper `app` instance

ocaisa: ...or perhaps what is already tested is good enough?

easybuild/framework/easyblock.py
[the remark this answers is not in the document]
boegel: Certainly the block of code that is added now, which is big enough to stand on its own.

Maybe also the whole logic that deals with `reprod` should be a dedicated function...

easybuild/framework/easyblock.py
boegel: This is for backward compatiblity... If you include an easyblock with `--include-easyblocks generic/configuremake.py`, it should still work like it did before.

It's true that `--include-easyblocks configuremake.py` will works as expected though, but that doesn't mean we should break people's configurations where they put generic easyblocks in a `generic` subdirectory because that's what EasyBuild required up to some point...

We can deprecate this block of logic though, and make it clear it's no longer required to do this. But on the other hand, people may *like* having the generic easyblock separate from the rest, so it also makes sense to keep supporting this.

easybuild/framework/easyblock.py
boegel: @ocaisa Once you have a parsed easyconfig (for example for `toy-0.0.eb` in the tests, you can just create an easyblock instance like `MyEasyBlock(parsed_easyconfig)` (e.g. `Toy(toy_ec)`).

But I guess there's no strong need to test this function in isolation, since it's always being called. Just make sure whatever it produces it being checked, which you're already doing I think.

boegel:
@ocaisa I'll check why @boegelbot isn't picking up on this, but there's a failing test:

```
ERROR: Test toy build produces expected reproducability files
  File "/tmp/453993240/lib/python2.6/site-packages/easybuild_framework-3.7.2.dev0-py2.6.egg/test/framework/toy_build.py", line 1557, in test_reproducability
    for framework_easyblock in FRAMEWORK_EASYBLOCK_MODULES:
NameError: global name 'FRAMEWORK_EASYBLOCK_MODULES' is not defined
```

easybuild/framework/easyblock.py
ocaisa: Hmm, from our configuration I noticed that we now have `--include-easyblocks=$CUSTOM/*.py,$CUSTOM/generic/*.py`. Does `--include-easyblocks=$CUSTOM/*` still work? I'd like to make the reproducability command simple and generic:
```
eb --include-easyblocks=.../reprod/* .../reprod/*.eb
```

ocaisa:
To address https://github.com/easybuilders/easybuild-easyblocks/issues/1574 , should I be doing the reproducability dump *before* `app.run_all_steps()`? That should save me from changes made to the instance by the easyblocks while preserving everything else. I'll have to do it in the temporary space and then move it if we succeed with the build.

ocaisa:
@boegel After a quick test, the approach in the last comment *looks* like it would solve https://github.com/easybuilders/easybuild-easyblocks/issues/1574 but I would like your opinion before I implement.

ocaisa:
Ok, I've implemented the move. This covers the case where modifications come from the command line (or configuration) and dependency resolution but doesn't cover what hooks might do. I think I'd also like to archive those but I don't know how to access them.

ocaisa:
Urgh, I already see a problem with the hook example https://easybuild.readthedocs.io/en/latest/Hooks.html?highlight=hooks#example-hook-to-inject-a-custom-patch-file
I guess patch files should also be copied over to the reprod folder?

ocaisa:
I've reverted this back to simply also dumping the used easyblocks. Fixing the deeper problems of this approach (as in https://github.com/easybuilders/easybuild-easyblocks/issues/1574) is for another PR.

boegel:
lgtm, thanks @ocaisa!

PR #2657, opened 2018-11-12
changed easyconfig.update to be able to specify not to allow duplicate values

mboisson:
This is to address https://github.com/easybuilders/easybuild-easyblocks/issues/1574

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -442,17 +442,18 @@ class EasyConfig(object):
 
         return ec
 
-    def update(self, key, value):
+    def update(self, key, value, allow_duplicate = True):
         """
         Update a string configuration value with a value (i.e. append to it).
         """
         prev_value = self[key]
-        if isinstance(prev_value, basestring):
-            self[key] = '%s %s ' % (prev_value, value)
-        elif isinstance(prev_value, list):
-            self[key] = prev_value + value
-        else:
-            raise EasyBuildError("Can't update configuration value for %s, because it's not a string or list.", key)
+        if allow_duplicate or not value in prev_value:
+            if isinstance(prev_value, basestring):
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -442,17 +442,18 @@ def copy(self):
...
         Update a string configuration value with a value (i.e. append to it).
         """
         prev_value = self[key]
-        if isinstance(prev_value, basestring):
-            self[key] = '%s %s ' % (prev_value, value)
-        elif isinstance(prev_value, list):
-            self[key] = prev_value + value
-        else:
-            raise EasyBuildError("Can't update configuration value for %s, because it's not a string or list.", key)
+        if allow_duplicate or value not in prev_value:
```
boegel: @mboisson This check isn't entirely correct in the case of `prev_value` (and hence `value`) being a list, since:

```python
>>> [2,3] in [1,2,3]
False
>>> [2,3] not in [1,2,3]
True
```

mboisson: Ah, I hadn't considered that `value` could be a list too, I assumed that we'ld be adding a single element to `prev_value`. I'll fix that.

mboisson: fixed in latest commit

mboisson:
Tests included. Fixed the list problem.

mboisson:
Mmmm, @boegel, am I noticing boegelbot acting stupid ? I don't understand the error log.

boegel:
@mboisson No, it's correctly reporting a failing test. Try running the failing test locally with `python -O -m test.framework.suite test_update`.

mboisson:
Ok, seems the tests all passed. Ready for another review @boegel

boegel:
lgtm, thanks @mboisson!

PR #2660, opened 2018-11-15
fix order of keys in 'toolchain' value for dumped easyconfig file (name, version) + run style check on dumped easyconfigs in dump tests

boegel:
Minor bug fix for something silly that has been annoying me...


boegel:
@ocaisa Not currently, no, the automated style check doesn't take into account order of easyconfig parameters...

The fix for that issue is probably to add `checksums` to `GROUPED_PARAMS` in `framework/easyconfig/format/format.py`, but let's make that a separate PR?

ocaisa:
Sure, just noticed it when browsing through reprod stuff during the week

PR #2661, opened 2018-11-15
Ensure checksums beside sources in dumped easyconfigs

ocaisa:


easybuild/framework/easyconfig/format/format.py
```diff
@@ -62,8 +62,7 @@
     ['name', 'version', 'versionprefix', 'versionsuffix'],
     ['homepage', 'description'],
     ['toolchain', 'toolchainopts'],
-    ['sources', 'source_urls'],
-    ['patches'],
+    ['source_urls', 'sources', 'patches', 'checksums'],
```
boegel: Hmm, there was a reason why `patches` was separate here...

Does this affect anything else than dumped easyconfigs?

boegel: I checked, this (`GROUPED_PARAMS`) is only used in the `dump` method, so only affects dumping of easyconfigs

boegel:
@ocaisa I've pushed in a test for this, please take a look? If you're OK with it and Travis gives green light, this is good to go.

boegel:
Thanks @ocaisa!

Issue #2658, opened 2018-11-12
--module-only --force messes with the reprod folder


mboisson:
When running eb with the options `--module-only --force`, it recreates the easyconfig stored in the `reprod` folder. Unfortunately, this means that this easyconfig may actually no longer be reproducible. 

This happened to me with GDAL, when the `start_dir` that is listed in the EasyConfig in the `reprod` folder went from 
```
start_dir = '%(builddir)s/%(namelower)s-%(version)s/'
``` 
to 
```
start_dir = '%(builddir)s/'
``` 

This is because the EasyBlock updates its config, and the full EasyBlock is not being run.

mboisson:
@boegel would avoiding to over-write the `reprod/*.eb` if `--module-only`  is used make sense to you ?

boegel:
@mboisson It probably does, but I guess it also depends: you may have made a change in the easyconfig file to make the sanity check pass for example. Not including that in the easyconfig file in `reprod` is also problematic...

cc @ocaisa

mboisson:
I would say we should also avoid dumping
exts_filter

When installing Platypus, I get the following error : 
```
== sanity checking...
ERROR: Traceback (most recent call last):
  File "/cvmfs/soft.computecanada.ca/easybuild/lib/python2.7/site-packages/easybuild-framework/easybuild/main.py", line 126, in build_and_install_software
    (ec_res['success'], app_log, err) = build_and_install_one(ec, init_env)
  File "/cvmfs/soft.computecanada.ca/easybuild/lib/python2.7/site-packages/easybuild-framework/easybuild/framework/easyblock.py", line 2860, in build_and_install_one
    result = app.run_all_steps(run_test_cases=run_test_cases)
  File "/cvmfs/soft.computecanada.ca/easybuild/lib/python2.7/site-packages/easybuild-framework/easybuild/framework/easyblock.py", line 2776, in run_all_steps
    self.run_step(step_name, step_methods)
  File "/cvmfs/soft.computecanada.ca/easybuild/lib/python2.7/site-packages/easybuild-framework/easybuild/framework/easyblock.py", line 2642, in run_step
    step_method(self)()
  File "/cvmfs/soft.computecanada.ca/easybuild/lib/python2.7/site-packages/easybuild-easyblocks/easybuild/easyblocks/generic/pythonpackage.py", line 538, in sanity_check_step
    (success, fail_msg) = super(PythonPackage, self).sanity_check_step(*args, **kwargs)
[...]

PR #2664, opened 2018-11-16
fix problems with easyconfig file saved in 'reprod' directory by copying it before running any installation steps

ocaisa:
Builds on ~~#2653~~


easybuild/framework/easyblock.py
```diff
@@ -2853,13 +2855,15 @@ def build_and_install_one(ecdict, init_env):
     start_time = time.time()
     try:
         run_test_cases = not build_option('skip_test_cases') and app.cfg['tests']
+        # create our reproducability files before carrying out the easyblock steps
+        reprod_dir_root = os.path.dirname(app.logfile)
+        reprod_dir = reproduce_build(app, reprod_dir_root)
```
ocaisa: Somehow, moving this causes the yeb format builds to fail

easybuild/framework/easyblock.py
```diff
@@ -2982,6 +2985,43 @@ def build_and_install_one(ecdict, init_env):
...
+    """
+
+    ec_filename = '%s-%s.eb' % (app.name, det_full_ec_version(app.cfg))
+
+    # for reproducability we dump out the fully processed easyconfig since the contents can be affected
+    # by subtoolchain resolution (and related options) and/or hooks
+    reprod_dir = os.path.join(reprod_dir_root, 'reprod')
+    reprod_spec = os.path.join(reprod_dir, ec_filename)
+    try:
+        app.cfg.dump(reprod_spec)
```
ocaisa: Confirmed, doing this early causes the yeb build to barf

boegel:
@ocaisa Please sync this with current `develop` now #2653 has been merged?

easybuild/framework/easyblock.py
```diff
@@ -2855,6 +2857,10 @@ def build_and_install_one(ecdict, init_env):
     start_time = time.time()
     try:
         run_test_cases = not build_option('skip_test_cases') and app.cfg['tests']
+        if not dry_run:
+            # create our reproducability files before carrying out the easyblock steps
+            reprod_dir_root = os.path.dirname(app.logfile)
+            reprod_dir = reproduce_build(app, reprod_dir_root)
```
ocaisa: This will definitely impact how the archived easyconfig looks, with many parameters introduced by the easyblock steps no longer recorded in the dump (which is a good thing). Also recording the hooks should sidestep *most* problems with the fact that we also would no longer record any of the changes made by the hooks in the easyblock steps.

ocaisa:
The final reprod easyconfig still includes at least 3 changes that are (potentially) unwanted:
```
parallel = 2
start_dir = '%(builddir)s/%(name)s-%(version)s/'
unwanted_env_vars = {}
```
using `zlib-1.2.8.eb`

ocaisa:
I'm ok with `parallel` and `unwanted_env_vars`, they could easily impact the build but from https://github.com/easybuilders/easybuild-framework/issues/2658 we know `start_dir` can be a problem (though it is now no longer an absolute path).

boegel:
@ocaisa Shouldn't we drop the 2nd call to `reproduce_build` in `build_and_install_one`?

Also, make sure the tests fail when `start_dir` in included in the easyconfig file that is dumped to `reprod/`

ocaisa:
Ah, the second call is a hangover from the merge...fixing

ocaisa:
Ok, now I understand things better, the dump is actually failing but this was being masked because the dumping was happening in a try/except and then the second call to the function was succeeding.

ocaisa:
Ok, the merge from develop reverted a number of changes, putting them back in place makes things look a lot more reasonable. No longer seeing the changes described in https://github.com/easybuilders/easybuild-framework/pull/2664#issuecomment-440954082

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -898,7 +898,12 @@ def dump(self, fp, always_overwrite=True, backup=False):
             if self.template_values[key] not in templ_val and len(self.template_values[key]) > 2:
                 templ_val[self.template_values[key]] = key
 
-        ectxt = self.parser.dump(self, default_values, templ_const, templ_val)
+        try:
+            ectxt = self.parser.dump(self, default_values, templ_const, templ_val)
+        except NotImplementedError as err:
+            self.enable_templating = orig_enable_templating
+            raise NotImplementedError(err)
```
boegel: @ocaisa: Can you clarify why this is needed? It seems useless since you're re-raising the error anyway?

ocaisa: This error is being caught in a try/except for yeb files and turned into a log error, if the templating value is not restored it results in a problem when doing anything with the ec instance since templating is switched off otherwise.

ocaisa: To clarify, I `reproduce_build()` catches this error and turns it into a warning. It then continues with the build (which can still succeed) but requires that the original `enable_templating` value was restored.

ocaisa: I've added a comment that hopefully makes this a bit clearer

easybuild/framework/easyblock.py
```diff
@@ -2892,9 +2894,12 @@ def build_and_install_one(ecdict, init_env):
             buildstats = get_build_stats(app, start_time, build_option('command_line'))
             _log.info("Build stats: %s" % buildstats)
 
-            # for reproducability we dump out the fully processed easyconfig since the contents can be affected
-            # by subtoolchain resolution (and related options) and/or hooks
-            reprod_dir = reproduce_build(app, new_log_dir)
+            # move the reproducability files to the final log directory
+            archive_reprod_dir = os.path.join(new_log_dir, os.path.basename(reprod_dir))
+            if os.path.exists(archive_reprod_dir):
+                print_warning("Reproducability directory %s already exists, cannot overwrite", archive_reprod_dir)
```
boegel: @ocaisa This will never actually print a warning as is, cfr. https://github.com/ocaisa/easybuild-framework/pull/40 for a fix.

However, I'm a bit worried about this: when `--module-only` is used, this means that the stuff in the `reprod` dir will never get refreshed...

I don't think that's our intention, is it? We *should* be overwriting in that case, no?

It's a tricky one, I guess, since we probably don't want to blindly blow away what is there already either...

ocaisa: I was considering doing a bit of inception here and creating a `module_only` subdirectory in the reprod dir for that case and dumping everything there instead. That one would be allowed to be overwritten if the conditions were met again.

boegel: @ocaisa How about a "safe" variant of `copy_dir` which backs up existing files before overwriting them? That would come in useful here I think, since then you solve the problem of whether or not to replace the existing files under `--module-only`.

ocaisa: This doesn't exist yet right? Are you suggesting that I make a back_up directory and move things there? That would be a lot easier than doing it file by file

easybuild/framework/easyblock.py
```diff
@@ -2892,10 +2895,15 @@ def build_and_install_one(ecdict, init_env):
...
             _log.info("Build stats: %s" % buildstats)
 
-            # for reproducability we dump out the fully processed easyconfig since the contents can be affected
-            # by subtoolchain resolution (and related options) and/or hooks
-            reprod_dir = reproduce_build(app, new_log_dir)
-            _log.info("Wrote files for reproducability to %s", reprod_dir)
+            # move the reproducability files to the final log directory
+            archive_reprod_dir = os.path.join(new_log_dir, os.path.basename(reprod_dir))
+            if os.path.exists(archive_reprod_dir):
+                _log.info("Reproducability directory %s already exists, backing up" % archive_reprod_dir)
```
ocaisa: @boegel This is now just a log entry since we don't remove any information.

easybuild/framework/easyblock.py
```diff
@@ -2892,10 +2895,15 @@ def build_and_install_one(ecdict, init_env):
...
+                backup_dir = find_backup_name_candidate(archive_reprod_dir)
+                copy_dir(archive_reprod_dir, backup_dir)
+                rmtree2(archive_reprod_dir)
```
boegel: @ocaisa You can probably use `move_file` here rather than copy & remove; despite it's name `move_file` should also work on directories.

Also, add a log entry (or move down the one above):

```python
_log.info("Existing reprod directory %s backed up to %s", archive_reprod_dir, backup_dir)

boegel:
@ocaisa Broken test is fixed with https://github.com/ocaisa/easybuild-framework/pull/41

PR #2668, opened 2018-11-22
use github_user in test_list_prs

migueldiascosta:


boegel:
Thanks @migueldiascosta!

PR #2670, opened 2018-11-22
make print functions more robust w.r.t. arguments being passed to format the message to be printed

boegel:
This avoids that `print_warning("Example: %s", foo)` gets interpreted as `print_warning("Example: %s", silent=True)`, but as `print_warning("Example: %s" % foo)`.

Strictly speaking, this is backwards incompatible, but if any code out there now uses `print_warning(msg, arg`) it's most likely not working as intended. And using `print_warning` with `silent=True` is quite silly anyway, that only makes sense in the tests.


migueldiascosta:
Going in, thanks @boegel!

PR #2673, opened 2018-11-27
Python 2.6 requires python-daemon 1.x as dep for GC3Pie

boegel:
Somehow the release of GC3Pie 2.5.1 is pulling in `python-daemon` 2.2.0, which is not compatible with Python 2.6, causing tests to fail.

This is an attempt to fix that issue...

cc @riccardomurri

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,8 @@ before_install:
     - if [ "x$TRAVIS_PYTHON_VERSION" == 'x2.6' ]; then pip install 'paramiko<2.4.0'; fi
     # SQLAlchemy 1.2.0 (dep for GC3Pie) & more recent doesn't work with Python 2.6
     - if [ "x$TRAVIS_PYTHON_VERSION" == 'x2.6' ]; then pip install 'SQLAlchemy<1.2.0'; fi
+    # python 2.0 (dep for GC3Pie) & more recent doesn't work with Python 2.6
+    - if [ "x$TRAVIS_PYTHON_VERSION" == 'x2.6' ]; then pip install 'python-daemon<2.0'; fi
     # autopep8 1.3.4 is last one to support Python 2.6
     - if [ "x$TRAVIS_PYTHON_VERSION" == 'x2.6' ]; then pip install 'autopep8<1.3.5'; else pip install autopep8; fi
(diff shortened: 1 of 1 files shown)


riccardomurri:
I think something changed in the way `pip` handles dependencies or installs packages... I spent quite some time yesterday trying to fix GC3Pie/Py2.6 tests (which fail for a similar reason).

GC3Pie tests fail because `pycparser` (which is a dependency of PyNaCl, which in turns is required by Paramiko, which GC3Pie uses) uses 2.7-only syntax in version 2.19.  So if you depend on `pycparser==2.18` it should work, right?  Well, it turns out that `pip` *does* install `pycparser==2.18` according to the Travis CI logs, but then ignores it and still tries to install the latest `pycparser` when installing `PyNaCl`... even if one lists `PyNaCl==1.0.1` (last version to support Py2.6) in the dependencies!

I am really puzzled.

boegel:
@riccardomurri Which version of `pip` are you using in that context? We're not seeing that issue with the `pip` 9.0.3 (which is the latest that supports Python 2.6, it seems, since that was the version available after `pip install --upgrade pip`) being used by Travis in our tests (see https://travis-ci.org/easybuilders/easybuild-framework/jobs/460323743).
After `pip install 'pycparser<2.19'` and `pip install 'python-daemon<2.0'` now, `pip install GC3Pie` seems to be working just fine...

riccardomurri:
> After pip install 'pycparser<2.19' and pip install 'python-daemon<2.0' now, pip install GC3Pie seems to be working just fine...

But you do this in Travis' script *before* installing GC3Pie, you do not do it from the `setup.py` script, right?

boegel:
@riccardomurri That's right, and then `pip` picks up on those pre-installed packages when installing `GC3Pie` since they fit the requirements, see: 

```
Requirement already satisfied: python-daemon in /home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages (from GC3Pie)
...
Requirement already satisfied: pycparser in /home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages (from cffi>=1.1->bcrypt>=3.1.3->paramiko->GC3Pie)
```

PR #2674, opened 2018-11-28
Add dependencies in the prepare step instead of check_readiness.

bartoldeman:
Using the prepare step we will be able to iterate over multiple
build-dependencies such as multiple Python versions.

~~This also removes the dependencies field from the Toolchain class which
handles them completely functionally now. The EasyBlock class now has
self.dependencies to remember them for outputting in modulefiles.~~


easybuild/tools/toolchain/toolchain.py
```diff
@@ -120,7 +120,6 @@ def __init__(self, name=None, version=None, mns=None, class_constants=None, tcde
         """
         self.base_init()
 
-        self.dependencies = []
```
boegel: @bartoldeman We can't just drop this, since existing easyblocks are using this, see `generic/craytoolchain.py` and `wien2k.py`...

bartoldeman:
ping @boegel ?

boegel:
@bartoldeman The problem with the tests is "fixed" in #2678, I'll retrigger the tests for this PR...

easybuild/tools/toolchain/toolchain.py
```diff
@@ -688,8 +693,18 @@ def prepare(self, onlymod=None, silent=False, loadmod=True, rpath_filter_dirs=No
         :param rpath_filter_dirs: extra directories to include in RPATH filter (e.g. build dir, tmpdir, ...)
         :param rpath_include_dirs: extra directories to include in RPATH
         """
+
+        # do all dependencies have a toolchain version?
+        if deps is None:
+            deps = []
+        tcdeps = self._check_dependencies(deps)
```
boegel: @bartoldeman Why not make this `self.dependencies = self._check_dependencies(deps)`?

Then there's no need to change `_load_modules`, `_load_dependencies_modules` and `_add_dependency_variables` to pass down `tcdeps`, they can just keep using `self.dependencies` like they were before?

It's a bit silly to pass them around if we have `self.dependencies` anyway (which we can't remove since easyblocks may be using it)...

boegel:
Thanks @bartoldeman!

PR #2676, opened 2018-11-29
deprecate goolf and goolfc toolchains

boegel:
Deprecating the `goolf` and `goolfc` toolchains triggers *a lot* of (trivial, in the end) changes in the tests.

The semantics of the tests haven't changed, it's basically just a modernization sweep...


verdurin:
@boegel - do you want testing of this, or a visual inspection?

boegel:
@verdurin There's little point in testing this imho, since it's mainly about updating the framework tests to avoid the use of `goolf` and `goolfc`, and Travis already gives green light on that.

So it boils down to visual inspection for any "suspicious" changes imho.

PR #2678, opened 2018-12-04
stick to idna<2.8 with Python 2.6 in Travis + distable broken test for HgRepository

boegel:


boegel:
HgRepo test failes because of:

WARNING: Mercurial local repo initialization failed, it might already exist: (255, '', 'warning: connecting to bitbucket.org using legacy security technology (TLS 1.0); see https://mercurial-scm.org/wiki/SecureConnections for more info\nabort: error: _ssl.c:510: error:1408F10B:SSL routines:SSL3_GET_RECORD:wrong version number')

boegel:
That test fails because of bitbucket.org disabling TLS 1.0/1.1, while Python 2.6 doesn't support newer TLS versions.

That's a blocker for the `test_hgrepo` test, I see no other way to disable that test for now until we've dropped support for Python 2.6...

Issue #2662, opened 2018-11-15
New --try-toolchain is too strict


mboisson:
We have been using `--try-toolchain` a lot in Compute Canada, and we are always trying to minimize the dependencies, i.e. we don't build with MPI, BLAS, etc. if the software package does not need it. 

This worked great for us until recently. With the new try-toolchain behavior, since upstream does the opposite of us (i.e. compile everything with intel/2018a for example), something like this does not work : 
`eb CDO-1.9.5-intel-2018a.eb --try-toolchain=iompi,2018.3` 

Even though it should work. Nothing in CDO or its dependencies actually requires or uses MKL/BLAS. Yet, I now get the error : 
```
ERROR: No possible mapping from source toolchain spec {'lapack_family': 'IntelMKL', 'name': 'intel', 'blas_family': 'IntelMKL', 'version': '2018a', 'cuda': None, 'comp_family': 'Intel', 'mpi_family': 'IntelMPI'} to target toolchain hierarchy specs [{'lapack_family': None, 'name': 'GCCcore', 'blas_family': None, 'version': '7.3.0', 'cuda': None, 'comp_family': 'GCC', 'mpi_family': None}, {'lapack_family': None, 'name': 'iccifort', 'blas_family': None, 'version': '2018.3', 'cuda': None, 'comp_family': 'Intel', 'mpi_family': None}, {'lapack_family': None, 'name': 'iompi', 'blas_family': None, 'version': '2018.3.312', 'cuda': None, 'comp_family': 'Intel', 'mpi_family': 'OpenMPI'}]
```

mboisson:
Tagging @bartoldeman @ocaisa

ocaisa:
You can revert to previous behaviour with `--force`, see https://github.com/easybuilders/easybuild-framework/pull/2605

boegel:
So, already fixed in EasyBuild v3.7.1 @mboisson?

mboisson:
Looks like we didn't have 3.7.1. Bart will udpate it on our node and I'll report back.

mboisson:
Though I got to admin, `--force` is starting to become used for quite a few varied things. It might be useful to have different options for different forced things ?

boegel:
@mboisson Yeah, you have a good point there... Do you have a particular suggestion in mind?

mboisson:
On a case-by-case basis, maybe. Something like `--try-toolchain-mode=legacy` would have been better than `--force` in this specific example ?

[...]

PR #2682, opened 2018-12-05
add support for disabling mapping of (sub)toolchains when --try-toolchain is used

boegel:
Fix for #2662 reported by @mboisson , or at least it provides a replacement for `--try-toolchain --force`: you can use `--disable-map-toolchains --try-toolchain`, or configure EasyBuild with `export EASYBUILD_DISABLE_MAP_TOOLCHAINS=1`.

cc @bartoldeman (since this changes what got merged via #2605), @ocaisa

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -120,26 +120,26 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
             else:
                 target_toolchain['version'] = source_toolchain['version']
 
-            try:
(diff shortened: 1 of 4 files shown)


mboisson:
It looks alright code-wise. It would take some documentation as to what "mapping of toolchain" means vs "not mapping of toolchains".

boegel:
@mboisson I agree, but we don't have documentation on `--try-toolchain` yet either...

Should this PR be blocked until that is in place?

mboisson:
Probably not

PR #2686, opened 2018-12-06
use 'tar xzf' (gzip) rather than 'tar xZf' (ancient compress) to unpack *.tar.Z source files

boegel:
fix for failure to unpack a `.tar.Z` file because `compress` utility is not available:

```
tar (child): compress: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
```

(see also https://bugs.launchpad.net/ubuntu/+source/tar/+bug/70610)


boegel:
(close/re-open to trigger Travis, there was.a hickup when opening the PR)

Issue #2672, opened 2018-11-27
toolchains/fft/intelfftw.py need to correctly set FFT_LIB_MT to use threaded mkl


akesandgren:
Current behavior is that FFT_LIB_MT is unset and in tools/toolchain/fft.py then gets set to FFT_LIB

boegel:
@akesandgren What would be the correct value for `$FFT_LIB_MT`, and does it depend on the Intel MKL version being used?

akesandgren:
Correct value is (probably) to use the threaded mkl libraries, just like BLAS_LIB_MT from linalg,
it should be all it needs.
I.e. the line 
        fftw_libs.extend(self.variables['LIBBLAS'].flatten())  # add BLAS libs (contains dft)
need to be extended for FFT_LIB_MT
But I'm not quite sure how to do this one

akesandgren:
Solved in https://github.com/easybuilders/easybuild-framework/pull/2688

PR #2688, opened 2018-12-07
correctly define $LIBFFT_MT for Intel MKL

boegel:
fix for #2672

@akesandgren I'm not 100% sure this is correct, so please check.
I assume you have a particular use case in mind for this?

### DIFF
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -870,8 +870,15 @@ class ToolchainTest(EnhancedTestCase):
         libblas_mt_intel3 += " -Wl,--end-group -Wl,-Bdynamic -liomp5 -lguide -lpthread"
 
         # no -lguide
+        libblas_intel4 = "-Wl,-Bstatic -Wl,--start-group -lmkl_intel_lp64 -lmkl_sequential -lmkl_core"
+        libblas_intel4 += " -Wl,--end-group -Wl,-Bdynamic"
         libblas_mt_intel4 = "-Wl,-Bstatic -Wl,--start-group -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core"
         libblas_mt_intel4 += " -Wl,--end-group -Wl,-Bdynamic -liomp5 -lpthread"
+
+        libfft_intel4 = libblas_intel4.replace('-lmkl_intel_lp64', '-lfftw3xc_intel -lmkl_intel_lp64')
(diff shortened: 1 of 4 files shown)


boegel:
@akesandgren Tested with `QuantumESPRESSO-6.2-intel-2017b.eb` and `QuantumESPRESSO-6.3-intel-2018b.eb` where `$LIBFFT_MT` is picked up, seems to work (it builds without issue, don't know how to test QE).

test/framework/toolchain.py
```diff
@@ -870,9 +870,16 @@ def test_old_new_iccifort(self):
...
         # no -lguide
+        libblas_intel4 = "-Wl,-Bstatic -Wl,--start-group -lmkl_intel_lp64 -lmkl_sequential -lmkl_core"
+        libblas_intel4 += " -Wl,--end-group -Wl,-Bdynamic"
         libblas_mt_intel4 = "-Wl,-Bstatic -Wl,--start-group -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core"
         libblas_mt_intel4 += " -Wl,--end-group -Wl,-Bdynamic -liomp5 -lpthread"
 
+        libfft_intel4 = libblas_intel4.replace('-lmkl_intel_lp64', '-lfftw3xc_intel -lmkl_intel_lp64')
+        libfft_mt_intel4 = libblas_mt_intel4.replace('-lmkl_intel_lp64', '-lfftw3xc_intel -lmkl_intel_lp64')
+        libfft_mt_intel4 = "-Wl,-Bstatic -Wl,--start-group -lfftw3xc_intel -lmkl_intel_lp64 -lmkl_intel_thread"
+        libfft_mt_intel4 += " -lmkl_core -liomp5 -lpthread -Wl,--end-group -Wl,-Bdynamic"
```
akesandgren: These two lines doesn't make sense, you're overwriting the value you just set libfft_mt_intel4 to.

akesandgren: Hmmm, I think the replace would be more correct? I.e. just as libfft_intel4 uses replace on libblas, libfft_mt_intel4 should use replace on libblas_mt ?? Shorter and the -liomp5 -lpthread should be outside the grouping as far as I can see from the code in toolchain/fft.py above

boegel: Agreed, and fixed in 269d6ef

akesandgren:
Going in, thanks @boegel!

Issue #2375, opened 2018-01-12
make robot only require toolchain definitions to resolve dependencies with subtoolchains 


boegel:
Currently, the `--robot` implementation will only resolve dependencies via a subtoolchain if the matching easyconfig file is available in the robot search path (even if the corresponding module is available already).

However, I don't think there's a strict need for it, it should suffice that just the easyconfigs that define the toolchain are available, since that's sufficient to figure out the subtoolchains to consider (w.r.t. versions).

Once the subtoolchains are known, available modules can be sufficient to resolve a dependency via a subtoolchain (unless there are other aspects that require the easyconfigs to be available, e.g. the use of `HierarchicalMNS` which requires an easyconfig file to determine the full module name).

Thoughts @ocaisa?

ocaisa:
This approach also wouldn't work if `modaltsoftname` was used in the (now unparsed) dep easyconfig as the module name wouldn't be correct.

paulmelis:
To clearify, is this bug applicable only when `--robot` is used?

boegel:
@paulmelis Hmm, good question... It's actually the process of finding modules for each of the dependencies, which is not *exactly* the same as `--robot` (which does more than that).

@ocaisa I hope to find time to try and fix this bug soon. You've been doing quite a bit of work related to subtoolchains recently, any idea how difficult this would be?

boegel:
I took a quick look at this, and it seems we can avoid the requirement that the easyconfig file for the dependencies are available via the robot search path by not always calling `robot_find_easyconfig` in `robot_find_minimal_toolchain_of_dependency`, which takes care of figuring out which subtoolchain should be used for each dependency.

There's a complication though... The alternative for considering subtoolchains is looking at existing modules to pick a subtoolchain for a particular dependency, which currently requires that `--use-existing-modules` is enabled (it's disabled by default).

Also, the problem reported in #2535 and the proposed fix in #2536 potentially complicates the situation even further...

@ocaisa I'd like to pick your brain on this one, can we set up a conf call on this sometime soon?

casparvl:
+1, I have some new colleagues and it is hard to explain to them why this is happening (‘but… the module is already there!’). If EasyBuild can do without the EasyConfig, that would definitely be more intuitive behaviour :)

akesandgren:
And you're not telling EB to use the ebfiles_repo to search for easyconfigs? If there is a module there should (normally) be an easyconfig in there that matches.

[...]

PR #2690, opened 2018-12-12
don't always require easyconfig files to resolve dependencies

boegel:
potential fix for #2375 

Currently, we always call `robot_find_easyconfig` in `robot_find_minimal_toolchain_of_dependency` which implies that an easyconfig file must be available for each dependency that was installed with a subtoolchain. This is done because determine the module name for this dependency *may* require a full easyconfig file, as enforced by `ActiveMNS.det_full_module_name` (via `ActiveMNS.check_ec_type`).

There's no need for this however, since in some situations a full easyconfig file is not required at all to determine the module name, for example when using the default (flat) module naming scheme `EasyBuildMNS`.

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1121,11 +1121,8 @@ class EasyConfig(object):
                         # try finding subtoolchain for dep for which an easyconfig file is available
                         # this may fail, since it requires that the easyconfigs for parent toolchain
                         # and subtoolchains are available
-                        try:
-                            tc = robot_find_subtoolchain_for_dep(dep, self.modules_tool, parent_first=True)
-                            self.log.debug("Using subtoolchain %s for dep %s", tc, dep_str)
-                        except EasyBuildError as err:
-                            self.log.debug("Ignoring error while looking for subtoolchain for dep %s: %s", dep_str, err)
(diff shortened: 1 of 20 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1644,58 +1644,90 @@ def verify_easyconfig_filename(path, specs, parsed_ec=None):
...
-                minimal_toolchain = filtered_possibilities[-1]['toolchain']
-    else:
+                mod_exists = mod_name in avail_modules or modtool.exist([mod_name], skip_avail=True)[0]
+
+            # add the subtoolchain to list of candidates
+            cand_subtcs.append({'toolchain': tc, 'mod_exists': mod_exists})
+
+    _log.debug("List of possible subtoolchains for %s: %s", dep, cand_subtcs)
+
+    cand_subtcs_with_mod = [tc for tc in cand_subtcs if tc.get('mod_exists', False)]
```
ocaisa: Shouldn't this be
```
cand_subtcs_with_mod = [tc for tc in cand_subtcs if tc.get('mod_exists', True)]
```
or else the list name is bad

ocaisa: Whoops, sorry, should really know by now how `get()` works!

ocaisa:
Fixes #2375

PR #2691, opened 2018-12-14
Add GCCcore 7.3.0 easyconfig for --review-pr test

ocaisa:


boegel:
@ocaisa This is not just a problem with the tests, it's a signal that behavior has changed via #2690, and we should fix it properly.

I don't see a reason why the `GCCcore` easyconfig is required for a `--review-pr` on a PR that just has a `CMake` easyconfig (see https://github.com/easybuilders/easybuild-easyconfigs/pull/6660, that's the PR being used in the test).

PR #2697, opened 2018-12-15
significantly speed up robot_find_subtoolchain_for_dep by specifying whether dep is resolved via an external module

boegel:
I noticed that the easyconfigs test suite became significantly slower after #2690 got merged, so I did a profile run to figure out where most time was being spent...

It turns out that a whopping 75% of time was spent in the `mod_exist_via_show` inner function of `ModulesTool.exist`, which is only used as a fallback check in case a (visible) module is not found via the list produced by `module avail`.
This is required because a module name can be partial, in the case of external modules (cfr. https://easybuild.readthedocs.io/en/latest/Using_external_modules.html).

`mod_exist_via_show` is expensive since it involves actually calling out to the modules tool (e.g. Lmod) every time. We already have a "show cache", which helps quite a bit (it cuts down actual `module show` calls to with about 75% in the case of the easyconfigs test suite), but we still call `module show` once for every module name that is not found in the output of `module avail`.

Since `robot_find_subtoolchain_for_dep` knows when a dependency is resolved via an external module or not, it can pass down this information to `ModulesTool.exist` to avoid all these needless `module show` calls when we are certain that the module name is not partial (since that can only occur with dependencies marked as external modules).

It's worth mentioning that `ModulesTool.exist` already handles hidden module names separately, and so a fallback to `show` is already in place for module names for which the actual module file starts with a `.`.
For modules that are hidden via Lmod's feature to hide modules via `.modulerc` were also save, since we run `module --show-hidden avail` with Lmod, and so hidden modules *will* be listed.

This change has a dramatic effect on the time needed for easyconfig test suite: it was cut down from ~2550s to ~550s (on my laptop).

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1696,9 +1696,12 @@ def robot_find_subtoolchain_for_dep(dep, modtool, parent_tc=None, parent_first=F
             # check whether module already exists or not (but only if that info will actually be used)
             mod_exists = None
             if parent_first or use_existing_modules:
+                mod_exists = mod_name in avail_modules
                 # fallback to checking with modtool.exist is required,
                 # for hidden modules and external modules where module name may be partial
-                mod_exists = mod_name in avail_modules or modtool.exist([mod_name], skip_avail=True)[0]
+                if not mod_exists:
+                    maybe_partial = dep.get('external_module', True)
+                    mod_exists = modtool.exist([mod_name], skip_avail=True, maybe_partial=maybe_partial)[0]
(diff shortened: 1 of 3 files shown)


ocaisa:
Ah ok, this impacts the easyconfig test suite, compare https://travis-ci.org/easybuilders/easybuild-easyconfigs/builds/468182334 to https://travis-ci.org/easybuilders/easybuild-easyconfigs/builds/443093057

ocaisa:
Going in, thanks for reducing our while-that's-chugging-away-I'll-just-go-have-a-coffee window @boegel !

PR #2698, opened 2018-12-17
fix logic in robot_find_subtoolchain_for_dep: consider easyconfigs first, before (maybe) considering available modules

boegel:
This fixes a backwards-incompatible change that was introduced in #2690 .

The logic in `robot_find_minimal_toolchain_of_dependency` was such that the available easyconfig files determined which subtoolchain was used for each dependency, while now in `robot_find_subtoolchain_for_dep` it is available modules first and only then available easyconfig files.

This creates problems in situations where there are multiple easyconfig files available with different (sub)toolchains that can be used to resolve a dependency, and when only some of them are installed.

For example, if `zlib` is installed with `GCCcore` as toolchain, this will determine the subtoolchain used for `zlib` that's listed as a dependency. When later `zlib` is also installed with `foss`, the selected toolchain for a `zlib` dependency changes when `foss` is used as a toolchain, which results in a toolchain conflict on `zlib`.

This problem surfaced during the regression test for EasyBuild v3.8.0, in particular for easyconfigs used the `foss/2016*` and `intel/2016*` toolchains (which is when we started pushing "down" dependencies like `zlib` to the compiler-only subtoolchain rather than the full toolchain).

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1650,8 +1650,6 @@ def robot_find_subtoolchain_for_dep(dep, modtool, parent_tc=None, parent_first=F
     :param parent_first: reverse order in which subtoolchains are considered: parent toolchain, then subtoolchains
     :return: minimal toolchain for which an easyconfig exists for this dependency (and matches build_options)
     """
-    minimal_toolchain = None
-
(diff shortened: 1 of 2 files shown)


boegel:
@ocaisa dedicated test added in f0c0c6d

PR #2708, opened 2018-12-27
ingest vsc-base & vsc-install

boegel:
This set of changes basically ingests `vsc-base` & `vsc-install` into the EasyBuild framework, either by copying the modules from the `vsc.utils` namespace into `easybuild.base`, or by copying the functionality into an existing module in the `easybuild.tools` namespace.

As such, this effectively removes `vsc-base` and `vsc-install` as dependencies.

I tried to only make minimal changes to the code being copied in, this is mostly about making the EasyBuild framework independent from `vsc-base`, not about also cleaning up that code. Therefore, any complaints made by Hound CI will be ignored.
I did strip out a bunch of code from the `vsc-base` modules that is currently not required by the EasyBuild framework.

Once this change is merged, we can start working on making EasyBuild compatible with Python 3.

Note that the target branch here is `4.x` (which was forked from the current `develop` branch), so these changes will only be included in EasyBuild 4.0.
It's quite likely there will be one or more 3.x releases (e.g. 3.8.1, 3.9.0) before 4.0 is released.

The current `develop` branch remains the right place for changes that should be included in the next EasyBuild (3.x) release. I'll keep an eye on merging `develop` into the `4.x` branch on a regular basis.


boegel:
Some further clarification w.r.t. my motivation to ingest `vsc-base` (as requested by @wpoely86):

* It helps with working towards Python 3 compatibility because:
  * Not depending on `vsc-base` anymore means that we can work on Python 3 compatibility on our own pace. Porting `vsc-base` to Python 3 will have to be done with great care because it's a dependency for a whole bunch of scripts & tools inside of HPC-UGent, including a bunch of private repositories, so that process will be slow (and it has barely started).
  * A significant amount of code that is not required by EasyBuild was stripped out when ingesting `vsc-base`, so there's less code to port to Python 3.

* No longer depending on `vsc-base` means that we can also look into getting rid of `setuptools` as a dependency for EasyBuild. Since `setuptools` is also required for `vsc-base` itself (to define the `vsc` namespace that is spread across multiple directories), this is only possible without depending on `vsc-base`.

* If we manage to also drop `setuptools` as a dependency, there are no Python packages not included in the Python standard library that we require (although there are still several optional dependencies that are only required for specific functionality).

setup.py
```diff
@@ -118,11 +119,7 @@ def find_rel_test():
     platforms="Linux",
     test_suite="test.framework.suite",
     zip_safe=False,
-    install_requires=[
-        'setuptools >= 0.6',
-        "vsc-install >= 0.9.19",
-        "vsc-base >= 2.5.8",
-    ],
+    install_requires=['setuptools >= 0.6'],
```
wpoely86: so where are we still using it?

wpoely86:
It's big but I didn't find anything that caught my attention.

setup.py
boegel: `setuptools` is still uses for now, yes, that needs more work

easybuild/base/exceptions.py
```diff
@@ -0,0 +1,126 @@
...
+        finally:
+            # make very sure that reference to frame object is removed, to avoid reference cycles
+            # see https://docs.python.org/2/library/inspect.html#the-interpreter-stack
+            del frame
+
+    return logger
+
+
+class LoggedException(Exception):
+    """Exception that logs it's message when it is created."""
```
itkovian: `its`

PR #2713, opened 2019-01-10
move mk_rst_table function from easybuild.tools.docs to easybuild.tools.utilities

boegel:
`mk_rst_table` was thrown into `easybuild.tools.docs` when ingesting `vsc-base` (see #2708), but that causes some problems because it's needed in `easybuild.base.generaloption` when porting the modules in the `easybuild.base` namespace to Python 3 as a first step

It'll be easier to have `mk_rst_table` parked in `easybuild.tools.utilities` for now, and keep `easybuild.base` more isolated from `easybuild.tools` (only `easybuild.tools.utilities` is required by `easybuild.base.*` after this change)

### DIFF
--- a/easybuild/tools/utilities.py
+++ b/easybuild/tools/utilities.py
@@ -40,6 +40,7 @@ from easybuild.tools.config import build_option
 
 _log = fancylogger.getLogger('tools.utilities')
 
+INDENT_4SPACES = ' ' * 4
 
 # a list of all ascii characters
 ASCII_CHARS = string.maketrans('', '')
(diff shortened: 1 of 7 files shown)


easybuild/tools/utilities.py
```diff
@@ -40,6 +40,7 @@
 
 _log = fancylogger.getLogger('tools.utilities')
 
+INDENT_4SPACES = ' ' * 4
```
wpoely86: you add this but don't remove it from anywhere else?

boegel: Yeah, good point, fixed in cb82587

PR #2719, opened 2019-01-13
fix import with Python 3 for easybuild.tools.* modules and easybuild.main

boegel:


easybuild/base/fancylogger.py
```diff
@@ -277,7 +277,7 @@ def log_method(self, msg):
 
     # method definition as it is in logging, can't change this
     # pylint: disable=unused-argument
-    def makeRecord(self, name, level, pathname, lineno, msg, args, excinfo, func=None, extra=None):
+    def makeRecord(self, name, level, pathname, lineno, msg, args, excinfo, func=None, extra=None, sinfo=None):
```
wpoely86: some documentation on the arguments wouldn't go amiss

easybuild/tools/filetools.py
```diff
@@ -513,14 +513,15 @@ def download_file(filename, url, path, forced=False):
...
+    used_urllib = std_urllib
 
     while not downloaded and attempt_cnt < max_attempts:
         try:
-            if used_urllib is urllib2:
-                # urllib2 does the right thing for http proxy setups, urllib does not!
-                url_fd = urllib2.urlopen(url_req, timeout=timeout)
+            if used_urllib is std_urllib:
+                # urllib2 (Python 2) / urllib.request (Python 3) does the right thing for http proxy setups,
+                # urllib does not!
```
wpoely86: is this still relevant? Do we use urllib anywhere?

boegel: No, but we used to, and I guess the comment serves as a reminder why we use `urllib2` rather than `urllib`, doesn't hurt to keep it there imho.

easybuild/base/fancylogger.py
boegel: Yeah, I agree, but the Python documentation itself is rather light on this too...

This is only used internally in `fancylogger`, I'd rather not spend too much time on this.

boegel: See https://docs.python.org/2/library/logging.html#logging.Logger.makeRecord

PR #2720, opened 2019-01-14
fix check_pr_eligible_to_merge to only consider the status of the last test report

migueldiascosta:
without this, if the last report failed it does output

```last test report is successful: FAILED => not eligible for merging!```

but keeps looking for test reports until it finds a successful one (or runs out of comments).


boegel:
@migueldiascosta Thanks for the fix! Let's also enhance the existing test for `check_pr_eligible_to_merge` to catch this bug, done in https://github.com/migueldiascosta/easybuild-framework/pull/8

boegel:
Thanks for the fix @migueldiascosta!

PR #2721, opened 2019-01-14
consistently use 'string_type' from easybuild.tools.py2vs3 rather than 'basestring'

boegel:


migueldiascosta:
Going in, thanks @boegel!

Issue #2556, opened 2018-08-30
Support for buildininstalldir=True and multiple compilations


victorusu:
Easybuild always deletes the install dir, if one attempts to combine several compilations and buildinstalldir = True inside the same easyconfig.

Example
```
buildininstalldir=True
configopts = [
    " configopts1 ",
    " configopts2",
    " configopts3",
]
```

The installation process will perform the installation using `configopts1`, then using`configopts2`, then `configopts3`. But the `installdir` will only contain the build that used `configopts3`. 

Even if I set to the easyblock
```
def install_step(self):
    ...
    self.cfg["keeppreviousinstall"] = True
    self.cfg["cleanupoldbuild"] = False
    self.cfg["cleanupoldinstall"] = False
```

boegel:
fixed in #2724, thanks for reporting this @victorusu!

PR #2724, opened 2019-01-16
don't recreate build directory when buildininstalldir is enabled for iterative installations

boegel:
fixes for bug reported by @victorusu, see #2556

### DIFF
--- a/test/framework/easyblock.py
+++ b/test/framework/easyblock.py
@@ -872,11 +872,17 @@ class EasyBlockTest(EnhancedTestCase):
 
         # by default, make_builddir will re-create the build directory (i.e. remove existing & re-create)
         eb.make_builddir()
-        builddir_inode1 = os.stat(eb.builddir).st_ino
+        builddir = eb.builddir
+        testfile = os.path.join(builddir, 'test123', 'foobar.txt')
+        write_file(testfile, 'test123')
+        self.assertTrue(os.path.exists(testfile))
+
         eb.make_builddir()
-        builddir_inode2 = os.stat(eb.builddir).st_ino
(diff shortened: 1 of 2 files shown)


test/framework/easyblock.py
```diff
@@ -854,6 +855,53 @@ def test_gen_dirs(self):
...
 
+    def test_make_builddir(self):
+        """Test make_dir method."""
+        self.contents = '\n'.join([
+            'easyblock = "ConfigureMake"',
+            "name = 'pi'",
+            "version = '3.14'",
+            "homepage = 'http://example.com'",
+            "description = 'test easyconfig'",
+            "toolchain = {'name': 'dummy', 'version': 'dummy'}",
```
akesandgren: A bit of nitpicking, should version be '' or dummy here?

boegel: It doesn't matter at all here, since no `dependencies`

akesandgren: Yes, but since we're trying to move away from using 'dummy' as version...

akesandgren:
Going in, thanks @boegel!

test/framework/easyblock.py
boegel: We intend to get rid of `dummy` as a *toolchain* for EasyBuild v4.0, see https://github.com/easybuilders/easybuild/issues/447 + https://github.com/easybuilders/easybuild-framework/issues/1181 + https://github.com/easybuilders/easybuild-framework/pull/1233.

But it's better to do that all at once, little point in trying to avoid use of `dummy` in tests for EasyBuild 3.x

PR #2725, opened 2019-01-16
filter (CUDA) lib*/stubs paths in RPATH wrapper script

### LINKED ISSUE #2683: CUDA stubs directory must not be added to rpath/runpath
This is a follow up on https://github.com/easybuilders/easybuild-easyblocks/pull/1464 

 adding stubs to LIBRARY_PATH is causing us problems, because it ends up in our binaries RUNPATH.

If this happens and you don't use LD_LIBRARY_PATH (because you are using RUNPATH), you will have problems at runtime.

This is probably something you need to check when you are using runpath/rpath with easybuild (for Compute Canada, @bartoldeman had to adjust our linker wrapper to exclude these paths)

```
A stub object is a shared object, built entirely from mapfiles, that supplies the same linking interface as the real object, while containing no code or data. Stub objects cannot be used at runtime.
```

boegel:
fixes #2683

### DIFF
--- a/easybuild/scripts/rpath_args.py
+++ b/easybuild/scripts/rpath_args.py
@@ -98,12 +98,13 @@ while idx < len(args):
             cmd_args.append('-L%s' % lib_path)
 
     # replace --enable-new-dtags with --disable-new-dtags if it's used;
-    # --enable-new-dtags would result in copying rpath to runpath, meaning that $LD_LIBRARY_PATH is taken into account again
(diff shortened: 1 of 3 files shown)


bartoldeman:
lgtm, well tested!

bartoldeman:
oops, I clicked the wrong place.

boegel:
To be clear: tests actually passed (see also https://travis-ci.org/easybuilders/easybuild-framework/builds/480537908), but Travis wasn't happy with the rests being retriggered for a closed PR...

PR #2728, opened 2019-01-18
add support for %(arch)s template value in easyconfig files (x86-64, aarch64, ppc64le, ...)

boegel:
cfr. discussion in https://github.com/easybuilders/easybuild-easyconfigs/issues/6866


akesandgren:
Going in, thanks @boegel!

PR #2729, opened 2019-01-18
make filetools tests pass with Python 3

boegel:


easybuild/tools/multidiff.py
```diff
@@ -224,7 +227,7 @@ def limit(text, length):
 
         diff = False
         for i in range(len(self.base_lines)):
-            lines = filter(None, self.get_line(i))
+            lines = [line for line in self.get_line(i) if line]
```
wpoely86: is this exactly the same? Shouldn't you use `if line is not None`?

easybuild/tools/multidiff.py
```diff
@@ -169,10 +169,13 @@ def get_line(self, line_no):
                     # track meta info (which filenames are relevant)
                     changes_dict.setdefault(diff_line, set()).add(meta)
 
-            # sort: lines with most changes last, limit number to MAX_DIFF_GROUPS
-            lines = sorted(lines, key=lambda line: len(changes_dict[line]))[:MAX_DIFF_GROUPS]
+            # sort lines with most changes last;
+            # sort lines with equal number of changes alphabetically to ensure consistent output;
+            # limit number to MAX_DIFF_GROUPS
+            lines = sorted(lines, key=lambda line: (len(changes_dict[line]), line))[:MAX_DIFF_GROUPS]
```
wpoely86: I don't understand this? Your lambda returns a tuple? What does happen then by default if you compare tuples?

easybuild/tools/multidiff.py
boegel: No, with `None` as first element to `function` this is exactly equivalent, it's even mentioned explicitly to be so in the documentation: https://docs.python.org/2/library/functions.html#filter and https://docs.python.org/3/library/functions.html#filter

Issue #2731, opened 2019-01-21
eb --software fails with "Unable to find an easyconfig for the given specifications"


jose-d:
attempts to install software with "--software" switch fail with
`ERROR: Unable to find an easyconfig for the given specifications`

reproducible with framework + easyblocks on 3.8.0.

example:

```
[jose@koios1 ~]$ eb --software=Mathematica,11.3.0
== temporary log file in case of crash /tmp/eb-im2Vxx/easybuild-sFcqB3.log
ERROR: Unable to find an easyconfig for the given specifications: {'version': '11.3.0', 'name': 'Mathematica'}; to make EasyBuild try to generate a matching easyconfig, use the --try-X options 
[jose@koios1 ~]$
```

debug logs:
`eb -d --software=Mathematica,11.3.0`: [debuglog](https://www.fzu.cz/~jose/temp/mathematica_eb_install_debug.log)
`eb -d --software=git,2.19.1 --toolchain=foss,2018b`: [debuglog](https://www.fzu.cz/~jose/temp/git_eb_install_debug.log)

and stdout of:
`eb --search mathematica > ~/temp/mathematica_eb_search.stdout`: [stdout](https://www.fzu.cz/~jose/temp/git_eb_search.stdout)
`eb --search git > ~/temp/git_eb_search.stdout`[stdout](https://www.fzu.cz/~jose/temp/mathematica_eb_search.stdout)

ocaisa:
Hmm, from the mathematica debug log it looks like it's filtering on the wrong version.

For `git` the failure it is expected since there is no git with that toolchain (the toolchain for `2.19.1` is `GCCcore,7.3.0`). For this case, we could probably make the toolchain requirement softer and search through the entire hierarchy.

jose-d:
> For `git` the failure it is expected 
uh, then it's my fault of-course.

ocaisa:
No problem, it's reasonable to expect that to work (providing specs on the commandline should be as forgiving as possible).

ocaisa:
Ok, the initial problem should be fixed in #2732

jose-d:
> ..specs on the commandline should be as forgiving as possible..

I see, GCCcoreXXX is subset of fossNNNN.

ocaisa:
[...]

PR #2732, opened 2019-01-21
Sort available version list when selecting/generating easyconfig

ocaisa:
Fix for #2731

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -483,9 +483,12 @@ def select_or_generate_ec(fp, paths, specs):
     # TOOLCHAIN NAME
 
     # we can't rely on set, because we also need to be able to obtain a list of unique lists
-    def unique(lst):
+    def unique(lst, sortkey=None):
         """Retain unique elements in a sorted list."""
-        lst = sorted(lst)
+        if sortkey:
+            lst = sorted(lst, key=sortkey)
+        else:
+            lst = sorted(lst)
         if len(lst) > 1:
             res = [lst[0]]
             for x in lst:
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyconfig/tweak.py
```diff
@@ -389,12 +389,12 @@ def pick_version(req_ver, avail_vers):
...
             selected_ver = avail_vers[0]
         else:
-            retained_vers = [v for v in avail_vers if v <= LooseVersion(ver)]
+            retained_vers = [v for v in avail_vers if LooseVersion(v) <= LooseVersion(ver)]
             if retained_vers:
                 selected_ver = retained_vers[-1]
             else:
                 # if no versions are available that are less recent, take the least recent version
-                selected_ver = sorted([LooseVersion(v) for v in avail_vers])[0]
+                selected_ver = sorted(avail_vers, key=LooseVersion)[0]
```
ocaisa: This sorting isn't actually necessary with the updated unique function

ocaisa: @boegel, why was the original list the LooseVersion return values, this is causing a failing test now but I don't see a reason not to simply retain the version string

damianam:
This lgtm. But I haven't tested it. @ocaisa, @jose-d can you confirm that this change indeed fixes it?

easybuild/framework/easyconfig/tweak.py
boegel: That was most likely a bug, not a feature, since the other code paths in `pick_version` to return a string rather than a `LooseVersion` instance...

boegel:
Going in, thanks for figuring out the fix @ocaisa!

PR #2741, opened 2019-01-30
Make builddependencies iterable.

bartoldeman:
This is a list of list of buildependencies. It gets iterated on top
of the common builddependencies. This will allow e.g. to build with
multiple Python versions. A toy test is included.

The "prepare" step is now fully iterated which needed a few adjustments.
Hopefully this doesn't break anything.


bartoldeman:
I have now implemented proper module unloading. Hopefully it works fine with Tmod.

Actually the way things are now it may not be to hard after all to iterate over builddependencies itself.
There's some duplication but probably little time is spent vs the actual build.

bartoldeman:
Just in case, I've put a patch vs develop for builddependencies without the iterate_builddependencies here keyword here:
https://gist.github.com/d4f03ab150e65237f5cbdf501e93b818

I think it works well but the hiddendependencies code is getting rather cumbersome, so I'll work on a new PR to address that (I suppose we could set dep['hidden']=True instead of needing to remove them and then adding them back in the dump code).

bartoldeman:
#2748 cleans up hiddendependencies.

bartoldeman:
@boegel I am happy about this patch now. Can you have a look please when you have time?

boegel:
@bartoldeman I did some code cleanup for this, see https://github.com/ComputeCanada/easybuild-framework/pull/14

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -834,7 +854,11 @@ def builddependencies(self):
         """
         return the parsed build dependencies
         """
-        return self['builddependencies']
+        builddeps = self['builddependencies']
+        if 'builddependencies' in self.iterate_options and not isinstance(builddeps[0], dict):
+            # flatten if not iterating yet
+            builddeps = flatten(builddeps)
```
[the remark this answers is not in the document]
bartoldeman: in the beginning, before iterating, eb calls `process_easyconfig`, which also parses all dependencies. In this case you just need to parse them, not iterate them yet. The code is this:
```
            # add dependencies (including build & hidden dependencies)                                                                                                                                                                                  
            for dep in ec.dependencies():
                _log.debug("Adding dependency %s for app %s." % (dep, name))
                easyconfig['dependencies'].append(dep)
```
if you have any ideas how to make this more elegant, let me know :)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -834,7 +854,11 @@ def builddependencies(self):
         """
         return the parsed build dependencies
         """
-        return self['builddependencies']
+        builddeps = self['builddependencies']
+        if 'builddependencies' in self.iterate_options and not isinstance(builddeps[0], dict):
```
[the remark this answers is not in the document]
bartoldeman: Checking first element is sufficient, though `all` would be more thorough. Either all of them are dicts (after processing) or none of them. Perhaps either:
* use a seperate "flattendependencies()"
* or keep this but check if iteration started using a different method.
I am not sure now.

easybuild/framework/easyblock.py
```diff
@@ -863,10 +860,22 @@ def make_builddir(self):
...
 
         trace_msg("build dir: %s" % self.builddir)
 
+    def reset_changes(self):
+        """
+        Reset environment
+        """
+        env.reset_changes()
+        if self.reset_environ is None:
+            self.reset_environ = copy.deepcopy(os.environ)
```
boegel: @bartoldeman Please add a comment to explain why keeping tracking of the environment in `reset_environ` is needed, since that's not very obvious here...

easybuild/framework/easyconfig/easyconfig.py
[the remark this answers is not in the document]
boegel: @bartoldeman During today's conf call you mentioned that one option could be to create two separate methods for this, to clearly discriminate the two situations, and avoid mixing them?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1102,11 +1124,20 @@ def _finalize_dependencies(self):
...
         for key in DEPENDENCY_PARAMETERS:
             # loop over a *copy* of dependency dicts (with resolved templates);
-            # to update the original dep dict, we need to index with idx into self._config[key][0]...
-            for idx, dep in enumerate(self[key]):
+            deps = self[key]
+
+            # to update the original dep dict, we need to get a reference with templating disabled...
+            deps_ref = self.get_ref(key)
+
+            # take into account that this *dependencies parameters may be iterated over
```
boegel: `parameter`?

boegel:
@bartoldeman Problem with failing tests with Python 2.6 should be fixed with #2807, but seems like you broke something else as well.

boegel:
(close/re-open to trigger new Travis run now that #2807 is merged)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1102,11 +1133,20 @@ def _finalize_dependencies(self):
...
-            for idx, dep in enumerate(self[key]):
+            deps = self[key]
+
+            # to update the original dep dict, we need to get a reference with templating disabled...
+            deps_ref = self.get_ref(key)
+
+            # take into account that this *dependencies parameter may be iterated over
+            if key in self.iterate_options:
+                deps = flatten(deps)
+                deps_ref = flatten(deps_ref)
```
boegel: @bartoldeman This looks wrong to me... `flatten` will create a new list, so `deps_ref` is most definitely no longer a reference to the original list...

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -832,9 +852,20 @@ def dependencies(self, build_only=False):
...
         """
-        return self['builddependencies']
+        builddeps = self['builddependencies']
+        if 'builddependencies' in self.iterate_options and not self.iterating:
+            tmp = flatten(builddeps)
+            # remove duplicates
+            builddeps = []
+            for builddep in tmp:
+                if builddep not in builddeps:
+                    builddeps.append(builddep)
```
boegel: You're re-inventing the `nub` function here :)

easybuild/framework/easyblock.py
```diff
@@ -863,10 +860,25 @@ def make_builddir(self):
         # otherwise we wipe the already partially populated installation directory,
         # see https://github.com/easybuilders/easybuild-framework/issues/2556
         if not (self.build_in_installdir and self.iter_idx > 0):
+            # make sure we no longer sit in the build directory before cleaning it.
+            change_dir(self.orig_workdir)
             self.make_dir(self.builddir, self.cfg['cleanupoldbuild'])
 
         trace_msg("build dir: %s" % self.builddir)
 
+    def reset_changes(self):
```
boegel: `reset_env` makes more sense here imho...

easybuild/framework/easyconfig/tweak.py
```diff
@@ -820,21 +820,29 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
         # loop over a *copy* of dependency dicts (with resolved templates);
-        # to update the original dep dict, we need to index with idx into self._config[key][0]...
-        for idx, dep in enumerate(parsed_ec['ec'][key]):
+        # to update the original dep dict, we need to get a reference with templating disabled...
+        val = parsed_ec[key]
+        orig_val = parsed_ec.get_ref(key)
+
+        if key in parsed_ec.iterate_options:
+            val = flatten(val)
+            orig_val = flatten(orig_val)
```
boegel: @bartoldeman Same here, `flatten` creates a new list, so this is no longer a reference to the original list...

easybuild/framework/easyconfig/easyconfig.py
boegel: Actually, it's OK. `deps_ref` will be a new list, but the entries inside are still references to the original dependency `dict` values that we aim to modify, so it's all good...

easybuild/framework/easyconfig/tweak.py
boegel: Same here, it's OK to use `flatten` here, since we're still using references to the original `dict` values, and that's what wer'e changing (not the list itself).

boegel:
Tested with modified `Eigen` easyconfig, which lists two different versions of `CMake` (there's no point in that, other than testing the support for iterating of a list of lists of build dependencies implemented here):

```
name = 'Eigen'
version = '3.3.7'

homepage = 'http://%(namelower)s.tuxfamily.org/index.php?title=Main_Page'

description = """Eigen is a C++ template library for linear algebra: matrices, vectors, numerical solvers,
 and related algorithms."""

# only includes header files, so no need for a non-dummy toolchain
toolchain = {'name': 'dummy', 'version': ''}

source_urls = [BITBUCKET_SOURCE]
sources = ['%(version)s.tar.bz2']
checksums = ['9f13cf90dedbe3e52a19f43000d71fdf72e986beb9a5436dddcd61ff9d77a3ce']

# stick to latest CMake 3.9.x, since more recent CMake versions require a C++ compiler that supports C++11,
# which may not be available yet in older OSs (e.g. CentOS 6.x)
builddependencies = [
    [('CMake', '3.9.6')],
    [('CMake', '3.12.1')],
]

moduleclass = 'math'
```

Works as intended, so I consider this good to go.

Thanks you very much for your hard work on this @bartoldeman, I'm well aware that this was far from trivial...

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -539,13 +540,28 @@ def parse(self):
 
         run_hook(PARSE, hooks, args=[self], msg=parse_hook_msg)
 
+        # create a list of all options that are actually going to be iterated over
+        # builddependencies are always a list, need to look deeper down below
+        self.iterate_options = [opt for opt in ITERATE_OPTIONS
+                                if opt != 'builddependencies' and isinstance(self[opt], (list, tuple))]
```
boegel: @bartoldeman I only realize now that this is done too early...

We're assuming here that once the easyconfig file is parsed, we know whether which easyconfig parameters are to be iterated over. That breaks backwards compatibility in some sense, since before this change we had the ability to tweak the easyconfig parameter values before they were actually being used...

And that's exactly what we're doing with `FFTW`, where we define the list of configure options to iterate over in `self.cfg['configopts']` via `run_all_steps`...
That value is not picked up for defining `self.iterate_options`, and hence stuff is broken.

I think we'll need to postpone this to avoid breaking backwards compatibility (yet still handle `builddependencies` here)...

boegel: problem should be fixed with #2810

PR #2746, opened 2019-02-01
use assertEqual rather than assertEquals (which is deprecated)

boegel:
see https://docs.python.org/2/library/unittest.html#deprecated-aliases


akesandgren:
Going in, thanks @boegel!

PR #2748, opened 2019-02-01
Replace deps by hidden deps instead of removing them from the lists.

bartoldeman:
This a cleanup so that hidden deps do not need to be re-added when
dumping the easyconfig. The build only needs to consider
builddependencies and dependencies, with integrated hidden
dependencies.

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -697,16 +697,15 @@ class EasyConfig(object):
                         # see comments in resolve_template
                         enable_templating = self.enable_templating
                         self.enable_templating = False
-
                         # track whether this hidden dep is listed as a build dep
                         orig_hidden_dep = self['hiddendependencies'][hidden_idx]
                         if key == 'builddependencies':
                             orig_hidden_dep['build_only'] = True
                         # actual replacement
                         self[key][idx] = orig_hidden_dep
-                        replaced = True
-
                         self.enable_templating = enable_templating
+
+                        replaced = True
                         if dep_mod_name == visible_mod_name:
                             msg = "Replaced (build)dependency matching hidden dependency %s"
                         else:
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -174,7 +174,7 @@ def det_subtoolchain_version(current_tc, subtoolchain_name, optional_toolchains,
 
 @toolchain_hierarchy_cache
 def get_toolchain_hierarchy(parent_toolchain, incl_capabilities=False):
-    """
+    r"""
     Determine list of subtoolchains for specified parent toolchain.
     Result starts with the most minimal subtoolchains first, ends with specified toolchain.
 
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -677,33 +677,36 @@ def validate_iterate_opts_lists(self):
...
-                    self[key] = [d for d in self[key] if d['full_mod_name'] != visible_mod_name]
-                self.log.debug("Removed (build)dependency matching hidden dependency %s", hidden_dep)
+                    self[key] = [d if d['full_mod_name'] != visible_mod_name else orig_hidden_dep for d in self[key]]
+                self.log.debug("Replaced (build)dependency matching hidden dependency %s", hidden_dep)
             elif hidden_mod_name in dep_mod_names:
                 for key in ['builddependencies', 'dependencies']:
-                    self[key] = [d for d in self[key] if d['full_mod_name'] != hidden_mod_name]
+                    self[key] = [d if d['full_mod_name'] != hidden_mod_name else orig_hidden_dep for d in self[key]]
                 self.log.debug("Hidden (build)dependency %s is already marked to be installed as a hidden module",
                                hidden_dep)
```
boegel: @bartoldeman The list comprehension with `if... else` makes this hard to read, and there's a lot of code duplication here as well.

How about something like this:

```python
# replace (build) dependencies with their equivalent hidden (build) dependency (if any) 
for key in ['builddependencies', 'dependencies']:
    deps = []
    for dep in self[key]:
        dep_mod_name = dep['full_mod_name']
        # FIXME: explain
        if visible_mod_name in dep_mod_names and dep_mod_name != visible_mod_name:
            deps.append(dep)
        # FIXME: explain
        elif hidden_mod_name in dep_mod_names and dep_mod_name != hidden_mod_name:
            deps.append(dep)
        # FIXME: explain fallback case (this is where deps are actually replaced by the equivalent hidden dep?)
        else:
            deps.append(orig_hidden_dep)
    self[key] = deps
    self.log.debug("<sensible log message goes here>")

Comment (bartoldeman) [reply easybuild/framework/easyconfig/easyconfig.py]: I have completely reorganized this code now to be iterative and replace in-place.

File: easybuild/framework/easyconfig/easyconfig.py [hunk 2]
```diff
@@ -677,43 +677,51 @@ def validate_iterate_opts_lists(self):
...
+
+                        # track whether this hidden dep is listed as a build dep
+                        orig_hidden_dep = self['hiddendependencies'][hidden_idx]
+                        if key == 'builddependencies':
+                            orig_hidden_dep['build_only'] = True
+                        # actual replacement
+                        self[key][idx] = orig_hidden_dep
+                        replaced = True
+
+                        self.enable_templating = enable_templating
```
boegel: @bartoldeman Style nitpicking: do this in the block above, after the `self[key][idx]`, since that's where it actually matters?

So, drop empty line above, add one below this line, to make it a bit less confusing.

boegel:
Thanks for the effort @bartoldeman!

PR #2756, opened 2019-02-08
fix running of build_log/config/containers/docs tests with Python 3

boegel:


akesandgren:
Going in, thanks @boegel!

PR #2757, opened 2019-02-10
fix undefined variable 'prefix' in Compiler._set_compiler_vars

boegel:


akesandgren:
Going in, thanks @boegel!

PR #2760, opened 2019-02-12
fix test account for GitHub-related tests

boegel:
This slipped in unintended with #2700, and it effectively disabled the tests for the GitHub integration features. The tests still pass though, so no damage done luckily...


migueldiascosta:
Going in, thanks @boegel!

PR #2761, opened 2019-02-12
fix running of github tests with Python 3

boegel:


test/framework/github.py
```diff
@@ -97,7 +96,7 @@ def test_read_api(self):
             return
 
         try:
-            self.assertEqual(self.ghfs.read("a_directory/a_file.txt").strip(), "this is a line of text")
+            self.assertEqual(self.ghfs.read("a_directory/a_file.txt").strip(), b"this is a line of text")
```
akesandgren: In the previous changes you removed the "b" in front of some strings. Since I still don't know enough python, what does that "b" do?

boegel: `b"..."` is a byte-string in Python 3 (as opposed to `"..."` which is a regular string of Unicode characters); see also https://docs.python.org/3/library/stdtypes.html#bytes.
In Python 2, `b"..."` is also accepted, but there's no real difference from `"..."` (exact same type of value).

I'm going with the easiest way of fixing the test here for now, but I may need to revise that later by making that `read` method return a regular string rather than a byte-string.

But for now, it seems like there's no need to change that, since all tests for GitHub-related features pass.

akesandgren:
Going in, thanks @boegel!

PR #2763, opened 2019-02-13
make REST client actually return headers when http method is HEAD

migueldiascosta:
ported from https://github.com/hpcugent/vsc-base/pull/276

(not sure where to put the test now)

### DIFF
--- a/easybuild/base/rest.py
+++ b/easybuild/base/rest.py
@@ -170,11 +170,14 @@ class Client(object):
         # TODO: in recent python: Context manager
         conn = self.get_connection(method, url, body, headers)
         status = conn.code
-        body = conn.read()
-        try:
-            pybody = json.loads(body)
-        except ValueError:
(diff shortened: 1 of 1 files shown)


boegel:
@migueldiascosta W.r.t. your question on the test: the tests for the modules ingested from `vsc-base` were not copied into the framework repository.

I think that's OK for now, since we will have tests for the EasyBuild functionality that relies on the `easybuild.base.*` modules anywhere.

Another option is to start a new `test/framework/base.py` test suite to cover changes/fixes made in `easybuild.base.*`.

In this case however, the change made here well eventually be covered (indirectly) via #2553, so I think it's fine.

migueldiascosta:
@boegel yeah, those were precisely the options I was wondering about, and I agree that either way it should be fine

boegel:
Thanks for the backport @migueldiascosta!

PR #2765, opened 2019-02-14
fix running of 'asyncprocess' & 'run' test suite with Python 3

boegel:
more string vs bytes trouble fixed in Python 3, while functionality in Python 2 is unaffected...


akesandgren:
Going in, thanks @boegel!

PR #2766, opened 2019-02-14
fix running of 'easyconfig' test suite with Python 3

boegel:


easybuild/framework/easyconfig/tweak.py
```diff
@@ -270,7 +270,7 @@ def __repr__(self):
         _log.warning("Tweaking version: checksums cleared, verification disabled.")
 
     # we need to treat list values separately, i.e. we prepend to the current value (if any)
-    for (key, val) in tweaks.items():
+    for (key, val) in list(tweaks.items()):
```
bartoldeman: This is another case of a double copy for Python 2. You can use
```
for key in tweaks:
     val = tweaks[key]
```

boegel: The reason I have to make this change is because there's a `tweaks.pop(...)` below, which is why I need a *copy* of the dict elements, rather than a reference (which is what I get in Python 3 with `.items()`).

So `for key in tweaks` isn't going to work, same problem. And `for key in list(tweaks):` is only marginally better?

bartoldeman: `for key in list(tweaks):` is better then, as that's a single copy in Python 2, unlike `list(...items())` which is a double copy in Python 2.

boegel: OK, fixed in 899c526df

PR #2768, opened 2019-02-15
fix running of 'robot' test suite with Python 3

boegel:


migueldiascosta:
Going in, thanks @boegel!

PR #2772, opened 2019-02-16
make test for dep_graph function robust against changing order of lines in resulting dot file

boegel:
In some circumstances, the order of lines in the `*.dot` file created via `dep_graph` is slightly different, e.g.:

```
digraph graphname {
toy;
"GCC/6.4.0-2.28 (EXT)";
intel;
toy -> intel;
toy -> "GCC/6.4.0-2.28 (EXT)";
}
```

vs

```
digraph graphname {
toy;
intel;
"GCC/6.4.0-2.28 (EXT)";
toy -> intel;
toy -> "GCC/6.4.0-2.28 (EXT)";
}
```

The difference doesn't matter though, since it represents the same dependency graph.


akesandgren:
Going in, thanks @boegel!

PR #2773, opened 2019-02-16
ensure non-zero exit code for all test subsuites (+ fix trivial style issues)

boegel:
Running a test subsuite always results in a zero exit code (suggesting success), regardless of whether any tests failed or not, which is stupid/annoying.

I fixed a whole bunch of style issues in the test modules while I was fixing this, which makes this PR look scary, but the changes are all trivial...

No tests were harmed in the process, and all changes are limited to `test/*`

### DIFF
--- a/test/framework/type_checking.py
+++ b/test/framework/type_checking.py
@@ -297,7 +297,7 @@ class TypeCheckingTest(EnhancedTestCase):
                               to_dependency, {'toolchain': 'lib, 1.2.8', 'versionsuffix': 'suff'})
         # too many values
         dep_spec = {'lib': '1.2.8', 'foo': '1.3', 'toolchain': 'lib, 1.2.8', 'versionsuffix': 'suff'}
-        self.assertErrorRegex(EasyBuildError, "Found unexpected \(key, value\) pair: .*", to_dependency, dep_spec)
+        self.assertErrorRegex(EasyBuildError, r"Found unexpected \(key, value\) pair: .*", to_dependency, dep_spec)
 
     def test_to_dependencies(self):
         """Test to_dependencies function."""
--- a/test/framework/type_checking.py
+++ b/test/framework/type_checking.py
@@ -257,18 +257,19 @@ class TypeCheckingTest(EnhancedTestCase):
         lib_dict.update({'versionsuffix': ''})
 
         # to_dependency doesn't touch values of non-dict type
-        self.assertEqual(to_dependency(('foo', '1.3')), ('foo','1.3'))
-        self.assertEqual(to_dependency(('foo', '1.3', '-suff', ('GCC', '4.8.2'))), ('foo', '1.3', '-suff', ('GCC','4.8.2')))
+        self.assertEqual(to_dependency(('foo', '1.3')), ('foo', '1.3'))
+        expected = ('foo', '1.3', '-suff', ('GCC', '4.8.2'))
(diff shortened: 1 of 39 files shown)


test/framework/type_checking.py
```diff
@@ -257,18 +257,19 @@ def test_to_dependency(self):
         lib_dict.update({'versionsuffix': ''})
 
         # to_dependency doesn't touch values of non-dict type
-        self.assertEqual(to_dependency(('foo', '1.3')), ('foo','1.3'))
-        self.assertEqual(to_dependency(('foo', '1.3', '-suff', ('GCC', '4.8.2'))), ('foo', '1.3', '-suff', ('GCC','4.8.2')))
+        self.assertEqual(to_dependency(('foo', '1.3')), ('foo', '1.3'))
+        expected = ('foo', '1.3', '-suff', ('GCC', '4.8.2'))
+        self.assertEqual(to_dependency(('foo', '1.3', '-suff', ('GCC', '4.8.2'))), expected)
```
akesandgren: isn't this just self.assertEqual(to_dependency(expected), expecter) ?? And the same for all of them...
Or am I blind??

test/framework/type_checking.py
```diff
@@ -293,10 +294,10 @@ def test_to_dependency(self):
 
         # no name/version
         self.assertErrorRegex(EasyBuildError, "Can not parse dependency without name and version: .*",
-            to_dependency, {'toolchain': 'lib, 1.2.8', 'versionsuffix': 'suff'})
+                              to_dependency, {'toolchain': 'lib, 1.2.8', 'versionsuffix': 'suff'})
         # too many values
-        self.assertErrorRegex(EasyBuildError, "Found unexpected \(key, value\) pair: .*",
-            to_dependency, {'lib': '1.2.8', 'foo':'1.3', 'toolchain': 'lib, 1.2.8', 'versionsuffix': 'suff'})
+        dep_spec = {'lib': '1.2.8', 'foo': '1.3', 'toolchain': 'lib, 1.2.8', 'versionsuffix': 'suff'}
+        self.assertErrorRegex(EasyBuildError, r"Found unexpected \(key, value\) pair: .*", to_dependency, dep_spec)
```
akesandgren: There is a subtle difference here, the original does not have r"Found... but only "Found...
How much of a difference does this result in in reality?

test/framework/type_checking.py
boegel: No, you're right, we're checking here whether `to_dependency` leaves correct dependency specs as they are (see comment above)

test/framework/type_checking.py
boegel: The `r"..."` only makes it a raw string, which is necessary to shut up the Hound w.r.t. the escaped `(` and `)` characters.

This check verifies whether the produced exception matches the specified pattern, so if the test passes (which it does), all is good.

test/framework/type_checking.py
akesandgren: So why not use to_dep(exp), exp)? much easier to grasp.

test/framework/type_checking.py
akesandgren: ok, just checking.

test/framework/type_checking.py
boegel: Good point, will change...

akesandgren:
Going in, thanks @boegel!

PR #2774, opened 2019-02-16
test all configuration with Python 3.6

boegel:
extend testing with Python 3.6 with all different configurations


akesandgren:
Going in, thanks @boegel!

PR #2775, opened 2019-02-18
fix running of 'easyblock' test suite with Python 3

boegel:


easybuild/tools/version.py
```diff
@@ -61,8 +61,10 @@ def get_git_revision():
     try:
         path = os.path.dirname(__file__)
         gitrepo = git.Git(path)
-        # 'encode' is required to make sure a regular string is returned rather than a unicode string
-        res = gitrepo.rev_list('HEAD').splitlines()[0].encode('ascii')
+        res = gitrepo.rev_list('HEAD').splitlines()[0]
+        # 'encode' is required to make sure a regular string is returned rather than a unicode string (only in Python 2)
```
bartoldeman: comment is a little strange now since regular strings are unicode strings in Python 3.

boegel: that why I added the `(only in Python 2)`

boegel:
Hmm, seems like the Travis notification didn't come through, closing & re-opening to re-trigger Travis...

PR #2776, opened 2019-02-20
handle github api pagination when present

migueldiascosta:
alternative to #2553 

needs `vsc-base` 2.8.4

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -243,10 +243,32 @@ def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
     url = request_f(RestClient(GITHUB_API_URL, username=github_user, token=token))
 
     try:
-        status, data = url.get(**kwargs)
+        status, headers = url.head(**kwargs) 
     except socket.gaierror as err:
         _log.warning("Error occurred while performing get request: %s", err)
-        status, data = 0, None
+        status, headers = 0, None
+
+    links = headers.getheader('Link')
+
+    if links:
+        match = re.search(r"page=(\d+)>; rel=\"last\"", links)
+        last = int(match.groups()[0]) if match else 1
+        data = []
+        for page in range(1, last+1):
+            kwargs['page'] = page
+            try:
+                status, page_data = url.get(**kwargs)
+            except socket.gaierror as err:
+                _log.warning("Error occurred while performing get request: %s", err)
+                status, page_data = 0, None
+                break
+            data.extend(page_data)
+    else:
+        try:
+            status, data = url.get(**kwargs)
+        except socket.gaierror as err:
+            _log.warning("Error occurred while performing get request: %s", err)
+            status, data = 0, None
 
     _log.debug("get request result for %s: status: %d, data: %s", url, status, data)
     return (status, data)


migueldiascosta:
I'm not particularly happy about the parsing of `Links` and traversing of pages here, it's certainly not pretty and I don't know if it is sufficiently robust. (It would be much cleaner using `requests`, but I suppose we are avoiding adding it as a dependency.) Open to suggestions/corrections.

easybuild/tools/github.py
```diff
@@ -243,10 +243,32 @@ def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
...
+                _log.warning("Error occurred while performing get request: %s", err)
+                status, page_data = 0, None
+                break
+            data.extend(page_data)
+    else:
+        try:
+            status, data = url.get(**kwargs)
+        except socket.gaierror as err:
+            _log.warning("Error occurred while performing get request: %s", err)
+            status, data = 0, None
```
boegel: @migueldiascosta We can avoid quite a bit of duplicate code here:

```python
if links:
    ...
    pages = range(1, last+1)
else:
    pages = [1]

data = []
for page in pages:
    try:
        status, page_data = url.get(**kwargs)
    except socket.gaierror as err:
        _log.warning("Error occurred while performing get request: %s", err)
        status, page_data = 0, []
        break
    data.extend(page_data)

File: easybuild/tools/github.py [hunk 2]
```diff
@@ -243,10 +243,32 @@ def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
...
 
     try:
-        status, data = url.get(**kwargs)
+        status, headers = url.head(**kwargs) 
     except socket.gaierror as err:
         _log.warning("Error occurred while performing get request: %s", err)
-        status, data = 0, None
+        status, headers = 0, None
+
+    links = headers.getheader('Link')
```
boegel: @migueldiascosta This will fail miserably if `headers` was set to `None` above...

easybuild/tools/github.py
migueldiascosta: not all API endpoints return lists, we would need something like 

```python
if isinstance(page_data, list):
    data.extend(page_data)
else:
    data = page_data
```
?

easybuild/tools/github.py
migueldiascosta: @boegel tfw you realise you created two branches at different times to work on the same thing and pushed the older one...

easybuild/tools/github.py
boegel: Ugh, what a mess... :)

This works too:
```python
if not isinstance(page_data, list):
    page_data = [page_data]
data.extend(page_data)

File: easybuild/tools/github.py (hunk 2 above)
Comment (boegel) [reply easybuild/tools/github.py]: hehe

File: easybuild/tools/github.py [hunk 3]
```diff
@@ -243,10 +243,34 @@ def github_api_get_request(request_f, github_user=None, token=None, **kwargs):
...
+            status, page_data = url.get(**kwargs)
+        except socket.gaierror as err:
+            _log.warning("Error occurred while performing get request: %s", err)
+            status, page_data = 0, None
+            break
+        if isinstance(page_data, list):
+            data.extend(page_data)
+        else:
+            data = page_data
+            break
```
boegel: Hmm, why the break? Aren't we ignore all remaining pages that way?

easybuild/tools/github.py
migueldiascosta: @boegel I think this will break the uses of `github_api_get_request` that weren't getting a list to begin with

easybuild/tools/github.py
migueldiascosta: @boegel well, if the output of `get` was not a list, it certainly isn't paginated

migueldiascosta: @boegel in a sense, the initial version with some code duplication was clearer about these two cases...

boegel: If there is no pagination, we would only have `pages = [1]`, right?

migueldiascosta: right :)

wpoely86:
The code on which the rest code from vsc-base is based on has full pagination support since the beginning of this year: https://github.com/mozilla/agithub/pull/49

So we might as well update the code in vsc-base and use that?

PR #2781, opened 2019-02-21
silence printed deprecation warnings for easyconfigs/toolchains based on 'silent' build option

boegel:
This should get rid of the stream of `WARNING: Deprecated functionality` messages being printed in the easyconfigs tests, while still printing them during normal use of `eb`.

### DIFF
--- a/test/framework/build_log.py
+++ b/test/framework/build_log.py
@@ -109,6 +109,7 @@ class BuildLogTest(EnhancedTestCase):
         log.deprecated("anotherwarning", newer_ver)
         log.deprecated("onemorewarning", '1.0', '2.0')
         log.deprecated("lastwarning", '1.0', max_ver='2.0')
+        log.deprecated("thisisnotprinted", '1.0', max_ver='2.0', silent=True)
         log.error("kaput")
         log.error("err: %s", 'msg: %s')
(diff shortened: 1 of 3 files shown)


test/framework/build_log.py
```diff
@@ -138,6 +139,7 @@ def test_easybuildlog(self):
             r"%s.test_easybuildlog \[WARNING\] :: Deprecated functionality.*anotherwarning.*" % root,
             r"%s.test_easybuildlog \[WARNING\] :: Deprecated functionality.*onemorewarning.*" % root,
             r"%s.test_easybuildlog \[WARNING\] :: Deprecated functionality.*lastwarning.*" % root,
+            r"%s.test_easybuildlog \[WARNING\] :: Deprecated functionality.*thisisnotprinted.*" % root,
```
ocaisa: Are you missing the `silent` test where this shouldn't be printed?

boegel: No, it's there already...

The test checks both the messages printed to stderr (which haven't changed), and the messages included in the log file (which gets one additional line).

The log message isn't silenced, the one that is printed to stderr is.

ocaisa: Ah ok

Issue #2771, opened 2019-02-15
Warnings during tests


surak:
Are they expected? How can one have a 'clean' test?

```
WARNING: Failed to determine toolchain hierarchy for GCC/6.4.0-2.28 when determining subtoolchain for dependency 'hwloc': 'Could not find easyconfig for GCC toolchain version 6.4.0-2.28'

.Skipping test_dep_graph, since pygraph is not available
.
WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'first': 'Could not find easyconfig for GCC toolchain version 4.6.3'


WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'second': 'Could not find easyconfig for GCC toolchain version 4.6.3'


WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'first': 'Could not find easyconfig for GCC toolchain version 4.6.3'


WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'second': 'Could not find easyconfig for GCC toolchain version 4.6.3'


WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'foo': 'Could not find easyconfig for GCC toolchain version 4.6.3'


WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'bar': 'Could not find easyconfig for GCC toolchain version 4.6.3'

.....Skipping test_dump_autopep8, since autopep8 is not available
./tmp/eb-7eG3Bv/eb-JcnMcd/eb-yxEReVFWTgiE.eb:24:19: W292 no newline at end of file
........
WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'first': 'Could not find easyconfig for GCC toolchain version 4.6.3'


WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'second': 'Could not find easyconfig for GCC toolchain version 4.6.3'


WARNING: Failed to determine toolchain hierarchy for GCC/4.6.3 when determining subtoolchain for dependency 'first': 'Could not find easyconfig for GCC toolchain version 4.6.3'


[...]

PR #2782, opened 2019-02-21
avoid printing of messages/warnings in framework test suite

boegel:
The `silent` build option is only set to `True` when the tests are run (and since there's no direct option for it, it can't be controlled easily during normal use of `eb`)

fixes complaint made in #2771

### DIFF
--- a/test/framework/lib.py
+++ b/test/framework/lib.py
@@ -54,7 +54,8 @@ class EasyBuildLibTest(EnhancedTestCase):
         super(EasyBuildLibTest, self).setUp()
 
         # make sure BuildOptions instance is re-created
-        del BuildOptions._instances[BuildOptions]
+        if BuildOptions in BuildOptions._instances:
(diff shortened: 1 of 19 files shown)


test/framework/lib.py
```diff
@@ -54,7 +54,8 @@ def setUp(self):
         super(EasyBuildLibTest, self).setUp()
 
         # make sure BuildOptions instance is re-created
-        del BuildOptions._instances[BuildOptions]
+        if BuildOptions in BuildOptions._instances:
+            del BuildOptions._instances[BuildOptions]
```
akesandgren: Shouldn't the "del BuildOptions..." in configure further down be protected too just in case?

boegel: sure, why not, fixed in 852cdd3

akesandgren:
Going in, thanks @boegel!

PR #2783, opened 2019-02-22
add support for reverse dependencies of installed modules (WIP)

migueldiascosta:


easybuild/tools/docs.py
```diff
@@ -707,6 +708,44 @@ def list_software_txt(software, detailed=False):
...
 
+def gather_reverse_dependencies():
+    """
+    Gather reverse dependencies from installed modules.
+
+    :return: dictionary with reverse dependencies
+    """
+    reverse_dependencies = {}
+
+    module_avail_cmd = "module -t avail 2>&1 | grep -v ':$'"
```
ocaisa: How would this work with a hierarchical scheme (or any scheme that involved extending the module path)?

easybuild/tools/docs.py
```diff
@@ -707,6 +708,44 @@ def list_software_txt(software, detailed=False):
...
+    module_purge_load_list_cmd = "module purge; module load %s; module -t list 2>&1 | grep -v %s"
+
+    available_modules, _ = run_cmd(module_avail_cmd)
+
+    for module in available_modules.splitlines():
+        dependencies, _ = run_cmd(module_purge_load_list_cmd % (module,module))
```
boegel: @migueldiascosta probably a good idea to use `trace=False` here, otherwise you get very amusing output when using `eb --trace` (which is default for me)

easybuild/tools/docs.py
```diff
@@ -707,6 +708,44 @@ def list_software_txt(software, detailed=False):
...
+def gather_reverse_dependencies():
+    """
+    Gather reverse dependencies from installed modules.
+
+    :return: dictionary with reverse dependencies
+    """
+    reverse_dependencies = {}
+
+    module_avail_cmd = "module -t avail 2>&1 | grep -v ':$'"
+    module_purge_load_list_cmd = "module purge; module load %s; module -t list 2>&1 | grep -v %s"
```
boegel: Using `purge` is troublesome on some systems, and fairly expensive with Lmod...

How about just running this in a subshell, since we only care about the output, not the changes to the environment?

Or use `ml show` (which only gives direct deps though), that would also make it significantly faster...

easybuild/tools/docs.py
boegel: That shouldn't be a problem when using `modules_tool().available()`, which spits out a list of *full* module names in a hierarchy (i.e. things like `Core/GCC/6.4.0` rather than just `GCC/6.4.0`), so you'll see *all* modules regardless of where they are in the hierarchy...

ocaisa:
I was thinking about this recently but with a very different approach, you can generate a full dependency graph for your installation tree with
```
eb -D --dep-graph=depgraph.dot $EASYBUILD_INSTALLPATH/software/*/*/easybuild/*.eb
```
but it only works with certain caveats
- You use a fixed install dir naming scheme (or use `find` which is very expensive)
- There's no mechanism to check that the software is actually available in the module tree
- It doesn't take into account user installed modules (but could easily)
- It is probably pretty slow

ocaisa:
> I was thinking about this recently but with a very different approach, you can generate a full dependency graph for your installation tree with
> 
> ```
> eb -D --dep-graph=depgraph.dot $EASYBUILD_INSTALLPATH/software/*/*/easybuild/*.eb
> ```
> but it only works with certain caveats
> 
> * You use a fixed install dir naming scheme (or use `find` which is very expensive)
> * There's no mechanism to check that the software is actually available in the module tree (but it wouldn't be hard to build)
> * It doesn't take into account user installed modules (but could easily)
> * It is probably pretty slow

I tried it for a software in a stage of JURECA, and it (currently) doesn't work, graphdot throws errors like
```
pygraph.classes.exceptions.AdditionError: Edge (Core/JUnit/4.12, Core/Java/1.8) already in digraph
```

ocaisa:
Sorry for bombing your PR but with a small tweak to tools.py
```
            try:
               dgr.add_edge((spec['module'], dep))
            except Exception as e:
               if "already in digraph" in str(e):
                   continue
               else:
                  raise e
```
my method succeeds and I get
```
[ocais1@jrl01 temp]$ time eb --detect-loaded-modules=ignore --robot=./ --dep-graph=depgraph.dot *.eb
== temporary log file in case of crash /tmp/eb-ogaUb9/easybuild-YAPyiN.log
== resolving dependencies ...
Wrote dependency graph for 1238 easyconfigs to depgraph.dot

real	1m42.163s
user	1m26.528s
sys	0m10.211s
```

boegel:
@ocaisa One of the design principles here, which we also briefly discussed during the user meeting, is that the existing module files present the current "state" of things, and that's really all the matters from a users perspective.

Your approach is interesting, but not with the feature we have in mind that will be implemented on top of this (i.e. `eb --remove`).

ocaisa:
But it would be easy to prune out non-installed easyconfigs since you have this info from the functionality of `-D`.
I kind of see where you are coming from but my approach allows you to see where something fits in the global tree, and even visualise it, which is exactly what you need with `--remove`.

migueldiascosta:
@ocaisa bomb away :), this WIP PR was created for precisely for discussing the issue

My original approach a while back was also based on easyconfigs in the robot path, and I still think there's a use case for that, but since then I came to agree with @boegel that for uninstall purposes the installed modules are the source of truth from the perspective of the users. For instance, there may be modules that were not created by easybuild but depend on modules created by easybuild, and we should know about those before removing a module created by easybuild, or even modules may have been modified after being created by easybuild...

Maybe there's a case for having both "easyconfig reverse dependencies", e.g. for removing easyconfigs from a repository, and "module reverse dependencies", e.g. for removing installed software?

As to this particular implementation of "module reverse dependencies", it's *very* slow, I'll look into using module show instead of module load + list to speed it up

migueldiascosta:
@boegel by caching dependencies and, crucially, using `$MODULEPATH` + full module name to get the modulefile path instead of using `module show`, which is the default behaviour, this is now pretty fast (dependencies of ~2000 modules in ~20s, on pretty old and slow system and filesystem), the question is if it is correct in all cases, particularly different module tools and naming schemes...

easybuild/tools/docs.py
```diff
@@ -61,6 +61,7 @@
 from easybuild.tools.filetools import read_file
 from easybuild.tools.modules import modules_tool
 from easybuild.tools.ordereddict import OrderedDict
+from easybuild.tools.run import run_cmd
```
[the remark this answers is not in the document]
boegel: @migueldiascosta No need to import `run_cmd` if you're not going to use it?

easybuild/tools/docs.py
```diff
@@ -707,6 +710,66 @@ def list_software_txt(software, detailed=False):
...
+            for full_mod_name in all_reverse_deps:
+                if full_mod_name.startswith(mod_name):
+                    output_mod_names.append(full_mod_name)
+
+    if not output_mod_names:
+        return "No reverse dependencies found for %s" % ', '.join(mod_names)
+    else:
+        return '\n'.join(["%s: %s" % (mod_name, reverse_deps)
+                          for mod_name, reverse_deps in all_reverse_deps.items()
+                          if mod_name in output_mod_names])
```
boegel: This can be simplified using:

```python
return '\n'.join(["%s: %s" % (m, all_reverse_deps[m]) for m in output_mod_names])
```

We should also ensure sorted output?

easybuild/tools/docs.py
```diff
@@ -707,6 +710,66 @@ def list_software_txt(software, detailed=False):
...
+    """
+    Show reverse dependencies of installed software
+
+    :param output_format: output format to use
+    :return: multi-line string presenting requested info
+    """
+
+    all_reverse_deps = gather_reverse_dependencies()
+
+    # TODO: format output
```
boegel: No longer TODO?

easybuild/tools/docs.py
```diff
@@ -707,6 +710,66 @@ def list_software_txt(software, detailed=False):
...
+    available_modules = modtool.available()
+
+    for idx, mod_name in enumerate(available_modules):
+
+        mod_deps = dependencies_for(mod_name, modtool, alldeps=dependencies, full_module_names=True)
+
+        for dep_mod_name in mod_deps:
+            if dep_mod_name not in reverse_dependencies:
+                reverse_dependencies[dep_mod_name] = []
+            reverse_dependencies[dep_mod_name].append(mod_name)
```
boegel: This can be simplified as:

```python
reverse_dependencies.set_default(dep_mod_name, []).append(mod_name)

File: easybuild/tools/module_generator.py [hunk 8]
```diff
@@ -85,32 +85,43 @@ def module_load_regex(modfilepath):
     return re.compile(regex, re.M)
 
 
-def dependencies_for(mod_name, modtool, depth=None):
+def dependencies_for(mod_name, modtool, depth=None, alldeps=None, full_module_names=False):
     """
     Obtain a list of dependencies for the given module, determined recursively, up to a specified depth (optionally)
+    :param alldeps: recursion depth (default is None, which corresponds to infinite recursion depth)
```
boegel: incorrect docstring for `alldeps`?

easybuild/tools/module_generator.py
```diff
@@ -85,32 +85,43 @@ def module_load_regex(modfilepath):
...
     :param depth: recursion depth (default is None, which corresponds to infinite recursion depth)
+    :param full_module_names: use $MODULEPATH to build filepath instead of get_value_from_modulefile
     """
-    mod_filepath = modtool.modulefile_path(mod_name)
+
+    if alldeps is None:
+        alldeps = {}
+
+    if mod_name not in alldeps:
+        alldeps[mod_name] = []
```
boegel: No need for this if you use `alldeps.setdefault(mod_name, []).extend(newdeps)` below

easybuild/tools/module_generator.py
```diff
@@ -85,32 +85,43 @@ def module_load_regex(modfilepath):
...
+
+    mod_filepath = modtool.modulefile_path(mod_name, full_module_names=full_module_names)
     modtxt = read_file(mod_filepath)
     loadregex = module_load_regex(mod_filepath)
```
boegel: @migueldiascosta What about `depends_on` and `prereq`?

migueldiascosta:
This currently works for a flat module naming scheme, but not for one that extends the module path.

For HierarchicalMNS, it is possible to recurse through the hierarchy levels, keeping track of extended module paths, short and full module names for each dependency, but it's messy and even then I'm not sure it would work for other naming schemes...

PR #2786, opened 2019-02-23
 fix problem with import_available_modules when running from easybuild-framework directory

boegel:


migueldiascosta:
Going in, thanks @boegel!

PR #2788, opened 2019-02-25
prepend location of test easyblocks to $PYTHONPATH to test_generate_software_list, rather than append

boegel:
`test_generate_software_list` fails when `easybuild-easyblocks` repo is also included in `$PYTHONPATH`


migueldiascosta:
Going in, thanks @boegel!

PR #2800, opened 2019-03-06
fix test_import_available_modules on top of Python 3.7 by making sure sys.path has an empty entry

boegel:


vanzod:
@boegel You need to add the `GC3Pie` exclusion for Python 3.7 too

akesandgren:
Going in, thanks @boegel!

PR #2801, opened 2019-03-06
fix UnicodeEncodeError raised from run_cmd with Python 2.7

boegel:
Fix for this issues that occurred when installing `libreadline` on top of Python 2.7.5 when using `4.x` branch and `eb --trace libreadline-7.0-GCCcore-7.3.0.eb`:

```
== installing...
  >> running command:
	[started at: 2019-03-06 08:33:27]
	[output logged in /tmp/eb-Tmn2Kx/easybuild-run_cmd-6ll1rS.log]
	make install
  File "/home/kehoste/easybuild-framework/easybuild/main.py", line 112, in build_and_install_software
  File "/home/kehoste/easybuild-framework/easybuild/framework/easyblock.py", line 2861, in build_and_install_one
  File "/home/kehoste/easybuild-framework/easybuild/framework/easyblock.py", line 2771, in run_all_steps
  File "/home/kehoste/easybuild-framework/easybuild/framework/easyblock.py", line 2637, in run_step
    step_method(self)()
  File "/home/kehoste/easybuild-easyblocks/easybuild/easyblocks/generic/configuremake.py", line 294, in install_step
    (out, _) = run_cmd(cmd, log_all=True, simple=False)
  File "/home/kehoste/easybuild-framework/easybuild/tools/run.py", line 88, in cache_aware_func
  File "/home/kehoste/easybuild-framework/easybuild/tools/run.py", line 217, in run_cmd
    cmd_log.write(output)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 4570-4572: ordinal not in range(128)
```

This problem was reproduced in the tests before fixing the issue...

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -43,8 +43,8 @@ import tempfile
 import time
 from datetime import datetime
 
+import easybuild.tools.asyncprocess as asyncprocess
 from easybuild.base import fancylogger
-from easybuild.tools.asyncprocess import PIPE, STDOUT, Popen, recv_some, send_all
 from easybuild.tools.build_log import EasyBuildError, dry_run_msg, print_msg, time_str_since
 from easybuild.tools.config import ERROR, IGNORE, WARN, build_option
 from easybuild.tools.py2vs3 import string_type
@@ -107,7 +107,7 @@ def get_output_from_process(proc, read_size=None, asynchronous=False):
     """
 
     if asynchronous:
-        output = recv_some(proc)
+        output = asyncprocess.recv_some(proc)
     elif read_size:
         output = proc.stdout.read(read_size)
     else:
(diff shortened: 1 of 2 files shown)


easybuild/tools/run.py
```diff
@@ -97,6 +97,32 @@ def cache_aware_func(cmd, *args, **kwargs):
...
+    :param read_size: number of bytes of output to read (if None: read all output)
+    :param asynchronous: get output asynchronously
+    """
+
+    if asynchronous:
+        output = recv_some(proc)
+    elif read_size:
+        output = proc.stdout.read(read_size)
+    else:
+        output = proc.stdout.read()
```
vanzod: Do we need some sort of check in the case the function is called with `read_size=1234, asynchronous=True`?

boegel: That's not a valid use case currently... `asynchronous=True` is used by `run_cmd_qa` to get intermediate output, `read_size` is used by `run_cmd` to get output in a buffered way

boegel: I added a dedicated test for `get_output_from_process` (cfr. a37a08408), which shows that `read_size` is effectively ignored when using `asynchronous=True`...

PR #2804, opened 2019-03-08
avoid relying on order in which keys are processed in to_dependency

boegel:
When looking into extending the support in the `4.x` to other Python 3.x versions beyond 3.6, I ran into a failing test for `to_dependency` because of the order in which the keys were being iterated over.

If a bogus key-pair value is included in the dict passed to `to_dependency` (which is supposed to trigger an exception `Found unexpected (key, value) pair`), and it gets processed *before* `name` or `version` are, then it is accepted as a valid key, which is incorrect.

This patch fixes it by first considering `name`/`version` before the other keys, so the order in which the remaining keys are processed doesn't matter...


migueldiascosta:
Going in, thanks @boegel!

PR #2806, opened 2019-03-09
ensure correct value type in data field of Request object in Python 2/3

boegel:
Without this fix, the added test fails as below when using Python 3.

This is relevant for features like `--upload-test-report`.

```
ERROR: test_restclient (__main__.GithubTest)
Test use of RestClient.
  File "/Volumes/work/easybuild-framework/test/framework/github.py", line 575, in test_restclient
    status, body = client.user.emails.post(body='test@example.com')
  File "/Volumes/work/easybuild-framework/easybuild/base/rest.py", line 136, in post
    return self.request(self.POST, url, json.dumps(body), headers, content_type='application/json')
  File "/Volumes/work/easybuild-framework/easybuild/base/rest.py", line 167, in request
    conn = self.get_connection(method, url, body, headers)
  File "/Volumes/work/easybuild-framework/easybuild/base/rest.py", line 201, in get_connection
    connection = self.opener.open(request)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 523, in open
    req = meth(req)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 1247, in do_request_
    raise TypeError(msg)
TypeError: POST data should be bytes, an iterable of bytes, or a file object. It cannot be of type str.
```

### DIFF
--- a/test/framework/github.py
+++ b/test/framework/github.py
@@ -28,6 +28,7 @@ Unit tests for talking to GitHub.
 @author: Jens Timmerman (Ghent University)
 @author: Kenneth Hoste (Ghent University)
 """
+import base64
 import os
 import random
 import re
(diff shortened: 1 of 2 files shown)


test/framework/github.py
```diff
@@ -551,6 +552,40 @@ def test_det_patch_specs(self):
...
+        """Test use of RestClient."""
+        if self.skip_github_tests:
+            print("Skipping test_restclient, no GitHub token available?")
+            return
+
+        client = RestClient('https://api.github.com', username=GITHUB_TEST_ACCOUNT, token=self.github_token)
+
+        status, body = client.repos['hpcugent']['testrepository'].contents.a_directory['a_file.txt'].get()
+        self.assertEqual(status, 200)
+        # dGhpcyBpcyBhIGxpbmUgb2YgdGV4dAo= == 'this is a line of text' in base64 encoding
```
wpoely86: why don't you encode this on the fly instead of this comment?

boegel: I stole this from the `vsc-base` checks, I have to admit I don't know why it was done like this.

Will try and encode on the fly...

boegel: fixed in e0e1e16ab

PR #2807, opened 2019-03-14
update Travis config to reflect that PyYAML 5.x is no longer compatible with Python 2.6

boegel:


migueldiascosta:
Going in, thanks @boegel!

Issue #2808, opened 2019-03-15
support for .yeb not compatible with PyYAML 5.x


boegel:
see https://github.com/easybuilders/easybuild-framework/pull/2807#issuecomment-473085856

@ocaisa If you want to keep support for `.yeb` alive (sort of), here's your chance... ;)

ocaisa:
Solved in #2809

boegel:
fixed with #2809, thanks a lot @ocaisa!

PR #2809, opened 2019-03-15
Fix yeb format to work with PyYaml 5

ocaisa:
Solves #2808


ocaisa:
Hmm, these changes fail with Python 2.6, @boegel any examples of tests that are restricted to Python 2.7?

PR #2812, opened 2019-03-18
take into account that builddependencies is always a list in EasyBlock.det_iter_cnt

boegel:
@bartoldeman: we overlooked this in #2741

It's only problematic if there is more than one build dependency listed, but it's an important bug to fix...

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1529,8 +1529,18 @@ class EasyBlock(object):
 
     def det_iter_cnt(self):
         """Determine iteration count based on configure/build/install options that may be lists."""
-        iter_cnt = max([1] + [len(self.cfg[opt]) for opt in ITERATE_OPTIONS
-                              if isinstance(self.cfg[opt], (list, tuple))])
+        iter_opt_lens = [len(self.cfg[opt]) for opt in ITERATE_OPTIONS
+                         if opt not in ['builddependencies'] and isinstance(self.cfg[opt], (list, tuple))]
+
+        # we need to take into account that builddependencies is always a list
+        # we're only iteraing over it if it's a list of lists
+        builddeps = self.cfg['builddependencies']
+        if all(isinstance(x, list) for x in builddeps):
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1529,8 +1529,18 @@ def restore_iterate_opts(self):
 
     def det_iter_cnt(self):
         """Determine iteration count based on configure/build/install options that may be lists."""
-        iter_cnt = max([1] + [len(self.cfg[opt]) for opt in ITERATE_OPTIONS
-                              if isinstance(self.cfg[opt], (list, tuple))])
+        iter_opt_lens = [len(self.cfg[opt]) for opt in ITERATE_OPTIONS
+                         if opt not in ['builddependencies'] and isinstance(self.cfg[opt], (list, tuple))]
+
+        # we need to take into account that builddependencies is always a list
+        # we're only iteraing over it if it's a list of lists
```
ocaisa: `iterating`

easybuild/framework/easyblock.py
```diff
@@ -1529,8 +1529,18 @@ def restore_iterate_opts(self):
 
     def det_iter_cnt(self):
         """Determine iteration count based on configure/build/install options that may be lists."""
-        iter_cnt = max([1] + [len(self.cfg[opt]) for opt in ITERATE_OPTIONS
-                              if isinstance(self.cfg[opt], (list, tuple))])
+        iter_opt_lens = [len(self.cfg[opt]) for opt in ITERATE_OPTIONS
```
ocaisa: Not a big fan of this variable name, given the function name how about `iter_opt_counts`

easybuild/framework/easyblock.py
boegel: fixed in 23fd3a12c

easybuild/framework/easyblock.py
boegel: fixed in 23fd3a12c

PR #2817, opened 2019-03-22
fix typo in error message for use of unknown easyconfig parameter

ocaisa:


boegel:
kicking Travis now #2819 is merged which fixes the problems with the bootstrap script...

PR #2818, opened 2019-03-22
enhance log_file_format to fix problem when %(name)s template value is used for log directory

### LINKED ISSUE #2815: Can't use string templates such as %(name)s in the DIR part of --logfile-format
If I use e.g.

```
--logfile-format="easybuild-%(name)s,%(name)s-%(version)s-%(date)s.%(time)s.ebout.log"
```

However, the %(name)s in the directory-part is not replaced. So, I get folders that are actually called `'easybuild-%(name)s'`.

The documentation claims that these string templates should be usable for both the filename and directory: 

> The logfile format configuration setting contains a tuple specifying a log directory name and a template log file name. In both of these values, using the following string templates is supported:

See: https://easybuild.readthedocs.io/en/latest/Configuration.html#logfile-format

boegel:
fixes #2815

### DIFF
--- a/test/framework/config.py
+++ b/test/framework/config.py
@@ -639,6 +639,11 @@ class EasyBuildConfigTest(EnhancedTestCase):
         res = log_file_format(return_directory=True, ec=ec, date='20190322', timestamp='094356')
         self.assertEqual(res, 'eb-foo-20190322')
 
+        # test handling of incorrect setting for --logfile-format
+        init_config(args=['--logfile-format=easybuild,log.txt,thisiswrong'])
+        error_pattern = "Incorrect log file format specification, should be 2-tuple"
+        self.assertErrorRegex(EasyBuildError, error_pattern, log_file_format)
+
     def test_log_path(self):
         """Test for log_path()."""
         # default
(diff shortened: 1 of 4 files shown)


test/framework/config.py
```diff
@@ -570,6 +572,89 @@ def test_strict(self):
...
+        # default
+        self.assertEqual(log_path(), 'easybuild')
+
+        # providing template values doesn't affect the default
+        ec = {'name': 'foo', 'version': '1.2.3'}
+        res = log_path(ec=ec)
+        self.assertEqual(res, 'easybuild')
+
+        # reconfigure with value for log directory that includes templates
+        init_config(args=['--logfile-format=easybuild-%(name)s-%(version)s-%(date)s-%(time)s,log,txt'])
```
ocaisa: I don't think this option should be valid, isn't your log file going to be `log,txt` in this case?

ocaisa: Nope, just `log`, everything from the last comma is ignored

boegel: Oops, yeah. It'll probably be `log` rather than `log,txt`, but you're right, this doesn't make sense...

boegel: fixed in 884413b1c

boegel:
kicking Travis now #2819 is merged which fixes the problems with the bootstrap script...

Issue #2797, opened 2019-03-04
Incorrect definition of FFTW_STATIC_LIBS_MT and other FFT(W) related vars


akesandgren:
With foss toolchain I can see:
environment.py:97 INFO Environment variable FFTW_STATIC_LIBS_MT set to -fftw3_omp -fftw3 -pthread (previously undefined)
and with intel:
environment.py:97 INFO Environment variable FFTW_STATIC_LIBS_MT set to -fftw3xc_intel -fftw3x_cdft_lp64 -mkl_cdft_core -mkl_blacs_intelmpi_lp64 -mkl_intel_lp64 -mkl_intel_thread -mkl_core -iomp5 -pthread (previously undefined)

Scalapack on the other hand produces:
environment.py:97 INFO Environment variable LIBSCALAPACK_MT set to -Wl,-Bstatic -Wl,--start-group -lmkl_scalapack_lp64 -lmkl_blacs_intelmpi_lp64 -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core -Wl,--end-group -Wl,-Bdynamic -liomp5 -lpthread (previously undefined)
and
environment.py:97 INFO Environment variable EBVARLIBSCALAPACK_MT set to -lscalapack -lopenblas -lpthread -lgfortran (previously undefined)

Note the missing "l" (-l) for FFT libs.

boegel:
fixed in #2822

PR #2822, opened 2019-03-27
Added FFTW_STATIC_LIBS_MT to FFTW_MAP_CLASS. Otherwise FFTW_STATIC_LIBS_MT are incorrectly defined

damianam:
This PR partly addresses this issue: https://github.com/easybuilders/easybuild-framework/issues/2797

But not completely yet. `FFT_STATIC_LIBS_MT` and `FFTW_STATIC_LIBS_MT` are now both correctly defined to some linkable libraries (previously just `FFT_STATIC_LIBS_MT` was defined, `FFTW_STATIC_LIBS_MT` was wrongly defined to just `-fftw...`). However, they are wrong, as they are not really "`MT`" libraries. They link the sequential version of MKL, and I suspect it is the same for the `foss` toolchain.

@akesandgren did you also observed `FFT_STATIC_LIBS_MT` to pick up the wrong libraries?

### DIFF
--- a/easybuild/tools/toolchain/constants.py
+++ b/easybuild/tools/toolchain/constants.py
@@ -194,6 +194,7 @@ FFTW_MAP_CLASS = {
     ],
     CommaStaticLibs: [
         ('FFTW_STATIC_LIBS', 'Comma-separated list of static FFTW libraries'),
+        ('FFTW_STATIC_LIBS_MT', 'Comma-separated list of static multithreaded FFTW libraries'),
     ],
 }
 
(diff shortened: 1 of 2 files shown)


akesandgren:
I don't remember checking FFT_STATIC_LIBS_MT so no idea there.

boegel:
@damianam The value for `FFT_STATIC_LIBS_MT` looks fine to me for MKL, see also test added in https://github.com/damianam/easybuild-framework/pull/13

For FFTW, we link to `libfftw3.a` though, which maybe should be `libfftw3f_omp.a` or `libfftw3_threads.a` instead? But if so, which one should it be?

That should be dealt with in a separate issue imho, what's being fixed here is worse since the current value of `$FFT_STATIC_LIBS_MT` is just completely broken...

akesandgren:
I'd say to use the _omp since the _MT is usually used with toolchainopts openmp.

boegel:
@akesandgren The `_omp` libraries are actually used when the `openmp` toolchain option is set, I'm working on enhancing the test to verify that too...

PR #2824, opened 2019-03-28
Add support for GitHub token file to store a GitHub token.

ghuls:
Add support for GitHub token file to store a GitHub token as an
alternative to storing the GitHub token in a keyring.

The GitHub token file should be stored at:

  ~/.config/easybuild/github_token

and should only be readable by the owner of the file, which
should be the same user than the invoker of easybuild (same
permissions as private SSH keys).


ghuls:
I am unable to install the keyring module, so this workaround lets me us `eb --new-pr` and other commands that require a github token. The token is stored quite savely (as save as a SSH private key, without passfrase).

akesandgren:
"The token is stored quite savely (as save as a SSH private key, without passfrase)."
In other words not safe at all. This is something that should be banned in my opinion.

boegel:
There's a need for an alternative to using the `keyring` module, which is not easy to deal with.

I'm not a big fan of this approach either though; having the token in clear text in a file is not a good idea, and we shouldn't encourage it.

The contents of the file should be encrypted somehow. One option could be using SSH keys, see https://bjornjohansen.no/encrypt-file-using-ssh-key (there's probably a Pythonic equivalent of that).

ghuls:
The `keyring` module doesn't install at all, so at least I need a workaround.

boegel:
@ghuls OK, but that workaround should also be secure, storing the GitHub token in clear text should be avoided...

ocaisa:
There's an old PR that can install `keyring` for you with system python using eb: https://github.com/easybuilders/easybuild-easyconfigs/pull/2696
It could do with an update but it probably still works.

PR #2827, opened 2019-04-03
fix crash in dep_graph when easyconfig that uses multi_deps is used

boegel:
Fix for another problem caused by #2813 (similar to problem fixed in #2826):

```
  File "/home/travis/build/easybuilders/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 130, in test_dep_graph
    dep_graph(fn, self.ordered_specs)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/easybuild_framework-3.9.0.dev0-py2.6.egg/easybuild/framework/easyconfig/tools.py", line 200, in dep_graph
    spec['ec'].build_dependencies = [mk_node_name(s) for s in spec['ec']['builddependencies']]
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/easybuild_framework-3.9.0.dev0-py2.6.egg/easybuild/framework/easyconfig/tools.py", line 182, in mk_node_name
    if spec.get('external_module', False):
AttributeError: 'list' object has no attribute 'get'
```

### DIFF
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -197,7 +197,7 @@ def dep_graph(filename, specs):
         all_nodes.update(spec['ec'].all_dependencies)
 
         # Get the build dependencies for each spec so we can distinguish them later
-        spec['ec'].build_dependencies = [mk_node_name(s) for s in spec['ec']['builddependencies']]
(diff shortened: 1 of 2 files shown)


akesandgren:
Going in, thanks @boegel!

PR #2828, opened 2019-04-03
add support for loading first version listed in multi_deps by default in generated module file

boegel:


test/framework/toy_build.py
```diff
@@ -2025,6 +2027,140 @@ def test_toy_build_hooks(self):
...
+        if self.modtool.supports_depends_on:
+
+            remove_file(toy_mod_file)
+            self.test_toy_build(ec_file=test_ec, extra_args=['--module-depends-on'])
+
+            toy_mod_txt = read_file(toy_mod_file)
+
+            # check whether (guarded) load statement for first version listed in multi_deps is there
+            if get_module_syntax() == 'Lua':
+                cond = 'mode() == "unload" or ( not ( isloaded("GCC/4.6.3") ) and not ( isloaded("GCC/7.3.0-2.30") ) )'
```
bartoldeman: I am not sure I follow the lmod depends_on logic, what happens if one of the two modules is already loaded, then you get no depends_on on load but a depends_on on unload.

And what happens if the user switches from one to another GCC version via module load.

I'll need to try that myself.. but will do so tomorrow morning.

boegel: Good point...

When the module is being loaded, and no `GCC` is loaded yet, then the first listed version should be loaded as default; that bit is OK with the current approach, both with `load` and `depends_on`.

When the module is being unloaded, the `GCC` that is loaded should be unloaded; that bit is not OK with `depends_on` right now, since we will blindly lower the reference count for the "default" version, not for the one that is loaded...

So, on `unload` we just need a `depends_on("GCC")`? Would that work?

We do *not* want to use that on `load`, since we want to control which version gets loaded (it may not be the latest version/default-in-the-Lmod-sense).

So, something like this:

```lua
if not ( isloaded("GCC/4.6.3") ) and not ( isloaded("GCC/7.3.0-2.30") ) then
    depends_on("GCC/4.6.3")
elseif mode() == "unload" then
    depends_on("GCC")
end
```

boegel: Actually, it can be done a lot cleaner:

```lua
if isloaded("Python/2.7.15-GCCcore-8.2.0") then
    depends_on("Python/2.7.15-GCCcore-8.2.0")
else
    depends_on("Python/3.7.2-GCCcore-8.2.0")
end
```

This way, we bump the ref count for `Python/2.7.15-GCCcore-8.2.0` if it's loaded already. Same for `Python/3.7.2-GCCcore-8.2.0`, but since there's no guard there, it'll also get loaded as a default if neither are loaded.

On unloading, it also works as expected: it will lower the ref count for `Python/2.7.15-GCCcore-8.2.0` if that one is loaded, or lower the ref count (perhaps also unloading) `Python/3.7.2-GCCcore-8.2.0` otherwise.

bartoldeman: Yes I was thinking on something like that too (there needs to be a `depends_on` for all `if` branches for sure), but I think there may still be an issue if the user switches the python version after loading scipy-bundle; a `depends_on("Python")` for `unload` as you did before may fix this. To be tested...
```
if mode() == "unload" then
    depends_on("Python")
elseif isloaded("Python/2.7.15-GCCcore-8.2.0") then
    depends_on("Python/2.7.15-GCCcore-8.2.0")
else
    depends_on("Python/3.7.2-GCCcore-8.2.0")
end
```

boegel: @bartoldeman I think my last comment that just uses `isloaded("GCC")` also covers that...

Just pushed the necessary changes with c6387cd31, please re-review?

bartoldeman: no, now it will behave differently if an incompatible python version is already loaded, and keep using that.

bartoldeman: This is the shortest I can make it:
```
if mode() == "unload" or isloaded("Python/2.7.15-GCCcore-8.2.0") then
    depends_on("Python")
else
    depends_on("Python/3.7.2-GCCcore-8.2.0")
end
```
List any other non-default version in the or statement, eg. `if mode() == "unload" or isloaded("Python/2.7.15-GCCcore-8.2.0") or isloaded("Python/3.6.6-GCCcore-8.2.0") then`

mboisson:
I ran into an issue from this PR. I found `multi_deps_load_default` to disable that feature, but I think that feature still raises the question of why EasyBuild is loading a specific version, hence overriding  whatever might have been set as default version in the module system ? Why does it choose to mark a version of python as "default" when the module system could say something entirely different ?

mboisson:
If one wants to limit to the known versions, it could be
`depends_on(between("python",<min version>,<max version>))`

boegel:
@mboisson Leaving it up to the actual default is equally problematic, since that tends to change over time (and it's outside of the control of EasyBuild)...

The naming may be a bit unfortunate, but the help message is pretty clear:

```
$ eb -a | grep multi_deps_load_default
multi_deps_load_default   Load module for first version listed in multi_deps by default [default: True]
```

mboisson:
I can understand the rationale, but I disagree with it. We will probably always disable the automatic loading. In my opinion, setting a default should not be something that requires hard coding in a module, and réinstallation to change the default. There are mechanism in module systems, like deprecation, hiding, default versions, especially made to handle this.
  

    
Anyway, I’m happy that there is a flag to disable this. I would probably be slightly happier if it were a global config flag rather than an easy config specific flag.... actually, does it exist as a global flag that we can turn off ?
  

  

  
  

  
  

  
  
>

mboisson:
The other problem is the inconsistency that it creates. Lmod already shows what is the default version of a module, and now the module comes and hard code a potentially different version. That is just confusing. I would like to even remove the mention in the help message that there is a specific default version, but it seems that the flag to disable the loading of defaults does not remove that mention in the module help.
  

  

  
  

  

  
  
>

boegel:
There's no global flag to turn this off, but we can certainly consider adding that.

The fact that the mention in module help stays there even if this is disabled sounds like a bug...

Please open separate issues on both to follow up (comments in a merged PR make good follow-up difficult...).

mboisson:
Done, https://github.com/easybuilders/easybuild-framework/issues/2961
https://github.com/easybuilders/easybuild-framework/issues/2962

PR #2833, opened 2019-04-16
Lower required version in slurm job backend to 16.05

geimer:
This has been tested on our test cluster running SLURM 16.05.9 and works like a charm.  It is also quite likely that earlier releases may work as well, but testing this is up to someone with an even more ancient setup ;-)


boegel:
Makes sense to me, thanks @geimer!

PR #2834, opened 2019-04-17
only call 'scontrol release' when there's at least one job submitted

boegel:
Fix for following issue:

```
$ eb foo-1.2.3.eb --job
== temporary log file in case of crash /tmp/eb-7fOWcP/easybuild-KIt4oj.log
== foo/1.2.3 is already installed (module found), skipping
== No easyconfigs left to be built.
ERROR: cmd "scontrol release " exited with exit code 1 and output:
too few arguments for keyword:release
```


akesandgren:
Going in, thanks @boegel!

PR #2835, opened 2019-04-17
fix small bug in find_eb_script + various minor issues with tests (related to environment in which tests are run)

boegel:


test/framework/options.py
```diff
@@ -2248,7 +2270,11 @@ def test_include_module_naming_schemes(self):
         args = [
             '--avail-module-naming-schemes',
         ]
-        logtxt, _ = run_cmd("cd %s; eb %s" % (self.test_prefix, ' '.join(args)), simple=False)
+
+        env_cmd = "export PATH='%s'; export PYTHONPATH='%s'; " % (path, pythonpath)
```
akesandgren: Don't really need the "; export" part here you can export both with the same export

test/framework/options.py
```diff
@@ -2326,7 +2366,10 @@ def test_include_toolchains(self):
         args = [
             '--list-toolchains',
         ]
-        logtxt, _ = run_cmd("cd %s; eb %s" % (self.test_prefix, ' '.join(args)), simple=False)
+        env_cmd = "export PATH='%s'; export PYTHONPATH='%s'; " % (path, pythonpath)
```
akesandgren: Same here

test/framework/options.py
```diff
@@ -2271,7 +2271,7 @@ def test_include_module_naming_schemes(self):
             '--avail-module-naming-schemes',
         ]
 
-        env_cmd = "export PATH='%s'; export PYTHONPATH='%s'; " % (path, pythonpath)
+        env_cmd = "export PATH='%s' export PYTHONPATH='%s'; " % (path, pythonpath)
```
akesandgren: No, env_cmd = "export PATH='%s' PYTHONPATH='%s'; " % (path, pythonpath)

boegel: needs more coffee...

boegel: fixed in 776993e6e with force push

akesandgren:
Going in, thanks @boegel!

PR #2836, opened 2019-04-19
drop requirement for setuptools as runtime dependency

boegel:
`setuptools` was only required because of the  `pkg_resources.declare_namespace` statements that were used to declare the various `easybuild.*` (sub)namespaces.

There's a replacement in the Python standard library however, being `__import__('pkgutil').extend_path`, and it's even the recommend approach (see https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages).

One side-effect is that the `__init__.py` files should only contain the `extend_path` statement, nothing else. Because of this, some things had to be moved, i.e.:

* the `DUMMY_TOOLCHAIN_NAME` and `DUMMY_TOOLCHAIN_VERSION` constants were moved from `easybuild.tools.toolchain` to `easybuild.tools.toolchain.toolchain`
* the `DEVEL_MODULE_SUFFIX` constant was moved from `easybuild.tools.module_naming_scheme` to `easybuild.tools.module_naming_scheme.mns`
* the `ModuleNamingScheme` class is no longer available from the `easybuild.tools.module_naming_scheme` namespace, only from the `easybuild.tools.module_naming_scheme.mns` namespace (cfr. changes in #953)


akesandgren:
Going in, thanks @boegel!

PR #2837, opened 2019-04-22
only import from pkg_resources (provided by setuptools) to check GC3Pie version

boegel:
Since `setuptools` is a requirement for GC3Pie, this will work out fine in practice. It allows us to avoid requiring `setuptools` for EasyBuild itself...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,10 +113,10 @@ script:
     - export PYTHONPATH=/tmp/$TRAVIS_JOB_ID/lib/python$TRAVIS_PYTHON_VERSION/site-packages
     # install easybuild-framework to unique temporary location using prepared sdist tarball
     - mkdir -p $PYTHONPATH; easy_install --prefix /tmp/$TRAVIS_JOB_ID $TRAVIS_BUILD_DIR/dist/easybuild-framework*tar.gz
-    # make sure there are no "import setuptools" or "import pkg_resources" statements,
+    # make sure there are no (top-level) "import setuptools" or "import pkg_resources" statements,
     # since EasyBuild should not have a runtime requirement on setuptools
-    - SETUPTOOLS_IMPORTS=$(egrep -RI '^(from|import) pkg_resources|^(from|import) setuptools' easybuild/ || true)
-    - test "x$SETUPTOOLS_IMPORTS" = "x" || (echo "Found setuptools imports in easybuild/:\n${SETUPTOOLS_IMPORTS}" && exit 1)
+    - SETUPTOOLS_IMPORTS=$(egrep -RI '^(from|import)[ \t]*pkg_resources|^(from|import)[ \t]*setuptools' easybuild/ || true)
+    - test "x$SETUPTOOLS_IMPORTS" = "x" || (echo "Found setuptools and/or pkg_resources imports in easybuild/:\n${SETUPTOOLS_IMPORTS}" && exit 1)
     # move outside of checkout of easybuild-framework repository,
     # to run tests on an *installed* version of the EasyBuild framework;
     # this is done to catch possible packaging issues
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,11 +113,10 @@ script:
     - export PYTHONPATH=/tmp/$TRAVIS_JOB_ID/lib/python$TRAVIS_PYTHON_VERSION/site-packages
     # install easybuild-framework to unique temporary location using prepared sdist tarball
     - mkdir -p $PYTHONPATH; easy_install --prefix /tmp/$TRAVIS_JOB_ID $TRAVIS_BUILD_DIR/dist/easybuild-framework*tar.gz
-    # actively break "import setuptools" and "import pkg_resources",
+    # make sure there are no "import setuptools" or "import pkg_resources" statements,
     # since EasyBuild should not have a runtime requirement on setuptools
-    - echo 'raise ImportError("setuptools should no longer be relied on")' > /tmp/$TRAVIS_JOB_ID/setuptools.py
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -113,11 +113,10 @@ script:
     - export PYTHONPATH=/tmp/$TRAVIS_JOB_ID/lib/python$TRAVIS_PYTHON_VERSION/site-packages
     # install easybuild-framework to unique temporary location using prepared sdist tarball
     - mkdir -p $PYTHONPATH; easy_install --prefix /tmp/$TRAVIS_JOB_ID $TRAVIS_BUILD_DIR/dist/easybuild-framework*tar.gz
-    # actively break "import setuptools" and "import pkg_resources",
+    # make sure there are no "import setuptools" or "import pkg_resources" statements,
     # since EasyBuild should not have a runtime requirement on setuptools
-    - echo 'raise ImportError("setuptools should no longer be relied on")' > /tmp/$TRAVIS_JOB_ID/setuptools.py
-    - echo 'raise ImportError("setuptools is not available")' > /tmp/$TRAVIS_JOB_ID/pkg_resources.py
-    - export PYTHONPATH=/tmp/$TRAVIS_JOB_ID:$PYTHONPATH
+    - SETUPTOOLS_IMPORTS=$(egrep -RI '^(from|import) pkg_resources|^(from|import) setuptools' easybuild/ || true)
```
akesandgren: To make this 100% correct you need
'^ *(from|import) *pkg_resources|^ *(from|import) *setuptools'
and in fact it should be '^[ \t]*...'
where \t is an actual tab.

.travis.yml
```diff
@@ -113,11 +113,10 @@ script:
...
+    - test "x$SETUPTOOLS_IMPORTS" = "x" || (echo "Found setuptools imports in easybuild/:\n${SETUPTOOLS_IMPORTS}" && exit 1)
```
akesandgren: "Found setuptools and/or pkg_resources..."

.travis.yml
boegel: No, I only want to catch top-level imports, so the `^(` is deliberate.

Otherwise we would also hit the `from pkg_resources import ...` in `gc3pie.py`.

You're right about the wildcard before `pkg_resources` though I guess, in theory...

.travis.yml
boegel: Since `pkg_resources` is provided  by `setuptools`, technically this is correct, but I see the point, so I'll tweak it, thanks.

.travis.yml
akesandgren: Ah yes, i keep forgetting that python is indentation language.

.travis.yml
```diff
@@ -113,10 +113,10 @@ script:
     - export PYTHONPATH=/tmp/$TRAVIS_JOB_ID/lib/python$TRAVIS_PYTHON_VERSION/site-packages
     # install easybuild-framework to unique temporary location using prepared sdist tarball
     - mkdir -p $PYTHONPATH; easy_install --prefix /tmp/$TRAVIS_JOB_ID $TRAVIS_BUILD_DIR/dist/easybuild-framework*tar.gz
-    # make sure there are no "import setuptools" or "import pkg_resources" statements,
+    # make sure there are no (top-level) "import setuptools" or "import pkg_resources" statements,
     # since EasyBuild should not have a runtime requirement on setuptools
-    - SETUPTOOLS_IMPORTS=$(egrep -RI '^(from|import) pkg_resources|^(from|import) setuptools' easybuild/ || true)
-    - test "x$SETUPTOOLS_IMPORTS" = "x" || (echo "Found setuptools imports in easybuild/:\n${SETUPTOOLS_IMPORTS}" && exit 1)
+    - SETUPTOOLS_IMPORTS=$(egrep -RI '^(from|import)[ \t]*pkg_resources|^(from|import)[ \t]*setuptools' easybuild/ || true)
```
akesandgren: the \t must be an actual tab char. egrep doesn't handle \t, at least mine doesn't.

boegel: In that case I'll just drop the tab character, `import` statements that involve tab characters are going to be an issue regardless of this...

.travis.yml
```diff
@@ -113,9 +113,10 @@ script:
     - export PYTHONPATH=/tmp/$TRAVIS_JOB_ID/lib/python$TRAVIS_PYTHON_VERSION/site-packages
     # install easybuild-framework to unique temporary location using prepared sdist tarball
     - mkdir -p $PYTHONPATH; easy_install --prefix /tmp/$TRAVIS_JOB_ID $TRAVIS_BUILD_DIR/dist/easybuild-framework*tar.gz
-    # actively break "import setuptools", since EasyBuild should not have a runtime requirement on setuptools
-    - echo 'raise NotImplementedError("setuptools is not available")' > /tmp/$TRAVIS_JOB_ID/setuptools.py
-    - export PYTHONPATH=/tmp/$TRAVIS_JOB_ID:$PYTHONPATH
+    # make sure there are no (top-level) "import setuptools" or "import pkg_resources" statements,
+    # since EasyBuild should not have a runtime requirement on setuptools
+    - SETUPTOOLS_IMPORTS=$(egrep -RI '^(from|import)[ ]*pkg_resources|^(from|import)[ ]*setuptools' easybuild/ || true)
```
akesandgren: Without the tab you don't need [] around the space.

boegel: Strictly speaking, no, but it's awkward not to do so imho...

akesandgren: Ok, will leave it so then.

akesandgren:
Going in, thanks @boegel!

PR #2838, opened 2019-04-23
add function to determine whether specified easyblock is generic or not

boegel:


akesandgren:
Going in, thanks @boegel!

PR #2840, opened 2019-04-25
Arch specific deps - fosscuda 2018b

edmondac:


edmondac:
Ooops - wrong target branch

Issue #2655, opened 2018-11-06
Architecture-specific dependencies


edmondac:
I propose adding an alternative way of specifying a dependency as a dict rather than a tuple. The dict should handle all the existing info from the tuple.

Then adding the ability to say which architecture it's for, like this:

To specify that mkl_dnn should only be included as a dep on x86_64:
```python
dependencies = [{'name': 'mkl_dnn', 'version': '0.16', 'arch': 'x86_64'}, ...]
```

To specify which Java should be used:
```python
dependencies = [
    {'name': 'Java', 'version': '%(version)s_191-b26-OpenJDK'', 'arch': 'ppc64le'},
    {'name': 'Java', 'version': '%(version)s.0_181', 'arch': 'x86_64'},
]
```

edmondac:
This is related to https://github.com/easybuilders/easybuild-framework/issues/2654 and https://github.com/easybuilders/easybuild-easyconfigs/pull/7110

boegel:
@edmondac Good news: specifying dependencies through a dictionary is actually already supported in the EasyBuild framework... See the `_parse_dependency` method in `easybuild/framework/easyconfig/easyconfig.py`.

It's there but never used in the central repository, mostly because of brevity. But if additional stuff needs to be specified in dependencies, then extending the dict format is way better than extending the tuple format...

The `name`, `version`, `versionsuffix` and `toolchain` keys already work as expected.

For example, this:

```python
dependencies = [('foo', '1.2.3, '-example', ('GCC', '4.8.4')]
```

is equivalent with

```python
dependencies = [{
  'name': 'foo',
  'version': '1.2.3,
  'versionsuffix': '-example',
  'toolchain': {'name': 'GCC', 'version': '4.8.4'},
}]
```

So, adding support for an additional `'arch'` wouldn't be *so* much work, but you may hit some surprises when extending the dict value that is produced by `_parse_dependency`... The best way of figuring it out is by trying it. ;)

edmondac:
Excellent - I'll have a go at that.

edmondac:
Half of the solution is complete in PRs related to CUDA:

-     #2946
-     easybuilders/easybuild-easyblocks#1766
-     easybuilders/easybuild-easyconfigs#8136

Just a solution for Java now required...

edmondac:
This was all done

PR #2841, opened 2019-04-25
Architecture-specific dependencies

edmondac:
Specify different deps depending on what architecture we're running on


edmondac:
Fixes #2655

edmondac:
@boegel I'm getting this error: ```ImportError: cannot import name OrderedDict``` - do we still need to be Python 2.6 compatible? 2.7 goes EOL in less than a year! Any advice gratefully received :-)

ocaisa:
You can use
```
```

edmondac:
Fab - thanks!

edmondac:
Is there a pattern for "assertRaisesRegexp" in the EasyBuild world?

edmondac:
Example easyconfigs can be found in https://github.com/easybuilders/easybuild-easyconfigs/pull/8136

boegel:
@edmondac My apologies for not giving this a thorough review yet, it's been difficult to find time for non-trivial external contributions lately.

The changes in https://github.com/easybuilders/easybuild-easyconfigs/pull/8136/files#diff-ced0387c9ba3d51cd26a9aeab6dd5ef0 make me wonder whether there's a cleaner way to support this though...
It seems like the current approach would lead to a lot of duplicate work.

For CUDA (in https://github.com/easybuilders/easybuild-easyconfigs/pull/8136), it seems like it's just a matter of picking a different source file depending on the system architecture.

For Java (in https://github.com/easybuilders/easybuild-easyconfigs/pull/8408), it looks more complicated...

Maybe we should set up a conf call at some point to discuss the motivation for these changes, and see what we can come up with? Realistically, we're talking end of June here though...
Please fire me an email if this sounds interesting to you.

edmondac:
Re. CUDA I was thinking a dictionary for checksums might solve it, so we could include a checksum for both arch, and then just use %(arch)s in the filename for the source code.

edmondac:
1) CUDA - checksums dictionary instead of list would work, since the filename can use %(arch)s. Could use easyblock for more complicated source filenames using a custom template variable.

edmondac:
2. For Java - could do this:
```
dependencies = [('Java', {'arch=x86_64': '%(version)s_202', 'arch=ppc64le': '%(version)s_191-b26-OpenJDK'})]
```

edmondac:
> 
> 
>     1. CUDA - checksums dictionary instead of list would work, since the filename can use %(arch)s. Could use easyblock for more complicated source filenames using a custom template variable.

https://github.com/easybuilders/easybuild-framework/issues/2853 can be extended for this

boegel:
@edmondac As discussed, I think looking into supporting something like:

```python
dependencies = [('Java', {'arch=x86_64': '%(version)s_202', 'arch=ppc64le': '%(version)s_191-b26-OpenJDK'})]
```

makes more sense, since it side-steps a couple of things, incl. having multiple entries in `dependencies` for the same dependency which has possible impact in a bunch of places.

In addition, it also makes the implementation quite a bit simpler I think, since all that needs to be done (I think) is checking the type for `version` part of the dependency spec in the `_parse_dependency` method at https://github.com/easybuilders/easybuild-framework/blob/master/easybuild/framework/easyconfig/easyconfig.py#L1178, and then do some more work in case it's a `dict` value.

Something like:

```python
# new function/method
def find_dep_version_match(dep_version_candidates):
    """Sensible docstring goes here"""
    # figure out matches based on dict keys (after splitting on '=')
    matches = [...]

    res = None
    # if there's a single match, we can continue with the corresponding string version
    if len(matches) == 1:
        res = matches[0]
    elif matches:
        raise EasyBuildError("Multiple matches found for dep version using %s: %s", dep_version_candidates, matches)
    else:
        raise EasyBuildError("No matches for dep version using %s", dep_version_candidates)

    return res

def _parse_dependency(...):
    ...
    if isinstance(dependency['version'], basestring):
        self.log.debug("Dependency version is already a string ('%s'), OK", dependency['version'])
    elif if isinstance(dependency['version'], dict):
        # call out to a function that figures out which entries in the dict matches
        selected_dep_version = find_dep_version_match(dependency['version'])
        self.log.info("Dependency version selected from %s: %s", dependency['version'], selected_dep_version)
        dependency['version'] = selected_dep_version
    else:
        raise EasyBuildError("Unknown value type for dependency version: %s", dependency['version'])

edmondac:
> > ```
> > 1. CUDA - checksums dictionary instead of list would work, since the filename can use %(arch)s. Could use easyblock for more complicated source filenames using a custom template variable.
> > ```
> 
> #2853 can be extended for this

CUDA PRs:

- https://github.com/easybuilders/easybuild-framework/pull/2946
- https://github.com/easybuilders/easybuild-easyblocks/pull/1766
- https://github.com/easybuilders/easybuild-easyconfigs/pull/8136

edmondac:
> > > ```
> > > 1. CUDA - checksums dictionary instead of list would work, since the filename can use %(arch)s. Could use easyblock for more complicated source filenames using a custom template variable.
> > > ```
> > 
> > 
> > #2853 can be extended for this
> 
> CUDA PRs:
> 
>     * #2946
> 
>     * [easybuilders/easybuild-easyblocks#1766](https://github.com/easybuilders/easybuild-easyblocks/pull/1766)
> 
>     * [easybuilders/easybuild-easyconfigs#8136](https://github.com/easybuilders/easybuild-easyconfigs/pull/8136)

These CUDA PRs have all been merged now - so fosscuda 2019a will build happily on POWER9.

edmondac:
Only Java is now required, which will happen in a separate PR. Closing this one now.

edmondac:
> @edmondac As discussed, I think looking into supporting something like:
> 
> ```python
> dependencies = [('Java', {'arch=x86_64': '%(version)s_202', 'arch=ppc64le': '%(version)s_191-b26-OpenJDK'})]
> ```
> 
> makes more sense, since it side-steps a couple of things, incl. having multiple entries in `dependencies` for the same dependency which has possible impact in a bunch of places.
> 
> In addition, it also makes the implementation quite a bit simpler I think, since all that needs to be done (I think) is checking the type for `version` part of the dependency spec in the `_parse_dependency` method at https://github.com/easybuilders/easybuild-framework/blob/master/easybuild/framework/easyconfig/easyconfig.py#L1178, and then do some more work in case it's a `dict` value.
> 
> Something like:
> 
> ```python
> # new function/method
> def find_dep_version_match(dep_version_candidates):
>     """Sensible docstring goes here"""
>     # figure out matches based on dict keys (after splitting on '=')
>     matches = [...]
> 
>     res = None
>     # if there's a single match, we can continue with the corresponding string version
>     if len(matches) == 1:
>         res = matches[0]
>     elif matches:
>         raise EasyBuildError("Multiple matches found for dep version using %s: %s", dep_version_candidates, matches)
>     else:
>         raise EasyBuildError("No matches for dep version using %s", dep_version_candidates)
> 
>     return res
> 
> def _parse_dependency(...):
>     ...
>     if isinstance(dependency['version'], basestring):
>         self.log.debug("Dependency version is already a string ('%s'), OK", dependency['version'])
>     elif if isinstance(dependency['version'], dict):
>         # call out to a function that figures out which entries in the dict matches
>         selected_dep_version = find_dep_version_match(dependency['version'])
>         self.log.info("Dependency version selected from %s: %s", dependency['version'], selected_dep_version)
>         dependency['version'] = selected_dep_version
>     else:
>         raise EasyBuildError("Unknown value type for dependency version: %s", dependency['version'])
> ```

I'm basing my new PRs on this: https://github.com/easybuilders/easybuild-framework/pull/3047 and https://github.com/easybuilders/easybuild-easyconfigs/pull/9081

PR #2842, opened 2019-04-26
allow setting parallel to 0 or False to disable adding the -j argument

migueldiascosta:
motivation: `ConfigureMake`, `MakeCp`, etc. easyblocks allow setting `build_cmd` to something other than `make`, but without this change `-j PARALLEL` is always added anyway

this change is enough because `self.cfg['parallel']` is tested again in the `configuremake` build step, before adding `-j`

### DIFF
--- a/test/framework/easyblock.py
+++ b/test/framework/easyblock.py
@@ -1349,6 +1349,10 @@ class EasyBlockTest(EnhancedTestCase):
         os.close(handle)
         write_file(toy_ec2, toytxt + "\nparallel = 123\nmaxparallel = 67")
 
+        handle, toy_ec3 = tempfile.mkstemp(prefix='easyblock_test_file_', suffix='.eb')
+        os.close(handle)
+        write_file(toy_ec3, toytxt + "\nparallel = False")
+
         # default: parallellism is derived from # available cores + ulimit
         test_eb = EasyBlock(EasyConfig(toy_ec))
         test_eb.check_readiness_step()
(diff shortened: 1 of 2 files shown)


boegel:
@migueldiascosta We should cover the use case of using `parallel = False` in the tests, see `test_parallel` in `test/framework/easyblock.py`

test/framework/easyblock.py
```diff
@@ -1380,6 +1384,11 @@ def test_parallel(self):
         test_eb.check_readiness_step()
         self.assertEqual(test_eb.cfg['parallel'], 67)
 
+        # make sure 'parallel = False' is not overriden
+        test_eb = EasyBlock(EasyConfig(toy_ec3))
+        test_eb.check_readiness_step()
+        self.assertEqual(test_eb.cfg['parallel'], False)
```
boegel: @migueldiascosta Looks good, but you should also do the same test in the context where `--parallel` is not used (i.e. copy-paste this stuff above line 1372 where `--parallel 97` is put in place, to ensure `parallel = False` works as expected in both contexts.

Side-note: this test only passes because `0 == False` evaluates to `True` in Python...

When `--parallel` is also specified, the current codes takes the minimum of that value (`97` in the test) and the `int` values of the `parallel` easyconfig parameter; with `int(False)` returning `0`, you get `min(97, 0)` or just `0`.

That's not necessarily a problem though, but maybe we should check for `0` explicitly here instead?
We should keep checking against `False` in the context where `--parallel` is not set though...

boegel:
Going in, thanks for the fix @migueldiascosta!

PR #2843, opened 2019-04-26
 add support for marking extensions with 'only_if_missing', such that they're skipped automatically if already installed

boegel:
This may come in useful to automatically skip Python packages that are available in Python 3.x but not in Python 2.x, for multi-Python installations.

My use case for this was `numba`, which requires `funcsigs` and `singledispatch` only when built on top of Python 2.x, but this approach doesn't cover that because of simple `improt singledispatch` is not enough to skip the installation for `singledispatch` (since it actually lives in `functools.singledispatch` in Python 3.x); see also https://github.com/easybuilders/easybuild-easyconfigs/pull/8164 .

Nevertheless, this could be useful in other contexts...

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1426,7 +1426,7 @@ class EasyBlock(object):
 
         # only consider actually skipping extensions when --skip is used,
         # or when one or more individual extensions are marked 'only_if_missing'
-        do_skip = self.skip or any(ext['options'].get('only_if_missing', False) for ext in self.exts)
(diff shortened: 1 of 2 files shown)


damianam:
Couple of thoughts:

* Shouldn't this be the default behaviour? Why install something twice if it is available already?
* What about versions? Does it make sense to install something `if missing or version <X.Y`? 
* I think the name is misleading. EB installs extensions only if missing already, when using `--skip`. If I understand this correctly, this expands it by skipping it also in a "normal" installation. For this use case I would use `install_if_python2`

akesandgren:
@boegel  conflict resolution please

Issue #2845, opened 2019-04-26
Problem with nested quoting in environment variables


martin-lueders:
I am trying to modify a working easyconfig file by changing the toolchain from ("foss", "2018a") to ("fosscuda", "2018a"), leaving everything else the same.

The eb command now fails in the final (sanity checking) stage. I managed to track down the problem a bit further by locating the temporarily created module file. This contains the lines:
```
setenv  CUDA_CFLAGS             '-Xcompiler="-O2 -ftree-vectorize -march=native -fno-math-errno" -Xlinker="-L/home/octobld/eb-test/EasyBuild/software/CUDA/9.1.85-GCC-6.4.0-
2.28/lib64 -L/home/octobld/eb-test/EasyBuild/software/GCCcore/6.4.0/lib64 -L/home/octobld/eb-test/EasyBuild/software/GCCcore/6.4.0/lib -lm -lrt -lcudart -lpthread"'
```
and 
```
setenv  CUDA_CXXFLAGS           '-Xcompiler="-O2 -ftree-vectorize -march=native -fno-math-errno" -Xlinker="-L/home/octobld/eb-test/EasyBuild/software/CUDA/9.1.85-GCC-6.4.0-
2.28/lib64 -L/home/octobld/eb-test/EasyBuild/software/GCCcore/6.4.0/lib64 -L/home/octobld/eb-test/EasyBuild/software/GCCcore/6.4.0/lib -lm -lrt -lcudart -lpthread"'
```

These two lines give rise to an error
```
cmdSetenv.c(124):ERROR:11: Usage is 'setenv[-force] variable value'
bb-fosscuda-2018a(49):ERROR:102: Tcl command execution failed: setenv   CUDA_CFLAGS             '-Xcompiler="-O2 -ftree-vectorize -march=native -fno-math-errno" -Xlinker="-L/home/octobld/eb-test/EasyBuild/software/CUDA/9.1.85-GCC-6.4.0-2.28/lib64 -L/home/octobld/eb-test/EasyBuild/software/GCCcore/6.4.0/lib64 -L/home/octobld/eb-test/EasyBuild/software/GCCcore/6.4.0/lib -lm -lrt -lcudart -lpthread"'
```
when trying a module show or module load.

The similar looking line:
```
setenv           CUDA_CC 'nvcc -ccbin="g++"'
```
however works without problems. It seems related to the fact that the lines above contain more than one inner quotations.

Is there any easy way to fix this?

[...]

PR #2847, opened 2019-04-29
Be more careful when using single quotes for Tcl syntax

ocaisa:
Fixes https://github.com/easybuilders/easybuild-framework/pull/2845


ocaisa:
I'm nervous about this change, using brackets changes the Tcl treatment of strings: "When quoting with braces, no substitutions are performed."

ocaisa:
I've gone for just escaping the double quotes in the value rather than introduce behaviour changing brackets.

easybuild/tools/utilities.py
```diff
@@ -59,7 +59,7 @@ def flatten(lst):
     return res
 
 
-def quote_str(val, escape_newline=False, prefer_single_quotes=False):
+def quote_str(val, escape_newline=False, prefer_single_quotes=False, syntax=None):
```
boegel: @ocaisa Please add a `:param` line to the docstring for the new `syntax` parameter (and mention `Tcl` explicitly, since that's actually the only value that matters).

With that in mind, maybe you should be adding a `tcl=False` option instead, rather than checking for a magic value (and then use `tcl=self.SYNTAX == TCL` in `module_generator.py`)?

ocaisa:
@boegel Have you got an example of a test that checks if a module is load-able?

boegel:
@ocaisa see `test/framework/toy_build.py`, there's plenty in there

ocaisa:
@boegel So there is an interesting(?) problem with the tests. Recent versions of Lmod (`7.8.5` in the tests) seem to sanitise the string so that when I compare values I get a failure:
```
AssertionError: '-Xflags1="foo\\ bar"\\ -Xflags2="more\\ flags"\\ ' != '-Xflags1="foo bar" -Xflags2="more flags" ':
DIFF:
- -Xflags1="foo\ bar"\ -Xflags2="more\ flags"\ ?              -     -               -       ^^
+ -Xflags1="foo bar" -Xflags2="more flags" 
```

ocaisa:
There are two failing tests that occur with recent versions of Lmod. The contents of the test module are
```
#%Module
proc ModulesHelp { } {
    puts stderr {

Description
===========
gzip (GNU zip) is a popular data compression program as a replacement for compress


More information
================
 - Homepage: http://www.gzip.org/
    }
}

module-whatis {Description: gzip (GNU zip) is a popular data compression program as a replacement for compress}
module-whatis {Homepage: http://www.gzip.org/}

set root /tmp/eb-t6RPaR/eb-Edj_cW/eb-UQ3StE/easybuild-modgen-test-tHkXQq

conflict gzip
setenv	TEST_FLAGS		"-Xflags1=\"foo bar\" -Xflags2=\"more flags\" "
```
This is indeed correct so I don't really know what to do.

ocaisa:
I checked this against Lmod `8.0.2` and the tests are passing there. Unfortunately this looks like an Lmod bug, @boegel what do I do in this case?

boegel:
@ocaisa https://github.com/TACC/Lmod/blob/master/README.old#L14 suggests that the issue was fixed in Lmod 7.8.8.

Let's try and update the Lmod 7.x versions used in Travis to the latest 7.x?

We should also add an entry for latest 8.x, but let's do that in a separate PR.

And I guess we should considering dropping support for Lmod 6.x in EasyBuild 4.0?
Last Lmod 6.x release was Nov'16...

ocaisa:
Ok, the failing tests are fixed with Lmod 7.8.8

.travis.yml
```diff
@@ -5,8 +5,8 @@ env:
     # purposely specifying slowest builds first, to gain time overall
     - LMOD_VERSION=6.5.1
     - LMOD_VERSION=6.5.1 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
-    - LMOD_VERSION=7.8.5
-    - LMOD_VERSION=7.8.5 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
+    - LMOD_VERSION=7.8.22
```
ocaisa: I chose `7.8.22` rather than `7.8.90` because I saw issues on my local machine with `7.8.90` (with my `Devel` module that is valid with every other release)

boegel: 7.8.22 makes sense, that's the actual latest 7.8.x based on https://github.com/TACC/Lmod/blob/master/README.old#L31, 7.8.90 is more like a 8.0 beta...

boegel:
Looking good, going in, thanks for your efforts on this @ocaisa!

PR #2848, opened 2019-04-30
append to existing .modulerc file rather than overwriting it (if --force is used)

boegel:
fix for https://github.com/easybuilders/easybuild-easyblocks/issues/1699


akesandgren:
Going in, thanks @boegel!

PR #2849, opened 2019-04-30
Add Lmod 8 to test suite

ocaisa:


ocaisa:
`8.0.6` is the last Lmod 8 release where the test suite passes, in `8.0.7` something was introduced that breaks the modulerc tests

ocaisa:
Lmod introduced "extended defaults" in `v8.0.7` (if `Python/2.7.16` exists `Python/2` is loadable even if that specific module file does not exist), I've configured the tests to disable this feature.

boegel:
@ocaisa I'm not sure disabling a feature in the test setup that's enabled by default and breaks something in EasyBuild is the best way forward here...

In what way do the test fail? Are they simply not taking something into account, or are they signaling that something is broken in framework when using Lmod 8.0.7 with the "extended defaults" feature enabled?

ocaisa:
It breaks the modulerc test at https://github.com/easybuilders/easybuild-framework/blob/master/test/framework/module_generator.py#L390 because with extended defaults the central module also evaluated to true. Either we retain the strictness that exists in the other module tools or consider switching this particular test off for Lmod.

boegel:
@ocaisa Thart's basically a bug in the tests, for the case where Lmod 8.0.8 or newer is configured with "extended defaults", it's not a bug in EasyBuild itself I guess (although it could lead to weird situations...).

There's probably a runtime switch we can enable to tell Lmod to not consider extended defaults when being used by EasyBuild (much like we disable the Lmod spider cache by setting `$LMOD_IGNORE_CACHE` to `1`, see https://github.com/easybuilders/easybuild-framework/blob/master/easybuild/tools/modules.py#L1187).

So we should just figure out which magic `$LMOD_*` variable to define to disable the "extended default" feature of Lmod 8.0.8 at runtime, and (always) set it?

ocaisa:
I'm for using the runtime switch, but I suspect that will be a feature request (the configure option to disable extended defaults appeared in `8.0.8` if I remember correctly from yesterday)

ocaisa:
@boegel When  tests pass this should be good to go.

boegel:
Thanks @ocaisa!

Issue #1597, opened 2016-02-07
template PYPI_SOURCE cannot be used to download python extensions


pescobar:
I am writing a python bundle for scikit-bio and I tried to add this extension

```
exts_list = [
_genutils', '0.1.0', {
        'source_urls': [PYPI_SOURCE],
    }),
]
```

but the download fails because the url uses the main download url in the easyconfig

```
runpy.filetools WARNING Download of https://pypi.python.org/packages/source/s/scikit-bio/ipython_genutils-0.1.0.tar.gz to /export/soft/source/s/scikit-bio/extensions/ipython_genutils-0.1.0.tar.gz failed,
```

it would be nice if `PYPI_SOURCE` and `PYPI_LOWER_SOURCE` could be used for python extensions. It would simplify a lot writing python bundles

serverhorror:
Could it be that this is the culprit:

* https://github.com/hpcugent/easybuild-framework/blob/027438dffb67121e9feb2a4764a72ccaa338f1c7/easybuild/framework/easyconfig/templates.py#L102-L105

Current:

```python
    ('PYPI_SOURCE', 'https://pypi.python.org/packages/source/%(nameletter)s/%(name)s',
     'pypi source url'),  # e.g., Cython, Sphinx
    ('PYPI_LOWER_SOURCE', 'https://pypi.python.org/packages/source/%(nameletterlower)s/%(namelower)s',
     'pypi source url (lowercase name)'),  # e.g., Greenlet, PyZMQ
```
Should be:
```python
    ('PYPI_SOURCE', 'https://files.pythonhosted.org/packages/source/%(nameletter)s/%(name)s',
     'pypi source url'),  # e.g., Cython, Sphinx
    ('PYPI_LOWER_SOURCE', 'files.pythonhosted.org/packages/source/%(nameletterlower)s/%(namelower)s',
     'pypi source url (lowercase name)'),  # e.g., Greenlet, PyZMQ
```

I gave it a quick try but it seems that `nameletter*` is never actually resolved and I couldn't find why this would be happening as it will be populated for the actual easyconfig but not for the entries in `exts_list`.

boegel:
@serverhorror The actual URLs are not the problem. For PyPI specifically, we translate the URLs obtained through `PYPI_SOURCE` via `derive_alt_pypi_url` to an actual download URL (see https://bitbucket.org/pypa/pypi/issues/438 why).

[...]

Issue #1138, opened 2015-01-15
easyconfig templates for source_urls don't work with extensions


boegel:
cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/1306

boegel:
fixed in #2852

PR #2852, opened 2019-05-01
correctly resolve template values used for extensions

### LINKED ISSUE #1445: string values for extensions are templated with wrong name/version when queried early
for example in [this easyconfig file](https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/p/Python/Python-2.7.10-intel-2015b.eb#L77) the highlighted line will be templated like `python-Python-2.7.10.tar.gz` instead of `python-dateutil-2.4.2.tar.gz`

maybe use ext_name and ext_version in exts_list to avoid this (see
http://easybuild.readthedocs.org/en/latest/Removed-functionality.html#extension-filter-template)

boegel:
This fixes a long-standing issue where using templates in `exts_list` either did not work at all, or where incorrect values were being used (i.e. the ones inherited by the 'parent').

fixes #1445, #1597, #1138

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -58,6 +58,7 @@ from easybuild.tools import config, filetools
 from easybuild.framework.easyconfig import EASYCONFIGS_PKG_SUBDIR
 from easybuild.framework.easyconfig.easyconfig import ITERATE_OPTIONS, EasyConfig, ActiveMNS, get_easyblock_class
 from easybuild.framework.easyconfig.easyconfig import get_module_path, letter_dir_for, resolve_template
+from easybuild.framework.easyconfig.templates import template_constant_dict
 from easybuild.framework.easyconfig.format.format import INDENT_4SPACES
 from easybuild.framework.easyconfig.parser import fetch_parameters_from_easyconfig
 from easybuild.framework.easyconfig.style import MAX_LINE_LENGTH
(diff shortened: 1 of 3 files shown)


boegel:
@akesandgren While working on this I realized there's another problem with templates like `%(pymajver)s` when used in the context of `multi_deps`; that issue is not fixed yet in this PR (which fixes a long-standing more general problem).

More specifically: templates like `%(pymajver)s` are currently not correctly resolved yet during the `fetch` step when `Python` is listed via `multi_deps`.

The reason for this is that the `fetch` step is not included in the list of steps that are iterated over.
Fixing that issue is not that trivial I think, which I why I prefer tackling that in a separate PR...

akesandgren:
Going in, thanks @boegel!

easybuild/framework/easyblock.py
```diff
@@ -498,17 +497,24 @@ def fetch_extension_sources(self, skip_checksums=False):
                         'options': ext_options,
                     }
 
+                    # construct dictionary with template values;
+                    # inherited from parent, except for name/version templates which are specific to this extension
+                    template_values = copy.deepcopy(self.cfg.template_values)
+                    template_values.update(template_constant_dict(ext_src))
```
ocaisa: Missing an option here, this should be
```
template_values.update(template_constant_dict(ext_src, skip_lower=False))
```

boegel: @ocaisa alternative fix via #2856

PR #2855, opened 2019-05-03
Treat components like extensions in fetch_step

ocaisa:


easybuild/framework/easyblock.py
```diff
@@ -458,68 +458,85 @@ def fetch_patches(self, patch_specs=None, extension=False, checksums=None):
...
+                    extra_options = resolve_template(extra_options, template_values)
 
-                    checksums = ext_options.get('checksums', [])
+                    checksums = extra_options.get('checksums', [])
 
                     # use default template for name of source file if none is specified
                     default_source_tmpl = resolve_template('%(name)s-%(version)s.tar.gz', template_values)
-                    fn = ext_options.get('source_tmpl', default_source_tmpl)
+                    if components:
+                        fn = resolve_template(extra_src['options'].get('sources', default_source_tmpl)[0],
```
ocaisa: Problem here is that for components this is a list, there may be more than one source file

ocaisa:
Injecting checksums into a bundle does not work, their syntax is very similar (the same even) to extensions, but they do have have the same status so that makes it hard

ocaisa:
Closing this since there is a more straightforward solution to the particular problem in https://github.com/easybuilders/easybuild-easyblocks/pull/1711

PR #2856, opened 2019-05-03
deprecate useless 'skip_lower' named argument in template_constant_dict, always define *lower templates

boegel:
For some reason unclear to me the template values are generated in two passes: once without  `%(namelower)s`, and then again with `%(namelower)s` (where the set of template values is mostly the same, except for `%(namelower)s` and `%(nameletterlower)s`).

This was introduced a long time ago as a part of a big refactor in #526, but I don't see a reason (anymore) for this.

Besides the double work being done without good reason, having the `skip_lower` named argument that is enabled by default in the `template_constant_dict` function is confusing, it leads to bugs where `template_constant_dict` is called without `skip_lower=False` which results in the `%(namelower)s` template not getting resolved (cfr. https://github.com/easybuilders/easybuild-framework/pull/2852#issuecomment-280746825)


easybuild/framework/easyconfig/templates.py
```diff
@@ -237,19 +240,19 @@ def template_constant_dict(config, ignore=None, skip_lower=True):
...
+            continue
+
+        value = config.get(name) or template_values.get(name)
+
+        if value is None:
+            continue
+        try:
+            template_values[TEMPLATE_NAMES_LOWER_TEMPLATE % {'name': name}] = value.lower()
+        except Exception:
+            _log.debug("_getitem_string: can't get .lower() for name %s value %s (type %s)", name, value, type(value))
```
akesandgren: _getitem_string??

boegel: Yeah, good point, I'll fix that...

akesandgren:
Going in, thanks @boegel!

PR #2861, opened 2019-05-06
inject dependencies specified via multi_deps before normal build dependencies

boegel:
Dependencies listed via `multi_deps`(for example `Python`) must be prepended to the list of build dependencies, to ensure that they get loaded first when setting up the build environment in the `prepare` step.

This is important if any of the build dependencies were also installed using `multi_deps`.

Example: installing `h5py` with Python 3.7.2 + 2.7.15 which requires (the Python bindings of) `pkgconfig` as a build dependency, which was also installed via `multi_deps` to support both Python 3.7.2 & 2.7.15; cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/8135


akesandgren:
Going in, thanks @boegel!

PR #2870, opened 2019-05-09
Update HierarchicalMNS for GCCcore toolchain

ocaisa:
@Micket With our custom naming scheme (related to HMNS) I was also seeing an issue with the Clang easyconfig in https://github.com/easybuilders/easybuild-easyconfigs/pull/8254 This PR fixes that for me, however the reported error (for me) was:
```
File "/tmp/eb-KByFqx/included-module-naming-schemes/easybuild/tools/module_naming_scheme/custom_hierarchical_mns.py", line 143, in det_modpath_extensions
KeyError: 'GCC'
```
This is quite different from what you reported in 
https://github.com/easybuilders/easybuild-easyconfigs/pull/8254#issuecomment-490150288
Does this PR fix your problem?

### DIFF
--- a/easybuild/tools/module_naming_scheme/hierarchical_mns.py
+++ b/easybuild/tools/module_naming_scheme/hierarchical_mns.py
@@ -33,6 +33,7 @@ import os
 import re
 from vsc.utils import fancylogger
 
+from easybuild.toolchains.gcccore import GCCcore
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.module_naming_scheme import ModuleNamingScheme
 from easybuild.tools.module_naming_scheme.toolchain import det_toolchain_compilers, det_toolchain_mpi
@@ -44,6 +45,8 @@ MPI = 'MPI'
 
 MODULECLASS_COMPILER = 'compiler'
 MODULECLASS_MPI = 'mpi'
+
+GCCCORE = GCCcore.NAME
 
 # note: names in keys are ordered alphabetically
 COMP_NAME_VERSION_TEMPLATES = {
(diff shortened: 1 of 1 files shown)


ocaisa:
Ok, I reproduced the error that was seen for @Micket , this fixes it.

akesandgren:
This change can perhaps also make it possible to combine CUDA with GCCcore instead of GCC to make the GCC-CUDA based stuff.... might be worth looking at in the future...

easybuild/tools/module_naming_scheme/hierarchical_mns.py
```diff
@@ -189,6 +192,9 @@ def det_modpath_extensions(self, ec):
 
                         if non_dummy_tc:
                             tc_comp_name, tc_comp_ver = tc_comp_info
+                            # Stick to name GCC for GCCcore
+                            if tc_comp_name == GCCCORE:
+                                tc_comp_name = 'GCC'
```
akesandgren: Are you sure this won't cause things built with GCCcore to end up in .../Compiler/GCC instead of Compiler/GCCcore

I'm not sure exactly what this affects.

ocaisa: No, it won't affect that since GCCcore is built with the dummy compiler (so the condition `non_dummy_tc` 4 lines up is not satisfied). This is only relevant to compilers built with GCCcore and only affects the searching for matches in `COMP_NAME_VERSION_TEMPLATES` (see the for loop around this block).

I'm not overly confident that this has no impact though but I don't believe it does. As far as I know Clang is the only compiler build from source using GCCcore (rather than GCC).

Micket: I just tried this out with Clang, and then with a normal package
`eb Ghostscript-9.27-GCCcore-8.2.0.eb --include-module-naming-schemes hierarchical_mns.py`
and it was placed correctly (under `software/Compiler/GCCcore/Ghostscript` )

Micket:
> lgtm, but I'd like to get confirmation from @Micket on this before merging...
@boegel
Building clang (on centos7) with this worked. I can't say I understand this part of the code, but, I have no complaints.

boegel:
Thanks for confirming the fix @Micket!

PR #2873, opened 2019-05-10
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

### DIFF
--- a/easybuild/scripts/port_easyblock.py
+++ b/easybuild/scripts/port_easyblock.py
@@ -43,6 +43,8 @@ Usage: check_code_cleanup.py
 :author: Pieter De Baets (Ghent University)
 :author: Jens Timmerman (Ghent University)
 """
+from __future__ import print_function
+
 import re
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


boegel:
@cclauss Thank you very much for your contribution!

I appreciate your effort, but the `port_easyblock.py` script is of little use now, so there's little point it making changes to it...
It was added to the repository a long time ago to help early adopters of EasyBuild to migrate to the (I believe) EasyBuild v1.0 stable release.

This script of no use anymore, so it should just be removed for the upcoming EasyBuild v4.0 release, together with a couple of other scripts located in the same directory. I'll take care of that in a separate PR.

The code we do still care about has already been updated  in the `4.x` branch to support both Python 2.x (2.6 or 2.7) and Python 3.x (3.5 or newer). We aim to release EasyBuild 4.0 from this branch sometime in the coming weeks/months (ETA is summer 2019).

cclauss:
Should we close this PR?

boegel:
Eventually, yes, but let’s keep it open for now to serve as a reminder for me to clean up those old scripts...

boegel:
`port_easyblock.py` script removed in #2897

PR #2874, opened 2019-05-10
pip install -r requirements.txt

cclauss:
https://pip.pypa.io/en/stable/reference/pip_install/?highlight=python_version#requirement-specifiers

Moving to a __requirements.txt__ file will move the package logic out of the Continuous Integration file and will allow local installs, Docker installs, and other CI installs to all utilize the same file.  The other advantage is that this approach will ease any future migration to pyproject.toml, pipenv, flit, poetry, etc.


boegel:
I fully agree that this is a lot better than the mess we have now in .travis.yml, and also see the value of this outside of Travis. Thank you very much for the effort here @cclaus!

One minor thing: I assume pip doesn’t mind to have empty lines in the requirements.txt?
I would like to have separate “blocks” to make the various packages stand out a bit more (since there is no syntax highlighting). So, an empty line after lines 6, 9, etc.

Do you mind making that trivial change?
If the tests then pass, I think we can go ahead and merge this...

cclauss:
Whitespace change made...  Let's wait on the build.

boegel:
Thank you very much for your contribution @cclauss, I really appreciate this, much cleaner...

PR #2876, opened 2019-05-12
fix Python classifiers in setup.py, should be (only) Python 2.6 & 2.7

boegel:


akesandgren:
Going in, thanks @boegel!

PR #2877, opened 2019-05-13
deprecate 'dummy' toolchain, replace with 'system' toolchain

### LINKED ISSUE #1181: Please rename the `dummy` toolchain

boegel:
(fixes #1181)

The main motivation of this change is i) to get rid of the silly/meaningless '`dummy`' name, ii) to get rid of the confusing behavior of the `dummy` toolchain of not loading dependencies in the build environment when '`dummy`' is used as a toolchain version (cfr. https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#loading-of-modules-for-dependencies-with-a-dummy-toolchain).

A big part of this PR is replacing the use of '`dummy`' with '`system`' throughout the code base and tests, so let me highlight the main changes:

* A new `system` toolchain was introduced (see also `System` toolchain definition + `SystemCompiler` compiler used by it).

* In contrast with the `dummy` toolchain, the `version` of the `system` toolchain is totally meaningless (so no more special behavior like we had with using '`dummy`' as a `version` for the `dummy` toolchain).

* The use of the `dummy` toolchain is deprecated: it will produce a big fat warning in EasyBuild 4.x, and will result in an error being raised starting with EasyBuild 5.0 (see `Toolchain.__init__` in `easybuild/tools/toolchain/toolchain.py`).
   In addition, whenever the `dummy` toolchain is specified, it will be replaced by the `system` toolchain under the covers

* The `--add-dummy-to-minimal-toolchains` configuration option has been renamed to `--add-system-to-minimal-toolchains` (so `--add-dummy-to-minimal-toolchains` is no longer a valid configuration option).

* A new function `is_system_toolchain` (+ utility method `Toolchain.is_system_toolchain()`) has been added, which should be used wherever we need to check whether or not a system toolchain is used (rather than checking the toolchain name against `DUMMY_TOOLCHAIN_NAME`).

* A new constant `SYSTEM` was added which allows for using `toolchain = SYSTEM` in easyconfig files, to make the use of a system toolchain stand out a bit more, and to make it cleaner (compared to using `toolchain = {'name': 'system', 'version': 'system'}`.

* A new (internal) constant `SYSTEM_TOOLCHAIN_NAME` was introduced, equivalent with `DUMMY_TOOLCHAIN_NAME` (but easyblocks should be updated to use `is_system_toolchain` instead).

* `dummy` is filtered from the output of `--list-toolchains` (since it's deprecated)


easybuild/tools/config.py
```diff
@@ -193,7 +193,7 @@ def mk_full_default_path(name, prefix=DEFAULT_PREFIX):
         'zip_logs',
     ],
     False: [
-        'add_dummy_to_minimal_toolchains',
+        'add_system_to_minimal_toolchains',
```
akesandgren: This one will be important to mention in the release notes!

easybuild/tools/module_naming_scheme/toolchain.py
```diff
@@ -89,8 +88,8 @@ def det_toolchain_compilers(ec):
     :param ec: a parsed EasyConfig file (an AttributeError will occur if a simple dict is passed)
     """
     tc_elems = ec.toolchain.definition()
-    if ec.toolchain.name == DUMMY_TOOLCHAIN_NAME:
-        # dummy toolchain has no compiler
+    if ec.toolchain.is_system_toolchain():
```
akesandgren: Missing import of is_system_toolchain?

easybuild/tools/package/utilities.py
```diff
@@ -112,7 +111,7 @@ def package_with_fpm(easyblock):
         cmdlist.append('--debug')
 
     deps = []
-    if easyblock.toolchain.name != DUMMY_TOOLCHAIN_NAME:
+    if not easyblock.toolchain.is_system_toolchain():
```
akesandgren: Missing import here too?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -178,14 +190,18 @@ def __init__(self, name=None, version=None, mns=None, class_constants=None, tcde
         self.mod_full_name = None
         self.mod_short_name = None
         self.init_modpaths = None
-        if self.name != DUMMY_TOOLCHAIN_NAME:
+        if not self.is_system_toolchain():
```
akesandgren: Does that really work? don't you need is_system_toolchain(self.name) ?

test/framework/toolchain.py
```diff
@@ -95,23 +96,69 @@ def test_unknown_toolchain(self):
...
+
+        self.setup_sandbox_for_intel_fftw(self.test_prefix)
+        self.modtool.prepend_module_path(self.test_prefix)
+        tc = self.get_toolchain('intel', version='2018a')
+        self.assertFalse(tc.is_system_toolchain())
+
+        # using dummy toolchain is deprecated, so to test for that we need to explicitely allow using deprecated stuff
+        error_pattern = "Use of 'dummy' toolchain is deprecated"
+        for ver in ['dummy', '']:
+            self.assertErrorRegex(EasyBuildError, error_pattern, self.get_toolchain, 'dummy', version=ver)
```
akesandgren: Is that line correct?

easybuild/tools/config.py
boegel: Yep, which is why I also mentioned it in the PR description (of course this is only for EasyBuild v4.0)

easybuild/tools/module_naming_scheme/toolchain.py
boegel: No, we have both a method in the `Toolchain` class (which doesn't need an argument since a `Toolchain` instance knows its toolchain name), and a stand-alone function (which requires a toolchain name as argument, and which is used by `Toolchain.is_system_toolchain`)

easybuild/tools/package/utilities.py
boegel: Nope, method call here

easybuild/tools/toolchain/toolchain.py
boegel: Yeah, method call here :)

test/framework/toolchain.py
boegel: What would be wrong with it?

By default, the tests run in a setup where triggering deprecated stuff results in an error being raised, which is exactly what we're checking here.

boegel: So, with EasyBuild 4.0, you'll get:

```
$ eb bzip2-1.0.6.eb
== temporary log file in case of crash /tmp/eb-Oxg9IR/easybuild-RZK2aJ.log

WARNING: Deprecated functionality, will no longer work in v5.0: Use of 'dummy' toolchain is deprecated, use 'system' toolchain instead; see http://easybuild.readthedocs.org/en/latest/Deprecated-functionality.html for more information

== processing EasyBuild easyconfig /easybuild-easyconfigs/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb
...
```

But with EasyBuild 5.0 and newer, this will result in an error, which you can already mimic using `--deprecated` (the tests basically run with `--deprecated=100000` unless told otherwise to signal unintended triggering of deprecated functionality)

```
$ eb bzip2-1.0.6.eb --deprecated=5.0
== temporary log file in case of crash /tmp/eb-7bD941/easybuild-x66BGe.log
ERROR: Failed to process easyconfig /easybuild-easyconfigs/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb: DEPRECATED (since v5.0) functionality used: Use of 'dummy' toolchain is deprecated, use 'system' toolchain instead; see http://easybuild.readthedocs.org/en/latest/Deprecated-functionality.html for more information
```

akesandgren: I didn't bother looking at what assertErrorRegex takes for argument and expected self.get_toolchain to be a function call, not the function pointer it apparently is...

PR #2886, opened 2019-05-17
bump iteration index counter at the correct time

### LINKED ISSUE #2885: self.iter_idx is bumped early
While working on https://github.com/easybuilders/easybuild-easyblocks/pull/1718 I noticed that `self.iter_idx` is bumped before the iteration is actually fully complete (see https://github.com/easybuilders/easybuild-framework/blob/706a40b77087952f3a61a200ca6f7ddf16b1525d/easybuild/framework/easyblock.py#L1556).

Can we add another variable that gets set at the beginning of the same function? This would make leveraging the value more consistent in easyblocks. For example:
```
self.curr_idx = self.iter_idx
```
done anywhere before the bump.

ocaisa:
Fixes #2885

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1561,6 +1561,8 @@ class EasyBlock(object):
         # re-enable templating before self.cfg values are used
         self.cfg.enable_templating = prev_enable_templating
 
+        # add a reference parameter for easyblocks for the current iteration
+        self.current_iteration = self.iter_idx
         # prepare for next iteration (if any)
         self.iter_idx += 1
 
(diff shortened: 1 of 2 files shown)


boegel:
@ocaisa Hmm, I'm not sure this is the best approach, I feel it can lead to continued confusion because we have two variables with different semantics.

Moreover, since `self.iter_idx` will be `>=1` after the first call to `handle_iterate_opts`, the `if self.iter_idx > 0` in `prepare_step` is wrong (but it doesn't do much harm, it just resets the toolchain also during the first iteration).

Maybe the best way is to have a `next_iter` method that (currently) simply bumps `self.iter_idx` (and spits out a log message than the iteration is done)?

PR #2887, opened 2019-05-17
don't specify --ntasks-per-node when submitting Slurm jobs

boegel:
Currently, jobs to a `Slurm` backend are being submitted with "`sbatch --nodes 1 --ntasks 5 --ntasks-per-node 5`" when "`eb --job --job-cores 5`" is used.

That works fine, until you change your mind after the jobs were submitted and you want to increase the number of cores using `scontrol update job=<jobid> NumTasks=10 NumCPUS=10`.

The `--ntasks-per-node 5` that was used at submission time can not be changed afterwards, and thus it prevents you from increasing the number of requested cores for that job (if you do try to increase the number of cores/tasks, the job gets stuck due to `BadConstraints`).

I see no need for using `--ntasks-per-node` at all, so let's just remove it...
Afaik, this shouldn't cause any problems.

### DIFF
--- a/easybuild/tools/job/slurm.py
+++ b/easybuild/tools/job/slurm.py
@@ -184,6 +184,5 @@ class SlurmJob(object):
         if cores:
             self.job_specs['nodes'] = 1
             self.job_specs['ntasks'] = cores
-            self.job_specs['ntasks-per-node'] = cores
         else:
             self.log.warn("Number of cores to request not specified, falling back to whatever Slurm does by default")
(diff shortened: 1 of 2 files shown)


akesandgren:
Yes, I agree that --ntasks-per-node should not be used. But nor should --nodes be, you should only use -n and -c and let the batch system take care of the rest.

ocaisa:
Not really, you need all the cores on the same node, `make` still has a ways before to go before it can manage multiple nodes at once!

akesandgren:
Yes, but mpi tests don't. And it's up to the user to specify sane things. EB shouldn't get in my way...
Say for instance that i have to build something that needs to run mpi during testing, and the test requires 4 MPI tasks. If I only have single core nodes then I still need -n 4 and EB should NOT add -N 1 to that.

PR #2888, opened 2019-05-19
fix order of easyconfig parameters in output generated by 'eb --avail-easyconfig-params --output-format rst'

boegel:
Without this fix, the available easyconfig parameters are in no particular order, which affects the order in the auto-updated documentation at https://easybuild.readthedocs.io/en/latest/version-specific/easyconfig_parameters.html...


akesandgren:
Going in, thanks @boegel!

PR #2897, opened 2019-05-22
remove old scripts that are no longer useful (+ minor fixes to others)

boegel:
The scripts I'm removing here were useful at some point, but no longer...


boegel:
close/re-open to trigger Travis...

PR #2898, opened 2019-05-22
make --fix-deprecated-easyconfigs a bit less strict

boegel:
Some existing easyconfigs don't have `name` and `version` in that order in the `toolchain` spec, so we need to be a bit more flexible in `--fix-deprecated-easyconfigs` (cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/8369).

### DIFF
--- a/test/framework/easyconfig.py
+++ b/test/framework/easyconfig.py
@@ -2865,11 +2865,19 @@ class EasyConfigTest(EnhancedTestCase):
         test_ec = os.path.join(self.test_prefix, 'test.eb')
         tc_regex = re.compile('^toolchain = .*', re.M)
 
-        for dummy_ver in ['dummy', '', '1.2.3']:
-            write_file(test_ec, tc_regex.sub("toolchain = {'name': 'dummy', 'version': '%s'}" % dummy_ver, toy_ectxt))
+        tc_strs = [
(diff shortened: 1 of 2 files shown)


test/framework/easyconfig.py
```diff
@@ -2865,11 +2865,19 @@ def test_fix_deprecated_easyconfigs(self):
...
+            "{'version': 'dummy', 'name': 'dummy'}",
+        ]
+
+        for tc_str in tc_strs:
+            write_file(test_ec, tc_regex.sub("toolchain = %s" % tc_str, toy_ectxt))
 
             test_ec_txt = read_file(test_ec)
-            regex = re.compile("^toolchain = {'name': 'dummy', .*$", re.M)
+            regex = re.compile("^toolchain = {.*'name': 'dummy'.*$", re.M)
             self.assertTrue(regex.search(test_ec_txt), "Pattern '%s' found in: %s" % (regex.pattern, test_ec_txt))
```
wpoely86: I'm not sure I understand this test? What exactly is being tested here?

boegel: It's a sanity check for the test itself, it checks whether the test easyconfig file really uses a dummy toolchain.

akesandgren:
Going in, thanks @boegel!

PR #2903, opened 2019-05-24
add support for specifying non-image based bootstrap agents in Singularity container recipes via --container-base-config

### LINKED ISSUE #2901: Add support for Sylabs Library as singularity source
Add
- https://cloud.sylabs.io/home

at:
- https://github.com/easybuilders/easybuild-framework/blob/master/easybuild/tools/containers/singularity.py#L44 

There is a client at:
- https://github.com/sylabs/scs-library-client

boegel:
Things that work now that did not before:

```
eb -C --container-base-config bootstrap=yum,osversion=7.6.1810 --experimental bzip2-1.0.6.eb
```

This is only partially useful for now, since the assumption is still that EasyBuild is installed in the container image (which it won't be if you start from a minimal OS), along with `pip`, Lmod, etc., and that an `easybuild` user exists that has write access to `/app/` and `/scratch/`.
I plan to fix that in a follow-up PR, but I prefer keeping the scope of PRs limited to make them easier to review/merge.

This fixes #2901 along the way too (cc @ArangoGutierrez):

```
eb -C --container-base-config bootstrap=library,from=sylabsed/examples/lolcow:latest --experimental bzip2-1.0.6.eb
```

When an existing image is being pulled from a repository like Singularity Hub (`shub`), Sylabs Container Library (`library`) or Docker Hub (`docker`), or if a local container image (`localimage`) is used as base, the assumptions w.r.t. EasyBuild, Lmod and the `easybuild` user are probably going to stand (for now)...

### DIFF
--- a/easybuild/tools/containers/singularity.py
+++ b/easybuild/tools/containers/singularity.py
@@ -162,7 +162,7 @@ class SingularityContainer(ContainerGenerator):
                 raise EasyBuildError("Unknown key for base container configuration: %s", key)
 
         # make sure correct bootstrap agent is specified
-        bootstrap = template_data['bootstrap']
+        bootstrap = template_data.get('bootstrap')
         if bootstrap:
             if bootstrap not in SINGULARITY_BOOTSTRAP_AGENTS:
                 raise EasyBuildError("Unknown value specified for 'bootstrap' keyword: %s (known: %s)",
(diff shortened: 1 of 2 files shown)


easybuild/tools/containers/singularity.py
```diff
@@ -115,26 +137,63 @@ def resolve_template_data_base_config(self):
...
+            if key in base_config_known_keys:
                 template_data[key] = value
             else:
                 raise EasyBuildError("Unknown key for base container configuration: %s", key)
 
-        if sorted(base_config_keys) != sorted(template_data.keys()):
-            raise EasyBuildError("Not all keys for base configuration were specified! Found %s, expected %s",
-                                 ', '.join(sorted(base_config_keys)), ', '.join(sorted(template_data.keys())))
+        # make sure correct bootstrap agent is specified
+        bootstrap = template_data['bootstrap']
```
wpoely86: get?

boegel: Nice catch, fixed in ffd814423

PR #2905, opened 2019-05-28
Preserve interpreter options when fixing script shebang

vanzod:
With the current implementation, the shebang is not fixed if anything is specified after the path, as in `#!/usr/bin/perl -w`.
The proposed solution captures whatever follows the interpreter path and adds it back into the new shebang. If nothing is specified, nothing is added, making it fully backwards compatible.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2167,9 +2167,12 @@ class EasyBlock(object):
                     paths = glob.glob(os.path.join(self.installdir, glob_pattern))
                     self.log.info("Fixing '%s' shebang to '%s' for files that match '%s': %s",
                                   lang, shebang, glob_pattern, paths)
-                    regex = r'^#!.*/%s[0-9.]*$' % lang
+                    # Use a regex group to capture any option passed to the interpreter
+                    # in the original shebang to preserve them in the new one.
+                    regex = r'^#!.*/%s[0-9.]?([^\.]*)$' % lang
+                    replace = r'%s\1' % shebang
                     for path in paths:
-                        apply_regex_substitutions(path, [(regex, shebang)], backup=False)
+                        apply_regex_substitutions(path, [(regex, replace)], backup=False)
 
     def post_install_step(self):
         """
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2167,9 +2167,12 @@ def fix_shebang(self):
                     paths = glob.glob(os.path.join(self.installdir, glob_pattern))
                     self.log.info("Fixing '%s' shebang to '%s' for files that match '%s': %s",
                                   lang, shebang, glob_pattern, paths)
-                    regex = r'^#!.*/%s[0-9.]*$' % lang
+                    # Use a regex group to capture any option passed to the interpreter
+                    # in the original shebang to preserve them in the new one.
+                    regex = r'^#!.*/%s[0-9.]?([^\.]*)$' % lang
```
boegel: @vanzod `[0-9.]*` should remain exactly that, not become `[0-9.]?` (`?` is zero or one repeats, while `*` is zero or more repeats).
That change is also what broke the `test_fix_shebang` test:

```
$ python -O -m test.framework.toy_build test_fix_shebang
INFO: This is (based on) vsc.install.shared_setup 0.12.7
Filtered ToyBuildTest tests using 'test_fix_shebang', retained 1/43 tests: test_fix_shebang
F
FAIL: test_fix_shebang (__main__.ToyBuildTest)
Test use of fix_python_shebang_for & co.
  File "/Volumes/work/easybuild-framework/test/framework/toy_build.py", line 2255, in test_fix_shebang
    "Pattern '%s' found in %s: %s" % (py_shebang_regex.pattern, pybin_path, pybin_txt))
AssertionError: None is not true : Pattern '^#!/usr/bin/env python\n# test$' found in /var/folders/d9/k7y8zpx91nlfdmf3mdgp8vgw0000gn/T/eb-TFWN1J/eb-AJhQw1/tmpfBKV8j/software/toy/0.0/bin/t2.py: #!/software/Python/3.6.6-foss-2018b/bin/python3.6
# test


Ran 1 test in 4.205s

FAILED (failures=1)
```

In addition, please extend the `test_fix_shebang` test to also check that shebang lines that use `/.../perl -w` are being replaced correctly.

Note that an easier way could have been to simply drop the `$` at the end in the regex, which should result in a match only up until the `/perl` (since `$` marks the end of a line).

boegel:
Thanks for the fix @vanzod!

PR #2906, opened 2019-06-03
Remove shebang options - Improve apply_regex_substitution log output

vanzod:
The problem with the changes in #2905 is that with coreutils < 0.30 everything that follows the first word in the shebang is treated as a single argument. Hence something like `#!/usr/bin/env perl -w` fails with:
```
/usr/bin/env: 'perl -w': No such file or directory
```
This fix removes any flag from the shebang.

Changes to `filetools.py` are also proposed to provide a complete log of the strings replaced by `apply_regex_substitutions`. The log output will be:
```
INFO Fixing 'perl' shebang to '#!/usr/bin/env perl' for files that match 'bin/SVDetect': ['/opt/easybuild/software/Compiler/GCC/6.4.0-2.28/SVDetect/0.8b-Perl-5.26.0/bin/SVDetect']
INFO Applying following regex substitutions to /opt/easybuild/software/Compiler/GCC/6.4.0-2.28/SVDetect/0.8b-Perl-5.26.0/bin/SVDetect: [('^#!.*/perl[0-9.]*\\s+.*', '#!/usr/bin/env perl')]
INFO Replacing line 1 in /opt/easybuild/software/Compiler/GCC/6.4.0-2.28/SVDetect/0.8b-Perl-5.26.0/scripts/BAM_preprocessingPairs.pl: '#!/usr/bin/perl -w' -> '#!/usr/bin/env perl'
```

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1041,13 +1041,11 @@ def apply_regex_substitutions(path, regex_subs, backup='.orig.eb'):
             backup_ext = ''
 
         try:
-            counter = 0
-            for line in fileinput.input(path, inplace=1, backup=backup_ext):
-                counter += 1
+            for line_id, line in enumerate(fileinput.input(path, inplace=1, backup=backup_ext)):
                 for regex, subtxt in regex_subs:
                     match = regex.search(line)
                     if match:
-                        _log.info("Replacing line %d in %s: '%s' -> '%s'", counter, path, match.group(0), subtxt)
+                        _log.info("Replacing line %d in %s: '%s' -> '%s'", (line_id + 1), path, match.group(0), subtxt)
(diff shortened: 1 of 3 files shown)


easybuild/tools/filetools.py
```diff
@@ -1041,8 +1041,13 @@ def apply_regex_substitutions(path, regex_subs, backup='.orig.eb'):
             backup_ext = ''
 
         try:
+            counter = 0
             for line in fileinput.input(path, inplace=1, backup=backup_ext):
```
boegel: You can use `enumerate` here rather than keeping track of a `counter` yourself (and I would name it `line_id` instead):

```python
for line_id, line in enumerate(fileinput.input(path, inplace=1, backup=backup_ext)):

PR #2914, opened 2019-06-06
fix crash on iterated installation when using Cray toolchains

### LINKED ISSUE #2911: Error "_set_blas_variables: blas_lib not set" in EasyBuild 3.9.1
I am getting the following error when I am trying to build LAMMPS with EasyBuild 3.9.1. 
For an extended dry run, the following is included in the logs:
```
WARNING: ignoring error '_set_blas_variables: BLAS_LIB not set'
```

Using EasyBuild 3.8.1 the build succeeds. The eb recipe is this https://github.com/eth-cscs/production/blob/master/easybuild/easyconfigs/l/LAMMPS/LAMMPS-22Aug2018-CrayGNU-18.08.eb,

ocaisa:
Fixes #2911

### DIFF
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -179,12 +179,16 @@ class ToolchainTest(EnhancedTestCase):
         self.check_vars_foss_usempi(tc)
 
     def test_cray_reset(self):
+        """Test toolchain preparation after reset for Cray* toolchain."""
+        # cfr. https://github.com/easybuilders/easybuild-framework/issues/2911
         init_config(build_options={'optarch': 'test', 'silent': True})
-        tc = self.get_toolchain('CrayGNU', version='2015.06-XC')
-        tc.set_options({'dynamic': True})
-        tc.prepare()
-        tc.reset()
-        tc.prepare()
+
+        for tcname in ['CrayGNU', 'CrayCCE', 'CrayIntel']:
+            tc = self.get_toolchain(tcname, version='2015.06-XC')
+            tc.set_options({'dynamic': True})
+            tc.prepare()
+            tc.reset()
+            tc.prepare()
 
     def test_get_variable_seq_compilers(self):
         """Test get_variable function to obtain compiler variables."""
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -178,6 +178,14 @@ class ToolchainTest(EnhancedTestCase):
         tc.prepare()
         self.check_vars_foss_usempi(tc)
 
+    def test_cray_reset(self):
+        init_config(build_options={'optarch': 'test', 'silent': True})
+        tc = self.get_toolchain('CrayGNU', version='2015.06-XC')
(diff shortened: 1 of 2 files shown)


ocaisa:
@teojgo @gppezzi In my case this causes all the BLAS and LAPACK related variables to be removed from the build environment (which you can see during the extended dry run):
```
alanc@~/EasyBuild_Git/easybuild-framework/test/framework(compiler_reset_bug)$ diff without_changes.out with_changes.out 
1,8d0
<   export BLAS_INC_DIR=''
<   export BLAS_LAPACK_INC_DIR=''
<   export BLAS_LAPACK_LIB_DIR=''
<   export BLAS_LAPACK_MT_STATIC_LIBS=''
<   export BLAS_LAPACK_STATIC_LIBS=''
<   export BLAS_LIB_DIR=''
<   export BLAS_MT_STATIC_LIBS=''
<   export BLAS_STATIC_LIBS=''
25,28d16
<   export LAPACK_INC_DIR=''
<   export LAPACK_LIB_DIR=''
<   export LAPACK_MT_STATIC_LIBS=''
<   export LAPACK_STATIC_LIBS=''
30,35d17
<   export LIBBLAS=''
<   export LIBBLAS_MT=''
<   export LIBLAPACK=''
<   export LIBLAPACK_MT=''
<   export LIBLAPACK_MT_ONLY=''
<   export LIBLAPACK_ONLY=''
```
So there is definitely a change here but since these vars were all empty, there should be no real consequence to this change. This could easily cause problems somewhere else though maybe if these empty variables are actually expected in the environment.

ocaisa:
I've trivialised the necessary change. By adding an empty element to the list, it is not equivalent to None and the build environment remains the same for multiple iterations

gppezzi:
I confirm that this fix solves our problem (https://github.com/easybuilders/easybuild-framework/issues/2911).

Thanks @ocaisa !

boegel:
@ocaisa Is there any way in which we could enhance the tests to catch this issue (and prevent from re-introducing it)?

ocaisa:
@boegel I guess we could check that all toolchain are `reset`-able...I'll see how hard that is.

test/framework/toolchain.py
```diff
@@ -178,6 +178,14 @@ def test_prepare_iterate(self):
...
         self.check_vars_foss_usempi(tc)
 
+    def test_cray_reset(self):
+        init_config(build_options={'optarch': 'test', 'silent': True})
+        tc = self.get_toolchain('CrayGNU', version='2015.06-XC')
+        tc.set_options({'dynamic': True})
+        tc.prepare()
+        tc.reset()
+        tc.prepare()
+
```
ocaisa: This set of calls raise an EasyBuildError without the change...@boegel is this enough?

boegel: That certainly covers it, yeah.

Slighty improved in e2bda0a0f, now good to go imho...

Will merge when Travis gives green light.

PR #2915, opened 2019-06-06
add support to pass additional arguments/options to 'eb' command in generated container recipe

boegel:


test/framework/containers.py
```diff
@@ -244,7 +246,18 @@ def test_end2end_singularity_recipe_config(self):
...
 
-        regex = re.compile("^id easybuild", re.M)
+        for pattern in ["id easybuild", eb_pattern]:
+            regex = re.compile('^' + pattern, re.M)
+            self.assertTrue(regex.search(txt), "Pattern '%s' should be found in: %s" % (regex.pattern, txt))
+
+        remove_file(test_container_recipe)
+
+        # options can be passed to 'eb' command in recipe via 'eb_args' keyword
+        args[-1] = 'bootstrap=yum,osversion=7.6.1810,eb_args=--debug -l'
```
ocaisa: Looks strange not to have
```
args[-1] = 'bootstrap=yum,osversion=7.6.1810,eb_args="--debug -l"'
```

ocaisa: Or does that have an impact in the regex later?

boegel: It looks strange indeed, but that's mainly because there's no shell instance that is interpreting this (so the `"` quotes wouldn't get stripped out).

Outside of the tests, you would indeed have to use quotes to preserve the space in the value for `eb_args`:

```
eb --container-config bootstrap=yum,osversion=7.6.1810,eb_args="--debug -l" ...
```

or

```
eb --container-config "bootstrap=yum,osversion=7.6.1810,eb_args=--debug -l" ...
```

The latter explains it better: we're basically only passing the value within the `"..."` to `args[-1]` in the test.

test/framework/containers.py
```diff
@@ -492,7 +505,7 @@ def test_container_config_template_recipe(self):
             "from: foobar",
             "",
             "%post",
-            "eb toy-0.0.eb --robot --installpath=/app/ --prefix=/scratch --tmpdir=/scratch/tmp",
+            "eb toy-0.0.eb --robot --debug -l",
```
wpoely86: what happened to those dropped arguments?

boegel: I'm just changing the template recipe being used here, to make it a more realistic use case we're testing with. The test would still work without this change.

Configuration options like `--installpath` are now handled through the corresponding environment variable (cfr. the changes in #2913), which allows for overriding those settings by injecting another value via the corresponding CLI option through `eb_args`.

Issue #2912, opened 2019-06-05
Using a variable named buildcmd on easyconfigs shouldn't be fatal


gppezzi:
This [vmd easyconfig](https://github.com/eth-cscs/production/blob/d542cafe39008cdae4913a56370426ba5b855220/easybuild/easyconfigs/v/VMD/VMD-1.9.3-egl.eb) was working happily until we upgraded to EB 3.9.1.

With EB 3.9.1 we get a fatal error:
```
ERROR: Failed to process easyconfig /users/perettig/VMD-1.9.3-egl.eb: You may have some typos in your easyconfig file: buildcmd -> build_cmd
``` 
Maybe it would be more useful to have a warning and proceed with the build?

boegel:
This issue is caused by the `build_cmd` easyconfig parameter that was added to the generic `ConfigureMake` easyblock via https://github.com/easybuilders/easybuild-easyblocks/pull/1628.

In this case, `buildcmd` was basically just being used as a local variable, the intention was never to use the newly `build_cmd` introduced easyconfig parameter at all (in fact, in this context it's entirely meaningless, since `CmdCp` doesn't use `build_cmd` at all).

With this in mind, it probably makes sense to let the typo detection issue a clear warning rather than producing an error, since in this case it shouldn't be fatal at all...

I'm a bit torn on whether making that change is a good idea though, since this is actually a sign of a bigger issue where innocent local variables that are not known easyconfig parameters in EasyBuild version X (and thus are basically totally ignored by EasyBuild itself, they only serve a purpose while the easyconfig file is parsed) can suddenly be assigned meaning in EasyBuild version X+1 where additional easyconfig parameters were defined. The typo detection causes trouble here is just bad luck, but it could've been "worse" in theory.

I'm not sure how we can avoid the bigger issue from occurring going forward, other than introducing a requirement that local variables follow a particular naming scheme (e.g. `local_xxx` rather than `xxx`), and make EasyBuild complain if any local variables that don't follow this scheme are being used. That would be a fairly painful change to make, since it would break any easyconfig file that uses local variables (although it may be feasible to auto-fix those).

vanzod:
@boegel This is definitely a potential source of issues. I agree on your last paragraph.

victorusu:
IMHO, @boegel, I agree with the idea of `local_xxx` and I think it should be a guideline, but I also have sympathies for an imposition by the framework to force local variable names.

[...]

PR #2917, opened 2019-06-07
Warn users that local vars in easyconfigs should be prefixed with local_

ocaisa:
For #2912, this doesn't truly fix this, just adds to the error message for typos in case the use of the variable was intentional (as a local variable within an easyconfig).

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -515,7 +515,9 @@ class EasyConfig(object):
 
         typos = [(key, guesses[0]) for (key, guesses) in possible_typos if len(guesses) == 1]
         if typos:
-            raise EasyBuildError("You may have some typos in your easyconfig file: %s",
+            raise EasyBuildError("You may have some typos in your easyconfig file: %s\n(if your use of a variable in an"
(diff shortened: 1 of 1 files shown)


boegel:
@ocaisa I think we should take this one step further (perhaps in the upcoming EasyBuild 4.0), by:

* enforcing that all local variables in easyconfig files (anything that isn't a known easyconfig parameter) start with `local_` (or equivalent), rather than silently ignoring them like we do now (cfr. https://github.com/easybuilders/easybuild-framework/blob/master/easybuild/framework/easyconfig/easyconfig.py#L530)

* fixing easyconfig files that get broken because of that change

* making sure that no easyconfig parameters with a name like `local_*` get defined (since that clashes with the above), by checking the list of known easyconfig parameters at parse time

ocaisa:
@boegel I've no problem agreeing that this should be done but I don't have the time to do that, this was a stop-gap PR so it would've easier for the user to diagnose the issue in #2912

boegel:
@ocaisa Sure, I understand. There's no reason to rush in a half-baked solution imho, this problem is quite rare (it's the first time an issue like this is reported).

boegel:
@ocaisa obsolete thanks to #2938?

PR #2918, opened 2019-06-08
disable checking of read/exec permissions when checking for availability of 'sudo' command

boegel:
On CentOS7:
```
$ which sudo
/usr/bin/sudo
$ file /usr/bin/sudo
/usr/bin/sudo: setuid executable, regular file, no read permission
$ ls -l /usr/bin/sudo
---s--x--x. 1 root root 147392 Oct 31  2018 /usr/bin/sudo
```

Without this, using `--container-build-image` fails with:

```
ERROR: sudo not found on your system
```

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -376,7 +376,7 @@ def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced
     return find_base_dir()
 
 
-def which(cmd, retain_all=False):
+def which(cmd, retain_all=False, check_perms=True):
     """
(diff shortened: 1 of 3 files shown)


easybuild/tools/filetools.py
```diff
@@ -376,7 +376,7 @@ def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced
     return find_base_dir()
 
 
-def which(cmd, retain_all=False):
+def which(cmd, retain_all=False, check_perms=True):
```
wpoely86: document new argument

boegel: fixed in 48f0323b8

PR #2928, opened 2019-06-12
stick to Ubuntu Trusty when testing with Python 2.6 in Travis

boegel:


migueldiascosta:
Going in, thanks @boegel!

PR #2930, opened 2019-06-14
MKL: add -L$EBROOTIMKL/lib/intel64 to LDFLAGS

bartoldeman:
It tried to add -L$EBROOTIMKL/compiler/lib/intel64 instead but that
path is only valid for certain old versions: 10.3.4 .. 11.0.x
(see imkl easyblock).

### DIFF
--- a/easybuild/toolchains/linalg/intelmkl.py
+++ b/easybuild/toolchains/linalg/intelmkl.py
@@ -155,9 +155,9 @@ class IntelMKL(LinAlg):
             else:
                 self.BLAS_LIB_DIR = ['mkl/lib/intel64']
                 if ver >= LooseVersion('10.3.4') and ver < LooseVersion('11.1'):
-                    self.BLAS_LIB_DIR += ['compiler/lib/intel64']
+                    self.BLAS_LIB_DIR.append('compiler/lib/intel64')
                 else:
-                    self.BLAS_LIB_DIR += ['lib/intel64']
+                    self.BLAS_LIB_DIR.append('lib/intel64')
 
             self.BLAS_INCLUDE_DIR = ['mkl/include']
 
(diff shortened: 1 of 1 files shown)


boegel:
Do you have a specific reason/use case for this @bartoldeman, or is there a particular problem you fixed by making this change?

easybuild/toolchains/linalg/intelmkl.py
```diff
@@ -152,7 +153,11 @@ def _set_blas_variables(self):
                 raise EasyBuildError("_set_blas_variables: 32-bit libraries not supported yet for IMKL v%s (> v10.3)",
                                      found_version)
             else:
-                self.BLAS_LIB_DIR = ['mkl/lib/intel64', 'compiler/lib/intel64' ]
+                self.BLAS_LIB_DIR = ['mkl/lib/intel64']
+                if ver >= LooseVersion('10.3.4') and ver < LooseVersion('11.1'):
+                    self.BLAS_LIB_DIR += ['compiler/lib/intel64']
```
boegel: please use `.append('...')` rather than `+= ['...']`

bartoldeman:
@boegel the issue is the same as in https://github.com/easybuilders/easybuild-framework/pull/2931

what happens is that the `imkl` module includes this path (which has compiler support libraries that are used by MKL such as libiomp5, not the main MKL libraries) in `LIBRARY_PATH` and `LD_LIBRARY_PATH`, but the `LDFLAGS` have higher priority if they are used. Without this change there is an inconsistency between `LIBRARY_PATH` and `LDFLAGS`: with `icc`, the icc-supplied libraries have priority, and with `gcc` it will use LIBRARY_PATH instead of LDFLAGS to find the `imkl` support libraries.

Usually the issue is masked by `LD_LIBRARY_PATH` at runtime, but if you use rpath without LD_LIBRARY_PATH, and one shared object has the imkl libiomp5 first and the other icc libiomp5 first it gets messy.

bartoldeman:
@boegel ping?

boegel:
re-triggering Travis tests by close & re-open...

PR #2937, opened 2019-06-26
honor --tmp-logdir when determining location of temporary log file

### LINKED ISSUE #2935: `--tmp-logdir` not working
When setting `--tmp-logdir`, I was expecting the build logs etc. to be placed into the directory. However, nothing of the sort happens: The logs are still put into `--tmpdir`. A simple example:

```bash
mkdir $HOME/tmplogs
eb -l --debug --tmp-logdir=$HOME/tmplogs zlib-1.2.11.eb
```

Nothing is ever in `$HOME/tmplogs` during or after the build.

Bootstrapped EasyBuild 3.9.2 on CentOS 7.6.

ocaisa:
Fixes #2935

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -65,7 +65,8 @@ from easybuild.tools.config import DEFAULT_PATH_SUBDIRS, DEFAULT_PKG_RELEASE, DE
 from easybuild.tools.config import DEFAULT_PNS, DEFAULT_PREFIX, DEFAULT_REPOSITORY, EBROOT_ENV_VAR_ACTIONS, ERROR
 from easybuild.tools.config import FORCE_DOWNLOAD_CHOICES, IGNORE, JOB_DEPS_TYPE_ABORT_ON_ERROR
 from easybuild.tools.config import JOB_DEPS_TYPE_ALWAYS_RUN, LOADED_MODULES_ACTIONS, WARN
-from easybuild.tools.config import get_pretend_installpath, init, init_build_options, mk_full_default_path
+from easybuild.tools.config import get_build_log_path, get_pretend_installpath
+from easybuild.tools.config import init, init_build_options, mk_full_default_path
 from easybuild.tools.configobj import ConfigObj, ConfigObjError
 from easybuild.tools.docs import FORMAT_TXT, FORMAT_RST
 from easybuild.tools.docs import avail_cfgfile_constants, avail_easyconfig_constants, avail_easyconfig_licenses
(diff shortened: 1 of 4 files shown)


easybuild/tools/options.py
```diff
@@ -1293,7 +1294,7 @@ def set_up_configuration(args=None, logfile=None, testing=False, silent=False):
     # initialise logging for main
     log, logfile = init_logging(logfile, logtostdout=options.logtostdout,
                                 silent=(testing or options.terse or search_query or silent),
-                                colorize=options.color)
+                                colorize=options.color, tmp_logdir=get_build_log_path())
```
ocaisa: This is failing with
```
  File "/home/alanc/EasyBuild_Git/easybuild-framework/easybuild/main.py", line 457, in <module>
    main()
  File "/home/alanc/EasyBuild_Git/easybuild-framework/easybuild/main.py", line 193, in main
    eb_go, cfg_settings = set_up_configuration(args=args, logfile=logfile, testing=testing)
  File "/home/alanc/EasyBuild_Git/easybuild-framework/easybuild/tools/options.py", line 1297, in set_up_configuration
    colorize=options.color, tmp_logdir=get_build_log_path())
  File "/home/alanc/EasyBuild_Git/easybuild-framework/easybuild/tools/config.py", line 639, in get_build_log_path
    if variables['tmp_logdir'] is not None:
  File "/home/alanc/EasyBuild_Git/vsc-base/lib/vsc/utils/missing.py", line 277, in __getitem__
    raise KeyError(err)
KeyError: KeyError('tmp_logdir',)
```

boegel: It's a chicken-egg situation I guess...

You're trying to query the configuration before it's set up.

boegel: @ocaisa How about passing down `options.tmp_logdir`?

test/framework/config.py
```diff
@@ -661,6 +661,14 @@ def test_log_path(self):
         self.assertTrue(regex.match(res), "Pattern '%s' matches '%s'" % (regex.pattern, res))
         self.assertEqual(log_file_format(), 'log.txt')
 
+    def test_get_build_log_path(self):
+        """Test for build_log_path()"""
+        init_config()
+        self.assertEqual(get_build_log_path(), tempfile.gettempdir())
+        build_log_path = os.path.join(os.getcwd(), 'chicken')
```
boegel: Please use `self.test_prefix` here, to avoid creating a directory named `chicken` somewhere and making people wonder where it comes from (since the relation between `chicken` and EasyBuild isn't that clear).

`self.test_prefix` is a temporary directory unique to each test which gets cleaned automatically...

easybuild/tools/options.py
ocaisa: Yup, works, PR updated

boegel:
Thanks a lot for the fix @ocaisa!

PR #2940, opened 2019-07-03
Allowing dict-type checksums

edmondac:
Added the ability for checksums to be a dictionary, of:
key=filename
value=checksum or checksums

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -331,10 +331,12 @@ class EasyBlock(object):
                 return checksums[index]
             else:
                 return None
+        elif isinstance(checksums, dict):
+            return checksums.get(filename)
         elif checksums is None:
             return None
         else:
-            raise EasyBuildError("Invalid type for checksums (%s), should be list, tuple or None.", type(checksums))
+            raise EasyBuildError("Invalid type for checksums (%s), should be list, tuple, dict or None.", type(checksums))
 
     def fetch_sources(self, sources=None, checksums=None):
         """
@@ -1812,10 +1814,16 @@ class EasyBlock(object):
             msg += "vs %d checksums" % checksum_cnt
             checksum_issues.append(msg)
 
-        for fn, checksum in zip(sources + patches, checksums):
-            if not is_sha256_checksum(checksum):
-                msg = "Non-SHA256 checksum found for %s: %s" % (fn, checksum)
-                checksum_issues.append(msg)
+        if isinstance(checksums, (tuple, list)):
+            for fn, checksum in zip(sources + patches, checksums):
+                if not is_sha256_checksum(checksum):
+                    msg = "Non-SHA256 checksum found for %s: %s" % (fn, checksum)
+                    checksum_issues.append(msg)
+        elif isinstance(checksums, dict):
+            for fn, checksum in checksums.items():
+                if not is_sha256_checksum(checksum):
+                    msg = "Non-SHA256 checksum found for %s: %s" % (fn, checksum)
+                    checksum_issues.append(msg)
 
         return checksum_issues
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -336,7 +336,8 @@ class EasyBlock(object):
         elif checksums is None:
             return None
         else:
-            raise EasyBuildError("Invalid type for checksums (%s), should be list, tuple, dict or None.", type(checksums))
+            raise EasyBuildError("Invalid type for checksums (%s), should be list, tuple, dict or None.",
+                                 type(checksums))
 
     def fetch_sources(self, sources=None, checksums=None):
         """
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -514,8 +514,7 @@ SANITY_CHECK_PATHS_DICT = (dict, as_hashable({
     'opt_keys': [],
     'req_keys': ['files', 'dirs'],
 }))
-CHECKSUMS_DICT = (dict, as_hashable({'elem_types': [STRING_OR_TUPLE_LIST]}))
-CHECKSUMS = (list, as_hashable({'elem_types': [STRING_OR_TUPLE_LIST, CHECKSUMS_DICT]}))
+CHECKSUMS = (list, as_hashable({'elem_types': [STRING_OR_TUPLE_LIST, dict]}))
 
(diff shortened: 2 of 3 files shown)


edmondac:
See also https://github.com/easybuilders/easybuild-easyconfigs/pull/8136 and https://github.com/easybuilders/easybuild-easyblocks/pull/1764

easybuild/framework/easyblock.py
```diff
@@ -331,10 +331,12 @@ def get_checksum_for(self, checksums, filename=None, index=None):
                 return checksums[index]
             else:
                 return None
+        elif isinstance(checksums, dict):
+            return checksums.get(filename)
```
boegel: @edmondac Please ensure proper error reporting if `filename` is not a key in `checksums`

easybuild/framework/easyblock.py
```diff
@@ -331,10 +331,12 @@ def get_checksum_for(self, checksums, filename=None, index=None):
...
         elif checksums is None:
             return None
         else:
-            raise EasyBuildError("Invalid type for checksums (%s), should be list, tuple or None.", type(checksums))
+            raise EasyBuildError("Invalid type for checksums (%s), should be list, tuple, dict or None.", type(checksums))
```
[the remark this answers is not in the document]
boegel: One way to fix this:

```python
error_msg = "Invalid type for checksums (%s), should be list, tuple, dict or None." % type(checksums)

File: easybuild/framework/easyblock.py [hunk 3]
```diff
@@ -1812,10 +1814,16 @@ def check_checksums_for(self, ent, sub='', source_cnt=None):
...
+        if isinstance(checksums, (tuple, list)):
+            for fn, checksum in zip(sources + patches, checksums):
+                if not is_sha256_checksum(checksum):
+                    msg = "Non-SHA256 checksum found for %s: %s" % (fn, checksum)
+                    checksum_issues.append(msg)
+        elif isinstance(checksums, dict):
+            for fn, checksum in checksums.items():
+                if not is_sha256_checksum(checksum):
+                    msg = "Non-SHA256 checksum found for %s: %s" % (fn, checksum)
+                    checksum_issues.append(msg)
```
boegel: This doesn't look correct...

`checksums` here is the full list of checksums, so it should be something like:

```python
for fn, checksum in zip(sources + patches, checksums):
    if isinstance(checksum, basestring):
        if not is_sha256_checksum(checksum):
            ...
    elif isinstance(checksum, dict):
        non_sha256_checksums = [(fn, c) for (fn, c) in checksum.values() if is_sha256_checksum(c)]
        if non_sha256_checksums:
            msg = "Non-SHA256 checksums found: %s" % non_sha256_checksums
            checksum_issues.append(msg)
```

(probably needs a bit more tweaking for proper formatting)

easybuild/framework/easyconfig/types.py
```diff
@@ -420,20 +420,39 @@ def to_dependencies(dep_list):
...
+            if isinstance(checksum, basestring):
+                res.append(checksum)
+            elif isinstance(checksum, (list, tuple)):
+                # 2 elements + only string/int values => a checksum tuple
+                if len(checksum) == 2 and all(isinstance(x, (basestring, int)) for x in checksum):
+                    res.append(tuple(checksum))
+                else:
+                    res.append(to_checksums(checksum))
+
+    elif isinstance(checksums, dict):
```
boegel: This check should be *in* the for loop, `checksums` is the overarching list of checksums, you intend to add support for `dict` values *in* `checksums`, right?

edmondac: Actually, I had intended to let the top-level checksums variable be a dict... since that's the only way that you can have e.g. more individual checksums than source files and patches...

edmondac: I've not answered the previous comments, since this one seems crucial to agree first!

boegel: @edmondac This is what I had in mind:

```python
sources = [
    'foo-%(version)s.tar.gz',
    'example-%(version)s-%(arch)s.tar.gz',
]
patches = ['example.patch']
checksums = [
    'def47ee13a30a7698718c4f9ee1569406c8135ffb7febec5158c015a0b0b7590',  # foo-1.2.3.tar.gz
    {
        'example-1.2.3-x86_64.tar.gz': 'f0d5f7cd6137637ef9065431bebeb14333ef13018458135c74024c16161ca1ec',
        'example-1.2.3-ppc64le.tar.gz': '5f24fd6084268f5ffea0efd2027fb4319ba7a6bd3c3659bca1d283cba06b2864',
    },
    '7ce1913278203c45a667df9b7cb6192f46d2a6516fa6e1c1e269a27473d8413c',  # example.patch
]
```

Here, the 2nd item in `checksums` is a dict that holds two checksums for the 2nd source file, depending on the `%(arch)s` value.

What you had in mind is probably this:

```python
checksums = {
    'foo-1.2.3.tar.gz': 'def47ee13a30a7698718c4f9ee1569406c8135ffb7febec5158c015a0b0b7590',
    'example-1.2.3-x86_64.tar.gz': 'f0d5f7cd6137637ef9065431bebeb14333ef13018458135c74024c16161ca1ec',
    'example-1.2.3-ppc64le.tar.gz': '5f24fd6084268f5ffea0efd2027fb4319ba7a6bd3c3659bca1d283cba06b2864',
    'example.patch': '7ce1913278203c45a667df9b7cb6192f46d2a6516fa6e1c1e269a27473d8413c',
}
```

That would work too, but it's more verbose since the filenames of *all* sources/patches need to be mentioned again, even if there's only one checksum for them.

There's another option, which is to extend the "alternative" format we already support for `sources` to also be able to list `checksums` directly together with the `sources` (see https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#common-easyconfig-param-sources-alt):

```python
sources = [
    'foo-%(version)s.tar.gz',
    {
        'filename': 'example-%(version)s-%(arch)s.tar.gz',
        'checksums': {
            'example-1.2.3-x86_64.tar.gz': 'f0d5f7cd6137637ef9065431bebeb14333ef13018458135c74024c16161ca1ec',
            'example-1.2.3-ppc64le.tar.gz': '5f24fd6084268f5ffea0efd2027fb4319ba7a6bd3c3659bca1d283cba06b2864',
        },
    },
]
```

I think both the ones I had in mind make sense (it comes down to personal preference I guess), and they're probably relatively equally easy to add support for.

edmondac: Ahah - got it. Yes, I can do the first one. I'll probably make a fresh branch for it next week...

edmondac:
OK - https://github.com/easybuilders/easybuild-framework/pull/2946 will replace this.

PR #2946, opened 2019-07-11
add support for specifying checksum via dict providing a filename-to-checksum mapping

edmondac:
New approach to dict-based checksums, from https://github.com/easybuilders/easybuild-framework/pull/2940#discussion_r300823895


edmondac:
https://github.com/easybuilders/easybuild-easyconfigs/pull/8136 has an example easyconfig (CUDA) for this.

ocaisa:
I'd be reluctant to merge this without touching base with @boegel first (so it may have to wait).

Anyway, it is definitely going to require some tests, I guess the architecture dependent have a special home in the test suite?

ocaisa:
> Anyway, it is definitely going to require some tests, I guess the architecture dependent have a special home in the test suite?
If they don't then they probably should have since you have other PRs in the pipeline

easybuild/tools/filetools.py
```diff
@@ -760,6 +760,17 @@ def verify_checksum(path, checksums):
         checksums = [checksums]
 
     for checksum in checksums:
+        if isinstance(checksum, dict):
+            filename = os.path.split(path)[-1]
```
boegel: You should use `os.path.basename` here:

```python
filename = os.path.basename(path)
```

Or even better, do this way on top of this function so you can also use `filename` above.

easybuild/tools/filetools.py
```diff
@@ -760,6 +760,17 @@ def verify_checksum(path, checksums):
...
+            if filename in checksum:
+                checksum = checksum[filename]
+            else:
+                # No checksum for this file
+                if build_option('enforce_checksums'):
+                    raise EasyBuildError("Missing checksum for %s", os.path.basename(path))
+                else:
+                    continue
+
         if isinstance(checksum, basestring):
```
boegel: This should be an `elif` now?

I would also prefer having the more common case of a checksum as a string considered first, so:

```python
if isinstance(checksum, basestring):
    ...
elif isinstance(checksum, dict):
    ...

File: easybuild/tools/filetools.py [hunk 3]
```diff
@@ -760,6 +760,17 @@ def verify_checksum(path, checksums):
...
 
     for checksum in checksums:
+        if isinstance(checksum, dict):
+            filename = os.path.split(path)[-1]
+            if filename in checksum:
+                checksum = checksum[filename]
+            else:
+                # No checksum for this file
+                if build_option('enforce_checksums'):
+                    raise EasyBuildError("Missing checksum for %s", os.path.basename(path))
```
boegel: You can use `filename` here

easybuild/tools/filetools.py
```diff
@@ -760,6 +760,17 @@ def verify_checksum(path, checksums):
...
+                else:
+                    continue
```
boegel: This can be trimmed:

```python
if filename in checksum:
    checksum = checksum[filename]
elif build_option('enforce_checksums'):
    raise EasyBuildError("Missing checksum for %s", filename)
else:
    continue
```

I'm not sure the `continue` part makes sense here though...

We should probably just set `checksum` to `None`, and let it fail elsewhere?

easybuild/framework/easyconfig/types.py
```diff
@@ -495,10 +500,17 @@ def ensure_iterable_license_specs(specs):
     'opt_keys': [],
     'req_keys': ['files', 'dirs'],
 }))
-CHECKSUMS = (list, as_hashable({'elem_types': [STRING_OR_TUPLE_LIST]}))
-
-CHECKABLE_TYPES = [CHECKSUMS, DEPENDENCIES, DEPENDENCY_DICT, TOOLCHAIN_DICT, SANITY_CHECK_PATHS_DICT,
-                   STRING_OR_TUPLE_LIST, TUPLE_OF_STRINGS]
+CHECKSUM = (list, as_hashable({'elem_types': [str, TUPLE_OF_STRINGS]}))
```
boegel: When can a single checksum be a list of strings? It's either a string or a tuple of strings, not a list?

easybuild/framework/easyconfig/types.py
```diff
@@ -434,6 +434,11 @@ def to_checksums(checksums):
                 res.append(tuple(checksum))
             else:
                 res.append(to_checksums(checksum))
+        elif isinstance(checksum, dict):
+            d = {}
+            for k, v in checksum.items():
+                d[k] = to_checksums(v)
+            res.append(d)
```
boegel: Please avoid using single-letter variables outside of list comprehensions. `k` should be `key`, `v` should be `value`, `d` should be `...` (not creative enough to fill in the latter myself, `checksums` is clearly not an option here)

easybuild/framework/easyconfig/types.py
edmondac: Not sure what to change this to...?

boegel: So, an element in `checksums` can be either:

(1) a single string
(2) a single 2-tuple with (<type>, <checksum>)
(3) once #2958 is merged, a single n-tuple with alternative checksums
(4) a list of checksums that should *all* pass (where each checksum can be in formats 1-3)
(5) a dict with filename-to-checksum mappings (this PR)

So, with that in mind, I'm not sure if the original `CHECKSUMS` here actually was correct, since that only (partially) covers (4)?

I need to check this again when I'm more awake. :)

boegel:
One other thing to take into account is how this will affect `--inject-checksums`...

edmondac:
Running `eb --inject-checksums=sha256 --force CUDA-10.1.105.eb` results in the following change (on x86):

```
-checksums = [
-    {
-        '%(namelower)s_%(version)s_418.39_linux.run': '33ac60685a3e29538db5094259ea85c15906cbd0f74368733f4111eab6187c8f',
-        '%(namelower)s_%(version)s_418.39_linux_ppc64le.run': 'b81290b834483847e317ff4eafacc14d96bae8ad873aa9b1adddf2567318d658',
-    }
-]
+checksums = ['33ac60685a3e29538db5094259ea85c15906cbd0f74368733f4111eab6187c8f']
```

Personally, I think that's fine since `--force` has been used. What do you think @boegel ?

boegel:
@edmondac It would be nice to make `--inject-checksums` a bit smarter so it maintains a dict entry and simply updates it rather than blowing it away.

That shouldn't block this PR though, so let's open an issue on that to remind ourselves?

I'll go ahead and merge this to be included in EasyBuild v3.9.4, enhancing `--inject-checksums` can be done separately since that's not life-threatening...

boegel:
@edmondac I started working on enhancing `--inject-checksums` to support the new formats for checksums (the `dict`-typed values for which support was added in this PR + the `tuple`-typed values to specify alternative checksums, cfr. #2958), but I quickly realized that it doesn't make much sense...

Usually `--inject-checksums` is used after updating a software version, so the checksums that are there in the easyconfig file are basically useless, and should be replaced (which is exactly what `--inject-checksums` already does).

If a `dict` value is specified for the checksum of a particular source file, then there's little else that `--inject-checksums` can do to just blow it away and replace it with the one checksum it obtained for the filename it obtained on the current system...

If a tuple of checksums was in place that specified alternative checksums for a particular file (cfr. #2958), then I was looking into retaining those checksums and simply adding the obtained checksum as an additional alternative.
But in the usual context that `--inject-checksums` is used in (after updating a software version), that doesn't make any sense at all, since the checksums that were there are simply *not* valid anymore...

In short, I think the best we can do is to leave `--inject-checksums` as is: let it compute the checksum for the files it finds, and hard inject those, replacing whatever was there.

If that was a checksum specified via a `dict`-typed value, that will simply require human intervention to figure out the other checksums for the other filenames somehow...

PR #2955, opened 2019-08-06
be more forgiving in check_easy_install_cmd function of bootstrap script w.r.t. harmless warnings in output of 'easy_install --version'

boegel:
This helps to debug a failing bootstrap in case harmless locale warnings are included in the output of `easy_install --version`...


akesandgren:
@boegel conflict resolution needed

boegel:
EasyBuild bootstrap script was removed in #4233, so this is no longer relevant.

PR #2956, opened 2019-08-06
enhance 'completed' message with how much time was needed the installation

boegel:
Before:

```
== COMPLETED: Installation ended successfully
```

now:

```
== COMPLETED: Installation ended successfully (took 5 sec)
```

For long-running builds the time will be shown in minutes/seconds or hours/minutes/seconds (not days), see example outputs in tests.

### DIFF
--- a/easybuild/tools/utilities.py
+++ b/easybuild/tools/utilities.py
@@ -183,3 +183,22 @@ def trace_msg(message, silent=False):
     """Print trace message."""
     if build_option('trace'):
         print_msg('  >> ' + message, prefix=False)
+
+
+def time2str(time_secs):
+    """Return string representing provided amount of time (in seconds) in human-readable form."""
+    res = None
+    if time_secs < 60:
+        res = '%s sec' % time_secs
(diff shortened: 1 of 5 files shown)


easybuild/tools/utilities.py
```diff
@@ -183,3 +184,28 @@ def trace_msg(message, silent=False):
...
+
+
+def time2str(delta):
+    """Return string representing provided datetime.timedelta value in human-readable form."""
+    res = None
+
+    if not isinstance(delta, datetime.timedelta):
+        raise EasyBuildError("Incorrect value type provided to time2str, should be datetime.timedelta: %s", type(delta))
+
+    delta_secs = delta.days * 3600 * 24 + delta.seconds + delta.microseconds / 10**6
```
wpoely86: `total_seconds()` ?

wpoely86: A py2.7 only, nevermind

boegel: Jup, indeed... :)

PR #2968, opened 2019-08-14
replace --strict-local-var-naming with the more flexible --local-var-naming-check={error,log,warn}

boegel:
Still `WIP` because `error-eb` and `warn-eb` are not working as suggested yet...


boegel:
@vanzod Is this what you had in mind (cfr. our discussion during the last EB conf call)?

Easiest way to digest this is to look at the help text added in `easybuild/tools/options.py` first ;)

Issue #2546, opened 2018-08-10
update-pr fails to find where to add a new patch when the patch belongs to a bundle extension


akesandgren:
Example:
```
easyblock = 'Bundle'

name = 'mordecai'

exts_defaultclass = 'PythonPackage'

exts_list = [
    ('cymem', '1.31.2', {
        'source_urls': ['https://pypi.python.org/packages/source/c/cymem'],
        'checksums': ['f06d9b50da0474d7405674d8101c319d89a17d33792d6d429fe3d5c64f0d9df1'],
    }),
    (name, version, {
        'patches': ['mordecai-2.0.1_fix_bad_path_handling.patch'],
        'source_urls': ['https://pypi.python.org/packages/source/%(nameletter)s/%(name)s'],
    }),
]
```
when using --update-pr to add that patch it fails with:
```
Matching easyconfig for mordecai-2.0.1_fix_bad_path_handling.patch not found on the first try: scanning all easyconfigs to determine where patch file belongs (this may take a while)...
2259 of 2259 easyconfigs checked
ERROR: Failed to determine software name to which patch file mordecai-2.0.1_fix_bad_path_handling.patch relates
```

smoors:
I was bitten by the same bug. The same problem occurs when using `--new-pr`.

jose-d:
from obvious reasons reproducible with `PythonBundle` too.

boegel:
Bug reproduced in tests and fixed in #2969...

akesandgren:
Now solved with #2969

PR #2969, opened 2019-08-15
also consider extension patches when determining for which easyconfig a given patch file is intended in --new-pr/--update-pr

boegel:
fix for #2546


akesandgren:
Going in, thanks @boegel!

PR #2970, opened 2019-08-15
unset $CDPATH early on if it is defined (fixes #2944)

### LINKED ISSUE #2944: setting CDPATH breaks libxsmm build
If you have `CDPATH` environmental variable, it breaks the build of `libxsmm`. (version 1.8.3 confirmed, second run, when shared objects created from static libraries). It is probably caused by `patsubst` make function or automake or something else, as issuing `cd` command might generate an output if `CDPATH` is set. It makes sense to unset `CDPATH` as it might cause problems in other places as well.

boegel:


akesandgren:
Going in, thanks @boegel!

PR #2979, opened 2019-08-22
use 'command -v' to avoid requiring 'which' in 'eb' command

### LINKED ISSUE #2976: Undocumented hard requirement for the `which` command
The `eb` wrapper script has a hard requirement on the `which` command to figure out what Python to run, but the command is not present on my container installation of CentOS 7.6 pulled from Docker.

This results in `ERROR: python not available in $PATH?` errors during sanity check when installing EasyBuild itself, or at run-time for EasyBuild-develop.

This should probably be documented in the guide as mandatory.

boegel:
fixes #2976

cc @zao

### DIFF
--- a/eb
+++ b/eb
@@ -47,7 +47,7 @@ for python_cmd in ${EB_PYTHON} 'python2' 'python3' 'python'; do
 
     verbose "Considering '$python_cmd'..."
 
-    which $python_cmd &> /dev/null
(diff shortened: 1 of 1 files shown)


boegel:
@akesandgren target branch fixed

akesandgren:
Going in, thanks @boegel!

PR #2980, opened 2019-08-22
first attempt at using GitHub CI (WIP)

boegel:


.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,12 @@
+name: EasyBuild framework unit tests
+on: pull_request
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [2.7, 3.5, 3.6, 3.7]
```
boegel: Hmm, this doesn't seem to be working as expected... The `python` command that `eb` picks up on is `2.7.15+` for all 4 of these environment (and `eb` considers using `python3` before `python`).

boegel: fixed in c067736

boegel:
Will play around some more with this in my own account, since I also got access to GitHub Actions there now...

PR #2981, opened 2019-08-22
include docs URL in warning/error about non-conforming names of local variables

boegel:
It'll have to be the `4.x` URL for now, should be changed to `latest` shortly before the EasyBuild v4.0.0 release...


boegel:
replaced with #2984

PR #2984, opened 2019-08-27
include docs URL in warning/error about non-conforming names of local variables

boegel:
replacement for #2981, now to correct target branch (`develop`, not `4.x`)


migueldiascosta:
Going in, thanks @boegel!

PR #2985, opened 2019-08-27
fix broken --from-pr tests 

boegel:
Tests for `--from-pr` got broken due to archiving the eaysconfigs that were used in the easyconfigs PR that was being used in these tests (since for a closed/merged PR, `--from-pr` grabs the easyconfigs from the `develop` branch instead).

Fixed by switching to a more recent PR, where the easyconfigs are still available in `develop` (and will be for a while).

I had to add a couple of dummy easyblocks since `ConfigureMake` used to be used for `FFTW` and `OpenBLAS`, but not anymore; nor will it be for `OpenMPI` soon (see https://github.com/easybuilders/easybuild-easyblocks/pull/1789).

Tests involving GitHub are skipped for PRs (to keep the GitHub token that is required to test them safe), but this does in fact fix the broken tests, see the results for testing the branch used for this PR at https://travis-ci.org/boegel/easybuild-framework/builds/577478785


test/framework/robot.py
```diff
@@ -705,21 +702,21 @@ def test_det_easyconfig_paths_from_pr(self):
         gompi_2015a_txt = '\n'.join([
```
migueldiascosta: for consistency, variable name should now be `gompi_2018b_txt`?

test/framework/robot.py
```diff
@@ -705,21 +702,21 @@ def test_det_easyconfig_paths_from_pr(self):
...
             "name = 'gompi'",
-            "version = '2015a'",
+            "version = '2018b'",
             "versionsuffix = '-test'",
             "homepage = 'foo'",
             "description = 'bar'",
             "toolchain = SYSTEM",
         ])
-        write_file(os.path.join(self.test_prefix, 'gompi-2015a-test.eb'), gompi_2015a_txt)
+        write_file(os.path.join(self.test_prefix, 'gompi-2018b-test.eb'), gompi_2015a_txt)
```
migueldiascosta: see above

test/framework/robot.py
boegel: good point, fixed in f6ae5c688

PR #2987, opened 2019-08-29
run contrib style checks when using new/update/preview_pr (WIP)

migueldiascosta:
is there a reason not to do this? Anything that doesn't pass these checks at this point will make the tests fail later...

I suppose one could allow overriding with `-f`, if there is some corner case where one would have a reason to submit/update a PR that doesn't pass the tests (yet)

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -331,8 +331,10 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     easyconfigs, generated_ecs = parse_easyconfigs(paths, validate=not options.inject_checksums)
 
     # handle --check-contrib & --check-style options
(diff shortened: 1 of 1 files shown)


migueldiascosta:
this will make at least `test_new_pr` fail precisely because it uses easyconfigs that don't pass the contrib style checks (which means those can actually be used to test this enchancement :) )

I will work on the tests if people agree this is a good thing

migueldiascosta:
I guess one problem with this is that it requires `pycodestyle` or `pep8`, but while I understand that it is an optional dependency for using `eb` in general, it would make sense (?) to make it a requirement for using github integration, especially `--new-pr` and `--update-pr` (it already requires `keyring`, `git`/'GitPython', etc....)

boegel:
close/re-open to get updated CI results

PR #2988, opened 2019-08-29
stop trying to use setuptools.setup in setup.py, always use distutils.core.setup instead

### LINKED ISSUE #2941: try to avoid requiring setuptools to install EasyBuild
cfr. #2428

boegel:
Using `setuptools` in `setup.py` has done more harm than good. It's frequently a cause of problems because of old/ancient versions being available in the OS, broken versions floating around everywhere, etc.
One recent example is https://lists.ugent.be/wws/arc/easybuild/2019-08/msg00057.html

In addition, it yields little to no benefit over just installing with the `setup` function from `distutils.core` instead.

So, let's abolish the `setuptools` requirement fully, following no longer requiring it as a runtime dependency (cfr. #2836).

(fixes #2941)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,7 @@ before_install:
 install:
     # install environment modules or Lmod
     - export INSTALL_DEP=$TRAVIS_BUILD_DIR/easybuild/scripts/install_eb_dep.sh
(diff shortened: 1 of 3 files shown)


migueldiascosta:
Going in, thanks @boegel!

PR #2994, opened 2019-09-06
Add support for iccifort compiler-only toolchain without icc/ifort deps.

bartoldeman:
Relates to easybuilders/easybuild-easyconfigs#8879 and easybuilders/easybuild-easyblocks#1799.

The toolchain components need to be adjusted dynamically depending on 'moduleclass' (which is also used by HMNS to determine if this is a compiler-toolchain that needs MODULEPATH extensions.

### DIFF
--- a/easybuild/toolchains/compiler/inteliccifort.py
+++ b/easybuild/toolchains/compiler/inteliccifort.py
@@ -34,6 +34,7 @@ from distutils.version import LooseVersion
 import easybuild.tools.systemtools as systemtools
 from easybuild.framework.easyconfig.easyconfig import process_easyconfig, robot_find_easyconfig
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.config import build_option
 from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
 from easybuild.tools.toolchain.compiler import Compiler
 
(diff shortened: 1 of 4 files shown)


bartoldeman:
Retry Travis. Something fails with v2 easyconfigs only with Python 3.5. I only see it sometimes locally.

boegel:
@bartoldeman Regardless of the test that sometimes fails (which may not be related to this PR, but we shouldn't be ignoring it), I'm not a huge fan of this approach.
I can live with relying on `moduleclass` since we already do this in module naming schemes, but it does imply that you need to i) find the easyconfig for the `iccifort` toolchain being used (which is otherwise not required except when using HMNS), ii) parse the easyconfig file to figure out the `moduleclass` (which is fairly expensive).

In addition, changing the `COMPILER_MODULE_NAME` *constant* dynamically is also something we should avoid, since it's also being used to interpret the `--optarch` value (without additional changes, specifying custom flags when `icc` is being used as compiler won't work anymore due to your changes).

I appreciate the effort you've put into this (I know how complex the toolchain implementation is), but I think we need to come up with something better... Here's my stab at it: https://github.com/easybuilders/easybuild-framework/pull/2995

bartoldeman:
ok, yes your way of defining `$EBROOTICC` and `$EBROOTIFORT` side-steps the issue quite nicely.

I liked having a proof of concept even though I wasn't too happy myself either about redefining `COMPILER_MODULE_NAME`. I'll close this now.

PR #2996, opened 2019-09-09
HMNS: Define mapping for iccifort->intel for iccifort compiler-only.

bartoldeman:


bartoldeman:
test to come...

bartoldeman:
I added the bug fix label since the HMNS was broken for using iccifort compiler-only as a toolchain (it complained about the 'icc' component missing).

boegel:
Thanks a lot for tackling this @bartoldeman!

PR #2999, opened 2019-09-13
Make installation independent of module naming scheme by default

ocaisa:


ocaisa:
Setting this as a default will open the door to us allowing people to "replay" module creation using different MNS's.

ocaisa:
It's now 99% your PR, does that mean I can approve it?

boegel:
@ocaisa I don't think you can approve your own PR.

Fair game is to get a 3rd party to review this...

Issue #1376, opened 2015-09-01
(build)dependencies only get loaded after extracting sources


boegel:
This is a problem when unpacking `.iso` files using `7z` (`p7zip`), which is generally not available.

Including `p7zip` as a build dependency doesn't help, as the module only gets loaded *after* trying to unpack the sources...

So, the `prepare_step` should be moved up? Or build dependencies should be loaded earlier in a separate step?

wpoely86:
I would not move the `prepare_steps`. Seems like a huge can of worms.
We add a `prepare_fetch` in which we load all the dependencies? So we can use them for both fetching and extracting?

boegel:
@wpoely86: I'm not sure if that's any less of a can of worms than moving up `prepare_step`...

Why would moving `prepare_step` cause problems?

I don't see a real issue in `EasyBlock.prepare_step` or the easyblocks that customise `prepare_step`.

The only thing that needs to be handled is making sure that `self.guess_start_dir()` is done after extracting the sources, so it should move to `extract_step`.

wpoely86:
I don't like it. Other people can depend on this order. We have no way to see the repercussions of this. It should only be done in a major version update.

Why would loading the deps before fetch be any trouble?

boegel:
Well, it wouldn't (it's exactly what's needed), but that would be basically equivalent to moving up `prepare_step` imho.

The only way that people could rely on `prepare_step` running after `fetch_step` is assuming that the sources are already available?

Anyway, if you feel strongly about it, I'm OK with loading the dependencies (i.e., calling `self.toolchain.prepare())` before fetching sources without moving `prepare_step`.

But, I wouldn't add a separate full step for it.

How about defining a `set_up_build_environment` method (which just calls `toolchain.prepare()`), and adding it to the list of `ready_substeps` in `get_steps`?

boegel:
I started looking into actually moving `prepare_step` before `extract_step`, and quickly had to conclude that @wpoely86 was right: it's too painful; see #3004 .

So, how about this as an alternative: we add support for `load_build_deps_before_extract_step = True` (disabled by default), which triggers EasyBuild to do exactly that: load (only) the build dependencies (so *not* including the toolchain) early on, before unpacking the sources.

Since this can be done in a backward-compatible way (by having it disabled by default), there is no need to do this in a major EasyBuild release, and it doesn't impact installations where `load_build_deps_before_extract_step` is unused (it'll be opt-in, so only used where actually needed).

[...]

PR #3004, opened 2019-09-14
move prepare_step before extract_step

boegel:
cfr. #1376

I don't actually intend to get this merged, I just want to get this on record...

I started looking into this, and it quickly became clear that @wpoely86's view was right: moving up the `prepare_step` such that dependencies required for unpacking sources is quite intrusive to do.

The `guess_start_dir()` call had to be moved from `prepare_step` to the end of `extract_step`, since it obviously requires that the sources are unpacked. That's a fairly minor issue, more like a logical consequence (you could even argue that `guess_start_dir()` is out of place in `prepare_step`, I guess).

As anticipated in #1376, this impacts easyblocks where the `prepare_step` is customized and it is assumed that the sources are already unpacked when it is being called. At first sight, this affects at least two software-specific easyblocks in the central repository: `DL_POLY_Classic` and `OpenCV`.
Although it's fairly easy to fix them to take into account that `prepare_step` is run before `extract_step`, it highlights the impact that the change has (we should take into account that there are other (customized) easyblocks outside of the central repository that may be affected).

This change also breaks a test (`test_guess_start_dir`), which may be easy to fix (it looks like a problem with assumptions made by the test itself, rather than an actual problem being caught), but it's still a sign of trouble...

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2887,9 +2887,9 @@ class EasyBlock(object):
         steps_part1 = [
             (FETCH_STEP, 'fetching files', [lambda x: x.fetch_step], False),
             ready_step_spec(True),
+            prepare_step_spec,
             source_step_spec(True),
             patch_step_spec,
-            prepare_step_spec,
             configure_step_spec,
             build_step_spec,
             test_step_spec,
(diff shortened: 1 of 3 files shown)


boegel:
Closing this, since I don't actually intend to get this merged...

PR #3006, opened 2019-09-15
print warning when 'eb' command is not found in $PATH and for empty robot search path

boegel:
Using EasyBuild with an empty robot search path is likely to end in tears, so we should print a warning for it.

The location of the `eb` command is used as a fallback to determine the default robot search path, and having a setup where EasyBuild works (when run with `python -m easybuild.main`) without having the `eb` command available is weird, so deserves a warning as well.


akesandgren:
Going in, thanks @boegel!

PR #3007, opened 2019-09-16
change order of 'python*' commands considered by 'eb' command: python, python3, python2

boegel:
* `python` is considered first since that's assumed to be the preferred Python version
* `python3` is considered before `python2` since Python 2 is EOL on Dec 31st 2019

### DIFF
--- a/eb
+++ b/eb
@@ -43,7 +43,7 @@ function verbose() {
 }
 
 PYTHON=
-for python_cmd in ${EB_PYTHON} 'python2' 'python3' 'python'; do
+for python_cmd in ${EB_PYTHON} 'python' 'python3' 'python2'; do
 
     verbose "Considering '$python_cmd'..."
 


wpoely86:
What's the way this is handle in Debian/Ubuntu? With RHEL 8 you have to use either python2 or python3. I guess Debian handles this with `update-alternatives` which changes to what the `python` commands points to?

boegel:
Maybe @akesandgren or @vanzod can pitch in on @wpoely86's question?

In any case, you can always use `$EB_PYTHON` to specify which Python command `eb` should use...

akesandgren:
On Ubuntu we have:
ls -l /usr/bin/python
lrwxrwxrwx 1 root root 9 Nov 24  2017 /usr/bin/python -> python2.7

boegel:
@bartoldeman I'll let you do the honors on this one, as MotW...

Issue #2952, opened 2019-07-31
Locking mechanism to avoid competing builds to run


mboisson:
At the very beginning of a build, EasyBuild should create the installation directory and touch a file lock in it, and delete it at the end. This is to avoid having two builds that are happening at the same time for the same installed path.

boegel:
Currently the installation directory is only created relatively late in the installation process (as a part of the install step).
But we could touch a temporary file somewhere else in the install tree, or name it `.<installdir>.lock`?

mboisson:
In our case, it could be, since we are using a single build node, but as soon as you start using multiple nodes to build, you want it to be on a shared filesystem, and you cannot assume the build directory to be on the shared filesystem. Where would you put this file ?

damianam:
If you are using multiple nodes to build software and installing in non-shared directories (let's say, via RPMs), then probably the best would be to create the lock in the directory where the easyconfigs are coming from. Of course if that doesn't come from a shared directory you are pretty much on the same point.

In general, I think it makes sense to do `.<installdir>.lock`. Optionally one could make it configurable with `--lockpath` or something like that, but I think that'd be a corner case that I am not sure would affect anyone.

akesandgren:
Yes please.... and if there is a lock one need to be able to define the behavior, wait or die.
Mostly when I push build jobs to the cluster I sometime accidentally run eb --job multiple times with targets that share no-yet-built dependencies. I do not want it to fail in that case but wait until the lock is gone and then detect that it is already built and exit gracefully with a SUCCESS state.

mboisson:
Hum... waiting does not make sense to me. It could potentially wait indefinitely... There is no easy way to guarantee that the lock is deleted in case of failure.

akesandgren:
Since a normal batch job has a timelimit, so that's not a problem. And if interactive you'll probably notice sooner or later. Especially if it prints a warning, and the default should be to not wait.

mboisson:
Mmm, ok. Presumably, if we are speaking about sleeping, we need to define a rate at which the presence of a file would be checked. Should this be configurable, or a constant (say once per minute) ?

mboisson:
I just opened a PR for this.  Now, I know you will ask for tests for this, but I have no damn idea how to write a test for something that requires two parallel builds of the same thing at the same time to test the feature.

mboisson:
This has been closed by the merged PR.

PR #3009, opened 2019-09-16
adding locking to prevent two parallel builds of the same installation directory

mboisson:
... with an option --wait-on-lock which will cause EB to wait until the lock file is removed, checking every minute for the presence of the lock file.

This implements a feature for https://github.com/easybuilders/easybuild-framework/issues/2952


mboisson:
Mmmm, so I understand that this new code caused tests to fail, but I have trouble figuring out why... Is it because of additional output in the build output ?

mboisson:
Seeking guidance here @boegel and @akesandgren. I believe we want to add the mention of the lock file in the interactive output of EB, but it seems to make all sorts of tests fail because it is unexpected output. How do I fix that ?

mboisson:
I think I found it. I noticed I missed a parameter for print_msg, which (presumably) makes the output silent while in test mode. Submitted a new commit and testing with travis now.

easybuild/framework/easyblock.py
```diff
@@ -2943,20 +2943,42 @@ def run_all_steps(self, run_test_cases):
...
-                    else:
-                        print_msg("%s..." % descr, log=self.log, silent=self.silent)
-                    self.current_step = step_name
-                    self.run_step(step_name, step_methods)
 
-        except StopException:
-            pass
+        lockpath = build_option('lockpath') or os.path.join(install_path('software'), '.locks')
+        if not os.path.exists(lockpath):
+            mkdir(lockpath)
```
boegel: @mboisson Please change this to `mkdir(lockpath, parents=True)` (equivalent with `mkdir -p`)

mboisson: done

mboisson:
Looks like all checks have passed. Ready to merge @boegel  ?

easybuild/framework/easyblock.py
```diff
@@ -3038,20 +3038,42 @@ def run_all_steps(self, run_test_cases):
...
-                    self.current_step = step_name
-                    self.run_step(step_name, step_methods)
 
-        except StopException:
-            pass
+        lockpath = build_option('lockpath') or os.path.join(install_path('software'), '.locks')
+        if not os.path.exists(lockpath):
+            mkdir(lockpath, parents=True)
+        lockfile_name = os.path.join(lockpath, ".%s.lock" % self.installdir.replace('/', '_'))
+        if os.path.exists(lockfile_name):
```
akesandgren: You do know that the only thing that works reliably as a lock is a directory. A file doesn't really work as a lock.
Even if you do check for existing lockfile and then create if it isn't there, you can't be sure it was you who created it.

So a file system lock entity should be a directory, not a file. mkdir is atomic, it either succeeeds becuse it didn't exist or it fails because someone else was fast enough to make it just between your exist-check and your mkdir.

easybuild/framework/easyblock.py
```diff
@@ -3038,20 +3038,42 @@ def run_all_steps(self, run_test_cases):
...
+                        else:
+                            print_msg("%s..." % descr, log=self.log, silent=self.silent)
+                        self.current_step = step_name
+                        self.run_step(step_name, step_methods)
+
+            except StopException:
+                pass
+            finally:
+                print_msg("Removing lock file %s" % lockfile_name, silent=self.silent)
+                os.remove(lockfile_name)
```
akesandgren: If the lock is kept as the file then lockpath should also be removed

PR #3019, opened 2019-09-20
fix broken test_download_repo due to archiving of easyconfigs

boegel:
I'm deliberately opening this to the `master` branch, since we don't want a big red cross in Travis for a silly issue like this...


akesandgren:
Going in, thanks @boegel!

PR #3025, opened 2019-09-21
avoid that --inject-checksums introduces list of patches for extensions as a single long line (fixes #2734)

### LINKED ISSUE #2734: inject-checksum creates "too long line"
On a python extension with multiple patches over multiple lines (due to being too long for a single line), inject-checksum merges those patches into a single line, causing travis to complain with "too long line"

See for instance, TensorFlow-1.12.0-fosscuda-2018b-Python-2.7.15.eb (easyconfig PR #7246)

boegel:
problem reported in #2734 reproduced in tests, and then fixed

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3475,7 +3475,7 @@ def inject_checksums(ecs, checksum_type):
                             strval = quote_str(val, prefer_single_quotes=True)
                             line = "%s'%s': %s," % (INDENT_4SPACES * 2, key, strval)
                             # fix long lines for list-type values (e.g. patches)
-                            if len(line) > MAX_LINE_LENGTH and isinstance(val, list):
+                            if len(val) > 1 and isinstance(val, list):
                                 exts_list_lines.append("%s'%s': [" % (INDENT_4SPACES * 2, key))
                                 exts_list_lines.extend(make_list_lines(val, indent_level=3))
                                 exts_list_lines.append(INDENT_4SPACES * 2 + '],',)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3332,7 +3332,13 @@ def inject_checksums(ecs, checksum_type):
     :param ecs: list of EasyConfig instances to inject checksums into corresponding files
     :param checksum_type: type of checksum to use
     """
+    def make_list_lines(values, indent_level):
+        """Make lines for list of values."""
+        line_indent = INDENT_4SPACES * indent_level
+        return [line_indent + "'%s'," % x for x in values]
+
     def make_checksum_lines(checksums, indent_level):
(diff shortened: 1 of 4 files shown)


easybuild/framework/easyblock.py
```diff
@@ -3466,8 +3472,15 @@ def make_checksum_lines(checksums, indent_level):
 
                     for key, val in sorted(ext_options.items()):
                         if key != 'checksums' and val != exts_default_options.get(key):
-                            val = quote_str(val, prefer_single_quotes=True)
-                            exts_list_lines.append("%s'%s': %s," % (INDENT_4SPACES * 2, key, val))
+                            strval = quote_str(val, prefer_single_quotes=True)
+                            line = "%s'%s': %s," % (INDENT_4SPACES * 2, key, strval)
```
damianam: When are we dropping python 2 support? This would look so much better with f-strings :-P

easybuild/framework/easyblock.py
```diff
@@ -3466,8 +3472,15 @@ def make_checksum_lines(checksums, indent_level):
...
+                            # fix long lines for list-type values (e.g. patches)
+                            if len(line) > MAX_LINE_LENGTH and isinstance(val, list):
```
damianam: What about this?

`if len(val) > 1 and isinstance(val, list):`

easybuild/framework/easyblock.py
boegel: It's not *that* bad... It's not pretty, sure, but does it have to be?

We use `%s` string templating pretty consistently across the codebase, which is more important imho...

When we drop Python 2 support (perhaps with EasyBuild 5, time will tell), I'll count on you to port all of that to f-strings so we can keep things consistent. ;)

Would this be better?

```python
line = INDENT_4SPACES * 2 + "'" + key + "': " + strval + ","

easybuild/framework/easyblock.py
boegel: makes sense, fixed in b2ee70abf

easybuild/framework/easyblock.py
damianam: That's much more readable IMO, but I am ok keeping the other approach. My comment was a general comment, nothing that have to be necessarily changed now.

damianam:
@akesandgren you were also requested for review. I can merge it right now, but since @boegel is counting on you I rather wait ;-)

boegel:
To be clear, I requested @akesandgren's review since the reported the bug being fixed here, not because we'd block the PR... ;)

easybuild/framework/easyblock.py
akesandgren: Nah, the "%s'%s': %s," is much easier to read. I'm a C-programmer by heart....

akesandgren:
Going in, thanks @boegel!

PR #3028, opened 2019-09-23
make bootstrap script compatible with Python 3 (WIP)

boegel:
TODO:

* [ ] only *check* for `setuptools` in stage 0 when an older EasyBuild version is being installed (via `$EASYBUILD_BOOTSTRAP_FORCE_VERSION`), don't try installing it
* [ ] stop using `easy_install` (provided by `setuptools` in stage 1
* [ ] figure out EasyBuild version when installing from provided source tarballs

### DIFF
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -49,7 +49,6 @@ import site
 import sys
 import tempfile
 import traceback
-import urllib2
 from distutils.version import LooseVersion
 from hashlib import md5
 
@@ -63,6 +62,15 @@ try:
     HAVE_ARGPARSE = True
 except ImportError:
     import optparse
+
+try:
+    # Python 3
+    from io import StringIO
+    from urllib.request import HTTPError, URLError, urlopen
+except ImportError:
+    # Python 2
+    from cStringIO import StringIO
+    from urllib2 import HTTPError, URLError, urlopen
 
 PYPI_SOURCE_URL = 'https://pypi.python.org/packages/source'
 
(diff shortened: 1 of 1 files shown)


arielzn:
@boegel as i was dealing briefly with this today, just thinking out loud here.

Instead of having easybuild bootstrapping itself. 
To reproduce the same resylt as the bootstrap.py install,  wouldn't it make sense to have instead a bash script leveraging the normal python install, something as:

```
EASYBUILD_PREFIX=$1
EB_VERSION=4.3.4

wget [url]/easybuild-framework-$EB_VERSION.tar.gz
wget [url]/easybuild-easyblocks-$EB_VERSION.tar.gz
wget [url]/easybuild-easyconfigs-$EB_VERSION.tar.gz

extract tarballs

mkdir -p $EASYBUILD_PREFIX/software/EasyBuild/$EB_VERSION

do "cd each dir"
    python{2,3} setup.py install --prefix=$EASYBUILD_PREFIX/software/EasyBuild/$EB_VERSION
done

create corresponding $EASYBUILD_PREFIX/modules/EasyBuild/$EB_VERSION.lua
```

is just a very schematic thing to transmit the idea. 

there's more stuff to add for sure, as checking if an install is there already and so on, also something to check if there's lmod or tcl that we have on the system. 

dunno if adding these latter checks will make things more cumbersome than just finishing the bootstrap.py work.

or maybe a script as above can be used to replace the stage0? so pointed to a temp folder, and do the real install from there?

boegel:
Bootstrap script was deprecated in #3742, not planning to spend more time on it...

PR #3029, opened 2019-09-23
enhance regex in fix_shebang method to fix more Python/Perl shebangs + avoid patching binary files

boegel:
fix for two problems that surfaced with @smoors' PR https://github.com/easybuilders/easybuild-easyconfigs/pull/8968

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2168,7 +2168,7 @@ class EasyBlock(object):
                     paths = glob.glob(os.path.join(self.installdir, glob_pattern))
                     self.log.info("Fixing '%s' shebang to '%s' for files that match '%s': %s",
                                   lang, shebang, glob_pattern, paths)
-                    regex = r'^#!.*/%s[0-9.]*.*$' % lang
+                    regex = r'^#![ ]*.*[/ ]%s[0-9.]*.*$' % lang
                     for path in paths:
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2158,6 +2158,7 @@ def package_step(self):
     def fix_shebang(self):
         """Fix shebang lines for specified files."""
         for lang in ['perl', 'python']:
+            shebang_regex = re.compile(r'^#![ ]*.*[/ ]%s[0-9.]*.*' % lang)
```
smoors: should shebangs like `#!/usr/bin/env python3` and `#!/usr/bin/env python3.6` also be fixed?

boegel: They are, see enhanced tests. The `[0-9.]*` part allows that the `python` is followed by `3` or `3.6`

smoors: since there is a dot in the `[0-9.]*` I think this can be simplified: `r'^#![ ]*.*[/ ]%s.*' % lang`

boegel: Yeah, good point, there's no point in using `[0-9.]*.*` indeed, might as well just use `.*`.

boegel: fixed in a1cea138c

smoors:
fix confirmed for https://github.com/easybuilders/easybuild-easyconfigs/pull/8968

PR #3034, opened 2019-09-26
fix traceback with --inject-checksums for extensions that set a boolean easyconfig parameter

smoors:
fix for EB crashing during checksum generation in some cases.
this can happen when the value of a dictionary item is not iterable, for example: `'modulename': False`
error:
```
  File "/work/smoors/hpc/easybuild-dev/easybuild-framework/easybuild/framework/easyblock.py", line 3488, in inject_checksums
    if len(val) > 1 and isinstance(val, list):
TypeError: object of type 'bool' has no len()

```

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3485,7 +3485,7 @@ def inject_checksums(ecs, checksum_type):
                             strval = quote_str(val, prefer_single_quotes=True)
                             line = "%s'%s': %s," % (INDENT_4SPACES * 2, key, strval)
                             # fix long lines for list-type values (e.g. patches)
-                            if len(val) > 1 and isinstance(val, list):
+                            if isinstance(val, list) and len(val) > 1:
                                 exts_list_lines.append("%s'%s': [" % (INDENT_4SPACES * 2, key))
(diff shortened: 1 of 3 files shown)


boegel:
Just ran into this as well, thanks for the fix @smoors!

The problem was introduced in #3025.

Your fix is correct, but we should get this covered by the tests too, to avoid re-introducing the issue... I'll look into that right now.

boegel:
Slightly modifying the test easyconfig used in `test_inject_checksums` is sufficient to trigger the bug (without the fix, of course):

```
ERROR: test_inject_checksums (__main__.CommandLineOptionsTest)
Test for --inject-checksums
  File "/Volumes/work/easybuild-framework/test/framework/options.py", line 3895, in test_inject_checksums
    stdout, stderr = self._run_mock_eb(args, raise_error=True, strip=True)
  File "/Volumes/work/easybuild-framework/test/framework/options.py", line 2813, in _run_mock_eb
    self.eb_main(args, **kwargs)
  File "test/framework/utilities.py", line 297, in eb_main
    raise myerr
TypeError: object of type 'bool' has no len()

```

The problem only manifests itself for easyconfigs that include an extension for which a boolean value is set...

Issue #2929, opened 2019-06-13
FR: Template value for number of available threads


orbsmiv:
Please can a template value be provided to give the number of available threads for e.g. the `make` command? Currently this value is automatically passed to the first make target invoked by _buildopts_ but if one were to e.g.

`buildopts = "foo && make bar"`
... then the following example command would be issued: `make -j 8 foo && make bar`

Instead, the following would be convenient:

`buildopts = "foo && make -j %(parallel)s bar"`
... which would render: `make -j 8 foo && make -j 8 bar`

boegel:
Support for a `%(parallel)s` template value was added in #3036, will be included in next EasyBuild release (v4.0.1)

PR #3036, opened 2019-09-27
add 'parallel' to list of config templates

migueldiascosta:
answers #2929?


boegel:
@migueldiascosta Wait, does `%(parallel)s` actually ever get resolved when used in an easyconfig file?

migueldiascosta:
@boegel with this change, yes... (just tried it again)

boegel:
@migueldiascosta OK, but then we should get that properly covered in the tests too...

The non-`None` value for `parallel` is only available after a certain step (`ready`), see `set_parallel` in `easyblock.py`.

Done in https://github.com/migueldiascosta/easybuild-framework/pull/9

boegel:
Thanks @migueldiascosta!

PR #3039, opened 2019-09-30
add GitHub Actions workflow to run easybuild-framework test suite

boegel:
This enables running the easybuild-framework test suite in the native CI that is now supported by GitHub through GitHub actions.

For now, I would like to have the tests running in both Travis and GitHub (since GitHub Actions is still in beta)
In the short term, we can change the setup to only keep testing with Python 2.6 in Travis (since it's a bit painful to do that in GitHub CI).
In the long term, switching to GitHub entirely is probably the best way forward...

This should significantly help with getting CI results more quickly, since:
* testing 21 configuration in Travis CI takes about ~2h
* testing 30 configuration in GitHub Actions takes about 40min (mostly due to more configurations being run in parallel)

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -1,0 +1,161 @@
+# documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
+name: EasyBuild framework unit tests
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python: [2.7, 3.5, 3.6, 3.7]
+        modules_tool: [Lmod-6.6.3, Lmod-7.8.22, Lmod-8.1.14, modules-tcl-1.147, modules-3.2.10, modules-4.1.4]
+        module_syntax: [Lua, Tcl]
+        # exclude some configuration for non-Lmod modules tool:
+        # - don't test with Lua module syntax (only supported in Lmod)
+        # - don't test with Python 3.5 and 3.7 (2.7 and 3.7), to limit test configurations
+        exclude:
+          - modules_tool: modules-tcl-1.147
+            module_syntax: Lua
+          - modules_tool: modules-3.2.10
+            module_syntax: Lua
+          - modules_tool: modules-4.1.4
+            module_syntax: Lua
+          - modules_tool: modules-tcl-1.147
+            python: 3.5
+          - modules_tool: modules-tcl-1.147
+            python: 3.7
+          - modules_tool: modules-3.2.10
+            python: 3.5
+          - modules_tool: modules-3.2.10
+            python: 3.7
+          - modules_tool: modules-4.1.4
+            python: 3.5
+          - modules_tool: modules-4.1.4
+            python: 3.7
+      fail-fast: false
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{matrix.python}}
+        architecture: x64
+
+    - name: install OS & Python packages
+      run: |
+        # for modules tool
+        sudo apt-get install lua5.1 liblua5.1-0-dev lua-filesystem lua-posix lua-filesystem tcl tcl-dev
+        sudo ln -s /usr/lib/x86_64-linux-gnu/lua/5.1/posix_c.so /usr/lib/x86_64-linux-gnu/lua/5.1/posix.so
+        # for GitPython, python-hglib
+        sudo apt-get install git mercurial
+        # dep for GC3Pie
+        sudo apt-get install time
+        # Python packages
+        pip --version
+        pip install --upgrade pip
+        pip --version
+        pip install -r requirements.txt
+        # git config is required to make actual git commits (cfr. tests for GitRepository)
+        git config --global user.name "Travis CI"
+        git config --global user.email "travis@travis-ci.org"
+        git config --get-regexp 'user.*'
+
+    - name: install GitHub token (if available)
+      env:
+        # see https://github.com/<username>/easybuild-framework/settings/secrets
+        GITHUB_TOKEN: ${{secrets.TEST_GITHUB_TOKEN}}
+      run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,161 @@
...
+  build:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python: [2.7, 3.5, 3.6, 3.7]
+        modules_tool: [Lmod-6.6.3, Lmod-7.8.22, Lmod-8.1.14, modules-tcl-1.147, modules-3.2.10, modules-4.1.4]
+        module_syntax: [Lua, Tcl]
+        # exclude some configuration for non-Lmod modules tool:
+        # - don't test with Lua module syntax (only supported in Lmod)
+        # - don't test with Python 3.5 and 3.7 (2.7 and 3.7), to limit test configurations
```
akesandgren: This comment is confusing, the () specifically.

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,161 @@
...
+            python: 3.7
+      fail-fast: false
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{matrix.python}}
+        architecture: x64
```
akesandgren: x64? is that something predefined in github? The architecture is really called amd64 since AMD was first to create it.
It is sometimes called x64 but that not something one should use unless necessary.

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,161 @@
...
+
+    - name: install OS & Python packages
+      run: |
+        # for modules tool
+        sudo apt-get install lua5.1 liblua5.1-0-dev lua-filesystem lua-posix lua-filesystem tcl tcl-dev
```
akesandgren: Should probably use lua5.2 instead but not important.

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,161 @@
...
+          # changes in environment are not passed to other steps, so need to create files...
+          echo $MOD_INIT > mod_init
+          echo $PATH > path
+          if [ ! -z $MODULESHOME ]; then echo $MODULESHOME > moduleshome; fi
+
+    - name: check sources
+      run: |
+          # make sure there are no (top-level) "import setuptools" or "import pkg_resources" statements,
+          # since EasyBuild should not have a runtime requirement on setuptools
+          SETUPTOOLS_IMPORTS=$(egrep -RI '^(from|import)[ ]*pkg_resources|^(from|import)[ ]*setuptools' * || true)
```
akesandgren: Are your sure you don't want to handle spaces before "from|import"?
And what files is * matching here?

.github/workflows/unit_tests.yml
bartoldeman: Originally called x86-64 (and still called that way by many places): https://web.archive.org/web/20120308030806/http://www.amd.com/us/press-releases/Pages/Press_Release_751.aspx
but in the Windows world it's called x64 and guess who bought github :)

akesandgren: Ah, yes.... forgot that AMD themselves originally called it that. Everybody else quickly adopted amd64 as the name :-)

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,161 @@
...
+          EB_BOOTSTRAP_EXPECTED="20190922.01 778762479d77df641c247aba33c13203c8c5b8275e47f59464bab93917aaaf22"
+          test "$EB_BOOTSTRAP_FOUND" = "$EB_BOOTSTRAP_EXPECTED" || (echo "Version check on bootstrap script failed $EB_BOOTSTRAP_FOUND" && exit 1)
+
+          # test bootstrap script (only compatible with Python 2 for now)
+          if [[ ${{matrix.python}} =~ '2.' ]]; then
+              export PREFIX=/tmp/$USER/$GITHUB_SHA/eb_bootstrap
+              python easybuild/scripts/bootstrap_eb.py $PREFIX
+              # unset $PYTHONPATH to avoid mixing two EasyBuild 'installations' when testing bootstrapped EasyBuild module
+              unset PYTHONPATH
+              # simply sanity check on bootstrapped EasyBuild module (skip when testing with Python 3, for now)
```
akesandgren: Do you mean "simple sanity check on" or "simply sanity check the ..."

.github/workflows/unit_tests.yml
boegel: Well, I can update it to `lua5.2`, but we're not testing Lmod here... ;)

Why would you bump it to Lua 5.2?

.github/workflows/unit_tests.yml
boegel: `simple`, will fix

.github/workflows/unit_tests.yml
akesandgren: Mostly because 5.1 is old, and 5.3 has been out for quite a while... and Lmod supports 5.2
So as long as it's running on Ubuntu18.04 5.2 is a more "natural" choice.

But as I said, not important, just a suggestion.

.github/workflows/unit_tests.yml
boegel: I can probably drop this fairly safely, but since it doesn't hurt I prefer leaving it in.

Documentation doesn't clearly mention the possible values.

.github/workflows/unit_tests.yml
boegel: Well, ok, doesn't hurt to bump to 5.2

.github/workflows/unit_tests.yml
boegel: fixed, thx

.github/workflows/unit_tests.yml
akesandgren: okidoki

akesandgren:
Going in, thanks @boegel!

PR #3044, opened 2019-10-03
Error out if some GC3Pie job failed.

riccardomurri:
As requested by Will Furnass on the EB chat:
https://easybuild.slack.com/archives/C34UA1HT7/p1570098913442400

### DIFF
--- a/easybuild/tools/job/gc3pie.py
+++ b/easybuild/tools/job/gc3pie.py
@@ -255,23 +255,26 @@ class GC3Pie(JobBackend):
             self._engine.progress()
 
             # report progress
-            self._print_status_report()
+            stats = self._engine.counts(only=Application)
+            self._print_status_report(stats)
 
             # Wait a few seconds...
             time.sleep(self.poll_interval)
 
         # final status report
         print_msg("Done processing jobs", log=self.log, silent=build_option('silent'))
-        self._print_status_report()
+        self._print_status_report(stats)
 
         # fail if at least one job has failed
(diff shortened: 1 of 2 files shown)


boegel:
@riccardomurri I'd like to enhance the existing test we have for using GC3Pie as a backend, to verify that a failing build indeed results in a raised error (and that it stays that way).

I started looking into this myself, but noticed that `test_build_easyconfigs_in_parallel_gc3pie` basically hangs on my system (macOS) for some reason (it works fine in Travis CI & GitHub Actions though), which makes that very annoying to work on. :)

I'd like to pick your brain on that... Can you think of something that could be causing that, from the top of your head (some different behavior between macOS & Linux would be my first guess)?

riccardomurri:
Added report of names of jobs that failed, as further requested by Will on chat.

riccardomurri:
> I started looking into this myself, but noticed that test_build_easyconfigs_in_parallel_gc3pie basically hangs on my system (macOS)

@boegel Maybe this should be posted as a bug on GC3Pie?  If you Ctrl+C that script while it hangs, what traceback do you get?

riccardomurri:
> Can you think of something that could be causing that, from the top of your head (some different behavior between macOS & Linux would be my first guess)?

If you have GC3Pie configured to use the `shellcmd` ("localhost") backend, then it may hang while running some local discovery command... Maybe Apple has changed something in the latest releases, or some command has a hidden resource deadlock that causes it to break when run in parallel...

Issue #3045, opened 2019-10-03
EB 4.0.0 breaks when there is another "eb" command before itself in the PATH despite being called with a full path


akesandgren:
Setup:
```
grep robot-paths /hpc2n/eb/easybuild.cfg 
robot-paths=/hpc2n/eb/custom/easyconfigs:%(DEFAULT_ROBOT_PATHS)s
```

/hpc2n/eb -> /cvmfs/ebsw.hpc2n.umu.se/amd64_ubuntu1604_skx/

(Last part arch dependent)

EB 3.9.4:
```
eb --show-config
robot-paths                 (F) = /hpc2n/eb/custom/easyconfigs, /hpc2n/eb/software/Core/EasyBuild/3.9.4/lib/python2.7/site-packages/easybuild_easyconfigs-3.9.4-py2.7.egg/easybuild/easyconfigs
```

EB 4.0.0:
```
eb --show-config
robot-paths                    (F) = /hpc2n/eb/custom/easyconfigs, /cvmfs/ebsw.hpc2n.umu.se/amd64_ubuntu1604_common/software/Core/EasyBuild/4.0.0/easybuild/easyconfigs
```

(Note that it also resolves the DEFAULT_ROBOT_PATHS which it shouldn't do)

```
cat /scratch/q/eb
#!/bin/bash
echo IIeeekkk, Im getting called
```

EB 3.9.4:
```
env PATH=/scratch/q:$PATH $EBROOTEASYBUILD/bin/eb --show-config
robot-paths                 (F) = /hpc2n/eb/custom/easyconfigs, /hpc2n/eb/software/Core/EasyBuild/3.9.4/lib/python2.7/site-packages/easybuild_easyconfigs-3.9.4-py2.7.egg/easybuild/easyconfigs
```

Still correct, but

EB 4.0.0:
```
env PATH=/scratch/q:$PATH $EBROOTEASYBUILD/bin/eb --show-config
robot-paths                    (F) = /hpc2n/eb/custom/easyconfigs
```

This breaks our setup completely and I'm now stuck.

boegel:
This is related to no longer relying on `setuptools`, sort of...

The location of the `eb` command is used as a fallback to determine the location of the easyconfig files included with EasyBuild (that's done in `get_paths_for` from `easybuild.framework.easyconfigs.tools`).

[...]

PR #3046, opened 2019-10-03
define $EB_SCRIPT_PATH in 'eb' wrapper script, and consider it before location of 'eb' determined via $PATH in get_paths_for function

boegel:
fix for #3045 (cc @akesandgren)

### DIFF
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -287,8 +287,7 @@ def get_paths_for(subdir=EASYCONFIGS_PKG_SUBDIR, robot_path=None):
             if eb_path != resolved_eb_path:
                 install_prefix = os.path.dirname(os.path.dirname(resolved_eb_path))
                 path_list.append(install_prefix)
-                _log.info("Also considering installation prefix %s (via resolved path to 'eb' script)...",
-                           install_prefix)
+                _log.info("Also considering installation prefix %s (via resolved path to 'eb')...", install_prefix)
 
     # look for desired subdirs
     for path in path_list:
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -255,17 +255,31 @@ def get_paths_for(subdir=EASYCONFIGS_PKG_SUBDIR, robot_path=None):
     path_list.extend(sys.path)
 
     # figure out installation prefix, e.g. distutils install path for easyconfigs
-    eb_path = which('eb')
+
+    # prefer using path specified in $EB_SCRIPT_PATH (if defined), which is set by 'eb' wrapper script
+    eb_path = os.getenv('EB_SCRIPT_PATH')
+    if eb_path is None:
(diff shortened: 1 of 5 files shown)


easybuild/framework/easyconfig/tools.py
```diff
@@ -255,18 +255,32 @@ def get_paths_for(subdir=EASYCONFIGS_PKG_SUBDIR, robot_path=None):
...
 
     # figure out installation prefix, e.g. distutils install path for easyconfigs
-    eb_path = which('eb')
+
+    # prefer using path specified in $EB_SCRIPT_PATH (if defined), which is set by 'eb' wrapper script
+    eb_path = os.getenv('EB_SCRIPT_PATH')
+    if eb_path is None:
+        # try to determine location of 'eb' script via $PATH, as fallback mechanism
+        eb_path = which('eb')
+
```
akesandgren: Why not do this the same way as below? getenv or which?

easybuild/framework/easyconfig/tools.py
```diff
@@ -255,18 +255,32 @@ def get_paths_for(subdir=EASYCONFIGS_PKG_SUBDIR, robot_path=None):
...
-        eb_path = resolve_path(eb_path)
+        # eb_path is location to 'eb' wrapper script, e.g. <install_prefix>/bin/eb
+        # so installation prefix is usually two levels up
         install_prefix = os.path.dirname(os.path.dirname(eb_path))
         path_list.append(install_prefix)
         _log.debug("Also considering installation prefix %s..." % install_prefix)
 
+        # also consider fully resolved location to 'eb' wrapper
+        # see https://github.com/easybuilders/easybuild-framework/pull/2248
+        resolved_eb_path = resolve_path(eb_path)
```
akesandgren: I would argue that the resolved path should only be used if the unresolved path doesn't contain easybuild/subdir.

Sysadmins usually frown upon code that lets users see paths that they have "hidden" with symlinks for no good reason.

Code gets somewhat more complex but would in my opinion produce a cleaner result.

As is it looks now it would for --search show both our /hpc2n/eb/software/Core/EasyBuild/... path and the resolved /cvmfs/ebsw.hpc2n.umu.se/amd64_ubutnu1604_bdw/software/.... path (unless i totally misunderstood this code)

easybuild/framework/easyconfig/tools.py
boegel: Well, I could, doesn't matter.

But this way, we can probably log how we obtained the full path to `eb` (which I'm not doing yet here, but I should)

akesandgren: okidoki

easybuild/framework/easyconfig/tools.py
boegel: Yeah, good point I guess, this should really be a fallback mechanism used "in times of need".

See changes in 1a7c4fd26

akesandgren:
Going in, thanks @boegel!

PR #3047, opened 2019-10-04
add support for specifying different dependency version based on architecture

edmondac:
Allow different versions of software to be used depending on the cpu architecture

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1226,33 +1226,26 @@ class EasyConfig(object):
 
         return multi_deps
 
-    def find_dep_version_match(self, ec):
-        """Identify the correct version for this system from the choices provided. This overwrites ec['version']"""
-        if isinstance(ec['version'], string_type):
-            self.log.debug("Version is already a string ('%s'), OK", ec['version'])
+    def find_dep_version_match(self, dep):
+        """Identify the correct version for this system from the choices provided. This overwrites dep['version']"""
+        if isinstance(dep['version'], string_type):
+            self.log.debug("Version is already a string ('%s'), OK", dep['version'])
             return
-        elif ec['version'] is None:
-            self.log.debug("Version is None, OK", ec)
-        elif isinstance(ec['version'], dict):
+        elif dep['version'] is None:
+            self.log.debug("Version is None, OK", dep)
+        elif isinstance(dep['version'], dict):
             # figure out matches based on dict keys (after splitting on '=')
-            matches = []
-            my_arch = get_cpu_architecture()
-            for key, value in ec['version'].items():
-                if '=' in key:
-                    inner_key, inner_value = key.split('=', 1)
-                    if inner_key == 'arch' and inner_value == my_arch:
-                        matches.append(value)
-
-            # if there's a single match, we can continue with the corresponding string version
-            if len(matches) == 1:
-                self.log.info("Version selected from %s: %s", ec['version'], matches[0])
-                ec['version'] = matches[0]
-            elif matches:
-                raise EasyBuildError("Multiple matches found for version using %s: %s", ec['version'], matches)
-            else:
-                raise EasyBuildError("No matches for version using %s", ec['version'])
+            my_arch_key = 'arch=%s' % get_cpu_architecture()
+            has_arch_keys = [x for x in dep['version'].keys() if x.startswith('arch=')]
+            if has_arch_keys:
+                if my_arch_key in dep['version']:
+                    ver = dep['version'][my_arch_key]
+                    self.log.info("Version selected from %s: %s", dep['version'], ver)
+                    dep['version'] = ver
+                else:
+                    raise EasyBuildError("No matches for version using %s (looking for %s)", dep['version'], my_arch_key)
         else:
-            raise EasyBuildError("Unknown value type for version: %s", ec['version'])
+            raise EasyBuildError("Unknown value type for version: %s", dep['version'])
 
     # private method
     def _parse_dependency(self, dep, hidden=False, build_only=False):
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1331,7 +1331,7 @@ class EasyConfig(object):
         if isinstance(dependency['version'], string_type):
             self.log.debug("Dependency version is already a string ('%s'), OK", dependency['version'])
         elif dependency['version'] is None:
-            self.log.debug("Dependency version is None, OK")
+            self.log.debug("Dependency %s version is None, OK", dependency)
         elif isinstance(dependency['version'], dict):
             # call out to a function that figures out which entries in the dict matches
             selected_dep_version = self.find_dep_version_match(dependency['version'])
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1238,13 +1238,13 @@ class EasyConfig(object):
             # figure out matches based on dict keys (after splitting on '=')
             my_arch_key = 'arch=%s' % get_cpu_architecture()
             arch_keys = [x for x in dep['version'].keys() if x.startswith('arch=')]
-            other_keys = [x for x in dep['version'].keys() if '=' in x and x not in arch_keys]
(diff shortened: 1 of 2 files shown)


edmondac:
The design comes from https://github.com/easybuilders/easybuild-framework/pull/2841#issuecomment-538465120

edmondac:
OK - this is ready for review :-)

https://github.com/easybuilders/easybuild-easyconfigs/pull/9081 has an example easyconfig file

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1226,6 +1226,29 @@ def get_parsed_multi_deps(self):
...
+    # new function/method
+    def find_dep_version_match(self, dep_version_candidates):
+        """Identify the correct version for this system from the choices provided"""
+        # figure out matches based on dict keys (after splitting on '=')
+        matches = []
+        my_arch = get_cpu_architecture()
+        for key, value in dep_version_candidates.items():
+            if '=' in key:
+                inner_key, inner_value = key.split('=', 1)
+                if inner_key == 'arch' and inner_value == my_arch:
```
bartoldeman: This could be done more simply via
```
if key == 'arch='+my_arch:
```
You could also iterate `key in dep_version_candidates:`, and then use `dep_version_candidate[key]` for `value`. That is more efficient in Python 2 mainly (iterator vs copy)

bartoldeman: or even
```
matches = [dep_version_candidates[key] for key in dep_version_candidates if key == 'arch'+my_arch]
```

edmondac: I've added a test - then if that passes on here I'll look at the changes.

edmondac: Simplified the code further, having thought for 1 minute about what a dictionary is...

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1226,6 +1226,27 @@ def get_parsed_multi_deps(self):
...
+        if isinstance(dep['version'], string_type):
+            self.log.debug("Version is already a string ('%s'), OK", dep['version'])
+            return
+        elif dep['version'] is None:
+            self.log.debug("Version is None, OK", dep)
+        elif isinstance(dep['version'], dict):
+            # figure out matches based on dict keys (after splitting on '=')
+            my_arch_key = 'arch=%s' % get_cpu_architecture()
+            has_arch_keys = [x for x in dep['version'].keys() if x.startswith('arch=')]
+            if has_arch_keys:
```
boegel: @edmondac I think this should be made a bit more strict...

Right now, only `arch=` keys are supported, so we should make sure nothing else is used (and if so, an appropriate error should be raised).

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1305,6 +1327,15 @@ def _parse_dependency(self, dep, hidden=False, build_only=False):
         else:
             raise EasyBuildError("Dependency %s of unsupported type: %s", dep, type(dep))
 
+        if isinstance(dependency['version'], string_type):
+            self.log.debug("Dependency version is already a string ('%s'), OK", dependency['version'])
+        elif dependency['version'] is None:
```
boegel: Both these conditions are also handled in `find_dep_version_match`, so just call `self.find_dep_version_match(dependency)` here?

Also, it's probably better to make `find_dep_version_match` return the picked version, rather than updating `dependency` in place, so:

```python
dependency['version'] = self.find_dep_version_match(dependency)

Comment (edmondac) [reply easybuild/framework/easyconfig/easyconfig.py]: done

File: easybuild/framework/easyconfig/easyconfig.py [hunk 4]
```diff
@@ -1226,6 +1226,31 @@ def get_parsed_multi_deps(self):
...
+            self.log.debug("Version is already a string ('%s'), OK", dep['version'])
+            return dep['version']
+        elif dep['version'] is None:
+            self.log.debug("Version is None, OK", dep)
+            return None
+        elif isinstance(dep['version'], dict):
+            # figure out matches based on dict keys (after splitting on '=')
+            my_arch_key = 'arch=%s' % get_cpu_architecture()
+            arch_keys = [x for x in dep['version'].keys() if x.startswith('arch=')]
+            other_keys = [x for x in dep['version'].keys() if '=' in x and x not in arch_keys]
```
boegel: @edmondac I think we should drop the `'=' in x` part.

If I use the following:

```python
dependencies = [('example', {'foo': '1.2.3'}]
```

Then you should still get an error for `Unexpected keys`, even though `foo` doesn't include `=`?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1226,6 +1226,31 @@ def get_parsed_multi_deps(self):
...
+            if other_keys:
+                raise EasyBuildError("Unexpected keys in version: %s. Only 'arch=' keys are supported", other_keys)
+            if arch_keys:
+                if my_arch_key in dep['version']:
+                    ver = dep['version'][my_arch_key]
+                    self.log.info("Version selected from %s: %s", dep['version'], ver)
```
boegel: also mention key in log message?

test/framework/easyconfig.py
```diff
@@ -3136,6 +3136,49 @@ def test_local_vars_detection(self):
...
+
+        ec = EasyConfig(test_ec)
+        self.assertEqual(ec.dependencies()[0]['version'], expected_version)
+
+    def test_unexpected_version_keys_caught(self):
+        """Tests that unexpected keys in a version dictionary are caught"""
+
+        my_arch = st.get_cpu_architecture()
+        expected_version = '1.2.3'
+        dep_str = "[('foo', {'bar=%s': '%s', 'arch=Foo': 'bar'})]" % (my_arch, expected_version)
```
boegel: Test with unknown key not including `=`?

easybuild/framework/easyconfig/easyconfig.py
edmondac: done

easybuild/framework/easyconfig/easyconfig.py
edmondac: done

test/framework/easyconfig.py
edmondac: Done

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1226,6 +1226,31 @@ def get_parsed_multi_deps(self):
 
         return multi_deps
 
+    def find_dep_version_match(self, dep):
+        """Identify the correct version for this system from the choices provided. This returns the version to use."""
+        if isinstance(dep['version'], string_type):
+            self.log.debug("Version is already a string ('%s'), OK", dep['version'])
+            return dep['version']
+        elif dep['version'] is None:
+            self.log.debug("Version is None, OK", dep)
```
boegel: the `, dep` shouldn't be there, unless you want to print the whole `dep` dict:

So, either:

```python
self.log.debug("Version is None, OK")
```

or

```python
self.log.debug("Version is None, OK (dep: %s)", dep)
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1226,6 +1226,31 @@ def get_parsed_multi_deps(self):
 
         return multi_deps
 
+    def find_dep_version_match(self, dep):
+        """Identify the correct version for this system from the choices provided. This returns the version to use."""
+        if isinstance(dep['version'], string_type):
```
boegel: @edmondac I think it makes sense to only pass down `dep['version']` to this method, and then rename `dep` to `dep_version` in the method body...

boegel:
@edmondac Sorry for the back & forth on this, but we're really close now to getting this merged... :)

easybuild/framework/easyconfig/easyconfig.py
edmondac: Agreed

easybuild/framework/easyconfig/easyconfig.py
edmondac: Done

edmondac:
> @edmondac Sorry for the back & forth on this, but we're really close now to getting this merged... :)

No probs!

Issue #3026, opened 2019-09-22
Going from EB 3.9.4 with HMNS to 4.0 can cause ghost installs at rebuild


akesandgren:
When going from fixed-installdir-naming-scheme=False at EB 3.9.4 using a HMNS to fixed-installdir-naming-scheme=True at EB 4.0.0, rebuilding a package will generate a ghost installation under the HMNS software tree.

The removing of the old installdir should take this into consideration, perhaps using the EBROOT* from the module, or picking up the installdir using fixed-installdir-naming-scheme temporarily set to False to check whether it's there or not.

boegel:
Removing an existing installation under `--force` should be done using the `$EBROOT*` value taken from the existing module file that is being replaced.

akesandgren:
Using EBROOT* assumes that the module file exists, that may not be the case.
Consider having tried (and failed) installing Amber (which build in the installdir) with 3.9.4 and changing to 4.0 before trying again. Then there is no module file to look in. So using get_software_root with fixed-install-dir...=False is the only way to find the previous attempt to be able to remove it. And doesn't a build always start with a removal of the old installdir?
 
I suggest to just add a small piece of, if g_s_r() exist with fixed-install-dir=False then remove to be added before the current removal line.

boegel:
@akesandgren I'm currently working on a change where the installation directory is extracted from the module file that is being replaced, and this directory is removed in case it differs from the one that will be used in the current configuration.

Under `--force`, the existing module file is removed early on, but if the installation directory already exists that's only removed when the `install` step starts.

I'm not sure if `--force` should remove an existing installation directory that is there if there's no module file found for it (the module file for that installation directory could be elsewhere, in theory).

boegel:
fixed in #3050, so closing...

PR #3050, opened 2019-10-14
add support for --remove-ghost-install-dirs, and warn about (potential) ghost install dirs by default when --force/--rebuild is used

boegel:
fix for #3026 reported by @akesandgren


akesandgren:
Going in, thanks @boegel!

PR #3060, opened 2019-10-24
fix for unsetting ...-%(pyver)s in job context

cmeesters:
Hi,

When running eb on configs referring to specific python versions, with v4.0.1 in a multideps context the log contains something like:

/var/tmp/slurmd_spool/job13282805/slurm_script: line 4: syntax error near unexpected token `('
/var/tmp/slurmd_spool/job13282805/slurm_script: line 4: `unset TMPDIR && 
...
foss-2019a-Python-%(pyver)s'

For this I single quoted the --testoutput parameter in line 138 of tools/parallelbuild.py .

This is not the canonical way, of course. shlex.quote would be the preferred approach in python3, but is not feasible for python2, right?

Perhaps there are side-effects of better approaches / places for this fix. Just let me know.

Best regards,
Christian

### DIFF
--- a/easybuild/tools/parallelbuild.py
+++ b/easybuild/tools/parallelbuild.py
@@ -135,7 +135,7 @@ def submit_jobs(ordered_ecs, cmd_line_opts, testing=False, prepare_first=True):
     # compose string with command line options, properly quoted and with '%' characters escaped
     opts_str = ' '.join(opts).replace('%', '%%')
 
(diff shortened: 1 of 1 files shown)


boegel:
@cmeesters Which easyconfig file were you using to trigger the bug you're fixing here?

I don't think it's specific to `--job` at all, the issue seems to be that you're using `%(pyver)s` in combination with `multi_deps = {'Python': ...}`?

cmeesters:
@boegel: Yes, indeed. The easyconfig was a multi-dep.

boegel:
@cmeesters I'm not sure this change is a good idea, it may lead to very weird output dirs being used, I'm not sure we want that... Thoughts/

PR #3062, opened 2019-10-26
Init git repo with git.repo.clone() instead of copy_dir()

lexming:
Currently, `init_repo()` does a recursive copy of all files in the existing repo in `git_working_dirs_path` to the temporary git repo in `repo_path`. This can cause issues with `eb --new-pr` at later stages if (like me) you have some untracked and/or unstaged files in the repo. For instance, if the `.eb` file of the new easyconfig is untracked, it will be falsely detected as already existing in the branch.
This patch solves this issue by instead using `git.Repo.clone()` to clone the local repo in `git_working_dirs_path`. This ensures that only tracked files are transferred to the temporary repo.

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -556,6 +556,7 @@ def init_repo(path, repo_name, silent=False):
         raise EasyBuildError("Failed to init git repo at %s: %s", repo_path, err)
 
     _log.debug("temporary git working directory ready at %s", repo_path)
+
     return repo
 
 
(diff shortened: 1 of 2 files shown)


boegel:
@lexming This makes perfect sense, thanks a lot for fixing this!

We should get this covered in the tests too, I looked into this myself, see https://github.com/lexming/easybuild-framework/pull/1

lexming:
@boegel merged your changes for the tests :+1:

Issue #3058, opened 2019-10-23
Error building toolchains with --module-depends-on and maybe also --module-syntax=Tcl


douglasgscofield:
We're a centre new to EasyBuild.  We are trying to install the foss-2018b toolchain but it keeps failing attempting to install FFTW-3.3.8-gompi-2018b with what seems to be a straightforward config mismatch, but with our *very* limited experience we can't see anything obvious.  It doesn't install with EasyBuild 3.8.1 either.

I emailed the EasyBuild mailing list and Bart Oldeman suggested opening an issue.  He wrote it might be a framework bug associated with using `--module-depends-on` perhaps together with `--module-syntax=Tcl`.

> this smells like a framework bug in the parsing of depends_on in Tcl module files to figure out the dependencies.

```
$ eb --module-depends-on --robot --module-syntax=Tcl foss-2018b.eb
== temporary log file in case of crash /scratch/eb-dFaRLw/easybuild-eFncBe.log
== resolving dependencies ...
== processing EasyBuild easyconfig /sw/EasyBuild/software/EasyBuild/4.0.1/easybuild/easyconfigs/f/FFTW/FFTW-3.3.8-gompi-2018b.eb
== building and installing FFTW/3.3.8-gompi-2018b...
== fetching files...
== creating build dir, resetting environment...
== starting iteration #0 ...
== unpacking...
== patching...
== preparing...
== FAILED: Installation ended unsuccessfully (build directory: /sw/EasyBuild/build/FFTW/3.3.8/gompi-2018b): build failed (first 300 chars): List of toolchain dependency modules and toolchain definition do not match (found [] vs expected set(['GCC', 'OpenMPI'])) (took 17 sec)
== Results of the build can be found in the log file(s) /scratch/eb-dFaRLw/easybuild-FFTW-3.3.8-20191023.101730.GDYpv.log
ERROR: Build of /sw/EasyBuild/software/EasyBuild/4.0.1/easybuild/easyconfigs/f/FFTW/FFTW-3.3.8-gompi-2018b.eb failed (err: "build failed (first 300 chars): List of toolchain dependency modules and toolchain definition do not match (found [] vs expected set(['GCC', 'OpenMPI']))")
```

At the suggestion of Alan O'Cais, here is `eb --dry-run gompi-2018b.eb`:
```
$ eb  --dry-run gompi-2018b.eb
== temporary log file in case of crash /scratch/eb-qUHeQG/easybuild-8PK2Hu.log
[...]

PR #3065, opened 2019-10-30
fix regular expression so depends-on statements are recognized correctly in Tcl module files

boegel:
fix for #3058 reported by @douglasgscofield


akesandgren:
Going in, thanks @boegel!

PR #3071, opened 2019-11-05
skip validation when copying EasyConfig object for extension

boegel:
Currently when an extension is initialized, the `EasyConfig` instance of the "parent" is copied into the extension (since stuff like `toolchain` etc. is also relevant for installing the extension).

By default, this results in re-validating the easyconfig file (which includes re-checking the OS dependencies), which is silly.
Even worse, it leads to problems like the ones reported in https://github.com/easybuilders/easybuild-easyconfigs/issues/8661 since commands like `rpm -q` may break when particular modules (like `XZ` are loaded).

cc @zao, @terjekv, @akesandgren

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -530,12 +530,15 @@ class EasyConfig(object):
                 self.mandatory.append(key)
         self.log.debug("Updated list of mandatory easyconfig parameters: %s", self.mandatory)
 
-    def copy(self):
+    def copy(self, validate=None):
         """
         Return a copy of this EasyConfig instance.
         """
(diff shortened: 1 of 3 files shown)


boegel:
@akesandgren Basically, yes.

I'd like to get this covered in the tests somehow too, need to think about how to do that exactly...

terjekv:
This resolves the issues I've seen with OS dependencies on RHEL8. Thanks!

boegel:
Thanks a lot for confirming the fix @terjekv!

I've enhanced one of the tests to try and ensure that we don't accidentally reverse this fix, which makes the PR good to go...

akesandgren:
Going in, thanks @boegel!

PR #3073, opened 2019-11-08
correctly specify that ActiveMNS & co are singleton classes

boegel:
This was overlooked when porting to Python 3.

While the `__metaclass__ = Singleton` that specifies a singleton class in Python 2 does no harm when using Python 3, it's blatantly ignored when using Python 3, so `ActiveMNS` isn't a singleton class at all then...

This is important w.r.t. the time required for running the easyconfig test suite, to avoid that 475k (identical) `ActiveMNS` instances are created...


akesandgren:
Going in, thanks @boegel!

PR #3074, opened 2019-11-08
don't call out to prohibitively expensive getRootLoggerName in getLogger, just use 'root' instead

boegel:
The calls to `getRootLoggerName` are very expensive when `__debug__` is `True`, since then it relies heavily on Python's `inspect` module.

For some reason unclear to me, the easyconfig tests that run in GitHub CI with Python 3.6 and 3.7 trigger the path that corresponds with `__debug__` being `False`, despite running under `python -O` (and also defining `$PYTHONOPTIMIZE` doesn't help).

Regardless, I think it's fine to strip out the call to `getRootLoggerName` and just use a fixed `'root'` name instead.

For normal use of EasyBuild via the `eb` command, this doesn't make a difference (since then `python -O` is used under the covers, and the name of the root logger isn't actually visible anywhere by default unless the `%(name)s` template is used in the log format set via `setLogFormat`).

This minor change will significantly speed up the easyconfig tests though, when `getRootLoggerName` follows the path to `inspect`...

### DIFF
--- a/easybuild/base/fancylogger.py
+++ b/easybuild/base/fancylogger.py
@@ -452,7 +452,9 @@ def getLogger(name=None, fname=False, clsname=False, fancyrecord=None):
     nameparts = []
 
     if not is_fancyroot():
-        nameparts.append(getRootLoggerName())
+        # deliberately not calling getRootLoggerName function to determine actual root logger name,
+        # because it is prohibitively expensive in some texts (even when using 'python -O')
+        nameparts.append('root')
 
     if fancyrecord is None:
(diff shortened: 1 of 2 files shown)


boegel:
Some more details: currently ~65% of the time running the easyconfig test suite in GitHub CI with Python 3.6 and 3.7 is spent in `getRootLoggerName`, so replacing the call with a hardcoded string value is going to result in cutting the test time in half (more even)...

boegel:
close/re-open to wake up the Hound...

akesandgren:
Going in, thanks @boegel!

PR #3075, opened 2019-11-11
support use of glob patterns for paths to files with external modules metadata

boegel:
This is useful to better integrate with the pre-packaged software provided by OpenHPC.

@koomie @adrianreber have plans to include metadata files in the software RPMs, so EasyBuild can pick up on them.

For this to work, EasyBuild needs to support a file *pattern* to be passed to `$EASYBUILD_EXTERNAL_MODULES_METADATA` (rather than a list of absolute paths to files, which would have to be continuously updated).

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1471,14 +1471,25 @@ def parse_external_modules_metadata(cfgs):
     """
     Parse metadata for external modules.
 
-    :param cfgs: list of config files providing metadata for external modules
+    :param cfgs: list of (glob patterns for) paths to config files providing metadata for external modules
     :return: parsed metadata for external modules
     """
+    if cfgs is None:
+        cfgs = []
+
+    # expand glob patterns, and report error for faulty paths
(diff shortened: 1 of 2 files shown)


easybuild/tools/options.py
```diff
@@ -1471,14 +1471,25 @@ def parse_external_modules_metadata(cfgs):
     """
     Parse metadata for external modules.
 
-    :param cfgs: list of config files providing metadata for external modules
+    :param cfgs: list of (glob patterns for) paths to config files providing metadata for external modules
```
migueldiascosta: https://github.com/easybuilders/easybuild-framework/blob/4ac2f6d2add3aed853430f97e87415a7e7b717af/easybuild/tools/options.py#L452 should also be updated?

boegel:
corresponding docs update: https://github.com/easybuilders/easybuild/pull/558

migueldiascosta:
Going in, thanks @boegel!

PR #3077, opened 2019-11-11
deprecate support for using Lmod 6.x

boegel:
The last EasyBuild user survey (conducted in Dec 2018) indicated that about 25% of the EasyBuild community was using Lmod 6 at the time. The most recent release of Lmod 6.x was in November 2016, so it's fair to assume that most people have upgraded to Lmod 7 or 8 by now.

We won't actively break support for Lmod 6 for now, but we need to trim down on the test configurations we're using in Travis CI & GitHub Actions to avoid waiting forever on test results.

With this change included, we will:

* only test with Lmod 6 (& Python 2.7) in a single test configuration in Travis CI (and no more in GitHub Actions)
* produce a deprecation warning when Lmod 6 is used:

```
$ eb example.eb

WARNING: Deprecated functionality, will no longer work in v5.0:
Support for Lmod version < 7.0.0 is deprecated, found version 6.6.3;
see http://easybuild.readthedocs.org/en/latest/Deprecated-functionality.html for more information
```

If desired, this warning can be avoided by configuring EasyBuild to skip the modules tool version check via `--disable-modules-tool-version-check` or by setting `$EASYBUILD_DISABLE_MODULES_TOOL_VERSION_CHECK=1` (which is, for obvious reasons, *not* recommended).

In addition, I've updated some of the modules tool log messages to clearly mention which modules tool is being used (see the `NAME` class variable that was set for each type of modules tool).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,13 @@
 language: python
-python: 2.6
-dist: trusty
+python: 2.7
+dist: xenial
 env:
   matrix:
     # purposely specifying slowest builds first, to gain time overall
-    - LMOD_VERSION=6.5.1
-    - LMOD_VERSION=6.5.1 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
     - LMOD_VERSION=7.8.22
     - LMOD_VERSION=7.8.22 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
-    - LMOD_VERSION=8.0.9
-    - LMOD_VERSION=8.0.9 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
+    - LMOD_VERSION=8.2.3
+    - LMOD_VERSION=8.2.3 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
     - ENV_MOD_VERSION=3.2.10 TEST_EASYBUILD_MODULES_TOOL=EnvironmentModulesC TEST_EASYBUILD_MODULE_SYNTAX=Tcl
     - ENV_MOD_TCL_VERSION=1.147 TEST_EASYBUILD_MODULES_TOOL=EnvironmentModulesTcl TEST_EASYBUILD_MODULE_SYNTAX=Tcl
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,15 +1,13 @@
...
   matrix:
     # purposely specifying slowest builds first, to gain time overall
-    - LMOD_VERSION=6.5.1
-    - LMOD_VERSION=6.5.1 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
     - LMOD_VERSION=7.8.22
     - LMOD_VERSION=7.8.22 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
-    - LMOD_VERSION=8.0.9
-    - LMOD_VERSION=8.0.9 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
+    - LMOD_VERSION=8.2.3
+    - LMOD_VERSION=8.2.3 TEST_EASYBUILD_MODULE_SYNTAX=Tcl
```
wpoely86: why use 8.2.3 here and 8.1.14 in the github workflow?

easybuild/tools/modules.py
```diff
@@ -1110,6 +1128,7 @@ def update(self):
 
 class EnvironmentModulesTcl(EnvironmentModulesC):
     """Interface to (Tcl) environment modules (modulecmd.tcl)."""
+    NAME = "ancient Tcl-only Environment Modules"
```
wpoely86: nothing shorter came to mind?

.travis.yml
boegel: should be 8.2.3 (latest release) in both, will fix...

easybuild/tools/modules.py
boegel: I'm open to suggestions. `Environment Modules` is not specific enough, and this type of modules tool doesn't have proper versioning.

We should probably deprecate support for this one too, since I doubt anyone is still using it...
I'll do that in a separate PR once this one is merged.

PR #3078, opened 2019-11-11
take into account that platform.linux_distribution and platform.dist was removed in Python 3.8

boegel:
see https://bugs.python.org/issue28167


akesandgren:
Going in, thanks @boegel!

PR #3083, opened 2019-11-14
take into account that external modules may not be visible directly (due to module hierarchy)

boegel:
Modules provided through OpenHPC packages are organized hierarchically, where for example a `gnu*` module must be loaded before an `openmpi*` or `openblas` module becomes available for loading, an `openmpi*` must be loaded before `fftw` can be loaded, etc.

These minor enhancements that basically loosen up the restrictions for using external modules (see https://easybuild.readthedocs.io/en/latest/Using_external_modules.html) a bit are sufficient to let EasyBuild also leverage external modules that are organized hierarchically.

The assumption here is that the dependencies (or toolchain components) are ordered correctly, i.e.  that if module Y requires module X to be loaded first, X is listed before Y.

Although the checks here are loosened up a bit, incorrectly ordering things will still result in an installation failure, since the sanity check which tries to load the resulting module will fail in that case.

I've also extended `parse_external_modules_metadata` a bit so no unknown keys are listed in the metadata for external modules (to avoid silly mistakes like typos going undetected).

cc @koomie @adrianreber

### DIFF
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -125,10 +125,10 @@ def find_resolved_modules(easyconfigs, avail_modules, modtool, retain_all_deps=F
         for dep in easyconfig['dependencies']:
             dep_mod_name = dep.get('full_mod_name', ActiveMNS().det_full_module_name(dep))
 
-            # treat external modules as resolved when retain_all_deps is enabled (e.g., under --dry-run),
+            # always treat external modules as resolved,
             # since no corresponding easyconfig can be found for them
-            if retain_all_deps and dep.get('external_module', False):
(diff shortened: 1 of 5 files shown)


adrianreber:
@boegel Thanks for working on this!

boegel:
@adrianreber I got a proof of concept to work where I build `TopHat` using EasyBuild on top of a toolchain that leverages OpenHPC components + using an OpenHPC-provided `boost` as a dependency.

I need to open one or two more PRs in EasyBuild to get all required changes in, I'll send you & @koomie more details via email later to follow up on this.

The upcoming EasyBuild v4.1.0 (to be released around the end of the month) should have everything in place what is needed...

PR #3085, opened 2019-11-15
adds support for including 'extensions' statement in Lua modules with Lmod 8.2+

wpoely86:
Lmod 8.2 supports now extensions, this PR makes EB generate the needed lines in the module files for it.


easybuild/tools/module_generator.py
```diff
@@ -1110,6 +1123,11 @@ def get_description(self, conflict=True):
         for line in self._generate_whatis_lines():
             whatis_lines.append("whatis(%s%s%s)" % (self.START_STR, self.check_str(line), self.END_STR))
 
+        provide_list = self._generate_provides_list()
+        if self.modules_tool.supports_extensions and provide_list:
+            provide_list_str = ', '.join(['"%s"' % x for x in provide_list])
+            lines.extend(['', 'extensions(%s)' % provide_list_str])
```
boegel: We have two options here:

* either we include this if the Lmod version used at installation time is recent enough (which is what you're doing here)
* or we add a conditional in the module file itself, using `LmodVersion()` (that's what Robert recommended)

The benefit of the latter approach is that the module files will still work when loaded with an older Lmod version, and there's basically no overhead in using the conditional.

wpoely86: It's a design choice, I don't have a strong preference for either.

But adding support in a module file for something that doesn't work with the current Lmod seems a bit weird to me? Especially as it's pretty easy to regenerate a module.

wpoely86: Also, it might surprise people. You update Lmod and suddenly the output of av etc changes

boegel: It's a protection mechanism too though. You may be installing software on a system with a slightly more recent version of Lmod (8.2+), and then consume the modules elsewhere with a slightly older Lmod (<8.2).
That wouldn't be very nice either... It's certainly less serious than being a bit careful and only using `extensions` conditionally.

Output of `ml av` will change anyway as soon as you install something with extensions after updating to Lmod 8.2, and you're probably not going to notice straight away (I certainly rarely check the generated module files mannually ;) ).

boegel:
@wpoely86 Tests are still broken when using Lmod 8.2+ (@boegelbot only reports broken tests from Travis CI).

```
FAIL: test_module_extensions (test.framework.module_generator.LuaModuleGeneratorTest)
test the extensions() for extensions
  File "test/framework/module_generator.py", line 585, in test_module_extensions
    self.assertTrue(re.search(pattern, desc, re.M), "Pattern '%s' found in: %s" % (pattern, desc))
AssertionError: None is not true : Pattern '^extensions\("bar/0.0", "barbar/0.0", "l/s", "toy/0.0"\)' found in: help([==[

Description
===========
Toy C program, 100% toy.

More information
================
 - Homepage: https://easybuilders.github.io/easybuild


Included extensions
===================
bar-0.0, barbar-0.0, l-s, toy-0.0
]==])

whatis([==[Description: Toy C program, 100% toy.]==])
whatis([==[Homepage: https://easybuilders.github.io/easybuild]==])
whatis([==[URL: https://easybuilders.github.io/easybuild]==])
whatis([==[Extensions: bar-0.0, barbar-0.0, l-s, toy-0.0]==])

local root = "/tmp/eb-CRdgXt/eb-Qajkhe/eb-FkUxoO/tmp9r3vc4/software/toy/0.0-gompi-2018a-test"

conflict("toy")

if convertToCanonical(LmodVersion()) > convertToCanonical("8.2.0") then
    extensions("bar/0.0", "barbar/0.0", "l/s", "toy/0.0")
end

Ran 653 tests in 874.006s

FAILED (failures=1)
ERROR: Not all tests were successful.
```

easybuild/tools/module_generator.py
```diff
@@ -519,6 +519,15 @@ def _generate_extension_list(self):
 
         return extensions
 
+    def _generate_provides_list(self):
```
boegel: rename `provides` to `extensions`?

easybuild/tools/module_generator.py
```diff
@@ -984,6 +993,8 @@ class ModuleGeneratorLua(ModuleGenerator):
     LOAD_TEMPLATE_DEPENDS_ON = 'depends_on("%(mod_name)s")'
     IS_LOADED_TEMPLATE = 'isloaded("%s")'
 
+    VERSION_CHECK = 'convertToCanonical(LmodVersion()) > convertToCanonical("%(ver_maj)s.%(ver_min)s")'
```
boegel: rename to `LMOD_VERSION_CHECK_EXPR`?

easybuild/tools/module_generator.py
```diff
@@ -998,6 +1009,15 @@ def __init__(self, *args, **kwargs):
             if self.modules_tool.version and LooseVersion(self.modules_tool.version) >= LooseVersion('7.7.38'):
                 self.DOT_MODULERC = '.modulerc.lua'
 
+    def check_version(self, minimal_version_maj, minimal_version_min):
```
boegel: You should also add this method to the `ModuleGenerator` class, and make it `raise NotImplementedError` there in case someone trigger it for `ModuleGeneratorTcl`?

Either that, or rename to something with a leading underscore to make it "private", like `_check_lmod_version`

easybuild/tools/module_generator.py
```diff
@@ -998,6 +1009,15 @@ def __init__(self, *args, **kwargs):
...
+        """
+        Check the minimal version of the moduletool in the module file
+        """
+        return self.VERSION_CHECK % {
```
boegel: if `VERSION_CHECK` is only used in one place, I'm not sure it's worth adding a class constant for, just embed the `VERSION_CHECK` in this method (which could be a static method btw)

easybuild/tools/module_generator.py
```diff
@@ -1110,6 +1130,12 @@ def get_description(self, conflict=True):
         for line in self._generate_whatis_lines():
             whatis_lines.append("whatis(%s%s%s)" % (self.START_STR, self.check_str(line), self.END_STR))
 
+        provide_list = self._generate_provides_list()
```
boegel: `provides_list` -> `extensions_list`

easybuild/tools/module_generator.py
```diff
@@ -519,6 +519,15 @@ def _generate_extension_list(self):
 
         return extensions
 
+    def _generate_provides_list(self):
+        """
+        Generate a list of all extensions in name/version format
+        """
+        exts_list = self.app.cfg['exts_list']
+        exts_ver_list = sorted(['%s/%s' % (ext[0], ext[1]) for ext in exts_list], key=str.lower)
```
boegel: clarify here that this is `<extenion_name>/<extension_version>`

easybuild/tools/module_generator.py
```diff
@@ -1110,6 +1130,12 @@ def get_description(self, conflict=True):
         for line in self._generate_whatis_lines():
             whatis_lines.append("whatis(%s%s%s)" % (self.START_STR, self.check_str(line), self.END_STR))
 
+        provide_list = self._generate_provides_list()
+
+        if provide_list:
+            provide_list_mod = 'extensions(%s)' % ', '.join(['"%s"' % x for x in provide_list])
+            lines.extend(['', self.conditional_statement(self.check_version("8", "2"), provide_list_mod)])
```
boegel: add a comment explaining why we're wrapping this under an Lmod version check (bonus points if you include a reference to the Lmod docs)

PR #3086, opened 2019-11-15
trim set of test configurations in Travis CI

boegel:
removed:

* testing with Lmod 8 (still covered in GitHub CI)
* testing with Python 3.x (except for Lmod 7.x + Lua module syntax)

All these are still covered in GitHub CI, where test budget is less limited.


akesandgren:
Going in, thanks @boegel!

PR #3088, opened 2019-11-16
flesh out env_vars_external_module from Toolchain._simulated_load_dependency_module

boegel:
... so we can leverage `env_vars_external_module` in https://github.com/easybuilders/easybuild-easyblocks/pull/1849


akesandgren:
Going in, thanks @boegel!

Issue #2441, opened 2018-03-14
BUG: Backup modules created with --module-only


shahzebsiddiqui:
@damianam @boegel 

Looks like whenever I want to run --module-only I get a backup module created that is part of ``module av``

The backup should be hidden but I assume you had to specify ``--backup-modules`` in order to do this. Right now it is doing by default.

I am running Lmod 6.5.1 on Redhat 7.3

```
[siddis14@amrndhl1157 config]$ eb M4-1.4.17.eb --module-only --rebuild
== temporary log file in case of crash /hpc/grid/hpcws/hpcswbuild/easybuild/IvyBridge/redhat/7.3/tmp/eb-4tZ2m3/easybuild-2j68H7.log
== processing EasyBuild easyconfig /lustre/workspace/home/hpcswbuild/siddis14/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.17.eb
== building and installing M4/1.4.17...
== fetching files [skipped]
== creating build dir, resetting environment...
== backup of existing module file stored at /nfs/grid/software/easybuild/IvyBridge/redhat/7.3/modules/all/M4/1.4.17.lua.bak_20180314160540
== unpacking [skipped]
== patching [skipped]
== preparing...
== configuring [skipped]
== building [skipped]
== testing [skipped]
== installing [skipped]
== taking care of extensions [skipped]
== postprocessing [skipped]
== sanity checking [skipped]
== cleaning up [skipped]
== creating module...
== comparing module file with backup /nfs/grid/software/easybuild/IvyBridge/redhat/7.3/modules/all/M4/1.4.17.lua.bak_20180314160540; no differences found
== permissions [skipped]
== packaging [skipped]
== COMPLETED: Installation ended successfully
== Results of the build can be found in the log file(s) /nfs/grid/software/easybuild/IvyBridge/redhat/7.3/software/M4/1.4.17/easybuild/easybuild-M4-1.4.17-20180314.160541.log
== Build succeeded for 1 out of 1
== Temporary log file(s) /hpc/grid/hpcws/hpcswbuild/easybuild/IvyBridge/redhat/7.3/tmp/eb-4tZ2m3/easybuild-2j68H7.log* have been removed.
[siddis14@amrndhl1157 config]$ ml av M4/1.4.17
------------------------------------------------------------------- /nfs/grid/software/easybuild/IvyBridge/redhat/7.3/modules/all --------------------------------------------------------------------
[...]

PR #3089, opened 2019-11-17
hide backup module file when using Lmod 6.x (fixes #9302)

boegel:
It looks like the fix that was made in #2463 for the problem reported in #2441 was not correct, since the same issue surfaced again in #9302.

Apparently Lmod consider any file it finds as a module file, regardless of whether it has `.lua` in the filename, so we must always hide backup module files when using Lmod 6.x.


akesandgren:
Going in, thanks @boegel!

PR #3094, opened 2019-11-23
fix semantics of --set-default-module: only set default for specified easyconfigs, not for the ones that are installed as dependencies via --robot

### LINKED ISSUE #3090: unexpected behaviour with --set-default-module --robot
The EasyBuild option `--set-default-module` will apply to all the recipes in the same command line: this means that every dependency of the build using `--set-default-module` will be set default, overwriting existing  defaults for different versions of the same module name. I paste here the info on the flag:
```
eb -H | grep default
    --set-default-module
                        Set the generated module as default (def False)
```
This is happening frequently with `CMake`, since there are builds using different `CMake` versions.

I provide an example below with Bison, that depends on M4 (the lower case naming scheme is enabled on the build system and I skip lines of output using the symbol `[...]` to make it shorter): 
```
eb Bison-3.3.2.eb -Dr
[...]
 * [ ] $CFGS/m/M4/M4-1.4.18.eb (module: m4/1.4.18)
 * [ ] $CFGS/b/Bison/Bison-3.3.2.eb (module: bison/3.3.2)

eb Bison-3.3.2.eb -r --set-default-module
 [...]

module avail
------------------------------------------------------------------ 
bison/3.3.2(default)  m4/1.4.18(default)
```

boegel:
fixes #3090

### DIFF
--- a/test/framework/options.py
+++ b/test/framework/options.py
@@ -3980,6 +3980,60 @@ class CommandLineOptionsTest(EnhancedTestCase):
         else:
             self.assertTrue(False, "Uknown module syntax: %s" % get_module_syntax())
 
+    def test_set_default_module_robot(self):
+        """Test use of --set-default-module --robot."""
+        # create two test easyconfigs, one depending on the other
+        # (using dummy Toolchain easyblock included in the tests)
+        test_ec = os.path.join(self.test_prefix, 'test.eb')
+        write_file(test_ec, '\n'.join([
+            "easyblock = 'Toolchain'",
+            "name = 'test'",
+            "version = '1.0'",
+            "homepage = 'https://example.com'",
+            "description = 'this is just a test'",
+            "toolchain = SYSTEM",
+            "dependencies = [('thisisjustatestdep', '3.14')]",
+        ]))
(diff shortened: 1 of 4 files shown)


test/framework/options.py
```diff
@@ -3980,6 +3980,60 @@ def test_set_default_module(self):
         else:
             self.assertTrue(False, "Uknown module syntax: %s" % get_module_syntax())
 
+    def test_set_default_module_robot(self):
+        """Test use of --set-default-module --robot."""
+        # create two test easyconfigs, one depending on the other
+        # (using dummy Toolchain easyblock included in the tests)
+        test_ec = os.path.join(self.test_prefix, 'test.eb')
```
akesandgren: Won't verify-easyconfig-filenames=True in site easyconfig.cfg (or users) cause problems with that file name for someone running the tests locally?

boegel: Could be, but if that's so then we need to isolate the tests from the user's configuration.

The tests are assumed to run with the default configuration.

akesandgren: ok

akesandgren:
Going in, thanks @boegel!

PR #3097, opened 2019-11-24
fix generated module statements in case only a single version is listed in multi_deps

### LINKED ISSUE #3080: Invalid module code when multi_deps contains a single version
For consistency across two recipes, one that work with Python 3.5, 3.6, 3.7, 3.8, and the other one with 2.7, I wanted to use 
```
multi_deps = {
	'Python': ['2.7.14']
}
``` 

This resulted in the following invalid Module code : 
```
if mode() == "unload" or  then
    depends_on("python")
else
    depends_on("python/2.7.14")
end
````

@boegel @bartoldeman

boegel:
fixes #3080 

cc @mboisson


migueldiascosta:
Going in, thanks @boegel!

PR #3102, opened 2019-11-26
add template %(module_name_full)s

smoors:
for background, see discussion here: https://github.com/easybuilders/easybuild-easyblocks/pull/1857/files

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2799,7 +2799,10 @@ class EasyBlock(object):
         """Update the the easyconfig template dictionary with easyconfig.TEMPLATE_NAMES_EASYBLOCK_RUN_STEP names"""
 
         for name in TEMPLATE_NAMES_EASYBLOCK_RUN_STEP:
-            self.cfg.template_values[name[0]] = str(getattr(self, name[0], None))
+            if name[0] == 'module_name_full':
+                self.cfg.template_values[name[0]] = ActiveMNS().det_full_module_name(self.cfg)
+            else:
+                self.cfg.template_values[name[0]] = str(getattr(self, name[0], None))
         self.cfg.generate_template_values()
 
     def _skip_step(self, step, skippable):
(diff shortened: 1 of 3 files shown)


smoors:
for some reason travis is still indicated as pending even though all its tests have passed.
will close and reopen..

easybuild/framework/easyblock.py
```diff
@@ -2799,7 +2799,10 @@ def update_config_template_run_step(self):
         """Update the the easyconfig template dictionary with easyconfig.TEMPLATE_NAMES_EASYBLOCK_RUN_STEP names"""
 
         for name in TEMPLATE_NAMES_EASYBLOCK_RUN_STEP:
-            self.cfg.template_values[name[0]] = str(getattr(self, name[0], None))
+            if name[0] == 'module_name_full':
+                self.cfg.template_values[name[0]] = ActiveMNS().det_full_module_name(self.cfg)
```
boegel: You should be able to use `self.cfg.full_mod_name` here (no need to re-determine the module name)

easybuild/framework/easyconfig/templates.py
```diff
@@ -73,6 +73,7 @@
 TEMPLATE_NAMES_EASYBLOCK_RUN_STEP = [
     ('installdir', "Installation directory"),
     ('builddir', "Build directory"),
+    ('module_name_full', "Full module name"),
```
boegel: wouldn't `full_module_name` make more sense?

smoors:
> @smoors I think you actually want/need the "short" module name for the SCOOP use case, which is the "user-facing" module name (for example `SCOOP/0.7.1.1` when using a module hierarchy, rather than `Compile/GCCcore/8.2.0/SCOOP/0.7.1.1`, which should never be loaded using that full name).

@boegel ok, I am not sure how HMNS works exactly. but in order to load SCOOP like that, don't you have to load the toolchain first to make it available?:
```
module load GCCcore/8.2.0 SCOOP/0.7.1.1
```

easybuild/framework/easyconfig/templates.py
smoors: depends how you look at it :) either way is fine to me.

smoors:
this is going to be a pain to get right for every possible setup, so I will just define the module load command in the easyconfig instead.

PR #3107, opened 2019-11-30
make including 'extensions' statement in generated Lua module files opt-in for now

boegel:
Lmod apparently chokes on a module file is the list of extensions is too long (which is the case for `Perl`) after including an `extensions` statement (see @wpoely86's PR #3085):
```
== 2019-11-30 08:50:24,617 build_log.py:164 ERROR EasyBuild crashed with an error (at easybuild/base/exceptions.py:124 in __init__): Module command 'module load Perl/5.30.0-GCCcore-8.3.0' failed with exit code 1; stderr: Lmod has detected the following error: Unable to load module because of error when evaluating modulefile:
     /tmp/eb-idlfoc6k/tmp2s1c6_jv/all/Perl/5.30.0-GCCcore-8.3.0.lua: [string "help([==[..."]:118: function or expression needs too many registers near '"Test::Simple/1.302168"'
     Please check the modulefile and especially if there is a the line number specified in the above message
While processing the following module(s):
    Module fullname            Module Filename
    ---------------            ---------------
    Perl/5.30.0-GCCcore-8.3.0  /tmp/eb-idlfoc6k/tmp2s1c6_jv/all/Perl/5.30.0-GCCcore-8.3.0.lua
; stdout:
false
 (at easybuild/tools/modules.py:758 in run_module)
```

It may have something to do with the complete module file being too large, since we include the list of extensions 3 times now (in description, in `whatis`, in `extensions`)...

So for now, let's make including an `extensions` statement in the generated module file opt-in, until we figure out this issue (or have a fix in Lmod).

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -228,6 +228,7 @@ BUILD_OPTIONS_CMDLINE = {
         'lib64_fallback_sanity_check',
         'logtostdout',
         'minimal_toolchains',
+        'module_extensions',
         'module_only',
         'package',
         'read_only_installdir',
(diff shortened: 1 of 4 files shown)


boegel:
Also reported this to Lmod: https://github.com/TACC/Lmod/issues/428

akesandgren:
Going in, thanks @boegel!

PR #3111, opened 2019-12-03
take into account that dependency version could be a dict rather than a string value in template_constant_dict

boegel:
`template_constant_dict` didn't take into account yet that dependency versions can now also be dictionaries (see https://github.com/easybuilders/easybuild-framework/pull/3047), which results in this crash when running `eb --list-software` when it hits the modified `Java` easyconfig from https://github.com/easybuilders/easybuild-easyconfigs/pull/9081:

```
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Volumes/work/easybuild-framework/easybuild/main.py", line 485, in <module>
    main()
  File "/Volumes/work/easybuild-framework/easybuild/main.py", line 256, in main
    print(list_software(output_format=options.output_format, detailed=options.list_software == 'detailed'))
  File "easybuild/tools/docs.py", line 544, in list_software
    template_values = template_constant_dict(ec)
  File "/Volumes/work/easybuild-framework/easybuild/framework/easyconfig/templates.py", line 232, in template_constant_dict
    dep_version_parts = dep_version.split('.')
AttributeError: 'dict' object has no attribute 'split'
```

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -68,7 +68,7 @@ from easybuild.tools.module_naming_scheme.utilities import avail_module_naming_s
 from easybuild.tools.module_naming_scheme.utilities import det_hidden_modname, is_valid_module_name
 from easybuild.tools.modules import modules_tool
 from easybuild.tools.py2vs3 import OrderedDict, create_base_metaclass, string_type
-from easybuild.tools.systemtools import check_os_dependency, get_cpu_architecture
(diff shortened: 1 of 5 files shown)


migueldiascosta:
Going in, thanks @boegel!

PR #3118, opened 2019-12-10
Add check_log_for_errors to detect and handle multiple errors

Flamefire:
In preparation for https://github.com/easybuilders/easybuild-easyblocks/issues/157 I created this PoC.

The specification is (as documented):

```
    Check logTxt for messages matching regExps in order and do appropriate action
    :param logTxt: String containing the log, will be split into individual lines
    :param regExps: List of: regular expressions (as RE or string) to error on,
                    or tuple of regular expression and action (any of [IGNORE, WARN, ERROR])
```

In the end you'll get (if appropriate) a logged warning and an exception about the found issues.

I guess this function could even be split: 1 that returns the `(warnings, errors)` tuple and one that also increases `errors_found_in_log`, logs the warning and raises the exception (which would be the most common usage), but the name `parse_log_for_error` is already taken. Maybe name it `parse_log_for_errors`? The "old" `parse_log_for_error` can be implemented in terms of that but in can probably be simply deprecated. It is strange, does to much and yet almost nothing. Similar for `parse_cmd_output`. So for backwards compatibility I'd rather have EasyBlocks call `check_log_for_errors` directly and pass `None` as the `regExp` to `run_cmd* `

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -539,7 +539,7 @@ def parse_cmd_output(cmd, stdouterr, ec, simple, log_all, log_ok, regexp):
     if use_regexp or regexp:
         res = parse_log_for_error(stdouterr, regexp, msg="Command used: %s" % cmd)
         if len(res) > 0:
-            message = "Found %s errors in command output (output: %s)" % (len(res), ", ".join([r[0] for r in res]))
+            message = "Found %s errors in command output (output: %s)" % (len(res), "\n\t".join([r[0] for r in res]))
             if use_regexp:
                 raise EasyBuildError(message)
             else:
@@ -643,7 +643,7 @@ def check_log_for_errors(log_txt, reg_exps):
     errors_found_in_log += len(warnings) + len(errors)
     if warnings:
         _log.warning("Found %s potential error(s) in command output (output: %s)",
-                     len(warnings), ", ".join(warnings))
+                     len(warnings), "\n\t".join(warnings))
     if errors:
         raise EasyBuildError("Found %s error(s) in command output (output: %s)",
-                             len(errors), ", ".join(errors))
+                             len(errors), "\n\t".join(errors))
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -589,3 +589,59 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
                   (regExp, '\n'.join([x[0] for x in res])))
 
     return res
+
+
+def check_log_for_errors(logTxt, regExps):
+    """
+    Check logTxt for messages matching regExps in order and do appropriate action
+    :param logTxt: String containing the log, will be split into individual lines
+    :param regExps: List of: regular expressions (as RE or string) to error on,
+                    or tuple of regular expression and action (any of [IGNORE, WARN, ERROR])
+    """
+    global errors_found_in_log
+
+    def is_regexp_object(objToTest):
+        try:
+            objToTest.match('')
+            return True
+        except AttributeError:
+            return False
+
+    # Avoid accidentally passing a single element
+    assert isinstance(regExps, list), "regExps must be a list"
+    regExpTuples = []
+    for cur in regExps:
+        try:
+            if isinstance(cur, str):
+                regExpTuples.append((re.compile(cur), ERROR))
+            elif is_regexp_object(cur):
+                regExpTuples.append((cur, ERROR))
+            elif len(cur) != 2:
+                raise TypeError("Invalid tuple")
+            elif not isinstance(cur[0], str) and not is_regexp_object(cur[0]):
+                raise TypeError("Invalid RegExp in tuple")
+            elif cur[1] not in (IGNORE, WARN, ERROR):
+                raise TypeError("Invalid action in tuple")
+            else:
+                regExpTuples.append((re.compile(cur[0]) if isinstance(cur[0], str) else cur[0], cur[1]))
+        except TypeError:
+            raise EasyBuildError("Invalid input: No RegExp or tuple of RegExp and action: %s" % str(cur))
+    warnings = []
+    errors = []
+    for l in logTxt.split('\n'):
+        for regExp, action in regExpTuples:
+            m = regExp.search(l)
+            if m:
+                if action == ERROR:
+                    errors.append(l)
+                elif action == WARN:
+                    warnings.append(l)
+                break
+    errors_found_in_log += len(warnings) + len(errors)
+    if warnings:
+        _log.warning("Found %s potential errors in command output (output: %s)" %
+                     (len(warnings), ", ".join(warnings)))
+    if errors:
+        raise EasyBuildError("Found %s errors in command output (output: %s)" %
+                             (len(errors), ", ".join(errors)))
+
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -600,8 +600,10 @@ def extract_errors_from_log(log_txt, reg_exps):
     :return (warnings, errors) as lists of lines containing a match
     """
 
-    # Avoid accidentally passing a single element
-    assert isinstance(reg_exps, list), "reg_exps must be a list"
+    # promote single string value to list, since code below expects a list
+    if isinstance(reg_exps, string_type):
+        reg_exps = [reg_exps]
+
     re_tuples = []
     for cur in reg_exps:
         try:
@@ -614,6 +616,7 @@ def extract_errors_from_log(log_txt, reg_exps):
             re_tuples.append((re.compile(reg_exp), action))
         except Exception as e:
             raise EasyBuildError("Invalid input: No RegExp or tuple of RegExp and action '%s' (%s)", str(cur), e)
+
     warnings = []
     errors = []
     for line in log_txt.split('\n'):
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -591,57 +591,56 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
     return res
 
 
-def check_log_for_errors(logTxt, regExps):
+def extract_errors_from_log(log_txt, reg_exps):
     """
-    Check logTxt for messages matching regExps in order and do appropriate action
-    :param logTxt: String containing the log, will be split into individual lines
-    :param regExps: List of: regular expressions (as RE or string) to error on,
+    Check log_txt for messages matching regExps and return warnings and errors
+    :param log_txt: String containing the log, will be split into individual lines
+    :param reg_exps: List of: regular expressions (as strings) to error on,
+                    or tuple of regular expression and action (any of [IGNORE, WARN, ERROR])
+    :return (warnings, errors) as lists of lines containing a match
+    """
+
+    # Avoid accidentally passing a single element
+    assert isinstance(reg_exps, list), "reg_exps must be a list"
+    re_tuples = []
+    for cur in reg_exps:
+        try:
+            if isinstance(cur, str):
+                reg_exp, action = cur, ERROR
+            else:
+                reg_exp, action = cur
+            if not isinstance(reg_exp, str) or action not in (IGNORE, WARN, ERROR):
+                raise TypeError("Invalid types")
+            re_tuples.append((re.compile(reg_exp), action))
+        except Exception as e:
+            raise EasyBuildError("Invalid input: No RegExp or tuple of RegExp and action '%s' (%s)", str(cur), e)
+    warnings = []
+    errors = []
+    for line in log_txt.split('\n'):
+        for reg_exp, action in re_tuples:
+            if reg_exp.search(line):
+                if action == ERROR:
+                    errors.append(line)
+                elif action == WARN:
+                    warnings.append(line)
+                break
+    return warnings, errors
+
+
+def check_log_for_errors(log_txt, reg_exps):
+    """
+    Check log_txt for messages matching regExps in order and do appropriate action
+    :param log_txt: String containing the log, will be split into individual lines
+    :param reg_exps: List of: regular expressions (as strings) to error on,
                     or tuple of regular expression and action (any of [IGNORE, WARN, ERROR])
     """
     global errors_found_in_log
+    warnings, errors = extract_errors_from_log(log_txt, reg_exps)
 
-    def is_regexp_object(objToTest):
-        try:
-            objToTest.match('')
-            return True
-        except AttributeError:
-            return False
-
-    # Avoid accidentally passing a single element
-    assert isinstance(regExps, list), "regExps must be a list"
-    regExpTuples = []
-    for cur in regExps:
-        try:
-            if isinstance(cur, str):
-                regExpTuples.append((re.compile(cur), ERROR))
-            elif is_regexp_object(cur):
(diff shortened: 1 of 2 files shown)


easybuild/tools/run.py
```diff
@@ -589,3 +589,59 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
...
+                raise TypeError("Invalid RegExp in tuple")
+            elif cur[1] not in (IGNORE, WARN, ERROR):
+                raise TypeError("Invalid action in tuple")
+            else:
+                regExpTuples.append((re.compile(cur[0]) if isinstance(cur[0], str) else cur[0], cur[1]))
+        except TypeError:
+            raise EasyBuildError("Invalid input: No RegExp or tuple of RegExp and action: %s" % str(cur))
+    warnings = []
+    errors = []
+    for l in logTxt.split('\n'):
```
boegel: code style remarks after a quick review:

* expect for older parts of the code we haven't touched/cleaned up in a while, we usually don't use camel-case for variable, so please use `logtxt` or `log_txt` rather than `logTxt` (same goes for `regExps`, `regExpTuples`, `objToTest`
* avoid using single-letter variables outside of list comprehensions (like `l` here and `m` below)

easybuild/tools/run.py
```diff
@@ -589,3 +589,59 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
...
+                elif action == WARN:
+                    warnings.append(l)
+                break
+    errors_found_in_log += len(warnings) + len(errors)
+    if warnings:
+        _log.warning("Found %s potential errors in command output (output: %s)" %
+                     (len(warnings), ", ".join(warnings)))
+    if errors:
+        raise EasyBuildError("Found %s errors in command output (output: %s)" %
+                             (len(errors), ", ".join(errors)))
```
boegel: better/cleaner/faster (no need to use `%` for `log` methods or in `EasyBuildError`)

```python
raise EasyBuildError("Found %s errors in command output (output: %s)", len(errors), ", ".join(errors))

File: test/framework/run.py [hunk 3]
```diff
@@ -508,6 +509,59 @@ def test_run_cmd_stream(self):
...
+        self.assertErrorRegex(EasyBuildError, "Invalid input:", check_log_for_errors, "", [("42", "invalid-mode")])
+        self.assertErrorRegex(EasyBuildError, "Invalid input:", check_log_for_errors, "", [("42", IGNORE, "")])
+
+        input_text = "\n".join([
+            "OK",
+            "error found",
+            "test failed",
+            "msg: allowed-test failed",
+            "enabling -Werror",
+            "the process crashed with 0"
```
boegel: Typical genuine error messages:

* `error: ...` (compiler errors)
* `Error 1`, `Error 2`, ... (`make` errors)

easybuild/tools/run.py
```diff
@@ -589,3 +589,59 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
...
+
+    def is_regexp_object(objToTest):
+        try:
+            objToTest.match('')
+            return True
+        except AttributeError:
+            return False
+
+    # Avoid accidentally passing a single element
+    assert isinstance(regExps, list), "regExps must be a list"
```
boegel: Why not promote a single string value to a single-element list, so you can use `check_log_for_errors("error:")` for example?

```python
# promote single string value to list, since code below expects a list
if isinstance(regexps, string_type):
    regexps = [regexps]
```

easybuild/tools/run.py
```diff
@@ -589,3 +589,59 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
...
+                    errors.append(l)
+                elif action == WARN:
+                    warnings.append(l)
+                break
+    errors_found_in_log += len(warnings) + len(errors)
+    if warnings:
+        _log.warning("Found %s potential errors in command output (output: %s)" %
+                     (len(warnings), ", ".join(warnings)))
+    if errors:
+        raise EasyBuildError("Found %s errors in command output (output: %s)" %
```
boegel: `output:` should be `errors:`?

`errors` should be `error(s)` (since there could be only one, see your last test case)

easybuild/tools/run.py
```diff
@@ -589,3 +589,59 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
...
+            return False
+
+    # Avoid accidentally passing a single element
+    assert isinstance(regExps, list), "regExps must be a list"
+    regExpTuples = []
+    for cur in regExps:
+        try:
+            if isinstance(cur, str):
+                regExpTuples.append((re.compile(cur), ERROR))
+            elif is_regexp_object(cur):
```
boegel: Is there a huge benefit in also accepting compiled regular expressions as inputs to this function?

Why not enforce to only pass strings, and then compile the regexp patterns here always, that could simplify the code here a bit?

easybuild/tools/run.py
Flamefire: thx, done

easybuild/tools/run.py
Flamefire: ok

test/framework/run.py
Flamefire: Is this important for a (artificial) test? Only thing is that it needs to match the regexps used below

easybuild/tools/run.py
Flamefire: Because I'd expect usually more than 1 regexp. At least a few "IGNORE" in front. And on call site you can always use `check_log_for_errors(["error:"])`. This also avoids additional complexity (I was already hesitant to allow either a tuple or a string/re)

easybuild/tools/run.py
Flamefire: It was like that before and actually it is the output containing the error message. I'd just leave the name

Flamefire: Oh and what'd you say about separating warning/error output lines by `\n` instead? Currently it is hard to read:
`== 2019-12-11 13:14:12,211 run.py:546 WARNING Found 26 errors in command output (output: /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/document.h:1881: warning: preprocessing issue while doing constant expression evaluation: syntax error, Searching for files in directory /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error, Preprocessing /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/en.h..., Parsing file /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/en.h..., Preprocessing /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h..., Parsing file /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h..., Parsing file /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/document.h:1881: warning: preprocessing issue while doing constant expression evaluation: syntax error, /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h:123: warning: Member operator==(const ParseResult &that) const  (function) of class rapidjson::ParseResult is not documented., /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h:124: warning: Member operator==(ParseErrorCode code) const  (function) of class rapidjson::ParseResult is not documented., /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h:125: warning: Member operator==(ParseErrorCode code, const ParseResult &err) (friend) of class rapidjson::ParseResult is not documented., Generating code for file include/rapidjson/error/en.h..., Generating code for file include/rapidjson/error/error.h..., Generating docs for file include/rapidjson/error/error.h..., /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/document.h:1881: warning: preprocessing issue while doing constant expression evaluation: syntax error, Searching for files in directory /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error, Preprocessing /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/en.h..., Parsing file /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/en.h..., Preprocessing /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h..., Parsing file /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h..., Parsing file /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1/tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/document.h:1881: warning: preprocessing issue while doing constant expression evaluation: syntax error, /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h:123: warning: Member operator==(const ParseResult &that) const  (function) of class rapidjson::ParseResult is not documented., /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h:124: warning: Member operator==(ParseErrorCode code) const  (function) of class rapidjson::ParseResult is not documented., /tmp/s3248973-EasyBuild/RapidJSON/1.1.0/GCCcore-8.2.0/rapidjson-1.1.0/include/rapidjson/error/error.h:125: warning: Member operator==(ParseErrorCode code, const ParseResult &err) (friend) of class rapidjson::ParseResult is not documented., Generating code for file include/rapidjson/error/en.h..., Generating code for file include/rapidjson/error/error.h..., Generating docs for file include/rapidjson/error/error.h...)
`

easybuild/tools/run.py
Flamefire: I've just seen one can use e.g. `(?i)` for enabling case-insensitive mode in the re string.
So I'll just make it string then.

easybuild/tools/run.py
Flamefire: So basically: `raise EasyBuildError("Found %s errors in command output. Output:\n\t%s)", ..., "\n\t".join(...))`

boegel: one error per line makes sense, that's indeed a lot easier to read...

easybuild/tools/run.py
boegel: I still think it's convenient to promote a single value to a list, we do this in several other places too, so you can just use `check_log_for_errors("error:")`

test/framework/run.py
boegel: not really I guess...

easybuild/tools/run.py
Flamefire: Ok, done

easybuild/tools/run.py
Flamefire: Done

test/framework/run.py
```diff
@@ -520,6 +527,56 @@ def test_run_cmd_stream(self):
...
+        self.assertErrorRegex(EasyBuildError, expected_error_msg, check_log_for_errors, input_text,
+                              r"\b(error|crashed)\b")
+        # List of string(s)
+        self.assertErrorRegex(EasyBuildError, expected_error_msg, check_log_for_errors, input_text,
+                              [r"\b(error|crashed)\b"])
+        # List of tuple(s)
+        self.assertErrorRegex(EasyBuildError, expected_error_msg, check_log_for_errors, input_text,
+                              [(r"\b(error|crashed)\b", ERROR)])
+
+        expected_error_msg = "Found 2 potential error(s) in command output " \
```
boegel: rename to `expected_warning_msg`?

test/framework/run.py
```diff
@@ -520,6 +527,56 @@ def test_run_cmd_stream(self):
...
+        write_file(logfile, '')
+        init_logging(logfile, silent=True)
+        self.assertErrorRegex(EasyBuildError, expected_error_msg, check_log_for_errors, input_text, [
+            r"\berror\b",
+            (r"\ballowed-test failed\b", IGNORE),
+            (r"(?i)\bCRASHED\b", WARN),
+            "fail"
+        ])
+        stop_logging(logfile)
+        expected_error_msg = "Found 1 potential error(s) in command output (output: the process crashed with 0)"
```
boegel: rename to `expected_warning_msg`?

easybuild/tools/run.py
```diff
@@ -589,3 +589,61 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
...
+    :param log_txt: String containing the log, will be split into individual lines
+    :param reg_exps: List of: regular expressions (as strings) to error on,
+                    or tuple of regular expression and action (any of [IGNORE, WARN, ERROR])
+    """
+    global errors_found_in_log
+    warnings, errors = extract_errors_from_log(log_txt, reg_exps)
+
+    errors_found_in_log += len(warnings) + len(errors)
+    if warnings:
+        _log.warning("Found %s potential error(s) in command output (output: %s)",
```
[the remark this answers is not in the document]
Flamefire: It was meant as a replacement of the old method which does the same: Log the warnings/potential errors and show a summary "Found N potential errors, see log" at the end of the build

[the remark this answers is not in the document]
boegel: OK then :)

easybuild/tools/run.py
```diff
@@ -589,3 +589,61 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
...
+
+    re_tuples = []
+    for cur in reg_exps:
+        try:
+            if isinstance(cur, str):
+                reg_exp, action = cur, ERROR
+            else:
+                reg_exp, action = cur
+            if not isinstance(reg_exp, str) or action not in (IGNORE, WARN, ERROR):
+                raise TypeError("Invalid types")
```
boegel: This should provide more details: invalid types for what, which types were expected, which were obtained

Flamefire: This exception is caught 2 lines below and enhanced with more information including what you requested. But I enhanced it anyway a bit to distinguish the 2 failures

PR #3119, opened 2019-12-10
Use correct module for errors_found_in_log

Flamefire:
The (important) warning message was never printed due to a wrong module being used after a rename/move some time ago

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -61,7 +61,7 @@ from easybuild.framework.easyconfig.parser import fetch_parameters_from_easyconf
 from easybuild.framework.easyconfig.style import MAX_LINE_LENGTH
 from easybuild.framework.easyconfig.tools import get_paths_for
 from easybuild.framework.easyconfig.templates import TEMPLATE_NAMES_EASYBLOCK_RUN_STEP, template_constant_dict
-from easybuild.tools import config, filetools
+from easybuild.tools import config, run
 from easybuild.tools.build_details import get_build_stats
 from easybuild.tools.build_log import EasyBuildError, dry_run_msg, dry_run_warning, dry_run_set_dirs
 from easybuild.tools.build_log import print_error, print_msg, print_warning
@@ -3082,7 +3082,7 @@ def build_and_install_one(ecdict, init_env):
 
     # restore original environment, and then sanitize it
     _log.info("Resetting environment")
-    filetools.errors_found_in_log = 0
+    run.errors_found_in_log = 0
     restore_env(init_env)
     sanitize_env()
 
@@ -3233,9 +3233,9 @@ def build_and_install_one(ecdict, init_env):
     print_msg("%s: Installation %s %s (took %s)" % (summary, ended, succ, req_time), log=_log, silent=silent)
 
     # check for errors
-    if filetools.errors_found_in_log > 0:
+    if run.errors_found_in_log > 0:
         print_msg("WARNING: %d possible error(s) were detected in the "
-                  "build logs, please verify the build." % filetools.errors_found_in_log,
+                  "build logs, please verify the build." % run.errors_found_in_log,
                   _log, silent=silent)
 
     if app.postmsg:
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3234,9 +3234,8 @@ def build_and_install_one(ecdict, init_env):
 
     # check for errors
     if run.errors_found_in_log > 0:
-        print_msg("WARNING: %d possible error(s) were detected in the "
-                  "build logs, please verify the build." % run.errors_found_in_log,
-                  log=_log, silent=silent)
+        _log.warning("%d possible error(s) were detected in the "
+                     "build logs, please verify the build.", run.errors_found_in_log)
 
     if app.postmsg:
         print_msg("\nWARNING: %s\n" % app.postmsg, log=_log, silent=silent)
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyblock.py
```diff
@@ -62,7 +62,7 @@
 from easybuild.framework.easyconfig.style import MAX_LINE_LENGTH
 from easybuild.framework.easyconfig.tools import get_paths_for
 from easybuild.framework.easyconfig.templates import TEMPLATE_NAMES_EASYBLOCK_RUN_STEP, template_constant_dict
-from easybuild.tools import config, filetools
+from easybuild.tools import config, filetools, run
```
ocaisa: There's another line that needs to be changed at https://github.com/easybuilders/easybuild-framework/blob/9fd2b836ae9df5a5bff3f1bc926b115818b297fd/easybuild/framework/easyblock.py#L3047
and then you can remove `filetools` from this import.

Flamefire: Done

ocaisa:
> For me this looks good, I will take a look tomorrow to see how hard it would be to come up a test

Actually the test is there, this was really just incorrect import/usage.

Flamefire:
Hmm, haven't found one by grepping for `errors_found_in_log `

boegel:
Nice catch @Flamefire...

We indeed overlooked this `errors_found_in_log` global variable when moving `parse_log_for_error` from `easybuild.tools.filetools` to `easybuild.tools.run` a (very) long time ago (see #842).

I'm a bit reluctant to re-instate the warning like this though, since the pattern used by `parse_log_for_error` for error seems very generic to me now that I take a closer look at it...

It basically considers any occurrence of `error` or `failed` as a clear indication that something went wrong when running a command, while we know now that this often leads to many false positives.

This is from the CMake log file for example, after checking the output of `./configure ...`:

```
== 2019-11-03 23:36:42,996 run.py:589 INFO parse_log_for_error (some may be harmless) regExp (?<![(,-]|\w)(?:error|segmentation fault|failed)(?![(,-]|\.?\w) found:
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed                                                                                                                                                         -- Check size of __int64 - failed
-- Check size of unsigned __int64 - failed                                                                                                                                                                  -- Performing Test HAVE_WORKING_EXT2_IOC_GETFLAGS - Failed
-- Performing Test HAVE_STRUCT_VFSCONF - Failed                                                                                                                                                             -- Performing Test HAVE_STRUCT_XVFSCONF - Failed
-- Performing Test MAJOR_IN_MKDEV - Failed                                                                                                                                                                  -- Performing Test HAVE_LZMA_STREAM_ENCODER_MT - Failed
-- Performing Test HAVE_STRUCT_TM___TM_GMTOFF - Failed                                                                                                                                                      -- Performing Test HAVE_STRUCT_STATFS_F_NAMEMAX - Failed
-- Performing Test HAVE_STRUCT_STAT_ST_BIRTHTIME - Failed                                                                                                                                                   -- Performing Test HAVE_STRUCT_STAT_ST_BIRTHTIMESPEC_TV_NSEC - Failed
-- Performing Test HAVE_STRUCT_STAT_ST_MTIMESPEC_TV_NSEC - Failed                                                                                                                                           -- Performing Test HAVE_STRUCT_STAT_ST_MTIME_N - Failed
-- Performing Test HAVE_STRUCT_STAT_ST_UMTIME - Failed                                                                                                                                                      -- Performing Test HAVE_STRUCT_STAT_ST_MTIME_USEC - Failed
-- Performing Test HAVE_STRUCT_STAT_ST_FLAGS - Failed                                                                                                                                                       -- Performing Test HAVE_STRUCT_STATVFS_F_IOSIZE - Failed
```

So, I'm not sure that re-instating this without doing a better job at filtering out false positives (which can be hard to distinguish from actual problems in practice)...

Flamefire:
Found another bug in the code, so it seems the log message is not covered by any test.

> I'm a bit reluctant to re-instate the warning like this though, since the pattern used by parse_log_for_error for error seems very generic to me now that I take a closer look at it...
>
> It basically considers any occurrence of error or failed as a clear indication that something went wrong when running a command, while we know now that this often leads to many false positives.

Then my proposal would be:
- Merge this, it fixes a regression either way
- Sort out and merge https://github.com/easybuilders/easybuild-framework/pull/3118
- Use #3118 from `parse_log_for_error` in the default case to produce less false-positives. Eg. `-- Performing Test` can be added as an "IGNORE" match
- Use #3118 in configure EB to fix #157

Flamefire:
Together with #3118 I could use it on our system to install a big package chain and sort out the most common false positives. I expect most of them to be the same, so should be easy to reduce the amount of false-positives considerably

boegel:
Reducing the false positives is nice, but I still feel that producing a big fat warning for every occurrence of something that looks like an error but may not be is not a good idea.

In some cases you may not even be able to automatically distinguish false positives from actual errors...

How about changing the `print_warning` to `log.warning`, to just log in the log file rather than spit it out in the output of the `eb` command?

Flamefire:
Sure. This PR was just meant to fix a regression and not question the initial intention of that feature. I'm all for replacing this by something more advanced.

On the other hand: Providing this warnings to someone testing a EC as a "warning" could still be useful. Checking for false-positives can be done by the user and common false-positives can be suppressed by EB (see my other PR). I mean: In the end this is a warning that something might be wrong. We can't know if this is a false-positive or not. For the second case it would be wrong **not** to show it.

boegel:
I understand your point, but I think there will be a lot more false positives than genuine errors that go totally undetected through other means (like a failing sanity check).

We certainly have had errors that went undetected for a while, but I'd say they are rare, and so I'm reluctant to start raising concerns by highlighting false positives...

Flamefire:
Ok so _log.warning instead

PR #3120, opened 2019-12-11
ignore imports from vsc namespace made from pkgutil.py

boegel:
The fake `vsc` Python package that is put in place after `vsc-base` was ingested (see #2790) apparently gets tripped by `pkgutil` (part of Python standard library) in some situations (probably when a `vsc-base` is installed system-wide), see https://lists.ugent.be/wws/arc/easybuild/2019-12/msg00044.html.

There's no point in exiting with an error in that case, so let's ignore imports from `vsc` triggered by `pkgutil.py`...

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1902,6 +1902,7 @@ def install_fake_vsc():
     fake_vsc_path = os.path.join(tempfile.mkdtemp(prefix='fake_vsc_'))
 
     fake_vsc_init = '\n'.join([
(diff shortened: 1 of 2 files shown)


boegel:
@ocaisa `eb --hooks test.py` where `test.py` has `import vsc`. That should still trigger the problem (unless you use `pkgutil.py` as filename).

The enhanced test covers it quite well though.

ocaisa:
Ah sorry, I didn't see the test!

PR #3122, opened 2019-12-11
Add a check for files not existing in the verify_checksum function

i-mtz:
Before the patch I got this error for a patch file when its name was not correct in the easyconfig file:
```ERROR: Build of /scicore/home/scicore/easybuild/easyconfigs/PSIPRED/PSIPRED-4.02-foss-2018b-BLAST-2.2.26.eb failed (err: 'build failed (first 300 chars): Checksum verification for /export/soft/source/p/PSIPRED/PSIRED-scripts-path.patch using ed712bb2d536ebe4d4acf18a1c75ab253915dc8f77c1c12a37a5183fbfb6510f failed.')```

After the change, the error points to the filename and not the checksum:
```ERROR: Build of /scicore/home/scicore/easybuild/easyconfigs/PSIPRED/PSIPRED-4.02-foss-2018b-BLAST-2.2.26.eb failed (err: 'build failed (first 300 chars): File PSIRED-scripts-path.patch does not exist')```

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -771,6 +771,10 @@ def verify_checksum(path, checksums):
 
     filename = os.path.basename(path)
 
+    # if file does not exist (typo) report it as such
+    if not os.path.isfile(filename):
+        raise EasyBuildError("File %s does not exist", filename)
+
     # if no checksum is provided, pretend checksum to be valid, unless presence of checksums to verify is enforced
     if checksums is None:
         if build_option('enforce_checksums'):
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -772,8 +772,8 @@ def verify_checksum(path, checksums):
     filename = os.path.basename(path)
 
     # if file does not exist (typo) report it as such
-    if not os.path.isfile(filename):
-        raise EasyBuildError("File %s does not exist", filename)
+    if not os.path.isfile(path):
+        raise EasyBuildError("File %s does not exist", path)
 
     # if no checksum is provided, pretend checksum to be valid, unless presence of checksums to verify is enforced
     if checksums is None:

### CODE REVIEW

File: easybuild/tools/filetools.py [hunk 1]
```diff
@@ -771,6 +771,10 @@ def verify_checksum(path, checksums):
 
     filename = os.path.basename(path)
 
+    # if file does not exist (typo) report it as such
+    if not os.path.isfile(filename):
```
ocaisa: This should be `path` not `filename`

boegel:
@i-mtz Are you up for also covering this in the test we have for `verify_checksums`?

See `test_checksums` in `test/framework/filetools.py`.

Adding a couple of lines like this should be enough:

```python
nosuchfile = os.path.join(self.test_prefix, 'nosuchfile')
self.assertFalse(os.path.exists(nosuchfile))
error_pattern =  "File .*/nosuchfile does not exist"
self.assertErrorRegex(EasyBuildError, error_pattern, ft.verify_checksum, nosuchfile, 'abcdef')

Flamefire:
I think this is the wrong place. How come a non-existing file is passed to the checksum function? Shouldn't existance be checked earlier? In this function existence should be asserted/assumed.

For sources the existence is already verified by the download/fetch functions. So it can only(?) apply to patches. So rather check it there instead?

i-mtz:
@Flamefire I think you are right, this should go in ```framework/easyblock.py``` just after https://github.com/easybuilders/easybuild-framework/blob/364590cfb5be6cac959f024fd60f8ebb8fad7d92/easybuild/framework/easyblock.py#L547 ?

As otherwise it tries to get both the ```md5``` and ```sha256sum```.

Flamefire:
You you get a backtrace of your issue? Because file existence should be already checked. See e.g. https://github.com/easybuilders/easybuild-framework/blob/364590cfb5be6cac959f024fd60f8ebb8fad7d92/easybuild/framework/easyblock.py#L454

i-mtz:
So after digging further, I found that the file check works as expected. The filename typo in the easyconfig file had  ```PSIRED-scripts-path.patch``` as patch file name and my patch file was ```PSIPRED-scripts-path.patch```. However EasyBuild (I didn't know) matched a patch file inside our sources ```/export/soft/source/p/PSIPRED/PSIRED-scripts-path.patch``` first, and I had overlooked it while parsing the log.

If I use a patch filename in the easyconfig that does not match any file, EasyBuild reports that the file does not exist.

To sum it up: I didn't properly read the error log

PR #3125, opened 2019-12-12
only actually change permissions using os.chmod in adjust_permissions if the current permissions are not correct already

boegel:
Using `eb --skip` on an installation that was originally performed by someone else currently fails in the permissions step, even if both account are a member of the same group and `--group-writable-installdir` is used.

This change/optimization to `adjust_permissions` fixes that problem...

cc @verdurin, @migueldiascosta

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1130,10 +1130,21 @@ def convert_name(name, upper=False):
         return name
 
 
-def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs=False, recursive=True,
+def adjust_permissions(provided_path, permission_bits, add=True, onlyfiles=False, onlydirs=False, recursive=True,
                        group_id=None, relative=True, ignore_errors=False, skip_symlinks=None):
     """
-    Add or remove (if add is False) permissionBits from all files (if onlydirs is False)
+    Change permissions for specified path, using specified permission bits
+
+    :param add: add permissions relative to current permissions (only relevant if 'relative' is set to True)
+    :param onlyfiles: only change permissions on files (not directories)
+    :param onlydirs: only change permissions on directories (not files)
+    :param recursive: change permissions recursively (only makes sense if path is a directory)
+    :param group_id: also change group ownership to group with this group ID
+    :param relative: add/remove permissions relative to current permissions (if False, hard set specified permissions)
+    :param ignore_errors: ignore errors that occur when changing permissions
+                          (up to a maximum ratio specified by --max-fail-ratio-adjust-permissions configuration option)
+
+    Add or remove (if add is False) permission_bits from all files (if onlydirs is False)
     and directories (if onlyfiles is False) in path
     """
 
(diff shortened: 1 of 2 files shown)


easybuild/tools/filetools.py
```diff
@@ -1130,10 +1130,21 @@ def convert_name(name, upper=False):
...
+
+    :param add: add permissions relative to current permissions (only relevant if 'relative' is set to True)
+    :param onlyfiles: only change permissions on files (not directories)
+    :param onlydirs: only change permissions on directories (not files)
+    :param recursive: change permissions recursively (only makes sense if path is a directory)
+    :param group_id: also change group ownership to group with this group ID
+    :param relative: add/remove permissions relative to current permissions (if False, hard set specified permissions)
+    :param ignore_errors: ignore errors that occur when changing permissions
+                          (up to a maximum ratio specified by --max-fail-ratio-adjust-permissions configuration option)
+
```
akesandgren: Missing :param skip_symlinks

easybuild/tools/filetools.py
```diff
@@ -1170,34 +1181,47 @@ def adjust_permissions(name, permissionBits, add=True, onlyfiles=False, onlydirs
...
+                current_perms = os.lstat(path)[stat.ST_MODE]
+                _log.debug("Current permissions for %s: %s", path, oct(current_perms))
 
+                if relative:
                     # relative permissions (add or remove)
-                    perms = os.lstat(path)[stat.ST_MODE]
-
                     if add:
-                        os.chmod(path, perms | permissionBits)
+                        _log.debug("Adding permissions for %s: %s", path, oct(permission_bits))
```
akesandgren: I would move this below new_perms and print both the permission_bits and the final result.
Same for removing

easybuild/tools/filetools.py
boegel: That one is deprecated (and if you try using it in EasyBuild v4.x you'll got an error), so I excluded that one on purpose.

easybuild/tools/filetools.py
boegel: But then I need another `if add:` to discriminate below adding/removing of the `permission_bits`?

The log messages now mention everything: original permissions, added/removed permissions, changed permissions. For (fictional) example:

```
== 2019-12-12 21:51:39,936 filetools.py:1189 DEBUG Current permissions for /Users/kehoste/software/bzip2/1.0.6/bin/bzip2: 0100777
== 2019-12-12 21:51:39,936 filetools.py:1197 DEBUG Removing permissions for /Users/kehoste/software/bzip2/1.0.6/bin/bzip2: 022
== 2019-12-12 21:51:39,936 filetools.py:1207 DEBUG Changing permissions for /Users/kehoste/software/bzip2/1.0.6/bin/bzip2 to 0100755
```

akesandgren: Nah, I meant moving it one line down after the new_perms = ... line.
I didn't think of the log below with the actual chmod

akesandgren:
Going in, thanks @boegel!

PR #3126, opened 2019-12-12
increase default number of chars/bytes read in asyncprocess.Popen

boegel:
I was grinding my teeth on trying to figure out why the interactive `./configure` command that is run while installing `WRF` was failing when running EasyBuild on top of Python 3, and it turns out it boils down to a too small default number of bytes/characters that is read every time the output of the interactive command is read.

It seems that if the command produces a significant amount of output shortly after it was started, only the first chunk of the output (basically exactly the first 1024 characters) is picked up when running on top of Python 3. Subsequent polls then don't receive the rest of the output (which is where the actual bug is, probably).

It's not entirely clear yet why to me this happens, it doesn't occur at all when using Python 2 (in fact, there it seems that all output is received at the first poll, even though it's significantly more than 1204 characters, it's on the order of 3.5k).
I hope to dive in deeper some other time, but for now this fix will have to do.

The current default of only picking up max. 1024 bytes/characters seems rather low to me anyway, so we might as well increase it...

### DIFF
--- a/easybuild/tools/asyncprocess.py
+++ b/easybuild/tools/asyncprocess.py
@@ -89,7 +89,7 @@ class Popen(subprocess.Popen):
 
     def get_conn_maxsize(self, which, maxsize):
         if maxsize is None:
-            maxsize = 1024
+            maxsize = 10240
         elif maxsize < 1:
             maxsize = 1
         return getattr(self, which), maxsize


akesandgren:
Could explicitly setting bufsize=0 to subprocess.Popen to force it unbuffered help?

boegel:
@akesandgren Very good suggestion, it seems like the default changed from `bufsize=0` (no buffering) in Python 2 to `bufsize=-1` (use system default) in recent Python 3 versions, which sure sounds like the culprit for what I was observing...

akesandgren:
As I understood the documentation, it was -1 in python 2 too, changed to 0 in some v3.x and then they changed it back to -1 to match what codes expected from py2.

Exactly which py3 version where you seeing the problem in?

akesandgren:
never mind, they changed it back in 3.3.1

boegel:
It is definitely `0` by default in Python 2, see https://docs.python.org/2/library/subprocess.html#subprocess.Popen:

```
 The default value for bufsize is 0 (unbuffered).
```

Talk about subtle differences between Python 2 & 3...

boegel:
closing this, proper fix in https://github.com/easybuilders/easybuild-framework/pull/3130

PR #3127, opened 2019-12-13
use shutil.copyfile to just copy file contents if target path exists and is owned by someone else

boegel:
This fixes an issue we had on `generoso`, @verdurin can confirm the fix works (on top of the fix in #3125).

It's difficult to get this case covered in the tests though, because I can't think of a way to create a temporary file owned by another use where we have write permissions to...

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -138,6 +138,9 @@ jobs:
           eb --show-system-info
           # check GitHub configuration
           eb --check-github --github-user=easybuild_test
+          # create file owned by root but writable by anyone (used by test_copy_file)
+          sudo touch /tmp/file_to_overwrite_for_easybuild_test_copy_file.txt
+          sudo chmod o+w /tmp/file_to_overwrite_for_easybuild_test_copy_file.txt
           # run test suite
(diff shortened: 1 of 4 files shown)


boegel:
I think I know a way to have this covered by the tests after all, don't merge yet please...

boegel:
Test to verify the use of `shutil.copyfile` in `copy_file` when it's needed is added in commit 017f3a9.

Travis & GitHub should report a failure for commit 017f3a9 because `/tmp/file_to_overwrite_for_easybuild_test_copy_file.txt` is not created yet in the Travis or GitHub CI configurations, which will trigger printing of the message that signals that part of `test_copy_file` being skipped, which results in a failing test run because the tests are not allowed to produce unexpected output...

If `test_copy_file` is run without `/tmp/file_to_overwrite_for_easybuild_test_copy_file.txt` being there, it won't make the test suite fail (the test suite output is an additional separate check in the Travis & GitHub CI configs).

Next push will fix that by putting `/tmp/file_to_overwrite_for_easybuild_test_copy_file.txt` in place using `sudo` (so it's root-owned), but I'll let it fail first to show this approach works, for future reference...

PR #3132, opened 2019-12-14
implement support for 'eb --show-ec'

boegel:
Small but useful feature, I find myself using `eb --search` often just to obtain the location of an easyconfig file so I can inspect it's contents...


akesandgren:
Going in, thanks @boegel!

PR #3133, opened 2019-12-17
also update $XDG_DATA_DIR (share/) and $GI_TYPELIB_PATH environment variables (lib*/girepository-*)

Flamefire:
Those are required for GIR libs to find *.gir files (usually in /share, XDG_DATA_DIR used)
and *.typelib files (usually in /lib/girepository-1.0, GI_TYPELIB_PATH used)

Adresses https://github.com/easybuilders/easybuild-easyconfigs/issues/9520

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -44,7 +44,6 @@ import glob
 import inspect
 import os
 import re
-import shutil
 import stat
 import tempfile
 import time
@@ -71,7 +70,7 @@ from easybuild.tools.config import build_option, build_path, get_log_filename, g
 from easybuild.tools.config import install_path, log_path, package_path, source_paths
 from easybuild.tools.environment import restore_env, sanitize_env
 from easybuild.tools.filetools import CHECKSUM_TYPE_MD5, CHECKSUM_TYPE_SHA256
-from easybuild.tools.filetools import adjust_permissions, apply_patch, apply_regex_substitutions, back_up_file
+from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file
 from easybuild.tools.filetools import change_dir, convert_name, compute_checksum, copy_file, derive_alt_pypi_url
 from easybuild.tools.filetools import diff_files, download_file, encode_class_name, extract_file
 from easybuild.tools.filetools import find_backup_name_candidate, get_source_tarball_from_git, is_alt_pypi_url
@@ -1317,15 +1316,18 @@ class EasyBlock(object):
         """
         A dictionary of possible directories to look for.
         """
+        lib_paths = ['lib', 'lib32', 'lib64']
         return {
             'PATH': ['bin', 'sbin'],
-            'LD_LIBRARY_PATH': ['lib', 'lib32', 'lib64'],
-            'LIBRARY_PATH': ['lib', 'lib32', 'lib64'],
+            'LD_LIBRARY_PATH': lib_paths,
+            'LIBRARY_PATH': lib_paths,
             'CPATH': ['include'],
             'MANPATH': ['man', os.path.join('share', 'man')],
-            'PKG_CONFIG_PATH': [os.path.join(x, 'pkgconfig') for x in ['lib', 'lib32', 'lib64', 'share']],
+            'PKG_CONFIG_PATH': [os.path.join(x, 'pkgconfig') for x in lib_paths + ['share']],
             'ACLOCAL_PATH': [os.path.join('share', 'aclocal')],
             'CLASSPATH': ['*.jar'],
+            'XDG_DATA_DIRS': ['share'],
+            'GI_TYPELIB_PATH': [os.path.join(x, 'girepository-*') for x in lib_paths],
         }
 
     def load_module(self, mod_paths=None, purge=True, extra_modules=None):


easybuild/framework/easyblock.py
```diff
@@ -1317,15 +1316,18 @@ def make_module_req_guess(self):
...
+            'LD_LIBRARY_PATH': lib_paths,
+            'LIBRARY_PATH': lib_paths,
             'CPATH': ['include'],
             'MANPATH': ['man', os.path.join('share', 'man')],
-            'PKG_CONFIG_PATH': [os.path.join(x, 'pkgconfig') for x in ['lib', 'lib32', 'lib64', 'share']],
+            'PKG_CONFIG_PATH': [os.path.join(x, 'pkgconfig') for x in lib_paths + ['share']],
             'ACLOCAL_PATH': [os.path.join('share', 'aclocal')],
             'CLASSPATH': ['*.jar'],
+            'XDG_DATA_DIRS': ['share'],
+            'GI_TYPELIB_PATH': [os.path.join(x, 'girepository-*') for x in lib_paths],
```
akesandgren: I think we should not add to XDG_DATA_DIRS unconditionally. But only when there is a gir-* dir in share.

Flamefire: I didn't want to make that GIR specific because the docu is more general: https://specifications.freedesktop.org/basedir-spec/0.6/ar01s03.html

>  $XDG_DATA_DIRS defines the preference-ordered set of base directories to search for data files in addition to the $XDG_DATA_HOME base directory. The directories in $XDG_DATA_DIRS should be seperated with a colon ':'.
>
>If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used. 

This sounds like more than GIR  can benefit from having access to the share folders of installed modules.

If this is still unwanted: It will be more complex. I don't think one can change it at this point to add `share` if `share/gir-*` exists. Would need to be done at another place then.

akesandgren: True, it just felt a bit too much. But it is what the variable and dir is for so i'll withdraw that comment.

Flamefire: Fully understood, I felt the same. But I guess in the end this is better.

akesandgren: There is a fairly big risk that XDG_DATA_DIRS will end up being VERY long.
There are LOTS of packages that have a share dir.

Flamefire: It still sounds like this is exactly the use case for this var: To point to all non-default-location share folders. Will this long var be a problem? Should we be more cautious and implement e.g. my proposal above with `check-expr`?

akesandgren: In our current tree there are > 1000 packages with "share", (just counting one arch), though there are multiple versions of a lot of them.
And due to HMNS there probably wont be more than ~150 modules loaded at any point in time. 
Scipion is the one we have that has the biggest dep tree and that's > 100 modules if I remember correctly.

Flamefire: So this is OK?

akesandgren: Don't really know. It mainly depends on how it is used and if that has any significant performance impact.

I would guess that it won't be a real problem. But it is something we should be aware of and preferably come up with some performance test code that actually uses that var for something...

boegel: Same applies to `$PATH`, `$LD_LIBRARY_PATH`, etc., I don't think we've run into any serious issues there either because of a long list of entries there...

paulmelis: Just wanted to make a remark on this patch now that is has been applied in newer EB releases and we run into some side-effect of it. This might be relevant for other folks.

The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) mandates: 

> If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used. 

On many system `$XDG_DATA_DIRS` is empty (or not set) by default, which means the default directories `/usr/local/share/:/usr/share/` are assumed implicitly for searching data files. But when an EB-based module simply uses `prepend-path` to alter `$XDG_DATA_DIRS` those directories are no longer implicitly used, as only the value set by the module is present in `$XDG_DATA_DIRS`. So loading a module has the effect of the global data directories no longer being searched, which has all kinds of side-effects (I ran into this while installing a VNC environment, whose desktop startup failed in strange ways). 

I'm not sure this should (or can be easily) fixed on the EB side. One workaround is to set `$XDG_DATA_DIRS=`/usr/local/share/:/usr/share/` in the shell skeleton files, for example. But many other system scripts I found that alter `$XDG_DATA_DIRS` make sure to add the `/usr/local/share/:/usr/share/` entries when the current value for `$XDG_DATA_DIRS` is empty. 

All in all the way the value is mandated in the specification is a bit troubled, due these kinds of nasty side-effects.

akesandgren: It's not enforced, the text above uses "should" not "must". But yes, it does hint at problems...

paulmelis: Fine, it's not mandatory, but then there need to be good reasons to deviate. Plus the real world *does* assume the stated principle in many cases (either explicitly or implicitly).

boegel: Follow-up in #3368 on this issue...

PR #3141, opened 2019-12-21
get_software_libdir: take into account that lib64 could be a symlink to lib (or vice versa)

### LINKED ISSUE #3139: get_software_libdir fails for symlinked lib64->lib
An installation of SuperLU fails due to it calling `openblas_lib = os.path.join(get_software_root('OpenBLAS'), get_software_libdir('OpenBLAS'), "libopenblas.a")` with an implicit `only_one=True` parameter to `get_software_libdir` but that finds `lib` and `lib64` and fails with an error.

The problem is that (our?) installation of `OpenBLAS/0.2.20-GCC-6.4.0-2.28` contains a symlink `lib64 -> lib`.

There are multiple solutions to this:
1. Reinstall OpenBLAS w/o the symlink. I guess it was manually added because I can't see it in the reinstalled one. But maybe this makes something else fail
2. SuperLU should be a) using `get_software_libdir('OpenBLAS', fs=["libopenblas.a"]) and b) then allow multiple dirs to be returned and simply pick the first
3. `get_software_libdir` should not return/check paths which are the same real path (resolving symlinks). I'd prefer this as this is the more universal solution that avoids problems in other EBs

boegel:
fixes #3139 reported by @Flamefire

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1435,9 +1435,9 @@ def get_software_libdir(name, only_one=True, fs=None):
                 # take into account that lib64 could be a symlink to lib (or vice versa)
                 # see https://github.com/easybuilders/easybuild-framework/issues/3139
                 if any(os.path.samefile(lib_dir_path, os.path.join(root, x)) for x in res):
-                    continue
+                    _log.debug("%s is the same as one of the other paths, so skipping it", lib_dir_path)
 
-                if fs is None or any(os.path.exists(os.path.join(lib_dir_path, f)) for f in fs):
+                elif fs is None or any(os.path.exists(os.path.join(lib_dir_path, f)) for f in fs):
                     _log.debug("Retaining library subdir '%s' (found at %s)", lib_subdir, lib_dir_path)
                     res.append(lib_subdir)
 
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1434,16 +1434,11 @@ def get_software_libdir(name, only_one=True, fs=None):
             if os.path.exists(lib_dir_path):
                 # take into account that lib64 could be a symlink to lib (or vice versa)
                 # see https://github.com/easybuilders/easybuild-framework/issues/3139
-                new_lib_dir = True
-                for retained_subdir in res:
-                    retained_path = os.path.join(root, retained_subdir)
-                    if os.path.samefile(lib_dir_path, retained_path):
-                        _log.info("%s is identical to retained path %s, so not retaining it...")
-                        new_lib_dir = False
-                        break
+                if any(os.path.samefile(lib_dir_path, os.path.join(root, x)) for x in res):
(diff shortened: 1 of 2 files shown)


easybuild/tools/modules.py
```diff
@@ -1430,9 +1430,22 @@ def get_software_libdir(name, only_one=True, fs=None):
     res = []
     if root:
         for lib_subdir in lib_subdirs:
-            if os.path.exists(os.path.join(root, lib_subdir)):
-                if fs is None or any([os.path.exists(os.path.join(root, lib_subdir, f)) for f in fs]):
+            lib_dir_path = os.path.join(root, lib_subdir)
+            if os.path.exists(lib_dir_path):
+                # take into account that lib64 could be a symlink to lib (or vice versa)
+                # see https://github.com/easybuilders/easybuild-framework/issues/3139
+                new_lib_dir = True
```
Flamefire: IMO the verbosity/log is not required here. Then it can be simplified to `if any(os.path.samefile(lib_dir_path, os.path.join(root, subdir) for subdir in res): continue`

easybuild/tools/modules.py
```diff
@@ -1430,9 +1430,22 @@ def get_software_libdir(name, only_one=True, fs=None):
...
+                for retained_subdir in res:
+                    retained_path = os.path.join(root, retained_subdir)
+                    if os.path.samefile(lib_dir_path, retained_path):
+                        _log.info("%s is identical to retained path %s, so not retaining it...")
+                        new_lib_dir = False
+                        break
+
+                if new_lib_dir and (fs is None or any([os.path.exists(os.path.join(lib_dir_path, f)) for f in fs])):
```
Flamefire: As you touched this line: Can you remove the explicit list creation in `any`? It is not required, will save 2 chars and some performance (EB is already slow enough ;-) )

easybuild/tools/modules.py
boegel: The simplified version you propose is not good enough, because the `os.path.exists(lib_dir_path)` check if really needed (calling `os.path.samefile` with a non-existing path is trouble).

I'll change the log statements to `log.debug` though to make it less verbose.

boegel: Ah, wait, what you propose goes under the `if os.path.exists(lib_dir_path):`, so it does make sense, thanks.

easybuild/tools/modules.py
boegel: Good suggestion, but I'm a bit offended by the "slow enough" part ;)

Flamefire: Well it takes a couple seconds to run `eb --help`, how would you call it? ;)

Jokes aside: It might be worth to explore where and why that much time is spend and do some focused optimizations. Besides the startup latency I see the same output multiple times in the log. So I guess there is duplicate work done somewhere.

PR #3142, opened 2019-12-23
add support for --copy-ec 

boegel:


boegel:
@bartoldeman I often find myself first determining the location of an easyconfig file just to copy & tweak it manually, that would become a little bit easier using `eb --copy-ec` (which is sort of a companion for `eb --show-ec`, see #3132, and an attempt to gradually implement what I had in mind in #2665)

PR #3146, opened 2019-12-28
fix --list-software=detailed when using Python 3 by leveraging sort_looseversions function from py2vs3 module

### LINKED ISSUE #3112: --list-software --output-format=rst fails when using python 3
it fails at

https://github.com/easybuilders/easybuild-framework/blob/64f0b30d3e1b74e7bf6106e8778212c1595bed61/easybuild/tools/docs.py#L638-L645

with
```
File "/usr/lib64/python3.6/distutils/version.py", line 46, in __eq__
    c = self._cmp(other)
  File "/usr/lib64/python3.6/distutils/version.py", line 337, in _cmp
```

boegel:
fixes #3112


migueldiascosta:
Going in, thanks @boegel!

PR #3152, opened 2020-01-06
Only remove folders that contain no files recursively

Flamefire:
Otherwise 'include/libfoo/foo.h'  would be removed from `$CPATH`

Bug introduced in https://github.com/easybuilders/easybuild-framework/pull/3145

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -681,6 +681,14 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False, filen
     return var_defs, hits
 
 
+def is_empty_folder(folder_path):
+    """Return true if the given folder does not contain any files"""
+    for _root, _dirs, files in os.walk(folder_path):
+        if files:
+            return False
+    return True
+
+
 def find_eb_script(script_name):
     """Find EasyBuild script with given name (in easybuild/scripts subdirectory)."""
     filetools, eb_dir = __file__, None
@@ -746,7 +754,7 @@ def calc_block_checksum(path, algorithm):
     try:
         # in hashlib, blocksize is a class parameter
         blocksize = algorithm.blocksize * 262144  # 2^18
-    except AttributeError as err:
+    except AttributeError:
         blocksize = 16777216  # 2^24
     _log.debug("Using blocksize %s for calculating the checksum" % blocksize)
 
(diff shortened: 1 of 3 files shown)


test/framework/filetools.py
```diff
@@ -1724,6 +1724,32 @@ def test_search_file(self):
         for pattern in ['*foo', '(foo', ')foo', 'foo)', 'foo(']:
             self.assertErrorRegex(EasyBuildError, "Invalid search query", ft.search_file, [test_ecs], pattern)
 
+    def test_is_empty_folder(self):
+        tmpdir = tempfile.mkdtemp()
```
boegel: @Flamefire There's a temporary directory unique to every test already available in `self.test_prefix`, you can use that one as a sandbox (it's also cleaned up automatically).

Flamefire: Ok, thanks. Done and improved the readability of the test

easybuild/tools/filetools.py
```diff
@@ -681,6 +681,14 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False, filen
     return var_defs, hits
 
 
+def is_empty_folder(folder_path):
```
boegel: * I would prefer `directory` over `folder` (we don't use `folder` anywhere else)
* A directory containing an empty subdirectory is not really empty, so this should be renamed...
  How about `dir_contains_files` (and flipping the `True`/`False` too, of course)?

Flamefire: Renaming makes sense.

I was thinking about future extensions, e.g. adding parameters like `non_empty_if_contains_subdir` or so. But I didn't see it necessary, at least ATM. I still stand by the name: A directory containing only directories is still empty. But I see how your name avoids the ambiguity. Changing...

easybuild/tools/filetools.py
```diff
@@ -681,6 +681,11 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False, filen
     return var_defs, hits
 
 
+def dir_contains_files(folder_path):
+    """Return true if the given folder does not contain any files"""
```
boegel: @Flamefire Docstring needs fixing too

Flamefire: :+1:

PR #3153, opened 2020-01-07
take into account that a checksum value may be a tuple of valid checksum in EasyBlock.check_checksums

boegel:


easybuild/framework/easyblock.py
```diff
@@ -1857,8 +1857,15 @@ def check_checksums_for(self, ent, sub='', source_cnt=None):
             if isinstance(checksum, dict):
                 checksum = checksum.get(fn)
 
-            if not is_sha256_checksum(checksum):
-                msg = "Non-SHA256 checksum found for %s: %s" % (fn, checksum)
+            # take into account that we may encounter a tuuple of valid checksums
+            # (see https://github.com/easybuilders/easybuild-framework/pull/2958)
+            if isinstance(checksum, tuple):
```
Flamefire: What about `(algo, checksum)` tuples and `{file:checksums}` dicts?

easybuild/framework/easyblock.py
```diff
@@ -1857,8 +1857,15 @@ def check_checksums_for(self, ent, sub='', source_cnt=None):
             if isinstance(checksum, dict):
                 checksum = checksum.get(fn)
 
-            if not is_sha256_checksum(checksum):
-                msg = "Non-SHA256 checksum found for %s: %s" % (fn, checksum)
+            # take into account that we may encounter a tuuple of valid checksums
```
Flamefire: typo: `tuuple`

easybuild/framework/easyblock.py
boegel: Good point, fixed in ba413cfe3

easybuild/framework/easyblock.py
boegel: fixed in ba413cfe3, thanks

PR #3156, opened 2020-01-09
Adding CrayIntelMKL toolchain (without libsci)

gppezzi:
We have a few apps in production that perform better with MKL (wrt libsci), so I thought we could add this new toolchain to avoid unloading libsci in the easyconfigs where we need MKL (example [here](https://github.com/eth-cscs/production/blob/master/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.4.1-CrayIntel-19.10.eb)).

Here’s an example easyconfig for [CrayIntelMKL]( https://gist.github.com/gppezzi/92ca6b8d5e06c1a24d7c9150e31ccc96)

The code is 'inspired' in the original CrayIntel TC, I think the only reason we need a new TC is because the original CrayIntel checks if the module `cray-libsci` is loaded and it fails if it is not the case.


easybuild/toolchains/craymkl.py
```diff
@@ -0,0 +1,93 @@
...
+:author: Kenneth Hoste (Ghent University)
+:author: Guilherme Peretti-Pezzi (CSCS)
+"""
+import os
+
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.toolchain.linalg import LinAlg
+
+
+CRAY_MKL_MODULE_NAME = 'intel'
```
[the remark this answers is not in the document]
boegel: No idea what this is about, opened an issue: https://github.com/houndci/hound/issues/1769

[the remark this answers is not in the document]
ocaisa: Black would change this to `"intel"` is I think what is meant here.

easybuild/toolchains/craymkl.py
```diff
@@ -0,0 +1,93 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+Support for Cray/Intel MKL library, which provides BLAS/LAPACK support.
```
boegel: @gppezzi Looks like you mixed up the docstrings between both files?

easybuild/toolchains/craymkl.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class CrayMKL(LinAlg):
+    """Support for Cray/Intel MKL library, which provides BLAS/LAPACK support."""
+    # BLAS/LAPACK support
+    # via intel (MKL) module, which gets loaded via the PrgEnv module
+    BLAS_MODULE_NAME = [CRAY_MKL_MODULE_NAME]
+
+    # no need to specify libraries, compiler driver takes care of linking the right libraries
+    # FIXME: need to revisit this, on numpy we ended up with a serial BLAS through the wrapper.
```
boegel: is this `FIXME` still valid?

easybuild/toolchains/craymkl.py
```diff
@@ -0,0 +1,93 @@
...
+        pass
+
+    def _set_scalapack_variables(self):
+        """Skip setting ScaLAPACK related variables"""
+        pass
+
+    def definition(self):
+        """
+        Filter BLAS module from toolchain definition.
+        The cray-libsci module is loaded indirectly (and versionless) via the PrgEnv module,
```
boegel: not `cray-libsci`, but `intel`?

easybuild/toolchains/craymkl.py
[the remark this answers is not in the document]
boegel: Yeah, that could be it... 🤦‍♂ 

I'll look into disabling `flake8-black` though (as suggested in https://github.com/houndci/hound/issues/1769), since unless you're actually using the `black` code formatter (maybe we should, but it's a bit of effort to start using it & enforcing it), that's going to be a nightmare...

easybuild/toolchains/craymkl.py
gppezzi: It is much worse than that: I mixed up both files entirely. It should be fixed now.

easybuild/toolchains/craymkl.py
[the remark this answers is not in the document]
boegel: disabled via #3162

easybuild/toolchains/crayimkl.py
```diff
@@ -0,0 +1,41 @@
...
+:author: Kenneth Hoste (Ghent University)
+:author: Guilherme Peretti-Pezzi (CSCS)
+"""
+from easybuild.toolchains.compiler.craype import CrayPEIntel
+from easybuild.toolchains.linalg.craymkl import CrayMKL
+from easybuild.toolchains.mpi.craympich import CrayMPICH
+from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME
+
+
+class CrayIMKL(CrayPEIntel, CrayMPICH, CrayMKL):
```
boegel: @gppezzi Let's keep this consistent and name this `CrayIntelMKL`?

Changed in https://github.com/gppezzi/easybuild-framework/pull/2

boegel:
HPL in https://github.com/easybuilders/easybuild-easyconfigs/pull/9660 fails with current `CrayIntelMKL` toolchain definition, and @gppezzi also observed issues with `magma` that need to be figured out first...

migueldiascosta:
@gppezzi would it make sense to define a subtoolchain consisting of `CrayPEIntel` and `CrayMPICH` and then define `CrayIntel` and `CrayIntelMKL` by adding `LibSci` and `CrayMKL`, respectively, to it, so that dependencies that don't require math libraries can be shared?

akesandgren:
@migueldiascosta that does make sense, not sure how easy it is with the Cray toolchain since that one is a bit odd compared to the other ones.

Issue #3143, opened 2019-12-23
DEB package version doesn't follow dpkg policy


Mycron:
Hi!
I'm using EasyBuild v4.0.0 and creating a deb package to distribute it in such way.
I faced a problem while installing package with en error like this:

> debconf: delaying package configuration, since apt-utils is not installed
dpkg: error processing archive /tmp/apt-dpkg-install-RmYRki/0-m4-1.4.17_eb-4.0.0-48_amd64.deb (--unpack):
 parsing file '/var/lib/dpkg/tmp.ci/control' near line 2 package 'm4-1.4.17':
 error in 'Version' field string 'eb-4.0.0-48': version number does not start with digit

I found that by the default(using **EasyBuildPNS** class) value of a version field is composing from eb-**EASYBUILD_VERSION** which doesn't comply with [Debian policy manual](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-version) and **should start with a digit**

Mycron:
The same error happens using EasyBuild 4.1.0 version

boegel:
Thanks for the problem report @Mycron!

Fixing this is sort of painful now, since people may be expecting that we stick to the `eb-<easybuild_version>` format for the package version.

What would be an acceptable format? Would `<easybuild_version>-eb` work? Any other suggestions?

Mycron:
Hi, @boegel !
We fixed this issue in our project just creating separate package naming scheme class and point to it using cmd option --package-naming-scheme=<PNS_name>. In our project, we decided to use a form of <STACK_RELEASE>.<BUILD_DATE> and PACKAGE_RELEASE equals to build number(using Jenkins), where:
STACK_RELEASE - digit, like a major part of the commonly used version scheme. Because we are using a bunch of modules as one thing(we call it - software stack)
BUILD_DATE - just a date in form of YYYYMMDD

boegel:
@Mycron That makes a lot of sense...

So with that in mind, is there anything we should change in EasyBuild itself? Doesn't seem like there is?

Mycron:
Hi, @boegel !
I can make a proposal(changes, that can be useful) using my fork and create PR into the base branch.
Sorry for the delay, it was a Christmas holiday in Russia for a few days

boegel:
@Mycron Sounds good!

PR #3157, opened 2020-01-09
Fix deb package version field

Mycron:
issue: #3143 
The suggestion to resolve https://github.com/easybuilders/easybuild-framework/issues/3143

### DIFF
--- a/easybuild/tools/package/package_naming_scheme/easybuild_pns.py
+++ b/easybuild/tools/package/package_naming_scheme/easybuild_pns.py
@@ -30,8 +30,7 @@ Implementation of the EasyBuild packaging naming scheme
 """
 from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
 from easybuild.tools.package.package_naming_scheme.pns import PackageNamingScheme
-from easybuild.tools.version import VERSION as EASYBUILD_VERSION
-
+from datetime import datetime
 
 class EasyBuildPNS(PackageNamingScheme):
     """Class implmenting the default EasyBuild packaging naming scheme."""
@@ -42,12 +41,6 @@ class EasyBuildPNS(PackageNamingScheme):
         return '%s-%s' % (ec['name'], det_full_ec_version(ec))
 
     def version(self, ec):
-        """Determine package version: EasyBuild version used to build & install."""
-        ebver = str(EASYBUILD_VERSION)
-        if ebver.endswith('dev'):
-            # try and make sure that 'dev' EasyBuild version is not considered newer just because it's longer
-            # (e.g., 2.2.0 vs 2.2.0dev)
-            # cfr. http://rpm.org/ticket/56,
-            # https://debian-handbook.info/browse/stable/sect.manipulating-packages-with-dpkg.html (see box in 5.4.3)
-            ebver.replace('dev', '~dev')
-        return 'eb-%s' % ebver
+        """Determine package version: Date used while build & install."""
+        date_stamp = datetime.today().strftime('%Y%m%d')
+        return '%s' % date_stamp


easybuild/tools/package/package_naming_scheme/easybuild_pns.py
```diff
@@ -42,12 +41,6 @@ def name(self, ec):
...
-        ebver = str(EASYBUILD_VERSION)
-        if ebver.endswith('dev'):
-            # try and make sure that 'dev' EasyBuild version is not considered newer just because it's longer
-            # (e.g., 2.2.0 vs 2.2.0dev)
-            # cfr. http://rpm.org/ticket/56,
-            # https://debian-handbook.info/browse/stable/sect.manipulating-packages-with-dpkg.html (see box in 5.4.3)
-            ebver.replace('dev', '~dev')
-        return 'eb-%s' % ebver
+        """Determine package version: Date used while build & install."""
+        date_stamp = datetime.today().strftime('%Y%m%d')
```
ocaisa: @boegel Is this what we want? I imagine we might still like to see the EB version here, since that holds a bit more useful information than a timestamp?

As I understood it, the problem was the version starting with `eb-`, could we just make append `~eb` instead?

Mycron: Hi, @ocaisa ! It make sense.
I was just thinking about my version with datestamp to have a clear look when module was built. And IMHO, from user perspective, datestamp in a version looks more useful rather than just eb version

boegel: I agree with @ocaisa that it makes sense to keep tracking the EasyBuild version that was used to build the package.

I can see why a datestamp would be useful, but it's mostly meaningless (it doesn't give any information about what's inside the package, just when it was created).

Mycron: Ok. So I could revert some part of changes and come back to the version based on `ebver~eb`, where ebver determines as usual

easybuild/tools/package/package_naming_scheme/easybuild_pns.py
```diff
@@ -50,4 +49,4 @@ def version(self, ec):
             # cfr. http://rpm.org/ticket/56,
             # https://debian-handbook.info/browse/stable/sect.manipulating-packages-with-dpkg.html (see box in 5.4.3)
             ebver.replace('dev', '~dev')
```
ocaisa: I'm not expert here but shouldn't we somehow concat this with the change below so the function returns either `4.x.x~eb` or `4.x.x~ebdev`...or is it ok to have 2 `~`?

Mycron: Good information to clarify and analyze it may be here: https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
As I understand, having two tildas is allowed. And in the case of having `4.x.x~eb` and `4.x.x~dev~eb`, the first one seems to be the latest

Mycron: To be honest, I didn't check such form for the case of rpm packages

ocaisa: For me then this seems ok, but maybe @boegel should give the final ok

ocaisa:
@Mycron You will need to update the tests in `test/framework/package.py` to reflect the updated naming

easybuild/tools/package/package_naming_scheme/easybuild_pns.py
```diff
@@ -50,4 +49,4 @@ def version(self, ec):
...
-        return 'eb-%s' % ebver
+        return '%s~eb' % ebver
```
akesandgren, line 52: What you should do is add a timestamp after the ~eb, as in ~eb-yyyymmdd to make it possible to update the eb built package automatically. Using the ```date_stamp = datetime.today().strftime('%Y%m%d')``` code from the first commit.

akesandgren (changes requested):

PR #3160, opened 2020-01-10
support use of %(startdir)s template in easyconfig files (WIP)

### LINKED ISSUE #3158: Make start_dir available for steps
The config parameter start_dir is useful for finding the path to extracted contents. Example: In https://github.com/easybuilders/easybuild-easyconfigs/pull/9604/ `%(builddir)s/KjellKod-g2log-d9a55a4a6154` could be simplified to `%(start_dir)s` if that was supported.

Maybe something better named would be good too but having the path to the extracted contents available as a template param would already help

boegel:
intends to fix #3158, but not finished (hence WIP)

TODO:
* better test? (where `%(startdir)s` is actually used in an easyconfig file)
* also support use of `startdir` easyconfig parameter (equivalent to `start_dir`)
* deprecate `start_dir` easyconfig parameter

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1490,6 +1490,11 @@ class EasyBlock(object):
     #
 
     @property
+    def startdir(self):
+        """Alias for self.start_dir."""
+        return self.start_dir
+
+    @property
     def start_dir(self):
         """Start directory in build directory"""
         return self.cfg['start_dir']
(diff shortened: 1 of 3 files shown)


Flamefire:
Small ping here. The change looks good as is to me and provides what is needed. 

The last 2 points are renaming start_dir to startdir in ECs, right? I'd suggest to just open an issue for that and leave that for later as it is'nt required for this, is it?

akesandgren:
@boegel should we move forward with this and handle startdir/start_dir later?

smoors:
the logs for the failing checks have expired..
closing/reopening to get them back

akesandgren:
@boegel conflict resolution needed

PR #3162, opened 2020-01-12
ignore "Black would make changes" produced by flake8-black

boegel:
see also https://github.com/houndci/hound/issues/1769 and #3156 where the confusing `"Black would make changes"` comments produced by @houndci-bot first popped up...


akesandgren:
Going in, thanks @boegel!

PR #3168, opened 2020-01-20
also mention working directory + input passed via stdin (if any) in trace output of run_cmd

boegel:


boegel:
close + re-open to re-trigger Hound, no idea why it's failing...

PR #3174, opened 2020-01-23
Add probe of PREFIX and VERSION variable in external modules

victorusu:
This commit handles metadata for external module dependencies when there is no entry in the metadata file

It should look for the pair of variables definitions in the modules in the given order
  1. CRAY_XXXX_PREFIX and CRAY_XXXX_VERSION
  2. CRAY_XXXX_DIR and CRAY_XXXX_VERSION
  2. CRAY_XXXX_ROOT and CRAY_XXXX_VERSION
  5. XXXX_PREFIX and XXXX_VERSION
  4. XXXX_DIR and XXXX_VERSION
  5. XXXX_ROOT and XXXX_VERSION
  3. XXXX_HOME and XXXX_VERSION

If no pair is found, an empty dependency dictionary is returned, as is the default behavior.

This avoids the need for a very large metadata file.


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1152,6 +1153,52 @@ def _validate(self, attr, values):  # private method
...
+            ('%s_DIR' % short_ext_modname_upper, '%s_VERSION' % short_ext_modname_upper),
+            ('%s_ROOT' % short_ext_modname_upper, '%s_VERSION' % short_ext_modname_upper),
+            ('%s_HOME' % short_ext_modname_upper, '%s_VERSION' % short_ext_modname_upper),
+        ]
+
+        for p, v in allowed_pairs:
+            prefix = self.modules_tool.get_variable_from_modulefile(dep_name, p)
+            version = self.modules_tool.get_variable_from_modulefile(dep_name, v)
+
+            if prefix and version:
```
migueldiascosta: if prefix is found but not version, and if the version is in `dep_name`, can't we simply use that one?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1185,15 +1192,15 @@ def handle_external_module_metadata_by_probing_modules(self, dep_name):
...
+            module_prefix = self.modules_tool.get_variable_from_modulefile(dep_name, prefix)
+            module_version = self.modules_tool.get_variable_from_modulefile(dep_name, version)
 
-            if prefix and version:
+            if module_prefix and module_version:
                 dependency = {
-                    'name': [short_ext_modname],
-                    'version': [version],
-                    'prefix': p
+                    'name': [short_ext_modname_upper],
```
migueldiascosta: `short_ext_modname` (`= dep_name.split('/')[0]`) seemed to be ok as `dependency['name']`, why the change to `short_ext_modname_upper` (which is used in environment variables)?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1163,7 +1210,9 @@ def handle_external_module_metadata(self, dep_name):
             self.log.info("Updated dependency info with available metadata for external module %s: %s",
                           dep_name, dependency['external_module_metadata'])
         else:
-            self.log.info("No metadata available for external module %s", dep_name)
+            self.log.info("No metadata available for external module %s. Attempting to read from available modules",
+                dep_name)
+            dependency['external_module_metadata'] = self.handle_external_module_metadata_by_probing_modules(dep_name)
```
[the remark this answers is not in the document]
migueldiascosta: can you add this and look at the output when running `eb` without the metadata file, to get a better picture of what metadata is being picked up and for which external modules the probe is failing? When I do this in the test system I have access to, there are still quite a few...

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1170,38 +1170,38 @@ def handle_external_module_metadata_by_probing_modules(self, dep_name):
...
-        dependency = {}
-
         short_ext_modname = dep_name.split('/')[0]
-
-        if short_ext_modname.startswith('craype-'):
-            short_ext_modname = short_ext_modname.split('craype-')[1]
-        elif short_ext_modname.startswith('cray-'):
-            short_ext_modname = short_ext_modname.split('cray-')[1]
-
-        short_ext_modname.replace('-', '_')
```
migueldiascosta: this block is still needed, just before `convert_name` below

easybuild/framework/easyconfig/easyconfig.py
victorusu: @migueldiascosta, what I wrote is wrong...
Actually, there is no way around it... The name maps the external module name to EasyBuild's internal one. E.g., if the module is called `cudatoolkit`, then the user has to tell in the metadafile that this represents `CUDA`. So, I changed the function behavior in order to only fill in the missing terms...
This way, the user can set in the metadata file
```
[cudatoolkit]
name = CUDA
```
Then, the code will load the default `cudatoolkit` module and attempt to capture the version and prefix.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1152,19 +1153,109 @@ def _validate(self, attr, values):  # private method
...
+            self.log.info("No metadata available for external module %s. Attempting to read from available modules",
+                          dep_name)
+            external_metadata = self._handle_ext_module_metadata_by_probing_modules(dep_name)
+            if external_metadata:
+                self.log.info("Updated dependency info with metadata from available modules for external module %s: %s",
+                              dep_name, external_metadata)
+            else:
+                self.log.info("No metadata available for external module %s.", dep_name)
 
+        dependency['external_module_metadata'] = external_metadata
```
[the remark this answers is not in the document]
victorusu: That was not the case before. I will fix it!
Thank @boegel

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1152,18 +1153,110 @@ def _validate(self, attr, values):  # private method
...
+          3. XXXX_HOME and XXXX_VERSION
+
+        If neither of the pairs is found, then an empty dictionary is returned
+        """
+        if dependency is None:
+            dependency = dict()
+
+        short_ext_modname = dep_name.split('/')[0]
+        if not 'name' in dependency:
+            dependency['name'] = [short_ext_modname]
```
migueldiascosta: the test is failing because this is added regardless of finding the rest of the metadata, if this is moved inside the `if dep_prefix and dep_version:` block then it doesn't create a problem

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1152,18 +1153,110 @@ def _validate(self, attr, values):  # private method
...
-            self.log.info("No metadata available for external module %s", dep_name)
+            self.log.info("No metadata available for external module %s. Attempting to read from available modules",
+                          dep_name)
+            external_metadata = self._handle_ext_module_metadata_by_probing_modules(dep_name)
+            if external_metadata:
+                dependency['external_module_metadata'] = external_metadata
+                self.log.info("Updated dependency info with metadata from available modules for external module %s: %s",
+                              dep_name, external_metadata)
+            else:
+                self.log.info("No metadata available for external module %s.", dep_name)
```
migueldiascosta: what about simplifying a bit to something like

```python
        if dep_name in self.external_modules_metadata:
            external_metadata = self.external_modules_metadata[dep_name]
        elif dep_name_no_version in self.external_modules_metadata:
            external_metadata = self.external_modules_metadata[dep_name_no_version]

        if external_metadata:
            if all(d in external_metadata for d in metadata_fields):
                dependency['external_module_metadata'] = external_metadata
            else:
                self.log.info("Not all metadata available for external module %s. Attempting to read from available modules",
                              dep_name)
                external_metadata = self._handle_ext_module_metadata_by_probing_modules(dep_name,
                                                                                        dependency=external_metadata)
                if external_metadata:
                    dependency['external_module_metadata'] = external_metadata
                    self.log.info("Updated dependency info with metadata from available modules for external module %s: %s",
                                  dep_name, external_metadata)
                else:
                    self.log.info("No metadata available for external module %s.", dep_name)
        else:
            self.log.info("No metadata available for external module %s. Attempting to read from available modules",
                          dep_name)
            external_metadata = self._handle_ext_module_metadata_by_probing_modules(dep_name)
            if external_metadata:
                dependency['external_module_metadata'] = external_metadata
                self.log.info("Updated dependency info with metadata from available modules for external module %s: %s",
                              dep_name, external_metadata)
            else:
                self.log.info("No metadata available for external module %s.", dep_name)
```

boegel: Fully agreed, it's way too copy-pasty right now.
The `if` and `elif` are exactly the same, except for `dep_name` vs `dep_name_no_version`...

We can do better, I'm looking into it.

easybuild/framework/easyconfig/easyconfig.py
boegel: fixed, thanks for pointing this out @migueldiascosta!

easybuild/framework/easyconfig/easyconfig.py
boegel: The point is to only trust the metadata obtained by probing the module if it provides both the installation prefix and software version.
If it only provides one of both, we could have a "lucky hit", and the obtained metadata may be less trustworthy...

Let's go forward as is with this strict requirement, which can easily be loosened up later (the other way around is more difficult).

The other scenario, where the prefix is already known from the metadata file, but (only) the version is available through the module, may actually be more likely (since the prefix can be specified as the name of the environment variable specifying the installation prefix, not an actual hardcoded prefix, which can be done generically in the metadata file).

PR #3177, opened 2020-01-27
advise PR labels in --review-pr and add support for --add-pr-labels

migueldiascosta:


easybuild/framework/easyconfig/tools.py
```diff
@@ -538,6 +539,16 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
...
             lines.extend(['', "(no related easyconfigs found for %s)\n" % os.path.basename(ec['spec'])])
 
+    labels = []
+    file_info = det_file_info(pr_files, download_repo_path)
+    if any(file_info['new_folder']):
+        labels.append('new')
+    if any(file_info['new_file_in_existing_folder']):
+        labels.append('update')
+
+    if labels:  # TODO: compare to actual PR labels?
```
boegel: @migueldiascosta Makes sense to fix this first?

migueldiascosta: Done... now thinking that since this is likely being run by a maintainer, in that case it could actually set the labels when they are missing...

easybuild/tools/github.py
```diff
@@ -1271,6 +1271,41 @@ def merge_url(gh):
         print_warning("Review indicates this PR should not be merged (use -f/--force to do so anyway)")
 
 
+@only_if_module_is_available('git', pkgname='GitPython')
```
akesandgren: How does this affect calls to post_pr_labels?

easybuild/tools/github.py
```diff
@@ -1271,6 +1271,41 @@ def merge_url(gh):
...
+    pr_target_repo = build_option('pr_target_repo')
+
+    # fetch GitHub token (required to perform actions on GitHub)
+    github_user = build_option('github_user')
+    if github_user is None:
+        raise EasyBuildError("GitHub user must be specified to open a pull request")
+
+    github_token = fetch_github_token(github_user)
+    if github_token is None:
+        raise EasyBuildError("GitHub token for user '%s' must be available to open a pull request", github_user)
```
akesandgren: Is an error really the right thing here?
Anyone can do review-pr, they don't need an active token for that.
If it's a review only then it should just return False and not error out.

And we're not necessarily opening a pull request, we're might be reviewing one...

migueldiascosta:
@akesandgren thanks, too much copy-pasting...

I'd like to know if you (or any other maintainer) might think it's not adequate for `--review-pr` to post labels - for instance, I can imagine arguing that it's expected to be a read-only operation...

easybuild/framework/easyconfig/tools.py
```diff
@@ -538,6 +539,28 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
         else:
             lines.extend(['', "(no related easyconfigs found for %s)\n" % os.path.basename(ec['spec'])])
 
+    labels = []
+    file_info = det_file_info(pr_files, download_repo_path)
+    if any(file_info['new_folder']):
+        labels.append('new')
+    if any(file_info['new_file_in_existing_folder']):
+        labels.append('update')
```
boegel, line 547: @migueldiascosta This is copy-pasted from `github.py`, can we flesh this out into a small separate function that takes `file_info` as input?

Perhaps we later enhance the logic here somehow to add labels more intelligently (or use other labels).

easybuild/framework/easyconfig/tools.py
```diff
@@ -538,6 +539,28 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
...
+    pr_data, _ = fetch_pr_data(pr, github_account, github_repo, github_user)
+    pr_labels = [label['name'] for label in pr_data['labels']]
+
+    missing_labels = []
+    for label in labels:
+        if label not in pr_labels:
+            missing_labels.append(label)
+
+    if missing_labels:
+        if not post_pr_labels(pr, labels):
```
boegel, line 561: I wouldn't let `--review-pr` add labels automagically, I feel this should remain a read-only option, as you mentioned.

But we can add support for a separate `--add-pr-labels` option?

easybuild/framework/easyconfig/tools.py
migueldiascosta: done

easybuild/framework/easyconfig/tools.py
migueldiascosta: reverted adding labels. It's probably best to leave `--add-pr-labels` to a separate PR then

easybuild/framework/easyconfig/tools.py
```diff
@@ -538,6 +544,23 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
...
             lines.extend(['', "(no related easyconfigs found for %s)\n" % os.path.basename(ec['spec'])])
 
+    file_info = det_file_info(pr_files, download_repo_path)
+
+    pr_target_account = build_option('pr_target_account')
+    github_user = build_option('github_user')
+    pr_data, _ = fetch_pr_data(pr, pr_target_account, pr_target_repo, github_user)
+    pr_labels = [label['name'] for label in pr_data['labels']]
+
+    labels = det_labels(file_info, pr_target_repo)
```
boegel, line 554: rename to `expected_labels`?

easybuild/framework/easyconfig/tools.py
```diff
@@ -538,6 +544,23 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
...
+
+    missing_labels = []
+    for label in labels:
+        if label not in pr_labels:
+            missing_labels.append(label)
+
+    if missing_labels:
+        lines.extend(['', "This PR should be labeled %s" % ', '.join(missing_labels)])
```
boegel, line 562: Add quotes around names of missing labels?

```python
if missing_labels:
    lines.extend(['', "This PR should be labeled with %s" % ', '.join(["'%s'" % l for l in missing_labels])])

File: easybuild/tools/github.py [hunk 8]
```diff
@@ -1313,6 +1313,59 @@ def merge_url(gh):
         print_warning("Review indicates this PR should not be merged (use -f/--force to do so anyway)")
 
 
+def det_labels(file_info, pr_target_repo):
```
boegel, line 1316: Maybe rename this to `det_pr_labels` (to be in sync with `post_pr_labels`)?

easybuild/tools/github.py
```diff
@@ -1313,6 +1313,59 @@ def merge_url(gh):
...
+    """
+    Determine labels from file_info.
```
boegel, line 1318: `Determine labels for a pull request based on provided information on files changed by that pull request.`

test/framework/github.py
```diff
@@ -566,6 +566,38 @@ def run_check(expected_result=False):
         expected_warning = ''
         self.assertEqual(run_check(True), '')
 
+    def test_det_labels(self):
+        """Test for det_labels function."""
```
boegel, line 570: also rename to `det_pr_labels` here

test/framework/github.py
```diff
@@ -566,6 +566,38 @@ def run_check(expected_result=False):
...
+
+        file_info = {'new_folder': [False], 'new_file_in_existing_folder': [True]}
+        self.mock_stdout(True)
+        res = gh.det_labels(file_info, GITHUB_EASYCONFIGS_REPO)
+        self.mock_stdout(False)
```
boegel, line 575: No need for `mock_stdout` at all, since `det_pr_labels` doesn't produce any output at all?

test/framework/github.py
```diff
@@ -566,6 +566,38 @@ def run_check(expected_result=False):
...
+        self.assertEqual(len(res), 1)
+        self.assertEqual(res[0], 'update')
```
boegel, line 577: shorter (one check rather than two)

```python
self.assertEqual(res, ['update'])

File: test/framework/github.py [hunk 13]
```diff
@@ -566,6 +566,38 @@ def run_check(expected_result=False):
...
+        self.assertEqual(len(res), 1)
+        self.assertEqual(res[0], 'new')
+
+        file_info = {'new_folder': [True, False], 'new_file_in_existing_folder': [False, True]}
+        self.mock_stdout(True)
+        res = gh.det_labels(file_info, GITHUB_EASYCONFIGS_REPO)
+        self.mock_stdout(False)
+        self.assertEqual(len(res), 2)
+        self.assertTrue('new' in res)
+        self.assertTrue('update' in res)
```
boegel, line 592: 1 check:

```python
self.assertEqual(sorted(res), ['new', 'update'])

File: test/framework/options.py [hunk 14]
```diff
@@ -2942,6 +2942,21 @@ def test_review_pr(self):
         regex = re.compile(r"^Comparing gzip-1.10-\S* with gzip-1.10-")
         self.assertTrue(regex.search(txt), "Pattern '%s' not found in: %s" % (regex.pattern, txt))
 
+        self.mock_stdout(True)
+        self.mock_stderr(True)
+        # closed, unlabeled PR for gzip 1.2.8 easyconfig,
+        # see https://github.com/easybuilders/easybuild-easyconfigs/pull/5365
```
boegel, line 2948: Hmm, not sure if this is a good idea, since if some maintainer thinks "hey, this PR never got properly labeled, let me fix that", this test will be broken...

Can we do better, by somehow ignoring the labels that were set, to ensure the test will never fail?

Maybe even go as far as supporting an `--ignore-pr-labels` option?

migueldiascosta: is there another use case for `--ignore-pr-labels`?

what about something like

```diff
diff --git a/easybuild/framework/easyconfig/tools.py b/easybuild/framework/easyconfig/tools.py
index 2df6450..0ae7ea0 100644
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -506,7 +506,7 @@ def find_related_easyconfigs(path, ec):
     return sorted(res)
 
 
-def review_pr(paths=None, pr=None, colored=True, branch='develop'):
+def review_pr(paths=None, pr=None, colored=True, branch='develop', testing=False):
     """
     Print multi-diff overview between specified easyconfigs or PR and specified branch.
     :param pr: pull request number in easybuild-easyconfigs repo to review
@@ -549,7 +549,7 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
     pr_target_account = build_option('pr_target_account')
     github_user = build_option('github_user')
     pr_data, _ = fetch_pr_data(pr, pr_target_account, pr_target_repo, github_user)
-    pr_labels = [label['name'] for label in pr_data['labels']]
+    pr_labels = [label['name'] for label in pr_data['labels']] if not testing else []
 
     expected_labels = det_pr_labels(file_info, pr_target_repo)
 
diff --git a/easybuild/main.py b/easybuild/main.py
index 415321d..8bd9ed5 100644
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -247,7 +247,7 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         merge_pr(options.merge_pr)
 
     elif options.review_pr:
-        print(review_pr(pr=options.review_pr, colored=use_color(options.color)))
+        print(review_pr(pr=options.review_pr, colored=use_color(options.color), testing=testing))
 
     elif options.list_installed_software:
         detailed = options.list_installed_software == 'detailed'
diff --git a/test/framework/options.py b/test/framework/options.py
index 0230bf5..c76841c 100644
--- a/test/framework/options.py
+++ b/test/framework/options.py
@@ -2944,14 +2944,14 @@ class CommandLineOptionsTest(EnhancedTestCase):
 
         self.mock_stdout(True)
         self.mock_stderr(True)
-        # closed, unlabeled PR for gzip 1.2.8 easyconfig,
+        # closed PR for gzip 1.2.8 easyconfig,
         # see https://github.com/easybuilders/easybuild-easyconfigs/pull/5365
         args = [
             '--color=never',
             '--github-user=%s' % GITHUB_TEST_ACCOUNT,
             '--review-pr=5365',
         ]
-        self.eb_main(args, raise_error=True)
+        self.eb_main(args, raise_error=True, testing=True)
         txt = self.get_stdout()
         self.mock_stdout(False)
         self.mock_stderr(False)
```

?

boegel: @migueldiascosta That's acceptable imho, just add a comment in `review_pr` to explain why the `testing` named argument is there, and where it is used.

boegel:
@migueldiascosta Just in case you missed my suggestion, adding support for `--add-pr-labels` should be fairly trivial now, no?

easybuild/framework/easyconfig/tools.py
```diff
@@ -538,6 +545,24 @@ def review_pr(paths=None, pr=None, colored=True, branch='develop'):
...
+        github_user = build_option('github_user')
+        pr_data, _ = fetch_pr_data(pr, pr_target_account, pr_target_repo, github_user)
+        pr_labels = [label['name'] for label in pr_data['labels']] if not testing else []
+
+        expected_labels = det_pr_labels(file_info, pr_target_repo)
+
+        missing_labels = []
+        for label in expected_labels:
+            if label not in pr_labels:
+                missing_labels.append(label)
```
boegel, line 561: This block is identical to a large chunk of `add_pr_labels`, so we should flesh that out to a separate function `det_missing_labels`?

migueldiascosta:
some of the tests are fragile in that they will fail if the labels of the (closed) PRs used ever change...

boegel:
@migueldiascosta Should we open PRs in https://github.com/easybuilders/testrepository to prevent us some breaking the tests?
We could mention in the PR title to leave the labels and add a reference to the test that relies on it?

migueldiascosta:
@boegel `add_pr_labels` uses `fetch_easyconfigs_from_pr` which expects an `easybuild-easyconfigs` repo

We could use any fork of `easybuild-easyconfigs`, e.g. `boegelbot`'s, `ebtutorial`'s or a new test account

boegel:
@migueldiascosta We can open a PR in @boegelbot's fork for this, sure. Can you look into this?

boegel:
@migueldiascosta This now also needs a sync with `develop` to fix the merge conflicts...

migueldiascosta:
`easystack_basic` test is failing with `unbound method error() must be called with EasyStackParser instance as first argument (got str instance instead)`, not sure how this is related to this PR...

boegel:
@migueldiascosta I ran into that while working on #3479, but that should be fixed in `develop`...

I don't see that error popping up in the CI logs though, are you only seeing that when running that test locally?

migueldiascosta:
> I don't see that error popping up in the CI logs though, are you only seeing that when running that test locally?

it is popping up in the CI tests, e.g. https://github.com/easybuilders/easybuild-framework/pull/3177/checks?check_run_id=1477507461

migueldiascosta:
> I don't see that error popping up in the CI logs though, are you only seeing that when running that test locally?

sorry, yes, the actual error message I printed locally, it should show up in the CI now that your #3511 is merged (both to develop and to this branch)

migueldiascosta:
except now it doesn't fail in the CI... 

fwiw, that test still fails for me locally (with python 2.7; it passes with python 3.6)

boegel:
@migueldiascosta Can you open an issue with more info on how the test fails exactly with current `develop`?

boegel (approved):

PR #3183, opened 2020-01-31
Fix shebang even if first line doesn't start with '#!'

### LINKED ISSUE #3179: Allow injecting shebang (now it only replaces - eg. touching files starting with '#!')
The shebang regex is here:
https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/framework/easyblock.py#L2189

Darkless012:
fixes #3179

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2209,6 +2209,9 @@ class EasyBlock(object):
                         if should_patch:
                             contents = shebang_regex.sub(shebang, contents)
                             write_file(path, contents)
+                        else:
+                            contents = shebang + "\n" + contents
+                            write_file(path, contents)
 
     def post_install_step(self):
         """
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2209,7 +2209,9 @@ class EasyBlock(object):
                         if should_patch:
                             contents = shebang_regex.sub(shebang, contents)
                             write_file(path, contents)
-                        else:
+                        elif not contents.startswith('#!'):
+                            self.log.info("The file '%s' doesn't have any shebang present, inserting it as first line.",
+                                          path)
                             contents = shebang + "\n" + contents
                             write_file(path, contents)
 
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2209,6 +2209,9 @@ def fix_shebang(self):
                         if should_patch:
                             contents = shebang_regex.sub(shebang, contents)
                             write_file(path, contents)
+                        else:
```
boegel: @Darkless012 This is not fully correct...

Imagine I have a bash script called `bin/foo` that has `#!/bin/bash`, and `fix_python_shebang_for = ['bin/*']` is used.

You don't want to inject `#!/usr/bin/env python` in that case and break the script.

We should make sure there's *no* shebang at all before adding it. Maybe like this:

```python
elif not contents.startswith('#!/'):

File: test/framework/toy_build.py [hunk 2]
```diff
@@ -2390,6 +2392,8 @@ def test_fix_shebang(self):
             "   'echo \"#!/usr/bin/perl -w\\n# test\" > %(installdir)s/bin/t4.pl',",
             # space after #! + 'env perl5'
             "   'echo \"#!/usr/bin/env perl5\\n# test\" > %(installdir)s/bin/t5.pl',",
+            # no shebang perl
+            "   'echo \"# test\" > %(installdir)s/bin/t6.pl',",
```
boegel: Also add a test which has `#!/bin/bash` as shebang, to make sure you don't break scripts that do have a shebang?

easybuild/framework/easyblock.py
akesandgren: Does the code path also verify that it is a script in the first place?? And not a random binary?

Darkless012: @boegel Yup, I've also thought about having wrong files "fixed". Also, there is optional space possible after `#!` I will add that to tests as well.

Darkless012: @akesandgren Yup, the files are read and checked for shebang regex or `#!` string. So binary files will not be touched. https://github.com/easybuilders/easybuild-framework/blob/1d112fb3aac5c62b58e80dc5bd15b2ebb1188513/easybuild/framework/easyblock.py#L2200-L2205

test/framework/toy_build.py
Darkless012: Done

easybuild/framework/easyblock.py
boegel: @akesandgren Confirmed, that was fixed in #3029

PR #3184, opened 2020-02-01
enabling urlchecker workflow

shahzebsiddiqui:
@boegel this may be useful. If you are interested check the automated github action in buildtest.


ocaisa:
@shahzebsiddiqui I don't see this urlchecker test actually being run, is it being triggered at all?

shahzebsiddiqui:
It will work once its merged. See on marketplace https://github.com/marketplace/actions/urls-checker. 

Any incoming push will get a github action to trigger url check workflow and it will show up in https://github.com/easybuilders/easybuild-framework/actions 

To see it in action you can see https://github.com/HPC-buildtest/buildtest-framework/actions

ocaisa:
@shahzebsiddiqui I updated it to also work on PRs and I still don't see it in the CI

shahzebsiddiqui:
I think the PR needs to be merged first then have any incoming PR  should trigger the URLchecker action. I think reason why it is not working now is because the base branch (i.e ``develop``) does not have this workflow tied in.

shahzebsiddiqui:
@ocaisa  the developer has released ``v0.1.3``  I can't tell much of the changes that were done and no release notes. See https://github.com/SuperKogito/URLs-checker for more details.

If you want to upgrade to latest version let me know and i can bump the version.

I do note there were few false positive see https://github.com/HPC-buildtest/buildtest-framework/runs/428310246?check_suite_focus=true 

if you look at one of the red icons 

image


But the link is valid https://www.hpcwire.com/2019/01/17/pfizer-hpc-engineer-aims-to-automate-software-stack-testing/

I will report it to developer and see if it was fixed. Most of them seemed to work pretty well.

ocaisa:
Now that I think about it, is this really the right place to be running this? Shouldn't we be running this in the documentation repo?

ocaisa:
It was missing the `.yml` extension, that's why it didn't run

ocaisa:
I'm really not sure this test is worth it, it's quite slow and doesn't cache successful URLs (so we check things like `http://ugent.be/hpc/en` a hundred times). The only time I see it having value is when we are making a release (so pushing to master)

ocaisa:
Ok, I think this is a no-go, it took 11m and is failing for all our URLs with formatting. For posterity, to only implement it on master:
```
on:
  # Trigger the workflow on push or pull request,
  # but only for the master branch
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```

@shahzebsiddiqui If you think you could actually get this check to pass then feel free to reopen. I'm going to test it on the docs repo, I think it has more value there.

shahzebsiddiqui:
@ocaisa yes i have a PR for easybuild docs also https://github.com/easybuilders/easybuild/pull/591/files I see i will fix the .yml extension

PR #3185, opened 2020-02-01
black workflow  for python formatting

### LINKED ISSUE #3165: use Black for code formatting
from the website https://black.readthedocs.io/en/stable/
"Black makes code review faster by producing the smallest diffs possible. Blackened code looks the same regardless of the project you’re reading. Formatting becomes transparent after a while and you can focus on the content instead."

using Black would take away all the review comments and discussions about code formatting, which are often subjective and dependent on the preference of the reviewer.

formatting with Black could even be done on all PRs automatically, so the contributor does not even have to do anything.

Black also has a few options to relax the constraints a bit if desired:
-changing the max line length to 120 chars (as is the case now)
-ignoring changes to single and double quotes with --skip-string-normalization (by default, black uses double quotes everywhere).

the idea would be to start 'blackening' the framework code and test if everything works ok, next require Black formatting for all new PRs.

shahzebsiddiqui:
fix #3165

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,12 @@
+  
+name: Black Formatter
+on: [push, pull_request]
+jobs:
+  black-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Black Code Formatter
+        uses: lgeiger/black-action@master
+        with:
+          args: "."


.github/workflows/black.yml
```diff
@@ -0,0 +1,12 @@
...
+on: [push, pull_request]
+jobs:
+  black-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Black Code Formatter
+        uses: lgeiger/black-action@master
+        with:
+          args: "."
```
ocaisa: This should be `args: ". --check"`, see https://github.com/lgeiger/black-action

ocaisa: If you want to automatically "black" PRs, then you would need something like https://github.com/cclauss/autoblack

shahzebsiddiqui: @ocaisa it seems like this may be more troublesome than it looks. The regression test failed when running black manually.

shahzebsiddiqui:
i can see the black format is running in github action https://github.com/easybuilders/easybuild-framework/actions?query=workflow%3A%22Black+Formatter%22 though it may help to get all code base blacked and work through regtest if they broke. See #3187  before merging this

ocaisa:
This can only do a style check, and that is not really much good without a black pre-commit hook. I'm positive about going that way but this would need a bit more thought and work...and we would need to get the initial `black` of the repo to be merged first

shahzebsiddiqui:
i agree @ocaisa  first black the entire repo and merge to ``develop`` then once black pre-commit hook is configured in Github workflow then all pull request/push will be blacked. Only the commit diff will be blacked out so having that as part of GitHub check should be ok.

boegel:
@shahzebsiddiqui This would require all contributors to set up the black pre-commit hook?

#3187 is a bit of a nightmare btw, I'm getting an angry pink unicorn when visiting that PR (because the page takes too long to load)...

shahzebsiddiqui:
@boegel yes i am aware. Since it was first time it had to reformat alot of code. Reason why I created #3187 was to see if there would be any issues merging it for first time. I am eager to try out  the autoblack https://github.com/cclauss/autoblack mentioned by @ocaisa.

If it works it would solve the pre-commit issue. If you want I can close #3187 its up to you guys and we can update this PR to use autoblack instead.

ocaisa:
Autoblack cannot work as a GitHub action, see https://github.com/cclauss/autoblack#tldr-it-does-not-work

shahzebsiddiqui:
i would not close this issue yet. There may be one more option I want to try. We are discussing this in https://github.com/HPC-buildtest/buildtest-framework/issues/174 if you are interested.

shahzebsiddiqui:
@boegel I am going to close this PR since it doesn't look like this would work for you guys.

PR #3191, opened 2020-02-06
read patch files as bytestring to avoid UnicodeDecodeError for patches that include funky characters

### LINKED ISSUE #3190: hard crash with UnicodeDecodeError when reading patch file that contains backtick character when using Python 3
While testing https://github.com/easybuilders/easybuild-easyconfigs/pull/9832 on top of Python 3.6, `eb` crashed hard with a `UnicodeDecodeError`.

The underlying cause is a backtick character in the patch file `Grace-5.1.25-5build1.patch` on line 7238:

```
$  sed -n '7238p' easybuild/easyconfigs/g/Grace/Grace-5.1.25-5build1.patch
+  ximage->byte_order=T1_byte_order; /* Set t1lib�s byteorder */
```

Here's the error reproduced with a small script that basically just calls `det_patched_files` on the patch file:

```
  File "t.py", line 5, in <module>
    det_patched_files('easybuild/easyconfigs/g/Grace/Grace-5.1.25-5build1.patch')
  File "/work/easybuild-framework/easybuild/tools/filetools.py", line 996, in det_patched_files
    txt = read_file(path)
  File "/work/easybuild-framework/easybuild/tools/filetools.py", line 189, in read_file
    txt = handle.read()
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb4 in position 241240: invalid start byte
```

boegel:
fixes #3190

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -945,8 +945,8 @@ def det_patched_files(path=None, txt=None, omit_ab_prefix=False, github=False, f
 
     if path is not None:
         # take into account that file may contain non-UTF-8 characters;
-        # so, read a byyte string, and decode to ascii string (ignoring any non-ascii characters);
-        txt = read_file(path, mode='rb').decode('ascii', 'ignore')
+        # so, read a byte string, and decode to UTF-8 string (ignoring any non-UTF-8 characters);
+        txt = read_file(path, mode='rb').decode(encoding='utf-8', errors='ignore')
     elif txt is None:
         raise EasyBuildError("Either a file path or a string representing a patch should be supplied")
 
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -944,7 +944,9 @@ def det_patched_files(path=None, txt=None, omit_ab_prefix=False, github=False, f
     patched_regex = re.compile(patched_regex, re.M)
 
     if path is not None:
-        txt = read_file(path)
+        # take into account that file may contain non-UTF-8 characters;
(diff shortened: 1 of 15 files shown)


easybuild/tools/filetools.py
```diff
@@ -944,7 +944,9 @@ def det_patched_files(path=None, txt=None, omit_ab_prefix=False, github=False, f
     patched_regex = re.compile(patched_regex, re.M)
 
     if path is not None:
-        txt = read_file(path)
+        # take into account that file may contain non-UTF-8 characters;
+        # so, read a byyte string, and decode to ascii string (ignoring any non-ascii characters);
+        txt = read_file(path, mode='rb').decode('ascii', 'ignore')
```
zao: "`byyte`" is misspelled.
Would `'replace'` be slightly better than `'ignore'` as error mode for `decode`? It would provide slightly more disambiguation in the case of filenames like `"foo"` and `"foo\xa0"`.
Something like `'backslashreplace'` would be more informative, but that's not available in older Python.

boegel: Good point, fixed in 3f369d6f1

zao:
Regarding the Travis failure for 2.6 - the docs for [str.encode](https://docs.python.org/2/library/stdtypes.html?highlight=encode#str.encode) say:
> Changed in version 2.7: Support for keyword arguments added.

PR #3192, opened 2020-02-07
increase timeout & use api.github.com for connectivity check in check_github

Flamefire:
This should fix flaky test failures like in https://github.com/easybuilders/easybuild-framework/pull/3191

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1611,7 +1611,7 @@ def check_github():
     # check whether we're online; if not, half of the checks are going to fail...
     try:
         print_msg("Making sure we're online...", log=_log, prefix=False, newline=False)
-        urlopen(GITHUB_URL, timeout=30)
+        urlopen(GITHUB_API_URL, timeout=30)
         print_msg("OK\n", log=_log, prefix=False)
     except URLError as err:
         print_msg("FAIL (%s)", err, log=_log, prefix=False)
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1611,10 +1611,10 @@ def check_github():
     # check whether we're online; if not, half of the checks are going to fail...
     try:
         print_msg("Making sure we're online...", log=_log, prefix=False, newline=False)
-        urlopen(GITHUB_URL, timeout=5)
+        urlopen(GITHUB_URL, timeout=30)
         print_msg("OK\n", log=_log, prefix=False)
     except URLError as err:
-        print_msg("FAIL")
+        print_msg("FAIL", log=_log, prefix=False)
         raise EasyBuildError("checking status of GitHub integration must be done online")
 
     # GitHub user
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1614,7 +1614,7 @@ def check_github():
         urlopen(GITHUB_URL, timeout=30)
         print_msg("OK\n", log=_log, prefix=False)
     except URLError as err:
-        print_msg("FAIL", log=_log, prefix=False)
+        print_msg("FAIL (%s)", err, log=_log, prefix=False)
         raise EasyBuildError("checking status of GitHub integration must be done online")
 
     # GitHub user


boegel:
@Flamefire Good idea to bump the timeout & show the actual error, but doesn't help...

```
Making sure we're online...FAIL (HTTP Error 406: Not Acceptable)
```

Now what could cause a 406.... 🤔

boegel:
I was able to trigger it locally too just a minute ago, but not anymore now...

```
$ python3
Python 3.7.2 (default, Dec 27 2018, 07:35:06)
[Clang 10.0.0 (clang-1000.11.45.5)] on darwin
>>> from urllib.request import urlopen
>>> urlopen('https://github.com', timeout=30)
<http.client.HTTPResponse object at 0x105750588>
>>> urlopen('https://github.com', timeout=30)
<http.client.HTTPResponse object at 0x105750b70>
>>> urlopen('https://github.com', timeout=30)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 641, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 503, in _call_chain
    result = func(*args)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 406: Not Acceptable
```

ocaisa:
The suggestion I saw in https://github.com/renovatebot/renovate/issues/3026 is to add `api/v3/` to our GitHub urls.

easybuild/tools/github.py
```diff
@@ -1611,10 +1611,10 @@ def check_github():
     # check whether we're online; if not, half of the checks are going to fail...
     try:
         print_msg("Making sure we're online...", log=_log, prefix=False, newline=False)
-        urlopen(GITHUB_URL, timeout=5)
+        urlopen(GITHUB_URL, timeout=30)
```
ocaisa: Should we be checking `GITHUB_API_URL` here instead?

boegel: That shouldn't hurt, but it may not matter.

Checking `https://github.com` has worked fine until now, so why is it broken now?

@Flamefire Can you change the PR to use `GITHUB_API_URL` instead, to see if it helps?

Flamefire: I had the same idea, but we also use GITHUB_URL. Another test incoming

Flamefire: See https://github.com/easybuilders/easybuild-framework/pull/3192/commits/c495404a24425a3368e11b28a07454932debfe2f

easybuild/tools/github.py
```diff
@@ -1591,6 +1591,24 @@ def update_pr(pr_id, paths, ecs, commit_msg=None):
...
 
+def check_online_status():
+    """
+    Check whether we currently are online
+    Return True if online, else an URLError instance
+    """
+    # Try repeatedly and with different URLs to cater for flaky connections or servers
+    # Timeout and repeats set so total 1 minute
+    urls = [GITHUB_URL, GITHUB_API_URL]
+    for i in range(6):
```
boegel: @Flamefire This may be a step too far...

If the connection is flaky, we should recognize it as such, and not try again with increasingly larger timeout settings imho.

The `406` issue we're seeing is probably just because we're opening `github.com` rather than `api.github.com`?

Flamefire: > If the connection is flaky, we should recognize it as such, and not try again with increasingly larger timeout settings imho.

The timeout stays the same and isn't even triggered for the current issue. I could locally reproduce the the 406 is flaky: Running it multiple times works, so it might be a server issue with that specific URL.

> The 406 issue we're seeing is probably just because we're opening github.com rather than api.github.com

Why wouldn't github.com be correct? We use it in `download_repo`. But seems we mostly use GITHUB_API_URL so maybe we should check for that?

boegel: Well, in theory both should be fine, shouldn't matter much...

We can even check `google.com` instead (but that makes less sense).

It's very unclear why we're even hitting the 406 all of a sudden, this has never popped up before, and the way we're checking whether we're online hasn't changed recently.

I'm reluctant to start using a creative workaround for something being broken upstream...

easybuild/tools/github.py
```diff
@@ -1591,6 +1591,24 @@ def update_pr(pr_id, paths, ecs, commit_msg=None):
...
+        try:
+            urlopen(urls[i % len(urls)], timeout=10)
+            result = True
+            break
+        except URLError as err:
```
boegel: @Flamefire This won't even help with the 406 we were seeing, since that was a `HTTPError`, not a `URLError`...

Flamefire: This is Python3 then I guess. Because you verified that URLError was caught "somewhere": https://github.com/easybuilders/easybuild-framework/pull/3192#issuecomment-583295750

Flamefire: And a HTTPError is-a URLError: https://docs.python.org/3/library/urllib.error.html

easybuild/tools/github.py
Flamefire: Ok, then let's use the API-URL

ocaisa:
There was a timeout error  for a git clone on one of the travis tests, so there is something going on on github...but at least it was only one failure, I restarted that test.

ocaisa:
I'm going to close and reopen to retrigger all the tests and see if we succeed all the way through

Flamefire:
Failed again, but with a timeout this time. Seems github.com has problems...

Increase it to 60s to reduce test failures on CI? Try GITHUB_URL and GITHUB_API_URL and see if any works?

Note that CI tests succeed if the connectivity checks succeeds. There are no tests for any other failures as they are converted to warnings

boegel:
> Failed again, but with a timeout this time. Seems github.com has problems...
> 
> Increase it to 60s to reduce test failures on CI? Try GITHUB_URL and GITHUB_API_URL and see if any works?
> 
> Note that CI tests succeed if the connectivity checks succeeds. There are no tests for any other failures as they are converted to warnings

Updating timeout to 60s is fine by me, but we need to be careful with not ignoring the failing GitHub tests for a long time though...

I wouldn't consider both URLs and pretend everything is fine when just one of them works.

boegel:
Also, @boegelbot re-triggers flake test jobs in Travis (it did just now for the one that was failing...) :)

Flamefire:
> I wouldn't consider both URLs and pretend everything is fine when just one of them works.

But we have a variety of failures:
- HTTP Error 403: Forbidden
- HTTP Error 406
- Timeout

The first 2 are solely caused by flaky github servers, the last one might be or it might be the user. IMO the function I had in https://github.com/easybuilders/easybuild-framework/commit/c495404a24425a3368e11b28a07454932debfe2f was reasonable: After all we are checking whether there is internet connectivity. And that is the case if any of those URLs is reachable in a reasonable amount of time and it is somewhat imune to the flaky HTTP errors.

> Also, @boegelbot re-triggers flake test jobs in Travis (it did just now for the one that was failing...) :)

Yes, but now another one failed...

> we need to be careful with not ignoring the failing GitHub tests for a long time though...

Sure, but then the tests need to be written for that and the test user needs a fork and a github token.

boegel:
> > I wouldn't consider both URLs and pretend everything is fine when just one of them works.
> 
> But we have a variety of failures:
> 
>     * HTTP Error 403: Forbidden
> 
>     * HTTP Error 406
> 
>     * Timeout
> 
> 
> The first 2 are solely caused by flaky github servers, the last one might be or it might be the user. IMO the function I had in [c495404](https://github.com/easybuilders/easybuild-framework/commit/c495404a24425a3368e11b28a07454932debfe2f) was reasonable: After all we are checking whether there is internet connectivity. And that is the case if any of those URLs is reachable in a reasonable amount of time and it is somewhat imune to the flaky HTTP errors.

But usually with flaky errors other tests will fail, no? Some require internet connectivity...

> 
> > Also, @boegelbot re-triggers flake test jobs in Travis (it did just now for the one that was failing...) :)
> 
> Yes, but now another one failed...

Actually the same test config failed, just different. @boegelbot only re-triggers the test runs that failed because of flaky issues.

> 
> > we need to be careful with not ignoring the failing GitHub tests for a long time though...
> 
> Sure, but then the tests need to be written for that and the test user needs a fork and a github token.

For PR tests, there is no GitHub token available that can be used for the tests (it doesn't have the necessary permissions). That's a security problem, since having a token like that available makes it vulnerable to leaking (anyone could open a malicious PR to expose the token so they can copy & use it).
Only in a test suite run for branches (after a PR was merged) is the full test suite, incl. tests for the GitHub integration, run.

Flamefire:
> But usually with flaky errors other tests will fail, no? Some require internet connectivity...

They **may** fail. Maybe their API endpoints are not affected? Or they use different servers? Or even different URLs. The purpose of the function is to verify connectivity, not the reliability of GH servers. So IMO retrying and using different URLs is fine as this ensures exactly that: Connectivity

> Only in a test suite run for branches (after a PR was merged) is the full test suite, incl. tests for the GitHub integration, run.

Currently not done though: https://travis-ci.org/easybuilders/easybuild-framework/jobs/638300997
> *  GitHub token...(no token found) => FAIL

boegel:
> > Only in a test suite run for branches (after a PR was merged) is the full test suite, incl. tests for the GitHub integration, run.
> 
> Currently not done though: https://travis-ci.org/easybuilders/easybuild-framework/jobs/638300997
> 
> > * GitHub token...(no token found) => FAIL

That's misleading, the token is actually there for the tests (which pick up the token slightly different from usual, so it works fine in a headless environment).

Compare the output of the test suite (mostly `...`) with the one for a PR (where you see lots of `Skipping text_xxx, no GitHub token available?`).

Flamefire:
Ah yes, it is only `--check-github` that is not tested (as in: behavior checked)

boegel:
@Flamefire I'll go ahead and merge this since it at least helps resolve the `406` issue.

If more changes are needed, we can follow-up in another PR.

PR #3193, opened 2020-02-07
Restore flake8 default ignores

ocaisa:


boegel:
@ocaisa Can you clarify this?

Did we actually start complaining about more style issues because of the change made in #3162?

Where did you get this default list from?

ocaisa:
> Did we actually start complaining about more style issues because of the change made in #3162?

Yes, by giving an explicit ignore we've overridden the default ignores leading to annoying loops as found in https://github.com/easybuilders/easybuild-framework/pull/3186#discussion_r374118184

> Where did you get this default list from?
It's the flake8 default for v3.5.0, see `flake8 --help` and check `--ignore=errors`

ocaisa:
I updated this to only extend the warnings, this requires `flake8 >= 3.6`

PR #3194, opened 2020-02-07
fix online check in --check-github: try repeatedly and with different URLs to cater for HTTP issues

Flamefire:
Follow up to #3192

There can be many failures:

-    HTTP Error 403: Forbidden
-    HTTP Error 406
-    Timeout

The first 2 are solely caused by flaky github servers, the last one might be or it might be the user.
So retrying is reasonable: After all we are checking whether there is internet connectivity. And that is the case if any of those URLs is reachable in a reasonable amount of time. This makes the check somewhat immune to the flaky HTTP errors.

Note: This helps a lot for faster contributions. If GH currently returns a HTTP error then boegelbot will restart the whole job (or even many jobs) wasting a lot of time spend in setup and other tests. So not erroring out unnecessarily reduces stress on the test servers

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1594,21 +1594,24 @@ def update_pr(pr_id, paths, ecs, commit_msg=None):
 def check_online_status():
     """
     Check whether we currently are online
-    Return True if online, else an URLError instance of the last failure
+    Return True if online, else a list of error messages
     """
     # Try repeatedly and with different URLs to cater for flaky servers
     # E.g. Github returned "HTTP Error 403: Forbidden" and "HTTP Error 406: Not Acceptable" randomly
     # Timeout and repeats set to total 1 minute
     urls = [GITHUB_URL, GITHUB_API_URL]
     num_repeats = 6
+    errors = set()  # Use set to record only unique errors
     for attempt in range(num_repeats):
+        # Cycle through URLs
+        url = urls[attempt % len(urls)]
         try:
-            urlopen(urls[attempt % len(urls)], timeout=10)
-            result = True
+            urlopen(url, timeout=10)
+            errors = None
             break
         except URLError as err:
-            result = err
-    return result
+            errors.add('%s: %s' % (url, err))
+    return sorted(errors) if errors else True
 
 
 def check_github():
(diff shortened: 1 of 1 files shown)


easybuild/tools/github.py
```diff
@@ -1591,6 +1591,25 @@ def update_pr(pr_id, paths, ecs, commit_msg=None):
...
+def check_online_status():
+    """
+    Check whether we currently are online
+    Return True if online, else an URLError instance of the last failure
+    """
+    # Try repeatedly and with different URLs to cater for flaky servers
+    # E.g. Github returned "HTTP Error 403: Forbidden" and "HTTP Error 406: Not Acceptable" randomly
+    # Timeout and repeats set to total 1 minute
+    urls = [GITHUB_URL, GITHUB_API_URL]
+    for i in range(6):
```
boegel: Don't use single-letter variables outside of list comprehensions please. Maybe use `idx` or `iter` or `attempt_id`?

Also clarify the `6`: `total 1 minute (6 * 10s)`?

Flamefire: Done: https://github.com/easybuilders/easybuild-framework/pull/3194/commits/87fd425566bbe3827a7ff5ade948dbaadc40a005

easybuild/tools/github.py
```diff
@@ -1591,6 +1591,26 @@ def update_pr(pr_id, paths, ecs, commit_msg=None):
...
+    # Timeout and repeats set to total 1 minute
+    urls = [GITHUB_URL, GITHUB_API_URL]
+    num_repeats = 6
+    for attempt in range(num_repeats):
+        try:
+            urlopen(urls[attempt % len(urls)], timeout=10)
+            result = True
+            break
+        except URLError as err:
+            result = err
```
boegel: I hate to be pedantic, but we're only keeping track of the last failure here, how about keeping track of al off them (incl. the URLs we tried)?

So, make `result` a list of error messages in case of failures, and mention the URL that was tested in there...

Flamefire: You really want to see all 6 error messages and the 2 URLs repeated 3 times each? Sounds overkill to me.

I can do that, but not til Monday. So maybe merge this and hope we never see an error again?

boegel: Well, it just feels wrong to hide the earliest 5 errors...
Are they all the same? Is it a different issue on every attempt? Is only one of the URLs broken?

Those are all questions you may want an answer to when the check fails after 6 attempts.

Flamefire: Current test will look like this (prefixed urls with "fo://" to force failure):

> Making sure we're online...FAIL (fo://https://api.github.com: <urlopen error unknown url type: fo>, fo://https://github.com: <urlopen error unknown url type: fo>)

Flamefire: See https://github.com/easybuilders/easybuild-framework/pull/3194/commits/2e11b4829a6c804750a954a13846730b552d612b

PR #3197, opened 2020-02-10
Add an option to git_config to retain the .git directory

### LINKED ISSUE #3196: git_clone and the .git directory
When using git_config on sources the tar file created excludes the .git directory in:

.../EasyBuild/4.1.1/lib/python2.7/site-packages/easybuild/tools/filetools.py

on the line:

\# create an archive and delete the git repo directory
tar_cmd = ['tar', 'cfvz', targetpath, '--exclude', '.git', repo_name]

Some software uses the .git files to determine its own version and the directory is crucial for its correct compilation. There is no flag to alter this behaviour. Can this be changed?

ocaisa:
Fixes #3196


akesandgren:
Going in, thanks @ocaisa!

PR #3198, opened 2020-02-10
don't crash when GitPython is not installed in Python3

Flamefire:
When there is a folder (or tar file) named "git" in PYTHONPATH
(includes PWD) Python3 imports that as a namespace package
So to check if we actually have GitPython we need to import something
from that package to trigger an import error.

### DIFF
--- a/easybuild/tools/version.py
+++ b/easybuild/tools/version.py
@@ -55,18 +55,18 @@ def get_git_revision():
     relies on GitPython (see http://gitorious.org/git-python)
     """
     try:
-        import git
+        from git import Git, GitCommandError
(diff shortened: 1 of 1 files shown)


boegel:
@Flamefire Same issue in `easybuild/tools/repository/gitrepo.py`, `easybuild/tools/github.py`?

Flamefire:
Already check, no they all import GitCommandError (probably because of this)

boegel:
> Already check, no they all import GitCommandError (probably because of this)

Ah, right, they *also* have the `from git import`... Probably not for this exact reason though, but fine I guess.

PR #3199, opened 2020-02-11
Support for lower case module naming scheme

lucamar:
As discussed already by @gppezzi with @boegel, I'm submitting a pull request to support the lowercase module naming scheme in EasBuild (release 4).

### DIFF
--- a/easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
+++ b/easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
@@ -1,0 +1,32 @@
+import os
+import re
+
+from easybuild.tools.module_naming_scheme.mns import ModuleNamingScheme
+from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
+
+class LowercaseModuleNamingScheme(ModuleNamingScheme):
+    """Class implementing a lowercase module naming scheme."""
+
+    REQUIRED_KEYS = ['name', 'version', 'versionsuffix', 'toolchain']
+
+    def det_full_module_name(self, ec):
+        """
+        Determine full module name from given easyconfig, according to the EasyBuild module naming scheme.
+        :param ec: dict-like object with easyconfig parameter values (e.g. 'name', 'version', etc.)
+        :return: string with full module name <name>/<installversion>, e.g.: 'gzip/1.5-goolf-1.4.10'
+        """
+        return os.path.join(ec['name'], det_full_ec_version(ec)).lower()
+
+    def is_short_modname_for(self, short_modname, name):
+        """
+        Determine whether the specified (short) module name is a module for software with the specified name.
+        Default implementation checks via a strict regex pattern, and assumes short module names are of the form:
+            <name>/<version>[-<toolchain>]
+        """
+        modname_regex = re.compile('^%s(/\S+)?$' % re.escape(name.lower()))
+        res = bool(modname_regex.match(short_modname))
+
+        self.log.debug("Checking whether '%s' is a module name for software with name '%s' via regex %s: %s",
+                       short_modname, name, modname_regex.pattern, res)
+
+        return res


easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
```diff
@@ -0,0 +1,32 @@
...
+        """
+        return os.path.join(ec['name'], det_full_ec_version(ec)).lower()
+
+    def is_short_modname_for(self, short_modname, name):
+        """
+        Determine whether the specified (short) module name is a module for software with the specified name.
+        Default implementation checks via a strict regex pattern, and assumes short module names are of the form:
+            <name>/<version>[-<toolchain>]
+        """
+        modname_regex = re.compile('^%s(/\S+)?$' % re.escape(name.lower()))
```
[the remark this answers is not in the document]
ocaisa: I ran into a few of these yesterday, just use raw strings:
```
        modname_regex = re.compile(r'^%s(/\S+)?$' % re.escape(name.lower()))
```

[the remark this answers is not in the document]
lucamar: Thanks @ocaisa, I've just commited the fix

smoors:
@lucamar we probably need a test for this new MNS, see `test/framework/module_generator.py`.
are you up for that?

easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
```diff
@@ -0,0 +1,33 @@
+import os
+import re
+
+from easybuild.tools.module_naming_scheme.mns import ModuleNamingScheme
+from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
+
+
+class LowercaseModuleNamingScheme(ModuleNamingScheme):
```
boegel, line 8: @lucamar Wouldn't it make sense to derive this from `EasyBuildMNS`, and rename it to `LowercaseEasyBuildMNS`?

easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
```diff
@@ -0,0 +1,33 @@
...
+
+    REQUIRED_KEYS = ['name', 'version', 'versionsuffix', 'toolchain']
+
+    def det_full_module_name(self, ec):
+        """
+        Determine full module name from given easyconfig, according to the EasyBuild module naming scheme.
+        :param ec: dict-like object with easyconfig parameter values (e.g. 'name', 'version', etc.)
+        :return: string with full module name <name>/<installversion>, e.g.: 'gzip/1.5-goolf-1.4.10'
+        """
+        return os.path.join(ec['name'], det_full_ec_version(ec)).lower()
```
boegel, line 19: This could be simplified to

```python
return super(EB_LowercaseEasyBuildMNS, self).det_full_module_name().lower()
```

easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
lucamar: @boegel Sure, I can change it to a subclass of `EasyBuildMNS` and the filename `lowercase_mns.py`: I think that initially we just copied the structure of `categorized_mns.py` as a template

easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
lucamar: Thanks for the hint, however if I change it as you suggested, I keep getting the `NameError` below: 
```
NameError: global name 'EB_LowercaseEasyBuildMNS' is not defined
```
Maybe I'm missing something trivial...

lucamar:
> @lucamar we probably need a test for this new MNS, see `test/framework/module_generator.py`.
> are you up for that?
Thanks for your feedback @smoors and sorry for my late reply. Should I add a test for the lowercase module naming scheme defining something similar to  `test_mns()` in `module_generator.py`?

easybuild/tools/module_naming_scheme/lowercase_module_naming_scheme.py
smoors: @lucamar that's because you did not yet implement the suggested change by @boegel to `LowercaseEasyBuildMNS` (see above).

smoors:
@lucamar the test should probably be on the same level as `test_module_naming_scheme()`, which is the test for the default `EasyBuildMNS`

lucamar:
I have closed this pull request as it is quite old and likely outdated.

PR #3203, opened 2020-02-14
significantly speed up -D/--dry-run by avoiding useless 'module show'

boegel:
`ModulesTool.exist` falls back to use `module show` by default since the module names it is passed may be partial module names which won't show up in `module avail`.

In `skip_available` we're always passing full module names though, so there we can tell it the module names are *not* partial...

This results in a speedup of about 2x for `eb -D`!


akesandgren:
Going in, thanks @boegel!

boegel:
@shahzebsiddiqui This should help quite a bit with `eb -D` being slow...

shahzebsiddiqui:
thanks @boegel this will certainly help. I will upgrade and test it out once you release 4.1.2

PR #3206, opened 2020-02-20
add support for --include-easyblocks-from-pr (REVIEW)

migueldiascosta:


easybuild/main.py
```diff
@@ -199,6 +200,10 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     eb_go, cfg_settings = set_up_configuration(args=args, logfile=logfile, testing=testing)
     options, orig_paths = eb_go.options, eb_go.args
 
+    if options.include_easyblocks_from_pr:
+        included_easyblocks = fetch_easyblocks_from_pr(options.include_easyblocks_from_pr)
+        include_easyblocks(options.tmpdir, included_easyblocks)
```
migueldiascosta: not sure if this is the correct `tmpdir`

easybuild/main.py
```diff
@@ -199,6 +200,10 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     eb_go, cfg_settings = set_up_configuration(args=args, logfile=logfile, testing=testing)
     options, orig_paths = eb_go.options, eb_go.args
 
+    if options.include_easyblocks_from_pr:
+        included_easyblocks = fetch_easyblocks_from_pr(options.include_easyblocks_from_pr)
```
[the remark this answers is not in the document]
boegel: We rely heavily on `build_option` in `tools/github.py`, so only using `fetch_easyblocks_from_pr` after setting up the configuration indeed makes most sense.

I would move this inside of `set_up_configuration` though, since it's really part of setting up the configuration?
As long as you do it after `init_build_options`, you're good.
Do add a comment to explain why `include_easyblocks` is called there too though, and not in `_postprocess_include`.

That way people who leverage EasyBuild as a library (which requires calling `set_up_configuration` first) can also pull in easyblocks from a pull requests (by defining `$EASYBUILD_INCLUDE_EASYBLOCKS_FROM_PR`)...

easybuild/main.py
boegel: You better create a subdirectory in `self.tmpdir`, to avoid that `--include-easyblocks` and `--include-easyblocks-from-pr` step on each others toes if both options are used at the same time.

Or we tweak the `include_easyblocks` function to use `tempfile.mkdtemp` to create a subdirectory in the path that it is given, that's probably a better options in the long run:

```python
easyblocks_path = tempfile.mkdtemp(prefix=tmpdir, os.path.join(dir=tmpdir, prefix='included-easyblocks-')
```

Ideally, we also have this use case covered of using both `--include-easyblocks*` options in the tests...

easybuild/main.py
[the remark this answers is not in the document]
boegel: We rely heavily on `build_option` in `tools/github.py`, so only using `fetch_easyblocks_from_pr` after setting up the configuration indeed makes most sense.

I would move this inside of `set_up_configuration` though, since it's really part of setting up the configuration?
As long as you do it after `init_build_options`, you're good.
Do add a comment to explain why `include_easyblocks` is called there too though, and not in `_postprocess_include`.

That way people who leverage EasyBuild as a library (which requires calling `set_up_configuration` first) can also pull in easyblocks from a pull requests (by defining `$EASYBUILD_INCLUDE_EASYBLOCKS_FROM_PR`)...

easybuild/tools/github.py
```diff
@@ -384,9 +395,15 @@ def fetch_easyconfigs_from_pr(pr, path=None, github_user=None):
...
     github_account = build_option('pr_target_account')
-    github_repo = GITHUB_EASYCONFIGS_REPO
 
-    _log.debug("Fetching easyconfigs from %s/%s PR #%s into %s", github_account, github_repo, pr, path)
+    if github_repo is None:
+        github_repo = GITHUB_EASYCONFIGS_REPO
+    elif github_repo not in [GITHUB_EASYBLOCKS_REPO, GITHUB_EASYCONFIGS_REPO]:
+        raise EasyBuildError("Don't know how to fetch files from repo %s", github_repo)
+
+    easyfiles = 'easyconfigs' if github_repo == GITHUB_EASYCONFIGS_REPO else 'easyblocks'
```
boegel: A bit easier to digest, I think:

```python
if github_repo is None:
    github_repo = GITHUB_EASYCONFIGS_REPO

if github_repo == GITHUB_EASYCONFIGS_REPO:
    easyfiles = 'easyconfigs'
elif github_repo == GITHUB_EASYBLOCKS_REPO:
    easyfiles = 'easyblocks'
else:
    raise EasyBuildError("Don't know how to fetch files from repo %s", github_repo)

easybuild/main.py
migueldiascosta: hm, when looking into adding a test for the use of both `--include-easyblocks*` options, I noticed that `--list-easyblocks` doesn't take into account the easyblocks from a PR, because it's done in the `options.postprocess` step, before `set_up_configuration`...

PR #3210, opened 2020-02-21
add support for creating an index & using it when searching for easyconfigs

boegel:
With `--create-index` you can create an index file that significantly speeds up searching for (easyconfig) files (& patches) on slow filesystems. You can control how long the index is valid via `--index-max-age` (in seconds), where 1 week (604800 seconds) is the default, and `0` indicates that the index remains valid indefinitely.

If an index is available, it's used both for `--search` and `--robot`.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -764,8 +764,11 @@ def search_file(paths, query, short=False, ignore_dirs=None, silent=False, filen
 
         path_index = load_index(path, ignore_dirs=ignore_dirs)
         if path_index is None:
-            _log.info("No index found for %s, creating one...", path)
-            path_index = create_index(path, ignore_dirs=ignore_dirs)
+            if os.path.exists(path):
+                _log.info("No index found for %s, creating one...", path)
+                path_index = create_index(path, ignore_dirs=ignore_dirs)
+            else:
+                path_index = []
         else:
             _log.info("Index found for %s, so using it...", path)
 
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -56,7 +56,7 @@ from xml.etree import ElementTree
 from easybuild.base import fancylogger
 from easybuild.tools import run
 # import build_log must stay, to use of EasyBuildLog
-from easybuild.tools.build_log import EasyBuildError, dry_run_msg, print_msg
+from easybuild.tools.build_log import EasyBuildError, dry_run_msg, print_msg, print_warning
 from easybuild.tools.config import build_option
 from easybuild.tools.py2vs3 import std_urllib, string_type
 from easybuild.tools.utilities import nub
@@ -618,15 +618,29 @@ def create_index(path, ignore_dirs=None):
     return index
 
 
-def dump_index(path):
+def dump_index(path, max_age_sec=None):
     """
     Create index for files in specified path, and dump it to file (alphabetically sorted).
     """
+    if max_age_sec is None:
+        max_age_sec = build_option('index_max_age')
 
     index_fp = os.path.join(path, PATH_INDEX_FILENAME)
     index_contents = create_index(path)
 
-    write_file(index_fp, '\n'.join(sorted(index_contents)))
+    curr_ts = datetime.datetime.now()
+    if max_age_sec == 0:
+        end_ts = datetime.datetime.max
+    else:
+        end_ts = curr_ts + datetime.timedelta(0, max_age_sec)
+
+    lines = [
+        "# created at: %s" % str(curr_ts),
+        "# valid until: %s" % str(end_ts),
+    ]
+    lines.extend(sorted(index_contents))
+
+    write_file(index_fp, '\n'.join(lines), always_overwrite=False)
 
     return index_fp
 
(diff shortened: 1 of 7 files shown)


smoors:
@boegel is there an easy way to remove the index as well?

boegel:
> @boegel is there an easy way to remove the index as well?

You mean `rm`? ;)

Or do you want a `--remove-index`?

smoors:
I was thinking about how this is going to be used in practice.
I can imagine this feature being used a lot, but in some cases you really want to see the latest of the latest. maybe adding an option `--ignore-index` will do the trick?

boegel:
@smoors Having an `--ignore-index` makes a lot of sense, I'll look into that...

easybuild/tools/filetools.py
```diff
@@ -596,6 +597,113 @@ def download_file(filename, url, path, forced=False):
...
+        ignore_dirs = []
+
+    index = set()
+
+    if not os.path.exists(path):
+        raise EasyBuildError("Specified path does not exist: %s", path)
+    elif not os.path.isdir(path):
+        raise EasyBuildError("Specified path is not a directory: %s", path)
+
+    for (dirpath, dirnames, filenames) in os.walk(path, topdown=True):
```
smoors: follow symlinks too?
```
for (dirpath, dirnames, filenames) in os.walk(path, topdown=True, followlinks=True):
```

easybuild/tools/filetools.py
```diff
@@ -596,6 +597,113 @@ def download_file(filename, url, path, forced=False):
...
+        for filename in filenames:
+            # use relative paths in index
+            index.add(os.path.join(dirpath[len(path) + 1:], filename))
+
+        # do not consider (certain) hidden directories
+        # note: we still need to consider e.g., .local !
+        # replace list elements using [:], so os.walk doesn't process deleted directories
+        # see http://stackoverflow.com/questions/13454164/os-walk-without-hidden-folders
```
smoors: nit: use https

easybuild/tools/filetools.py
```diff
@@ -596,6 +597,113 @@ def download_file(filename, url, path, forced=False):
...
+
+    if not os.path.exists(path):
+        raise EasyBuildError("Specified path does not exist: %s", path)
+    elif not os.path.isdir(path):
+        raise EasyBuildError("Specified path is not a directory: %s", path)
+
+    for (dirpath, dirnames, filenames) in os.walk(path, topdown=True):
+        for filename in filenames:
+            # use relative paths in index
+            index.add(os.path.join(dirpath[len(path) + 1:], filename))
```
smoors: I think this gives a wrong relative path if the full path ends with '/'

test/framework/options.py
```diff
@@ -776,6 +776,49 @@ def test_search(self):
...
+        copy_file(toy_ec, self.test_prefix)
+
+        # install index that list more files than are actually available,
+        # so we can check whether it's used
+        index_txt = '\n'.join([
+            'toy-0.0.eb',
+            'toy-1.2.3.eb',
+            'toy-4.5.6.eb',
+        ])
+        write_file(os.path.join(self.test_prefix, '.eb-path-index'), index_txt)
```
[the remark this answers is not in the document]
boegel: Because then we wouldn't easily catch a change to the `PATH_INDEX_FILENAME` constant.

I'm deliberately hardcoding the `.eb-path-index` here, since if the constant would be changed for whatever weird reason, that should make the tests fail (since it changes the EasyBuild behavior, existing indexes would suddenly be ignored).

easybuild/tools/filetools.py
boegel: Doesn't hurt, sure.

easybuild/tools/filetools.py
boegel: You're absolutely right, nice catch!

```
>>> path = '/Users/kehoste/work/easybuild-easyconfigs'
>>> dirpath, _, _ = list(os.walk(path))[551]
>>> dirpath[len(path) + 1:]
'easybuild/easyconfigs/r/ReFrame'
```
```
>>> path = '/Users/kehoste/work/easybuild-easyconfigs/'
>>> dirpath, _, _ = list(os.walk(path))[551]
>>> dirpath[len(path) + 1:]
'asybuild/easyconfigs/r/ReFrame'
```

boegel: I should be using `os.path.relpath` instead:

```
>>> os.path.relpath('/Users/kehoste/work/easybuild-easyconfigs/easybuild/easyconfigs/r/ReFrame', '/Users/kehoste/work/easybuild-easyconfigs')
'easybuild/easyconfigs/r/ReFrame'
>>> os.path.relpath('/Users/kehoste/work/easybuild-easyconfigs/easybuild/easyconfigs/r/ReFrame', '/Users/kehoste/work/easybuild-easyconfigs/')
'easybuild/easyconfigs/r/ReFrame'
>>> os.path.relpath('/Users/kehoste/work/easybuild-easyconfigs/easybuild/easyconfigs/r/ReFrame', '/Users/kehoste/work/easybuild-easyconfigs////')
'easybuild/easyconfigs/r/ReFrame'
```

easybuild/tools/filetools.py
boegel: fixed

easybuild/tools/filetools.py
boegel: fixed in b5c5b38fa

test/framework/options.py
```diff
@@ -4681,6 +4722,51 @@ def test_cuda_compute_capabilities(self):
...
+        for pattern in patterns:
+            regex = re.compile(pattern, re.M)
+            self.assertTrue(regex.search(index_txt), "Pattern '%s' found in: %s" % (regex.pattern, index_txt))
+
+        # existing index is not overwritten without --force
+        error_pattern = "File exists, not overwriting it without --force: .*/.eb-path-index"
+        self.assertErrorRegex(EasyBuildError, error_pattern, self._run_mock_eb, args, raise_error=True)
+
+        # also test creating index that's infinitely valid
+        args.extend(['--index-max-ag=0', '--force'])
```
smoors:
```suggestion
        args.extend(['--index-max-age=0', '--force'])
```

test/framework/options.py
```diff
@@ -4681,6 +4722,51 @@ def test_cuda_compute_capabilities(self):
...
+        self._run_mock_eb(args, raise_error=True)
+        index_txt = read_file(index_fp)
+        regex = re.compile(r"^# valid until: 9999-12-31 23:59:59", re.M)
```
smoors: good to know time will come to a standstill that date :)

test/framework/options.py
boegel: indeed, it doesn't matter much, but I fixed it anyway in 72c6b3c01

test/framework/options.py
boegel: It might be sooner, it's just a guess ;)

smoors:
Going in, thanks @boegel!

PR #3212, opened 2020-02-24
Fix gitdb dependency on Python 2.6

Flamefire:
GitDB 4.0.1 was recently released (23.2.20) which does not work with Python 2.6 making CI tests fail
Pin to last known working version

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,10 @@ keyring==5.7.1; python_version < '2.7'
 keyring<=9.1; python_version >= '2.7'
 keyrings.alt; python_version >= '2.7'
 
+# GitDB 4.0.1 no longer supports Python 2.6
+gitdb==0.6.4; python_version < '2.7'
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -5,6 +5,10 @@ keyring==5.7.1; python_version < '2.7'
 keyring<=9.1; python_version >= '2.7'
 keyrings.alt; python_version >= '2.7'
 
+# GitDB 4.0.1 no longer supports Python 2.6
+gitdb==0.6.4; python_version < '2.7'
```
boegel: `==2.0.4` should work fine according to https://pypi.org/project/gitdb2/2.0.4/?

How come this only starts failing now though?

Flamefire: As mentioned, 4.0.1 was released on 23rd Feb. And you linked gitdb**2** but we(GitPython) need gitdb

boegel: Ah, my bad...

PR #3213, opened 2020-02-26
Allow use of SYSTEM as --try-toolchain option

### LINKED ISSUE #3104: `--try-toolchain=SYSTEM` does not work
It results in an error stating that `--try-toolchain` requires a toolchain name and a version number.

ocaisa:
Fixes #3104


akesandgren:
Going in, thanks @ocaisa!

PR #3216, opened 2020-02-26
make ModulesTool.exist more robust w.r.t. module wrappers, aliases, defaults, etc.

### LINKED ISSUE #3215: Module alias/wrapper detection is broken/incomplete
Some modules have wrapper modules or aliases. E.g. `Java/11` is not a real modules but implemented via a `.modulerc.lua` as `module_version("Java/11.0.6-ppc64le", "11")`

Similar thing exists for aliases: `module_alias(...)`

When EasyBuild checks for an existing module it runs `module show <modname>` and checks the output with a regexp `^\s*\S*/<modname>.*(\.lua)?:\s*$`

This isn't exactly reliable. E.g. one might have a module `Java/site-default-11` or so. Additionally not all possible modulerc locations are considered and aliases are completely unsupported. See https://lmod.readthedocs.io/en/latest/093_modulerc.html

The obvious solution would be to start parsing the modulerc files correctly.

But: Of course it isn't as this is what the modules tool can do way better.

Hence the better solution would be to check for `_mlstatus = False` and if that is found the module isn't available. Getting the filename isn't required for the correct return value of `<ModulesTool>.exist` so this can be greatly simplified.

Edit: This doesn't seem to work reliably for some reason. Sometimes lmod seems to return only `_mlstatus = True` even though the module isn't found. But only in EB, can't reproduce in shell. Seems to be that `LMOD_QUIET=1` also suppresses the error code

In case we need the filename sometime later we should split the stdout/stderr and parse only stderr for a filename (LMod, not sure about EnvironmentModules)

Flamefire:
Allows to find Java/whatver-11 from "module show Java/11" as well as `AIModule` aliased to `TensorFlow/2.1.0` or similar.

The scope widened during the course of this PR trying to fix various test failures with different module tools resulting in a complete implementation which is much more robust.

This now fixes #3215

Proposal: Remove the previous check via `module avail` and solely rely on `module show` as the true source which handles broken aliases by returning false instead of True for `module avail` in Lmod and False otherwise.

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -489,7 +489,8 @@ class ModulesTool(object):
         Only .modulerc file in Tcl syntax is considered here.
         DEPRECATED. Use exists()
         """
-        self.log.deprecated('module_wrapper_exists is unreliable and should no longer be used', '5.0')
+        self.log.deprecated('module_wrapper_exists is unreliable and should no longer be used. ' +
+                            'Use exists instead to check for an existing module or alias.', '5.0')
 
         if mod_wrapper_regex_template is None:
             mod_wrapper_regex_template = "^[ ]*module-version (?P<wrapped_mod>[^ ]*) %s$"
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -531,12 +531,12 @@ class ModulesTool(object):
 
         return wrapped_mod
 
-    def exist(self, mod_names, mod_exists_regex_template=r'^\s*\S*/%s.*:\s*$', skip_avail=False, maybe_partial=True):
+    def exist(self, mod_names, mod_exists_regex_template=None, skip_avail=False, maybe_partial=True):
         """
         Check if modules with specified names exists.
 
         :param mod_names: list of module names
-        :param mod_exists_regex_template: template regular expression to search 'module show' output with
+        :param mod_exists_regex_template: DEPRECATED and unused
         :param skip_avail: skip checking through 'module avail', only check via 'module show'
         :param maybe_partial: indicates if the module name may be a partial module name
         """
@@ -546,21 +546,25 @@ class ModulesTool(object):
 
             :param mod_name: module name
             """
-            txt = self.show(mod_name)
+            stderr = self.show(mod_name)
             res = False
-            names_to_check = [mod_name]
-            # The module might be an alias where the target can be arbitrary
-            # As a compromise we check for the base name of the module so we find
-            # "Java/whatever-11" when searching for "Java/11" (--> basename="Java")
-            basename = os.path.dirname(mod_name)
-            if basename:
-                names_to_check.append(basename)
-            for name in names_to_check:
-                mod_exists_regex = mod_exists_regex_template % re.escape(name)
-                if re.search(mod_exists_regex, txt, re.M):
-                    res = True
+            # Parse the output:
+            # - Skip whitespace
+            # - Any error -> Module does not exist
+            # - Check first non-whitespace line for something that looks like an absolute path terminated by a colon
+            mod_exists_regex = r'\s*/.+:\s*'
+            for line in stderr.split('\n'):
+                if OUTPUT_MATCHES['whitespace'].search(line):
+                    continue
+                if OUTPUT_MATCHES['error'].search(line):
                     break
+                if re.match(mod_exists_regex, line):
+                    res = True
+                break
             return res
+
+        if mod_exists_regex_template is not None:
+            self.log.deprecated('mod_exists_regex_template is no longer used', '5.0')
 
         if skip_avail:
             avail_mod_names = []
@@ -1417,19 +1421,6 @@ class Lmod(ModulesTool):
             res = super(Lmod, self).module_wrapper_exists(mod_name)
 
         return res
-
-    def exist(self, mod_names, skip_avail=False, maybe_partial=True):
-        """
-        Check if modules with specified names exists.
-
-        :param mod_names: list of module names
-        :param skip_avail: skip checking through 'module avail', only check via 'module show'
-        """
-        # module file may be either in Tcl syntax (no file extension) or Lua sytax (.lua extension);
-        # the current configuration for matters little, since the module may have been installed with a different cfg;
-        # Lmod may pick up both Tcl and Lua module files, regardless of the EasyBuild configuration
-        return super(Lmod, self).exist(mod_names, mod_exists_regex_template=r'^\s*\S*/%s.*(\.lua)?:\s*$',
-                                       skip_avail=skip_avail, maybe_partial=maybe_partial)
 
     def get_setenv_value_from_modulefile(self, mod_name, var_name):
         """
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -541,6 +541,9 @@ class ModulesTool(object):
         :param skip_avail: skip checking through 'module avail', only check via 'module show'
         :param maybe_partial: indicates if the module name may be a partial module name
         """
+        if mod_exists_regex_template is not None:
+            self.log.deprecated('mod_exists_regex_template is no longer used', '5.0')
+
         def mod_exists_via_show(mod_name):
             """
             Helper function to check whether specified module name exists through 'module show'.
@@ -563,9 +566,6 @@ class ModulesTool(object):
                     res = True
                 break
             return res
-
-        if mod_exists_regex_template is not None:
-            self.log.deprecated('mod_exists_regex_template is no longer used', '5.0')
 
         if skip_avail:
             avail_mod_names = []
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -487,7 +487,10 @@ class ModulesTool(object):
         """
         Determine whether a module wrapper with specified name exists.
         Only .modulerc file in Tcl syntax is considered here.
+        DEPRECATED. Use exists()
         """
+        self.log.deprecated('module_wrapper_exists is unreliable and should no longer be used', '5.0')
+
         if mod_wrapper_regex_template is None:
             mod_wrapper_regex_template = "^[ ]*module-version (?P<wrapped_mod>[^ ]*) %s$"
 
@@ -581,15 +584,6 @@ class ModulesTool(object):
                 # hidden modules are not visible in 'avail', need to use 'show' instead
                 self.log.debug("checking whether hidden module %s exists via 'show'..." % mod_name)
                 mod_exists = mod_exists_via_show(mod_name)
-
-            # if no module file was found, check whether specified module name can be a 'wrapper' module...
-            if not mod_exists:
-                self.log.debug("Module %s not found via module avail/show, checking whether it is a wrapper", mod_name)
-                wrapped_mod = self.module_wrapper_exists(mod_name)
-                if wrapped_mod is not None:
-                    # module wrapper only really exists if the wrapped module file is also available
-                    mod_exists = wrapped_mod in avail_mod_names or mod_exists_via_show(wrapped_mod)
-                    self.log.debug("Result for existence check of wrapped module %s: %s", wrapped_mod, mod_exists)
 
             self.log.debug("Result for existence check of %s module: %s", mod_name, mod_exists)
 
@@ -1408,7 +1402,7 @@ class Lmod(ModulesTool):
     def module_wrapper_exists(self, mod_name):
         """
         Determine whether a module wrapper with specified name exists.
-        First check for wrapper defined in .modulerc.lua, fall back to also checking .modulerc (Tcl syntax).
+        DEPRECATED. Use exists()
         """
         res = None
 
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -771,8 +771,8 @@ class ModulesTool(object):
         # also catch and check exit code
         exit_code = proc.returncode
         if kwargs.get('check_exit_code', True) and exit_code != 0:
-            raise EasyBuildError("Module command 'module %s' failed with exit code %s; stderr: %s; stdout: %s",
-                                 ' '.join(cmd_list[2:]), exit_code, stderr, stdout)
+            raise EasyBuildError("Module command '%s' failed with exit code %s; stderr: %s; stdout: %s",
+                                 ' '.join(cmd_list), exit_code, stderr, stdout)
 
         if kwargs.get('check_output', True):
             self.check_module_output(full_cmd, stdout, stderr)
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -543,9 +543,21 @@ class ModulesTool(object):
 
             :param mod_name: module name
             """
-            mod_exists_regex = mod_exists_regex_template % re.escape(mod_name)
             txt = self.show(mod_name)
-            return bool(re.search(mod_exists_regex, txt, re.M))
+            res = False
+            names_to_check = [mod_name]
+            # The module might be an alias where the target can be arbitrary
+            # As a compromise we check for the base name of the module so we find
+            # "Java/whatever-11" when searching for "Java/11" (--> basename="Java")
+            basename = os.path.dirname(mod_name)
+            if basename:
+                names_to_check.append(basename)
+            for name in names_to_check:
+                mod_exists_regex = mod_exists_regex_template % re.escape(name)
+                if re.search(mod_exists_regex, txt, re.M):
+                    res = True
+                    break
+            return res
 
         if skip_avail:
             avail_mod_names = []
@@ -643,7 +655,7 @@ class ModulesTool(object):
             ans = MODULE_SHOW_CACHE[key]
             self.log.debug("Found cached result for 'module show %s' with key '%s': %s", mod_name, key, ans)
         else:
-            ans = self.run_module('show', mod_name, check_output=False, return_output=True)
+            ans = self.run_module('show', mod_name, check_output=False, return_stderr=True)
             MODULE_SHOW_CACHE[key] = ans
             self.log.debug("Cached result for 'module show %s' with key '%s': %s", mod_name, key, ans)
 
@@ -765,7 +777,9 @@ class ModulesTool(object):
         if kwargs.get('check_output', True):
             self.check_module_output(full_cmd, stdout, stderr)
 
-        if kwargs.get('return_output', False):
+        if kwargs.get('return_stderr', False):
+            return stderr
+        elif kwargs.get('return_output', False):
             return stdout + stderr
         else:
             # the module command was run with an outdated selected environment variables (see LD_ENV_VAR_KEYS list)
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -114,17 +114,34 @@ class ModulesTest(EnhancedTestCase):
         self.modtool.run_module(['load', 'GCC/6.4.0-2.28'])
         self.assertEqual(os.environ['EBROOTGCC'], '/prefix/software/GCC/6.4.0-2.28')
 
-        # by default, exit code is checked and an error is raised if we run something that fails
-        error_pattern = "Module command 'module thisdoesnotmakesense' failed with exit code [1-9]"
-        self.assertErrorRegex(EasyBuildError, error_pattern, self.modtool.run_module, 'thisdoesnotmakesense')
+        # skip tests that rely on exit codes when using EnvironmentModulesTcl modules tool,
+        # because it doesn't use proper exit codes
+        if not isinstance(self.modtool, EnvironmentModulesTcl):
 
-        error_pattern = "Module command 'module load nosuchmodule/1.2.3' failed with exit code [1-9]"
-        self.assertErrorRegex(EasyBuildError, error_pattern, self.modtool.run_module, 'load', 'nosuchmodule/1.2.3')
+            # by default, exit code is checked and an error is raised if we run something that fails
+            error_pattern = "Module command '.*thisdoesnotmakesense' failed with exit code [1-9]"
+            self.assertErrorRegex(EasyBuildError, error_pattern, self.modtool.run_module, 'thisdoesnotmakesense')
+
+            # we need to use a different error pattern here with EnvironmentModulesC,
+            # because a load of a non-existing module doesnt' trigger a non-zero exit code...
+            # it will still fail though, just differently
+            if isinstance(self.modtool, EnvironmentModulesC):
+                error_pattern = "Unable to locate a modulefile for 'nosuchmodule/1.2.3'"
+            else:
+                error_pattern = "Module command '.*load nosuchmodule/1.2.3' failed with exit code [1-9]"
+            self.assertErrorRegex(EasyBuildError, error_pattern, self.modtool.run_module, 'load', 'nosuchmodule/1.2.3')
 
         # we can choose to blatently ignore the exit code,
-        # and also disable the output check that serves as a fallback
-        self.modtool.run_module('thisdoesnotmakesense', check_exit_code=False, check_output=False)
-        self.modtool.run_module('load', 'nosuchmodule/1.2.3', check_exit_code=False, check_output=False)
+        # and also disable the output check that serves as a fallback;
+        # we also enable return_output here, because trying to apply the environment changes produced
+        # by a faulty command is bound to cause trouble...
+        kwargs = {
+            'check_exit_code': False,
+            'check_output': False,
+            'return_output': True,
+        }
+        self.modtool.run_module('thisdoesnotmakesense', **kwargs)
+        self.modtool.run_module('load', 'nosuchmodule/1.2.3', **kwargs)
 
         # by default, the output (stdout+stderr) produced by the command is processed;
         # result is a list of useful info (module names in case of list/avail)
@@ -1185,7 +1202,7 @@ class ModulesTest(EnhancedTestCase):
     def test_exit_code_check(self):
         """Verify that EasyBuild checks exit code of executed module commands"""
         if isinstance(self.modtool, Lmod):
-            error_pattern = "Module command 'module load nosuchmoduleavailableanywhere' failed with exit code"
+            error_pattern = "Module command '.*load nosuchmoduleavailableanywhere' failed with exit code"
         else:
             # Tcl implementations exit with 0 even when a non-existing module is loaded...
             error_pattern = "Unable to locate a modulefile for 'nosuchmoduleavailableanywhere'"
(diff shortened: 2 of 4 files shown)


boegel:
@Flamefire This would basically also consider the output of `module show Java`, but I don't see how that helps in general, since this will only produce output for whatever module version is the default?

In addition, this would make things like `eb -D` significantly slower, since it will do an additional `module show` for every non-existing module (see also #3203)

Also, the tests should be enhanced.

Flamefire:
I don't understand. If you call `module_exist("Java")` and `module show Java` returns a module, then the result of `module_exist` is True, isn't it?

Currently the problem is: If "Java/11" is an alias for "Java/whatever-11" then `module show Java/11` will return `...Java/whetever-11.lua` but the before this patch the result of `module_exist` would be false because it searches for `Java/11*.lua` (which fails)   
With this patch only the part before "/" must match: `Java/*.lua`

> In addition, this would make things like eb -D significantly slower, since it will do an additional module show for every non-existing module (see also #3203)

Why? There is only an additional regexp check, not another `module show` call

boegel:
Ugh, I clearly didn't look at this closely enough... Thanks for clarifying @Flamefire!
And you're right, there's no extra `show` at all, you're just doing a double take on the output of `module show`.

In that case, I would:

* enhance the existing test for `exist` in `test/framework/modules.py` (see `def test_exist`) to cover the case you mentioned
* add a comment in the code to clarify why we're also considering the basename of the module name in `mod_exists_via_show`

Flamefire:
@boegel Done as suggested. I also implemented the show command so that it returns stderr not the combined stdout+stderr to reduce false positives (which can be fatal). Reasoning is that the output we want is contained in stderr (at least for LMod) and checking stderr to can have a path matching the regexp which we don't want.

easybuild/tools/modules.py
```diff
@@ -643,7 +654,7 @@ def show(self, mod_name):
             ans = MODULE_SHOW_CACHE[key]
             self.log.debug("Found cached result for 'module show %s' with key '%s': %s", mod_name, key, ans)
         else:
-            ans = self.run_module('show', mod_name, check_output=False, return_output=True)
+            ans = self.run_module('show', mod_name, check_output=False, return_stderr=True)
```
boegel: I don't think this is correct, since Lmod may be configured to spit out the output of `module show` to stdout (via `$LMOD_REDIRECT` set to `1` for example); that's how we have configured Lmod...

I also don't fully understand the motivation for this. You mentioned reducing of false positives, can you give an example of what you mean?

easybuild/tools/modules.py
```diff
@@ -543,9 +543,20 @@ def mod_exists_via_show(mod_name):
...
+            # The module might be an alias where the target can be arbitrary
+            # As a compromise we check for the base name of the module so we find
+            # "Java/whatever-11" when searching for "Java/11" (--> basename="Java")
+            basename = os.path.dirname(mod_name)
+            if basename:
+                names_to_check.append(basename)
+            for name in names_to_check:
+                mod_exists_regex = mod_exists_regex_template % re.escape(name)
+                if re.search(mod_exists_regex, txt, re.M):
+                    res = True
```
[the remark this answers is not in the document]
Flamefire: Doesn't matter much but the break will save 1 regexp check

easybuild/tools/modules.py
Flamefire: But we don't call module show, do we? We call the underlying lmod and that outputs configuration stuff (like the current module table) to stdout and user facing stuff to stderr.

Basically: the state can contain something which make the match succeed. I'm currently on leave so just check the current output stored here and you'll see what I mean. 

Alternatively we could actually use "module show" but that is a bash function. So I don't think that's reasonable

easybuild/tools/modules.py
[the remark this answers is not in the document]
Flamefire: Added the break, rebased and squashed

easybuild/tools/modules.py
Flamefire: An example what is printed to stdout:

```
import os
os.environ["MODULEPATH"] = "/lustre/ssd/ws/s3248973-EasyBuild/easybuild-haswell/modules/all:/home/s3248973/.local/EasyBuildDev/modules:/sw/modules/scs5/classic:/sw/modules/scs5/vis:/sw/modules/scs5/tools:/sw/modules/scs5/toolchain:/sw/modules/scs5/system:/sw/modules/scs5/phys:/sw/modules/scs5/perf:/sw/modules/scs5/numlib:/sw/modules/scs5/mpi:/sw/modules/scs5/math:/sw/modules/scs5/lib:/sw/modules/scs5/lang:/sw/modules/scs5/geo:/sw/modules/scs5/devel:/sw/modules/scs5/debugger:/sw/modules/scs5/data:/sw/modules/scs5/compiler:/sw/modules/scs5/chem:/sw/modules/scs5/cae:/sw/modules/scs5/bio:/sw/modules/scs5/base:/sw/modules/taurus/environment";
os.environ["_ModuleTable001_"] = "X01vZHVsZVRhYmxlXz17WyJNVHZlcnNpb24iXT0zLFsiY19yZWJ1aWxkVGltZSJdPTcyMDAwMCxbImNfc2hvcnRUaW1lIl09MC40MzAxNzY5NzMzNDI5LGRlcHRoVD17fSxmYW1pbHk9e30sbVQ9e1siRWFzeUJ1aWxkLWRldmVsb3AiXT17WyJmbiJdPSIvaG9tZS9zMzI0ODk3My8ubG9jYWwvRWFzeUJ1aWxkRGV2L21vZHVsZXMvRWFzeUJ1aWxkLWRldmVsb3AiLFsiZnVsbE5hbWUiXT0iRWFzeUJ1aWxkLWRldmVsb3AiLFsibG9hZE9yZGVyIl09Mixwcm9wVD17fSxbInN0YWNrRGVwdGgiXT0wLFsic3RhdHVzIl09ImFjdGl2ZSIsWyJ1c2VyTmFtZSJdPSJFYXN5QnVpbGQtZGV2ZWxvcCIsfSxtb2RlbnY9e1siZm4iXT0iL3N3L21vZHVsZXMvdGF1cnVzL2Vudmlyb25tZW50L21v";
os.environ["_ModuleTable002_"] = "ZGVudi9zY3M1Lmx1YSIsWyJmdWxsTmFtZSJdPSJtb2RlbnYvc2NzNSIsWyJsb2FkT3JkZXIiXT0xLHByb3BUPXtsbW9kPXtbInN0aWNreSJdPTEsfSx9LFsic3RhY2tEZXB0aCJdPTAsWyJzdGF0dXMiXT0iYWN0aXZlIixbInVzZXJOYW1lIl09Im1vZGVudi9zY3M1Iix9LH0sbXBhdGhBPXsiL2x1c3RyZS9zc2Qvd3MvczMyNDg5NzMtRWFzeUJ1aWxkL2Vhc3lidWlsZC1oYXN3ZWxsL21vZHVsZXMvYWxsIiwiL2hvbWUvczMyNDg5NzMvLmxvY2FsL0Vhc3lCdWlsZERldi9tb2R1bGVzIiwiL3N3L21vZHVsZXMvc2NzNS9jbGFzc2ljIiwiL3N3L21vZHVsZXMvc2NzNS92aXMiLCIvc3cvbW9kdWxlcy9zY3M1L3Rvb2xzIiwiL3N3L21vZHVsZXMvc2NzNS90b29sY2hhaW4iLCIvc3cv";
os.environ["_ModuleTable003_"] = "bW9kdWxlcy9zY3M1L3N5c3RlbSIsIi9zdy9tb2R1bGVzL3NjczUvcGh5cyIsIi9zdy9tb2R1bGVzL3NjczUvcGVyZiIsIi9zdy9tb2R1bGVzL3NjczUvbnVtbGliIiwiL3N3L21vZHVsZXMvc2NzNS9tcGkiLCIvc3cvbW9kdWxlcy9zY3M1L21hdGgiLCIvc3cvbW9kdWxlcy9zY3M1L2xpYiIsIi9zdy9tb2R1bGVzL3NjczUvbGFuZyIsIi9zdy9tb2R1bGVzL3NjczUvZ2VvIiwiL3N3L21vZHVsZXMvc2NzNS9kZXZlbCIsIi9zdy9tb2R1bGVzL3NjczUvZGVidWdnZXIiLCIvc3cvbW9kdWxlcy9zY3M1L2RhdGEiLCIvc3cvbW9kdWxlcy9zY3M1L2NvbXBpbGVyIiwiL3N3L21vZHVsZXMvc2NzNS9jaGVtIiwiL3N3L21vZHVsZXMvc2NzNS9jYWUiLCIvc3cvbW9kdWxlcy9zY3M1L2Jp";
os.environ["_ModuleTable004_"] = "byIsIi9zdy9tb2R1bGVzL3NjczUvYmFzZSIsIi9zdy9tb2R1bGVzL3RhdXJ1cy9lbnZpcm9ubWVudCIsfSxbInN5c3RlbUJhc2VNUEFUSCJdPSIvc3cvbW9kdWxlcy90YXVydXMvZW52aXJvbm1lbnQiLH0=";
os.environ["_ModuleTable_Sz_"] = "4";
_mlstatus = True
```
We don't want any of this for this task, do we?

boegel: Indeed, `$LMOD_CMD python show` still prints the output we're interested in to `stderr`, even if Lmod is configured to redirect to stdout (for example by setting `$LMOD_REDIRECT`).

easybuild/tools/modules.py
```diff
@@ -543,9 +543,21 @@ def mod_exists_via_show(mod_name):
...
             txt = self.show(mod_name)
-            return bool(re.search(mod_exists_regex, txt, re.M))
+            res = False
+            names_to_check = [mod_name]
+            # The module might be an alias where the target can be arbitrary
+            # As a compromise we check for the base name of the module so we find
+            # "Java/whatever-11" when searching for "Java/11" (--> basename="Java")
+            basename = os.path.dirname(mod_name)
+            if basename:
+                names_to_check.append(basename)
```
boegel, line 554: While looking into this, I noticed this change is not required at all to make the enhanced test pass, because of the fallback mechanism with `module_wrapper_exists` in `ModulesTool.show` that was added in #2606 .

So, I'm still puzzled why this change is required at al...

I enhanced the checks in `test_exist` a bit more, and roll back the changes you made to `mod_exists_via_show` here, and it still works fine.

See https://github.com/Flamefire/easybuild-framework/compare/improve_module_exist...boegel:improve_module_exist. 

I can open a PR to your branch with these changes once I get the `test_run_module` I added passing with all different module tools.

boegel: I propose we roll back this change for now, which is done as a part of https://github.com/Flamefire/easybuild-framework/pull/2 .

We can further extend `test_exist` to cover the use cases that are actually broken (for example module aliases as discussed in #3215), before looking into a fix.

easybuild/tools/modules.py
Flamefire: That is because `$LMOD_CMD` is the underlying command. `module` uses this, splits stdout/stderr and processes those differently. Then it may pass through stderr or redirect to stdout

easybuild/tools/modules.py
Flamefire: > So, I'm still puzzled why this change is required at al...

Because as explained in #3215 the fallback shouldn't be there: It is fuzzy and incomplete. If `module show` returns a module, then the module exists else it does not, no fallback required.

Just looked at why the tests pass and comparing the #3215 and the docu at https://lmod.readthedocs.io/en/latest/093_modulerc.html:

Fallback considers only: 
- `module_version` command
- only module paths and 1 given directory

Fallback hence fails for:
- `module_alias`
- system MODULERC files
- ~/.modulerc.lua or ~/.modulerc

The last one was the problem I found. Users can't change the files in modulepath but they may want to change aliases via a home-folder modulerc.

Counter-proposal: We keep this change and print a fat warning when a module is not found via `module show` but via the fallback and gradually get rid of the fallback (My assumption is the warning will never trigger with this change)

Or do you see any reason to keep and improve a hand-rolled implementation of part of LMod/envModules?

More arguments: The fallback is particularly tricky as you noted in https://github.com/easybuilders/easybuild-framework/pull/2606#issuecomment-426916059 and I'm not sure that this is addressed: If a module wrapper exists but the underlying module does not, what does happen? I can't find such a 2nd check in https://github.com/easybuilders/easybuild-framework/blob/3ff16d7c3fe5378ed92118f372e289f6bd820a98/easybuild/tools/modules.py#L486
Using `module show` only avoids this problem altogether.

To go forward faster: Is there anything wrong with this PR? Any possible failure you can see? Any test that fails with it? If you want a test which passes after but not before I guess using `module_alias` will be enough. Can add that Tuesday.

boegel: I don't have any strong objections against your approach, other than that it partially solves the same issue as `module_wrapper_exists` does.

I agree that your approach is better, thanks a lot for the clarification.
`module_wrapper_exists` indeed has shortcomings (quite a lot actually, as you point them out), and it's indeed not EasyBuild's task to re-implement the modules tool.

With that in mind, I suggest proceeding as follows:

1) Merge https://github.com/Flamefire/easybuild-framework/pull/2 to update this PR, mostly for the additional tests (which are largely not directly related to the changes made in this PR, but are still useful). I spent quite a bit of time on the test for `run_module` in particular, it would be great if that effort doesn't go to waste...

2) Revert commit a10a6ac that undoes your changes in `mod_exists_via_show` (or do a `git rebase --interactive` to kick it out entirely). I can also kick out that commit from https://github.com/Flamefire/easybuild-framework/pull/2 before you merge it, if you prefer.

3) Add more checks in `test_exist` to cover the use cases that are not dealt with by `module_wrapper_exists`, preferably not only `module_alias` but also the others, if possible. For `~/.modulerc.lua` and `~/.modulerc`, you can probably cover those in the test by temporarily redefining `$HOME`?

4) Remove `module_wrapper_exists` entirely, since it's no longer needed with your enhanced `mod_exists_via_show`. I see no real reason to be careful there, the checks in `test_exist` should be good enough to ensure that nothing gets broken that was covered by `module_wrapper_exists`.

    If you prefer being cautious, I guess adding a `log.deprecated` statement in `module_wrapper_exists` makes sense too, to leave the fallback in place for now but print a clear warning if it's still triggered. If we take this route, the message should clearly mention to report the use case that triggers the warning, so we know we missed something before we go ahead and remove that block of code.

5) Make your enhanced `mod_exists_via_show` a little bit stricter, as follows:
    ```python
    names_to_check.append(basename + '/')
    ```
    The regular expression used to check the output of `module show` is a bit too liberal now in combination with only using the directory part of the module name.

    Using just `Java` in the fallback check would result in checking with `'^\s*\S*/Java.*(\.lua)?:\s*$'` for example, while it should really be `'^\s*\S*/Java/.*(\.lua)?:\s*$'`.
    Maybe we should even use `basename + '/.*'`, and remove the `.*` from the `mod_exists_regex_template` values?
    I'm not sure why it's even there at all, we expect an exact match (except for the `.lua` part), don't we?

6) Rename `basename` to something else, since it causes confusion. You use `os.path.dirname` to define a variable named `basename`, which clashes directly with `os.path.basename` which does the exact opposite (return just the filename). :)

Flamefire: Will work on this Tuesday, some comments now to leave time for decisions:

> For ~/.modulerc.lua and ~/.modulerc, you can probably cover those in the test by temporarily redefining $HOME?

I didn't add tests yet for fear of accidentally removing the real file or e.g. missing to undefine it or so... I'll see what I can come up with

> Remove module_wrapper_exists entirely,

I'm all in for that. But what about your usual reasoning someone else might be using it?

> print a clear warning if it's still triggered. If we take this route, the message should clearly mention to report the use case that triggers the warning, so we know we missed something before we go ahead and remove that block of code.

We must not call it within EasyBuild or make the message only trigger if `module_wrapper_exists` returns anything because it is currently called if the previous check does not find anything, which is valid: When the module indeed does not exist. Problem: An alias might exist but the underlying module not. `module show` would return nothing but `module_wrapper_exists` would return something. So I'd never call it from EB and deprecate it like other functions

> Make your enhanced mod_exists_via_show a little bit stricter, as follows:

Why? If `$LMOD_CMD show foo/1.2` returns `bar_3.4.lua` because someone set the alias as such, then this is valid. AFAIK `module show` returns nothing when the module is not found (maybe an error? But even that not on stderr of the module tool as that is up to the wrapper)      
Hence I don't think it needs to be made any stricter. Use case for the loose version: IBM has a custom TensorFlow. Maybe someone makes an alias `TensorFlow_PowerAI` to `TensorFlow/1.13_Anaconda_...` or so to distinguish it from official TF

Do you know of anything where the strict check would be required? Then we should add negative tests too that cover that. But until then I would keep it loose

Flamefire:
@boegel Tests failing due to the tests for the fallback printing deprecated messages. Shall I remove those tests?

Flamefire:
Rebased and removed the succeeding but echoing "DEPRECATED" module_wrapper_exists tests

Flamefire:
@boegel Found multiple issues:
- `module-alias` support is different for different versions and tools. 
  - `module-alias JavaAlias Java/1.8.0_181` is possible for LMod 7+ only
  - `module-alias Java/Alias toy/42.1337` works for all but fails the current, restrictive regexp
- `module avail` may or may not list aliases (shown for e.g. LMod). If it does, the pointed to modules may not exist, hence `module show <mod>` would return failure.

I fixed and tested the first case. The second may need a major change and deprecation warning. However it is not totally clear what `exists` should return for a non-existant aliased module. I'd say False because it cannot be loaded which is usually what is wanted.

easybuild/tools/modules.py
```diff
@@ -487,7 +487,10 @@ def module_wrapper_exists(self, mod_name, modulerc_fn='.modulerc', mod_wrapper_r
         """
         Determine whether a module wrapper with specified name exists.
         Only .modulerc file in Tcl syntax is considered here.
+        DEPRECATED. Use exists()
         """
+        self.log.deprecated('module_wrapper_exists is unreliable and should no longer be used', '5.0')
```
boegel, line 492: Maybe also mention that the `exists` method should be used?

```python
self.log.deprecated('module_wrapper_exists is unreliable, use exists method instead', '5.0')

File: easybuild/tools/modules.py [hunk 5]
```diff
@@ -543,9 +546,25 @@ def mod_exists_via_show(mod_name):
...
+                    continue
+                if OUTPUT_MATCHES['error'].search(line):
+                    break
+                if re.match(mod_exists_regex, line):
+                    res = True
+                break
+            return res
+
+        if mod_exists_regex_template is not None:
+            self.log.deprecated('mod_exists_regex_template is no longer used', '5.0')
```
boegel, line 567: Move this to the top of the method (above `stderr = self.show(mod_name)` please.

easybuild/tools/modules.py
```diff
@@ -543,9 +546,25 @@ def mod_exists_via_show(mod_name):
...
+            # - Check first non-whitespace line for something that looks like an absolute path terminated by a colon
+            mod_exists_regex = r'\s*/.+:\s*'
+            for line in stderr.split('\n'):
+                if OUTPUT_MATCHES['whitespace'].search(line):
+                    continue
+                if OUTPUT_MATCHES['error'].search(line):
+                    break
+                if re.match(mod_exists_regex, line):
+                    res = True
+                break
```
boegel, line 563: If there's no match, you'll just continue processing, and may hit a match later, which disagrees with what the comments above claim (first non-whitespace line).

Maybe use:

```python
            for line in stderr.split('\n'):
                if OUTPUT_MATCHES['whitespace'].search(line):
                    continue
                elif OUTPUT_MATCHES['error'].search(line):
                    break
                else:
                    res = bool(re.match(mod_exists_regex, line))
                    break
            return res

File: easybuild/tools/modules.py [hunk 7]
```diff
@@ -543,9 +546,25 @@ def mod_exists_via_show(mod_name):
...
-            mod_exists_regex = mod_exists_regex_template % re.escape(mod_name)
-            txt = self.show(mod_name)
-            return bool(re.search(mod_exists_regex, txt, re.M))
+            stderr = self.show(mod_name)
+            res = False
+            # Parse the output:
+            # - Skip whitespace
+            # - Any error -> Module does not exist
+            # - Check first non-whitespace line for something that looks like an absolute path terminated by a colon
+            mod_exists_regex = r'\s*/.+:\s*'
```
boegel, line 555: Hmm, this quite liberal now... No reference to part of `mod_name` requires at all anymore?

test/framework/modules.py
```diff
@@ -56,6 +58,17 @@
...
+@contextmanager
+def temporary_home_dir():
+    tmpdir = tempfile.mkdtemp()
+    orig_home = os.environ['HOME']
+    os.environ['HOME'] = tmpdir
+    try:
+        yield tmpdir
+    finally:
+        os.environ['HOME'] = orig_home
+        shutil.rmtree(tmpdir)
```
boegel, line 70: This is a bit much imho, just setting `os.environ['HOME'] = self.test_prefix` is the test is enough (`self.test_prefix` is cleaned up automatically in the tests, and the environment is restored before every test starts)

test/framework/modules.py
```diff
@@ -168,18 +279,26 @@ def test_exist(self):
...
+                '    module-version Java/1.8.0_181 site_default',
+                '}',
             ])
         else:
             modulerc_tcl_txt = '\n'.join([
                 '#%Module',
                 'module-version Java/1.8.0_181 1.8',
+                'module-version Java/1.8.0_181 site_default',
+                'module-alias Java/Alias toy/42.1337',
+                # 'module-alias Java/NonExist non_existant/1',  # (only) LMod has this in module avail, disable for now
```
boegel, line 300: Should still this be commented out?

You're running a test with `Java/NonExist` below, so it shouldn't be commented?

You can only test whether `Java/NonExist` is included in the output of `avail` when `Lmod` is being used?

easybuild/tools/modules.py
boegel: Actually, I think skipping over non-matching values is correct, see the example in https://github.com/easybuilders/easybuild/issues/368#issuecomment-324910159: the `------` line must be part of `stderr` (since the stuff sent to `stdout` is supposed to be Python syntax).

So, you'll need to make sure you process all lines, not just stop at the first non-whitespace/error line.

```python
for line in stderr.split('\n'):
    if OUTPUT_MATCHES['whitespace'].search(line):
        continue
    elif OUTPUT_MATCHES['error'].search(line):
        break

    res = bool(re.match(mod_exists_regex, line))
    # if we found a match, we're done;
    # if not, we keep processing the "show" output
    if res:
        break

    return res
```

Wit this change, you should be able to reinstate the test you removed in https://github.com/easybuilders/easybuild-framework/pull/3216/commits/1454dd34da9bfc252f8c77337fcd3efc1892b1f7 ?

test/framework/modules.py
Flamefire: Interesting...
However I did have failures due to test_prefix being used for other stuff too so a whole new ~world~ folder was needed. I'm also very much in favor for local judging of code and this is simple enough to understand (6 lines of code) and meaning in usage is very clear. So I'd keep this if you don't mind. Otherwise I'll change too `os.environ['HOME'] = tempfile.mkdtemp()` at least (IIRC tempfile is changed so the folder is a subfolder of test_prefix)

test/framework/modules.py
Flamefire: > You're running a test with Java/NonExist below, so it shouldn't be commented?

Yes, with expected "False", so this also test if no other alias is used.

And no, LMod will always fail the test as exists checks avail first returning True although using "show" would return False. Hence my (now strong) suggestion to remove the exists-via-avail code

easybuild/tools/modules.py
Flamefire: No. As the test with the alias shows: You can alias `foo123`->`bar/4.5.6` if you wish to do so.

As all tests (including the negative tests) show that this works, it is fine.

Stronger argument: LMods stderr is empty if the module does not exist. EnvironmentModules output contains the "ERROR" first (and only) which is also caught. Furthermore only the first non-comment line is checked and the patter is `<space><absolute-path>:<space>` which is quite restrictive

easybuild/tools/modules.py
Flamefire: > If there's no match, you'll just continue processing, and may hit a match later, which disagrees with what the comments above claim

No, you missed the break which is triggered for the first non-empty line. And only the first non-empty line must be checked, so your proposed code would be wrong (to permissive)

> Actually, I think skipping over non-matching values is correct, see the example in easybuilders/easybuild#368 (comment):

No. The python code goes to stdout. Previously (to this PR) the `show` method return `stdout + stderr` which explains the bug in #368.   
The `module` shell function works by `eval`ing stdout and printing stderr to stdout. This is what I'm using here

easybuild/tools/modules.py
Flamefire: That would be the inner method which is wrong. It should trigger from the outer method

Another point for collapsing this monstrosity and just use exists-via-show ;)

easybuild/tools/modules.py
Flamefire: Done a bit more verbose dace69d1d

easybuild/tools/modules.py
boegel: Oh, ok, well, then move it above `def mod_exists_via_show` please

easybuild/tools/modules.py
boegel: The `------------` line, which is what your implementation was tripping over, must have been part of the `stderr` output, so that test case remains relevant now you're only grabbing the `stderr` output from `module show`?

So, you can't just stop processing the `show` stderr output lines after the first non-whitespace line that's not an error, or you'll base potentially base the result on an output line like `-------------` (which is why the test you removed was failing).

easybuild/tools/modules.py
boegel: OK, the alias thing prevents us from making this more restrictive, indeed.

But maybe we should check whether the path for which the patterns matches actually exists? Not sure if that would take us too far...

test/framework/modules.py
boegel: I'd prefer the latter if you really need an empty dir for `$HOME` (less code to worry about, and you're just cleaning up what's already being cleaned up here)

easybuild/tools/modules.py
Flamefire: Still no. The `OUTPUT_MATCHES['whitespace']` regex was designed to remove those `-------` lines (great idea BTW! wondering why this isn't used more)

The test stuffed in a mix of stdout+stderr into the cache which can never occur (anymore). The test then (rightly) failed due to ` "import os"` being the first non-empty line

boegel: OK, got it, I forgot about the `whitespace` regex also handling `----` lines...

easybuild/tools/modules.py
boegel: Yeah, ok, it's probably overkill indeed...

PR #3219, opened 2020-02-29
use checkout@v2 in GitHub Actions to fix broken re-triggered tests

boegel:
see https://github.com/actions/checkout/issues/23 and https://github.com/easybuilders/easybuild-easyconfigs/pull/9925


akesandgren:
Going in, thanks @boegel!

PR #3225, opened 2020-03-03
don't filter out None values in to_checksums, leave them in place

boegel:
This is required for https://github.com/easybuilders/easybuild-easyconfigs/pull/9963, where we need to use `None` as checksum value for the source tarball because it's created from a Git repository (no direct downloadable source tarball available), and we do want proper SHA256 checksums for the patch files...

### DIFF
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -446,9 +446,11 @@ def to_checksums(checksums):
     res = []
     for checksum in checksums:
         # each list entry can be:
-        # * a string (MD5 checksum)
+        # * None (indicates no checksum)
+        # * a string (MD5 or SHA256 checksum)
         # * a tuple with 2 elements: checksum type + checksum value
         # * a list of checksums (i.e. multiple checksums for a single file)
+        # * a dict (filename to checksum mapping)
         if isinstance(checksum, string_type):
             res.append(checksum)
         elif isinstance(checksum, (list, tuple)):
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/types.py
```diff
@@ -462,6 +462,8 @@ def to_checksums(checksums):
             for key, value in checksum.items():
                 validated_dict[key] = to_checksums(value)
             res.append(validated_dict)
+        else:
+            res.append(checksum)
```
akesandgren: Wouldn't it be prudent to add None to the list of possible values in the "each list entry can be" list above (the comment that is)

akesandgren: And also, doesn't this construct open up for having just an integer as the checksum value? which will probably cause problems...

boegel: good point, fixed in a2786d66c

akesandgren: what about my second comment? 
in other words, shouldn't that else: really check if it None instead of just being an else:?

boegel: This is just the type auto-conversion stuff, not the checking of checksum values.
It takes care of changing `[['sha256', '<sha256_checksum>']]` to `[('sha256', '<sha256_checksum>')]` for example (which is partially done because of the `.yep` easyconfigs format).

Anything that is not a valid checksum value (like an integer) will still result in a failure when the checksum values are actually picked up in `easyblock.py`.

akesandgren:
Going in, thanks @boegel!

migueldiascosta:
shouldn't `check_checksums_for` in `easyblock.py` also filter out None values, so that `--check-contrib` doesn't fail?

boegel:
@migueldiascosta Done in #3244

PR #3227, opened 2020-03-03
mention CPU arch name in comment for uploaded test report, if it's known by archspec

boegel:
leveraging `archspec` if it's installed (https://github.com/archspec/archspec)

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -95,8 +95,8 @@ from easybuild.tools.package.utilities import avail_package_naming_schemes
 from easybuild.tools.toolchain.compiler import DEFAULT_OPT_LEVEL, OPTARCH_MAP_CHAR, OPTARCH_SEP, Compiler
 from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME
 from easybuild.tools.repository.repository import avail_repositories
-from easybuild.tools.systemtools import check_python_version, get_cpu_architecture, get_cpu_family, get_cpu_features
-from easybuild.tools.systemtools import get_system_info
+from easybuild.tools.systemtools import UNKNOWN, check_python_version, get_cpu_architecture, get_cpu_family
+from easybuild.tools.systemtools import get_cpu_features, get_system_info
 from easybuild.tools.version import this_is_easybuild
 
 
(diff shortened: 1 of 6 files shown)


migueldiascosta:
worth also showing the arch name in `eb --show-system-info`, if it is known?

easybuild/tools/testing.py
```diff
@@ -264,6 +264,11 @@ def post_easyconfigs_pr_test_report(pr_nr, test_report, msg, init_session_state,
         'os_version': system_info['os_version'],
         'pyver': system_info['python_version'].split(' ')[0],
     }
+
+    # also mention CPU architecture name, but only if it's known
+    if system_info['cpu_arch_name'] != UNKNOWN:
+        short_system_info['cpu_model'] += " (%s)" % system_info['cpu_arch_name']
```
migueldiascosta: `short_system_info` is a string, not a dictionary

boegel: nice catch, thanks...

fixed in 8a1115f

boegel:
> worth also showing the arch name in `eb --show-system-info`, if it is known?

Good point, fixed in 1d571f0b7, should be good to go now?

easybuild/tools/options.py
```diff
@@ -1166,6 +1167,11 @@ def show_system_info(self):
             "  -> vendor: %s" % system_info['cpu_vendor'],
             "  -> architecture: %s" % get_cpu_architecture(),
             "  -> family: %s" % get_cpu_family(),
+        ]
+        if cpu_arch_name:
+            lines.append("  -> arch name: %s" % cpu_arch_name)
```
migueldiascosta: I think there is a small inconsistency... if here you mean exactly this, then without archspec it will still be printed, but UNKNOWN, and in `test_show_system_info` there is no reason to check for HAVE_ARCHSPEC, since the pattern will always be there

otherwise this should be `if cpu_arch_name != UNKNOWN`, no?

boegel: Maybe we should hint that `archspec` is not installed when `cpu_arch__name == UNKNOWN`?

migueldiascosta:
Going in, thanks @boegel!

PR #3228, opened 2020-03-04
Improve raised error in remove_dir and deprecate rmtree2

Flamefire:
A look like https://gist.github.com/migueldiascosta/05c804a26b22ef266d4e577db26f19cf is not really helpful as the error is very generic:
`
ERROR EasyBuild crashed with an error (at easybuild/base/exceptions.py:124 in __init__): Failed to remove path /work/maintainers/CO7/broadwell/software/R/3.6.2-fosscuda-2019b with shutil.rmtree, even after 3 attempts. (at easybuild/tools/filetools.py:1399 in rmtree2)`

This adds the actual errors thrown for further investigation like:
```
Failed to remove path /work/maintainers/CO7/broadwell/software/R/3.6.2-fosscuda-2019b with shutil.rmtree, even after 3 attempts. 
Reasons: [FileNotFoundError(2, 'No such file or directory'), FileNotFoundError(2, 'No such file or directory'), FileNotFoundError(2, 'No such file or directory')]
```

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1384,6 +1384,7 @@ def rmtree2(path, n=3):
     """Wrapper around shutil.rmtree to make it more robust when used on NFS mounted file systems."""
 
     ok = False
+    errors = []
     for i in range(0, n):
         try:
             shutil.rmtree(path)
@@ -1391,12 +1392,14 @@ def rmtree2(path, n=3):
             break
         except OSError as err:
             _log.debug("Failed to remove path %s with shutil.rmtree at attempt %d: %s" % (path, n, err))
+            errors.append(err)
             time.sleep(2)
 
             # make sure write permissions are enabled on entire directory
             adjust_permissions(path, stat.S_IWUSR, add=True, recursive=True)
     if not ok:
-        raise EasyBuildError("Failed to remove path %s with shutil.rmtree, even after %d attempts.", path, n)
+        raise EasyBuildError("Failed to remove path %s with shutil.rmtree, even after %d attempts.\nReason(s): %s",
+                             path, n, errors)
     else:
         _log.info("Path %s successfully removed." % path)
 
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -303,11 +303,27 @@ def remove_dir(path):
         dry_run_msg("directory %s removed" % path, silent=build_option('silent'))
         return
 
-    try:
-        if os.path.exists(path):
-            rmtree2(path)
-    except OSError as err:
-        raise EasyBuildError("Failed to remove directory %s: %s", path, err)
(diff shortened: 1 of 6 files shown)


boegel:
@Flamefire Wherever `rmtree2` is still being used, we should use `remove_dir` instead.

We should also i) move the code from `rmtree2` into `remove_dir` (which still calls `rmtree2`) and ii) deprecate `rmtree2`.

Are you up for tackling this while you're at it?

Flamefire:
@boegel Done in https://github.com/easybuilders/easybuild-framework/pull/3228/commits/9249291e31dd78f1611c5c187b603e91a8e1861e

easybuild/tools/filetools.py
```diff
@@ -303,11 +303,27 @@ def remove_dir(path):
...
-    try:
-        if os.path.exists(path):
-            rmtree2(path)
-    except OSError as err:
-        raise EasyBuildError("Failed to remove directory %s: %s", path, err)
+    if os.path.exists(path):
+        ok = False
+        errors = []
+        # Try multiple times to cater for temporary failures on e.g. NFS mounted paths
+        max_attempts = 3
```
migueldiascosta: make this an argument with default value 3?

easybuild/tools/filetools.py
```diff
@@ -1383,22 +1399,8 @@ def path_matches(path, paths):
...
-            time.sleep(2)
-
-            # make sure write permissions are enabled on entire directory
-            adjust_permissions(path, stat.S_IWUSR, add=True, recursive=True)
-    if not ok:
-        raise EasyBuildError("Failed to remove path %s with shutil.rmtree, even after %d attempts.", path, n)
-    else:
-        _log.info("Path %s successfully removed." % path)
+    _log.deprecated("Use 'remove_dir' rather than 'rmtree2'", '5.0')
+    remove_dir(path)
```
migueldiascosta: pass `n` to `remove_dir`? In theory, until `5.0` someone else may still be using `rmtree2` and expecting passing `n` to work...

easybuild/tools/filetools.py
Flamefire: @boegel said to use `remove_dir`. And this function uses the max_attempts as an implementation detail which isn't changeable by the user. I'd rather keep this behavior

easybuild/tools/filetools.py
Flamefire: The function removes a directory. The attempts was made an implementation detail when adding `remove_dir`. So it is very unlikely that anyone even notices that another value of n was used. I haven't found any case in EasyBuild that changed the default so there it was not required.

Adding the param to `remove_dir` would mean we have to keep this parameter for a long time and I don't see the gain for that.

migueldiascosta:
Going in, thanks @Flamefire!

PR #3235, opened 2020-03-07
reorganize EasyBlock.extensions_step to ensure correct filtering of extensions 

### LINKED ISSUE #3167: eb --skip --rebuild isn't filtering out already built modules correctly
Doing a --skip --rebuild on Python-2.7.16 will rebuild things like:
more-itertools wcwidth MarkupSafe Jinja2 Pygments Babel Sphinx Click

The modulename conversioni of those lowercase(), sed /-/_/g, is not taken into account during skip filtering.

boegel:
(fixes #3167)

Already installed Python package were getting reinstalled under `--skip` when the Python module name was slightly different that the name of the extension, i.e. when the name only had to lower cased and `-` replaced with `_`, like `PythonPackage` does to determine the name that should be used in the `import` sanity check (see also https://github.com/easybuilders/easybuild-easyblocks/pull/1560).

The problem was that the `EasyBlock.extensions_step` method was using the raw "dict" value that specifies the extension name, version & extra info, not a proper `Extension` instance, and hence the changed `modulename` that is put in place by `PythonPackage` (which derives from `Extension`) was being blatantly ignored.

This is fixed by breaking up the implementation of `extensions_step` a bit, to first create `Extension` instances, and only then skipping already installed extensions, before installing the missing extensions.

The `resolve_exts_filter_template` function already supports both passing a proper `Extension` instance (which we're switching to now) or a `dict` value (which we're moving away from in `extensions_step`).

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1473,17 +1473,18 @@ class EasyBlock(object):
             raise EasyBuildError("Skipping of extensions, but no exts_filter set in easyconfig")
 
         res = []
-        for ext in self.exts:
-            cmd, stdin = resolve_exts_filter_template(exts_filter, ext)
+        for ext_inst in self.ext_instances:
+            cmd, stdin = resolve_exts_filter_template(exts_filter, ext_inst)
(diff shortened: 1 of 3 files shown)


boegel:
@akesandgren Since you reported this bug...

akesandgren:
Going in, thanks @boegel!

PR #3236, opened 2020-03-07
make --merge-pr take into account --pr-target-branch

boegel:
While merging easyconfigs PRs to the `2020a` branch, I'm quite annoyed that `--merge-pr` fails because the target branch isn't `develop`.

So, let's make `--merge-pr` honor `--pr-target-branch`, so we can merge PRs that target `2020a` using `eb --pr-target-branch 2020a --merge-pr`...

Of course the default is still `develop` (which is the default value for `--pr-target-branch`).


akesandgren:
Going in, thanks @boegel!

PR #3238, opened 2020-03-09
make tests use easybuilders/testrepository rather than hpcugent/testrepository after it was moved

boegel:
https://github.com/hpcugent/testrepository was moved out of the `hpcugent` organization to https://github.com/easybuilders/testrepository .

Even though this doesn't break the tests (because GitHub does an auto-redirect), it's better to change the tests accordingly...


migueldiascosta:
Going in, thanks @boegel!

PR #3242, opened 2020-03-11
Change the way that set_default_module get module paths

### LINKED ISSUE #3241: set default version not working with Categorized Module Naming Scheme
**Enviroment**
 - Lua modules
 - Categorized naming scheme

**Issue**
Running `eb HPL-2.3-foss-2019b.eb --robot --module-naming-scheme=CategorizedModuleNamingScheme --prefix=/home/jenkins/teste/prefix/  --set-default-module` fails to create the default symlink. 

**Probable cause** 
This was propably added with #3021 or with some change in the naming scheme.
When using  CategorizedModuleNamingScheme the `EASYBUILD_INSTALLPATH_MODULES/<moduleclass>` is becoming `EASYBUILD_INSTALLPATH_MODULES/<moduleclass>/<moduleclass>` and for some reason the setdefault method is trying to create for the first.
So the issue is either:
- `EASYBUILD_INSTALLPATH_MODULES/<moduleclass>/<moduleclass>` should not happen
- Set default method should try to link to `EASYBUILD_INSTALLPATH_MODULES/<moduleclass>/<moduleclass>`

thiagoabdo:
possible fix #3241

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -1373,9 +1373,8 @@ class ModuleGeneratorLua(ModuleGenerator):
             mod_symlink_paths = []
 
         for mod_symlink_path in mod_symlink_paths:
-            mod_dir_name = os.path.basename(module_dir_path)
-            mod_symlink_dir = os.path.join(install_path('mod'), mod_symlink_path, mod_dir_name)
-            create_default_symlink(mod_symlink_dir)
+            mod_dir_name = os.path.dirname(self.get_module_filepath(mod_path_suffix=mod_symlink_path))
+            create_default_symlink(mod_dir_name)
 
     def set_environment(self, key, value, relpath=False):
         """


boegel:
@thiagoabdo My apologies for not picking up on this sooner...

Can you provide some more information on this? What was going wrong exactly, how do your changes fix it?

Also, we'll need to get this covered in the tests before merging this...

thiagoabdo:
@boegel 
This is a possible fix for the problem I explained in # 3241, the main problem is that the directory that the --set-default-module is generating for the installation is not the same as the one used at installation time.

During installation the generated directory is <module_class> / <module_class> and not <module_class> as the set-default-module is trying to use.

This problem only occurs in installations using the categorized module naming scheme.

boegel:
@thiagoabdo Are you up for trying to get this fix covered by the tests?

thiagoabdo:
@boegel I can do that, but I am going to have some free time for that end of December

PR #3244, opened 2020-03-13
make --check-contrib print a warning when None is used for checksums

boegel:


migueldiascosta:
Going in, thanks @boegel!

PR #3246, opened 2020-03-14
Adding sorting to rst output of list_toolchains to match txt and none

justbennet:
When requesting `--output-format=rst`, the list of available toolchains was
not being sorted.  That output will be used to create a table of available
toolchains for the documentation and needs to be sorted.  I wrapped the
list of keys in `sorted()` to match the ouput of `eb --list-toolchains`
with no options and when `txt` is requested.

### DIFF
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -744,30 +744,58 @@ def list_toolchains_rst(tcs):
     """ Returns overview of all toolchains in rst format """
     title = "List of known toolchains"
 
-    # figure out column names
-    table_titles = ['name', 'compiler', 'MPI']
-    for tc in tcs.values():
-        table_titles.extend(tc.keys())
+    # Specify the column names for the table
+    table_titles = ['name', 'compiler', 'MPI', 'BLAS', 'FFT']
 
+    # Set up column name : display name pairs
     col_names = {
-        'COMPILER_CUDA': 'CUDA compiler',
-        'SCALAPACK': 'ScaLAPACK',
+        'name': 'Name',
+        'compiler': 'Compiler(s)',
+        'BLAS': "BLAS/LAPACK/ScaLAPACK",
     }
 
+    # removes duplicate entries
     table_titles = nub(table_titles)
 
+    # Initialize an empty list of lists for the table data
     table_values = [[] for i in range(len(table_titles))]
-    table_values[0] = ['**%s**' % tcname for tcname in sorted(tcs.keys())]
 
+    # Fill in the first column with the names of the toolchains
+    table_values[0] = ['**%s**' % tcname for tcname in tcs.keys()]
+
+    # Fill in the other table values
     for idx in range(1, len(table_titles)):
         for tc in tcs.values():
-            table_values[idx].append(', '.join(tc.get(table_titles[idx].upper(), [])))
+            table_values[idx].append(', '.join(tc.get(table_titles[idx].upper(), ["(*none*)"])))
 
+    # We want to combine the display for linear algebra, so create a new list
+    # with a list of BLAS, LAPACK, and ScaLAPACK included.
+    blas = []
+    for tc in tcs.values():
+        entry = list(tc.get('BLAS', ['(*none*)']))
+        entry.extend(list(tc.get('LAPACK', [])))
+        entry.extend(list(tc.get('SCALAPACK', [])))
+        blas.append(", ".join(nub(entry)))
+
+    # Put the combined BLAS/LAPACK/ScaLAPACK entry in the proper column
+    # of table_values
+    table_values[3] = blas
+
+    # To sort, we need to first transpose the column major list, sort on
+    # the toolchain name, then transpose back.
+    tmp_values = list(map(list, zip(*table_values)))
+    sorted_values = sorted(tmp_values, key=lambda x: str.lower(x[0]))
+    table_values = list(map(list, zip(*sorted_values)))
+
+    # Set the table titles to the pretty ones
     table_titles = [col_names.get(col, col) for col in table_titles]
+
+    # Pass the data to the rst formatter, wich is returned as a list, each element
+    # is an rst formatted text row.
     doc = rst_title_and_table(title, table_titles, table_values)
 
+    # Make a string with line endings suitable to write to document file
     return '\n'.join(doc)
-
 
 def list_toolchains_txt(tcs):
     """ Returns overview of all toolchains in txt format """
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -745,47 +745,55 @@ def list_toolchains_rst(tcs):
     title = "List of known toolchains"
 
     # Specify the column names for the table
-    table_titles = ['name', 'compiler', 'MPI', 'BLAS', 'FFT']
+    table_titles = ['name', 'COMPILER', 'MPI', 'linalg', 'FFT']
 
     # Set up column name : display name pairs
     col_names = {
         'name': 'Name',
         'compiler': 'Compiler(s)',
-        'BLAS': "BLAS/LAPACK/ScaLAPACK",
+        'linalg': "Linear algebra",
     }
 
-    # removes duplicate entries
-    table_titles = nub(table_titles)
+    # Create sorted list of toolchain names
+    sorted_tc_names = sorted(tcs.keys(), key=str.lower)
+
+    # Create text placeholder to use for missing entries.
+    # Not good documentation style to italicize the parentheses enclosing
+    # italic text unless surrouning text is also italic.
+    none_txt = '(*none*)'
 
     # Initialize an empty list of lists for the table data
     table_values = [[] for i in range(len(table_titles))]
 
-    # Fill in the first column with the names of the toolchains
-    table_values[0] = ['**%s**' % tcname for tcname in tcs.keys()]
-
-    # Fill in the other table values
-    for idx in range(1, len(table_titles)):
-        for tc in tcs.values():
-            table_values[idx].append(', '.join(tc.get(table_titles[idx].upper(), ["(*none*)"])))
-
-    # We want to combine the display for linear algebra, so create a new list
-    # with a list of BLAS, LAPACK, and ScaLAPACK included.
-    blas = []
-    for tc in tcs.values():
-        entry = list(tc.get('BLAS', ['(*none*)']))
-        entry.extend(list(tc.get('LAPACK', [])))
-        entry.extend(list(tc.get('SCALAPACK', [])))
-        blas.append(", ".join(nub(entry)))
-
-    # Put the combined BLAS/LAPACK/ScaLAPACK entry in the proper column
-    # of table_values
-    table_values[3] = blas
-
-    # To sort, we need to first transpose the column major list, sort on
-    # the toolchain name, then transpose back.
-    tmp_values = list(map(list, zip(*table_values)))
-    sorted_values = sorted(tmp_values, key=lambda x: str.lower(x[0]))
-    table_values = list(map(list, zip(*sorted_values)))
+    for col_id, col_name in enumerate(table_titles):
+        if col_name == 'name':
+            # toolchain names column gets bold face entry
+            table_values[col_id] = ['**%s**' % tcname for tcname in sorted_tc_names]
+        else:
+            for tc_name in sorted_tc_names:
+                tc = tcs[tc_name]
+                # Cray is a special snowflake
+                if tc_name.find('Cray') >= 0:
+                    if col_name == 'COMPILER':
+                        entry = 'PrgEnv-cray'
+                    elif col_name == 'MPI':
+                        entry = 'cray-MPI'
+                    elif col_name == 'linalg':
+                        entry = 'cray-libsci'
+                # Combine the linear algebra libraries into a single column
+                elif col_name == 'linalg':
+                    linalg = []
+                    for col in ['BLAS', 'LAPACK', 'SCALAPACK']:
+                        linalg.extend(tc.get(col, [none_txt]))
+                    # Doing this by iterating over the columns results in multiple 'none'
+                    # entries, so reduce those to one.
+                    linalg = nub(linalg)
+                    entry = ', '.join(linalg)
+                else:
+                    # for other columns, we can grab the values via 'tc'
+                    # key = col_name
+                    entry = ', '.join(tc.get(col_name, [none_txt]))
+                table_values[col_id].append(entry)
 
     # Set the table titles to the pretty ones
     table_titles = [col_names.get(col, col) for col in table_titles]
@@ -796,6 +804,7 @@ def list_toolchains_rst(tcs):
 
     # Make a string with line endings suitable to write to document file
     return '\n'.join(doc)
+
 
 def list_toolchains_txt(tcs):
     """ Returns overview of all toolchains in txt format """
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -757,7 +757,7 @@ def list_toolchains_rst(tcs):
     table_titles = nub(table_titles)
 
     table_values = [[] for i in range(len(table_titles))]
-    table_values[0] = ['**%s**' % tcname for tcname in tcs.keys()]
+    table_values[0] = ['**%s**' % tcname for tcname in sorted(tcs.keys())]
 
     for idx in range(1, len(table_titles)):
         for tc in tcs.values():
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -777,7 +777,7 @@ def list_toolchains_rst(tcs):
                     if col_name == 'COMPILER':
                         entry = 'PrgEnv-cray'
                     elif col_name == 'MPI':
-                        entry = 'cray-MPI'
+                        entry = 'cray-mpich'
                     elif col_name == 'linalg':
                         entry = 'cray-libsci'
                 # Combine the linear algebra libraries into a single column
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -750,17 +750,15 @@ def list_toolchains_rst(tcs):
     # Set up column name : display name pairs
     col_names = {
         'name': 'Name',
-        'compiler': 'Compiler(s)',
+        'COMPILER': 'Compiler(s)',
         'linalg': "Linear algebra",
     }
 
     # Create sorted list of toolchain names
     sorted_tc_names = sorted(tcs.keys(), key=str.lower)
 
-    # Create text placeholder to use for missing entries.
-    # Not good documentation style to italicize the parentheses enclosing
-    # italic text unless surrouning text is also italic.
-    none_txt = '(*none*)'
+    # Create text placeholder to use for missing entries
+    none_txt = '*(none)*'
 
     # Initialize an empty list of lists for the table data
     table_values = [[] for i in range(len(table_titles))]
@@ -772,10 +770,9 @@ def list_toolchains_rst(tcs):
         else:
             for tc_name in sorted_tc_names:
                 tc = tcs[tc_name]
-                # Cray is a special snowflake
-                if tc_name.find('Cray') >= 0:
+                if 'cray' in tc_name.lower():
                     if col_name == 'COMPILER':
-                        entry = 'PrgEnv-cray'
+                        entry = ', '.join(tc[col_name.upper()])
                     elif col_name == 'MPI':
                         entry = 'cray-mpich'
                     elif col_name == 'linalg':
@@ -784,11 +781,8 @@ def list_toolchains_rst(tcs):
                 elif col_name == 'linalg':
                     linalg = []
                     for col in ['BLAS', 'LAPACK', 'SCALAPACK']:
-                        linalg.extend(tc.get(col, [none_txt]))
-                    # Doing this by iterating over the columns results in multiple 'none'
-                    # entries, so reduce those to one.
-                    linalg = nub(linalg)
-                    entry = ', '.join(linalg)
+                        linalg.extend(tc.get(col, []))
+                    entry = ', '.join(linalg) or none_txt
                 else:
                     # for other columns, we can grab the values via 'tc'
                     # key = col_name
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -745,13 +745,13 @@ def list_toolchains_rst(tcs):
     title = "List of known toolchains"
 
     # Specify the column names for the table
-    table_titles = ['name', 'COMPILER', 'MPI', 'linalg', 'FFT']
+    table_titles = ['NAME', 'COMPILER', 'MPI', 'LINALG', 'FFT']
 
     # Set up column name : display name pairs
     col_names = {
-        'name': 'Name',
+        'NAME': 'Name',
         'COMPILER': 'Compiler(s)',
-        'linalg': "Linear algebra",
+        'LINALG': "Linear algebra",
     }
 
     # Create sorted list of toolchain names
@@ -764,7 +764,7 @@ def list_toolchains_rst(tcs):
     table_values = [[] for i in range(len(table_titles))]
 
     for col_id, col_name in enumerate(table_titles):
-        if col_name == 'name':
+        if col_name == 'NAME':
             # toolchain names column gets bold face entry
             table_values[col_id] = ['**%s**' % tcname for tcname in sorted_tc_names]
         else:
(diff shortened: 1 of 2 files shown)


ocaisa:
This fix only sorts the first column but leaves the other columns unchanged (and, on a side note you can only see the problem in Python2). I can't figure out a good way of doing this though...

ocaisa:
Ok, this seems to work:
```
     table_values[0] = ['**%s**' % tcname for tcname in sorted(tcs.keys())]      

     for idx in range(1, len(table_titles)):                                     
         for _, tc in sorted(tcs.items()):     
```

justbennet:
@ocaisa Thanks for finding that it didn't work in Python 2 (I think).

It seems like this shouldn't take two sorts, somehow.  It also seems like a common approach for both `rst` and `txt` would be desirable, at least for sorting the entries, as it's the same data.  Maybe that's not feasible because of the extra columns.

Having just stared at it a bit, it looks like it's producing two colums with the same compiler information, perhaps that redundancy should be removed, too?

I will not have time to look today, but I should be able to return to this tomorrow.

ocaisa:
The other option is to turn `tcs` into an OrderedDict type at the beginning. I agree that is probably a cleaner solution.

I should clarify that in Python3 the dictionary was already ordered so there isn't the same problem. It was only with Python 2 that I could reproduce the problem.

ocaisa:
You can get to it with
```
from easybuild.tools.py2vs3 import OrderedDict
```

justbennet:
Hi, @ocaisa and @boegel,

I've been looking at this more closely, and there seem to be multiple issues.  It appears that the data for the rst table comes out as a bunch of columns, which makes sorting all of them together a key.  I think this code might be more complex than it needs to be, too.

It looked to me as though your sorting of tcs.items() does not work because the data is extracted into lists in what seems to me to be a complex way.

If you look at the current output, does it have two columns for compiler information?  I am seeing one for 'compiler' and one for 'COMPILER', both containing the same information.  That redundancy seems like it could be eliminated. I also question the value of have a 'CUDA' column that only lists 'CUDA' as its only possible value when that information is also included in the compiler/COMPILER column.  I think I would advocate for dropping the CUDA column from the rst output as redundant, as well.

I think I have two excerpts here

    https://gist.github.com/justbennet/916fad2b2c62a02fc063cd96a25fded3

to illustrate this.

I propose that trying to refactor the code to make it simpler and more easily read and to eliminate the second compiler column and the CUDA column.

I've spent several hours looking at this, and I have an inefficient way to get this done.  I think that the scale this might attain will never make that inefficiency matter particularly, but I think that taking an alternate approach to constructing the table might end up with more readable and more easily maintained code and provide the same information.

Thoughts?

justbennet:
Actually, I just looked at https://easybuild.readthedocs.io/en/latest/eb_list_toolchains.html again, and the format of the included table is quite different from the output of `eb --list-toolchains`.
```
Toolchain     Components that comprise the toolchain
--------------------------------------------------------
Name          Compilers    MPI stack  Included Libraries
--------------------------------------------------------
ClangGCC      Clang/GCC
GCC           GCC
cgmpich       Clang/GCC    MPICH
cgmpolf       Clang/GCC    MPICH      OpenBLAS/LAPACK, ScaLAPACK(/BLACS), FFTW
```
Should the output of `eb --list-toolchains --output-format` match that in the old file currently being included, which is https://raw.githubusercontent.com/easybuilders/easybuild/develop/docs/version-specific/eb_list_toolchains.txt

That seems to have the same information as in the current output but seems more cleanly formatted and easier to read.

ocaisa:
@boegel I don't really know how the docs get generated, and I don't think I'd be the one to make decisions on how that table should look but there what comes out of this function does indeed have redundancy and the documentation page is markedly different.

boegel:
I'll pick up on this again. The current proposed change only sorts the first column, so it definitely needs a bot more work.

We can just sort the toolchain names once, and then walk through the `tcs` dict using the sorted keys (toolchain names), no need to sort twice.

The page in the documentation hasn't been updated for ages, because it's not getting updated by the `update-all-docs.sh` in https://github.com/easybuilders/easybuild/tree/develop/docs/scripts.

boegel:
Some thoughts on this:

* the table in https://easybuild.readthedocs.io/en/latest/eb_list_toolchains.html should be removed, and a pointer to https://easybuild.readthedocs.io/en/latest/version-specific/toolchains.html should be added instead;

* the table in https://easybuild.readthedocs.io/en/latest/version-specific/toolchains.html should formatted better, indeed to sorted by toolchain names (case-insensitive), exclude duplicate columns, etc.;

* w.r.t. the CUDA column, that could be made more generic to something like "`Extra`", where then mentioning `CUDA` for toolchains that include it could make sense;

* maybe *`(none)`* should be mentioned in places where a toolchain has no BLAS or MPI for example, rather than just leaving it blank?

boegel:
To get a good idea of what the internal data structure looks like that holds all the toolchain details, making this temporary change in the code this is useful:

```diff
diff --git a/easybuild/tools/docs.py b/easybuild/tools/docs.py
index 1d1342e2c..a5d6a8ae1 100644
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -744,6 +744,9 @@ def list_toolchains_rst(tcs):
     """ Returns overview of all toolchains in rst format """
     title = "List of known toolchains"

+    import pprint
+    pprint.pprint(tcs)
+
     # figure out column names
     table_titles = ['name', 'compiler', 'MPI']
     for tc in tcs.values():
```

With that in place, `eb --list-toolchains --output-format=rst` produces something like:

```
$ eb --list-toolchains --output-format=rst
...
 'GCC': {'COMPILER': ['GCC']},
...
 'foss': {'BLACS': ['BLACS'],
          'BLAS': ['OpenBLAS'],
          'COMPILER': ['GCC'],
          'FFT': ['FFTW'],
          'LAPACK': ['OpenBLAS'],
          'MPI': ['OpenMPI'],
          'SCALAPACK': ['ScaLAPACK']},
...
 'gompi': {'COMPILER': ['GCC'], 'MPI': ['OpenMPI']},
...

```

It would be nice if it could produce output like this:

```
List of known toolchains
------------------------

============    =================    ===========    ========    =====================    ==========    
name            compilers            MPI            BLAS        LAPACK                   FFT
============    =================    ===========    ========    =====================    ==========    
**foss**        GCC                  OpenMPI        OpenBLAS    OpenBLAS, ScaLAPACK      FFTW
**fosscuda**    GCC, CUDA            OpenMPI        OpenBLAS    OpenBLAS, ScaLAPACK      FFTW
**GCC**         GCC                  *(none)*       *(none)*    *(none)*                 *(none)*        
**gompi**       GCC                  OpenMPI        *(none)*    *(none)*                 *(none)*        
**intel**       icc, ifort           impi           imkl        imkl                     imkl
**intel**       icc, ifort, CUDA     impi           imkl        imkl                     imkl
============    =================    ===========    ========    =====================    ==========
```

justbennet:
Ace!

I'll try to make that so.

justbennet:
Please correct me if I am wrong, but if there is no BLAS, then there is no LAPACK, and if there is no LAPACK, then there can be no ScaLAPACK.  Those are all linear algebra, so I took the liberty of combining them into a single column.  I have something now that is modified to produce a table that begins with this.
```
List of known toolchains
------------------------

================    ================    ===========    =======================    ========
Name                Compiler(s)         MPI            BLAS/LAPACK/ScaLAPACK      FFT     
================    ================    ===========    =======================    ========
**cgmpich**         Clang, GCC          MPICH          (*none*)                   (*none*)
**cgmpolf**         Clang, GCC          MPICH          OpenBLAS, ScaLAPACK        FFTW
**cgmvapich2**      Clang, GCC          MVAPICH2       (*none*)                   (*none*)
**cgmvolf**         Clang, GCC          MVAPICH2       OpenBLAS, ScaLAPACK        FFTW
**cgompi**          Clang, GCC          OpenMPI        (*none*)                   (*none*)
**cgoolf**          Clang, GCC          OpenMPI        OpenBLAS, ScaLAPACK        FFTW
**ClangGCC**        Clang, GCC          (*none*)       (*none*)                   (*none*)
**CrayCCE**         PrgEnv-cray                                                   (*none*)
**CrayGNU**         PrgEnv-gnu                                                    (*none*)
**CrayIntel**       PrgEnv-intel                                                  (*none*)
**CrayPGI**         PrgEnv-pgi                         (*none*)                   (*none*)
**foss**            GCC                 OpenMPI        OpenBLAS, ScaLAPACK        FFTW
. . . .
```
So, you get no duplicate compiler column, no redundant CUDA column, sorted case-insensitive on name.

I do not know what is up with the Cray entries.  I am using
```
tc.get('BLAS', ['(*none*)'])
```
and that works as expected with `ClangGCC` in the entry above `CrayCCE`, but not with the `CrayCCE` entry.  Perhaps it is because the `ClangGCC` does not have those keys defined, whereas the `CrayCCE` entry does, but they are empty?  The pretty print dump for those two looks like this.

```
{'ClangGCC': {'COMPILER': ['Clang', 'GCC']},
 'CrayCCE': {'BLACS': [],
             'BLAS': [],
             'COMPILER': ['PrgEnv-cray'],
             'LAPACK': [],
             'MPI': [],
             'SCALAPACK': []},
```

In any event, I have something that I believe is close to what you want, and seems to mostly work.  I tested with python 2.7.5 and 3.6.8.

```
[bennet@studentpc2 ~]$ eb --list-toolchains --output-format=rst
>> Considering 'python'...
>> 'python' version: 2.7.5, which matches Python 2 version requirement (>= 2.6)
>> Selected Python command: python (/usr/bin/python)
>> python -m easybuild.main --list-toolchains --output-format=rst
. . . .

[bennet@studentpc2 ~]$ eb --list-toolchains --output-format=rst
>> Considering 'python3'...
>> 'python3' version: 3.6.8, which matches Python 3 version requirement (>= 3.5)
>> Selected Python command: python3 (/usr/bin/python3)
>> python3 -m easybuild.main --list-toolchains --output-format=rst
```

If this looks good enough, I can update the PR with this version of `docs.py`.

justbennet:
Full examples of two alternate formats are at.

https://gist.github.com/justbennet/f5a405a61665feef7bde1df6b39ae5fc

easybuild/tools/docs.py
```diff
@@ -744,31 +744,59 @@ def list_toolchains_rst(tcs):
...
+            table_values[idx].append(', '.join(tc.get(table_titles[idx].upper(), ["(*none*)"])))
 
+    # We want to combine the display for linear algebra, so create a new list
+    # with a list of BLAS, LAPACK, and ScaLAPACK included.
+    blas = []
+    for tc in tcs.values():
+        entry = list(tc.get('BLAS', ['(*none*)']))
+        entry.extend(list(tc.get('LAPACK', [])))
+        entry.extend(list(tc.get('SCALAPACK', [])))
+        blas.append(", ".join(nub(entry)))
```
boegel, line 778: I would collapse this with the for loop above, and also handle the first "name" column in the same loop.

Also, I would first create a sorted list of toolchain names, which avoids the transposing below.

```python
# first obtained a sorted list of toolchain names
sorted_tc_names = sorted(tc.keys(), key=str.lower)

# fallback value, to avoid empty table elements
none = '*(none)*'

for col_id, col_name enumerate(table_titles):
    if col_name == 'name':
        # toolchain names column
        table_values[col_id] = ['**%s**' % tcname for tcname in sorted_tc_names]
    else:
        for tc_name in sorted_tc_names:
            tc = tcs[tc_name]
            if col_name == 'linalg':
                # column for linear algebra toolchain elements is combination of BLAS/LAPACK/ScaLAPACK values
                entry = ', '.join(tc.get(key) or none for key in ['BLAS', 'LAPACK', 'SCALAPACK'])
            else:
                # for other columns, we can grab the values via 'tc', using the upper case column name as key
                key = col_name.upper()
                entry = ', '.join(tc.get(key)) or none

            table_values[col_id].append(entry)

File: easybuild/tools/docs.py [hunk 2]
```diff
@@ -744,31 +744,59 @@ def list_toolchains_rst(tcs):
...
     table_titles = [col_names.get(col, col) for col in table_titles]
+
+    # Pass the data to the rst formatter, wich is returned as a list, each element
+    # is an rst formatted text row.
     doc = rst_title_and_table(title, table_titles, table_values)
 
+    # Make a string with line endings suitable to write to document file
     return '\n'.join(doc)
 
-
```
boegel, line 771: style-wise, it's intentional to have two empty lines in between function definitions, so please leave this in place

justbennet: Restored.

easybuild/tools/docs.py
justbennet: This took a bit of massaging, but thanks for the framework.  Hopefully the new version is OK.

easybuild/tools/docs.py
```diff
@@ -744,31 +744,59 @@ def list_toolchains_rst(tcs):
...
+    # Specify the column names for the table
+    table_titles = ['name', 'compiler', 'MPI', 'BLAS', 'FFT']
 
+    # Set up column name : display name pairs
     col_names = {
-        'COMPILER_CUDA': 'CUDA compiler',
-        'SCALAPACK': 'ScaLAPACK',
+        'name': 'Name',
+        'compiler': 'Compiler(s)',
+        'BLAS': "BLAS/LAPACK/ScaLAPACK",
```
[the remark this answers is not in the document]
justbennet: Changed the table heading to 'Linear algebra' and used `linalg` internal to the script.

easybuild/tools/docs.py
```diff
@@ -744,31 +744,59 @@ def list_toolchains_rst(tcs):
...
     }
 
+    # removes duplicate entries
     table_titles = nub(table_titles)
```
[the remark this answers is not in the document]
justbennet: Removed

easybuild/tools/docs.py
```diff
@@ -744,31 +744,59 @@ def list_toolchains_rst(tcs):
     """ Returns overview of all toolchains in rst format """
     title = "List of known toolchains"
 
-    # figure out column names
-    table_titles = ['name', 'compiler', 'MPI']
-    for tc in tcs.values():
-        table_titles.extend(tc.keys())
+    # Specify the column names for the table
+    table_titles = ['name', 'compiler', 'MPI', 'BLAS', 'FFT']
```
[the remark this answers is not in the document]
justbennet: Removed.

easybuild/tools/docs.py
```diff
@@ -744,28 +744,65 @@ def list_toolchains_rst(tcs):
...
-    table_titles = nub(table_titles)
+    # Create sorted list of toolchain names
+    sorted_tc_names = sorted(tcs.keys(), key=str.lower)
 
-    table_values = [[] for i in range(len(table_titles))]
-    table_values[0] = ['**%s**' % tcname for tcname in tcs.keys()]
+    # Create text placeholder to use for missing entries.
+    # Not good documentation style to italicize the parentheses enclosing
+    # italic text unless surrouning text is also italic.
+    none_txt = '(*none*)'
```
boegel, line 763: also making the parentheses italic was deliberate actually, I think it looks awkward otherwise...

compare '*(none)*' (fully italic) with '(*none*)' (partial italic)

anyway, this is [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality), definitely not a blocker for this PR :)

easybuild/tools/docs.py
```diff
@@ -744,28 +744,65 @@ def list_toolchains_rst(tcs):
     """ Returns overview of all toolchains in rst format """
     title = "List of known toolchains"
 
-    # figure out column names
-    table_titles = ['name', 'compiler', 'MPI']
-    for tc in tcs.values():
-        table_titles.extend(tc.keys())
+    # Specify the column names for the table
+    table_titles = ['name', 'COMPILER', 'MPI', 'linalg', 'FFT']
```
boegel, line 748: Should be `'compiler'` to ensure correct column name is used in the end, no?

Now you get `COMPILER` as column name, I think...

easybuild/tools/docs.py
```diff
@@ -744,28 +744,65 @@ def list_toolchains_rst(tcs):
...
+    for col_id, col_name in enumerate(table_titles):
+        if col_name == 'name':
+            # toolchain names column gets bold face entry
+            table_values[col_id] = ['**%s**' % tcname for tcname in sorted_tc_names]
+        else:
+            for tc_name in sorted_tc_names:
+                tc = tcs[tc_name]
+                # Cray is a special snowflake
+                if tc_name.find('Cray') >= 0:
+                    if col_name == 'COMPILER':
```
boegel, line 777: also change to `'compiler'` here

easybuild/tools/docs.py
```diff
@@ -744,28 +744,65 @@ def list_toolchains_rst(tcs):
...
 
+    for col_id, col_name in enumerate(table_titles):
+        if col_name == 'name':
+            # toolchain names column gets bold face entry
+            table_values[col_id] = ['**%s**' % tcname for tcname in sorted_tc_names]
+        else:
+            for tc_name in sorted_tc_names:
+                tc = tcs[tc_name]
+                # Cray is a special snowflake
+                if tc_name.find('Cray') >= 0:
```
boegel, line 776:
```python
if 'cray' in tc_name.lower():

File: easybuild/tools/docs.py [hunk 10]
```diff
@@ -744,28 +744,65 @@ def list_toolchains_rst(tcs):
...
+                    if col_name == 'COMPILER':
+                        entry = 'PrgEnv-cray'
```
boegel, line 778: This won't always be `PrgEnv-cray`, only for `CrayCCE` this is correct.
You can grab the correct name via `tc` though:

```python
entry = tc[col_name.upper()]

File: easybuild/tools/docs.py [hunk 11]
```diff
@@ -744,28 +744,65 @@ def list_toolchains_rst(tcs):
...
+                    for col in ['BLAS', 'LAPACK', 'SCALAPACK']:
+                        linalg.extend(tc.get(col, [none_txt]))
+                    # Doing this by iterating over the columns results in multiple 'none'
+                    # entries, so reduce those to one.
+                    linalg = nub(linalg)
+                    entry = ', '.join(linalg)
+                else:
+                    # for other columns, we can grab the values via 'tc'
+                    # key = col_name
+                    entry = ', '.join(tc.get(col_name, [none_txt]))
```
boegel, line 795: use `col_name.upper()` as key (important w.r.t. changing `'COMPILER'` to `'compiler'` above):

```python
key = col_name.upper()
entry = ', '.join(tc.get(key, [none_txt]))
```

or

```python
entry = ', '.join(tc.get(col_name.upper(), [none_txt]))

File: easybuild/tools/docs.py [hunk 12]
```diff
@@ -744,28 +744,65 @@ def list_toolchains_rst(tcs):
...
+                        entry = 'cray-libsci'
+                # Combine the linear algebra libraries into a single column
+                elif col_name == 'linalg':
+                    linalg = []
+                    for col in ['BLAS', 'LAPACK', 'SCALAPACK']:
+                        linalg.extend(tc.get(col, [none_txt]))
+                    # Doing this by iterating over the columns results in multiple 'none'
+                    # entries, so reduce those to one.
+                    linalg = nub(linalg)
+                    entry = ', '.join(linalg)
```
boegel, line 791: alternate approach:

```python
linalg = []
for col in ['BLAS', 'LAPACK', 'SCALAPACK']:
    linalg.extend(tc.get(col, []))
entry = ', '.join(linalg) or none_txt
```

(the `join` will give empty string `''` if `linalg` is an empty list)

easybuild/tools/docs.py
justbennet: Ah, you are right.  I think table_titles is OK, but the change should be to
```
    col_names = {
        'name': 'Name',
        'COMPILER': 'Compiler(s)',
        'linalg': "Linear algebra",
    }
```
to match the change you requested to `table_titles` before.  Sorry.  Will do that.

easybuild/tools/docs.py
justbennet: I was trained as a typesetter and spent years as a proofreader and (English) copy editor, and I have the opposite reaction to _(none)_ and (_none_)
https://english.stackexchange.com/questions/9878/should-punctuation-surrounding-italicised-words-be-italicised
But, this is your baby, so I will do it your way.

easybuild/tools/docs.py
justbennet: I think that would unfix it.  If it is upper in `table_titles` it should be upper here.  I think right now it is consistently `COMPILER` and it seems to work, so should remain that way.  No?
```
    table_titles = ['name', 'COMPILER', 'MPI', 'linalg', 'FFT']
. . .
    col_names = {
        'name': 'Name',
        'COMPILER': 'Compiler(s)',
        'linalg': "Linear algebra",
    }
. . . 
                    if col_name == 'COMPILER':
```

easybuild/tools/docs.py
justbennet: Will change.

easybuild/tools/docs.py
justbennet: That returns a list from which the string must be extracted for printing.  If this is done for Cray the same way it is for the other toolchains, then that would be
```
entry = ', '.join(tc[col_name.upper()])
```
Is that how you would like it done?

easybuild/tools/docs.py
justbennet: Your first comment wants to change `COMPILER` to `compiler`.  I think this is a matter of consistency throughout the block.  Whatever is used in `table_titles` should be used throughout for things to work properly.  The column heading was coming out `COMPILER` because I initially failed to make it a matching 'COMPILER' in `col_names`.  That is fixed, and it is `COMPILER` everywhere, which works.

easybuild/tools/docs.py
justbennet: Changed.

easybuild/tools/docs.py
boegel: Oh boy, please don't agree with me if you're better trained than me in this stuff! :)

I was just going by feeling, if there are clear arguments against using `*(none)*`, then please don't!

easybuild/tools/docs.py
boegel: I think it looks awkward to have `name` vs `COMPILER` in `table_titles`.

For `MPI` & `FFT` it's OK, since those are abbreviations.

justbennet: They are just placeholders internal to the code.  How about _all_ of the entries in `table_titles` go upper, and the `col_names` are used to set the headings in the display?  I tested that now, and that works, and that will make things consistent?

easybuild/tools/docs.py
boegel: It should definitely be consistent for it to work correctly, that's also why we need a test... ;)

Using `'compiler'` instead everywhere (and `.upper()` where needed seems cleaner to me, but again, it's mostly bikeshedding...

easybuild/tools/docs.py
boegel: Yes, I overlooked the list part, sorry.

easybuild/tools/docs.py
justbennet: If it would lead to inconsistencies elsewhere, it's fine.  It's just a convention.  ;-)  Just trying to explain why my instincts are what they are.  And that it was not entirely idiosyncratic.  ;-)

I suspect that most everyone else on the project would do what you did and put the *s around the parentheses, and it's probably more important to be consistent throughout.  I changed them, so let's not change back and get this out the door.   :-)

easybuild/tools/docs.py
boegel: It sort of works by accident now, without the `.upper()`, because `name` and `linalg` are handled separate as special cases...

If someone would change `table_titles` (to use `'compiler'` rather than `'COMPILER'` for example ;) ), this would fail with a `.upper()` (and having the `.upper()` doesn't hurt).

easybuild/tools/docs.py
justbennet: Same with `linalg`, then?  So,
```
table_titles = ['NAME', 'COMPILER', 'MPI', 'LINALG', 'FFT']
. . . 
    col_names = {
        'NAME': 'Name',
        'COMPILER': 'Compiler(s)',
        'LINALG': "Linear algebra",
    }
```
and all references in the code to the text contents of `table_titles` would have the upper version.

boegel:
@justbennet I've added a test for the output produced by `--list-toolchains --output-format=rst`.

It fails now, but with good reason imho, for the following cases:

* for toolchains using `OpenBLAS`, `OpenBLAS` is listed twice in the `linalg` column, we should try and avoid that (missing a call `nub` somewhere?);
* for toolchains using `imkl`, `imkl` is listed 3 times in the `linalg` column;
* for the `system` toolchain, we're missing a `(none)` in the compiler column;

To run the test yourself locally, try using this command:

```
python -O -m test.framework.options list_toolchains
```

If the tests for `--list-toolchains` pass, you should see output like:

```
Ran 2 tests in 0.931s

ok
```

If the test still fails, you'll get a `AssertionError: None is not true : Pattern '<pattern>' should be found in: `, followed by the full output produced by `--list-toolchains --output-format=rst`.

Let me know if you need any help!

easybuild/tools/docs.py
boegel: +1 for consistency, so going all upper is fine by me

boegel:
Thanks a lot for your efforts on this @justbennet!

Issue #3223, opened 2020-03-03
Rewrite bootstrap script using pip


Flamefire:
There are currently 15 open issues that the boostrap script is broken on some environments: https://github.com/easybuilders/easybuild-framework/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+bootstrap+in%3Atitle

One of the main reasons is the patched distribute and setuptools package used. Furthermore it makes the whole install process very hacky and has several issues I described in a partial fix PR: #3211 

Summary: 
- PYTHONPATH and sys.path not kept in sync leading to different behavior based on the setuptools used (spawned by loading it inside the current env [sys.path] but it may spawn a new process which uses PYTHONPATH)
- Resulting instillation of unwanted versions is possible (partially due to the above but also because EB < 4 uses vsc-utils with a minimum version but requires a maximum version [manually installed on boostrap])
- Using `setup.py install` is deprecated as is using the ancient disutils, easy_install or ez_setup.py, see e.g. https://github.com/pypa/setuptools/issues/581
- The only valid way of installing something is `pip install`

Proposal: Use pip and virtualenv for installation

I created a Proof Of Concept in a branch: https://github.com/Flamefire/easybuild-framework/blob/test_pip_for_bootstrap/easybuild/scripts/bootstrap_eb.py

Details:
- If `pip` exists, use it to install a specific, up-to-date pip into a temporary folder, else use `get-pip.py` (the official way to install pip) to do the same
- Use this pip to add setuptools and wheels packages required to build source packages (stage 0)
- install easybuild and its prerequisites into a new temporary folder (stage 1)
- reset PYTHONPATH to stage0 (maybe not even required if the EC is correctly handled)
- import easybuild from stage1 and build EasyBuild from generated EasyConfig (stage 2)

Potential problem: System installed packages. If a package exists, `pip` will uninstall it before installing the new one. There is `--ignore-installed` which avoids this, but this will also ignore existing dependencies leading to the following situation:
- `pip install vsc-install<0.11.4 --ignore-installed` -> No dependencies, all good
- `pip install vsc-base<2.9.0  --ignore-installed` -> Ignores installed vsc-install and pulls in the latest available one because vsc-base requests `vsc-install>=0.9.0`
- Finally fails due to incompatibilites

[...]

PR #3211, opened 2020-02-24
[bootstrap] Fix building setuptools

Flamefire:
Fixes #3200
Fixes #2428

This started as a fix for #3200 but falls a bit short of solving the installation for a Python3 environment without setuptools already installed.

What works/got fixed:
- Installation on Python2
  - Use https for PyPi (they dropped support for http) to fix installing/using distribute
- Use correct string type for setup_distribute unpacked blob (#3200)
- Reduce some noise when checking for setuptools
- Use system toolchain instead of dummy toolchain for EB >= 4.0

Problems to be resolved in follow-up:   
TLDR: Don't use easy_install anymore

The `distribute` package used installs an incredibly ancient version of setuptools which doesn't work out of the box anymore (see above and the need for patching setuptools and the setup_distribute script). There was/is even an ez_setup.py script which seems to superseed this: https://bootstrap.pypa.io/ez_setup.py   
However even that is deprecated: https://github.com/pypa/setuptools/issues/581

I started to use the https://bootstrap.pypa.io/get-pip.py to install pip (if missing) and then use `pip` to install setuptools just as `distribute` would have. But I hit another, currently hidden problem: The EB `bootstrap` script has a function `prep` which sets up PYTHONPATH for some temporary location where `stage{0,1,2}` are going to be installed to. It then proceeds to **import** easy_install and start installing dependencies and packages of EB (e.g. vsc-*) into that environment. The problem: This environment is **not** part of the current environment (`sys.path`). Hence easy_install fails to find the dependencies it just installed and tries to redownload them, which then fails because it doesn't find the `wheels` package which is also part of the stage0 environment (and would be part of a `system()` invoked environment) This is (at least partially) due to the mismatch of the environments, i.e. python core libs check for support of features and then run `system()` but those features are not part of that sub-process to to mismatch of PYTHONPATH vs sys.path.

IMO the easiest solution would be just use pip run via `system` and install pip if it is missing. Potentially (very probably required) also upgrade pip and setuptools into the stage1 directory.

Summary of proposed EB bootstrap procedure:
- Prepare stage1 environment in new temporary folder
  - Check for pip and if non-existant install into stage1 via get-pip script
  - Upgrade pip and setuptools with `--prefix` set to the temporary folder using pip
  - Install EB and its dependencies into that stage1 folder using pip
- In stage2
  - Use `system("eb ...")` to install the real EasyBuild via stage1 EB
[...]

PR #3247, opened 2020-03-15
Changing 'dummy' toolchain in bootstrap to 'system'

justbennet:
The `dummy` toolchain was being used in the bootstrap script and generating a deprecation warning.

Changed to `system`.  Warning is no longer issued.


ocaisa:
Hi @justbennet thanks for the PR, there is actually an open PR #3211 that also fixes this along with some other issues with the bootstrap script. That PR takes into account that people may still want to bootstrap in an older version (3.x) of EB where `system` is not understood.

In general, there are quite a few open issues with the bootstrap script (see #3223) and we probably need a deeper fix.

justbennet:
Thanks, @ocaisa .  I will close this one then.

PR #3252, opened 2020-03-21
unset $GITHUB_TOKEN in Travis after installing token, to avoid failing test_from_pr_token_log

boegel:
`test_from_pr_token_log` was failing outside of the context of testing a PR, because the contents of the `$GITHUB_TOKEN` environment variable set in the Travis environment was included in the environment dump that is included in EasyBuild debug logs...


zao:
Going in, thanks @boegel!

PR #3258, opened 2020-03-25
Show EC names for parallel build

Flamefire:
Currently the class is show which means some generic output like the address of the object. This shows the actual name(s) of the EC/module

### DIFF
--- a/easybuild/tools/parallelbuild.py
+++ b/easybuild/tools/parallelbuild.py
@@ -68,7 +68,7 @@ def build_easyconfigs_in_parallel(build_command, easyconfigs, output_dir='easybu
     :param output_dir: output directory
     :param prepare_first: prepare by runnning fetch step first for each easyconfig
     """
-    _log.info("going to build these easyconfigs in parallel: %s", easyconfigs)
+    _log.info("going to build these easyconfigs in parallel: %s", [os.path.basename(ec['spec']) for ec in easyconfigs])
 
     active_job_backend = job_backend()
     if active_job_backend is None:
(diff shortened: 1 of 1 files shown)


easybuild/tools/parallelbuild.py
```diff
@@ -68,7 +68,7 @@ def build_easyconfigs_in_parallel(build_command, easyconfigs, output_dir='easybu
     :param output_dir: output directory
     :param prepare_first: prepare by runnning fetch step first for each easyconfig
     """
-    _log.info("going to build these easyconfigs in parallel: %s", easyconfigs)
+    _log.info("going to build these easyconfigs in parallel: %s", (ec['full_mod_name'] for ec in easyconfigs))
```
migueldiascosta: @Flamefire does this generator get iterated here?

e.g.
```python
import logging
logging.basicConfig(level=logging.INFO)
logging.info("%s", (i for i in [1,2,3]))
logging.info("%s", [i for i in [1,2,3]])
```
returns
```
INFO:root:<generator object <genexpr> at 0x7fc6b37cf550>
INFO:root:[1, 2, 3]
```

Flamefire:
```suggestion
    _log.info("going to build these easyconfigs in parallel: %s", [ec['full_mod_name'] for ec in easyconfigs])
```

In my test it did. But you are right: Better safe than sorry in case anything changes later on.

migueldiascosta:
nitpicking, especially since anything is better than the current status of showing the class object, but wouldn't `ec['spec']` (or `os.path.basename(ec['spec'])`) be more appropriate than `ec['full_mod_name']`, since it's talking about easyconfigs, not modules?

Flamefire:
Excellent idea. Updated and rebased.

migueldiascosta:
Going in, thanks @Flamefire!

PR #3259, opened 2020-03-26
Refactor mpi_cmd_for so we can just grab an MPI execution command

ocaisa:


ocaisa:
@boegel I've set this up so that it *could* be templated in an easyconfig but to do that I would need a toolchain instance for the easyconfig since I would rely on knowing the MPI family. Not sure where to do that though

test/framework/toolchain.py
```diff
@@ -948,6 +948,32 @@ def test_nosuchtoolchain(self):
         tc = self.get_toolchain('intel', version='1970.01')
         self.assertErrorRegex(EasyBuildError, "No module found for toolchain", tc.prepare)
 
+    def test_mpi_exec_nranks(self):
+        """Test mpi_exec_nranks function."""
+        self.modtool.prepend_module_path(self.test_prefix)
+
+        tc = self.get_toolchain('gompi', version='2018a')
+        tc.prepare()
+        self.assertEqual(tc.mpi_exec_nranks(nr_ranks=2), "mpirun -n 2")
```
boegel: Also check the default?

```python
self.assertEqual(tc.mpi_exec_nranks(), "mpirun -n 1")
```

And maybe make sure that we don't get a nasty traceback when a string value is passed:

```python
error_pattern = "pattern for error message goes here"
self.assertErrorRegex(EasyBuildError, error_pattern, tc.mpi_exec_nranks, '2')

File: test/framework/toolchain.py [hunk 2]
```diff
@@ -948,6 +948,32 @@ def test_nosuchtoolchain(self):
...
+        self.modtool.purge()
+
+        self.setup_sandbox_for_intel_fftw(self.test_prefix)
+        tc = self.get_toolchain('intel', version='2018a')
+        tc.prepare()
+        self.assertEqual(tc.mpi_exec_nranks(nr_ranks=2), "mpirun -n 2")
```
boegel: also check default `tc.mpi_exec_nranks()`

test/framework/toolchain.py
```diff
@@ -948,6 +948,32 @@ def test_nosuchtoolchain(self):
...
+        self.modtool.purge()
+
+        self.setup_sandbox_for_intel_fftw(self.test_prefix, imklver='10.2.6.038')
+        tc = self.get_toolchain('intel', version='2012a')
+        tc.prepare()
+
+        mpi_exec_nranks_re = re.compile("^mpirun --file=.*/mpdboot -machinefile .*/nodes -np 4")
+        self.assertTrue(mpi_exec_nranks_re.match(tc.mpi_exec_nranks(nr_ranks=4)))
```
boegel: also check default `tc.mpi_exec_nranks()`

test/framework/toolchain.py
```diff
@@ -948,6 +948,32 @@ def test_nosuchtoolchain(self):
...
+
+        # test specifying custom template for MPI commands
+        init_config(build_options={'mpi_exec_template': "mpiexec -np %(nr_ranks)s --", 'silent': True})
+        self.assertEqual(tc.mpi_exec_nranks(nr_ranks="7"), "mpiexec -np 7 --")
```
boegel: Also check default `tc.mpi_exec_nranks()`

Passing a string argument '7' here only works because your custom template uses `%(nr_ranks)s` rather than `%(nr_ranks)d`.

It probably makes sense to just also use `%(nr_ranks)s` in the default template, so both `7` and `'7'` can be passed as an argument to `mpi_exec_nranks`...

easybuild/tools/toolchain/mpi.py
```diff
@@ -168,40 +168,67 @@ def mpi_family(self):
...
+        if mpi_exec_template is None:
 
             if mpi_family == toolchain.INTELMPI:
 
                 # for old versions of Intel MPI, we need to use MPD
                 impi_ver = self.get_software_version(self.MPI_MODULE_NAME)[0]
                 if LooseVersion(impi_ver) <= LooseVersion('4.1'):
 
-                    mpi_cmds[toolchain.INTELMPI] = "mpirun %(mpdbf)s %(nodesfile)s -np %(nr_ranks)d %(cmd)s"
+                    mpi_cmds[toolchain.INTELMPI] = "mpirun %(mpdbf)s %(nodesfile)s -np %(nr_ranks)d"
```
boegel: better use `%(nr_ranks)s`

easybuild/tools/toolchain/mpi.py
```diff
@@ -168,40 +168,67 @@ def mpi_family(self):
...
+            'nr_ranks': nr_ranks,
+        }
+
+        mpi_exec_template = build_option('mpi_exec_template')
+        if mpi_exec_template:
+            self.log.info("Using specified template for MPI commands: %s", mpi_exec_template)
         else:
             # different known mpirun commands
-            mpirun_n_cmd = "mpirun -n %(nr_ranks)d %(cmd)s"
+            mpirun_n = "mpirun -n %(nr_ranks)d"
```
boegel: better use `%(nr_ranks)s`

easybuild/tools/toolchain/mpi.py
```diff
@@ -168,40 +168,67 @@ def mpi_family(self):
...
             mpi_cmds = {
-                toolchain.OPENMPI: mpirun_n_cmd,
-                toolchain.QLOGICMPI: "mpirun -H localhost -np %(nr_ranks)d %(cmd)s",
-                toolchain.INTELMPI: mpirun_n_cmd,
-                toolchain.MVAPICH2: mpirun_n_cmd,
-                toolchain.MPICH: mpirun_n_cmd,
-                toolchain.MPICH2: mpirun_n_cmd,
+                toolchain.OPENMPI: mpirun_n,
+                toolchain.QLOGICMPI: "mpirun -H localhost -np %(nr_ranks)d",
```
boegel: better use `%(nr_ranks)s`

easybuild/tools/toolchain/mpi.py
```diff
@@ -168,40 +168,67 @@ def mpi_family(self):
...
+            else:
+                self.log.info("Using specified template for MPI commands: %s", mpi_cmd_template)
+        else:
+            params['mpi_exec_nranks'] = self.mpi_exec_nranks(nr_ranks=nr_ranks)
+            mpi_cmd_template = "%(mpi_exec_nranks)s %(cmd)s"
+
+        try:
+            res = mpi_cmd_template % params
+        except KeyError as err:
+            raise EasyBuildError("Failed to complete MPI cmd template '%s' with %s: %s", mpi_cmd_template, params, err)
```
boegel: It's probably better if we explicitly check that `%(nr_ranks)s` and %(cmd)s` are both included in the `mpi_cmd_template` value?

```python
missing = []
for key in params:
    tmpl = '%(' + key + ')s'
    if tmpl not in mpi_cmd_template:
        missing.append(tmpl)
if missing:
    raise EasyBuildError("Missing templates in mpi-cmd-template value '%s': %s", mpi_cmd_template, ', '.join(missing))

**Comment by boegel** [comment]
> @boegel I've set this up so that it _could_ be templated in an easyconfig but to do that I would need a toolchain instance for the easyconfig since I would rely on knowing the MPI family. Not sure where to do that though

template values are collected via `template_constant_dict`, which is called from `EasyConfig._generate_template_values`.

You can pass down the toolchain instance from the `EasyConfig` instance down to `template_constant_dict` and put it to good use there, something like this:

```python
template_values = template_constant_dict(self, ignore=ignore, toolchain=self.toolchain)
```

and then in `template_constant_dict`:

```python
def template_constant_dict(..., toolchain=None):
    ...
    if toolchain is not None and hasattr(toolchain, 'mpi_cmd_for'):
        # use empty string as command, to only get prefix for commands to be run with mpirun
        mpi_cmd_prefix = toolchain.mpi_cmd_for('', 1)
        template_values['mpi_cmd_prefix'] = mpi_cmd_prefix

ocaisa:
> template values are collected via `template_constant_dict`, which is called from `EasyConfig._generate_template_values`.
> 
> You can pass down the toolchain instance from the `EasyConfig` instance down to `template_constant_dict` and put it to good use there, something like this:
> 
> ```python
> template_values = template_constant_dict(self, ignore=ignore, toolchain=self.toolchain)
> ```
> 
> and then in `template_constant_dict`:
> 
> ```python
> def template_constant_dict(..., toolchain=None):
>     ...
>     if toolchain is not None and hasattr(toolchain, 'mpi_cmd_for'):
>         # use empty string as command, to only get prefix for commands to be run with mpirun
>         mpi_cmd_prefix = toolchain.mpi_cmd_for('', 1)
>         template_values['mpi_cmd_prefix'] = mpi_cmd_prefix
>        
>     
> ```

I could have done this without making any other changes...I think I might just replace this PR rather row back all the other changes.

ocaisa:
Closed in favour of #3264

PR #3260, opened 2020-03-27
Toolchain hierarchy aware dump

ocaisa:
Don't you wish life could be so simple...the import required to make this work is not allowed so this requires some refactoring.


boegel:
replaced with #3261

PR #3268, opened 2020-04-06
use apt-spy2 before using apt in GitHub Actions

boegel:


lexming:
Going in, thanks @boegel !

PR #3270, opened 2020-04-08
Update copy_dir to include option to merge directories

Darkless012:
copy_dir is using shutils.copy_tree which throws error when target directory exists.
In Python 3.8 new `dirs_exist_ok` parameter was added to be able to merge folders as well.

The cleanest solution was to mimic this option with distutils copy_tree which does exactly that.

This is needed for Tarball installation in Bundles, because Tarball removes the whole folder, effectively wiping previous installed bundle components. (Fix for Tarball EB will follow)

Please Squash the merge if possible.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -40,7 +40,6 @@ Set of file tools.
 """
 import datetime
 import difflib
-import distutils
 import fileinput
 import glob
 import hashlib
@@ -52,6 +51,7 @@ import sys
 import tempfile
 import time
 import zlib
+from distutils.dir_util import copy_tree
 from xml.etree import ElementTree
 
 from easybuild.base import fancylogger
@@ -1847,7 +1847,7 @@ def copy_dir(path, target_path, force_in_dry_run=False, dirs_exist_ok=False, **k
                 # Check if there are other named arguments
                 if len(kwargs) > 0:
                     raise EasyBuildError("You can't use 'dirs_exist_ok=True' with other named arguments: %s", kwargs)
-                distutils.dir_util.copy_tree(path, target_path, preserve_symlinks=preserve_symlinks)
+                copy_tree(path, target_path, preserve_symlinks=preserve_symlinks)
 
             else:
                 # Use shutil.copytree WITHOUT 'dirs_exist_ok'
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -40,6 +40,7 @@ Set of file tools.
 """
 import datetime
 import difflib
+import distutils
 import fileinput
 import glob
 import hashlib
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1834,23 +1834,24 @@ def copy_dir(path, target_path, force_in_dry_run=False, dirs_exist_ok=False, **k
             if not dirs_exist_ok and os.path.exists(target_path):
                 raise EasyBuildError("Target location %s to copy %s to already exists", target_path, path)
 
-            # On Python >= 3.8
             if sys.version_info >= (3, 8):
-                # Use the shutil.copytree WITH 'dirs_exist_ok'
+                # on Python >= 3.8, shutil.copytree works fine, thanks to availability of dirs_exist_ok named argument
                 shutil.copytree(path, target_path, dirs_exist_ok=dirs_exist_ok, **kwargs)
 
             elif dirs_exist_ok:
-                # Get symlinks named argument and use distutils.dir_util.copy_tree instead.
+                # use distutils.dir_util.copy_tree with Python < 3.8 if dirs_exist_ok is enabled
+
+                # first get value for symlinks named argument (if any)
                 preserve_symlinks = kwargs.pop('symlinks', False)
 
-                # Check if there are other named arguments
+                # check if there are other named arguments (there shouldn't be, only 'symlinks' is supported)
                 if kwargs:
-                    raise EasyBuildError("You can't use 'dirs_exist_ok=True' with other named arguments: %s",
-                                         list(kwargs.keys()))
+                    raise EasyBuildError("Unknown named arguments passed to copy_dir with dirs_exist_ok=True: %s",
+                                         ', '.join(sorted(kwargs.keys())))
                 distutils.dir_util.copy_tree(path, target_path, preserve_symlinks=preserve_symlinks)
 
             else:
-                # Use shutil.copytree WITHOUT 'dirs_exist_ok'
+                # if dirs_exist_ok is not enabled, just use shutil.copytree
                 shutil.copytree(path, target_path, **kwargs)
 
             _log.info("%s copied to %s", path, target_path)
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -40,6 +40,7 @@ Set of file tools.
 """
 import datetime
 import difflib
+import distutils.dir_util
 import fileinput
 import glob
 import hashlib
@@ -51,7 +52,6 @@ import sys
 import tempfile
 import time
 import zlib
-from distutils.dir_util import copy_tree
 from xml.etree import ElementTree
 
 from easybuild.base import fancylogger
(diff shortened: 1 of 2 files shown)


easybuild/tools/filetools.py
```diff
@@ -1808,24 +1809,51 @@ def copy_files(paths, target_dir, force_in_dry_run=False):
...
         dry_run_msg("copied directory %s to %s" % (path, target_path))
     else:
         try:
-            if os.path.exists(target_path):
+            if not dirs_exist_ok and os.path.exists(target_path):
                 raise EasyBuildError("Target location %s to copy %s to already exists", target_path, path)
 
-            shutil.copytree(path, target_path, **kwargs)
+            # On Python >= 3.8
+            if (sys.version_info[0] == 3 and sys.version_info[1] >= 8) or sys.version_info[0] > 3:
```
boegel, line 1838: This can be simplified to `sys.version_info >= (3, 8):`

easybuild/tools/filetools.py
```diff
@@ -1808,24 +1809,51 @@ def copy_files(paths, target_dir, force_in_dry_run=False):
...
+            elif dirs_exist_ok:
+                preserve_symlinks = False
+                # Get symlinks named argument and use distutils.dir_util.copy_tree instead.
+                if 'symlinks' in kwargs:
+                    preserve_symlinks = kwargs.pop('symlinks', False)
+                # Check if there are other named arguments
+                if len(kwargs) > 0:
+                    raise EasyBuildError("You can't use 'dirs_exist_ok=True' with other named arguments: %s",
+                                         list(kwargs.keys()))
+                copy_tree(path, target_path, preserve_symlinks=preserve_symlinks)
```
boegel, line 1851: please use `distutils.dir_util.copy_tree` explicitly here, otherwise it's confusing compared to `shutil.copytree`

easybuild/tools/filetools.py
```diff
@@ -1808,24 +1809,51 @@ def copy_files(paths, target_dir, force_in_dry_run=False):
...
+                # Use the shutil.copytree WITH 'dirs_exist_ok'
+                shutil.copytree(path, target_path, dirs_exist_ok=dirs_exist_ok, **kwargs)
+
+            elif dirs_exist_ok:
+                preserve_symlinks = False
+                # Get symlinks named argument and use distutils.dir_util.copy_tree instead.
+                if 'symlinks' in kwargs:
+                    preserve_symlinks = kwargs.pop('symlinks', False)
+                # Check if there are other named arguments
+                if len(kwargs) > 0:
```
boegel, line 1848: just `if kwargs:` is enough :)

easybuild/tools/filetools.py
```diff
@@ -1808,24 +1809,51 @@ def copy_files(paths, target_dir, force_in_dry_run=False):
...
+            # On Python >= 3.8
+            if (sys.version_info[0] == 3 and sys.version_info[1] >= 8) or sys.version_info[0] > 3:
+                # Use the shutil.copytree WITH 'dirs_exist_ok'
+                shutil.copytree(path, target_path, dirs_exist_ok=dirs_exist_ok, **kwargs)
+
+            elif dirs_exist_ok:
+                preserve_symlinks = False
+                # Get symlinks named argument and use distutils.dir_util.copy_tree instead.
+                if 'symlinks' in kwargs:
+                    preserve_symlinks = kwargs.pop('symlinks', False)
```
boegel, line 1846: no need to initialize or the `if`

test/framework/filetools.py
```diff
@@ -1470,6 +1470,24 @@ def test_copy_dir(self):
...
+        self.assertTrue(sorted(os.listdir(to_copy)) == sorted(os.listdir(testdir)))
+
+        # if the directory already exists and 'dirs_exist_ok' is True and there is another named argument (ignore)
+        # we expect clean error on Python < 3.8 and pass the test on Python >= 3.8
+        # NOTE: reused ignore from previous test
+        shutil.rmtree(testdir)
+        ft.mkdir(testdir)
+        if (sys.version_info[0] == 3 and sys.version_info[1] >= 8) or sys.version_info[0] > 3:
+            ft.copy_dir(to_copy, testdir, dirs_exist_ok=True,
+                        ignore=lambda src, names: [x for x in names if '6.4.0-2.28' in x])
```
boegel, line 1484: this is more readable if you add a variable for the lambda function

test/framework/filetools.py
```diff
@@ -1470,6 +1470,24 @@ def test_copy_dir(self):
...
+        # if the directory already exists and 'dirs_exist_ok' is True, copy_dir should succeed
+        ft.copy_dir(to_copy, testdir, dirs_exist_ok=True)
+        self.assertTrue(sorted(os.listdir(to_copy)) == sorted(os.listdir(testdir)))
+
+        # if the directory already exists and 'dirs_exist_ok' is True and there is another named argument (ignore)
+        # we expect clean error on Python < 3.8 and pass the test on Python >= 3.8
+        # NOTE: reused ignore from previous test
+        shutil.rmtree(testdir)
+        ft.mkdir(testdir)
+        if (sys.version_info[0] == 3 and sys.version_info[1] >= 8) or sys.version_info[0] > 3:
```
boegel, line 1482: `sys.version_info >= (3, 8)`

easybuild/tools/filetools.py
Darkless012: I've tried but it has interesting loading mechanims:
```
ERROR: test_copy_dir (test.framework.filetools.FileToolsTest)
Test copy_dir function.
  File "test/framework/filetools.py", line 1474, in test_copy_dir
    ft.copy_dir(to_copy, testdir, dirs_exist_ok=True)
  File "easybuild/tools/filetools.py", line 1850, in copy_dir
    distutils.dir_util.copy_tree(path, target_path, preserve_symlinks=preserve_symlinks)
AttributeError: 'module' object has no attribute 'dir_util'
```

easybuild/tools/filetools.py
Darkless012: true :)

test/framework/filetools.py
Darkless012: Then I need to define function otherwise it will not be according to PEP8 .

easybuild/tools/filetools.py
Darkless012: `import distutils.dir_util` might do the trick

ocaisa:
@Darkless012 Can you please take a look at https://github.com/easybuilders/easybuild-easyblocks/pull/1989 that is covering similar ground with a different approach, I get the impression we will need to synchronise our approach among that and your upcoming tarball PR

lexming:
Having the option to merge folders is a nice addition. However, I'd like to point out that it won't solve the issue with `tarball` easyblock as it is very unlikely that two source code trees can be merged in the same folder. I proposed to add an option to unpack the tarballs in specific subdirectories in https://github.com/easybuilders/easybuild-easyblocks/pull/1989.
I think that both options are compatible and should be added to `tarball` easyblock as long as none of them is made default. @Darkless012 If there is anything I can do to help synchronize those changes to `tarball` let me know.

Darkless012:
Thank you for noticing, agreed :)
Lets discuss this in mentioned Tarball Easyblock PR.

boegel:
> However, I'd like to point out that it won't solve the issue with `tarball` easyblock as it is very unlikely that two source code trees can be merged in the same folder.

Why is that an issue? It's say it's quite uncommon that two different software packages would include the same files, and hence problems would ensue when copying the files into an already existing directory?

This is exactly what traditional package managers do by throwing everything in `/usr` for example...

Am I overlooking something? (I haven't looked at https://github.com/easybuilders/easybuild-easyblocks/pull/1989 in detail yet)

lexming:
@boegel one thing is to put all executables in a common `bin` folder or all libs in a common `lib` as it can be assumed that they will have different names. But merging whole tarballs is a different matter in my opinion as it is unpredictable if files might share the same name and relative path. I see it more closely related to what is done by  traditional package managers in `/usr/share`.

PR #3281, opened 2020-04-16
also mention CPU architecture (x86_64, POWER) in comment for test reports

Flamefire:
In https://github.com/easybuilders/easybuild-easyconfigs/pull/9830 I noted that the architecture is now included in the test report, so it's hard to tell which reports are from x86 and which are from power.

Yes there is CPU model stuff via archspec, but that needs an (undocumented?) external dependency.

This PR uses the already available `get_cpu_architecture` which is also used in various EasyBlocks to change behavior and adds it to the report.

Example: https://github.com/easybuilders/easybuild-easyconfigs/pull/10397#issuecomment-614557953

> Build succeeded for 8 out of 8 (8 easyconfigs in this PR)
taurusml1 - Linux RHEL 7.6, POWER, 8335-GTX, Python 2.7.5

and 

> Build succeeded for 8 out of 8 (8 easyconfigs in this PR)
taurusa5 - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2603 v4 @ 1.70GHz, Python 2.7.5

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -770,6 +770,7 @@ def get_system_info():
     return {
         'core_count': get_avail_core_count(),
         'total_memory': get_total_memory(),
+        'cpu_arch': get_cpu_architecture(),
         'cpu_arch_name': get_cpu_arch_name(),
         'cpu_model': get_cpu_model(),
         'cpu_speed': get_cpu_speed(),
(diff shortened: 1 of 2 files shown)


smoors:
to be fair, `archspec` is in the requirements.txt file of the framework

Flamefire:
Which is still to be installed manually. Our site doesn't and the boostraped and EB installed modules don't contain it either. (And I personally wasn't even aware of its existance until recently)

Anything wrong with this PR? IMO It is simple and a good improvement. Especially because it shows a variable which changes behavior of EasyBlock. The archspec output doesn't.

smoors:
As I said I don't mind adding this change, but I'll leave the decision to someone else (@boegel ?)

PR #3282, opened 2020-04-17
fix code style issues in easybuild.tools + add flake8 linting test

Flamefire:
First step towards getting flake8 pass for the whole repo to inspect what `autopep8` can do.

Command run: `autopep8 -r -i --max-line-length 120 easybuild/tools` and `autopep8 -a -r -i --max-line-length 120 easybuild/tools`

Note: Don't use aggressive level 2 (`-aa`) as this will change `if x == True` to `if x` which is a semantic change: https://github.com/hhatto/autopep8/issues/539


boegel:
@Flamefire This should be accompanied by running `flake8` on the list of files that should now fully comply with PEP8 in the tests, to ensure future PRs don't introduce regressions that diverge from PEP8 again?

Flamefire:
@boegel Done. Again I'm using a separate job (even a separate yaml file) to make it cleaner and run only once. For flake8 it doesn't make sense to run it with multiple pythons. latest is enough.

Also see the name. IMO "Static Analysis / python-linting" is way more concise then the rather verbose "EasyBuild framework unit tests / build" which hides the more important parts. I'd suggest renaming those to `framework / unit tests`

easybuild/tools/asyncprocess.py
```diff
@@ -65,6 +65,8 @@
 :author: Jens Timmerman (Ghent University)
 """
 
+import fcntl  # @UnresolvedImport
+import select  # @UnresolvedImport
 import errno
 import os
```
boegel, line 71: @Flamefire Please maintain alphabetical order after moving.

Also, are those `@UnresolvedImport` still needed? Those are modules from the Python std lib, so OK to remove imho.

```python
import errno
import fcntl
import os
import select
import subprocess
```

.github/workflows/linting.yml
```diff
@@ -0,0 +1,22 @@
...
+        python-version: 3.8
+
+    - name: install Python packages
+      run: |
+        pip install --upgrade pip
+        pip install --upgrade flake8
+
+    - name: Run flake8
+      run: |
+        flake8 easybuild/tools
```
boegel, line 21: Can we also do a full run, but ignore any errors it produces, to assess how bad things are?

```shell
# run flake8 on full easybuild directory, but ignore the exit code until whole codebase is PEP8 compliant
flake8 easybuild || true

File: easybuild/tools/configobj.py [hunk 3]
```diff
@@ -1506,14 +1464,13 @@ def _decode(self, infile, encoding):
             # NOTE: Could raise a ``UnicodeDecodeError``
             return infile.decode(encoding).splitlines(True)
         for i, line in enumerate(infile):
-            if not isinstance(line, unicode):
+            if isinstance(line, str):
```
boegel, line 1467: Hmm, that's not fully equivalent, is it?

Especially not in Python 3, where `str` is also used for Unicode?

easybuild/tools/filetools.py
```diff
@@ -1671,7 +1672,7 @@ def copytree(src, dst, symlinks=False, ignore=None):
     class Error(EnvironmentError):
         pass
     try:
-        WindowsError  # @UndefinedVariable
+        WindowsError = WindowsError # noqa
```
boegel, line 1675: I'd rather just remove this whole body (and clean up the docstring to basically just mention "use copy_dir instead".

It's dead code anyway, since the `log.deprecated` will raise an error now we're in EasyBuild 4.x...

easybuild/tools/job/gc3pie.py
```diff
@@ -69,7 +69,7 @@ class AbortingDependentTaskCollection(AbortOnError, DependentTaskCollection):
         """
         pass
 
-except ImportError as err:
+except ImportError:
```
boegel, line 72: I would restore this, and actually mention the error in the log message:

```python
except ImportError as err:
    _log.debug("Failed to import gc3libs from GC3Pie ('%s'). " % err
               "Silently ignoring, this is a real issue only when GC3Pie is used as backend for --job")

File: easybuild/tools/job/pbs_python.py [hunk 6]
```diff
@@ -54,7 +54,7 @@
     from PBSQuery import PBSQuery
     KNOWN_HOLD_TYPES = [pbs.USER_HOLD, pbs.OTHER_HOLD, pbs.SYSTEM_HOLD]
 
-except ImportError as err:
+except ImportError:
```
boegel, line 57: Same here, please restore and include error message in log message

easybuild/tools/job/pbs_python.py
```diff
@@ -171,7 +171,8 @@ def _get_ppn(self):
 
             # return most frequent
             freq_count, freq_np = max([(j, i) for i, j in res.items()])
-            self.log.debug("Found most frequent np %s (%s times) in interesting nodes %s" % (freq_np, freq_count, interesting_nodes))
+            self.log.debug("Found most frequent np %s (%s times) in interesting nodes %s" %
+                           (freq_np, freq_count, interesting_nodes))
```
boegel, line 175: better:

```python
self.log.debug("Found most frequent np %s (%s times) in interesting nodes %s", 
               freq_np, freq_count, interesting_nodes)

File: easybuild/tools/job/pbs_python.py [hunk 8]
```diff
@@ -453,7 +455,7 @@ def info(self, types=None):
             return None
 
         # convert single type into list
-        if type(types) is str:
+        if isinstance(types, str):
```
boegel, line 458: Better use the `is_string` from `easybuild.tools.py2vs3`?

We should probably deprecate `pbs_python`, Torque is pretty much dead anyway...

easybuild/tools/run.py
```diff
@@ -569,10 +569,10 @@ def parse_log_for_error(txt, regExp=None, stdout=True, msg=None):
     """
     global errors_found_in_log
 
-    if regExp and type(regExp) == bool:
+    if regExp and isinstance(regExp, bool):
         regExp = r"(?<![(,-]|\w)(?:error|segmentation fault|failed)(?![(,-]|\.?\w)"
         _log.debug('Using default regular expression: %s' % regExp)
-    elif type(regExp) == str:
+    elif isinstance(regExp, str):
```
boegel, line 575: please use `is_string` function from `py2vs3` here

easybuild/tools/toolchain/toolchain.py
```diff
@@ -331,8 +331,7 @@ def show_variables(self, offset='', sep='\n', verbose=False):
         if self.vars is None:
             self.generate_vars()
 
-        var_names = self.variables.keys()
-        var_names.sort()
+        var_names = sorted(self.variables.keys())
         res = []
         for v in var_names:
```
boegel, line 336: Do you mind getting rid of this single-letter variable here while you're at it?

Those shouldn't be used outside of list comprehensions...

```python
for var_name in var_names:
```

or

```python
for key in var_names:
```

or even

```python
for key in sorted(self.variables.keys()):

File: easybuild/tools/variables.py [hunk 11]
```diff
@@ -52,7 +52,7 @@ def get_class(name, default_class, map_class=None):
     if name is not None:
         try:
             klass = map_class[name]
-        except:
+        except BaseException:
             for k, v in map_class.items():
```
boegel, line 56: More single-letter variables that would be nice to get rid of...

.github/workflows/linting.yml
Flamefire: I'd say, do this manually locally. This is a CI script so it shouldn't contain pointless stuff. I mean: What would you do with this information? Why do you need it for every CI run? You wouldn't even see it unless you look into the output and it may confuse users because it produces many messages and contributors are meant to look at this output to see what they need to fix to make a PR pass.

easybuild/tools/configobj.py
Flamefire: Not really. But the code works on Python2 only anyway (no str.decode on Python 3) and https://github.com/easybuilders/easybuild-framework/pull/3282/files#diff-70567741de31395daaee88557c7c390dL1521 does it this way. This is the only place that uses `unicode`

As mentioned in Slack this code seems to be unused and has no tests or they would fail for Python3, so fixing those should be done later (or removed)

easybuild/tools/filetools.py
Flamefire: Good point. BTW: When are those deprecated messages meant to be remove? 2 versions after?

Flamefire:
@boegel Added a couple of comments. For the remaining ones: Can we agree to keep this series of PRs focused on PEP8 compliance so flake8 can be enabled? If one is supposed to improve/fix/change every code piece that is neighbouring to code touched by autopep8 we won't finish anytime soon.

So I would really prefer to just fix the PEP8 failures and only change where it is useful for pep8 or an introduced regression.

Examples:
- Ordering of imports: OK to redo as introduced here
- `is_string ` instead of `isinstance(regExp, str):`: actual change in semantic, so not OK as we would have to check if it is really ok to include python 2 strings
- variable renaming: definitely out of scope of this PR

I'm a fan of the boy scout rule but not for a (semi-)automatic style fix which is big enough already.

Final note on e.g. variable renaming: Linters can be used to detect single-letter variables too. This could be done after flake8 passes :)

.github/workflows/linting.yml
boegel: Good point with the latter, so let's leave it as is.

easybuild/tools/configobj.py
boegel: Still, this change annoys me...

Why was it actually done, do you know?
What's wrong with the original `not isinstance(line, unicode)`?

easybuild/tools/filetools.py
boegel: The `log.deprecated` resulting in a warning in EasyBuild 3.x, now it results in an error.
So, it should definitely stay for now (to avoid a nasty crash rather than a clear error), but I guess it could be dropped in EasyBuild 5.0.

boegel:
@Flamefire I understand that you want to avoid that the scope of this PR keeps expanding, so feel free to disagree with some suggested changes (and maybe consider opening a separate PR (or issue) for them).

Agreed on `is_string` vs `isinstance(xxx, str)`, it's probably fine to leave it as is in hindsight.

W.r.t. single-letter variables outside of list comprehensions: OK to consider outside the scope of this PR. Do you know of a linter that has a check for this specifically?

Flamefire:
Ok then I'll define the scope as to not introduce new issues. This should be easily checkable.

There is a flake8 plugin which checks general naming: https://pypi.org/project/pep8-naming/

And more generically (superset of that plugin): pylint. http://pylint.pycqa.org/en/latest/user_guide/options.html See e.g. https://stackoverflow.com/questions/21833872/why-does-pylint-object-to-single-character-variable-names

easybuild/tools/configobj.py
Flamefire: It was done manually because flake8 complained about the unknown name. IMO it is right to change it as it is consistent within the file (Line 1521 uses the same [new] check)

Alternative: Supress with `//noqa`

easybuild/tools/filetools.py
Flamefire: Ok, so 2 versions (or 1 version after "expiry"/version mentioned)

easybuild/tools/job/gc3pie.py
Flamefire: Automatic change w/o behavior change. Can be done later

easybuild/tools/job/pbs_python.py
Flamefire: Automatic change w/o behavior change. Can be done later

easybuild/tools/job/pbs_python.py
Flamefire: Automatic change w/o behavior change. Can be done later

easybuild/tools/run.py
Flamefire: Automatic change w/o behavior change. Can be done later

easybuild/tools/toolchain/toolchain.py
Flamefire: Unrelated

easybuild/tools/variables.py
Flamefire: Unrelated

Flamefire:
Closes all open discussions but 1 either by fixing or mentioning cause for not fixing according to above. Need your call on how to handle https://github.com/easybuilders/easybuild-framework/pull/3282#discussion_r411142227

easybuild/tools/configobj.py
boegel: OK

PR #3286, opened 2020-04-17
cast CPU arch name provided by archspec to a regular string

### LINKED ISSUE #3284: UnicodeDecodeError when adding easyconfig to archive (if archspec is available)
I ran into this crash when installing `MAFFT-7.453-iccifort-2019.5.281-with-extensions.eb` (when using `eb` on top of Python 2):

```
  File "/lib/python2.7/site-packages/easybuild_framework-4.2.0.dev0-py2.7.egg/easybuild/main.py", line 114, in build_and_install_software
  File "/lib/python2.7/site-packages/easybuild_framework-4.2.0.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 3239, in build_and_install_one
    repo.add_easyconfig(spec, app.name, det_full_ec_version(app.cfg), buildstats, currentbuildstats)
  File "/lib/python2.7/site-packages/easybuild_framework-4.2.0.dev0-py2.7.egg/easybuild/tools/repository/filerepo.py", line 120, in add_easyconfig
    txt += statstxt
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 668: ordinal not in range(128)
```

The culprit is that `archspec` returns a `unicode`-type value as CPU arch name, which gets embedded in `statstxt`, making it also a `unicode`-type value (rather than a regular `str`).

When `statstxt` is added to `txt` (which is just a `str` value), it triggers a decoding using the `ascii` of the `txt` value, which fails because the `description` value of `MAFFT-7.453-iccifort-2019.5.281-with-extensions.eb` contains non-ASCII characters (the `~` in "`for alignment of <∼200 sequences`"`).

It's easy to fix this, by making sure we cast the `unicode` value produced by `archspec` to a regular `str` value, but it does raise the question why `archspec` uses `unicode` values at all, there should be no need for that...

boegel:
fixes #3284 (see also https://github.com/archspec/archspec/issues/13)

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -360,7 +360,7 @@ def get_cpu_arch_name():
     if HAVE_ARCHSPEC:
         res = archspec_cpu_host()
         if res:
-            cpu_arch_name = res.name
+            cpu_arch_name = str(res.name)
 
     if cpu_arch_name is None:
         cpu_arch_name = UNKNOWN
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -1,4 +1,5 @@
-# #
+# -*- coding: utf-8 -*-
+##
 # Copyright 2013-2020 Ghent University
 #
 # This file is part of EasyBuild,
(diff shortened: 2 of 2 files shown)


smoors:
Going in, thanks @boegel!

PR #3287, opened 2020-04-17
Introduce contextmanager for disabling templating and reduce resolving errors

Flamefire:
The context manager is a better alternative to the currently widely used pattern of
```
old = value
value = False
# do work
value = old
```

Advantage: Shorter and safer as the reset cannot be forgotten even in case of exceptions

Afterwards some template resolving failures are removed which are observed in https://github.com/easybuilders/easybuild-framework/pull/3285

Diff looks larger than it is as indentation has changed and Github seems to be bad in handling this. All changes are trivial.

### DIFF
--- a/easybuild/framework/extension.py
+++ b/easybuild/framework/extension.py
@@ -127,14 +127,14 @@ class Extension(object):
         # make sure they are merged into self.cfg so they can be queried;
         # unknown easyconfig parameters are ignored since self.options may include keys only there for extensions;
         # this allows to specify custom easyconfig parameters on a per-extension basis
-        for key in self.options:
+        for key, value in self.options.items():
             if key in self.cfg:
-                self.cfg[key] = resolve_template(self.options[key], self.cfg.template_values)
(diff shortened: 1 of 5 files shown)


easybuild/framework/extension.py
```diff
@@ -127,14 +127,14 @@ def __init__(self, mself, ext, extra_params=None):
         # make sure they are merged into self.cfg so they can be queried;
         # unknown easyconfig parameters are ignored since self.options may include keys only there for extensions;
         # this allows to specify custom easyconfig parameters on a per-extension basis
-        for key in self.options:
+        for key, value in self.options.items():
             if key in self.cfg:
-                self.cfg[key] = resolve_template(self.options[key], self.cfg.template_values)
```
boegel, line 132: This was a bit silly indeed, since templates in `self.options` are already resolved above... Nice catch!

akesandgren:
Going in, thanks @Flamefire!

PR #3288, opened 2020-04-17
add support for enhancing existing sanity check

boegel:
Following the issue that was raised by @ocaisa in https://github.com/easybuilders/easybuild-easyconfigs/pull/10418, where defining `sanity_check_commands` in a LAMMPS easyconfig causes other checks that are run by the easyblock to be skipped, I think it's time to add support for specifying that the `sanity_check_commands` and `sanity_check_paths` specified in the easyconfig file should be run *in addition* to what the easyblock already checks, rather than overriding them.

This PR adds support for doing so, via:

```python
enhance_sanity_check = True

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2409,20 +2409,36 @@ class EasyBlock(object):
             SANITY_CHECK_PATHS_DIRS: ("(non-empty) directory", lambda dp: os.path.isdir(dp) and os.listdir(dp)),
         }
 
-        # prepare sanity check paths
-        paths = self.cfg['sanity_check_paths']
-        if not paths:
+        enhance_sanity_check = self.cfg['enhance_sanity_check']
+        ec_commands = self.cfg['sanity_check_commands']
+        ec_paths = self.cfg['sanity_check_paths']
+
+        # if enhance_sanity_check is not enabled, only sanity_check_paths specified in the easyconfig file are used,
+        # the ones provided by the easyblock (via custom_paths) are ignored
+        if ec_paths and not enhance_sanity_check:
+            paths = ec_paths
+            self.log.info("Using (only) sanity check paths specified by easyconfig file: %s", paths)
+        else:
+            # if no sanity_check_paths are specified in easyconfig,
+            # we fall back to the ones provided by the easyblock via custom_paths
             if custom_paths:
                 paths = custom_paths
-                self.log.info("Using customized sanity check paths: %s" % paths)
+                self.log.info("Using customized sanity check paths: %s", paths)
+            # if custom_paths is empty, we fall back to a generic set of paths:
+            # non-empty bin/ + /lib or /lib64 directories
             else:
                 paths = {}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by boegel** [comment]
> I only have one remark, I think that it is counter-intuitive having to forcefully add and empty `dirs` to `sanity_check_files` when we only want to add files to the sanity check with `enhance_sanity_check = True`. Adding an empty `dirs` does not break the sanity checks, it just gets added to the list of paths, but I would make it mandatory only if `enhance_sanity_check = False`.

@lexming That restriction is actually lifted implicitly when `enhanced_sanity_check` is enabled, because either the `sanity_check_paths` from the easyblock or the default one (`bin` + `lib`/`lib64`) will be enhanced, so both `files` and `dirs` will already be there, implying that only specifying `files` or `keys` is fine as long as `enhanced_sanity_check = True` is used...

I've covered that in the test added in 4ee9d25, and also fixed two minor bugs I ran into because of adding that test in 23167d5...

**Comment by lexming** [comment]
@boegel this is not the case on my side. For instance, if I add to LAMMPS-3Mar2020-intel-2019b-Python-3.7.4-kokkos.eb the following `sanity_check_paths`
```
enhance_sanity_check = True
 
sanity_check_paths = {                                                                                                                                                   
    'files': ['log.lammps'],
}
```

I get the following error
>$ ebt LAMMPS-3Mar2020-intel-2019b-Python-3.7.4-kokkos.eb 
== temporary log file in case of crash /local/3036178.master01.hydra.brussel.vsc/eb-RENhlb/easybuild-DDvnmA.log
== resolving dependencies ...
== processing EasyBuild easyconfig /theia/home/brussel/101/vsc10122/src/easybuild-easyconfigs/easybuild/easyconfigs/LAMMPS-3Mar2020-intel-2019b-Python-3.7.4-kokkos.eb
== FAILED: Installation ended unsuccessfully (build directory: /user/brussel/101/vsc10122/.local/easybuild-ivybridge/build/LAMMPS/3Mar2020/intel-2019b-Python-3.7.4-kokkos): build failed (first 300 chars): **Missing mandatory key 'dirs' in sanity_check_paths.** (took 0 sec)
== Results of the build can be found in the log file(s) /local/3036178.master01.hydra.brussel.vsc/eb-RENhlb/easybuild-LAMMPS-3Mar2020-20200420.194814.YSvyG.log
ERROR: Build of /theia/home/brussel/101/vsc10122/src/easybuild-easyconfigs/easybuild/easyconfigs/LAMMPS-3Mar2020-intel-2019b-Python-3.7.4-kokkos.eb failed (err: "build failed (first 300 chars): **Missing mandatory key 'dirs' in sanity_check_paths**.")

boegel:
@lexming Ah, that's failing in an entirely different part of the code. Apparently we have that same restriction in two places...

I like to keep it there though, for the case where `enhance_sanity_check` is not enabled.

I've reproduced your issue in a test and fixed the problem in 96ec4e9 .

lexming:
Going in, thanks @boegel !

Issue #3280, opened 2020-04-15
Lock folder not cleaned up after (Slurm) job is killed


Flamefire:
The following happened:
- Create interactive job with SLURM 
- Change to node
- Start build with EB
- Get killed (Connection closed in this case, but can be due to timeout [slurm time exceeded])

Afterwards the lock was still there so the next build failed too and the lock had to be removed manually

@boegel @mboisson

mboisson:
Copy from a discussion with Todd Gamblin on Slack : 

```
we use fcntl locks
they are released when the process dies
and we use one lockfile for everything — no mess.
the locks are “bytes” in the file
but the file’s zero-length
it’s one empty lockfile, we take a 63-bit prefix of the SHA-1 of the DAG hash and lock that byte for each build
the advantage is that a) fcntl locks are supported for on most filesystems (NFS3 with lock server, NFS4, GPFS, Lustre) and b) they’re released when the process dies
which prevents the annoying SVN locked-forever-because-nfs-is-busy issue

if you want to steal ours, here you go: https://github.com/spack/spack/blob/develop/lib/spack/llnl/util/lock.py (edited) 
tests are here: https://github.com/spack/spack/blob/develop/lib/spack/spack/test/llnl/util/lock.py 
I think you will need our little barrier implementation if you want to test on python 2.6: https://github.com/spack/spack/blob/develop/lib/spack/llnl/util/multiproc.py — or you can just run MPI tests. (edited) 
the only thing beyond that lock you need is the little trick we use to lock prefixes with one lock file: https://github.com/spack/spack/blob/develop/lib/spack/spack/database.py#L489 (edited) 
spack lets you install “anything” so every installation has a hash.  so we map a hash to a byte in a lockfile and lock that byte (fcntl and that lock class I linked support byte-range locking)

boegel:
The problem with `fcntl` locks is that not all filesystems support them though, so it's not a perfect solution either. If we can figure out a way to detect whether `fcntl` locks can be used, then it seems that's a better solution, but I'm not sure that can be done reliably...

I've also been bitten by the problem that @Flamefire reported, it's certainly annoying.

How about installing a signal handler that cleans up the locks that were created in that EasyBuild session in case a `SIGTERM` signal (and possible other signals) is received?

mboisson:
Does a crash necessarily get a signal ?
  

  
Not all signals can be catches either.
  

  
What about implementing both, and letting the configuration decide which implementation is used ?
  

[...]

PR #3291, opened 2020-04-19
clean up locks when EasyBuild session is cancelled with signal like SIGTERM

boegel:
cfr. #3280

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -60,7 +60,7 @@ from easybuild.base import fancylogger
 from easybuild.tools import run
 # import build_log must stay, to use of EasyBuildLog
 from easybuild.tools.build_log import EasyBuildError, dry_run_msg, print_msg, print_warning
-from easybuild.tools.config import GENERIC_EASYBLOCK_PKG, build_option, install_path
+from easybuild.tools.config import DEFAULT_WAIT_ON_LOCK_INTERVAL, GENERIC_EASYBLOCK_PKG, build_option, install_path
 from easybuild.tools.py2vs3 import std_urllib, string_type
 from easybuild.tools.utilities import nub, remove_unwanted_chars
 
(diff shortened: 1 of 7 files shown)


easybuild/tools/filetools.py
```diff
@@ -1476,6 +1480,103 @@ def mkdir(path, parents=False, set_gid=None, sticky=None):
...
+    """
+    lock_path = det_lock_path(lock_name)
+    if os.path.exists(lock_path):
+        _log.info("Lock %s exists!", lock_path)
+        wait_on_lock = build_option('wait_on_lock')
+        if wait_on_lock:
+            while os.path.exists(lock_path):
+                print_msg("lock %s exists, waiting %d seconds..." % (lock_path, wait_on_lock),
+                          silent=build_option('silent'))
+                time.sleep(wait_on_lock)
```
akesandgren, line 1522: I would expect wait_on_lock to be the max time to wait, not how long between checks. That's what you would normally do in this situation, i.e. specify a max time to wait.

easybuild/tools/filetools.py
```diff
@@ -1476,6 +1480,103 @@ def mkdir(path, parents=False, set_gid=None, sticky=None):
...
+    """
+    Clean up all still existing locks that were created in this session.
+    """
+    for lock_name in list(global_lock_names):
+        remove_lock(lock_name)
+
+
+def clean_up_locks_signal_handler(signum, frame):
+    """
+    Signal handler, cleans up locks & exists with received signal number.
```
akesandgren, line 1551: exits not exists

easybuild/tools/filetools.py
boegel: Yeah, we need a way to limit the time you're waiting for a lock, and the semantics of `--wait-on-lock` is a bit weird right now.

So, I changed things in d879cdadc:

* `--wait-on-lock` still works as before (`0` (the default) implies no waiting (raise an error if lock is already there), a value implies enabling waiting with interval of specified number of seconds), but is deprecated;
* `--wait-on-lock-limit` can be used to i) enable waiting for lock to be released, ii) specify a maximum waiting time (if set to `0` is implies no waiting (raise error if lock is there), if set to `-1` it implies waiting indefinitely);
* `--wait-on-lock-interval` can be used to specify the interval for re-checking whether the lock has been released (default 60 sec.)

easybuild/tools/filetools.py
```diff
@@ -1493,6 +1497,131 @@ def mkdir(path, parents=False, set_gid=None, sticky=None):
...
+        # wait limit could be zero (no waiting), -1 (no waiting limit) or non-zero value (waiting limit in seconds)
+        if wait_limit != 0:
+            wait_time = 0
+            while os.path.exists(lock_path) and (wait_limit == -1 or wait_time < wait_limit):
+                print_msg("lock %s exists, waiting %d seconds..." % (lock_path, wait_interval),
+                          silent=build_option('silent'))
+                time.sleep(wait_interval)
+                wait_time += wait_interval
+
+            if wait_limit != -1 and wait_time >= wait_limit:
```
akesandgren, line 1563: This must check if the lock still exists before throwing an error.
In case the lock is removed during the last sleep..

akesandgren:
Going in, thanks @boegel!

PR #3292, opened 2020-04-19
add 'change_into_dir' named argument to 'extract_file' + print deprecation warning if it's not specified

boegel:
cfr. 2858

If this gets merged, we'll need to make all calls to `extract_file` in easyblocks also specify `change_into_dir`...

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -372,7 +372,7 @@ def change_dir(path):
     return cwd
 
 
-def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced=False):
+def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced=False, change_into_dir=None):
     """
     Extract file at given path to specified directory
     :param fn: path to file to extract
@@ -381,8 +381,16 @@ def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced
     :param extra_options: extra options to pass to extract command
     :param overwrite: overwrite existing unpacked file
     :param forced: force extraction in (extended) dry run mode
+    :param change_into_dir: change into resulting directory;
+                          None (current default) implies True, but this is deprecated,
+                          this named argument should be set to False or True explicitely
+                          (in a future major release, default will be changed to False)
     :return: path to directory (in case of success)
     """
+    if change_into_dir is None:
+        _log.deprecated("extract_file function was called without specifying value for change_into_dir", '5.0')
+        change_into_dir = True
+
     if not os.path.isfile(fn) and not build_option('extended_dry_run'):
         raise EasyBuildError("Can't extract file %s: no such file", fn)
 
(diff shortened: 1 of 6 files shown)


easybuild/tools/filetools.py
```diff
@@ -408,7 +416,18 @@ def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced
...
 
-    return find_base_dir()
+    # note: find_base_dir also changes into the base dir!
+    base_dir = find_base_dir()
+
+    # if changing into obtained directory is not desired,
+    # change back to where we came from (unless that was a non-existing directory)
+    if not change_into_dir:
+        if cwd is None:
+            _log.warning("Can't change back to non-existing directory after extracting %s in %s", fn, dest)
```
akesandgren, line 426: Are you sure this shouldn't be an fatal error?

boegel: Yeah, it doesn't make much sense if this isn't fatal... fixed in 4bb149d31

easybuild/tools/filetools.py
```diff
@@ -392,8 +400,8 @@ def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced
     abs_dest = os.path.abspath(dest)
 
     # change working directory
-    _log.debug("Unpacking %s in directory %s.", fn, abs_dest)
-    change_dir(abs_dest)
+    _log.debug("Unpacking %s in directory %s", fn, abs_dest)
+    cwd = change_dir(abs_dest)
```
akesandgren, line 404: There is a problem here from test_guess_start_dir which does:
```
        # default (no start_dir specified): use unpacked dir as start dir
        self.assertEqual(ec['ec']['start_dir'], None)
        check_start_dir('toy-0.0')

        # using start_dir equal to the one we're in is OK
        ec['ec']['start_dir'] = '%(name)s-%(version)s'
        self.assertEqual(ec['ec']['start_dir'], 'toy-0.0')
        check_start_dir('toy-0.0')
```
That second check_start_dir('toy-0.0') has cwd %(builddir)s/toy-0.0 when it starts, the run_all_steps kicks off
ready_steps which is make_builddir which removes the old builddir thus causing the change_dir(abs_dest) to return None since builddir no longer exists.

Solution would perhaps be to compare abs_dest and cwd before that change_dir to see if they are the same, or if cwd is below abs_dest, and then set cwd to abs_dest

boegel: This is just an issue in the test, easy fix is to change back to where the test started from on each call to `check_start_dir`, done in 3a4bc9860

easybuild/tools/filetools.py
```diff
@@ -408,7 +416,18 @@ def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced
...
 
-    return find_base_dir()
+    # note: find_base_dir also changes into the base dir!
+    base_dir = find_base_dir()
+
+    # if changing into obtained directory is not desired,
+    # change back to where we came from (unless that was a non-existing directory)
+    if not change_into_dir:
+        if cwd is None:
+            raise EasyBuildError("Can't change back to non-existing directory after extracting %s in %s", fn, dest)
```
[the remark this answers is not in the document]
boegel: We don't have to the path to the non-existing directory, since you can't determine it...

```
$ mkdir /tmp/thiswontbeherelong
$ cd /tmp/thiswontbeherelong
$ python -c 'import os; print(os.getcwd())'
/private/tmp/thiswontbeherelong

$ rm -rf $PWD
$ python -c 'import os; print(os.getcwd())'
OSError: [Errno 2] No such file or directory
```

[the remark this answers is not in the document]
akesandgren: yeah hadn't figured out exactly what was going on there yet...

boegel:
@akesandgren easyblocks fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2051

akesandgren:
Going in, thanks @boegel!

PR #3294, opened 2020-04-22
Add 'git_config' support within 'exts_list'

### LINKED ISSUE #3251: Installing extensions from git repository
Usually, it is possible to install from a git repository directly, and without looking at the mechanisms in too much detail, I believe that this works by providing `git_config` at

https://github.com/easybuilders/easybuild-framework/blob/210743d0e3618a8ac0a56eb9c0f4fa4fd8ae53b9/easybuild/framework/easyblock.py#L379

However, I was not able to figure out how to do the same for an extension (defined within the `exts_list` easyconfig parameter), and I suspect that might be due to the fact that no `git_config` from an extension's options is passed at 

https://github.com/easybuilders/easybuild-framework/blob/0b7a627f28131633870148d643c0286bc348206c/easybuild/framework/easyblock.py#L520

Why not? Wouldn't that be useful, i.e. for installing a custom python repository via git together with its dependencies (say, standard pipy packages) in a `PythonBundle`?

kelseymh:
Modify `fetch_extension_sources()` so that when a source dictionary is included in an `exts_list` entry, it looks for `git_config`, and passes the latter through to `obtain_file()`.  This allows us to pull extensions from Git the same way we pull regular sources, including creating a local tar-ball on the fly.

~I plan to make this a "draft" (WIP) PR initially, because I haven't been able to explicitly test my code changes.~

edit (@boegel): fixes #3251


kelseymh:
I've successfully tested this feature with one of our internal Git packages.  In my EasyConfig file, I have an extensions block:

    exts_defaultclass = 'PythonPackage'
    exts_list = [
        ('pyCAP', 'master', {
            'filename': 'pyCAP-master.tar.gz',
            'git_config': {
                'url': local_git + '/Analysis',
                'repo_name': 'pyCAP',
                'tag': 'master',
            }
        }),
    ]

When I run with the production `eb` (EasyBuild/2020a-sprint), I get the expected failure:

    == fetching files...
    == FAILED: Installation ended unsuccessfully (build directory: /scratch/group/mi
    tchcomp/eb/tmp/build/cdmsPyGlobal/0.5/system-system-Python-3.6.6test):
    build failed (first 300 chars): Couldn't find file pyCAP-master.tar.gz anywhere,
    and downloading it didn't work either... Paths attempted (in order): /scratch/gr
    oup/mitchcomp/eb/eb-files/c/cdmsPyGlobal/extensions/pyCAP-master.tar.gz, /scratc
    h/group/mitchcomp/eb/eb-files/c/cdmsPyGlobal/packages/pyCAP-master.tar.gz, /scra
    tch/group/m (took 0 sec)

but when I run with my personal checkout, it succeeds:

    == fetching files...
    [...]
    == installing extension pyCAP master (1/1)...
    [...]

Is there documentation for how to create a new unit test?  I'd love to exercise this in some way as part of the pull request.

kelseymh:
@boegel Suggested that a unit test could be added to either the [easyblock.py](https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/easyblock.py) or [toy_build.py]( https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/toy_build.py) tests.  

I think I can do this, but neither one appears to actually perform a download of sources.  What I would want to test is a dictionary entry like

    exts_list = [('toy', '1.0', {
        'filename':'toy-build.tar.gz', 
        'git_config': {
            'url':'https://github.com/something-something',
            'repo_name':'toy', 'tag':'1.0',
        }
    })]

and see that either the download actually occurs (ideal) or at least that the contents of 'git_config' is seen and is validated.

kelseymh:
I don't think I need to actually " see that ... the download actually occurs (ideal)."  The [filetools.py](https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/filetools.py) tests already include `test_get_source_tarball_from_git()`.  That exercises both the different components of `git_config` as well as the downloading procedure.

For this situation, I just need to test whether having a 'git_config' dictionary in `exts_list` does not generate an error.  This should be easier for me to figure out how to implement.

kelseymh:
I just noticed that [`test_obtain_file()`](https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/easyblock.py#L1247) in the `easyblock.py` test does not exercise the `git_config` option.  Should I try to add that as a unit test as well?

akesandgren:
@boegel @kelseymh Shouldn't the format for this be
```
exts_list = [
  ('pyCAP', 'master', {
    'sources': [{
      'filename':...,
      'git_config'
    }],
  }),
]
```
To make it clear that it's the source specification, and make it behave like non-ext-list sources, allowing multiple sources files...

kelseymh:
@akesandgren That's a good question!  The allowed content of `exts_list` is [undocumented](https://github.com/easybuilders/easybuild/pull/617), so I constructed that example block by trial and error.

In the code of this PR, I didn't want to try to change how exts_list parsed the `extra_options` dictionary, just added extraction of a `git_config` element.  

The EasyConfigs which make use of `'sources'` in a dictionary all seem to have it either at the top level or in `default_component_specs` (see, e.g., [SpliceMap](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/s/SpliceMap/SpliceMap-3.3.5.2-GCC-7.3.0-2.30.eb)).

My write up of `exts_list` behaviour (see [Framework PR #617](https://github.com/easybuilders/easybuild/pull/617) follows this searc of existing examples.  If putting a whole `'sources':` dictionary into `extra_options` does work, then I should update that attempt at documentation accordingly.

test/framework/toy_build.py
```diff
@@ -1194,6 +1194,40 @@ def test_toy_extension_patches(self):
...
+        ext_src = 'int main() { return 0; }'
+        ext_cfile = 'exts-git.c'
+
+        test_ec = os.path.join(self.test_prefix, 'test.eb')
+        test_ec_txt = '\n'.join([
+            toy_ec_txt,
+            'prebuildopts = "echo \\\"%s\\\" > %s && ",' % (ext_src, ext_cfile),
+            'exts_list = [',
+            '   ("exts-git", "0.0", {',
+            '       "buildopts": "&& ls -l %s",' % ext_tarball,
```
boegel, line 1217: @kelseymh Maybe (also) use `ls -l a_directory/a_file.txt` here (which should be there, since that's part of `testrepository`)? That can't just come out of thin air, it must have resulted from downloading the git repo and unpacking the resulting tarball?

test/framework/toy_build.py
```diff
@@ -1194,6 +1194,40 @@ def test_toy_extension_patches(self):
...
+            '       "source_tmpl": "%s",' % ext_tgz,
+            '       "git_config": {',
```
boegel, line 1219: For non-extensions, `git_config` is a way of specifying an entry in `sources`.
Should we do that here too?

I think long-term we want to move away from the `source_tmpl` (which is custom to extensions), and move to using `sources` for extensions as well. For now, we can restrict it to a single-element list for an extension, we've never run into a case where an extension requires multiple sources afaik.

So, we should:

* introduce support for `sources` in an extension (which should have preference over `source_tmpl` in case both are specified);
* support `git_config` as a part of `sources`, like:
  ```
  exts_list = [
    ('exts-git', '0.0', {
        'sources': [{
            'filename': 'exts-git-0.0.tar.gz',
            'git_config': {
                ...
            }],
    }),
  ]
  ```

Let me know if you're up for this...

test/framework/toy_build.py
kelseymh: Sure, I can do that!  I figured that since the tarball name was "unexpected", it couldn't have come into existence automatically either.

(I'll "resolve" this once I updated the code.)

test/framework/toy_build.py
kelseymh: This needs some thinking.  I imagine it would require continuing to support the existing interpretation of `extra_options`, and having an alternative if-block where `extra_options` contains only `sources`.

Perhaps the expansion of `sources` could be encapsulated and just called from here?

kelseymh: Okay, breakfast and lunch gave me time to digest this more.  I'm thinking that since there's at least one other user out there (#3251) who would like this functionality, it might serve the community better to roll out the functionality here, and open a separate Issue or PR to support the nested `extra_options = { 'sources': [{...}] }` functionality.  

I realize that means `git_config` is apparently being handled in two different ways, but that's not true behind the scenes.  In both cases, the `git_config` dictionary value is captured and passed through to `obtain_file`, without being separately parsed, so the behaviour would be consistent.

kelseymh:
@boegel wrote:
> An enhanced test for obtain_file to cover the use of git_config makes sense too, but that can also be tackled in a separate PR I think, up to you...

No need.  I got most of the way through writing it, and discovered that `test/framework/filetools.py` already contains `test_test_get_source_tarball_from_git()`.  I've closed the issue (#3322) I created for this.

kelseymh:
@boegel Would you prefer to hold this off until `'sources'` is supported in `exts_list`, or do it in two stages?

kelseymh:
Okay, this will need some rewriting.  It looks as though `fetch_extension_sources()` is written to be analogous to `fetch_sources()`, which takes top-level 'sources' and 'checksums' as arguments.

kelseymh:
I've implemented the use of a single-dictionary `'sources'` key in `ext_options`, along with a unit test for it in `toy_build.py`.  It does not work, and I'd appreciate some assistance from the experts to understand why.  Here's an outline of what I did and why it doesn't work correctly.

* I factored most of the action in `fetch_sources()` out to a single-item `fetch_source()` function.  This function returns a "src" dictionary with keys `[ 'name', 'path', 'cmd', 'checksum', 'finalpath' ]`.
* In `fetch_sources()`, those returned dictionaries are collected into `self.src` as before.
* I added the ability to pass a simple dictionary to `fetch_sources()`, but I think that was unnecessary.
* In `fetch_extension_sources()`, I check for a `sources` key in `ext_options`, and if found, it gets passed to `fetch_source()` along with the checksum, if there was one.
* I notice that in `fetch_extension_sources()`, in the block where `obtain_file()` is called directly, the resulting file is put into `ext_src` with the key `'src'`, rather than the key `'path'`.

This is the problem, I think. In places where the extensions are actually processed, I guess that they look for a key `'src'` in the output from `fetch_extension_sources()`.  But by moving to my new `fetch_source()` function, what comes back is a key `'path'`.  I tried doing a global replace, but that just seemed to cause more problems.

In `easyblock.py`, I've added "`# TODO MHK`" comments in the places where I'm not sure how to proceed.  I could use some insight from the experts on how extensions actually get built, to understand where my changes are having an effect elsewhere.

kelseymh:
After discussion with @boegel , I think we can use both 'src' and 'path' as keys in the `ext_src` dictionary.  `fetch_source()` will return the "standard" dictionary expected by `fetch_sources()`.  In `fetch_extension_sources()` we copy the 'path' value into a new 'src' key.  The former should just be silently ignored during the extensions step.  If it's not, then in the same copy action we can delete the 'path' key from `ext_src`.

kelseymh:
This isn't working.  If I use a `sources` dictionary within `exts_list`, and do what I described above, putting the name of the fetched source file into an `ext_sources` entry with a `src` key, then the extension-building step fails:

The command that tries to get executed is (this is from running test.framework.toy_build):

    gcc exts-git.tar.gz.c -o exts-git.tar.gz

where it's plugged in the name of the tarball itself as the thing to be compiled!

kelseymh:
Solved!  Instead of the whole data-structure output from the new `fetch_source()`, ext_source really only wants the downloaded filename put into the `src` key in ext_src.  With this change the `toy_build_extension_sources` test runs successfully.

However, `toy_build_download_sources` is failing (for me, when I specify just the "sources" tests) because it doesn't find EASYBUILD_BUILDPATH in the environment variables.

test/framework/toy_build.py
kelseymh: Following up, since it's showing that there are still "unresolved changes" from the review, but I can't find them.

I have successfully (I think!) implemented putting a full 'sources' block inside an 'exts_list' entry.  I did not remove the bare 'git_config' entry in 'exts_list', because it already worked, and because (I think) each 'exts_list' is supposed to represent a single source, not a complex multisource compilation.

This does mean, as I noted above, that this gives two syntactic ways of achieving the same end result.  If you would like the bare 'git_config' support removed, please let me know.

easybuild/framework/easyblock.py
```diff
@@ -337,6 +341,59 @@ def get_checksum_for(self, checksums, filename=None, index=None):
...
+        Will be tested for existence or can be located
+
+        :param source: source to be found (single dictionary in 'sources' list, or filename)
+        :param checksum: checksum corresponding to source
+        :param extension: flag if being called from fetch_extension_sources()
+        """
+        filename, download_filename, extract_cmd, source_urls, git_config = None, None, None, None, None
+
+        if source is None:
+            raise EasyBuildError("Fetch_source called without 'source' argument")
```
boegel, line 356: `s/Fetch_source/fetch_source/g` (getting the function name correct is superior to start a log message with a capital letter imho)

easybuild/framework/easyblock.py
```diff
@@ -350,46 +407,22 @@ def fetch_sources(self, sources=None, checksums=None):
...
-                    'name': filename,
-                    'path': path,
-                    'cmd': extract_cmd,
-                    'checksum': self.get_checksum_for(checksums, filename=filename, index=index),
-                    # always set a finalpath
-                    'finalpath': self.builddir,
-                })
+        # Loop over the list of sources; list of checksums must match >= in size
+        for index, source in enumerate(sources):
+            if source:
```
boegel, line 418: Why the `if source:`? How can this ever be equivalent to `False`?

easybuild/framework/easyblock.py
```diff
@@ -350,46 +407,22 @@ def fetch_sources(self, sources=None, checksums=None):
...
+                got_src = self.fetch_source(source, self.get_checksum_for(checksums=checksums, index=index))
+                if got_src:
```
boegel, line 420: `got_src` is a bit confusing to me, maybe `src_spec` or `src_details` (or just `src` if the `if source:` above is not required)?

easybuild/framework/easyblock.py
```diff
@@ -514,10 +547,19 @@ def fetch_extension_sources(self, skip_checksums=False):
...
+                        sources = ext_options['sources']
+                        src = self.fetch_source(sources, checksums, extension=True)
+                        # Copy 'path' entry to 'src' for use with extensions
+                        ext_src.update({'src': src['path']})
+                        exts_sources.append(ext_src)
                     else:
                         source_urls = ext_options.get('source_urls', [])
                         force_download = build_option('force_download') in [FORCE_DOWNLOAD_ALL, FORCE_DOWNLOAD_SOURCES]
-                        src_fn = self.obtain_file(fn, extension=True, urls=source_urls, force_download=force_download)
+                        git_config = ext_options.get('git_config', None)
```
boegel, line 559: This doesn't make much sense to me, to support `git_config` outside of `sources`.

In the long run, we want to deprecate the use of `source_tmpl`, and replace it with just using `sources`, which is more consistent with the top-level `sources` easyconfig parameter. There is no `git_config` easyconfig parameter either, it only exists in the context of `sources`.

easybuild/framework/easyblock.py
```diff
@@ -337,6 +341,59 @@ def get_checksum_for(self, checksums, filename=None, index=None):
         else:
             raise EasyBuildError("Invalid type for checksums (%s), should be list, tuple or None.", type(checksums))
 
+    def fetch_source(self, source=None, checksum=None, extension=False):
```
boegel, line 344: This should just be ` def fetch_source(self, source, ...):`, it doesn't make sense to call `fetch_source` without passing a value for `source`

easybuild/framework/easyblock.py
kelseymh: Yes, this makes sense.  The idea that each `exts_list` entry is like a miniature easyconfig, immediately implies that it shouldn't have a top-level `git_config`.  I plan to add this explanation to the documentation (docs PR #[617](https://github.com/easybuilders/easybuild/pull/617)).

easybuild/framework/easyblock.py
kelseymh: This is a sanity check for a user who puts a dead entry into their `sources` list.  I want to keep the error check here with a clear message, as well as down in `fetch_source()` itself.  The latter is there because it can be called directly, as we do with extensions, so I want an explicit test if someone passes None.

I did rearrange the code slightly so it has the error condition at the top:
```
for index, source in enumerate(sources):
        if source is None:
            raise EasyBuildError("Empty source in sources list at index %d", index)

        got_src = self.fetch_source(source, self.get_checksum_for(checksums=  [...]
```
I tend to prefer this layout when the failure condition is simple, and the success is a big block of code.  Unless you like it for coding style, the "else:" is unnecessary since the error acts like a return/exit.

easybuild/framework/easyblock.py
```diff
@@ -514,9 +547,19 @@ def fetch_extension_sources(self, skip_checksums=False):
 
                     if ext_options.get('nosource', None):
                         exts_sources.append(ext_src)
+                    elif ext_options.get('sources', None):
+                        sources = ext_options['sources']
+                        if isinstance(sources, (list, tuple)):
+                            raise EasyBuildError("'sources' entry to 'exts_list' must be a single dictionary.")
```
boegel, line 553: I would allow single-element lists, to minimize the gap with the `sources` easyconfig parameter, see https://github.com/kelseymh/easybuild-framework/pull/2

boegel:
Close & re-open to try and get Travis to wake up... You're running on fumes here Travis, seriously...

PR #3297, opened 2020-04-22
added function find_glob_pattern to filetools.py

mboisson:
This function is used in the bender and the ROOT easyblocks. I also have a need for it in the qscintilla EasyBlock. I generalized it slightly to allow to not raise an exception if no match is found (but instead return None).


kelseymh:
I'm probably being too simple-minded here.  I think the "`self`" argument to find_glob_pattern needs to be dropped (since filetools.py isn't a class), and the "`self.dry_run`" condition should be removed, or modified to be a passed-in argument from `root.py` and `blender.py`.

Micket:
sorry about that, my github was out-of-date, i totally missed the conversation

Micket:
So, dry-run check at all seems a little bit strange to me; it's mostly used for functions that actually modify files and directories. This is just read only. I suppose it's just to avoid exceptions being thrown when expected directories aren't created due to a dry run?

mboisson:
The dryrun check is inherited from the code that was already in the two EasyBlocks. I did not want to remove it and risk breaking something.

Micket:
Yeah I'm just not sure if the easyblock `self.dry_run` is really equivalent to if `build_option('extended_dry_run')`. Perhaps a input parameter is called for in that case.
As I don't really see the need for it in either case, I'd like input from a second maintainer.

mboisson:
Maybe @kelseymh can comment ?

kelseymh:
@mboisson and @Micket The two are exactly the same:
https://github.com/easybuilders/easybuild-framework/blob/7c1c2885168a16ad6407c9e09cdd949ae4406250/easybuild/framework/easyblock.py#L235

PR #3304, opened 2020-04-27
fix broken test for --include-easyblocks-from-pr

boegel:
Test for `--include-easyblocks-from-pr` which was relying on easyconfigs PR https://github.com/easybuilders/easybuild-easyconfigs/pull/9979 to `2020a` branch got broken when cleanup PR https://github.com/easybuilders/easybuild-easyconfigs/pull/10488 got merged

Easy to fix by using https://github.com/easybuilders/easybuild-easyconfigs/pull/10487 instead...


lexming:
Going in, thanks @boegel !

PR #3305, opened 2020-04-28
added code to support multi_deps for extensions only

mboisson:
This can save valuable time when only the extensions require multi_deps. For example, this could be used to install Matlab, or TensorRT with python bindings, without re-installing those for each version of python. 

Ideally, in this case, I would want to skip extracting the source, and keep the version that was built in the previous iteration, but I could not figure out the right combination of options for that, so I kept extracting the source here. 

I tested this with the TensorRT recipe, which installs python wheels that are part of the extracted binary. 

I would like @bartoldeman and @boegel's comments on this PR.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2832,7 +2832,7 @@ class EasyBlock(object):
             raise StopException(step)
 
     @staticmethod
-    def get_steps(run_test_cases=True, iteration_count=1):
+    def get_steps(run_test_cases=True, iteration_count=1, multi_deps_extensions_only=False):
         """Return a list of all steps to be performed."""
 
         def get_step(tag, descr, substeps, skippable, initial=True):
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/default.py
```diff
@@ -152,6 +152,7 @@
     'hiddendependencies': [[], "List of dependencies available as hidden modules", DEPENDENCIES],
     'multi_deps': [{}, "Dict of lists of dependency versions over which to iterate", DEPENDENCIES],
     'multi_deps_load_default': [True, "Load module for first version listed in multi_deps by default", DEPENDENCIES],
+    'multi_deps_extensions_only': [False, "Use multiple dependencies for extensions only", DEPENDENCIES],
```
boegel, line 155: I would make this `"Use dependencies specified via multi_deps only to install extensions"`

easybuild/framework/easyblock.py
```diff
@@ -2898,17 +2898,25 @@ def install_step_spec(initial):
...
-            configure_step_spec,
-            build_step_spec,
-            test_step_spec,
-            install_step_spec(False),
-            extensions_step_spec,
-        ] * (iteration_count - 1)
+        if multi_deps_extensions_only:
+            steps_part2 = [
+                ready_step_spec(False),
+                source_step_spec(False),
```
boegel, line 2904: @mboisson What problem did you encounter when kicking out `ready_step_spec` and `source_step_spec`?

mboisson: Well, for example, in the case of TensorRT, the extensions are actually bundled with the extracted binaries. So they need to be there. One of the steps removes them after the installation, so they need to be re-extracted. 

The ready_step is the one that handles having the different dependencies reloaded, so it's needed.

easybuild/framework/easyconfig/default.py
mboisson: That was too long for a single line. I went with a compromise.

easybuild/framework/easyblock.py
mboisson: I managed to get rid of the `source_step_spec` by modifying `ready_step_spec` conditionnaly to avoir recreating the build directory if `multi_deps_extensions_only` is used.

akesandgren:
Looks sane but I'll defer to @boegel

mboisson:
Will this be included in 4.2.1 @boegel ?

boegel:
@mboisson Tests are missing, so this is not ready to go in yet imho...

I haven't had a chance to closely follow up on this, but i don't think this should block the 4.2.1 release (which we hope to push out any day now).

I'm moving this to the milestone for the release after 4.2.1...

mboisson:
@boegel, what kind of test would you like to see ? I'm not sure how to test this.

boegel:
@mboisson To get this covered in the tests, see how we test the use of `multi_deps`?

Issue #3231, opened 2020-03-06
Errors and warnings in unit tests.


shrinny:
Following the installation instructions [here](https://easybuild.readthedocs.io/en/latest/Installation.html#installing-easybuild), running the unit tests after installation results in a lot of errors and warnings. Not sure if these are expected or not.

Link to full [output](https://gist.github.com/shrinny/76e87569ace811710e1bc33378648aff)

The majority of the warnings are `WARNING: 'eb' not found in $PATH, failed to determine installation prefix!`, which is weird because I _do_ have eb on my `$PATH`. It seems like the tests change `PATH` and `PYTHONPATH` environment variables. Moreover, some of the test failures also seem to be related to the changed environment variables as running some of the failing tests in a python terminal (without the testing framework) seems to work.

On the other hand, most of the errors seem to be complaining about some toolchains not being available e.g
`'Toolchain GCC not found, available toolchains: GCCcore,system'`. Not sure if I was meant to install other toolchains first before running the unit tests? Doesn't seem to be mentioned in the documentation in any case.

___
### System Information
<details><summary></summary>

```console
user@system:~$ uname -a
Linux berni 5.3.0-24-generic #26-Ubuntu SMP Thu Nov 14 01:33:18 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
user@system:~$ python -V
Python 3.7.5
user@system:~$ type module
module is a function
module ()
{
    eval $($LMOD_CMD bash "$@") && eval $(${LMOD_SETTARG_CMD:-:} -s sh)
}
user@system:~$ type -f module
-bash: type: module: not found
user@system:~$ module --version

Modules based on Lua: Version 8.3.3  2020-03-02 15:56 -06:00
    by Robert McLay mclay@tacc.utexas.edu

user@system:~$ module av EasyBuild
-------------------------- /opt/modules/all --------------------------   
EasyBuild/4.1.1 (L)

  Where:
   L:  Module is loaded

Use "module spider" to find all possible modules and extensions. 
Use "module keyword key1 key2 ..." to search for all possible modules 
matching any of the "keys"

user@system:~$ which -a eb
/opt/software/EasyBuild/4.1.1/bin/eb
user@system:~$ eb --version
This is EasyBuild 4.1.1 (framework: 4.1.1, easyblocks: 4.1.1) on host system.
user@system:~$
```
</details>

boegel:
[...]

PR #3309, opened 2020-04-29
Constants for osdependecies.

terjekv:
- The osdependencies relating to ibverbs and openssl are repeated
    ad nauseam in the code. Occasionally with minor mistakes added
    in.
  - ibverbs / infiniband and openssl are the most frequent
    package lists used, but their use is inconsistent as
    packages change name over time.
  - When packages change names, old easyconfigs will stop working
    unless the dependency statement is updated.
  - Moving these lists into constants allows for the lists to be
    updated in one place and at one time.

### DIFF
--- a/easybuild/framework/easyconfig/constants.py
+++ b/easybuild/framework/easyconfig/constants.py
@@ -51,4 +51,7 @@ EASYCONFIG_CONSTANTS = {
     'OS_VERSION': (get_os_version(), "System version"),
     'SYS_PYTHON_VERSION': (platform.python_version(), "System Python version (platform.python_version())"),
     'SYSTEM': ({'name': 'system', 'version': 'system'}, "System toolchain"),
+
+    'OSPKGS_IBVERBS': (('libibverbs-dev', 'libibverbs-devel', 'rdma-core-devel'), "OS packages providing ibverbs support"),
+    'OSPKGS_OPENSSL': (('openssl-devel', 'libssl-dev', 'libopenssl-devel'), "OS packages providing openSSL support"),
 }
--- a/easybuild/framework/easyconfig/constants.py
+++ b/easybuild/framework/easyconfig/constants.py
@@ -53,7 +53,7 @@ EASYCONFIG_CONSTANTS = {
     'SYSTEM': ({'name': 'system', 'version': 'system'}, "System toolchain"),
 
     'OSPACKAGES_IBVERBS': (('libibverbs-dev', 'libibverbs-devel', 'rdma-core-devel'),
-                        "OS packages providing ibverbs support"),
+                           "OS packages providing ibverbs support"),
     'OSPACKAGES_OPENSSL': (('openssl-devel', 'libssl-dev', 'libopenssl-devel'),
-                        "OS packages providing openSSL support"),
+                           "OS packages providing openSSL support"),
 }
--- a/easybuild/framework/easyconfig/constants.py
+++ b/easybuild/framework/easyconfig/constants.py
@@ -52,6 +52,8 @@ EASYCONFIG_CONSTANTS = {
     'SYS_PYTHON_VERSION': (platform.python_version(), "System Python version (platform.python_version())"),
     'SYSTEM': ({'name': 'system', 'version': 'system'}, "System toolchain"),
 
-    'OSPKGS_IBVERBS': (('libibverbs-dev', 'libibverbs-devel', 'rdma-core-devel'), "OS packages providing ibverbs support"),
-    'OSPKGS_OPENSSL': (('openssl-devel', 'libssl-dev', 'libopenssl-devel'), "OS packages providing openSSL support"),
+    'OSPKGS_IBVERBS': (('libibverbs-dev', 'libibverbs-devel', 'rdma-core-devel'),
+                        "OS packages providing ibverbs support"),
+    'OSPKGS_OPENSSL': (('openssl-devel', 'libssl-dev', 'libopenssl-devel'),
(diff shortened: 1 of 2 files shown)


terjekv:
Usage: `osdependencies = [OSPACKAGES_IBVERBS]`. For perl, one can change `osdependencies = ['openssl', ('openssl-devel', 'libssl-dev', 'libopenssl-devel')]` into `osdependencies = ['openssl', OSPACKAGES_OPENSSL]`.

terjekv:
`Found printed messages in output of test suite\nERROR: test_easyconfig_constants (test.framework.easyconfigparser.EasyConfigParserTest)
File "/tmp/681175494/lib/python2.7/site-packages/easybuild_framework-4.2.1.dev0-py2.7.egg/test/framework/easyconfigparser.py", line 187, in test_easyconfig_constants
self.assertTrue(isinstance(val, (string_type, dict)), "Constant value %s is a string or dict" % val)`

Does this mean that constants can't be lists? If so, either the test needs to be changed, or I'll have to wrap the list in a dict and then unpack at a later stage, but that just seems like circumventing the intent of the test.

boegel:
@terjekv Using a `list` type value (`[1,2,3]`) as a constant is a bad idea because lists are mutable, but using a tuple (`(1,2,3)`) should be fine (and thus changing that check to also accept tuples as constant values is OK).

terjekv:
> @terjekv Using a `list` type value (`[1,2,3]`) as a constant is a bad idea because lists are mutable, but using a tuple (`(1,2,3)`) should be fine (and thus changing that check to also accept tuples as constant values is OK).

Blargh. I fail at terminology. Sorry about that. I ment tuple. I have updated the test code. Let's see if that helps.

easybuild/framework/easyconfig/constants.py
```diff
@@ -51,4 +51,9 @@
     'OS_VERSION': (get_os_version(), "System version"),
     'SYS_PYTHON_VERSION': (platform.python_version(), "System Python version (platform.python_version())"),
     'SYSTEM': ({'name': 'system', 'version': 'system'}, "System toolchain"),
+
+    'OSPACKAGES_IBVERBS': (('libibverbs-dev', 'libibverbs-devel', 'rdma-core-devel'),
+                           "OS packages providing ibverbs support"),
+    'OSPACKAGES_OPENSSL': (('openssl-devel', 'libssl-dev', 'libopenssl-devel'),
```
boegel, line 57: @terjekv To avoid confusion with the `openssl` package, should we rename this constant to `OSPACKAGES_OPENSSL_DEV` or something a bit shorter, like `OS_PKG_OPENSSL_DEV`?

Maybe we should also add `OS_PKG_OPENSSL` (for `'openssl'`) and `OS_PKG_OPENSSL_LIB` (for `('libssl', 'libopenssl')`)?

easybuild/framework/easyconfig/constants.py
```diff
@@ -51,4 +51,9 @@
     'OS_VERSION': (get_os_version(), "System version"),
     'SYS_PYTHON_VERSION': (platform.python_version(), "System Python version (platform.python_version())"),
     'SYSTEM': ({'name': 'system', 'version': 'system'}, "System toolchain"),
+
+    'OSPACKAGES_IBVERBS': (('libibverbs-dev', 'libibverbs-devel', 'rdma-core-devel'),
```
boegel, line 55: To be in sync with `OS_PKG_OPENSSL_DEV` (see below), maybe make this `OS_PKG_IBVERBS_DEV`?

terjekv: Resovled in https://github.com/easybuilders/easybuild-framework/pull/3309/commits/2f967ae6496125390d1d94bda9ecba13cccd7563

easybuild/framework/easyconfig/constants.py
terjekv: Resolved in https://github.com/easybuilders/easybuild-framework/pull/3309/commits/2f967ae6496125390d1d94bda9ecba13cccd7563

test/framework/easyconfigparser.py
```diff
@@ -180,11 +180,12 @@ def test_easyconfig_constants(self):
...
         self.assertEqual(system_constant, {'name': 'system', 'version': 'system'})
 
-        # make sure both keys and values are only strings
+        # make sure both keys and values are of appropriate types
         for constant_name in constants:
             self.assertTrue(isinstance(constant_name, string_type), "Constant name %s is a string" % constant_name)
             val = constants[constant_name]
-            self.assertTrue(isinstance(val, (string_type, dict, tuple)), "Constant value %s is string, dict or tuple" % val)
+            self.assertTrue(isinstance(val, (string_type, dict, tuple)),
+                            "The constant %s has an acceptable type" % constant_name)
```
boegel, line 188: @terjekv Do you mind rewording this to `should have` rather than `has`, which leads to less confusion w.r.t. failing tests (see #3231).

Also, you can still print the value, just wrap it in `str(...)` to avoid problems with tuple values:

```python
fail_msg = "The constant %s should have an acceptable type, found %s (%s)" % (constant_name, type(val), str(val))
self.assertTrue(isinstance(val, (string_type, dict, tuple)), fail_msg)

terjekv: Sorted in https://github.com/easybuilders/easybuild-framework/pull/3309/commits/3b193743172636ebf398dd7db7ee69a8638c8606

boegel:
Thanks a lot @terjekv!

PR #3310, opened 2020-04-30
don't use distutils.dir_util in copy_dir

### LINKED ISSUE #3306: Replace usage of distutils.dir_util
User fizwit reported on Slack that distutils may not be installed, see https://github.com/spack/spack/issues/8968#issuecomment-412647398

From tgamblin

> There is also this issue with distutils.dir_util that led us to stop using it entirely — it’s only designed to be used once in a process:
https://github.com/spack/spack/pull/8182#pullrequestreview-121560040
it actually caches directory creations in a global, so if you somehow end up having to create the same directory twice, it won’t work the second time.

Hence remove the usage of `distutils.dir_util` introduced in https://github.com/easybuilders/easybuild-framework/pull/3270

Possible replacements:

- https://stackoverflow.com/questions/1868714/how-do-i-copy-an-entire-directory-of-files-into-an-existing-directory-using-pyth
- https://spack.readthedocs.io/en/latest/llnl.util.html#llnl.util.filesystem.copy_tree

boegel:
fixes #3306

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -40,7 +40,6 @@ Set of file tools.
 """
 import datetime
 import difflib
-import distutils.dir_util
 import fileinput
 import glob
 import hashlib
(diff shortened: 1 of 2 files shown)


Flamefire:
May I suggest not to switch implementations based on Python version? Why not use a single implementation which behaves as the Python 3.8 version always? Or a defined subset of it. I see no advantage of the usage of Python 3.8 function if a hand-rolled version must still exist. Just makes the test-surface larger

lexming:
Tests with this PR can be found in https://github.com/easybuilders/easybuild-easyblocks/pull/2049

boegel:
> May I suggest not to switch implementations based on Python version? Why not use a single implementation which behaves as the Python 3.8 version always? Or a defined subset of it. I see no advantage of the usage of Python 3.8 function if a hand-rolled version must still exist. Just makes the test-surface larger

The idea is that we can get rid of our own code whenever we reach only having to support Python 3.8 & newer...

I sort of see your point though.

Thoughts on this @lexming?

lexming:
@boegel If our own implementation for Python < 3.8 works just as well in Python 3.8 and onwards, we could indeed use it indiscriminately.

boegel:
@lexming I reproduced your broken use case in the tests and fixed it in e086444; special handling for Python >= 3.8 removed in 3c80fef

lexming:
New tests in https://github.com/easybuilders/easybuild-easyblocks/pull/2049#issuecomment-621758045

lexming:
Going in, thanks @boegel !

PR #3312, opened 2020-04-30
flesh out get_mpi_cmd_template function from Mpi.mpi_cmd_for method

boegel:
@lexming This should come in useful for https://github.com/easybuilders/easybuild-easyblocks/pull/2045


ocaisa:
LGTM Thanks @boegel

PR #3313, opened 2020-04-30
gitignore .mypy_cache dirs

willfurnass:


boegel:
@willfurnass I don't mind merging this, but maybe a better approach for you is to create a top-level `.gitignore` in `$HOME`?

Can you clarify where files like this come from?

willfurnass:
@boegel The [mypy](http://mypy-lang.org/) static type checker.  

I didn't know one could have global `.gitignore`s (but should probably have realised that's possible!  Feel free to reject and close.

boegel:
Thanks for clarifying, doesn't hurt to have this there, so merging...

PR #3314, opened 2020-04-30
Improve install_eb_dep.sh

Flamefire:
- Use correct paths (stray $HOME instead of $PREFIX)
- Use bash parameter expansion instead of `echo | sed`
- Correct quoting
- Cleanup archive and folder
- Reset -e

This script helped me a lot for #3216, so here some improvements for local use.

### DIFF
--- a/easybuild/scripts/install_eb_dep.sh
+++ b/easybuild/scripts/install_eb_dep.sh
@@ -1,77 +1,81 @@
 #!/bin/bash
 
-set -e
-
 if [ $# -ne 2 ]; then
     echo "Usage: $0 <name>-<version> <prefix>"
     exit 1
 fi
-PKG=$1
-PREFIX=$2
 
-PKG_NAME=`echo $PKG | sed 's/-[^-]*$//g'`
-PKG_VERSION=`echo $PKG | sed 's/.*-//g'`
+set -eu
+
+PKG="$1"
+PREFIX="$2"
+
+PKG_NAME="${PKG%-*}"
+PKG_VERSION="${PKG##*-}"
 
 CONFIG_OPTIONS=
 PRECONFIG_CMD=
 
-if [ x$PKG_NAME == 'xmodules' ] && [ x$PKG_VERSION == 'x3.2.10' ]; then
+if [ "$PKG_NAME" == 'modules' ] && [ "$PKG_VERSION" == '3.2.10' ]; then
     PKG_URL="http://prdownloads.sourceforge.net/modules/${PKG}.tar.gz"
     BACKUP_PKG_URL="https://easybuilders.github.io/easybuild/files/${PKG}.tar.gz"
-    export PATH=$PREFIX/Modules/$PKG_VERSION/bin:$PATH
-    export MOD_INIT=$HOME/Modules/$PKG_VERSION/init/bash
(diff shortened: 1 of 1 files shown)


easybuild/scripts/install_eb_dep.sh
```diff
@@ -1,77 +1,81 @@
...
+PREFIX="$2"
+
+PKG_NAME="${PKG%-*}"
+PKG_VERSION="${PKG##*-}"
 
 CONFIG_OPTIONS=
 PRECONFIG_CMD=
 
-if [ x$PKG_NAME == 'xmodules' ] && [ x$PKG_VERSION == 'x3.2.10' ]; then
+if [ "$PKG_NAME" == 'modules' ] && [ "$PKG_VERSION" == '3.2.10' ]; then
```
akesandgren, line 19: Use = and not == for POSIX comformance in all if [ statements comparing strings

easybuild/scripts/install_eb_dep.sh
```diff
@@ -1,77 +1,81 @@
...
-mkdir -p ${PREFIX}
-set +e
-wget ${PKG_URL} && tar xfz *${PKG_VERSION}.tar.gz
-if [ $? -ne 0 ] && [ ! -z $BACKUP_PKG_URL ]; then
-    rm -f *${PKG_VERSION}.tar.gz
-    wget ${BACKUP_PKG_URL} && tar xfz *${PKG_VERSION}.tar.gz
+mkdir -p "${PREFIX}"
+if ! wget "${PKG_URL}" && [ -n "$BACKUP_PKG_URL" ]; then
+    rm -f ./*"${PKG_VERSION}".tar.gz
+    wget "${BACKUP_PKG_URL}"
```
akesandgren, line 57: If this fails, should probably bail out. Same for other wget's below

easybuild/scripts/install_eb_dep.sh
```diff
@@ -1,77 +1,81 @@
...
-if [ x$PKG_NAME == 'xmodules-tcl' ]; then
-    mv modules $PREFIX/${PKG}
+if [ "$PKG_NAME" == 'modules-tcl' ]; then
+    mv modules "$PREFIX/${PKG}"
 else
-    cd ${PKG}
-    if [[ ! -z $PRECONFIG_CMD ]]; then
-        eval ${PRECONFIG_CMD}
+    cd "${PKG}"
+    if [[ -n "$PRECONFIG_CMD" ]]; then
```
akesandgren, line 73: Why the [[ here?

easybuild/scripts/install_eb_dep.sh
Flamefire: The script has a bash shebang and AFAIK == is bash and worked before. Don't really want to change something like that unless you see value in doing so

easybuild/scripts/install_eb_dep.sh
Flamefire: The script has `set -e` which bails out on any error already

easybuild/scripts/install_eb_dep.sh
Flamefire: Was there before, didn't change that. Just changed the "not not" to "is" semantic

akesandgren:
Going in, thanks @Flamefire!

PR #3316, opened 2020-04-30
print trace message for sanity check command before running it

boegel:
For some installations, we run sanity check commands that take a while, so it's good to have `--trace` print the command that's being run before it starts running...


akesandgren:
Going in, thanks @boegel!

PR #3320, opened 2020-05-01
make test_find_eb_script more robust in case $EB_SCRIPT_PATH is already set

boegel:


test/framework/filetools.py
```diff
@@ -2027,19 +2027,35 @@ def makedirs_in_test(*paths):
 
     def test_find_eb_script(self):
         """Test find_eb_script function."""
+
+        # make sure $EB_SCRIPT_PATH is not set already (used as fallback mechanism in find_eb_script)
+        if 'EB_SCRIPT_PATH' in os.environ:
```
akesandgren, line 2032: Don't you need to reset this to the original value at the end of this test?

boegel: The environment is already automatically reset to the original at the end of each test (see `tearDown` in `test/framework/utilities.py`).

akesandgren:
Going in, thanks @boegel!

PR #3321, opened 2020-05-01
fix broken test_toy_lock_cleanup_signals & properly clean up after using SIGALRM signal in tests

boegel:
2nd test case in `test_toy_lock_cleanup_signals` was failing with `Can't change back to non-existing directory` (due to changes to `extract_file` in #3292).

Before the alarm wasn't disabled properly, which caused the alarm to trip when the next test (`test_toy_modaltsoftname`) was running, leading to a lot of confusion...


akesandgren:
Going in, thanks @boegel!

Issue #3323, opened 2020-05-04
Specify dependency search path within an EasyConfig file


kelseymh:
I recently had a situation where I need to pick up a "hand-written" (script generated) Lmod module file that is included in a group-specific software build.  This file isn't meant to be installed in a sitewide modules area, and we are not wanting our users to have to know or specify the path to this module themselves.

I wrote an EasyBlock file to support creating a "public" EasyBuild Bundle to wrap this special module, along with the infrastructure it needs (toolchain, Python, etc.).  The EasyBlock only implements an overload of `make_module_deps()`:

    def make_module_dep(self, unload_info=None):
        dependencydir = self.cfg['dependencydir']
        if dependecydir:
            txt += self.module_generator.prepend_paths('MODULEPATH', dependencydir, allow_abs=True)

        txt += super(EB_scdms_minus_release, self).make_module_dep()        
        return txt

(the real code has existence checks, and currently only supports a single directory for my particular case).

I wonder if this would be a useful feature to add to `easyblock.py`.  I'm thinking of a new parameter, `moddeppath = '/path/to/dependencies'` or `moddeppath = [ '/one/path', '/two/path' ]`.  These would typically be absolute paths, and get added to `MODULEPATH` as I did above, before the dependencies' `load()` actions were generated.  I'd like to also include a directive in the output module file to strip these paths back out of `MODULEPATH` after the `load()` actions (so the location isn't exposed to end users), but I didn't find a function to do that in `module_generator.py`.

Does this seem like useful functionality for EasyConfigs?  Or is my use case sufficiently obscure that it doesn't need a general solution?

boegel:
@kelseymh I think this is a sensible use case to support.

It hasn't popped up before, as far as I know, but that doesn't mean it's not relevant for others.

And it should be relatively easy to implement support for this without introducing problems elsewhere.

kelseymh:
Okay!  Since I've already got the code in my local EasyBlock, I'll work on porting into `easyblock.py`, and see how well I can do treating all of the various type cases properly.

boegel:
Closed since support for `moddependpaths` easyconfig paramter was implemented in #3324 

Thanks a lot for your efforts on this @kelseymh! :+1:

PR #3324, opened 2020-05-05
New variable 'moddependpaths' to resolve dependencies at load time.

kelseymh:
Following #3323 , a new EasyConfig parameter is introduced in the DEPENDENCIES section to allow EasyConfig writers to specify an additional path to find dependency modules within the EasyConfig.  The `moddependpaths` path (or paths, it may be either a string or a list/tuple) is prepended(*) to MODULEPATH _before_ load actions in the generated Lmod file.

A new function in `easyblock.py`, `make_module_deppaths()` is introduced to support this, and a unit test for it is provided.

(*) `module_generator.use()` is invoked, rather than explicitly specifying MODULEPATH.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1022,8 +1022,8 @@ class EasyBlock(object):
         deppaths = self.cfg['moddependpaths']
         if not deppaths:
             return ''
-        elif not isinstance(deppaths, (str, tuple, list)):
-            raise EasyBuildError("moddependpaths value %s (type: %s) is not a string or collection",
+        elif not isinstance(deppaths, (str, list, tuple)):
+            raise EasyBuildError("moddependpaths value %s (type: %s) is not a string, list or tuple",
                                  deppaths, type(deppaths))
 
         if isinstance(deppaths, str):
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1014,6 +1014,25 @@ class EasyBlock(object):
         # cleanup: unload fake module, remove fake module dir
         self.clean_up_fake_module(fake_mod_data)
 
+    def make_module_deppaths(self):
+        """
+        Add specific 'module use' actions to module file, in order to find
+        dependencies outside the end user's MODULEPATH.
+        """
+        deppaths = self.cfg['moddependpaths']
+        if not deppaths:
+            return ''
+        elif not isinstance(deppaths, (str, tuple, list)):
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1014,6 +1014,25 @@ def make_devel_module(self, create_in_builddir=False):
...
+    def make_module_deppaths(self):
+        """
+        Add specific 'module use' actions to module file, in order to find
+        dependencies outside the end user's MODULEPATH.
+        """
+        deppaths = self.cfg['moddependpaths']
+        if not deppaths:
+            return ''
+        elif not isinstance(deppaths, (str, tuple, list)):
+            raise EasyBuildError("moddependpaths value %s (type: %s) is not a string or collection",
```
ocaisa, line 1026: Sorry to be nit-picky here but `collection` is a loaded term ([at least in Lmod](https://lmod.readthedocs.io/en/latest/010_user.html#user-collections)), maybe just say `string, list or tuple`?

easybuild/framework/easyconfig/default.py
```diff
@@ -154,6 +154,7 @@
     'multi_deps': [{}, "Dict of lists of dependency versions over which to iterate", DEPENDENCIES],
     'multi_deps_load_default': [True, "Load module for first version listed in multi_deps by default", DEPENDENCIES],
     'osdependencies': [[], "OS dependencies that should be present on the system", DEPENDENCIES],
+    'moddependpaths': [None, "Absolute path or paths that should be searched for dependencies", DEPENDENCIES],
```
ocaisa, line 157: I think we need to be more explicit about what this is, this could potentially be confusing (are we talking about EB or the module tool when we talk about dependencies?). How about something like:
```
Absolute path or paths to prepend to MODULEPATH before loading the resulting module file
```

easybuild/framework/easyblock.py
kelseymh: Thanks.  I was hoping to be concise, and have one word for "a bunch of things."

easybuild/framework/easyconfig/default.py
kelseymh: Hmmm.  Since it's included in the DEPENDENCIES section, shouldn't it be clear to an EC author what "dependencies" refers to?  How about

    Absolute path or paths to prepend to MODULEPATH before loading the module's dependencies

ocaisa:
Going in, thanks @kelseymh !

PR #3326, opened 2020-05-06
Only calculate the mapping of version suffixes if they are actually used

mboisson:
Otherwise, at our site, when using multiple paths in `EASYBUILD_ROBOT_PATHS`, this fails with 
`ERROR: No unique versionsuffix mapping for -Python-2.7.10 in {'name': 'GCC', 'version': '7.3.0'} toolchain hierarchy to {'name': 'GCC', 'version': '9.3.0'} toolchain hierarchy` 

While this may or may not be a bug in the easyconfigs we have in our robot_paths (we may have duplicated -Python version suffixes for example), we don't usually use versionsuffixes, and for the easyconfig I was trying to build, there are absolutely no need to have `versionsuffix` mapped, since none is used.

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -1073,9 +1073,6 @@ def list_deps_versionsuffixes(ec_spec):
 
     versionsuffix_list = []
     for key in DEPENDENCY_PARAMETERS:
-        # loop over a *copy* of dependency dicts (with resolved templates);
-
-        # to update the original dep dict, we need to get a reference with templating disabled...
         val = parsed_ec[key]
 
         if key in parsed_ec.iterate_options:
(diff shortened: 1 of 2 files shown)


mboisson:
Asking for @ocaisa's review of this PR.

easybuild/framework/easyconfig/tweak.py
```diff
@@ -1059,6 +1060,34 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+
+    :return: The list of versionsuffixes used by the dependencies of this recipe
+    """
+    # Fully parse the original easyconfig
+    parsed_ec = process_easyconfig(ec_spec, validate=False)[0]['ec']
+
+    versionsuffix_list = []
+    for key in DEPENDENCY_PARAMETERS:
+        # loop over a *copy* of dependency dicts (with resolved templates);
+
```
ocaisa, line 1077: These comments are a bit off, remove altogether or update

mboisson: oops, indeed, that's a copy from another function. Removed.

ocaisa:
See https://github.com/ComputeCanada/easybuild-framework/pull/23 to improve the error reporting that triggered this PR

mboisson:
> See [ComputeCanada#23](https://github.com/ComputeCanada/easybuild-framework/pull/23) to improve the error reporting that triggered this PR

merged.

ocaisa:
Going in as soon as the tests pass, thanks @mboisson !

PR #3330, opened 2020-05-08
check easyconfig itself as well as deps for versionsuffix when using --try-update-deps

ocaisa:
This is required as we may need to use the versionsuffix mapping for the easyconfig versionsuffix.

Fix bug introduced by #3326:
```
alanc@~$ eb --experimental --dry-run --robot --try-toolchain=foss,2019a --minimal-toolchains --try-update-deps --try-software-version=2.2.5 Keras-2.2.4-foss-2018b-Python-3.6.6.eb
== temporary log file in case of crash /tmp/eb-omkGKx/easybuild-rhrl3H.log
ERROR: Failed to process easyconfig /tmp/eb-omkGKx/tweaked_easyconfigs/Keras-2.2.5-foss-2019a-Python-3.7.2.eb: Failed to determine minimal toolchain for dep Theano 1.0.3-Python-3.7.2
```

This was due to the filename of Theano not being correct when writing the tweaked easyconfig.

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -958,7 +958,7 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
     versonsuffix_mapping = {}
     # We only need to map versionsuffixes if we are updating dependency versions and if there are
     # versionsuffixes being used in dependencies
-    if update_dep_versions and list_deps_versionsuffixes(ec_spec):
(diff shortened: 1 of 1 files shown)


ocaisa:
@mboisson Can you check that the problem you were trying to avoid is still avoided with this fix?

mboisson:
I don't have much time today to test this, but reading the code, I don't see how it could break. And it makes sense to keep checking versionsuffix if the recipe you are trying to build includes such

PR #3332, opened 2020-05-08
allow copying of tweaked easyconfigs when using --try-* with --copy-ec

ocaisa:
Requires ~~#3330~~

Example usage:
```
eb --experimental --copy-ec --dry-run --robot --try-toolchain=foss,2019a --minimal-toolchains --try-update-deps --try-software-version=2.2.5 Keras-2.2.4-foss-2018b-Python-3.6.6.eb my_copy_of_tweaked_ecs
```
This will only copy the tweaked easyconfigs that are (or would be) actually used to resolve dependencies.


mboisson:
@ocaisa, this does not call `clean_up_easyconfigs`. Is it possible to add this ?

ocaisa:
I don't know what that is yet...but yes

mboisson:
https://github.com/easybuilders/easybuild-framework/blob/018337e94442d83ec32fd9a9a242273b69996b1b/easybuild/framework/easyconfig/easyconfig.py#L2250

PR #3333, opened 2020-05-08
avoid empty entries in $LD_LIBRARY_PATH and other path-like environment variables

boegel:
fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/9843

### DIFF
--- a/easybuild/tools/environment.py
+++ b/easybuild/tools/environment.py
@@ -189,13 +189,13 @@ def sanitize_env():
     """
     Sanitize environment.
 
-    This function
+    This function:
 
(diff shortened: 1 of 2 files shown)


easybuild/tools/environment.py
```diff
@@ -189,18 +189,39 @@ def sanitize_env():
     """
     Sanitize environment.
 
-    This function undefines all $PYTHON* environment variables,
-    since they may affect the build/install procedure of Python packages.
+    This function
```
ocaisa, line 193: Either remove this line or change below to 'filters', 'undefines'

boegel: fixed in 772d823

PR #3337, opened 2020-05-16
reinstate fallback to ModulesTool.module_wrapper_exists in ModulesTool.exist

### LINKED ISSUE #3335: Java/11 not considered present with HMNS on develop
Att: @Flamefire 

The changes in #3216 seem to have caused a regression on my Ubuntu 18.04 machine, where something depending on `Java/11` fails to notice the installed module.

```
$ eb R-4.0.0-foss-2020a.eb -rD
*snip*
 * [x] $CFGS/j/Java/Java-11.0.2.eb (module: Core | Java/11.0.2)
 * [ ] $CFGS/j/Java/Java-11.eb (module: Core | Java/11)
 * [x] $CFGS/n/NLopt/NLopt-2.6.1-GCCcore-9.3.0.eb (module: Compiler/GCCcore/9.3.0 | NLopt/2.6.1)
 * [x] $CFGS/l/libsndfile/libsndfile-1.0.28-GCCcore-9.3.0.eb (module: Compiler/GCCcore/9.3.0 | libsndfile/1.0.28)
*snip*
 * [x] $CFGS/s/ScaLAPACK/ScaLAPACK-2.1.0-gompi-2020a.eb (module: MPI/GCC/9.3.0/OpenMPI/4.0.3 | ScaLAPACK/2.1.0)
 * [x] $CFGS/f/foss/foss-2020a.eb (module: Core | foss/2020a)
 * [ ] $CFGS/r/R/R-4.0.0-foss-2020a.eb (module: MPI/GCC/9.3.0/OpenMPI/4.0.3 | R/4.0.0)
```

This works fine on 4.2.0, but fails on 4.2.0 w/ 3216 applied and on current develop.

The build attempts to find `Core/Java/11` which doesn't resolve with my module path, while `Java/11` does.

boegel:
This reinstates the fallback mechanism that was removed in #3216, since it's required to correctly recognize existing module wrappers when using Lmod and a hierarchical module naming scheme.

fixes #3335

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -487,10 +487,7 @@ class ModulesTool(object):
         """
         Determine whether a module wrapper with specified name exists.
         Only .modulerc file in Tcl syntax is considered here.
-        DEPRECATED. Use exists()
         """
-        self.log.deprecated('module_wrapper_exists is unreliable and should no longer be used. ' +
-                            'Use exists instead to check for an existing module or alias.', '5.0')
 
         if mod_wrapper_regex_template is None:
             mod_wrapper_regex_template = "^[ ]*module-version (?P<wrapped_mod>[^ ]*) %s$"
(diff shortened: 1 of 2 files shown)


zao:
This PR makes it work on the machine mentioned in #3335.

akesandgren:
BTW, having a "." as first char of module name isn't the only way to hide it...
I don't know if using Lmods hide_modulefile causes any problems here.
We do use if but mostly to get rid of builds we deem problematic.

boegel:
> BTW, having a "." as first char of module name isn't the only way to hide it...
> I don't know if using Lmods hide_modulefile causes any problems here.
> We do use if but mostly to get rid of builds we deem problematic.

No issue, because we always use Lmod's `--show-hidden` option.

akesandgren:
Going in, thanks @boegel!

PR #3341, opened 2020-05-20
fix crash in --avail-easyconfig-constants when using --output-format=rst + ensure sorted output

boegel:
Fix for:
```
$ eb --avail-easyconfig-constants --output-format rst
ERROR: Failed to parse configuration options: not all arguments converted during string formatting
```

Side effect of #3309, went unnoticed because there was no test yet for `--avail-easyconfig-constants`...


easybuild/tools/docs.py
```diff
@@ -159,10 +159,12 @@ def avail_easyconfig_constants_rst():
         "Description",
     ]
 
+    sorted_keys = sorted(EASYCONFIG_CONSTANTS.keys())
```
bartoldeman, line 162: why not `sorted(EASYCONFIG_CONSTANTS)` ?

PR #3346, opened 2020-05-25
always take into account builddependencies when generating template values, also when we're not iteraring over builddependencies

### LINKED ISSUE #2114: pyshortver deferencing fails when Python is not a full dependency
I was testing an easyconfig with `Python` as a builddependency, not a full dependency.

In `postinstallcmds`, I tried to use `pyshortver` but this wasn't resolved and instead was passed as a raw value, causing a crash.

This should be picked up earlier, because it's a somewhat problematic configuration anyway.

### LINKED ISSUE #3300: Templates not being resolved if the triggering package is a build dependency, when using minimal toolchains
Python as build dependency, with and without `EASYBUILD_MINIMAL_TOOLCHAINS`

```
$ cat build.eb
easyblock = 'ConfigureMake'

name = 'test'
version = 'test'

homepage = ''
description = """ """

toolchain = {'name': 'GCCcore', 'version': '8.3.0'}

builddependencies = [
    ('Python', '3.6.8'),
    ('libxml2', '2.9.9', '-Python-%(pyver)s'),
]

moduleclass = 'system'
$ eb --check-style -r . build.eb
== temporary log file in case of crash /path/to/log

Running style check on 1 easyconfig(s)...

[PASS] /path/to/build.eb

>> All style checks PASSed!
...
$ EASYBUILD_MINIMAL_TOOLCHAINS=1 eb --check-style -r . build.eb
== temporary log file in case of crash /path/to/log
ERROR: Failed to process easyconfig /path/to/build.eb: Failed to determine minimal toolchain for dep libxml2 2.9.9-Python-%(pyver)s
```
Python as build dependency, without using templates, with `EASYBUILD_MINIMAL_TOOLCHAINS`

```
$ cat build2.eb
easyblock = 'ConfigureMake'

name = 'test'
version = 'test'

homepage = ''
description = """ """

toolchain = {'name': 'GCCcore', 'version': '8.3.0'}

builddependencies = [
    ('Python', '3.6.8'),
    ('libxml2', '2.9.9', '-Python-3.6.8'),
]

moduleclass = 'system'
$ EASYBUILD_MINIMAL_TOOLCHAINS=1 eb --check-style -r . build2.eb
== temporary log file in case of crash /path/to/log

Running style check on 1 easyconfig(s)...

[PASS] /path/to/build2.eb

>> All style checks PASSed!
...
```
Python as normal dependency, with `EASYBUILD_MINIMAL_TOOLCHAINS`

```
cat build3.eb
easyblock = 'ConfigureMake'

name = 'test'
version = 'test'

homepage = ''
description = """ """

toolchain = {'name': 'GCCcore', 'version': '8.3.0'}

dependencies = [
    ('Python', '3.6.8'),
    ('libxml2', '2.9.9', '-Python-%(pyver)s'),
]

moduleclass = 'system'
$ EASYBUILD_MINIMAL_TOOLCHAINS=1 eb --check-style -r . build3.eb
== temporary log file in case of crash /path/to/log

Running style check on 1 easyconfig(s)...

[PASS] /path/to/build3.eb

>> All style checks PASSed!
```

ekouts:
fixes #3300, fixes #2114

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -223,7 +223,8 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
         # first, determine if we have an EasyConfig instance
         # (indirectly by checking for 'iterating' and 'iterate_options' attributes,
         #  because we can't import the EasyConfig class here without introducing
-        #  a cyclic import...)
+        #  a cyclic import...);
+        # we need to know to determine whether we're iterating over a list of build dependencies
         is_easyconfig = hasattr(config, 'iterating') and hasattr(config, 'iterate_options')
 
         if is_easyconfig:
@@ -234,11 +235,6 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
                     deps += config.get('builddependencies', [])
             else:
                 deps += config.get('builddependencies', [])
-
-        # if we're not dealing with an EasyConfig instance,
-        # we can just add list of builddependencies directly (no list of lists in that case)
-        else:
-            deps += config.get('builddependencies', [])
 
         for dep in deps:
             if isinstance(dep, dict):
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -227,7 +227,7 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
         is_easyconfig = hasattr(config, 'iterating') and hasattr(config, 'iterate_options')
 
         if is_easyconfig:
-           # if we're iterating over different lists of build dependencies,
+            # if we're iterating over different lists of build dependencies,
             # only consider build dependencies when we're actually in iterative mode!
             if 'builddependencies' in config.iterate_options:
                 if config.iterating:
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -227,7 +227,7 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
         is_easyconfig = hasattr(config, 'iterating') and hasattr(config, 'iterate_options')
 
         if is_easyconfig:
-            # if we're iterating of different lists of build dependencies,
+           # if we're iterating over different lists of build dependencies,
             # only consider build dependencies when we're actually in iterative mode!
             if 'builddependencies' in config.iterate_options:
                 if config.iterating:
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyconfig/templates.py
```diff
@@ -216,9 +216,12 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
...
 
-        # only consider build dependencies for defining *ver and *shortver templates if we're in iterative mode
-        if hasattr(config, 'iterating') and config.iterating:
-            deps += config.get('builddependencies', [])
+        if hasattr(config, 'iterating'):
+            if 'builddependencies' in config.iterate_options:
+                if config.iterating:
+                    deps += config.get('builddependencies', [])
+            else:
+                deps += config.get('builddependencies', [])
```
boegel, line 224: @ekouts I think we can simplify this whole block, there's no point in being overly careful I think...

The iterating part is now basically irrelevant, we can always consider build deps to define template values:

```suggestion
        deps += config.get('builddependencies', [])
```

boegel: No, my suggestion is too simplistic... We need to take into account that `builddependencies` can be a list of lists while we're doing an iterative installation (like when `multi_deps` is used), but we also just want to add all builddependencies if we're not doing an iterated installation...

easybuild/framework/easyconfig/templates.py
```diff
@@ -216,8 +216,28 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
...
+        # (when multi_deps is used for example) the builddependencies value may be a list of lists
+
+        # first, determine if we have an EasyConfig instance
+        # (indirectly by checking for 'iterating' and 'iterate_options' attributes,
+        #  because we can't import the EasyConfig class here without introducing
+        #  a cyclic import...)
+        is_easyconfig = hasattr(config, 'iterating') and hasattr(config, 'iterate_options')
+
+        if is_easyconfig:
+            # if we're iterating of different lists of build dependencies,
```
boegel, line 230: fix typo:

```suggestion
           # if we're iterating over different lists of build dependencies,
```

easybuild/framework/easyconfig/templates.py
```diff
@@ -216,8 +216,28 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
...
+        # (when multi_deps is used for example) the builddependencies value may be a list of lists
+
+        # first, determine if we have an EasyConfig instance
+        # (indirectly by checking for 'iterating' and 'iterate_options' attributes,
+        #  because we can't import the EasyConfig class here without introducing
+        #  a cyclic import...)
+        is_easyconfig = hasattr(config, 'iterating') and hasattr(config, 'iterate_options')
+
+        if is_easyconfig:
+           # if we're iterating over different lists of build dependencies,
```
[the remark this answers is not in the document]
boegel:
```suggestion
            # if we're iterating over different lists of build dependencies,

PR #3350, opened 2020-05-26
Include Perl as a potential case for versionsuffix mapping

### LINKED ISSUE #3349: --test-update-deps + --try-toolchain fails with Perl versionsuffix
`eb --try-toolchain=intel,2019b --robot --experimental --try-update-deps BioPerl-1.7.2-intel-2018b-Perl-5.28.0.eb`
Gives this with minimal-toolchains=False: 
`ERROR: Missing dependencies: XML-LibXML/2.0132-intel-2019b-Perl-5.30.0`
And this with minimal-toolchains=True:
`Failed to determine minimal toolchain for dep XML-LibXML 2.0132-Perl-5.30.0`

I think the expected behavior should be to find XML-LibXML-2.0201-GCCcore-8.3.0.eb, as this is compatible with intel/2019b.

ocaisa:
...when using `--try-update-deps`

Fixes #3349

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -961,6 +961,7 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
     if update_dep_versions and (list_deps_versionsuffixes(ec_spec) or parsed_ec['versionsuffix']):
         # We may need to update the versionsuffix if it is like, for example, `-Python-2.7.8`
         versonsuffix_mapping = map_common_versionsuffixes('Python', parsed_ec['toolchain'], toolchain_mapping)
+        versonsuffix_mapping.update(map_common_versionsuffixes('Perl', parsed_ec['toolchain'], toolchain_mapping))
 
     if update_build_specs is not None:
         if 'version' in update_build_specs:


ocaisa:
This allows the dependency to be resolved in the particular use case:
```
eb --try-toolchain=intel,2019b --robot --experimental --try-update-deps BioPerl-1.7.2-intel-2018b-Perl-5.28.0.eb -D
```
but an alternative match actually exists, `XML-LibXML-2.0201-GCCcore-8.3.0.eb`, this isn't chosen because we changed convention and decided to drop the `Perl` suffix recently. We can't predict a change in convention, howver you could build in logic to account for this...but we would first need to agree on exactly what our conventions are :P

ocaisa:
If this were me I would recommend doing a dry run with the `--copy-ec` parameter:
```
eb --try-toolchain=intel,2019b --robot --experimental --try-update-deps --copy-ec -D BioPerl-1.7.2-intel-2018b-Perl-5.28.0.eb my_tweaked_ecs
```
then going in and fiddling with the dep to use the latest version and then building things.

ocaisa:
I guess I could add a feature to provide a warning if there is a case where a version exists without a versionsuffix and is newer (would be particularly useful for Python I suspect).

Issue #3348, opened 2020-05-26
self.cfg.update() appears to not handle a dictionary value


kelseymh:
In a custom (project specific) EasyBlock, I am overloading the `make_module_extra()` action to set a large number of required envvars.  To avoid the long, cumbersome lines of "`txt += self.module_generator.set_environment()`" with even longer string constructions as arguments, I thought I could just add to self.cfg['modextravars'] and let the base class take care of it.

In the EB, I wrote:

        batsdir = os.path.join(self.installdir, 'share', self.name)
        batsset = os.path.join(batsdir, 'UserSettings', 'BatRootSettings')
        bats_envvars = {
            'CDMSBATSDIR': batsdir,
            'BATCALIB_RQDATA': os.path.join(batsdir, 'NoiseFiles'),
            'BATCALIB_RRQDATA': os.path.join(batsdir, 'NoiseFiles'),
            'BATNOISE_TEMPLATES': os.path.join(batsdir, 'PulseTemplates'),
            'BATROOT_NOISEFILES': os.path.join(batsdir, 'FilterFiles'),
            'BATROOT_RQDATA': os.path.join(batsdir, 'RQData'),
            'BATROOT_PROC': os.path.join(batsset, 'processing'),
            'BATROOT_CONST': os.path.join(batsset, 'analysis'),
            'BATROOT_CALIB': os.path.join(batsset, 'calibration'),
        }

        self.cfg.update('modextravars', bats_envvars)

Which should have worked (thanks, @boegel !).  Instead, it complained that `bats_envvars` "is not a string or list":

    == 2020-05-26 00:11:36,043 build_log.py:169 ERROR EasyBuild crashed with an erro
    r (at easybuild/base/exceptions.py:124 in __init__): Can't update configuration 
    value for modextravars, because the attempted update value, '{'BATROOT_CALIB': '
    [...]
    msbats/FilterFiles'}', is not a string or list. (at easybuild/framework/easyconf
    ig/easyconfig.py:586 in update)

Since 'modextravars' is supposed to be a dictionary in the EasyConfig file, a dictionary value supplied in `self.cfg.update()` should work.

For the record, I worked around the problem using an assignment:

    self.cfg['modextravars'] = bats_envvars

and produced a module file with all of my needed envvars properly set.

boegel:
I guess we never ran into a case where passing a `dict` value to `self.cfg.update` was needed...

[...]

PR #3356, opened 2020-05-28
Add support for updating dictionary or tuple keys in self.cfg

kelseymh:
Following #3348 , I've added the necessary `elif` blocks to allow updating of dictionary-type keys (like 'sources' or 'modextravars') in EasyBlock `self.cfg`.  For completeness, I've also added support for extending tuples (using the cast-to-list-and-back-again method).

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -587,7 +587,7 @@ class EasyConfig(object):
             raise EasyBuildError(msg, key, value)
 
         # For dictionaries, input value cannot be a string; must be iterable
-        if isinstance(self[key], dict) and not isinstance(value, string_type):
+        if isinstance(self[key], dict) and isinstance(value, string_type):
             msg = "Can't update configuration value for %s, because the attempted"
             msg += "update value, '%s', is not iterable (list, tuple, dict)."
             raise EasyBuildError(msg, key, value)
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -575,14 +575,15 @@ class EasyConfig(object):
     def update(self, key, value, allow_duplicate=True):
         """
         Update a string configuration value with a value (i.e. append to it).
+        NOTE: For dictionaries, 'allow_duplicate' will be ignored.
         """
         if isinstance(value, string_type):
             lval = [value]
-        elif isinstance(value, list):
+        elif isinstance(value, (list, dict, tuple)):
             lval = value
         else:
-            msg = "Can't update configuration value for %s, because the "
-            msg += "attempted update value, '%s', is not a string or list."
+            msg = "Can't update configuration value for %s, because the attempted"
+            msg += " update value, '%s', is not a string, list or dictionary."
             raise EasyBuildError(msg, key, value)
 
         param_value = self[key]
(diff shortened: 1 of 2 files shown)


kelseymh:
The last round of tests was interesting, and taught me something about Python I didn't know.  I implemented the dictionary update test along the lines of the others:

https://github.com/kelseymh/easybuild-framework/blob/282b739e4b9a087ac6f050a65ee2a30091d19563/test/framework/easyconfig.py#L1747

This assumes (as was the case when tried it by hand) that the dictionary keeps the keys in the order in which they were inserted, not in an arbitrary hashed order.  As explained to me by [StackOverflow](https://stackoverflow.com/questions/5629023/the-order-of-keys-in-dictionaries), that's only true for Python 3.6 and later; it's not guaranteed in earlier versions.

I confirmed this using Python/3.5.2:
```
>>> mydict = { 'a':'vala', 'b':'valb', 'c':'valc' }
>>> print(mydict)
{'a': 'vala', 'b': 'valb', 'c': 'valc'}
>>> mydict.update({'d':'vald'})
>>> print(mydict)
{'a': 'vala', 'd': 'vald', 'b': 'valb', 'c': 'valc'}
```
So here, using .endswith() would fail, as it did in three of the four Python 3.5 tests.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -595,8 +596,17 @@ def update(self, key, value, allow_duplicate=True):
             for item in lval:
                 if allow_duplicate or item not in param_value:
                     param_value = param_value + [item]
+        elif isinstance(param_value, tuple):
+            param_value = list(param_value)
+            for item in lval:
+                if allow_duplicate or item not in param_value:
+                    param_value.append(item)
+            param_value = tuple(param_value)
```
boegel, line 604: This could actually be collapsed with the case above, since this approach works both for `list` and `tuple` values.

This line:

```python
param_value = list(param_value)
```

will result in a copy of the original in case of a `list` value, to which you can safely `.append` (we don't do that in the code above to avoid changing the original `param_value` in place.

```python
elif isinstance(param_value, (list, tuple)):
    param_value = list(param_value)
    ...

File: easybuild/framework/easyconfig/easyconfig.py [hunk 2]
```diff
@@ -595,8 +596,17 @@ def update(self, key, value, allow_duplicate=True):
...
+        elif isinstance(param_value, dict):
+            param_value.update(lval)
```
boegel, line 606: Two things here:

* You shouldn't do an in-place update of `param_value` here, just like we avoid above in the `list` case, but you should take a copy and update that instead. Below we replace the actual easyconfig parameter value (`self[key]`) with the updated `param_value`.

* You shouldn't assume `lval` is a `dict` here, just because `param_value` is. What if `lval` is a `list` of `tuple`?
   In that case we can't actually do the update (since we don't have key-value pairs.
   Similarly above: what if `param_value` is a `list`, but `lval` is a `dict`?

easybuild/framework/easyconfig/easyconfig.py
kelseymh: Nice!  Collapsing if blocks is something I really like to do in my own code.

easybuild/framework/easyconfig/easyconfig.py
kelseymh: I thought param_value was already a local copy, made with the intent of doing in-place updates before pushing it back onto the original?

I see your point about handling all of the crossed cases where the user's input isn't necessarily the same as the target type (lval).

easybuild/framework/easyconfig/easyconfig.py
kelseymh: See my updated comment below.  I think ([line 589](https://github.com/kelseymh/easybuild-framework/blob/4fd2205a007f37ac2efbd713636fa206fa4a7e1a/easybuild/framework/easyconfig/easyconfig.py#L589) that param_value is already a local copy that the function manipulates, then assigns back into `self[key]`.

easybuild/framework/easyconfig/easyconfig.py
kelseymh: See also above.  I am going to change the name, because (at least to me, coming from C++) the name "lval" is really misleading; to me, it implies "left-hand value".  I think it's supposed to mean "*l*ocal *val*ue" or something.  I'd rather call it "inval" ("*in*put *val*ue") or something.

boegel:
Thanks a lot @kelseymh!

Issue #3359, opened 2020-05-29
Crash when using --amend of a local_* variable together with `--try-toolchain`


Flamefire:
When using `eb toy.eb --try-amend=local_foo=123 --try-toolchain=GCC,9.2.0-2.32` it crashes with
> Use of unknown easyconfig parameter 'local_foo' when setting parameter value to '123'

Stacktrace:
```
  File "/opt/software/Core/EasyBuild/4.2.1/lib/python2.6/site-packages/easybuild/main.py", line 533, in <module>
    main()
  File "/opt/software/Core/EasyBuild/4.2.1/lib/python2.6/site-packages/easybuild/main.py", line 386, in main
    easyconfigs = tweak(easyconfigs, build_specs, modtool, targetdirs=tweaked_ecs_paths)
  File "/opt/software/Core/EasyBuild/4.2.1/lib/python2.6/site-packages/easybuild/framework/easyconfig/tweak.py", line 200, in tweak
    update_dep_versions=update_dependencies)
  File "/opt/software/Core/EasyBuild/4.2.1/lib/python2.6/site-packages/easybuild/framework/easyconfig/tweak.py", line 993, in map_easyconfig_to_target_tc_hierarchy
    parsed_ec[key] = update_build_specs[key]
  File "/opt/software/Core/EasyBuild/4.2.1/lib/python2.6/site-packages/easybuild/framework/easyconfig/easyconfig.py", line 124, in new_ec_method
    return ec_method(self, key, *args, **kwargs)
  File "/opt/software/Core/EasyBuild/4.2.1/lib/python2.6/site-packages/easybuild/framework/easyconfig/easyconfig.py", line 1673, in __setitem__
```

A workaround is to use `--disable-map-toolchains` which uses regex to change the EC

Flamefire:
As a quick fix I tried to skip local_ variables in the loop at https://github.com/easybuilders/easybuild-framework/blob/c89945d571ea90256e35c1aa23ced38609dbed01/easybuild/framework/easyconfig/tweak.py#L992

This does not work as it means that the updated variable is not used in the install_cmd leading to an empty README. I assume this affects more cases like that and am wondering if the approach of modifying the EC instance is good/safe at all.

ocaisa:
It's possible to only allow amend to use regex mode, I can make a PR for that next week...but then you can't couple it with some of the `try-*` options

Flamefire:
[...]

PR #3370, opened 2020-06-22
Try a simple regex on tweaked files rather than rely on dumping

ocaisa:
First efforts on resolving #3359

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -1060,9 +1060,34 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
     tweaked_spec = os.path.join(targetdir or tempfile.gettempdir(), ec_filename)
 
     parsed_ec.dump(tweaked_spec, always_overwrite=False, backup=True)
+    regex_attempt_on_ecspec = attempt_regex_substitution(ec_spec, tweaked_spec)
+    if regex_attempt_on_ecspec:
+        # Prefer a simple regex substitution version
+        pass
     _log.debug("Dumped easyconfig tweaked via --try-* to %s", tweaked_spec)
 
     return tweaked_spec
+
+
+def attempt_regex_substitution(original_spec, tweaked_spec):
+    regexed_spec = None
+
+    original_ecdict = EasyConfig(original_spec).asdict()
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyconfig/tweak.py
```diff
@@ -1060,11 +1060,36 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
...
+        pass
     _log.debug("Dumped easyconfig tweaked via --try-* to %s", tweaked_spec)
 
     return tweaked_spec
 
 
+def attempt_regex_substitution(original_spec, tweaked_spec):
+    regexed_spec = None
+
+    original_ecdict = EasyConfig(original_spec).asdict()
```
ocaisa, line 1075: Unfortunately the templating is not always resolved in parameters such as `postinstallcmds`  in this approach so the matching is not good enough. We need to completely remove possible templates from the raw txt to be able to compare like with like. With this I get
```
alanc@~/EasyBuild_Git/easybuild-framework(regex_tweaking)$ time eb --experimental --dry-run --robot --try-toolchain=foss,2019a --try-update-deps mpiP-3.4.1-iimpi-2019a.eb
== temporary log file in case of crash /tmp/eb-73awbqw5/easybuild-n5kshsrz.log

WARNING: Your request will result in a compiler family switch (Intel to GCC). Here be dragons!


WARNING: Your request will result in a compiler family switch (Intel to GCC). Here be dragons!

{}
{}
{'toolchain': ({'name': 'iimpi', 'version': '2019a'}, {'name': 'gompi', 'version': '2019a'}), 'postinstallcmds': (["sed -e 's,MPIP_LIB_DIR=.*,MPIP_LIB_DIR=%(installdir)s/lib,' -i bin/mpirun-mpip", "sed -e 's,MPIP_DIR=.*,MPIP_DIR=%(installdir)s,' -i bin/srun-mpip"], ["sed -e 's,MPIP_LIB_DIR=.*,MPIP_LIB_DIR=%(installdir)s/lib,' -i bin/mpirun-%(namelower)s", "sed -e 's,MPIP_DIR=.*,MPIP_DIR=%(installdir)s,' -i bin/srun-%(namelower)s"])}

```

ocaisa: It's probably some evil escaping because even if I try to resolve it by hand with `resolve_template()` it still doesn't work

PR #3375, opened 2020-07-01
Also put easyblocks used by extensions in `reprod` directory

### LINKED ISSUE #3373: reprod dir doesn't get copy of easyblock used for extensions
The tensorflow.py is missing from the reprod dir of a recent install of TensorFlow

ocaisa:
Fix #3373

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3278,6 +3278,7 @@ def build_and_install_one(ecdict, init_env):
     start_time = time.time()
     try:
         run_test_cases = not build_option('skip_test_cases') and app.cfg['tests']
+
         if not dry_run:
             # create our reproducability files before carrying out the easyblock steps
             reprod_dir_root = os.path.dirname(app.logfile)
@@ -3435,7 +3436,6 @@ def copy_easyblocks_for_reprod(easyblock_instances, reprod_dir):
                 break
             else:
                 easyblock_paths.add(easyblock_path)
-
     for easyblock_path in easyblock_paths:
         easyblock_basedir, easyblock_filename = os.path.split(easyblock_path)
         copy_file(easyblock_path, os.path.join(reprod_easyblock_dir, easyblock_filename))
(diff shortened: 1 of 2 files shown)


ocaisa:
@boegel Any idea how I could test this?

ocaisa:
Ok, I have a test, trying to make it pass now...

akesandgren:
Going in, thanks @ocaisa!

easybuild/framework/easyblock.py
```diff
@@ -3416,6 +3424,24 @@ def build_and_install_one(ecdict, init_env):
...
 
+def copy_easyblocks_for_reprod(easyblock_instances, reprod_dir):
+    reprod_easyblock_dir = os.path.join(reprod_dir, 'easyblocks')
+    easyblock_paths = set()
+    for easyblock_instance in easyblock_instances:
+        for easyblock_class in inspect.getmro(type(easyblock_instance)):
+            easyblock_path = inspect.getsourcefile(easyblock_class)
+            # if we reach EasyBlock or ExtensionEasyBlock class, we are done
+            # (ExtensionEasyblock is hardcoded to avoid a cyclical import)
+            if easyblock_class.__name__ in [EasyBlock.__name__, 'ExtensionEasyBlock']:
```
boegel, line 3435: @ocaisa To be on the safe side, we also should consider `'Extension'` here... In theory there could be extensions that are installed with a class that directly derives from `Extension` rather than `ExtensionEasyBlock` (we don't have any like these currently in the central easyblocks repository though....)

easybuild/framework/easyblock.py
```diff
@@ -3437,18 +3463,8 @@ def reproduce_build(app, reprod_dir_root):
...
-        easyblock_path = inspect.getsourcefile(easyblock_class)
-        easyblock_basedir, easyblock_filename = os.path.split(easyblock_path)
-        # if we reach EasyBlock or ExtensionEasyBlock class, we are done
-        # (ExtensionEasyblock is hardcoded to avoid a cyclical import)
-        if easyblock_class.__name__ in [EasyBlock.__name__, 'ExtensionEasyBlock']:
-            break
-        else:
-            copy_file(easyblock_path, os.path.join(reprod_easyblock_dir, easyblock_filename))
-            _log.info("Dumped easyblock %s required for reproduction to %s", easyblock_filename, reprod_easyblock_dir)
+    # also archive all the relevant easyblocks (including any used by extensions)
```
boegel, line 3466: Comment is incorrect, archiving of easyblocks used by extensions is not done here?

easybuild/framework/easyblock.py
ocaisa: @boegel This comes at no cost right, I should do this

PR #3379, opened 2020-07-06
skip lines that start with 'module-version' when determining whether a module exists in ModulesTool.exist

### LINKED ISSUE #3376: Java-1.8.eb modulerc file fails to install with EasyBuild-4.2.1 due to changes in modules.py
**Issue**

With EasyBuild-4.2.0 `eb Java-1.8.eb -f` worked without problems, but with EasyBuild-4.2.1 I get:

```
== 2020-07-01 13:12:01,272 modules.py:596 DEBUG Module Java/1.8.0_192 not found via module avail/show, checking whether it is a wrapper
== 2020-07-01 13:12:01,273 modules.py:502 DEBUG Found /home/casparl/.local/easybuild/RedHatEnterpriseServer7/2019/modulefiles/all/Java/.modulerc that may define Java/1.8.0_192 as a wrapper for a module file
== 2020-07-01 13:12:01,273 modules.py:502 DEBUG Found /sw/arch/RedHatEnterpriseServer7/EB_production/2019/modulefiles/lang/Java/.modulerc that may define Java/1.8.0_192 as a wrapper for a module file
== 2020-07-01 13:12:01,274 modules.py:603 DEBUG Result for existence check of Java/1.8.0_192 module: False
== 2020-07-01 13:12:01,382 build_log.py:169 ERROR EasyBuild crashed with an error (at easybuild/base/exceptions.py:124 in __init__): Can't get value from a non-existing module Java/1.8.0_192 (at easybuild/tools/modules.py:691 in get_value_from_modulefile)
```

I have pinpointed this to the `modules.py`, because if I copy the `modules.py` from 4.2.0 into my 4.2.1 installation, my EasyBuild 4.2.1 can suddenly complete `eb Java-1.8.eb -f` succesfully.

Quite a number of things changed in the `modules.py` with the last update, but most of it is logging. I suspect the change in `mod_exists_via_show(mod_name)` to be the breaking change. I will proceed to modify the modules.py to only roll back that change.

boegel:
fixes #3376

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -575,12 +575,13 @@ class ModulesTool(object):
 
                 # skip lines that start with 'module-' (like 'module-version'),
                 # see https://github.com/easybuilders/easybuild-framework/issues/3376
-                if line.startswith('module- '):
+                if line.startswith('module-'):
                     self.log.debug("Skipping line '%s' since it starts with 'module-version'", line)
                     continue
 
                 # if line matches pattern that indicates an existing module file, the module file exists
-                self.log.debug("Determining whether module %s exists using line '%s'", mod_name, line)
+                self.log.debug("Determining whether module %s exists using 'module show' output line '%s'",
+                               mod_name, line)
                 res = bool(re.match(mod_exists_regex, line))
                 self.log.debug("Result for existence check of %s based on 'module show' output: %s", mod_name, res)
                 break
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -547,7 +547,7 @@ class ModulesTool(object):
 
             :param mod_name: module name
             """
-            self.log.info("Checking whether %s exists based on output of 'module show", mod_name)
+            self.log.info("Checking whether %s exists based on output of 'module show'", mod_name)
             stderr = self.show(mod_name)
             res = False
             # Parse the output:
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -598,7 +598,8 @@ class ModulesTool(object):
                 if mod_exists:
                     self.log.info("Module %s exists (found in list of available modules)", mod_name)
                 elif not mod_exists and maybe_partial:
-                    self.log.info("Module %s not found in list of available modules, checking via 'module show'...", mod_name)
+                    self.log.info("Module %s not found in list of available modules, checking via 'module show'...",
+                                  mod_name)
                     mod_exists = mod_exists_via_show(mod_name)
             else:
                 # hidden modules are not visible in 'avail', need to use 'show' instead
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -547,6 +547,7 @@ class ModulesTool(object):
 
             :param mod_name: module name
             """
+            self.log.info("Checking whether %s exists based on output of 'module show", mod_name)
             stderr = self.show(mod_name)
             res = False
             # Parse the output:
@@ -560,18 +561,21 @@ class ModulesTool(object):
                 if OUTPUT_MATCHES['whitespace'].search(line):
                     self.log.debug("Treating line '%s' as whitespace, so skipping it", line)
                     continue
-                # skip lines that start with 'module-version',
+
+                # if any errors occured, conclude that module doesn't exist
+                if OUTPUT_MATCHES['error'].search(line):
+                    break
+
+                # skip lines that start with 'module-' (like 'module-version'),
                 # see https://github.com/easybuilders/easybuild-framework/issues/3376
-                elif line.startswith('module-version '):
+                if line.startswith('module- '):
                     self.log.debug("Skipping line '%s' since it starts with 'module-version'", line)
                     continue
-                # if any errors occured, conclude that module doesn't exist
-                elif OUTPUT_MATCHES['error'].search(line):
-                    break
+
                 # if line matches pattern that indicates an existing module file, the module file exists
-                elif re.match(mod_exists_regex, line):
-                    res = True
+                res = bool(re.match(mod_exists_regex, line))
                 break
+
             return res
 
         if skip_avail:
@@ -587,10 +591,14 @@ class ModulesTool(object):
 
         mods_exist = []
         for (mod_name, visible) in mod_names:
+            self.log.info("Checking whether %s exists...", mod_name)
             if visible:
                 mod_exists = mod_name in avail_mod_names
                 # module name may be partial, so also check via 'module show' as fallback
-                if not mod_exists and maybe_partial:
+                if mod_exists:
+                    self.log.info("Module %s exists (found in list of available modules)", mod_name)
+                elif not mod_exists and maybe_partial:
+                    self.log.info("Module %s not found in list of available modules, checking via 'module show'...", mod_name)
                     mod_exists = mod_exists_via_show(mod_name)
             else:
                 # hidden modules are not visible in 'avail', need to use 'show' instead
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -547,7 +547,7 @@ class ModulesTool(object):
 
             :param mod_name: module name
             """
-            self.log.info("Checking whether %s exists based on output of 'module show'", mod_name)
+            self.log.debug("Checking whether %s exists based on output of 'module show'", mod_name)
             stderr = self.show(mod_name)
             res = False
             # Parse the output:
@@ -609,13 +609,13 @@ class ModulesTool(object):
                 # module name may be partial, so also check via 'module show' as fallback
                 if mod_exists:
                     self.log.info("Module %s exists (found in list of available modules)", mod_name)
-                elif not mod_exists and maybe_partial:
+                elif maybe_partial:
                     self.log.info("Module %s not found in list of available modules, checking via 'module show'...",
                                   mod_name)
                     mod_exists = mod_exists_via_show(mod_name)
             else:
                 # hidden modules are not visible in 'avail', need to use 'show' instead
-                self.log.debug("checking whether hidden module %s exists via 'show'..." % mod_name)
+                self.log.info("Checking whether hidden module %s exists via 'show'..." % mod_name)
                 mod_exists = mod_exists_via_show(mod_name)
 
             # if no module file was found, check whether specified module name can be a 'wrapper' module...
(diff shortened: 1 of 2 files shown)


boegel:
@casparvl Can you verify this fix?

@Flamefire Thoughts?

easybuild/tools/modules.py
```diff
@@ -555,11 +555,21 @@ def mod_exists_via_show(mod_name):
...
             for line in stderr.split('\n'):
+                # skip whitespace lines
+                self.log.debug("Checking line '%s' to determine whether %s exists...", line)
                 if OUTPUT_MATCHES['whitespace'].search(line):
+                    self.log.debug("Treating line '%s' as whitespace, so skipping it", line)
                     continue
-                if OUTPUT_MATCHES['error'].search(line):
+                # skip lines that start with 'module-version',
+                # see https://github.com/easybuilders/easybuild-framework/issues/3376
+                elif line.startswith('module-version '):
```
Flamefire, line 565: Doesn't ModuleTCL also have things like `module-alias` or so? They should be handled the same, so I'd propose `line.startswith('module-'):` here.

I'd also move this check after the error check so we don't accidentally pick up an error. I think the whitespace and error regexps are a good safe guard before any "manual evalulation".

I also have mixed feelings about the `elif`. It doesn't mix well with the `continue` and `break`. Maybe remove the `continue` and make all a "real` `if-elif-else` chain? I guess the last one can be: `else: res = re.match(mod_exists_regex, line); break`

boegel: I prefer keeping the `continue` and `break`, it helps making to logic more clear, so I ditched the `elif` again, see f237afe

easybuild/tools/modules.py
```diff
@@ -577,10 +591,15 @@ def mod_exists_via_show(mod_name):
...
         for (mod_name, visible) in mod_names:
+            self.log.info("Checking whether %s exists...", mod_name)
             if visible:
                 mod_exists = mod_name in avail_mod_names
                 # module name may be partial, so also check via 'module show' as fallback
-                if not mod_exists and maybe_partial:
+                if mod_exists:
+                    self.log.info("Module %s exists (found in list of available modules)", mod_name)
+                elif not mod_exists and maybe_partial:
+                    self.log.info("Module %s not found in list of available modules, checking via 'module show'...",
```
Flamefire, line 601: As you are using the (verbose) info method I find this (or the other) superflous as you will print `Checking whether %s exists based on output of 'module show` next (BTW: There is a `'` missing)

boegel: @Flamefire The other log message is in another (sub)method, which could also be called from other places, so it's not superfluous imho. Missing single quote fixed in e483fe4

easybuild/tools/modules.py
```diff
@@ -557,23 +557,32 @@ def mod_exists_via_show(mod_name):
...
                     continue
 
                 # if any errors occured, conclude that module doesn't exist
                 if OUTPUT_MATCHES['error'].search(line):
+                    self.log.debug("Line '%s' looks like an error, so concluding that %s doesn't exist", mod_name)
                     break
 
+                # skip warning lines, which may be produced by modules tool but should not be used
+                # to determine whether a module file exists
+                if line.startswith('WARNING: '):
```
Flamefire, line 572: Do you have a use case and sample output for this? Seems to haven't come up yet.

boegel: @casparvl Shared `module show Java/1.8.0_192` stderr output with me that starts with:

```
module-version    Java/1.8.0_192 1.8
WARNING: Duplicate version symbol '1.8' found
module-version  Java/1.8.0_192 1.8
/sw/arch/Debian10/EB_production/2019/modulefiles/lang/Java/1.8.0_192:
```

easybuild/tools/modules.py
```diff
@@ -557,23 +557,32 @@ def mod_exists_via_show(mod_name):
...
                 # skip lines that start with 'module-' (like 'module-version'),
                 # see https://github.com/easybuilders/easybuild-framework/issues/3376
                 if line.startswith('module- '):
                     self.log.debug("Skipping line '%s' since it starts with 'module-version'", line)
                     continue
 
                 # if line matches pattern that indicates an existing module file, the module file exists
+                self.log.debug("Determining whether module %s exists using line '%s'", mod_name, line)
                 res = bool(re.match(mod_exists_regex, line))
+                self.log.debug("Result for existence check of %s based on 'module show' output: %s", mod_name, res)
```
Flamefire, line 585: Maybe collapse the 2 outputs into one? I know it's just 1 more line but it adds up and the log is already so large that you can't read it anymore but need to grep/search it.

easybuild/tools/modules.py
```diff
@@ -555,13 +556,35 @@ def mod_exists_via_show(mod_name):
             # - Check first non-whitespace line for something that looks like an absolute path terminated by a colon
             mod_exists_regex = r'\s*/.+:\s*'
             for line in stderr.split('\n'):
+                # skip whitespace lines
+                self.log.debug("Checking line '%s' to determine whether %s exists...", line, mod_name)
```
Flamefire, line 560: As you now already output on every decision IMO this line can go. If you really want to keep it move it above the comment please

easybuild/tools/modules.py
```diff
@@ -555,13 +556,35 @@ def mod_exists_via_show(mod_name):
...
                 if OUTPUT_MATCHES['whitespace'].search(line):
+                    self.log.debug("Treating line '%s' as whitespace, so skipping it", line)
                     continue
+
+                # if any errors occured, conclude that module doesn't exist
                 if OUTPUT_MATCHES['error'].search(line):
+                    self.log.debug("Line '%s' looks like an error, so concluding that %s doesn't exist", mod_name)
```
Flamefire, line 567: Missing format argument

easybuild/tools/modules.py
Flamefire: Hm... What happens if they are different versions? Shall we really skip those?

easybuild/tools/modules.py
```diff
@@ -555,13 +556,35 @@ def mod_exists_via_show(mod_name):
...
+                # skip warning lines, which may be produced by modules tool but should not be used
+                # to determine whether a module file exists
+                if line.startswith('WARNING: '):
+                    self.log.debug("Skipping warning line '%s'", line)
+                    continue
+
+                # skip lines that start with 'module-' (like 'module-version'),
+                # see https://github.com/easybuilders/easybuild-framework/issues/3376
+                if line.startswith('module- '):
+                    self.log.debug("Skipping line '%s' since it starts with 'module-version'", line)
```
Flamefire, line 579: Please update the message and maybe remove the first comment line and move the 2nd (same/similar to log message which is enough)

boegel: fixed in 03fbb3a, but keeping the comments (since that refers to `module-version`)

easybuild/tools/modules.py
boegel: fixed missing format argument (but kept the comment, still helps with explaining the logic)

easybuild/tools/modules.py
boegel: I prefer keeping it, so I moved it

easybuild/tools/modules.py
boegel: Log messages collapsed in 03fbb3a

easybuild/tools/modules.py
boegel: Also mentioned this in #3376, and I'll add a test case for this...

casparvl:
I can confirm that the current version of `modules.py` in this PR fixes my original issue (as proven by the attached log).

[easybuild-Java-1.8-20200706.124325.log](https://github.com/easybuilders/easybuild-framework/files/4878377/easybuild-Java-1.8-20200706.124325.log)

easybuild/tools/modules.py
```diff
@@ -577,10 +603,15 @@ def mod_exists_via_show(mod_name):
...
         mods_exist = []
         for (mod_name, visible) in mod_names:
+            self.log.info("Checking whether %s exists...", mod_name)
             if visible:
                 mod_exists = mod_name in avail_mod_names
                 # module name may be partial, so also check via 'module show' as fallback
-                if not mod_exists and maybe_partial:
+                if mod_exists:
+                    self.log.info("Module %s exists (found in list of available modules)", mod_name)
+                elif not mod_exists and maybe_partial:
```
Flamefire, line 612: Minor: the `not mod_exists` is superflous as it is in the `else` branch.

easybuild/tools/modules.py
```diff
@@ -555,13 +556,38 @@ def mod_exists_via_show(mod_name):
             # - Check first non-whitespace line for something that looks like an absolute path terminated by a colon
             mod_exists_regex = r'\s*/.+:\s*'
             for line in stderr.split('\n'):
+
+                self.log.debug("Checking line '%s' to determine whether %s exists...", line, mod_name)
+
+                # skip whitespace lines
                 if OUTPUT_MATCHES['whitespace'].search(line):
+                    self.log.debug("Treating line '%s' as whitespace, so skipping it", line)
```
wpoely86, line 564: isn't this overly verbose, even for debugging?

easybuild/tools/modules.py
```diff
@@ -593,14 +624,14 @@ def mod_exists_via_show(mod_name):
...
-                self.log.debug("Module %s not found via module avail/show, checking whether it is a wrapper", mod_name)
+                self.log.info("Module %s not found via module avail/show, checking whether it is a wrapper", mod_name)
                 wrapped_mod = self.module_wrapper_exists(mod_name)
                 if wrapped_mod is not None:
                     # module wrapper only really exists if the wrapped module file is also available
                     mod_exists = wrapped_mod in avail_mod_names or mod_exists_via_show(wrapped_mod)
                     self.log.debug("Result for existence check of wrapped module %s: %s", wrapped_mod, mod_exists)
 
-            self.log.debug("Result for existence check of %s module: %s", mod_name, mod_exists)
+            self.log.info("Result for existence check of %s module: %s", mod_name, mod_exists)
```
wpoely86, line 634: why bump all these debugs to info?

easybuild/tools/modules.py
boegel: It helped a lot with figuring out the problem reported in #3376, so no.

easybuild/tools/modules.py
boegel: Because the non-debug log wasn't verbose enough to make it clear what was going on.

easybuild/tools/modules.py
casparvl: I know the PR has already been merged, but just to log this: I don't think it is an issue if the versions are different.

I hit this issue when trying to install the wrapper (i.e. `Java-1.8`). The EasyConfig specifies the _specific_ version (here: `1.8.0_192`) that you want to wrap. The log show that in the sanity check, the `module show` command is called on that _specific_ version (`Checking whether Java/1.8.0_192 exists based on output of 'module show'`). That will just return the correct output (i.e. it will either find a module, or not), regardless of what is set in the `.modulerc`. (since the module show is called on an actual module, and not the wrapper, I'm not even entirely sure why a `module show 1.8.0_192` even involves reading the two `.modulerc` files, but who cares)

Note that if any EasyBuild command would ever do a `module show Java/1.8` to check if thát module exists, I agree that it would be a concern to check which underlying specific module is actually checked for. I tried `module show Java/1.8`, and it just picks up on the `.modulerc` that comes first in the `MODULEPATH`. Of course, it _could_ happen that the `.modulerc` that comes first in the `MODULEPATH` is not the the one that is in EasyBuilds `installpath-modules`. I guess you could argue if that is the desired behaviour, or if you want EasyBuild to check based on the version defined in the `.modulerc` that IS in the `installpath-modules`. However, that nuance true in general for any check that would rely on `module show` - not just for modules that are wrapped - and I therefore don't think we should try to resolve that here.

PR #3382, opened 2020-07-07
add templates for CUDA compute capabilities

Flamefire:
The idea is to allow customization of the CCs without requiring to write an EasyBlock just for that.

They can be specified in the EC which is used as a default and can be overwritten by the build_opts (commandline or config)
It defines:
- `cuda_compute_capabilities`, a comma-separated list of CC: 3.1,7.2
- `cuda_sm_list`, a comma-separated list of sm-values: sm_31,sm_72
- `cuda_sm_values`, a space-separated list of sm-values: sm_31 sm_72

We likely need more than that. E.g. PyTorch wants space or semicolon separated CCs. Not sure how to do that. Maybe add a suffix: `_comma`, `_space`, `_semicolon` or so?


easybuild/framework/easyconfig/templates.py
```diff
@@ -299,6 +300,14 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
             # don't fail just because we couldn't resolve this template
             _log.warning("Failed to create mpi_cmd_prefix template, error was:\n%s", err)
 
+    # step 6. CUDA compute capabilities
+    cuda_compute_capabilities = build_option('cuda_compute_capabilities') or config.get('cuda_compute_capabilities')
+    if cuda_compute_capabilities:
+        template_values['cuda_compute_capabilities'] = ','.join(cuda_compute_capabilities)
+        sm_values = ['sm_' + cc.replace('.', '') for cc in cuda_compute_capabilities]
+        template_values['cuda_sm_list'] = ','.join(sm_values)
+        template_values['cuda_sm_values'] = ' '.join(sm_values)
```
boegel, line 309: @Flamefire How about renaming these to `cuda_sm_comma_sep` and `cuda_sm_space_sep`?

easybuild/framework/easyconfig/templates.py
```diff
@@ -299,6 +300,14 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
             # don't fail just because we couldn't resolve this template
             _log.warning("Failed to create mpi_cmd_prefix template, error was:\n%s", err)
 
+    # step 6. CUDA compute capabilities
+    cuda_compute_capabilities = build_option('cuda_compute_capabilities') or config.get('cuda_compute_capabilities')
```
boegel, line 304: Maybe add a comment mentioning that the `--cuda-compute-capabilities` values overrides value of the `cuda_compute_capabilities` parameter in the easyconfig file?

easybuild/framework/easyconfig/templates.py
```diff
@@ -299,6 +300,14 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
...
+    if cuda_compute_capabilities:
+        template_values['cuda_compute_capabilities'] = ','.join(cuda_compute_capabilities)
```
boegel, line 306: Should we rename this to `cuda_cc`?

Flamefire: I'm hesitant. Using the same name as the EC param is more intuitive as we do that with other like `%(prebuildopts)s` too.

easybuild/framework/easyconfig/templates.py
Flamefire: Good idea. Done

easybuild/framework/easyconfig/templates.py
Flamefire: Does this suffice: https://github.com/easybuilders/easybuild-framework/pull/3382/commits/e0a64144bdab3c7b5230dcdfc07728a4360ede58

Flamefire:
> urllib.error.URLError: <urlopen error [Errno 99] Cannot assign requested address>

WTF? From `pkg_simple = std_urllib.urlopen('https://pypi.python.org/simple/%s' % pkg, timeout=10).read()`

akesandgren:
@boegel are all your change requests done?

easybuild/framework/easyconfig/templates.py
boegel: yep!

easybuild/framework/easyconfig/templates.py
boegel: Yeah, I agree, let's keep it as is, it's mostly cosmetic anyway and may lead to confusion.

boegel (approved):

PR #3386, opened 2020-07-14
also escape backslashes in quote_py_str()

lexming:
This PR enhances `quote_py_str()` to also escape all backslashes. Otherwise parse tests may fail due to divergences between the provided and dumped easyconfig. That's because `quote_py_str()` is limited to quotes and new lines, while several options such as `preconfigopts`, `prebuildopts` or `postintallcmds` can potentially contain any string literal.

This fixes the failed tests of https://github.com/easybuilders/easybuild-easyconfigs/pull/10965?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDk4ODg3MTQ0Mzo0OTczNzk0#issuecomment-658010559

### DIFF
--- a/easybuild/tools/utilities.py
+++ b/easybuild/tools/utilities.py
@@ -54,7 +54,7 @@ def flatten(lst):
     return res
 
 
-def quote_str(val, escape_newline=False, prefer_single_quotes=False, tcl=False):
+def quote_str(val, escape_newline=False, prefer_single_quotes=False, escape_backslash=False, tcl=False):
     """
     Obtain a new value to be used in string replacement context.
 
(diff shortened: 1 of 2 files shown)


lexming:
My previous approach with `repr()` did work with tests in `test.easyconfigs.suite` but it messed up some tests in `test.framework.easyconfig`. So I revised it and now I propose to fix `quote_py_str()`.

easybuild/tools/utilities.py
```diff
@@ -92,7 +92,13 @@ def quote_str(val, escape_newline=False, prefer_single_quotes=False, tcl=False):
 
 def quote_py_str(val):
     """Version of quote_str specific for generating use in Python context (e.g., easyconfig parameters)."""
-    return quote_str(val, escape_newline=True, prefer_single_quotes=True)
+    quoted_str = quote_str(val, escape_newline=True, prefer_single_quotes=True)
+
+    if isinstance(val, string_type):
+        # escape backslashes in strings
+        quoted_str = quoted_str.replace('\\', '\\\\')
```
boegel, line 99: @lexming Why not add support for using `escape_backslash=True` to `quote_str`?

lexming: sure, I can move it to `quote_str` :+1:

Issue #109, opened 2012-08-19
generated devel module needs work


boegel:
Generating the devel module is already started in `prepare`, so before `configure` and friends (why is it already created there?).

For easyblocks that redefine `make_module_req_guess`, this could be a problem, because in various occasions this function relies on `self.x` variables that were set during the build (e.g. for SLEPc).

So, the devel module may contain incorrect values because of that.

Also, the devel module also loads the devel module of all dependencies, which seems to be causing issues (e.g. try loading the current devel module for DOLFIN).

citibob:
Issue #203 (now closed) is a duplicate.

boegel:
@citibob: I think you intended to mention that issue #1531 is a duplicate of #203? Seems like you misplaced your last comment?

citibob:
Yes I think you are right.

PR #3388, opened 2020-07-20
add EB option --generate-devel-module

smoors:
this option is useful for example when creating a module in combination with `--module-only` and `-modules-footer` where the footer generates an `LmodError` upon loading the module.

using `LmodError` causes EB to fail because it tries to load the module unless make_devel_module() is skipped. skipping this function is fine in this case because the module is not supposed to be loaded anyway.

if this option is well-received I'll try to write a test for it as well.


smoors:
closing-reopening

boegel:
@smoors Can you clarify the use case for this? What's the point of generating a module file you can't load?

Also, this is for using together with `--force --module-only`, right? Since otherwise the generated module will be loaded during the sanity check, which will result in a failure if the module can't be loaded.

I would rename it too, since it's a boolean option. It should be something like `--generate-devel-module` that is enabled by default, and then you can use `--disable-generate-devel-module`.

That being said, I think we need to re-evaluate generating those devel modules, I doubt they're actually useful at all (see also #109).

smoors:
@boegel my use case is discoverability:
you could make a module visible to the user e.g. when they do 'module spider' or 'module av' on the login node, without allowing them to actually load/use it on the login node itself. when the user tries to load it they then get an error message saying 'this module can only be used on this type of node'. currently we work around this by creating a `LmodWarning` but imho this is confusing for the users because then they have loaded a module that does not do anything, while still having loaded its dependencies.

it should indeed be used together with `--force`.

renaming the is fine to me, I was not happy about the name in the first place.

about the usefulness of devel modules, I never used them myself, if they work at all. maybe we can ask the community to see if anyone is using them?

smoors:
closing-reopening

easybuild/tools/options.py
```diff
@@ -1063,6 +1065,10 @@ def _postprocess_config(self):
         if self.options.inject_checksums:
             self.options.pre_create_installdir = False
 
+        # imply --force with --disable-generate-devel-module
+        if not self.options.generate_devel_module:
+            self.options.force = True
```
boegel, line 1070: @smoors This seems weird to me, I don't think `--disable-generate-devel-module` should always imply `--force`?

boegel (changes requested):

easybuild/tools/options.py
smoors: yeah I was not sure about that, I'll remove it.

boegel (approved):

PR #3394, opened 2020-07-28
also consider $EB_INSTALLPYTHON to specify 'python' command to use for running EasyBuild

ocaisa:
...which has lower precedence than `EB_PYTHON`, this envvar can be leveraged by an EasyBuild module file rather than it setting `EB_PYTHON`

### DIFF
--- a/eb
+++ b/eb
@@ -43,7 +43,7 @@ function verbose() {
 }
 
 PYTHON=
-for python_cmd in ${EB_PYTHON} 'python' 'python3' 'python2'; do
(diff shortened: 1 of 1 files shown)


boegel:
@ocaisa I'm not sure this is the best way forward (see my comment in https://github.com/easybuilders/easybuild-easyblocks/issues/2078#issuecomment-669009467), but if we do this, we should add some comments to clarify the purpose of `$EB_PYTHON` and `$EB_INSTALLPYTHON` in the `eb` script, so people checking the contents of the script are no left wondering why two environment variables are being considered...

ocaisa:
Replacing this with #3428 since the CI is not failing there...

PR #3397, opened 2020-08-04
Use one argument 'module swap' statements in Tcl modulefiles (required by Modules 4.2.3+)

### LINKED ISSUE #3396: Can't generate working Tcl modulefiles with Modules 4.2.3+
Using EasyBuild with newer versions of EnvironmentModules fails as it generates Tcl modules that can't load. [Changes to Modules in 4.2.3](https://modules.readthedocs.io/en/latest/NEWS.html#modules-4-2-3-2019-03-23) mean that modules unloaded as part of a swap/switch operation are now flagged as conflicting with the module, so when an EasyBuild module does e.g. `module swap gcc gcc/9.3.0`, 'gcc' becomes conflicting with the main module, 'gcc/9.3.0' can't be loaded, and the main module fails to load.

The problem appeared for me while setting up a new CrayGNU toolchain on a system with Modules 4.3.1. The modulefile generated has several swap statements to ensure the correct module versions are loaded. Installing the toolchain fails at the sanity check stage when the module can't be loaded.

Modulefiles with the simple following form will fail to load for the same reason:

```
#%Module
module swap some_module some_module/x.y.z
```

[The Modules 4.2.3 release notes](https://modules.readthedocs.io/en/latest/NEWS.html#modules-4-2-3-2019-03-23) also state 'If `module switch` has only one argument, do not define a conflict toward switched-off modulefile.' If EasyBuild generates Tcl modulefiles with swap statements like this, the problem should be resolved. However, it also states 'it is not recommended to use `module switch` command in modulefiles' so there might be more to this.

welucas2:
This should address #3396 by generating modulefiles which use `module swap` commands with the only argument being the module to load in. `module` is intelligent enough to know which module should be unloaded, and this doesn't result in conflicting modules (and so unusable modulefiles) which occurs with Modules 4.2.3 and newer.

edit: fixes #3396

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -963,8 +963,8 @@ class ModuleGeneratorTcl(ModuleGenerator):
         :param guarded: guard 'swap' statement, fall back to 'load' if module being swapped out is not loaded
         """
         # In Modules 4.2.3+ a 2-argument swap 'module swap foo foo/X.Y.Z' will fail as the unloaded 'foo'
-        # means all 'foo' modules conflict and 'foo/X.Y.Z' will not load.  A 1-argument swap like 
-        # 'module swap foo/X.Y.Z' will unload any currently loaded 'foo' without it becoming conflicting 
+        # means all 'foo' modules conflict and 'foo/X.Y.Z' will not load.  A 1-argument swap like
+        # 'module swap foo/X.Y.Z' will unload any currently loaded 'foo' without it becoming conflicting
         # and successfully load the new module.
         # See: https://modules.readthedocs.io/en/latest/NEWS.html#modules-4-2-3-2019-03-23
         body = "module swap %s" % (mod_name_in)
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -962,7 +962,7 @@ class ModuleGeneratorTcl(ModuleGenerator):
         :param mod_name_in: name of module to load (swap in)
         :param guarded: guard 'swap' statement, fall back to 'load' if module being swapped out is not loaded
         """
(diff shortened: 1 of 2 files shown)


welucas2:
Looks like it fails with older Modules versions. Something else is needed?

easybuild/tools/module_generator.py
```diff
@@ -962,7 +962,7 @@ def swap_module(self, mod_name_out, mod_name_in, guarded=True):
         :param mod_name_in: name of module to load (swap in)
         :param guarded: guard 'swap' statement, fall back to 'load' if module being swapped out is not loaded
         """
-        body = "module swap %s %s" % (mod_name_out, mod_name_in)
+        body = "module swap %s" % (mod_name_in)
```
boegel, line 965: @welucas2 It's probably worth adding a comment above here to explain why we use single argument swap commands.

Before we didn't have a good reason to use `swap foo foo/<version>` rather than `swap foo/<version>` I think, but now we clearly do, so it's worth mentioning that I think...

I'll also check with the other module tools we support whether single argument swap works for those, I'm not sure that's covered by the test suite right now (we just do a text comparison of the generated module file in the tests currently).

boegel (approved):
I've updated this PR with a small update to `test_swap` that verifies for which module tools the single-argument swap works.

It's fine for everything except for the ancient Tcl-only modules tool version 1.x, which we don't really care about anymore (I'll deprecate the support for that in a separate PR).

I'll go ahead and merge this PR as soon as CI gives the green light, thanks for diving into this yourself @welucas2!

boegel:
I've also verified that single-arg swap in `Cray*` toolchain modules works fine on Piz Daint (cc @victorusu, @gppezzi), so I'll go ahead and merge this.

Thanks again @welucas2!

PR #3400, opened 2020-08-11
fix copying of (non-existing) file with apply_patch

boegel:
Copying a file through patches was failing if the target location is a non-existing file. Example:

```python
patches = [('foo.txt', 'src/bar.txt')]
```

resulting error:

```
Can't patch directory /tmp/example/1.2.3/intel-2020a-Python-3.8.2/example-1.2.3/src/bar.txt: no such directory
```

The fix is basically to postpone the directory check until after copying of files (when `copy` is `True`) is dealt with.


smoors:
Going in, thanks @boegel!

PR #3401, opened 2020-08-13
create symlink from lib64 to lib subdir in installation directories to avoid that GCC prefers /lib64 system directories

Flamefire:
This avoids using the system libs from /lib64 instead of libs from LIBRARY_PATH
Fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/5776 but needs a rebuild of the affected modules

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2352,10 +2352,11 @@ class EasyBlock(object):
         # However for each <dir> in $LIBRARY_PATH (where <dir> is often <prefix>/lib) it searches <dir>/../lib64 first.
         # So we create <prefix>/lib64 as a symlink to <prefix>/lib to make it prefer EB installed libraries.
         # See https://github.com/easybuilders/easybuild-easyconfigs/issues/5776
-        lib_dir = os.path.join(self.installdir, 'lib')
-        lib64_dir = os.path.join(self.installdir, 'lib64')
-        if os.path.exists(lib_dir) and not os.path.exists(lib64_dir):
-            symlink(lib_dir, lib64_dir)
+        if build_option('lib64_lib_symlink'):
+            lib_dir = os.path.join(self.installdir, 'lib')
+            lib64_dir = os.path.join(self.installdir, 'lib64')
+            if os.path.exists(lib_dir) and not os.path.exists(lib64_dir):
+                symlink(lib_dir, lib64_dir)
 
     def sanity_check_step(self, *args, **kwargs):
         """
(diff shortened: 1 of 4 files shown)


Flamefire:
Ok this seems to conflict with the test for `disable-lib64-fallback-sanity-check`. How to proceed?

easybuild/framework/easyblock.py
```diff
@@ -2348,6 +2348,13 @@ def post_install_step(self):
                 run_cmd(cmd, simple=True, log_ok=True, log_all=True)
 
         self.fix_shebang()
+        # GCC linker searches system /lib64 path before the LIBRARY_PATH paths
```
boegel, line 2351: `$LIBRARY_PATH` (to make it clear we're referring to the environment variable)

easybuild/framework/easyblock.py
```diff
@@ -2348,6 +2348,13 @@ def post_install_step(self):
...
+        # However foreach path x in LIBRARY_PATH it searches x/../lib64 so we create that as a symlink to lib
```
boegel, line 2352: `foreach x` -> `for each <dir>/lib`
`x/../lib64` -> `<dir>/lib/../lib64`
`symlink to lib` -> `symlink to <dir>/lib/`
and please use `$LIBRARY_PATH`

boegel:
> Ok this seems to conflict with the test for `disable-lib64-fallback-sanity-check`. How to proceed?

We could add a configuration option like `--symlink-lib-lib64` which is enabled by default, and then use `--disable-symlink-lib-lib64` in the test for `--disable-lib64-fallback-sanity-check`?

easybuild/framework/easyblock.py
Flamefire: Slightly revised in https://github.com/easybuilders/easybuild-framework/pull/3401/commits/db3b5e656766b63bcbddfdefe6b0f10ce8cc7ad4

easybuild/framework/easyblock.py
Flamefire: Done: https://github.com/easybuilders/easybuild-framework/pull/3401/commits/db3b5e656766b63bcbddfdefe6b0f10ce8cc7ad4

Flamefire:
Ok added `--lib64-lib-symlink` option in https://github.com/easybuilders/easybuild-framework/pull/3401/commits/6ccd097cff4a76b3239dbab24ed35af2f9e21832 to make it show up next to the other lib64 option

Flamefire:
Fixed the merge conflict.

Flamefire:
Well it was tested although hidden in another test: https://github.com/easybuilders/easybuild-framework/pull/3401/files#diff-ce4ad0ca611df220e49671774fa169a9R1982

But yours is better of course, thanks!

mboisson:
Could this be made to be a symlink to a relative path, rather than an absolute path ?

PR #3402, opened 2020-08-14
fix default value for 'lib64_fallback_sanity_check' build option

boegel:
`--lib64-fallback-sanity-check` is enabled by default, so the correct default value for the corresponding build option is `True`, not `False`

This has no impact in normal circumstances, since the default values for build options are not actually used during normal configuration (when using `eb` or after calling `set_up_configuration` when leveraging EasyBuild as a library).

Thanks to @Flamefire for raising this issue...


migueldiascosta:
Going in, thanks @boegel!

PR #3409, opened 2020-08-17
stop installing ccache wrapper for Fortran compiler

boegel:
There's little point in installing a `ccache` wrapper for the Fortran compiler, because `ccache` has dropped support for Fortran a while ago because of trouble with Fortran modules (see https://github.com/ccache/ccache/issues/342).


akesandgren:
Going in, thanks @boegel!

PR #3412, opened 2020-08-18
Add patch step dict notation

Darkless012:
Allow patches to be defined as dicts.
This allows passing additional options for patch command.

This was needed in past because patching file with CRLF endings requires use of `--binary` option. 
(it was MOABS, but recent versions doesn't have the issue)

More about this for example here:
https://stackoverflow.com/questions/2076688/how-to-use-patches-created-in-windows-with-crlf-in-linuxi
https://github.com/conda/conda-build/issues/1973

The general work is done, and I don't want it to be lost.
Tests are included as well, but I couldn't find a way to test some aspects.

This is why I'm writing this long description.
I will add links to the changed files with commented code to bring insight.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -38,7 +38,7 @@ The EasyBlock class should serve as a base class for all easyblocks.
 :author: Maxime Boissonneault (Compute Canada)
 :author: Davide Vanzo (Vanderbilt University)
 """
-
+import collections
 import copy
 import glob
 import inspect
@@ -408,6 +408,9 @@ class EasyBlock(object):
             copy_file = False
             suff = None
             level = None
+            patch_opts = None
+
+            # patch_spec is list or tuple
             if isinstance(patch_spec, (list, tuple)):
                 if not len(patch_spec) == 2:
                     raise EasyBuildError("Unknown patch specification '%s', only 2-element lists/tuples are supported!",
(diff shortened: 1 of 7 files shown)


test/framework/easyblock.py
```diff
@@ -1512,6 +1512,73 @@ def test_patch_step(self):
...
+        eb.extract_step()
+
+        # # patch step with captured output
+        # self.mock_stdout(True)
+        eb.patch_step()
+        # stdout = self.get_stdout()
+        # self.mock_stdout(False)
+        #
+        # # verify that patch cmd included opts ('--verbose')
+        # self.assertIn('opts: --verbose', stdout)
```
Darkless012, line 1544: How to check if additional options were included in the patch command?

boegel: @Darkless012 The easiest way is probably by injecting a fake `patch` command that just writes its options to a file in `self.test_prefix`?

```python
fake_patch = os.path.join(self.test_prefix, 'patch')
fake_patch_out = os.path.join(self.test_prefix, 'patch.out')
write_file(fake_patch, "#!/bin/bash echo "$@" > %s" % fake_patch_out)
adjust_permissions(fake_patch, stat.S_IRUSR | stat.S_IXUSR)
os.environ['PATH'] = self.test_prefix + ':' + os.environ['PATH']

Comment (Darkless012) [reply test/framework/easyblock.py:1544]: Thanks, will test it out :)

**Comment by akesandgren** [comment]
@Darkless012 this needs a conflict resolution.

File: easybuild/framework/easyblock.py [hunk 2]
```diff
@@ -426,6 +429,30 @@ def fetch_patches(self, patch_specs=None, extension=False, checksums=None):
                 else:
                     raise EasyBuildError("Wrong patch spec '%s', only int/string are supported as 2nd element",
                                          str(patch_spec))
+
+            # patch_spec is dict
+            elif isinstance(patch_spec, collections.Mapping):
```
Flamefire, line 434: Suggestion: As below a `patchspec` dict is created (mind the almost name clash), maybe don't do this checking here (I'd assume this is done by the parser already?) and simply coerce the tuple into a dict in the above `if` and also below. So basically: If it isn't already a dict, make it one, else assume it is valid. In any case add the remaining keys (name, path, checksum, ...)

akesandgren:
@Darkless012 This needs conflict resolution and a look at flamefires suggestion.

PR #3414, opened 2020-08-18
fix issues with applying PR patch in --from-pr

### LINKED ISSUE #3411: --from-pr does not (correctly) set the permissions
Use case: A pull request for an easyconfig contains an executable script which is to be run via `tests` which expects a list of commands or executables.

When trying this with `eb --from-pr` the executable permission from that file is lost making the test fail at the very end.

When checking the `*.patch` file from the PR I see `new file mode 100755`

Flamefire:
When PRs introduce only new files determining the patch level fails which makes it skip the whole `git am` part even though the level is not even required. Hence as a first part of the fix the level determination is skipped when using git to apply the patch.

Secondly `git am` requires an active git repo but we download the base branch via the archive, not a clone. Also the parameter `patch` is likely wrong as I can't find it in the docu. Hence replace that with `git apply` which does work for regular folders too.

Fixes #3411

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1186,7 +1186,7 @@ def guess_patch_level(patched_files, parent_dir):
     return patch_level
 
 
-def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=False):
+def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git=False):
     """
     Apply a patch to source code in directory dest
     - assume unified diff created with "diff -ru old new"
@@ -1237,8 +1237,8 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
             change_dir(apatch_dir)
             apatch = os.path.join(apatch_dir, apatch_name)
 
-    if use_git_am:
-        patch_cmd = "git am patch %s" % apatch
+    if use_git:
+        patch_cmd = "git apply %s" % apatch
     else:
         if level is None and build_option('extended_dry_run'):
             level = '<derived>'
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1228,19 +1228,19 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
     abs_dest = os.path.abspath(dest)
 
     # Attempt extracting the patch if it ends in .patch.gz, .patch.bz2, .patch.xz
-    # split in name + extension
-    patch_filestem, patch_extension = os.path.splitext(os.path.split(abs_patch_file)[1])
+    # split in stem (filename w/o extension) + extension
+    patch_stem, patch_extension = os.path.splitext(os.path.split(abs_patch_file)[1])
     # Supports only bz2, gz and xz. zip can be archives which are not supported.
     if patch_extension in ['.gz', '.bz2', '.xz']:
         # split again to get the second extension
-        patch_subextension = os.path.splitext(patch_filestem)[1]
+        patch_subextension = os.path.splitext(patch_stem)[1]
         if patch_subextension == ".patch":
             workdir = tempfile.mkdtemp(prefix='eb-patch-')
             _log.debug("Extracting the patch to: %s", workdir)
             # extracting the patch
             extracted_dir = extract_file(abs_patch_file, workdir, change_into_dir=False)
             change_dir(extracted_dir)
-            abs_patch_file = os.path.join(extracted_dir, patch_filestem)
+            abs_patch_file = os.path.join(extracted_dir, patch_stem)
 
     if use_git:
         verbose = '--verbose ' if build_option('debug') else ''
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1239,7 +1239,6 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
             _log.debug("Extracting the patch to: %s", workdir)
             # extracting the patch
             extracted_dir = extract_file(abs_patch_file, workdir, change_into_dir=False)
-            change_dir(extracted_dir)
             abs_patch_file = os.path.join(extracted_dir, patch_stem)
 
     if use_git:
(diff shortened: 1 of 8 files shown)


easybuild/tools/filetools.py
```diff
@@ -1186,7 +1186,7 @@ def guess_patch_level(patched_files, parent_dir):
     return patch_level
 
 
-def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=False):
+def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git=False):
```
boegel, line 1189: @Flamefire We shouldn't drop `use_git_am` like this, but deprecate it to avoid breaking backwards compatibility:

```python
if use_git_am:
    _log.deprecated('5.0', "Setting 'use_git_am' named argument for apply_patch is deprecated, please set 'use_git=True' instead")
    use_git = True
```

easybuild/tools/filetools.py
```diff
@@ -1237,43 +1237,41 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
...
-        patch_cmd = "git am patch %s" % apatch
-    else:
         patch_cmd = "patch -b -p%s -i %s" % (level, apatch)
 
     out, ec = run.run_cmd(patch_cmd, simple=False, path=adest, log_ok=False, trace=False)
 
     if ec:
         raise EasyBuildError("Couldn't apply patch file %s. Process exited with code %s: %s", patch_file, ec, out)
 
-    return ec == 0
```
boegel, line 1275: @Flamefire Did you drop this by accident? It's important to have it here, since otherwise the return value is always `None`.

Flamefire: I considered it useless as it is either True or an exception but forgot about backwards compat. Changed to True and documented that

easybuild/tools/filetools.py
```diff
@@ -1218,61 +1224,60 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
...
-    apatch = os.path.abspath(patch_file)
-    adest = os.path.abspath(dest)
+    abs_patch_file = os.path.abspath(patch_file)
+    abs_dest = os.path.abspath(dest)
 
     # Attempt extracting the patch if it ends in .patch.gz, .patch.bz2, .patch.xz
     # split in name + extension
-    apatch_root, apatch_file = os.path.split(apatch)
-    apatch_name, apatch_extension = os.path.splitext(apatch_file)
+    patch_filestem, patch_extension = os.path.splitext(os.path.split(abs_patch_file)[1])
```
boegel, line 1232: I would go with `patch_filename` (or `patch_name`), `filestem` is so close to `filesystem` that it's *very* confusing

easybuild/tools/filetools.py
```diff
@@ -1186,12 +1186,18 @@ def guess_patch_level(patched_files, parent_dir):
...
+def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=False, use_git=False):
     """
     Apply a patch to source code in directory dest
     - assume unified diff created with "diff -ru old new"
+
+    Raises EasyBuildError on any error and returns True on success
     """
 
+    if use_git_am:
+        _log.deprecated('5.0', "'use_git_am' named argument has been renamed to 'use_git'")
```
boegel, line 1198:
```suggestion
        _log.deprecated('5.0', "'use_git_am' named argument in apply_patch function has been renamed to 'use_git'")

File: easybuild/tools/filetools.py (hunk 3 above)
Comment (Flamefire) [reply easybuild/tools/filetools.py:1232]: then I'd rather use `patch_stem`. "stem" is commonly used as the filename without the extension which is why I wanted to highlight it here. Or maybe `patch_name` but I'd like to avoid `filename` because that usually includes the extension

File: easybuild/tools/filetools.py (hunk 1 above)
Comment (Flamefire) [reply easybuild/tools/filetools.py]: https://github.com/easybuilders/easybuild-framework/pull/3414/commits/e5cd2ce2c384285b98b5d1a3bc59f3f91b6f3faf

File: easybuild/tools/filetools.py (hunk 3 above)
Comment (boegel) [reply easybuild/tools/filetools.py:1232]: Then I'd converge on `patch_name`, `patch_stem` is fine too.

Comment (Flamefire) [reply easybuild/tools/filetools.py]: Ok, will go with `patch_stem` and a comment above it.

Note: I also removed the `change_dir` below: https://github.com/easybuilders/easybuild-framework/pull/3414/commits/9956c30917e0e20b6357b4c3cd905da4bdb1e6c7

File: easybuild/tools/filetools.py [hunk 5]
```diff
@@ -1186,12 +1186,18 @@ def guess_patch_level(patched_files, parent_dir):
...
+def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=False, use_git=False):
     """
     Apply a patch to source code in directory dest
     - assume unified diff created with "diff -ru old new"
+
+    Raises EasyBuildError on any error and returns True on success
     """
 
+    if use_git_am:
+        _log.deprecated('5.0', "'use_git_am' named argument in apply_patch function has been renamed to 'use_git'")
```
boegel, line 1198: @Flamefire Arguments should be the other way around here:

```suggestion
        _log.deprecated("'use_git_am' named argument in apply_patch function has been renamed to 'use_git'", '5.0')

File: test/framework/options.py [hunk 6]
```diff
@@ -5056,7 +5056,7 @@ def test_inject_checksums(self):
             "^== backup of easyconfig file saved to .*/test\.eb\.bak_[0-9]+_[0-9]+\.\.\.$",
             "^== injecting md5 checksums for sources & patches in test\.eb\.\.\.$",
             "^== \* toy-0.0\.tar\.gz: be662daa971a640e40be5c804d9d7d10$",
-            "^== \* toy-0\.0_fix-silly-typo-in-printf-statement\.patch: e6785e1a721fc8bf79892e3ef41557c0$",
+            "^== \* toy-0\.0_fix-silly-typo-in-printf-statement\.patch: a99f2a72cee1689a2f7e3ace0356efb1$",
```
Flamefire, line 5059:
```suggestion
            r"^== \* toy-0\.0_fix-silly-typo-in-printf-statement\.patch: a99f2a72cee1689a2f7e3ace0356efb1$",
```

boegel (approved):

PR #3416, opened 2020-08-19
make flake8 code style checks pass in test/

boegel:
(cfr. https://github.com/easybuilders/easybuild/issues/643)

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -19,4 +19,4 @@ jobs:
     - name: Run flake8
       run: |
         flake8 easybuild/tools
- 
+        flake8 test/
--- a/test/framework/github.py
+++ b/test/framework/github.py
@@ -89,7 +89,7 @@ class GithubTest(EnhancedTestCase):
         try:
             expected = [(None, ['a_directory', 'second_dir'], ['README.md']),
                         ('a_directory', ['a_subdirectory'], ['a_file.txt']), ('a_directory/a_subdirectory', [],
-                        ['a_file.txt']), ('second_dir', [], ['a_file.txt'])]
+                                                                              ['a_file.txt']), ('second_dir', [], ['a_file.txt'])]
             self.assertEqual([x for x in self.ghfs.walk(None)], expected)
         except IOError:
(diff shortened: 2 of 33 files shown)


wpoely86 (approved):

.github/workflows/linting.yml
```diff
@@ -19,4 +19,4 @@ jobs:
     - name: Run flake8
       run: |
         flake8 easybuild/tools
- 
\ No newline at end of file
+        flake8 test/
```
Flamefire, line 22: Please do not change this during the sprint. My suggestion is to enable this at the very end for `.`. Otherwise all other PRs will conflict when changing this too.
Oh and according to GH it is missing the final newline and the trailing slash is not required :)

Flamefire: Sure if you want to make those dependent PRs but IMO it is easier to verify locally via flake8 and don't change this yet.

test/framework/github.py
```diff
@@ -89,7 +89,7 @@ def test_walk(self):
         try:
             expected = [(None, ['a_directory', 'second_dir'], ['README.md']),
                         ('a_directory', ['a_subdirectory'], ['a_file.txt']), ('a_directory/a_subdirectory', [],
-                        ['a_file.txt']), ('second_dir', [], ['a_file.txt'])]
+                                                                              ['a_file.txt']), ('second_dir', [], ['a_file.txt'])]
```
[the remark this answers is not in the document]
boegel: Nice job there `autopep8`...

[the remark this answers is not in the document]
Flamefire: Intresting... I see why though. The tuple starts in the previous line and autopep8 doesn't try to hard... Maybe with `-aaa` or so to make it more agressive but I'd adivse against that

.github/workflows/linting.yml
boegel: For `easyblocks` repo, I definitely agree we should hold off touching (actually adding) `linting.yml` until we're done.

wpoely86 (approved):

PR #3418, opened 2020-08-20
limit Travis to only test with Python 2.6 + Lomod 7

wpoely86:
All other combination of python/Tmod/Lmod are covered by github
workflows.


boegel (approved):
Agreed, Travis is flake enough that this makes sense, especially now `@boegelbot` is keeping an eye on GitHub Actions rather than Travis...

PR #3419, opened 2020-08-24
add --sysroot configuration option

boegel:
This will come in useful when building certain software (GCC, binutils, ...) in an environment where you don't want to pick up *anything* from the host environment, for example when cross-compiling or when installing software on top of an alternative environment like Gentoo Prefix.

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -212,6 +212,7 @@ BUILD_OPTIONS_CMDLINE = {
         'skip',
         'stop',
         'subdir_user_modules',
+        'sysroot',
         'test_report_env_filter',
         'testoutput',
         'wait_on_lock',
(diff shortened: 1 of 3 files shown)


casparvl:
Looks pretty straightforward to me. Just to clarify your ideas here: you don't envision the `--sysroot` option will _do_ anything on the framework side I guess? I mean, I guess we mostly want to _define_ this option in framework so that we can query its setting in the relevant EasyBlocks and customize behaviour accordingly there?

boegel:
@casparvl Exactly, see https://github.com/easybuilders/easybuild-easyblocks/pull/2142 for example.

I'm not planning anything specific on the framework side for now, although it's possible we could enhance the sanity check when `--sysroot` is set, to make sure nothing outside the specified sysroot is picked up (by checking `ldd` output, etc.).

Whether that's a viable approach remains to be seen, but it could be interesting...

casparvl:
> I'm not planning anything specific on the framework side for now, although it's possible we could enhance the sanity check when --sysroot is set, to make sure nothing outside the specified sysroot is picked up (by checking ldd output, etc.).

Interesting idea. I guess we could try and see with a couple of `ldd`'s if 'all' cases where it picks up something from outside the sysroot are 'wrong'. I mean, the things outside the sysroot it should pick up on are drivers and such, but those shouldn't typically show up with an `ldd` I guess... It IS very hard to otherwise find out that you're linking against stuff outside of your sysroot...

casparvl (approved):
LGTM

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3422, opened 2020-08-26
make RPATH wrapper script more robust by using 'python -E -s -S' to run rpath_args.py

### LINKED ISSUE #3421: broken installation of 'cffi' extension in Python 2 easyconfig when running EasyBuild with Python 3 and using --rpath
I ran into this weird problem when installing the `cffi` extension in `Python-2.7.18-GCCcore-9.3.0.eb`:

```shell
    building '_cffi_backend' extension
  creating build/temp.linux-x86_64-2.7
  creating build/temp.linux-x86_64-2.7/c
  gcc -fno-strict-aliasing -O2 -ftree-vectorize -march=native -fno-math-errno -fPIC -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -O2 -ftree-vectorize -march=native -fno-math-errno -fPIC -fPIC -I/software/Python/2.7.18-GCCcore-9.3.0/include/python2.7 -c c/_cffi_backend.c -o build/temp.linux-x86_64-2.7/c/_cffi_backend.o
  Fatal Python error: initsite: Failed to import the site module
    File "/software/Python/2.7.18-GCCcore-9.3.0/lib/python2.7/site-packages/site.py", line 73, in <module>
      __boot()
    File "/software/Python/2.7.18-GCCcore-9.3.0/lib/python2.7/site-packages/site.py", line 26, in __boot
      import imp  # Avoid import loop in Python 3
    File "/usr/lib/python-exec/python3.7/../../../lib/python3.7/imp.py", line 27, in <module>
      import tokenize
    File "/usr/lib/python-exec/python3.7/../../../lib/python3.7/tokenize.py", line 33, in <module>
      import re
    File "/usr/lib/python-exec/python3.7/../../../lib/python3.7/re.py", line 145, in <module>
      class RegexFlag(enum.IntFlag):
  AttributeError: module 'enum' has no attribute 'IntFlag'
  error: command 'gcc' failed with exit status 1
```

(relevant) details about the environment:

* EasyBuild is being run with `python3` (by setting `$EB_PYTHON` to `python3`)
* EasyBuild is configured to use RPATH (via `eb --rpath`)
* both `setuptools` and `enum34` are also installed as extensions in `Python-2.7.18-GCCcore-9.3.0.eb`, before `cffi`
* the default `python` command is actually Python 3 (but I don't think that's relevant here)

It took me a looong time to figure out what is actually going wrong here, so I'm documenting this for eternity...

When installing Python extensions, `PythonPackage.install_step` makes sure that the installation prefix is prepended to the `$PYTHONPATH` environment variable; in this case, that's `$EBROOTPREFIXlib/python2.7/site-packages`.

Installing `cffi` from source (with `pip install`) kicks of a Cython installation, which involves compiling stuff from source using `gcc`.

The first `gcc` in `$PATH` is our RPATH wrapper script, which is a bash script that runs the `rpath_args.py` Python script: https://github.com/easybuilders/easybuild-framework/blob/0da2122f350eff50181d55c7ffc5a0ecb0a7021d/easybuild/scripts/rpath_wrapper_template.sh.in#L50)

`rpath_args.py` is run using the same `python` command as used by EasyBuild itself (so `python3`). But `$PYTHONPATH` is set up for Python 2 (since we're installing Python packages as extensions for Python 2), and that's where the problem lies:

* `pip install` triggers a a compilation with `gcc`;
* `gcc` is our RPATH wrapper script, which runs `python3 rpath_args.py ...`;
* the `rpath_args.py` script triggers an `import site`;
* `site` is imported from `$EBROOTPREFIXlib/python2.7/site-packages/site.py`, since that location is listed first in `$PYTHONPATH` (so it wins over the standard library locations for `python3`, uh-oh #1);
  * this `site.py` was installed by our dear friend `setuptools` (which is installed as an extension before `cffi`)
* `site.py` does an `import imp`, in a `try`-`catch`:
   ```python
    try:
        import imp
        ...
    except ImportError:
        continue
    ```
    This code is used in both Python 2 & 3 it seems, but actually only makes sense in Python 3 (since there's no `imp` module in Python 2).
    In this case however the `import imp` actually works, since it's being done in a Python 3 environment (since the `rpath_args.py` script is being run with `python3`).
* the `import imp` leads to `import tokenize`, which leads to `import re`, which are all imported from the Python 3 standard library;
* the `import re` triggers an `import enum`, which is resolved by the `enum34` extension that is installed in `$EBROOTPREFIXlib/python2.7/site-packages` (uh-oh #2);
* this triggers the `AttributeError` because the `enum` package provided by `enum34` is not what Python 3 expects (it expects the `enum` included in Python 3 standard library), and so things go boom 💥 ...

And this is why friends don't let friends combine Python 2 and 3 in the same environment.

This is easy to fix: we just need to make sure that the `rpaths_args.py` script is run using `python -E` (to ignore `$PYTHONPATH`) or `python -S` (to skip the `import site`). Doing both is probably the best way forward, to make the RPATH wrapper scripts more robust... The `rpaths_args.py` script is pretty basic, doesn't need anything special (Python 2/3  standard library is sufficient), so it's fine to limit what it has access to.

edit: side note: https://github.com/pypa/pip/issues/8214 looks very similar, but it really a different problem (this bug has nothing to do with `pip` perse)

boegel:
fixes #3421

### DIFF
--- a/easybuild/scripts/rpath_wrapper_template.sh.in
+++ b/easybuild/scripts/rpath_wrapper_template.sh.in
@@ -46,8 +46,14 @@ CMD=`basename $0`
 log "found CMD: $CMD | original command: %(orig_cmd)s | orig args: '$(echo \"$@\")'"
 
 # rpath_args.py script spits out statement that defines $CMD_ARGS
-log "%(python)s -O %(rpath_args_py)s $CMD '%(rpath_filter)s' '%(rpath_include)s' $(echo \"$@\")'"
-rpath_args_out=$(%(python)s -O %(rpath_args_py)s $CMD '%(rpath_filter)s' '%(rpath_include)s' "$@")
+# options for 'python' command (see https://docs.python.org/3/using/cmdline.html#miscellaneous-options)
+# * -E: ignore all $PYTHON* environment variables that might be set (like $PYTHONPATH);
+# * -O: run Python in optimized mode (remove asserts, ignore stuff under __debug__ guard);
+# * -s: don’t add the user site-packages directory to sys.path;
+# * -S: disable the import of the module site and the site-dependent manipulations of sys.path that it entails;
+# (once we only support Python 3, we can (also) use -I (isolated mode)
+log "%(python)s -E -O -s -S %(rpath_args_py)s $CMD '%(rpath_filter)s' '%(rpath_include)s' $(echo \"$@\")'"
+rpath_args_out=$(%(python)s -E -O -s -S %(rpath_args_py)s $CMD '%(rpath_filter)s' '%(rpath_include)s' "$@")
 
 log "rpath_args_out:
 $rpath_args_out"
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -1975,6 +1975,43 @@ class ToolchainTest(EnhancedTestCase):
         # any other available 'g++' commands should not be a wrapper or our fake g++
         self.assertFalse(any(os.path.samefile(x, fake_gxx) for x in res[2:]))
 
+        # RPATH wrapper should be robust against Python environment variables & site-packages magic,
+        # so we set up a weird environment here to verify that
+        # (see https://github.com/easybuilders/easybuild-framework/issues/3421)
+
+        # redefine $HOME so we can put up a fake $HOME/.local/lib/python*/site-packages,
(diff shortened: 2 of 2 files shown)


casparvl (approved):
LGTM, this solves some nice headaches.

PR #3425, opened 2020-08-31
fix lack of -mno-recip on aarch64 by tweaking GCC options used when 'precise' toolchain option is enabled

terjekv:


boegel (approved):
lgtm, thanks @terjekv!

PR #3426, opened 2020-09-01
make sure self.start_dir is set in ExtensionEasyBlock

Flamefire:
Currently it is confusing whether you should/can use `self.start_dir` in an EasyBlock (or EC) leading to subtile errors such as in https://github.com/easybuilders/easybuild-easyconfigs/pull/11109 where `start_dir` is `None` or workarounds such as `topdir = self.start_dir or self.ext_dir` which are repetitive and require comments to explain.

This solution maintains current behavior such that `ext_dir` is set to the extracted folder and the working directory is changed to start_dir or if that is not set to ext_dir iff `unpack_src=True`.
Additionally in any case `start_dir` is now set to either the current start_dir or ext_dir whichever is set and exists

### DIFF
--- a/easybuild/framework/extensioneasyblock.py
+++ b/easybuild/framework/extensioneasyblock.py
@@ -97,6 +97,17 @@ class ExtensionEasyBlock(EasyBlock, Extension):
 
         self.ext_dir = None  # dir where extension source was unpacked
 
+    def _set_start_dir(self):
+        # Don't change start_dir if it is already set and exists
+        # else if ext_dir (path to extracted source) is set and exists use that, similar to guess_start_dir
+        possible_dirs = (self.start_dir, self.ext_dir)
+        for possible_dir in possible_dirs:
+            if possible_dir and os.path.isdir(possible_dir):
+                self.cfg['start_dir'] = possible_dir
+                self.log.debug("Using start_dir: %s", self.start_dir)
+                return
+        self.log.debug("Unable to determine start_dir as none of these paths is set and exists: %s", possible_dirs)
+
     def run(self, unpack_src=False):
         """Common operations for extensions: unpacking sources, patching, ..."""
 
@@ -105,11 +116,11 @@ class ExtensionEasyBlock(EasyBlock, Extension):
             targetdir = os.path.join(self.master.builddir, remove_unwanted_chars(self.name))
             self.ext_dir = extract_file(self.src, targetdir, extra_options=self.unpack_options,
                                         change_into_dir=False)
-            change_dir(self.ext_dir)
 
-            if self.start_dir and os.path.isdir(self.start_dir):
-                self.log.debug("Using start_dir: %s", self.start_dir)
-                change_dir(self.start_dir)
+            self._set_start_dir()
+            change_dir(self.start_dir)
+        else:
+            self._set_start_dir()
 
         # patch if needed
         EasyBlock.patch_step(self, beginpath=self.ext_dir)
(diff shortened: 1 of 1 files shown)


easybuild/framework/extensioneasyblock.py
```diff
@@ -105,11 +116,11 @@ def run(self, unpack_src=False):
             targetdir = os.path.join(self.master.builddir, remove_unwanted_chars(self.name))
             self.ext_dir = extract_file(self.src, targetdir, extra_options=self.unpack_options,
                                         change_into_dir=False)
-            change_dir(self.ext_dir)
 
-            if self.start_dir and os.path.isdir(self.start_dir):
-                self.log.debug("Using start_dir: %s", self.start_dir)
-                change_dir(self.start_dir)
+            self._set_start_dir()
+            change_dir(self.start_dir)
```
ocaisa, line 121: Shouldn't this be
```
change_dir(self.cfg['start_dir'])
```
since you never alter the value of `self.start_dir`?

Flamefire: `self.start_dir` is a property which accesses `self.cfg['start_dir']` so they are synonym (but you can't set the former)

boegel: Indeed, see https://github.com/easybuilders/easybuild-framework/blob/186eaf9a243cdea4a0c147cada6e3c6b8239db2e/easybuild/framework/easyblock.py#L1569-L1572 (this is inherited here since `ExtensionEasyBlock` derives from `EasyBlock`

boegel (changes requested):
Small suggestion, makes sense to me overall.

easybuild/framework/extensioneasyblock.py
```diff
@@ -97,6 +97,17 @@ def __init__(self, *args, **kwargs):
 
         self.ext_dir = None  # dir where extension source was unpacked
 
+    def _set_start_dir(self):
+        # Don't change start_dir if it is already set and exists
```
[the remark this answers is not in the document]
Flamefire: Used a slightly different solution by incorporating the comments into the docstring. See https://github.com/easybuilders/easybuild-framework/pull/3426/commits/31dba1089260d3160212ff8cf50d6566d51a442f

ocaisa (approved):
I'm ok with this, any final comments @boegel

boegel (approved):
lgtm

Flamefire:
FTR: It is still possible that `self.start_dir` is not set: When the EC does not specify it and the sources are *not* extracted then it will be None (as before)

easybuild/framework/extensioneasyblock.py
```diff
@@ -105,11 +119,11 @@ def run(self, unpack_src=False):
             targetdir = os.path.join(self.master.builddir, remove_unwanted_chars(self.name))
             self.ext_dir = extract_file(self.src, targetdir, extra_options=self.unpack_options,
                                         change_into_dir=False)
-            change_dir(self.ext_dir)
 
-            if self.start_dir and os.path.isdir(self.start_dir):
-                self.log.debug("Using start_dir: %s", self.start_dir)
-                change_dir(self.start_dir)
+            self._set_start_dir()
+            change_dir(self.start_dir)
```
boegel, line 124: Turns out this is a breaking change, because the `isdir(self.start_dir)` check is now being done from a different working directory.

That matters (a lot), because the `self.start_dir` value is usually a path relative to the unpacked sources for the extension.

I'll look into fixing this, and covering it in the tests, this was way too easy to overlook...

Flamefire: Good catch! Suggestion (pseudo code): 
```
start_dir = self.start_dir or '.'
ext_dir = self.ext_dir or os.pwd()
os.path.make_absolute(start_dir, ext_dir)
then check for is_dir and use
```

boegel: problem fixed in #3435

PR #3428, opened 2020-09-02
consider $EB_INSTALLPYTHON to specify 'python' command to use for running EasyBuild

ocaisa:
Replaces #3394 (which was created on a branch in the easybuilders repo by accident)

This replaces the use of the `EB_PYTHON` in an EasyBuild module (generated by EasyBuild). `EB_PYTHON` still takes precedence

### DIFF
--- a/eb
+++ b/eb
@@ -43,7 +43,7 @@ function verbose() {
 }
 
 PYTHON=
-for python_cmd in ${EB_PYTHON} 'python' 'python3' 'python2'; do
+for python_cmd in ${EB_PYTHON} ${EB_INSTALLPYTHON} 'python' 'python3' 'python2'; do
 
     verbose "Considering '$python_cmd'..."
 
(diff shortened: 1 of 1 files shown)


migueldiascosta:
according to the comments in https://github.com/easybuilders/easybuild-framework/pull/3394 and https://github.com/easybuilders/easybuild-easyblocks/issues/2078#issuecomment-669009467, this solution is not consensual yet?

Or are the clarifications in the eb script sufficient now?

akesandgren:
This is clearly the best solution for the problem at hand in my opinion.

akesandgren (approved):
LGTM

akesandgren:
Won't merge though until I get the go-ahead from @boegel

boegel:
Makes sense, the clarifying comment clearly explains why we consider two environment variables now. :+1:

ocaisa:
Thanks @akesandgren don't forget to also review/merge https://github.com/easybuilders/easybuild-easyblocks/pull/2109 or this won't do much!

PR #3429, opened 2020-09-02
Architecture-specific filtering of dependencies.

terjekv:
- Done in the superclass for all EasyConfigs
  - Supports a list of dependencies to hard ignore for a given
    architecture.
  - Required to avoid Yasm on aarch64, but Yasm is required on
    x86_64 for performance when building x26[45].

Fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/11190

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -74,6 +74,7 @@ from easybuild.tools.module_naming_scheme.utilities import det_hidden_modname, i
 from easybuild.tools.modules import modules_tool
 from easybuild.tools.py2vs3 import OrderedDict, create_base_metaclass, string_type
 from easybuild.tools.systemtools import check_os_dependency, pick_dep_version
+from easybuild.tools.systemtools import get_cpu_architecture, AARCH64
 from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME, is_system_toolchain
 from easybuild.tools.toolchain.toolchain import TOOLCHAIN_CAPABILITIES, TOOLCHAIN_CAPABILITY_CUDA
 from easybuild.tools.toolchain.utilities import get_toolchain, search_toolchain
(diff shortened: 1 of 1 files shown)


ocaisa:
@boegel This looks ok to me, but I'd like to get your input on the approach

ocaisa:
@boegel Is it possible for us to override `ARCH_HARD_FILTER` in the tests so a test can be added for this in https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/easyconfig.py#L1332?

terjekv:
The warning about hard filtering will be added irrespective of the filtered easyconfig being a dependency or not. Would it be preferable to only add the warning if the dependency is actually in use? (And if so, is there a pretty way to know that?)

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -96,6 +97,9 @@
 # prefix for names of local variables in easyconfig files
 LOCAL_VAR_PREFIX = 'local_'
 
+ARCH_HARD_FILTER = {
+    AARCH64: ['Yasm'],
```
boegel, line 101: Hmm, I'm not a big fan of this approach...

I would rather implement something more flexible, where you can configure EasyBuild to opt-in to or opt-out of optional dependencies based on certain condition (i.e. `aarch64`, GPU system, etc.).

In this particular case, configuring EasyBuild with `--filter-deps=Yasm` should be sufficient?

terjekv: It's ugly for sure. Using `--filter-deps=Yasm` should work, but it would need to always be done so for aarch64.

Dynamic dependencies would be interesting, could spiral into "we depend on cuda if we see a compatible GPU", but absolute exceptions like Yasm would need to be defined somewhere in EasyBuild itself. I have no good/better ideas on how to achieve that.

akesandgren:
https://github.com/easybuilders/easybuild-framework/pull/3506 solves this on a case by case basis which should be enough to move forward.

terjekv:
I agree, https://github.com/easybuilders/easybuild-framework/pull/3506 is a much better solution. Closing this.

PR #3432, opened 2020-09-03
 More tuning of GCC options for aarch64.

terjekv:
There are plenty more things missing from gcc on aarch64...

  - '-mno-recip', '-mieee-fp', and '-mfno-math-errno' are all gone.
  - A full pass over all the options needs to be done.


terjekv:
Broken due to not being branched from develop. Will redo.

PR #3433, opened 2020-09-04
Further GCC toolchain fixes for aarch64.

terjekv:
- Both strict and precise use no-recip.
  - strict wants -mieee-fp. :-(


migueldiascosta:
lgtm - should it be made clearer somewhere that for aarch64 `strict` and `precise` will be the same thing?

terjekv:
There needs to be more fixes all around as other options may still not work, but I'll keep that in mind.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @terjekv!

PR #3436, opened 2020-09-07
Exclude protobuf and pybind11 in createSubmoduleDeps

Flamefire:
With the new EasyBlock those can be installed with EasyBuild and hence are no longer required to download


Flamefire:
@akesandgren Please review. I used this successfully in TF 2.2

boegel (approved):
lgtm

PR #3439, opened 2020-09-09
add option to link MKL dynamically instead of statically

migueldiascosta:


migueldiascosta:
hm, it seems using a build_option here breaks a few things that run before set_up_configuration()...

boegel:
> hm, it seems using a build_option here breaks a few things that run before set_up_configuration()...

Yes, indeed, triggering a `build_option` call too early means trouble, since it relies on the configuration being set up already.

akesandgren:
Why limit this to MKL? Why not just select dynamic libs all over?

migueldiascosta:
@akesandgren I think the other LinAlgs are already using dynamic libs. I don't know why eb linkl MKL statically by default, but there is at least one use case for linking dynamically (in order to use LD_PRELOAD...), so adding an option instead of changing the default should be better... (?)

akesandgren:
I frequently see openblas being linked in as .../libopenblas.a so I don't think it's doing shared libs by default for the other libs either.

So perhaps we should check what the defaults really are?

PR #3440, opened 2020-09-09
Improve apply_regex_substitutions

Flamefire:
- Fix a bug where the passed list was modified leading to outputs of '<_sre.SRE_Pattern..>' when the same list is passed in again
- Minor speedup by only using replacements when a match was found
- Make the log more readable by aggregating the replacements into a single output
- Add `on_missing_match` parameter with `warn` as default (from `--strict`) to catch outdated or faulty replacements but give invokers control over what should be done


test/framework/filetools.py
```diff
@@ -1159,6 +1159,7 @@ def test_apply_regex_substitutions(self):
             (r"^(FC\s*=\s*).*$", r"\1${FC}"),
             (r"^(.FLAGS)\s*=\s*-O3\s-g(.*)$", r"\1 = -O2\2"),
         ]
+        regex_subs_str = str(regex_subs)
```
Micket, line 1162:
```suggestion
        regex_subs_copy = regex_subs.copy()
```

test/framework/filetools.py
```diff
@@ -1169,6 +1170,8 @@ def test_apply_regex_substitutions(self):
         ])
         new_testtxt = ft.read_file(testfile)
         self.assertEqual(new_testtxt, expected_testtxt)
+        # Must not have touched the list -> Formatted output is still the same
+        self.assertEqual(regex_subs_str, str(regex_subs))
```
Micket, line 1174:
```suggestion
        self.assertEqual(regex_subs_copy, regex_subs)
```

test/framework/filetools.py
Micket: this confused me a bit, i think a copy will work fine for this purpose?

Micket (changes requested):

test/framework/filetools.py
Flamefire, line 1174:
```suggestion
        # Must not have touched the list
        self.assertEqual(regex_subs_copy, regex_subs)
```

Flamefire:
Just wanted to highlight that this is expected to print warnings on current ECs/EasyBlocks. E.g. just got one from Bazel 3.4.1: 

> WARNING: Nothing found to replace in tools/cpp/cc_configure.bzl

It's not a hard failure but incentivizes checking why those replacements don't apply: Either an upstream fix (so no longer required), an upstream change (so need to be adjusted) or a mistake in the regexps.

Question also: Currently I warn only when none of the replacements applied. Maybe this should be enhanced to be more fine granular: Like `WARNING: 1 out of 4 replacements not found in tools/cpp/cc_configure.bzl`

akesandgren:
That would probably be a good idea. But perhaps leave for a later enhancement unless it is trivial?

test/framework/filetools.py
```diff
@@ -1159,6 +1159,7 @@ def test_apply_regex_substitutions(self):
             (r"^(FC\s*=\s*).*$", r"\1${FC}"),
             (r"^(.FLAGS)\s*=\s*-O3\s-g(.*)$", r"\1 = -O2\2"),
         ]
+        regex_subs_copy = regex_subs.copy()
```
boegel, line 1162:
```suggestion
        regex_subs_copy = regex_subs[:]
```

easybuild/tools/filetools.py
```diff
@@ -1279,14 +1279,23 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
...
 
-def apply_regex_substitutions(path, regex_subs, backup='.orig.eb'):
+def apply_regex_substitutions(path, regex_subs, backup='.orig.eb', on_missing_match=None):
     """
     Apply specified list of regex substitutions.
 
     :param path: path to file to patch
     :param regex_subs: list of substitutions to apply, specified as (<regexp pattern>, <replacement string>)
     :param backup: create backup of original file with specified suffix (no backup if value evaluates to False)
+    :param on_missing_match: Define what to do when the file when no match was found in the file.
```
boegel, line 1289:
```suggestion
    :param on_missing_match: Define what to do when no match was found in the file.
```

easybuild/tools/filetools.py
```diff
@@ -1279,14 +1279,23 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
...
+                             Can be 'error' to raise an error, 'warn' to print a warning or 'ignore' to do nothing
+                             Defaults to value of --strict
     """
+    if on_missing_match is None:
+        on_missing_match = build_option('strict')
+    ALLOWED_VALUES = (run.ERROR, run.WARN, run.IGNORE)
+    if on_missing_match not in ALLOWED_VALUES:
+        raise EasyBuildError('Invalid value passed to on_missing_match: %s', on_missing_match)
```
boegel, line 1297: Also mention allowed values:

```suggestion
        raise EasyBuildError('Invalid value passed to on_missing_match: %s (allowed: %s)', on_missing_match, ', '.join(allowed_values))
```

easybuild/tools/filetools.py
```diff
@@ -1279,14 +1279,23 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
...
     :param regex_subs: list of substitutions to apply, specified as (<regexp pattern>, <replacement string>)
     :param backup: create backup of original file with specified suffix (no backup if value evaluates to False)
+    :param on_missing_match: Define what to do when the file when no match was found in the file.
+                             Can be 'error' to raise an error, 'warn' to print a warning or 'ignore' to do nothing
+                             Defaults to value of --strict
     """
+    if on_missing_match is None:
+        on_missing_match = build_option('strict')
+    ALLOWED_VALUES = (run.ERROR, run.WARN, run.IGNORE)
+    if on_missing_match not in ALLOWED_VALUES:
```
boegel, line 1296: This is a local variable, not a proper constant, so should just be lowercase:

```suggestion
    allowed_values = (run.ERROR, run.WARN, run.IGNORE)
    if on_missing_match not in allowed_values:

**Review by boegel — CHANGES_REQUESTED** [review]

**Comment by Flamefire** [comment]
> That would probably be a good idea. But perhaps leave for a later enhancement unless it is trivial?

It's not to hard but likely leads to more warnings. As mentioned I think that's ok. 
Shall I go ahead?

**Comment by boegel** [comment]
> > That would probably be a good idea. But perhaps leave for a later enhancement unless it is trivial?
> 
> It's not to hard but likely leads to more warnings. As mentioned I think that's ok.
> Shall I go ahead?

We don't want to start producing a large amount of warnings for what's basically no-op actions. If the patching no longer works but doesn't cause trouble, it shouldn't result in a verbose warning in the `eb` output I think.

That already applies here, regardless of whether we make the warnings more fine-grained or not...

**Comment by Flamefire** [comment]
> If the patching no longer works but doesn't cause trouble, it shouldn't result in a verbose warning in the eb output I think.

But what if the patching is important and the trouble is only detected much later when the software is actually used? E.g. for most *Make builds we don't even run tests...

**Comment by boegel** [comment]
@Flamefire Can you revive this by fixing the merge conflict?

**Comment by Flamefire** [comment]
Rebased and resolved, hope that works

**Comment by boegel** [comment]
> > If the patching no longer works but doesn't cause trouble, it shouldn't result in a verbose warning in the eb output I think.
> 
> But what if the patching is important and the trouble is only detected much later when the software is actually used? E.g. for most *Make builds we don't even run tests...

That's a fair point, but the only way to catch issues like that is to get better at testing, not producing scary looking warnings for things that most likely don't matter. Usually if the patching silently failed it'll either trigger an installation error or it doesn't matter at all, I expect. There may be other cases, but we need to detect those in some other way imho.

Just *logging* a warning (`log.warning` rather than `print_warning`) would be fine though...

File: easybuild/tools/filetools.py [hunk 7]
```diff
@@ -1319,13 +1328,25 @@ def apply_regex_substitutions(path, regex_subs, backup='.orig.eb'):
...
+                        line = regex.sub(subtxt, line)
                 sys.stdout.write(line)
+            if replacement_msgs:
+                _log.info('Applied the following substitutions to %s:\n%s', path, '\n'.join(replacement_msgs))
+            else:
+                msg = 'Nothing found to replace in %s' % path
+                if on_missing_match == run.ERROR:
+                    raise EasyBuildError(msg)
+                elif on_missing_match == run.WARN:
+                    print_warning(msg)
```
Flamefire, line 1347:
```suggestion
                    log.warning(msg)
```

@boegel So like this?

boegel: Yes.

boegel:
@Flamefire This will need some conflict resolving after the merge of #3507

Flamefire:
Rebased. Wasn't particularly great to resolve the conflicts. Please doublecheck the result

easybuild/tools/filetools.py
```diff
@@ -1333,14 +1333,24 @@ def apply_patch(patch_file, dest, fn=None, copy=False, level=None, use_git_am=Fa
...
+def apply_regex_substitutions(paths, regex_subs, backup='.orig.eb', on_missing_match=None):
     """
     Apply specified list of regex substitutions.
 
     :param paths: list of paths to files to patch (or just a single filepath)
     :param regex_subs: list of substitutions to apply, specified as (<regexp pattern>, <replacement string>)
     :param backup: create backup of original file with specified suffix (no backup if value evaluates to False)
+    :param on_missing_match: Define what to do when no match was found in the file.
+                             Can be 'error' to raise an error, 'warn' to print a warning or 'ignore' to do nothing
+                             Defaults to value of --strict
```
boegel, line 1345: `None` should be mentioned there too, since that's the default value.
And perhaps the default should be `'strict'` (which corresponds to following `--strict`)

easybuild/tools/filetools.py
```diff
@@ -1379,14 +1387,31 @@ def apply_regex_substitutions(paths, regex_subs, backup='.orig.eb'):
...
+                    elif on_missing_match == run.WARN:
+                        _log.warning(msg)
+                    else:
+                        msg = 'Nothing found to replace in %s' % path
+                        if on_missing_match == run.ERROR:
+                            raise EasyBuildError(msg)
+                        elif on_missing_match == run.WARN:
+                            print_warning(msg)
+                        else:
+                            _log.info(msg)
```
boegel, line 1414: @Flamefire Duplicated block of code, only `_log.info(msg)` should be under `else:`?

test/framework/filetools.py
```diff
@@ -1201,10 +1205,34 @@ def test_apply_regex_substitutions(self):
...
+        # Error
+        self.assertErrorRegex(EasyBuildError, error_pat, ft.apply_regex_substitutions, testfile, regex_subs_no_match,
+                              on_missing_match=run.ERROR)
+
+        fancylogger.logToFile(self.logfile)
+
+        # Warn
+        ft.apply_regex_substitutions(testfile, regex_subs_no_match, on_missing_match=run.WARN)
+        with open(self.logfile, 'r') as f:
+            logtxt = f.read()
```
boegel, line 1225: please use `read_file` from `filetools` here

test/framework/filetools.py
```diff
@@ -1201,10 +1205,34 @@ def test_apply_regex_substitutions(self):
...
+        self.assertTrue('WARNING ' + error_pat in logtxt)
+
+        # Ignore
+        ft.apply_regex_substitutions(testfile, regex_subs_no_match, on_missing_match=run.IGNORE)
+        with open(self.logfile, 'r') as f:
+            logtxt = f.read()
```
boegel, line 1231: please use `read_file` from `filetools` here

test/framework/filetools.py
```diff
@@ -1201,10 +1205,34 @@ def test_apply_regex_substitutions(self):
...
+        self.assertTrue('INFO ' + error_pat in logtxt)
+
+        fancylogger.logToFile(self.logfile, enable=False)
```
boegel, line 1234: This should (also) be in `tearDown`, to make sure it's always done, also if the tests fails somewhere in the middle?

boegel (changes requested):

easybuild/tools/filetools.py
Flamefire: I didn't want to mention `None` as that shouldn't be used. In a perfect world there would be no default value, I only added it for backwards compatibility. The caller should make a conscious decision what to do.
Hence also the use of `None` to represent "nothing passed".

test/framework/filetools.py
Flamefire: `tearDown` is tricky. It would be run for all tests, which might have side effects. I added a contextmanager which solves that better.

akesandgren:
@Flamefire More conflicts to resolve.

@boegel What is your take on this now?

boegel (approved):

Issue #3289, opened 2020-04-18
--upload-test-report should post comment with test report in easyblocks PR if --include-easyblock-from-pr was used


boegel:
(but only if it's used without `--from-pr`)

That would be useful to have a record of with which existing easyconfigs testing was done for PRs that change existing easyblocks (like https://github.com/easybuilders/easybuild-easyblocks/pull/2030 for example).

Test case, using `Score-P-6.0-gompi-2019a.eb` easyconfig file that is already included in EasyBuild releases:

```shell
eb Score-P-6.0-gompi-2019a.eb -f --include-easyblocks-from-pr 2030
```

migueldiascosta:
in the `--from-pr` case, we could reference the easyblock PR in the easyconfig PR comment?

PR #3446, opened 2020-09-16
take into account --include-easyblocks-from-pr when uploading test reports

migueldiascosta:
closes https://github.com/easybuilders/easybuild-framework/issues/3289

### DIFF
--- a/easybuild/tools/testing.py
+++ b/easybuild/tools/testing.py
@@ -249,15 +249,12 @@ def upload_test_report_as_gist(test_report, descr=None, fn=None):
     return gist_url
 
 
-def post_pr_test_report(pr_nr, repo, test_report, msg, init_session_state, success):
+def post_pr_test_report(pr_nr, repo_type, test_report, msg, init_session_state, success):
     """Post test report in a gist, and submit comment in easyconfigs or easyblocks PR."""
 
     github_user = build_option('github_user')
     pr_target_account = build_option('pr_target_account')
-    if repo == 'easyconfigs':
-        pr_target_repo = build_option('pr_target_repo') or GITHUB_EASYCONFIGS_REPO
-    else:
-        pr_target_repo = build_option('pr_target_repo') or GITHUB_EASYBLOCKS_REPO
+    pr_target_repo = build_option('pr_target_repo') or repo_type
 
     # create gist with test report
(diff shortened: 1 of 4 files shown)


easybuild/tools/testing.py
```diff
@@ -248,18 +249,21 @@ def upload_test_report_as_gist(test_report, descr=None, fn=None):
...
+def post_pr_test_report(pr_nr, repo, test_report, msg, init_session_state, success):
+    """Post test report in a gist, and submit comment in easyconfigs or easyblocks PR."""
 
     github_user = build_option('github_user')
     pr_target_account = build_option('pr_target_account')
-    pr_target_repo = build_option('pr_target_repo') or GITHUB_EASYCONFIGS_REPO
+    if repo == 'easyconfigs':
+        pr_target_repo = build_option('pr_target_repo') or GITHUB_EASYCONFIGS_REPO
+    else:
+        pr_target_repo = build_option('pr_target_repo') or GITHUB_EASYBLOCKS_REPO
```
akesandgren, line 260: If we manage to enhance this to framework in the future (test reports that is) then we should protect against that by making this a case-stmt instead (of elif) and error out if repo isn't easyconfigs/easyblocks. That way we'll easily catch this one.

easybuild/tools/testing.py
```diff
@@ -276,18 +280,29 @@ def post_easyconfigs_pr_test_report(pr_nr, test_report, msg, init_session_state,
...
-        "Test report by @%s" % github_user,
-        ('**FAILED**', '**SUCCESS**')[success],
+    comment_lines = ["Test report by @%s" % github_user]
+
+    easyblocks_pr_nr = build_option('include_easyblocks_from_pr')
+    if easyblocks_pr_nr:
+        if repo == 'easyconfigs':
+            comment_lines.append("Using easyblocks from https://github.com/%s/%s/pull/%s" % (
+                pr_target_account, GITHUB_EASYBLOCKS_REPO, easyblocks_pr_nr))
+        else:
```
akesandgren, line 290: Same here ...

boegel (approved):
lgtm

boegel:
Thanks a lot @migueldiascosta, this is going to be really useful...

PR #3451, opened 2020-09-25
add path to pkg-config files in sysroot to $PKG_CONFIG_PATH when --sysroot is specified

boegel:
Setting `$PKG_CONFIG_PATH` to include the alternate sysroot location fixes problems with `pkg-config` not finding libraries like `zlib` in case they are available in the alternate sysroot and have been filtered out as proper EasyBuild dependency via `--filter-deps`.

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -1039,7 +1039,7 @@ class Toolchain(object):
             # update $PKG_CONFIG_PATH to include sysroot location to pkg-config files (*.pc)
             sysroot_pc_paths = [os.path.join(sysroot, 'usr', libdir, 'pkgconfig') for libdir in ['lib', 'lib64']]
 
-            pkg_config_path = [p for p in os.getenv('PKG_CONFIG_PATH', '').split(':') if p]
+            pkg_config_path = [p for p in os.getenv('PKG_CONFIG_PATH', '').split(os.pathsep) if p]
 
             for sysroot_pc_path in sysroot_pc_paths:
                 if os.path.exists(sysroot_pc_path):
@@ -1048,7 +1048,7 @@ class Toolchain(object):
                         pkg_config_path.append(sysroot_pc_path)
 
             if pkg_config_path:
-                setvar('PKG_CONFIG_PATH', ':'.join(pkg_config_path))
+                setvar('PKG_CONFIG_PATH', os.pathsep.join(pkg_config_path))
 
     def _add_dependency_variables(self, names=None, cpp=None, ld=None):
         """ Add LDFLAGS and CPPFLAGS to the self.variables based on the dependencies
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -821,6 +821,9 @@ class Toolchain(object):
         :param rpath_include_dirs: extra directories to include in RPATH
         """
 
+        # take into account --sysroot configuration setting
+        self.handle_sysroot()
+
         # do all dependencies have a toolchain version?
         if deps is None:
             deps = []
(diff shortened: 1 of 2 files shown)


easybuild/tools/toolchain/toolchain.py
```diff
@@ -1025,6 +1028,28 @@ def prepare_rpath_wrappers(self, rpath_filter_dirs=None, rpath_include_dirs=None
...
+    def handle_sysroot(self):
+        """
+        Extra stuff to be done when alternate system root is specified via --sysroot EasyBuild configuration option.
+
+        * Update $PKG_CONFIG_PATH to include sysroot location to pkg-config files (*.pc).
+        """
+        sysroot = build_option('sysroot')
+        if sysroot:
+            # update $PKG_CONFIG_PATH to include sysroot location to pkg-config files (*.pc)
+            sysroot_pc_paths = [os.path.join(sysroot, 'usr', libdir, 'pkgconfig') for libdir in ['lib', 'lib64']]
```
akesandgren, line 1040: There is actually reasons to do this even when sysroot is not used.
I.e. picking up the system pkgconfig files as a last resort... It's really no different than when using sysroot.

easybuild/tools/toolchain/toolchain.py
```diff
@@ -1025,6 +1028,28 @@ def prepare_rpath_wrappers(self, rpath_filter_dirs=None, rpath_include_dirs=None
...
+
+            pkg_config_path = [p for p in os.getenv('PKG_CONFIG_PATH', '').split(':') if p]
```
akesandgren, line 1042: Split on os.pathsep instead of explicitly using ":". Windows uses ";"

test/framework/toolchain.py
```diff
@@ -166,6 +166,53 @@ def test_is_system_toolchain(self):
...
+        sysroot = sysroot.replace('usr/lib/pkgconfig', 'usr/lib64/pkgconfig')
+        mkdir(sysroot_pkgconfig, parents=True)
+        init_config(build_options={'sysroot': sysroot})
+
+        del os.environ['PKG_CONFIG_PATH']
+        tc.prepare()
+        self.assertEqual(os.getenv('PKG_CONFIG_PATH'), sysroot_pkgconfig)
+
+        # existing $PKG_CONFIG_PATH value is retained
+        test_pkg_config_path = ':'.join([self.test_prefix, '/foo/bar'])
```
akesandgren, line 199: os.,pathsep

easybuild/tools/toolchain/toolchain.py
```diff
@@ -1025,6 +1028,28 @@ def prepare_rpath_wrappers(self, rpath_filter_dirs=None, rpath_include_dirs=None
...
+
+            for sysroot_pc_path in sysroot_pc_paths:
+                if os.path.exists(sysroot_pc_path):
+                    # avoid adding duplicate paths
+                    if not any(os.path.exists(x) and os.path.samefile(x, sysroot_pc_path) for x in pkg_config_path):
+                        pkg_config_path.append(sysroot_pc_path)
+
+            if pkg_config_path:
+                setvar('PKG_CONFIG_PATH', ':'.join(pkg_config_path))
```
akesandgren, line 1051: use os.pathsep.join instead

test/framework/toolchain.py
```diff
@@ -166,6 +166,53 @@ def test_is_system_toolchain(self):
...
+        os.environ['PKG_CONFIG_PATH'] = test_pkg_config_path
+        tc.prepare()
+        self.assertEqual(os.getenv('PKG_CONFIG_PATH'), test_pkg_config_path + ':' + sysroot_pkgconfig)
+
+        # no duplicate paths are added
+        test_pkg_config_path = ':'.join([self.test_prefix, sysroot_pkgconfig, '/foo/bar'])
```
akesandgren, line 205: os.pathsep

test/framework/toolchain.py
```diff
@@ -166,6 +166,53 @@ def test_is_system_toolchain(self):
...
+
+        del os.environ['PKG_CONFIG_PATH']
+        tc.prepare()
+        self.assertEqual(os.getenv('PKG_CONFIG_PATH'), sysroot_pkgconfig)
+
+        # existing $PKG_CONFIG_PATH value is retained
+        test_pkg_config_path = ':'.join([self.test_prefix, '/foo/bar'])
+        os.environ['PKG_CONFIG_PATH'] = test_pkg_config_path
+        tc.prepare()
+        self.assertEqual(os.getenv('PKG_CONFIG_PATH'), test_pkg_config_path + ':' + sysroot_pkgconfig)
```
akesandgren, line 202: os.pathsep

akesandgren (changes requested):

easybuild/tools/toolchain/toolchain.py
boegel: I think `pkg-config` will check the system paths by default, no?

akesandgren: No, it will only check what's in PKG_CONFIG_PATH

boegel: I'm not sure it's a good idea to always set `$PKG_CONFIG_PATH`, since generally we want to *avoid* that stuff is picked up from the OS? Unless `--filter-deps` is used, maybe...

You can always just set `$PKG_CONFIG_PATH` outside of `eb` to make sure it `*.pc` files are picked up from the OS (since we're careful here no to overwrite existing paths in `$PKG_CONFIG_PATH`)...

akesandgren: True, but why then set it to sysroot in the first place??

boegel: Because using `--sysroot` (in combination with `--filter-deps`) should imply that `$PKG_CONFIG_PATH` is set to the same path.

akesandgren: With filter-deps perhaps but this doesn't check filter-deps being set at all.

And sysroot set or not isn't really relevant. It should behave the same in both cases.

So, I'd vote for not setting it at all. Set it from the outside if you need it.

akesandgren: Perhaps a config option to decide if PKG_CONFIG gets set or not when filter-deps is used??
But that's a bit too specific for my taste.

boegel: Yeah, that's very specific indeed...

To me it makes sense to do this automatically when `--sysroot` is used, since then you're clearly building on top of something else where you want to use libraries/tools from.

That's not the case when building without a custom sysroot?

akesandgren: Well, that's one way to look at it.

Setting it only when filter-deps is in use would make more sense to me though, but I'm not going to argue that.

Should we merge as is then?

akesandgren (approved):

akesandgren:
Going in, thanks @boegel!

PR #3454, opened 2020-09-28
add support for NVHPC compiler + toolchain (based on PGI)

AndiH:
PGI beyond 20.4 is not called PGI any more but _NVIDIA HPC SDK_, with 20.7 being the first version; https://developer.nvidia.com/hpc-sdk.

We call the (set of) software(s) _NVHPC_, after internal discussions and on recommendation by NVHPC's product owners. They call NVIDIA HPC SDK _NVHPC_ internally as well. 

This pull request is part of a set of three pull requests to introduce NVHPC to EasyBuild ([1](https://github.com/easybuilders/easybuild-framework/pull/3454), [2](https://github.com/easybuilders/easybuild-easyblocks/pull/2190), [3](https://github.com/easybuilders/easybuild-easyconfigs/pull/11391)). The stack has been successfully deployed at JSC and is under use by our HPC users.  
The EasyBuild files are created after discussions with the NVHPC product owners, taking into account HPC practices and EasyBuild conventions.

The toolchain and compiler introduced with this pull request are closely based on PGI, cleaned up for unsupported flags.


easybuild/toolchains/compiler/nvhpc.py
```diff
@@ -0,0 +1,111 @@
...
+
+:author: Bart Oldeman (McGill University, Calcul Quebec, Compute Canada)
+:author: Damian Alvarez (Forschungszentrum Juelich GmbH)
+:author: Andreas Herten (Forschungszentrum Juelich GmbH)
+"""
+
+from distutils.version import LooseVersion
+
+import easybuild.tools.systemtools as systemtools
+from easybuild.tools.build_log import EasyBuildError
```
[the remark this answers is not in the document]
AndiH: Can I get rid of this import? I am not sure.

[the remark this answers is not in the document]
boegel: @AndiH Please remove this line, since `EasyBuildError` is not used?

easybuild/toolchains/compiler/nvhpc.py
```diff
@@ -0,0 +1,111 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+Support for NVIDIA HPC SDK ('NVHPC') compilers (nvc, nvc++, nvfortran) as toolchain compilers. NVHPC is the successor of the PGI compilers, on which this file is based upon.
```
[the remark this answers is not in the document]
boegel: @AndiH All lines should be shorter than 120 characters, so let's split this:

```suggestion
Support for NVIDIA HPC SDK ('NVHPC') compilers (nvc, nvc++, nvfortran) as toolchain compilers.
NVHPC is the successor of the PGI compilers, on which this file is based upon.
```

boegel (changes requested):

easybuild/toolchains/compiler/nvhpc.py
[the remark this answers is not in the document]
AndiH: Do I really need to change this? I find it stupid in 2020 to enforce line length limits, buuuut I follow suit if I must :).

easybuild/toolchains/compiler/nvhpc.py
```diff
@@ -0,0 +1,106 @@
...
+
+import easybuild.tools.systemtools as systemtools
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.toolchain.compiler import Compiler
+
+
+TC_CONSTANT_NVHPC = "NVHPC"
+
+
+class Nvhpc(Compiler):
```
boegel, line 44: @AndiH Doesn't matter much, but `NVHPC` makes more sense here imho.

```suggestion
class NVHPC(Compiler):

File: easybuild/toolchains/compiler/nvhpc.py [hunk 4]
```diff
@@ -0,0 +1,106 @@
...
+        (systemtools.X86_64, systemtools.INTEL): 'tp=host',
+    }
+    # used with --optarch=GENERIC
+    COMPILER_GENERIC_OPTION = {
+        (systemtools.X86_64, systemtools.AMD): 'tp=px',
+        (systemtools.X86_64, systemtools.INTEL): 'tp=px',
+    }
+
+    COMPILER_CC = 'nvc'
+    # C++ compiler command is version-dependent, see below
```
boegel, line 86: @AndiH This comment is no longer correct, so should be removed?

easybuild/toolchains/nvhpc.py
```diff
@@ -0,0 +1,46 @@
...
+:author: Bart Oldeman (McGill University, Calcul Quebec, Compute Canada)
+:author: Andreas Herten (Forschungszentrum Juelich)
+"""
+
+from easybuild.toolchains.compiler.nvhpc import Nvhpc
+from easybuild.toolchains.gcccore import GCCcore
+from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME
+
+
+class NvhpcToolchain(Nvhpc):
```
boegel, line 40: I would keep using `NVHPC` in all caps (also requires fixing the `import` statement above):

```
class NVHPCToolchain(NVHPC):

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/toolchains/compiler/nvhpc.py (hunk 3 above)
Comment (AndiH) [reply easybuild/toolchains/compiler/nvhpc.py:44]: This came from `pgi.py`:
```
class Pgi(Compiler):
```
I don't mind at all. But I need to propagate this to the other files, right?

boegel: Yeah, it has to be consistent in `import` statements & such, indeed.

For `pgi.py` it should've been `PGI` I guess (it doesn't matter much in practice, but `Nvhpc` looks weird to me)

AndiH: Alright! I'll change and propagate to the other files when fixing our other comments.

Thanks!

AndiH:
I think that should be it.

Tests on our JSC machine with the current state of the files (including renaming "Nvhpc" --> "NVHPC") seemed to work well.

boegel:
Hmm, that failing test has nothing to do with the changes made here, I've re-triggered the tests, some kind of race condition?

boegel (approved):

boegel:
lgtm, thanks @AndiH!

PR #3460, opened 2020-10-05
Check for _VERSION and _PREFIX Cray environment variables with both software and module name

ekouts:
Cray should include the variables `<MODULE_NAME>_VERSION` and `<MODULE_NAME>_PREFIX` in the module. The module name is sometimes different than the software name, so EasyBuild should check for both cases.


boegel (approved):
@ekouts I cleaned this up a bit, and enhanced the tests to cover the use case you're fixing here... Thanks for making the effort to look into the fix yourself!

ekouts:
@boegel Thanks a lot for fixing it!

PR #3461, opened 2020-10-09
Fix missing string template on Error for extension sources

Flamefire:
There are 2 values but only one `%s`

Bug introduced in https://github.com/easybuilders/easybuild-framework/commit/79aa779b07d9bfc9e25b3990f0b2527a4540208e by @boegel (you might want to have a look)


ocaisa (approved):

ocaisa:
Thanks for the fix @Flamefire !

PR #3473, opened 2020-10-19
Allow use of --copy-ec with --from-pr

### LINKED ISSUE #3178: eb --copy-ec does not work with --from-pr yet
would be nice to have that.

ocaisa:
edit: fixes #3178


boegel:
@ocaisa Are you up for enhancing the test for `--copy-ec` to cover the use case (combo with `--from-pr`) you're fixing here?

ocaisa:
I'm working on it now

ocaisa:
@boegel I need a hint here, using `--from-pr` seems to require all the GitHub setup

ocaisa:
This falls short, it won't also copy over the required patch files (having said that, the option `--copy-ec` never promised that)

ocaisa:
@boegel Ready for review

test/framework/options.py
```diff
@@ -1072,6 +1072,20 @@ def check_copied_files():
         self.assertTrue(os.path.exists(copied_toy_cwd))
         self.assertEqual(read_file(copied_toy_cwd), toy_ec_txt)
 
+        # Test --copy-ec coupled with --from-pr
+        all_ecs_pr8007 = [
```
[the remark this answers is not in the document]
boegel: @ocaisa Please fix this minor issue first, so we have a green light in the tests?

easybuild/main.py
```diff
@@ -357,6 +373,12 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
                     log=_log, opt_parser=eb_go.parser, exit_on_error=not testing)
     _log.debug("Paths: %s", paths)
 
+    # if we are only copying ec's for a specific PR we can do that now
+    if options.copy_ec and options.from_pr:
+        determined_paths = [path[0] for path in paths]
+        copy_ecs_to_target(determined_paths, target_path)
```
ocaisa, line 379: The change to use an absolute path for `target_path` was required because at this point, if we use `--from-pr`, we are inside a temporary directory so a relative path points to the wrong place.

easybuild/main.py
```diff
@@ -357,6 +373,12 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
                     log=_log, opt_parser=eb_go.parser, exit_on_error=not testing)
     _log.debug("Paths: %s", paths)
 
+    # if we are only copying ec's for a specific PR we can do that now
+    if options.copy_ec and options.from_pr:
+        determined_paths = [path[0] for path in paths]
```
boegel, line 378: @ocaisa This looks a bit strange, but the entries in `paths` are basically tuples of `(<path>, <bool>)`, where the `<bool>` indicates whether the path is generated or not.

I know you copy-pasted this from parts below, but this is probably clearer:

```python
determined_paths = [path for (path, _) in paths]
copy_ecs_to_target(determined_paths, target_path)
```

or just

```python
copy_ecs_to_target([path for (path, _) in paths], target_path)
```

or event


```python
copy_ecs_to_target([x for (x, _) in paths], target_path)
```

Please change it below too, to keep things consistent.

easybuild/main.py
```diff
@@ -176,6 +176,24 @@ def run_contrib_style_checks(ecs, check_contrib, check_style):
     return check_contrib or check_style
 
 
+def copy_ecs_to_target(determined_paths, target_path, prefix=False):
```
boegel, line 179: Doesn't this belong in `easybuild.framework.easyconfig.tools`?

Ideally we don't add much stuff to `main.py` (since then it can only be used here, nowhere else)

easybuild/main.py
```diff
@@ -321,17 +343,11 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     # determine paths to easyconfigs
     determined_paths = det_easyconfig_paths(categorized_paths['easyconfigs'])
 
-    if (options.copy_ec and not tweaked_ecs_paths) or options.fix_deprecated_easyconfigs or options.show_ec:
+    if (options.copy_ec and not (tweaked_ecs_paths or options.from_pr)) or options.fix_deprecated_easyconfigs or \
+            options.show_ec:
```
boegel, line 347: We should clean this up a bit... How about this:

```python
# only copy easyconfigs here if we're not using --try-* or --from-pr (that's are handled below)
copy_ec = options.copy_ec and not (tweaked_ecs_paths or options.from_pr)
if copy_ec or options.fix_deprecated_easyconfigs or options.show_ec:

File: easybuild/main.py [hunk 6]
```diff
@@ -357,6 +373,12 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
                     log=_log, opt_parser=eb_go.parser, exit_on_error=not testing)
     _log.debug("Paths: %s", paths)
 
+    # if we are only copying ec's for a specific PR we can do that now
```
boegel, line 376: `for a specific PR` -> `from a specific PR`

test/framework/options.py
```diff
@@ -1072,6 +1072,49 @@ def check_copied_files():
...
+        stdout = mocked_main(args)
+        self.assertEqual(stdout, '2 file(s) copied to %s' % test_working_dir)
+        # check that the two easyconfigs exist
+        self.assertTrue(os.path.exists(os.path.join(test_working_dir, all_ecs_pr8007[0])))
+        remove_file(os.path.join(test_working_dir, all_ecs_pr8007[0]))
+        self.assertTrue(os.path.exists(os.path.join(test_working_dir, all_ecs_pr8007[2])))
+        remove_file(os.path.join(test_working_dir, all_ecs_pr8007[2]))
+        # ...but the patch doesn't
+        self.assertFalse(os.path.exists(os.path.join(test_working_dir, all_ecs_pr8007[1])))
+        remove_file(os.path.join(test_working_dir, all_ecs_pr8007[1]))
```
boegel, line 1096: Hmm, would it be difficult to fix this? I would expect that the patch gets copied too...

We should also test copying a single easyconfig from a PR where multiple are touched?
Like `eb --copy-ec --from-pr 8007 Arrow-0.7.1-intel-2017b-Python-3.6.3.eb <target_dir>`

easybuild/main.py
```diff
@@ -303,12 +321,16 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
             eb_file = find_easybuild_easyconfig()
             orig_paths.append(eb_file)
 
-    if len(orig_paths) == 1:
+    if not orig_paths:
+        # if no easyconfig files are specified and we are using --from-pr, use current directory as target directory
+        target_path = os.getcwd() if (options.copy_ec and options.from_pr) else None
```
boegel, line 326: How about:

```python
# if only one easyconfig is specified, or if none are specified and we are using --from-pr,
# use current directory as target directory
if len(orig_paths) == 1 or (options.copy_ec and options.from_pr and len(orig_paths) == 0):
    target_path = os.getcwd()
elif orig_paths:
    ...

File: easybuild/main.py [hunk 9]
```diff
@@ -303,12 +321,16 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+    elif len(orig_paths) == 1 and not (options.copy_ec and options.from_pr):
         # if only one easyconfig file is specified, use current directory as target directory
         target_path = os.getcwd()
     elif orig_paths:
         # last path is target when --copy-ec is used, so remove that from the list
-        target_path = orig_paths.pop() if options.copy_ec else None
+        # use the absolute path as some use cases occur in a temporary directory
```
boegel, line 332: In which case are we copying to the target path from a different directory?
I'm guessing that's with `--from-pr` (we should mention it here rather than leave ourselves guessing in the future)

Perhaps that should be considered a bug?

test/framework/options.py
```diff
@@ -2067,15 +2110,15 @@ def test_try_with_copy(self):
 
         self.mock_stdout(True)
         self.mock_stderr(True)
-        tweaked_ecs_dir = os.path.join(self.test_buildpath, 'my_tweaked_ecs')
-        self.eb_main(args + ['--try-software=foo,1.2.3', '--try-toolchain=gompi,2018a', tweaked_ecs_dir],
-                     verbose=True, raise_error=True)
+        tweaked_ecs_dir = os.path.join(self.test_buildpath)
+        os.chdir(tweaked_ecs_dir)
```
boegel, line 2114: Why was this changed? Doesn't this signal a (breaking) change in behavior?

boegel (changes requested):

test/framework/options.py
ocaisa: I can easily copy *all* patches that are alongside an easyconfig, but I don't know how to (easily) filter the ones that are part of a PR

easybuild/main.py
ocaisa: Hmm, not sure what you mean by doing it below, the case there is very different

easybuild/main.py
ocaisa: I agree that this behaviour is probably a bit of a bug

test/framework/options.py
ocaisa: Yeah, it does change behaviour but it's not a breaking change. For the case where you were tweaking things, it always copied the tweaked files to a new directory (even if there was only one path in the list, which is the case in the test). This is different behaviour to what happens in the original use case, where if there was only one file, it would do a simple copy to the destination path. I thought it was probably more important to make the behaviour consistent in all cases.

test/framework/options.py
ocaisa: Regarding the additional test, it is not going to pick that up, according to the existing logic, it is going to assume that the additional argument is the target path. I'm not sure what logic I could use to get around that, I'd need to know if the filename(s) appear(s) in the PR.

ocaisa: Maybe I can do it before I copy the files, if the basename of the target path matches one of the paths then I assume that is what was meant and only copy that single file to the basedir of the target path

easybuild/main.py
boegel: `regtest([path[0] for path in paths] or ...` -> `regtest([x for (x, _) in paths] or`

test/framework/options.py
ocaisa: I can revert this, but I should probably take the same approach with `--from-pr`, where you don't necessarily know how many ecs you will end up with.

ocaisa:
I think I've made this overly complicated, now that I've had a look at existing GitHub functionality I could have just used `fetch_files_from_pr()` and this would be done

boegel (changes requested):
I spent quite a bit of time cleaning things up on top of these changes to clean things up a bit, see #3482

PR #3477, opened 2020-10-21
unset $LD_LIBRARY_PATH when checking for OS dependencies with 'rpm' & co

boegel:
This fixes issue like the one we danced around in https://github.com/easybuilders/easybuild-easyconfigs/pull/4179 by patching XZ, once and for all.

We saw the issue popping up again on RHEL8, because of missing symbols in the vanilla XZ installation that is picked up by the `rpm` command when the module is loaded...

```
$ rpm -q rdma-core-devel
rpm: /software/XZ/5.2.4-GCCcore-8.3.0/lib/liblzma.so.5: version `XZ_5.2' not found (required by /lib64/librpmio.so.8)
```

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -677,8 +677,15 @@ def check_os_dependency(dep):
 
     for pkg_cmd in pkg_cmds:
         if which(pkg_cmd):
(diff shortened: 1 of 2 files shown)


bartoldeman (approved):
lgtm

PR #3479, opened 2020-10-22
add (experimental) support for specifying easyconfig files via "easystack" file

deniskristak:
Added support for easybuild-ing from provided specsfile. Currently works with yaml, can be extended to json etc. 
Original task: https://github.com/easybuilders/easybuild-framework/issues/3468


boegel:
@deniskristak (very) quick feedback:

* I don't think there's a need to spread the new code you have across different files in `easybuild/tools/specsfile/*.py` , just add a single file `easybuild/tools/specsfile.py` with all of it (it'll still be short).
* clean it up a bit so the code style checks are happy (you can check yourself by running flake8, see https://flake8.pycqa.org).
* try to add tests for the functions you've implemented in `specsfile.py`

Maybe we also need a better name than "specsfile", which sounds very generic... Any suggestions on that?

cc @casparvl

casparvl:
Nice start Denis! Just for fun, I tried to run it on the example yaml file I provided in my original issue and got

```
Bioconductor-3.11-foss-2020a

 full_path: []

Bioconductor does not have clearly specified parameters - 0 matches found. Skipping.

R-4.0.0-foss-2020a

 full_path: []

R does not have clearly specified parameters - 0 matches found. Skipping.

GROMACS-2020.1-foss-2020a

 full_path: []

GROMACS does not have clearly specified parameters - 0 matches found. Skipping.

GROMACS-2020.3-foss-2020a

 full_path: ['/hpc/eb/easyconfigs-surfsara/g/GROMACS/GROMACS-2020.3-foss-2020a.eb']

GROMACS-2020.1-fosscuda-2020a

 full_path: []

GROMACS does not have clearly specified parameters - 0 matches found. Skipping.

EasyBuild-4.3.1-SYSTEM-

 full_path: []

EasyBuild does not have clearly specified parameters - 0 matches found. Skipping.

OpenFOAM-8-foss-2020a

 full_path: ['/hpc/eb/easyconfigs-surfsara/o/OpenFOAM/OpenFOAM-8-foss-2020a.eb']

OpenFOAM-v2006-foss-2020a

 full_path: []

OpenFOAM does not have clearly specified parameters - 0 matches found. Skipping.


 determined_paths: ['/hpc/eb/easyconfigs-surfsara/g/GROMACS/GROMACS-2020.3-foss-2020a.eb', '/hpc/eb/easyconfigs-surfsara/o/OpenFOAM/OpenFOAM-8-foss-2020a.eb']


WARNING: Failed to determine toolchain hierarchy for foss/2020a when determining subtoolchain for dependency 'CMake': 'Could not find easyconfig for foss toolchain version 2020a'
```

looks like what you intended :)

> Maybe we also need a better name than "specsfile", which sounds very generic... Any suggestions on that?

Hm, good one. Everything sounds unclear, or too long to me. Buildspec? Buildspecfile? Buildlist? Buildlistspec? The problem with the first two is that it doesn't reflect that this concerns (typically) building _more_ than one software. "Buildlist" is again quite generic, and doesn't cover the level of detail with which you can specify build behaviour (once we allow e.g. specifying command line arguments per item in the yaml file). "Buildlistspec" covers quite well _what_ the yaml file is, but may be a bit long. And I find _all_ of my suggestions ugly, so I'm open to other suggestions too ;-)

I've checked a bit the code. And I have two major points

(1) I see that you're currently calling search_easyconfigs. I'm not sure if that is a wise road to go down: what if multiple EasyConfigs are found? Or none? How is EasyBuild supposed to continue from that? In my view, the specfile should specify unambiguously which EasyConfig is intended. I.e. to me

```
software:
 GROMACS:
    toolchains:
      foss-2020a:
        exclude-labels: system:gpu
        versions:
          2020.1:
            from_pr: 1234
```

Should quite litterally do the equivalent of

`eb GROMACS-2020.1-foss-2020a.eb --from-pr 1234`

since the specfile really specifies _all_ (except the `.eb` extension) the information that is needed. A spec missing e.g. a toolchain version is simply an invalid spec and should just error out - just like e.g. `eb GROMACS-2020.1-foss.eb` would error out saying that that EasyConfig can't be found.

Currently, it does more of a

`eb -S GROMACS-2020.1-foss-20202a`

Grab the resulting full path (if a unique one is returned) and pass that on to eb.main's `determined_path` parameter.

Why not just parse the spec file to compose a new `orig_paths` and pass that back to main? The only reason I can imagine  _might_ be our request for supporting wildcards, which you may have intended to do through the `search_eascyonfigs`, but... I would probably do that at a more generic EB level so that a `eb GROMACS-*-foss-2020a.eb` would _also_ just build all versions of gromacs with that toolchain that are available. In other words: I'd make `det_easyconfig_paths` more intelligent so that it can deal with wildcards - not try to implement that (only) for our specfile option.

(2) The way it is currently set up, a call to `eb example.yaml` would lead to a single call to EB's main, parse the YAML file, extract the `determined_paths` and then proceed. But: how will that enable us to allow specification of additional arguments (such as a `--from-pr`) for some easyconfigs, but not for others? I don't think that is possible: I think one call to EB's main will always result in one set of arguments that is applied to _all_ EasyConfigs.

What I would do is probably: parse the yaml file, and _compose new calls to eb.main_ from that (with appropriate arguments). That keeps the additional specfile option simpler (the only thing it needs to do is extract an EasyConfig name and related options, and create a loop over that parsed list that simply calls main again).

@boegel what do you think?

casparvl:
Just discussed in a teams meeting, since it is faster than in this PR. Conclusions from the meeting:

- Change `handle_specfile` so that it returns `orig_paths` instead of `determined_path`. `orig_paths` just contains the easyconfig names, so we can eliminate the `search_easyconfigs` functionality from `handle_specfile` itself (the searching is done later in eb's main).
- For now, we limit the scope of the requested funcionality. It will not include support for wildcards, neither will it allow setting options _per_ EasyConfig. The global options (at the top of the specfile) will be returned by `handle_specfile`, but we print a warning that we don't do anything with those (yet).
- Add `log.experimental`, so that the `--specfile` option is seen as an experimental feature
- Add some more tests
-  One to make sure that `handle_specfile` can parse `yaml` files with wildcards (even if we don't do anything with them yet in main)
- Add one end-to-end test that actually installs a dummy EasyConfig (ask Kenneth)
- Add one end-to-end test called on a non-existing file to check that is gracefully handled.
- Add license header

In later (and seperate) iterations, we can (seperately) implement support for

- wildcards, probably by making `det_easyconfig_paths` smarter, so that it can handle wildcards when resolvuing easyconfig names to full paths
- options per item in the YAML file. This requires `eb.main` to be enhanced so that it loops over the returned `orig_paths` from `handle_specfile`, and IN the loop resets EB's configuration, and calls `eb.main` again with the correct (per easyconfig) list of arguments.
- filter by `include` and `exclude` labels in the YAML file. This will require adding a `--label` option to EB that is also passed to `handle_specfile`

Seperating out these three items keeps the current PR as simple as possible, and means that we can potentially work on those three enhancements in parallel. We'll create issues / PRs for this once we get there, but at least we have the ideas written down :)

easybuild/main.py
```diff
@@ -70,6 +70,7 @@
 from easybuild.tools.parallelbuild import submit_jobs
 from easybuild.tools.repository.repository import init_repository
 from easybuild.tools.testing import create_test_report, overall_test_report, regtest, session_state
+from easybuild.tools.build_from_specsfile import handle_specsfile
```
boegel, line 73: rename:
`easybuild.tools.build_from_specsfile` -> `easybuild.tools.easystack` (and maybe it should be under `easybuild.framework` instead, next to `easyconfig` and `easyblock`?)

`handle_specsfile` -> `parse_easystack`

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+_log = fancylogger.getLogger('specsfile', fname=False)
+
+
+# general specs applicable to all commands
+class Specsfile(object):
+    def __init__(self):
+        self.easybuild_version = None
+        self.robot = False
```
boegel, line 14: `robot` is just one of the possible options that can be passed, so we need to handle this in a more generic way.

Leave it as is for now, and open an issue to follow up on this ("add general support for defining configuration options in easystack files")

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+        self.software_list = []
```
boegel, line 15: Just `self.software`?

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # returns list of all commands - finished
+    def compose_full_paths(self):
```
boegel, line 18: this should be renamed, since we only want easyconfig filenames (not paths to them)

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+            path_to_append = self.get_ec_path(sw)
+
+            if path_to_append is None:
+                continue
+            else:
+                easyconfigs_full_paths.append(path_to_append)
+        return easyconfigs_full_paths
+
+    # single command
+    def get_ec_path(self, sw):
```
boegel, line 30: This function can be removed (and the import for `search_easyconfigs` removed)

Above, you should probably use the `det_full_ec_version` version to determine the "full version" part of the easyconfig filename (the part except the software name and the `.eb` extension).
For the `.eb` extension, use the `EB_FORMAT_EXTENSION` constant.

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+        general_options = {}
+        general_options['robot'] = self.robot
+        general_options['easybuild_version'] = self.easybuild_version
+        return general_options
+
+
+# single sw command
+class SoftwareSpecs(object):
+    def __init__(self, software, version, toolchain, toolchain_version, toolchain_name):
+        self.software = software
```
boegel, line 52: `self.software` should be `self.name` (since it's the software *name*)

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+        self.version = version
+        self.toolchain = toolchain
+        self.toolchain_version = toolchain_version
+        self.toolchain_name = toolchain_name
+        self.toolchain = toolchain
+
+        self.version_suffix = None
```
boegel, line 59: `version_suffix` -> `versionsuffix` (to match the easyconfig parameter)

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def get_version_suffix(self):
+        return self.version_suffix or ''
+
+
+# implement this to your own needs - to create custom yaml/json/xml parser
+class GenericSpecsParser(object):
```
boegel, line 66: I would rename this to `EasyStackParser`, and assume we'll only support YAML format.

If we change our minds later, it should be easy to refactor to also support other formats...

"Premature optimizations is the root of all evil"

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+        raise NotImplementedError
+
+
+class YamlSpecParser(GenericSpecsParser):
+    @staticmethod
+    def parse(filename):
+
+        try:
+            with open(filename, 'r') as f:
+                spec_dict = yaml.safe_load(f)
```
boegel, line 78: Please use the `read_file` function (from `easybuild.tools.filetools`) to read the contents of the file, and then pass that to `yaml.safe_load`.

`read_file` already takes care of proper error reporting, so then you don't need to worry about the file not being there here.

What happens if parsing of the YAML file fails?
You should catch the corresponding exception, and raise an `EasyBuildError` in that case, to ensure proper error reporting and not triggering a nasty Python traceback?

Also, rename `spec_dict` to `easystack_dict`?

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+
+            eb = Specsfile()
```
boegel, line 80: Don't use `eb` as local variable, that's very confusing. I would go with `easystack` (`es` is also short and confusing imho)

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+        except FileNotFoundError:
+            raise EasyBuildError("Could not read provided specsfile.")
```
boegel, line 82: Also show the actual error too, so in this case:

```python
except FileNotFoundError as err:
    raise EasyBuildError("Could not read provided specsfile: %s", err)
```

But you should change this to catch parsing errors instead (since `read_file` already handles error reporting for missing files)

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+
+        sw_dict = spec_dict["software"]
```
boegel, line 84: `sw_dict` -> `software`?

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+
+        # assign software-specific EB attributes
+        for software in sw_dict:
```
boegel, line 87: `for key in software` or `for name in software`?

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+            try:
+                # iterates through toolchains to find out what sw version is needed
+                for yaml_toolchain in sw_dict[software]['toolchains']:
+                    # retrieves version number
+                    for yaml_version in sw_dict[software]['toolchains'][yaml_toolchain]['versions']:
```
boegel, line 92: add local variables for stuff you pull from the dict:

```python
try:
    toolchains = software[name]['toolchains']
    for toolchain in toolchains:
        software_versions = toolchains[toolchain_name]['versions']
        for key in software_versions:
            # ensure we have a string value (YAML may provide float value types for a version like '2.3' for example)
            key = str(key)

File: easybuild/tools/build_from_specsfile.py [hunk 15]
```diff
@@ -0,0 +1,150 @@
...
+        general_options['easybuild_version'] = self.easybuild_version
+        return general_options
+
+
+# single sw command
+class SoftwareSpecs(object):
+    def __init__(self, software, version, toolchain, toolchain_version, toolchain_name):
+        self.software = software
+        self.version = version
+        self.toolchain = toolchain
```
boegel, line 54: keeping track of both the toolchain name+version combo and the toolchain name and version separately doesn't make much sense, so drop `self.toolchain` and only use `self.toolchain_version` and `self.toolchain_name`?

Or maybe rework to make `self.toolchain` a dict value with values like `{'name': 'foss', 'version': '2020a'}`, which matches how we usually keep track of toolchains in other places (incl. in easyconfigs).

So:

```python
self.toolchain = {
    'name': toolchain_name,
    'version': toolchain_version,
}

File: easybuild/tools/build_from_specsfile.py [hunk 16]
```diff
@@ -0,0 +1,150 @@
...
+            try:
+                # iterates through toolchains to find out what sw version is needed
+                for yaml_toolchain in sw_dict[software]['toolchains']:
+                    # retrieves version number
+                    for yaml_version in sw_dict[software]['toolchains'][yaml_toolchain]['versions']:
+
+                        # if among versions there is anything else than known flags or version flag itself, it is wrong
+                        # identification of version strings is vague
+                        if str(yaml_version)[0].isdigit() \
+                                or str(yaml_version)[-1].isdigit():
```
boegel, line 97: Is dangerous to make a lot of assumptions here, you wouldn't believe the things we've seen... :)

Even assuming that a software version starts or ends with a digit can lead to limitations we don't want.

I would just handle the special cases (like `versionsuffix` and `exclude-labels`/`include-labels`) first, and if there are not other matches assume you've found a software version, without checking the value at all...

So:

```python
if key == 'versionsuffix':
    ...
elif key in ('exclude-labels', 'include-labels'):
    ...
else:
    # assume the value is a software version
```

In the last case you can add a check to ensure the software version wasn't found yet.
If you hit two keys that would be interpreted as a software version, something is wrong?

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+                            # creates a sw class instance
+                            try:
+                                yaml_toolchain_name = str(yaml_toolchain).split('-', 1)[0]
+                                yaml_toolchain_version = str(yaml_toolchain).split('-', 1)[1]
```
boegel, line 101: This can be simplified a lot I think...
```python
if '-' in toolchain:
    toolchain_name, toolchain_version = toolchain.split('-', 1)
else:
    toolchain_name, toolchain_version = toolchain, ''
```

No more need for `try`/`except` in that case?

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+                                               toolchain_version=yaml_toolchain_version)
+
+                            # append newly created class instance to the list inside EbFromSpecs class
+                            eb.software_list.append(sw)
+
+                        elif str(yaml_version) == 'versionsuffix':
+                            try:
+                                version_info = sw_dict[software]['toolchains'][yaml_toolchain]['versions'][yaml_version]
+                                if version_info['versionsuffix'] is not None:
+                                    sw.version_suffix = version_info['versionsuffix']
```
boegel, line 117: Only thing to keep in mind here is that you shouldn't make assumptions over the order in which keys in the dict are being processed, so you may be assigning the `versionsuffix` to the wrong `SoftwareSpecs` instance here if `'versionsuffix'` is processed first, before the software version for this entry.

So, I think you need to first try and locate the software version to construct the `SoftwareSpecs`, and then iterate over the remaining keys.

Or you postpone creating the actual instance until after this entry is processed.

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+                            except (KeyError, TypeError, IndexError):
```
boegel, line 118: This looks weird, why would it be OK to silently ignore errors like this?

If `versionsuffix` may not be there, use `version_info.get('versionsuffix')`, and check for `None`?

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+                                continue
+                        elif str(yaml_version) == 'exclude-labels' \
```
boegel, line 120: use `in` rather than using `==` twice (see above)

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+                                or str(yaml_version) == 'include-labels':
+                            continue
+
+                        else:
+                            print('Software % s has wrong yaml structure!' % (str(software)))
```
boegel, line 125: No need for the print, raising an error is enough.

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+                            raise EasyBuildError('Wrong yaml structure')
```
boegel, line 126: Provide some more info in the error message: what was being parsed when this happened? Which software name for example...

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+
+            except (KeyError, TypeError, IndexError):
+                print('Software % s has wrong yaml structure!' % (str(software)))
+                raise EasyBuildError('Wrong yaml structure')
```
boegel, line 130:
```python
except Exception as err:
    raise EasyBuildError("Failed to interpret easystack entry: %s", err)
```

If you can provide any specific info at this point (software name?), do so.

easybuild/tools/build_from_specsfile.py
```diff
@@ -0,0 +1,150 @@
...
+
+        # assign general EB attributes
+        eb.easybuild_version = spec_dict.get('easybuild_version', None)
+        eb.robot = spec_dict.get('robot', False)
+        return eb
+
+
+def handle_specsfile(filename):
```
boegel, line 138: rename to `parse_easystack`?

easybuild/tools/options.py
```diff
@@ -604,6 +604,9 @@ def informative_options(self):
             'show-full-config': ("Show current EasyBuild configuration (all settings)", None, 'store_true', False),
             'show-system-info': ("Show system information relevant to EasyBuild", None, 'store_true', False),
             'terse': ("Terse output (machine-readable)", None, 'store_true', False),
+            'specsfile': (
+                "Accepts file containing build specifications (i.e. yaml), parses it and prints all eb commands to run",
```
boegel, line 608: Doesn't print?

boegel (changes requested):
Please add the license header on top of the new module file, see existing module files.
Use `Copyright 2020-2020 Ghent University`, and add yourself as author in the docstring below the license header, also mentioning `INUITS`.

Also, add docstrings for all classes, functions and methods (so they will be covered nicely in https://easybuild.readthedocs.io/en/latest/api/easybuild.html when that gets updated)

Please remove the `binutils-2.25.tar.gz` binary files that got added by accident.

casparvl:
Some clarification examples on the wildcard usage that the function should pass (and that can thus be integrated in the tests):
```
software:
 GROMACS:
    toolchains:
      foss-2020a:
        versions:
          2020.*:
```
```
software:
 GROMACS:
    toolchains:
      foss-2020a:
        versions:
          *:
```
```
software:
 GROMACS:
    toolchains:
      foss-2020*:
        versions:
          2020.1:
```
```
software:
 GROMACS:
    toolchains:
      *:
        versions:
          2020.1:
```
I guess these four variants would be the most important. I hope that the YAML parser supports asterisks in those positions...

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+        """Returns general options (flags applicable to all sw (i.e. --robot))"""
+        general_options = {}
+        # TODO add support for general_options
+        # general_options['robot'] = self.robot
+        # general_options['easybuild_version'] = self.easybuild_version
+        return general_options
+
+
+class SoftwareSpecs(object):
+    """Contains information about every SW that should be installed"""
```
boegel, line 70: `SW` -> `software` (don't use shorthands in docstrings and leave people guessing what is meant exactly)

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+    def __init__(self, software, version, toolchain, toolchain_version, toolchain_name):
+        self.software = software
+        self.version = version
+        self.toolchain = toolchain
```
boegel, line 74: Do we still need `self.toolchain`, since we have both toolchain name and version separately?

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+        self.toolchain_version = toolchain_version
+        self.toolchain_name = toolchain_name
+        self.toolchain = toolchain
+        self.versionsuffix = None
```
boegel, line 78: Use `''` as default value (otherwise you'll run into trouble in`compose_ec_filenames` above)

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class Easystack(object):
+    """One class instance per easystack. General options + list of all SoftwareSpecs instances"""
+    def __init__(self):
+        self.easybuild_version = None
+        self.robot = False
+        self.software_list = []
+
+    def compose_ec_names(self):
```
boegel, line 47: rename to `compose_ec_filenames`?

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+class GenericSpecsParser(object):
+    """Parent of all implemented parser classes"""
+    @ staticmethod
+    def parse(filename):
+        """Parent of all implemented parser functions (i.e. JSON, XML...)"""
+        raise NotImplementedError
+
+
+class YamlSpecParser(GenericSpecsParser):
+    """YAML file parser"""
```
boegel, line 95: Since we call `YamlSpecParser` below directly, I don't see the point of having both `GenericSpecsParser` and `YamlSpecParser`.

I doubt we'll actually look into alternate formats in the future (sticking to a single format avoids confusion), so please rework to:

```python
class EasyStackParser(object):
    """Parser for easystack files."""
    @staticmethod
    def parse(filename):
        ...

File: easybuild/tools/build_from_easystack.py [hunk 31]
```diff
@@ -0,0 +1,173 @@
...
+    @ staticmethod
```
boegel, line 96: remove space after `@`

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+    def parse(filename):
```
boegel, line 97: argument is a *path* to a file, not just the name?

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+        """Parses YAML file and assigns obtained values to SW config instances as well as general config instance"""
+        try:
+            with open(filename, 'r') as f:
```
boegel, line 100: Why not use `read_file`?

```python
def parse(filepath):
    """Parse easystack file and return EasyStack instance."""
    easystack_raw = yaml.safe_load(read_file(path))
```

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+                spec_dict = yaml.safe_load(f)
+
+            eb = Easystack()
```
boegel: rename `Easystack` to `EasyStack`

don't use `eb` for local variables, that's confusing (`eb` is the name of the EasyBuild command, and could also be "easyblock")

Please use `easystack` (or `es`, but that's also confusing imho)

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+        except Exception:
```
boegel, line 104: No need for `try`/`except` here if you use `read_file`

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+            raise EasyBuildError("Could not read provided easystack.")
+
+        sw_dict = spec_dict["software"]
```
boegel, line 107: I would rename `spec_dict` to `easystack_raw` or `easystack_raw_dict`?

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # assign software-specific EB attributes
+        for software in sw_dict:
+            try:
+                # iterates through toolchains to find out what sw version is needed
+                for yaml_toolchain in sw_dict[software]['toolchains']:
+                    # retrieves version number
+                    for yaml_version in sw_dict[software]['toolchains'][yaml_toolchain]['versions']:
```
boegel: Please use a variable for stuff you're using in multiple places, and avoid using `yaml_` (at this point, it doesn't matter anymore where the data comes from?)

```python
for software_name, software_data in easystack_raw_dict.items():
    # make very sure we have a string value (parsed YAML data may have other value types)
    software_name = str(software_name)
    toolchains = software_data['toolchains']
    for toolchain_name, toolchain_data in toolchains.items():
        toolchain_name = str(toolchain_name)
        for key in toolchain_data['versions']:
            key = str(key)
```

I used `key` here in the inner loop since at this point you're not sure yet it's a version, if could be `'versionsuffix'` or `'exclude-labels'` (in the future).

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+
+                        # if among versions there is anything else than known flags or version flag itself, it is wrong
+                        # identification of version strings is vague
+                        if str(yaml_version)[0].isdigit() \
+                                or str(yaml_version)[-1].isdigit():
```
boegel, line 135: As I already mentioned, making assumptions about what a software version looks like are pointless. Even a simple assumption like you're making here (starts or ends with a digit) may not be correct. In some cases a version could be all letters (`latest`) or a Git commit ID (`abc12def`).

So don't check anything here, try the other cases first (only `versionsuffix`), and when you've find nothing else, assume you've find a software version.

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+                                yaml_toolchain_version = ''
+
+                            sw = SoftwareSpecs(software=software, version=yaml_version,
+                                               toolchain=yaml_toolchain, toolchain_name=yaml_toolchain_name,
+                                               toolchain_version=yaml_toolchain_version)
+
+                            # append newly created class instance to the list inside EbFromSpecs class
+                            eb.software_list.append(sw)
+
+                        elif str(yaml_version) == 'versionsuffix':
```
boegel: As I also mentioned already, you're making assumptions about the order in which keys are being iterated over, which you shouldn't.

What if you hit `versionsuffix` first, and then the corresponding software version entry afterwards? (then you would assign the versionsuffix to the *previous* entry, that's wrong).

Maybe this indicates that there's another level of nesting required in the easystack structure?

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+                            try:
+                                version_info = sw_dict[software]['toolchains'][yaml_toolchain]['versions'][yaml_version]
```
boegel: This is why you use variables for stuff you use in multiple places. Use `toolchain_data` here instead of `sw_dict[software]['toolchains'][yaml_toolchain]`, that will makes this a lot more readable.

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+                                if version_info['versionsuffix'] is not None:
+                                    sw.version_suffix = version_info['versionsuffix']
```
boegel: `sw.version_suffix` is wrong, and it indicates a gap in the tests (they should fail on this), should be `sw.versionsuffix`.

And I think you can skip the `if` check with:

```python
sw.versionsuffix = version_info.get('versionsuffix', '')
```

(the `''` is the fallback value in case `'versionsuffix'` is not in `version_info`)

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+                            except (KeyError, TypeError, IndexError):
```
boegel: Why is this `except` needed? It shouldn't be (and silently ignoring it is not OK)

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+                                continue
+                        elif str(yaml_version) == 'exclude-labels' \
+                                or str(yaml_version) == 'include-labels':
```
boegel: I would remove this for now, and complain when you hit these?

We'll revisit this in a future PR.

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+                            continue
+                        else:
+                            raise EasyBuildError('Software % s has wrong yaml structure!' % (str(software)))
+
+            except (KeyError, TypeError, IndexError):
```
boegel, line 153: Make this more general: if any error happens while interpreting the data, give up (and mention the error that occurred):

```python
except Exception as err:
    raise EasyBuildError("An error occurred when interpreting the data for software %s: %s", software, err)

File: easybuild/tools/build_from_easystack.py [hunk 45]
```diff
@@ -0,0 +1,173 @@
...
+        # assign general EB attributes
+        eb.easybuild_version = spec_dict.get('easybuild_version', None)
+        eb.robot = spec_dict.get('robot', False)
+        return eb
+
+
+def parse_easystack(filename):
+    """Parses through easystack file, returns what EC are to be installed together with their options."""
+    _log.experimental("Support for easybuild-ing from multiple easyconfigs based on \
+information obtained from provided file (easystack) with build specifications.")
```
boegel, line 165: please avoid the ugly line wrapping:

```python
log_msg = "Support for easybuild-ing from multiple easyconfigs based on "
log_msg += "information obtained from provided file (easystack) with build specifications."
_log.experimental(log_msg)

File: easybuild/tools/build_from_easystack.py [hunk 46]
```diff
@@ -0,0 +1,173 @@
...
+            except (KeyError, TypeError, IndexError):
+                raise EasyBuildError('Software % s has wrong yaml structure!' % (str(software)))
+
+        # assign general EB attributes
+        eb.easybuild_version = spec_dict.get('easybuild_version', None)
+        eb.robot = spec_dict.get('robot', False)
+        return eb
+
+
+def parse_easystack(filename):
```
boegel, line 162: `filename` -> `filepath`

easybuild/tools/build_from_easystack.py
```diff
@@ -0,0 +1,173 @@
...
+    _log.info("Building from easystack: '%s'" % filename)
+
+    # class instance which contains all info about planned build
+    eb = YamlSpecParser.parse(filename)
+
+    easyconfig_names = eb.compose_ec_names()
+
+    general_options = eb.get_general_options()
+
+    _log.debug("Easystack parsed. Proceeding to install these Easyconfigs: \n'%s'" % ',\n'.join(easyconfig_names))
```
boegel, line 175: Also log general options?

easybuild/tools/options.py
```diff
@@ -604,6 +604,9 @@ def informative_options(self):
             'show-full-config': ("Show current EasyBuild configuration (all settings)", None, 'store_true', False),
             'show-system-info': ("Show system information relevant to EasyBuild", None, 'store_true', False),
             'terse': ("Terse output (machine-readable)", None, 'store_true', False),
+            'easystack': (
+                "Parses file containing build specifications (i.e. yaml) and initiates build accordingly",
```
boegel, line 608:
```python
            'easystack': ("Path to easystack file in YAML format, specifying details of a software stack",
                          None, 'store', None),

File: test/framework/easystacks/test_easystack_1.yaml [hunk 49]
```diff
@@ -0,0 +1,33 @@
+easybuild_version: 4.3.0
+robot: True
+software:
+  Bioconductor:
+    toolchains:
+      foss-2020a:
+        versions:
+            3.11:
+            versionsuffix: '-R-4.0.0'
+            exclude-labels: arch:aarch64
```
boegel: these should be a level deeper, *under* `3.11`

```yaml
software:
  Bioconductor:
    toolchains:
      foss-2020a:
        versions:
            3.11:
              versionsuffix: '-R-4.0.0'
              exclude-labels: arch:aarch64
```

boegel (changes requested):

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+        self.toolchain = toolchain
+        self.versionsuffix = None
+
+    # to be implemented
+    # def get_versionsuffix(self):
+    #     return self.versionsuffix or ''
+
+
+# implement this to your own needs - to create custom yaml/json/xml parser
+class GenericSpecsParser(object):
```
Darkless012, line 87: I would rename this to `GenericEasystackParser`

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+    """Parent of all implemented parser classes"""
+    @ staticmethod
+    def parse(filepath):
+        """Parent of all implemented parser functions (i.e. JSON, XML...)"""
+        raise NotImplementedError
+
+
+class YamlSpecParser(GenericSpecsParser):
```
Darkless012, line 95: Same here as `YamlEasystackParser`

easybuild/tools/easystack.py
deniskristak: good catch :-)

easybuild/tools/easystack.py
deniskristak: also good catch :-)

easybuild/main.py
```diff
@@ -70,6 +70,7 @@
 from easybuild.tools.parallelbuild import submit_jobs
 from easybuild.tools.repository.repository import init_repository
 from easybuild.tools.testing import create_test_report, overall_test_report, regtest, session_state
+from easybuild.tools.easystack import parse_easystack
```
boegel, line 73: @deniskristak I think it makes sense to move `easybuild.tools.easystack` to `easybuild.framework.easystack` (since also the code to deal with easyconfigs and easyblocks lives there).

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+    def __init__(self):
+        self.easybuild_version = None
+        self.robot = False
+        self.software_list = []
+
+    def compose_ec_names(self):
+        """Returns a list of all easyconfig names"""
+        ec_names = []
+        for sw in self.software_list:
+            ec_to_append = '%s-%s-%s-%s.eb' % (str(sw.software), str(sw.version),
```
boegel, line 52: Please use `det_full_ec_version`, rename `sw.software` to `sw.name`, and take into account `versionsufifx`.

Also the `str(...)` should not be needed here anymore (you should deal with that in `YamlEasystackParser.parse`):

```python
full_ec_version = det_full_ec_version({
    'toolchain': {'name': sw.toolchain_name, 'version': sw.toolchain_version},
    'version': sw.version,
    'versionsuffix': sw.versionsuffix,
})
ec_name = '%s-%s.eb' % (sw.name, full_ec_version)

File: easybuild/tools/easystack.py [hunk 54]
```diff
@@ -0,0 +1,186 @@
...
+                                               str(sw.toolchain_name), str(sw.toolchain_version))
+            if ec_to_append is None:
+                continue
+            else:
+                ec_names.append(ec_to_append)
```
boegel, line 57: I would rename `ec_to_append` to `ec_name` (or `ec_filename`)

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+        # TODO add support for general_options
+        # general_options['robot'] = self.robot
+        # general_options['easybuild_version'] = self.easybuild_version
+        return general_options
+
+
+class SoftwareSpecs(object):
+    """Contains information about every SW that should be installed"""
+    def __init__(self, software, version, toolchain, toolchain_version, toolchain_name):
+        self.software = software
```
boegel, line 73: Please rename `self.software` to `self.name`

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+        self.version = version
+        self.toolchain = toolchain
+        self.toolchain_version = toolchain_version
+        self.toolchain_name = toolchain_name
+        self.toolchain = toolchain
+        self.versionsuffix = None
```
boegel, line 79: use `''` as default value for `versionsuffix`?

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # to be implemented
+    # def get_versionsuffix(self):
+    #     return self.versionsuffix or ''
```
boegel, line 83: You can drop this if you use `''` as default value above...

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+        raise NotImplementedError
+
+
+class YamlEasystackParser(GenericEasystackParser):
+    """YAML file parser"""
+    @ staticmethod
+    def parse(filepath):
+        """Parses YAML file and assigns obtained values to SW config instances as well as general config instance"""
+        try:
+            with open(filepath, 'r') as f:
```
boegel, line 101: Please use the `read_file` function here, and just feed the text you obtain to `yaml.safe_load` (and get rid of the `try`/`except`, `read_file` will handle problems like missing file, etc.

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.base import fancylogger
+try:
+    import yaml
+except ImportError:
+    pass
+_log = fancylogger.getLogger('easystack', fname=False)
+
+
+class Easystack(object):
```
boegel, line 41: Please rename to `EasyStack`, with capital `S` (like we do for `EasyConfig`, `EasyBlock`, etc.)

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+        self.toolchain = toolchain
+        self.versionsuffix = None
+
+    # to be implemented
+    # def get_versionsuffix(self):
+    #     return self.versionsuffix or ''
+
+
+# implement this to your own needs - to create custom yaml/json/xml parser
+class GenericEasystackParser(object):
```
boegel, line 87: Again, as I mentioned already, get rid of this, there's no point in trying to make this generic, it's very unlikely we'll implement support in a format different than YAML anyway.

Please get rid of this.

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+    """Parent of all implemented parser classes"""
+    @ staticmethod
+    def parse(filepath):
+        """Parent of all implemented parser functions (i.e. JSON, XML...)"""
+        raise NotImplementedError
+
+
+class YamlEasystackParser(GenericEasystackParser):
```
boegel, line 95: Please rename and derive from `object`:

```python
class EasyStackParser(object):

File: easybuild/tools/easystack.py [hunk 62]
```diff
@@ -0,0 +1,186 @@
...
+    """YAML file parser"""
```
boegel, line 96: Please fix docstring

```python
"""Parser for easystack files (in YAML syntax)."""
```

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+                spec_dict = yaml.safe_load(f)
+
+                eb = Easystack()
+        except Exception:
+            raise EasyBuildError("Could not read provided easystack.")
+
+        sw_dict = spec_dict["software"]
+
+        # assign software-specific EB attributes
+        for software in sw_dict:
```
boegel, line 111: Use `name` for the key, and make sure it's a string value before you start using it below:

```python
for name in software:
    name = str(name)
```

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+        """Parses YAML file and assigns obtained values to SW config instances as well as general config instance"""
+        try:
+            with open(filepath, 'r') as f:
+                spec_dict = yaml.safe_load(f)
+
+                eb = Easystack()
+        except Exception:
+            raise EasyBuildError("Could not read provided easystack.")
+
+        sw_dict = spec_dict["software"]
```
boegel, line 108:
```python
software = spec_dict["software"]

File: easybuild/tools/easystack.py [hunk 65]
```diff
@@ -0,0 +1,186 @@
...
+                                toolchain=yaml_tc, toolchain_name=yaml_toolchain_name,
+                                toolchain_version=yaml_toolchain_version)
+
+                            # append newly created class instance to the list inside EbFromSpecs class
+                            eb.software_list.append(sw)
+                        else:
+                            raise EasyBuildError(wrong_structure_err)
+
+            except (KeyError, TypeError, IndexError):
+                raise EasyBuildError(wrong_structure_err)
```
boegel, line 158: Don't hide the underlying error please:

```python
            except (KeyError, TypeError, IndexError) as err:
                raise EasyBuildError(wrong_structure_err + ": %s" % err)
```

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+        sw_dict = spec_dict["software"]
+
+        # assign software-specific EB attributes
+        for software in sw_dict:
+            asterisk_err = "Easystack specifications of '%s' contain asterisk. " % (str(software))
+            asterisk_err += "Wildcard feature is not supported yet."
+            wrong_structure_err = "Easystack specifications of '%s' have wrong yaml structure." % (str(software))
+            try:
+                # iterates through toolchains to find out what sw version is needed
+                for yaml_tc in sw_dict[software]['toolchains']:
```
boegel, line 117: Please make the names of variables match exactly with what they refer to, that makes things a lot easier to follow:

```python
toolchains = software[name]['toolchains']
for toolchain in toolchains:
    toolchain = str(toolchain)

File: easybuild/tools/easystack.py [hunk 67]
```diff
@@ -0,0 +1,186 @@
...
+                    # if version string containts asterisk or labels, raise error (not implemented yet)
+                    if '*' in str(sw_dict[software]['toolchains'][yaml_tc]['versions']):
```
boegel, line 119: If you access the same part of the parsed file, use local variables.

Also, this check is in the wrong place, you're now just checking whether `'*'` is an entry in the list of versions, while you probably want to check whether `*` appears in *any* version?

```python
versions = toolchains[toolchain]['versions']
for version in versions:
    version = str(version)
    if '*' in version:

File: easybuild/tools/easystack.py [hunk 68]
```diff
@@ -0,0 +1,186 @@
...
+                        raise EasyBuildError(asterisk_err)
+                    for yaml_version in sw_dict[software]['toolchains'][yaml_tc]['versions']:
+                        try:
+                            yaml_ver_specs = sw_dict[software]['toolchains'][yaml_tc]['versions'][yaml_version]
+                            if 'versionsuffix' in str(yaml_ver_specs):
+                                ver_suf_err = "Easystack specifications of '%s' contain versionsuffix. " % str(software)
+                                ver_suf_err += "This isn't supported yet."
+                                raise EasyBuildError(ver_suf_err)
```
boegel, line 127: Isn't it trivial to also support `versionsuffix`?

Just make sure that `versionsuffix` is always defined as `''` if it's not specified, and that you can do `SoftwareSpecs(..., versionsuffix=versionsuffix)`?

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+                            elif 'exclude-labels' in str(yaml_ver_specs) or 'include-labels' in str(yaml_ver_specs):
+                                lab_err = "Easystack specifications of '%s' contain labels. " % str(software)
+                                lab_err += "Labels aren't supported yet."
+                                raise EasyBuildError(lab_err)
+                        except TypeError:
```
boegel, line 132: Again, please don't do this. If there's a `TypeError`, something is very wrong, and you should silently ignore it?!

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+                            pass
+                        if '*' in str(yaml_version):
+                            raise EasyBuildError(asterisk_err)
+                        # TODO - think of better ID of versions
+                        elif str(yaml_version)[0].isdigit() \
+                                or str(yaml_version)[-1].isdigit():
```
boegel, line 138: You're still checking here whether this looks like a version, please don't, you shouldn't make *any* assumptions of what a version looks like. It may be something like `'develop'` or `'latest'`, which is perfectly valid.

Jst use `else:` here, and assume it's a valid version, don't try to interpret it.

easybuild/tools/easystack.py
```diff
@@ -0,0 +1,186 @@
...
+                            # creates a sw class instance
+                            try:
+                                yaml_toolchain_name = str(yaml_tc).split('-', 1)[0]
+                                yaml_toolchain_version = str(yaml_tc).split('-', 1)[1]
```
boegel, line 142: This splitting should be done higher up, right below `toolchain = str(toolchain)`?

boegel (changes requested):

easybuild/main.py
deniskristak: done

easybuild/tools/easystack.py
deniskristak: i believe you are asking if this applies to i.e. `2.2.*`
it does, since the str() func takes the version (which might be of type dict or float at this point) and transfers it to string, which is iterable, therefore `if '*' in .....` is possible

deniskristak: true, it returns error even if there is a correct version spec (without asterisk) alongside the incorrect one, but I suppose it is supposed to stop either way

deniskristak: also, appears I did think about it. there is another condition that covers this - `if '*' in yaml_version:`
which is obsolete, when i think about it (just bcs of what i wrote in previous comments)

easybuild/tools/easystack.py
deniskristak: i think this contradicts with the comment above, so ill name it ec_filename and ec_filenames

easybuild/framework/easystack.py
```diff
@@ -0,0 +1,243 @@
...
+        """Parses YAML file and assigns obtained values to SW config instances as well as general config instance"""
+        try:
+            with open(filepath, 'r') as f:
+                spec_dict = yaml.safe_load(f)
+
+                eb = Easystack()
+        except Exception:
+            raise EasyBuildError("Could not read provided easystack.")
+
+        try:
```
casparvl, line 112: This is a huge `try` block. PEP recommends keeping try blocks as short as possible "Additionally, for all try/except clauses, limit the try clause to the absolute minimum amount of code necessary. Again, this avoids masking bugs." https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations 

I agree, since it is hard to see what failure mode this `try` is testing for, and the `except` can thus only be very generic.

I guess here, you just want to `try` if the `software` keyword is in the spec_dict? That's what I would limit myself too then, so you can throw a super clear error ("Not a valid EasyStack YAML file: no 'software' key found" or something of that nature)

casparvl (changes requested):
Made a few extra notes, but I do see that a lot of Kenneth's comments on the previous version (https://github.com/easybuilders/easybuild-framework/pull/3479/files/e3b3c8248e45c59a8e7ecebd3a873c98ea826ba8) are indeed also still valid.

easybuild/main.py
```diff
@@ -223,6 +224,13 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         last_log = find_last_log(logfile) or '(none)'
         print_msg(last_log, log=_log, prefix=False)
 
+    # if easystack is provided with the command, commands with arguments from it will be executed
+    if options.easystack:
+        # TODO add general_options (i.e. robot) to build options
+        orig_paths, general_options = parse_easystack(options.easystack)
+        if general_options:
+            raise EasyBuildError("Support for general options (flags) is not supported yet.")
```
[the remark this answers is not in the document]
deniskristak: done

easybuild/tools/easystack.py
deniskristak: added support for versionsuffix

easybuild/tools/easystack.py
deniskristak: please check if it is ok in the next iteration

easybuild/tools/easystack.py
deniskristak: the original idea was to mark anything that wasnt known as error, but I get what you are saying...changed

easybuild/framework/easystack.py
```diff
@@ -0,0 +1,243 @@
...
+
+class SoftwareSpecs(object):
+    """Contains information about every SW that should be installed"""
+    def __init__(self, name, version, versionsuffix, toolchain, toolchain_version, toolchain_name):
+        self.name = name
+        self.version = version
+        self.toolchain = toolchain
+        self.toolchain_version = toolchain_version
+        self.toolchain_name = toolchain_name
+        self.toolchain = toolchain
```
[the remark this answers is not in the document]
deniskristak: thanks, nice catch

easybuild/tools/easystack.py
deniskristak: done

easybuild/framework/easystack.py
```diff
@@ -0,0 +1,243 @@
...
+    def __init__(self):
+        self.easybuild_version = None
+        self.robot = False
+        self.software_list = []
+
+    def compose_ec_names(self):
+        """Returns a list of all easyconfig names"""
+        ec_filenames = []
+        for sw in self.software_list:
+            full_ec_version = self.det_full_ec_version({
```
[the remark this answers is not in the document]
deniskristak: lol, i thought that one sounded weird :D

easybuild/framework/easystack.py
deniskristak: you are right, it has been there more or less because I was hoping to avoid doing try-except for all possible index errors that could occur and have one for all missing keywords...but i've already dived into it, so it does make sense to finish for all...

easybuild/tools/easystack.py
deniskristak: I think it's better not to have a default for versionsuffix (see latest iteration), since the code is more readable if I see `yaml_versionsuffix = ''` after `else`, rather than just `pass`... just an idea, we can change it if you want...

easybuild/tools/easystack.py
deniskristak: done

easybuild/tools/easystack.py
deniskristak: done

easybuild/tools/easystack.py
deniskristak: done

easybuild/tools/build_from_easystack.py
deniskristak: probably not

easybuild/tools/build_from_easystack.py
deniskristak: I think after the last version, this is solved and the only requirement should be consistency.
if i ever use `versionsuffix` for a software-toolchain combination, I shouldn't write versions in a format without ':' at the end, like `2.25`...didn't figure out how to check this though...

easybuild/tools/build_from_easystack.py
deniskristak: now it returns err, is it ok?

easybuild/framework/easystack.py
```diff
@@ -0,0 +1,218 @@
...
+information obtained from provided file (easystack) with build specifications.
+
+:author: Denis Kristak (Inuits)
+:author: Pavel Grochal (Inuits)
+"""
+
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.base import fancylogger
+from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
+from easybuild.tools.filetools import read_file
```
boegel, line 35: @deniskristak Please order these alphabetically:

```
from easybuild.base import fancylogger
from easybuild.tools.filetools import read_file
from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
```

easybuild/framework/easystack.py
```diff
@@ -0,0 +1,218 @@
...
+        self.versionsuffix = versionsuffix
+
+
+class EasyStackParser(object):
+    """Parser for easystack files (in YAML syntax)."""
+    @staticmethod
+    def parse(filepath):
+        """Parses YAML file and assigns obtained values to SW config instances as well as general config instance"""
+        f = read_file(filepath)
+        easystack_raw = yaml.safe_load(f)
```
boegel, line 89: Please don't use single-letter variables outside of list comprehensions:

```python
yaml_txt = read_file(filepath)
easystack_raw = yaml.safe_load(yaml_txt)

File: easybuild/framework/easystack.py [hunk 78]
```diff
@@ -0,0 +1,218 @@
...
+        easystack = EasyStack()
+
+        try:
+            software = easystack_raw["software"]
+        except (KeyError):
```
boegel, line 94:
```python
except KeyError:

File: easybuild/framework/easystack.py [hunk 79]
```diff
@@ -0,0 +1,218 @@
...
+            wrong_structure_file = "Not a valid EasyStack YAML file: no 'software' key found"
+            raise EasyBuildError(wrong_structure_file)
+
+        # assign software-specific easystack attributes
+        for name in software:
+            name = str(name)
```
boegel, line 100: add a comment above to explain this:

```python
# ensure we have a string value (YAML parser may return string values like '0.5' as a double value, for example)
name = str(name)

File: easybuild/framework/easystack.py [hunk 80]
```diff
@@ -0,0 +1,218 @@
...
+            try:
+                toolchains = software[name]['toolchains']
+            except (KeyError):
```
boegel, line 103: `except KeyError:`

easybuild/framework/easystack.py
```diff
@@ -0,0 +1,218 @@
...
+                raise EasyBuildError("Toolchains for software '%s' are not defined" % name)
+            for toolchain in toolchains:
+                toolchain = str(toolchain)
+                try:
+                    toolchain_name = toolchain.split('-', 1)[0]
+                    toolchain_version = toolchain.split('-', 1)[1]
```
boegel, line 109: Only split once, ensure you get the correct number of parts, complain if not (and no need for `try`):

```python
toolchain_parts = toolchain.split('-', 1)
if len(toolchain_parts) == 2:
    toolchain_name, toolchain_version = toolchain_parts
elif len(toolchain_parts) == 1:
    toolchain_name, toolchain_version = toolchain, ''
else:
    raise EasyBuildError("Incorrect toolchain specification, too many parts: %s", toolchain_parts)
```

boegel (changes requested):

boegel (approved):

PR #3480, opened 2020-10-23
allow including easyblocks from multiple PRs

migueldiascosta:
closes #3476

PR #3481, opened 2020-10-25
take into account multiple easyblock PRs when uploading test reports

migueldiascosta:
follow-up to #3480

when used with --from-pr, it simply lists the multiple easyblocks in the comment (e.g. https://github.com/easybuilders/easybuild-easyconfigs/pull/11489#issuecomment-716089715)

when used without --from-pr and with multiple easyblock PRs, this leads to some duplication (e.g. https://github.com/easybuilders/easybuild-easyblocks/pull/2191#issuecomment-716087172, https://github.com/easybuilders/easybuild-easyblocks/pull/2205#issuecomment-716087176), these comments should probably only list the easyconfigs related to the easyblocks in each PR...

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1454,11 +1454,11 @@ def set_up_configuration(args=None, logfile=None, testing=False, silent=False):
     init_build_options(build_options=build_options, cmdline_options=options)
 
     # done here instead of in _postprocess_include because github integration requires build_options to be initialized
-    try:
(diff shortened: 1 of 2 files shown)


boegel:
@migueldiascosta Can we cover this fix in the tests?

boegel (approved):
lgtm

boegel:
I overlooked that this was a follow-up to #3480, already covered well enough in the tests, so going in, thanks @migueldiascosta!

PR #3482, opened 2020-10-25
fix combination of --copy-ec and --from-pr

boegel:
This continues from the work done by @ocaisa in #3473, and fixes the combination of `--copy-ec` and `--from-pr`.

The scope of this PR is pretty big, and I feel it should be split up into several PRs, so I made this a draft PR for now.

High level overview of additional changes:

* [x] enhance `--copy-files` function so it can be used to implement `--copy-ec` (both with and without `--from-pr`)
  * => fleshed out to PR ~~#3483~~
* [x] implement caching of result of `fetch_files_from_pr` calls (mostly to significantly speed up tests)
   * => fleshed out to PR ~~#3484~~
* [x] add `locate_files` function to `easybuild.tools.filetools` module which can be leverared in `det_easyconfig_paths` function
   * => fleshed out to PR ~~#3485~~
* [x] avoid that `download_repo` function from `easybuild.tools.github` module changes working directory
   * => fleshed out to PR ~~#3486~~
* add `det_copy_ec_specs` function to leverage from `main.py`
* clean exit after handling combination of `--copy-ec` and `--try-*`
* fix default for `pr_target_account` build option (only really relevant for tests)
* enhance tests to cover all these changes & fixes


easybuild/framework/easyconfig/tools.py
```diff
@@ -728,3 +698,65 @@ def avail_easyblocks():
...
+        # assume that files need to be copied to current working directory for now
+        target_path = os.getcwd()
+
+        if orig_paths:
+            last_path = orig_paths[-1]
+
+            # check files touched by PR and see if the target directory for --copy-ec
+            # corresponds to the name of one of these files;
+            # if so we should copy the specified file(s) to the current working directory,
+            # since interpreting the last argument as target location is very unlikely incorrect in this case
```
ocaisa, line 739: "is very unlikely to be correct"

ocaisa (changes requested):
Looks good, just a typo, will take a deeper look while you fix that

test/framework/options.py
```diff
@@ -1065,18 +1069,136 @@ def check_copied_files():
...
+        remove_dir(test_working_dir)
+
+        # test with only one ec in the PR (final argument is taken as a filename)
+        test_ec = os.path.join(self.test_prefix, 'test.eb')
+        args = ['--copy-ec', '--from-pr', '11521', test_ec]
+        ec_pr11521 = "ExifTool-12.00-GCCcore-9.3.0.eb"
+        stdout = self.mocked_main(args)
+        regex = re.compile(r'.*/%s copied to %s' % (ec_pr11521, test_ec))
+        self.assertTrue(regex.search(stdout), "Pattern '%s' found in: %s" % (regex.pattern, stdout))
+        self.assertTrue(os.path.exists(test_ec))
```
ocaisa, line 1199: Check the contents as well:
```
self.assertTrue("name = 'ExifTool'" in read_file(test_ec))
```

ocaisa (changes requested):
Looks good to me, one minor request

migueldiascosta:
also lgtm, fwiw (and more importantly, it works! :) )

test/framework/options.py
ocaisa: @boegel I injected it so we can hit merge on this

boegel: :+1:

ocaisa (approved):

PR #3483, opened 2020-10-25
enhance copy_files: single file target, error on empty input list, verbose mode

boegel:
fleshed out from #3482

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2057,7 +2057,7 @@ def copy_files(paths, target_path, force_in_dry_run=False, target_single_file=Fa
     elif len(paths) == 1 and target_single_file:
         copy_file(paths[0], target_path)
         if verbose:
-            print_msg("%s copied to %s" % (os.path.basename(paths[0]), target_path), prefix=False)
+            print_msg("%s copied to %s" % (paths[0], target_path), prefix=False)
 
     elif paths:
         # check target path: if it exists it should be a directory; if it doesn't exist, we create it
@@ -2076,7 +2076,7 @@ def copy_files(paths, target_path, force_in_dry_run=False, target_single_file=Fa
             print_msg("%d file(s) copied to %s" % (len(paths), target_path), prefix=False)
 
     elif not allow_empty:
-        raise EasyBuildError("One of more files to copy should be specified!")
+        raise EasyBuildError("One or more files to copy should be specified!")
 
 
 def copy_dir(path, target_path, force_in_dry_run=False, dirs_exist_ok=False, **kwargs):
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2034,26 +2034,49 @@ def copy_file(path, target_path, force_in_dry_run=False):
             raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
 
 
-def copy_files(paths, target_dir, force_in_dry_run=False):
+def copy_files(paths, target_path, force_in_dry_run=False, target_single_file=False, allow_empty=True, verbose=False):
     """
-    Copy list of files to specified target directory (which is created if it doesn't exist yet).
+    Copy list of files to specified target path.
+    Target directory is created if it doesn't exist yet.
 
-    :param filepaths: list of files to copy
-    :param target_dir: target directory to copy files into
+    :param paths: list of filepaths to copy
+    :param target_path: path to copy files to
     :param force_in_dry_run: force copying of files during dry run
+    :param target_single_file: if there's only a single file to copy, copy to a file at target path (not a directory)
+    :param allow_empty: allow empty list of paths to copy as input (if False: raise error on empty input list)
+    :param verbose: print a message to report copying of files
     """
+    # dry run: just report copying, don't actually copy
     if not force_in_dry_run and build_option('extended_dry_run'):
-        dry_run_msg("copied files %s to %s" % (paths, target_dir))
-    else:
-        if os.path.exists(target_dir):
-            if os.path.isdir(target_dir):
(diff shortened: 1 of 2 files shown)


easybuild/tools/filetools.py
```diff
@@ -2034,26 +2034,49 @@ def copy_file(path, target_path, force_in_dry_run=False):
...
+        if len(paths) == 1:
+            dry_run_msg("copied %s to %s" % (paths[0], target_path))
+        else:
+            dry_run_msg("copied %d files to %s" % (len(paths), target_path))
+
+    # special case: single file to copy and target_single_file is True => copy to file
+    elif len(paths) == 1 and target_single_file:
+        copy_file(paths[0], target_path)
+        if verbose:
+            print_msg("%s copied to %s" % (os.path.basename(paths[0]), target_path), prefix=False)
```
akesandgren, line 2060: Not sure that basename-ing is a good thing to do here. At least I usually want to know the full paths as often as possible...

easybuild/tools/filetools.py
```diff
@@ -2034,26 +2034,49 @@ def copy_file(path, target_path, force_in_dry_run=False):
...
-            if os.path.isdir(target_dir):
-                _log.info("Copying easyconfigs into existing directory %s...", target_dir)
+        if len(paths) == 1:
+            dry_run_msg("copied %s to %s" % (paths[0], target_path))
+        else:
+            dry_run_msg("copied %d files to %s" % (len(paths), target_path))
+
+    # special case: single file to copy and target_single_file is True => copy to file
+    elif len(paths) == 1 and target_single_file:
+        copy_file(paths[0], target_path)
```
akesandgren, line 2058: Should really verify that target_path is a directory before copying, or does copy_file handle that properly already?

test/framework/filetools.py
```diff
@@ -1551,6 +1551,89 @@ def test_copy_files(self):
         error_pattern = "/toy-0.0.eb exists but is not a directory"
         self.assertErrorRegex(EasyBuildError, error_pattern, ft.copy_files, [bzip2_ec], copied_toy_ec)
 
+        # by default copy_files allows empty input list, but if allow_empty=False then an error is raised
+        ft.copy_files([], self.test_prefix)
+        error_pattern = 'One of more files to copy should be specified!'
```
akesandgren, line 1556: Typo, s/of/or/

easybuild/tools/filetools.py
```diff
@@ -2034,26 +2034,49 @@ def copy_file(path, target_path, force_in_dry_run=False):
...
+
         for path in paths:
-            copy_file(path, target_dir)
+            copy_file(path, target_path)
+
+        if verbose:
+            print_msg("%d file(s) copied to %s" % (len(paths), target_path), prefix=False)
+
+    elif not allow_empty:
+        raise EasyBuildError("One of more files to copy should be specified!")
```
akesandgren, line 2079: Typo, s/of/or/

test/framework/filetools.py
```diff
@@ -1551,6 +1551,89 @@ def test_copy_files(self):
...
+        # test special case: copying a single file to a file target via target_single_file=True
+        target = os.path.join(self.test_prefix, 'target')
+        self.assertFalse(os.path.exists(target))
+        ft.copy_files([toy_ec], target, target_single_file=True)
+        self.assertTrue(os.path.exists(target))
+        self.assertTrue(os.path.isfile(target))
+        self.assertEqual(toy_ec_txt, ft.read_file(target))
+
+        ft.remove_file(target)
+
```
akesandgren, line 1568: Add test for target_single_file into path with a missing directory in the path?

test/framework/filetools.py
```diff
@@ -1551,6 +1551,89 @@ def test_copy_files(self):
...
+        # default behaviour is to copy single file list to target *directory*
+        self.assertFalse(os.path.exists(target))
+        ft.copy_files([toy_ec], target)
+        self.assertTrue(os.path.exists(target))
+        self.assertTrue(os.path.isdir(target))
+        copied_toy_ec = os.path.join(target, 'toy-0.0.eb')
```
akesandgren, line 1574: Shouldn't this really use basename(toy_ec) instead of hardcoding toy-0.0.eb?

And the same for all other places that does this?

test/framework/filetools.py
```diff
@@ -1551,6 +1551,89 @@ def test_copy_files(self):
...
+        self.assertEqual(stderr, '')
+        regex = re.compile(r"^toy-0\.0\.eb copied to .*/target")
+        self.assertTrue(regex.match(stdout), "Pattern '%s' should be found in: %s" % (regex.pattern, stdout))
+
+        ft.remove_file(target)
+
+        # check behaviour under -x: only printing, no actual copying
+        init_config(build_options={'extended_dry_run': True})
+        self.assertFalse(os.path.exists(target))
+        self.assertFalse(os.path.exists(os.path.join(target, 'test.eb')))
```
akesandgren, line 1608: Isn't this somewhat redundant, if target doesn't exist (as per the assertFalse above), then target/test.eb can't exist

akesandgren (changes requested):

easybuild/tools/filetools.py
boegel: `copy_file` takes care of that

test/framework/filetools.py
boegel: hardcoding in the tests is fine, it's often done to ensure we're not just re-implementing the function we're testing (and making the same mistake on both ends, causing the tests to pass while there's actually a problem)

boegel:
@akesandgren Please re-review, should be OK now.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3484, opened 2020-10-25
cache result of fetch_files_from_pr function

boegel:
mostly done to speed up the test suite (in the context of #3482)

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -33,6 +33,7 @@ import base64
 import copy
 import getpass
 import glob
+import functools
 import os
 import random
 import re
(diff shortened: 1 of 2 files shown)


easybuild/tools/github.py
```diff
@@ -377,16 +378,34 @@ def download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='master', account=GITHUB_
...
+            _log.debug("Using cached value for fetch_files_from_pr for PR #%s (%s)", pr, kwargs)
+            return cache[key]
+        else:
+            res = func(pr, *args, **kwargs)
+            cache[key] = res
+            return res
+
+    # expose clear/update methods of cache to wrapped function
+    cache_aware_func.clear_cache = cache.clear
+    cache_aware_func.update_cache = cache.update
```
migueldiascosta, line 403: just to check if I understand correctly, these two methods are not (yet) used anywhere, right?

easybuild/tools/github.py
```diff
@@ -377,16 +378,34 @@ def download_repo(repo=GITHUB_EASYCONFIGS_REPO, branch='master', account=GITHUB_
...
-    return fetch_files_from_pr(pr, path, github_user, github_repo=GITHUB_EASYBLOCKS_REPO)
+def pr_files_cache(func):
+    """
+    Decorator to cache result of fetch_files_from_pr.
+    """
+    cache = {}
 
+    @functools.wraps(func)
+    def cache_aware_func(pr, *args, **kwargs):
+        """Retrieve cached resul, or fetch files from PR & cache result."""
```
akesandgren, line 389: Typo s/resul/result/

akesandgren (changes requested):

easybuild/tools/github.py
boegel: Indeed, they're not, but they could come in useful at some point.

In the enhanced tests for example (which are still missing in this PR).

akesandgren (approved):
LGTM

easybuild/tools/github.py
boegel: fixed, tests are using `clear_cache` and `update_cache` now

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3485, opened 2020-10-25
add locate_files function to filetools module

boegel:


easybuild/tools/filetools.py
```diff
@@ -792,6 +792,57 @@ def find_easyconfigs(path, ignore_dirs=None):
     return files
 
 
+def locate_files(files, paths, ignore_subdirs=None):
+    """
+    Determine full path for list of files, in given list of paths (directories).
+    """
```
akesandgren, line 798: Since the code seem to be strictly targeted at easyconfig files, the name and description should say so.

easybuild/tools/filetools.py
```diff
@@ -792,6 +792,57 @@ def find_easyconfigs(path, ignore_dirs=None):
...
+    # find missing easyconfigs by walking paths in robot search path
+    for path in paths:
+        _log.debug("Looking for missing files (%d left) in %s..." % (len(files_to_find), path))
+
+        # try to load index for current path, or create one
+        path_index = load_index(path, ignore_dirs=ignore_subdirs)
+        if path_index is None or build_option('ignore_index'):
+            if os.path.exists(path):
+                _log.info("No index found for %s, creating one...", path)
+                path_index = create_index(path, ignore_dirs=ignore_subdirs)
```
akesandgren, line 815: Not sure what create_index does exactly, but if it is creating an index file, does it verify that it actually can (permissions) and if it doesn't then what?
For instance, our local ec repo tree doesn't have an index and is not writable by users, so what happens if a user runs this?

And the log message above, should not say "creating one" if it isn't allowed to...

easybuild/tools/filetools.py
```diff
@@ -792,6 +792,57 @@ def find_easyconfigs(path, ignore_dirs=None):
...
+    _log.debug("List of files to find: %s" % files_to_find)
+
+    # find missing easyconfigs by walking paths in robot search path
+    for path in paths:
+        _log.debug("Looking for missing files (%d left) in %s..." % (len(files_to_find), path))
+
+        # try to load index for current path, or create one
+        path_index = load_index(path, ignore_dirs=ignore_subdirs)
+        if path_index is None or build_option('ignore_index'):
+            if os.path.exists(path):
```
akesandgren, line 813: This would probably be better to have before the log_debug, i.e. first check if the path exists at all and if not just skip it completely, i.e. "if not os.path.exists(path): next" or however that is done in python

akesandgren (changes requested):

easybuild/tools/filetools.py
boegel: `create_index` just creates an index in memory, it doesn't dump it to file (that's done via `dump_index`), so as long as there's read access, that's enough

easybuild/tools/filetools.py
boegel: Nothing in this function is specific to easyconfigs. I've renamed the `ec_file` local variable to `filepath`, maybe that made you reach that conclusion?

boegel:
@akesandgren Tackled or replied to remarks, please re-review

easybuild/tools/filetools.py
boegel: Very good point, fixed in bb18de4

easybuild/tools/filetools.py
akesandgren: ok

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3486, opened 2020-10-25
don't change directory in download_repo function

### LINKED ISSUE #3283: using --include-easyblocks-from-pr together with a local easyconfig file fails
Example:

```
$ ls ABAQUS-2019.eb
ABAQUS-2019.eb

$ eb ABAQUS-2019.eb --installpath /tmp/$USER --include-easyblocks-from-pr 2034
...
ERROR: Can't find path /tmp/eb-d2fKqd/tmpA9Jiv_/easybuilders/ABAQUS-2019.eb
```

Using absolute path to easyconfig instead via `eb $PWD/AABQUS-2019.eb ...` does work...

boegel:
fleshed out from #3482

The `change_dir` done by the `download_repo` function is a nasty side effect, and not actually needed at all for the implementation of it. In addition, none of the functions that call out to `download_repo` rely on this side effect, so it's a change that's safe to make as a bug fix.

fixes #3283

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -50,7 +50,7 @@ from easybuild.framework.easyconfig.easyconfig import process_easyconfig
 from easybuild.framework.easyconfig.parser import EasyConfigParser
 from easybuild.tools.build_log import EasyBuildError, print_msg, print_warning
 from easybuild.tools.config import build_option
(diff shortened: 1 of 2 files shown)


migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3487, opened 2020-10-25
take source_urls, checksums, patches into account when extension source is specified via 'sources'

### LINKED ISSUE #3462: Checksums ignored for `sources` of `exts_list`
When an extension uses the `sources` key the checksum is ignored.

See https://github.com/easybuilders/easybuild-framework/blob/8291a1b97f18db7c83df189e23e251ce5f19efae/easybuild/framework/easyblock.py#L551-L566

A bit below it can be seen how checksums are checked when not using `sources`

### LINKED ISSUE #3478: source_urls aren't used for extensions using the source field
Trying to use an extension where controlling the filename was desired:
``` 
    ('batchspawner', '1.0.0', {
        'source_urls': ['https://github.com/jupyterhub/batchspawner/archive/'],
        'sources': [{'download_filename': 'v%(version)s.tar.gz', 'filename': SOURCE_TAR_GZ}],
        'checksums': ['861648f8b39d7c564582e0b705325008461487a9bc656c6b96f85230127a3702'],
    }),
```
turns out the `source_urls` aren't considered (no attempts to download). Only works when added directly into the dictionary, e.g:
``` 
    ('batchspawner', '1.0.0', {
        'sources': [{'download_filename': 'v%(version)s.tar.gz', 'filename': SOURCE_TAR_GZ, 'source_urls': ['https://github.com/jupyterhub/batchspawner/archive/']}],
        'checksums': ['861648f8b39d7c564582e0b705325008461487a9bc656c6b96f85230127a3702'],
    }),
```

boegel:
This aspect was overlooked in #3294...

The diff may look scary, but it's mostly moving part "up" from underneath the `else` which only happens the `source_tmpl` case for extensions. The changes are well covered by the enhanced test.

fixes #3462, fixes #3478

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -539,6 +539,7 @@ class EasyBlock(object):
                     # resolve templates in extension options
                     ext_options = resolve_template(ext_options, template_values)
 
+                    source_urls = ext_options.get('source_urls', [])
                     checksums = ext_options.get('checksums', [])
 
                     # use default template for name of source file if none is specified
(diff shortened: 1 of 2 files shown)


Micket (approved):
lgtm, probably, diff was hard to read

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3491, opened 2020-10-29
also ignore vsc.* imports coming from from pkg_resources/__init__.py (setuptools) in fake vsc namespace

boegel:
Fix for problem that pops up when installing EasyBuild with `eb` on our RHEL8 systems (where we have `vsc-base` & co installed via RPMs). Apparently the more recent `setuptools` we have there triggers an `import vsc` while scanning for namespace packages or something along those lines...

```
ERROR: Detected import from 'vsc' namespace in /usr/lib/python3.6/site-packages/pkg_resources/__init__.py (line 2105)
vsc-base & vsc-install were ingested into the EasyBuild framework in EasyBuild v4.0
```


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3495, opened 2020-11-02
also inject -rpath options for all entries in $LIBRARY_PATH in RPATH wrappers

boegel:
This is required to make a TensorFlow installation work when using `eb --rpath --filter-env-vars=LD_LIBRARY_PATH` (at least in combination with https://github.com/easybuilders/easybuild-easyblocks/pull/2218)

### DIFF
--- a/easybuild/scripts/rpath_args.py
+++ b/easybuild/scripts/rpath_args.py
@@ -81,7 +81,6 @@ while idx < len(args):
             add_rpath_args = False
         cmd_args.append(arg)
 
-    # FIXME: also consider $LIBRARY_PATH?
     # FIXME: support to hard inject additional library paths?
     # FIXME: support to specify list of path prefixes that should not be RPATH'ed into account?
     # FIXME skip paths in /tmp, build dir, etc.?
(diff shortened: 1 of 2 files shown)


easybuild/scripts/rpath_args.py
```diff
@@ -119,6 +118,14 @@
 
     idx += 1
 
+# also inject -rpath options for all entries in $LIBRARY_PATH,
+# unless they are there already
+for lib_path in os.getenv('LIBRARY_PATH', '').split(os.pathsep):
+    if lib_path and (rpath_filter is None or not rpath_filter.match(lib_path)):
```
ocaisa, line 124: How come you don't do the absolute path checking here?

boegel: Well, I could I guess, but I'm assuming only absolute paths are listed in `$LIBRARY_PATH`, which seems like a fair assumption?

boegel: Actually, consistent is better, so fixed in c35647f

ocaisa (approved):

PR #3497, opened 2020-11-05
add support for %(module_name)s template value

boegel:


ocaisa (approved):

easybuild/framework/easyconfig/templates.py
```diff
@@ -47,6 +47,7 @@
 # derived from easyconfig, but not from ._config directly
 TEMPLATE_NAMES_EASYCONFIG = [
     ('arch', "System architecture (e.g. x86_64, aarch64, ppc64le, ...)"),
+    ('mod_name', "Module name"),
```
ocaisa, line 50: I'm on the fence but why not be fully explicit here and use `module_name`, most of the others are pretty explicit

ocaisa (changes requested):

easybuild/framework/easyconfig/templates.py
boegel: OK, makes sense.

ocaisa (approved):

PR #3505, opened 2020-11-25
Add Support for BLIS+libFLAME toolchain

SebastianAchilles:
This pull request adds support for a toolchain using GNU with OpenMPI, BLIS, libFLAME, ScaLAPACK and FFTW.

I am not sure if the name `gobff` is the best. I tried to be in line with the other naming of toolchains, like `goblf`.

### DIFF
--- a/easybuild/toolchains/linalg/blis.py
+++ b/easybuild/toolchains/linalg/blis.py
@@ -51,8 +51,7 @@ class Blis(LinAlg):
         # This assumes that AMD's BLIS has ver > 2.1 and vanilla BLIS < 2.1
 
         found_version = self.get_software_version(self.BLAS_MODULE_NAME)[0]
-        ver = LooseVersion(found_version)
-        if ver > LooseVersion('2.1'):
+        if LooseVersion(found_version) > LooseVersion('2.1'):
             self.BLAS_LIB_MT = ['blis-mt']
 
         super(Blis, self)._set_blas_variables()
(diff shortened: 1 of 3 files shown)


easybuild/toolchains/linalg/blis.py
```diff
@@ -42,3 +44,15 @@ class Blis(LinAlg):
...
     BLAS_FAMILY = TC_CONSTANT_BLIS
+
+    def _set_blas_variables(self):
+        """AMD's fork with version number > 2.1 names the MT library blis-mt, while vanilla BLIS doesn't."""
+
+        # This assumes that AMD's BLIS has ver > 2.1 and vanilla BLIS < 2.1
+
+        found_version = self.get_software_version(self.BLAS_MODULE_NAME)[0]
+        ver = LooseVersion(found_version)
+        if ver > LooseVersion('2.1'):
```
akesandgren, line 55: no need to create a variable if it's only used once, so reduce to
if LooseVersion(found_version) > LooseVersion('2.1'):

akesandgren (changes requested):

easybuild/toolchains/linalg/blis.py
SebastianAchilles: Thanks for the comment! I changed it.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @SebastianAchilles!

akesandgren:
Now we just need some easyconfigs to go with it... or are they already there?

SebastianAchilles:
Yes, I create a PR for the easyconfigs:
https://github.com/easybuilders/easybuild-easyconfigs/pull/11761

PR #3506, opened 2020-11-27
Specifying False for a version removes the dep

edmondac:
Useful for architecture-specific deps, e.g.:

```
dependencies = [
    ('VTK', {'arch=x86_64': '8.2.0', 'arch=POWER': False}, local_python_versionsuffix),
]
```

So in this example, on POWER don't include a VTK dependency at all

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -702,10 +702,11 @@ class EasyConfig(object):
             # parse dependency specifications
             # it's important that templating is still disabled at this stage!
             self.log.info("Parsing dependency specifications...")
+
             def remove_false_versions(deps):
                 ret = []
                 for dep in deps:
-                    if isinstance(dep, dict) and dep['version'] == False:
+                    if isinstance(dep, dict) and dep['version'] is False:
                         continue
                     ret.append(dep)
                 return ret
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -702,10 +702,10 @@ class EasyConfig(object):
             # parse dependency specifications
             # it's important that templating is still disabled at this stage!
             self.log.info("Parsing dependency specifications...")
-            self['dependencies'] = [self._parse_dependency(dep) for dep in self['dependencies']]
-            self['hiddendependencies'] = [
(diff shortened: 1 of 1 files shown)


akesandgren:
Hmm.... wouldn't it be enough to write
```
dependencies = [
    ('VTK', {'arch=x86_64': '8.2.0'}, local_python_versionsuffix),
]
```
to get that effect?
I.e., default being False?

edmondac:
> Hmm.... wouldn't it be enough to write
> 
> ```
> dependencies = [
>     ('VTK', {'arch=x86_64': '8.2.0'}, local_python_versionsuffix),
> ]
> ```
> 
> to get that effect?
> I.e., default being False?
> 
> And then we could perhaps do,
> 
> ```
> dependencies = [
>     ('VTK', {'arch=default': '8.2.0', 'arch=POWER': False}, local_python_versionsuffix),
> ]
> ```
> 
> To only turn it off for POWER, or something, replace "default" with something better...

I do agree with that, but those changes would be in a different part of the codebase, so I'd prefer not to do it in this PR if possible. The syntax for arch-specific deps exists already - I'm just adding support for specifying version=False (for any dep) so I can say "this arch doesn't want this dep at all".

akesandgren:
Yes, I was just throwing out ideas...

Flamefire:
I do like the `arch=default` idea however I'm strongly against the "implicit default" idea for the version: This will lead to skipping the dependency for an arch not yet added. So IMO this is good as-is and should be followed-up by the `arch=default` which allows the same but being explicit: `{'arch=x86_64': '8.2.0', 'arch=default': False}`

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -702,10 +702,18 @@ def parse(self):
...
-                self._parse_dependency(dep, hidden=True) for dep in self['hiddendependencies']
-            ]
+
+            def remove_false_versions(deps):
+                ret = []
+                for dep in deps:
+                    if isinstance(dep, dict) and dep['version'] is False:
+                        continue
+                    ret.append(dep)
+                return ret
```
Flamefire, line 712: How about a (faster) list comprehension:
```suggestion
                return [dep for dep in deps if not (isinstance(dep, dict) and dep['version'] is False)]
```

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @edmondac!

PR #3509, opened 2020-11-30
clarify input format for --cuda-compute-capabilities

### LINKED ISSUE #3500: --cuda-compute-capabilites help should show desired format
Since there are quite a number of ways that CUDA capabilities are defined (e.g. `sm_75`, `75` and `7.5` define the same level), it would be nice if the help message for `--cuda-compute-capabilites` mentioned the desired format.

boegel:
fixes #3500


bartoldeman (approved):
LGTM

PR #3511, opened 2020-12-01
tweak test_easystack_basic to get better error reporting if it fails

boegel:


migueldiascosta:
@boegel this will still not show the actual error, maybe first check that `err` is empty and print it if it is not?

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3515, opened 2020-12-01
fix error when --easystack is used without having PyYAML installed

### LINKED ISSUE #3514: easystack tests in test/framework/option.py fail when PyYAML is not installed
with python3, there is a proper error message mentioning the requirement, but with python2 the error is `TypeError: unbound method error() must be called with EasyStackParser instance as first argument (got str instance instead)`

this will likely also happen outside of the tests

boegel:
This fixes the following ugly crash when trying to use `--easystack` without having PyYAML installed (which is a problem I introduced with the commits I added to #3479 before merging it), and also improves the `@only_if_module_is_available` decorator a bit (which was the underlying cause of the trouble).

closes #3514

### DIFF
--- a/easybuild/framework/easystack.py
+++ b/easybuild/framework/easystack.py
@@ -86,7 +86,6 @@ class SoftwareSpecs(object):
 class EasyStackParser(object):
     """Parser for easystack files (in YAML syntax)."""
 
-    @only_if_module_is_available('yaml', pkgname='PyYAML')
     @staticmethod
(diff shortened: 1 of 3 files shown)


migueldiascosta:
confirm that it fixes #3514 (the `yaml` requirement is now clear with both python 2 and 3)

migueldiascosta (dismissed):
lgtm

bartoldeman (approved):
lgtm

PR #3518, opened 2020-12-04
fix GitHub Actions workflow for test suite: run outside of repo checkout + also test bootstrap script with Python 3.9

boegel:
This probably needs to fix from #3517 for the tests to pass...


bartoldeman (approved):
lgtm

boegel:
I've updated this with the fix from #3517, tests should pass now...

This packaging problem (fixed in #3517) was actually caught in Travis (which is still running the tests with Python 2.6), but not in GitHub Actions, at least not until I made the change I'm making here...

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3521, opened 2020-12-07
use --dry-run in test_easystack_basic to avoid downloading of binutils sources

boegel:
This is mainly to avoid downloading of binutils sources in the test, which is sometimes failing due to problems with the GNU mirrors:

```
Unexpected error occurred when trying to download https://ftpmirror.gnu.org/gnu/binutils/binutils-2.25.tar.gz 
```

### DIFF
--- a/easybuild/framework/easystack.py
+++ b/easybuild/framework/easystack.py
@@ -225,7 +225,7 @@ def parse_easystack(filepath):
 
     general_options = easystack.get_general_options()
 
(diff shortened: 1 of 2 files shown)


ocaisa (dismissed):

ocaisa:
@boegel The tests failed due to the order of the easyconfigs in the printed output, just add a sort on the `easyconfig_names
`

ocaisa (approved):

PR #3524, opened 2020-12-09
Add option to skip the unit tests (test_step)

Flamefire:
Add --skip-test-step to skip the test_step

### DIFF
--- a/test/framework/easyconfigs/test_ecs/t/toy/toy-0.0-test.eb
+++ b/test/framework/easyconfigs/test_ecs/t/toy/toy-0.0-test.eb
@@ -1,0 +1,35 @@
+name = 'toy'
+version = '0.0'
+versionsuffix = '-test'
+
+homepage = 'https://easybuilders.github.io/easybuild'
+description = "Toy C program, 100% toy."
+
+toolchain = SYSTEM
+
+sources = [SOURCE_TAR_GZ]
+checksums = [[
+    'be662daa971a640e40be5c804d9d7d10',  # default (MD5)
+    '44332000aa33b99ad1e00cbd1a7da769220d74647060a10e807b916d73ea27bc',  # default (SHA256)
+    ('adler32', '0x998410035'),
+    ('crc32', '0x1553842328'),
+    ('md5', 'be662daa971a640e40be5c804d9d7d10'),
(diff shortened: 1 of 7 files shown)


test/framework/easyconfigs/test_ecs/t/toy/toy-0.0-test.eb
```diff
@@ -0,0 +1,35 @@
...
+    'toy-0.0_fix-silly-typo-in-printf-statement.patch',
+    ('toy-extra.txt', 'toy-0.0'),
+]
+
+sanity_check_paths = {
+    'files': [('bin/yot', 'bin/toy')],
+    'dirs': ['bin'],
+}
+
+runtest = "make_test dummy_cmd" # Provide some value which is unique enough to be checked for
```
ocaisa, line 30: Updating the test suite for this file is going to be a bit painful. Maybe just add this line to an existing test eb and use that? Of course this could also have negative consequences...

ocaisa: Maybe not, I see you only had two failures

Flamefire: What do you mean by updating the test suite? What else (besides this PR) needs to be done?

I haven't found any other test_ec with a `runtest` key and adding to an existing might be problematic as some of them are actually installed so they would choke on the "faulty" command.

ocaisa: One of your failing tests is
```
FAIL: test_index_functions (test.framework.filetools.FileToolsTest)
```
which does some kind of counting test. When you add new files to the test suite, these kinds of tests need to be updated (even though they are not related). Your case may not be so bad, but I remember this being tedious for me in the past.

ocaisa: The other failing test is a required update to https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/docs.py#L248

Flamefire: I see, thanks!

ocaisa (approved):

boegel:
I'll squeeze this in for EasyBuild v4.3.2...

PR #3535, opened 2020-12-23
add support for --accept-eula configuration option + 'accept_eula' easyconfig parameter

boegel:
This will come in useful for software like [AOCC](https://github.com/easybuilders/easybuild-easyconfigs/pull/11868) and Intel's oneAPI that require accepting an End User License Agreement (EULA) on installation...

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1636,6 +1636,28 @@ class EasyBlock(object):
         change_dir(self.start_dir)
         self.log.debug("Changed to real build directory %s (start_dir)", self.start_dir)
 
+    def check_accepted_eula(self, more_info=None):
+        """Check whether EULA for this software is accepted in current EasyBuild configuration."""
+
+        accepted_eulas = build_option('accept_eula') or []
+        if self.cfg['accept_eula'] or self.name in accepted_eulas:
+            self.log.info("EULA for %s is accepted", self.name)
+        else:
+            error_lines = [
+                "The End User License Argreement (EULA) for %(name)s is currently not accepted!",
+            ]
+            if more_info:
+                error_lines.append("(see %s for more information)" % more_info)
+
+            error_lines.extend([
+                "You should either:",
+                "- add --accept-eula-for=%(name)s to the 'eb' command;",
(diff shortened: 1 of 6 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1636,6 +1636,28 @@ def guess_start_dir(self):
         change_dir(self.start_dir)
         self.log.debug("Changed to real build directory %s (start_dir)", self.start_dir)
 
+    def check_accepted_eula(self, more_info=None):
+        """Check whether EULA for this software is accepted in current EasyBuild configuration."""
+
+        accepted_eulas = build_option('accept_eula') or []
+        if self.cfg['accept_eula'] or self.name in accepted_eulas:
```
akesandgren, line 1643: For ease of use, shouldn't the self.name in accepted_eulas be a case insensitive match?

boegel: I'm in favor of keeping it strict actually.

You don't want to accidentally accept a EULA either, imagine the consequences... ;)

akesandgren: ok.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

easybuild/framework/easyblock.py
```diff
@@ -1636,6 +1636,28 @@ def guess_start_dir(self):
...
+        else:
+            error_lines = [
+                "The End User License Argreement (EULA) for %(name)s is currently not accepted!",
+            ]
+            if more_info:
+                error_lines.append("(see %s for more information)" % more_info)
+
+            error_lines.extend([
+                "You should either:",
+                "- add --accept-eula-for=%(name)s to the 'eb' command;",
```
migueldiascosta, line 1654: either this should say `--accept-eula=` instead of `--accept-eula-for=`

boegel: Yes, you're right, I renamed the option to `--accept-eula` while working on it...

boegel: fixed in #3536

PR #3538, opened 2021-01-04
filter out duplicate paths in RPATH wrapper script

boegel:
Currently we're (potentially) including the same path multiple times via RPATH, either via both `-L` and `$LIBRARY_PATH`, or via `lib`/`lib64` symlinks.

These changes should fix that...

### DIFF
--- a/easybuild/scripts/rpath_args.py
+++ b/easybuild/scripts/rpath_args.py
@@ -38,7 +38,7 @@ import sys
 
 def is_new_existing_path(new_path, paths):
     """
-    Check whether specified path exsits and is a new path compared to provided list of paths.
+    Check whether specified path exists and is a new path compared to provided list of paths.
     """
 
     # assume path is new, until proven otherwise
--- a/easybuild/scripts/rpath_args.py
+++ b/easybuild/scripts/rpath_args.py
@@ -36,6 +36,26 @@ import re
 import sys
 
 
+def is_new_existing_path(new_path, paths):
+    """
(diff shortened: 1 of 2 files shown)


easybuild/scripts/rpath_args.py
```diff
@@ -36,6 +36,26 @@
 import sys
 
 
+def is_new_existing_path(new_path, paths):
+    """
+    Check whether specified path exsits and is a new path compared to provided list of paths.
```
bartoldeman, line 41: s/exsits/exists/

easybuild/scripts/rpath_args.py
```diff
@@ -36,6 +36,26 @@
...
+            if os.path.exists(path) and os.path.samefile(new_path, path):
+                res = False
+                break
+    else:
+        # path doesn't exist
+        res = False
+
+    return res
+
+
```
bartoldeman, line 58: I can't help thinking this could be shorter.
```
    if not os.path.exists(new_path):
        return False

    for path in paths:
        if os.path.exists(path) and os.path.samefile(new_path, path):
            return False

    return True
```
though it's a different coding style, so i'll leave this as a suggestion

bartoldeman (changes requested):

easybuild/scripts/rpath_args.py
boegel: We usually don't do inline `return`s, but have just a single `return` per function

bartoldeman (approved):
lgtm

PR #3541, opened 2021-01-06
Add toolchain deps to template resolution

Micket:
Adds the indirect toolchain dependencies to templates so that it works for system level CUDAcore packages.
Also includes CUDAcore -> cudaver. 

Cf. https://github.com/easybuilders/easybuild-easyconfigs/pull/11907

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -80,6 +80,7 @@ TEMPLATE_NAMES_EASYBLOCK_RUN_STEP = [
 TEMPLATE_SOFTWARE_VERSIONS = [
     # software name, prefix for *ver and *shortver
     ('CUDA', 'cuda'),
+    ('CUDAcore', 'cuda'),
     ('Java', 'java'),
     ('Perl', 'perl'),
     ('Python', 'py'),
@@ -241,6 +242,10 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
                     deps += config.get('builddependencies', [])
             else:
                 deps += config.get('builddependencies', [])
+
+        # Include all toolchain deps (e.g. CUDAcore template in fosscuda)
+        if config.toolchain.tcdeps is not None:
+            deps += config.toolchain.tcdeps
 
         for dep in deps:
             if isinstance(dep, dict):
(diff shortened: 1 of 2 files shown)


Micket:
This is frustrating. This particular test uses templates as part of the toolchain string, so we get a circular dependency of template -> toolchain -> template. As far as I can tell, we don't do this in any easyconfig, so I'm in favour of changing the test.

ping @boegel

easybuild/framework/easyconfig/templates.py
```diff
@@ -242,6 +243,10 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
             else:
                 deps += config.get('builddependencies', [])
 
+            # Include all toolchain deps (e.g. CUDAcore template in fosscuda)
+            if config.toolchain.tcdeps is not None:
+                deps += config.toolchain.tcdeps
```
boegel, line 248: Using `.extend` here (and above) would actually make more sense:

```python
deps.extend(config.toolchain.tcdeps)

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/framework/easyconfig/templates.py (hunk 1 above)
Comment (Micket) [reply easybuild/framework/easyconfig/templates.py:248]: Well, OK, I could do that change if you want, not that I think it makes any difference (in my own mental model += already means extend)

Comment (boegel) [reply easybuild/framework/easyconfig/templates.py]: With `+=` you're created a new list instance, with `.extend` you're updating the existing one.

In this context, it doesn't matter much, but it helps to have a consistent style w.r.t. updating lists.

Thanks for changing it!

Comment (Micket) [reply easybuild/framework/easyconfig/templates.py]: that's not correct actually. += modifies in place, and is identical to expand for lists as far as I know:
```
x = [1,2,3]
y = x
x += [4,5]
print(y)
```
specifically
```
a += b  # expands to:
a = a.__iadd__(b)
```

boegel: Hmm, ok, then it's mostly a matter of style...

lexming:
Successful test with using the templates from this PR in https://github.com/easybuilders/easybuild-easyconfigs/pull/12066#issuecomment-771309198

lexming (approved):
LGTM

lexming:
Going in, thanks @Micket !

PR #3543, opened 2021-01-08
detect 'SYSTEM' toolchain as special case in easystack files

boegel:
This adds support for more easily specifying software that is installed with a system toolchain in an easystack file, for example:

```yaml
software:
  EasyBuild:
    toolchains:
      SYSTEM:
        versions: [4.3.2]
```


bartoldeman (approved):
lgtm

PR #3544, opened 2021-01-10
don't clean up imported modules after verifying imports of included Python modules

### LINKED ISSUE #3542: Occasional failures with Python 2, triggered by cleanup of imported modules in verify_imports
When using EasyBuild with Python 2, I'm experiencing occasional failures with a traceback like this:
```
[...]
  File ".../easybuild/framework/easyconfig/easyconfig.py", line 2537, in det_full_module_name
    require_result=require_result)
  File ".../easybuild/framework/easyconfig/easyconfig.py", line 2509, in _det_module_name_with
    mod_name = mns_method(ec)
  File ".../easybuild/tools/module_naming_scheme/hierarchical_mns.py", line 89, in det_full_module_name
    return os.path.join(self.det_module_subdir(ec), self.det_short_module_name(ec))
AttributeError: 'NoneType' object has no attribute 'path'
```
The exact circumstances still seem to be unclear, but according to @Flamefire it may be related to https://stackoverflow.com/questions/29107470/imported-modules-becomes-none-when-replacing-current-module-in-sys-modules-using

boegel:
fixes #3542

### DIFF
--- a/easybuild/tools/include.py
+++ b/easybuild/tools/include.py
@@ -125,10 +125,16 @@ def set_up_eb_package(parent_path, eb_pkg_name, subpkgs=None, pkg_init_body=None
 
 def verify_imports(pymods, pypkg, from_path):
     """Verify that import of specified modules from specified package and expected location works."""
-    saved_modules = sys.modules.copy()
 
     for pymod in pymods:
         pymod_spec = '%s.%s' % (pypkg, pymod)
+
+        # force re-import if the specified modules was already imported;
+        # this is required to ensure that an easyblock that is included via --include-easyblocks-from-pr
+        # gets preference over one that is included via --include-easyblocks
+        if pymod_spec in sys.modules:
+            del sys.modules[pymod_spec]
+
         try:
             pymod = __import__(pymod_spec, fromlist=[pypkg])
         # different types of exceptions may be thrown, not only ImportErrors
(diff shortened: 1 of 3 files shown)


ocaisa:
@Flamefire This looks ok to me since the `clear()` and `update()` are only being done in the tests, so we shouldn't run into these runtime issues any more. I don't trust that I fully understand the issue though, can you review this PR?

Flamefire:
I have found https://stackoverflow.com/questions/43181440/what-does-del-sys-modulesmodule-actually-do which references the python docs:

> [`sys.modules`] is a dictionary that maps module names to modules which have already been loaded. This can be manipulated to force reloading of modules and other tricks. However, replacing the dictionary will not necessarily work as expected and deleting essential items from the dictionary may cause Python to fail.

So IMO even though it is test code this shouldn't be done as it seems to be unreliable (as shown by the issue: adding an extra import seems to have triggered the GC causing the issue to become a bug).
Why was the test code changed? Wasn't the old stuff (`del` and `reload`) enough?

I assume the `del` to be fine, so the module code change is good as far as I can tell, thanks!

boegel:
I added the `sys.modules.clear()` in the test at some point because it seemed necessary, but it actually isn't, the `del sys.modules[*]` statements indeed work just as well; fixed in 98578934b .

ocaisa (approved):
LGTM

Flamefire (approved):

PR #3545, opened 2021-01-10
enhance extract_cmd functio to use 'cp -a' for shell scripts (.sh)

boegel:


easybuild/tools/filetools.py
```diff
@@ -153,6 +153,8 @@
     '.iso': "7z x %(filepath)s",
     # tar.Z: using compress (LZW), but can be handled with gzip so use 'z'
     '.tar.z': "tar xzf %(filepath)s",
+    # shell scripts don't need unpacked, just copy there
```
ocaisa, line 156:
```suggestion
    # shell scripts don't need to be unpacked, just copy there
```

ocaisa (changes requested):
LGTM apart from comment

test/framework/filetools.py
```diff
@@ -95,6 +95,7 @@ def test_extract_cmd(self):
             ('test.txz', "unxz test.txz --stdout | tar x"),
             ('test.iso', "7z x test.iso"),
             ('test.tar.Z', "tar xzf test.tar.Z"),
+            ('test.foo.bar.sh', "cp -a test.foo.bar.sh ."),
```
ocaisa, line 98: Is there any chance that we would run such a command? Because it would fail...
```
[EESSI pilot 2020.12] $ cp -a modulerc.lua .
cp: ‘modulerc.lua’ and ‘./modulerc.lua’ are the same file
```

boegel: Yes, this is required for https://github.com/easybuilders/easybuild-easyconfigs/pull/11982.

The generated extract command actually uses a template that is filled in with the full path to the source file (in the `sourcepath`), and is executed in the build directory.

So the actual command would actually look like "`cp -a /sources/t/test/test.foo.bar.sh .`" after a "`cd %(builddir)s`"

ocaisa: Yeah, that's what I thought, just wanted to be sure

ocaisa (approved):
LGTM

PR #3546, opened 2021-01-11
add support for specifying custom name to use when checking accepted EULA

boegel:
With this trivial enhancement for #3535, we can use a custom name to check for when calling `check_accepted_eula`.

This is useful in the context of the Intel oneAPI tools for example, so we can accept the EULA just once by including `Intel-oneAPI` in `--accept-eula`, rather than having to include separate entries for the compilers, MPI, MKL, etc.


ocaisa (approved):
LGTM

PR #3551, opened 2021-01-22
add create_unused_dir function to create a directory which does not yet exist

Flamefire:
This will append _%d until a new folder could be created.
If the base folder name does not exists, it is used directly, so the simple, most common case is the cleanest

Note that I didn't use our `mkdir` function to reduce the possibility of a race condition where the folder gets created between the check and the `os.mkdir` call. This solution pushes the burden to Python, which at least reduces the window for such an issue.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1667,6 +1667,25 @@ def patch_perl_script_autoflush(path):
         write_file(path, newtxt)
 
 
+def set_gid_sticky_bits(path, set_gid=None, sticky=None, recursive=False):
+    """Set GID/sticky bits on specified path."""
+    if set_gid is None:
+        set_gid = build_option('set_gid_bit')
+    if sticky is None:
+        sticky = build_option('sticky_bit')
+
+    bits = 0
+    if set_gid:
+        bits |= stat.S_ISGID
+    if sticky:
+        bits |= stat.S_ISVTX
+    if bits:
+        try:
+            adjust_permissions(path, bits, add=True, relative=True, recursive=recursive, onlydirs=True)
+        except OSError as err:
+            raise EasyBuildError("Failed to set groud ID/sticky bit: %s", err)
+
+
 def mkdir(path, parents=False, set_gid=None, sticky=None):
     """
     Create a directory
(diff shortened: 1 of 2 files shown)


boegel:
@Flamefire What's the use case for this? Always good to know.

Also, I think it makes sense to re-use the existing `find_backup_name_candidate` function from `filetools` here, which already has logic to determine a new path (which was added in #2134)?

Flamefire:
This is for the build directories in meson, cmake, ... where for each build (iteration, extension, bundle component) a new, empty build dir is (usually) required. We had that on Slack where someone reported a failure and I wanted to get this function into framework instead of hacking it into meson only

As for `find_backup_name_candidate`: I think it makes sense for that function to use the proposed function here (in a follow-up I'd say) as that has 2 issues: First it might be slow, as a sleep is used instead of simply adding a number and second it is unsafe: A file/folder with that name might be created in the meantime. The function here is as safe as possible.

And finally: I'd like to have a simple folder name, e.g. `build` in the regular case, not something timestamped. This is especially important for the use case where ECs/builds rely on previous builds/iterations or for debugability: A bundle of 3 components will have component 1 in `build`, component 2 in `build_1`, component 3 in `build_2`, ...
It might also matter to keep the pathnames short to avoid failures due to long paths

easybuild/tools/filetools.py
```diff
@@ -2473,3 +2473,43 @@ def copy_framework_files(paths, target_dir):
...
+    Create a new folder in parent_folder using name as the name.
+    When a folder of that name already exists, '_0' is appended which is retried for increasing numbers until
+    an unused name was found
+    """
+    if not os.path.isabs(parent_folder):
+        parent_folder = os.path.abspath(parent_folder)
+
+    start_path = os.path.join(parent_folder, name)
+    number = None
+    while True:
```
boegel, line 2489: This `while True` makes me a bit nervous...

How about giving up after say 1000 attempts? So use `for _ in range(1000):` rather than `while True`?

test/framework/filetools.py
```diff
@@ -2888,6 +2888,24 @@ def test_locate_files(self):
         error_pattern = r"One or more files not found: 2\.txt \(search paths: \)"
         self.assertErrorRegex(EasyBuildError, error_pattern, ft.locate_files, ['2.txt'], [])
 
+    def test_create_unused_dir(self):
```
boegel, line 2891: missing docstring

test/framework/filetools.py
```diff
@@ -2888,6 +2888,24 @@ def test_locate_files(self):
...
+            self.assertEqual(path, os.path.join(self.test_prefix, 'folder_%s' % i))
+            self.assertTrue(os.path.exists(path))
+        # Not influenced by similar folder
+        path = ft.create_unused_dir(self.test_prefix, 'folder2')
+        self.assertEqual(path, os.path.join(self.test_prefix, 'folder2'))
+        self.assertTrue(os.path.exists(path))
+        for i in range(10):
+            path = ft.create_unused_dir(self.test_prefix, 'folder2')
+            self.assertEqual(path, os.path.join(self.test_prefix, 'folder2_%s' % i))
+            self.assertTrue(os.path.exists(path))
```
boegel, line 2907: Also check handling of errors when trying to create the directory, for example by calling `create_unused_dir` in a read-only directory?

Also, what happens if `path` point to an already existing *file*?

boegel (changes requested):

test/framework/filetools.py
Flamefire: Hm, I don't think any of the docstrings here is useful, but added for consistency

test/framework/filetools.py
Flamefire: Done.
> Also, what happens if path point to an already existing file?

No different than if it was a directory. Added a test for that

easybuild/tools/filetools.py
Flamefire: Ok, done

boegel (approved):

PR #3553, opened 2021-01-27
Avoid no-op changes to $LD_* environment variables in ModulesTool

Flamefire:
LD_ENV_VAR_KEYS variables are set and restore prior and after running any module command. Often they don't have any change leading to many messages like:
> INFO Environment variable LD_LIBRARY_PATH set to <long value> (previous value: '<same value>')

This just clutters the log without any benefit.
This change skips such messages and changes


boegel (approved):
lgtm

PR #3555, opened 2021-01-29
Performance improvements for easyconfig parsing and eb startup

Flamefire:
With cProfile I identified 2 hotspots:
 - TEMPLATE_SOFTWARE_VERSIONS resolving was inefficient duplicating lot's of work
 - the `Toolchain.get_module_version` raises `EasyBuildError` which is a particularly expensive exception due to stack tracing, logging, ... There is a seemingly prominent case where failure is expected. Hence I added a param so None is returned instead
The other function could be changed too, but it is not required or useful (yet).

My Python also complained about a leaked handle in `run` which might become a problem when it is called often (which we do) leading to anything from nothing over slowdown to crash.


boegel (approved):
Thoroughly reviewed, lgtm.

Also tested this a bit in the wild, didn't run into any surprises, so going in.

Thanks a lot @Flamefire, this probably took quite a bit of time to figure out...

PR #3557, opened 2021-02-02
avoid running expensive 'module use' command when using Lmod as modules tool, update $MODULEPATH directly instead

Flamefire:
I discovered an issue: The priority.
There are 2 problems:
- module use foo won't put foo to the front if any other path has a (big) priority
- We have a function `prepend_module_path` with an optional priority. Hence if we don't pass a priority then this function will not "prepend" the path

@boegel 
> First bullet is not a problem, that's just how module use works?
> Second one: yeah, if we really want to be sure, we should give it a (very) high priority, or at least detect that we need to do so.

It is a problem if I replace module use by setting MODULEPATH because that would yield a different behavior. I currently check if any priority is in use and fallback to module use instead
We use the priority in `load_fake_module`, but I guess we should not. The problem is: It doesn't tell what is intended. In `prepend_module_path` we could (hard, but possible) check current priorities and choose a higher one. Who knows that the 10000 we currently use isn't in use already or a user used 10001?
hence I'd deprecate the priority argument

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1428,7 +1428,20 @@ class Lmod(ModulesTool):
         if priority:
             self.run_module(['use', '--priority', str(priority), path])
         else:
-            self.run_module(['use', path])
+            # LMod allows modifying MODULEPATH directly. So do that to avoid the costly module use
+            # unless priorities are in use already
+            if os.environ.get('__LMOD_Priority_MODULEPATH'):
+                self.run_module(['use', path])
+            else:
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm

Flamefire:
Follow up issue for the unadressed points: https://github.com/easybuilders/easybuild-framework/issues/3631
BTW: As you added tests for `module unuse` I think we should add the same handling for the `unuse` function: If no priority in effect, simply remove it.

PR #3558, opened 2021-02-02
Allow use of alternate envvar(s) to $HOME for user modules

ocaisa:
The use of `$HOME` for user modules is quite restrictive, it makes it very difficult to share user space installations. Even if we use a symlink to an accessible space it is still problematic as the `root` directory is still placed into the module (with $HOME expanded).

The nice thing about `$HOME` is that it is guaranteed to exist and right now that is required by module file that we create. Nevertheless, if a site is willing to take on using another (configurable) variable instead (with a sensible default value, such as `$HOME`), it would make user space installations with a module hierarchy a lot more flexible.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -66,6 +66,7 @@ from easybuild.tools import config, run
 from easybuild.tools.build_details import get_build_stats
 from easybuild.tools.build_log import EasyBuildError, dry_run_msg, dry_run_warning, dry_run_set_dirs
 from easybuild.tools.build_log import print_error, print_msg, print_warning
+from easybuild.tools.config import DEFAULT_ENVVAR_USERS_MODULES
 from easybuild.tools.config import FORCE_DOWNLOAD_ALL, FORCE_DOWNLOAD_PATCHES, FORCE_DOWNLOAD_SOURCES
 from easybuild.tools.config import build_option, build_path, get_log_filename, get_repository, get_repositorypath
 from easybuild.tools.config import install_path, log_path, package_path, source_paths
@@ -1351,7 +1352,7 @@ class EasyBlock(object):
             # add user-specific module path; use statement will be guarded so no need to create the directories
             user_modpath = build_option('subdir_user_modules')
             if user_modpath:
-                user_envvars = build_option('envvars_user_modules') or ['HOME']
+                user_envvars = build_option('envvars_user_modules') or [DEFAULT_ENVVAR_USERS_MODULES]
                 user_modpath_exts = ActiveMNS().det_user_modpath_extensions(self.cfg)
                 self.log.debug("Including user module path extensions returned by naming scheme: %s", user_modpath_exts)
                 for user_envvar in user_envvars:
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1356,9 +1356,8 @@ class EasyBlock(object):
                 user_modpath_exts = ActiveMNS().det_user_modpath_extensions(self.cfg)
                 self.log.debug("Including user module path extensions returned by naming scheme: %s", user_modpath_exts)
                 for user_envvar in user_envvars:
-                    if not os.getenv(user_envvar):
-                        self.log.warning("Requested environment variable $%s as an additional branch for user "
-                                         "modules does not exist in current environment", user_envvar)
+                    self.log.debug("Requested environment variable $%s to host additional branch for modules",
+                                   user_envvar)
                     default_value = user_envvar + "_NOT_DEFINED"
                     txt += self.module_generator.use(user_modpath_exts,
                                                      prefix=self.module_generator.getenv_cmd(user_envvar,
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1359,9 +1359,8 @@ class EasyBlock(object):
                     self.log.debug("Requested environment variable $%s to host additional branch for modules",
                                    user_envvar)
                     default_value = user_envvar + "_NOT_DEFINED"
-                    txt += self.module_generator.use(user_modpath_exts,
-                                                     prefix=self.module_generator.getenv_cmd(user_envvar,
-                                                                                             default=default_value),
+                    getenv_txt = self.module_generator.getenv_cmd(user_envvar, default=default_value)
+                    txt += self.module_generator.use(user_modpath_exts, prefix=getenv_txt,
                                                      guarded=True, user_modpath=user_modpath)
         else:
             self.log.debug("Not including module path extensions, as specified.")
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1357,8 +1357,8 @@ class EasyBlock(object):
                 self.log.debug("Including user module path extensions returned by naming scheme: %s", user_modpath_exts)
                 for user_envvar in user_envvars:
                     if not os.getenv(user_envvar):
-                        raise EasyBuildError("Requested environment variable $%s as an additional branch for user "
-                                             "modules does not exist in current environment", user_envvar)
+                        self.log.warning("Requested environment variable $%s as an additional branch for user "
+                                         "modules does not exist in current environment", user_envvar)
                     txt += self.module_generator.use(user_modpath_exts,
                                                      prefix=self.module_generator.getenv_cmd(user_envvar), guarded=True,
                                                      user_modpath=user_modpath)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1360,8 +1360,8 @@ class EasyBlock(object):
                         self.log.warning("Requested environment variable $%s as an additional branch for user "
                                          "modules does not exist in current environment", user_envvar)
                     txt += self.module_generator.use(user_modpath_exts,
-                                                     prefix=self.module_generator.getenv_cmd(user_envvar), guarded=True,
-                                                     user_modpath=user_modpath)
+                                                     prefix=self.module_generator.getenv_cmd(user_envvar, safe=True),
+                                                     guarded=True, user_modpath=user_modpath)
         else:
             self.log.debug("Not including module path extensions, as specified.")
         return txt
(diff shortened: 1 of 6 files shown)


ocaisa:
@boegel This is also interesting for EESSI since it would allow sites to easily configure a stack built on top of the EESSI baseline

easybuild/framework/easyblock.py
```diff
@@ -1345,10 +1345,12 @@ def make_module_extend_modpath(self):
             # add user-specific module path; use statement will be guarded so no need to create the directories
             user_modpath = build_option('subdir_user_modules')
             if user_modpath:
+                user_envvar = build_option('envvar_user_modules')
                 user_modpath_exts = ActiveMNS().det_user_modpath_extensions(self.cfg)
                 self.log.debug("Including user module path extensions returned by naming scheme: %s", user_modpath_exts)
-                txt += self.module_generator.use(user_modpath_exts, prefix=self.module_generator.getenv_cmd('HOME'),
-                                                 guarded=True, user_modpath=user_modpath)
+                txt += self.module_generator.use(user_modpath_exts,
+                                                 prefix=self.module_generator.getenv_cmd(user_envvar), guarded=True,
```
ocaisa, line 1352: Ideally what I would like to have here is to not have an option at all for this but instead have a generic environment variable that the module checks for and if it doesn't exist then just default it to `$HOME`. In Lmod I can do that with
```
local foo=os.getenv("USER_MODULES") or os.getenv("FOO")
```
but this would require some changes to `self.module_generator.use`

ocaisa: However, the nice thing about using environment variables is that I could potentially have a list of them, allowing me to fork out into different paths, e.g., for example a user module built on top of system modules which are built on top of EESSI.

akesandgren:
Ha, nice. I have had vague thoughts in this direction a few weeks ago...

easybuild/framework/easyblock.py
akesandgren: From an EESSI user (site) perspective it would maybe make slightly more sense to have SITE_MODULES and USER_MODULES, and HOME. The two first could be a ":" separated list of paths as you say.

ocaisa: @akesandgren The way I've implemented it for now, that would mean giving a list on the command line (and order matters) `--envvars-user-modules=USER_MODULES,SITE_MODULES --subdir-user-modules=easybuild`

ocaisa: The `EasyBuildMNS` may need an update to support this though, right now it does not fiddle with `MODULEPATH`

ocaisa: @damianam Would this work for us at JSC?

easybuild/framework/easyblock.py
```diff
@@ -1345,10 +1345,16 @@ def make_module_extend_modpath(self):
...
             user_modpath = build_option('subdir_user_modules')
             if user_modpath:
+                user_envvars = build_option('envvars_user_modules') or ['HOME']
                 user_modpath_exts = ActiveMNS().det_user_modpath_extensions(self.cfg)
                 self.log.debug("Including user module path extensions returned by naming scheme: %s", user_modpath_exts)
-                txt += self.module_generator.use(user_modpath_exts, prefix=self.module_generator.getenv_cmd('HOME'),
-                                                 guarded=True, user_modpath=user_modpath)
+                for user_envvar in user_envvars:
+                    if not os.getenv(user_envvar):
+                        raise EasyBuildError("Requested environment variable %s as an additional branch for user"
```
boegel, line 1353: Use `$` when referring to environment modules, add missing space after `user`

```suggestion
                        raise EasyBuildError("Requested environment variable $%s as an additional branch for user "

File: easybuild/tools/options.py [hunk 3]
```diff
@@ -489,6 +489,9 @@ def config_options(self):
             'buildpath': ("Temporary build path", None, 'store', mk_full_default_path('buildpath')),
             'containerpath': ("Location where container recipe & image will be stored", None, 'store',
                               mk_full_default_path('containerpath')),
+            'envvars-user-modules': ("List of environment variables that hold the base paths for which user-specific "
+                                     "modules will be installed relative to", 'strlist', 'store',
+                                     ['HOME']),
```
boegel, line 494: You should use `DEFAULT_ENVVAR_USERS_MODULES` here?

test/framework/easyblock.py
```diff
@@ -301,6 +301,51 @@ def test_make_module_extend_modpath(self):
...
+        }
+        init_config(build_options=build_options)
+        eb = EasyBlock(EasyConfig(self.eb_file))
+        eb.installdir = config.install_path()
+
+        txt = eb.make_module_extend_modpath()
+        for envvar in list_of_envvars:
+            if get_module_syntax() == 'Tcl':
+                regexs = [r'^module use ".*/modules/funky/Compiler/pi/3.14/%s"$' % c for c in modclasses]
+                home = r'\$::env\(%s\)' % envvar
```
boegel, line 323: Don't use `home` here, rename this variable...

test/framework/easyblock.py
```diff
@@ -301,6 +301,51 @@ def test_make_module_extend_modpath(self):
...
+                regexs.extend([
+                    # extension for user modules is guarded
+                    r'if { \[ file isdirectory \[ file join %s \[ %s \] \] \] } {$' % (home, fj_usermodsdir),
+                    # no per-moduleclass extension for user modules
+                    r'^\s+module use \[ file join %s \[ %s \] \]$' % (home, fj_usermodsdir),
+                ])
+            elif get_module_syntax() == 'Lua':
+                regexs = [r'^prepend_path\("MODULEPATH", ".*/modules/funky/Compiler/pi/3.14/%s"\)$' % c
+                          for c in modclasses]
+                home = r'os.getenv\("%s"\)' % envvar
```
boegel, line 334: Same here, please rename this

easybuild/framework/easyblock.py
boegel: Actually, do we really need to be this restrictive?
Why does the environment variable need to be defined at this point?

It only matters when the module file is consumed, and whatever the value of the environment variable is here, it doesn't say much about the value it could have when the module is loaded.

test/framework/easyblock.py
```diff
@@ -301,6 +301,51 @@ def test_make_module_extend_modpath(self):
             regex = re.compile(regex, re.M)
             self.assertTrue(regex.search(txt), "Pattern '%s' found in: %s" % (regex.pattern, txt))
 
+        # Repeat this but using an alternate envvars (instead of $HOME)
+        list_of_envvars = ['TURKEY', 'HAM']
+        os.environ['TURKEY'] = "1"
+        os.environ['HAM'] = "1"
```
boegel, line 307: Use more sensible values and variable names here?

`1` doesn't make sense at all, make it `self.test_prefix` (since that's a path)?

boegel (changes requested):
@ocaisa It would be nice to have a more complete test for this, where the generated module file actually gets loaded, and the effect on the changes in the environment are checked.

That may be a significant additional effort though...

easybuild/framework/easyblock.py
ocaisa: No strictly not, I was trying to think of a way to hammer home that the environment variable must exist or the module file will be broken.

ocaisa: I wanted to guard checking the environment variable and that is really easy in Lua
```
os.getenv("ENVVAR") or ""
```
in fact, in our use case, if the variable doesn't exist the return value is `nil` and this is equivalent since we are calling this in a guarded statement that is using `pathJoin`.

The problem is with Tcl modules, in this case if the  envvar does not exist you get an error like:
```
alanc@~/EasyBuild_Git/easybuild-framework(alt-envvar-user-mod)$ module show test2
Lmod has detected the following error:  /home/alanc/EasyBuild_Git/EB_Devel/test2: (test2): can't read "::env(HOME2)": no such variable 
While processing the following module(s):
    Module fullname  Module Filename
    ---------------  ---------------
    test2            /home/alanc/EasyBuild_Git/EB_Devel/test2
```
This is because we are trying to reference the value in an array when we use `$env::("HOME2")`. I tried to put a guard around this but my Tcl-foo is not making it work

ocaisa:
@boegel Regarding having an end-to-end test for this, that is indeed additional effort and I don't really see the point. We currently check for the exact text we expect to find in the module file, the only additional thing that could be done is to make sure of the order.

easybuild/framework/easyblock.py
ocaisa: Ok, I found a Tcl syntax that can give the same behaviour as Lua:
```
[if { [info exists ::env(HOME)] } { concat $::env(HOME) } ]
```

easybuild/tools/module_generator.py
```diff
@@ -780,7 +780,7 @@ def getenv_cmd(self, envvar):
         """
         Return module-syntax specific code to get value of specific environment variable.
         """
-        return '$::env(%s)' % envvar
+        return '[if { [info exists ::env(%s)] } { concat $::env(%s) } ]' % (envvar, envvar)
```
ocaisa, line 783: This implies the module file no longer errors if the variable is not in the environment

ocaisa:
> @ocaisa It would be nice to have a more complete test for this, where the generated module file actually gets loaded, and the effect on the changes in the environment are checked.
> 
> That may be a significant additional effort though...

@boegel Ok, I found a relatively easy way to do this

ocaisa:
@akesandgren Can you take a look at this to see if it makes sense for you too? We kind of need this so that we can create additional installations shared among a *project* as opposed to (or as well as) just for a single *user*.

easybuild/framework/easyblock.py
```diff
@@ -1351,10 +1352,16 @@ def make_module_extend_modpath(self):
...
             user_modpath = build_option('subdir_user_modules')
             if user_modpath:
+                user_envvars = build_option('envvars_user_modules') or [DEFAULT_ENVVAR_USERS_MODULES]
                 user_modpath_exts = ActiveMNS().det_user_modpath_extensions(self.cfg)
                 self.log.debug("Including user module path extensions returned by naming scheme: %s", user_modpath_exts)
-                txt += self.module_generator.use(user_modpath_exts, prefix=self.module_generator.getenv_cmd('HOME'),
-                                                 guarded=True, user_modpath=user_modpath)
+                for user_envvar in user_envvars:
+                    if not os.getenv(user_envvar):
+                        self.log.warning("Requested environment variable $%s as an additional branch for user "
```
ocaisa, line 1360: Not even sure we even need this warning any more since I can now handle the case where the environment variable is not defined (for both Tcl and Lua modules)

boegel (approved):

PR #3559, opened 2021-02-02
avoid metadata greedy behaviour when probing for external module metadata

victorusu:
This PR attempts to fix the greedy behaviour from the metadata handler.
The greedy behaviour causes this type of issues (https://github.com/eth-cscs/production/pull/2106).

Currently, the metadata file parser populates the `PACKAGE_PREFIX` and `PACKAGE_VERSION` variables at the moment it reads the file. This might be ok for non Cray systems, due to the fact that the modules are may still be present and have the same values. For Cray systems, the value of the fields change with the change in the programming environment. In this case, one needs to let EB populate the values at the very end.


boegel:
@victorusu Can you look into the failing test?

boegel (approved):

PR #3560, opened 2021-02-02
avoid initializing Toolchain instance when taking into account toolchain dependencies for templates

boegel:
This was missing in #3541.

The test fails without the changes in #3541...


lexming:
As expected, the test fails if `cudaver` is not defined:

```
ERROR: test_templating_cuda_toolchain (__main__.EasyConfigTest)
Test templates via toolchain component, like setting %(cudaver)s with fosscuda toolchain.
  File "/theia/home/brussel/101/vsc10122/src/easybuild-framework/test/framework/easyconfig.py", line 1094, in test_templating_cuda_toolchain
    self.assertEqual(ec.template_values['cudaver'], '10.1.243')
KeyError: 'cudaver'

Ran 82 tests in 625.336s

FAILED (errors=1)
```

lexming (approved):
LGTM

PR #3565, opened 2021-02-04
Fix UTF-8 encoding errors when running EasyBuild with Python 3.0.x-3.6.x

Flamefire:
When the system encoding is not UTF-8 Python 3 will try to decode files as ASCII which fails
This adds a test to CI with LC_ALL=C to enforce this situation and test cases reproducing failures in read_file, write_file and apply_regex_substitutions as seen in the wild

Example: https://github.com/easybuilders/easybuild-easyblocks/issues/2279 the failing text is valid UTF-8 (likely)

It seems only Python 3 < 3.7 is affected because in 3.7 a "UTF-8" mode was introduced which opens files as UTF-8 by default: https://www.python.org/dev/peps/pep-0540/

Fixes https://github.com/easybuilders/easybuild-easyblocks/issues/2279
fixed https://github.com/easybuilders/easybuild-easyblocks/issues/2329

This change makes sure files are opened with utf-8 encoding in Python 3 (Python 2 doesn't really care). This applies to read_file/write_file and friends as well as the logger which also opens files and if no encoding is specified it will run into the same situation: It uses the default encoding which may be ASCII and then fail to log UTF-8 encoded strings. Example: Dumping an EC with an UTF-8 char in the description

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -169,6 +169,12 @@ jobs:
           PRINTED_MSG=$(egrep -v "${IGNORE_PATTERNS}" test_framework_suite.log | grep '\.\n*[A-Za-z]' || true)
           test "x$PRINTED_MSG" = "x" || (echo "ERROR: Found printed messages in output of test suite\n${PRINTED_MSG}" && exit 1)
 
+          # There may be encoding errors in Python 3 which are hidden when an UTF-8 encoding is set
+          # Hence run the tests again with LC_ALL=C
+          if [[ ${{matrix.python}} =~ "3." ]]; then
+            LC_ALL=C python -O -m test.framework.suite
+          fi
+
     - name: test bootstrap script
       run: |
           # (re)initialize environment for modules tool
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -9,6 +9,7 @@ jobs:
         python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
         modules_tool: [Lmod-7.8.22, Lmod-8.2.9, modules-tcl-1.147, modules-3.2.10, modules-4.1.4]
         module_syntax: [Lua, Tcl]
+        lc_all: [""]
         # exclude some configuration for non-Lmod modules tool:
         # - don't test with Lua module syntax (only supported in Lmod)
         # - exclude Python 3.x versions other than 3.6, to limit test configurations
@@ -51,6 +52,13 @@ jobs:
             python: 3.8
           - modules_tool: Lmod-7.8.22
             python: 3.9
+        # There may be encoding errors in Python 3 which are hidden when an UTF-8 encoding is set
+        # Hence run the tests (again) with LC_ALL=C and Python 3.6 (or any < 3.7)
+        include:
+          - python: 3.6
+            modules_tool: Lmod-8.2.9
+            module_syntax: Lua
+            lc_all: C
       fail-fast: false
     steps:
     - uses: actions/checkout@v2
@@ -129,6 +137,7 @@ jobs:
         EB_VERBOSE: 1
         EASYBUILD_MODULE_SYNTAX: ${{matrix.module_syntax}}
         TEST_EASYBUILD_MODULE_SYNTAX: ${{matrix.module_syntax}}
+        LC_ALL: ${{matrix.lc_all}}
       run: |
           # run tests *outside* of checked out easybuild-framework directory,
           # to ensure we're testing installed version (see previous step)
@@ -168,12 +177,6 @@ jobs:
           # '|| true' is needed to avoid that Travis stops the job on non-zero exit of grep (i.e. when there are no matches)
           PRINTED_MSG=$(egrep -v "${IGNORE_PATTERNS}" test_framework_suite.log | grep '\.\n*[A-Za-z]' || true)
           test "x$PRINTED_MSG" = "x" || (echo "ERROR: Found printed messages in output of test suite\n${PRINTED_MSG}" && exit 1)
-
-          # There may be encoding errors in Python 3 which are hidden when an UTF-8 encoding is set
-          # Hence run the tests again with LC_ALL=C
-          if [[ ${{matrix.python}} =~ "3." ]]; then
-            LC_ALL=C python -O -m test.framework.suite
-          fi
 
     - name: test bootstrap script
       run: |
--- a/easybuild/base/fancylogger.py
+++ b/easybuild/base/fancylogger.py
@@ -580,6 +580,8 @@ def logToFile(filename, enable=True, filehandler=None, name=None, max_bytes=MAX_
         'maxBytes': max_bytes,
         'backupCount': backup_count,
     }
+    if sys.version_info[0] >= 3:
+        handleropts['encoding'] = 'utf-8'
     # logging to a file is going to create the file later on, so let's try to be helpful and create the path if needed
     directory = os.path.dirname(filename)
     if not os.path.exists(directory):
--- a/easybuild/base/optcomplete.py
+++ b/easybuild/base/optcomplete.py
@@ -594,8 +594,8 @@ def autocomplete(parser, arg_completer=None, opt_completer=None, subcmd_complete
         if isinstance(debugfn, logging.Logger):
             debugfn.debug(txt)
         else:
-            with open(debugfn, 'a') as f:
-                f.write(txt)
+            with open(debugfn, 'a') as fh:
+                fh.write(txt)
 
     # Exit with error code (we do not let the caller continue on purpose, this
     # is a run for completions only.)
--- a/easybuild/base/optcomplete.py
+++ b/easybuild/base/optcomplete.py
@@ -594,9 +594,8 @@ def autocomplete(parser, arg_completer=None, opt_completer=None, subcmd_complete
         if isinstance(debugfn, logging.Logger):
             debugfn.debug(txt)
         else:
-            f = open(debugfn, 'a')
-            f.write(txt)
-            f.close()
+            with open(debugfn, 'a') as f:
+                f.write(txt)
 
     # Exit with error code (we do not let the caller continue on purpose, this
     # is a run for completions only.)
--- a/easybuild/scripts/fix_docs.py
+++ b/easybuild/scripts/fix_docs.py
@@ -57,13 +57,12 @@ for tmp in py_files:
         continue
     with open(tmp) as f:
         temp = "tmp_file.py"
-        out = open(temp, 'w')
-        for line in f:
-            if "@author" in line:
-                out.write(re.sub(r"@author: (.*)", r":author: \1", line))
-            elif "@param" in line:
-                out.write(re.sub(r"@param ([^:]*):", r":param \1:", line))
-            else:
-                out.write(line)
-        out.close()
+        with open(temp, 'w') as out:
+            for line in f:
+                if "@author" in line:
+                    out.write(re.sub(r"@author: (.*)", r":author: \1", line))
+                elif "@param" in line:
+                    out.write(re.sub(r"@param ([^:]*):", r":param \1:", line))
+                else:
+                    out.write(line)
         os.rename(temp, tmp)
--- a/easybuild/scripts/fix_docs.py
+++ b/easybuild/scripts/fix_docs.py
@@ -55,10 +55,10 @@ for tmp in py_files:
     # exclude self
     if os.path.basename(tmp) == os.path.basename(__file__):
         continue
-    with open(tmp) as f:
+    with open(tmp) as fh:
         temp = "tmp_file.py"
         with open(temp, 'w') as out:
-            for line in f:
+            for line in fh:
                 if "@author" in line:
                     out.write(re.sub(r"@author: (.*)", r":author: \1", line))
                 elif "@param" in line:
--- a/easybuild/scripts/mk_tmpl_easyblock_for.py
+++ b/easybuild/scripts/mk_tmpl_easyblock_for.py
@@ -225,8 +225,8 @@ try:
     dirpath = os.path.dirname(easyblock_path)
     if not os.path.exists(dirpath):
         os.makedirs(dirpath)
-    with open(easyblock_path, "w") as f:
-        f.write(txt)
+    with open(easyblock_path, "w") as fh:
+        fh.write(txt)
 except (IOError, OSError) as err:
     sys.stderr.write("ERROR! Writing template easyblock for %s to %s failed: %s" % (name, easyblock_path, err))
     sys.exit(1)
--- a/easybuild/scripts/mk_tmpl_easyblock_for.py
+++ b/easybuild/scripts/mk_tmpl_easyblock_for.py
@@ -225,9 +225,8 @@ try:
     dirpath = os.path.dirname(easyblock_path)
     if not os.path.exists(dirpath):
         os.makedirs(dirpath)
-    f = open(easyblock_path, "w")
-    f.write(txt)
-    f.close()
+    with open(easyblock_path, "w") as f:
+        f.write(txt)
 except (IOError, OSError) as err:
     sys.stderr.write("ERROR! Writing template easyblock for %s to %s failed: %s" % (name, easyblock_path, err))
     sys.exit(1)
--- a/easybuild/tools/configobj.py
+++ b/easybuild/tools/configobj.py
@@ -1213,9 +1213,8 @@ class ConfigObj(Section):
         if isinstance(infile, string_type):
             self.filename = infile
             if os.path.isfile(infile):
-                h = open(infile, 'r')
-                infile = h.read() or []
-                h.close()
+                with open(infile, 'r') as h:
+                    infile = h.read() or []
             elif self.file_error:
                 # raise an error if the file doesn't exist
                 raise IOError('Config file not found: "%s".' % self.filename)
@@ -1224,9 +1223,8 @@ class ConfigObj(Section):
                 if self.create_empty:
                     # this is a good test that the filename specified
                     # isn't impossible - like on a non-existent device
-                    h = open(infile, 'w')
-                    h.write('')
-                    h.close()
+                    with open(infile, 'w') as h:
+                        h.write('')
                 infile = []
 
         elif isinstance(infile, (list, tuple)):
@@ -2052,9 +2050,8 @@ class ConfigObj(Section):
         if outfile is not None:
             outfile.write(output)
         else:
-            h = open(self.filename, 'wb')
-            h.write(output)
-            h.close()
+            with open(self.filename, 'wb') as h:
+                h.write(output)
 
     def validate(self, validator, preserve_errors=False, copy=False,
                  section=None):
--- a/easybuild/tools/configobj.py
+++ b/easybuild/tools/configobj.py
@@ -1213,8 +1213,8 @@ class ConfigObj(Section):
         if isinstance(infile, string_type):
             self.filename = infile
             if os.path.isfile(infile):
-                with open(infile, 'r') as h:
-                    infile = h.read() or []
+                with open(infile, 'r') as fh:
+                    infile = fh.read() or []
             elif self.file_error:
                 # raise an error if the file doesn't exist
                 raise IOError('Config file not found: "%s".' % self.filename)
@@ -1223,8 +1223,8 @@ class ConfigObj(Section):
                 if self.create_empty:
                     # this is a good test that the filename specified
                     # isn't impossible - like on a non-existent device
-                    with open(infile, 'w') as h:
-                        h.write('')
+                    with open(infile, 'w') as fh:
+                        fh.write('')
                 infile = []
 
         elif isinstance(infile, (list, tuple)):
@@ -2050,8 +2050,8 @@ class ConfigObj(Section):
         if outfile is not None:
             outfile.write(output)
         else:
-            with open(self.filename, 'wb') as h:
-                h.write(output)
+            with open(self.filename, 'wb') as fh:
+                fh.write(output)
 
     def validate(self, validator, preserve_errors=False, copy=False,
                  section=None):
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -40,7 +40,6 @@ Set of file tools.
 """
 import datetime
 import difflib
-import fileinput
 import glob
 import hashlib
 import imp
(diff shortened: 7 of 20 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -169,6 +169,12 @@ jobs:
           PRINTED_MSG=$(egrep -v "${IGNORE_PATTERNS}" test_framework_suite.log | grep '\.\n*[A-Za-z]' || true)
           test "x$PRINTED_MSG" = "x" || (echo "ERROR: Found printed messages in output of test suite\n${PRINTED_MSG}" && exit 1)
 
+          # There may be encoding errors in Python 3 which are hidden when an UTF-8 encoding is set
+          # Hence run the tests again with LC_ALL=C
+          if [[ ${{matrix.python}} =~ "3." ]]; then
+            LC_ALL=C python -O -m test.framework.suite
```
boegel, line 175: Oh may, there's a bit extreme... The test suite already runs quite long.

How about just adding a test configuration with Python 3.x + `$LC_ALL` set to `C`?

easybuild/base/fancylogger.py
```diff
@@ -580,6 +580,8 @@ def logToFile(filename, enable=True, filehandler=None, name=None, max_bytes=MAX_
         'maxBytes': max_bytes,
         'backupCount': backup_count,
     }
+    if sys.version_info[0] >= 3:
+        handleropts['encoding'] = 'utf-8'
```
boegel, line 584: This deserves a comment to explain why this is needed?

boegel (changes requested):

.github/workflows/unit_tests.yml
Flamefire: Yeah, it's another 16 minutes on top of the 18 minutes...
IMO we need this for Python 3.5 and any Python >= 3.6. At least till we find another issue... Hence this is the most complete and the safe bet.
But running tests with LC_ALL=C is also potentially problematic. E.g. I found issues where the tests relied on an English locale which gets hidden by LC_ALL=C.
But yes, I guess it makes sense to refine the test matrix. But then we should talk about it in more general. E.g.

> exclude Python 3.x versions other than 3.6, to limit test configurations

This is not what's happening now and the `exclude` is overly complicated. `include` is much better here. So maybe we should first define which test configurations to test exactly. This should include Python 3.5 and any Python >= 3.6 run twice for e.g. LC_ALL and a non-English UTF-8 locale

easybuild/base/fancylogger.py
Flamefire: Erm, to make sure sure logs are utf-8 encoded? I don't really know what else to explain which isn't already stated by `encoding = utf-8`

All the gory details about default encoding in Python 3, 3.6, 3.7 would IMO be overkill here.

boegel: But *why* do logs need to be utf-8 encoded all of a sudden :)

Flamefire: Because we log potentially UTF-8 encoded strings and if the default doesn't happen to be utf-8 we get encoding errors. This change makes us not rely on the default to be UTF-8 (e.g. on Python 3.6) but choose it explicitly. It accidentally worked prior to the change because we didn't test that case (i.e. no LC_ALL=C), so this is a bugfix
I'll add that to the PR description to answer why we need the change, however I'd avoid adding a more or less trivial comment to the code.

.github/workflows/unit_tests.yml
boegel: We currently have 3 "dimensions":

* Python version (currently 6)
* modules tool (5)
* module syntax (2)

So in theory we have 60 combinations, but some don't make sense (non-Lmod modules with Lua syntax), and we already strip out some to limit the number of test configurations (we only test Lmod 8 with all Python 3 versions), which leads to 22 test configurations currently (+ 1 code style check).

Adding environment variables like `$LC_ALL` in the mix blows things up, especially if want want to consider multiple values for `$LC_ALL` (default, `C`, non-English UTF-8).

I would be fine, for now, with just adding `LC_ALL=C` with Python 3.6, Lmod 8 and Lua, since that would be enough to catch the problem being fixed, and it's the most common setup in which EasyBuild is being used (the exact Python 3.x version doesn't matter much).

Doing that would indeed require a rewrite of the test matrix, which I'd like to keep for a separate PR (we really want this one merged for the next EasyBuild release, which I want to push out ASAP now).

So how about we make the `if` test here a stricter so that we only also check with `LC_ALL=C` in *one* test configuration? Then that configuration would be "expensive", but it's only a single one...

Ideally we also change the order of the values in the test matrix being tested, so the long-running test configuration is picked up first . Not all test configurations start in parallel, so the expensive one(s) should start first so they can overlap with others and not block the outcome of the whole test suite.

boegel: Another option would be to just add a specific test for using `read_file` & co when `LC_ALL=C` is set, but that may not be entirely trivial (you probably have to run the tests in a Python subshell to actually reproduce the problem being fixed here).

easybuild/base/optcomplete.py
```diff
@@ -594,9 +594,8 @@ def autocomplete(parser, arg_completer=None, opt_completer=None, subcmd_complete
         if isinstance(debugfn, logging.Logger):
             debugfn.debug(txt)
         else:
-            f = open(debugfn, 'a')
-            f.write(txt)
-            f.close()
+            with open(debugfn, 'a') as f:
```
boegel, line 597: @Flamefire We shouldn't use single-letter variables outside of list comprehensions, so please use `fh` (file handle) rather than `f`.

Bonus points if you can make `flake8` trip over this first (by tweaking `setup.cfg` so the `linting` workflow we already have will fail).

easybuild/scripts/fix_docs.py
```diff
@@ -57,13 +57,12 @@
...
-        for line in f:
-            if "@author" in line:
-                out.write(re.sub(r"@author: (.*)", r":author: \1", line))
-            elif "@param" in line:
-                out.write(re.sub(r"@param ([^:]*):", r":param \1:", line))
-            else:
-                out.write(line)
-        out.close()
+        with open(temp, 'w') as out:
+            for line in f:
```
boegel, line 61: Please rename `f` to `fh` while at it.

easybuild/scripts/mk_tmpl_easyblock_for.py
```diff
@@ -225,9 +225,8 @@ def make_module_extra(self):
     dirpath = os.path.dirname(easyblock_path)
     if not os.path.exists(dirpath):
         os.makedirs(dirpath)
-    f = open(easyblock_path, "w")
-    f.write(txt)
-    f.close()
+    with open(easyblock_path, "w") as f:
```
boegel, line 228: `f` -> `fh`

easybuild/tools/configobj.py
```diff
@@ -1213,9 +1213,8 @@ def _load(self, infile, configspec):
         if isinstance(infile, string_type):
             self.filename = infile
             if os.path.isfile(infile):
-                h = open(infile, 'r')
-                infile = h.read() or []
-                h.close()
+                with open(infile, 'r') as h:
```
boegel, line 1216: `h` -> `fh`

easybuild/tools/configobj.py
```diff
@@ -1224,9 +1223,8 @@ def _load(self, infile, configspec):
                 if self.create_empty:
                     # this is a good test that the filename specified
                     # isn't impossible - like on a non-existent device
-                    h = open(infile, 'w')
-                    h.write('')
-                    h.close()
+                    with open(infile, 'w') as h:
```
boegel, line 1226: `h` -> `fh`

easybuild/tools/configobj.py
```diff
@@ -2052,9 +2050,8 @@ def write(self, outfile=None, section=None):
         if outfile is not None:
             outfile.write(output)
         else:
-            h = open(self.filename, 'wb')
-            h.write(output)
-            h.close()
+            with open(self.filename, 'wb') as h:
```
boegel, line 2053: `h` -> `fh`

.github/workflows/unit_tests.yml
Flamefire: And the issue is in more than just read_file. So we need to run all of it to be sure.

easybuild/tools/filetools.py
```diff
@@ -189,11 +188,19 @@ def is_readable(path):
         raise EasyBuildError("Failed to check whether %s is readable: %s", path, err)
 
 
+def _open(path, mode):
```
boegel, line 191: Maybe the function name could be a bit better here, like `_smart_open_file` or `_safe_open_file`?

easybuild/tools/filetools.py
```diff
@@ -189,11 +188,19 @@ def is_readable(path):
...
+    """Open a file. If mode is not binary, then utf-8 encoding will be selected for Python3"""
```
boegel, line 192: `will be selected` -> `will be used`
`Python3` -> `Python 3.x`

easybuild/tools/filetools.py
```diff
@@ -1011,10 +1018,9 @@ def calc_block_checksum(path, algorithm):
     _log.debug("Using blocksize %s for calculating the checksum" % blocksize)
 
     try:
-        f = open(path, 'rb')
-        for block in iter(lambda: f.read(blocksize), b''):
-            algorithm.update(block)
-        f.close()
+        with open(path, 'rb') as f:
```
boegel, line 1021: `f` -> `fh`

easybuild/tools/filetools.py
```diff
@@ -1381,14 +1381,21 @@ def apply_regex_substitutions(paths, regex_subs, backup='.orig.eb'):
...
 
-                for line_id, line in enumerate(fileinput.input(path, inplace=1, backup=backup_ext)):
-                    for regex, subtxt in compiled_regex_subs:
-                        match = regex.search(line)
-                        if match:
-                            origtxt = match.group(0)
-                            _log.info("Replacing line %d in %s: '%s' -> '%s'", (line_id + 1), path, origtxt, subtxt)
-                        line = regex.sub(subtxt, line)
-                    sys.stdout.write(line)
+                backup_path = path + (backup or '.bak')
```
boegel, line 1384: The `or '.bak` part is new, and implies always backing up?

That's a change in behavior, see code removed above that mentions *no* backup is created when an empty string is passed in as `backup` value...

I understand this complicates things below w.r.t. "streaming", but we should be careful here imho, in some places (like easyblocks beyond our control), the may rely on the `backup=''` behavior...

We can grab the text when doing the `.read()` above?

easybuild/tools/filetools.py
```diff
@@ -1381,14 +1381,21 @@ def apply_regex_substitutions(paths, regex_subs, backup='.orig.eb'):
...
+                            for regex, subtxt in compiled_regex_subs:
+                                match = regex.search(line)
+                                if match:
+                                    origtxt = match.group(0)
+                                    _log.info("Replacing line %d in %s: '%s' -> '%s'",
+                                              (line_id + 1), path, origtxt, subtxt)
+                                line = regex.sub(subtxt, line)
+                            out_file.write(line)
+                if not backup:
+                    remove_file(backup_path)
```
boegel, line 1398: Ah, ok, so we're not changing behavior after all... But we are reading the file twice though, that can be easily avoided now?

easybuild/tools/filetools.py
```diff
@@ -2042,7 +2049,7 @@ def find_flexlm_license(custom_env_vars=None, lic_specs=None):
             if lic_files:
                 for lic_file in lic_files:
                     try:
-                        open(lic_file, 'r')
+                        open(lic_file, 'r').close()
```
boegel, line 2052: This deserves a comment I think...
```
# just try to open file for reading, no need to actually read it
```

test/framework/utilities.py
```diff
@@ -264,9 +264,8 @@ def eb_main(self, args, do_build=False, return_error=False, logfile=None, verbos
             logfile = self.logfile
         # clear log file
         if logfile:
-            f = open(logfile, 'w')
-            f.write('')
-            f.close()
+            with open(logfile, 'w') as f:
```
boegel, line 267: `f` -> `fh`

test/framework/filetools.py
```diff
@@ -1212,6 +1233,15 @@ def test_apply_regex_substitutions(self):
         path = os.path.join(self.test_prefix, 'nosuchfile.txt')
         self.assertErrorRegex(EasyBuildError, error_pat, ft.apply_regex_substitutions, path, regex_subs)
 
+        # make sure apply_regex_substitutions can patch files that include UTF-8 characters
+        testtxt = b"foo \xe2\x80\x93 bar"  # This is an UTF-8 "-"
+        ft.write_file(testfile, testtxt)
+        ft.apply_regex_substitutions(testfile, [('foo', 'FOO')])
+        txt = ft.read_file(testfile)
+        if sys.version_info[0] == 3:
+            testtxt = testtxt.decode('utf-8')
```
boegel, line 1242: Hmm, shouldn't this `decode` be done in `read_file` itself, to maintain backward compatibility?

boegel (changes requested):

easybuild/tools/filetools.py
Flamefire: Couldn't think of anything useful. It is a wrapper around `open`, so maybe `_open_wrapper`? I wanted to keep the name simple as it basically just does call `open` and using a consistent default encoding, which is even basically a no-op in most cases. Nothing smart here ;)

easybuild/tools/filetools.py
Flamefire: Yes and no. I basically copied the logic from fileinput which does the same. The backup is removed if it is not wanted later but we need it for the stream-processing. So there is no change in behavior in practice: It is just more visible what also happened before

Yes using what we read above makes sense. This would likely be even faster. Maybe. However this means we'd need to do the line splitting ourselves. Not sure how different line endings are handled by the file iterable. So maybe move that to an issue and do later? As mentioned: No behavior change by now here.

test/framework/filetools.py
Flamefire: It does. This is only the decode on the  `testtxt` not `txt`

easybuild/tools/filetools.py
Flamefire: Ok it seems we can split the lines simply be `\n` and not reread the file. Sounds easy, will try.

easybuild/tools/filetools.py
Flamefire: Done

.github/workflows/unit_tests.yml
Flamefire: I added an extra job for that

easybuild/base/optcomplete.py
Flamefire: This requires a flake8 plugin. Not doing here as maybe more will pop up, maybe open an issue to introduce that later

boegel: Follow-up on this via #3568

.github/workflows/unit_tests.yml
```diff
@@ -9,6 +9,7 @@ jobs:
         python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
         modules_tool: [Lmod-7.8.22, Lmod-8.2.9, modules-tcl-1.147, modules-3.2.10, modules-4.1.4]
         module_syntax: [Lua, Tcl]
+        lc_all: [""]
```
boegel, line 12: Shouldn't this be `lc_all: ""`?

.github/workflows/unit_tests.yml
```diff
@@ -129,6 +137,7 @@ jobs:
         EB_VERBOSE: 1
         EASYBUILD_MODULE_SYNTAX: ${{matrix.module_syntax}}
         TEST_EASYBUILD_MODULE_SYNTAX: ${{matrix.module_syntax}}
+        LC_ALL: ${{matrix.lc_all}}
```
boegel, line 140: Now you're always setting `$LC_ALL`. Setting it to an empty value is *not* the same as not setting it at all.

So please make this conditional below (if `${{matrix.lc_all}}` is not empty, use it to define `$LC_ALL` via `export`

easybuild/base/fancylogger.py
boegel: OK, fine.

easybuild/tools/filetools.py
boegel: I prefer `_open_wrapper` over just `_open` then, I can't think of a better name myself either. :)

easybuild/tools/filetools.py
```diff
@@ -1011,10 +1018,9 @@ def calc_block_checksum(path, algorithm):
     _log.debug("Using blocksize %s for calculating the checksum" % blocksize)
 
     try:
-        f = open(path, 'rb')
-        for block in iter(lambda: f.read(blocksize), b''):
-            algorithm.update(block)
-        f.close()
+        with open(path, 'rb') as fh:
```
boegel, line 1021: also use `_open_wrapper` here, for consistency?

easybuild/tools/filetools.py
```diff
@@ -1360,35 +1366,37 @@ def apply_regex_substitutions(paths, regex_subs, backup='.orig.eb'):
...
-                        if match:
-                            origtxt = match.group(0)
-                            _log.info("Replacing line %d in %s: '%s' -> '%s'", (line_id + 1), path, origtxt, subtxt)
-                        line = regex.sub(subtxt, line)
-                    sys.stdout.write(line)
+                if backup:
+                    copy_file(path, path + backup)
+                with _open(path, 'w') as out_file:
+                    lines = txt_utf8.split('\n')
+                    del txt_utf8
```
boegel, line 1389: Why the `del` here already? Just let it go out of scope so garbage collector will pick it up?

This looks weird/dangerous to me, so if you have a good reason for it, you should add a comment.

easybuild/tools/filetools.py
```diff
@@ -2042,7 +2050,8 @@ def find_flexlm_license(custom_env_vars=None, lic_specs=None):
             if lic_files:
                 for lic_file in lic_files:
                     try:
-                        open(lic_file, 'r')
+                        # just try to open file for reading, no need to actually read it
+                        open(lic_file, 'r').close()
```
boegel, line 2054: Use `_open_wrapper`?

easybuild/tools/filetools.py
boegel: Maybe it should even be `open_wrapper`, so it's a public function that can be used in other places outside of the `filetools` module?

test/framework/filetools.py
boegel: ok, gotcha

test/framework/modulestool.py
```diff
@@ -39,7 +39,7 @@
 from easybuild.base import fancylogger
 from easybuild.tools import modules
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import which, write_file
+from easybuild.tools.filetools import which, write_file, read_file
```
boegel, line 42: `read_file` should go before `which` to retain alphabetical order

boegel (changes requested):

.github/workflows/unit_tests.yml
Flamefire: No. The matrix expects a list

.github/workflows/unit_tests.yml
Flamefire: I think this is fine. I tested `locale` and `LC_ALL= locale` and got the same output. I haven't found any information telling me otherwise but found https://www.baeldung.com/linux/locale-environment-variables
```
To reset it, we could export an empty value to LC_ALL:

$ export LC_ALL=
```

As this solution is the cleanest, easiest (code wise), I'd suggest to keep it

easybuild/tools/filetools.py
Flamefire: The scope is the function and the file read might be large. We basically copy the contents multiple times: input string, list and joined list. I wanted to keep the scope of that short to limit the amount of memory used similar to what `fileinput` did: Processing the file line-wise not as a whole. We don't do that anymore but can at least avoid having 3 copies in memory or accidentally using the wrong variable, so this is rather safe than dangerous

Note that the `del` is safe: It is simply that: It ends the variables lifetime prior to its scope, may lead to freeing its resources by the GC and if you were to use that variable you'd get an error "variable does not exist".
As `del` is a core Python feature I wouldn't add a comment here. The statement alone says "I'm done with this variable"

easybuild/tools/filetools.py
Flamefire: Went with `open_file` which is similar to read_file, write_file. We could name it `open_text_file` which would be more correct and tell, that we don't need it for binary files. But then it might be confusing, so I just changed the docstring and added a comment

easybuild/tools/filetools.py
Flamefire: As discussed in Slack: Not using that function for binary files

easybuild/tools/filetools.py
Flamefire: Changed to 'rb'-mode and hence not using the function intended for text files

boegel: :+1:

.github/workflows/unit_tests.yml
boegel: OK, good enough then.

.github/workflows/unit_tests.yml
boegel: Ugh, ok, fine.

easybuild/tools/filetools.py
boegel: OK, I'd expect it to be fine too, but it's still a pattern you don't see very often. Let's leave it as is, thanks for clarifying...

boegel (approved):

easybuild/tools/modules.py
```diff
@@ -1403,16 +1403,12 @@ def update(self):
...
-                        mkdir(cache_dir, parents=True)
-                    with open(cache_fp, 'w') as cache_file:
-                        cache_file.write(stdout)
-                except (IOError, OSError) as err:
-                    raise EasyBuildError("Failed to update Lmod spider cache %s: %s", cache_fp, err)
+                cache_fp = os.path.join(self.USER_CACHE_DIR, 'moduleT.lua')
+                self.log.debug("Updating Lmod spider cache %s with output from '%s'" % (cache_fp, ' '.join(cmd)))
+                cache_dir = os.path.dirname(cache_fp)
+                if not os.path.exists(cache_dir):
+                    mkdir(cache_dir, parents=True)
```
Flamefire, line 1410: FTR: write_file already does this

PR #3566, opened 2021-02-04
Create the lib64 symlink as a relative symlink

### LINKED ISSUE #3564: Symbolic link lib64 to lib should be a relative link
This is a comment from 
https://github.com/easybuilders/easybuild-framework/pull/3401

The symlink created for lib64 is a symlink to the absolute path. It should be a symlink with the relative path.

Flamefire:
This makes moving the install tree easier.

Closes #3564

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2405,6 +2405,8 @@ class EasyBlock(object):
             lib_dir = os.path.join(self.installdir, 'lib')
             lib64_dir = os.path.join(self.installdir, 'lib64')
             if os.path.exists(lib_dir) and not os.path.exists(lib64_dir):
+                # create *relative* 'lib64' symlink to 'lib';
+                # see https://github.com/easybuilders/easybuild-framework/issues/3564
                 symlink('lib', lib64_dir, use_abspath_source=False)
 
     def sanity_check_step(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2405,7 +2405,7 @@ def post_install_step(self):
             lib_dir = os.path.join(self.installdir, 'lib')
             lib64_dir = os.path.join(self.installdir, 'lib64')
             if os.path.exists(lib_dir) and not os.path.exists(lib64_dir):
-                symlink(lib_dir, lib64_dir)
+                symlink('lib', lib64_dir, use_abspath_source=False)
```
boegel, line 2408:
```suggestion
                # create *relative* 'lib64' symlink to 'lib';
                # see https://github.com/easybuilders/easybuild-framework/issues/3564
                symlink('lib', lib64_dir, use_abspath_source=False)
```

boegel (changes requested):
@Flamefire It looks "wrong" at first glance, so worth a comment imho.

easybuild/framework/easyblock.py
Flamefire: FTR: I'm not a big fan of such comments: The first is superflous as it states the same as `use_abspath_source=False` and if the second (mentioning issues) becomes a habit the code might be cluttered by those in the future.
Added anyway, just to raise awereness

boegel (approved):
lgtm

PR #3569, opened 2021-02-08
create CUDA cache (for JIT compiled PTX code) in build dir instead of $HOME

Flamefire:
Add option to control the cache size and to disable it

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1027,6 +1027,24 @@ class EasyBlock(object):
 
         mkdir(dir_name, parents=True)
 
+    def setup_cuda_cache(self):
+        cuda_cache_maxsize = build_option('cuda_cache_maxsize')
+        if cuda_cache_maxsize is None:
+            cuda_cache_maxsize = 1 * 1024  # 1 GiB default value
+        else:
+            cuda_cache_maxsize = int(cuda_cache_maxsize)
+        if cuda_cache_maxsize == 0:
+            self.log.info('Disabling CUDA PTX cache as per request')
+            env.setvar('CUDA_CACHE_DISABLE', '1')
+        else:
+            cuda_cache_dir = build_option('cuda_cache_dir')
+            if not cuda_cache_dir:
+                cuda_cache_dir = os.path.join(self.builddir, 'eb-cuda-cache')
+            self.log.info('Enabling CUDA PTX cache of size %s MiB at %s', cuda_cache_maxsize, cuda_cache_dir)
+            env.setvar('CUDA_CACHE_DISABLE', '0')
+            env.setvar('CUDA_CACHE_PATH', cuda_cache_dir)
+            env.setvar('CUDA_CACHE_MAXSIZE', str(cuda_cache_maxsize * 1024 * 1024))
+
     #
     # MODULE UTILITY FUNCTIONS
     #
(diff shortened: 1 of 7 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2150,6 +2150,22 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
             self.log.info("Loading extra modules: %s", extra_modules)
             self.modules_tool.load(extra_modules)
 
+        # Setup CUDA cache if required. If we don't do this, CUDA will use the $HOME for its cache files
+        if get_software_root('CUDA') or get_software_root('CUDAcore'):
```
boegel, line 2154: @Flamefire This belongs in a separate function imho, which we just call here.

How about a `set_up_cuda_cache` function in filetools?

easybuild/tools/options.py
```diff
@@ -357,6 +357,9 @@ def override_options(self):
             'consider-archived-easyconfigs': ("Also consider archived easyconfigs", None, 'store_true', False),
             'containerize': ("Generate container recipe/image", None, 'store_true', False, 'C'),
             'copy-ec': ("Copy specified easyconfig(s) to specified location", None, 'store_true', False),
+            'cuda-cache-maxsize': ("Maximum size of the CUDA cache (in bytes) used for JIT compilation of PTX code. "
```
boegel, line 360: Would a size in MB make more sense?

I'd say it's very unlikely that people would want to use a cache size smaller than 1MB?

easybuild/framework/easyblock.py
```diff
@@ -2150,6 +2150,22 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
...
+            cuda_cache_maxsize = build_option('cuda_cache_maxsize')
+            if cuda_cache_maxsize is None:
+                cuda_cache_maxsize = 1 * 1024 * 1024 * 1024  # 1 GB default value
+            if cuda_cache_maxsize == 0:
+                self.log.info('Disabling CUDA PTX cache as per request')
+                env.setvar('CUDA_CACHE_DISABLE', '1')
+            else:
+                cuda_cache_dir = os.path.join(self.builddir, 'eb-cuda-cache')
```
boegel, line 2162: Don't we want to make this configurable too?

I can see myself using `/dev/shm` for this for example, but always using `/dev/shm` for `--buildpath` is probably not a good idea...

boegel (changes requested):

easybuild/framework/easyblock.py
Flamefire: Added --cuda-cache-dir

easybuild/tools/options.py
Flamefire: Ok done, changed this to be MiB

easybuild/framework/easyblock.py
Flamefire: Locations considered for that function:
- filetools: This is not reading/writing of files but simply changing some env vars
- environment: That contains only the most basic functions
- toolchain: Likely very good fit but excludes allow_system_deps and extra_modules options if done in `prepare` and needs access to EasyBlock.builddir for the default
- EasyBlock: Easiest and also makes sense. Could make it private if we want to decide on the exact location later

boegel (approved):

PR #3570, opened 2021-02-08
Making symlinking of posix_c.so to posix.so in test suite configuration conditional

justbennet:
The missing lua posix.so symlink has been fixed in Ubuntu 2.04.1 so
the creating of it breaks testing.  I added a conditional that will
only create it if the posix.so file does not exist as either a file
or a symlink.


boegel (approved):
lgtm, thanks @justbennet!

PR #3572, opened 2021-02-13
use http://sources/easybuild.io as fallback source URL

boegel:
As discussed recently, it's useful to bake in a fallback source URL, so we can easily and quickly fix disappearing upstream source tarballs (assuming we're allowed to redistribute them of course), see for example https://github.com/easybuilders/easybuild-easyconfigs/pull/12156 .

For now, only UDUNITS source tarballs are available at http://sources/easybuild.io (which is a VPS hosted by [fosshost](https://fosshost.org/)).


bartoldeman (approved):
lgtm

PR #3576, opened 2021-02-16
use https for sources.easybuild.io fallback URL

boegel:
follow-up to #3572, SSL certificate is now in place for https://sources.easybuild.io


bartoldeman (approved):
lgtm

bartoldeman:
Confirmed, I can see the let's encrypt certificate. Merging :)

PR #3578, opened 2021-02-17
Add cuda_cc_space_sep and cuda_cc_semicolon_sep templates

Flamefire:
The latter is required for e.g. PyTorch and TorchVision which need a
CMake list, which is semicolon separated

### DIFF
--- a/test/framework/easyconfig.py
+++ b/test/framework/easyconfig.py
@@ -4053,7 +4053,7 @@ class EasyConfigTest(EnhancedTestCase):
             "cuda_compute_capabilities = ['5.1', '7.0', '7.1']",
             "installopts = '%(cuda_compute_capabilities)s'",
             "preinstallopts = '%(cuda_cc_space_sep)s'",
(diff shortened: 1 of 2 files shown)


test/framework/easyconfig.py
```diff
@@ -4052,20 +4052,26 @@ def test_cuda_compute_capabilities(self):
             "toolchain = SYSTEM",
             "cuda_compute_capabilities = ['5.1', '7.0', '7.1']",
             "installopts = '%(cuda_compute_capabilities)s'",
+            "preinstallopts = '%(cuda_cc_space_sep)s'",
+            "postinstallcmds = '%(cuda_cc_semicolon_sep)s'",
```
boegel, line 4056: please use `prebuildopts` here instead, `postinstallcmds` doesn't make any sense at all (that should be a list of strings representing commands, not a single string value)

boegel (changes requested):

test/framework/easyconfig.py
Flamefire: Done

boegel (approved):

PR #3580, opened 2021-02-18
symlink 'lib' to 'lib64' if it doesn't exist

### LINKED ISSUE #3549: symlink 'lib' to 'lib64' if 'lib' directory is missing
Similar to #3401, but the other way around...

Should fix some problems on OpenSUSE, see https://github.com/easybuilders/easybuild-easyblocks/issues/2309 and https://github.com/easybuilders/easybuild-easyconfigs/issues/11976

boegel:
fixes #3549

### DIFF
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -3145,7 +3145,7 @@ class ToyBuildTest(EnhancedTestCase):
         lib_path = os.path.join(toy_installdir, 'lib')
         lib64_path = os.path.join(toy_installdir, 'lib64')
 
-        # lib6 subdir exists, is not a symlink
+        # lib64 subdir exists, is not a symlink
         self.assertTrue(os.path.exists(lib_path))
         self.assertTrue(os.path.isdir(lib_path))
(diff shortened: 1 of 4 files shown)


test/framework/toy_build.py
```diff
@@ -3133,25 +3133,29 @@ def test_toy_build_unicode_description(self):
...
 
-        # by default, lib64 symlink is created
+        # by default, lib64 -> lib symlink is created (--lib64-lib-symlink is enabled by default)
         self.test_toy_build()
 
         toy_installdir = os.path.join(self.test_installpath, 'software', 'toy', '0.0')
         lib_path = os.path.join(toy_installdir, 'lib')
         lib64_path = os.path.join(toy_installdir, 'lib64')
 
+        # lib6 subdir exists, is not a symlink
```
ocaisa, line 3148: lib64

ocaisa (changes requested):
Tiny typo

ocaisa (approved):

PR #3581, opened 2021-02-18
add support for intel-compiler compiler toolchain (>= 2021.x versions, oneAPI)

boegel:
For now this just inherits most stuff from `IntelIccIfort`, so it still uses `icc`, `icpc`, `ifort`.

Eventually, we'll need to come up with a mechanism to allow switching to the new compiler commands (`icx`, `icpx`,`ifx`), as well as supporting `dpcpp`, perhaps via `toolchainopts`?


bartoldeman (approved):
lgtm

PR #3582, opened 2021-02-18
fix Python module filename for intel-compilers toolchain

boegel:
This is required to ensure we can import from `easybuild.toolchains.intel_compilers`...


bartoldeman (approved):
lgtm

PR #3589, opened 2021-02-23
use 'main' rather than 'master' branch in GitHub integration functionality

boegel:
We're making the necessary changes to rename the `master` branch to `main`.

Currently both are available (and they're identical), but we'll remove `master` as soon as we're done with making the necessary changes.

This shouldn't have a lot of impact, since the most important branch for ongoing development is `develop` anyway, the `master` (now `main`) branch is only there to correspond with the latest EasyBuild release...


migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3591, opened 2021-02-23
sync with main branch + bump version to 4.3.4dev

boegel:


easybuild/tools/github.py
```diff
@@ -311,6 +317,14 @@ def fetch_latest_commit_sha(repo, account, branch='master', github_user=None, to
...
     :return: latest SHA1
     """
+    if branch is None:
+        # use 'main' as default branch for 'easybuilders' organisation,
+        # otherwise use 'master'
+        if account == GITHUB_EB_MAIN:
+            branch = 'main'
+        else:
+            branch = 'master'
+
```
akesandgren, line 327: Should probably split out to a separate func since it's used 3 times.

boegel: That should be done in a separate PR, I can take care of that after merging this one...

akesandgren: ok.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3600, opened 2021-03-01
add support for using fallback 'arch=*' key in dependency version specified as arch->version mapping

boegel:
follow-up for #3506

### DIFF
--- a/test/framework/easyconfig.py
+++ b/test/framework/easyconfig.py
@@ -72,7 +72,7 @@ from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
 from easybuild.tools.options import parse_external_modules_metadata
 from easybuild.tools.py2vs3 import OrderedDict, reload
 from easybuild.tools.robot import resolve_dependencies
-from easybuild.tools.systemtools import get_cpu_architecture, get_shared_lib_ext
+from easybuild.tools.systemtools import AARCH64, POWER, X86_64, get_cpu_architecture, get_shared_lib_ext
 from easybuild.tools.toolchain.utilities import search_toolchain
 from easybuild.tools.utilities import quote_str, quote_py_str
 from test.framework.utilities import find_full_path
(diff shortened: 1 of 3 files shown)


test/framework/easyconfig.py
```diff
@@ -305,6 +307,68 @@ def test_dependency(self):
...
+            'description = "test easyconfig"',
+            'toolchain = {"name":"GCC", "version": "4.6.3"}',
+            'dependencies = ['
+            '   ("not_on_power", {"arch=*": "1.2.3", "arch=POWER": False}),',
+            ']',
+        ])
+        self.prep()
+
+        # only non-POWER arch, dependency is retained
+        for arch in (AARCH64, X86_64):
```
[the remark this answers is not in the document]
boegel: Oh nice catch Hound!

test/framework/easyconfig.py
```diff
@@ -305,6 +307,69 @@ def test_dependency(self):
...
+            'description = "test easyconfig"',
+            'toolchain = {"name":"GCC", "version": "4.6.3"}',
+            'dependencies = ['
+            '   ("first", "1.0"),',
+            '   ("second", {"arch=%s": False}),' % arch,
+            ']',
+            'builddependencies = [',
+            '   ("first_build", {"arch=%s": False}),' % arch,
+            '   ("second_build", "2.0"),',
+            ']',
```
akesandgren, line 335: Style nitpicking, we usually keep builddeps before deps, and if would make it less confusing regarding the assert checks below where second_build comes before first in the deps array.

akesandgren (changes requested):

akesandgren (approved):
LGTM

test/framework/easyconfig.py
boegel: fixed in 3c1eb7f

akesandgren:
Going in, thanks @boegel!

Issue #3606, opened 2021-03-11
Cannot merge a PR where a 'changes requested' review has not been dismissed


branfosj:
Regression from #3604

In https://github.com/easybuilders/easybuild-easyconfigs/pull/12379
* I requested changes
* Changes were made
* I approved the PR

Then I get that there are still pending changes.
```
$ eb --merge-pr 12379 -D
== Temporary log file in case of crash /scratch/branfosj-admin/eb-9kaboihu/easybuild-cbqtnno8.log

easybuilders/easybuild-easyconfigs PR #12379 was submitted by sassy-crick, you are using GitHub account 'branfosj'

Checking eligibility of easybuilders/easybuild-easyconfigs PR #12379 for merging...
* targets develop branch: OK
* test suite passes: OK
* last test report is successful: OK
* no pending change requests: FAILED (changes requested by branfosj) => not eligible for merging!
* approved review: OK (by branfosj)
* milestone is set: OK (4.x)
* mergeable state is clean: OK

WARNING: Review indicates this PR should not be merged (use -f/--force to do so anyway)
```

I also then dismissed the approval to see if I could dismiss the original review - but I could not see a way to dismiss the original review.

PR #3607, opened 2021-03-11
ignore undismissed 'changes requested' review if there is an 'approved' review by the same user

migueldiascosta:
fixes https://github.com/easybuilders/easybuild-framework/issues/3606 (if I understand github's review workflow correctly)


boegel (approved):
lgtm

PR #3609, opened 2021-03-12
Sort output of eb --search in natural order (respecting numbers)

Flamefire:
This makes the output of `eb --search` more approachable by sorting versions in natural order. I.e. Foo-10.1 comes **after* Foo-9.2

Had to create and move a test file to add a test for natural_keys

Code is based on https://stackoverflow.com/a/5967539/1930508 which is based on http://nedbatchelder.com/blog/200712/human_sorting.html


boegel (changes requested):
@Flamefire You also need to include the new test subsuite to the general `test/suite.py`, or those tests will not be run in CI, cfr. https://github.com/Flamefire/easybuild-framework/pull/5

Flamefire:
Thanks. I'm too used to pytest which makes writing unit tests so much easier...

boegel (approved):

PR #3610, opened 2021-03-14
enhance 'eb' command to ensure that easybuild.main can be imported before settling on python* command to use

boegel:
This is a small enhancement for the `eb` command: only accepted the `python*` command if it's able to import `easybuild.main`.

This fixes the following strange/confusing situation, which occurs currently because `eb` considers `python` before `python3` (unless `$EB_PYTHON` is set):

```
# default Python command is Python 2
$ python -V
Python 2.7.18

# use pip3 to install EasyBuild, with the intention to run it via 'python3'
$ pip3 install easybuild

# try running 'eb' fails because default 'python' command is not Python 3
$ eb --version
/usr/bin/python: No module named easybuild
```

### DIFF
--- a/.github/workflows/eb_command.yml
+++ b/.github/workflows/eb_command.yml
@@ -1,0 +1,88 @@
+# documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
+name: Tests for the 'eb' command
+on: [push, pull_request]
+jobs:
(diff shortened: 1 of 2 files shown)


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #2878, opened 2019-05-13
add support for 'fix_python_shebang_for' and 'fix_perl_shebang_for' easyconfig parameters

boegel:
fixes #2869 (cc @smoors)

PR #3613, opened 2021-03-16
added configuration option to define the env command to use for shebangs

mboisson:
With the feature introduced in PR https://github.com/easybuilders/easybuild-framework/pull/2878, it can break some scripts, because the arguments following the command are passed as a single argument. This is fixed in recent `env` by providing the option `-S`. This is however not available on all operating systems. Notably, the `/usr/bin/env` provided by CentOS7 does not support that. 

This PR makes this a configuration option that sites can use to point to a different `env` command and optionally specify the `-S` argument if supported by their OS.

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -375,6 +375,7 @@ class EasyBuildOptions(GeneralOption):
                           None, 'store', None, 'e', {'metavar': 'CLASS'}),
             'enforce-checksums': ("Enforce availability of checksums for all sources/patches, so they can be verified",
                                   None, 'store_true', False),
+            'env-for-shebang': ("Define the env command to use when fixing shebangs", None, 'store', '/usr/bin/env'),
             'experimental': ("Allow experimental code (with behaviour that can be changed/removed at any given time).",
                              None, 'store_true', False),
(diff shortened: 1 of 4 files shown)


easybuild/tools/options.py
```diff
@@ -375,6 +375,7 @@ def override_options(self):
                           None, 'store', None, 'e', {'metavar': 'CLASS'}),
             'enforce-checksums': ("Enforce availability of checksums for all sources/patches, so they can be verified",
                                   None, 'store_true', False),
+            'env-for-shebang': ("Define the env command to use when fixing shebangs", None, 'store', '/usr/bin/env'),
```
boegel, line 378: @mboisson You should use the `DEFAULT_ENV_FOR_SHEBANG` constant here to specify the default value, not hardcode `/usr/bin/env`

boegel (changes requested):

boegel (approved):
lgtm

boegel (changes requested):
@mboisson Test problem is fixed with https://github.com/ComputeCanada/easybuild-framework/pull/24

boegel (approved):

PR #3616, opened 2021-03-18
Refactor the CI configuration to use inclusion instead of exclusion

Flamefire:
This provides more clarity on what exactly is run

The jobs have not changed, the following configurations are tested before (and after):
```
build (2.7, Lmod-7.8.22, Lua)
build (2.7, Lmod-7.8.22, Tcl)
build (2.7, Lmod-8.2.9, Lua)
build (2.7, Lmod-8.2.9, Tcl)
build (2.7, modules-tcl-1.147, Tcl)
build (2.7, modules-3.2.10, Tcl)
build (2.7, modules-4.1.4, Tcl)

build (3.6, Lmod-7.8.22, Lua)
build (3.6, Lmod-7.8.22, Tcl)
build (3.6, Lmod-8.2.9, Lua)
build (3.6, Lmod-8.2.9, Tcl)
build (3.6, modules-tcl-1.147, Tcl)
build (3.6, modules-3.2.10, Tcl)
build (3.6, modules-4.1.4, Tcl)

build (3.5, Lmod-8.2.9, Lua)
build (3.5, Lmod-8.2.9, Tcl)

build (3.7, Lmod-8.2.9, Lua)
build (3.7, Lmod-8.2.9, Tcl)

build (3.8, Lmod-8.2.9, Lua)
build (3.8, Lmod-8.2.9, Tcl)

build (3.9, Lmod-8.2.9, Lua)
build (3.9, Lmod-8.2.9, Tcl)

build (3.6, Lmod-8.2.9, Lua, C)
```

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         python: [2.7, 3.6]
-        modules_tool: [Lmod-7.8.22, Lmod-8.2.9, modules-tcl-1.147, modules-3.2.10, modules-4.1.4]
+        modules_tool: [Lmod-7.8.22, Lmod-8.4.27, modules-tcl-1.147, modules-3.2.10, modules-4.1.4]
         module_syntax: [Lua, Tcl]
         lc_all: [""]
         # don't test with Lua module syntax (only supported in Lmod)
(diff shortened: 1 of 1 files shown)


boegel (dismissed):
@Flamefire lgtm, that looks a bit cleaner indeed.
Let's bump to the latest Lmod 8.x while we're at it: https://github.com/Flamefire/easybuild-framework/pull/7

Flamefire:
I'd like to try something more: Avoid the repetition of the lmod version.
It seems you can generate a whole matrix from within a job step: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-returning-a-json-object

But what we just need here is something we can access with `${{}}`, e.g. an env variable. I'll try that later, so marking this as a draft

boegel:
> I'd like to try something more: Avoid the repetition of the lmod version.
> It seems you can generate a whole matrix from within a job step: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-returning-a-json-object
> 
> But what we just need here is something we can access with `${{}}`, e.g. an env variable.
> I'll try that later, so hold up the merge please :)

OK! :)

boegel:
@Flamefire Once you're happy with the changes, I'll need to re-configure the test configurations that are required to pass (which is why there's a bunch of orange dots sitting there "waiting" right now).

Flamefire:
I think this works now. What do you think about this approach?
Alternative:
```
env:
  LMOD_7: Lmod-...
  ...
  modules_tool: [${env.LMOD_7}, ...]
```
That's a bit shorter but sets those as actual env variables, so my current approach feels "more right"

.github/workflows/unit_tests.yml
```diff
@@ -2,61 +2,68 @@
...
   build:
+    needs: setup
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
-        modules_tool: [Lmod-7.8.22, Lmod-8.2.9, modules-tcl-1.147, modules-3.2.10, modules-4.1.4]
+        python: [2.7, 3.6]
+        modules_tool:
+          - ${{needs.setup.outputs.lmod7}}
```
boegel, line 22: I think it doesn't hurt to add a comment here:

```
# use variables defined by 'setup' job above,
# see also https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#needs-context

File: .github/workflows/unit_tests.yml [hunk 2]
```diff
@@ -2,61 +2,68 @@
...
-          - modules_tool: Lmod-7.8.22
-            python: 3.7
-          - modules_tool: Lmod-7.8.22
-            python: 3.8
-          - modules_tool: Lmod-7.8.22
-            python: 3.9
-        # There may be encoding errors in Python 3 which are hidden when an UTF-8 encoding is set
-        # Hence run the tests (again) with LC_ALL=C and Python 3.6 (or any < 3.7)
         include:
+          # Test different Python versions with Lmod 8.x (with both Lua and Tcl module syntax)
```
boegel, line 38:
```suggestion
          # Test different Python 3 versions with Lmod 8.x (with both Lua and Tcl module syntax)
```

boegel (changes requested):

boegel (approved):
lgtm

PR #3618, opened 2021-03-24
add templates for architecture independent Python wheels

Flamefire:
Wheels with the -none-any suffix are basically archives of source code which can often be installed with pip where using the tar.gz would fail for reasons such as missing versions

Follow-up to https://github.com/easybuilders/easybuild-easyconfigs/pull/12455


boegel (approved):
lgtm

PR #3622, opened 2021-03-29
mention easyblocks PR in gist when uploading test report for it + fix clean_gists.py script

Flamefire:
Also enhance and fix the clean_gists script to not confuse easyconfigs and easyblocks repo which leads to deleting gists to keep

This is kinda crucial after the ability to create test reports for easyblocks PRs was introduced

An example of an upload after this is in: https://github.com/easybuilders/easybuild-easyblocks/pull/2366#issuecomment-809217280


boegel (approved):
lgtm

PR #3629, opened 2021-04-05
use develop branch when testing push access in --check-github

### LINKED ISSUE #3590: Github hooks should check against develop branch
Since `master` will be switched to `main`, it breaks --check-github.

Using the `develop` branch for these checks is the obvious fix; since that's a branch that all the people using github integration needs to have on their cloned repos as well.

There are numerous mentions of "master" in links, and branch names here as well, which should all use develop instead.
https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/github.py

boegel:
fixes #3590

This will make the switch from `master` to `main` a bit less painful, and we should be using `develop` instead anyway for this...

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1457,7 +1457,7 @@ def post_pr_labels(pr, labels):
         return True
 
 
-def add_pr_labels(pr, branch='develop'):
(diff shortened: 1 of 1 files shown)


migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3630, opened 2021-04-05
deprecate --accept-eula, rename to --accept-eula-for + also accept regular expression value

### LINKED ISSUE #3595: Misleading --accept-eula configuration option and a wildcard proposal
Hi guys,

I think the '--accept-eula' configuration option is misleading.
There is already an easyconfig parameter of a very similar name, 'accept_eula', and of different data types, string vs bool.
Using the name '--accept-eula-for' instead would be more self-explanatory. The use case would be: '--accept-eula-for=AOCC,Intel-OneAPI'. I noticed that Kenneth already used such a name in PR #3535 but renamed it to --accept-eula.

To me, the configuration option naming, '--accept-eula', implies boolean type. The first time I saw the option, I thought I need to specify True to accept it. I did it and noticed that it didn't work. Checking Github and the documentation, I figured out that I need to specify the package for which I accept EULA.

On top of that, one should specify for which packages he accepts EULA. In most cases, one doesn't care and thus wants to specify that he accepts them all but in the current state cannot. Therefore, I propose to add a wildcard '*' which means that he accepts EULA for all packages he installs.

Best regards,
Robert

boegel:
fixes #3595

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1655,7 +1655,7 @@ class EasyBlock(object):
         if name is None:
             name = self.name
 
-        accepted_eulas = build_option('accept_eula') or []
+        accepted_eulas = build_option('accept_eula_for') or []
         if self.cfg['accept_eula'] or name in accepted_eulas:
             self.log.info("EULA for %s is accepted", name)
         else:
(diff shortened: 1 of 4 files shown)


Micket (approved):
lgtm

PR #3633, opened 2021-04-06
Avoid module call for unuse() for Lmod and set $MODULEPATH directly 

Flamefire:
Followup to #3557
No check for priorities is required as LMod correctly handles that
Also adds a check for an empty path which is technically possible and handled by LMod like any other path


boegel (approved):
lgtm

Flamefire:
Followup: https://github.com/easybuilders/easybuild-framework/pull/3634

PR #3634, opened 2021-04-06
Simplify use/unuse in Lmod

Flamefire:
This basically moves the common functionality of use and unuse into `_set_module_path` which takes care of proper logging and handling empty paths. Note that an empty string is different to no path, i.e. `$MODULEPATH=''` is different to `unset MODULEPATH` in Lmod. Checked with  `module use ''` which yields a set, but empty MODULEPATH and a `module unused $MODULEPATH` remove that variable completely

Not fully sure if there was a bug before, but now there provably is not due to the enhanced tests :)


Flamefire:
Closed as included in https://github.com/easybuilders/easybuild-framework/pull/3636

PR #3636, opened 2021-04-07
Avoid using a priority in prepend_module_path (Lmod) to avoid costly module calls

Flamefire:
Includes https://github.com/easybuilders/easybuild-framework/pull/3634 due to the enhanced `curr_module_paths`, can be merged here, then I'll close the other.

This PR avoids using a priority in `prepend_module_path` (Lmod) which means that $MODULEPATH can be directly modified by this and all subsequent calls to either  `prepend_module_path` or `use` which is a lot faster than calling Lmod.

The test has been changed so it now expects that `prepend_module_path` actually does prepend even when no priority is passed and another priority is in effect.
For compatibility I kept the constant `10000` as `HIGH_PRIORITY` so user sites that really want other module paths to take preference can use any higher priority.

As discussed in Slack this change is unlikely to effectively change any behavior and will very likely continue to work, see https://github.com/TACC/Lmod/issues/509#issuecomment-814502541


boegel:
@Flamefire Let's close #3634 and continue the discussion here. When this PR (#3636) gets merged, #3634 will also be merged automatically (since the commits included in here are also included in there), but context w.r.t. the discussion will be missing...

easybuild/tools/modules.py
```diff
@@ -1469,7 +1475,17 @@ def prepend_module_path(self, path, set_mod_paths=True, priority=None):
...
         if not modulepath or os.path.realpath(modulepath[0]) != os.path.realpath(path):
+            # If no explicit priority is set, but priorities are already in use we need to use a high
+            # priority to make sure the path (very likely) ends up at the front
+            if priority is None and os.environ.get('__LMOD_Priority_MODULEPATH'):
+                priority = self.HIGH_PRIORITY
             self.use(path, priority=priority)
+            modulepath = curr_module_paths(clean=False)
+            path_idx = modulepath.index(path)
+            if path_idx != 0:
+                self.log.warn("Path '%s' could not be prepended to $MODULEPATH. The following paths have a higher "
```
boegel, line 1486: Make this `print_warning` so it pops out better? You can pass `self.log` via `log=` so the warning is logged too.

easybuild/tools/modules.py
```diff
@@ -1469,7 +1475,17 @@ def prepend_module_path(self, path, set_mod_paths=True, priority=None):
         # Lmod pushes a path to the front on 'module use', no need for (costly) 'module unuse'
         modulepath = curr_module_paths()
         if not modulepath or os.path.realpath(modulepath[0]) != os.path.realpath(path):
+            # If no explicit priority is set, but priorities are already in use we need to use a high
+            # priority to make sure the path (very likely) ends up at the front
+            if priority is None and os.environ.get('__LMOD_Priority_MODULEPATH'):
+                priority = self.HIGH_PRIORITY
             self.use(path, priority=priority)
+            modulepath = curr_module_paths(clean=False)
+            path_idx = modulepath.index(path)
```
boegel, line 1484: This could trigger a `ValueError` if (somehow...) `path` is not in `$MODULEPATH`, so maybe wrap this in a `try`/`except` to avoid a nasty traceback?

It would be really weird to not have `path` at all in `modulepath` though, I must admit.

easybuild/tools/modules.py
```diff
@@ -1603,12 +1619,18 @@ def get_software_version(name):
     return version
 
 
-def curr_module_paths():
+def curr_module_paths(clean=True):
```
boegel, line 1622: I wouldn't make cleaning optional at all, I think weird things may happen when empty values are included in `$MODULEPATH` and EasyBuild starts using those. There's a reason we're filtering those here...

test/framework/easyconfig.py
```diff
@@ -2211,7 +2211,7 @@ def test_dump_extra(self):
             '',
         ])
 
-        handle, testec = tempfile.mkstemp(prefix=self.test_prefix, suffix='.eb')
+        handle, testec = tempfile.mkstemp(suffix='.eb')
```
boegel, line 2214: Why not change this to `dir=self.test_prefix`, so we're still using the tmpdir that's specific to this test?
Same below

easybuild/tools/modules.py
```diff
@@ -1433,30 +1457,12 @@ def use(self, path, priority=None):
...
-                    new_mod_path = [path] + [p for p in cur_mod_path.split(':') if p != path]
-                    new_mod_path = ':'.join(new_mod_path)
-                self.log.debug('Changing MODULEPATH from %s to %s' %
-                               ('<unset>' if cur_mod_path is None else cur_mod_path, new_mod_path))
-                os.environ['MODULEPATH'] = new_mod_path
+                self._set_module_path([path] + [p for p in curr_module_paths(clean=False) if p != path])
 
     def unuse(self, path):
         """Remove a module path"""
         # We can simply remove the path from MODULEPATH to avoid the costly module call
```
boegel, line 1464: `costly "module unuse" call`

Maybe also clarify that modifying `$MODULEPATH` directly implies that Lmod won't reconsider loaded modules, but in the context of an EasyBuild session that should be OK...

Maybe we should also add an option to force running the actual `module unuse` command (same above for `use`), in case people are using the EasyBuild interface to the modules tool in Python scripts?

easybuild/tools/modules.py
```diff
@@ -1433,30 +1457,12 @@ def use(self, path, priority=None):
             if os.environ.get('__LMOD_Priority_MODULEPATH'):
```
boegel, line 1457: We're using `__LMOD_Priority_MODULEPATH` in multiple places, so this should become a class constant in the `Lmod` class imho, and maybe refer to https://github.com/TACC/Lmod/issues/509 to clarify that it's OK to rely on this?

easybuild/tools/modules.py
```diff
@@ -1469,7 +1475,17 @@ def prepend_module_path(self, path, set_mod_paths=True, priority=None):
...
         modulepath = curr_module_paths()
         if not modulepath or os.path.realpath(modulepath[0]) != os.path.realpath(path):
+            # If no explicit priority is set, but priorities are already in use we need to use a high
+            # priority to make sure the path (very likely) ends up at the front
+            if priority is None and os.environ.get('__LMOD_Priority_MODULEPATH'):
+                priority = self.HIGH_PRIORITY
             self.use(path, priority=priority)
+            modulepath = curr_module_paths(clean=False)
+            path_idx = modulepath.index(path)
+            if path_idx != 0:
```
boegel, line 1485: If we notice the path isn't first, can't we try again with `self.use(path, priority=priority, force_module_command=True)`?
The warning should suffice to make it clear something fishy is going on, but we can still try to do the correct thing?

easybuild/tools/modules.py
```diff
@@ -1415,6 +1417,28 @@ def update(self):
...
+        new_mod_path can be None or an iterable of paths
+        """
+        if new_mod_path is not None:
+            if not isinstance(new_mod_path, list):
+                assert not isinstance(new_mod_path, string_type)  # Just to be sure
+                new_mod_path = list(new_mod_path)  # Expand generators
+            new_mod_path = mk_module_path(new_mod_path) if new_mod_path else None
+        cur_mod_path = os.environ.get('MODULEPATH')
+        if new_mod_path != cur_mod_path:
+            self.log.debug(
```
boegel, line 1432: This is worth `log.info` perhaps?

boegel (changes requested):

easybuild/tools/modules.py
Flamefire: True, done

easybuild/tools/modules.py
Flamefire: It should be virtually impossible for this to happen as it should be either caught already by our tests or is an actual (and serious) bug in Lmod. So I'd say the "nasty traceback" is fine in this rare case so we don't need to litter our code :)

easybuild/tools/modules.py
Flamefire: This is required to match Lmod behavior: Lmod uses any path as specified, so non-existing, empty or "unusual" ones are just used.
So the "fast" module use/unuse must use the non-cleaned path or behavior will differ

> There's a reason we're filtering those here...

Remember which? Also the default is still to clean, so all of EB except for the new functions is unaffected

test/framework/easyconfig.py
Flamefire: Because at the same time that `self.test_prefix` is set the temp directory is set too, so `mkstemp` will use that already. The same is used in other places already, so I think this is a left-over from a change some time ago.

easybuild/tools/modules.py
Flamefire: Done with default set to False to provide the fast implementation by default

easybuild/tools/modules.py
Flamefire: I refactored this into `_has_module_paths_with_priority` so make it even more clear and provide a single point of use

easybuild/tools/modules.py
Flamefire: This wouldn't change anything: If `priority` is `None`, then the path will be at the front already due to the fast path, if `priority` is not None or any priorities are in use then the module command will be used already.

easybuild/tools/modules.py
Flamefire: This matches behavior of current `module use/unuse` which does a debug log already. That's ok IMO, as all this changing of variables back and forth is not something you need to know for a normal, successful installation but only when debugging an issue.

easybuild/tools/modules.py
Flamefire: From Slack:
> we do need a `clean_existing_paths` for `curr_module_paths`, but it should still filter out empty paths

Why? I don't even think we should clean out non-existing paths by default and filtering out empty paths would be different from what Lmod does, although I'm not sure about the semantic of that. And if we do filter those, then the fast and regular `module.use` would be different.

akesandgren:
@Flamefire Conflict resolution needed

Flamefire:
Rebased and conflicts resolved. That was quite a lot... Please double-check the changes

akesandgren (changes requested):

akesandgren:
This now looks good to me, but since it's large i'll wait for @boegel to review before merging.

akesandgren (dismissed):
LGTM

test/framework/modules.py
```diff
@@ -1229,6 +1287,22 @@ def test_module_use_unuse(self):
             self.assertEqual(os.environ['MODULEPATH'], test_dir1)
             self.modtool.unuse(test_dir1)
             self.assertFalse('MODULEPATH' in os.environ)
+            # Unuse when the MODULEPATH is already empty
+            self.modtool.unuse(test_dir1)
+            self.assertFalse('MODULEPATH' in os.environ)
+            os.environ['MODULEPATH'] = old_module_path  # Restore
+
+            # Forcing to use the module command reloads modules (Only Lmod does this)
+            old_module_path = os.environ['MODULEPATH']
```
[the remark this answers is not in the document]
Flamefire: You are right, but those are more or less different (sub-)testcases. I wanted to make them self-contained, so it would still work if we remove one of those. Maybe best done via a fixture or context-manager but yeah...

akesandgren:
This one has a conflict

easybuild/tools/modules.py
```diff
@@ -1464,35 +1504,26 @@ def use(self, path, priority=None):
...
-                    os.environ['MODULEPATH'] = new_mod_path
+    def unuse(self, path, force_module_command=False):
+        """
+        Remove a module path
+
+        :param path: path to remove from $MODULEPATH
+        :param force_module_command: If False running the module command may be skipped which is faster
+                                     but does not reload modules
+        """
+        if force_module_command:
```
akesandgren, line 1521: Are you sure you don't need "or self._has_module_paths_with_priority()" here for the unuse case?

test/framework/modules.py
```diff
@@ -489,15 +489,23 @@ def test_curr_module_paths(self):
...
+        self.assertEqual(curr_module_paths(clean=False), [''])
 
-        os.environ['MODULEPATH'] = '%s:%s:%s' % (test1, test2, test3)
+        os.environ['MODULEPATH'] = os.pathsep.join([test1, test2, test3])
         self.assertEqual(curr_module_paths(), [test1, test2, test3])
+        self.assertEqual(curr_module_paths(clean=False), [test1, test2, test3])
 
         # empty entries and non-existing directories are filtered out
-        os.environ['MODULEPATH'] = '/doesnotexist:%s::%s:' % (test2, test1)
+        os.environ['MODULEPATH'] = os.pathsep.join(['/doesnotexist', test2, '', test1, ''])
```
akesandgren, line 505: strictly speaking it should be
os.sep + 'doesnotexist'
but we're failing on that account in soooo many places...

easybuild/tools/modules.py
Flamefire: Yes, quite sure, see b5e154c95be5dfc709f6738a1fbca0fb10805a1c

akesandgren (dismissed):
LGTM

akesandgren:
@boegel A final review by you needed.

boegel:
> There is one failing test where we should consider what we want:
> 
> ```
> self.prepend_module_path(aa)
> self.prepend_module_path(bb, 10000)
> self.prepend_module_path(cc)
> ```
> 
> If the caller of `cc` doesn't know anything about the former states he reasonably expects to be first but is not.
> 
> This currently happens in a test: `load(module, mod_paths=[aa, (bb,10000), cc])` and checks for bb being first while with this patch cc is first.

If a path was added with priority, that was likely done with good reason, so we should respect that...

This recently popped up actually, see https://github.com/easybuilders/easybuild-framework/commit/c8e723a699faaabd60286b4549177221e0b197df (which is part of https://github.com/easybuilders/easybuild-framework/pull/4991)

test/framework/modules.py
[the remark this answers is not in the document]
Flamefire: FWIW: https://github.com/easybuilders/easybuild-framework/pull/5015 introduces a context manager that can be used here

PR #3637, opened 2021-04-07
Deprecate adding a non-existing path to $MODULEPATH

Flamefire:
The `module.use` function is not actually supposed to create a directory, but we require that the directories of `$MODULEPATH` exist in `curr_module_paths`

Hence deprecate adding a path which does not exist

As a side effect we need to explicitly disallow the empty path `''`, as that also does not exist. EnvironmentModules ignores such paths, LMod does create an empty entry in `$MODULEPATH` but doesn't really use it. We also filter it in  `curr_module_paths` so it makes sense to fully disallow adding it.

During the tests on CI another bug was discovered: The module tools (at least Lmod) strip trailing slashes before adding a path to $MODULEPATH. However we sometimes do `add_module_path(os.path.join(some_prefix, modpath))` where `modpath` could be empty which leads to a trailing slash and `add_module_path` (and `remove_module_path`) not finding the path in `curr_module_paths` (which do not contain the trailing slashes) This means that paths wrongly were retained in $MODULEPATH or unnecessarily added again.

I added a test to catch this issue and fixed it.

I also had to add checks before adding Hierarchical MNS paths, such as `Core` if they exist (already). Not adding non-existing ones is fine as far as I can tell and even leads to more speed by avoiding module calls, cache invalidation and work for the modules tool (repeatably checking a non-existing folder)

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2155,7 +2155,7 @@ class EasyBlock(object):
         curr_modpaths = curr_module_paths()
         for init_modpath in init_modpaths:
             full_mod_path = os.path.join(self.installdir_mod, init_modpath)
-            if full_mod_path not in curr_modpaths:
+            if os.path.exists(full_mod_path) and full_mod_path not in curr_modpaths:
                 self.modules_tool.prepend_module_path(full_mod_path)
 
         # prepare toolchain: load toolchain module and dependencies, set up build environment
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -1089,7 +1089,7 @@ class ModulesTest(EnhancedTestCase):
         """Test module caches and invalidate_module_caches_for function."""
         self.assertEqual(mod.MODULE_AVAIL_CACHE, {})
 
-        # purposely extending $MODULEPATH with non-existing path, should be handled fine
+        # purposely extending $MODULEPATH with an empty path, should be handled fine
         nonpath = os.path.join(self.test_prefix, 'nosuchfileordirectory')
         mkdir(nonpath)
         self.modtool.use(nonpath)
--- a/test/framework/modules.py
+++ b/test/framework/modules.py
@@ -531,14 +531,15 @@ class ModulesTest(EnhancedTestCase):
         os.environ['MODULEPATH'] = test2
         modtool.check_module_path()
         self.assertEqual(modtool.mod_paths, [mod_install_dir, test1, test2])
-        self.assertEqual(os.environ['MODULEPATH'], mod_install_dir + ':' + test1 + ':' + test2)
+        self.assertEqual(os.environ['MODULEPATH'], os.pathsep.join([mod_install_dir, test1, test2]))
 
         # check behaviour if non-existing directories are included in $MODULEPATH
         os.environ['MODULEPATH'] = '%s:/does/not/exist:%s' % (test3, test2)
         modtool.check_module_path()
         # non-existing dir is filtered from mod_paths, but stays in $MODULEPATH
         self.assertEqual(modtool.mod_paths, [mod_install_dir, test1, test3, test2])
-        self.assertEqual(os.environ['MODULEPATH'], ':'.join([mod_install_dir, test1, test3, '/does/not/exist', test2]))
+        self.assertEqual(os.environ['MODULEPATH'],
+                         os.pathsep.join([mod_install_dir, test1, test3, '/does/not/exist', test2]))
 
     def test_check_module_path_hmns(self):
         """Test behaviour of check_module_path with HierarchicalMNS."""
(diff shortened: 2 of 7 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2155,7 +2155,7 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
         curr_modpaths = curr_module_paths()
         for init_modpath in init_modpaths:
             full_mod_path = os.path.join(self.installdir_mod, init_modpath)
-            if full_mod_path not in curr_modpaths:
+            if os.path.exists(full_mod_path) and full_mod_path not in curr_modpaths:
```
akesandgren, line 2158: Are you sure that this doesn't cause problems when starting from an empty installation.
Consider the situation of a new install, there is just the easybuild-prefix directory, no modules in the HMNS and no module tree, not even $prefix/modules. Bootstrap easybuild into /scratch and use that eb to install foss.
Would this change then cause problems?

test/framework/modules.py
```diff
@@ -1091,6 +1091,7 @@ def test_module_caches(self):
 
         # purposely extending $MODULEPATH with non-existing path, should be handled fine
         nonpath = os.path.join(self.test_prefix, 'nosuchfileordirectory')
+        mkdir(nonpath)
```
akesandgren, line 1094: The comment above no longer makes sense since it is no longer a non-existing path.

easybuild/framework/easyblock.py
Flamefire: I'm pretty sure, yes. This is done in the prepare step of each installation. If the path does not (yet) exist, then it is of no use for the current EC. And for the next EC (when the path does exist) this check will return true and it will work as usual.

However I just tested it with no MODULEPATH set at all and successfully installed the full foss/2019b tree from scratch (EB MNS, not HMNS but probably doesn't matter, does it?)

test/framework/modules.py
```diff
@@ -1091,6 +1091,7 @@ def test_module_caches(self):
 
         # purposely extending $MODULEPATH with non-existing path, should be handled fine
```
Flamefire, line 1092:
```suggestion
        # purposely extending $MODULEPATH with an empty path, should be handled fine
```

easybuild/framework/easyblock.py
akesandgren: It's HMNS I'm worried about, not standard MNS. Since there it matters for modules/all/Core, could you run such a check too?

Flamefire: Sure. For reference: Starting with MODULEPATH only containing the EB-devel module in $HOME/..., empty installpath and `eb foss-2019b.eb -r --installpath=/tmp/ebinstall --module-naming-scheme=HierarchicalMNS`

akesandgren: ok then, if that works i guess we can move foreward here

Flamefire: Yes, currently at "Compiler/GCCcore/8.3.0/Perl/5.30.0", so I'd say it does work

test/framework/modules.py
```diff
@@ -1223,17 +1229,66 @@ def test_module_use_unuse(self):
...
+        self.assertEqual(os.environ['MODULEPATH'], test_dir2 + ':' + test_dir1)
+        # Adding the same path does not change the path
+        self.modtool.add_module_path(test_dir1)
+        self.assertEqual(os.environ['MODULEPATH'], test_dir2 + ':' + test_dir1)
+        self.modtool.add_module_path(test_dir2)
+        self.assertEqual(os.environ['MODULEPATH'], test_dir2 + ':' + test_dir1)
+        # Even when a (meaningless) slash is added
+        # This occurs when using an empty modules directory name
+        self.modtool.add_module_path(os.path.join(test_dir1, ''))
+        self.assertEqual(os.environ['MODULEPATH'], test_dir2 + ':' + test_dir1)
```
akesandgren, line 1250: These should really use join using os.pathsep for test_dir2 + ':' + test_dir1
Same thing below. And perhaps a var that contains that join instead of repeating it.

test/framework/modules.py
```diff
@@ -1223,17 +1229,66 @@ def test_module_use_unuse(self):
...
+        # And with empty last part
+        self.modtool.remove_module_path(os.path.join(test_dir1, ''))
+        self.assertEqual(os.environ.get('MODULEPATH', ''), '')
+
+        # And with some more trickery
+        # Lmod seems to remove empty paths: /foo//bar/. -> /foo/bar
+        # Environment-Modules 4.x seems to resolve relative paths: /foo/../foo -> /foo
+        # Hence we can only check the real paths
+        def get_resolved_module_path():
+            return ':'.join(os.path.realpath(p) for p in os.environ['MODULEPATH'].split(':'))
```
akesandgren, line 1267: This should also be os.pathsep and not ':'

akesandgren (changes requested):

test/framework/modules.py
Flamefire: Makes sense. Done

akesandgren (approved):
LGTM
Still more places where we should do os.pathsep.join and os.path.join but let's leave that for another cleanup PR someday.

easybuild/framework/easyblock.py
Flamefire: Finished successfully

akesandgren:
Going in, thanks @Flamefire!

PR #3644, opened 2021-04-09
Catch problems early on if --github-user is not specified for --new-pr & co

### LINKED ISSUE #3642: `--new-pr` does not check for valid GitHub user name
I was just opening an easyconfig PR and noticed that `--new-pr` doesn't seem to check that there is a valid GitHub user name set (any more?..EB v4.3.4) and tries to use `None`

```
alanc@~$ eb --new-pr line-prof.eb
== Temporary log file in case of crash /tmp/eb-za20dzv9/easybuild-vg4d9x65.log
== fetching branch 'develop' from https://github.com/easybuilders/easybuild-easyconfigs.git...
== copying files to /tmp/eb-za20dzv9/git-working-dirdxszylm8/easybuild-easyconfigs...
== pushing branch '20210409123147_new_pr_line_profiler310' to remote 'github_None_VeRJa' (git@github.com:None/easybuild-easyconfigs.git)
ERROR: Failed to push branch '20210409123147_new_pr_line_profiler310' to GitHub (git@github.com:None/easybuild-easyconfigs.git): Cmd('git') failed due to: exit code(128)
  cmdline: git push --porcelain github_None_VeRJa 20210409123147_new_pr_line_profiler310
  stderr: 'fatal: Could not read from remote repository.
```

ocaisa:
Fixes #3642 but probably could error earlier for the case of `--new-pr` rather than waste time cloning the repo

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -688,8 +688,10 @@ def setup_repo_from(git_repo, github_url, target_account, branch_name, silent=Fa
     # git fetch
     # can't use --depth to only fetch a shallow copy, since pushing to another repo from a shallow copy doesn't work
     print_msg("fetching branch '%s' from %s..." % (branch_name, github_url), silent=silent)
+    res = None
     try:
-        res = origin.fetch()
+        if target_account != None:
+            res = origin.fetch()
     except GitCommandError as err:
         raise EasyBuildError("Failed to fetch branch '%s' from %s: %s", branch_name, github_url, err)
     if res:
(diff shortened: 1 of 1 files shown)


easybuild/tools/github.py
```diff
@@ -971,21 +973,24 @@ def push_branch_to_github(git_repo, target_account, target_repo, branch):
     if dry_run:
         print_msg(push_branch_msg + ' [DRY RUN]', log=_log)
     else:
-        print_msg(push_branch_msg, log=_log)
-        try:
-            res = remote.push(branch)
-        except GitCommandError as err:
-            raise EasyBuildError("Failed to push branch '%s' to GitHub (%s): %s", branch, github_url, err)
+        if target_account is None:
+            raise EasyBuildError("No valid GitHub username (--github-user) given, pushing branch will fail!")
```
boegel, line 977: `target_account` could also be set some other way, there's no direct relation to `--github-user` here, so this error shouldn't be raised here...

boegel (changes requested):

easybuild/tools/github.py
boegel: see https://github.com/ocaisa/easybuild-framework/pull/46

ocaisa: Merged

boegel (approved):

PR #3645, opened 2021-04-09
Don't skip sanity check for --module-only --rebuild

Flamefire:
The only difference to now us that the sanity check is run when `--rebuild` is used.
For "proper" use, i.e. regenerating a module, this does not cause any problems and might even show issues that user was not aware of.
For those that used it for something else, the sanity check can fail and those need to use `--force` instead.
So that is IMO very low impact.

I'd even say it is a bugfix as the comment states: "allow skipping sanity check too when only generating module and force is used"

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3096,7 +3096,7 @@ class EasyBlock(object):
     def skip_step(self, step, skippable):
         """Dedice whether or not to skip the specified step."""
         module_only = build_option('module_only')
-        force = build_option('force') or build_option('rebuild')
+        force = build_option('force')
         skip = False
 
         # under --skip, sanity check is not skipped
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -1583,16 +1583,11 @@ class ToyBuildTest(EnhancedTestCase):
 
         os.remove(toy_mod)
 
-        # --module-only --rebuild should be equivalent with --module-only --force
-        self.eb_main(args + ['--rebuild'], do_build=True, raise_error=True)
-        self.assertTrue(os.path.exists(toy_mod))
-
-        # make sure load statements for dependencies are included in additional module file generated with --module-only
-        modtxt = read_file(toy_mod)
-        self.assertTrue(re.search('load.*intel/2018a', modtxt), "load statement for intel/2018a found in module")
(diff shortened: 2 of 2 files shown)


smoors:
I'm in favor of this change, it's indeed useful to still (re)do the sanity checks with `--module-only`.
it would be nice if the same could be done for extensions, as the sanity checks for those are also skipped currently.

boegel:
@smoors The sanity check for extensions are always skipped when using `--module-only` right now (even without `--force` or `--rebuild`), which is sort of a feature (although you could argue it's a bug).

If we change/fix that, we would definitely need an `--skip-exts` or something too.

ocaisa:
I'm a bit confused by this because I spent all day yesterday using `--module-only --rebuild` exactly because it does the sanity checks.

Flamefire:
@ocaisa : What exactly is the behavior you want? NOT running the sanity check when module-only is used? You can use `--module-only --force` for that. Everything else (e.g. `--module-only`) does run the sanity check (after this change also for `--module-only --rebuild`)
Maybe bring it up in the conf call in a few mins?

ocaisa:
> @ocaisa : What exactly is the behavior you want? NOT running the sanity check when module-only is used? You can use `--module-only --force` for that. Everything else (e.g. `--module-only`) does run the sanity check (after this change also for `--module-only --rebuild`)
> Maybe bring it up in the conf call in a few mins?

I updated my comment, I was wrong in what I originally said. I definitely agree that running the sanity check should be done when combining `--module-only` and `--rebuild`, see https://github.com/easybuilders/easybuild-easyconfigs/issues/12760 for why I think this would be useful

ocaisa (approved):

ocaisa:
@Flamefire Do you think you will implement the `--skip-exts` option? For me this is a nice to have and if you don't have time we can create an issue for it.

Personally I believe you should run the sanity checks for extensions too. The R extensions in `R/4.0.0-foss-2020a` have changed multiple times, so if you built with 4.3.2 but do a rebuild with 4.3.4, the module will report modules that are in fact not installed.

Flamefire:
@ocaisa Makes sense to run the sanity check for extensions too. Not sure if or when I have time for that, so do create an issue for this and if someone works on that soon, then the better

PR #3648, opened 2021-04-10
Correct cryptography comment in requirements.txt

branfosj:
This makes the comment consistent with that for the other packages above and that means that we no longer have the mismatch between the version being installed and the comment (which #3643 introduced).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -53,7 +53,7 @@ requests
 
 archspec; python_version >= '2.7'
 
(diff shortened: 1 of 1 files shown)


ocaisa (approved):

boegel:
Ah, totally overlooked this... I hate it when bots only do a half-ass job. ;)

Thanks for catching this @branfosj!

PR #3650, opened 2021-04-13
Allow for overriding RPATH with higher priority paths

ocaisa:
This addresses https://github.com/EESSI/software-layer/issues/88 where EESSI wants the ability to override what is linked in rpath (for example to inject an ABI compatible MPI).


easybuild/framework/easyblock.py
```diff
@@ -2144,6 +2144,24 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
...
+                _log.debug("Converted RPATH override directories ('%s') to a list of paths: %s" % (rpath_overrides,
+                                                                                                   rpath_override_dirs))
+                for path in rpath_override_dirs:
+                    if not os.path.isabs(path):
+                        raise EasyBuildError(
+                            "Path used in rpath_override_dirs is not an absolute path: %s", path)
+            else:
+                raise EasyBuildError("Value for rpath_override_dirs has invalid type (%s), should be string: %s",
+                                     type(rpath_overrides), rpath_overrides)
+            self.rpath_include_dirs.extend(rpath_override_dirs)
```
boegel, line 2163: @ocaisa Don't we want to add these paths at the very front, even before `<installdir>/lib` and `$ORIGIN` & co?

With the `/opt/eessi/lib` use case in mind, we want to *always* pick up libraries from there, even for binaries that come with that library in the corresponding installation directory?

boegel (changes requested):

easybuild/framework/easyblock.py
ocaisa: Yeah, I struggled with that and went the other way, assuming that if you are inside the installation then you want to stay consistent...but I don't know, you could argue either way

boegel: Let's take an MPICH library as an example...

If the MPICH installation comes with binaries that link to `libmpi`, do you want to stick with the `libmpi.so` that comes with the MPICH installation, or would you want to use the alternate `libmpi.so` you provide in (for example) `/opt/eessi/lib`?

I'd say it's the latter.

It's certainly an edge case though...

ocaisa: Ok, I reversed it

boegel (approved):

PR #3651, opened 2021-04-14
Allow amending easyconfig parameters which are not the default

Flamefire:
Overwriting parameters which are not in the EC and not one of a few standard ones is currently impossible.
However it is useful when e.g. EasyBlocks introduce new params where the user wants to overwrite the default value without changing the EC manually.
This change allows this by simply appending the new parameter when it is not in the default list.

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -330,7 +330,19 @@ def tweak_one(orig_ec, tweaked_ec, tweaks, targetdir=None):
             tweaks.pop(key)
 
     # add parameters or replace existing ones
+    special_values = ('True', 'False', 'None')
+    quoted_special_values = ['"%s"' % val for val in special_values] + ["'%s'" % val for val in special_values]
     for (key, val) in tweaks.items():
+        # if the value is True/False/None then take that
+        # if e.g. (literal) True is wanted, then it can be passed as "True"/'True'
+        if val in special_values:
+            str_val = val
+            val = eval(val)
+        elif val in quoted_special_values:
+            str_val = val
+            val = val.strip('"\'')
+        else:
+            str_val = quote_str(val)
 
         regexp = re.compile(r"^(?P<key>\s*%s)\s*=\s*(?P<val>.*)$" % key, re.M)
         _log.debug("Regexp pattern for replacing '%s': %s" % (key, regexp.pattern))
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyconfig/tweak.py
```diff
@@ -324,13 +324,25 @@ def __repr__(self):
...
+        # if the value is True/False/None then take that
+        # if e.g. (literal) True is wanted, then it can be passed as "True"/'True'
+        if val in special_values:
+            str_val = val
+            val = eval(val)
+        elif val in quoted_special_values:
+            str_val = val
+            val = val.strip('"\'')
+        else:
+            str_val = quote_str(val)
```
smoors, line 345: this can be simplified a bit:
```suggestion
    special_values = {
        'True': True, "'True'": 'True', '"True"': 'True',
        'False': False, "'False'": 'False', '"False"': 'False',
        'None': None, "'None'": 'None', '"None"': 'None',
    }
    for (key, val) in tweaks.items():
        # if the value is True/False/None then take that
        # if e.g. (literal) True is wanted, then it can be passed as "True"/'True'
        if val in special_values:
            str_val = val
            val = special_values[val]
        else:
            str_val = quote_str(val)
```

test/framework/easyconfig.py
```diff
@@ -839,13 +857,6 @@ def test_obtain_easyconfig(self):
         self.assertEqual(ec['start_dir'], specs['start_dir'])
         remove_file(res[1])
 
-        specs.update({
-            'foo': 'bar123'
-        })
-        self.assertErrorRegex(EasyBuildError, "Unknown easyconfig parameter: foo",
```
smoors, line 845: I would think this test is still useful for parameters that are not defined in the easyblock, or is that taken care of somewhere else?

Flamefire: That is taken care of elsewhere. I don't think we can trigger this error anymore at all, as that is specific to using non-default parameters which we allow now.
The EasyBlock instantiation will trigger e.g. "Use of 2 unknown easyconfig parameters detected in test.eb: foo, some_list", see test_fix_deprecated_easyconfigs

Flamefire:
@smoors Took your suggestion but had to add an is-string check or it will fail for unhashable types (e.g. other dicts) as values. I also reordered the dict and interspersed the comments which IMO is easier to understand. Check if you agree please.

smoors (approved):
lgtm

smoors:
Going in, thanks @Flamefire!

PR #3655, opened 2021-04-17
add support for --sanity-check-only + also run sanity check for extensions when using --module-only

boegel:


easybuild/framework/easyblock.py
```diff
@@ -3119,9 +3130,14 @@ def skip_step(self, step, skippable):
...
             skip = True
 
+        elif sanity_check_only and step != SANITYCHECK_STEP:
+            self.log.info("Skipping %s step because of sanity-check-only mode", step)
+            skip = True
+
         else:
-            self.log.debug("Not skipping %s step (skippable: %s, skip: %s, skipsteps: %s, module_only: %s, force: %s",
-                           step, skippable, self.skip, self.cfg['skipsteps'], module_only, force)
+            msg = "Not skipping %s step (skippable: %s, skip: %s, skipsteps: %s, module_only: %s, force: %s, "
```
bartoldeman, line 3138: just wondering, for the `msg +=` is there any style that's better, e.g. 
```
msg = ("Not skipping %s step (skippable: %s, skip: %s, skipsteps: %s, module_only: %s, force: %s, "
            "sanity_check_only: %s)")
```
or  using `''.join()` or something else?

Also there is a `)` missing in the string itself, behind the last `%s`.

bartoldeman (changes requested):

boegel:
This probably needs more work to ensure that sanity checks for extensions are also done when using `eb --sanity-check-only`...

ocaisa:
> This probably needs more work to ensure that sanity checks for extensions are also done when using `eb --sanity-check-only`...

Yes, I just noticed that in general sanity checks for extensions are not run when using `--module-only` so this is a more general shortcoming.

easybuild/framework/easyblock.py
```diff
@@ -3119,9 +3130,14 @@ def skip_step(self, step, skippable):
             self.log.info("Skipping %s step because of forced module-only mode", step)
             skip = True
 
+        elif sanity_check_only and step != SANITYCHECK_STEP:
+            self.log.info("Skipping %s step because of sanity-check-only mode", step)
+            skip = True
```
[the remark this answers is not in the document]
boegel: I prefer to assume that there won't be any skipping, and only setting `skip` to `True` when we do want to skip.

Seems less likely to introduce bugs like that to me...

easybuild/framework/easyblock.py
boegel: We use the `+=` approach all throughout the code, I think it's clearer sticking to that...

easybuild/framework/easyblock.py
```diff
@@ -2332,6 +2309,44 @@ def extensions_step(self, fetch=False, install=True):
...
+
+        self.prepare_for_extensions()
+
+        if fetch:
+            self.exts = self.fetch_extension_sources()
+
+        self.exts_all = self.exts[:]  # retain a copy of all extensions, regardless of filtering/skipping
+
+        # actually install extensions
+        self.log.info("Installing extensions")
```
ocaisa, line 2341: I guess this should be guarded?
```suggestion
        if install:
            self.log.info("Installing extensions")
```

easybuild/framework/easyblock.py
```diff
@@ -3281,9 +3315,14 @@ def skip_step(self, step, skippable):
...
 
+        elif sanity_check_only and step != SANITYCHECK_STEP:
+            self.log.info("Skipping %s step because of sanity-check-only mode", step)
+            skip = True
+
         else:
-            self.log.debug("Not skipping %s step (skippable: %s, skip: %s, skipsteps: %s, module_only: %s, force: %s",
-                           step, skippable, self.skip, self.cfg['skipsteps'], module_only, force)
+            msg = "Not skipping %s step (skippable: %s, skip: %s, skipsteps: %s, module_only: %s, force: %s, "
+            msg += "sanity_check_only: %s"
```
akesandgren, line 3324: Still missing the end ")" in that msg string

akesandgren (changes requested):

akesandgren (approved):
LGTM

test/framework/toy_build.py
```diff
@@ -1682,6 +1682,58 @@ def test_module_only(self):
...
+            "exts_list = [",
+            "    ('barbar', '0.0', {",
+            "        'start_dir': 'src',",
+            "        'exts_filter': ('ls -l lib/lib%(ext_name)s.a', ''),",
+            "    })",
+            "]",
+        ])
+        write_file(test_ec, test_ec_txt)
+
+        # clean up $MODULEPATH to only modules in test prefix dir are found
```
ocaisa, line 1712:
```suggestion
        # clean up $MODULEPATH so only modules in test prefix dir are found
```

ocaisa (changes requested):
LGTM, typo in a comment

PR #3656, opened 2021-04-18
add support for running commands asynchronously via run_cmd + complete_cmd functions

boegel:
This is a required step towards support for installing extensions in parallel...

This is mostly breaking up the current `run_cmd` into `run_cmd` + `complete_cmd`.
In combination with `get_output_from_process`, this is sufficient to run installation commands for extensions in the background...


bartoldeman (approved):
LGTM

PR #3657, opened 2021-04-20
Improve logging when failing to load class from exts_classmap

Flamefire:
Posted in Slack:

    TypeError: __init__() takes 2 positional arguments but 3 were given

This was not caught and "enhanced" by the information which class wastried to load.
This change catches the base Exception class, which includes TypeError.
It also logs the class found before instantiating it.


bartoldeman (approved):
LGTM

PR #3659, opened 2021-04-26
Allow module-only on existing installations (without write permissions)

ocaisa:
While attempting to rebuild a module tree from EESSI, I realised there are a couple of barriers in the code right now that don't permit this. As things currently stand `--module-only` requires write permissions on the existing installation (for log files and other goodies), this PR fixes this.


smoors:
@ocaisa do you think it's doable/useful to make a test for this?

ocaisa:
@smoors I was afraid someone was going to ask that, I'll see if I can figure it out

ocaisa:
@smoors The test I added fails on current `develop`. Not sure about using `subprocess` but the `os` option is a lot of extra lines of code

smoors (approved):
lgtm

smoors:
Going in, thanks @ocaisa!

PR #3660, opened 2021-04-27
Correct help text for subdir-user-modules

Flamefire:
Fix the name of the referenced option


ocaisa (approved):
Thanks for catching this!

PR #3661, opened 2021-04-27
Replace log_error parameter of which() by on_error

Flamefire:
This allows to additionally make the warning an error instead of only being able to ignore it.
This is useful if the command must be found.


smoors (approved):
lgtm

smoors:
Going in, thanks @Flamefire!

PR #3664, opened 2021-04-28
Make which(on_error change backward compatible with log_error=False

Flamefire:
Fix for #3661 which led to `log_error=False` getting ignored

Uses `None` now to detect if the new param was set.

@smoors


smoors (approved):
lgtm

smoors:
Going in, thanks @Flamefire!

PR #3665, opened 2021-04-28
add support for using oneAPI versions of 'intel' toolchain components

boegel:
The `iimpi` toolchain definition had to be enhanced to discriminate between the oneAPI generation (`iimpi` version 2020.12 and newer), and the "classic" generation. No changes are needed for the `intel` toolchain itself (since that just derives from `iimpi` and adds MKL).

The `det_subtoolchain_version` function was enhanced to accept alternative subtoolchain names, since we don't know whether `intel-compilers` (for the oneAPI generation) or `iccifort` is the actual subtoolchain for `iimpi` until we can check the version, which we can't until there's an actual *instance* of the `Toochain` class...

Other toolchains that use the Intel compilers will need to be tweaked in a similar way, but `iimpi` is the more important one in the short term (for `intel/2021a`).

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -220,7 +220,7 @@ def toolchain_hierarchy_cache(func):
     return cache_aware_func
 
 
-def det_subtoolchain_version(current_tc, subtoolchain_name, optional_toolchains, cands, incl_capabilities=False):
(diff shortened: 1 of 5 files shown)


migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3667, opened 2021-04-28
add support for installing extensions in parallel (WIP)

boegel:
WIP because needs enhancements in the test suite + some more testing with R


boegel:
Tests in latest commit are failing because `update_progress_bar` gets called without calling `start_progress_bar` first.

To fix that without jumping through lots of code hoops, I've proposed some trivial changes in https://github.com/easybuilders/easybuild-framework/pull/3867...

PR #3668, opened 2021-04-28
move disable_templating function into the EasyConfig class

Flamefire:
Allows easier use from outside as e.g. a few tests and easyblocks use it
The free function now just redirects to the member function

This also changes all uses if `ec.enable_templating = False` to use the context manager which avoids "forgetting" to reset it on error


boegel (approved):
lgtm

PR #3669, opened 2021-05-03
allow tweaking of easyconfigs from different toolchains

Flamefire:
This is useful to bulk-install e.g. Python with enable_lto=False set for all ECs.
The restriction is only required when changing the toolchain so move the check down and add a simple test.


easybuild/framework/easyconfig/tweak.py
```diff
@@ -115,6 +107,16 @@ def tweak(easyconfigs, build_specs, modtool, targetdirs=None):
                 revert_to_regex = True
 
         if not revert_to_regex:
+            # make sure easyconfigs all feature the same toolchain (otherwise we *will* run into trouble)
```
bartoldeman, line 110: can you explain why this only applies when `revert_to_regex = False` ?

bartoldeman (changes requested):

easybuild/framework/easyconfig/tweak.py
Flamefire: TBH: I don't know why this restriction really exists.
The only reason I could find is the `source_toolchain` which must (by definition) be a single toolchain. (It is set to the first toolchain found assuming all are the same.) This is only used in this code part, similar `target_toolchain` is also only used here. So after moving those variables down into this codepath (to limit their scope as much as possible) I see no reason to disallow multiple toolchains for the other code paths

On a more general explanation: This code path is used when the toolchain is about to be changed AND map-toolchains is enabled. I can imagine, how this can fail, when multiple toolchains are in effect as the mapping will be much harder.
On the other hand when tweaking ECs via regex-replacement (i.e. NOT this codepath) then all this does not matter: A regex-substitution is a text operation and hence semantic concepts like toolchains do not matter.

FWIW: I tested this on a single install of about 8 different Python from 4 different toolchain generations where I used try-amend to change an EC param and it worked well.

akesandgren:
This looks sane to me, but I think @boegel needs to take a look at it too...

Flamefire:
Rebased again

akesandgren:
@boegel I think your eyes are needed on this

Flamefire:
ping

ocaisa (approved):
I think this is ok, and the test is there

easybuild/framework/easyconfig/tweak.py
ocaisa: I think the reasoning is fine here, it's dangerous to have multiple toolchains in play when we need to deal with a hierarchy, but that won't be the case for a regex tweak

ocaisa:
@Flamefire Not sure why the full test suite hasn't run on this, it won't let me merge as is

Flamefire:
> @Flamefire Not sure why the full test suite hasn't run on this, it won't let me merge as is

I guess some required checks have since been removed. I merged develop to restart CI and see if that solves it.

PR #3670, opened 2021-05-04
add support for using ARCH constant in easyconfig files

boegel:


test/framework/options.py
```diff
@@ -492,12 +492,16 @@ def run_test(fmt=None):
...
                 pattern_lines = [
+                    r'^``ARCH``\s*``(aarch64|ppc64le|x86_64)``\s*CPU architecture .*',
+                    r'^``EXTERNAL_MODULE``.*',
                     r'^``HOME``.*',
                     r'``OS_NAME``.*',
                     r'``OS_PKG_IBVERBS_DEV``.*',
                 ]
             else:
                 pattern_lines = [
+                    r'^\s*ARCH: (aarch64|ppc64le|x86_64) \(CPU architecture .*',
```
Micket, line 503: I'll admit I don't understand where these patterns occur, but it does look like an unmatched `\)` here?

bartoldeman: I think missing the `\)` is probably sufficient to match but it would look nicer and tighter with the `\)`

boegel: The `\)` (an escaped `)` is not required thanks to the `.*`, but I can use `.*\)` to avoid confusion, sure...

bartoldeman (approved):
LGTM

PR #3674, opened 2021-05-12
honor keyboard interrupt in 'eb' command

boegel:
Quickly cancelling an `eb` command right after starting it currently requires hitting `Ctrl-C` multiple times in a row, because you're only cancelling the initial Python version check, and there are multiple in a row.

We can trap the `SIGINT` signal that occurs when hitting `Ctrl-C` though, and honor that, so a single `Ctrl-C` is always sufficient to cancel the `eb` command.

### DIFF
--- a/eb
+++ b/eb
@@ -33,6 +33,13 @@
 # @author: Pieter De Baets (Ghent University)
 # @author: Jens Timmerman (Ghent University)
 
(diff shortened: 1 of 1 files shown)


ocaisa (approved):
Tested and seems to work fine

PR #3675, opened 2021-05-13
Pin GitPython version for Python<3.6, don't test bootstrap script against Python 3.8/3.9

ocaisa:


ocaisa:
@boegel Now the unit tests but the bootstrap is failing 😢

ocaisa:
Version bumps in Python 3.8 and 3.9 since yesterday that are causing the failures

ocaisa:
```
Warning
Easy Install is deprecated. Do not use it. Instead use pip. If you think you need Easy Install, please reach out to the PyPA team (a ticket to pip or setuptools is fine), describing your use-case.
```

boegel (approved):
lgtm

PR #3677, opened 2021-05-14
add toolchain definition for Fujitsu ARM toolchain

migueldiascosta:
being tested at Fugaku

### DIFF
--- a/easybuild/toolchains/compiler/fujitsu.py
+++ b/easybuild/toolchains/compiler/fujitsu.py
@@ -76,5 +76,6 @@ class FujitsuCompiler(Compiler):
     def _set_compiler_vars(self):
         super(FujitsuCompiler, self)._set_compiler_vars()
 
+        # enable clang compatibility mode
         self.variables.nappend('CFLAGS', ['Nclang'])
         self.variables.nappend('CXXFLAGS', ['Nclang'])
--- a/easybuild/toolchains/compiler/fujitsu.py
+++ b/easybuild/toolchains/compiler/fujitsu.py
@@ -29,18 +29,22 @@ The basic concept is the same as for the Cray Programming Environment.
 
 :author: Miguel Dias Costa (National University of Singapore)
 """
+import os
+
+import easybuild.tools.environment as env
 import easybuild.tools.systemtools as systemtools
 from easybuild.tools.toolchain.compiler import Compiler, DEFAULT_OPT_LEVEL
 
 TC_CONSTANT_FUJITSU = 'Fujitsu'
+TC_CONSTANT_MODULE_NAME = 'lang'
+TC_CONSTANT_MODULE_VAR = 'FJSVXTCLANGA'
 
 
 class FujitsuCompiler(Compiler):
     """Generic support for using Fujitsu compiler drivers."""
     TOOLCHAIN_FAMILY = TC_CONSTANT_FUJITSU
 
-    # compiler module name is lang (with version e.g. tcsds-1.2.31)
-    COMPILER_MODULE_NAME = ['lang']
+    COMPILER_MODULE_NAME = [TC_CONSTANT_MODULE_NAME]
     COMPILER_FAMILY = TC_CONSTANT_FUJITSU
 
     # make sure fcc is always called in clang compatibility mode
@@ -81,6 +85,14 @@ class FujitsuCompiler(Compiler):
     COMPILER_GENERIC_OPTION = {
         (systemtools.AARCH64, systemtools.ARM): '-mcpu=generic -mtune=generic',
     }
+
+    def prepare(self, *args, **kwargs):
+        super(FujitsuCompiler, self).prepare(*args, **kwargs)
+
+        # make sure the fujitsu module libraries are found (and added to rpath by wrapper)
+        libdir = os.path.join(os.getenv(TC_CONSTANT_MODULE_VAR), 'lib64')
+        self.log.debug("Adding %s to $LIBRARY_PATH" % libdir)
+        env.setvar('LIBRARY_PATH', os.pathsep.join([os.getenv('LIBRARY_PATH'), libdir]))
 
     def _set_compiler_vars(self):
         super(FujitsuCompiler, self)._set_compiler_vars()
--- a/easybuild/toolchains/compiler/fujitsu.py
+++ b/easybuild/toolchains/compiler/fujitsu.py
@@ -85,6 +85,7 @@ class FujitsuCompiler(Compiler):
     def _set_compiler_vars(self):
         super(FujitsuCompiler, self)._set_compiler_vars()
 
-        # enable clang compatibility mode; moved to compiler constants to make sure it is always used
+        # enable clang compatibility mode
+        # moved to compiler constants to make sure it is always used
         # self.variables.nappend('CFLAGS', ['Nclang'])
         # self.variables.nappend('CXXFLAGS', ['Nclang'])
--- a/easybuild/toolchains/compiler/fujitsu.py
+++ b/easybuild/toolchains/compiler/fujitsu.py
@@ -43,17 +43,19 @@ class FujitsuCompiler(Compiler):
     COMPILER_MODULE_NAME = ['lang']
     COMPILER_FAMILY = TC_CONSTANT_FUJITSU
 
+    # make sure fcc is always called in clang compatibility mode
     COMPILER_CC = 'fcc -Nclang'
     COMPILER_CXX = 'FCC -Nclang'
+
     COMPILER_F77 = 'frt'
     COMPILER_F90 = 'frt'
     COMPILER_FC = 'frt'
 
     COMPILER_UNIQUE_OPTION_MAP = {
-        DEFAULT_OPT_LEVEL: 'O2 -Kfast',
+        DEFAULT_OPT_LEVEL: 'O2',
         'lowopt': 'O1',
         'noopt': 'O0',
-        'opt': 'O3 -Kfast',
+        'opt': 'Kfast',  # -O3 -Keval,fast_matmul,fp_contract,fp_relaxed,fz,ilfunc,mfunc,omitfp,simd_packed_promotion
         'optarch': '',  # Fujitsu compiler by default generates code for the arch it is running on
         'openmp': 'Kopenmp',
         'unroll': 'funroll-loops',
@@ -83,6 +85,6 @@ class FujitsuCompiler(Compiler):
     def _set_compiler_vars(self):
         super(FujitsuCompiler, self)._set_compiler_vars()
 
-        # enable clang compatibility mode
+        # enable clang compatibility mode; moved to compiler constants to make sure it is always used
         # self.variables.nappend('CFLAGS', ['Nclang'])
         # self.variables.nappend('CXXFLAGS', ['Nclang'])
--- a/easybuild/toolchains/compiler/fujitsu.py
+++ b/easybuild/toolchains/compiler/fujitsu.py
@@ -1,0 +1,80 @@
+##
+# Copyright 2014-2021 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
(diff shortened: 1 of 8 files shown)


easybuild/toolchains/fft/fftw.py
```diff
@@ -39,15 +39,18 @@ class Fftw(Fft):
...
-        elif LooseVersion(version) > LooseVersion('2'):
-            suffix = '3'
+        suffix = self.FFTW_API_VERSION
+        if not suffix:
+            version = self.get_software_version(self.FFT_MODULE_NAME)[0]
+            if LooseVersion(version) < LooseVersion('2') or LooseVersion(version) >= LooseVersion('4'):
+                raise EasyBuildError("_set_fft_variables: FFTW unsupported version %s (major should be 2 or 3)",
+                                     version)
+            elif LooseVersion(version) > LooseVersion('2'):
+                suffix = '3'
```
migueldiascosta, line 53: this is the only change in existing files in this PR. It's meant to reduce code duplication by allowing `FujitsuFFTW` to derive from `Fftw`, and it seems innocuous to me (and eventually even useful for other fftw variants that don't follow the API version, as the AMD one did for a while), but I can also make `FujitsuFFTW` derive from `Fft` instead and duplicate most of `Fftw` there

boegel: Thanks for pointing this out, I think this approach makes sense.

easybuild/toolchains/ffmpi.py
```diff
@@ -0,0 +1,38 @@
...
+EasyBuild support for fompi compiler toolchain (includes Fujitsu Compiler and MPI).
+
+:author: Miguel Dias Costa (National University of Singapore)
+"""
+from easybuild.toolchains.compiler.fujitsu import FujitsuCompiler
+from easybuild.toolchains.linalg.fujitsussl import FujitsuSSL
+from easybuild.toolchains.mpi.fujitsumpi import FujitsuMPI
+
+
+class ffmpi(FujitsuCompiler, FujitsuMPI, FujitsuSSL):
```
boegel, line 35: class name should always start with a capital letter, so please use `Ffmpi` here

easybuild/toolchains/ffmpi.py
```diff
@@ -0,0 +1,38 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for fompi compiler toolchain (includes Fujitsu Compiler and MPI).
```
boegel, line 26: `fompi` -> `ffmpi`?

easybuild/toolchains/compiler/fujitsu.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # used with --optarch=GENERIC
+    COMPILER_GENERIC_OPTION = {
+        (systemtools.AARCH64, systemtools.ARM): '-mcpu=generic -mtune=generic',
+    }
+
+    def _set_compiler_vars(self):
+        super(FujitsuCompiler, self)._set_compiler_vars()
+
+        self.variables.nappend('CFLAGS', ['Nclang'])
```
boegel, line 79: Is it worth clarifying why these are required with a comment?

easybuild/toolchains/linalg/fujitsussl.py
```diff
@@ -0,0 +1,115 @@
...
+    SCALAPACK_LIB = ['']
+    SCALAPACK_LIB_MT = ['']
+    SCALAPACK_FAMILY = TC_CONSTANT_FUJITSU_SSL
+
+    def _get_software_root(self, name):
+        """Get install prefix for specified software name; special treatment for Fujitsu modules."""
+        if name == 'lang':
+            # Fujitsu-provided  module
+            env_var = 'FJSVXTCLANGA'
+            root = os.getenv(env_var, None)
```
boegel, line 69: `None` is the default fallback value, so this is equivalent:

```python
root = os.getenv(env_var)

File: easybuild/toolchains/mpi/fujitsumpi.py [hunk 6]
```diff
@@ -0,0 +1,66 @@
...
+    MPI_TYPE = TC_CONSTANT_MPI_TYPE_OPENMPI
+
+    # OpenMPI reads from CC etc env variables
+    MPI_SHARED_OPTION_MAP = dict([('_opt_%s' % var, '') for var, _ in MPI_COMPILER_VARIABLES])
+
+    MPI_LINK_INFO_OPTION = '-showme:link'
+
+    def __init__(self, *args, **kwargs):
+        """Toolchain constructor"""
+        super(FujitsuMPI, self).__init__(*args, **kwargs)
```
boegel, line 52: This doesn't do anything, so custom `__init__` can be dropped?

boegel (changes requested):

migueldiascosta:
@boegel see https://github.com/easybuilders/easybuild-easyconfigs/pull/12863#issuecomment-841788384 about the hierarchy and the naming of `ffmpi`, do you have a better suggestion? (e.g. `FujitsuCore`? `FujitsuCoreMPI`?)

basically, this will be the toolchain used for most of what's in both `GCCcore` and `gompi`...

boegel:
I replied in that issue, but to be clear: the `Ffmpi` class name is just an internal thing, and sort of separate with the actual toolchain name (but it's better to keep those in sync, of course).

easybuild/toolchains/compiler/fujitsu.py
```diff
@@ -0,0 +1,81 @@
...
+    COMPILER_F90 = 'frt'
+    COMPILER_FC = 'frt'
+
+    COMPILER_UNIQUE_OPTION_MAP = {
+        DEFAULT_OPT_LEVEL: 'Kfast',
+        'optarch': '',  # Fujitsu compiler by default generates code for the arch it is running on
+        'openmp': 'Kopenmp',
+        'unroll': 'funroll-loops',
+        'strict': ['Kfp_precision'],
+        'precise': ['Kfp_precision'],
```
migueldiascosta, line 58: apparently `-Kfp_precision` doesn't work in clang mode, closest we can get seems to be `-Knoeval,nofast_matmul,nofp_contract,nofp_relaxed,noilfunc` (and at least the tests that require `precise` in `GMP` do pass with these flags)

easybuild/toolchains/linalg/fujitsussl.py
```diff
@@ -0,0 +1,115 @@
...
+                self.variables.nappend(flags_var, ['SSL2BLAMP'])
+        else:
+            for flags_var, _ in COMPILER_FLAGS:
+                self.variables.nappend(flags_var, ['SSL2'])
+
+    def _set_scalapack_variables(self):
+        """Setting FujitsuSSL specific SCALAPACK related variables"""
+        super(FujitsuSSL, self)._set_scalapack_variables()
+        for flags_var, _ in COMPILER_FLAGS:
+            self.variables.nappend(flags_var, ['SCALAPACK'])
```
migueldiascosta, line 93: `-SSL2`/`-SSL2BLAMP` and `-SCALAPACK` are only meant to be used when linking and generate a warning on compiling that is harmless but pollutes the logs

easybuild prepends `-L` to variables appended to `LDFLAGS`, so some refactoring seems to be necessary (?) in order to do this properly (is it worth it?)

easybuild/toolchains/compiler/fujitsu.py
```diff
@@ -0,0 +1,81 @@
...
+    COMPILER_GENERIC_OPTION = {
+        (systemtools.AARCH64, systemtools.ARM): '-mcpu=generic -mtune=generic',
+    }
+
+    def _set_compiler_vars(self):
+        super(FujitsuCompiler, self)._set_compiler_vars()
+
+        # enable clang compatibility mode
+        self.variables.nappend('CFLAGS', ['Nclang'])
+        self.variables.nappend('CXXFLAGS', ['Nclang'])
```
migueldiascosta, line 81: turns out having `-Nclang` only in these flags messes up with configure scripts that probe for the compiler without them, and so detect the traditional mode instead (hitting this with the `libffi` included in `_ctypes` in `Python 2.7.18`)

one solution that seems to work is to move them to `COMPILER_CC` and `COMPILER_CCX` above (but I think then it's crucial to always enclose `$CC` etc. in double quotes, since they will include a space?)

easybuild/toolchains/compiler/fujitsu.py
```diff
@@ -0,0 +1,81 @@
...
+        DEFAULT_OPT_LEVEL: 'Kfast',
+        'optarch': '',  # Fujitsu compiler by default generates code for the arch it is running on
+        'openmp': 'Kopenmp',
+        'unroll': 'funroll-loops',
+        'strict': ['Kfp_precision'],
+        'precise': ['Kfp_precision'],
+        'defaultprec': [],
+        'loose': ['Kfp_relaxed'],
+        'veryloose': ['Kfp_relaxed'],
+        'vectorize': {False: 'KNOSVE', True: 'KSVE'},
```
migueldiascosta, line 62: same as above, these don't work in clang mode...

migueldiascosta:
> I replied in that issue, but to be clear: the `Ffmpi` class name is just an internal thing, and sort of separate with the actual toolchain name (but it's better to keep those in sync, of course).

well, besides the name there is also the hierarchy, the reason I used only two levels is that the module provides compiler, mpi and linalg; 

but even though they are there, easybuild doesn't have to enable all at once, e.g. we can have an additional non-mpi level "below" `Ffmpi` and/or move `FujitsuSSL` to the full toolchain level, would that be clearer?

easybuild/toolchains/compiler/fujitsu.py
```diff
@@ -0,0 +1,88 @@
...
+    COMPILER_CXX = 'FCC -Nclang'
+    COMPILER_F77 = 'frt'
+    COMPILER_F90 = 'frt'
+    COMPILER_FC = 'frt'
+
+    COMPILER_UNIQUE_OPTION_MAP = {
+        DEFAULT_OPT_LEVEL: 'O2 -Kfast',
+        'lowopt': 'O1',
+        'noopt': 'O0',
+        'opt': 'O3 -Kfast',
```
migueldiascosta, line 56: `-Kfast` is actually `-O3` plus a bunch of other stuff, so planning to play it safe and change to:

```python
    DEFAULT_OPT_LEVEL: 'O2',
    'opt': 'Kfast',
```
In addition to the possibility of using `'opt': True` in easyconfigs, `-Kfast` will be explicitly enabled in some easyblocks, e.g. fftw, berkeleygw (, fortranpythonpackage?)

migueldiascosta:
> but even though they are there, easybuild doesn't have to enable all at once, e.g. we can have an additional non-mpi level "below" `Ffmpi` and/or move `FujitsuSSL` to the full toolchain level, would that be clearer?

this does seem clearer (and cleaner), done in https://github.com/migueldiascosta/easybuild-framework/commit/c095190d0dbc848e00c2efbbcd20f91f506bc128

migueldiascosta:
> but even though they are there, easybuild doesn't have to enable all at once, e.g. we can have an additional non-mpi level "below" `Ffmpi` and/or move `FujitsuSSL` to the full toolchain level, would that be clearer?

this does seem clearer (and cleaner), done in https://github.com/migueldiascosta/easybuild-framework/commit/c095190d0dbc848e00c2efbbcd20f91f506bc128

migueldiascosta:
(weird, commit https://github.com/easybuilders/easybuild-framework/pull/3677/commits/c095190d0dbc848e00c2efbbcd20f91f506bc128 only showed up here after I pushed https://github.com/easybuilders/easybuild-framework/pull/3677/commits/558487e4b99485fa38c02a6997cac7396282a267)

easybuild/toolchains/linalg/fujitsussl.py
boegel: @migueldiascosta If you think it's important enough, please open an issue to follow up on this.

boegel (approved):

boegel:
@migueldiascosta I'll assume this is fully functional in the current state (can't check myself, because I don't have access to Fugaku), so I'll go ahead and merge this, mostly because I want to ensure that the changes to `easybuild/toolchains/fft/fftw.py` don't have any negative side effects as we're ramping up to the extensive regression test for EasyBuild v4.4.0.

Follow-up PRs to fix specific issues are welcome, of course! :)

PR #3678, opened 2021-05-14
extend sanity check step to check whether specific libraries are not linked into installed binaries/libraries

boegel:
~~This is still work-in-progress, because it now hardcodes `openblas` as a banned library, which is obviously not the intention...~~

~~We should figure out where the list of required/banned libraries will come from.~~
~~One obvious way is to provide a dedicated configuration option like `--banned-libraries`, but we also need a way to pick up additional banned libraries from the toolchain, for example when using FlexiBLAS (since then all the backend libraries should be banned from linking to directly).~~

This will also be useful to ensure there are no direct links to the OpenSSL libraries provided by the OS (cfr. https://github.com/easybuilders/easybuild-easyconfigs/issues/11895#issuecomment-841301807), or to `libz.so`, `libcurl.so`, etc.

To do:

* [x] make list of required/banned libraries configurable
* [x] let toolchain specify additional required/banned libraries somehow
* [x] tests~~


boegel:
Tests are failing because the new test easyblock for `libtoy` isn't being picked up by the tests, but it's not clear to me why...

migueldiascosta (approved):
lgtm

Micket (commented):
As far as I can tell, this (understandably) requires opt-in, so, it certainly won't help to catch inadvertent linking?

boegel:
> As far as I can tell, this (understandably) requires opt-in, so, it certainly won't help to catch inadvertent linking?

It will help if we opt-in to checking stuff like `/usr/lib/libopenssl.*so.*` & co, on `generoso` for example.

I don't think we should be hardcoding things like this, it should be left as opt-in imho...

It does catch linking the `libopenblas` directly when `FlexiBLAS` is being used as a toolchain component though, see `banned_linked_shared_libs` in `easybuild/toolchains/linalg/flexiblas.py`.

Micket (approved):
lgtm

PR #3679, opened 2021-05-15
tweak foss toolchain definition to switch from OpenBLAS to FlexiBLAS in foss/2021a

boegel:
requires ~~#3678~~ (not really though, at least not for the tests to pass)


migueldiascosta (approved):
lgtm

Issue #3671, opened 2021-05-04
Clearer error message needed for the --check-github command when ssh keys are missing or wrong


rungitta:
I got this error when I had ssh keys that was not added to github. Removing the keys entirely from the host produced the same error.

```bash
palrosen@freebio2$ eb --check-github
== Temporary log file in case of crash /tmp/eb-l0zlpmu7/easybuild-dfa7z5sz.log
Checking status of GitHub integration...
Making sure we're online...OK
* GitHub user...rungitta => OK
Please enter password for encrypted keyring:
* GitHub token...ghp..mqV (len: 40) => OK (validated)
* git command...OK ("git version 2.27.0; ")
* GitPython module...OK (GitPython version 3.1.15)
* push access to rungitta/easybuild-easyconfigs repo @ GitHub...FAIL (unexpected exception: 'Checking out branch \'develop\' from git@github.com:rungitta/easybuild-easyconfigs.git failed: "Failed to fetch branch \'develop\' from git@github.com:rungitta/easybuild-easyconfigs.git: Cmd(\'git\') failed due to: exit code(128)\\n  cmdline: git fetch -v pr_target_account_rungitta_BWIGs\\n  stderr: \'b\'fatal: Could not read from remote repository.\\\\n\\\\nPlease make sure you have the correct access rights\\\\nand the repository exists.\'\'"')
* creating gists...OK
* location to Git working dirs... not found (suboptimal)
One or more checks FAILed, GitHub configuration not fully complete!
See http://easybuild.readthedocs.org/en/latest/Integration_with_GitHub.html#configuration for help.
Status of GitHub integration:
* --from-pr: OK
* --new-pr: not supported
* --review-pr: OK
* --update-pr: not supported
* --upload-test-report: OK
```

boegel:
@migueldiascosta I think we could mention something like `GitHub SSH key missing?` before showing the full error, or perhaps do some pattern matching on "`Could not read from remote repository`" to inject that suggestion?

PR #3681, opened 2021-05-17
suggest missing SSH key when not able to read from remote repository in --check-github

migueldiascosta:
fixes https://github.com/easybuilders/easybuild-framework/issues/3671


boegel (approved):
lgtm

PR #3682, opened 2021-05-17
add locate_solib function to locate Linux shared libraries

boegel:
fleshed out from https://github.com/easybuilders/easybuild-easyblocks/pull/2429


Micket (approved):
lgtm

PR #3683, opened 2021-05-17
add system agnostic function to locate shared libraries

lexming:
taken from https://github.com/easybuilders/easybuild-easyblocks/pull/2429

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -41,6 +41,12 @@ import termios
 from ctypes.util import find_library
 from socket import gethostname
 
+try:
+    # only needed on macOS, may not be available on Linux
+    import ctypes.macholib.dyld
+except ImportError:
+    pass
+
 from easybuild.base import fancylogger
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
@@ -849,6 +855,8 @@ def locate_solib(libobj):
     """
     Return absolute path to loaded library using dlinfo
     Based on https://stackoverflow.com/a/35683698
+
+    :params libobj: ctypes CDLL object
     """
     # early return if we're not on a Linux system
     if get_os_type() != LINUX:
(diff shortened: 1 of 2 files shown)


easybuild/tools/systemtools.py
```diff
@@ -873,6 +881,37 @@ class LINKMAP(ctypes.Structure):
...
+        if os_type == LINUX:
+            # find path to library with dlinfo
+            lib_abspath = locate_solib(lib_obj)
+        elif os_type == DARWIN:
+            # ctypes.macholib.dyld.dyld_find accepts file names and returns full path
+            lib_abspath = ctypes.macholib.dyld.dyld_find(lib_filename)
+        else:
+            raise EasyBuildError("Unknown host OS type: %s", os_type)
+
+        _log.info("Found absolute path to %s: %s" % (lib_filename, lib_abspath))
```
boegel, line 910: Please make this log statement "lazy" (only template when the log message is actually triggered)

```suggestion
      _log.info("Found absolute path to %s: %s", lib_filename, lib_abspath)
```

boegel (changes requested):

test/framework/systemtools.py
```diff
@@ -1052,6 +1052,19 @@ def test_locate_solib(self):
                 self.assertEqual(os.path.basename(libc_path), libname)
                 self.assertTrue(os.path.exists(libc_path), "%s should exist" % libname)
 
+    def test_find_library_path(self):
+        """Test find_library_path function (Linux and Darwin only)."""
+        if get_os_type() == LINUX:
+            libname = 'libc.so.6'
```
boegel, line 1058: This may bite us in the ass if the system on which the test is run doesn't have `libc.so.6`, but I think that's acceptable for now?

easybuild/tools/systemtools.py
Micket: Looks like you unintentionally unindented this suggestion?

lexming: I edited the suggestion by @boegel, fixed :+1:

test/framework/systemtools.py
lexming: Ijust copied your test from `locate_libso` :slightly_smiling_face: 
```
    def test_locate_solib(self):
        """Test locate_solib function (Linux only)."""
            if get_os_type() == LINUX:
                libname = 'libc.so.6'
```

boegel: But I made that test a bit less strict, by allowing that the library is not found. That's not the case here. ;)

test/framework/systemtools.py
```diff
@@ -1052,6 +1052,19 @@ def test_locate_solib(self):
...
+        elif get_os_type() == DARWIN:
+            libname = 'libSystem.dylib'
+        else:
+            libname = None
+
+        if libname:
+            lib_path = find_library_path(libname)
+            self.assertEqual(os.path.basename(lib_path), libname)
+            self.assertTrue(os.path.exists(lib_path), "%s should exist" % libname)
```
boegel, line 1068: @lexming Style check is failing, there should be two empty lines above `def suite`... 🙄

boegel (changes requested):

test/framework/systemtools.py
boegel: fixed it myself :)

test/framework/systemtools.py
boegel: Let's leave this as is, we can try and come up with a better mechanism if we hit test failures in the future.

boegel (approved):
lgtm

lexming:
re-triggering CI tets

PR #3686, opened 2021-05-18
update HierarchicalMNS to also return 'Toolchain/<name>/<version>' as $MODULEPATH extension for cpe* Cray toolchains

### LINKED ISSUE #3575: Hierarchical Module Naming Scheme with Cray toolchains
I have tried following the tutorial https://easybuilders.github.io/easybuild-tutorial/hmns to implement the Hierarchical Module Naming Scheme (MNS) with a modified Cray toolchain using Lmod 8.3.1, but the compiler modulefile created under the `Core` folder does not update the `MODULEPATH` once loaded, therefore the modules created cannot be searched (not even with `module spider`).

I was wondering if it is at all possible to use the Hierarchical MNS with the way Cray toolchains are configured: for instance I cannot prepare the environment as suggested in the tutorial, using the commands below, or I would unload essential modules:
```
module purge
module unuse $MODULEPATH
```

Thanks for your advice!

boegel:
fixes #3575

### DIFF
--- a/easybuild/tools/module_naming_scheme/hierarchical_mns.py
+++ b/easybuild/tools/module_naming_scheme/hierarchical_mns.py
@@ -70,6 +70,12 @@ COMP_NAME_VERSION_TEMPLATES = {
     'xlc,xlf': ('xlcxlf', '%(xlc)s'),
 }
 
+# possible prefixes for Cray toolchain names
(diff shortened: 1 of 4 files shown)


migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

Issue #3627, opened 2021-04-02
Recursive module unload with limited depth for Cray toolchains


lucamar:
I report here as a placeholder for further discussion a request to implement a limited depth recursive module unload when using the Cray toolchains: the relevant framework scripts are `easyblocks/generic/craytoolchain.py` and `tools/module_generator.py`.
The Cray toolchains load Cray external modules as dependencies that should not be unloaded, in order to keep a working programming environment. 
When you build a modulefile using a Cray toolchain with the recursive module unload feature enabled, the full list of dependencies is unrolled, therefore also the Cray external modules (dependencies of the Cray toolchain) are unloaded with the module. 
It would be preferable to stop the unloading at the depth of the Cray toolchain module, if possible.

boegel:
@lucamar Thanks to the changes in #3689 (which will be included in the upcoming EasyBuild v4.4.0 release), you can now selectively opt-out of recursive module unloading in specific easyconfig files using `recursive_module_unload = False`, so you can use this in the easyconfig files for the Cray toolchains to avoid that the (external) modules for the toolchain components are unloaded.

PR #3689, opened 2021-05-20
allow opting out of recursively unloaded of modules via recursive_module_unload easyconfig parameter

boegel:
cfr. #3627

### DIFF
--- a/test/framework/options.py
+++ b/test/framework/options.py
@@ -1974,10 +1974,15 @@ class CommandLineOptionsTest(EnhancedTestCase):
             self.eb_main(args, do_build=True, verbose=True)
 
             toy_module = os.path.join(self.test_installpath, 'modules', 'all', 'toy', '0.0-deps')
+
             if get_module_syntax() == 'Lua':
                 toy_module += '.lua'
+                is_loaded_regex = re.compile(r'if not \( isloaded\("gompi/2018a"\) \)', re.M)
+            else:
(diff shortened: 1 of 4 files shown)


test/framework/options.py
```diff
@@ -1974,10 +1974,15 @@ def test_recursive_module_unload(self):
...
+
             if get_module_syntax() == 'Lua':
                 toy_module += '.lua'
+                is_loaded_regex = re.compile(r'if not \( isloaded\("gompi/2018a"\) \)', re.M)
+            else:
+                # Tcl syntax
+                is_loaded_regex = re.compile(r"if { !\[is-loaded gompi/2018a\] }", re.M)
+
             toy_module_txt = read_file(toy_module)
-            is_loaded_regex = re.compile(r"if { !\[is-loaded gompi/2018a\] }", re.M)
```
migueldiascosta, line 1980: so, with `Lua` this test wasn't actually testing? (just checking if I understand)

boegel: Indeed. We were checking for a missing pattern using a Tcl regex pattern in a Lua module file.

That probably shows this test should be rewritten a bit, but I just wanted to fix the obvious problem when I noticed it.

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3692, opened 2021-05-21
make get_config_obj return a copy rather than a reference

### LINKED ISSUE #3691: get_config_dict returns parsed easyconfig file by reference
The `get_config_dict` method currently returns the parsed easyconfig file by reference, which leaves the door open for updating it in place. That's a problem because the next call the `get_config_dict` then doesn't correctly reflect the easyconfig file anymore...

Tweaking the easyconfig parameter values happens in the `PythonBundle` constructor for example for `exts_default_options`, where the default values are injected.
This causes problems with the easyconfigs test suite, since we don't allow `source_urls` to be set to `PYPI_SOURCE` anymore (cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/12453), but that's exactly what gets injected by the `PythonBundle` constructor if `source_urls` is not specified in `exts_default_options` in the easyconfig file...

boegel:
fixes #3691

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -672,8 +672,7 @@ class EasyConfig(object):
             for key in sorted(params.keys()):
                 # validations are skipped, just set in the config
                 if key in self._config.keys():
-                    # don't pass mutable values like lists/dicts by reference, copy them
-                    self[key] = copy.deepcopy(params[key])
+                    self[key] = params[key]
                     self.log.info("setting easyconfig parameter %s: value %s (type: %s)",
                                   key, self[key], type(self[key]))
                 else:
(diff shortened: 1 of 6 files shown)


Flamefire (approved):
I would have done it at a higher level (i.e. in the parser which has-a format object to avoid duplicating this, but this works too. Thanks!

For future reference as written in Slack:
The faster solution would be to copy the dict only on (potential) modification, e.g. when it is assigned into the EC (`set_keys`) because now we copy even if we only read from it (e.g. in the EC test suite). However it was decided to go for the safer solution and avoid the potential for accidental modification at the cost of a (comparatively cheap) copy of a dict

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #3663, opened 2021-04-27
add support for postinstallcmds for extensions

smoors:
(created using `eb --new-pr`)

PR #3696, opened 2021-05-25
run post-install commands specified for a specific extension 

boegel:
Without this change, specifying `postinstallcmds` for a specific extension is futile, they're just ignored because `post_install_step` is not run for extensions.

A workaround was implemented in the `PythonPackage` eaysblock for this in https://github.com/easybuilders/easybuild-easyblocks/pull/2381 which introduces a custom easyconfig paramter named `ext_postinstallcmds`, which is no longer needed thanks to the changes here.

Both #3663 and https://github.com/easybuilders/easybuild-easyblocks/pull/2404 can be closed once this is merged.

The custom `ext_postinstallcmds` easyconfig parameter for `PythonPackage` should be removed again (it's only there in `develop`, not included yet with an EasyBuild release), and the 2 easyconfigs that use it should be tweaked to use `postinstallcmds` instead (`DROP-1.0.3-foss-2020b-R-4.0.3.eb` and `CNVkit-0.9.8-foss-2020b-R-4.0.3.eb`).

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2436,20 +2436,33 @@ class EasyBlock(object):
                             contents = shebang + '\n' + contents
                             write_file(path, contents)
 
+    def run_post_install_commands(self, commands=None):
+        """
+        Run post install commands that are specified via 'postinstallcmds' easyconfig parameter.
+        """
+        if commands is None:
+            commands = self.cfg['postinstallcmds']
+
+        if commands:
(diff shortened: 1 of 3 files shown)


akesandgren (approved):
LGTM

smoors (approved):
lgtm
test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/12950
checked that it works with `postinstallcmds` for both the global easyconfig and extensions

PR #3701, opened 2021-05-26
enhance sched_getaffinity function to avoid early crash when counting available cores on systems with more than 1024 cores

Flamefire:
Dynamically determine the size of the cpu_set_t struct doubling it on each try

This is basically what the os module in Python 3.3+ does, see e.g. https://github.com/akheron/cpython/blob/f91d2f2e2e992c3006a5023eb4ba3cf0d082fde8/Modules/posixmodule.c#L5706

Error shown even on `eb --help` is:
> ERROR: sched_getaffinity failed for pid 675186 ec -1

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -29,6 +29,7 @@ Module with useful functions for getting system information
 @auther: Ward Poelmans (Ghent University)
 """
 import ctypes
+import errno
 import fcntl
 import grp  # @UnresolvedImport
 import os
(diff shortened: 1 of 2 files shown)


boegel:
@Flamefire Can you mention the error/crash you run into without this fix in the PR description?

Flamefire:
Added

easybuild/tools/systemtools.py
```diff
@@ -160,24 +161,35 @@ class SystemToolsException(Exception):
...
     pid = os.getpid()
-    cs = cpu_set_t()
-    ec = _libc.sched_getaffinity(os.getpid(), ctypes.sizeof(cpu_set_t), ctypes.pointer(cs))
-    if ec == 0:
-        _log.debug("sched_getaffinity for pid %s successful", pid)
-    else:
-        raise EasyBuildError("sched_getaffinity failed for pid %s ec %s", pid, ec)
+
+    cpu_setsize = 1024  # Max number of CPUs currently detectable
+    max_cpu_setsize = ctypes.c_ulong(-1).value // 4  # (INT_MAX / 2)
```
akesandgren, line 172: You could do max_cpu_setsize = cpu_mask_t(-1).value // 4 for consistency. And why the "//" instead of simply "/", asking since I don't know python enough.

akesandgren:
Apart from my comment above this code is sane and correct.

easybuild/tools/systemtools.py
boegel: `//` is integer division in Python 3, see https://riptutorial.com/python/example/2797/integer-division.

It's the correct way to ensure you get an integer both in Python 2 and 3, no float surprises :)

In Python 3:

```
>>> 3/2
1.5
>>> 3//2
1
```

akesandgren: good to know, learning by pestering those who know better :-)

test/framework/systemtools.py
```diff
@@ -454,7 +454,7 @@ def test_cpu_speed_native(self):
         """Test getting CPU speed."""
         cpu_speed = get_cpu_speed()
         self.assertTrue(isinstance(cpu_speed, float) or cpu_speed is None)
```
akesandgren, line 456: Shouldn't this one also get the same change in testing order as below?

boegel: No, that's fine. `isinstance(cpu_speed, float)` will be `False` if `cpu_speed = None`, so it'll go to the 2nd part of the `or`.

But `cpu_speed > 0.0` will fail hard if `cpu_speed = 0` in Python 3:

```
$ python3 -O -m test.framework.systemtools
.....................E.............................
ERROR: test_cpu_speed_native (__main__.SystemToolsTest)
Test getting CPU speed.
  File "/mnt/shared/home/boegel/easybuild/easybuild-framework/test/framework/systemtools.py", line 457, in test_cpu_speed_native
    self.assertTrue(cpu_speed > 0.0 or cpu_speed is None)
TypeError: '>' not supported between instances of 'NoneType' and 'float'
```

boegel (approved):
lgtm

easybuild/tools/systemtools.py
boegel: fixed in 3c1279e

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #3702, opened 2021-05-26
add support for --skip-extensions

### LINKED ISSUE #3672: Allow to skip extensions in sanity check
In #3645 the idea was raised to be able to skip the sanity check for extensions only, which might be useful if you are doing `--module-only --rebuild`

boegel:
#3655 fixed the behavior of `--module-only` to also cover extensions during the sanity check, which trumps the use case of generating a module file for an `R` installation when not all extensions have been installed.

To restore this capability, the new `--skip-extensions` options can be used in combination with `--module-only`.
It can also be used to install `R` (or `Python`, `Perl`, etc.) without extensions first, and then enhance the installation via `eb --skip` afterwards.

fixes #3672

### DIFF
--- a/test/framework/options.py
+++ b/test/framework/options.py
@@ -5858,7 +5858,8 @@ class CommandLineOptionsTest(EnhancedTestCase):
             self.assertTrue(regex.search(error_msg), "Pattern '%s' should be found in: %s" % (regex.pattern, error_msg))
 
         # failing sanity check for extension can be bypassed via --skip-extensions
-        self.eb_main(args + ['--skip-extensions'], do_build=True, return_error=True)
+        outtxt = self.eb_main(args + ['--skip-extensions'], do_build=True, raise_error=True)
+        self.assertTrue("Sanity check for toy successful" in outtxt)
 
     def test_skip_extensions(self):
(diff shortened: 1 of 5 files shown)


test/framework/options.py
```diff
@@ -5857,6 +5857,41 @@ def test_sanity_check_only(self):
             regex = re.compile(error_pattern)
             self.assertTrue(regex.search(error_msg), "Pattern '%s' should be found in: %s" % (regex.pattern, error_msg))
 
+        # failing sanity check for extension can be bypassed via --skip-extensions
+        self.eb_main(args + ['--skip-extensions'], do_build=True, return_error=True)
```
ocaisa, line 5861: There's no test here is there, so what's the point?

akesandgren: Yeah, what's the use of this?

boegel: If the sanity check fails, then an error is raised, so this would cause the test to fail.

I can add another quick check for the module file after this though, to make that a bit more clear.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3704, opened 2021-05-27
set LDFLAGS instead of LIBRARY_PATH and add new `required` keyword to `_get_software_root` for Fujitsu toolchain

migueldiascosta:
the `required` keyword was introduced in https://github.com/easybuilders/easybuild-framework/pull/3555


boegel (approved):
lgtm

PR #3705, opened 2021-05-27
correctly strip extension from filename in extract_cmd and back_up_file functions

branfosj:
If the original module file is named `2020a.lua` then the backup generated will start with `2020.bak_`. This is because `rstrip('.lua')` removes any of the characters `.`, `l`, `u`, or `a` found at the end of the string until it hits a character that is not one of those.

I've searched the framework and these look to be the only problematic uses of `rstrip`.


boegel (approved):
lgtm

PR #3712, opened 2021-05-31
don't use a different module name for FFTW in the Fujitsu toolchain

migueldiascosta:
a `versionsuffix` will be used instead (https://github.com/easybuilders/easybuild-easyconfigs/pull/12863#discussion_r639503172), so `FujitsuFFTW` can simply inherit `FFT_MODULE_NAME` from `Fftw`


boegel (approved):
lgtm

PR #3713, opened 2021-05-31
move -Nclang back to $C[XX]FLAGS in toolchain support for Fujitsu compiler

migueldiascosta:
having it in `CC` and `CXX` broke the RPATH wrappers

moving it back to`C[XX]FLAGS` has been tested with ~60 easyconfigs derived from the `GCCcore/10.3.0`/`foss/2021a` toolchain, both with and without RPATH, and the only place where it fails is within the bundled `_ctypes/libffi` of Python *2* 

this seems to be the best way to go for now until a better solution is found...((https://github.com/easybuilders/easybuild/issues/701#issuecomment-851316265)


boegel (approved):
lgtm

PR #3714, opened 2021-05-31
add Fujitsu compiler lib* subdir to both $LIBRARY_PATH and $LDFLAGS

migueldiascosta:
so that it's both always picked up by the RPATH wrappers and present in the toolchain variables, for instance for the `buildenv` modules


boegel (approved):
lgtm

PR #3715, opened 2021-05-31
drop support for Python 2.6

boegel:
The time is right to no longer support for Python 2.6, let's squeeze in this change for the upcoming EasyBuild v4.4.0 release...

The impact of this on the EasyBuild community should be very low, see results of latest EasyBuild User Survey (slide 19
 at https://easybuild.io/eum21/003_eum21_easybuild_state_of_the_union.pdf).

EasyBuild User Survey results on support for Python 2.6

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -50,7 +50,6 @@ except ImportError:
 
 from easybuild.base import fancylogger
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.config import build_option
 from easybuild.tools.filetools import is_readable, read_file, which
 from easybuild.tools.py2vs3 import string_type
 from easybuild.tools.run import run_cmd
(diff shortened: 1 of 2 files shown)


easybuild/tools/systemtools.py
```diff
@@ -1032,17 +1032,11 @@ def check_python_version():
     python_ver = '%d.%d' % (python_maj_ver, python_min_ver)
     _log.info("Found Python version %s", python_ver)
 
-    silence_deprecation_warnings = build_option('silence_deprecation_warnings') or []
-
     if python_maj_ver == 2:
         if python_min_ver < 6:
             raise EasyBuildError("Python 2.6 or higher is required when using Python 2, found Python %s", python_ver)
```
akesandgren, line 1037: Should say 2.7 here, or change the whole thing to say python_min_ver < 7

akesandgren (changes requested):

easybuild/tools/systemtools.py
```diff
@@ -1032,17 +1032,11 @@ def check_python_version():
...
         elif python_min_ver == 6:
```
Flamefire, line 1038:
```suggestion
        if python_min_ver < 7:
```

boegel: fixed with forced push in 8ba55f6c6

akesandgren (approved):
LGTM

branfosj (approved):
lgtm

PR #3716, opened 2021-05-31
Add per-step timing information

Flamefire:
Shows e.g.:
```
== fetching files...
== ... (took 2 secs)
== unpacking...
== ... (took 15 secs)
== patching...
== ... (took 7 secs)
```

Requires Python 2.7/3.2

### DIFF
--- a/easybuild/tools/utilities.py
+++ b/easybuild/tools/utilities.py
@@ -300,22 +300,21 @@ def time2str(delta):
     if not isinstance(delta, datetime.timedelta):
         raise EasyBuildError("Incorrect value type provided to time2str, should be datetime.timedelta: %s", type(delta))
 
-    delta_secs = delta.days * 3600 * 24 + delta.seconds + delta.microseconds / 10**6
+    delta_secs = delta.total_seconds()
 
-    if delta_secs < 60:
-        res = '%d sec' % int(delta_secs)
-    elif delta_secs < 3600:
-        mins = int(delta_secs / 60)
-        secs = int(delta_secs - (mins * 60))
-        res = '%d min %d sec' % (mins, secs)
-    else:
-        hours = int(delta_secs / 3600)
-        mins = int((delta_secs - hours * 3600) / 60)
-        secs = int(delta_secs - (hours * 3600) - (mins * 60))
-        hours_str = 'hours' if hours > 1 else 'hour'
-        res = '%d %s %d min %d sec' % (hours, hours_str, mins, secs)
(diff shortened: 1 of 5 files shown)


Flamefire:
@smoors Maybe you got time for this too?

easybuild/tools/utilities.py
```diff
@@ -300,22 +300,21 @@ def time2str(delta):
...
-        secs = int(delta_secs - (hours * 3600) - (mins * 60))
-        hours_str = 'hours' if hours > 1 else 'hour'
-        res = '%d %s %d min %d sec' % (hours, hours_str, mins, secs)
+    delta_secs = delta.total_seconds()
 
-    return res
+    hours = int(delta_secs / 3600)
+    delta_secs -= hours * 3600
+    mins = int(delta_secs / 60)
+    secs = int(delta_secs - mins * 60)
```
smoors, line 308: can be even more concise:
```suggestion
    hours, remainder = divmod(s, 3600)
    mins, secs = divmod(remainder, 60)

```

easybuild/tools/utilities.py
```diff
@@ -300,22 +300,21 @@ def time2str(delta):
...
+
+    res = []
+    if hours:
+        res.append('%d %s' % (hours, 'hour' if hours == 1 else 'hours'))
+    if mins or hours:
+        res.append('%d %s' % (mins, 'min' if mins == 1 else 'mins'))
+    res.append('%d %s' % (secs, 'sec' if secs == 1 else 'secs'))
```
smoors, line 315: I suspect this is going to break some users' grep scripts :)

smoors (commented):
minor suggestion

easybuild/tools/utilities.py
Flamefire: Well, I hope they don't grep for the "secs" part, but for "FINISHED IN" (or what that was) ;)

easybuild/tools/utilities.py
Flamefire: I like it, thanks! And no, the results of divmod are already ints :)

smoors: oh it depends on the type of delta_secs. if it's float, the result is also float ;)

Flamefire: Oh, yes. It's a float here... :/
Then you are right.

smoors (approved):
lgtm

smoors:
Going in, thanks @Flamefire!

PR #3724, opened 2021-06-05
avoid overwritting pr_nr in post_pr_test_report for reports with --in…

lexming:
Fixes issue discussed in https://github.com/easybuilders/easybuild-easyconfigs/pull/9902#issuecomment-855157975

The problem is that this `pr_nr` variable in the list comprehension is overwriting the `pr_nr` variable in the `post_pr_test_report` scope.

### DIFF
--- a/easybuild/tools/testing.py
+++ b/easybuild/tools/testing.py
@@ -295,7 +295,7 @@ def post_pr_test_report(pr_nr, repo_type, test_report, msg, init_session_state, 
 
     if build_option('include_easyblocks_from_pr'):
         if repo_type == GITHUB_EASYCONFIGS_REPO:
-            easyblocks_pr_nrs = [int(pr_nr) for pr_nr in build_option('include_easyblocks_from_pr')]
+            easyblocks_pr_nrs = [int(eb_pr_nr) for eb_pr_nr in build_option('include_easyblocks_from_pr')]
             comment_lines.append("Using easyblocks from PR(s) %s" %
                                  ", ".join(["https://github.com/%s/%s/pull/%s" %
                                             (pr_target_account, GITHUB_EASYBLOCKS_REPO, easyblocks_pr_nr)


migueldiascosta (approved):
lgtm!

migueldiascosta:
Going in, thanks @lexming!

easybuild/tools/testing.py
```diff
@@ -295,7 +295,7 @@ def post_pr_test_report(pr_nr, repo_type, test_report, msg, init_session_state,
 
     if build_option('include_easyblocks_from_pr'):
         if repo_type == GITHUB_EASYCONFIGS_REPO:
-            easyblocks_pr_nrs = [int(pr_nr) for pr_nr in build_option('include_easyblocks_from_pr')]
+            easyblocks_pr_nrs = [int(eb_pr_nr) for eb_pr_nr in build_option('include_easyblocks_from_pr')]
```
boegel, line 298: A safer fix would be using `for x in `, since we don't use single-letter variables outside of list comprehensions... :)

boegel: Also, can we come up with a way to make the tests trip over this?

JensTimmerman:
flake8 should catch this (https://www.flake8rules.com/rules/F812.html)
also, this is fixed in python3: 

https://www.python.org/dev/peps/pep-3100/
https://www.python.org/dev/peps/pep-0289/
```
List comprehensions also "leak" their loop variable into the surrounding scope. This will also change in Python 3.0, so that the semantic definition of a list comprehension in Python 3.0 will be equivalent to list(<generator expression>). Python 2.4 and beyond should issue a deprecation warning if a list comprehension's loop variable has the same name as a variable used in the immediately surrounding scope.
```
```
 ~  python3                                                                            Sun 06 Jun 2021 12:48:43 PM CEST
Python 3.9.5 (default, May 14 2021, 00:00:00) 
[GCC 11.1.1 20210428 (Red Hat 11.1.1-1)] on linux
>>> i = 6
>>> x = [i for i in [5]]
>>> x
[5]
>>> i
6
```

perhaps flake8 is ignoring this test because it is not an issue, since flake8 is run in python3 in the tests, I think adding an extra flake8 that runs as python2 would catch these...

JensTimmerman:
@boegel I've added linting for all supported python versions in https://github.com/easybuilders/easybuild-framework/pull/3725
This pr fails because it is now catching errors that were not caught before... so someone might have a look at it.

easybuild/tools/testing.py
boegel: done in #3726

PR #3726, opened 2021-06-06
enhance test for post_pr_test_report to cover combination of --from-pr and --include-easyblocks-from-pr

boegel:
follow-up for #3724

### DIFF
--- a/easybuild/tools/testing.py
+++ b/easybuild/tools/testing.py
@@ -193,7 +193,7 @@ def create_test_report(msg, ecs_with_res, init_session_state, pr_nrs=None, gist_
                 descr = "(partial) EasyBuild log for failed build of %s" % ec['spec']
 
                 if pr_nrs is not None:
-                    descr += " (PR #%s)" % ', #'.join(str(pr_nr) for pr_nr in pr_nrs)
+                    descr += " (PR #%s)" % ', #'.join(str(x) for x in pr_nrs)
 
                 if easyblock_pr_nrs:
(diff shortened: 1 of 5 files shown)


migueldiascosta (approved):
lgtm

easybuild/tools/testing.py
```diff
@@ -295,7 +295,7 @@ def post_pr_test_report(pr_nr, repo_type, test_report, msg, init_session_state,
 
     if build_option('include_easyblocks_from_pr'):
         if repo_type == GITHUB_EASYCONFIGS_REPO:
-            easyblocks_pr_nrs = [int(eb_pr_nr) for eb_pr_nr in build_option('include_easyblocks_from_pr')]
+            easyblocks_pr_nrs = [int(x) for x in build_option('include_easyblocks_from_pr')]
```
migueldiascosta, line 298: there are a few other very similar list comprehensions for pr numbers, is it worth also changing them? (given https://github.com/easybuilders/easybuild-framework/pull/3724#issuecomment-855379319 and https://github.com/easybuilders/easybuild-framework/pull/3725)

boegel: Probably worth also using single-letter variables there, yes...

boegel: fixed in 14715cb

smoors (approved):
lgtm

smoors:
Going in, thanks @boegel!

PR #3728, opened 2021-06-07
Add module-write hook

Flamefire:
This hook will be called anytime a module is written, so even for the fake modules used for sanity checks etc.
It gets passed the easyblock as the first argument and the filename and content to be written as the next 2 and may return a new text to be used for the module file. If nothing or None is returned the original text will be used.

### DIFF
--- a/easybuild/tools/hooks.py
+++ b/easybuild/tools/hooks.py
@@ -58,6 +58,7 @@ TESTCASES_STEP = 'testcases'
 
 START = 'start'
 PARSE = 'parse'
+MODULE_WRITE = 'module_write'
 END = 'end'
 
 PRE_PREF = 'pre_'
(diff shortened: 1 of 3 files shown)


easybuild/tools/hooks.py
```diff
@@ -162,18 +160,16 @@ def find_hook(label, hooks, pre_step_hook=False, post_step_hook=False):
...
     return res
 
 
 def run_hook(label, hooks, pre_step_hook=False, post_step_hook=False, args=None, msg=None):
     """
-    Run hook with specified label.
+    Run hook with specified label and return result of hook or None.
 
     :param label: name of hook
     :param hooks: list of defined hooks
```
smoors, line 175: here hooks is also a dict, right?

easybuild/tools/hooks.py
```diff
@@ -120,10 +121,7 @@ def load_hooks(hooks_path):
 
 def verify_hooks(hooks):
     """Check whether list of obtained hooks only includes known hooks."""
```
smoors, line 123: here hooks is also a dict, right?
and also in easybuild/framework/easyblock.py line 147?

easybuild/tools/hooks.py
```diff
@@ -162,18 +160,16 @@ def find_hook(label, hooks, pre_step_hook=False, post_step_hook=False):
...
+    if res:
+        _log.info("Found %s hook", hook_name)
 
     return res
 
 
 def run_hook(label, hooks, pre_step_hook=False, post_step_hook=False, args=None, msg=None):
     """
-    Run hook with specified label.
+    Run hook with specified label and return result of hook or None.
```
smoors, line 172:
```suggestion
    Run hook with specified label and return result of hook() or None.
```

smoors (commented):
looks good to me, apart from some comment updates.

easybuild/tools/hooks.py
Flamefire: Yes. Fixed

smoors (approved):
lgtm

smoors:
Going in, thanks @Flamefire!

PR #3729, opened 2021-06-07
fix crash in get_config_dict when copying modules that were imported in easyconfig file (like 'import os')

damianam:
can't be serialized. For instance `os`, if it is imported in a eb file

edit: fix for traceback:

```
  ...
  File "/Volumes/work/easybuild-framework/test/framework/easyconfig.py", line 4449, in test_easyconfig_import
    ec = EasyConfig(test_ec)
  File "/Volumes/work/easybuild-framework/easybuild/framework/easyconfig/easyconfig.py", line 514, in __init__
    self.parse()
  File "/Volumes/work/easybuild-framework/easybuild/framework/easyconfig/easyconfig.py", line 698, in parse
    ec_vars = self.parser.get_config_dict()
  File "/Volumes/work/easybuild-framework/easybuild/framework/easyconfig/parser.py", line 222, in get_config_dict
    cfg = self._formatter.get_config_dict()
  File "/Volumes/work/easybuild-framework/easybuild/framework/easyconfig/format/one.py", line 139, in get_config_dict
    if key != '__builtins__' and "'module'" not in type(cfg[key]):
TypeError: argument of type 'type' is not iterable
```

### DIFF
--- a/easybuild/framework/easyconfig/format/one.py
+++ b/easybuild/framework/easyconfig/format/one.py
@@ -135,10 +135,11 @@ class FormatOneZero(EasyConfigFormatConfigObj):
         # we can't use copy.deepcopy() directly because in Python 2 copying the (irrelevant) __builtins__ key fails...
         cfg_copy = {}
         for key in cfg:
-            try:
-                cfg_copy[key] = copy.deepcopy(cfg[key])
-            except TypeError:
-                self.log.info("Contents of key '%s' can't be pickled: '%s'", key, cfg[key])
+            if key != '__builtins__':
+                try:
+                    cfg_copy[key] = copy.deepcopy(cfg[key])
+                except TypeError:
+                    self.log.info("Contents of key '%s' can't be pickled: '%s'", key, cfg[key])
 
         return cfg_copy
 
--- a/easybuild/framework/easyconfig/format/one.py
+++ b/easybuild/framework/easyconfig/format/one.py
@@ -137,7 +137,12 @@ class FormatOneZero(EasyConfigFormatConfigObj):
         for key in cfg:
             # skip special variables like __builtins__, and imported modules (like 'os')
             if key != '__builtins__' and "'module'" not in str(type(cfg[key])):
-                cfg_copy[key] = copy.deepcopy(cfg[key])
+                try:
+                    cfg_copy[key] = copy.deepcopy(cfg[key])
+                except Exception as err:
+                    raise EasyBuildError("Failed to copy '%s' easyconfig parameter: %s" % (key, err))
+            else:
+                self.log.debug("Not copying '%s' variable from parsed easyconfig", key)
 
         return cfg_copy
 
--- a/easybuild/framework/easyconfig/format/one.py
+++ b/easybuild/framework/easyconfig/format/one.py
@@ -135,8 +135,10 @@ class FormatOneZero(EasyConfigFormatConfigObj):
         # we can't use copy.deepcopy() directly because in Python 2 copying the (irrelevant) __builtins__ key fails...
         cfg_copy = {}
         for key in cfg:
-            if key != '__builtins__':
+            try:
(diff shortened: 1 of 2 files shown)


boegel:
@damianam We should have an enhanced test for this (I can look into that).

I also wonder if we can come up with a better fix... I guess the failure was trigger by trying to pickle `os` itself?

damianam:
Yes, exactly.

easybuild/framework/easyconfig/format/one.py
```diff
@@ -136,7 +136,10 @@ def get_config_dict(self):
         cfg_copy = {}
         for key in cfg:
             if key != '__builtins__':
-                cfg_copy[key] = copy.deepcopy(cfg[key])
+                try:
```
boegel, line 139: I think it makes sure sense to check the value type of `cfg[key]` here, and skip the deepcopy for imported modules...

I can't think of anything else that would cause trouble, but if this starts failing because of other changes we make in the future, we don't want to ignore it silently...

easybuild/framework/easyconfig/format/one.py
```diff
@@ -136,7 +136,10 @@ def get_config_dict(self):
...
+                    cfg_copy[key] = copy.deepcopy(cfg[key])
+                except TypeError:
+                    self.log.info("Contents of key '%s' can't be pickled: '%s'", key, cfg[key])
```
boegel, line 142: This should be a warning?

boegel (dismissed):
I'm happy to follow up on this myself with a test + making the requested changes (and leave merging the PR to someone else)

easybuild/framework/easyconfig/format/one.py
```diff
@@ -135,7 +135,8 @@ def get_config_dict(self):
         # we can't use copy.deepcopy() directly because in Python 2 copying the (irrelevant) __builtins__ key fails...
         cfg_copy = {}
         for key in cfg:
-            if key != '__builtins__':
+            # skip special variables like __builtins__, and imported modules (like 'os')
+            if key != '__builtins__' and "'module'" not in str(type(cfg[key])):
```
smoors, line 139: maybe also skip class objects?
```python
... and "class 'type'" not in str(type(cfg[key])) and "type 'classobj'" not in str(type(cfg[key])):
```

boegel: When would we ever have a `class` definition in easyconfigs though?!

smoors: When something is possible, there will always be one person one day that will do it ;)

boegel: Well sure, but there's all kinds of weird stuff you could in theory do in an easyconfig file, since it's essentially Python code. That doesn't mean we support all those things.

In 558829b I've made sure you don't get an ugly traceback with a funky easyconfig file, but a hopefully useful clean error message.

smoors: that's perfectly fine, and I like that users will get a nice error message.

to be clear, we should indeed not support all possible funky stuff, but there's always a balance between what's useful to support and what can be implemented easily.
in this case the balance was not so clear-cut to me. some users prefer to put all logic into the easyconfig instead of the easyblock, so that could definitely be a useful use case (but again, I'm not saying we should support it)

smoors (approved):
lgtm

smoors (approved):
lgtm

smoors:
Going in, thanks @damianam!

easybuild/framework/easyconfig/format/one.py
damianam: Just to add a little bit of context: We are starting to rely on `import os` to have "smart" easyconfigs that do different things depending on the system where they are being installed. I would pretty much like to keep this working in the future 😅 . There are of course other ways to do the same thing (hooks, easyblocks, overlays with slightly different EB files), but this is pretty convenient in some cases.

PR #3730, opened 2021-06-08
make sure that tests requiring a github token have 'github' in the test name so that they can be easily filtered

migueldiascosta:
@boegel shall we also make sure all tests in `github.py` have `github` in the name, even though it looks redundant, so that all of those are also picked up by ` -m test.framework.suite github`?


boegel:
> @boegel shall we also make sure all tests in `github.py` have `github` in the name, even though it looks redundant, so that all of those are also picked up by ` -m test.framework.suite github`?

@migueldiascosta Yes, that makes sense to me, we should be able to easily run all tests related to the GitHub integration by filtering on `test_github`...

migueldiascosta:
@boegel with https://github.com/easybuilders/easybuild-framework/pull/3730/commits/5c521a9fc6c71acece84443d1b76b13cee3445ea, we can run all tests related to github integration by filtering on `github`. If we want to filter on `test_github`, there are a few more tests that need to be renamed, let me know

boegel (approved):
lgtm

PR #3732, opened 2021-06-08
Add option to ignore failing test step

Flamefire:
Users can now use --ignore-test-failure instead of --skip-test-step which will run the tests but does not abort on error and just prints it instead.

Using both options doesn't make sense and is hence disallowed and a warning is show to make users move to ignoring failures instead of skipping the test completely.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -258,6 +258,8 @@ class EasyBlock(object):
         if group_name is not None:
             self.group = use_group(group_name)
 
+        self.ignore_test_failure = build_option('ignore_test_failure')
+
         # generate build/install directories
         self.gen_builddir()
         self.gen_installdir()
@@ -1814,6 +1816,16 @@ class EasyBlock(object):
             self.log.info("Removing existing module file %s", self.mod_filepath)
             remove_file(self.mod_filepath)
 
+    def report_test_failure(self, msgOrError):
+        """Report a failing test either via an exception or warning depending on ignore-test-failure
+
+        msgOrError - Failure description. Either a string or an EasyBuildError"""
+        if self.ignore_test_failure:
+            print_warning("Test failure ignored: " + str(msgOrError), log=self.log)
+        else:
+            exception = msgOrError if isinstance(msgOrError, EasyBuildError) else EasyBuildError(msgOrError)
+            raise exception
+
     #
     # STEP FUNCTIONS
     #
@@ -2228,6 +2240,13 @@ class EasyBlock(object):
             (out, _) = run_cmd(unit_test_cmd, log_all=True, simple=False)
 
             return out
+
+    def _test_step(self):
+        """Run the test_step and handles failures"""
+        try:
+            self.test_step()
+        except EasyBuildError as e:
+            self.report_test_failure(e)
 
     def stage_install_step(self):
         """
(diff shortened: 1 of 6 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1814,6 +1816,16 @@ def remove_module_file(self):
             self.log.info("Removing existing module file %s", self.mod_filepath)
             remove_file(self.mod_filepath)
 
+    def report_test_failure(self, msgOrError):
```
boegel, line 1819: Don't use `camelCase` for variable names please: `msg_or_error`

easybuild/framework/easyblock.py
```diff
@@ -1814,6 +1816,16 @@ def remove_module_file(self):
...
+        """Report a failing test either via an exception or warning depending on ignore-test-failure
+
+        msgOrError - Failure description. Either a string or an EasyBuildError"""
```
boegel, line 1822: Please reformat the docstring:

```python
        """
        Report a failing test either via an exception or warning depending on ignore-test-failure

        :param msg_or_error: failure description (string value or an EasyBuildError instance)
        """

File: easybuild/framework/easyblock.py [hunk 3]
```diff
@@ -2229,6 +2241,13 @@ def test_step(self):
 
             return out
 
+    def _test_step(self):
+        """Run the test_step and handles failures"""
+        try:
+            self.test_step()
+        except EasyBuildError as e:
```
boegel, line 2248: Don't use single-letter variable names outside of list comprehensions please, use `err`?

easybuild/framework/easyblock.py
```diff
@@ -3363,10 +3382,12 @@ def run_step(self, step, step_methods):
         run_hook(step, self.hooks, pre_step_hook=True, args=[self])
 
         for step_method in step_methods:
-            self.log.info("Running method %s part of step %s" % (extract_method_name(step_method), step))
+            # Remove leading underscore from e.g. "_test_step"
+            method_name = extract_method_name(step_method).lstrip('_')
+            self.log.info("Running method %s part of step %s" % (method_name, step))
```
boegel, line 3387: use lazy style:

```python
self.log.info("Running method %s part of step %s", method_name, step)

File: easybuild/main.py [hunk 5]
```diff
@@ -304,6 +305,14 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     init_session_state.update({'module_list': modlist})
     _log.debug("Initial session state: %s" % init_session_state)
 
+    if options.skip_test_step:
+        if options.ignore_test_failure:
+            raise EasyBuildError("Found both ignore-test-failure and skip-test-step being set. "
```
boegel, line 310: `being set` -> `enabled`

test/framework/options.py
```diff
@@ -394,6 +394,28 @@ def test_skip_test_step(self):
...
+        toy_ec = os.path.join(topdir, 'easyconfigs', 'test_ecs', 't', 'toy', 'toy-0.0-test.eb')
+
+        args = [toy_ec, '--ignore-test-failure', '--force']
+
+        with self.mocked_stdout_stderr() as (_, stderr):
+            outtxt = self.eb_main(args, do_build=True)
+
+        msg = 'Test failure ignored'
+        self.assertTrue(re.search(msg, outtxt),
+                        "Ignored test failure message in log not found, outtxt: %s" % outtxt)
```
boegel, line 410: `not found` -> `should be found`

test/framework/options.py
```diff
@@ -394,6 +394,28 @@ def test_skip_test_step(self):
...
+        self.assertTrue(re.search(msg, stderr.getvalue()),
+                        "Ignored test failure message in stderr not found, stderr: %s" % stderr.getvalue())
```
boegel, line 412: `not found` -> `should be found`

test/framework/options.py
```diff
@@ -394,6 +394,28 @@ def test_skip_test_step(self):
         found = re.search(test_run_msg, outtxt)
         self.assertFalse(found, "Test execution command is NOT present, outtxt: %s" % outtxt)
 
+    def test_ignore_test_failure(self):
+        """Test ignore failing tests (--ignore-test-failure)."""
+
+        topdir = os.path.abspath(os.path.dirname(__file__))
+        toy_ec = os.path.join(topdir, 'easyconfigs', 'test_ecs', 't', 'toy', 'toy-0.0-test.eb')
+
+        args = [toy_ec, '--ignore-test-failure', '--force']
```
boegel, line 403: maybe add a comment to clarify why the test step is expected to fail when using `toy-0.0-test.eb`?

boegel (changes requested):

boegel (approved):
lgtm

PR #3734, opened 2021-06-10
Add per-extension timing

Flamefire:
This adds timing information for each extension. Especially useful to check which extension did hold up the build.

Follow-up to https://github.com/easybuilders/easybuild-framework/pull/3716

Choose to print the time with a tab, similar to the python-bundle prints its extension-steps, so it is not confused with the time for the extension step


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #3737, opened 2021-06-11
Eerror out when passing a list to run_cmd* to avoid running wrong command unintended, unless shell=True is used

Flamefire:
This is a serious pitfall as e.g. run_cmd(['python', '-c' , 'print(1)']) would be interpreted as `bash -c python -c 'print(1)'`, i.e. the arguments are passed to bash and not to python!

Allow to opt-in into that by passing shell=True explicitely.

edit (by @boegel): only occurrence in central easyblocks was in custom easyblock for TensorFlow, which was fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2469 (I went through all easyblocks, didn't find other instances of passing a list of strings to `run_cmd*`)

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -128,7 +128,7 @@ def get_output_from_process(proc, read_size=None, asynchronous=False):
 
 @run_cmd_cache
 def run_cmd(cmd, log_ok=True, log_all=False, simple=False, inp=None, regexp=True, log_output=False, path=None,
-            force_in_dry_run=False, verbose=True, shell=True, trace=True, stream_output=None, asynchronous=False):
+            force_in_dry_run=False, verbose=True, shell=None, trace=True, stream_output=None, asynchronous=False):
     """
     Run specified command (in a subshell)
     :param cmd: command to run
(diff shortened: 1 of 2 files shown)


easybuild/tools/run.py
```diff
@@ -155,6 +155,13 @@ def run_cmd(cmd, log_ok=True, log_all=False, simple=False, inp=None, regexp=True
     else:
         raise EasyBuildError("Unknown command type ('%s'): %s", type(cmd), cmd)
 
+    if shell is None:
+        shell = True
+        if isinstance(cmd, list):
+            raise EasyBuildError("When passing cmd as a list then `shell` must be set explictely! "
+                                 "Not that all elements of the list but the first are treated as arguments "
```
bartoldeman, line 162: s/Not/Note/

bartoldeman (dismissed):
just a little typo to fix, otherwise lgtm

easybuild/tools/run.py
Flamefire, line 162:
```suggestion
                                 "Note that all elements of the list but the first are treated as arguments "
```

Flamefire: Done.
PS: Check the "+-" button to directly make a suggestion of a code change.

boegel (approved):
lgtm

Issue #3688, opened 2021-05-20
surprising behavior with patch files that don't have .patch extension


smoors:
patches hat don't have .patch extension work the same as patches, unless a custom path is provided, eg.

`('patch_without_extension', 'custom/path')`

in that case the patch file is considered a normal file and is just copied to the custom path.

to avoid surprises for users who are unaware of this, it would be better to do one of the following options:
1. never apply the patch if it does not contain the .patch suffix.

2. always apply the patch, unless explicitly specified that it should only be copied, eg. with an extra list/tuple item:
    `('patch_without_extension', 'custom/path', 'copy-only')`.
    this would require updating easyconfigs that do an implicit copy.

3. raise error if it does not contain the .patch suffix and no destination path is given.
    this may break some easyconfigs that are not in the EB repo (although this is unlikely, and it's easily fixed)
    
I think option 3 is the best way to avoid surprises, . note also that the documentation specifically mentions that paches should have .patch extension: https://caylo-easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#source-files-and-patches

boegel:
I think the right way forward here is to print a clear deprecation warning whenever an actual patch file is specified that doesn't have a `.patch` extension, and then hard break that in the upcoming EasyBuild 5.0.

smoors:
I agree, just need to find time to implement it...

PR #3738, opened 2021-06-12
raise error for patches without .patch extension and no destination given (WIP)

smoors:
fixes https://github.com/easybuilders/easybuild-framework/issues/3688

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -443,7 +443,7 @@ class EasyBlock(object):
         for index, patch_spec in enumerate(patch_specs):
 
             # check if the patches can be located
-            copy_file = False
+            is_copy_file = False
             suff = None
             level = None
             if isinstance(patch_spec, (list, tuple)):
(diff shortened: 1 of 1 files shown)


boegel:
@smoors Isn't the copy location implicitly the current directory then, in the current implementation? Cfr. the failing test...

So, rather than making this a backwards-incompatible change, maybe we should deprecate it instead, and make it an error only in EasyBuild 5.0?

Can you also check whether any of the easyconfigs in the central repository will need to be updated?

smoors:
@boegel no, the failing test stops at `fetch_extension_sources()`, so it's not actually doing anything with the patch beyond fetching and checking.

In the current implementation, implicit copy is currently only done for patches without .patch suffix that have a path associated with them, as in tuple(filename, destination). In all other cases it is treated as a normal patch.

I also checked the central EB repository, and the few 'patches'  that don't have a .patch suffix all have a path, so there's nothing to update there.

It's only backward incompatible for the rare cases where users deviate from the documentation that says patches should have .patch suffix: https://caylo-easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#source-files-and-patches

For me the current behavior is problematic, especially for new users, because in some cases it's ok to have a patch without .patch suffix, while in other cases it just copies the patch without actually patching anything.

But if you prefer to deprecate it first, that's ok with me. Do you still prefer that?

smoors:
superseded by #3920

PR #3742, opened 2021-06-16
deprecate EasyBuild bootstrap script

branfosj:
related to https://github.com/easybuilders/easybuild/pull/718

### DIFF
--- a/easybuild/scripts/bootstrap_eb.py
+++ b/easybuild/scripts/bootstrap_eb.py
@@ -854,6 +854,8 @@ def main():
     self_txt = open(__file__).read()
     if IS_PY3:
         self_txt = self_txt.encode('utf-8')
+    info("DEPRECATED: Use of the EasyBuild boostrap script is deprecated and may not work.")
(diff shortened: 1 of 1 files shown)


easybuild/scripts/bootstrap_eb.py
```diff
@@ -854,6 +854,8 @@ def main():
     self_txt = open(__file__).read()
     if IS_PY3:
         self_txt = self_txt.encode('utf-8')
+    info("DEPRECATED: Use of the EasyBuild boostrap script is deprecated and may not work.")
+    info("Recommended install methods: https://docs.easybuild.io/en/latest/Installation.html\n")
```
boegel, line 858: I would take this one step further, and abort the bootstrap unless an environment variable like `$EASYBUILD_BOOTSTRAP_DEPRECATED` is defined, to actively discourage still using it.

boegel: see https://github.com/branfosj/easybuild-framework/pull/2

boegel (approved):
lgtm

Issue #3746, opened 2021-06-17
Checksums outside of EasyConfigs


mboisson:
At the moment, checksums for files are specified in EasyConfig files. This means that for every single copy of a recipe made for different toolchains, or even different versions that use the same patches, we repeat this information. 

My proposal is to implement a mechanism to get checksums from outside of the EasyConfig file. 

The file would be named : 
`<software name>.checksums` (or `checksums.json`)
and it would contain a dictionary of this format: 
```
{
   'filename1': 'checksum1',
   'filename2': 'checksum2',
}
``` 
The file would be located inside the EasyConfig directory for that software, as so to not obscure where the information is coming from (global parameters can be obscure). 

It could include patches as well as checksums for various versions of a given software. 

This would allow to completely eliminate checksums from the EasyConfig files (although we could support having both, with the EasyConfig-specified checksums having the highest priority). 

This is a competing proposal to https://github.com/easybuilders/easybuild-framework/issues/3744

ocaisa:
I like the general idea. It would also easily solve the issue we have of checksums changing over time (update a single location). It can also preserve backwards compatibility. 

One concern I would have is what to do with pull requests. If we had such a mechanism we would like to encourage/enforce it's use. That means that people would have to submit (at least) 2 files with every PR and there would be a high chance of a conflict in the sources yaml.

mboisson:
Maybe I am naive, but the learning curve of doing a PR is much steeper than the learning curve of adding more than 1 file to a PR ?

It could also occur that there are PRs *only* on the checksum file. For example, there is no need to create a separate EasyConfig if the only difference is the software version, you just add the checksum of the new version's source file to the checksum file, and use `--try-software-version=...`

ocaisa:
> Maybe I am naive, but the learning curve of doing a PR is much steeper than the learning curve of adding more than 1 file to a PR ?
> 

In general yes, but not with EB: `--new-pr`. Many, many people use that and this will require that they edit another file to make a PR. What we can do is make `--inject-checksums` do it for them but I don't see how you can get away from the likelihood of conflicts, and conflict resolution is not something we can handle right now.

> It could also occur that there are PRs _only_ on the checksum file. For example, there is no need to create a separate EasyConfig if the only difference is the software version, you just add the checksum of the new version's source file to the checksum file, and use `--try-software-version=...`
>
> Or in the case of evolving checksums, you just do a PR on the checksum file.

[...]

PR #3747, opened 2021-06-17
(WIP) External checksums

mboisson:
This is an admittedly naive implementation. In particular, problems left to be resolved: 

- [ ] `checksums.yaml` is not found when using `--try-` options
- [ ] no test implemented
- [ ] not sure what to do with `--inject-checksums` 

This is for issue https://github.com/easybuilders/easybuild-framework/issues/3746


mboisson:
Forgot to exclude our local modifications. Closing and opening a new PR

PR #3748, opened 2021-06-17
(WIP) Supporting checksums.json

mboisson:
This is an admittedly naive implementation (a proof of concept). In particular, problems left to be resolved: 

- [ ] no test implemented or validate
- [ ] not sure what to do with `--inject-checksums` 
- [ ] not sure what to do with `--new-pr` for the yaml file

This is for issue https://github.com/easybuilders/easybuild-framework/issues/3746

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -48,6 +48,7 @@ import stat
 import tempfile
 import time
 import traceback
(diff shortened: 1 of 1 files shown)


mboisson:
I'm not quite sure why the `'eb' command / test-eb` tests fail...

PR #3749, opened 2021-06-18
add support for checksums specified in external `checksums.json` file

mboisson:
Renamed the branch since we switch from yaml to json. 
This replaces https://github.com/easybuilders/easybuild-framework/pull/3748
to address 
https://github.com/easybuilders/easybuild-framework/issues/3746

- [x] no test implemented
- [X]  not sure what to do with `--inject-checksums`   (added `--inject-checksums-to-json`)


mboisson:
So far, I have assumed that one had either checksums in the EasyConfig file *or* a `checksums.json` file. The `checksums.json` file mechanism *only* takes effect if there are no checksums in the EasyConfig file. 

This is not really useful to reuse existing EasyConfigs with a bump `--try-software-version`, as the list provided in the EasyConfig will take precedence and the build will fail. 

I am rewriting the code such that *both* the checksums from the EasyConfig and the checksum from the Json file are loaded, and then providing a `build_option` to guide which one should be prioritized to validate the checksum if there are both.

mboisson:
@boegel, do you think we need to address `--new-pr` and the `--inject-checksums-to-json` when using `--try-software-version` issue in this PR ?

mboisson:
@boegel Should this be closed or fixed ? I won't spend time fixing its tests if it's not going to end up getting merged.

boegel:
@mboisson Having support for storing checksums externally makes sense, it can help a lot to declutter easyconfigs, so it would be great if we could get the tests fixed...
I can't make any promises on when this will go in though, sorry.

boegel:
@mboisson Some proposed cleanup for the test part of this in https://github.com/ComputeCanada/easybuild-framework/pull/26

I intend to make a pass over the non-test part too to propose some minor improvements, but this looks really good overall, thanks for all the effort you've put into this already! 👍

boegel (approved):
lgtm

PR #3754, opened 2021-06-22
add --review-pr-max and --review-pr-filter options to limit easyconfigs shown in multi-diff

migueldiascosta:
…


migueldiascosta:
without the sort, the tests run differently in the CI and locally, I suppose because the order of `glob` is not well defined. So it would seem we do need to sort somewhere, but making sure higher versions come first when using --review-pr-max (?)

PR #3755, opened 2021-06-23
use temporary download folder in test_http_header_fields_urlpat

migueldiascosta:


test/framework/options.py
```diff
@@ -2829,6 +2831,9 @@ def run_and_assert(args, msg, words_expected=None, words_unexpected=None):
         not_expected = [testdoval, testdonthdr, testdontval]
         run_and_assert(args, "case E", expected, not_expected)
 
+        # cleanup downloads
+        shutil.rmtree(tmpdir)
```
Flamefire, line 2835: I think this is not required. The test cases creates and removes the parent folder already

Flamefire (changes requested):

test/framework/options.py
boegel: Indeed, but it doesn't hurt either (this is cleanup of a subdir), so fine

boegel (approved):

PR #3762, opened 2021-07-03
honor specified easyblock for extensions

### LINKED ISSUE #3710: specified easyblock for extension is not taken into account
If we specify e.g. in a PythonBundle  easyconfig , to install an extension that has the same name of an existing EasyBlock , easybuild will try to use  that easyblock for it.

A check should be added to honor an `easyblock:` definition on the extension dict  as e.g. : 

```
    ('tbb', '2021.2.0', {
        'easyblock': 'PythonPackage',
        'source_tmpl': 'tbb-%(version)s-py2.py3-none-manylinux1_x86_64.whl',
        'unpack_sources': False,
        'checksums': ['bf54992d1b5d6135620d5f3f2649bf331e2c412a4af00effd35481a9ecb98ee6'],
    }),
```

boegel:
fixes #3710

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -540,6 +540,10 @@ class EasyBlock(object):
                         'options': ext_options,
                     }
 
+                    # if a particular easyblock is specified, make sure it's used
+                    # (this is picked up by init_ext_instances)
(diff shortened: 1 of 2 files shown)


Micket (approved):
lgtm

PR #3769, opened 2021-07-07
Don't add empty PATH or LD_LIBRARY_PATH

Flamefire:
While we already have detection of those containing any files recursively that is not enough: Those paths are NOT searched recursively, hence we need to check those folders only.

This happens a lot e.g. for Python packages where we have `lib/python3.7/site-packages/foo.py` with an otherwise empty `lib`.

And as each path added to those 2 env vars slows down execution/library loading it makes sense to clean those more.

### DIFF
--- a/test/framework/easyblock.py
+++ b/test/framework/easyblock.py
@@ -563,7 +563,7 @@ class EasyBlockTest(EnhancedTestCase):
         if get_module_syntax() == 'Tcl':
             self.assertFalse(re.search(r"prepend-path\s+LD_LIBRARY_PATH\s+\$%s\n" % sub_lib_path,
                                        txt, re.M))
-            self.assertFalse(re.search(r"prepend-path\s+PATH\s+\$%d\n" % sub_path_path, txt, re.M))
+            self.assertFalse(re.search(r"prepend-path\s+PATH\s+\$%s\n" % sub_path_path, txt, re.M))
         else:
             assert get_module_syntax() == 'Lua'
             self.assertFalse(re.search(r'prepend_path\("LD_LIBRARY_PATH", pathJoin\(root, "%s"\)\)\n' % sub_lib_path,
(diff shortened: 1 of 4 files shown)


test/framework/easyblock.py
```diff
@@ -551,6 +551,25 @@ def test_make_module_req(self):
...
+        eb.make_module_req_guess = lambda: {'LD_LIBRARY_PATH': sub_lib_path, 'PATH': sub_path_path}
+        for path in (sub_lib_path, sub_path_path):
+            full_path = os.path.join(eb.installdir, path, 'subpath')
+            os.makedirs(full_path)
+            write_file(os.path.join(full_path, 'any.file'), 'test')
+        txt = eb.make_module_req()
+        if get_module_syntax() == 'Tcl':
+            self.assertFalse(re.search(r"prepend-path\s+LD_LIBRARY_PATH\s+\$%s\n" % sub_lib_path,
+                                       txt, re.M))
+            self.assertFalse(re.search(r"prepend-path\s+PATH\s+\$%d\n" % sub_path_path, txt, re.M))
```
ocaisa, line 566:
```suggestion
            self.assertFalse(re.search(r"prepend-path\s+PATH\s+\$%s\n" % sub_path_path, txt, re.M))
```
typo

ocaisa (approved):
LGTM

Flamefire:
BTW: I'm not sure if we should do the same for `LIBRARY_PATH` as is done in e.g. PythonPackage, which would make that redundant and allow for removal of that. However for `LIBRARY_PATH=/tmp/foo gcc -l:bar/lib.so` it is possible to find `/tmp/foo/bar/lib.so`. This is a VERY unusual use and I'm not even sure if intended, but it is possible...

ocaisa:
The fact that that is possible is perhaps enough for me to not include `LIBRARY_PATH` in this PR. LIBRARY_PATH is only used at link time so I don't think it's going to have the same level of impact as the other two.

Your final test looks like it is stuck, I'm going to have to retrigger them all.

ocaisa:
Thanks for the improvement @Flamefire !

PR #3770, opened 2021-07-07
Filter out duplicate paths added to module files

Flamefire:
This basically fixes faulty EC files which result in modules like:
```
prepend_path("EBPYTHONPREFIXES", root)
prepend_path("QT_PLUGIN_PATH", pathJoin(root, "plugins"))
prepend_path("EBPYTHONPREFIXES", root)
```

The idea is to use a context manager around the creation of a module, store all added paths and reject duplicate ones that way clearing the stored paths at the end.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3192,21 +3192,18 @@ class EasyBlock(object):
         else:
             trace_msg("generating module file @ %s" % self.mod_filepath)
 
-        txt = self.module_generator.prepare_module_creation()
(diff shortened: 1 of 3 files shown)


akesandgren:
@Flamefire conflict resolution needed

Flamefire:
@akesandgren Rebased

akesandgren:
I can't find a new test that actually verifies that multiple copies of the same path actually gets filtered out.
Am I just not seeing it or is it missing?

Issue #3752, opened 2021-06-21
Skip of 'permissions' step does not work


Ghepardo:
When I add a line "skipsteps = ['permissions']" to an easyconfig, I would expect EB to skip its setting of permissions on the files and directories installed.  However, it does not skip this.  I attach an easyconfig containing this parameter, and the log file from running a build using it.  
[USEARCH-11.0.667-i86linux32.eb.txt](https://github.com/easybuilders/easybuild-framework/files/6687085/USEARCH-11.0.667-i86linux32.eb.txt)
[easybuild-USEARCH-11.0.667-20210621.142829.log](https://github.com/easybuilders/easybuild-framework/files/6687086/easybuild-USEARCH-11.0.667-20210621.142829.log)

The resulting installed file 'usearch' ends up with a permissions mode of 555.  It should retain the mode of 111 that is successfully set in the install_cmd commands, but which is then undone by EB.  

I believe there could be a bug here.  

Please note that my objective here is to end up with an installed file having a permissions mode of 111, and I am not really wanting EB to skip its entire 'permissions' step.  Skipping that step was simply suggested as a way to achieve my objective, in the Slack thread https://easybuild.slack.com/archives/C34UA1HT7/p1624266336284100.  Ideally I would have a way simply to specify the permissions mode of an installed file via a new easyconfig parameter.  I shall raise a separate issue for that.

PR #3771, opened 2021-07-08
Make logdir writable also when --stop/--fetch is given

Flamefire:
Same as for the usual installation: `--fetch` or `--stop` also write their logs into the installdir which fails when it is read-only.

Turns out (in a test) we did then also make the whole installdir writable but not write-only again when the permissions-step is skipped (e.g. via --fetch). So undo our temporary writeable modification in that case.
Related to https://github.com/easybuilders/easybuild-framework/issues/3752

While debugging I included 2 minor changes:
- Out of ECs when printed/logged as `<name> EasyConfig @ <path>`
- Output of unset $MODULEPATH as `<unset>`


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #3773, opened 2021-07-09
Fix forgotten renaming of 'l' to 'char'

Flamefire:


ocaisa (dismissed):

Flamefire:
@ocaisa Changed to the same as in the easyblocks package

easybuild/tools/include.py
```diff
@@ -72,9 +72,11 @@
...
     __path__ = pkgutil.extend_path(__path__, '%s.%s' % (__name__, subdir))
 
-del l, subdir, subdirs
+del subdir, subdirs
+if 'char' in dir():
+    del char
 
-__path__ = __import__('pkgutil').extend_path(__path__, __name__)
+__path__ = pkgutil.extend_path(__path__, __name__)
 """
```
ocaisa, line 80: Could there be a reason why this is not used in the other code?

Flamefire: The other code is literally a single line and the import is only used once. Here it is used multiple times

ocaisa (approved):

PR #3782, opened 2021-07-21
 Fix ignore_test_failure not set for Extensions 

Flamefire:
The ExtensionEasyBlock does not call `__init__` from EasyBlock and hence that value is not set making uses of report_test_failure fail.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -258,8 +258,6 @@ class EasyBlock(object):
         if group_name is not None:
             self.group = use_group(group_name)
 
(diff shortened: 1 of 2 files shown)


ocaisa (approved):

ocaisa:
I checked our shipped easyblocks and `self.ignore_test_failure` is not used anywhere

Flamefire:
> I checked our shipped easyblocks and self.ignore_test_failure is not used anywhere

This was very recently introduced (shortly before last release) and as the extension EB shows might be unreliable. So should be fine

ocaisa:
Thanks for the fix @Flamefire, going in!

PR #3784, opened 2021-07-23
Check for recursive symlinks by default before copying a folder 

Flamefire:
Avoids endless loops and creating infinite length paths when `symlinks=True` is NOT also passed to copy_dir

This turned up when naively testing https://github.com/easybuilders/easybuild-easyconfigs/pull/13469 without the `keepsymlinks` setting which ended up "hanging" in the install step until I eventually found out about the recursive symlink manually.
This PR should avoid such cases.

Downside: Traversing the whole source tree before a copy. Shouldn't be too bad as those are usually on fast, local disks and as the results of the traversal might even be cached it could make the actual copy happening later a bit faster reducing the impact of the check.
It will also follow symlinks to absolute paths and check there. Not fully sure about that but as the copy done afterwards does the same I don't think this is an additional problem.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -44,6 +44,7 @@ import glob
 import hashlib
 import imp
 import inspect
+import itertools
 import os
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #3786, opened 2021-07-28
Don't parse patch files as easyconfigs when searching for where patch file is used

### LINKED ISSUE #3781: find_software_name_for_patch can fail when non UTF8 files exist
Whilst trying
```
eb --update-pr 13453 Brotli-1.0.9_pc_link_flags.patch --pr-commit-msg="Add patch"
```
the code that tries to figure out where the patch belongs fails hard:
```
    "__main__", mod_spec)
    exec(code, run_globals)
  File "/apps/Common/software/EasyBuild/4.4.1/lib/python3.6/site-packages/easybuild/main.py", line 557, in <module>
    main()
  File "/apps/Common/software/EasyBuild/4.4.1/lib/python3.6/site-packages/easybuild/main.py", line 476, in main
    update_pr(options.update_pr, categorized_paths, ordered_ecs)
  File "/apps/Common/software/EasyBuild/4.4.1/lib/python3.6/site-packages/easybuild/tools/github.py", line 1876, in update_pr
    update_branch(branch_name, paths, ecs, github_account=github_account, commit_msg=commit_msg)
  File "/apps/Common/software/EasyBuild/4.4.1/lib/python3.6/site-packages/easybuild/tools/github.py", line 1848, in update_branch
    commit_msg=commit_msg)
  File "/apps/Common/software/EasyBuild/4.4.1/lib/python3.6/site-packages/easybuild/tools/github.py", line 861, in _easyconfigs_pr_common
    patch_specs = det_patch_specs(paths['patch_files'], file_info, [target_dir])
  File "/apps/Common/software/EasyBuild/4.4.1/lib/python3.6/site-packages/easybuild/tools/github.py", line 1051, in det_patch_specs
    soft_name = find_software_name_for_patch(patch_file, ec_dirs)
  File "/apps/Common/software/EasyBuild/4.4.1/lib/python3.6/site-packages/easybuild/tools/github.py", line 1079, in find_software_name_for_patch
    rawtxt = read_file(path)
  File "/apps/Common/software/EasyBuild/4.4.1/lib/python3.6/site-packages/easybuild/tools/filetools.py", line 209, in read_file
    txt = handle.read()
  File "/usr/lib64/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb4 in position 241240: invalid start byte
```

The file in question was
```
easybuild-easyconfigs/easybuild/easyconfigs/g/Grace/Grace-5.1.25-5build1.patch
```
which seems to be ISO-8859 encoded.

We have some options
1. I think this function should skip all patches, this just seems like a bug that it doesn't.
2. We could consider forcing UTF8 encoding? But it might be hard for patches..
3. We can just handle the errors here and proceed.
4. Make read_file encoding aware by some means?

Flamefire:
Exclude *.patch files during search
Remove C&P bug that needlessly reads the file a third time

Fixes #3781

While at it I also included a major speedup by checking ECs prefixed with the patch name (or parts of it) first.
Example: The `test_github_find_patches` test now parses exactly 1 EC to find a matching one.

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1069,11 +1069,13 @@ def find_software_name_for_patch(patch_name, ec_dirs):
 
     soft_name = None
 
+    ignore_dirs = build_option('ignore_dirs')
     all_ecs = []
     for ec_dir in ec_dirs:
         for (dirpath, dirnames, filenames) in os.walk(ec_dir):
-            # Don't visit any hidden folders, such as .git
-            dirnames[:] = [i for i in dirnames if not i.startswith('.')]
+            # Exclude ignored dirs
+            if ignore_dirs:
+                dirnames[:] = [i for i in dirnames if i not in ignore_dirs]
             for fn in filenames:
                 # TODO: In EasyBuild 5.x only check for '*.eb' files
                 if fn != 'TEMPLATE.eb' and os.path.splitext(fn)[1] not in ('.py', '.patch'):
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -839,7 +839,7 @@ def _easyconfigs_pr_common(paths, ecs, start_branch=None, pr_branch=None, start_
     # copy easyconfig files to right place
     target_dir = os.path.join(git_working_dir, pr_target_repo)
     print_msg("copying files to %s..." % target_dir)
-    file_info = COPY_FUNCTIONS[pr_target_repo](ec_paths, os.path.join(git_working_dir, pr_target_repo))
+    file_info = COPY_FUNCTIONS[pr_target_repo](ec_paths, target_dir)
 
     # figure out commit message to use
     if commit_msg:
@@ -1071,7 +1071,9 @@ def find_software_name_for_patch(patch_name, ec_dirs):
 
     all_ecs = []
     for ec_dir in ec_dirs:
-        for (dirpath, _, filenames) in os.walk(ec_dir):
+        for (dirpath, dirnames, filenames) in os.walk(ec_dir):
+            # Don't visit any hidden folders, such as .git
+            dirnames[:] = [i for i in dirnames if not i.startswith('.')]
             for fn in filenames:
                 # TODO: In EasyBuild 5.x only check for '*.eb' files
                 if fn != 'TEMPLATE.eb' and os.path.splitext(fn)[1] not in ('.py', '.patch'):
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1074,7 +1074,7 @@ def find_software_name_for_patch(patch_name, ec_dirs):
         for (dirpath, _, filenames) in os.walk(ec_dir):
             for fn in filenames:
                 # TODO: In EasyBuild 5.x only check for '*.eb' files
-                if fn != 'TEMPLATE.eb' and os.path.splitext(fn) not in ('.py', '.patch'):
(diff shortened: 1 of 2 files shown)


easybuild/tools/github.py
```diff
@@ -1073,17 +1073,35 @@ def find_software_name_for_patch(patch_name, ec_dirs):
     for ec_dir in ec_dirs:
         for (dirpath, _, filenames) in os.walk(ec_dir):
             for fn in filenames:
-                if fn != 'TEMPLATE.eb' and not fn.endswith('.py'):
+                # TODO: In EasyBuild 5.x only check for '*.eb' files
+                if fn != 'TEMPLATE.eb' and os.path.splitext(fn) not in ('.py', '.patch'):
```
Micket, line 1077: This returns a tuple
```suggestion
                if fn != 'TEMPLATE.eb' and os.path.splitext(fn)[1] not in ('.py', '.patch'):
```
I'm testing out this code with a PR where I'm currently getting stuck.

Micket (dismissed):

easybuild/tools/github.py
Micket: eh, actually, not even this is enough. 
I had git repo in path, and it's searching through `.git/index` and failing to parse it as utf-8. This should really have the same logic is robot does

Micket: I printed the paths just to see what it is trying to read, and it's a lot. E.g.
```
Reading /local/EB/build/eb-hqbrnpbu/git-working-dird1p5bheb/easybuild-easyconfigs/easybuild/easyconfigs/0/3to2/3to2-1.1.1-foss-2016b-Python-2.7.12.eb
Reading /local/EB/build/eb-hqbrnpbu/git-working-dird1p5bheb/easybuild-easyconfigs/.github/workflows/unit_tests.yml
Reading /local/EB/build/eb-hqbrnpbu/git-working-dird1p5bheb/easybuild-easyconfigs/.github/workflows/linting.yml
Reading /local/EB/build/eb-hqbrnpbu/git-working-dird1p5bheb/easybuild-easyconfigs/.git/index
```

Micket: To be clear, I'm printing all the paths that "slip through" the extension filtering here 
```python
                # TODO: In EasyBuild 5.x only check for '*.eb' files
                if fn != 'TEMPLATE.eb' and os.path.splitext(fn)[1] '.eb')[1] not in ('.py', '.patch'):
                    path = os.path.join(dirpath, fn)
                    print('Reading', path)
                    rawtxt = read_file(path)
                    if 'patches' in rawtxt:
                        all_ecs.append(path)
```

Micket: So, `ec_dirs = ['/local/EB/build/eb-7c1srdq1/git-working-dirgr_pgwkk/easybuild-easyconfigs']` is the problem then. It's 2 directories to high up.

Flamefire: I suppose we can add the 2 folders to `patch_specs = det_patch_specs(paths['patch_files'], file_info, [target_dir])` but for now, to keep it generic, I exclude hidden folders which also works

Flamefire: Using the ignore-dirs option now which includes .git by default. :)

boegel:
close + re-open in an attempt to get GitHub to get its act together w.r.t. commit vs PR status...

Flamefire:
@boegel That worked! :)

boegel (approved):
lgtm

PR #3788, opened 2021-07-28
Fully drop support code for Python < 2.7

Flamefire:
As Python <2.7 support was dropped in EasyBuild 4.4 I cleaned up the code that supports 2.6 which means removal of `OrderedDict` as well as some more unittest/CI code

Additionally I fixed the known issue that installing EB in Python3 shows a syntax error warning.


easybuild/tools/py2vs3/py2.py
```diff
@@ -63,9 +57,12 @@ def subprocess_popen_text(cmd, **kwargs):
     return subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs)
 
 
+# Wrapped in exec to avoid invalid syntax warnings for Python 3
```
boegel, line 60: Interesting approach to work around this invalid syntax error when installing EasyBuild with Python 3...

Maybe the comment should be clarified a bit though: the invalid syntax issue only occurs when this module is loaded when EasyBuild is installed with Python 3, and it's perfectly fine to ignore that because when running EasyBuild with Python 3, the `tools.py2vs3.py2` module will never be imported.

test/framework/repository.py
```diff
@@ -129,27 +128,6 @@ def test_svnrepo(self):
...
-        try:
-            import hglib  # noqa
-        except ImportError:
-            print("(skipping HgRepository test)")
-            return
-
-        # GitHub also supports SVN
-        test_repo_url = 'https://kehoste@bitbucket.org/kehoste/testrepository'
-
-        repo = HgRepository(test_repo_url)
```
boegel, line 148: Not testing this anymore means we should also deprecate the use of `HgRepository`, so we can drop support entirely in EasyBuild 5.0?

boegel (changes requested):

easybuild/tools/py2vs3/py2.py
Flamefire: I wouldn't spend to many letters on that corner case. I mean it is in "py2.py", the comment says "warnings" and "Python 3"
IMO that says enough already :)

requirements.txt
```diff
@@ -1,62 +1,34 @@
 # keyring is required to provide GitHub token to EasyBuild;
-# keyring v5.7.1 is last version to be compatible with py2.6;
 # for recent versions of keyring, keyrings.alt must be installed too
-keyring==5.7.1; python_version < '2.7'
-keyring<=9.1; python_version >= '2.7'
-keyrings.alt; python_version >= '2.7'
+keyring<=9.1
```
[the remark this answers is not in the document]
Flamefire: No idea. In the docs we have `pip install 'keyring<19.0'` for python 2, so I'll update this

test/framework/repository.py
Flamefire: We haven't tested this for ages, I just removed the dead code. No idea if you want to deprecate this as Mercury isn't exactly dead but I don't really care. If it works why not keep it. Might add a test later or deprecate it later.

PS: I love the comment above a mercury bitbucket URL about github and SVN ;)

requirements.txt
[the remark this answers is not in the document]
Flamefire: https://github.com/easybuilders/easybuild-framework/pull/3788/commits/dbd16466d2a29d77711ca6c520a66a9a1cedf9c8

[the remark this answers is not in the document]
branfosj: We have fixed keyring on one of our systems. The comment against that is `fix keyring to a version that does not require cryptography, which means no libffi required`

requirements.txt
```diff
@@ -1,67 +1,33 @@
...
-keyring<=9.1; python_version >= '2.7'
-keyrings.alt; python_version >= '2.7'
+# 19.0 dropped Python 2 support
+keyring<19.0; python_version < '3.0'
+keyring; python_version >= '3.0'
+keyrings.alt
 
-# GitDB 4.0.1 no longer supports Python 2.6
-gitdb==0.6.4; python_version < '2.7'
-gitdb; python_version >= '2.7'
```
akesandgren, line 10: Hmm, is gitdb not needed at all any longer?

akesandgren:
Apart from that I think this is good enough. @boegel do you want a deprecation warning for HgRepository?

requirements.txt
Flamefire: It is an indirect dependency of something. So not on of ours and doesn't need to be listed here. We only had it due to the python_version requirement

Flamefire:
Rebased and updated. Please review.

akesandgren:
Hmmm, this one has conflicts again...

boegel:
Since we're doing a broader code cleanup in #4229 for the upcoming EasyBuild 5.0, I'm not sure it's worth the effort to get this PR merged for EasyBuild 4.x...

Flamefire:
--> https://github.com/easybuilders/easybuild-framework/pull/4607 for 5.0x

boegel:
@Flamefire With #4607 merged, I guess we can close this?

Flamefire:
Python 2.6. is not supported by EB AFAIK but the `eb` binary still says otherwise.  This fixes it and does some IMO nice cleanup, which is easy to verify.

I'd still merge this but am fine with closing given the issues have been fixed in 5.x

easybuild/tools/py2vs3/py3.py
```diff
@@ -37,7 +37,6 @@
 import subprocess
 import sys
 import urllib.request as std_urllib  # noqa
-from collections import OrderedDict  # noqa
```
boegel, line 40: We can't/shouldn't make a breaking change like this in EasyBuild 4.x, it will break any `from easybuild.tools.py2vs3 import OrderedDict` out there...

Flamefire: Makes sense. Reverted that change

boegel (approved):
lgtm

PR #3794, opened 2021-08-03
add easybuild.tools.LooseVersion (since distutils is deprecated in Python 3.10)

Flamefire:
Port/Copy of `distutils.version.LooseVersion` as `distutils` will be deprecated and eventually removed from Python 3.10+

Slightly simplified and with tests
Also includes the fix from https://bugs.python.org/issue14894 to compare any LooseVersion instance in Python 3

BTW: There is a major pitfall we might actually run into at some points:
```
        # Careful here: 1.0 > 1 !!!
        self.assertGreater(LooseVersion('1.0'), LooseVersion('1'))
        self.assertLess(LooseVersion('1'), LooseVersion('1.0'))
```
With the included fix we could actually fix that such that missing components are considered equal. So the above would return equal which IMO is more sane for our usages
This is how e.g. CMake does its `if(ver VERSION_GREATER ver2)`


Flamefire:
@boegel Updated this. As this is required for #3963 I'd like to get this in. IMO having this class is very good, especially as we now have full control over it. This means:

- Fix the [bug](https://bugs.python.org/issue14894) when sorting 
- Drop-in replacement
- Easy use in comparisons e.g. `version = LooseVersion(self.version)` and later `if version > '1.4'  and version < '1.6'` reads very nice.

easybuild/tools/loose_version.py
```diff
@@ -0,0 +1,84 @@
...
+import re
+# Modified: Make this compatible with Python 2
+try:
+    from itertools import zip_longest
+except ImportError:
+    # Python 2
+    from itertools import izip_longest as zip_longest
+
+
+class LooseVersion(object):
```
boegel, line 21: Maybe we should rename this to something else than `LooseVersion`?

It's clearly based on that, but it's been fiddled with quite a bit, and behaves a bit different for some specific cases, so not naming it `LooseVersion` makes sense to me to avoid confusion with the original?

Flamefire: I find the name good and would like to keep it:

- It really is a "loose version" as it doesn't care much about the format or specifics like pre-releases etc.
- It is a drop-in replacement, i.e. only change is the `import` statement while everything else can be kept which reduces work, see this PR
- Only change is the comparison but that is basically a backported bugfix not a real behavior change, so I wouldn't say it behaves different, just consistent across different python versions (i.e. with and without the bugfix)

If you really want a different name, please suggest one you find better fitting. But note that the diff will be much larger as every use will have to be modified.

boegel:
Another (short-term) option here could be to keep using `LooseVersion` for Python versions where it is available, and fall back to the [`looseversion` Python package that is available via PyPI](https://pypi.org/project/looseversion/) for Python > = v3.12 .

Longer-term (EasyBuild v5.0?) we should probably fully take matters into our own hands, to avoid a required dependency outside of the Python standard library going forward...

Flamefire:
Why wait and use different versions on different Pythons creating possible hard-to-recreate bugs?

So why not just use this everywhere now? Is there any downside?

easybuild/tools/loose_version.py
boegel: Yeah, good points, thanks for clarifying! :+1:

boegel (approved):
This is a great step towards fixing #3963, and although I was a bit reluctant to take this approach at first, I have to admit it the most sensible solution.

Thanks a lot for your efforts on this @Flamefire!

PR #3795, opened 2021-08-04
make sure git clone with a tag argument actually downloads a tag

Flamefire:
The `git clone` command accept a `--branch` argument to clone a specific branch.
We use that to clone/download as tarball tags as it also accepts tags as the branch-argument.

However in case there is both a branch and a tag with the same name (happened in PyTorch) this will download the branch instead.

As contrary to `git checkout` `git clone --branch` does not accept neither a commit nor a direct tag reference (`refs/tags/foo`) we cannot reliably clone a tag only. But doing so (especially with `--depth 1` aka "shallow clone") is considerably faster and saves bandwidth.

So the current solution assumes that no branch with the same name as the tag exists and does the clone as before. It then checks, if we indeed cloned a tag
We could error out here, or (as done now) fetch the full history (as in a full clone) and check out the tag and submodules via `refs/tags/foo`.

--> In almost all cases the download is as fast as possible and it still works in the naughty case where maintainers screwed up and created a branch with the same name as a tag.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2477,7 +2477,7 @@ def get_source_tarball_from_git(filename, targetdir, git_config):
         clone_cmd.extend(['--depth', '1'])
 
     if tag:
-        clone_cmd.extend(['--branch', 'refs/tags/' + tag])
+        clone_cmd.extend(['--branch', tag])
         if recursive:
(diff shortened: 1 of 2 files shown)


boegel:
trying to trigger CI run again...

boegel (approved):
lgtm

PR #3796, opened 2021-08-05
add --filter-ecs options to filter out easyconfigs from set of easyconfigs to install

Flamefire:
Removes ECs to be build from the ECs passed on the cmdline or implicitely (and more useful) selected via the --from-pr option

Accepts a list of glob style patterns


boegel (approved):
lgtm

PR #3797, opened 2021-08-10
Allow optarch values to be partial maps including vector extensions

### LINKED ISSUE #3793: Add optimal optimization flags for Intel compilers on AMD CPUs
We have the following code for Intel toolchains:
```
        (systemtools.X86_64, systemtools.AMD): 'xHost',
        (systemtools.X86_64, systemtools.INTEL): 'xHost',
    }
```
However for AMD CPUs this is bad: `xHost` will use SSE only even when AVX2 is available. This then even fully fails installing software or installs it with worse optimizations.

As we usually care about CPU vendor and vector instructions supported, I'd optionally add a third argument here: The max supported vector instruction set.
In EB, we then need to define a list of supported types, e.g. "avx2, avx, sse2, sse" in that order. If a tuple `(<arch>, <vendor>, <vector>)` is found in the dict that flag is used, otherwise the next lower vec is tried. If all were tried, the last entry is removed and the remaining is tried
Extending this to the vendor part would allow this:

```
        (systemtools.X86_64, ): 'xHost',
        (systemtools.X86_64, systemtools.AMD, systemtools.AVX2): 'mavx2',
    }
```

We should also allow to set this via env vars, similar to `EASYBUILD_OPTARCH`: Instead of conditionally setting it to `EASYBUILD_OPTARCH="Intel:mavx2 -fma; GCC:march=native"` as we do on our site while having to check for AMD in the shell script we could do: 
`EASYBUILD_OPTARCH="Intel,x86:xHost; Intel,x86,AMD,AVX2:mavx2 -fma; GCC:march=native"`

This would allow some sort of future-proofing this.

For detecting the supported vector extensions we could use archspec or just use the cpu features query we already have and search for avx2 etc.

Flamefire:
Instead of only setting optimal compiler arguments based on architecture and CPU family/vendor also include the supported vector extensions as criteria to choose flags.
To simplify specifying generic flags allow partial matches, e.g. a fallback setting for any x86 arch which doesn't have more specific values.
Example:
```
        (systemtools.X86_64, ): 'xHost',
        (systemtools.X86_64, systemtools.AMD, systemtools.AVX2): 'mavx2',
    }
```

Closes #3793

### DIFF
--- a/easybuild/toolchains/compiler/inteliccifort.py
+++ b/easybuild/toolchains/compiler/inteliccifort.py
@@ -74,13 +74,15 @@ class IntelIccIfort(Compiler):
 
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
-        (systemtools.X86_64, systemtools.AMD): '-xHost',
-        (systemtools.X86_64, systemtools.INTEL): '-xHost',
+        systemtools.X86_64: 'xHost',
+        # Intel compilers don't auto-detect AMD features and default to SSE2
+        (systemtools.X86_64, systemtools.AMD, systemtools.SSSE3): 'mssse3',
+        (systemtools.X86_64, systemtools.AMD, systemtools.SSE4_2): 'msse4.2',
+        (systemtools.X86_64, systemtools.AMD, systemtools.AVX2): 'march=core-avx2',
     }
     # used with --optarch=GENERIC
     COMPILER_GENERIC_OPTION = {
-        (systemtools.X86_64, systemtools.AMD): '-xSSE2',
-        (systemtools.X86_64, systemtools.INTEL): '-xSSE2',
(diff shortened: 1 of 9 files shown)


Flamefire:
Ping on this as this came up in the EUM with respect to intel compilers

joeydumont:
To add to this, it seems like spack has already done some work in mapping CPU feature flags to compiler versions, see [microarchitectures.json](https://github.com/archspec/archspec-json/blob/edd3de72de240c332f6e349a20e287d547eb60dc/cpu/microarchitectures.json). There is some logic in portage as well, where `CPU_FLAGS` variable map to `USE` flags, but this probably maps less closely to the problem addressed in this PR than what archspec does.

Flamefire:
We just got hit by this not being implemented:
> tf_gen.F90(172): catastrophic error: Function return parameter requires SSE register while SSE is disabled.

When compiling HDF5 with the Intel compiler and `-xhost` on AMD EPYC 7702

PR #3798, opened 2021-08-10
fix CI by excluding GC3Pie 2.6.7 (which is broken with Python 2) and improve error reporting for option parsing

Flamefire:
GC3Pie 2.6.7 has a bug which makes it fail to import on Python2 fixed with https://github.com/gc3pie/gc3pie/commit/db8275470f2afa90192152994cc9cd73acb3c2af

Also our error reporting doesn't really help to find this as it just says: `Failed to parse configuration options: invalid syntax (__init__.py, line 2159)` which isn't exactly helpful
Hence only catch our error

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -50,7 +50,8 @@ pycodestyle; python_version < '2.7'
 flake8; python_version >= '2.7'
 
 # 2.6.7 uses invalid Python 2 syntax
-GC3Pie!=2.6.7
+GC3Pie!=2.6.7; python_version < '3.0'
+GC3Pie; python_version >= '3.0'
 python-graph-dot
 python-hglib
 requests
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -49,7 +49,8 @@ PyYAML; python_version >= '2.7'
 pycodestyle; python_version < '2.7'
 flake8; python_version >= '2.7'
 
-GC3Pie
+# 2.6.7 uses invalid Python 2 syntax
+GC3Pie!=2.6.7
```
boegel, line 53: We can make this a bit stricter, and keep testing with GC3Pie (current latest) when testing with Python 3 (for what it's worth...)

```
GC3Pie!=2.6.7; python_version <= '2.7'
GC3Pie; python_version > '2.7'

**Review by boegel — CHANGES_REQUESTED** [review]

File: requirements.txt (hunk 1 above)
Comment (Flamefire) [inline requirements.txt:53]:
```suggestion
GC3Pie!=2.6.7; python_version < '3.0'
GC3Pie; python_version >= '3.0'
```

Flamefire:
Done in the style of https://github.com/easybuilders/easybuild-framework/pull/3788 which will likely conflict with this

PR #3800, opened 2021-08-11
also define $BLAS_SHARED_LIBS & co in build environment (analogous to $BLAS_STATIC_LIBS)

boegel:
cfr. https://github.com/easybuilders/easybuild-easyblocks/issues/2521 for the need for this

### DIFF
--- a/easybuild/tools/toolchain/linalg.py
+++ b/easybuild/tools/toolchain/linalg.py
@@ -127,7 +127,9 @@ class LinAlg(Toolchain):
             self.variables.nappend('LIBBLAS', self.LIB_EXTRA, position=20)
             self.variables.nappend('LIBBLAS_MT', self.LIB_EXTRA, position=20)
 
+        self.variables.join('BLAS_SHARED_LIBS', 'LIBBLAS')
         self.variables.join('BLAS_STATIC_LIBS', 'LIBBLAS')
+        self.variables.join('BLAS_MT_SHARED_LIBS', 'LIBBLAS_MT')
         self.variables.join('BLAS_MT_STATIC_LIBS', 'LIBBLAS_MT')
         for root in self.get_software_root(self.BLAS_MODULE_NAME):
             self.variables.append_exists('BLAS_LIB_DIR', root, self.BLAS_LIB_DIR)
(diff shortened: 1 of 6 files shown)


easybuild/tools/toolchain/linalg.py
```diff
@@ -127,7 +127,9 @@ def _set_blas_variables(self):
             self.variables.nappend('LIBBLAS', self.LIB_EXTRA, position=20)
             self.variables.nappend('LIBBLAS_MT', self.LIB_EXTRA, position=20)
 
+        self.variables.join('BLAS_SHARED_LIBS', 'LIBBLAS')
         self.variables.join('BLAS_STATIC_LIBS', 'LIBBLAS')
+        self.variables.join('BLAS_MT_SHARED_LIBS', 'LIBBLAS_MT')
         self.variables.join('BLAS_MT_STATIC_LIBS', 'LIBBLAS_MT')
```
Micket, line 133: What is happening here? This looks strange to me that shared and static libs have something in common via LIBBLAS?

easybuild/tools/toolchain/variables.py
```diff
@@ -111,8 +111,16 @@ def change(self, separator=None, separator_begin_end=None, prefix=None, prefix_b
                 self.END.PREFIX = prefix_begin_end
 
 
+class CommaSharedLibs(LibraryList):
+    """Comma-separated list of shared libraries"""
+    SEPARATOR = ','
+
+    PREFIX = 'lib'
+    SUFFIX = '.so'
```
Micket, line 119: get_sharedlib_ext()  here

Micket (changes requested):

easybuild/tools/toolchain/variables.py
boegel: fixed in b2662fb1a

easybuild/tools/toolchain/linalg.py
boegel: This is strange internal stuff. The entry for `LIBBLAS_MT` is basically just a list of library names, at this point.

The values of each variable are later "formatted" according to the mapping in `easybuild/tools/toolchain/constants.py`, so the actual value you get for `$BLAS_MT_SHARED_LIBS` and `$LIBBLAS_MT` is different.

Micket (approved):
lgtm

PR #3802, opened 2021-08-12
speed up tests by caching checked paths in set_tmpdir + less test cases for test_compiler_dependent_optarch

Flamefire:
Noticed that one test was particularly slow in https://github.com/easybuilders/easybuild-framework/pull/3797 so I dug a bit and came up with these issues:
- Don't recheck temp folder for executability
- Reduce number of combinations tested in 1 test from 216 to 20

See the commits for details


boegel (approved):
lgtm

Debulknor:
~~~~

PR #3805, opened 2021-08-17
Correctly resolve templates for patches in extensions when uploading to GitHub

Flamefire:
name and version must be used from the extension, not the main software name

Also the components of Bundles were not handled at all -> Treat them as extensions


boegel (approved):
lgtm

PR #3807, opened 2021-08-20
add get_cuda_cc_template_value method to EasyConfig class

boegel:
This comes in useful in easyblocks for GPU-capable software like TensorFlow and jaxlib, since it results in a clean error message when neither `--cuda-compute-capabilities` (EasyBuild configuration option) nor `cuda_compute_capabilities` (easyconfig parameter) are defined.

See also the discussion at https://github.com/easybuilders/easybuild-easyblocks/pull/2545/files#r691193863

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1814,7 +1814,7 @@ class EasyConfig(object):
             try:
                 return self.template_values[key]
             except KeyError:
-                error_msg = "(get_cuda_cc_template_value) Template value '%s' is not defined!\n"
+                error_msg = "Template value '%s' is not defined!\n"
                 error_msg += "Make sure that either the --cuda-compute-capabilities EasyBuild configuration "
                 error_msg += "option is set, or that the cuda_compute_capabilities easyconfig parameter is defined."
                 raise EasyBuildError(error_msg, key)
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1819,7 +1819,7 @@ class EasyConfig(object):
                 error_msg += "option is set, or that the cuda_compute_capabilities easyconfig parameter is defined."
                 raise EasyBuildError(error_msg, key)
         else:
-            error_msg = "%s is not a template value baed on --cuda-compute-capabilities/cuda_compute_capabilities"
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1803,6 +1803,25 @@ def asdict(self):
...
         return res
 
+    def get_cuda_cc_template_value(self, key):
+        """
+        Get template value based on --cuda-compute-capabilities EasyBuild configuration option
+        and cuda_compute_capabilities easyconfig parameter.
+        Returns user-friendly error message in case neither are defined,
+        or if an unknown key is used.
+        """
+        if key.startswith('cuda_') and key in [x for (x, _) in TEMPLATE_NAMES_DYNAMIC]:
```
Flamefire, line 1813: A bit better and faster:
```suggestion
        if key.startswith('cuda_') and any(x[0] == key for x in TEMPLATE_NAMES_DYNAMIC):
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1803,6 +1803,25 @@ def asdict(self):
...
+            if key in self.template_values:
+                return self.template_values[key]
+            else:
```
Flamefire, line 1816: To be DRY:
```suggestion
            try:
                return self.template_values[key]
            except KeyError:
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1803,6 +1803,25 @@ def asdict(self):
...
+                error_msg = "(get_cuda_cc_template_value) Template value '%s' is not defined!\n"
```
Flamefire, line 1817: As this is a user facing error: I would not include this line in the error, maybe just log it and shorten it a bit: "You need to define --cuda-compute-capabilities and/or set the cuda_compute_capabilities EasyConfig parameter to a default value"

boegel: I would definitely include the "`Template value 'cuda_xxx' is not defined`" part, since that explain what goes wrong. The 2nd part of the error message explains how to fix the issue. Both bits are relevant information for the end user.

I agree that including `(get_cuda_cc_template_value)` might be overkill though, I'll kick that out.

branfosj (approved):
lgtm

PR #3808, opened 2021-08-23
add support for fix_bash_shebang_for

mboisson:
We have a weird case in which we need to fix the shebang to use `/usr/bin/env bash`


boegel (approved):
lgtm

PR #3812, opened 2021-08-30
speed up set_parallel method in EasyBlock class

Flamefire:
- Small refactoring to reduce dictionary access to cfg[parallel] option
- Cache default parallel value in det_parallelism to avoid systemcalls


boegel (approved):
lgtm

PR #3823, opened 2021-09-08
Use Rich (if available) to show progress bar when installing easyconfigs

### LINKED ISSUE #3711: Show progress counter during installation
When installing multiple EasyBuild configs it would be nice to show a simple progress counter of the number of completed configs against the total number of configs left to install.

One suggestion could be something like:
```bash
# Current: '== building and installing Python/3.8.6-GCCcore-10.2.0...'
== building and installing Python/3.8.6-GCCcore-10.2.0 (5 of 10)...
```

nordmoen:
Fixes #3711 

This implementation might be a bit overkill for what is needed, however, I wanted to show it as a potential implementation of how progress tracking could work.

The advantage of this approach is that the software being installed is always show together with a progressbar of the current installation. The disadvantage is that this implementation adds a new dependency ([`tqdm`](https://github.com/tqdm/tqdm)). One argument in favor of `tqdm` is that it is quite extensible so that we can customize it to suit the needs of this project.


boegel:
@nordmoen Looks cool, thanks a lot for looking into this!

Is there a way to avoid that the progress bar takes the whole width of the terminal with `tqdm`?
I found it very easy to overlook the ` | 0/2 [00:00<?, ?EB/s]` part on the right when building two easyconfigs, and felt myself wondering why the `0%` wasn't being updated at all as steps were progressing...

Also, the `EB/s` should probably say `easyconfigs` or `ec` rather than `EB`?

Maybe we want to count progress per installation step, not per easyconfig, so we also get a meaningful progress bar with just a single easyconfig?

I'm strongly in favor of not introducing any required dependencies, but I think this can be refactored with a minimal amount of additional work to only try using the progressbar if `tqdm` is actually installed.
We do similar stuff in other places, for example with `archspec` as optional dependency.

Finally, I've been trying to find time to brighten up the output produced by `eb` a bit using [`rich`](https://github.com/willmcgugan/rich). I don't mind using `tqdm`, but if we'll eventually use `rich` for other stuff anyway (like producing colorful error/warning/success messages), we mind as well use that here as well?
See `python3 -m rich.progress` for a demo if you have `rich` installed...

branfosj:
> @nordmoen Looks cool, thanks a lot for looking into this!
> 
> Is there a way to avoid that the progress bar takes the whole width of the terminal with `tqdm`?
> I found it very easy to overlook the ` | 0/2 [00:00<?, ?EB/s]` part on the right when building two easyconfigs, and felt myself wondering why the `0%` wasn't being updated at all as steps were progressing...
> 
> Also, the `EB/s` should probably say `easyconfigs` or `ec` rather than `EB`?
> 
> Maybe we want to count progress per installation step, not per easyconfig, so we also get a meaningful progress bar with just a single easyconfig?
> 
> I'm strongly in favor of not introducing any required dependencies, but I think this can be refactored with a minimal amount of additional work to only try using the progressbar if `tqdm` is actually installed.
> We do similar stuff in other places, for example with `archspec` as optional dependency.
> 
> Finally, I've been trying to find time to brighten up the output produced by `eb` a bit using [`rich`](https://github.com/willmcgugan/rich). I don't mind using `tqdm`, but if we'll eventually use `rich` for other stuff anyway (like producing colorful error/warning/success messages), we mind as well use that here as well?
> See `python3 -m rich.progress` for a demo if you have `rich` installed...

branfosj:
> @nordmoen Looks cool, thanks a lot for looking into this!

+1 on this.

> Is there a way to avoid that the progress bar takes the whole width of the terminal with `tqdm`?
> I found it very easy to overlook the ` | 0/2 [00:00<?, ?EB/s]` part on the right when building two easyconfigs, and felt myself wondering why the `0%` wasn't being updated at all as steps were progressing...
> 
> Also, the `EB/s` should probably say `easyconfigs` or `ec` rather than `EB`?
> 
> Maybe we want to count progress per installation step, not per easyconfig, so we also get a meaningful progress bar with just a single easyconfig?

The other option on this would be to have a two level progress bar - with level 1 being how many easyconfigs have been done and level 2 being the progress for installing the current easyconfig.

nordmoen:
Thank you both for the encouragement!

In the latest commit, I moved to the `rich` library since it has enough functionality for progress making no difference in my opinion. I also made the ticks more fine grained so that one gets progress during an install. Note that all steps advance the progress equally, which they probably should not, but this was done for ease of implementation and minimal intrusion.

Thanks to the `rich` library I could remove passing the progressbar to `print_msg`, with the down side that during heavy `print` output the progressbar loses its shape. We could of course go back to threading the progressbar to `print_msg` for prettier output, but that could be done later.

By passing both the progressbar and task of `rich` we also get the opportunity to create multiple layers of progress, which can be nice for Python packages and the `Bundle` easyblock.

boegel (dismissed):

boegel:
There's still a problem with the tests failing because a progress bar included in the generated output is messing up the expected output for some tests, I'm looking into fixing that...

boegel (changes requested):
@nordmoen Problem with the broken tests is fixed by making the progress bar transient, which makes sense anyway I think...

https://github.com/nordmoen/easybuild-framework/pull/2

nordmoen:
> @nordmoen Problem with the broken tests is fixed by making the progress bar transient, which makes sense anyway I think...
> 
> [nordmoen#2](https://github.com/nordmoen/easybuild-framework/pull/2)

Make sense

boegel:
@nordmoen One more proposed change in https://github.com/nordmoen/easybuild-framework/pull/3: using a spinner in the progress bar, rather than a simple static dot.
This makes things a bit more lively, as has the advantage that you can easily tell whether your session is still "alive" or not (which is useful when connecting to remote systems through SSH).

nordmoen:
Not sure if random choice is the best alternative as it could confuse users, but I don't mind 😊

boegel:
> Not sure if random choice is the best alternative as it could confuse users, but I don't mind 😊

I think it's just a funny detail, we can allow people to hard specify a specific spinner later, or take it even further and allow to define an EasyBuild "theme" which could control which colors are used for progress bars, error messages and warnings, etc.?
That's definitely for another PR though :P

One thing that is important though is to let people easily opt-out of getting progress bars, see https://github.com/nordmoen/easybuild-framework/pull/4

nordmoen:
Disabling the bar is necessary. Looks really good

boegel (approved):

boegel:
Thanks a lot for your efforts on this @nordmoen, this is the first step in something I've wanted to look into for a while now: leveraging Rich to make the EasyBuild user interface a bit more... rich.

PR #3832, opened 2021-09-13
ensure that path configuration options have absolute path values

### LINKED ISSUE #3619: --fetch fails for a git download when using a relative path for --sourcepath
Noticed while testing for https://github.com/easybuilders/easybuild-easyconfigs/pull/12347 - failure is not caused by `recursive = True`.

Using `eb PyTorch-1.8.0-foss-2020b.eb --fetch --sourcepath abc` fails with error
```
== 2021-03-25 09:59:22,577 run.py:225 INFO running cmd: tar cfvz abc/p/PyTorch/PyTorch-1.8.0.tar.gz --exclude .git pytorch
== 2021-03-25 09:59:22,712 build_log.py:169 ERROR EasyBuild crashed with an error (at easybuild/base/exceptions.py:124 in __init__): cmd "tar cfvz abc/p/PyTorch/PyTorch-1.8.0.tar.gz --exclude .git pytorch" exited with exit code -13 and output:
tar (child): abc/p/PyTorch/PyTorch-1.8.0.tar.gz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
pytorch/
pytorch/.bazelrc
pytorch/.bazelversion
pytorch/.circleci/
pytorch/.circleci/.gitignore
pytorch/.circleci/README.md
 (at easybuild/tools/run.py:537 in parse_cmd_output)
```

The directory `abc/p/PyTorch/` is created - it looks like the `tar` command is being run from the wrong place - so cannot find `abc/p/PyTorch/`.

Whereas `eb PyTorch-1.8.0-foss-2020b.eb --fetch --sourcepath /rds/bear-apps/devel/eb-sjb-up/src/easybuild-easyconfigs/easybuild/easyconfigs/tmp` worked fine.

### LINKED ISSUE #3830: Trouble with DB-18.1.40 in foss-2020b
Installation of foss-2020b fails at DB, and separate installation also fails:

`Failed to read ./generic/eb_v4.3.4/ConfigureMake/config.guess`

I have no idea..

boegel:
This could be considered a bug fix or an enhancement, depending on your point of view...

fixes #3619, fixes #3830

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1077,7 +1077,7 @@ class EasyBuildOptions(GeneralOption):
         # resolve relative paths for configuration options that specify a location
         path_opt_names = ['buildpath', 'containerpath', 'git_working_dirs_path', 'installpath',
                           'installpath_modules', 'installpath_software', 'prefix', 'packagepath',
-                          'robot_paths', 'sourcepath']
+                          'robot', 'robot_paths', 'sourcepath']
 
         # repositorypath is a special case: only first part is a path;
         # 2nd (optional) part is a relative subdir and should not be resolved to an absolute path!
@@ -1133,7 +1133,7 @@ class EasyBuildOptions(GeneralOption):
 
             # paths specified to --robot have preference over --robot-paths
             # keep both values in sync if robot is enabled, which implies enabling dependency resolver
-            self.options.robot_paths = [os.path.abspath(path) for path in self.options.robot + self.options.robot_paths]
+            self.options.robot_paths = self.options.robot + self.options.robot_paths
             self.options.robot = self.options.robot_paths
 
         # Update the search_paths (if any) to absolute paths
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1054,7 +1054,7 @@ class EasyBuildOptions(GeneralOption):
         else:
             abs_path = os.path.abspath(path)
             self.log.info("Relative path value for '%s' configuration option resolved to absolute path: %s",
-                          path, abs_path)
+                          opt_name, abs_path)
         return abs_path
 
     def _ensure_abs_path(self, opt_name):
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1135,7 +1135,7 @@ class EasyBuildOptions(GeneralOption):
 
             # paths specified to --robot have preference over --robot-paths
             # keep both values in sync if robot is enabled, which implies enabling dependency resolver
-            self.options.robot_paths = [os.path.abspath(p) for p in self.options.robot + self.options.robot_paths]
+            self.options.robot_paths = [os.path.abspath(p) for p in self.options.robot] + self.options.robot_paths
             self.options.robot = self.options.robot_paths
 
         # Update the search_paths (if any) to absolute paths
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1074,10 +1074,12 @@ class EasyBuildOptions(GeneralOption):
     def _postprocess_config(self):
         """Postprocessing of configuration options"""
 
-        # resolve relative paths for configuration options that specify a location
+        # resolve relative paths for configuration options that specify a location;
+        # ensuring absolute paths for 'robot' is handled separately below,
+        # because we need to be careful with the argument pass to --robot
         path_opt_names = ['buildpath', 'containerpath', 'git_working_dirs_path', 'installpath',
                           'installpath_modules', 'installpath_software', 'prefix', 'packagepath',
-                          'robot', 'robot_paths', 'sourcepath']
+                          'robot_paths', 'sourcepath']
 
         # repositorypath is a special case: only first part is a path;
         # 2nd (optional) part is a relative subdir and should not be resolved to an absolute path!
@@ -1133,7 +1135,7 @@ class EasyBuildOptions(GeneralOption):
 
             # paths specified to --robot have preference over --robot-paths
             # keep both values in sync if robot is enabled, which implies enabling dependency resolver
-            self.options.robot_paths = self.options.robot + self.options.robot_paths
+            self.options.robot_paths = [os.path.abspath(p) for p in self.options.robot + self.options.robot_paths]
             self.options.robot = self.options.robot_paths
 
         # Update the search_paths (if any) to absolute paths
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1047,8 +1047,40 @@ class EasyBuildOptions(GeneralOption):
 
         self.log.info("Checks on configuration options passed")
 
+    def _ensure_abs_path(self, opt_name):
+        """Ensure that path value for specified configuration option is an absolute path."""
+
+        def _ensure_abs_path(opt_name, path):
+            """Helper function to make path value for a configuration option an absolute path."""
+            if os.path.isabs(path):
+                abs_path = path
+            else:
+                abs_path = os.path.abspath(path)
+                self.log.info("Relative path value for '%s' configuration option resolved to absolute path: %s",
+                              path, abs_path)
+            return abs_path
+
+        opt_val = getattr(self.options, opt_name)
+        if opt_val:
+            if isinstance(opt_val, string_type):
+                setattr(self.options, opt_name, _ensure_abs_path(opt_name, opt_val))
+            elif isinstance(opt_val, list):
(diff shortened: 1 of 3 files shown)


easybuild/tools/options.py
```diff
@@ -1047,8 +1047,50 @@ def _postprocess_checks(self):
...
         self.log.info("Checks on configuration options passed")
 
+    def get_cfg_opt_abs_path(self, opt_name, path):
+        """Get path value of configuration option as absolute path."""
+        if os.path.isabs(path):
+            abs_path = path
+        else:
+            abs_path = os.path.abspath(path)
+            self.log.info("Relative path value for '%s' configuration option resolved to absolute path: %s",
+                          path, abs_path)
```
akesandgren, line 1057: judging form the info text that should be opt_name, abs_path

test/framework/options.py
```diff
@@ -6233,6 +6233,44 @@ def test_accept_eula_for(self):
...
+            '--show-config',
+        ]
+        txt, _ = self._run_mock_eb(args, do_build=True, raise_error=True, testing=False, strip=True)
+
+        patterns = [
+            r"^containerpath\s+\(F\) = .*/test_topdir/test_middle_dir$",
+            r"^installpath\s+\(E\) = .*/test_topdir$",
+            r"^prefix\s+\(C\) = .*/test_topdir/test_middle_dir$",
+            r"^repositorypath\s+\(F\) = \('/apps/easyconfigs_archive', ' somesubdir'\)$",
+            r"^sourcepath\s+\(C\) = .*/test_topdir/test_middle_dir/test_subdir$",
```
akesandgren, line 6268: The path regexes should really be /.*

easybuild/tools/options.py
```diff
@@ -1047,8 +1047,50 @@ def _postprocess_checks(self):
...
+
+    def _ensure_abs_path(self, opt_name):
+        """Ensure that path value for specified configuration option is an absolute path."""
+
+        opt_val = getattr(self.options, opt_name)
+        if opt_val:
+            if isinstance(opt_val, string_type):
+                setattr(self.options, opt_name, self.get_cfg_opt_abs_path(opt_name, opt_val))
+            elif isinstance(opt_val, list):
+                abs_paths = [self.get_cfg_opt_abs_path(opt_name, p) for p in opt_val]
```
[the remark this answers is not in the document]
boegel: But then the whole of `_ensure_abs_path` would basically be collapsed into `get_cfg_opt_abs_path`.

The point of `get_cfg_opt_abs_path` is to handle a *single* path, so it can also be used to handle the `repositorypath` special case.

I think it even makes sense to have multiple log messages for configuration options that are a list of paths.
The log message was carefully worded for that too, it never states that the absolute path set is *the* (only) value for that specific configuration option...

easybuild/tools/options.py
```diff
@@ -1047,8 +1047,50 @@ def _postprocess_checks(self):
...
+                error_msg = "Don't know how to ensure absolute path(s) for '%s' configuration option (value type: %s)"
+                raise EasyBuildError(error_msg, opt_name, type(opt_val))
+
     def _postprocess_config(self):
         """Postprocessing of configuration options"""
+
+        # resolve relative paths for configuration options that specify a location
+        path_opt_names = ['buildpath', 'containerpath', 'git_working_dirs_path', 'installpath',
+                          'installpath_modules', 'installpath_software', 'prefix', 'packagepath',
+                          'robot_paths', 'sourcepath']
```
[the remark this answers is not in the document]
boegel: Yes, but the path passed to `--robot` is already transformed to an absolute path elsewhere, see deeper down in this method where `self.options.robot_paths ` is set, so there's no need to do it again.

In fact, I can even kick out `robot_paths` from here too, so I will, and add a comment to clarify that `robot` and `robot_paths` are handled deeper down + make sure the test that verifies absolute paths covers both of those.

easybuild/tools/options.py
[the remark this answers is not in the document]
akesandgren: True, I just didn't like the idea of getting multiple messages for a single option... Not a big deal though.

easybuild/tools/options.py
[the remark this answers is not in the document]
akesandgren: well robot_paths below is only abs_path:ed if robot is !None so if you drop robot_paths from here you miss abs_pathing a --robot-paths which contains non-abs paths.

test/framework/options.py
```diff
@@ -6233,6 +6233,68 @@ def test_accept_eula_for(self):
...
+            r"^prefix\s+\(C\) = /.*/test_topdir/test_middle_dir$",
+            r"^repositorypath\s+\(F\) = \('/apps/easyconfigs_archive', ' somesubdir'\)$",
+            r"^sourcepath\s+\(C\) = /.*/test_topdir/test_middle_dir/test_subdir$",
+            r"^robot-paths\s+\(E\) = /.*/test_topdir$",
+        ]
+        for pattern in patterns:
+            regex = re.compile(pattern, re.M)
+            self.assertTrue(regex.search(txt), "Pattern '%s' should be found in: %s" % (pattern, txt))
+
+        # if --robot is also used, that wins and $EASYBUILD_ROBOT_PATHS doesn't matter anymore
```
akesandgren, line 6279: Eh what? EASYBUILD_ROBOT_PATHS should still matter when --robot is used?? Just like the configfile robot_paths still matters...

boegel: Ah, yes, it should state that the paths passed to `--robot` have precedence over those specified via `$EASYBUILD_ROBOT_PATHS`...

easybuild/tools/options.py
```diff
@@ -1047,8 +1047,50 @@ def _postprocess_checks(self):
...
+                error_msg = "Don't know how to ensure absolute path(s) for '%s' configuration option (value type: %s)"
+                raise EasyBuildError(error_msg, opt_name, type(opt_val))
+
     def _postprocess_config(self):
         """Postprocessing of configuration options"""
+
+        # resolve relative paths for configuration options that specify a location
+        path_opt_names = ['buildpath', 'containerpath', 'git_working_dirs_path', 'installpath',
+                          'installpath_modules', 'installpath_software', 'prefix', 'packagepath',
+                          'robot', 'robot_paths', 'sourcepath']
```
akesandgren, line 1080: Drop robot and return the "[os.path.abspath(path) for path in self.options.robot" part on (new) line 1136

akesandgren (changes requested):

easybuild/tools/options.py
boegel: Any reason why?
It's more consistent like this, and the absolute path resolution is done earlier, so less change for a relative path to cause trouble early on...

boegel: Oh, because of the failing test, we're now unintentionally promoting an argument like `example.eb` to an absolute path if it's used directly after `--robot`, which messes up the easyconfig searching...

And people wonder why testing is so important. :P

boegel: fixed in 6934dd8

akesandgren (approved):

easybuild/tools/options.py
```diff
@@ -1091,7 +1135,7 @@ def _postprocess_config(self):
 
             # paths specified to --robot have preference over --robot-paths
             # keep both values in sync if robot is enabled, which implies enabling dependency resolver
-            self.options.robot_paths = [os.path.abspath(path) for path in self.options.robot + self.options.robot_paths]
+            self.options.robot_paths = [os.path.abspath(p) for p in self.options.robot + self.options.robot_paths]
```
akesandgren, line 1138: Hmmm., doesn't this mean that robot_paths get run twice through abspath?? I don't know if that "+" op belongs to the for args or not...

boegel: `for p in self.options.robot + self.options.robot_paths` is equivalent with `for p in (self.options.robot + self.options.robot_paths)`, is that the question?

`robot_paths` is already covered higher up, sure, but `os.path.abspath` is *cheap* for absolute paths anyway, so that's fine?

akesandgren: Yeah, but then wouldn't [os.path.abspath(p) for p in self.options.robot] + self.options.robot_paths be even cheaper?

boegel: True, I can change it to that (but we're basically talking about skipping of `if p.startswith('/')`)

boegel: fixed in 94e987603

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3835, opened 2021-09-14
disable progress bars when running the tests to avoid messing up test suite output

boegel:


test/framework/utilities.py
```diff
@@ -137,6 +137,11 @@ def setUp(self):
 
         init_config()
 
+        # disable progress bars when running the tests,
+        # since it messes up with test suite progress output when test installations are performed
```
akesandgren, line 141: "since it messes with test suite progress output"
or
"since it messes up test suite progress output"

akesandgren (changes requested):

test/framework/utilities.py
boegel: fixed in d5ccae925

akesandgren (approved):

boegel:
Hmm, that failing test on Python 3.9.7 is very strange...
Looks like that could be a regression in that specific Python version?!

boegel:
> Hmm, that failing test on Python 3.9.7 is very strange...
> Looks like that could be a regression in that specific Python version?!

Not really a regression I guess, but indeed due to a change in Python 3.9.7, fixed in #3840

akesandgren (approved):

akesandgren:
Going in, thanks @boegel!

PR #3836, opened 2021-09-15
change copy_file function to raise an error when trying to copy non-existing file

nordmoen:
Previous behavior would simply skip copying a file if it did not exist
and report copy success. This commit changes that behavior to error out
early if the file does not exist and explicitly warn if trying to copy
something that is neither a file nor a symbolic link.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2270,7 +2270,8 @@ def copy_file(path, target_path, force_in_dry_run=False):
     :param target_path: path to copy the file to
     :param force_in_dry_run: force copying of file during dry run
     """
-    if not os.path.exists(path):
+    # NOTE: 'exists' will return False if 'path' is a broken symlink
+    if not os.path.exists(path) and not os.path.islink(path):
         raise EasyBuildError("could not copy '%s' it does not exist!" % path)
     if not force_in_dry_run and build_option('extended_dry_run'):
         dry_run_msg("copied file %s to %s" % (path, target_path))
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2272,7 +2272,7 @@ def copy_file(path, target_path, force_in_dry_run=False):
     """
     # NOTE: 'exists' will return False if 'path' is a broken symlink
     if not os.path.exists(path) and not os.path.islink(path):
-        raise EasyBuildError("could not copy '%s' it does not exist!" % path)
+        raise EasyBuildError("Could not copy '%s' it does not exist!" % path)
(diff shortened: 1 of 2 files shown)


easybuild/tools/filetools.py
```diff
@@ -2270,6 +2270,8 @@ def copy_file(path, target_path, force_in_dry_run=False):
     :param target_path: path to copy the file to
     :param force_in_dry_run: force copying of file during dry run
     """
+    if not os.path.exists(path):
```
boegel, line 2273: To avoid breaking the use case of copying a broken symlink (see broken test), you should add a `not os.path.islink(path)` here?

boegel (changes requested):

easybuild/tools/filetools.py
nordmoen: Fixed in latest commit

boegel:
@nordmoen Are you up for extending the `test_copy_file` we have in `test/framework/filetools.py` to verify the behavior when trying to copy a non-existing file?

nordmoen:
> @nordmoen Are you up for extending the `test_copy_file` we have in `test/framework/filetools.py` to verify the behavior when trying to copy a non-existing file?

Should be able to, yes

easybuild/tools/filetools.py
```diff
@@ -2270,6 +2270,9 @@ def copy_file(path, target_path, force_in_dry_run=False):
     :param target_path: path to copy the file to
     :param force_in_dry_run: force copying of file during dry run
     """
+    # NOTE: 'exists' will return False if 'path' is a broken symlink
+    if not os.path.exists(path) and not os.path.islink(path):
+        raise EasyBuildError("Could not copy '%s' it does not exist!" % path)
```
boegel, line 2275: Hmm, do we also want this to happen during dry run mode (`eb -x`)?

I'd say probably not, because there it's very common to not have the files actually exist, and it would trigger lots of failures when using `eb -x` (see the output of `eb -x Eigen-3.2.8.eb` for example, this change makes it trip during `install_step`)

So, an extra guard is required here:

```python
if not os.path.exists(path) and not os.path.islink(path):
    if force_in_dry_run:
        raise EasyBuildError("Could not copy '%s' it does not exist!", path)
    else:
        _log.debug("Ignoring non-existing file in copy_file because of dry run mode: %s", path)
```

The test should be enhanced accordingly, it should check that the error is not raised when in dry run mode (you can use `update_build_option('extended_dry_run', True)`, and also test with `force_in_dry_run=True`).

It's also better to use lazy logging: `, path` rather than `% path`

boegel (changes requested):

boegel (approved):

PR #3841, opened 2021-09-20
Add separate progress bar for extensions

nordmoen:
Slightly updated the way the progress bar is configured by adding a new
field value `software` describing the name that is currently installing
and using `description` for the act that is happening.

When installing extensions this PR creates a new task which is then removed once done installing extensions.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2433,6 +2433,7 @@ class EasyBlock(object):
             ext_task_id = self.progress_bar.add_task("Extension", total=len(self.ext_instances), software="")
             # Step size to advance progress bar for extensions
             ext_step_size = 1.0 / 4.0
+
         def _advance_ext_bar():
             """
             Helper method to update the extension progress bar
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -88,7 +88,6 @@ from easybuild.tools.module_naming_scheme.utilities import det_full_ec_version
 from easybuild.tools.modules import Lmod, curr_module_paths, invalidate_module_caches_for, get_software_root
 from easybuild.tools.modules import ROOT_ENV_VAR_NAME_PREFIX, VERSION_ENV_VAR_NAME_PREFIX, DEVEL_ENV_VAR_NAME_PREFIX
 from easybuild.tools.modules import get_software_root_env_var_name, get_software_version_env_var_name
-from easybuild.tools.output import create_progress_bar
 from easybuild.tools.package.utilities import package
 from easybuild.tools.py2vs3 import extract_method_name, string_type
 from easybuild.tools.repository.repository import init_repository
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2428,18 +2428,28 @@ def extensions_step(self, fetch=False, install=True):
...
 
             self.log.debug("Starting extension %s" % ext.name)
+            if self.progress_bar:
+                self.progress_bar.update(ext_task_id, software=ext.name)
 
             # always go back to original work dir to avoid running stuff from a dir that no longer exists
             change_dir(self.orig_workdir)
 
-            tup = (ext.name, ext.version or '', idx + 1, exts_cnt)
-            print_msg("installing extension %s %s (%d/%d)..." % tup, silent=self.silent)
```
boegel, line 2440: We shouldn't remove this output, since Rich is an optional dependency.

Still printing this when using Rich may even still makes sense (perhaps in a slightly different form), to give an overview of what was installed already.

The progress bar only shows what's currently being installed, it probably makes sense to keep showing what was installed already too...

easybuild/framework/easyblock.py
```diff
@@ -2428,18 +2428,28 @@ def extensions_step(self, fetch=False, install=True):
         if self.skip:
             self.skip_extensions()
 
-        exts_cnt = len(self.ext_instances)
-        for idx, ext in enumerate(self.ext_instances):
+        # Create progress bar for extensions
+        if self.progress_bar:
+            ext_task_id = self.progress_bar.add_task("Extension", total=len(self.ext_instances), software="")
+            # Step size to advance progress bar for extensions
+            ext_step_size = 1.0 / 4.0
```
boegel, line 2435: I see why you break up the installation of a single extension in 4 (prepare, pre-run, run, post-run), but perhaps that's a bit too fine-grained, since the prepare/pre-run and post-run are usually very quick (or empty) steps...

boegel (changes requested):
@nordmoen Your approach is pretty basic, since you're essentially just "forking" off a 2nd progress bar that's almost identical to the overall progress bar (which doesn't mean it's a bad approach, let me be clear).

I took things a bit further in #3844 by using *different* progress for different "levels" of progress (overall, per easyconfig, downloading, and eventually also extensions), which also includes some re-factoring of how the progress bar is handled (it avoids having to pass it around, leading to cleaner code).

Regardless, I'm happy to merge this anyway and deal with the fallout (merge conflicts) in my PR, since it's a nice step forward, and I don't want to let your efforts go to waste (we really appreciate contributions like this).

We do need to avoid impacting the output for people who don't have Rich though, and definitely avoid breaking the tests...

If you're up for looking into the suggested changes, we can get this merged soon, but if you're short on time it's fine too, since a progress bar for extensions will be done pretty differently in #3844 anyway.

nordmoen:
> If you're up for looking into the suggested changes, we can get this merged soon, but if you're short on time it's fine too, since a progress bar for extensions will be done pretty differently in #3844 anyway.

I don't think it is worth bringing this into the new setup. I think it will just be faster to re-implement this in the new setup.

PR #3845, opened 2021-09-21
Add option to make sanity_check_paths arch dependent

Flamefire:
- Factor out `pick_system_specific_value` to be used by `pick_dep_version`
- Allow arch specific entries in sanity check paths

Example: `'files': [{'arch=x86_64': 'correct.a'}, 'all_archs.a']`

Note that the arch-specific stuff are single files or alternatives (tuple), not a full list of files (again) to keep it simple.

### DIFF
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -405,7 +405,8 @@ def to_sanity_check_paths_entry(spec):
                                          value, type(value))
             result.append(elem)
         else:
-            raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)", elem, type(elem))
+            raise EasyBuildError("Expected elements to be of type string, tuple/list or dict, got %s (%s)",
+                                 elem, type(elem))
 
     return result
 
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -375,6 +375,43 @@ def to_list_of_strings_and_tuples_and_dicts(spec):
     return str_tup_list
 
 
+def to_sanity_check_paths_entry(spec):
+    """
+    Convert a 'list of lists and strings' to a 'list of tuples and strings' while allowing dicts of lists or strings
+
+    Example:
+        ['foo', ['bar', 'baz'], {'f42': ['a', 'b']}]
+        to
+        ['foo', ('bar', 'baz'), {'f42': ('a', 'b')}]
+    """
+    result = []
+
+    if not isinstance(spec, (list, tuple)):
+        raise EasyBuildError("Expected value to be a list, found %s (%s)", spec, type(spec))
+
+    for elem in spec:
+        if isinstance(elem, (string_type, tuple)):
+            result.append(elem)
+        elif isinstance(elem, list):
+            result.append(tuple(elem))
+        elif isinstance(elem, dict):
+            result.append({})
+            for key, value in elem.items():
+                if not isinstance(key, string_type):
+                    raise EasyBuildError("Expected keys to be of type string, got %s (%s)", key, type(key))
+                if isinstance(value, (string_type, tuple)):
+                    result[-1][key] = value
+                elif isinstance(value, list):
+                    result[-1][key] = tuple(value)
+                else:
+                    raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)",
+                                         value, type(value))
+        else:
+            raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)", elem, type(elem))
+
+    return result
+
+
 def to_sanity_check_paths_dict(spec):
     """
     Convert a sanity_check_paths dict as received by yaml (a dict with list values that contain either lists or strings)
@@ -389,7 +426,7 @@ def to_sanity_check_paths_dict(spec):
 
     sanity_check_dict = {}
     for key in spec:
-        sanity_check_dict[key] = to_list_of_strings_and_tuples(spec[key])
+        sanity_check_dict[key] = to_sanity_check_paths_entry(spec[key])
     return sanity_check_dict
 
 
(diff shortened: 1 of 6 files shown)


boegel:
@Flamefire Are the failing tests making sense to you?

Flamefire:
@boegel They are secondary. The primary issue is that I basically changed the allowed format to accept dicts in addition to strings and tuples in the list. See https://github.com/easybuilders/easybuild-framework/pull/3845/files#diff-c63ede120b91884bda5e7d0df599ac170866c050ccec036bdf61c3d456846f43R572
However I never really understood that part of the code. I'd think it's better to have simple "conversion" functions (like I did with https://github.com/easybuilders/easybuild-framework/pull/3845/files#diff-c63ede120b91884bda5e7d0df599ac170866c050ccec036bdf61c3d456846f43R352) which errors out on wrong formats instead of having to specify a conversion function and some kind of type-specification.

The conversion function works, however it seems for yeb-files (the YAMLs) the type-spec is used and fails.
So I'll need help here as I thought the way I did it was correct.

Flamefire:
Ah I needed to add `SANITY_CHECK_PATHS_ENTRY` to `CHECKABLE_TYPES`. Now those tests work for me

easybuild/framework/easyconfig/types.py
```diff
@@ -375,6 +375,43 @@ def to_list_of_strings_and_tuples_and_dicts(spec):
...
+            for key, value in elem.items():
+                if not isinstance(key, string_type):
+                    raise EasyBuildError("Expected keys to be of type string, got %s (%s)", key, type(key))
+                if isinstance(value, (string_type, tuple)):
+                    result[-1][key] = value
+                elif isinstance(value, list):
+                    result[-1][key] = tuple(value)
+                else:
+                    raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)",
+                                         value, type(value))
```
akesandgren, line 408: This would be somewhat easier to read using
tmpdict = {}
...
tmpdict[key] = value
...
result.append(tmpdict)

using reault[-1] is kind of weird to python noobs

Flamefire, line 408: What about this even easier solution:
```suggestion
            for key, value in elem.items():
                if not isinstance(key, string_type):
                    raise EasyBuildError("Expected keys to be of type string, got %s (%s)", key, type(key))
                elif isinstance(value, list):
                    elem[key] = tuple(value)
                elif not isinstance(value, (string_type, tuple)):
                    raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)",
                                         value, type(value))
                result.append(elem)
```

akesandgren: Yeah that's better :-)

easybuild/framework/easyconfig/types.py
```diff
@@ -375,6 +375,43 @@ def to_list_of_strings_and_tuples_and_dicts(spec):
...
+        else:
+            raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)", elem, type(elem))
```
akesandgren, line 410: Strictly speaking this error text is wrong (since dicts with the same type of values are also ok) but I don't know if we should change it.

easybuild/framework/easyconfig/types.py
```diff
@@ -375,6 +375,41 @@ def to_list_of_strings_and_tuples_and_dicts(spec):
...
+                if not isinstance(key, string_type):
+                    raise EasyBuildError("Expected keys to be of type string, got %s (%s)", key, type(key))
+                elif isinstance(value, list):
+                    elem[key] = tuple(value)
+                elif not isinstance(value, (string_type, tuple)):
+                    raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)",
+                                         value, type(value))
+            result.append(elem)
+        else:
+            raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)", elem, type(elem))
```
Flamefire, line 408: Yeah. Maybe:
```suggestion
            raise EasyBuildError("Expected elements to be of type string, tuple/list or dict, got %s (%s)", elem, type(elem))
```

akesandgren: probably better that way yes.

easybuild/framework/easyconfig/types.py
```diff
@@ -375,6 +375,42 @@ def to_list_of_strings_and_tuples_and_dicts(spec):
...
+        elif isinstance(elem, dict):
+            for key, value in elem.items():
+                if not isinstance(key, string_type):
+                    raise EasyBuildError("Expected keys to be of type string, got %s (%s)", key, type(key))
+                elif isinstance(value, list):
+                    elem[key] = tuple(value)
+                elif not isinstance(value, (string_type, tuple)):
+                    raise EasyBuildError("Expected elements to be of type string, tuple or list, got %s (%s)",
+                                         value, type(value))
+            result.append(elem)
```
akesandgren, line 406: Strictly speaking, every value in a list (turned tuple) or tuple, should be of string_type, or at least not list/tuple/dicts, 
 and we should perhaps check that too, or?

The question then becomes, are we doing that strictly in all the other to_xxx functions?

I don't think we need to change this here, but rather take a look at all of them at the same time in a separate PR.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

easybuild/framework/easyconfig/types.py
Flamefire: You are right that should be done as one should (be able to) assume that what comes out from those functions is valid.
Semi-related: https://github.com/easybuilders/easybuild-framework/issues/4177

boegel:
@Flamefire Can you also update the documentation to mention that this is now supported?
Cfr. https://docs.easybuild.io/writing-easyconfig-files/#sanity-check

Flamefire:
> @Flamefire Can you also update the documentation to mention that this is now supported? Cfr. https://docs.easybuild.io/writing-easyconfig-files/#sanity-check

Not in the next couple weeks at least. A quick message is all I can do at the moment, a rebase when required is pushing my time limits already.

PR #3850, opened 2021-09-27
Ensure newer location of CUDA stubs is taken into account by RPATH  filter

### LINKED ISSUE #2683: CUDA stubs directory must not be added to rpath/runpath

ocaisa:
Fixes #2683 (again)

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -960,9 +960,11 @@ class Toolchain(object):
 
         # always include filter for 'stubs' library directory,
         # cfr. https://github.com/easybuilders/easybuild-framework/issues/2683
-        lib_stubs_pattern = '.*/lib(64)?/stubs/?'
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm

boegel:
Tested with `OSU-Micro-Benchmarks-5.7.1-gompi-2021a-CUDA-11.3.1.eb`, works as expected, thanks @ocaisa!

PR #3851, opened 2021-09-28
add support for collecting GPU info (via nvidia-smi), and include it in --show-system-info and test report

### LINKED ISSUE #3825: Report available GPUs and NVIDIA driver version in PR test reports
When looking at test reports it would be useful to know what GPUs were available during the build and what NVIDIA driver version is installed.

This would (probably) only be needed when testing software with a CUDA dependency.

Also, the equivalent for non-NVIDIA GPUs.

branfosj:
closes #3825

Add GPU info to
 * `eb --show-system-info`
 * the PR post for a test report

Only supports NVIDIA GPUs on linux as that is all I have.

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -588,15 +588,16 @@ def get_gpu_info():
             out, ec = run_cmd(cmd, force_in_dry_run=True, trace=False, stream_output=False)
             if ec == 0:
                 for line in out.strip().split('\n'):
-                    if 'NVIDIA' in gpu_info and line in gpu_info['NVIDIA']:
-                        gpu_info['NVIDIA'][line] += 1
-                    else:
-                        gpu_info['NVIDIA'] = {}
-                        gpu_info['NVIDIA'][line] = 1
-        except Exception:
-            _log.debug("No NVIDIA GPUs detected")
+                    nvidia_gpu_info = gpu_info.setdefault('NVIDIA', {})
+                    nvidia_gpu_info.setdefault(line, 0)
+                    nvidia_gpu_info[line] += 1
+            else:
+                _log.debug("None zero exit (%s) from nvidia-smi: %s" % (ec, out))
+        except Exception as err:
+            _log.debug("Exception was raised when running nvidia-smi: %s", err)
+            _log.info("No NVIDIA GPUs detected")
     else:
-        _log.debug("Only know how to get GPU info on Linux")
+        _log.info("Only know how to get GPU info on Linux, assuming no GPUs are present")
 
     return gpu_info
 
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -574,6 +574,33 @@ def get_cpu_features():
     return cpu_feat
 
 
+def get_gpu_info():
+    """
+    Get the GPU info
+    """
+    gpu_info = {}
+    os_type = get_os_type()
+
+    if os_type == LINUX:
+        try:
+            cmd = "nvidia-smi --query-gpu=gpu_name,driver_version --format=csv,noheader"
+            _log.debug("Trying to determine NVIDIA GPU info on Linux via cmd '%s'", cmd)
+            out, ec = run_cmd(cmd, force_in_dry_run=True, trace=False, stream_output=False)
+            if ec == 0:
+                for line in out.strip().split('\n'):
+                    if 'NVIDIA' in gpu_info and line in gpu_info['NVIDIA']:
+                        gpu_info['NVIDIA'][line] += 1
+                    else:
+                        gpu_info['NVIDIA'] = {}
+                        gpu_info['NVIDIA'][line] = 1
+        except Exception:
+            _log.debug("No NVIDIA GPUs detected")
+    else:
+        _log.debug("Only know how to get GPU info on Linux")
+
+    return gpu_info
+
+
 def get_kernel_name():
     """NO LONGER SUPPORTED: use get_os_type() instead"""
     _log.nosupport("get_kernel_name() is replaced by get_os_type()", '2.0')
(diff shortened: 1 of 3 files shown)


easybuild/tools/systemtools.py
```diff
@@ -574,6 +574,33 @@ def get_cpu_features():
...
+            cmd = "nvidia-smi --query-gpu=gpu_name,driver_version --format=csv,noheader"
+            _log.debug("Trying to determine NVIDIA GPU info on Linux via cmd '%s'", cmd)
+            out, ec = run_cmd(cmd, force_in_dry_run=True, trace=False, stream_output=False)
+            if ec == 0:
+                for line in out.strip().split('\n'):
+                    if 'NVIDIA' in gpu_info and line in gpu_info['NVIDIA']:
+                        gpu_info['NVIDIA'][line] += 1
+                    else:
+                        gpu_info['NVIDIA'] = {}
+                        gpu_info['NVIDIA'][line] = 1
```
boegel, line 595: `setdefault` is your friend here, no need for the `if`/`else`:

```python
for line in out.strip().split('\n'):
    nvidia_gpu_info = gpu_info.setdefault('NVIDIA', {})
    nvidia_gpu_info.setdefault(line, 0)
    nvidia_gpu_info[line] += 1
```

easybuild/tools/systemtools.py
```diff
@@ -574,6 +574,33 @@ def get_cpu_features():
...
+    """
+    gpu_info = {}
+    os_type = get_os_type()
+
+    if os_type == LINUX:
+        try:
+            cmd = "nvidia-smi --query-gpu=gpu_name,driver_version --format=csv,noheader"
+            _log.debug("Trying to determine NVIDIA GPU info on Linux via cmd '%s'", cmd)
+            out, ec = run_cmd(cmd, force_in_dry_run=True, trace=False, stream_output=False)
+            if ec == 0:
```
boegel, line 589: Shouldn't we at least log something if we hit a non-zero exit code?

easybuild/tools/systemtools.py
```diff
@@ -574,6 +574,33 @@ def get_cpu_features():
...
+                for line in out.strip().split('\n'):
+                    if 'NVIDIA' in gpu_info and line in gpu_info['NVIDIA']:
+                        gpu_info['NVIDIA'][line] += 1
+                    else:
+                        gpu_info['NVIDIA'] = {}
+                        gpu_info['NVIDIA'][line] = 1
+        except Exception:
+            _log.debug("No NVIDIA GPUs detected")
```
boegel, line 597: Log which exception was hit?

```python
except Exception as err:
    _log.debug("Exception was raised when running nvidia-smi: %s", err)
    _log.info("No NVIDIA GPUs detected")

File: easybuild/tools/systemtools.py [hunk 4]
```diff
@@ -574,6 +574,33 @@ def get_cpu_features():
...
+    else:
+        _log.debug("Only know how to get GPU info on Linux")
```
boegel, line 599:
```python
_log.info("Only know how to get GPU info on Linux, assuming no GPUs are present")

File: easybuild/tools/systemtools.py [hunk 5]
```diff
@@ -574,6 +574,33 @@ def get_cpu_features():
...
+def get_gpu_info():
+    """
+    Get the GPU info
+    """
+    gpu_info = {}
+    os_type = get_os_type()
+
+    if os_type == LINUX:
+        try:
+            cmd = "nvidia-smi --query-gpu=gpu_name,driver_version --format=csv,noheader"
```
boegel, line 586: Maybe check first whether `nvidia-smi` command is available (via `which` function from `filetools`)?

boegel (changes requested):

easybuild/tools/systemtools.py
branfosj: Thanks. Done in 2e34420

easybuild/tools/systemtools.py
branfosj: Yes. See 2e34420

easybuild/tools/systemtools.py
branfosj: Done in 2e34420

easybuild/tools/systemtools.py
branfosj: 2e34420

easybuild/tools/systemtools.py
branfosj: We are inside a `try` / `except` block, so that will catch and report `nvidia-smi` being unavailable.

branfosj:
Currently looks like: https://github.com/easybuilders/easybuild-easyconfigs/pull/14069#issuecomment-929386339

branfosj:
Example PR test message at https://github.com/easybuilders/easybuild-easyconfigs/pull/14069#issuecomment-929399434

easybuild/tools/testing.py
```diff
@@ -295,11 +295,20 @@ def post_pr_test_report(pr_nrs, repo_type, test_report, msg, init_session_state,
     if system_info['cpu_arch_name'] != UNKNOWN:
         system_info['cpu_model'] += " (%s)" % system_info['cpu_arch_name']
 
+    # add GPU info, if known
+    gpu_info = get_gpu_info()
+    gpu_str = ""
+    if gpu_info:
+        for vendor in gpu_info:
+            for gpu, num in gpu_info[vendor].items():
+                gpu_str += ", %s %s x %s" % (vendor, num, gpu)
```
SebastianAchilles, line 304:
```suggestion
                gpu_str += ", %s x %s %s" % (num, vendor, gpu)
```

Personally I prefer `1 x NVIDIA Tesla V100-SXM2-16GB` because I think it is easier to read compared to `NVIDIA 1 x Tesla V100-SXM2-16GB`.

easybuild/tools/systemtools.py
boegel: Yeah, ok

boegel (changes requested):
Some minor additional suggested changes in https://github.com/branfosj/easybuild-framework/pull/4

With that included, this is good to imho.

Unless we want to go through the effort of extending the framework tests with a fake `nvidia-smi` command that produces output to be processed by `get_gpu_info`?

boegel (approved):

boegel:
Thanks a lot @branfosj!

PR #3852, opened 2021-09-29
replace which by command -v to avoid dependency on which

zemu-unile:
Same rational as on line 65, see  #2976


ocaisa (approved):
Thanks for catching this

PR #3889, opened 2021-10-30
only remove lock if it was created in the same EasyBuild session (not if it existed already)

boegel:


branfosj (approved):
lgtm

branfosj:
Going in, thanks @boegel!

PR #3890, opened 2021-10-31
determine which extensions can be skipped in parallel (if --parallel-extensions-install is enabled)

boegel:
When running sequentially, checking which extensions can be skipped in parallel for `R-4.1.0-foss-2021a.eb` takes about 30 minutes (incl. creating the `Extension` instances, which takes about 5 min.)

With this enhancement, this is sped up significantly using `--parallel-extensions-install --experimental` (measured on an AMD EPYC 7552 - Rome system, in a Slurm job with 24 cores available)

* Using all 24 cores: 11 min. 30 sec.
* Using `--parallel=6`: 14 min. sec.


branfosj (approved):
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/14299 and all is good

branfosj:
Going in, thanks @boegel!

PR #3893, opened 2021-11-07
take into account that repositorypath could specify remote repository via git@<server>:<org>/<repo>.git

### LINKED ISSUE #3892: Pushing easyconfigs to local git repo fails in 4.5.0
After we upgraded our EasyBuild installation to `4.5.0` we have noticed that the feature we use to send the easyconfigs and build logs to an internal git repo is no longer working. We have an ssh-key for our build user that works if we do a build with a previous EasyBuild version (`4.4.2` for example) but no longer works on `4.5.0`. At the end of the build we get:
```
/scicore/home/scicore/easybuil@bc2-gl.bc2.unibas.ch's password:
```
There seems to be some change in the way the `repositorypath` configuration option is processed, which in our config is:
```
repositorypath=git@bc2-gl.bc2.unibas.ch:easybuild/installed_easyconfigs_tests.git
```

@pescobar mentioned that this commit might be related https://github.com/easybuilders/easybuild-framework/commit/3252979f29b1b92a7223f5163cef02c7c03fa564

boegel:
fixes #3892

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1067,7 +1067,7 @@ class EasyBuildOptions(GeneralOption):
 
     def get_cfg_opt_abs_path(self, opt_name, path):
         """Get path value of configuration option as absolute path."""
-        if os.path.isabs(path):
+        if os.path.isabs(path) or path.startswith('git@'):
             abs_path = path
         else:
             abs_path = os.path.abspath(path)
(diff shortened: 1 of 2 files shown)


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #3899, opened 2021-11-18
don't try to ensure absolute path for path part of repositorypath

### LINKED ISSUE #3892: Pushing easyconfigs to local git repo fails in 4.5.0

boegel:
#3893 was an incomplete fix for #3892

The easy way out is to just leave the `repositorypath` value untouched (i.e. don't try to ensure it's an absolute path), which is difficult to do since it could specify the location to a *remote* repository in a variety of formats...

fixes #3892

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1092,21 +1092,19 @@ class EasyBuildOptions(GeneralOption):
     def _postprocess_config(self):
         """Postprocessing of configuration options"""
 
-        # resolve relative paths for configuration options that specify a location;
+        # resolve relative paths for configuration options that specify a location,
+        # to avoid incorrect paths being used when EasyBuild changes the current working directory
+        # (see https://github.com/easybuilders/easybuild-framework/issues/3619);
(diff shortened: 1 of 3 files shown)


bartoldeman (approved):
lgtm

PR #3900, opened 2021-11-19
introduce EasyBlock.post_init method to correctly define builddir variable when build-in-installdir mode is enabled in easyconfig or easyblock

### LINKED ISSUE #3895: builddir not set correctly when build_in_installdir = True and running --sanity-check-only
When `self.build_in_installdir = True` (c.f. OpenFOAM easyblock) and running `--sanity-check-only` then `self.builddir` is not set correctly.

Error:
```
cd: /dev/shm/build-branfosj-admin/branfosj-admin-up/OpenFOAM/8/foss-2020b: No such file or directory
```

That corresponds to
```
buildpath                 (E) = /dev/shm/build-branfosj-admin/branfosj-admin-up
```

boegel:
fixes #3895

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -268,11 +268,6 @@ class EasyBlock(object):
         # generate build/install directories
         self.gen_builddir()
         self.gen_installdir()
-        if self.build_in_installdir:
-            # self.builddir is set by self.gen_builddir(),
-            # but needs to be correct if the build is performed in the installation directory
-            self.log.info("Changing build dir to %s", self.installdir)
-            self.builddir = self.installdir
 
         self.ignored_errors = False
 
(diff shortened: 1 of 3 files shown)


branfosj (approved):

branfosj:
This looks correct, but is not solving the issue I am seeing with OpenFOAM.

```
eb OpenFOAM-8-foss-2020b.eb -Tr --sanity-check-only
```
with this PR is not hitting the right section of code.

branfosj:
So, setting `buildininstalldir = True` in the OpenFOAM easyconfig means this works correctly. However, the `self.build_in_installdir = True` in the OpenFOAM easyblock acts after this.

smoors:
a possible solution would then be to replace `self.build_in_installdir = True` with an EasyBlock method `self.build_in_installdir()`:
```python
build_in_installdir(self):
    if self.builddir != self.installdir:
        self.log.info("Changing build dir to %s", self.installdir)
        self.builddir = self.installdir
```

boegel:
> a possible solution would then be to replace `self.build_in_installdir = True` with an EasyBlock method `self.build_in_installdir()`

That's possible, but the downside is that easyblocks need to follow suite in order to pick up this bugfix...

I think a better option is to introduce a `post_init` method which is always run after initializing the `EasyBlock` instance (like I did in 96236290d), before any steps are run; not making that part of a step makes sure it's always run.

That way, existing easyblocks that set `build_in_installdir` to `True` in the constructor don't need to be changed.
There are only two exceptions in the central easybuild-easyblocks repository: the custom easyblocks for `PETSc` and `SLEPc` (which are among the oldest easyblocks we have).
I'll open a trivial PR to also make those two set `build_in_installdir` in the easyblock constructor.

branfosj (approved):
lgtm

branfosj:
Going in, thanks @boegel!

PR #3906, opened 2021-11-27
fall back to sequential installation for extensions with unknown dependencies

boegel:
This avoids a hard crash that occurs currently when trying to use `--parallel-extensions-install --experimental --robot` and an easyconfig that includes Python packages as extensions.

For example:

```
NotImplementedError: Don't know how to determine required dependencies for extension 'setuptools'
```

This occurs because the `PythonPackage` easyblock does not define the `required_deps` method, and hence the default implementation from the `Extension` easyblock is used.

The changes include here include:
* changing the default implementation of `required_deps` to just return `None`;
* fall back to sequential installation for an extension with unknown dependencies, that is wait until all preceding extensions have been installed first (since the order in the list of extensions should then be honored).

The added test verifies that this is working as expected.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1852,7 +1852,26 @@ class EasyBlock(object):
                 # check whether extension at top of the queue is ready to install
                 ext = exts_queue.pop(0)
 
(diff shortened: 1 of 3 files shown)


branfosj (approved):
lgtm

PR #3907, opened 2021-11-30
pick up custom extract_cmd specified for extension

boegel:


easybuild/tools/filetools.py
```diff
@@ -456,11 +456,14 @@ def extract_file(fn, dest, cmd=None, extra_options=None, overwrite=False, forced
...
+    if cmd:
         # complete command template with filename
         cmd = cmd % fn
+        _log.debug("Using specified command to unpack %s: %s", fn, cmd)
+    else:
+        cmd = extract_cmd(fn, overwrite=overwrite)
+        _log.debug("Using command derived from file extension to unpack %s: %s", fn, cmd)
+
     if not cmd:
         raise EasyBuildError("Can't extract file %s with unknown filetype", fn)
```
smoors, line 468: if `extract_cmd()` fails, this error will not be shown, right? if so, better to try-except instead:

```suggestion
        try:
            cmd = extract_cmd(fn, overwrite=overwrite)
            _log.debug("Using command derived from file extension to unpack %s: %s", fn, cmd)
        except KeyError as err:
            raise EasyBuildError("Can't extract file %s with unknown filetype", fn)
```

test/framework/toy_build.py
```diff
@@ -1362,6 +1362,34 @@ def test_toy_extension_sources(self):
...
             self.test_toy_build(ec_file=test_ec, raise_error=True)
 
+    def test_toy_extension_extract_cmd(self):
+        """Test for custom extract_cmd specified for an extension."""
+        test_ecs = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'test_ecs')
+        toy_ec = os.path.join(test_ecs, 't', 'toy', 'toy-0.0.eb')
+        toy_ec_txt = read_file(toy_ec)
+
+        # create file that we'll copy via 'patches'
+        write_file(os.path.join(self.test_prefix, 'test.txt'), 'test123')
```
smoors, line 1372: this not used in this test.

easybuild/tools/filetools.py
boegel: This `if not cmd` is a but superfluous, since the `file_extension` called in `extract_cmd` will already raise a meaningful error message (which I improved a bit in 1d4e45492).

I would like to keep this guard though, just in case we somehow accidentally make `extract_cmd` produce an empty extract command or something, it doesn't make sense to continue and let `eb` crash hard later with a cryptic error message in that case...

test/framework/toy_build.py
boegel: cleaned up in 1d4e45492981b80698aaa0702895191732b5ce3c

smoors:
tested with a custom PythonBundle easyconfig, and works as expected.

smoors (approved):
lgtm

smoors:
Going in, thanks @boegel!

PR #3911, opened 2021-12-09
add support for specifying a branch name in git_config + clarify warning message when trying to use 'tag' to specify a branch name

### LINKED ISSUE #3902: git_config confuses branches and tags
We used to install Spack on CSCS systems using EasyBuild. After the 4.4.2 release, the `tag` option from `git_config` doesn't change to the correct `develop` branch of Spack. EB outputs the warning and it fails to change branches:

```
...
... cmd "git checkout refs/tags/develop" exited with exit code 1 and output:
...
WARNING: Tag develop was not downloaded in the first try due to https://github.com/spack/spack containing a branch with the same name. You might want to alert the maintainers of spack about that issue.
```

I have tested and confirmed that versions are 4.4.2 and 4.5.0 give this error, while 4.4.1 works.

The easyconfig file that we use can be found [here](https://github.com/eth-cscs/production/blob/master/easybuild/easyconfigs/s/Spack/Spack-daint-git-develop.eb).

I suggest to try to reproduce the error using
```
eb Spack-daint-git-develop.eb --force-download -f
```

boegel:
closes #3902

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2578,8 +2578,11 @@ def get_source_tarball_from_git(filename, targetdir, git_config):
     url = git_config.pop('url', None)
     repo_name = git_config.pop('repo_name', None)
     commit = git_config.pop('commit', None)
+    branch = git_config.pop('branch', None)
     recursive = git_config.pop('recursive', False)
     keep_git_dir = git_config.pop('keep_git_dir', False)
+
+    commit_or_branch = commit or branch
 
     # input validation of git_config dict
     if git_config:
@@ -2588,11 +2591,11 @@ def get_source_tarball_from_git(filename, targetdir, git_config):
     if not repo_name:
         raise EasyBuildError("repo_name not specified in git_config parameter")
 
-    if not tag and not commit:
-        raise EasyBuildError("Neither tag nor commit found in git_config parameter")
+    if not tag and not commit_or_branch:
+        raise EasyBuildError("No tag, commit, or branch found in git_config parameter")
 
-    if tag and commit:
-        raise EasyBuildError("Tag and commit are mutually exclusive in git_config parameter")
+    if tag and commit_or_branch:
+        raise EasyBuildError("Tag and commit/branch are mutually exclusive in git_config parameter")
 
     if not url:
         raise EasyBuildError("url not specified in git_config parameter")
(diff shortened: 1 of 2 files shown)


Flamefire:
Note that checking out a branch should rather be done like checking out a tag. I mean the git param is named `--branch` for a reason and only additionally works for tags. This allows the shallow clone to work which is much faster for large repos. Only thing that: You need to make sure you checked out a branch and not a tag. I.e. the opposite as for the tag handling. Not fully sure how, but see https://github.com/easybuilders/easybuild-framework/issues/3902#issuecomment-989205202 for an idea.

Also we should somewhere say that `branch` is discouraged. I can't remember anything else where we support a download that changes over time, i.e. is literally impossible to reproduce and checksum. So I'm not sure we actually need that support.

boegel:
> Note that checking out a branch should rather be done like checking out a tag. I mean the git param is named `--branch` for a reason and only additionally works for tags. This allows the shallow clone to work which is much faster for large repos. 

Are sure sure that a shallow clone is sufficient to be able to check out a random branch?
The HEAD commit of that branch needs to be available, so how do you know you can do a shallow clone?

> Only thing that: You need to make sure you checked out a branch and not a tag. I.e. the opposite as for the tag handling. Not fully sure how, but see [#3902 (comment)](https://github.com/easybuilders/easybuild-framework/issues/3902#issuecomment-989205202) for an idea.

Oh boy...

> 
> Also we should somewhere say that `branch` is discouraged. I can't remember anything else where we support a download that changes over time, i.e. is literally impossible to reproduce and checksum. So I'm not sure we actually need that support.

Right, good point. I don't mind supporting it, but I guess it should come with a big fat warning since a branch name is essentially a moving target.
In practice, you always need to use `--force-download` to make sure you get the latest `develop`.

And for the Spack easyconfig specifically (cfr. #3902), there's actually no point at all in using `git_config`: they can just use https://github.com/spack/spack/archive/develop.tar.gz to download (since they're not retaining `.git`, or need a recursive checkout).

So should our advice for now be "use `'develop'` as value for `'commit'` (rather than `'tag'`)" (that does work as expected, I checked that), but be aware that a branch name is a moving target?
That's a viable workaround for the broken use case (which was actually working by accident before, in some sense).

Flamefire:
> Are sure sure that a shallow clone is sufficient to be able to check out a random branch?
The HEAD commit of that branch needs to be available, so how do you know you can do a shallow clone?

Assuming only. I mean you are not "checking out" a random branch but cloning a repo with a named object, i.e. a tag or a branch. And given that we do that for e.g. "develop" already (see the bug report) and the parameter name `--branch` I assume it just works. This is what I mean with: "checking out a branch should rather be done like checking out a tag"
So yeah, I guess that's the source of confusion here: You have to checkout a commit but you can clone a tag or branch. And the latter can be shallow. while the former cannot. Shallow is also not possible if there is a tag and branch with the same name and you want the tag.

PR #3920, opened 2021-12-28
deprecate use of patch files that don't have a filename ending with .patch

### LINKED ISSUE #3688: surprising behavior with patch files that don't have .patch extension

smoors:
fixes #3688

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1493,6 +1493,9 @@ def create_patch_info(patch_spec):
                                  str(patch_spec))
 
     elif isinstance(patch_spec, string_type):
+        if not patch_spec.endswith('.patch'):
+            _log.deprecated("Add '.patch' suffix to patch file, or use 2-element list/tuple to specify "
+                            "path to where non-patch file should be copied: %s" % patch_spec, '5.0')
         patch_info = {'name': patch_spec}
     else:
         error_msg = "Wrong patch spec, should be string of 2-tuple with patch name + argument: %s"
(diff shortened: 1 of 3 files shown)


easybuild/tools/filetools.py
```diff
@@ -1493,6 +1493,9 @@ def create_patch_info(patch_spec):
                                  str(patch_spec))
 
     elif isinstance(patch_spec, string_type):
+        if not patch_spec.endswith('.patch'):
+            _log.deprecated("Add '.patch' suffix to patch file, or use 2-element list/tuple to specify "
+                            "path to where non-patch file should be copied: %s" % patch_spec, '5.0')
```
boegel, line 1498: @smoors I would tweak this a bit, to have a cleaner deprecation warning message, see https://github.com/boegel/easybuild-framework/commit/0764aeb7ef4e0c874ae30ed2f2e5e13a0742dc5f

boegel (changes requested):

ocaisa (approved):

ocaisa:
@boegel This looks ok to me, I leave it to you to merge since you already reviewed it

boegel (approved):

Issue #3513, opened 2020-12-01
add (experimental) support for general options in Easystack


deniskristak:
related to: [easystack file pull request](https://github.com/easybuilders/easybuild-framework/pull/3479)

Additional command-line arguments to EasyBuild can be passed at any level (may even be included outside of `software` scope by specifying a keyword that matches the long argument name, e.g.

```
robot: True
software:
  GROMACS:
    parallel: 6
    toolchains:
      foss-2020a:
        versions:
          2020.1:
          2020.3:
            from_pr: 1234
```
Will be equivalent to running
```

eb GROMAS-2020.1-foss-2020a.eb --robot --parallel=6
eb GROMAS-2020.3-foss-2020a.eb --robot --parallel=6 --from-pr=1234
```

The deepest nested value will always take priority, i.e.
```

software:
  GROMACS:
    parallel: 6
    toolchains:
      foss-2020a:
        versions:
          2020.1:
          2020.3:
            parallel: 12
```
will be equivalent to running

```
eb GROMAS-2020.1-foss-2020a.eb --parallel=6
eb GROMAS-2020.3-foss-2020a.eb --parallel=12
```

boegel:
```yaml
robot: True
software:
  GROMACS:
    parallel: 6
    toolchains:
      foss-2020a:
        versions:
          2020.1:
          2020.3:
            from_pr: 1234
            include_easyblocks_from_pr: 123
            debug: True
            trace: True
```

```
# current approach:
# eb example.yaml
eb GROMACS-2020.1-foss-2020a.eb GROMACS-2020.3-foss-2020a.eb --robot --parallel 6
# 1) eb /path/to/GROMACS-2020.1-foss-2020a.eb --parallel 6
# 2) eb /path/to/GROMACS-2020.3-foss-2020a.eb --parallel 6 --from-pr 1234 --include-easyblocks-from-pr 123 --debug --trace
```

High-level overview of things to change to support using custom configuration options for specific installations

* [REQ] `parse_easystack` returns list of (easyconfig filename, cfg options for this easyconfig file)
* `det_copy_ec_specs` and `categorize_files_by_type` have to be aware of cfg options per easyconfig file
* [REQ for `--from-pr`] `det_easyconfig_paths` has to take into account things like `from-pr` cfg option per easyconfig file
* `determined_paths` -> `paths`
* [REQ] `parse_easyconfigs` has to store cfg options per easyconfig file in `EasyConfig` instance
[...]

PR #3923, opened 2022-01-11
add support for update_build_options + enhance update_build_options to return original value 

casparvl:
To be able to define (per-)EasyConfig specific options in an `EasyStack` file, as discussed [here](https://github.com/easybuilders/easybuild-framework/issues/3513#issuecomment-986990195) we need to easily be able to set entire dictionaries of new configuration options - and restore them to their original values later.

This PR provides that functionality by 

1) extending `update_build_option` so that it returns the original value (so it can later be restored)

2) implementing `update_build_options`, which takes a dictionary, calls `update_build_option` on each key-value pair in the dictionary, and collects & returns the original values for each in another dictionary. This allows one to call _another_ `update_build_options` on the returned dictionary to reset the original values.

Example usage:

```
# Have a dictionary of dictionaries, each specifying a list of options that need to be changed for that specific EasyConfig
    opts_per_ec = {
        'OpenFOAM-8-foss-2020a': {
            'parallel': '6',
            'robot': True,
        },
        'GROMACS-2020.4-foss-2020a-Python-3.8.2': {
            'debug': True,
            'trace': True,
            'from_pr': '1234',
            'include_easyblocks_from_pr': 123,
            'parallel': 6,
            'robot': True,
        },
    }
...
   old_val_dict = update_build_options(opts_per_ec['OpenFOAM-8-foss-2020a'])
...
   # actually build OpenFOAM-8-foss-202a, now using the options that were set of this specific EasyConfig
...
   # Restore the original configuration, so that the original configuration is used to build any _further_ EasyConfigs
   update_build_options(old_val_dict)
...
```

Of course, the `opts_per_ec` would normally not be hardcoded, but will (eventually) be returned by the `parse_easystack` function if it encounters EasyConfig-specific options in the EasyStack file.

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -593,6 +593,23 @@ def update_build_option(key, value):
     build_options._FrozenDict__dict[key] = value
     _log.warning("Build option '%s' was updated to: %s", key, build_option(key))
 
+    # Return original value, so they can be restored later if needed
+    return orig_value
+
+def update_build_options(key_value_dict):
+    """
+    Update all build options in the key_value_dict with the value given in that dictionary,
+    by calling update_build_option(key, value) repeatedly.
+    This function can be used e.g. when EasyConfig-specific build options are passed in an EasyStack file.
+    See https://github.com/easybuilders/easybuild-framework/issues/3513#issuecomment-986990195
+    """
+    orig_key_value_dict = {}
+    for key, value in key_value_dict.items():
+        orig_key_value_dict[key] = update_build_option(key, value)
+
+    # Return original key-value pairs in a dictionary.
+    # This way, they can later be restored by a single call to update_build_options(orig_key_value_dict)
+    return orig_key_value_dict
 
 def build_path():
     """
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -590,7 +590,7 @@ def update_build_option(key, value):
     """
     # BuildOptions() is a (singleton) frozen dict, so this is less straightforward that it seems...
     build_options = BuildOptions()
-    orig_value = build_options.__FrozenDict__dict[key]
+    orig_value = build_options._FrozenDict__dict[key]
     build_options._FrozenDict__dict[key] = value
     _log.warning("Build option '%s' was updated to: %s", key, build_option(key))
 
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -600,8 +600,8 @@ def update_build_option(key, value):
 
 def update_build_options(key_value_dict):
     """
-Update build options as specified by the given dictionary (where keys are assumed to be build option names).
-Returns dictionary with original values for the updated build options.
+    Update build options as specified by the given dictionary (where keys are assumed to be build option names).
+    Returns dictionary with original values for the updated build options.
     """
     orig_key_value_dict = {}
     for key, value in key_value_dict.items():
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -590,11 +590,13 @@ def update_build_option(key, value):
     """
     # BuildOptions() is a (singleton) frozen dict, so this is less straightforward that it seems...
     build_options = BuildOptions()
+    orig_value = build_options.__FrozenDict__dict[key]
     build_options._FrozenDict__dict[key] = value
     _log.warning("Build option '%s' was updated to: %s", key, build_option(key))
 
     # Return original value, so they can be restored later if needed
     return orig_value
+
 
 def update_build_options(key_value_dict):
     """
@@ -610,6 +612,7 @@ def update_build_options(key_value_dict):
     # Return original key-value pairs in a dictionary.
     # This way, they can later be restored by a single call to update_build_options(orig_key_value_dict)
     return orig_key_value_dict
+
 
 def build_path():
     """
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -600,10 +600,8 @@ def update_build_option(key, value):
 
 def update_build_options(key_value_dict):
     """
-    Update all build options in the key_value_dict with the value given in that dictionary,
-    by calling update_build_option(key, value) repeatedly.
-    This function can be used e.g. when EasyConfig-specific build options are passed in an EasyStack file.
-    See https://github.com/easybuilders/easybuild-framework/issues/3513#issuecomment-986990195
+Update build options as specified by the given dictionary (where keys are assumed to be build option names).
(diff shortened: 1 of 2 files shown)


easybuild/tools/config.py
```diff
@@ -590,9 +590,29 @@ def update_build_option(key, value):
     """
     # BuildOptions() is a (singleton) frozen dict, so this is less straightforward that it seems...
     build_options = BuildOptions()
+    orig_value = build_options._FrozenDict__dict[key]
     build_options._FrozenDict__dict[key] = value
     _log.warning("Build option '%s' was updated to: %s", key, build_option(key))
 
+    # Return original value, so they can be restored later if needed
```
boegel, line 597:
```suggestion
    # Return original value, so it can be restored later if needed
```

easybuild/tools/config.py
```diff
@@ -590,9 +590,29 @@ def update_build_option(key, value):
...
+    return orig_value
+
+
+def update_build_options(key_value_dict):
+    """
+    Update all build options in the key_value_dict with the value given in that dictionary,
+    by calling update_build_option(key, value) repeatedly.
+    This function can be used e.g. when EasyConfig-specific build options are passed in an EasyStack file.
+    See https://github.com/easybuilders/easybuild-framework/issues/3513#issuecomment-986990195
```
boegel, line 606: In which contexts this can be used is irrelevant for this particular imho, and I wouldn't mention variable or even function names, that's just more places to adjust if we happen to do some refactoring.

```suggestion
Update build options as specified by the given dictionary (where keys are assumed to be build option names).
Returns dictionary with original values for the updated build options.
```

boegel (changes requested):
@casparvl We should enhance the existing test for `update_build_option` in `test/framework/config.py` to verify the return value too, and we should add a test for `update_build_options` as well (can be added to the existing `test_update_build_option` imho)

casparvl:
> @casparvl We should enhance the existing test for `update_build_option` in `test/framework/config.py` to verify the return value too, and we should add a test for `update_build_options` as well (can be added to the existing `test_update_build_option` imho)

You're right. I added a test for the returned value. Also, I included a test for `update_build_options`. I actually made it a seperate one, sorry, overlooked your remark that you wanted to add it in the same test. If you want me to integrate them, we can. But I think it's quite nice to test it seperately, since a test failure will then immediately tell you if it's in the `update_build_option` (in which case both tests fail), or in the 'logic' (looping over the dictionary, adding the returned values to a dictionary) of `update_build_options`.

easybuild/tools/config.py
```diff
@@ -590,9 +590,27 @@ def update_build_option(key, value):
...
     _log.warning("Build option '%s' was updated to: %s", key, build_option(key))
 
+    # Return original value, so it can be restored later if needed
+    return orig_value
+
+
+def update_build_options(key_value_dict):
+    """
+Update build options as specified by the given dictionary (where keys are assumed to be build option names).
+Returns dictionary with original values for the updated build options.
```
boegel, line 604: Please indent to match the `"""` line above

test/framework/config.py
```diff
@@ -681,8 +682,29 @@ def test_update_build_option(self):
         """Test updating of a build option."""
         self.assertEqual(build_option('banned_linked_shared_libs'), None)
 
-        update_build_option('banned_linked_shared_libs', '/usr/lib64/libssl.so.1.1')
+        old_val = update_build_option('banned_linked_shared_libs', '/usr/lib64/libssl.so.1.1')
         self.assertEqual(build_option('banned_linked_shared_libs'), '/usr/lib64/libssl.so.1.1')
+        self.assertEqual(old_val, None)
```
boegel, line 687: Maybe extend this a bit to also update a build option that has a very specific default value, like:

```python
orig_cleanup_builddir = update_build_option('cleanup_builddir', False)
self.assertFalse(build_option('cleanup_builddir'))
self.assertTrue(orig_cleanup_builddir)

orig_pr_target_account = update_build_option('pr_target_account', 'test_pr_target_account')
self.assertEqual(build_option('pr_target_account'), 'test_pr_target_account')
self.assertEqual(orig_pr_target_account, 'easybuilders')
```

Because `None` is what you'd also get if there's no `return` at all...

test/framework/config.py
```diff
@@ -681,8 +682,29 @@ def test_update_build_option(self):
...
+
+    def test_update_build_options(self):
+        """Test updating of a dictionary of build options."""
+        self.assertEqual(build_option('banned_linked_shared_libs'), None)
+        self.assertEqual(build_option('filter_env_vars'), None)
```
boegel, line 692: Also use `cleanup_builddir` and `pr_target_account` here to test `update_build_options`?

boegel (changes requested):

easybuild/tools/config.py
casparvl: Haha, hey, I committed _your_ suggested change here! ;-) OK ok... will do :)

test/framework/config.py
casparvl: Done!

test/framework/config.py
casparvl: Sure, why not :) done!

casparvl:
Hm, I'm looking at the one failing test, but not sure what's wrong there. Seems to be a part of the code that I didn't touch at all, and only in one of the build configurations. A fluke? Can I trigger a rerun of the test suite (in some way more elegant than a recommit?)

boegel:
> Hm, I'm looking at the one failing test, but not sure what's wrong there. Seems to be a part of the code that I didn't touch at all, and only in one of the build configurations. A fluke? Can I trigger a rerun of the test suite (in some way more elegant than a recommit?)

```
FAIL: test_toy_lock_cleanup_signals (test.framework.toy_build.ToyBuildTest)
Test cleanup of locks after EasyBuild session gets a cancellation signal.
  File "/tmp/runner/a1a9cf08297f1c3dc8551fe19465b35dbcecf7ed/lib/python3.5/site-packages/test/framework/toy_build.py", line 3536, in test_toy_lock_cleanup_signals
    self.assertTrue(regex.search(stderr), "Pattern '%s' found in: %s" % (regex.pattern, stderr))
AssertionError: None is not true : Pattern '^WARNING: signal received \(3\), cleaning up locks \(.*software_toy_0.0\)\.\.\.' found in: WARNING: signal received (3), cleaning up locks ()...
```

Yes, that's a fluke... That test fails occasionally, and it has proven quite difficult to chase it down.

Please open an issue on it. I'll retrigger the tests for this PR.

boegel (approved):
lgtm

PR #3927, opened 2022-01-15
simplify requirements.txt by removing python 2.6 support

branfosj:
We removed Python 2.6 support in #3715, so we can simplify `requirements.txt` to no longer support that Python version.


requirements.txt
```diff
@@ -1,51 +1,28 @@
...
 
 # GitPython 3.1.15 deprecates Python 3.5
 GitPython==3.1.14; python_version >= '3.0' and python_version < '3.6'
-# GitPython 2.1.9 no longer supports Python 2.6
-GitPython==2.1.8; python_version < '2.7'
-GitPython; python_version >= '3.6' or (python_version <= '3.0' and python_version >= '2.7')
+GitPython; python_version >= '3.6' or python_version <= '3.0'
 
+# pydot
+pydot
```
boegel, line 14: @branfosj This can be removed, we only had this as a separate entry to control which `pydot` version is used as dependency for `python-graph-dot`

boegel (changes requested):

requirements.txt
branfosj: done in a920e85

boegel (approved):

PR #3928, opened 2022-01-15
use sources.easybuild.io as backup URL for downloading Environment Modules sources to run test suite

boegel:
Download of `modules-4.1.4.tar.gz` is failing due to trouble with SourceForge, we can work around that by using https://sources.easybuild.io/e/EnvironmentModules/ as backup download URL...

### DIFF
--- a/easybuild/scripts/install_eb_dep.sh
+++ b/easybuild/scripts/install_eb_dep.sh
@@ -18,12 +18,13 @@ PRECONFIG_CMD=
 
 if [ "$PKG_NAME" == 'modules' ] && [ "$PKG_VERSION" == '3.2.10' ]; then
     PKG_URL="http://prdownloads.sourceforge.net/modules/${PKG}.tar.gz"
-    BACKUP_PKG_URL="https://easybuilders.github.io/easybuild/files/${PKG}.tar.gz"
+    BACKUP_PKG_URL="https://sources.easybuild.io/e/EnvironmentModules/${PKG}.tar.gz"
     export PATH="$PREFIX/Modules/$PKG_VERSION/bin:$PATH"
     export MOD_INIT="$PREFIX/Modules/$PKG_VERSION/init/bash"
 
 elif [ "$PKG_NAME" == 'modules' ]; then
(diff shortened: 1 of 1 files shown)


branfosj:
lgtm

easybuild/scripts/install_eb_dep.sh
```diff
@@ -18,12 +18,13 @@ PRECONFIG_CMD=
...
-    BACKUP_PKG_URL="https://easybuilders.github.io/easybuild/files/${PKG}.tar.gz"
+    BACKUP_PKG_URL="https://sources.easybuild.io/e/EnvironmentModules/${PKG}.tar.gz"
     export PATH="$PREFIX/Modules/$PKG_VERSION/bin:$PATH"
     export MOD_INIT="$PREFIX/Modules/$PKG_VERSION/init/bash"
 
 elif [ "$PKG_NAME" == 'modules' ]; then
     PKG_URL="http://prdownloads.sourceforge.net/modules/${PKG}.tar.gz"
+    BACKUP_PKG_URL="https://sources.easybuild.io/e/EnvironmentModules/${PKG}.tar.gz"
     export PATH="$PREFIX/bin:$PATH"
     export MOD_INIT="$PREFIX/init/bash"
```
migueldiascosta, line 29: wouldn't this be clearer?

```bash
if [ "$PKG_NAME" == 'modules' ]; then
    PKG_URL="http://prdownloads.sourceforge.net/modules/${PKG}.tar.gz"
    BACKUP_PKG_URL="https://sources.easybuild.io/e/EnvironmentModules/${PKG}.tar.gz"
    if [ "$PKG_VERSION" == '3.2.10' ]; then
        export PATH="$PREFIX/Modules/$PKG_VERSION/bin:$PATH"
        export MOD_INIT="$PREFIX/Modules/$PKG_VERSION/init/bash"
    else
        export PATH="$PREFIX/bin:$PATH"
        export MOD_INIT="$PREFIX/init/bash"
    fi
```

boegel: Yes, indeed, fixed in 9658e4c4a

migueldiascosta (approved):
lgtm

PR #3930, opened 2022-01-18
also take /etc/os-release into account in get_os_name and get_os_version

boegel:
For recent Python versions, where `platform.linux_distribution` is no longer available, we're not getting useful info on OS name/version unless the 3rd party `distro` package is installed.

Falling back to `/etc/os-release` largely fixes that...

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -671,11 +671,15 @@ def get_os_name():
         # platform.linux_distribution is more useful, but only available since Python 2.6
         # this allows to differentiate between Fedora, CentOS, RHEL and Scientific Linux (Rocks is just CentOS)
         os_name = platform.linux_distribution()[0].strip()
-    elif HAVE_DISTRO:
+
+    # take into account that on some OSs, platform.distribution returns an empty string as OS name,
+    # for example on OpenSUSE Leap 15.2
+    if not os_name and HAVE_DISTRO:
         # distro package is the recommended alternative to platform.linux_distribution,
         # see https://pypi.org/project/distro
         os_name = distro.name()
-    elif os.path.exists(ETC_OS_RELEASE):
+
+    if not os_name and os.path.exists(ETC_OS_RELEASE):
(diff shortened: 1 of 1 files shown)


easybuild/tools/systemtools.py
```diff
@@ -669,11 +670,17 @@ def get_os_name():
     if hasattr(platform, 'linux_distribution'):
         # platform.linux_distribution is more useful, but only available since Python 2.6
         # this allows to differentiate between Fedora, CentOS, RHEL and Scientific Linux (Rocks is just CentOS)
-        os_name = platform.linux_distribution()[0].strip().lower()
+        os_name = platform.linux_distribution()[0].strip()
     elif HAVE_DISTRO:
```
SebastianAchilles, line 674:
```suggestion
    if HAVE_DISTRO and not os_name:
```

easybuild/tools/systemtools.py
```diff
@@ -700,6 +707,19 @@ def get_os_version():
         os_version = platform.dist()[1]
     elif HAVE_DISTRO:
```
SebastianAchilles, line 708:
```suggestion
    if HAVE_DISTRO and not os_version:
```

SebastianAchilles (changes requested):
On some OS platform.linux_distribution doesn't yield any relevant information, e.g. on OpenSUSE Leap 15.2:

```
>>> platform.linux_distribution()
('', '', '')
```

We need to catch this.

easybuild/tools/systemtools.py
boegel: fixed in 82cfcc8

easybuild/tools/systemtools.py
boegel: also fixed in 82cfcc8

SebastianAchilles (changes requested):
To fix `UnboundLocalError: local variable 'os_version' referenced before assignment`

SebastianAchilles (approved):
lgtm

PR #3936, opened 2022-01-23
take into account that patch files can also be zipped when checking filename extension for patches

boegel:
Support for zipped patch files was implemented in #2128, and we didn't take this into account in #3920 

This is rarely used (in the central easyconfigs), so it was easy to overlook...


migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

smoors (commented):
lgtm

PR #3945, opened 2022-01-27
switch to using pip3 for installing EasyBuild in Singularity definition file generated by EasyBuild

scimerman:
This is a solution for currently not working python2 pip problem. The problem happens when created definition file is being build, and it fails at the stage of updating pip. The update for pip on python 2 is failing with error, and the singularity sif file is not produced.
I do not see the reason why the container should even need to use the python 2. The support for python2 pip is expected to get even worse in the future.

The error was reported: easybuilders/easybuild-framework#3956

The users are currently either 1) creating the singularity definiton file, manually updating it, and then build with singulairty,   2) updating the template and then linking the creation of singularity definition on that one OR 3) not able to use this functionality.

I have checked and the current "[packagemanager] install python-pip" and "python" can be easily replaced with "[packagemanager] install python3-pip" and python3. I have checked, and the packages python3-pip and python3 exists on the ubuntu, centos/redhat/fedora and sles linux environments.

edit: fixes easybuilders/easybuild-framework#3956


ocaisa:
@scimerman The CI was failing because a test needed to be updated to reflect this change, hopefully should be fixed now.

scimerman:
I had a feeling, as test it was expecting a line with "pip ..." instead "pip3 ...". I am holding my fingers crossed  :)

easybuild/tools/config.py
```diff
@@ -134,7 +134,7 @@
 JOB_DEPS_TYPE_ABORT_ON_ERROR = 'abort_on_error'
 JOB_DEPS_TYPE_ALWAYS_RUN = 'always_run'
 
-DOCKER_BASE_IMAGE_UBUNTU = 'ubuntu:16.04'
+DOCKER_BASE_IMAGE_UBUNTU = 'ubuntu:20.04'
```
ocaisa, line 137: Ubuntu 16.04 is no longer supported since April 2021

ocaisa:
@scimerman Can you please merge https://github.com/scimerman/easybuild-framework/pull/1/files (which should fix the tests again)

ocaisa:
While we are doing this I wanted to actually see if I can fix the Docker recipes as well

scimerman:
> @scimerman Can you please merge [scimerman#1](https://github.com/scimerman/easybuild-framework/pull/1) (which should fix the tests again)

:+1: done

ocaisa:
@scimerman Another PR to merge here (fixes Docker support) scimerman#2

ocaisa:
For posterity, to check the Docker images I ran:
```
eb  --force --containerize --container-type=docker --experimental --container-config=ubuntu:20.04 zlib-1.2.11.eb
eb  --force --containerize --container-type=docker --experimental --container-config=centos:7 zlib-1.2.11.eb
```
and verified that I could get a shell inside the containers with the module pre-loaded

ocaisa (changes requested):
Once scimerman#2 is merged this should be good to go...but I am no longer the one who should merge it since a lot of the commits are now mine...@boegel?

boegel (approved):
lgtm, thanks for the work on this @scimerman and @ocaisa!

I've opened #3958 to catch the problems being fixed here in CI, so we at least know when it's broken through the test suite...

PR #3958, opened 2022-02-09
add end2end test for 'eb --containerize'

boegel:
Additional workflow to verify that experimental support for generating container recipes/images works.

This workflow reproduces the problem being fixed in #3945 (and without #3945 being merged, the test will fail)


.github/workflows/container_tests.yml
```diff
@@ -0,0 +1,92 @@
...
+          # create $HOME/.rpmmacros, see also https://github.com/apptainer/singularity/issues/241
+          echo '%_var /var' > $HOME/.rpmmacros
+          echo '%_dbpath %{_var}/lib/rpm' >> $HOME/.rpmmacros
+          # build CentOS 7 container image for bzip2 1.0.8 using EasyBuild;
+          # see https://docs.easybuild.io/en/latest/Containers.html
+          curl -OL https://raw.githubusercontent.com/easybuilders/easybuild-easyconfigs/develop/easybuild/easyconfigs/b/bzip2/bzip2-1.0.8.eb
+          export EASYBUILD_CONTAINERPATH=$PWD
+          export EASYBUILD_CONTAINER_CONFIG='bootstrap=yum,osversion=7'
+          eb bzip2-1.0.8.eb --containerize --experimental --container-build-image
+          test -f bzip2-1.0.8.sif
```
akesandgren, line 92: Why not a "singularity exec bzip2 -v" to finish things up?

boegel: How do I know I'm using the `bzip2` installed with `eb` :)

I can do a `which bzip2` in the container image, and check the path, maybe?

akesandgren: Yeah that would also work, or check bzip2 version outside and inside?
But some actual test of the container is probably in order here...

akesandgren: You could even first check version of bzip2 (if it's installed at all outside) and make sure to pick a version that isn't available outside...

boegel: I'll grep the output produce by a `which bzip2` , and then also run `bzip2 --version` in the container

.github/workflows/container_tests.yml
```diff
@@ -0,0 +1,92 @@
...
+          eb --version
+          # create $HOME/.rpmmacros, see also https://github.com/apptainer/singularity/issues/241
+          echo '%_var /var' > $HOME/.rpmmacros
+          echo '%_dbpath %{_var}/lib/rpm' >> $HOME/.rpmmacros
+          # build CentOS 7 container image for bzip2 1.0.8 using EasyBuild;
+          # see https://docs.easybuild.io/en/latest/Containers.html
+          curl -OL https://raw.githubusercontent.com/easybuilders/easybuild-easyconfigs/develop/easybuild/easyconfigs/b/bzip2/bzip2-1.0.8.eb
+          export EASYBUILD_CONTAINERPATH=$PWD
+          export EASYBUILD_CONTAINER_CONFIG='bootstrap=yum,osversion=7'
+          eb bzip2-1.0.8.eb --containerize --experimental --container-build-image
```
ocaisa, line 91: `bzip2` has a couple of deps including `binutils`, which  is why this CI is taking 10 mins. Are you concerned about ensuring `--robot` is working? If not, maybe `zlib` is enough...

boegel: I'm not too concerned about that, since this is a separate workflow, and the framework test suite also takes ~10min.

If it's really a nuisance, we can reconsider later?

.github/workflows/container_tests.yml
boegel: fixed in bdd34dd

ocaisa (approved):

Issue #3490, opened 2020-10-29
--merge-pr reporting "test suite passes: pending... => not eligible for merging!" even though all tests seem to have passed


migueldiascosta:
further inspection showed that GitHub is reporting two tests, 'https://github.com/apps/dependabot' and 'https://github.com/apps/hound', to have conclusion `None` instead of `Success`

this was in a PR to master, might be specific to this branch due to GitHub policies

follow up with GitHub issue?

migueldiascosta:
fwiw, this is still affecting https://github.com/easybuilders/easybuild-framework/pull/3526 and https://github.com/easybuilders/easybuild-easyblocks/pull/2277, at least (again, PRs to master)

boegel:
For #3526, see https://api.github.com/repos/easybuilders/easybuild-framework/commits/51535c5e131115fdaff06353e144f8228be05349/check-suites .

There's 3 "check suites" there that are stuck in `status: queued` and `conclusion: null`, for `dependabot`, `hound` and `travis-ci` (see `app.slug`).

The latter happens because we've run out of free credits in Travis CI, but we'll get rid of that by simply removing the `.travis.yml` configuration files going forward (which means we won't be tested with Python 2.6 anymore, but we've deprecated support for that a while ago anyway).

For the other two (`dependabot` and `hound`), I think we'll need to actively filter those out and simply ignore the results in `--merge-pr` for PRs to `master`, we can't just disable those checks somehow without having unwanted side affects (like not checking other branches, or totally disabling `dependabot`).

boegel:
@migueldiascosta Do we need to keep this open, after #3604 and #3607 ?

migueldiascosta:
@boegel I think this is a separate issue but I´ll check

migueldiascosta:
@boegel yeah, separate issue, and problem is still there. We may want to add something like

```diff
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -2274,6 +2274,7 @@ def det_commit_status(account, repo, commit_sha, github_user):
 
         # if any check suite hasn't completed yet, final result is still pending
         if status in ['queued', 'in_progress']:
+            print("%s (%s): %s" % (check_suite_data['app']['name'], check_suite_data['url'], check_suite_data['status']))
             result = STATUS_PENDING
 
         # if check suite is completed, take the conclusion into account
```

which in the case of https://github.com/easybuilders/easybuild-framework/pull/3640 outputted:

```
[...]

PR #3967, opened 2022-02-19
add a 'sync pr' message when the PR has a mergeable state but is showing a failed status for the test suite on the last commit

branfosj:
For #3490

If an easyconfigs PR fails the test suite, because a dep is in another PR or it needs a new / updated easyblock, then closing/reopening the PR to retrigger the test suite does not clear the `status_last_commit` even though the `mergeable_state` is fine (and the web UI will show the merge button). In this case we should give a suggestion to sync the PR.

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1155,12 +1155,14 @@ def check_pr_eligible_to_merge(pr_data):
 
     # check test suite result, Travis must give green light
     msg_tmpl = "* test suite passes: %s"
+    failed_status_last_commit = False
     if pr_data['status_last_commit'] == STATUS_SUCCESS:
         print_msg(msg_tmpl % 'OK', prefix=False)
     elif pr_data['status_last_commit'] == STATUS_PENDING:
         res = not_eligible(msg_tmpl % "pending...")
     else:
         res = not_eligible(msg_tmpl % "(status: %s)" % pr_data['status_last_commit'])
+        failed_status_last_commit = True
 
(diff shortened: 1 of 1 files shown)


branfosj:
The alternative to this is to treat `mergeable_state` as the test of if we can merge the PR, with `status_last_commit` only being used to determine a possible reason why the PR is not in a mergeable state - i.e. if the PR is not mergeable then that may because the CI is pending or failed.

easybuild/tools/github.py
```diff
@@ -1221,14 +1223,19 @@ def not_eligible(msg):
...
         print_msg(msg_tmpl % "PR is already merged", prefix=False)
     elif pr_data['mergeable_state'] == GITHUB_MERGEABLE_STATE_CLEAN:
         print_msg(msg_tmpl % "OK", prefix=False)
+        mergeable = True
     else:
         reason = "FAILED (mergeable state is '%s')" % pr_data['mergeable_state']
         res = not_eligible(msg_tmpl % reason)
 
+    if failed_status_last_commit and mergeable:
+        print_msg("\nThis PR is mergeable but the test suite has a failed status. Try syncing the PR.", prefix=False)
```
boegel, line 1237: Maybe `Try syncing the PR` should be a bit more specific, it could say "`Try syncing the PR with the develop branch using 'eb --sync-pr-with-develop XXX'`"?

boegel (changes requested):

easybuild/tools/github.py
branfosj: 20fdb74

boegel (approved):
lgtm

PR #3968, opened 2022-02-20
add end-to-end test for running EasyBuild in different Linux distros using containers

boegel:


.github/workflows/end2end.yml
```diff
@@ -0,0 +1,62 @@
...
+              echo ">>> sourcing /etc/profile.d/modules.sh"
+              . /etc/profile.d/modules.sh
+          else
+              echo ">>> sourcing /etc/profile.d/*lmod*.sh"
+              . /etc/profile.d/*lmod*.sh
+          fi
+
+          # tests are run with root privileges, so we need to tell EasyBuild that's OK...
+          export EASYBUILD_ALLOW_USE_AS_ROOT_AND_ACCEPT_CONSEQUENCES=1
+
```
lexming, line 50: I prefer to have atomic steps that do one thing. In this case we can limit this step to save environment variables in `GITHUB_ENV` for subsequent steps. I never tried to save the full environment, but the following should work

```suggestion

          while read line; do echo "$line" >> $GITHUB_ENV ; done < <(env)
          
      - name: Sanity check
        run: |
```

.github/workflows/end2end.yml
```diff
@@ -0,0 +1,62 @@
...
+            "eb --version"
+            "eb --show-system-info"
+            "eb --check-eb-deps"
+            "eb --show-config"
+            "eb bzip2-1.0.8.eb --trace --robot"
+          )
+          for cmd in "${cmds[@]}"; do
+              echo ">>> $cmd"
+              eval "$cmd"
+          done
```
lexming, line 62: Following the same idea of having atomic steps, I would limit this one to sanity checks. By the way, I think that we can also simplify this step by removing the array of commands and just execute them straight away. Is there any purpose for the array?

```suggestion
          env | sort
          eb --version
          eb --show-system-info
          eb --check-eb-deps
          eb --show-config
```

lexming, line 62: I would finally put the installation command in its own step:

```suggestion
          done
          
      - name: Test installation
        run : |
          eb bzip2-1.0.8.eb --trace --robot
```

lexming (dismissed):

.github/workflows/end2end.yml
```diff
@@ -0,0 +1,62 @@
...
+jobs:
+  build_publish:
+    name: End-to-end test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        container:
+          - centos-7.9
+          - centos-8.4
+          - fedora-35
```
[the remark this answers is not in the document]
boegel: For now, it works, and if it starts causing trouble it'll be very easy to kick it out from this workflow configuration.
So my vote goes to keeping it in, for now.

Testing in Fedora makes sense to me since it can act as a canary-in-the-coal-mine: if EasyBuild starts failing in Fedora, then that's something we should know ASAP, since that's going to trickly down into RHEL & co eventually...

.github/workflows/end2end.yml
```diff
@@ -0,0 +1,62 @@
...
+          - opensuse-15.4
```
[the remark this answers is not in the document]
boegel: Ah, ok, wasn't aware of this... Makes more sense to stick to openSUSE 15.3, indeed.

I'll get that container image built too, and then change this to 15.3

.github/workflows/end2end.yml
boegel: I don't really see the benefit of this (also seems like a nightmare to debug to me, especially with the setup for the modules tool)

.github/workflows/end2end.yml
boegel: I was using the array mostly so we can print the command being executed right before it's run.

That was useful when debugging this workflow, since the `eb` commands were failing in various interesting ways because the container images were rather minimal at first.

I prefer keeping it like this, since it's a bit easier to check the output produced by each of these commands if you also know which `eb` command was run...

boegel: Any specific reason, how would that help?

lexming: Having steps is more useful. Maybe in the current form of these tests there is little benefit, but if we ever want to expand these further, then we might want to have optional steps for instance (certainly in the case of some distros). It also allows to directly check the output of the failed step without all the preceding commands before it. It would also be easier to make a bot that triggers on some step but not others.

lexming: ok, having the commands printed will be handy indeed. I'm in the totally opposite mindset and I would put each of these commands in its own `run` :smile:

.github/workflows/end2end.yml
lexming: I agree that it is not super pretty due to lmod. But the environment in these containers should be rather minimal. There is also the option to print what has been injected into `GITHUB_ENV`

.github/workflows/end2end.yml
```diff
@@ -0,0 +1,62 @@
...
+              # for Rocky, CentOS 8, Fedora
+              echo ">>> sourcing /etc/profile.d/modules.sh"
+              . /etc/profile.d/modules.sh
+          else
+              echo ">>> sourcing /etc/profile.d/*lmod*.sh"
+              . /etc/profile.d/*lmod*.sh
+          fi
+
+          # tests are run with root privileges, so we need to tell EasyBuild that's OK...
+          export EASYBUILD_ALLOW_USE_AS_ROOT_AND_ACCEPT_CONSEQUENCES=1
```
boegel, line 49: We should add an `easybuild` user in the container image to avoid having to use `root`

boegel:
I did some extra work on this in a separate branch (see https://github.com/boegel/easybuild-framework/tree/end2test_test_TEST_TEST), but tests are acting up there, and it's not entirely clear to me why...

I'll get back to this later, this shouldn't block the release of EasyBuild v4.5.4...

.github/workflows/end2end.yml
boegel: More steps introduced in [5deae99](https://github.com/easybuilders/easybuild-framework/pull/3968/commits/5deae99370a27ade745df72e08f7e7ab71cb5b51)

.github/workflows/end2end.yml
boegel: More steps introduced in [5deae99](https://github.com/easybuilders/easybuild-framework/pull/3968/commits/5deae99370a27ade745df72e08f7e7ab71cb5b51)

.github/workflows/end2end.yml
boegel: Fixing this required `sudo` to be in the container images, see https://github.com/easybuilders/easybuild-containers/pull/23

.github/workflows/end2end.yml
```diff
@@ -0,0 +1,60 @@
...
+    name: End-to-end test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        container:
+          - centos-7.9
+          - centos-8.5
+          - fedora-36
+          - opensuse-15.4
+          - rockylinux-8.7
```
SebastianAchilles, line 14: Maybe we can also test on one Rocky 9 version?
```suggestion
          - rockylinux-8.7
          - rockylinux-9.1
```
I also opened an MR to add Rocky Linux 8.8 and 9.2 in case we want to update: https://github.com/easybuilders/easybuild-containers/pull/24

boegel: Yeah let's make it 8.8. and 9.2

```suggestion
          - rockylinux-8.8
          - rockylinux-9.2

SebastianAchilles (approved):
The end-to-end test looks good to me. :+1:

SebastianAchilles:
Going in, thanks @boegel!

SebastianAchilles:
Going in, thanks @boegel!

PR #3971, opened 2022-02-23
add gmpit toolchain definition (GCC + MPItrampoline)

ocaisa:


boegel (approved):
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/15018, works as advertised, and I don't see any reason to hold this back...

Thanks for looking into this @ocaisa!

PR #3973, opened 2022-02-28
use 'zypper search -i' to check whether specified OS dependency is installed on openSUSE + make sure that rpm is considered for checking OS dependencies on RHEL8

boegel:


easybuild/tools/systemtools.py
```diff
@@ -785,14 +787,17 @@ def check_os_dependency(dep):
     os_to_pkg_cmd_map = {
         'centos': RPM,
         'debian': DPKG,
+        'opensuse': ZYPPER,
         'redhat': RPM,
+        'rhel': RPM,
```
bartoldeman, line 792: Sneaked in?

bartoldeman (changes requested):
Should probably mention the rhel addition too?

easybuild/tools/systemtools.py
boegel: No, that was deliberately actually, sometimes `os_name` below is defined as `rhel` rather than `redhat`; on RHEL8 for example:

```
$ grep PRETTY /etc/os-release
PRETTY_NAME="Red Hat Enterprise Linux 8.4 (Ootpa)"
[13:42:54] {doduo} vsc40023@node3510:~ $ python3 -c 'from easybuild.tools.systemtools import get_os_name; print(get_os_name())'
RHEL
```

I've updated the PR title accordingly

bartoldeman (approved):
lgtm

PR #3974, opened 2022-03-01
add support for post-install patches

boegel:
Example use case here is a patch that we needed to apply for FLUENT, which can only be done after the installation was performed (because the files to patch don't exist in the unpacked sources, they're hidden deeper down in zipped files, etc.)


easybuild/framework/easyblock.py
```diff
@@ -455,12 +455,16 @@ def fetch_patches(self, patch_specs=None, extension=False, checksums=None):
...
+            post_install_patches = self.cfg['postinstallpatches']
+            patch_specs = self.cfg['patches'] + post_install_patches
+        else:
+            post_install_patches = []
 
         patches = []
         for index, patch_spec in enumerate(patch_specs):
 
             patch_info = create_patch_info(patch_spec)
+            patch_info['postinstall'] = patch_spec in post_install_patches
```
lexming, line 467: `patch_info['postinstall']` will only be true for post-install patches whenever `fetch_patches()` is called with `patch_specs=None`, but not when it will be called with some specific `patch_spec`. Is that ok?

easybuild/framework/easyblock.py
```diff
@@ -455,12 +455,16 @@ def fetch_patches(self, patch_specs=None, extension=False, checksums=None):
         All patches will be checked if a file exists (or can be located)
         """
         if patch_specs is None:
-            patch_specs = self.cfg['patches']
+            post_install_patches = self.cfg['postinstallpatches']
+            patch_specs = self.cfg['patches'] + post_install_patches
+        else:
+            post_install_patches = []
```
lexming, line 461: I have a minor suggestion for readability:
```
        post_install_patches = []

        if patch_specs is None:
            # if no patch_specs are specified, add all pre-install and post-install patches
            post_install_patches = self.cfg['postinstallpatches']
            patch_specs = self.cfg['patches'] + post_install_patches
```

lexming (commented):
Looks very good to me, I just have a question and minor suggestion

easybuild/framework/easyblock.py
boegel: `fetch_patches` is only passed fed `patch_specs` when fetching patch files for extensions, so it's fine if `postinstall` is always `False` then.

That's not really a limitation, since you can always use ` postinstallpatches` to patch any installed extensions.

easybuild/framework/easyblock.py
boegel: fixed in 9896bbb

easybuild/framework/easyblock.py
lexming: thanks for the clarification :+1:

lexming (approved):
LGTM

PR #3975, opened 2022-03-04
extend (experimental) support for generating container images with Apptainer

boegel:
With the release of [Apptainer v1.0.0](https://github.com/apptainer/apptainer/releases/tag/v1.0.0), we should also start to properly support it...

Although Apptainer also provides the `singularity` command, it doesn't work for EasyBuild because `singularity --version` prints `apptainer version 1.0.0`.


easybuild/tools/config.py
```diff
@@ -79,9 +79,10 @@
     CONT_IMAGE_FORMAT_SQUASHFS,
 ]
 
+CONT_TYPE_APPTAINER = 'apptainer'
 CONT_TYPE_DOCKER = 'docker'
 CONT_TYPE_SINGULARITY = 'singularity'
-CONT_TYPES = [CONT_TYPE_DOCKER, CONT_TYPE_SINGULARITY]
+CONT_TYPES = [CONT_TYPE_DOCKER, CONT_TYPE_SINGULARITY, CONT_TYPE_APPTAINER]
```
akesandgren, line 85: Why not alphabetical order here? You seem to insist on that in most other places where order is otherwise not important.

boegel: fixed in f9f2984

branfosj (changes requested):

branfosj (approved):

PR #3981, opened 2022-03-25
also print download instructions for extensions that use source_tmpl to specify name of source file

boegel:


bartoldeman (approved):
lgtm

PR #3991, opened 2022-04-01
deprecate use of CMake

boegelbot:
I'm sick of reporting installation problems that arise in easyconfig pull requests that can be basically brought down to the use of CMake as configuration tool, so it's time to take some measures...

Let's deprecate the use of CMake, and recommend users to use alternatives like SCons or Bazel, or to make a request to the developers of the software they want to install to do so.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2534,7 +2534,7 @@ class EasyBlock(object):
                 self.modules_tool.prepend_module_path(full_mod_path)
 
         deps = self.cfg.dependencies()
-        if any(d['name'].lower() == 'cmake' for d in deps):
+        if any(d['name'] and d['name'].lower() == 'cmake' for d in deps):
             msg = "Using CMake is deprecated, please make a request to the developers of %s " % self.name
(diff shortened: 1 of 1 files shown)


akesandgren:
While we're at it, let's deprecate Fortran and use Snowball instead.

migueldiascosta:
@boegelbot thank you for your first PR! your contributions to the easybuild community have been invaluable and we've been wondering when you'd take the natural step to contribute with actual code. Looking forward to see what else you have in store for us!

smoors (approved):
LGTM

klust:
Excellent idea, but I would skip SCons. The world would be so much better if there would be only one tool, so let's restrict ourselves to Bazel. After all, why would one want to die from CMake Cancer if the Bazel Bubonic Plague can kill you so much more efficiently? And if people want support: The answer is simple: "Ask Google".

dianaiusan:
This was fun, @boegelbot.

pescobar:
Fully agree with @klust . We should only support Bazel.

branfosj:
Closing the April Fools

PR #3994, opened 2022-04-14
Allowing searching for sources/patches in alternative location

### LINKED ISSUE #3990: option to point framework to another software's patches would be useful
Example is my PoC for OpenMPI-CUDA which has the same patches as OpenMPI, duplicating them would be error prone. So something like
```
find_patches_in = ['OpenMPI']
```
could be useful.

ocaisa:
Resolves #3990


ocaisa:
I'll update the MPItrampoline PR to use this PR (it currently symlinks to the OpenMPI sources)

casparvl:
I could review this (I'm maintainer of the week), but I think it might make more sense if @akesandgren does this, since he has a very similar case in #3990 that can serve as a test. @akesandgren do you have the opportunity to look at this PR?

On a side note: do I understand correctly that in EasyBuild, currently, patch files are only searched in the same directory that holds the EasyConfig? I've always assumed that the same paths were searched for patches that are also searched for EasyConfigs (as patch files also show up with `eb -S`), in which case it wouldn't matter in which subdirectories the patches reside... I guess that's currently not the case, but wouldn't it make more sense to just take that approach (i.e. just search the same searchpath used for EasyConfigs to also search for patches), rather than adding a specific, alternate search path?

I have to admit, I don't know `framework` well enough to know if this would make a reasonable approach. Also, you'd still have to decide what to do with sources: there, it does _not_ feel particularly intuitive if the whole sourcedir would be searched for a source tarball that happens to match the right name.

easybuild/framework/easyblock.py
```diff
@@ -698,11 +702,18 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
...
         """
         srcpaths = source_paths()
 
         update_progress_bar(PROGRESS_BAR_DOWNLOAD_ALL, label=filename)
 
+        if alt_location is None:
+            location = self.name
+        else:
+            location = alt_location
+
```
akesandgren, line 716: Does this meant that it ONLY looks in one of the locations?
And that you need to specify this explicitly for each patch/source?

I was more thinking of having a (possibly list of) global alt_location which it would also use to search for sources/patches.

ocaisa: Yes, that's what's there right now. In another scenario with multiple options you would also have to make a choice about where the sources should be downloaded to (and you increase the chance of a name clash for source files).

ocaisa: With this approach you will always find the OpenMPI sources in the same place. This will also work for components where you might have lots of different locations for source files.

akesandgren: By global I meant easyconfig file global

akesandgren: btw, I see no trace of this in the MPITrampoline PR (I may be half blind though)

easybuild/tools/filetools.py
```diff
@@ -1495,19 +1495,45 @@ def create_patch_info(patch_spec):
...
+            validate_patch_spec(patch_info['name'])
+        else:
+            if 'sourcepath' in patch_info.keys() or 'level' in patch_info.keys():
+                raise EasyBuildError("Wrong patch spec '%s', you can't use 'sourcepath' or 'level' with 'copy' (since "
+                                     "this implies you want to copy a file to the 'copy' location)",
+                                     str(patch_spec))
     else:
-        error_msg = "Wrong patch spec, should be string of 2-tuple with patch name + argument: %s"
+        error_msg = "Wrong patch spec, should be string, 2-tuple with patch name + argument, or a dict " \
+                    "(with possible keys ['name', 'copy', 'level', 'sourcepath', 'alt_location']): %s"
```
boegel, line 1523: This should definitely be derived from `valid_keys` rather than being hardcoded again here...

We're now already hardcoding the valid keys in two places (in `create_patch_info` + `fetch_patches`), which we should also try to avoid (but that's less easy I guess)

boegel (dismissed):

easybuild/tools/filetools.py
akesandgren: @ocaisa could you fix the first part of that so we can get this merged.

On the hardcoding part, that would involve defining macros for ALT_LOCATION etc, but I don't see the point really.

easybuild/framework/easyblock.py
ocaisa: I see what you mean by easyconfig global but for me that still has the same problem about where to store files: what happens in the case of bundles? For me, the least surprising thing is to be explicit. At least we can begin with that and then if someone wants to add additional possibilities they can.

akesandgren: yeah, I agree, this is better. I wasn't thinking about the download phase...

easybuild/tools/filetools.py
ocaisa: This should be fixed in [68ade05](https://github.com/easybuilders/easybuild-framework/pull/3994/commits/68ade05bc5a0c7006f74b2e694c288d2287d0c4a)

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @ocaisa!

Issue #3180, opened 2020-01-29
Poor startup time of 'eb'


casparvl:
I've talked to @boegel at the 5th EasyBuild user meeting about poor startup times of ```eb```. He mentioned it was something he was looking to improve as well and ask me to share what we encountered - hence this issue.

What we see on our systems:

```
[casparl@int1 ~]$ time eb --version
This is EasyBuild 4.1.1 (framework: 4.1.1, easyblocks: 4.1.1) on host int1.bullx.

real	0m4.421s
user	0m1.205s
sys	0m1.666s
```

```
casparl@login3:~$ time eb --version
This is EasyBuild 4.1.1 (framework: 4.1.1, easyblocks: 4.1.1) on host login3.lisa.surfsara.nl.

real	0m1.539s
user	0m1.032s
sys	0m0.184s
```

It seems to vary a lot, but overall performance isn't great.

My colleague @paulmelis has investigated it and made some traces. I don't have the time to organize it right now, but will simply dump those in this ticket so that you have a starting point.

casparvl:
Let's do some system call tracing to see where the time it spent:

```
paulm@int3 13:59 /dev/shm$ time strace -c python -m easybuild.main
== temporary log file in case of crash /tmp/paulm/eb-V4WOTx/easybuild-FhwyW3.log
...
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 45.35    0.183438           8     22392     20846 open
 26.79    0.108382           5     20452      8605 stat
 14.53    0.058762           4     15535         6 lstat
  2.99    0.012085           4      2930           read
  2.40    0.009702           3      3359         1 fstat
  1.39    0.005628         375        15           clone
  1.34    0.005408           3      1663           close
  1.18    0.004769           5       963           munmap
  1.06    0.004291           4      1182           mmap
  0.87    0.003525           3      1068         1 readlink
  0.83    0.003359           2      1569       144 lseek
  0.20    0.000820           4       189       151 openat
  0.20    0.000789          10        76           getdents
  0.14    0.000569           6        89           mprotect
  0.13    0.000512           3       167           fcntl
  0.13    0.000510           4       119           poll
  0.08    0.000320           2       156           brk
  0.07    0.000284          15        19           wait4
  0.06    0.000244           4        69           rt_sigaction
  0.05    0.000208           1       163       101 access
  0.05    0.000199           3        68           write
  0.04    0.000174           5        33           pipe
[...]

PR #3995, opened 2022-04-19
Change 'eb' command to import easybuild to check if it's working.

bartoldeman:
This is a much cheaper test than `import easybuild.main` and still
solves the same issue as #3610.

This reduces the wall time for `eb --version` from 1.35s to 0.73s
in our configuration.

### DIFF
--- a/eb
+++ b/eb
@@ -80,7 +80,7 @@ for python_cmd in ${EB_PYTHON} ${EB_INSTALLPYTHON} 'python' 'python3' 'python2';
 
         if [ ! -z $PYTHON ]; then
             # check whether easybuild.main is available for selected python command
-            $PYTHON -c "import $EASYBUILD_MAIN" 2> /dev/null
+            $PYTHON -c "import easybuild" 2> /dev/null
             if [ $? -eq 0 ]; then
                 verbose "'$python_cmd' is able to import '$EASYBUILD_MAIN', so retaining it"
             else
--- a/eb
+++ b/eb
@@ -82,10 +82,10 @@ for python_cmd in ${EB_PYTHON} ${EB_INSTALLPYTHON} 'python' 'python3' 'python2';
             # check whether easybuild.main is available for selected python command
             $PYTHON -c "import easybuild" 2> /dev/null
             if [ $? -eq 0 ]; then
-                verbose "'$python_cmd' is able to import '$EASYBUILD_MAIN', so retaining it"
+                verbose "'$python_cmd' is able to import 'easybuild', so retaining it"
             else
                 # if easybuild.main is not available, don't use this python command, keep searching...
-                verbose "'$python_cmd' is NOT able to import '$EASYBUILD_MAIN', so NOT retaining it"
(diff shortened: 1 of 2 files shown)


bartoldeman:
See also #3180

eb
```diff
@@ -80,7 +80,7 @@ for python_cmd in ${EB_PYTHON} ${EB_INSTALLPYTHON} 'python' 'python3' 'python2';
 
         if [ ! -z $PYTHON ]; then
             # check whether easybuild.main is available for selected python command
-            $PYTHON -c "import $EASYBUILD_MAIN" 2> /dev/null
+            $PYTHON -c "import easybuild" 2> /dev/null
             if [ $? -eq 0 ]; then
                 verbose "'$python_cmd' is able to import '$EASYBUILD_MAIN', so retaining it"
```
akesandgren, line 85: Would need to change this output too to be consistent.

akesandgren (dismissed):

eb
bartoldeman: done

eb
```diff
@@ -80,12 +80,12 @@ for python_cmd in ${EB_PYTHON} ${EB_INSTALLPYTHON} 'python' 'python3' 'python2';
 
         if [ ! -z $PYTHON ]; then
             # check whether easybuild.main is available for selected python command
```
casparvl, line 82: Probably should replace `easybuild.main` in this comment by `easybuild` in order not to create confusion

eb
```diff
@@ -80,12 +80,12 @@ for python_cmd in ${EB_PYTHON} ${EB_INSTALLPYTHON} 'python' 'python3' 'python2';
...
-            $PYTHON -c "import $EASYBUILD_MAIN" 2> /dev/null
+            $PYTHON -c "import easybuild" 2> /dev/null
             if [ $? -eq 0 ]; then
-                verbose "'$python_cmd' is able to import '$EASYBUILD_MAIN', so retaining it"
+                verbose "'$python_cmd' is able to import 'easybuild', so retaining it"
             else
                 # if easybuild.main is not available, don't use this python command, keep searching...
```
casparvl, line 87: Likewise, should probably replace `easybuild.main` with `easybuild in the comment

casparvl (approved):
Looks good to me. I've installed from the feature branch, and tested timings. For me it was about 25% faster to run `eb --version`. I've also tested a basic installation (which worked fine) just as a sanity check although this PR shouldn't change anything that could affect installations - only the check for which python to use was touched.

To me, the change looks sane and should still return the correct Python, namely the one used to install EasyBuild.

boegel:
This is actually not equivalent at all, since if somehow only the `eb` command and the `easybuild-easyblocks` package are available, `import easybuild` will also work, but `easybuild.main` (or anything else provided by framework) may not be, so I'm not sure this change is a good idea...

boegel:
An alternative could be to test the import of `easybuild.framework`, which is very light weight as well (doesn't import other things from `easybuild` namespace), and that's definitely a better check to verify that EasyBuild framework is available...

Follow-up PR at #3998

PR #3998, opened 2022-04-27
change 'eb' command to import easybuild.framework to check if EasyBuild framework is available

boegel:
follow-up for #3995, since just testing import of `easybuild` is way too generic


bartoldeman:
```
$ time python -c "import easybuild.framework"

real    0m0.030s
user    0m0.018s
sys     0m0.012s
$ time python -c "import easybuild"

real    0m0.029s
user    0m0.022s
sys     0m0.007s
$ time python -c "import easybuild.main"

real    0m30.081s
user    0m0.654s
sys     0m0.279s
$ time python -c "import easybuild.main"

real    0m0.605s
user    0m0.517s
sys     0m0.143s
```
it's cheap enough, so let's go for this!

Issue #3918, opened 2021-12-17
Extend framework to enable modules to ship RPATH wrappers


stderr-enst:
In this issue I want to track a couple of changes that are required on the framework side, to allow modules to ship RPATH wrapper scripts when they are loaded.

There are a couple of open points to address:
1. modify [`module_generator.py`](https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/module_generator.py):
   - [ ] Add conditional statement for set environment variables. Then we can ignore rpath wrappers while loading a module if something like `DISABLE_RPATH_WRAPPERS` is set
   - [ ] Ensure that wrappers are not put into `$PATH` if module is used within EasyBuild
1. modify [`prepare_rpath_wrappers()` in `toolchain.py`](https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/toolchain/toolchain.py#L947):
   - [ ] Add function parameter to `prepare_rpath_wrappers` to install wrappers in a caller-given directory
   - [ ] Return directory path where wrappers were installed
   - [ ] Add option to put all wrapper scripts in a single directory (with no subdirs like `gcc_wrapper/gcc`)
   - [ ] Don't write [`rpath_wrapper_log`](https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/toolchain/toolchain.py#L1014) to `tempfile.gettempdir()` if `debug` is set. This location does not necessarily exist if a module is loaded with the wrapper scripts in place
1. modify [`rpath_wrapper_template.sh.in`](https://github.com/easybuilders/easybuild-framework/blob/e7e04415797c1128f12a2d6b7dd18dbbb2b4bfce/easybuild/scripts/rpath_wrapper_template.sh.in):
   - [ ] Add option to read env variables `RPATH_INCLUDE_DIRS` and `RPATH_FILTER_DIRS` variables, to include/exclude directories in wrapper script (as comma-separated list) at runtime
1. Add `create_rpath_wrappers`:
   - [ ] Refactor `create_rpath_wrappers` from [easybuild-easyblocks/pull/2638](https://github.com/easybuilders/easybuild-easyblocks/pull/2638) into a more generic location within framework. This is probably used for other compilers in the future as well

Related PRs:
- [easybuild-easyblocks/pull/2638](https://github.com/easybuilders/easybuild-easyblocks/pull/2638)

ocaisa:
[...]

PR #4003, opened 2022-05-02
Update prepare_rpath_wrappers to enable wrapper shipping with a module (WIP)

stderr-enst:
This WIP PR is part of a suggested feature to ship rpath wrappers e.g. with GCCcore.
More information in related issues and PRs:
- [easybuilders/easybuild-framework/issues/3918](https://github.com/easybuilders/easybuild-framework/issues/3918)
- [easybuild-easyblocks/pull/2638](https://github.com/easybuilders/easybuild-easyblocks/pull/2638)

This PR contains a first working version, with a couple of dirty hacks, so it is definitely not ready to be merged.

`prepare_rpath_wrappers` is currently preparing the wrappers in temporary directories. Wrappers are then used in the module's build step so far!
Updating `prepare_rpath_wrappers` to also install wrappers in the module install is maybe stretching the purpose of a single function.
Would you suggest that I refactor the code a bit to produce 2 smaller functions? One prepares wrappers for build, the other does installation. There is probably a bit of shared code though.

I'll also point out a specific section of the code in the comments below.

Tagging @boegel

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -944,7 +944,7 @@ class Toolchain(object):
         calls_rpath_args = b'rpath_args.py $CMD' in read_file(path, mode='rb')
         return in_rpath_wrappers_dir and calls_rpath_args
 
-    def prepare_rpath_wrappers(self, rpath_filter_dirs=None, rpath_include_dirs=None):
+    def prepare_rpath_wrappers(self, rpath_filter_dirs=None, rpath_include_dirs=None, new_wrapper_dir=None, single_subdir=False, disable_wrapper_log=False):
         """
         Put RPATH wrapper script in place for compiler and linker commands
 
@@ -967,7 +967,12 @@ class Toolchain(object):
                 rpath_filter_dirs.append(lib_stubs_pattern)
 
         # directory where all wrappers will be placed
-        wrappers_dir = os.path.join(tempfile.mkdtemp(), RPATH_WRAPPERS_SUBDIR)
+        if new_wrapper_dir is None:
+            wrappers_dir = os.path.join(tempfile.mkdtemp(), RPATH_WRAPPERS_SUBDIR)
+        else:
+            wrappers_dir = new_wrapper_dir
+            if not os.path.exists(wrappers_dir):
+                os.mkdir(wrappers_dir)
 
         # must also wrap compilers commands, required e.g. for Clang ('gcc' on OS X)?
         c_comps, fortran_comps = self.compilers()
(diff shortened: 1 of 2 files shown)


easybuild/tools/toolchain/toolchain.py
```diff
@@ -989,6 +994,15 @@ def prepare_rpath_wrappers(self, rpath_filter_dirs=None, rpath_include_dirs=None
...
             orig_cmd = which(cmd)
+            if cmdDir is not None and os.path.exists(cmdDir):
+                orig_cmd = os.path.join(cmdDir,cmd)
+
+            # TODO: dirty hack to let module-shipped rpath wrappers point to EESSI software layer's ld
+            if cmdDir is not None and "ld" in cmd:
+                orig_cmd = which(cmd, retain_all=True)
+                orig_cmd = [a for a in orig_cmd if "/tmp" not in a][0]
+
+            self.log.debug("orig_cmd: %s", orig_cmd)
```
stderr-enst, line 1005: This section is a bit critical.
The original `prepare_rpath_wrappers` is using `which` to lookup `ld` and compiler commands to wrap with the newly created rpath wrapper.
But we want to wrap the compiler command of the e.g. GCCcore module that currently is build! So we cannot look it up in `PATH` and instead have to pass a new `cmdDir` to the function and decide to set it manually.
This might be a first good reason to split this function into two.

Second, in EESSI we ship `ld` in the software layer on CVMFS. We don't want to pick up any `ld` commands that exist in the current `/tmp` locations. They are still around during module building, so we have to put a dirty hack here.
I don't really like putting an EESSI special case into EasyBuild code.
Do we have a general approach to this kind of problem?
Maybe this calls for another manual path, from where to pick up `ld`?

casparvl: I'm confused which usage of `prepare_rpath_wrappers` you're trying to target here with the exception for `ld`: the usage by EasyBuild when installing with `--rpath`? Or is it to prevent the `ld` wrappers from being installed for EESSI? In the latter case, I'd say we should indeed make the `ld` wrapper installation part of `binutils` (as was suggested in https://github.com/easybuilders/easybuild-easyblocks/pull/2638). That means `--filter-deps binutils` is enough to (also) make sure the RPATH wrappers aren't installed (and I guess we filter `binutils` already in EESSI? Can't remember exactly...).

If it's more about EasyBuild when installing with `--rpath` (i.e. you're afraid that this will pick up the `ld` wrappers from the tempdir that EasyBuild creates), I'd have to think how to elegantly solve this. I agree with the statement that we don't want special case things for EESSI in the EasyBuild source code, it should be something generic. 

I'm generally a bit confused what we do currently in EESSI: do we set `RPATH` with our own `ld` wrapper? Or do we rely on EasyBuild to set RPATH? I thought it was the latter, but then you'd still want it to generate wrappers when installing with `--rpath`, right?

ocaisa:
@stderr-enst I'd like to take over what you've done here for a specific use case, I would like to create an updated `buildenv` easyblock that can ship the rpath wrappers (rather than inject them into `GCC`). It will also setup the typical build environment you'd need for `ConfigureMake`/`CMakeMake` (so Autotools, CMake, pkgconfig) and all the associated envvars. That should be good enough to allow people to build on top of EasyBuild without using EasyBuild itself (at least for the easier cases).

PR #4014, opened 2022-05-30
correctly identify Apple Silicon M1 as Arm 64-bit by also considering arm64 next to aarch64

boegel:


ocaisa:
Code looks ok, I have no way of checking this though so I'm going on faith alone

boegel:
Just to clarify: without this change, I was getting `AArch32` as a result from `get_cpu_architecture` on Apple M1:

```
$ python3 -c 'from easybuild.tools.systemtools import get_cpu_architecture; print(get_cpu_architecture())'
AArch32
```

with these changes, I'm getting `AArch64`, as it should be:

```
$ python3 -c 'from easybuild.tools.systemtools import get_cpu_architecture; print(get_cpu_architecture())'
AArch64
```

PR #4015, opened 2022-05-30
fix 'eb --show-system-info' on Apple M1 system

### LINKED ISSUE #4013: --show-system-info fails on Apple Silicon M1
Different system queries via `sysctl` are broken on Apple Silicon.

Known issues:

* "`sysctl -n hw.cpufrequency_max`" returns empty result, used in `get_cpu_speed`
  * see also https://github.com/sebhildebrandt/systeminformation/issues/564)
* `machdep.cpu.vendor` (used in `get_cpu_vendor`) and `machdep.cpu.extfeatures` (used in `get_cpu_features`) are not supported, triggers `sysctl: unknown oid` error

boegel:
This fixes various small issues that pop up on Apple M1:
* add fallback to "`sysctl -n machdep.cpu.brand_string`" in `get_cpu_vendor` to determine CPU type, since "`sysctl -n machdep.cpu.vendor`" is not supported on Apple M1 systems;
* add `Apple` as known CPU vendor;
* avoid that `get_cpu_speed` crashes when it fails to determine CPU speed via "`sysctl -n hw.cpufrequency_max`" (which returns empty string on Apple M1), and just return `None`;
* avoid crash in `get_cpu_features` because "`sysctl -n machdep.cpu.extfeatures`" is not supported on Apple M1 (and leave list of CPU features empty, for now);
* also includes fixes for `systemtools` tests so they pass on Apple M1;

Also requires #4014 for all `systemtools` tests to pass on Apple M1

fixes #4013

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -94,6 +94,7 @@ ARCH_KEY_PREFIX = 'arch='
 # Vendor constants
 AMD = 'AMD'
 APM = 'Applied Micro'
+APPLE = 'Apple'
 ARM = 'ARM'
 BROADCOM = 'Broadcom'
 CAVIUM = 'Cavium'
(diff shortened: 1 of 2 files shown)


ocaisa (approved):
This also looks ok, can you add an example of the output since I can't test this?

boegel:
Example output on Apple M1:

```
$ eb --show-system-info
System information (vpne020.ugent.be):

* OS:
  -> name: UNKNOWN
  -> type: Darwin
  -> version: UNKNOWN
  -> platform name: arm64-apple-darwin

* CPU:
  -> vendor: Apple
  -> architecture: AArch64
  -> family: ARM
  -> arch name: m1
  -> model: Apple M1
  -> speed: None
  -> cores: 8
  -> features:

* software:
  -> glibc version: UNKNOWN
  -> Python binary: /Library/Developer/CommandLineTools/usr/bin/python3
  -> Python version: 3.8.9
```

There's room for improvement, but this is *way* better than only producing a nasty traceback on macOS @ Apple M1...

ocaisa:
Agreed

PR #4018, opened 2022-06-05
make sure that ARCH constant has 'aarch64' (rather than 'arm64') as value on macOS ARM

boegel:
Fix for problems that occur when parsing an easyconfig that relies on the `ARCH` constant:

```
$ eb Pandoc-2.13.eb -D
== Temporary log file in case of crash /tmp/eb-vbgmw5al/easybuild-shentmri.log
ERROR: Failed to process easyconfig /home/example/easybuild-easyconfigs/easybuild/easyconfigs/p/Pandoc/Pandoc-2.13.eb: Parsing easyconfig file failed: 'arm64' (line 13)
```

### DIFF
--- a/test/framework/easyconfig.py
+++ b/test/framework/easyconfig.py
@@ -4597,6 +4597,12 @@ class EasyConfigTest(EnhancedTestCase):
         test_ec = EasyConfig(test_ec)
         self.assertEqual(test_ec.count_files(), 11)
 
+    def test_ARCH(self):
+        """Test ARCH easyconfig constant."""
+        arch = easyconfig.constants.EASYCONFIG_CONSTANTS['ARCH'][0]
+        known_arch_values = ('aarch64', 'ppc64le', 'riscv64', 'x86_64')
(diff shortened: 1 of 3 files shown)


test/framework/easyconfig.py
```diff
@@ -4597,6 +4597,12 @@ def test_count_files(self):
         test_ec = EasyConfig(test_ec)
         self.assertEqual(test_ec.count_files(), 11)
 
+    def test_ARCH(self):
+        """Test ARCH easyconfig constant."""
+        arch = easyconfig.constants.EASYCONFIG_CONSTANTS['ARCH'][0]
+        known_arch_values = ('aarch64', 'ppc64le', 'riscv64', 'x86_64')
```
migueldiascosta, line 4603: move this to a global constant?

migueldiascosta, line 4603: move this to a global constant?

boegel: fixed in 0ee0ddd55

boegel: fixed in 0ee0ddd55

migueldiascosta:
Going in, thanks @boegel!

PR #4021, opened 2022-06-07
add support for easystack file that contains easyconfig filenames + implement parsing of configuration options

boegel:


casparvl:
Let me see if I get this right: essentially, you want to support an additional yaml structure where we don't specify the packages to be installed using e.g.

```
# easystack.yaml
software:
  R-bundle-Bioconductor:
    toolchains:
      foss-2020a:
        versions:
          '3.11':
            versionsuffix: '-R-4.0.0'
  GROMACS:
    toolchains:
      foss-2020a:
        versions:
          '2020.1':
            versionsuffix: '-Python-3.8.2'
          '2020.4':
            versionsuffix: '-Python-3.8.2'
```

You would specify

```
# easystack.yaml
easyconfigs:
  - R-bundle-Bioconductor-foss-2020a-3.11-R-4.0.0
  - GROMACS-foss-2020a-2020.1-Python-3.8.2
  - GROMACS-foss-2020a-2020.4-Python-3.8.2
```

Correct?

Some thoughts:

It's a lot more compact, but one of the main reasons we opted for the first (hierarchical) structure, is that you probably want to add certain options at the software level (e.g. for all `GROMACS` installations, use some `--include-easyblocks-from-pr` or something). In the second example, you'd have to specify that twice, remove two items if the `easyblock` gets merged, etc. I'm fine in adding this as a second format, I mean, we're still able to use the first format if whenever that is more suitable.  However if we add a 3rd, 4th and 5th later on, things will get quite confusing. So I'm curious: what was your main reason for wanting _this_ format?

easybuild/framework/easystack.py
```diff
@@ -80,6 +85,11 @@ def compose_ec_filenames(self):
             })
             ec_filename = '%s-%s.eb' % (sw.name, full_ec_version)
             ec_filenames.append(ec_filename)
+
+        # entries specified via 'easyconfigs' top-level key
+        for ec in self.easyconfigs:
+            ec_filenames.append(ec + '.eb')
```
casparvl, line 91: Why did you decide to add the `.eb` extension here? Why not have people write in the YAML file:

```
robot: true
easyconfigs:
  - binutils-2.25-GCCcore-4.9.3.eb
  - foss-2018a.eb
  - R-bundle-Bioconductor-3.11-foss-2020a-R-4.0.0.eb
  - GROMACS-2020.4-foss-2020a-Python-3.8.2.eb
```
That way, it is immediately clear and recognizeable that these are (full) EasyConfig filenames

casparvl (commented):
Looks reasonable overall, but lets discuss the individual items I touched upon :)

boegel:
@casparvl 

> So I'm curious: what was your main reason for wanting this format?

Mainly because the current format is overly verbose for the common case where there's nothing special going on (no specific configuration options needed), etc.
I can probably rework this PR so it works with a "mixed" easystack, that uses both `easyconfigs` and `software` as top-level keys.

> Another question that came to mind: how will we specify options here?

Good question, I didn't think that through yet. I think something like this could work:

```yaml
easyconfigs:
    example-1.2.3:
        options: --from-pr 123
    example-4.5.6
```

I'll see if I can add dummy support for `options`, so we pick up on it, but don't do anything with it yet...

easybuild/framework/easystack.py
```diff
@@ -108,21 +118,53 @@ class EasyStackParser(object):
...
         except yaml.YAMLError as err:
             raise EasyBuildError("Failed to parse %s: %s" % (filepath, err))
 
-        easystack = EasyStack()
+        easystack_data = None
+        top_keys = ('easyconfigs', 'software')
+        for key in top_keys:
+            if key in easystack_raw:
+                easystack_data = easystack_raw[key]
+                break
```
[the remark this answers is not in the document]
boegel: We can either forbid it to combine both, and hard fail if an easystack file uses both, or support the use of both in a single easystack file (I don't think that's horribly complicated, as long as you let the 2nd key overrule entries from the first on, perhaps along with a warning)

easybuild/framework/easystack.py
```diff
@@ -108,21 +118,53 @@ class EasyStackParser(object):
...
-            wrong_structure_file = "Not a valid EasyStack YAML file: no 'software' key found"
-            raise EasyBuildError(wrong_structure_file)
+        if easystack_data is None:
+            msg = "Not a valid EasyStack YAML file: no 'easyconfigs' or 'software' top-level key found"
+            raise EasyBuildError(msg)
+        else:
+            parse_method_name = 'parse_by_' + key
+            parse_method = getattr(EasyStackParser, 'parse_by_%s' % key, None)
+            if parse_method is None:
+                raise EasyBuildError("Easystack parse method '%s' not found!", parse_method_name)
```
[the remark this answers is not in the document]
boegel: This is about making sure that there's a `parse_by_*` method that matches the key.

Maybe in the future we add support for another top-level key (for example to sort by toolchains first rather than software names), but then we don't add a corresponding `parse_by_*` method.
This `raise` just makes sure we don't get a horrible crash then, but a sensible error message.

easybuild/framework/easystack.py
```diff
@@ -108,21 +118,53 @@ class EasyStackParser(object):
...
+
+        # assign general easystack attributes
+        easybuild_version = easystack_raw.get('easybuild_version', None)
+        robot = easystack_raw.get('robot', False)
+
+        return parse_method(filepath, easystack_data, easybuild_version=easybuild_version, robot=robot)
```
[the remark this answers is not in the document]
boegel: Indeed, I've left that here for now, but we'll definitely get rid of `robot=` going forward when we add proper support for specifying options for specific entries. And we can since support for easystack files is still marked as experimental...

easybuild/framework/easystack.py
boegel: Because it's redundant. The top-level key should make it very clear what these strings are.

I'm happy to allow the use of `.eb` though, and we can guard the `append` with an `not ec.endswith('.eb')`

jfgrimm: personally I would find it more intuitive to specify the `.eb` extension, even if it's redundant. No reason not to allow both though, guarding the `append`

akesandgren: and it makes for easier C&P

casparvl: I took the liberty of pushing that change :)

easybuild/framework/easystack.py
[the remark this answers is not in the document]
casparvl: Ok, for now, I've implemented a hard-fail if multiple keys are specified. We can always improve support from there, but I don't think it's super essential to be able to combine both in a single file. 99% of users will probably opt for one format or the other anyway.

casparvl:
I've implemented the dummy support, will push in a second. The support that I've added is that the parser can now understand this format:
```
robot: true
easyconfigs:
  - pkgconf-1.8.0-GCCcore-11.3.0.eb:
      options: {
        'debug': True,
      }
  - ncurses-6.3-GCCcore-11.3.0.eb
```
i.e. `easyconfigs` is a list, but the items maybe be simple strings (and EasyConfig name), but may also be dictionaries. In case of a dictionary, it has this (raw) structure:
```
{'pkgconf-1.8.0-GCCcore-11.3.0.eb': None, 'options': {'debug': True}}
```
We can later quite easily have the parser store that in a data structure that satisfies the format proposed [here](https://github.com/easybuilders/easybuild-framework/issues/3513#issuecomment-986990195) for `opts_per_ec`: i.e. a dictionary, in which the keys are the easyconfig names, and the values are dictionaries with all the options.

easybuild/framework/easystack.py
```diff
@@ -108,21 +118,83 @@ class EasyStackParser(object):
...
+                    if not easyconf_name.endswith('.eb'):
+                        easyconf_name_with_eb = easyconf_name + '.eb'
+                    easystack.easyconfigs.append(easyconf_name_with_eb)
+                    # Add options to the ec_opts dict
+                    if 'options' in easyconfig[easyconf_name].keys():
+                        easystack.ec_opts[easyconf_name_with_eb] = easyconfig[easyconf_name]['options']
+                else:
+                    dict_keys = ', '.join(easyconfig.keys())
+                    msg = "Failed to parse easystack file: expected a dictionary with one key (the EasyConfig name). "
+                    msg += "Instead found keys: %s" % dict_keys
```
ocaisa, line 186: Consider a pointer to the docs again?

ocaisa (changes requested):

easybuild/framework/easystack.py
```diff
@@ -108,21 +118,83 @@ class EasyStackParser(object):
...
+                    raise EasyBuildError(msg)
+
+        return easystack
```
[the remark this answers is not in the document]
ocaisa: We could consider resolving these with the robot and returning the full path to the ecs (and outputting these in debug mode). we could consider an option top prefer easyconfig files that are stored in the same path as the easystack file (which may not be in the robot path)...but I think that is a question of taste

test/framework/easystacks/test_easystack_easyconfigs_opts.yaml
```diff
@@ -0,0 +1,11 @@
+easyconfigs:
+    - binutils-2.25-GCCcore-4.9.3:
+        options: {
+          'debug': True,
```
ocaisa, line 4: A test with more than one option set might be useful...and ultimately also one with an invalid setting!

easybuild/framework/easystack.py
```diff
@@ -108,21 +118,83 @@ class EasyStackParser(object):
...
+        for key in top_keys:
+            if key in easystack_raw:
+                keys_found.append(key)
+        # For now, we don't support mixing multiple top_keys, so check that only one was defined
+        if len(keys_found) > 1:
+            keys_string = ', '.join(keys_found)
+            msg = "Specifying multiple top level keys (%s) in one EasyStack file is not supported." % keys_string
+            raise EasyBuildError(msg)
+        elif len(keys_found) == 0:
+            msg = "Not a valid EasyStack YAML file: no 'easyconfigs' or 'software' top-level key found"
```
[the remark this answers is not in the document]
boegel: Yes, can mention https://docs.easybuild.io/en/latest/Easystack-files.html (but then we should also update it to reflect that `easyconfigs` is also supported as a top-level key now...

easybuild/framework/easystack.py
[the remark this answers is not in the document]
boegel: The job of the easystack file is to just provide a list of easyconfigs, nothing more. It basically replaces listing easyconfig filenames in the `eb` command.

Checking whether the easyconfigs can actually be found is done later through `main` (via `det_easyconfig_paths`).

Likewise for the options: passing those to the option parser, which will check if the options are valid, is done later (that's a part of @casparvl's PR #4057)

test/framework/easystacks/test_easystack_easyconfigs_opts.yaml
boegel: I'll extend this a bit so it uses multiple options, but handling of invalid options should be checked in #4057

boegel: done in 0550fc0

easybuild/framework/easystack.py
boegel: fixed in [c674005](https://github.com/easybuilders/easybuild-framework/pull/4021/commits/c6740056c21ed005f3a9c0d5ce06b400293546b5)

boegel:
Documentation will be updated to mention support for top-level `easyconfigs` key when #4057 is also merged (which adds support for actually applying the configuration options specific to a particular easystack entry)

easybuild/framework/easystack.py
```diff
@@ -243,12 +320,16 @@ def parse_easystack(filepath):
...
 
     _log.debug("EasyStack parsed. Proceeding to install these Easyconfigs: %s" % ', '.join(sorted(easyconfig_names)))
-    if len(general_options) != 0:
-        _log.debug("General options for installation are: \n%s" % str(general_options))
-    else:
-        _log.debug("No general options were specified in easystack")
-
-    return easyconfig_names, general_options
+    _log.debug("Using EasyConfig specific options based on the following dict:")
+    _log.debug(easystack.ec_opts)
```
boegel, line 332: We could use `pprint.pformat` here to make it more readable (let's do that in #4057)

easybuild/main.py
```diff
@@ -266,7 +265,7 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         # TODO add general_options (i.e. robot) to build options
         orig_paths, general_options = parse_easystack(options.easystack)
```
boegel, line 266: `general_options` is not correct here, but this is being fixed in #4057 (renamed to `opts_per_ec`)

casparvl (approved):
Went through this together with @boegel and looks good

Any polishing we still need to do can be done in https://github.com/easybuilders/easybuild-framework/pull/4057

PR #4025, opened 2022-06-11
add toolchain option to customize linker command that compiler should use (WIP)

boegel:
Usage to use `mold` as a linker:

```python
toolchainopts = {'linker': 'mold'}
```

For non-GCC compilers (Clang, Intel) and GCC 12.1, this is equivalent with adding `-fuse-ld=mold` to `$CFLAGS` and friends.
For GCC < 12.1, this is equivalent to adding `-B$EBROOTMOLD/libexec/mold` to `$CFLAGS` and co.
See also `classic way to use mold` at https://github.com/rui314/mold#how-to-use.

Of course this also requires that `mold` is actually available, so it should be added as a build dependency (or be available system-wide).

Marked as WIP because:
* these change broke one of the tests for some reason (to be figured out);
* need to make sure that tests cover old/new GCC, Intel, Clang;
* using this with Intel compilers doesn't actually seem to work; tested with GSL 2.7:
  ```
  $ ls software/GSL/*/bin/gsl-histogram | xargs readelf -p .comment | egrep '^File|mold'
  File: software/GSL/2.7-ClangGCC-13.0.1-11.2.0/bin/gsl-histogram
    [    53]  mold 1.2.1 (compatible with GNU ld)
  File: software/GSL/2.7-GCC-11.3.0/bin/gsl-histogram
    [    2c]  mold 1.2.1 (compatible with GNU ld)
  File: software/GSL/2.7-intel-compilers-2022.1.0/bin/gsl-histogram
  ```

### DIFF
--- a/easybuild/toolchains/compiler/gcc.py
+++ b/easybuild/toolchains/compiler/gcc.py
@@ -28,7 +28,7 @@ Support for GCC (GNU Compiler Collection) as toolchain compiler.
 :author: Stijn De Weirdt (Ghent University)
 :author: Kenneth Hoste (Ghent University)
 """
-
+import os
 import re
 from distutils.version import LooseVersion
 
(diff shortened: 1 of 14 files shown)


rui314:
@boegel A few bugs have been filed to make mold work with ICC. We fixed them, and the bug reporters seemed happy, so I believe it works with ICC. But I haven't tried it myself.

boegel:
@rui314 Is it correct that you need to use the `-B` option to make ICC use `mold` as linker (as opposed to `-fuse-ld=mold`)?
If so, I'll open a pull request to update mold's `README` file to mention that...

rui314:
@boegel I don't actually know. If icc doesn't take `-fuse-ld=mold`, then the answer is probably yes though.

boegel:
> @boegel I don't actually know. If icc doesn't take `-fuse-ld=mold`, then the answer is probably yes though.

Well, it accepts the option, but then silently doesn't use `mold` at all (based on output of `readelf -p .comment` on the binary)

PR #4030, opened 2022-06-17
add support for setting environment variables via 'pushenv' with modextravars

boegel:
Use case here is internal, but it's basically an environment variable we want to define via `pushenv` through a custom hook implementation.

General support for `pushenv` is useful, since it avoids the need to resort to `modluafooter`...

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -407,6 +407,29 @@ class ModuleGenerator(object):
 
         return res
 
+    def unpack_setenv_value(self, env_var_name, env_var_val):
+        """
+        Unpack value that specifies how to define an environment variable with specified name.
+        """
+        use_pushenv = False
+
+        # value may be specified as a string, or as a dict for special cases
+        if isinstance(env_var_val, string_type):
+            value = env_var_val
+
+        elif isinstance(env_var_val, dict):
+            use_pushenv = env_var_val.get('pushenv', False)
+            try:
+                value = env_var_val['value']
+            except KeyError:
+                raise EasyBuildError("Required key 'value' is missing in dict that specifies how to set $%s: %s",
+                                     env_var_name, env_var_val)
+        else:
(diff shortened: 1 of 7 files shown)


easybuild/tools/module_generator.py
```diff
@@ -407,6 +407,29 @@ def det_installdir(self, modfile):
...
+
+        elif isinstance(env_var_val, dict):
+            use_pushenv = env_var_val.get('pushenv', False)
+            try:
+                value = env_var_val['value']
+            except KeyError:
+                raise EasyBuildError("Required key 'value' is missing in dict that specifies how to set $%s: %s",
+                                     env_var_name, env_var_val)
+        else:
+            raise EasyBuildError("Incorrect value type for setting $%s environment variable (%s): %s",
```
akesandgren, line 428: Should this include list of valid types?

test/framework/easyblock.py
```diff
@@ -1158,8 +1159,20 @@ def test_make_module_step(self):
...
+            'FOO': 'bar',
+        }
+
+        # also test setting environment variables via pushenv for Lmod or recent Environment Modules
+        is_modtool_ver51 = LooseVersion(self.modtool.version) >= LooseVersion('5.1')
+        if isinstance(self.modtool, Lmod) or (isinstance(self.modtool, EnvironmentModules) and is_modtool_ver51):
+            modextravars['TEST_PUSHENV'] = {'value': '123', 'pushenv': True}
+
+        modextrapaths = {
+            'PATH': ('xbin', 'pibin'),
```
akesandgren, line 1173: Typo? xbin?? original was just bin

boegel: That's intentional, since `bin` is a special case (it gets added to `PATH` by default)

easybuild/tools/module_generator.py
boegel: Yes, that would be helpful indeed, I'll change that, thanks for the suggestion!

bartoldeman (approved):
LGTM

Issue #4009, opened 2022-05-17
Time to switch to `icx/icpx` for Intel C/C++ compilers


ocaisa:
The current recommendation from Intel is that we should be using the `icx/icpx` LLVM-based compilers by default

ocaisa:
I _think_ this would mean adding a version check to https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/toolchains/compiler/intel_compilers.py#L59 and overriding the values `self.variables['CXX']`/` self.variables['CC']` (that's assuming none of the compiler arguments have changed)

ocaisa:
Looking at https://www.intel.com/content/www/us/en/developer/articles/guide/porting-guide-for-icc-users-to-dpcpp-or-icx.html 
```
- ICX drivers icx and icpx will accept ICC Classic Compiler options or Clang*/LLVM Compiler options.
  - Clang*/LLVM Compiler options are interpreted directly.
  - Classic ICC Compiler options passed to ICX are translated to their Clang*/LLVM equivalents, wherever possible.
- Not all ICC Classic options are accepted and/or implemented in ICX.
```
We would also need to use `-fiopenmp` for OpenMP (according to that page).

I wonder if we will also end up with version differences compared to `ifort`?

akesandgren:
Should we do this for the 2022a toolchain version?

ocaisa:
Since #4031 includes (initial) support for these has been merged I think we can close this

PR #4031, opened 2022-06-21
Support for OneAPI compilers using 'oneapi': True

bartoldeman:
If toolchainopts = {'oneapi': True}, compilers change to icx/icpx/ifx.

The options `-fp-model source` and `-fp-model fast=1` no longer work with
OneAPI. Change them to the equivalent `-fp-model precise` and `-fp-model fast`
options.

See #4009

### DIFF
--- a/easybuild/toolchains/compiler/intel_compilers.py
+++ b/easybuild/toolchains/compiler/intel_compilers.py
@@ -39,6 +39,8 @@ class IntelCompilers(IntelIccIfort):
     """
 
     COMPILER_MODULE_NAME = ['intel-compilers']
+    COMPILER_UNIQUE_OPTS = dict(IntelIccIfort.COMPILER_UNIQUE_OPTS,
+                                oneapi = (False, "Use oneAPI compilers icx/icpx/ifx instead of classic compilers"))
 
     def _set_compiler_vars(self):
         """Intel compilers-specific adjustments after setting compiler variables."""
@@ -58,6 +60,19 @@ class IntelCompilers(IntelIccIfort):
 
     def set_variables(self):
         """Set the variables."""
+
+        if self.options.get('oneapi', False):
+            self.COMPILER_CXX = 'icpx'
+            self.COMPILER_CC = 'icx'
+            self.COMPILER_F77 = 'ifx'
+            self.COMPILER_F90 = 'ifx'
+            self.COMPILER_FC = 'ifx'
+            # fp-model source is not supported by icx but is equivalent to precise
+            self.options.options_map['defaultprec'] = ['ftz', 'fp-speculation=safe', 'fp-model precise']
+            # icx doesn't like -fp-model fast=1; fp-model fast is equivalent
+            self.options.options_map['loose'] = ['fp-model fast']
+            # recommended in porting guide
+            self.options.options_map['openmp'] = ['fiopenmp']
 
         # skip IntelIccIfort.set_variables (no longer relevant for recent versions)
         Compiler.set_variables(self)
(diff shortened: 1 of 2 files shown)


easybuild/toolchains/compiler/intel_compilers.py
```diff
@@ -59,5 +61,18 @@ def _set_compiler_vars(self):
...
+        if self.options.get('oneapi', False):
+            self.COMPILER_CXX = 'icpx'
+            self.COMPILER_CC = 'icx'
+            self.COMPILER_F77 = 'ifx'
+            self.COMPILER_F90 = 'ifx'
+            self.COMPILER_FC = 'ifx'
+            # fp-model source is not supported by icx but is equivalent to precise
+            self.options.options_map['defaultprec'] = ['ftz', 'fp-speculation=safe', 'fp-model precise']
+            # icx doesn't like -fp-model fast=1; fp-model fast is equivalent
+            self.options.options_map['loose'] = ['fp-model fast']
```
akesandgren, line 73: Does it handle -fp-model fast=2 then? I.e. veryloose

akesandgren: So we should perhaps override veryloose too?

bartoldeman: It wouldn't make a difference in practice apart from a line in the log, unless at some point in the future fast=2 will be supported again.

I'm happy to override it but then we may need to add a version check in the future to re-enable it I suppose

bartoldeman: https://www.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compiler-reference/compiler-options/floating-point-options/fp-model-fp.html

bartoldeman: "There is currently no difference between fast=1 and fast=2."

But for some reason fast=1 is flat out rejected with an error (so docs are wrong here), fast=2 as accepted with a warning

bartoldeman: done

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @bartoldeman!

easybuild/toolchains/compiler/intel_compilers.py
```diff
@@ -59,5 +61,20 @@ def _set_compiler_vars(self):
...
         """Set the variables."""
 
+        if self.options.get('oneapi', False):
+            self.COMPILER_CXX = 'icpx'
+            self.COMPILER_CC = 'icx'
+            self.COMPILER_F77 = 'ifx'
+            self.COMPILER_F90 = 'ifx'
+            self.COMPILER_FC = 'ifx'
+            # fp-model source is not supported by icx but is equivalent to precise
+            self.options.options_map['defaultprec'] = ['ftz', 'fp-speculation=safe', 'fp-model precise']
```
robogast, line 71: `ftz` doens't work for me in `ipx` and `ipcx`

robogast: @bartoldeman

bartoldeman: ftz works in icx 2022 but not in 2021 indeed. So we'll need a version check.

boegel: @bartoldeman Please open an issue on this ASAP, so we don't forget to fix it. Or, if it's an easy fix, a PR, then we can still squeeze in the fix for the EasyBuild v4.6.0 release we'll push out this week...

PR #4036, opened 2022-07-01
add support for using template values in name/version of extensions

Flamefire:
Allows to use the parent ECs values in the extension name and version.
Useful for e.g. TensorFlow where extension versions are often coupled to
the `%(version_major_minor)s` of the parent.


boegel (approved):
lgtm

PR #4049, opened 2022-08-02
fix code style issues reported by recent flake8 linter

Flamefire:
Recent flake8 complains about a missing space after keywords:

- C-style `while(condition)` instead of `while condition:`
- `not(condition)` (actually better: `return self.version is not None and self.operator is not None`, but that's personal preference)
- Rewrote the if-condition checking for `stream` attr to remove the `or`
- Missing error class in `raise`. Settled on `ValueError` as the hdlr has a missing or invalid `stream` value


boegel (approved):
lgtm

PR #4054, opened 2022-08-05
include 'sources' key to entries in list returned by collect_exts_file_info

Flamefire:
The returned dict per extension has a 'patches' key but no 'sources'.
Add this incoorporating `source_tmpl` or the default value.

Add test for that and also for `nosource: True`.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -725,6 +725,11 @@ class EasyBlock:
                                 # copy 'path' entry to 'src' for use with extensions
                                 'src': src['path'],
                             })
+                            filename = src['name']
+                        else:
+                            filename = source.get('filename')
+                        if filename is not None:
+                            ext_src['sources'] = [filename]
 
                     else:
 
@@ -738,6 +743,7 @@ class EasyBlock:
                             raise EasyBuildError(error_msg, type(src_fn).__name__, src_fn)
 
                         src_fn = resolve_template(src_fn, template_values)
+                        ext_src['sources'] = [src_fn]
 
                         if fetch_files:
                             src_path = self.obtain_file(src_fn, extension=True, urls=source_urls,
--- a/test/framework/easyblock.py
+++ b/test/framework/easyblock.py
@@ -2350,17 +2350,25 @@ class EasyBlockTest(EnhancedTestCase):
         toy_sources = os.path.join(testdir, 'sandbox', 'sources', 'toy')
         toy_ext_sources = os.path.join(toy_sources, 'extensions')
         toy_ec_file = os.path.join(testdir, 'easyconfigs', 'test_ecs', 't', 'toy', 'toy-0.0-gompi-2018a-test.eb')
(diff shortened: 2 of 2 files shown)


casparvl:
@Flamefire Can you provide a bit of context for this PR? Is there a related issue, and if not, can you shortly describe what was the previous behavior, why that was unwanted, and what the (intended) behavior of your fix is?

I saw your thread with @boegel on this in Slack, but it wasn't fully clear to me what 'inconsistency' you were referring to there. Plus, to make it easier to find the reason for this change (also in the future), it's probably good to log it here together with the PR :)

Flamefire:
This came up during development of a fix for a bug in the CI checks, see: https://github.com/easybuilders/easybuild-easyconfigs/pull/15973

Basically the function `collect_exts_file_info` returns a dict where the `patches` are included but not `sources` which is confusing at least and makes one wonder whether to use `['options']['patches']` or `['patches']` or `['options']['sources`]` etc respectively.

Especially as this functions is intended to handle the correct resolution  of templates in the extension sources and patches and there is even a special case (`nosource`) which needs to be handled this should be done completely in that function.

So the intended behavior is that the returned dict for each extensions contains a `sources` key with a list of sources if-and-only-if `nosource` is not set. The sources should have correctly resolved templates (i.e. using name and versions of the extension not the parent EC)

So actually this is a 2 part fix:

- Return the (correct) `sources` at all (i.e. especially using the default value when missing or derive from `source_tmpl`)
- Return `sources` and `patches` consistently in the main dict and not in the sub-dict `options`

easybuild/framework/easyblock.py
```diff
@@ -594,6 +594,7 @@ def collect_exts_file_info(self, fetch_files=True, verify_checksums=True):
                         if 'source_urls' not in source:
                             source['source_urls'] = source_urls
 
+                        ext_src['sources'] = sources
```
boegel, line 597: Hmm, shouldn't this be `[source]` (to always have a dict value here)?

Otherwise we'll sometimes have a dict value, sometimes a string value, which is annoying to deal with wherever this is used...

boegel (changes requested):

easybuild/framework/easyblock.py
Flamefire: The idea is that similar to `'patches'` this is always a list of filenames. See the test(s). This is especially important for the `nosource` option.

Flamefire:
CI failure is unrelated:

> `Unexpected error occurred when trying to download https://ftpmirror.gnu.org/gnu/binutils/binutils-2.37.tar.gz to /scratch/sources/b/binutils/binutils-2.37.tar.gz: 'Failed to write to /scratch/sources/b/binutils/binutils-2.37.tar.gz: The read operation timed out' (took 17 secs)`

boegel:
@Flamefire Please sync again with `develop` after merge of #4255

Flamefire:
resynced. Python 2 CI failure is unrelated

PR #4052, opened 2022-08-03
Support for passing from-pr in an easyconfig-based Easystack file [WIP]

casparvl:
Note that this PR builds on top of https://github.com/easybuilders/easybuild-framework/pull/4021 (because we need option parsing for EasyStack files, before we can do anything with the options...). Thus, that part might still change.

At this point in time, the behavior of this PR on the following example:

```
robot: true
easyconfigs:
  - ncurses-6.3-GCCcore-11.3.0.eb
  - expecttest-0.1.3-GCCcore-11.3.0.eb:
      options: {
        'from_pr': [15922,15923],
      }
  - FFmpeg-5.0.1-GCCcore-11.3.0.eb:
      options: {
        'from_pr': [15918],
      }
```

Is:

```
eb_github -D --easystack easystack_flat_ebs_from_pr.yaml -r --debug
== Temporary log file in case of crash /scratch/casparl/eb-xdu4273p/easybuild-uyvbixfc.log

WARNING: Specifying options in easystack files is not supported yet. They are parsed, but ignored.


WARNING: Using easyconfigs from closed PR #15923

Dry run: printing build status of easyconfigs and dependencies
 * [x] /home/casparl/.local/easybuild/Debian10/2022/software/EasyBuild/4.6.0-dev/easybuild/easyconfigs/m/M4/M4-1.4.19.eb (module: M4/1.4.19)
 * [x] /home/casparl/.local/easybuild/Debian10/2022/software/EasyBuild/4.6.0-dev/easybuild/easyconfigs/b/Bison/Bison-3.8.2.eb (module: Bison/3.8.2)
 * [x] /home/casparl/.local/easybuild/Debian10/2022/software/EasyBuild/4.6.0-dev/easybuild/easyconfigs/f/flex/flex-2.6.4.eb (module: flex/2.6.4)
 * [x] /home/casparl/.local/easybuild/Debian10/2022/software/EasyBuild/4.6.0-dev/easybuild/easyconfigs/z/zlib/zlib-1.2.12.eb (module: zlib/1.2.12)
 * [x] /home/casparl/.local/easybuild/Debian10/2022/software/EasyBuild/4.6.0-dev/easybuild/easyconfigs/b/binutils/binutils-2.38.eb (module: binutils/2.38)
 * [x] /home/casparl/.local/easybuild/Debian10/2022/software/EasyBuild/4.6.0-dev/easybuild/easyconfigs/g/GCCcore/GCCcore-11.3.0.eb (module: GCCcore/11.3.0)
[...]

PR #4057, opened 2022-08-10
take into account custom configuration options specified in easystack file + drop support for easystack files using 'software' top-level key

casparvl:
Replaces https://github.com/easybuilders/easybuild-framework/pull/4052

After discussion with @boegel we've decided that the the approach in #4052 is not maintainable. By pushing the loop over the entries in an EasyStack file to `main.py`, and by basically re-executing part of `main.py` for each item in the EasyStack file, we can properly reinitialize the configuration options for each item (if there are EasyConfig-specific options defined). This has the advantage that we can reuse the existing machinery of things like `set_up_configuration`.

It also means that EasyBuild builds a dependency tree _per_ item in the EasyStack file. On the one hand, that is desirable since two items might have different `--from-pr` arguments, and thus the same dependency `A.eb` might in theory resolve to two different PRs depending on the options passed for that item in the EasyStack file. The downside is that since we don't have a single dependency graph, we cannot e.g. submit a single set of jobs with `--job`: we'll need to make sure EasyBuild becomes aware of the order in the EasyStack file, and make sure that jobs for later items wait for _all_ jobs of earlier items. That's something we'll implement later though...

Note that this is really a work in progress, a lot more functions need to be put in place before this properly works...


easybuild/main.py
```diff
@@ -586,9 +481,198 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+        # for path in orig_paths:
+        #     validate_command_opts(args, opts_per_ec[path])
+        # Loop over each item in the EasyStack file, each time updating the config
+        # This is because each item in an EasyStack file can have options associated with it
+        for path in orig_paths:
+            _log.debug("Starting build for %s" % path)
+            # NOTE: not sure if this is needed. Is the EasyConfigs cache preserved throughout loops of this iteration?
+            # Current 'path' may have different options associated with it. Thus, resolution of EasyConfigs
+            # to full paths should not be read from cache, but redetermined. Thus, we wipe the cache
+            # wipe_cache()  # Wipes the easyconfig cache with _easyconfigs_cache.clear()  # TODO: implement
```
casparvl, line 642: Which of https://github.com/easybuilders/easybuild-framework/blob/d7409b3d68516a0bf70688d14e5750dedfbb8758/test/framework/utilities.py#L453 should it empty? All? Only the `_easyconfigs_cache`?

easybuild/main.py
```diff
@@ -586,9 +481,198 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+        if orig_paths:
+            raise EasyBuildError("Installing the latest EasyBuild release can not be combined with installing "
+                                 "other easyconfigs")
+        else:
+            eb_file = find_easybuild_easyconfig()
+            orig_paths.append(eb_file)
+
+    if options.easystack:
+        _log.debug("Start build loop over items in the EasyStack file: %s" % orig_paths)
+        # TODO: insert fast loop that validates if all command line options are valid. If there are errors in options,
```
casparvl, line 631: We could do this here. Alternatively, (and probably better) we could do it when the EasyStack gets parsed, i.e. right before `parse` returns, we check the options https://github.com/easybuilders/easybuild-framework/blob/56e3bfc21507b5b509e919cfc139345d186ed43b/easybuild/framework/easystack.py#L157

easybuild/framework/easystack.py
```diff
@@ -243,12 +317,16 @@ def parse_easystack(filepath):
 
     easyconfig_names = easystack.compose_ec_filenames()
 
-    general_options = easystack.get_general_options()
+    # Disabled general options for now. We weren't using them, and first want support for EasyConfig-specific options.
+    # Then, we need a method to resolve conflicts (specific options should win)
+    # general_options = easystack.get_general_options()
```
boegel, line 322: @casparvl We should clean up this commented out block, same below

easybuild/main.py
```diff
@@ -586,9 +481,198 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+        else:
+            print_warning("Will not run the test step as requested via skip-test-step. "
+                          "Consider using ignore-test-failure instead and verify the results afterwards")
+
+    # determine easybuild-easyconfigs package install path
+    easyconfigs_pkg_paths = get_paths_for(subdir=EASYCONFIGS_PKG_SUBDIR)
+    if not easyconfigs_pkg_paths:
+        _log.warning("Failed to determine install path for easybuild-easyconfigs package.")
+
+    if options.install_latest_eb_release:
```
boegel, line 621: this block should be moved into `rest_of_main` (since it uses `orig_paths`)

easybuild/main.py
```diff
@@ -586,9 +481,198 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+    modtool.check_loaded_modules()
+
+    if options.last_log:
+        # print location to last log file, and exit
+        last_log = find_last_log(logfile) or '(none)'
+        print_msg(last_log, log=_log, prefix=False)
+
+    # if easystack is provided with the command, commands with arguments from it will be executed
+    if options.easystack:
+        orig_paths, opts_per_ec = parse_easystack(options.easystack)
```
boegel, line 530: this should be moved down, into the lower `if options.easystack`:

boegel (changes requested):

easybuild/main.py
```diff
@@ -586,9 +490,191 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+                _log.info("Argument list for %s after merging command line arguments with EasyConfig specific "
+                          "options from the EasyStack file: %s" % (path, new_args))
+            else:
+                # If no EasyConfig specific arguments are defined, sse original args.
+                # That way,set_up_configuration restores the original config
+                new_args = args
+            eb_go, cfg_settings = set_up_configuration(args=new_args, logfile=logfile, testing=testing, reconfigure=True)
+
+            hooks = load_hooks(options.hooks)
+            overall_success = rest_of_main([path], eb_go.options, cfg_settings, modtool, testing, init_session_state,
```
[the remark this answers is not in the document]
boegel: Should be something like this:

```python
overall_success = True
if easystack:
  for path in orig_paths
    overall_success &= rest_of_main([path], ...)
else:
    overall_success = rest_of_main(...)
```

And the only do the `if overall_success` + `cleanup` outside of the `if/else`

boegel (changes requested):

easybuild/main.py
```diff
@@ -395,7 +269,10 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
                 print_msg("Contents of %s:" % path)
                 print_msg(read_file(path), prefix=False)
 
-        clean_exit(logfile, eb_tmpdir, testing)
+        if not skip_clean_exit:
+            clean_exit(logfile, eb_tmpdir, testing)
+        else:
+            return
```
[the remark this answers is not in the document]
boegel: We can just replace with `if`/`else` with `return True`, and then let the return value indicate whether or not `cleanup` should be run after running the `rest_of_main` function

easybuild/main.py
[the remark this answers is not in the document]
boegel: And `overall_success` should be renamed to `do_cleanup` here, since the return value of `rest_of_main` indicates whether or not cleanup should happen.

easybuild/main.py
```diff
@@ -213,149 +213,20 @@ def clean_exit(logfile, tmpdir, testing, silent=False):
     sys.exit(0)
 
 
-def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
+def rest_of_main(orig_paths, options, cfg_settings, modtool, testing, init_session_state, hooks, do_build):
```
boegel, line 216: I would rename `rest_of_main` to `process_eb_args`, rename `orig_paths` to `args` , and also tweak the description of this function:

```python
def process_eb_args(args, options, ...):
    """
    Remainder of main function, actually process provided arguments (list of files/paths),
    according to specified options.

    :param args: list of arguments that were specified to 'eb' command (or an easystack file);
                 includes filenames/paths of files to process
                 (mostly easyconfig files, but can also includes patch files, etc.)

    ...
    """
```

boegel (changes requested):

easybuild/tools/options.py
```diff
@@ -1444,16 +1445,7 @@ def det_location(opt, prefix=''):
...
-        # very early debug, to debug the generaloption itself
-        fancylogger.logToScreen(enable=True)
-        fancylogger.setLogLevel('DEBUG')
-
-    if args is None:
-        args = sys.argv[1:]
-
+def unroll_arguments(args):
     # unroll arguments that correspond to a combo of single-letter options
     # this is done to avoid interpreting -rD like "--robot D" instead of "--robot -D"
```
boegel, line 1450: comments should be change to a proper docstring (and we should add a test for this new function)

easybuild/tools/options.py
```diff
@@ -1878,3 +1898,28 @@ def set_tmpdir(tmpdir=None, raise_error=False):
             raise EasyBuildError("Failed to test whether temporary directory allows to execute files: %s", err)
 
     return current_tmpdir
+
+
+def dict_to_argslist(args_dict):
```
boegel, line 1903: We should also add a test for `dict_to_argslist`

easybuild/tools/options.py
```diff
@@ -1581,6 +1590,17 @@ def set_up_configuration(args=None, logfile=None, testing=False, silent=False):
         'try_to_generate': try_to_generate,
         'valid_stops': [x[0] for x in EasyBlock.get_steps()],
     }
+
+    # Remove existing singletons if reconfigure==True (allows reconfiguration when looping over EasyStack items)
+    if reconfigure:
+        BuildOptions.__class__._instances.clear()
+    elif len(BuildOptions.__class__._instances) > 0:
+        log.warn("set_up_configuration is about to call init() and init_build_options(). "
```
boegel, line 1598: better use `print_warning` here, otherwise the warning will only appear in the log file

```python
msg = '\n'.join([
    "...",
    "...",
])
print_warning(msg, log=log)
```

boegel (changes requested):

easybuild/main.py
```diff
@@ -590,10 +468,198 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+        #     cleanup(logfile, eb_tmpdir, testing)
+    else:
+        do_cleanup = process_eb_args(orig_paths, options, cfg_settings, modtool, testing, init_session_state, hooks,
+                                     do_build)
     # stop logging and cleanup tmp log file, unless one build failed (individual logs are located in eb_tmpdir)
     stop_logging(logfile, logtostdout=options.logtostdout)
-    if overall_success:
-        cleanup(logfile, eb_tmpdir, testing)
+    if do_cleanup:
+        print("Running cleanup")
```
boegel, line 661: please remove this `print` again, probably leftover of debugging

easybuild/main.py
```diff
@@ -213,149 +213,23 @@ def clean_exit(logfile, tmpdir, testing, silent=False):
...
-            print_warning("Will not run the test step as requested via skip-test-step. "
-                          "Consider using ignore-test-failure instead and verify the results afterwards")
-
-    # determine easybuild-easyconfigs package install path
-    easyconfigs_pkg_paths = get_paths_for(subdir=EASYCONFIGS_PKG_SUBDIR)
-    if not easyconfigs_pkg_paths:
-        _log.warning("Failed to determine install path for easybuild-easyconfigs package.")
-
     if options.install_latest_eb_release:
         if orig_paths:
```
boegel, line 234: You'll need to rename `orig_paths` to `args` throughout `process_eb_args`...

boegel (changes requested):

easybuild/tools/options.py
```diff
@@ -1581,6 +1590,17 @@ def set_up_configuration(args=None, logfile=None, testing=False, silent=False):
         'try_to_generate': try_to_generate,
         'valid_stops': [x[0] for x in EasyBlock.get_steps()],
     }
+
+    # Remove existing singletons if reconfigure==True (allows reconfiguration when looping over EasyStack items)
+    if reconfigure:
+        BuildOptions.__class__._instances.clear()
```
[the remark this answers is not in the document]
casparvl: Done

easybuild/tools/options.py
```diff
@@ -1581,6 +1590,20 @@ def set_up_configuration(args=None, logfile=None, testing=False, silent=False):
         'try_to_generate': try_to_generate,
         'valid_stops': [x[0] for x in EasyBlock.get_steps()],
     }
+
+    # Remove existing singletons if reconfigure==True (allows reconfiguration when looping over EasyStack items)
+    if reconfigure:
```
boegel, line 1589: Also for this new aspect of `set_up_configuration` we should make sure there's a test, to verify that this is indeed working as expected...

It seems like there's no dedicated test for `set_up_configuration` at all yet, that belongs in `easybuild.tools.options`...

boegel (changes requested):

easybuild/main.py
boegel: It basically boils down to this I think (wipe both the cache of `EasyConfig` instances + the cache of where easyconfig files are located (see `robot_find_easyconfig` function)

```python
from easybuild.framework.easyconfig import easyconfig
...
easyconfig._easyconfigs_cache.clear()
easyconfig._easyconfig_files_cache.clear()
```

boegel (changes requested):

easybuild/main.py
```diff
@@ -590,10 +468,198 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+                          "Consider using ignore-test-failure instead and verify the results afterwards")
+
+    # determine easybuild-easyconfigs package install path
+    easyconfigs_pkg_paths = get_paths_for(subdir=EASYCONFIGS_PKG_SUBDIR)
+    if not easyconfigs_pkg_paths:
+        _log.warning("Failed to determine install path for easybuild-easyconfigs package.")
+
+    # if EasyStack file is provided, parse it, and loop over the items in the EasyStack file
+    if options.easystack:
+        orig_paths, opts_per_ec = parse_easystack(options.easystack)
```
boegel, line 611: It probably makes sense to create a dedicated function for this block of code, so in `main` we just get:

```python
if options.easystack:
    process_easystack(path)
```

Maybe we should also print a warning if both `orig_paths` and `--easystack` are used?

Eventually we probably want to drop the `--easystack` option and simply do this instead:

```python
for path in easystack_paths:
    process_easystack(path)
```

boegel (changes requested):

boegel:
TODO for tests (I'll try to look into that):
- add test for `set_up_configuration` (which also covers `reconfigure`)
- add test for new `dict_to_argslist` function
- add end-to-end test with easystack file that sets options for some easyconfigs + verify result (using `toy` installs)

easybuild/main.py
boegel: fixed in 819138d1d532bf185ba2b63d880e8dfc1fb70ee6

easybuild/main.py
boegel: fixed in 3c0252eae690b675f5c3a9a6358fd6d9c63963e6

easybuild/main.py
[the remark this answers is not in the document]
boegel: fixed in 8afef99d9af118687955aec402eeaca7db39a5cb

easybuild/main.py
casparvl: Done

easybuild/main.py
boegel: fixed in e107a4b55670c9386ed518e91cc39dc431fa6267

easybuild/main.py
casparvl: Done

easybuild/tools/options.py
casparvl: Done

easybuild/tools/options.py
casparvl: This function was inlined again, so no description (or seperate test) needed anymore

easybuild/tools/options.py
boegel: fixed in d1298598e45abea579ae25c8a1d00ab827ceee89

easybuild/main.py
casparvl: Ok, we'll leave the TODO for now

easybuild/main.py
boegel: d98aa4673f9b7c2d46d49e39d6d336e0b5c8e7fe

easybuild/main.py
casparvl: Done

easybuild/main.py
```diff
@@ -590,10 +469,192 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+                _log.info("Argument list for %s after merging command line arguments with EasyConfig specific "
+                          "options from the EasyStack file: %s" % (path, new_args))
+            else:
+                # If no EasyConfig specific arguments are defined, use original args.
+                # That way,set_up_configuration restores the original config
+                new_args = args
+            eb_go, cfg_settings = set_up_configuration(args=new_args, logfile=logfile, testing=testing,
+                                                       reconfigure=True)
+
+            hooks = load_hooks(options.hooks)
```
casparvl, line 648: This should probably use `eb_go.options.hooks`, as otherwise it doesn't respect any hooks that might be specified in the EasyStack file

easybuild/main.py
```diff
@@ -590,10 +469,192 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+            do_cleanup &= process_eb_args([path], eb_go.options, cfg_settings, modtool, testing, init_session_state,
```
casparvl, line 649: This should probably not use the modtool that is already setup, but set up a new one using
```
modtool = modules_tool(testing=testing)
```
In theory, the EasyStack file could specify to use Tcl modules for one EasyConfig, and Lmod for another.

easybuild/main.py
```diff
@@ -213,162 +214,100 @@ def clean_exit(logfile, tmpdir, testing, silent=False):
...
+            # If no EasyConfig specific arguments are defined, use original args.
+            # That way,set_up_configuration restores the original config
+            new_args = args
+
+        # Reconfigure
+        eb_go, cfg_settings = set_up_configuration(args=new_args, logfile=logfile, testing=testing,
+                                                   reconfigure=True)
+        # Since we reconfigure, we should also reload hooks and get current module tools
+        hooks = load_hooks(eb_go.options.hooks)
         modtool = modules_tool(testing=testing)
```
casparvl, line 268: @boegel I figured that we should also re-fetch the modules-tool (didn't do that before) right after `set_up_configuration`, agreed? It could be that different items in the EasyStack file use different module tools. From what I understand, `modules_tool` fetches this from `ConfigurationVariables` (see [here](https://github.com/casparvl/easybuild-framework/blob/5096c4cc324bbd9642de7070cb87cc672a7d3bc0/easybuild/tools/config.py#L704)), which is reset by running `set_up_configuration` with `reconfigure=True`. Correct?

easybuild/main.py
```diff
@@ -213,162 +214,100 @@ def clean_exit(logfile, tmpdir, testing, silent=False):
...
+    # for path in orig_paths:
+    #     validate_command_opts(args, opts_per_ec[path])
+
+    # Loop over each item in the EasyStack file, each time updating the config
+    # This is because each item in an EasyStack file can have options associated with it
+    do_cleanup = True
+    for path in orig_paths:
+        _log.debug("Starting build for %s" % path)
+        # Whipe easyconfig caches
+        easyconfig._easyconfigs_cache.clear()
```
casparvl, line 243: @boegel I'm not sure this is actually needed. I created the following test:
```
cat test_clear_easyconfig_cache.yaml
robot: true
easyconfigs:
  - GCC-11.3.0.eb
  - binutils-2.38-GCCcore-11.3.0.eb:
      options: {
        'from-pr': '16386',
      }
  - GCC-11.3.0.eb
```
My assumption was that, without clearing the cache, the same EasyConfig file would be used every time for `binutils-2.38-GCCcore-11.3.0.eb`. But, even if I comment these two `clear()` lines out, I get

```
CFGS=/sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs
 * [x] $CFGS/m/M4/M4-1.4.19.eb (module: M4/1.4.19)
 * [x] $CFGS/b/Bison/Bison-3.8.2.eb (module: Bison/3.8.2)
 * [x] $CFGS/f/flex/flex-2.6.4.eb (module: flex/2.6.4)
 * [x] $CFGS/z/zlib/zlib-1.2.12.eb (module: zlib/1.2.12)
 * [x] $CFGS/b/binutils/binutils-2.38.eb (module: binutils/2.38)
 * [x] $CFGS/g/GCCcore/GCCcore-11.3.0.eb (module: GCCcore/11.3.0)
 * [x] $CFGS/z/zlib/zlib-1.2.12-GCCcore-11.3.0.eb (module: zlib/1.2.12-GCCcore-11.3.0)
 * [x] $CFGS/h/help2man/help2man-1.49.2-GCCcore-11.3.0.eb (module: help2man/1.49.2-GCCcore-11.3.0)
 * [x] $CFGS/m/M4/M4-1.4.19-GCCcore-11.3.0.eb (module: M4/1.4.19-GCCcore-11.3.0)
 * [x] $CFGS/b/Bison/Bison-3.8.2-GCCcore-11.3.0.eb (module: Bison/3.8.2-GCCcore-11.3.0)
 * [x] $CFGS/f/flex/flex-2.6.4-GCCcore-11.3.0.eb (module: flex/2.6.4-GCCcore-11.3.0)
 * [x] $CFGS/b/binutils/binutils-2.38-GCCcore-11.3.0.eb (module: binutils/2.38-GCCcore-11.3.0)
 * [x] $CFGS/g/GCC/GCC-11.3.0.eb (module: GCC/11.3.0)
== Temporary log file in case of crash /scratch/casparl/eb-l8wbjlkf/easybuild-164xa9n9.log
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/m/M4/M4-1.4.19.eb (module: M4/1.4.19)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/b/Bison/Bison-3.8.2.eb (module: Bison/3.8.2)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/f/flex/flex-2.6.4.eb (module: flex/2.6.4)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/z/zlib/zlib-1.2.12.eb (module: zlib/1.2.12)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/b/binutils/binutils-2.38.eb (module: binutils/2.38)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/g/GCCcore/GCCcore-11.3.0.eb (module: GCCcore/11.3.0)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/h/help2man/help2man-1.49.2-GCCcore-11.3.0.eb (module: help2man/1.49.2-GCCcore-11.3.0)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/m/M4/M4-1.4.19-GCCcore-11.3.0.eb (module: M4/1.4.19-GCCcore-11.3.0)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/z/zlib/zlib-1.2.12-GCCcore-11.3.0.eb (module: zlib/1.2.12-GCCcore-11.3.0)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/b/Bison/Bison-3.8.2-GCCcore-11.3.0.eb (module: Bison/3.8.2-GCCcore-11.3.0)
 * [x] /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs/f/flex/flex-2.6.4-GCCcore-11.3.0.eb (module: flex/2.6.4-GCCcore-11.3.0)
 * [x] /scratch/casparl/eb-blcb26tg/files_pr16386/b/binutils/binutils-2.38-GCCcore-11.3.0.eb (module: binutils/2.38-GCCcore-11.3.0)
== Temporary log file in case of crash /scratch/casparl/eb-l8wbjlkf/easybuild-164xa9n9.log
== found valid index for /sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs, so using it...
CFGS=/sw/noarch/Debian10/2022/software/EasyBuild/4.6.1/easybuild/easyconfigs
 * [x] $CFGS/m/M4/M4-1.4.19.eb (module: M4/1.4.19)
 * [x] $CFGS/b/Bison/Bison-3.8.2.eb (module: Bison/3.8.2)
 * [x] $CFGS/f/flex/flex-2.6.4.eb (module: flex/2.6.4)
 * [x] $CFGS/z/zlib/zlib-1.2.12.eb (module: zlib/1.2.12)
 * [x] $CFGS/b/binutils/binutils-2.38.eb (module: binutils/2.38)
 * [x] $CFGS/g/GCCcore/GCCcore-11.3.0.eb (module: GCCcore/11.3.0)
 * [x] $CFGS/z/zlib/zlib-1.2.12-GCCcore-11.3.0.eb (module: zlib/1.2.12-GCCcore-11.3.0)
 * [x] $CFGS/h/help2man/help2man-1.49.2-GCCcore-11.3.0.eb (module: help2man/1.49.2-GCCcore-11.3.0)
 * [x] $CFGS/m/M4/M4-1.4.19-GCCcore-11.3.0.eb (module: M4/1.4.19-GCCcore-11.3.0)
 * [x] $CFGS/b/Bison/Bison-3.8.2-GCCcore-11.3.0.eb (module: Bison/3.8.2-GCCcore-11.3.0)
 * [x] $CFGS/f/flex/flex-2.6.4-GCCcore-11.3.0.eb (module: flex/2.6.4-GCCcore-11.3.0)
 * [x] $CFGS/b/binutils/binutils-2.38-GCCcore-11.3.0.eb (module: binutils/2.38-GCCcore-11.3.0)
 * [x] $CFGS/g/GCC/GCC-11.3.0.eb (module: GCC/11.3.0)
```
I did not expect that to work with the EasyConfig cache not being whiped... Is it possible that something else (e.g. the `set_up_configuration`) is already clearing the cache / creating a new clean cache? In that case, these two lines could be removed.

easybuild/main.py
boegel: Makes sense, yes, since we're passing this down to `process_eb_args`

easybuild/main.py
boegel: Yes, correct.

easybuild/main.py
boegel: Exactly, and already fixed now.

easybuild/main.py
```diff
@@ -213,162 +214,100 @@ def clean_exit(logfile, tmpdir, testing, silent=False):
...
+        easyconfig._easyconfig_files_cache.clear()
+
+        # If EasyConfig specific arguments were supplied in EasyStack file
+        # merge arguments with original command line args
+        if path in opts_per_ec:
+            _log.debug("EasyConfig specific options have been specified for "
+                       "%s in the EasyStack file: %s" % (path, opts_per_ec[path]))
+            if args is None:
+                args = sys.argv[1:]
+            ec_args = dict_to_argslist(opts_per_ec[path])
```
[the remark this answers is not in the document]
boegel: I guess for now we should add a check that the same easyconfig can't be listed twice in the same easystack file?
This can be handled in a separate PR though, I agree.

easybuild/main.py
```diff
@@ -213,162 +214,109 @@ def clean_exit(logfile, tmpdir, testing, silent=False):
...
+    _log.debug("Start build loop over items in the EasyStack file: %s" % orig_paths)
+    # TODO: insert fast loop that validates if all command line options are valid. If there are errors in options,
+    # we want to know early on, and this loop potentially builds a lot of packages and could take very long
+    # for path in orig_paths:
+    #     validate_command_opts(args, opts_per_ec[path])
+
+    # Loop over each item in the EasyStack file, each time updating the config
+    # This is because each item in an EasyStack file can have options associated with it
+    do_cleanup = True
+    for path in orig_paths:
```
boegel, line 241: This should become a for loop over tuples, so we can support the use case of having the same easyconfig file listed multiple times in an easystack file, with different options (with and without `--from-pr`, for example):

```python
easystack = parse_easystack(easystack_path)
...
for (path, opts) in easystack:
    ...
    if opts:
        ...
        ec_args = opts_dict_to_eb_opts[opts]
```

boegel (changes requested):
In `parse_by_easyconfigs` we need to verify that `easyconfigs` is a *list*, not a *dict*

Test case:

- wrong (dict value):
```yaml
easyconfigs:
    bzip2-1.0.6:
        options:
            hidden: True
```
- right (list value):
```yaml
easyconfigs:
    - bzip2-1.0.6:
        options:
            hidden: True
```

test/framework/options.py
```diff
@@ -6542,6 +6544,173 @@ def test_easystack_basic(self):
             regex = re.compile(pattern)
             self.assertTrue(regex.search(stdout), "Pattern '%s' should be found in: %s" % (regex.pattern, stdout))
```
casparvl, line 6546: Also add test that uses incorrect format for EasyStack (i.e. with `-` before the EasyConfig name, so that `easyconfigs:` becomes a dict instead of a list). This format is invalid and won't be parsed correctly. It _should_ raise an error.

test/framework/easystack.py
```diff
@@ -71,10 +71,11 @@ def test_easystack_basic(self):
         self.assertEqual(sorted(ec_fns), sorted(expected))
         self.assertEqual(opts, {})
 
-    def test_easystack_easyconfigs(self):
-        """Test for easystack file using 'easyconfigs' key."""
+
+    def test_easystack_basic_dict(self):
```
casparvl, line 75: Check if the format in `test_easystack_basic_dict.yaml` _should_ be valid. If so, we should change the parser to make this pass. If not, we need to change this test and check for the failure

boegel: fixed in a5ce7d2

easybuild/main.py
boegel: The problem here was that you were listed the same easyconfig twice in the same easystack. That didn't work due to how we were looping over easystack items, which we changed in 63a879d

easybuild/tools/options.py
casparvl: Done!

easybuild/tools/options.py
casparvl: Done in https://github.com/easybuilders/easybuild-framework/pull/4057/commits/b11f346604599eea8e1aec6f8f9be4c9dea5382c

easybuild/main.py
casparvl: Dedicated function: https://github.com/easybuilders/easybuild-framework/pull/4057/commits/5096c4cc324bbd9642de7070cb87cc672a7d3bc0
Warning added: https://github.com/easybuilders/easybuild-framework/pull/4057/commits/63f9a3d5c310064f35eb42a821ff76caee0951f8

easybuild/main.py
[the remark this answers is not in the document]
casparvl: Repeating the same EasyConfig with different options is now supported, since we keep the EasyConfig name and EasyConfig specific options tegether in a tuple. Implemented in https://github.com/easybuilders/easybuild-framework/pull/4057/commits/7e2e6d60bc028033c6fe587b49901781e994950c

easybuild/main.py
casparvl: Done in https://github.com/easybuilders/easybuild-framework/pull/4057/commits/7e2e6d60bc028033c6fe587b49901781e994950c

test/framework/options.py
casparvl: Done in https://github.com/easybuilders/easybuild-framework/pull/4057/commits/cd8bb71c2432350943711c01b846f25217a81f0a

easybuild/main.py
```diff
@@ -213,162 +214,109 @@ def clean_exit(logfile, tmpdir, testing, silent=False):
...
-    (build_specs, _log, logfile, robot_path, search_query, eb_tmpdir, try_to_generate,
-     from_pr_list, tweaked_ecs_paths) = cfg_settings
-
-    # load hook implementations (if any)
-    hooks = load_hooks(options.hooks)
 
-    run_hook(START, hooks)
-
-    if modtool is None:
+    # TODO: insert fast loop that validates if all command line options are valid. If there are errors in options,
```
boegel, line 233: I opened an issue to follow up on this aspect: #4103

casparvl:
Todo, update documentation: https://github.com/easybuilders/easybuild/blob/develop/docs/Easystack-files.rst

boegel:
Merging #3749 caused a merge conflict here, which I've fixed just now.

I'll propose some tweaks to the docs update in https://github.com/easybuilders/easybuild/pull/832, so we can get this merged...

boegel (approved):

Issue #4056, opened 2022-08-10
Use of ifx in intel toolchains


hattom:
Since oneAPI, Intel ships 2 sets of compilers:
Classic: `icc`, `icpc`, `ifort`, and LLVM-based: `icx`, `icpx`, `ifx`.

Unlike `icx` and `icpx`, `ifx` is barely production-ready, still lagging behind `ifort` (it's finally out of beta, but that was not so long ago).

Two issues stand-out:
1) MPI: `mpiifort -fc=ifx` is the correct MPI command according to [1]
2) compatibility: presumably, fortran `.mod` files are not compatible between `ifort` and `ifx`. If this is indeed the case, does a separate toolchain make more sense than having a toolchain option?

(Related to https://github.com/easybuilders/easybuild-framework/issues/4009, but specifically relates to Fortran matters).

[1] https://www.intel.com/content/www/us/en/develop/documentation/mpi-developer-reference-linux/top/command-reference/compiler-commands/compilation-command-options.html

hattom:
"2." seems to not be a problem, `.mod` files seem compatible.

hattom:
For "1.", `https://github.com/easybuilders/easybuild-framework/pull/4032/files` should probably be updated, also for the mpi equivalents of `icx` and `icpx`: `mpiicc -cc=icx` and `mpiicpc -cxx=icpx` (assuming not DPC++).

hattom:
We can close. As @boegel pointed out on [#4062, ](https://github.com/easybuilders/easybuild-framework/pull/4062), this is already handled by env variables.

PR #4062, opened 2022-08-17
Modify MPI wrappers for oneAPI

hattom:
Not sure if this is these are the correct variables, and doubly bad if there's a `which $MPIF90` or similar being run later.

### DIFF
--- a/easybuild/toolchains/compiler/intel_compilers.py
+++ b/easybuild/toolchains/compiler/intel_compilers.py
@@ -69,6 +69,11 @@ class IntelCompilers(IntelIccIfort):
             self.COMPILER_F77 = 'ifx'
             self.COMPILER_F90 = 'ifx'
             self.COMPILER_FC = 'ifx'
+            self.COMPILER_MPICXX = 'mpiicpc -cxx=icpx'
+            self.COMPILER_MPICC = 'mpiicc -cc=icx'
+            self.COMPILER_MPIF77 = 'mpiifort -fc=ifx'
+            self.COMPILER_MPIF90 = 'mpiifort -fc=ifx'
+            self.COMPILER_MPIFC = 'mpiifort -fc=ifx'
             # fp-model source is not supported by icx but is equivalent to precise
             self.options.options_map['defaultprec'] = ['fp-speculation=safe', 'fp-model precise']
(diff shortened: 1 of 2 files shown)


hattom:
Not sure if the wrappers should be in here instead, but if so, I wouldn't know how to check for oneAPI there.
`https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/toolchains/mpi/intelmpi.py`

hattom:
Relates to Issue #4056

easybuild/toolchains/compiler/intel_compilers.py
```diff
@@ -69,6 +69,11 @@ def set_variables(self):
             self.COMPILER_F77 = 'ifx'
             self.COMPILER_F90 = 'ifx'
             self.COMPILER_FC = 'ifx'
+            self.COMPILER_MPICXX = 'mpiicpc -cxx=icpx'
```
boegel, line 72: @hattom This is actually not needed, I think.

EasyBuild already sets `$I_MPI_CC` & co to control which compiler is used by the MPI wrapper commands like `mpiicc`, and that seems to be working fine:

```
$ mpiicc --version
icc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '-diag-disable=10441' to disable this message.
icc (ICC) 2021.7.0 20220726
Copyright (C) 1985-2022 Intel Corporation.  All rights reserved.
```

```
$ I_MPI_CC=icx mpiicc --version
Intel(R) oneAPI DPC++/C++ Compiler 2022.2.0 (2022.2.0.20220730)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /software/intel-compilers/2022.2.0/compiler/2022.2.0/linux/bin-llvm
Configuration file: /software/intel-compilers/2022.2.0/compiler/2022.2.0/linux/bin/icx.cfg
```

hattom: Looks good. Let's close this PR.

PR #4064, opened 2022-08-22
update CI workflows to use Ubuntu 20.04 (since Ubuntu 18.04 is deprecated)

boegel:
see also https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/.github/workflows/container_tests.yml
+++ b/.github/workflows/container_tests.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 jobs:
   build:
     # stick to Ubuntu 18.04, where we can still easily install yum via 'apt-get install'
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


Micket (approved):
lgtm

.github/workflows/container_tests.yml
```diff
@@ -24,7 +23,7 @@ jobs:
         sudo apt-get update -yqq
         # for building Singularity images
         sudo apt-get install rpm
-        sudo apt-get install yum
+        sudo apt-get install dnf
```
boegel, line 26: We should also replace `apt-get` with `apt` while we're at it...

Micket (dismissed):
lgtm

.github/workflows/container_tests.yml
branfosj: Is there a follow up action to do this?

boegel: Fixed in 264b31f21

branfosj (approved):

PR #4065, opened 2022-08-23
Print potential errors after running a command only once

Flamefire:
`parse_cmd_output` calls `parse_log_for_error` which outputs potential
errors by default via `log.info`.
The former then outputs those via `log.warning` again.
Fix this by passing `stdout=False` to the latter.
Also minor rewording and variable renaming.

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -623,19 +623,19 @@ def parse_cmd_output(cmd, stdouterr, ec, simple, log_all, log_ok, regexp):
     """
     if strictness == IGNORE:
         check_ec = False
-        use_regexp = False
+        fail_on_error_match = False
     elif strictness == WARN:
         check_ec = True
-        use_regexp = False
+        fail_on_error_match = False
     elif strictness == ERROR:
         check_ec = True
-        use_regexp = True
+        fail_on_error_match = True
     else:
         raise EasyBuildError("invalid strictness setting: %s", strictness)
 
     # allow for overriding the regexp setting
     if not regexp:
-        use_regexp = False
+        fail_on_error_match = False
 
     if ec and (log_all or log_ok):
         # We don't want to error if the user doesn't care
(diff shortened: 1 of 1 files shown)


easybuild/tools/run.py
```diff
@@ -650,14 +650,16 @@ def parse_cmd_output(cmd, stdouterr, ec, simple, log_all, log_ok, regexp):
...
-            message = "Found %s errors in command output (output: %s)" % (len(res), "\n\t".join([r[0] for r in res]))
-            if use_regexp:
-                raise EasyBuildError(message)
+    if fail_on_error_match or regexp:
+        res = parse_log_for_error(stdouterr, regexp, stdout=False)
+        if res:
+            errors = "\n\t" + "\n\t".join([r[0] for r in res])
+            error_str = "error" if len(res) == 1 else "errors"
+            if fail_on_error_match:
+                raise EasyBuildError("Found %s %s in output of %s:%s", len(res), error_str, cmd, errors)
```
smoors, line 659: having only a colon between `cmd` and `errors` makes it difficult to visually detect the separation between the 2.
maybe use something like ` :: ` this instead? (a double colon surrounded by spaces does not occur often and is easy to search for)
```suggestion
                raise EasyBuildError("Found %s %s in output of %s :: %s", len(res), error_str, cmd, errors)
```

Flamefire: `errors` starts with a newline (see line 656) so sample output would be:
> ```
> Found 1 error in output of bar:
>    bar is not found
> ```

I think that newline is enough to separate, isn't it?

smoors: ah, yes indeed.

smoors (approved):
lgtm

Flamefire:
@smoors Could you merge those as the only failing checks are the now-removed bootstrap tests? Or shall I rebase to rerun the CI?

smoors:
> @smoors Could you merge those as the only failing checks are the now-removed bootstrap tests? Or shall I rebase to rerun the CI?

i cannot merge before all tests are green.
just restarted the 2 failing tests. if those fail again, then better to rebase indeed.

Flamefire:
They will fail. I rebased.

smoors:
Going in, thanks @Flamefire!

PR #4066, opened 2022-08-23
fix leaked handles in set_columns, complete_cmd, run_cmd_qa, det_terminal_size functions + tests

Flamefire:
`os.popen` is used without closing the returned handle.
As this function is deprecated anyway replace it by `subprocess_popen_text`.

Also fix some more leaked handles from opened processes and files, extracted from #3790


Flamefire:
@boegel I extracted some very similar commits from https://github.com/easybuilders/easybuild-framework/pull/3790 to this PR to make the other smaller.

boegel (approved):
lgtm

PR #4067, opened 2022-08-23
Fix `run_cmd` error handling

Flamefire:
Don't raise an error when `log_ok=False` is passed, i.e. the caller wants to handle the error.

The naming is confusing and the dependency on the strictness level is likely wrong. E.g. the caller may not expect a failing command to return from this function but to raise an error instead. However the user might have set the strictness to IGNORE which defeats the purpose of `log_ok`.

This change now allows to use `log_all=True, log_ok=False` to log the command and output and handle potential errors at the call site.


easybuild/tools/run.py
```diff
@@ -614,7 +614,13 @@ def parse_cmd_output(cmd, stdouterr, ec, simple, log_all, log_ok, regexp):
...
     if ec and (log_all or log_ok):
         # We don't want to error if the user doesn't care
-        if check_ec:
+        # Do not raise an error if we only want to log the command and output (i.e. log_all=True, log_ok=False)
+        # TODO: Refactor this confusing naming of parameters and dependency on strictness with EasyBuild 5.0
+        #       We need a way for the caller to:
+        #           - Never raise an error (e.g. caller handles that)
+        #           - Independently of the exit code log command and output
+        #           - Assure the command completed with no error (i.e. even for strictness=IGNORE)
+        if check_ec and log_ok:
```
boegel, line 623: I understand the current behavior of `run_cmd` is bloody confusing, and the meaning of the different parameters is unclear at best, but I'm not sure changes like this are a good idea. It's very difficult to assess the impact of this, I think.

I think we need a new `run*` function that properly supports the different use cases, like:
- just run a shell command, raise an error if it fails (based on exit code) -- default
- run a shell command, return output (stderr+stdout combined, or optionally split) and/or exit code

Logging behavior should probably always be the same (always log the output + exit code, no matter the use case?).

We can start slow, by implementing a new `run*` function that only supports the most common use case (or whichever one you have a need for), and gradually extend it's capabilities. Once we have a fully fleshed out new shiny `run*` implementation, we can deprecate `run_cmd` (and `run_cmd_qa` command) and make sure we use the new `run*` everywhere, both in framework itself and in easyblocks (that's perhaps a change to only make in EasyBuild v5.0, to avoid surprises).

Flamefire: I agree that a rewrite is overdue, but this is IMO a bugfix. See the description: In some places we expect that with `log_ok` set to False this will NOT raise an Exception.

The lengthy comment serves as a guidance for the future run-function

PR #4068, opened 2022-08-23
fix `quote_str` when string with both ' and " ends with a double quote

Flamefire:
When a string will be quoted with triple double quotes it must not end
in an unescaped double quote or the value will fail to parse as
4 double quotes are not recognized as the end-token for the string.
Handle that case by escaping the trailing double quote.

See CI failures in https://github.com/easybuilders/easybuild-easyconfigs/pull/15967 for where this happens or the new testcases which serve as examples.


boegel (approved):
lgtm

PR #4035, opened 2022-06-29
add support for `start_dir` easyconfig template

jfgrimm:
(created using `eb --new-pr`)

PR #4073, opened 2022-09-07
add support for start_dir easyconfig template

Flamefire:
As #4035 went stale with some issues remaining I tackled this. The test was quite tricky as this template can only be set once the source is extracted.
Hence I wrote the test so that it should be set in the configure and build step of a "regular" EasyConfig and in an extension which is where this will most likely be used.

The extension is very tricky:
- We must not resolve the `start_dir` in the `options` during `__init__` -> Move the relevant code down to the `Extension` class and clear the (new) template too
- The template value is then later set in a method I introduced in #3426 

@jfgrimm Please have a look here.


jfgrimm:
@Flamefire sorry for not looking at this sooner; mind fixing the conflict?

jfgrimm (approved):
lgtm

jfgrimm:
Going in, thanks @Flamefire!

Issue #3996, opened 2022-04-20
`--try-toolchain` still complains about missing dependencies


eburgueno:
I am trying to install R-4.1.0 using the latest foss toolchain (2021b). This is primarily because I already installed R-4.1.2 and I want to reuse dependencies as much as possible. Currently, EasyBuild complains that some dependencies are missing, even though presumably `--try-toolchain` should attempt to build them recursively if `--robot` is used (shouldn't it?):

```
$ eb -r --try-toolchain foss,2021b -D R-4.1.0-foss-2021a.eb
== Temporary log file in case of crash /tmp/eb-e0fwy6sw/easybuild-2x8c1uzi.log
== found valid index for /software/EasyBuild/4.5.4/easybuild/easyconfigs, so using it...
== found valid index for /software/EasyBuild/4.5.4/easybuild/easyconfigs, so using it...
ERROR: Missing dependencies: Python/3.9.5-GCCcore-11.2.0, Autotools/20210128-GCCcore-11.2.0, SQLite/3.35.4-foss-2021b, expat/2.2.9-GCCcore-11.2.0, Python/3.9.5-GCCcore-11.2.0-bare, CMake/3.20.1-GCCcore-11.2.0, cURL/7.76.0-foss-2021b, FFTW/3.3.9-foss-2021b, Python/3.9.5-foss-2021b, expat/2.2.9-foss-2021b, Autotools/20210128-gompi-2021b, CMake/3.20.1-gompi-2021b, cURL/7.76.0-gompi-2021b, SQLite/3.35.4-GCCcore-11.2.0, cURL/7.76.0-GCCcore-11.2.0 (no easyconfig file or existing module found)
```

Additional information:
```
$ eb --show-config
#
# Current EasyBuild configuration
# (C: command line argument, D: default value, E: environment variable, F: configuration file)
#
allow-modules-tool-mismatch (F) = True
buildpath                   (F) = /dev/shm
configfiles                 (E) = /software/powerPlant/etc/easybuild.cfg
containerpath               (D) = /home/user1/.local/easybuild/containers
detect-loaded-modules       (F) = purge
installpath                 (D) = /home/user1/.local/easybuild
installpath-modules         (F) = /software/powerPlant/modulefiles
installpath-software        (F) = /software
job-cores                   (F) = 4
module-syntax               (F) = Tcl
modules-tool                (F) = EnvironmentModules
repositorypath              (D) = /home/user1/.local/easybuild/ebfiles_repo
[...]

PR #4074, opened 2022-09-08
replaced distutils LooseVersion with packaging.version.parse

### LINKED ISSUE #4072: [bug|feature] Fixing version parsing of easyconfigs in `tweak.py`
### Context
I am running eb/4.6.0, and I was trying to update [OpenCV-4.5.5-foss-2021b-contrib.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.5-foss-2021b-contrib.eb) to `v4.6.0` on `foss/2022a` with updated deps, i.e.:
 ```
eb OpenCV-4.5.5-foss-2021b-contrib.eb --try-software-version=4.6.0 --try-toolchain-version=2022a --try-update-deps -Dr
```
But I quickly run into the following error:
```
    "__main__", mod_spec)
    exec(code, run_globals)
  File "/sw/noarch/Centos8/2021/software/EasyBuild/4.6.0/lib/python3.6/site-packages/easybuild/main.py", line 597, in <module>
    main()
  File "/sw/noarch/Centos8/2021/software/EasyBuild/4.6.0/lib/python3.6/site-packages/easybuild/main.py", line 440, in main
    easyconfigs = tweak(easyconfigs, build_specs, modtool, targetdirs=tweaked_ecs_paths)
  File "/sw/noarch/Centos8/2021/software/EasyBuild/4.6.0/lib/python3.6/site-packages/easybuild/framework/easyconfig/tweak.py", line 203, in tweak
    ignore_versionsuffixes=ignore_versionsuffixes)
  File "/sw/noarch/Centos8/2021/software/EasyBuild/4.6.0/lib/python3.6/site-packages/easybuild/framework/easyconfig/tweak.py", line 1064, in map_easyconfig_to_target_tc_hierarchy
    ignore_versionsuffixes=ignore_versionsuffixes
  File "/sw/noarch/Centos8/2021/software/EasyBuild/4.6.0/lib/python3.6/site-packages/easybuild/framework/easyconfig/tweak.py", line 1223, in find_potential_version_mappings
    LooseVersion(version) > LooseVersion(highest_version_ignoring_versionsuffix):
  File "/usr/lib64/python3.6/distutils/version.py", line 64, in __gt__
    c = self._cmp(other)
  File "/usr/lib64/python3.6/distutils/version.py", line 337, in _cmp
TypeError: '<' not supported between instances of 'int' and 'str'
```
I'm clearly not the first/only one who ran into this issue: #3996 

As I will show, the reason for the above error is two-fold:
1. EasyConfig (ec) version string substitution doesn't work in the regex which is used to retrieve the software version in `tweak.py`.
1. Correctly substituted strings still are not comparable 

### Issue 1. 
For a few ec's, version string substitution is used, e.g. [Java-1.8.0_292-OpenJDK.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/j/Java/Java-1.8.0_292-OpenJDK.eb).
The problem is that the un-substituted version string is obtained from `fetch_parameters_from_easyconfig`, which simply regex-es the ec for the version.
https://github.com/easybuilders/easybuild-framework/blob/b6daaacf08ac0b6ebec6c191f4359e2dcbfff5a0/easybuild/framework/easyconfig/tweak.py#L1213-L1214
For the above `Java-1.8.0_292-OpenJDK.eb`, the obtained version is:
`"'1.%s.0_%s' % (local_java_version, local_patch_version)"`

#### Proposed fix:
We could replace this function call with calling `process_easyconfig`, which actually does perform substitution:
```
# Replace the below:
# version, newversionsuffix = fetch_parameters_from_easyconfig(read_file(path), ['versionsuffix'])
#
ec = process_easyconfig(path)[0]['ec']
version, newversionsuffix = (
    ec.get('version', None),
    ec.get('versionsuffix', None)
)
```

- distutils is deprecated
  - https://peps.python.org/pep-0440/
  - packaging 20.9 for python == 2.7, 3.5
  - packaging 21.3 is the latest release

```python
$ python
Python 3.10.4 (main, Jul 13 2022, 12:16:05) [GCC 11.3.0] on linux
>>> from distutils.version import LooseVersion
>>> from packaging.version import parse
>>> 
>>> v1, v2 = "Java-1.8_265", "Java-1.8.0_292"
>>>
>>> LooseVersion(v1) < LooseVersion(v2)
<stdin>:1: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
  File "/home/robertsc/.local/easybuild/Centos8/2021/software/Python/3.10.4-GCCcore-11.3.0/lib/python3.10/site-packages/setuptools/_distutils/version.py", line 72, in __lt__
    c = self._cmp(other)
  File "/home/robertsc/.local/easybuild/Centos8/2021/software/Python/3.10.4-GCCcore-11.3.0/lib/python3.10/site-packages/setuptools/_distutils/version.py", line 357, in _cmp
>>>
>>> parse(v1) < parse(v2)
True
>>> 
```

Roadblocks:
- Requires an new python dependency (`packaging`, which is included in `setuptools`)

[Python/3.10.4-GCCcore-11.3.0](https://github.com/easybuilders/easybuild-framework/issues/3996)

robogast:
Fixes #4072, and #3996 partially.

This is my first PR for easybuild-framework, help is appreciated :)

### DIFF
--- a/easybuild/framework/easyconfig/tweak.py
+++ b/easybuild/framework/easyconfig/tweak.py
@@ -41,7 +41,7 @@ import glob
 import os
 import re
 import tempfile
-from distutils.version import LooseVersion
+from packaging.version import parse
 
 from easybuild.base import fancylogger
 from easybuild.framework.easyconfig.constants import EASYCONFIG_CONSTANTS
@@ -433,12 +433,12 @@ def pick_version(req_ver, avail_vers):
         if len(avail_vers) == 1:
             selected_ver = avail_vers[0]
         else:
-            retained_vers = [v for v in avail_vers if LooseVersion(v) <= LooseVersion(ver)]
+            retained_vers = [v for v in avail_vers if parse(v) <= parse(ver)]
             if retained_vers:
-                selected_ver = sorted(retained_vers, key=LooseVersion)[-1]
+                selected_ver = sorted(retained_vers, key=parse)[-1]
             else:
                 # if no versions are available that are less recent, take the least recent version
-                selected_ver = sorted(avail_vers, key=LooseVersion)[0]
+                selected_ver = sorted(avail_vers, key=parse)[0]
     else:
         # if no desired version is specified, just use last version
         ver = avail_vers[-1]
@@ -583,7 +583,7 @@ def select_or_generate_ec(fp, paths, specs):
 
     # TOOLCHAIN VERSION
     tcvers = unique([x[0]['toolchain']['version'] for x in ecs_and_files if x[0]['toolchain']['version']],
-                    sortkey=LooseVersion)
+                    sortkey=parse)
     _log.debug("Found %d unique toolchain versions: %s" % (len(tcvers), tcvers))
 
     tcver = specs.pop('toolchain_version', None)
@@ -606,7 +606,7 @@ def select_or_generate_ec(fp, paths, specs):
 
     # SOFTWARE VERSION
 
-    vers = unique([x[0]['version'] for x in ecs_and_files if x[0]['version']], sortkey=LooseVersion)
+    vers = unique([x[0]['version'] for x in ecs_and_files if x[0]['version']], sortkey=parse)
 
     _log.debug("Found %d unique software versions: %s" % (len(vers), vers))
 
@@ -925,7 +925,7 @@ def map_common_versionsuffixes(software_name, original_toolchain, toolchain_mapp
             version_matches = find_potential_version_mappings(software_as_dep, toolchain_mapping, quiet=True)
             if version_matches:
                 target_version = version_matches[0]['version']
-                if LooseVersion(target_version) > LooseVersion(version):
+                if parse(target_version) > parse(version):
                     original_suffix = '-%s-%s' % (software_name, version)
                     mapped_suffix = '-%s-%s' % (software_name, target_version)
                     # Make sure mapping is unique
@@ -1068,7 +1068,7 @@ def map_easyconfig_to_target_tc_hierarchy(ec_spec, toolchain_mapping, targetdir=
                 if potential_version_mappings:
                     highest_version_match = potential_version_mappings[0]['version']
                     highest_versionsuffix_match = potential_version_mappings[0]['versionsuffix']
-                    if LooseVersion(highest_version_match) > LooseVersion(dep['version']):
+                    if parse(highest_version_match) > parse(dep['version']):
                         _log.info("Updating version of %s dependency from %s to %s", dep['name'], dep['version'],
                                   highest_version_match)
                         _log.info("Depending on your configuration, this will be resolved with one of the following "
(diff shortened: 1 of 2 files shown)


Flamefire:
Adding an extra requirement is quite a lot to ask. I implemented a simple copy of `LooseVersion` in #3794 and is purposely simple.

And I mean from #4072

> comparing Java-1.8_265-b01 to Java-1.8.0_292, i.e. 1.8_265 to 1.8.0_292

What does it really mean to compare `1.8_265` to `1.8.0_292`? Without someone checking those shouldn't be (easily) comparable so an error in this case is good IMO.

robogast:
The pros/cons for the two ways of resolving the version comparison problem (feel free in adding/subtracting pros/cons from both lists, the following is just my opinion)
### `packaging.version.parse`:
#### Pros:
- Explicitly adheres to Python's extensive and well defined versioning scheme ([PEP-0440](https://peps.python.org/pep-0440/))
- Available for all Python versions (including 2.7)
- Is the advised way of replacing `disutils.version.LooseVersion` ([PEP-0632](https://peps.python.org/pep-0632/#migration-advice))
  - `packaging` is already included in `setuptools`
  - `distutils` (which is used in e.g. [`setup.py`](https://github.com/easybuilders/easybuild-framework/blob/develop/setup.py)) is deprecated and needs to be replaced by `setuptools` anyway by Python 3.12
  - PEP-0632 is also linked in the thread from which @Flamefire takes the fix for LooseVersion's str/int comparison issue #3996.

#### Cons:
- Adds a requirement
- Makes a (possibly wrong) decision in ambiguous cases such as comparing Java-1.8_265-b01 to Java-1.8.0_292, i.e. 1.8_265 to 1.8.0_292

### `easybuild-framework.tools.LooseVersion`
#### Pros:
- No additional requirement
- "Simple" implementation
#### Cons:
- Doesn't adhere to any standard(ized) versioning scheme
- Re-invents the wheel
- Doesn't handle complex cases (automatically)

I understand why the ambiguous case of comparing `1.8_265` to `1.8.0_292` is unfortunate, but breaking tools such as `tweak.py` in favor of manual decision making is (in my opinion) not a solution.

Micket:
We need to also consider the easyblocks and the many many places they use LooseVersion (and not only the ones we have, but also whatever people are using outside of the main repo). SI suspect noone is volunteering to changing all of those to parse and checking that nothing broke, so a switch to parse is something i think we have to introduce gradually there, and we regardless need to provide the easybuild-framework.tools.LooseVersion during that transition.

boegel:
My 2 cents on this: while it seems like we may not be able to avoid a dependency on `setuptools` from Python 3.12 onwards (to be checked, I still hope we can avoid that since we know it's going to be a PITA if not), my preference goes to "vendoring" `LooseVersion` as is done in #3794.

Not only does it avoid introducing a dependency on a Python package outside the standard library, but it also means that nothing really changes, so there won't be surprises with version checking suddenly being done slightly differently.

We have a broad set of test cases in the easyconfigs repository that we can test with, there are other interesting cases like OpenFOAM, LAMMPS, etc. which we should also check.

That said, I think whatever road we take, this is EasyBuild 5.0 material, since the potential impact is not small (and Python 3.12 is not available yet).

Flamefire:
> That said, I think whatever road we take, this is EasyBuild 5.0 material, since the potential impact is not small (and Python 3.12 is not available yet).

I don't see how #3794 is a breaking change. It is a more or less 1:1 copy of the existing LooseVersion with a corner case fixed which didn't work before simplifying code that wants to use it. I would suggest to merge that and transition to it soon so that we can test it extensively before the next release.

boegel:
> > That said, I think whatever road we take, this is EasyBuild 5.0 material, since the potential impact is not small (and Python 3.12 is not available yet).
> 
> I don't see how #3794 is a breaking change. It is a more or less 1:1 copy of the existing LooseVersion with a corner case fixed which didn't work before simplifying code that wants to use it. I would suggest to merge that and transition to it soon so that we can test it extensively before the next release.

I didn't call #3794 a breaking change, but it definitely is an intrusive one.
In theory, it's the same code (except, it's not really, since there's a bugfix on top).

We've discussed #3794 briefly during the EasyBuild conf call (again) today, and the general consensus is that we should just go ahead with it (without waiting for EasyBuild v5.0).

boegel:
Closing this since #3794 has been merged

PR #4076, opened 2022-09-11
make check_sha256_checksums verify all checksums if they're specified as a dict value

boegel:
https://github.com/easybuilders/easybuild-easyconfigs/pull/16234 showed that `check_sha256_checksums` overlooks faulty SHA256 checksums, this PR fixes that by checking *all* checksum values if a checksum is specified as a dict value, not just the checksum one of the possible source filenames (so in the case of CUDA, only for for the `x86_64` "source" file).


branfosj (approved):
lgtm

PR #4079, opened 2022-09-15
add support for `--extra-source-urls` to fetch sources from additional URLs

joeydumont:
This PR adds a --sources-url CLI option. It is a comma separated list of URLs that EasyBuild will fetch sources from. It replaces the hard-coded `EASYBUILD_SOURCES_URL`, but keeps it as a default value. The --sources-url-priority option is used to determine if these URLs are to be queried first, or last, i.e. as a fallback.

Unfortunately, this breaks 3 tests related to downloading the file, and I'm unsure of exactly why. It seems that `build_option('sources_url')` might return None when the function is run through the test harness, rather than getting populated with the default value. I've tried adding a `init_config(args=["--sources-url=https://sources.easybuild.io"])`, but that didn't work either.

If the feature is of interest, could you give me a hand in fixing the tests?

EDIT: After some discussion on Slack, I was able to fix the tests. I also renamed the options to `--extra-source-urls` and `--extra-source-urls-priority`. I changed the commit mesage to be imperative.


easybuild/tools/options.py
```diff
@@ -394,6 +396,11 @@ def override_options(self):
                              None, 'store_true', False),
             'extra-modules': ("List of extra modules to load after setting up the build environment",
                               'strlist', 'extend', None),
+            "extra-source-urls": ("Specify different URLs to fetch sources from", "strlist", "store",
```
Flamefire, line 399:
```suggestion
            "extra-source-urls": ("Specify URLs to fetch sources from in addition to those in the easyconfig", "strlist", "store",
```

easybuild/tools/options.py
```diff
@@ -394,6 +396,11 @@ def override_options(self):
                              None, 'store_true', False),
             'extra-modules': ("List of extra modules to load after setting up the build environment",
                               'strlist', 'extend', None),
+            "extra-source-urls": ("Specify URLs to fetch sources from in addition to those in the easyconfig",
+                                  "strlist", "store", DEFAULT_EXTRA_SOURCE_URLS),
+            "extra-source-urls-priority": ("Specify whether sources-url should be tried before or after the "
```
jfgrimm, line 401:
```suggestion
            "extra-source-urls-priority": ("Specify whether extra-source-urls should be tried before or after the "
```

valandil: Pushed that.

joeydumont: Pushed that.

joeydumont:
I see there are now merge conflicts for this. Let me know when you're ready for me to rebase, or if you prefer to handle that.

jfgrimm:
> I see there are now merge conflicts for this. Let me know when you're ready for me to rebase, or if you prefer to handle that.

go for it; if you have any issues/questions just ping me

Flamefire:
Suggestion here:
`--sources-url` supports multiple URLs, doesn't it? How are they separated?

For EasyConfigs we have --robot-paths which accepts a list of paths like  /path1:/path2 and works nicely mit the default paths if you append or prepend a colon: :/path1:/path2 appends it to the default, /path1:/path2: prepends it while the original replaces it.
How about replacing the priority option by something like this to have a uniform interface here?

joeydumont:
As suggested, I made sure that this used the same `add_flex` logic that `--robot-paths` uses. However, I think URLs should be separated by `|` rather than a colon (`:`) or a comma (,), as both are valid in URLs. I added `urllist` to the list of options in `ExtOption` to allow for this separator to be used.

easybuild/framework/easyblock.py
```diff
@@ -897,8 +895,13 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
                     source_urls = []
                 source_urls.extend(self.cfg['source_urls'])
 
-                # add https://sources.easybuild.io as fallback source URL
-                source_urls.append(EASYBUILD_SOURCES_URL + '/' + os.path.join(name_letter, location))
+                # add extra-source-urls CLI as either a first check, or a fallback.
+                self.log.warning("[extra_source_urls] %s", build_option('extra_source_urls'))
```
Flamefire, line 899: Why are those warnings? I'd suggest to only log if there are any at all, as (at most) `info` and the individual urls as `debug`. And maybe only log the final value, not each intermediate

easybuild/framework/easyblock.py
```diff
@@ -897,8 +895,13 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
                     source_urls = []
                 source_urls.extend(self.cfg['source_urls'])
 
-                # add https://sources.easybuild.io as fallback source URL
-                source_urls.append(EASYBUILD_SOURCES_URL + '/' + os.path.join(name_letter, location))
+                # add extra-source-urls CLI as either a first check, or a fallback.
```
Flamefire, line 898: The comment seems outdated. I don't see an either-or in the code below. Also CLI should rather be "option" or "command line option" as CLI is usually the whole set of (all) CL options

easybuild/framework/easyblock.py
```diff
@@ -897,8 +895,13 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
...
+                self.log.warning("[extra_source_urls] %s", build_option('extra_source_urls'))
+                for url in build_option("extra_source_urls"):
+                    url += "/" + name_letter + "/" + location
+                    self.log.warning("[extra_source_urls] url is %s", url)
+                    source_urls.extend([url])
```
Flamefire, line 903: It's only a single element:
```suggestion
                    source_urls.append(url)
```

easybuild/tools/options.py
```diff
@@ -407,6 +409,8 @@ def override_options(self):
                              None, 'store_true', False),
             'extra-modules': ("List of extra modules to load after setting up the build environment",
                               'strlist', 'extend', None),
+            "extra-source-urls": ("Specify URLs to fetch sources from in addition to those in the easyconfig",
+                                  "urltuple", "add_flex", DEFAULT_EXTRA_SOURCE_URLS, {'metavar': 'URL[%sURL]' % '|'}),
```
Flamefire, line 413: I'd rather not use string interpolation for a constant. Just makes loading EB (even) slower:
```suggestion
                                  "urltuple", "add_flex", DEFAULT_EXTRA_SOURCE_URLS, {'metavar': 'URL[|URL]'}),
```

Flamefire (changes requested):

easybuild/tools/options.py
joeydumont: Good point, done.

easybuild/framework/easyblock.py
joeydumont: Done.

easybuild/framework/easyblock.py
joeydumont: Thanks, I added those in for me to make sure the values were making their way in the expected places. I removed these logs completely, the debug logs already print
```
== 2024-07-16 20:58:08,626 generaloption.py:1604 DEBUG generate_cmd_line adding extra-source-urls value ('https://sources.easybuild.io/',) default found. Not adding to args.
```

easybuild/framework/easyblock.py
joeydumont: It was, fixed.

easybuild/framework/easyblock.py
```diff
@@ -897,8 +895,11 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
                     source_urls = []
                 source_urls.extend(self.cfg['source_urls'])
 
-                # add https://sources.easybuild.io as fallback source URL
-                source_urls.append(EASYBUILD_SOURCES_URL + '/' + os.path.join(name_letter, location))
+                # Insert --extra-source-urls command line option to the
+                # urls to try to download from.
```
Flamefire, line 899: As mentioned earlier this is not only a CLI option, but can be from 2 other places. I suggest something like:

```suggestion
                # Add additional URLs as configured
```

easybuild/base/generaloption.py
```diff
@@ -182,6 +187,7 @@ class ExtOption(CompleterOption):
           - strlist, strtuple : convert comma-separated string in a list resp. tuple of strings
           - pathlist, pathtuple : using os.pathsep, convert pathsep-separated string in a list resp. tuple of strings
               - the path separator is OS-dependent
+          - urllist, urltuple: using '|', convert urlsep separated string in a list resp. tuple of strings
```
Flamefire, line 190: Nit if you want to touch that: Although it is a copy of the above there is no definition of "urlsep". So maybe:

```suggestion
          - urllist, urltuple: convert string separated by '|', to a list resp. tuple of strings
```

easybuild/framework/easyblock.py
```diff
@@ -897,8 +895,10 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
                     source_urls = []
                 source_urls.extend(self.cfg['source_urls'])
 
-                # add https://sources.easybuild.io as fallback source URL
-                source_urls.append(EASYBUILD_SOURCES_URL + '/' + os.path.join(name_letter, location))
+                # Add additional URLs as configured.
+                for url in build_option("extra_source_urls") or []:
```
Flamefire, line 899: Looks like f4ec1825c740410f667342caa3403a4e37657c33 is not required after you fixed the issue with 1f5542e586ad3661c56d7394786627f6661a5bc5, i.e. "extra_source_urls" can never be None hence no need for the `or []`

I'd suggest to either revert that commit or rebase skipping it. Otherwise people might get confused later how it could be anything but a (possibly) empty list. But it can't even be an **empty** list, can it?

Or:
```suggestion
                for url in build_option("extra_source_urls"):
```

easybuild/tools/config.py
```diff
@@ -120,6 +120,7 @@
 DEFAULT_PR_TARGET_ACCOUNT = 'easybuilders'
 DEFAULT_PREFIX = os.path.join(os.path.expanduser('~'), ".local", "easybuild")
 DEFAULT_REPOSITORY = 'FileRepository'
+DEFAULT_EXTRA_SOURCE_URLS = ('https://sources.easybuild.io/',)
```
Flamefire, line 123: The trailing slash lead to a duplication later (added in the loop using it, wasn't there before) on which might cause issues, see above: `url += "/" + name_letter + "/" + location`

```suggestion
DEFAULT_EXTRA_SOURCE_URLS = ('https://sources.easybuild.io',)
```

easybuild/tools/options.py
```diff
@@ -69,10 +69,12 @@
...
 from easybuild.tools.config import DEFAULT_MINIMAL_BUILD_ENV, DEFAULT_MNS, DEFAULT_MODULE_SYNTAX, DEFAULT_MODULES_TOOL
 from easybuild.tools.config import DEFAULT_MODULECLASSES, DEFAULT_PATH_SUBDIRS, DEFAULT_PKG_RELEASE, DEFAULT_PKG_TOOL
-from easybuild.tools.config import DEFAULT_PKG_TYPE, DEFAULT_PNS, DEFAULT_PREFIX, DEFAULT_PR_TARGET_ACCOUNT
+from easybuild.tools.config import DEFAULT_PKG_TYPE, DEFAULT_PNS, DEFAULT_PREFIX, DEFAULT_EXTRA_SOURCE_URLS
+from easybuild.tools.config import DEFAULT_PR_TARGET_ACCOUNT
 from easybuild.tools.config import DEFAULT_REPOSITORY, DEFAULT_WAIT_ON_LOCK_INTERVAL, DEFAULT_WAIT_ON_LOCK_LIMIT
 from easybuild.tools.config import DEFAULT_FILTER_RPATH_SANITY_LIBS
-from easybuild.tools.config import EBROOT_ENV_VAR_ACTIONS, ERROR, FORCE_DOWNLOAD_CHOICES, GENERAL_CLASS, IGNORE
+from easybuild.tools.config import EBROOT_ENV_VAR_ACTIONS, ERROR
+from easybuild.tools.config import FORCE_DOWNLOAD_CHOICES, GENERAL_CLASS, IGNORE
```
Flamefire, line 77: Why this seemingly unrelated change?
```suggestion
from easybuild.tools.config import EBROOT_ENV_VAR_ACTIONS, ERROR, FORCE_DOWNLOAD_CHOICES, GENERAL_CLASS, IGNORE
```

easybuild/base/generaloption.py
```diff
@@ -103,6 +103,11 @@ def what_str_list_tuple(name):
         sep = os.pathsep
         helpsep = 'pathsep'
```
Flamefire, line 105: Can you remove this empty line please? And as we have 3 cases now, I think it would be great to put the comma part above into an `else` clause below this (again without empty line between previous "helpset" and "else")

Flamefire (changes requested):
A final round of comments. All minor things, so it's just polishing this up. :+1:

easybuild/framework/easyblock.py
joeydumont: My understanding is that it cannot be an empty list now it has a properly listed default value in `BUILD_OPTIONS_CMDLINE`, indeed. I'll remove it.

You mentioned rebasing. Would you prefer I squash everything to one commit here? Or do you want to keep the review history?

easybuild/tools/options.py
joeydumont: I had additional imports in the previous implementation, and they were subsequently removed. I've cleaned that up.

easybuild/tools/options.py
```diff
@@ -69,9 +69,9 @@
 from easybuild.tools.config import DEFAULT_JOB_EB_CMD, DEFAULT_LOGFILE_FORMAT, DEFAULT_MAX_FAIL_RATIO_PERMS
 from easybuild.tools.config import DEFAULT_MINIMAL_BUILD_ENV, DEFAULT_MNS, DEFAULT_MODULE_SYNTAX, DEFAULT_MODULES_TOOL
 from easybuild.tools.config import DEFAULT_MODULECLASSES, DEFAULT_PATH_SUBDIRS, DEFAULT_PKG_RELEASE, DEFAULT_PKG_TOOL
-from easybuild.tools.config import DEFAULT_PKG_TYPE, DEFAULT_PNS, DEFAULT_PREFIX, DEFAULT_PR_TARGET_ACCOUNT
+from easybuild.tools.config import DEFAULT_PKG_TYPE, DEFAULT_PNS, DEFAULT_PREFIX, DEFAULT_EXTRA_SOURCE_URLS
 from easybuild.tools.config import DEFAULT_REPOSITORY, DEFAULT_WAIT_ON_LOCK_INTERVAL, DEFAULT_WAIT_ON_LOCK_LIMIT
-from easybuild.tools.config import DEFAULT_FILTER_RPATH_SANITY_LIBS
+from easybuild.tools.config import DEFAULT_FILTER_RPATH_SANITY_LIBS, DEFAULT_PR_TARGET_ACCOUNT
```
Flamefire, line 74: Just FYI: We try to keep the imports sorted. So the ordering here would be the other way round. But not really important

easybuild/framework/easyblock.py
Flamefire: That is correct, yes

I'm not allowed to merge anyway so I can't really decide it. But my person preference for this is to squash. Most commits are just stylistic anyway, not some redesign or something. Only thing would be https://github.com/easybuilders/easybuild-framework/pull/4079/commits/1f5542e586ad3661c56d7394786627f6661a5bc5 to keep a mention why it is put there but even that is "obvious" enough. Hence a single commit is fine for me.

But I know that others like to keep all history or just don't want to do the extra work of squashing.

Flamefire (approved):

easybuild/tools/config.py
```diff
@@ -120,6 +120,7 @@
 DEFAULT_PR_TARGET_ACCOUNT = 'easybuilders'
 DEFAULT_PREFIX = os.path.join(os.path.expanduser('~'), ".local", "easybuild")
 DEFAULT_REPOSITORY = 'FileRepository'
+DEFAULT_EXTRA_SOURCE_URLS = ('https://sources.easybuild.io',)
```
boegel, line 123:
```suggestion
EASYBUILD_SOURCES_URL = 'https://sources.easybuild.io'
DEFAULT_EXTRA_SOURCE_URLS = (EASYBUILD_SOURCES_URL,)

File: easybuild/framework/easyblock.py [hunk 15]
```diff
@@ -106,8 +106,6 @@
 from easybuild.tools.version import this_is_easybuild, VERBOSE_VERSION, VERSION
 
 
-EASYBUILD_SOURCES_URL = 'https://sources.easybuild.io'
```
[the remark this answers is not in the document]
joeydumont: We can indeed define it in `easybuild.tools.config`, but then importing it in `easybuild.framework.easyblock` and not using it makes houndbot complain. I defined it separately in two locations to fix houndbot, but it's not optimal.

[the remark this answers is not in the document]
boegel: @joeydumont Hmm, I would be surprised if the bot complains about a constant not being used in `easybuild.tools.config`, but if it does, you can add a comment to skip the style check for that line:

```python
EASYBUILD_SOURCES_URL = 'https://sources.easybuild.io'  # noqa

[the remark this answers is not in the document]
Comment (joeydumont) [reply easybuild/framework/easyblock.py:109]: I meant in `easybuild.framework.easyblock`, if we had
```
from easybuild.tools.config import EASYBUILD_SOURCES_URL
```
so that it doesn't break the API. That's why I've done now, but added `# noqa`. 

In short, now `EASYBUILD_SOURCES_URL` is defined in `easybuild.tools.config`, where it is used, and imported in `easybuild.framework.easyblock`, where it used to be defined, so it can still be imported as `from easybuild.framework.easyblock`, but defined in a single place. The `# noqa` is required for HoundBot not to complain about the unused variable in `easybuild.framework.easyblock`.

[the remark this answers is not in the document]
boegel: :+1:

boegel (approved):

boegel:
Good to go, thanks a lot for your patience with this one @joeydumont, I'm happy it's finally getting merged :)

joeydumont:
> Good to go, thanks a lot for your patience with this one @joeydumont, I'm happy it's finally getting merged :)

And thank you! I appreciate the feedback!

PR #4080, opened 2022-09-20
skip over unset $EB_PYTHON/$EB_INSTALLPYTHON

Flamefire:
Check if the current $python_cmd is empty and skip all checks in that case. This avoids useless checks and log output in VERBOSE mode.

Also extend the CI test to catch this issue.


branfosj (approved):
lgtm

PR #4084, opened 2022-09-28
Add GITHUB_RELEASE and GITHUB_LOWER_RELEASE templates

Flamefire:
We use  `https://github.com/%(github_account)s/%(name)s/releases/download/v%(version)s` and the lower variant a lot in ECs, so add templates


boegel (approved):
lgtm

PR #4085, opened 2022-09-30
Make inject_checksums inject dictionaries

ocaisa:
`--inject-checksums` currently adds a comment beside the sha to indicate which file matches which checksum,
```
checksums = [
    '91844808532e5ce316b3c010929493c0244f3d37593afd6de04f71821d5136d9',  # zlib-1.2.12.tar.gz
    'f35eb05334a4f8d7b40b6c5610a6369f654863b5fa1a19c2507888f918025238',  # zlib-1.2.12_fix-CC-logic-in-configure.patch
]
```
dictionaries are supported though so prefer those to make the mapping explicit (and verifiable)
```
checksums = [
    {'zlib-1.2.12.tar.gz': '91844808532e5ce316b3c010929493c0244f3d37593afd6de04f71821d5136d9'},
    {'zlib-1.2.12_fix-CC-logic-in-configure.patch': 'f35eb05334a4f8d7b40b6c5610a6369f654863b5fa1a19c2507888f918025238'},
]
```

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -4416,11 +4416,11 @@ def inject_checksums(ecs, checksum_type):
         line_indent = INDENT_4SPACES * indent_level
         checksum_lines = []
         for fn, checksum in checksums:
-            checksum_line = "%s'%s',  # %s" % (line_indent, checksum, fn)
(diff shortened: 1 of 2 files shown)


bartoldeman:
Good idea, a dict is much more explicit than a list with comments!

Issue #4086, opened 2022-09-30
We should set CUDAARCHS environment variable (plus additional cuda_cc_ template)


Micket:
Credit to @mboisson for this recommendation

Ref.
https://cmake.org/cmake/help/latest/envvar/CUDAARCHS.html
https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html
https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html#prop_tgt:CUDA_ARCHITECTURES

The CUDAARCHS support was added into CMake just to "to avoid specifying it for every invocation" 
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5533
but as an environment variable it does also help in the case of nested builds. It's also less CMake specific, so who knows, maybe other builds systems would pick up on it.
Heads up though; **this will change all the default property for all cmake build targets if set**. So extensive testing would be required.

Regardless of whether or not we add this variable, we are desperately missing the no-dot semicolon separated compute-capablities 
We currently have;
```
cuda_cc_semicolon_sep='7.0;8.6'
cuda_cc_space_sep='7.0 8.6'
cuda_compute_capabilities='7.0,8.6'
cuda_sm_comma_sep='sm_70,sm_86'
cuda_sm_space_sep='sm_70 sm_86'
```
and we need something
```
cuda_cc_cmake='70;86'
```
as well.
It's especially important because there is simply no way at all do deal with this trivial removal of a dot, or switching to a semicolon from within an easyconfig. The template isn't expanded when that code runs, and we can't access the `cuda_compute_capabilities` either, forcing us to patch in custom code into buildscripts that does our made up format.

Micket:
Had to learn a bit about cmake; this variable is only used if the cmake scripts are using the new mode for building; 
1. Using `enable_language(CUDA)` or `project(Foo LANGUAGES CUDA ...)`
2. Doesn't use the old `find_package(CUDA)` or any of the  `cuda_add_xxxx` macros it defines. It should just be a normal `add_...`. It may optionall use `find_package(CUDAToolkit)`  for linking to cuFFT etc.

I don't recall actually coming across any cmakelists that does this yet, but doesn't hurt to support this from the start, possibly patch stuff to do the right thing.

Flamefire:
To add to that from Slack:

> I see some older easyconfigs making this mistake
> ```
[...]

PR #4087, opened 2022-10-05
add `%(cuda_cc_cmake)s` template

Flamefire:
This can be used with CMake projects via the env variable `$CUDAARCHS` (CMake 3.20+) or the CMake variable `CMAKE_CUDA_ARCHITECTURES` (CMake 3.18+) (via `-DCMAKE_CUDA_ARCHITECTURES=...`)
 
See https://cmake.org/cmake/help/v3.18/variable/CMAKE_CUDA_ARCHITECTURES.html

See #4086

Example uses (redundant):

```python
preconfigopts = 'CUDAARCHS="%(cuda_cc_cmake)s"'
pconfigopts = '-DCMAKE_CUDA_ARCHITECTURES="%(cuda_cc_cmake)s"'
```

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -90,6 +90,7 @@ TEMPLATE_NAMES_DYNAMIC = [
     ('mpi_cmd_prefix', "Prefix command for running MPI programs (with default number of ranks)"),
     ('cuda_compute_capabilities', "Comma-separated list of CUDA compute capabilities, as specified via "
      "--cuda-compute-capabilities configuration option or via cuda_compute_capabilities easyconfig parameter"),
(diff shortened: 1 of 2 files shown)


Micket (approved):
lgtm

As it might also be useful for other build tools that happen to use this format, but i still think we should also either set CUDAARCHS or at least set CMAKE_CUDA_ARCHITECTURES automatically in the cmakemake block.

Flamefire:
> but i still think we should also either set CUDAARCHS or at least set CMAKE_CUDA_ARCHITECTURES automatically in the cmakemake block.

I think it is valuable to do that explicitly in the EC as either the EC might not expect it to be set or the software might not support it while the EC-author thinks it is used.

boegel (approved):
lgtm

PR #4088, opened 2022-10-05
Fix trying to generate RPATH wrappers for Clang

casparvl:
[This](https://github.com/easybuilders/easybuild-framework/blob/541a645ab9c8e6fc60345e896a0471a2f75f71e8/easybuild/tools/toolchain/toolchain.py#L990) loop is supposed to loop over all compilers for which wrappers need to be created. However, [this line](https://github.com/easybuilders/easybuild-framework/blob/541a645ab9c8e6fc60345e896a0471a2f75f71e8/easybuild/tools/toolchain/toolchain.py#L973) returns (['clang', 'clang++'],[None, None, None]) for the `Clang` toolchain. The subsequent call to `which` happening [here](https://github.com/easybuilders/easybuild-framework/blob/541a645ab9c8e6fc60345e896a0471a2f75f71e8/easybuild/tools/toolchain/toolchain.py#L991) then fails.

This PR fixes that by explicitely checking if `cmd` is `None`. If so, it proceeds to the next loop iteration and doens't try to create an RPATH wrapper for this 'None' command.

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -988,6 +988,9 @@ class Toolchain(object):
 
         # create wrappers
         for cmd in nub(c_comps + fortran_comps + ['ld', 'ld.gold', 'ld.bfd']):
+            # Not all toolchains have fortran compilers (e.g. Clang), in which case they are 'None'
+            if cmd is None:
+                continue
             orig_cmd = which(cmd)
 
             if orig_cmd:
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -2622,7 +2622,7 @@ class ToolchainTest(EnhancedTestCase):
         fake_clang = os.path.join(self.test_prefix, 'fake', 'clang')
         write_file(fake_clang, '#!/bin/bash\necho "$@"')
         adjust_permissions(fake_clang, stat.S_IXUSR)
-        tc_clang = self.get_toolchain('Clang', version='13.0.1')
+        tc_clang = self.get_toolchain('ClangGCC', version='0.0.0')
         tc_clang.prepare()
 
         # Check that the clang wrapper is indeed in place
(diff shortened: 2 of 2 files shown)


boegel:
@casparvl It would be nice if we could enhance the `test_toolchain_prepare_rpath` test to cover this fix.

Without this change, what happens? A hard crash, or RPATH linking just doesn't happen?

casparvl:
Yes, without this, a `prepare_rpath_wrappers()` call on a `Clang` toolchain will fail (some complaint about a `Nonetype` blabla since it expects a string)

casparvl:
@boegel is this more or less what you had in mind in terms of test? Can you review this PR, in conjuction with this EasyBlock PR https://github.com/easybuilders/easybuild-easyblocks/pull/2799 ?

boegel:
@casparvl That test is perfect: if I run it after temporarily undoing the fix in `easybuild/tools/toolchain/toolchain.py`, then it reproduces the problem being fixed:

```
ERROR: test_toolchain_prepare_rpath (test.framework.toolchain.ToolchainTest)
Test toolchain.prepare under --rpath
  File "/Volumes/work/easybuild-framework/test/framework/toolchain.py", line 2627, in test_toolchain_prepare_rpath
    tc_clang.prepare_rpath_wrappers()
  File "/Volumes/work/easybuild-framework/easybuild/tools/toolchain/toolchain.py", line 994, in prepare_rpath_wrappers
    orig_cmd = which(cmd)
  File "/Volumes/work/easybuild-framework/easybuild/tools/filetools.py", line 521, in which
    cmd_path = os.path.join(path, cmd)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/posixpath.py", line 90, in join
    genericpath._check_arg_types('join', a, *p)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/genericpath.py", line 152, in _check_arg_types
    raise TypeError(f'{funcname}() argument must be str, bytes, or '
TypeError: join() argument must be str, bytes, or os.PathLike object, not 'NoneType'
```

boegel (approved):
lgtm

PR #4089, opened 2022-10-10
update to v3 of actions/checkout and actions/setup-python in CI workflows

boegel:
In response to the following warning popping up when using v2:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-python, actions/setup-python, actions/checkout
```

See also https://github.com/actions/checkout/pull/689

### DIFF
--- a/.github/workflows/bootstrap_script.yml
+++ b/.github/workflows/bootstrap_script.yml
@@ -54,10 +54,10 @@ jobs:
             lc_all: C
       fail-fast: false
     steps:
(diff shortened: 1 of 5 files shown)


lexming (approved):
LGTM

lexming:
Going in, thanks @boegel !

PR #4094, opened 2022-10-12
use SYSTEM template constant in dependencies instead of True in framework tests

jfgrimm:
update to reflect https://github.com/easybuilders/easybuild-easyconfigs/issues/16330, https://github.com/easybuilders/easybuild-easyconfigs/pull/16384, https://github.com/easybuilders/easybuild-easyconfigs/pull/16386


boegel (approved):
lgtm

PR #4097, opened 2022-10-17
first implementation of --try-dependencies and --try-builddependencies

mboisson:
The goal of this PR is to let someone easily modify existing dependencies or add new ones without having to rewrite an EasyConfig. The use-case that triggered this is being able to recompile CUDA software with different versions of CUDA. 

It works like this: 
```
eb AutoDock-GPU-1.5.3-GCC-10.3.0-CUDA-11.3.1.eb --try-dependency="('CUDA','11.4')" 
``` 

Multiple dependencies can be updated by repeating the flag: 
```
eb AutoDock-GPU-1.5.3-GCC-10.3.0-CUDA-11.3.1.eb --try-dependency="('CUDA','11.4')" --try-dependency="('CMake', '3.23.1')"
``` 

If a dependency with a matching name is found (i.e. CUDA 11.3.1) then the dependency is updated with the value specified. If no dependency with a matching name is found, then the dependency is added to the list.


mboisson:
Hum, I don't understand why there's already a test with `dependencies` and `builddependencies` when no options for `--try-dependencies` and `--try-builddependencies` existed before... 🤔

boegel:
> Hum, I don't understand why there's already a test with `dependencies` and `builddependencies` when no options for `--try-dependencies` and `--try-builddependencies` existed before... 🤔

Maybe via `--try-amend`?

mboisson:
> 

maybe. I renamed the options to `dependency` rather than `dependencies` to work around the test. They can only take a single dependency at a time anyway...

ocaisa:
Does this act recursively with `--robot`?

mboisson:
> Does this act recursively with `--robot`?

no idea, I don't use `--robot` and I don't know how it works :)

PR #4107, opened 2022-10-22
also use EasyConfig instances cache in process_easyconfig when build_specs is empty dict

boegel:
`test_easyconfigs_caches` added in dbf4697 fails without fix in 5c99f85ab


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #4108, opened 2022-10-22
fix build options that should have empty list as default value

boegel:
For build options like `from_pr` we assume throughout the code that they have an empty list as default value, see for example: https://github.com/easybuilders/easybuild-framework/blob/95bae20337781a06defe467faf487a122aae8512/easybuild/framework/easyconfig/tools.py#L343

There are situations where that's currently not the case though, primarily when EasyBuild is being used as a library (cfr. [this tutorial section](https://easybuilders.github.io/easybuild-tutorial/2021-lust/easybuild_library)).
That's fixed by the changes being made here, by implementing special treatment for build options that should have a *unique* empty list as default value.


branfosj (changes requested):
Can we get the lists in alphabetical order while we are doing this please?

See https://github.com/boegel/easybuild-framework/pull/54

branfosj (approved):

branfosj:
I fixed the conflict caused by #3749 and the reordering of the lists.

branfosj (approved):

Issue #4111, opened 2022-10-25
error when using `--ignore-checksums` on an easyconfig where checksums are specified in a tuple


jfgrimm:
trying to build Stata 17 with yet another different checksum from those listed. When running `eb --ignore-checksums Stata-17.eb`, I get:
```
== unpacking...
== ... (took 7 secs)

ERROR: Traceback (most recent call last):
  File "/opt/apps/easybuild/software/EasyBuild/4.6.2/lib/python3.6/site-packages/easybuild/main.py", line 127, in build_and_install_software
    (ec_res['success'], app_log, err) = build_and_install_one(ec, init_env)
  File "/opt/apps/easybuild/software/EasyBuild/4.6.2/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 4076, in build_and_install_one
    result = app.run_all_steps(run_test_cases=run_test_cases)
  File "/opt/apps/easybuild/software/EasyBuild/4.6.2/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 3959, in run_all_steps
    self.run_step(step_name, step_methods)
  File "/opt/apps/easybuild/software/EasyBuild/4.6.2/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 3794, in run_step
    step_method(self)()
  File "/opt/apps/easybuild/software/EasyBuild/4.6.2/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 2352, in checksum_step
    if verify_checksum(fil['path'], fil['checksum']):
  File "/opt/apps/easybuild/software/EasyBuild/4.6.2/lib/python3.6/site-packages/easybuild/tools/filetools.py", line 1307, in verify_checksum
    actual_checksum = compute_checksum(path, typ)
UnboundLocalError: local variable 'typ' referenced before assignment

```

PR #4112, opened 2022-10-25
add explicit return when no checksums match those specified in a tuple

jfgrimm:
closes https://github.com/easybuilders/easybuild-framework/issues/4111

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1300,7 +1300,7 @@ def verify_checksum(path, checksums):
                         return True
                     else:
                         _log.info("Ignoring non-matching checksum for %s (%s)...", path, cand_checksum)
-                
+
                 # no matching checksums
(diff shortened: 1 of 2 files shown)


boegel (changes requested):
@jfgrimm The test for `verify_checksum` in `test/framework/filetools.py` should be updated accordingly (i.e. check return value, or maybe even add a test case)

jfgrimm:
@boegel I've added a test, and verified that this test would have failed without the [c6d3c30](https://github.com/easybuilders/easybuild-framework/pull/4112/commits/c6d3c307451c922e3af0eb8c07ba44270a463587), [ebeddec](https://github.com/easybuilders/easybuild-framework/pull/4112/commits/ebeddecb53114f071971caf6cc41a4afaae9d346) commits

boegel (approved):

PR #4115, opened 2022-10-27
EasyBuild 4.6.2 Ubuntu friendly PNS

marbud0:
Simple new PNS that complies with the deb policy rule: version has to start with a number.

The standard EB PNS generates version strings beginning with `eb_`. This is violating the [Debian package version string policy](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version "Debian package version string policy") requirement of the version string always starting with a number. 
This is a implimentation of a new PNS called EasyBuildDebFriendlyPNS that postfix `eb` to the version string instead of prefixing and also add a `0` to any upstream version string NOT starting with a number

### DIFF
--- a/test/framework/package.py
+++ b/test/framework/package.py
@@ -192,7 +192,7 @@ class PackageTest(EnhancedTestCase):
         elif get_package_naming_scheme() == "EasyBuildDebFriendlyPNS":
             # default: EasyBuild deb friendly package naming scheme, pkg release 1
             self.assertEqual(pns.name(ec), 'OpenMPI-2.1.2-GCC-6.4.0-2.28')
-            self.assertEqual(pns.version(ec), '%s_eb' % EASYBUILD_VERSION)
+            self.assertEqual(pns.version(ec), '%s-eb' % EASYBUILD_VERSION)
             self.assertEqual(pns.release(ec), '1')
 
     def test_package(self):
--- a/test/framework/package.py
+++ b/test/framework/package.py
@@ -36,7 +36,7 @@ from test.framework.utilities import EnhancedTestCase, TestLoaderFiltered, init_
 from unittest import TextTestRunner
 
 from easybuild.framework.easyconfig.easyconfig import EasyConfig
-from easybuild.tools.config import log_path
+from easybuild.tools.config import get_package_naming_scheme, log_path
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import adjust_permissions, read_file, write_file
 from easybuild.tools.package.utilities import ActivePNS, avail_package_naming_schemes, check_pkg_support, package
@@ -154,7 +154,7 @@ class PackageTest(EnhancedTestCase):
 
     def test_avail_package_naming_schemes(self):
         """Test avail_package_naming_schemes()"""
-        self.assertEqual(sorted(avail_package_naming_schemes().keys()), ['EasyBuildPNS'])
+        self.assertEqual(sorted(avail_package_naming_schemes().keys()), ['EasyBuildDebFriendlyPNS', 'EasyBuildPNS'])
 
     def test_check_pkg_support(self):
         """Test check_pkg_support()."""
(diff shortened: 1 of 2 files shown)


boegel:
@marbud0 Are you up for making sure that this package naming scheme is covered by the tests?
See `test/framework/package.py`

marbud0:
Not sure where to add a test that would use the new deb friendly PNS. Is that controlled from the test execution env or should I somehow set the PNS from the test code?  (Assumed the first for now..)

marbud0:
Don't feel responsible for the [EasyBuild framework unit tests / build (3.6, Lmod-8.7.6, Tcl) (pull_request)](https://github.com/easybuilders/easybuild-framework/actions/runs/3386902539/jobs/5653762572) failure during the last run.. (Network unreachable...)... Not sure about [EasyBuild framework unit tests / build (2.7, Lmod-7.8.22, Tcl) (pull_request)](https://github.com/easybuilders/easybuild-framework/actions/runs/3386902539/jobs/5653762230)... Please advice.

branfosj:
> Don't feel responsible for the [EasyBuild framework unit tests / build (3.6, Lmod-8.7.6, Tcl) (pull_request)](https://github.com/easybuilders/easybuild-framework/actions/runs/3386902539/jobs/5653762572) failure during the last run.. (Network unreachable...)... Not sure about [EasyBuild framework unit tests / build (2.7, Lmod-7.8.22, Tcl) (pull_request)](https://github.com/easybuilders/easybuild-framework/actions/runs/3386902539/jobs/5653762230)... Please advice.

I triggered a rerun on both failing tests and they both passed. We do see intermittent failures.

test/framework/package.py
```diff
@@ -184,10 +184,16 @@ def test_active_pns(self):
...
-        self.assertEqual(pns.release(ec), '1')
+        if get_package_naming_scheme() == "EasyBuildPNS":
+            # default: EasyBuild package naming scheme, pkg release 1
+            self.assertEqual(pns.name(ec), 'OpenMPI-2.1.2-GCC-6.4.0-2.28')
+            self.assertEqual(pns.version(ec), 'eb-%s' % EASYBUILD_VERSION)
+            self.assertEqual(pns.release(ec), '1')
+        elif get_package_naming_scheme() == "EasyBuildDebFriendlyPNS":
+            # default: EasyBuild deb friendly package naming scheme, pkg release 1
+            self.assertEqual(pns.name(ec), 'OpenMPI-2.1.2-GCC-6.4.0-2.28')
+            self.assertEqual(pns.version(ec), '%s_eb' % EASYBUILD_VERSION)
```
ocaisa, line 195: Shoulding this be
```suggestion
            self.assertEqual(pns.version(ec), '%s-eb' % EASYBUILD_VERSION)
```
(or the return value from `pns.version(ec)` changed)?

test/framework/package.py
```diff
@@ -184,10 +184,16 @@ def test_active_pns(self):
...
-        # default: EasyBuild package naming scheme, pkg release 1
-        self.assertEqual(pns.name(ec), 'OpenMPI-2.1.2-GCC-6.4.0-2.28')
-        self.assertEqual(pns.version(ec), 'eb-%s' % EASYBUILD_VERSION)
-        self.assertEqual(pns.release(ec), '1')
+        if get_package_naming_scheme() == "EasyBuildPNS":
+            # default: EasyBuild package naming scheme, pkg release 1
+            self.assertEqual(pns.name(ec), 'OpenMPI-2.1.2-GCC-6.4.0-2.28')
+            self.assertEqual(pns.version(ec), 'eb-%s' % EASYBUILD_VERSION)
+            self.assertEqual(pns.release(ec), '1')
+        elif get_package_naming_scheme() == "EasyBuildDebFriendlyPNS":
```
ocaisa, line 192: I was wondering how this test could be passing when there's a clear typo...and now I know. `EasyBuildDebFriendlyPNS` is available as a package MNS, but it is never actually set in the test suite, (so this code path is never followed)

ocaisa: You can change the PNS by setting an environment variable, so that would mean adding a loop around the code inside the function:
```
    def test_active_pns(self):
        """Test use of ActivePNS."""
        for pns_type in ['EasyBuildDebFriendlyPNS', 'EasyBuildPNS']:
            os.environ['EASYBUILD_PACKAGE_NAMING_SCHEME'] = pns_type
            init_config(build_options={'silent': True})
            ....
            del os.environ['EASYBUILD_PACKAGE_NAMING_SCHEME']
```

ocaisa: I've updated the previous comment to include the full code content

ocaisa: It's the same as before but with a loop around the test function contents

marbud0: Implimented as suggested. Test passes for me..

marbud0: Should not 'os.environ['EASYBUILD_PACKAGE_NAMING_SCHEME']` really be `os.environ['TEST_EASYBUILD_PACKAGE_NAMING_SCHEME']`?

ocaisa: I got that from the module naming scheme tests so I don't think so, I think it matters that you're setting from within the test suite. A good check would be to reintroduce the typo and see if the tests break

marbud0: I ran the debugger on it. It is executed.. That's why I know it works.

ocaisa (approved):
LGTM

ocaisa:
Going in, thanks @marbud0 !

marbud0:
Thank you for the Review and help with the tests @ocaisa  !

Issue #833, opened 2014-02-03
Failing test due to config.cfg


wpoely86:
I've got 2 lines in my ```config.cfg```:
```
logtostdout=True
repository=GitRepository
```
This makes 11 tests fail. If I remove my config.cfg, all tests work.
All failing tests are related to ```test.framework.options*```. All of them (except the last one) are related to the first line in my ```config.cfg```. 

These fail:
```
======================================================================
FAIL: test_debug (test.framework.options.CommandLineOptionsTest)
Test enabling debug logging.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/framework/options.py", line 143, in test_debug
    self.assertTrue(res, "%s log messages are included when using %s: %s" % (log_msg_type, debug_arg, outtxt))
AssertionError: DEBUG log messages are included when using -d: 

======================================================================
FAIL: test_force (test.framework.options.CommandLineOptionsTest)
Test forcing installation even if the module is already available.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/framework/options.py", line 225, in test_force
    self.assertTrue(re.search(already_msg, outtxt), "Already installed message without --force, outtxt: %s" % outtxt)
AssertionError: Already installed message without --force, outtxt: 

======================================================================
FAIL: test_ignore_osdeps (test.framework.options.CommandLineOptionsTest)
Test ignoring of listed OS dependencies.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/framework/options.py", line 821, in test_ignore_osdeps
    self.assertTrue(regex.search(outtxt), "OS dependencies are checked, outtxt: %s" % outtxt)
AssertionError: OS dependencies are checked, outtxt: 

======================================================================
FAIL: test_info (test.framework.options.CommandLineOptionsTest)
Test enabling info logging.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/framework/options.py", line 165, in test_info
    self.assertTrue(res, "%s log messages are included when using %s (err: %s, out: %s)" % (log_msg_type, info_arg, myerr, outtxt))
AssertionError: INFO log messages are included when using --info (err: "EasyBuild crashed with an error (at easybuild/tools/build_log.py:155 in print_error): Unable to find an easyconfig for the given specifications: {'name': 'somethingrandom'}; to make EasyBuild try to generate a matching easyconfig, use the --try-X options ", out: )

======================================================================
[...]

PR #4116, opened 2022-10-30
add archlinux compatibility for os dependency resolution 

LukeLabrie:
per issue [#833](https://github.com/easybuilders/easybuild/issues/833#issuecomment-1292510148), added some logic to accommodate some archlinux distros (currently just arch and manjaro) when checking for os dependencies. some notes:

- should work for any arch distro that uses pacman as long as it's added to the `arch_distros` list
- i had to specifically define `EASYCONFIG_OS_DEPENDENCY_CONSTANTS` because an import of `EASYCONFIG_CONSTANTS` from `easybuild.framework.easyconfig.constants` results in a circular import... perhaps there's a better way to do it
- the names of the packages need to be changed when using those constants, since they have different names on arch. arch packages typically don't have the "...-dev" suffix, but it seems like the arch packages incorporate the dev features from other distros. I don't think you can just add the names to `EASYCONFIG_CONSTANTS` because then you might get the wrong version on other distros

works for me locally... let me know your thoughts

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -169,6 +169,7 @@ ARM_CORTEX_IDS = {
 RPM = 'rpm'
 DPKG = 'dpkg'
 ZYPPER = 'zypper'
+PACMAN = 'pacman'
 
 SYSTEM_TOOLS = {
     '7z': "extracting sources (.iso)",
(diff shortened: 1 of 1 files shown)


easybuild/tools/systemtools.py
```diff
@@ -211,6 +212,19 @@
     'setuptools': ('pkg_resources', "obtaining information on Python packages via pkg_resources module"),
 }
 
+# os dependency constants that can be used in easyconfig (for name updates on arch distros)
+ARCH_OS_DEP_CONSTANTS = {
+    'IBVERBS_DEV': (('libibverbs-dev', 'libibverbs-devel', 'rdma-core-devel'),
```
boegel, line 217: @LukeLabrie We already have constants for this in `easybuild/framework/easyconfig/constants.py`, we shouldn't copy that here.

If the package names for `openssl` & co are slightly different on Arch, then we should just add an item to the tuples in `easybuild/framework/easyconfig/constants.py`, since the tuples specify *alternatives* for the package names that are considered.

So for example, adding `rdma-core` to the tuple for `OS_PKG_IBVERBS_DEV`:

```python
EASYCONFIG_CONSTANTS = {
    ...
    'OS_PKG_IBVERBS_DEV': (('libibverbs-dev', 'libibverbs-devel', 'rdma-core-devel', 'rdma-core'),
                           "OS packages providing ibverbs/infiniband development support"),

boegel (changes requested):

easybuild/tools/systemtools.py
LukeLabrie: Thanks for the feedback @boegel. To your first comment, I tried that initially but since `easybuild/framework/easyconfig/constants.py` imports `easybuild/tools/systemtools.py` importing the constants from `constants.py` results in a circular import -- but as I mentioned I'm sure there's a cleaner way to avoid this so I can look into that. 

To your second comment, I also tried that initially, but wouldn't that create an issue on distros where there is a meaningful difference between the names `rdma-core` and `rdma-core-devel`? For example, I think these would be two different packages on redhat distros, so adding `rdma-core` to the tuple would mean a redhat user would pass a dependency check even if they didn't have `rdma-core-devel`. That's the only reason I added specific logic for arch distros.

boegel: Right, that's a good point on `rdma-core-devel` vs `rdma-core`.

I guess that forces us to do the override you're suggesting for Arch/Manjaro, at least with the current approach.

It's probably a good idea to add a comment above the code block you're currently adding to `constants.py`, since this complication is not immediately obvious, just to avoid that someone else goes "how silly, let's just add `rdma-core` to the tuple instead"...

PR #4119, opened 2022-11-02
add support for --filter-rpath-sanity-libs to skip RPATH sanity check for designated libraries

### LINKED ISSUE #4095: `sanity_check_rpath` should not check for `libcuda.so.1`
When installing Clang, e.g. `Clang-12.0.1-GCCcore-10.3.0.eb` with `--rpath`, it errors out with

```
== 2022-10-12 13:53:54,035 build_log.py:169 ERROR EasyBuild crashed with an error (at easybuild/RHEL8/2022/software/EasyBuild/4.6.1/lib/python3.6/site-packages/easybuild/base/exceptions.py:124 in __init__): Sanity check failed: One or more required libraries not found for /home/casparl/.local/easybuild/RHEL8/2022/software/Clang/12.0.1-GCCcore-10.3.0/lib/libomptarget.rtl.cuda.so:   linux-vdso.so.1 (0x00007fffb959f000)
        libcuda.so.1 => not found
        libelf.so.1 => /sw/arch/Centos8/EB_production/2021/software/elfutils/0.185-GCCcore-10.3.0/lib64/libelf.so.1 (0x0000152603766000)
        libstdc++.so.6 => /home/casparl/.local/easybuild/Centos8/2021/software/GCCcore/10.3.0/lib64/libstdc++.so.6 (0x000015260339b000)
        libm.so.6 => /lib64/libm.so.6 (0x0000152603019000)
        libgcc_s.so.1 => /home/casparl/.local/easybuild/Centos8/2021/software/GCCcore/10.3.0/lib64/libgcc_s.so.1 (0x000015260374a000)
        libc.so.6 => /lib64/libc.so.6 (0x0000152602c54000)
        /lib64/ld-linux-x86-64.so.2 (0x0000152603582000)
        libz.so.1 => /sw/arch/Centos8/EB_production/2021/software/zlib/1.2.11-GCCcore-10.3.0/lib/../lib64/libz.so.1 (0x0000152603726000)

One or more required libraries not found for /home/casparl/.local/easybuild/RHEL8/2022/software/Clang/12.0.1-GCCcore-10.3.0/lib64/libomptarget.rtl.cuda.so:     linux-vdso.so.1 (0x000014f10ef12000)
        libcuda.so.1 => not found
        libelf.so.1 => /sw/arch/Centos8/EB_production/2021/software/elfutils/0.185-GCCcore-10.3.0/lib64/libelf.so.1 (0x000014f10eec6000)
        libstdc++.so.6 => /home/casparl/.local/easybuild/Centos8/2021/software/GCCcore/10.3.0/lib64/libstdc++.so.6 (0x000014f10eb01000)
        libm.so.6 => /lib64/libm.so.6 (0x000014f10e77f000)
        libgcc_s.so.1 => /home/casparl/.local/easybuild/Centos8/2021/software/GCCcore/10.3.0/lib64/libgcc_s.so.1 (0x000014f10eeaa000)
        libc.so.6 => /lib64/libc.so.6 (0x000014f10e3ba000)
        /lib64/ld-linux-x86-64.so.2 (0x000014f10ece8000)
        libz.so.1 => /sw/arch/Centos8/EB_production/2021/software/zlib/1.2.11-GCCcore-10.3.0/lib/../lib64/libz.so.1 (0x000014f10ee86000)
 (at easybuild/RHEL8/2022/software/EasyBuild/4.6.1/lib/python3.6/site-packages/easybuild/framework/easyblock.py:3471 in _sanity_check_step)
```

However, `libcuda.so.1` should _never_ be `RPATH`-ed. Nvidia supports cross compilation by linking (at compile time) against `stubs`  libraries. Those `stubs` libraries contain function declarations, but no implementations. Thus, at runtime, they are _never_ meant to be used. See https://github.com/easybuilders/easybuild-framework/issues/2683 

We made sure that the `RPATH` wrappers allow specifying certain paths to _never_ be `RPATH`-ed, and we use that to exclude the `stubs`  libraries from `CUDA`. See for the implementation https://github.com/easybuilders/easybuild-framework/pull/2725 and  https://github.com/easybuilders/easybuild-framework/pull/3850

The problem that we _didn't_ tackle is that if you filter those directories from being `RPATH`-ed, the `sanity_check_rpath`, the libraries that are linked from those directories will fail the `sanity_check_rpath` (as above).

*Minimal reproducer*
Put the following cpp in a `tar.gz`:
```
# toy.cpp
// Your First C++ Program

#include <iostream>

int main() {
    std::cout << "Hello World!";
    return 0;
}
```
Then create an EasyConfig that does e.g.
```
# toy-1.0.0-GCCcore-11.3.0.eb

easyblock='CmdCp'

name = 'Toy'
version = '1.0.0'

homepage = 'https://whatever.org/'
description = """Toy"""

toolchain = {'name': 'GCCcore', 'version': '11.3.0'}

sources = [
    'toy.tar.gz',
]

cmds_map = [('toy', 'g++ -o toy -lcuda toy.cpp')]

files_to_copy = [(["toy"], 'bin')]

builddependencies = [
    ('binutils', '2.38'),
]

dependencies = [
    ('CUDA', '11.7.0', '', SYSTEM),
]

sanity_check_paths = {
    'files': ['bin/toy'],
    'dirs': [],
}

moduleclass = 'tools'
```
Then, try to install with
```
eb toy-1.0.0-GCCcore-11.3.0-eb --sourcepath=<path/containing/toy/tarball>
```
We discussed a bit on the Slack, and we should probably alter https://github.com/easybuilders/easybuild-framework/blob/0a506aa01407785429038f243941b2526bba646f/easybuild/framework/easyblock.py#L2966 to skip certain libraries, much like we hard-enabled filtering `RPATH` for de CUDA stubs dir in https://github.com/easybuilders/easybuild-framework/blob/6cab5e2aa6d03ffcd98288bfa648e0a264954cee/easybuild/tools/toolchain/toolchain.py#L961

casparvl:
fixes #4095

This PR will allow a (configurable) list of libraries to be ignored during the RPATH sanity check. This is e.g. needed when cross-compiling an executable that links against `libcuda.so.1`. The way NVIDIA supports cross-compilation is by compiling against a stub library, provided by the CUDA toolkit. The stub library only provides the symbols, but not the imlementation of the actual function that should be used at runtime. For example, `<prefix>/CUDA/11.7.0/lib/stubs/libcuda.so.1` contains all the symbols of the `libcuda.so.1`, but no implementations. Thus, it should _never_ be used at runtime: the `libcuda.so.1` that comes with the driver installation is supposed to be used at runtime.

That's also why we don't `RPATH` anything that's in a `stubs` directory (see e.g. https://github.com/easybuilders/easybuild-framework/issues/2683 and [this snippet](https://github.com/easybuilders/easybuild-framework/blob/5c81f748e8b2cf7c8d7fa0fbb003af552fa5113f/easybuild/tools/toolchain/toolchain.py#L961) of the `easybuild-framework` code).

The issue is that the `sanity_check_rpath()` step will _still_ try to unset `LD_LIBRARY_PATH` and see if all libraries are found correctly. On nodes that have the CUDA driver installed, that works fine, as `libcuda.so.1` is picked up from a default location (`/usr/lib64/libcuda.so.1`). However, on CPU nodes, this fails. The fact that `libcuda.so.1` is not found on CPU nodes when cross compiling is entirely legitimate though.

A point of discussion is what the default behaviour should be for EasyBuild, i.e. for which default list of libraries it will be 'accepted' if they are not found in the `sanity_check_rpath()` step. In my opinion, it makes sense to include `libcuda.so`, `libcuda.so.1`, `libnvidia-ml.so` and `libnvidia-ml.so.1` here, as these are the libraries for which _only_ stubs are provided in the CUDA toolkit and are thus supposed to be provided by the driver. For some other libraries (e.g. `libcublas.so`), there are stubs in the `stubs` directory, but also real ones in the  `<prefix>/CUDA/11.7.0/lib/` directory. Thus, I assume one would want those RPATH-ed against the latter locations.


casparvl:
For completeness: this is an example of what the drivers look like

```
ls -al /usr/lib64/libnvidia-ml.so*
lrwxrwxrwx. 1 root root      25 Apr 27  2022 /usr/lib64/libnvidia-ml.so -> libnvidia-ml.so.515.43.04
lrwxrwxrwx. 1 root root      25 Apr 27  2022 /usr/lib64/libnvidia-ml.so.1 -> libnvidia-ml.so.515.43.04
-rwxr-xr-x. 1 root root 1683960 Apr 26  2022 /usr/lib64/libnvidia-ml.so.515.43.04
```

i.e. here, the `*.so.1` and `*.so` are softlinks to the versioned-library

casparvl:
I'm puzzled by the CI failures of the unit tests... I added this to the `test_toy_rpath` test:

```
        # test sanity error when --rpath-filter is used to filter a required library
        toy_ec_txt = read_file(os.path.join(test_ecs, 't', 'toy', 'toy-0.0.eb'))
        toy_ec_txt += "\ndependencies = [('libtoy', '0.0', '', SYSTEM)]"
        toy_ec_txt += "\nbuildopts = '-ltoy'"
        toy_ec = os.path.join(self.test_prefix, 'toy.eb')
        write_file(toy_ec, toy_ec_txt)
        error_pattern = r"Sanity check failed\: Library libtoy\.so not found"
        self.assertErrorRegex(EasyBuildError, error_pattern, self.test_toy_build, ec_file=toy_ec,
                              extra_args=['--rpath', '--experimental', '--rpath-filter=.*libtoy.*'],
                              raise_error=True, verbose=False)
```

What this is _meant_ to do is:

- Add `libtoy` as a dependency
- Run with `--rpath`, but then filter `libtoy` from being `RPATH`-ed
- Fail on the `sanity_check_rpath`, as `libtoy` "failed" to be `RPATH`-ed (on purpose)
- Check that this produces the expected failure message

On my local machine, I run `python3 -m test.framework.toy_build rpath` (using the `skip_rpath_check` branch, of course) and it works fine. I've actually tried to check it out on another machine, just to make sure I didn't magically fail to check some change in or something.

Yet, in the EasyBuild CI, this produces a failure. If I understand it correctly, this failure tells me that the last line in the above addition did _not_ actually produce an error at all. That puzzles me: how can a `--rpath --rpath-filter=.*libtoy.*` build _not_ fail the `sanity_check_rpath`? The only way I can imagine is if `libtoy.so` is _also_ available somewhere in the default search path, but I have no idea how to check that (if I had the full build logs, I could...)

casparvl:
Hm, for some reason it doesn't seem to pick up on 

```
toy_ec_txt += "\nbuildopts = '-ltoy'"
```
I guess, since the `ldd` in the RPATH sanity check shows (see raw log of [this CI run](https://github.com/easybuilders/easybuild-framework/actions/runs/3469274027/jobs/5796071244)):
```
2022-11-15T09:58:27.4634375Z == 2022-11-15 09:58:26,617 run.py:682 DEBUG Using default regular expression: (?<![(,-]|\w)(?:error|segmentation fault|failed)(?![(,-]|\.?\w)
2022-11-15T09:58:27.4634882Z == 2022-11-15 09:58:26,617 run.py:215 DEBUG run_cmd: running cmd ldd /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/toy/0.0/bin/toy (in /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/toy/0.0)
2022-11-15T09:58:27.4635239Z == 2022-11-15 09:58:26,618 run.py:234 INFO running cmd: ldd /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/toy/0.0/bin/toy 
2022-11-15T09:58:27.4635652Z == 2022-11-15 09:58:26,662 run.py:648 DEBUG cmd "ldd /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/toy/0.0/bin/toy" exited with exit code 0 and output:
2022-11-15T09:58:27.4635828Z 	linux-vdso.so.1 (0x00007ffeaf974000)
2022-11-15T09:58:27.4636134Z 	libc.so.6 => /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libc.so.6 (0x00007f2b6e347000)
2022-11-15T09:58:27.4636338Z 	/lib64/ld-linux-x86-64.so.2 (0x00007f2b6e540000)
```

casparvl:
Ok, this confuses me:

```
2022-11-15T09:59:20.0339056Z == 2022-11-15 09:58:25,527 run.py:648 DEBUG cmd "readelf -d /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/bin/toy" exited with exit code 0 and output:
2022-11-15T09:59:20.0339241Z  0x0000000000000001 (NEEDED)             Shared library: [libtoy.so]
2022-11-15T09:59:20.0339368Z  0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
2022-11-15T09:59:20.0339781Z == 2022-11-15 09:58:25,527 run.py:682 DEBUG Using default regular expression: (?<![(,-]|\w)(?:error|segmentation fault|failed)(?![(,-]|\.?\w)
2022-11-15T09:59:20.0340252Z == 2022-11-15 09:58:25,527 easyblock.py:3051 DEBUG Output of 'readelf -d /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/bin/toy' checked, looks OK
2022-11-15T09:59:20.0340701Z == 2022-11-15 09:58:25,528 easyblock.py:3012 DEBUG Sanity checking RPATH for files in /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib
2022-11-15T09:59:20.0341219Z == 2022-11-15 09:58:25,528 easyblock.py:3015 DEBUG Sanity checking RPATH for /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib/libtoy.so
2022-11-15T09:59:20.0341798Z == 2022-11-15 09:58:25,528 run.py:215 DEBUG run_cmd: running cmd file /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib/libtoy.so (in /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0)
2022-11-15T09:59:20.0342262Z == 2022-11-15 09:58:25,528 run.py:234 INFO running cmd: file /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib/libtoy.so 
2022-11-15T09:59:20.0342738Z == 2022-11-15 09:58:25,566 run.py:648 DEBUG cmd "file /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib/libtoy.so" exited with exit code 0 and output:
2022-11-15T09:59:20.0343356Z /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib/libtoy.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=94fd5150a1f4a3d6fe72542856ebde879f3d033c, not stripped
2022-11-15T09:59:20.0343771Z == 2022-11-15 09:58:25,567 run.py:682 DEBUG Using default regular expression: (?<![(,-]|\w)(?:error|segmentation fault|failed)(?![(,-]|\.?\w)
2022-11-15T09:59:20.0344349Z == 2022-11-15 09:58:25,567 run.py:215 DEBUG run_cmd: running cmd ldd /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib/libtoy.so (in /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0)
2022-11-15T09:59:20.0344769Z == 2022-11-15 09:58:25,567 run.py:234 INFO running cmd: ldd /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib/libtoy.so 
2022-11-15T09:59:20.0345237Z == 2022-11-15 09:58:25,611 run.py:648 DEBUG cmd "ldd /tmp/eb-b2lZfe/eb-JYJdDP/eb-DiAzuk/tmpW7VnBk/software/libtoy/0.0/lib/libtoy.so" exited with exit code 0 and output:
2022-11-15T09:59:20.0345403Z 	linux-vdso.so.1 (0x00007ffe57947000)
2022-11-15T09:59:20.0345759Z 	libc.so.6 => /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libc.so.6 (0x00007fdc32b76000)
2022-11-15T09:59:20.0346032Z 	/lib64/ld-linux-x86-64.so.2 (0x00007fdc32d6f000)
```

The `readelf` shows `libtoy.so` is a `NEEDED` library. But then `ldd` doesn't list it?

casparvl:
The plot thickens: in CI, I see

```
2022-11-15T16:49:16.2419326Z == 2022-11-15 16:49:15,392 run.py:215 DEBUG run_cmd: running cmd  gcc toy.c -o toy -ltoy && ldd toy (in /tmp/eb-hsfctg_q/eb-u25uz390/eb-_1jhknof/tmpou1hv9y4/toy/0.0/system-system/toy-0.0)
2022-11-15T16:49:16.2419580Z == 2022-11-15 16:49:15,392 run.py:234 INFO running cmd:  gcc toy.c -o toy -ltoy && ldd toy 
2022-11-15T16:49:16.2419881Z == 2022-11-15 16:49:15,514 run.py:648 DEBUG cmd " gcc toy.c -o toy -ltoy && ldd toy" exited with exit code 0 and output:
2022-11-15T16:49:16.2420049Z 	linux-vdso.so.1 (0x00007ffd3f74f000)
2022-11-15T16:49:16.2420340Z 	libc.so.6 => /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libc.so.6 (0x00007fe0cf27f000)
2022-11-15T16:49:16.2420539Z 	/lib64/ld-linux-x86-64.so.2 (0x00007fe0cf478000)
```
Note that a local run produces:

```
== 2022-11-15 17:40:52,447 run.py:648 DEBUG cmd " gcc toy.c -o toy -ltoy && ldd toy" exited with exit code 0 and output:
        linux-vdso.so.1 (0x00001549fa9f4000)
        libtoy.so => /tmp/eb-j014pn4l/eb-tvdj052c/tmpaj4p1cd9/software/libtoy/0.0/lib/libtoy.so (0x00001549fa5c8000)
        libc.so.6 => /usr/lib/gcc/x86_64-redhat-linux/8/../../../../lib64/libc.so.6 (0x00001549fa203000)
        /lib64/ld-linux-x86-64.so.2 (0x00001549fa7ca000)
```
as expected.

casparvl:
Found the issue: Ubuntu passes `--as-needed` by default to the linker (see https://wiki.ubuntu.com/ToolChain/CompilerFlags#A-Wl.2C--as-needed ). You can see this when you run 

```
echo "int main(void) {}" | gcc $(dpkg-buildflags --get LDFLAGS) -o /dev/null -v -x c - &> /dev/stdout| grep collect
```

on an Ubuntu system. Most systems will _end_ with a `--no-as-needed` flag, so that any `-lmylib` passed by the user on command line will effectively default to `--no-as-needed` behavior. Ubuntu systems don't, they end with a `--as-needed` flag, which thus applies as the default to any of the `-l` arguments passed by the user. Annoying, because the `ld` manual states that "`--no-as-needed` restores the default behavior", implicitely saying this is the default.

Of course, I can pass the `-Wl,--no-as-needed` flag myself in the unit test, but a more robust fix is probably to just use an executable that _actually_ uses a symbol from `libtoy`, so that even with `--as-needed` it is still linked.

akesandgren (dismissed):
LGTM

easybuild/tools/config.py
```diff
@@ -113,6 +113,7 @@
 DEFAULT_PR_TARGET_ACCOUNT = 'easybuilders'
 DEFAULT_PREFIX = os.path.join(os.path.expanduser('~'), ".local", "easybuild")
 DEFAULT_REPOSITORY = 'FileRepository'
+DEFAULT_FILTER_RPATH_SANITY_LIBS = ['libcuda.so', 'libcuda.so.1', 'libnvidia-ml.so', 'libnvidia-ml.so.1']
```
boegel, line 116: We should add a comment here to explain why we're using these as default (as you explained in the PR description), and use a tuple rather than a list (since a list is mutable, which doesn't fit well with the idea of a "constant"; I know we're not consistent on this in the existing constants, but no need to make it worse ;) ).
Also, one per line to make it easier to consume for humans (and keep the group of `DEFAULT_*` constants sorted alphabetically):

```python
# this comment explains why we're including the libraries below
DEFAULT_FILTER_RPATH_SANITY_LIBS = (
    'libcuda.so',
    'libcuda.so.1',
    'libnvidia-ml.so',
    'libnvidia-ml.so.1',
)

File: easybuild/tools/options.py [hunk 2]
```diff
@@ -404,6 +405,11 @@ def override_options(self):
                            'strlist', 'extend', None),
             'filter-env-vars': ("List of names of environment variables that should *not* be defined/updated by "
                                 "module files generated by EasyBuild", 'strlist', 'extend', None),
+            'filter-rpath-sanity-libs': ("List of libraries that you want to ignore when the RPATH sanity check is "
```
boegel, line 408: `that you want ignore` -> `that should be ignored`

test/framework/sandbox/sources/toy2/toy2-0.0/toy2.c
```diff
@@ -0,0 +1,6 @@
+#include <toy.h>
+
+int main(int argc, char* argv[]){
+    toy_function();
```
boegel, line 4: Add a comment here that this is provided by the `libtoy` library?

Maybe we should also use a better name that `toy2`... How about `toy-cli` or `toy-app`?

test/framework/toy_build.py
```diff
@@ -2707,6 +2708,25 @@ def grab_gcc_rpath_wrapper_args():
...
+
+        # test sanity error when --rpath-filter is used to filter a required library
+        # In this test, libtoy.so will be linked, but not RPATH-ed due to the --rpath-filter
+        # Thus, the RPATH sanity check is expected to fail with libtoy.so not being found
+        test_ecs = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'test_ecs')
+        toy_ec = os.path.join(test_ecs, 't', 'toy2', 'toy2-0.0.eb')
+        error_pattern = r"Sanity check failed\: Library libtoy\.so not found"
+        self.assertErrorRegex(EasyBuildError, error_pattern, self.test_toy_build, ec_file=toy_ec,
+                              extra_args=['--rpath', '--experimental', '--rpath-filter=.*libtoy.*'],
+                              name='toy2', raise_error=True, verbose=False)
```
boegel, line 2722: Also run again without `--rpath-filter`, for completeness sake?
It's generally a good idea to make sure that the test setup is done fully correctly...

```suggestion
        args = ['--rpath', '--experimental', '--rpath-filter=.*libtoy.*']
        self.assertErrorRegex(EasyBuildError, error_pattern, self.test_toy_build, ec_file=toy_ec,
                              extra_args=args, name='toy2', raise_error=True, verbose=False)
        # works fine if --rpath-filter is not used (since then libtoy is RPATH'ed)
        self.test_toy_build(ec_file=toy_ec, name='toy2', extra_args=args, raise_error=True)
```

You will need to clean up the install dir to make sure the next part of the test still works as expected.

test/framework/toy_build.py
```diff
@@ -2707,6 +2708,25 @@ def grab_gcc_rpath_wrapper_args():
...
+
+        # test use of --filter-rpath-sanity-libs option. In this test, we use --rpath-filter to make sure libtoy.so is
+        # not rpath-ed. Then, we use --filter-rpath-sanity-libs to make sure the RPATH sanity checks ignores
+        # the fact that libtoy.so is not found. Thus, this build should complete succesfully
+        args = ['--rpath', '--experimental', '--rpath-filter=.*libtoy.*', '--filter-rpath-sanity-libs=libtoy.so']
+        self.test_toy_build(ec_file=toy_ec, name='toy2', extra_args=args, raise_error=True)
```
boegel, line 2728: Maybe here we should also do a check on the output of `readelf -d` on the binary, and check the RPATH section with a regex (to make sure it doesn't include `libtoy`)?
That mainly makes sense if my suggestion above to also run once without `--rpath-filter` is applied.

boegel (changes requested):

easybuild/framework/easyblock.py
```diff
@@ -2984,9 +2985,16 @@ def sanity_check_rpath(self, rpath_dirs=None):
         self.log.debug("$LD_LIBRARY_PATH during RPATH sanity check: %s", os.getenv('LD_LIBRARY_PATH', '(empty)'))
         self.log.debug("List of loaded modules: %s", self.modules_tool.list())
 
-        not_found_regex = re.compile('not found', re.M)
+        not_found_regex = re.compile(r'([^\s]*) \=\> not found')
```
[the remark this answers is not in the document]
casparvl: > Ah, no, it's because of the switch from search to findall below, OK.

Yes, indeed

> But my suggestion on [^\s]* vs \S+ remains.

Valid point, \S+ does the job, I'll replace it

easybuild/framework/easyblock.py
```diff
@@ -2984,9 +2985,16 @@ def sanity_check_rpath(self, rpath_dirs=None):
...
         readelf_rpath_regex = re.compile('(RPATH)', re.M)
 
+        # List of libraries that should be exempt from the RPATH sanity check
+        # E.g. libcuda.so.1 should never be RPATH-ed by design, see
+        # ttps://github.com/easybuilders/easybuild-framework/issues/4095
+        rpath_exception_libs = build_option('filter_rpath_sanity_libs')
```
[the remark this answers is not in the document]
casparvl: Agreed, let's keep it uniform. Will replace

easybuild/framework/easyblock.py
```diff
@@ -2984,9 +2985,16 @@ def sanity_check_rpath(self, rpath_dirs=None):
...
+        msg = "Ignoring the following libraries if they are not found by RPATH sanity check: %s" % rpath_exception_libs
+        self.log.info(msg)
```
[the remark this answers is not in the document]
casparvl: Ah, I keep forgetting about the lazy logging. But I didn't forget why it was important, so: I'll replace :)

easybuild/framework/easyblock.py
```diff
@@ -3013,10 +3021,18 @@ def sanity_check_rpath(self, rpath_dirs=None):
...
-                            self.log.warning(fail_msg)
-                            fails.append(fail_msg)
+                        matches = re.findall(not_found_regex, out)
+                        if len(matches) > 0:  # Some libraries are not found via 'ldd'
+                            # For each match, check if the library is in the exception list
+                            for match in matches:
+                                if match in rpath_exception_libs:
+                                    msg = "Library %s not found for %s, but ignored " % (match, path)
+                                    msg += "since it is on the rpath exception list: %s" % rpath_exception_libs
+                                    self.log.info(msg)
```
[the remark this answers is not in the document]
casparvl: Will do

easybuild/tools/options.py
casparvl: Yes, better, will fix

test/framework/sandbox/sources/toy2/toy2-0.0/toy2.c
casparvl: Done

easybuild/tools/config.py
casparvl: Done

test/framework/toy_build.py
casparvl: Why do we need to clean? A rebuild will clean the installdir anyway, doesn't it?

test/framework/toy_build.py
casparvl: Hm, good idea, but I need some help with this. How do I retrieve the installdir in a test? And what is the appropriate way to run `readelf -d` in a test, do we use the `run_cmd` function for that?  (if so, I need to find where that is defined again... :))

test/framework/toy_build.py
boegel: Yes, indeed

boegel (approved):
lgtm

Issue #4120, opened 2022-11-04
r8 options different for GCC and Intel


hattom:
the `r8` option for the Intel toolchain produces `-r8` for the `ifort` compiler.
This promotes reals to 64bit by default, but leaves doubles as 64bit.

The `r8` option for the GCC toolchain produces `-fdefault-real-8` for the `gfortran` compiler.
This promotes reals to 64bit by default AND doubles to 128bit.

For gfortran, the equivalent options would be `-fdefault-real-8 -fdefault-double-8`

hattom:
Reproducer:
```
thay@toki04:~/src/c4s/orb5> ifort ./test_prec.F90 && ./a.out 
          32          64
thay@toki04:~/src/c4s/orb5> ifort -r8 ./test_prec.F90 && ./a.out 
          64          64
thay@toki04:~/src/c4s/orb5> gfortran ./test_prec.F90 && ./a.out 
          32          64
thay@toki04:~/src/c4s/orb5> gfortran -fdefault-real-8 ./test_prec.F90 && ./a.out 
          64         128
thay@toki04:~/src/c4s/orb5> gfortran -fdefault-real-8 -fdefault-double-8 ./test_prec.F90 && ./a.out 
          64          64
```
Tested outside of EB with GCC 7.5.0 and Intel 19.1.3.304
```
thay@toki04:~/src/c4s/orb5> cat test_prec.F90 
program test_prec
  implicit none
  real :: sx
  double precision :: dx

  print *, storage_size(sx), storage_size(dx)
end program test_prec
```

hattom:
n.b. nvfortran is consistent with Intel already:
```
thay@toki04:> nvfortran ./test_prec.F90 && ./a.out 
nvfortran-Warning-CUDA_HOME has been deprecated. Please, use NVHPC_CUDA_HOME instead.
           32           64
thay@toki04:> nvfortran -r8 ./test_prec.F90 && ./a.out 
nvfortran-Warning-CUDA_HOME has been deprecated. Please, use NVHPC_CUDA_HOME instead.
           64           64
```
(using 22.7)

hattom:
FWIW, looks like the correct Cray option would be `-s real64` (currently not set), but I'll avoid adding this, since I don't use Cray.

PR #4121, opened 2022-11-04
add `fdefault-double-8` to `r8` toolchain compiler option for GCC (to be consistent with Intel)

hattom:
-r8 for intel promotes reals => 64bit, but leaves doubles as 64bit.
-fdefault-real-8 for gfotran promotes reals => 64bit and promotes doubles to 128bit.
-fdefault-real-8 -fdefault-double-8 yields the same effect as intel's -r8

https://github.com/easybuilders/easybuild-framework/issues/4120 contains examples.

### DIFF
--- a/easybuild/toolchains/compiler/gcc.py
+++ b/easybuild/toolchains/compiler/gcc.py
@@ -54,7 +54,7 @@ class Gcc(Compiler):
     }
     COMPILER_UNIQUE_OPTION_MAP = {
         'i8': 'fdefault-integer-8',
-        'r8': 'fdefault-real-8',
+        'r8': ['fdefault-real-8', 'fdefault-double-8'],
         'unroll': 'funroll-loops',
         'f2c': 'ff2c',
         'loop': ['ftree-switch-conversion', 'floop-interchange', 'floop-strip-mine', 'floop-block'],
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -872,7 +872,7 @@ class ToolchainTest(EnhancedTestCase):
                 tc = self.get_toolchain('foss', version='2018a')
                 tc.set_options({opt: enable})
                 tc.prepare()
-                flag = ' '.join('-%s' % x for x in  tc.COMPILER_UNIQUE_OPTION_MAP[opt])
+                flag = ' '.join('-%s' % x for x in tc.COMPILER_UNIQUE_OPTION_MAP[opt])
                 for var in flag_vars:
                     flags = tc.get_variable(var)
                     if enable:
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -872,7 +872,11 @@ class ToolchainTest(EnhancedTestCase):
                 tc = self.get_toolchain('foss', version='2018a')
                 tc.set_options({opt: enable})
                 tc.prepare()
-                flag = ' '.join('-%s' % x for x in tc.COMPILER_UNIQUE_OPTION_MAP[opt])
+                flag = tc.COMPILER_UNIQUE_OPTION_MAP[opt]
+                if isinstance(flag, list):
+                    flag = ' '.join('-%s' % x for x in flag)
+                else:
+                    flag = '-%s' % flag
                 for var in flag_vars:
                     flags = tc.get_variable(var)
                     if enable:


hattom:
https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/toolchain.py#L870

the toochain r8 test is failing. Not completely sure I can easily see what's being tested.

jfgrimm:
@hattom I think it's assuming it will be a string, and is tripping up because it's now a list
```
AssertionError: False is not true : r8: True means -['fdefault-real-8', 'fdefault-double-8'] in -O2 -ftree-vectorize -march=native -fno-math-errno -fdefault-real-8 -fdefault-double-8
```

hattom:
Slightly further down in the same file there's
```
        for prec in precs:
            prec_flags[prec] = ' '.join('-%s' % x for x in Gcc.COMPILER_UNIQUE_OPTION_MAP[prec])
```
Should the above r8 test be changed to support also:  `' '.join('-%s' % x for x)` ?

jfgrimm:
> Slightly further down in the same file there's
> 
> ```
>         for prec in precs:
>             prec_flags[prec] = ' '.join('-%s' % x for x in Gcc.COMPILER_UNIQUE_OPTION_MAP[prec])
> ```
> 
> Should the above r8 test be changed to support also: `' '.join('-%s' % x for x)` ?

@hattom yes that should work

easybuild/toolchains/compiler/gcc.py
```diff
@@ -54,7 +54,7 @@ class Gcc(Compiler):
     }
     COMPILER_UNIQUE_OPTION_MAP = {
         'i8': 'fdefault-integer-8',
```
akesandgren, line 56: If the tests is changed the way it is then both 'r8' and 'i8' must be lists, for all compilers.
Probably better to fix the test to handle both pure strings and lists of strings.

jfgrimm:
hmm, that's not quite right :smile: 
```
AssertionError: False is not true : i8: True means -f -d -e -f -a -u -l -t -- -i -n -t -e -g -e -r -- -8 in -O2 -ftree-vectorize -march=native -fno-math-errno -fdefault-integer-8
```

easybuild/toolchains/compiler/gcc.py
jfgrimm: yes, good point

test/framework/toolchain.py
```diff
@@ -872,7 +872,7 @@ def test_misc_flags_unique_fortran(self):
                 tc = self.get_toolchain('foss', version='2018a')
                 tc.set_options({opt: enable})
                 tc.prepare()
-                flag = '-%s' % tc.COMPILER_UNIQUE_OPTION_MAP[opt]
+                flag = ' '.join('-%s' % x for x in tc.COMPILER_UNIQUE_OPTION_MAP[opt])
```
jfgrimm, line 875:
```suggestion
                flag = tc.COMPILER_UNIQUE_OPTION_MAP[opt]
                if isinstance(flag, list):
                    flag = ' '.join('-%s' % x for x in flag)
                else:
                    flag = '-%s' % flag
```

hattom:
> hmm, that's not quite right 😄
> 
> ```
> AssertionError: False is not true : i8: True means -f -d -e -f -a -u -l -t -- -i -n -t -e -g -e -r -- -8 in -O2 -ftree-vectorize -march=native -fno-math-errno -fdefault-integer-8
> ```

The good news is that I tested this already locally.
The bad news is that I tested it with 1 character strings 🤦

hattom:
Is this sort of error normal?
```
== fetching files...
== ... (took 13 secs)
== FAILED: Installation ended unsuccessfully (build directory: /scratch/build/binutils/2.37/system-system): build failed (first 300 chars): Unexpected error occurred when trying to download https://ftpmirror.gnu.org/gnu/binutils/binutils-2.37.tar.gz to /scratch/sources/b/binutils/binutils-2.37.tar.gz: 'Failed to write to /scratch/sources/b/binutils/binutils-2.37.tar.gz: The read operation timed out' (took 13 secs)
== Results of the build can be found in the log file(s) /scratch/tmp/eb-rt_hozr7/easybuild-binutils-2.37-20221104.141736.CKNar.log
FATAL:   While performing build: while running engine: exit status 1
Error: Process completed with exit code 1.
```

branfosj:
We do see odd errors from time to time. I've kicked the failed tests to run again.

jfgrimm (approved):
LGTM

jfgrimm:
Going in, thanks @hattom!

PR #4125, opened 2022-11-16
add sanity_check_load_module method to EasyBlock to provide control over when module is loaded during sanity check step

boegel:
Some refactoring that allows making `PythonPackage` compatible with `--sanity-check-only` (cfr. https://github.com/easybuilders/easybuild-easyblocks/issues/2745)


akesandgren (approved):
LGTM

PR #4129, opened 2022-11-27
correct uses of :param name: and :return: in docstrings

branfosj:
One of the markdown API document generators I am trying:
* blows up on `:param:`
* does not recognise `:params ...`
* blows up on `:param name description` (missing `:` after the parameter name)
* blows up on `:return ...`


ocaisa (approved):
LGTM

ocaisa (approved):
Still LGTM

PR #4131, opened 2022-11-28
Avoid exception in `get_gpu_info`

Flamefire:
Currently the function calls `run_cmd` which throws on error AND checks the exit code which is redundant and causes a log message `"ERROR EasyBuild crashed with an error ..."` to be logged on error which is confusing as e.g it is VERY unlikely both `nvidia-smi` and `rocm-smi` are on the system.
So check for existance first and suppress output and error checking of `run_cmd`.


branfosj (approved):

branfosj:
Going in, thanks @Flamefire!

PR #4133, opened 2022-11-28
Add checksums.json support for Github interface

mboisson:
This uses a workflow very similar to how patches are handled with the github integration, with the difference that instead of copying the patches in place, it merges the `checksums.json` with the existing one if it does exist already


mboisson:
@boegel, @bartoldeman, I'ld like some initial feedback on this. I am not sure how I can test such a PR (I don't see tests for github integration itself, other than each sub functions)

PR #4135, opened 2022-12-06
auto-enable use of oneAPI C/C++ compilers for intel-compilers >= 2022.2.0

boegel:


easybuild/toolchains/compiler/intel_compilers.py
```diff
@@ -63,15 +70,45 @@ def _set_compiler_vars(self):
     def set_variables(self):
         """Set the variables."""
 
+        oneapi = False
+
+        # auto-enable use of oneAPI C/C++ compilers for sufficiently recent versions of Intel compilers
+        comp_ver = self.get_software_version(self.COMPILER_MODULE_NAME)[0]
+        print(comp_ver)
```
branfosj, line 77: remove this line

easybuild/toolchains/compiler/intel_compilers.py
```diff
@@ -63,15 +70,45 @@ def _set_compiler_vars(self):
...
+
         if self.options.get('oneapi', False):
+            oneapi = True
             self.COMPILER_CXX = 'icpx'
             self.COMPILER_CC = 'icx'
             self.COMPILER_F77 = 'ifx'
             self.COMPILER_F90 = 'ifx'
             self.COMPILER_FC = 'ifx'
+
+        # if both 'oenapi' and 'oneapi_*' are set, the latter are ignored
```
branfosj, line 92: `# if both 'oenapi'` should be `# if both 'oneapi'`

branfosj (changes requested):
Minor changes while we discuss if we should merge the C and C++ options into one.

branfosj (approved):

PR #4148, opened 2022-12-21
trim CI test configurations to avoid hitting GitHub rate limits

boegel:
Since we also started testing with #4092, we easily hit the GitHub rate limits due to having too many test configurations, so let's trim things down a bit by excluding less commonly used configurations (like Lmod + Tcl syntax)


akesandgren (approved):
LGTM

PR #4149, opened 2022-12-21
partially skip test_det_copy_ec_specs if no GitHub token is available

boegel:


test/framework/easyconfig.py
```diff
@@ -4408,6 +4413,10 @@ def test_det_copy_ec_specs(self):
             self.assertEqual(paths, args[:-1])
             self.assertEqual(target_path, args[-1])
 
+        if self.skip_github_tests:
+            print("Skipping test_det_copy_ec_specs using --from-pr, no GitHub token available?")
+            return
```
Flamefire, line 4418:
```suggestion
            self.skipTest("Not using --from-pr, no GitHub token available?")
```

See my longstanding work at #3790
```

Flamefire (changes requested):

akesandgren (approved):
LGTM

test/framework/easyconfig.py
Flamefire: Agreed

Flamefire (approved):

akesandgren:
Going in, thanks @boegel!

PR #1644, opened 2016-02-29
added option to EB to dump information about EasyConfig files

DirkdeDraak:
An option do dump Easyconfig file information in a standardized way (JSON/python structure) as developed during the EasyBuild user community session back in January 2016.

See what you think you can use.
I'm not planning on developing this feature any further.

regards,
 Fokko

PR #4117, opened 2022-10-31
add support for using --output-format=md (MarkDown)

boegel:
* [x] `eb --avail-cfgfile-constants --output-format md`
* [x] `eb --avail-easyconfig-constants --output-format md`
* [x] `eb --avail-easyconfig-licenses --output-format md`
* [x] `eb --avail-easyconfig-params --output-format md`
* [x] `eb --avail-easyconfig-templates --output-format md`
* [x] `eb --avail-toolchain-opts ... --output-format md`
* [x] `eb --list-easyblocks --output-format md`
* [x] `eb --list-software --output-format md`
* [x] `eb --list-toolchains --output-format md`
* [x] `eb --help-md`
* [x] `gen_easyblocks_overview_md`

PR #4152, opened 2022-12-29
add support for --list-software --output-format=json

APN-Pucky:
This could be a first step on getting easybuild on repology as described here:

* https://github.com/repology/repology-updater/issues/705

Next steps would be:

- [ ] handle name collisions
- [ ] serving the json to repology on some server/page (since they don't want to run third party software).

Possibly related:
* https://github.com/easybuilders/easybuild-framework/pull/1644

Please let me know, if the (estimated) effort to get easybuild on repology is worth it. 
The json format might be useful either way.

Cheers,
APN

- [x] eb --avail-cfgfile-constants --output-format json
- [x] eb --avail-easyconfig-constants --output-format json
- [x] eb --avail-easyconfig-licenses --output-format json
- [x] eb --avail-easyconfig-params --output-format json
- [x] eb --avail-easyconfig-templates --output-format json
- [x] eb --avail-toolchain-opts ... --output-format json
- [x] eb --list-easyblocks --output-format json
- [x] eb --list-software --output-format json
- [x] eb --list-toolchains --output-format json
~~- [ ] eb --help json~~
- [x] gen_easyblocks_overview_json

### DIFF
--- a/easybuild/tools/docs.py
+++ b/easybuild/tools/docs.py
@@ -39,6 +39,7 @@ Authors:
 import copy
 import inspect
 import os
+import json
 from easybuild.tools import LooseVersion
 
 from easybuild.base import fancylogger
@@ -75,6 +76,7 @@ SIMPLE = 'simple'
 FORMAT_MD = 'md'
 FORMAT_RST = 'rst'
 FORMAT_TXT = 'txt'
+FORMAT_JSON = 'json'
 
 
 def generate_doc(name, params):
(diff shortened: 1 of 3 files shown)


boegel:
@APN-Pucky It would definitely be interesting to get EasyBuild into repology!

High-level, the changes being made here look like a good start, but:

* tests should be enhanced as well to make sure that `--list-software --output-format=json` works (and keeps working) as intended;
* add `json` as a supported output format makes sense, but implies that all other options that also take into account the selected output format should be updated to support producing JSON. https://github.com/easybuilders/easybuild-framework/pull/4117 shows that this is no small effort, but maybe it makes sense to somehow only support `--output-format=json` for `--list-software`, without causing trouble for the other options;
* merge conflicts will appear once https://github.com/easybuilders/easybuild-framework/pull/4117 gets merged (although they'll likely be easy to resolve);

APN-Pucky:
1. resolved conflict with md format
2. added the other json format functions with `NotImplementedError` exceptions. They are also checked in the test, so that when I or someone else extends the json format one gets reminded of the tests.
3. added `--list-software --output-format=json` tests

APN-Pucky:
Seems to be only some newlines. While it worked for me I'll just replace the failed newlines with those that worked in the CI.

APN-Pucky:
So finally, after sorting and tuning the json for compatibility across different python versions, my CI runs through too... (https://github.com/APN-Pucky/easybuild-framework/actions/runs/7114626894)

boegel (approved):
lgtm

PR #4156, opened 2023-01-04
make vstring/version variables in LooseVersion read-only

Flamefire:
The exposed "public" members `.version` and `.vstring` must not be modified from outside as they belong together. Hence put them into properties.

I was also thinking about the case of `LooseVersion(None)` which can easily happen with `LooseVersion(get_software_version('Bazel'))`. One approach was to compare that equal to `None` or `LooseVersion(None)` and "less" otherwise. But that might be surprising and hence better/safer is `if LooseVersion(...).version is None`

I hence added that to the test suite to show an example and kept my WIP commit in the history if we later decide otherwise.

@boegel This should go into the next release which adds this class.

### DIFF
--- a/easybuild/base/fancylogger.py
+++ b/easybuild/base/fancylogger.py
@@ -341,10 +341,7 @@ class FancyLogger(logging.getLoggerClass()):
         loose_cv = LooseVersion(cur_ver)
         loose_mv = LooseVersion(max_ver)
 
-        loose_cv.version = loose_cv.version[:depth]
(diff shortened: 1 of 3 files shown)


Flamefire:
There was a use case that caused now test failures: `loose_cv.version = loose_cv.version[:depth]` in `log.deprecated` to only consider some parts of the version. I fixed that by comparing the list directly. I guess that works with the caveat that mixed-types in the components are no longer working. But I doubt we'd use that-

boegel (approved):
lgtm

PR #4157, opened 2023-01-04
add definition for nvofbf toolchain

Micket:
Completely untested

cf. https://github.com/easybuilders/easybuild-easyconfigs/pull/16724


boegel (approved):
I did a quick test with building HPL with `nvofbf/2022.07` on top of this, works fine, so good to go!

PR #4159, opened 2023-01-04
Fix `to_checksums` with `None` values in dicts and recursion

### LINKED ISSUE #4142: checksum dict with a checksum value of None doesn't work
This failes with --check-contrib, -D, and --new-pr
```
checksums = [
    {'v2.12.4.tar.gz': '31c1f4f91efbdb74619cebb36f57f999d6f1a57bb6f87b13e60d21e670c38f68'},
    {'CCCoreLib-20220714.tar.gz': None},
]
```

--check-contrib and -D fails on the above with
```
ERROR: Failed to process easyconfig /afs/hpc2n.umu.se/home/a/ake/Projects/git-repos/easybuild/easybuild_common/custom/easyconfigs/c/CloudCompare/CloudCompare-2.12.4-foss-2021b.eb: Converting type of [{'v2.12.4.tar.gz': '31c1f4f91efbdb74619cebb36f57f999d6f1a57bb6f87b13e60d21e670c38f68'}, {'CCCoreLib-20220714.tar.gz': None}] (<class 'list'>) to (<class 'list'>, (('elem_types', (<class 'str'>, <class 'tuple'>, (<class 'dict'>, (('elem_types', (<class 'str'>,)), ('key_types', (<class 'str'>,)))), (<class 'list'>, (('elem_types', (<class 'str'>, <class 'tuple'>, (<class 'dict'>, (('elem_types', (<class 'str'>,)), ('key_types', (<class 'str'>,)))))),)))),)) using <function to_checksums at 0x7fb0e86b4ee0> failed: 'NoneType' object is not iterable
```

Flamefire:
Having a `'src': None` entry in a dict for checksums is as valid as having a `None` entry directly in the list. However the current function didn't handle it and crashed.
Fix that as well as a few corner cases especially in the recursive case by introducing a new function for handling checksum entries in the checksum list and limiting the recursiveness.

It will now throw an `EasyBuildError` similar to e.g. `to_dependency` instead of trying to continue and fail at a random other place.   
I extended the testcases to cover the original issue (None as a value in a dict) and some more correct and error cases

Please carefully review and check if you can come up with any valid or invalid `checksums` entry that might break using the new function. Especially regarding the YEB compatibility this was pretty hard as for lists of strings you cannot know if those are 2 checksums without a type that should both match or a checksum and type. In EC-files (Python) one can use a tuple for that but then again it is hard to tell if that tuple is a checksum and type or 2 alternative checksums.

Fixes #4142

- [ ] Requires #4164 (contained here)
- [ ] https://github.com/easybuilders/easybuild-docs/pull/104

### DIFF
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -505,33 +505,51 @@ def to_dependencies(dep_list):
     return [to_dependency(dep) for dep in dep_list]
 
 
+def _to_checksum(checksum, list_level=0, allow_dict=True):
+    """Ensure the correct element type for each checksum in the checksum list"""
+    # each entry can be:
+    # * None (indicates no checksum)
+    # * a string (MD5, SHA256, ... checksum)
+    # * a list or tuple with 2 elements: checksum type + checksum value
+    # * a list or tuple of checksums (i.e. multiple checksums for a single file)
+    # * a dict (filename to checksum mapping)
+    if checksum is None or isinstance(checksum, string_type):
+        return checksum
+    elif isinstance(checksum, (list, tuple)):
+        if len(checksum) == 2 and isinstance(checksum[0], string_type) and isinstance(checksum[1], (string_type, int)):
+            # 2 elements so either:
+            #  - a checksum tuple (2nd element string or int)
+            #  - 2 alternative checksums (tuple)
+            #  - 2 checksums that must each match (list)
+            # --> Convert to tuple only if we can exclude the 3rd case
+            if not isinstance(checksum[1], string_type) or list_level > 0:
+                return tuple(checksum)
+            else:
+                return checksum
+        elif list_level < 2:
+            # Alternative checksums or multiple checksums for a single file
+            # Allowed to nest (at most) 2 times, e.g. [[[type, value]]] == [[(type, value)]]
+            # None is not allowed here
+            if any(x is None for x in checksum):
+                raise ValueError('Unexpected None in ' + str(checksum))
+            if isinstance(checksum, tuple) or list_level > 0:
+                # When we already are in a tuple no further recursion is allowed -> set list_level very high
+                return tuple(_to_checksum(x, list_level=99, allow_dict=allow_dict) for x in checksum)
+            else:
+                return list(_to_checksum(x, list_level=list_level+1, allow_dict=allow_dict) for x in checksum)
+    elif isinstance(checksum, dict) and allow_dict:
+        return {key: _to_checksum(value, allow_dict=False) for key, value in checksum.items()}
+
+    # Not returned -> Wrong type/format
+    raise ValueError('Unexpected type of "%s": %s' % (type(checksum), str(checksum)))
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyconfig/types.py
```diff
@@ -467,33 +467,39 @@ def to_dependencies(dep_list):
...
+    # * a string (MD5, SHA256, ... checksum)
+    # * a list or tuple with 2 elements: checksum type + checksum value
+    # * a list or tuple of checksums (i.e. multiple checksums for a single file)
+    # * a dict (filename to checksum mapping)
+    if checksum is None or isinstance(checksum, string_type):
+        return checksum
+    elif isinstance(checksum, (list, tuple)):
+        if len(checksum) == 2 and isinstance(checksum[0], string_type) and isinstance(checksum[1], (string_type, int)):
+            # 2 elements => a checksum tuple (2nd element string or int) or 2 alternative checksums
+            return tuple(checksum)
```
akesandgren, line 483: Does this only allow 2 "multiple checksums"? There could very well be more than 2 when pipy.org messes up things repeatedly

Flamefire:
I traced the failure to a difference in YEB and EB formats and the validator which seemingly has a bug:

The `toy-0.0.*` contains a single source with alternative checksums:

```
checksums: [[
    'be662daa971a640e40be5c804d9d7d10',  # default [MD5]
    '44332000aa33b99ad1e00cbd1a7da769220d74647060a10e807b916d73ea27bc',  # default (SHA256)
    ['adler32', '0x998410035'],
    ['crc32', '0x1553842328'],
    ['md5', 'be662daa971a640e40be5c804d9d7d10'],
    ['sha1', 'f618096c52244539d0e89867405f573fdb0b55b0'],
    ['size', 273],
]]
```

The difference is:
- YEB: `[['be...', '44...', [...], [...], [...], [...], [...]]]``
- EC: `[['be...', '44...', (...), (...), (...), (...), (...)]]`

I.e. the EC correctly uses tuples while the YEB uses lists. However actually both are wrong and need to be converted because alternative checksums must be specified as a tuple: https://github.com/easybuilders/easybuild-framework/blob/8ba299fde496a5b18fca8babee18da71a0c018d2/easybuild/tools/filetools.py#L1291

However the current definition of the checksums type is:
```
CHECKSUM_LIST = (list, as_hashable({'elem_types': [str, tuple, STRING_DICT]}))
CHECKSUMS = (list, as_hashable({'elem_types': [str, tuple, STRING_DICT, CHECKSUM_LIST]}))
```

I.e. the we wrongly allow a list where we expect a tuple.
Furthermore this doesn't cover all possible use cases (e.g. dicts with `None` entries or alternatives) and allows wrong entries, e.g. tuples of any types. As I don't see that it is possible to express all valid entries I'd purposely restrict this to less than is possible and make use of `to_checksums` to do the final verification.

Flamefire:
Closed for https://github.com/easybuilders/easybuild-framework/pull/4579

PR #4164, opened 2023-01-06
Fix the checksum type check

Flamefire:
The `None` case was missed and due to the unrestricted `tuple` elem_type it may return valid for actually invalid entries.
So restrict that beeing overly cautious so it may wrongly return invalid.   
But in that case the conversion function will be called which can do more elaborate verification.
    
Add test checking for None in checksums.

Extracted from #4159, see https://github.com/easybuilders/easybuild-framework/pull/4159#issuecomment-1373429332

Also required for #4142 as the check now correctly handles a `None` value in the dict

Requires resolution of
- [ ] https://github.com/easybuilders/easybuild-docs/pull/104

### DIFF
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -608,19 +608,33 @@ SANITY_CHECK_PATHS_DICT = (dict, as_hashable({
 }))
 # checksums is a list of checksums, one entry per file (source/patch)
 # each entry can be:
+# None
 # a single checksum value (string)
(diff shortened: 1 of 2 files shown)


Flamefire:
Converted to draft to first specify how checksums especially "None" should work

Flamefire:
Can this (and https://github.com/easybuilders/easybuild-docs/pull/104) be addressed? It is also a dependency of https://github.com/easybuilders/easybuild-framework/pull/4159

Flamefire:
Closed for https://github.com/easybuilders/easybuild-framework/pull/4578

PR #4166, opened 2023-01-06
fix use of `locate` in `check_os_dependency`

### LINKED ISSUE #4165: check_os_dependency wrongly succeeds using `locate`
`locate --regexp "/nosuchosdependency$"` returns with an exit code of zero for me making the `check_os_dependency` wrongly return True

Maybe we can use `-c` and check the output.

Flamefire:
`locate` may not set an error exit code when it does not find a dependency so the function may wrongly succeed.
Use the `-c` option to get a number of matches and check that this is more than zero (and an integer) in addition to the exit code.

Fixes #4165

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -873,9 +873,14 @@ def check_os_dependency(dep):
 
         # try locate if it's available
         if not found and which('locate'):
-            cmd = 'locate --regexp "/%s$"' % dep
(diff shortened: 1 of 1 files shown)


Flamefire:
I don't think so. It depends on how the OS `locate` behaves (if that even is used) so we would need to mock it which IMO isn't wort the effort.

boegel (approved):
lgtm

PR #4168, opened 2023-01-07
autogenerated docs formatting

branfosj:


easybuild/tools/docs.py
```diff
@@ -315,7 +315,7 @@ def avail_easyconfig_params_md(title, grouped_params):
         values = [grouped_params[grpname][key] for key in keys]
         table_values = [
             ['`%s`' % name for name in keys],  # parameter name
-            [x[0] for x in values],  # description
+            [x[0].replace('<', '[').replace('>', ']') for x in values],  # description
```
boegel, line 318: Hmm, shouldn't we escape `<` and `>` instead, rather than replace them with `[` and `]`?

boegel (changes requested):

easybuild/tools/docs.py
branfosj: done in dccb372

boegel (approved):
lgtm

PR #4171, opened 2023-01-09
add VERSION to easybuild.easyblocks namespace in test sandbox

Flamefire:
Extracted from #3790 and required for some tests (which happen to work now due to unrelated issues)


boegel (approved):
lgtm

PR #4172, opened 2023-01-09
Also apply filter to asyncprocess test

Flamefire:
Extracted from #3790 

Allows to run the test(s) as usual by filtering, i.e. only run the single test.


boegel (approved):
lgtm

PR #4175, opened 2023-01-09
filter out deprecation warnings for platform.dist in get_os_version and platform.linux_distribution in get_os_name

Flamefire:
Those 2 are deprecated and eventually removed which we do handle, so ignore the deprecation warnings.

Extracted from #3790


boegel (approved):
lgtm

PR #4176, opened 2023-01-09
update CI workflows to run 'apt-get update' if installation of packages via 'apt-get install' failed, likely due to stale apt cache

Flamefire:
Sometimes we need to update the APT cache before installing to avoid:

    E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/d/dh-autoreconf/dh-autoreconf_17_all.deb  503  Service Unavailable [IP: 52.252.75.106 80]
    E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

### DIFF
--- a/.github/workflows/bootstrap_script.yml
+++ b/.github/workflows/bootstrap_script.yml
@@ -75,7 +75,7 @@ jobs:
         # Avoid apt-get update, as we don't really need it,
         # and it does more harm than good (it's fairly expensive, and it results in flaky test runs)
         if ! sudo apt-get install $APT_PKGS; then
-          # Try to update cache, then try again
+          # Try to update cache, then try again to resolve 404s of old packages
           sudo apt-get update -yqq || true
           sudo apt-get install $APT_PKGS
         fi
--- a/.github/workflows/bootstrap_script.yml
+++ b/.github/workflows/bootstrap_script.yml
@@ -69,11 +69,17 @@ jobs:
 
     - name: install OS & Python packages
       run: |
-        # disable apt update, we don't really need it,
-        # and it does more harm than good (it's fairly expensive, and it results in flaky test runs)
-        # sudo apt update
         # for modules tool
-        sudo apt install lua5.2 liblua5.2-dev lua-filesystem lua-posix tcl tcl-dev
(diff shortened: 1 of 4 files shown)


jfgrimm:
@Flamefire is there a reason you're reverting the switch from `apt-get` to `apt` (PR #4064)?

.github/workflows/bootstrap_script.yml
```diff
@@ -69,11 +69,17 @@ jobs:
...
       run: |
-        # disable apt update, we don't really need it,
-        # and it does more harm than good (it's fairly expensive, and it results in flaky test runs)
-        # sudo apt update
         # for modules tool
-        sudo apt install lua5.2 liblua5.2-dev lua-filesystem lua-posix tcl tcl-dev
+        APT_PKGS="lua5.2 liblua5.2-dev lua-filesystem lua-posix tcl tcl-dev"
+
+        # Avoid apt-get update, as we don't really need it,
+        # and it does more harm than good (it's fairly expensive, and it results in flaky test runs)
```
boegel, line 76: @Flamefire Can you also extend the comment to explain why we do try to `apt-get update` if `apt-get install` failed, since it seems to go against what the current comment states...

boegel (changes requested):

.github/workflows/bootstrap_script.yml
Flamefire: I had already changed that comment to "avoid" but agree to be a bit more explicit why we then do it

Flamefire:
> @Flamefire is there a reason you're reverting the switch from `apt-get` to `apt` (PR #4064)?

Using `apt` doesn't seem to be what you should be doing: It prints something like "unstable CLI, don't use in scripts" or so. At the least it is noise in the CI logs hence the switch

boegel:
> > @Flamefire is there a reason you're reverting the switch from `apt-get` to `apt` (PR #4064)?
> 
> Using `apt` doesn't seem to be what you should be doing: It prints something like "unstable CLI, don't use in scripts" or so. At the least it is noise in the CI logs hence the switch

At some point we were probably running with Ubuntu 22.04 (because we were using `ubuntu-latest`), where `apt` does make sense, but not anymore since #4064 (we probably overlooked the "unstable CLI" messages when doing that)

PR #4180, opened 2023-01-16
accept single source as dict and single checksum in check_checksums_for

Flamefire:
Duplicate the behavior of `fetch_sources` in `check_checksums_for`


boegel (approved):
lgtm

PR #4182, opened 2023-01-18
only give read permissions in GitHub Actions workflows

boegel:
By default, CI workflows get full permissions (incl. write) to the repo, there's absolutely no need for this in our case.

cfr. https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs


ocaisa:
Yeah, this stuff is important. One thing to note is that you can set default permissions for the organisation (but it is also good to be explicit)

ocaisa:
See last setting in https://github.com/organizations/easybuilders/settings/actions

PR #4186, opened 2023-01-19
correctly restore sys.path in tests

Flamefire:
We need to modify the list not replace it.

From #3790

### DIFF
--- a/test/framework/config.py
+++ b/test/framework/config.py
@@ -499,7 +499,9 @@ class EasyBuildConfigTest(EnhancedTestCase):
 
         # prepend path to test easyconfigs into Python search path, so it gets picked up as --robot-paths default
         orig_sys_path = sys.path[:]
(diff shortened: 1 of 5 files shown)


Flamefire:
Flaky test as it seems to fail to download a file. Maybe we should have that already somewhere for the tests to find:

ERROR: Build of /usr/local/easybuild/easyconfigs/b/binutils/binutils-2.37.eb failed (err: "build failed (first 300 chars): Couldn't find file binutils-2.37.tar.gz anywhere, and downloading it didn't work either... Paths attempted (in order): /usr/local/easybuild/easyconfigs/b/binutils/b/binutils/binutils-2.37.tar.gz, /usr/local/easybuild/easyconfigs/b/binutils/binutils/binutils-2.37.tar.gz, /usr/local/easybuild/easyconf")

boegel (approved):
lgtm

PR #4187, opened 2023-01-19
suppress output of --skip-test-step test

Flamefire:
This test prints to stderr (`print_warning` is used) which we can ignore or better: Test for

From #3790


boegel (approved):
lgtm

PR #4188, opened 2023-01-19
Use skip markers in tests

Flamefire:
From #3790 

We can use the more conventional `@unittest.skipIf` or `self.skipTest` markers to skip tests instead of C&P of conditions which might be wrong (some were)
An approach taken from IIRC PyTorch is to define own decorators `requiresFoo` as a wrapper around a specific `skipIf` which is reusable and easier to understand.

Only issue I can see is with the change to CI to ignore all skipped tests as now that I look at it previously we were able to selectively ignore some skipped tests.
So maybe we should not allow any skipped tests on CI which usually means installing additional Python packages. And for the tests requiring a github token we then need to resort back to the old approach but maybe we can still use a decorator here

Split into
- #5264 
- #5263 
- #5262 
- #5265 
- #5266


.github/workflows/unit_tests.yml
```diff
@@ -212,7 +212,7 @@ jobs:
           # run test suite
           python -O -m test.framework.suite 2>&1 | tee test_framework_suite.log
           # try and make sure output of running tests is clean (no printed messages/warnings)
-          IGNORE_PATTERNS="no GitHub token available|skipping SvnRepository test|requires Lmod as modules tool|stty: 'standard input': Inappropriate ioctl for device|CryptographyDeprecationWarning: Python 3.[56]|from cryptography.*default_backend|CryptographyDeprecationWarning: Python 2|from cryptography.utils import int_from_bytes|Blowfish|GC3Pie not available, skipping test"
+          IGNORE_PATTERNS="no GitHub token available|skipping SvnRepository test|requires Lmod as modules tool|stty: 'standard input': Inappropriate ioctl for device|CryptographyDeprecationWarning: Python 3.[56]|from cryptography.*default_backend|CryptographyDeprecationWarning: Python 2|from cryptography.utils import int_from_bytes|Blowfish|GC3Pie not available, skipping test|\.s+"
```
boegel, line 215: Adding `\s+` to `IGNORE_PATTERNS` defeats the whole purpose of it?

Flamefire: Pythons unittest will print a literal dot for each test executed and a literal s for each test skipped.

I did add `\.s+` not `\s+` to the patterns, i.e. I want to ignore a dot followed by one or more "s" as otherwise those skipped tests would make CI fail. See my last paragraph in the PR description about a discussion starter on this.

boegel: Oh, right, I totally misinterpreting this, sorry... Makes sense

boegel:
> Only issue I can see is with the change to CI to ignore all skipped tests as now that I look at it previously we were able to selectively ignore some skipped tests. So maybe we should not allow any skipped tests on CI which usually means installing additional Python packages. And for the tests requiring a github token we then need to resort back to the old approach but maybe we can still use a decorator here

@Flamefire How about only allowing `\.s+` in the output when running the test suite in a PR?
Because only then should we allow skipped tests (because the GitHub token is missing)...

So we should conditionally extend `IGNORE_PATTERNS` with `|\.s+` if the test suite is being triggered by a `push` to `develop` branch (I think)...

Flamefire:
> So we should conditionally extend `IGNORE_PATTERNS` with `|\.s+` if the test suite is being triggered by a `push` to `develop` branch (I think)...

I guess you meant to insert a NOT somewhere as no tests should be skipped on push to develop.

> How about only allowing `\.s+` in the output when running the test suite in a PR?
> Because only then should we allow skipped tests (because the GitHub token is missing)...

Makes sense, however we should be more selective here. I hence added a dedicated decorator for that which doesn't produce output for PRs (i.e. silently succeeds) and make use of `$FORCE_EB_GITHUB_TESTS` to ensure github tests are run when the token is expected to be there.

Flamefire:
Resolved

Flamefire:
Ok, the other skip sign was caused by an incomplete conflict resolution which was impossible to find with the current approach.   
I force-pushed to fixup the conflict resolution failure directly

 I added a list of skipped tests at the end. The expected ones should be marked with `skip_silentCI*` (used by e.g. `requires_github_access`). Unexpected ones will show up in this list in addition to the "s" chars in the short output.

We could on CI completely ignore ".s." (already done) and (instead of `skip_silentCI`) filter the "sometimes expected" failures in the CI config (`$IGNORE_PATTERNS`)

Flamefire:
Rebased to fix

> Sorry for dragging this on so long, it's a huge PR to review...

I know sorry for that. If you view the diff without whitespace changes it gets better at least

Flamefire:
@boegel I rebased this one and split it into 5 smaller PRs which *should* contain all changes from this one, but they conflict with each other so I'm keeping this open to merge at some point

Would be great to transition to the skip-approach

PR #4190, opened 2023-01-20
fix flaky test_search_file due to accidental matches for `.*C++` search pattern

Flamefire:
`self.test_prefix` can contain log files which may have names that match the pattern `search_file` searches for resulting in unexpected additional matches failing the test. E.g.:
'eb-test-0hsdcdfm.log' contains "CDF" -> Match

See e.g. https://github.com/easybuilders/easybuild-framework/actions/runs/3956357537/jobs/6775489286

### DIFF
--- a/test/framework/filetools.py
+++ b/test/framework/filetools.py
@@ -2507,13 +2507,13 @@ class FileToolsTest(EnhancedTestCase):
         self.assertTrue(hits[4].endswith('/hwloc-1.11.8-GCC-7.3.0-2.30.eb'))
 
         # also test case-sensitive searching
-        var_defs, hits_bis = ft.search_file([test_ecs], 'HWLOC', silent=True, case_sensitive=True)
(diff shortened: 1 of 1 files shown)


Flamefire:
Hitting a rate limit here. See https://github.com/easybuilders/easybuild-framework/pull/4192 for an approach for mitigation.

test/framework/filetools.py
```diff
@@ -2507,13 +2507,13 @@ def test_search_file(self):
         self.assertTrue(hits[4].endswith('/hwloc-1.11.8-GCC-7.3.0-2.30.eb'))
 
         # also test case-sensitive searching
-        var_defs, hits_bis = ft.search_file([test_ecs], 'HWLOC', silent=True, case_sensitive=True)
+        var_defs, hits_ignore_case = ft.search_file([test_ecs], 'HWLOC', silent=True, case_sensitive=True)
```
boegel: `ignore_case` and `case_sensitive=True` disagree?

boegel (approved):
lgtm

PR #4192, opened 2023-01-20
reduce number of CI jobs by testing for Lua and Tcl module syntax in a single CI job

Flamefire:
We run into rate limits installing the dependencies. We can reduce the number of CI jobs to almost half by running the tests in a loop for each supported module syntax.

See e.g. https://github.com/easybuilders/easybuild-framework/actions/runs/3965853121/jobs/6796017484

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -103,16 +103,11 @@ jobs:
         # and are only run after the PR gets merged
         GITHUB_TOKEN: ${{secrets.CI_UNIT_TESTS_GITHUB_TOKEN}}
       run: |
-        # don't install GitHub token when testing with Lmod 7.x or non-Lmod module tools,
-        # and only when testing with Lua as module syntax,
-        # to avoid hitting GitHub rate limit;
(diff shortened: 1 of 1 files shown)


boegel:
Each test suite runs takes about 15min, if we run the test suite twice in a CI jobs, don't we risk effectively doubling the waiting time for testsuite runs in all configurations to complete?

Flamefire:
> Each test suite runs takes about 15min, if we run the test suite twice in a CI jobs, don't we risk effectively doubling the waiting time for testsuite runs in all configurations to complete?

Yes, **IF** formerly the jobs ran in parallel. But IIRC the free GHA plan has a limit on the number of parallel jobs which I think we still run into: I mean look at the number of checks the are run.

So in practice I'd say it even reduces the waiting time because the common stuff (installing dependencies) is only done once instead of twice. Furthermore jobs failing due to hitting rate limits IMO increase the time by much more or may erroneously completely block a PR

Flamefire:
Looks good so far (longest run is 37min) but required checks need updating for/after this.

boegel (approved):

PR #4193, opened 2023-01-21
add pre/post extension hook (triggered before/after individual extension installations)

### LINKED ISSUE #4191: feature request: Add pre/post extension hooks.
## Problem description

Currently it seems like the existing hooks cannot be used to amend configuration on extensions within an Easyconfig, e.g. for amending Perl extension `preconfigopts`.

I am currently wanting to override some settings on the SSLeay extension for Perl Easyconfigs which I don't think is directly achievable. I'm attempting to force dependency on base OS OpenSSL libraries which are versioned below the Easybuild OpenSSL configs for various toolchains (Centos 7 with no approval for EPEL OpenSSL v1.1 :cry: )

## Proposed solution

Add some hooks which execute in the post and pre context of extension installations.

## tl;dr

Please add pre/post extension hooks.

boegel:
While adding `EXTENSION_INSTALL` to `HOOKS_NAMES`, I noticed that the list of hooks wasn't being printed in order of being triggered, so I fixed that too...

fixes #4191


lexming (commented):
Looks good, my only concern is that `pre_extension_hook` and `pre_extensions_hook` are too similar. One typo and we will be executing code where we don't want to without any error. Following the name of the global, what about renaming `pre_extension_hook` to `pre_installextension_hook` ?

test/framework/options.py
```diff
@@ -659,6 +659,66 @@ def run_test(custom=None, extra_params=[], fmt=None):
...
+            "	pre_build_hook",
+            "	post_build_hook",
+            "	pre_test_hook",
+            "	post_test_hook",
+            "	pre_install_hook",
+            "	post_install_hook",
+            "	pre_extensions_hook",
+            "	pre_extension_hook",
+            "	post_extension_hook",
+            "	post_extensions_hook",
```
lexming, line 701: This test needs the name change as well
```suggestion
            "	pre_extensions_hook",
            "	pre_single_extension_hook",
            "	post_single_extension_hook",
            "	post_extensions_hook",
```

lexming (changes requested):

test/framework/options.py
boegel: fixed in 6a46414a8 (with force push)

lexming (approved):
LGTM

lexming:
Going in, thanks @boegel !

PR #4196, opened 2023-01-26
enforce absolute paths as start dir of extensions

lexming:
Currently, `self.start_dir` in regular packages is always an absolute path. So it can be easily used in easyblocks regardles of `cwd` in the same way as `self.installdir` and `self.builddir`.

However, `self.start_dir` of extensions is either `None` or the path set in `self.cfg['start_dir'] of the extensions, which is commonly a relative path.

This PR updates `ExtensionEasyBlock._set_start_dir` to always define `self.start_dir` as an absolute path. Standardizing its form across packages and extensions.

Breaking changes:
* breaks `RPackages` as it prepends stuff to `self.start_dir`

### DIFF
--- a/easybuild/framework/extensioneasyblock.py
+++ b/easybuild/framework/extensioneasyblock.py
@@ -100,18 +100,30 @@ class ExtensionEasyBlock(EasyBlock, Extension):
         self.ext_dir = None  # dir where extension source was unpacked
 
     def _set_start_dir(self):
-        """Set value for self.start_dir
+        """Set absolute path of self.start_dir similarly to EasyBlock.guess_start_dir
 
-        Uses existing value of self.start_dir if it is already set and exists
-        otherwise self.ext_dir (path to extracted source) if that is set and exists, similar to guess_start_dir
+        Uses existing value of self.start_dir if it is already set, an absolute path and exists
+        otherwise use self.ext_dir (path to extracted source) as base dir if that is set and exists.
         """
-        possible_dirs = (self.start_dir, self.ext_dir)
-        for possible_dir in possible_dirs:
-            if possible_dir and os.path.isdir(possible_dir):
-                self.cfg['start_dir'] = possible_dir
-                self.log.debug("Using start_dir: %s", self.start_dir)
+        start_dir = ''
+
+        # Use provided start dir if it is an absolute path
+        if self.start_dir:
+            start_dir = self.start_dir
+            if os.path.isabs(start_dir) and os.path.isdir(start_dir):
+                self.log.info("Using user provided start dir: %s", start_dir)
                 return
-        self.log.debug("Unable to determine start_dir as none of these paths is set and exists: %s", possible_dirs)
+
+        # Generate absolute start dir from ext_dir
+        if self.ext_dir:
+            ext_start_dir = os.path.join(self.ext_dir, start_dir)
+            if os.path.isdir(ext_start_dir):
+                self.cfg['start_dir'] = ext_start_dir
+                self.log.debug("Using start dir: %s", ext_start_dir)
+                return
+
+        tested_dirs = ', '.join([d for d in [start_dir, ext_start_dir] if d])
+        self.log.debug("Unable to determine extension start dir as none of the tentative dirs exist: %s" % tested_dirs)
 
     def run(self, unpack_src=False):
         """Common operations for extensions: unpacking sources, patching, ..."""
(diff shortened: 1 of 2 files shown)


easybuild/framework/extensioneasyblock.py
```diff
@@ -100,18 +100,30 @@ def __init__(self, *args, **kwargs):
...
+        # Use provided start dir if it is an absolute path
+        if self.start_dir:
+            start_dir = self.start_dir
+            if os.path.isabs(start_dir) and os.path.isdir(start_dir):
+                self.log.info("Using user provided start dir: %s", start_dir)
+                return
+
+        # Generate absolute start dir from ext_dir
+        if self.ext_dir:
+            ext_start_dir = os.path.join(self.ext_dir, start_dir)
```
ocaisa, line 119:
```suggestion
            # User may have provided a _relative_ path for self.start_dir (now stored in start_dir)
            ext_start_dir = os.path.join(self.ext_dir, start_dir)
```

ocaisa: It took me a while to figure out why this `join` was there, hence the comment

easybuild/framework/extensioneasyblock.py
```diff
@@ -100,18 +100,31 @@ def __init__(self, *args, **kwargs):
...
-        possible_dirs = (self.start_dir, self.ext_dir)
-        for possible_dir in possible_dirs:
-            if possible_dir and os.path.isdir(possible_dir):
-                self.cfg['start_dir'] = possible_dir
-                self.log.debug("Using start_dir: %s", self.start_dir)
+        start_dir, ext_start_dir = '', ''
+
+        # Use provided start dir if it is an absolute path
+        if self.start_dir:
+            start_dir = self.start_dir
```
ocaisa, line 112:
```suggestion
        if self.start_dir:
            start_dir = self.start_dir
            # Use provided start dir if it is an absolute path
```

easybuild/framework/extensioneasyblock.py
```diff
@@ -100,18 +100,31 @@ def __init__(self, *args, **kwargs):
...
+            if os.path.isabs(start_dir) and os.path.isdir(start_dir):
+                self.log.info("Using user provided start dir: %s", start_dir)
+                return
```
ocaisa, line 115: IIRC @boegel doesn't like using `return` to escape different code paths. Perhaps use an interim variable link `final_start_dir=None`, and then check `and not final_start_dir` below, make the assignment at the end then with an `if final_start_dir...else...`

ocaisa: I do see that the use of `return` was already there 🤷 so if @boegel want's to tell me I'm mistaken...

ocaisa (changes requested):

easybuild/framework/extensioneasyblock.py
boegel: Yeah, not a big fan of inline `return` statements, they're easy to overlook and misinterpret what the function is doing...

Especially empty `return` statements like this, which are equivalent with `return None`...

lexming: guys, we will have to grab a beer, sit and have a serious talk in the next EUM. Early returns are your friends :heart: 

I rewrote the code with an implicit return and simplified the logic a bit. It is indeed easier to read now.

easybuild/framework/extensioneasyblock.py
```diff
@@ -100,18 +100,27 @@ def __init__(self, *args, **kwargs):
...
+            # start dir is either empty or a _relative_ path provided by user through self.start_dir
+            # generate absolute path from ext_dir
+            ext_start_dir = os.path.join(self.ext_dir, ext_start_dir)
+
+        if os.path.isdir(ext_start_dir):
+            self.cfg['start_dir'] = ext_start_dir
+            self.log.debug("Using extension start dir: %s", ext_start_dir)
+        else:
+            # non-existing start dir means wrong input from user
+            self.log.debug("Provided start dir for extension does not exist: %s", ext_start_dir)
```
ocaisa, line 123: This is as it was before but I'm a bit curious here, why not inform the user that what they thought they were going to use is not going to be used (i.e., make this a warning)?

ocaisa (approved):
LGTM

easybuild/framework/extensioneasyblock.py
ocaisa: Right now, if I use this and make a typo, I'd have to create a debug log to discover it.

ocaisa: There's an example in https://github.com/easybuilders/easybuild-framework/commit/766347b27597413c180e87a6c75f862b29081231 of how that could be caught in the test

ocaisa: I'll open a PR for this myself

easybuild/framework/extensioneasyblock.py
ocaisa: I feel like I should be buying that beer

easybuild/framework/extensioneasyblock.py
ocaisa: Done in #4201

PR #4198, opened 2023-01-27
Enable github tokens without interaction.

terjekv:
- Add `--keyring-github-token` as a command line option.
  - This implies we get EASYBUILD_KEYRING_GITHUB_TOKEN for the option as well.
  - This allows for users to load the token into the environment on a session basis.

As an example, one can do:

````shell
function easybuild_github_token_print() {
        python -c 'from keyrings.cryptfile.cryptfile import CryptFileKeyring; print(CryptFileKeyring().get_password("github_token", "terjekv"))'
}

function easybuild_github_token_import() {
        if [ -z "$EASYBUILD_KEYRING_GITHUB_TOKEN" ]; then
                export EASYBUILD_KEYRING_GITHUB_TOKEN=$( easybuild_github_token_print )
        fi
}
````

Which leads to:

````
terjekv@foo:~$ eb --check-github
== Temporary log file in case of crash /tmp/eb-cd2gg12t/easybuild-zwfgp9d4.log

Checking status of GitHub integration...

Making sure we're online...OK

* GitHub user...terjekv => OK
Please enter password for encrypted keyring:
WARNING: signal received (2), cleaning up locks ()...

ERROR: Cancelled by user: keyboard interrupt
[...]
terjekv@foo:~$ easybuild_github_token_import
Please enter password for encrypted keyring:
terjekv@foo:~$ eb --check-github
== Temporary log file in case of crash /tmp/eb-qehdn69c/easybuild-2vxsk85d.log

Checking status of GitHub integration...

Making sure we're online...OK

* GitHub user...terjekv => OK
* GitHub token...ghp..vde (len: 40) => OK (validated)
* git command...OK ("git version 2.31.1; ")
* GitPython module...OK (GitPython version 3.1.30)
[...]
````

This also allows for tokens to be stored in Vault or similar tools.

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -223,6 +223,7 @@ BUILD_OPTIONS_CMDLINE = {
         'github_user',
         'github_org',
         'group',
+        'keyring_github_token',
         'hide_deps',
         'hide_toolchains',
         'http_header_fields_urlpat',
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -2164,7 +2164,12 @@ def check_github():
 
 
 def fetch_github_token(user):
-    """Fetch GitHub token for specified user from keyring."""
+    """Fetch GitHub token for specified user.
+
+    Sources used for the token are, in order:
+      - keyring option (--keyring-github-token or ENV{'EASYBUILD_KEYRING_GITHUB_TOKEN'})
+      - keyring.getpassword(KEYRING_GITHUB_TOKEN, user)
+    """
 
     token, msg = None, None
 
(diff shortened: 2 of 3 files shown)


boegel:
@terjekv I can see the motivation for this, but storing the token in the environment is making it really easy to accidentally leaking - too easy imho. In test reports that are submitting in easyconfig PRs, for example.

See also https://github.com/easybuilders/easybuild-framework/security/advisories/GHSA-2wx6-wc87-rmjm

So, I'm not sure that supporting this is a good idea...

terjekv:
Hmm. Right. The debug logs are a problem unless we explicitly filter data, which is always tricky.

The goals for me though are as follows:

1. Not be bound to using python keyring or develop keyring backends for other services. 
2. Allow for session-bound token usage, very useful looping over multiple, long-running, tests that are to be uploaded.
3. Allow other tools to control availability of the token (via export/unset).

However, the big pitfall is that we will need to do environment filtering. Now, people may have other secrets in their environment so some filtering has to happen either way, and we could create a namespace (ala `EASYBUILD_SECRETS...`) and filter out that from any reporting.

And yes, https://github.com/easybuilders/easybuild-framework/security/advisories/GHSA-2wx6-wc87-rmjm shows that having secrets is tricky in its own right and adding a scope like this will increase the risk surface. 

Hrmpf. :)

terjekv:
I'm going to look at a few different ways of resolving this without resorting to using ENV. Closing. :)

Issue #3963, opened 2022-02-11
stop relying on distutils (will no longer be available in Python 3.12)


boegel:
`distutils` is deprecated since Python 3.10, will be removed in Python 3.12; see https://www.python.org/dev/peps/pep-0632

Places we rely on `distutils` currently:

* `from distutils.version import LooseVersion` (framework + easyblocks)
  * we need a replacement to do software version comparison
  * attempt was made in #3794, ~but that has bitrotted a bit~ looks promising...
  * can/should we come up with a better solution than just copy-pasting `LooseVersion` from Python stdlib into EasyBuild?
  * maybe infix operators like `|before|` (see also http://tomerfiliba.com/blog/Infix-Operators/)?
* `from distutils.version import StrictVersion` in `easybuild/tools/modules.py` + tests
* in `setup.py`, to avoid depending on `setuptools`
  * it looks like `setuptools` will be a necesary evil (see also #2836)
* `from distutils.util import strtobool` in `easybuild/framework/easyconfig/types.py`

boegel:
All done, see #4484

haampie:
@boegel can you reopen? there are still distutils remnants

https://github.com/easybuilders/easybuild-framework/blob/d700144108b3a5508659a4d957c7594d35033e62/setup.py#L35-L36

having setuptools installed does not mean that distutils can be imported.

See https://github.com/cython/cython/pull/5754#issuecomment-1752102480

boegel:
Let's follow up in #4478, I've re-opened that, maybe we overlooked something indeed...

@haampie Thanks for raising this!

boegel:
fixed with #4478

PR #4204, opened 2023-02-01
Silence distutils deprecation warnings

Flamefire:
Currently CI fails due to deprecation warnings

Once due to a left-over use of  `LooseVersion` and 2nd `StrictVersion`.

I resolved the latter with a wrapper class that suppresses the warning.

See #3963

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -215,4 +215,4 @@ jobs:
           IGNORE_PATTERNS="no GitHub token available|skipping SvnRepository test|requires Lmod as modules tool|stty: 'standard input': Inappropriate ioctl for device|CryptographyDeprecationWarning: Python 3.[56]|from cryptography.*default_backend|CryptographyDeprecationWarning: Python 2|from cryptography.utils import int_from_bytes|Blowfish|GC3Pie not available, skipping test"
           # '|| true' is needed to avoid that Travis stops the job on non-zero exit of grep (i.e. when there are no matches)
           PRINTED_MSG=$(egrep -v "${IGNORE_PATTERNS}" test_framework_suite.log | grep '\.\n*[A-Za-z]' || true)
-          test "x$PRINTED_MSG" = "x" || (echo "ERROR: Found printed messages in output of test suite\n${PRINTED_MSG}" && exit 1)
+          test "x$PRINTED_MSG" = "x" || (echo "ERROR: Found printed messages in output of test suite" && echo "${PRINTED_MSG}" && exit 1)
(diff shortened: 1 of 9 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -212,7 +212,7 @@ jobs:
           # run test suite
           python -O -m test.framework.suite 2>&1 | tee test_framework_suite.log
           # try and make sure output of running tests is clean (no printed messages/warnings)
-          IGNORE_PATTERNS="no GitHub token available|skipping SvnRepository test|requires Lmod as modules tool|stty: 'standard input': Inappropriate ioctl for device|CryptographyDeprecationWarning: Python 3.[56]|from cryptography.*default_backend|CryptographyDeprecationWarning: Python 2|from cryptography.utils import int_from_bytes|Blowfish|GC3Pie not available, skipping test"
+          IGNORE_PATTERNS="no GitHub token available|skipping SvnRepository test|requires Lmod as modules tool|stty: 'standard input': Inappropriate ioctl for device|CryptographyDeprecationWarning: Python 3.[56]|from cryptography.* import |CryptographyDeprecationWarning: Python 2|Blowfish|GC3Pie not available, skipping test"
```
akesandgren, line 215: the "." in the "Python 3.[56]" should probably have a "\" in front of it to be fully correct.

easybuild/toolchains/iimkl.py
```diff
@@ -32,13 +32,13 @@
...
 * Bart Oldeman (McGill University, Calcul Quebec, Compute Canada)
 """
-from distutils.version import LooseVersion
 import re
 
 from easybuild.toolchains.iccifort import IccIfort
 from easybuild.toolchains.intel_compilers import IntelCompilersToolchain
 from easybuild.toolchains.fft.intelfftw import IntelFFTW
 from easybuild.toolchains.linalg.intelmkl import IntelMKL
+from easybuild.tools import LooseVersion
```
akesandgren, line 41: This import of LooseVerstion differs from the one below in tools/__init__.py, why?

.github/workflows/unit_tests.yml
Flamefire: I haven't touched that here.

easybuild/toolchains/iimkl.py
Flamefire: Because in order to do `from easybuild.tools import LooseVersion` we need to import it in the tools `__init__.py`

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #4205, opened 2023-02-02
Replace more usages of assertTrue/False

Flamefire:
I found yet another bug in the tests in `test_pypi_source_urls`:

> `self.assertTrue(eb340_url, res)`

This test will always succeed, `assertIn` was intended.

Hence I did another RegEx-powered Search&Replace and introduced `assert(Not)Exists` to replace the most common use of `assertTrue/False` which also fixes a few bugs in error messages.


test/framework/filetools.py
```diff
@@ -1009,7 +1009,7 @@ def test_back_up_file(self):
         new_file = [x for x in test_files if x not in known_files][0]
         self.assertTrue(new_file.startswith('test.txt.bak_'))
         known_files = os.listdir(os.path.dirname(fp))
-        self.assertTrue(ft.read_file(first_normal_backup), txt)
+        self.assertEqual(ft.read_file(first_normal_backup), txt)
```
jfgrimm, line 1012: these change the behaviour, but that looks intended

test/framework/filetools.py
```diff
@@ -2313,24 +2313,25 @@ def test_remove(self):
...
-            self.assertFalse(os.path.exists(test_dir) or os.path.isdir(test_dir))
+            self.assertNotExists(test_dir)
 
         # remove also takes a list of paths
         ft.write_file(testfile, 'bar')
         ft.mkdir(test_dir)
-        self.assertTrue(os.path.exists(testfile))
-        self.assertTrue(os.path.exists(test_dir) and os.path.isdir(test_dir))
+        self.assertExists(testfile)
+        self.assertExists(test_dir) and os.path.isdir(test_dir)
```
jfgrimm, line 2331:
```suggestion
        self.assertExists(test_dir)
        self.assertTrue(os.path.isdir(test_dir))
```

test/framework/systemtools.py
```diff
@@ -401,7 +401,7 @@ def tearDown(self):
     def test_avail_core_count_native(self):
         """Test getting core count."""
         core_count = get_avail_core_count()
-        self.assertTrue(isinstance(core_count, int), "core_count has type int: %s, %s" % (core_count, type(core_count)))
+        self.assertIsInstance(core_count, int), "core_count has type int: %s, %s" % (core_count, type(core_count))
```
jfgrimm, line 404: wrong brackets
```suggestion
        self.assertIsInstance(core_count, int, "core_count has type int: %s, %s" % (core_count, type(core_count)))
```

jfgrimm (changes requested):

test/framework/easyblock.py
```diff
@@ -1783,7 +1783,7 @@ def test_obtain_file(self):
             if res is not None:
                 loc = os.path.join(tmpdir, 't', 'toy', fn)
                 self.assertEqual(res, loc)
-                self.assertTrue(os.path.exists(loc), "%s file is found at %s" % (fn, loc))
+                self.assertExists(loc), "%s file is found at %s" % (fn, loc)
```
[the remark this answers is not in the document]
Flamefire: I actually intended to remove the message. It doesn't provide any value over the default (`fn` is part of `loc`)

test/framework/filetools.py
Flamefire: Yes, this was a bug before.

test/framework/systemtools.py
Flamefire: I missed to remove the message. Thanks!

jfgrimm (approved):
lgtm

jfgrimm:
Going in, thanks @Flamefire!

PR #4206, opened 2023-02-02
Avoid warning for non-existing start_dir when set to None

Flamefire:
Python packages may be installed from WHL files which do not get extracted in which case `self.start_dir, self.ext_dir` are both `None`. This leads to an unhelpful warning:
> WARNING: Provided start dir (None) for extension tensorboard-plugin-wit does not exist:

Check that case and only log a debug message. Also add test for this.

Note that #4196 changed semantics: Prior to this `start_dir` or `ext_dir` where used if set and existing resolved relative to the current directory. After that `start_dir` is resolved relative to `ext_dir` which is usually the same but does not need to.

### DIFF
--- a/easybuild/framework/extensioneasyblock.py
+++ b/easybuild/framework/extensioneasyblock.py
@@ -105,17 +105,24 @@ class ExtensionEasyBlock(EasyBlock, Extension):
 
         Uses existing value of self.start_dir if it is set.
         If self.ext_dir (path to extracted source) is set, it is used as the base dir for relative paths.
-        Otherwise (e.g. for non-extracted extensions like Python WHL files) or when the computed start_dir
-        does not exist the start dir is not changed.
+        Otherwise (e.g. for non-extracted extensions like Python WHL files) the current working directory
+        is used as the base dir.
+        When neither start_dir nor ext_dir are set or when the computed start_dir does not exist
+        the start dir is not changed.
         """
         ext_start_dir = self.start_dir
         if self.ext_dir:
             if ext_start_dir:
-                # If start_dir is relative, resolve it using ext_dir as a base
                 ext_start_dir = os.path.join(self.ext_dir, ext_start_dir)
             else:
-                # start_dir is not set -> Use ext_dir
+                # start_dir is not set or empty -> Use ext_dir
                 ext_start_dir = self.ext_dir
+        elif ext_start_dir is not None:
+            # Resolve relative to current dir
+            if ext_start_dir:
+                ext_start_dir = os.path.join(os.getcwd(), ext_start_dir)
+            else:
+                ext_start_dir = os.getcwd()
 
         if ext_start_dir and os.path.isdir(ext_start_dir):
             self.cfg['start_dir'] = ext_start_dir
--- a/easybuild/framework/extensioneasyblock.py
+++ b/easybuild/framework/extensioneasyblock.py
@@ -103,30 +103,25 @@ class ExtensionEasyBlock(EasyBlock, Extension):
     def _set_start_dir(self):
         """Set absolute path of self.start_dir similarly to EasyBlock.guess_start_dir
 
-        Uses existing value of self.start_dir if it is set.
+        Uses existing value of self.start_dir defaulting to self.ext_dir.
         If self.ext_dir (path to extracted source) is set, it is used as the base dir for relative paths.
-        Otherwise (e.g. for non-extracted extensions like Python WHL files) the current working directory
-        is used as the base dir.
+        Otherwise otherwise self.builddir is used as the base.
         When neither start_dir nor ext_dir are set or when the computed start_dir does not exist
         the start dir is not changed.
+        The computed start dir will not end in path separators
         """
         ext_start_dir = self.start_dir
         if self.ext_dir:
             if not os.path.isabs(self.ext_dir):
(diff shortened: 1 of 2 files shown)


easybuild/framework/extensioneasyblock.py
```diff
@@ -106,10 +106,7 @@ def _set_start_dir(self):
         Uses existing value of self.start_dir if it is already set, an absolute path and exists
         otherwise use self.ext_dir (path to extracted source) as base dir if that is set and exists.
```
lexming, line 107:
```suggestion
        Use existing value of self.start_dir if it is already set and return the corresponding absolute path
        For relative paths use self.ext_dir (path to extracted source) as base dir
        Non-extracted extensions (e.g. Python WHL) have a None start_dir
```

lexming (commented):
LGTM

I suggest to update the description of `_set_start_dir` to reflect the new logic

easybuild/framework/extensioneasyblock.py
Flamefire: Good suggestion! However "Non-extracted extensions (e.g. Python WHL) have a None start_dir" does not hold according to this function: They have an unset `ext_dir` and hence `start_dir` is not changed. I hence used
```
        Uses existing value of self.start_dir if it is set.
        If self.ext_dir (path to extracted source) is set, it is used as the base dir for relative paths.
        Otherwise (e.g. for non-extracted extensions like Python WHL files) or when the computed start_dir
        does not exist the start dir is not changed.
```

Of course we could further refine this to use the CWD to convert relative paths to absolute ones (I added tests: `start_dir` might be set to `.` or `..` in the EC) but not sure if that is worth it. But then we would have the invariant that the start_dir is always absolute

lexming: I agree that it would be nice to make `start_dir` always an absolute path, that was the intention of my initial PR. In the case where `self.start_dir=None` there is no `self.ext_dir` either, right? So one option would be to set `start_dir` to the absolute path of `builddir`. That will exist at this stage.

Flamefire: I wouldn't set start_dir if both `self.start_dir` and `self.ext_dir` are None. Accessing it in that case is likely an error and shouldn't be masked by setting it (kind of) arbitrary.

However I would agree to resolving start_dir relative to the CWD when it is set and ext_dir is not. That keeps the current behavior (using builddir would potentially change current behavior) and we have the invariant that it is either None or absolute

lexming:
As discussed with @Flamefire, these are the target outcomes for `_set_start_dir`:

| start_dir  |  ext_dir=None                |  ext_dir=relative path | ext_dir=absolute path |
-------------|------------------------------|------------------------|-----------------------|
| None       | None                         | error (1)              | absolute path |
| empty      | absolute path from build dir | error (1)              | absolute path from ext_dir |
| relative   | absolute path from build dir | error (1)              | absolute path from ext_dir |
| absolute   | keep as is                   | error (1)              | keep as is |

1: might be done by some custom easyblock, cannot be handled as we do not know the base dir
2: always remove trailing slashes from absolute paths (if `/` keep as is)

The result will be that `start_dir` is always an absolute path except where both `ext_dir` and `start_dir` are none (_e.g._ Python wheels)

test/framework/easyblock.py
```diff
@@ -2172,46 +2172,96 @@ def test_extension_set_start_dir(self):
...
+            if expected_start_dir is None:
+                self.assertIsNone(ext.start_dir)
+            else:
+                self.assertTrue(os.path.isabs(ext.start_dir))
+                if ext.start_dir != os.sep:
+                    self.assertFalse(ext.start_dir.endswith(os.sep))
+                if os.path.isabs(expected_start_dir):
+                    abs_expected_start_dir = expected_start_dir
+                else:
+                    abs_expected_start_dir = os.path.join(eb.builddir, expected_start_dir)
```
lexming, line 2194: @Flamefire as you have taught me, this is not necessary, isn't it?
```suggestion
                abs_expected_start_dir = os.path.join(eb.builddir, expected_start_dir)
```

test/framework/easyblock.py
```diff
@@ -2172,46 +2172,96 @@ def test_extension_set_start_dir(self):
...
+                self.assertEqual(ext.start_dir, abs_expected_start_dir)
+                if not os.path.exists(eb.builddir):
+                    eb.make_builddir()  # Required to exist for samefile
+                self.assertTrue(os.path.samefile(ext.start_dir, abs_expected_start_dir))
+            if unpack_src:
+                self.assertTrue(os.path.samefile(os.getcwd(), abs_expected_start_dir))
```
lexming, line 2200: `abs_expected_start_dir` might not be set if `expected_start_dir` is None

Flamefire: Yeah, I guess in that case this line would raise a TypeError. That is fine as it shouldn't ever happen: If we unpack then start_dir and hence `expected_start_dir` must not be `None` so an earlier assertion would already be triggered.

lexming: it's true that several other assertions will fail in such a case :+1:

lexming (approved):
LGTM

lexming:
Going in, thanks @Flamefire !

PR #4207, opened 2023-02-02
Add option to silence deprecation warning of ECs and toolchains

Flamefire:
We have the `silence_deprecation_warnings` option but that didn't support the warnings caused by deprecated ECs and toolchains.

Add this and enhance the error message.

While at it also make the build_option return an empty list instead of `None` so we can grep the code for any usage like `if 'foo' in build_option('silence_deprecation_warnings'):`

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -339,6 +339,8 @@ class EasyBuildOptions(GeneralOption):
         # override options
         descr = ("Override options", "Override default EasyBuild behavior.")
 
+        all_deprecations = ('python2', 'Lmod6', 'easyconfig', 'toolchain')
+
         opts = OrderedDict({
             'accept-eula': ("Accept EULA for specified software [DEPRECATED, use --accept-eula-for instead!]",
                             'strlist', 'store', []),
@@ -494,7 +496,9 @@ class EasyBuildOptions(GeneralOption):
             'set-default-module': ("Set the generated module as default", None, 'store_true', False),
             'set-gid-bit': ("Set group ID bit on newly created directories", None, 'store_true', False),
             'show-progress-bar': ("Show progress bar in terminal output", None, 'store_true', True),
-            'silence-deprecation-warnings': ("Silence specified deprecation warnings", 'strlist', 'extend', None),
+            'silence-deprecation-warnings': (
+                "Silence specified deprecation warnings out of (%s)" % ', '.join(all_deprecations),
+                'strlist', 'extend', []),
             'skip-extensions': ("Skip installation of extensions", None, 'store_true', False),
             'skip-test-cases': ("Skip running test cases", None, 'store_true', False, 't'),
             'skip-test-step': ("Skip running the test step (e.g. unit tests)", None, 'store_true', False),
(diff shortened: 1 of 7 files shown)


test/framework/easyconfig.py
```diff
@@ -3795,6 +3795,26 @@ def test_deprecated(self):
...
         self.assertErrorRegex(EasyBuildError, error_pattern, EasyConfig, test_ec)
+        with self.mocked_stdout_stderr():
+            # But this can be silenced
+            init_config(build_options={'silence_deprecation_warnings': ['easyconfig']})
+            EasyConfig(test_ec)
+            self.assertFalse(self.get_stderr())
+            self.assertFalse(self.get_stdout())
+
+    def test_deprecated_toolchain(self):
+        """Test use of deprecatd toolchain"""
```
jfgrimm, line 3806:
```suggestion
        """Test use of deprecated toolchain"""
```

easybuild/tools/options.py
```diff
@@ -907,6 +911,10 @@ def postprocess(self):
         if self.options.deprecated:
             build_log.CURRENT_VERSION = LooseVersion(self.options.deprecated)
 
+        # Convert this to an empty list if not set to simplify usage
+        if self.options.silence_deprecation_warnings is None:
+            self.options.silence_deprecation_warnings = []
+
```
jfgrimm, line 917: is this needed, since you're already changing the default in L499-501? I guess it doesn't hurt either way

Flamefire: I forgot to revert that change of the default. I decided against it as it looks ugly. Or maybe it did work because the default of `None` was still used in the "reverse-map"(?) in easybuild/tools/config.py for `silence_deprecation_warnings`

Changed the default back to `None` to be safe here.

Flamefire: @jfgrimm After doing the quick fix I (was) remembered why it was this way: I didn't want the `--help` to show `" ... (... default: )`, i.e. an empty default value which is shown for an empty list but not for `None`. Hence I cheated a bit to make it look pretty. But I just added a proper way of handling this.

Flamefire:
This now adds a new change:

Only print "default: " for str-list options if it is not empty.   
This changes `(type comma-separated list; default: )` to `(type comma-separated list)` similar to a default of `None`.

That changes the output of `--help` in the following places (done using `diff` over the output before and after):
>                         Parse (additional) configfiles (type comma-separated list)
>                         Enable dependency resolution, using easyconfigs in specified paths (type pathsep-separated list)
>                         Additional paths to consider by robot for easyconfigs (--robot paths get priority) (type pathsep-separated list)
>                         Additional locations to consider in --search (next to --robot and --robot-paths paths) (type pathsep-separated list)
>                         Location(s) of extra or customized easyblocks (type comma-separated list)
>                         Location(s) of extra or customized module naming schemes (type comma-separated list)
>                         Location(s) of extra or customized toolchains or toolchain components (type comma-separated list)
>     --from-pr=PR#       Obtain easyconfigs from specified PR (type comma-separated list)
>                         Include easyblocks from specified PR (type comma-separated list)
>                         Accept EULA for specified software [DEPRECATED, use --accept-eula-for instead!] (type comma-separated list)
>                         Accept EULA for specified software (type comma-separated list)
>                         Silence specified deprecation warnings out of (python2, Lmod6, easyconfig, toolchain) (type comma-separated list)

BTW: A colon after "type" would probably be nice.

easybuild/tools/options.py
jfgrimm: ah, that makes sense

jfgrimm (approved):
lgtm

PR #4212, opened 2023-02-09
added a test to raise an EasyBuildError if a value with a template is…

mboisson:
… replaced by a value without template

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1780,6 +1780,10 @@ class EasyConfig(object):
     def __setitem__(self, key, value):
         """Set value of specified easyconfig parameter (help text & co is left untouched)"""
         if key in self._config:
+            # raise an error if we are replacing a value with templates by a value with templates resolved
+            if "%(" in self._config[key][0] and not "%(" in value:
+                raise EasyBuildError("Replacing value of templated parameter '%s' with a value with templates
+                                      resolved. Old value: %s, new value: %s" % (key, self._config[key][0], value))
             self._config[key][0] = value
         else:
             raise EasyBuildError("Use of unknown easyconfig parameter '%s' when setting parameter value to '%s'",


lexming (commented):
I looked at the related issue and I wonder if we can define a more clear behaviour here. Ideally I would trigger a straight error when the template replacement is dangerous/unintended and not do anything otherwise.

Maybe we can check if templating is disabled through `EasyConfig.enable_templating` and in such a case trigger an error whenever `__setitem__` replaces a template. If templating is enabled probably is not an issue. What do you think? Would this approach work?

boegel:
There's various reasons why passing a "raw" string value (with unresolved template values) makes sense, and some template values (like `%(pyver)s`) can only be resolved until some installation step has been reached (like `prepare` step, where dependencies are loaded), so a warning like this would result in quite a lot of false positives, rendering it useless...

mboisson:
> There's various reasons why passing a "raw" string value (with unresolved template values) makes sense, and some template values (like `%(pyver)s`) can only be resolved until some installation step has been reached (like `prepare` step, where dependencies are loaded), so a warning like this would result in quite a lot of false positives, rendering it useless...

Hum ? That warning is precisely the opposite, i.e. when a "non-raw" string value (with *resolved* template values) is used to overwrite a raw one.

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1780,6 +1780,10 @@ def get_ref(self, key):
     def __setitem__(self, key, value):
         """Set value of specified easyconfig parameter (help text & co is left untouched)"""
         if key in self._config:
+            # raise an error if we are replacing a value with templates by a value with templates resolved
+            if "%(" in str(self._config[key][0]) and "%(" not in str(value):
```
Flamefire, line 1784: Maybe do this only when (both?) are already strings as it doesn't make sense IMO otherwise and just slows EB down (even more)
BTW: Outdated comment. Can be removed as the code (check and warning) already says what's done.

mboisson: I can change the comment, but I disagree that it does not make sense. Such erroneous replacements will definitely happen in config opts that are arrays (and not strings).

Flamefire: >happen in config opts that are arrays (and not strings).

Ah, true. Good point. I was thinking about the costly conversion and string search when setting int values etc. But I guess most are really strings or contain strings.

mboisson: comment has been adjusted

PR #4213, opened 2023-02-10
restore initial environment before processing each easystack item

### LINKED ISSUE #4194: Dependencies of first build are still loaded for all next builds with an easystack file
I was installing a whole bunch of apps using an easystack file, and suddenly it crashed at the installation of `zlib-1.2.12.eb` because `$EBROOTZLIB` was already set. I hadn't done that myself (started with a clean environment), and after inspecting the log file, I noticed that for all installations after the very first one Easybuild was warning that a bunch of modules were loaded. This list consisted of the (build) dependencies of the first application that it installed.

I was able to reproduce it with a simple example, using this easystack file:
```
easyconfigs:
  - attr-2.5.1-GCCcore-11.2.0.eb
  - zlib-1.2.12.eb
```

Installing this with `eb -f --experimental --easystack test.yml` (note that the dependencies of `attr` are already available on my system) results in a successful installation of `attr`, but:
```
== FAILED: Installation ended unsuccessfully (build directory: 
/data/eb/build/zlib/1.2.12/system-system): build failed (first 300 chars): 
Module is already loaded (EBROOTZLIB is set), installation cannot continue. 
(took 0 secs)
```

The `attr` log has:
```
== 2023-01-25 12:45:58,939 easyblock.py:2248 WARNING Loaded modules detected: ['EasyBuild/4.7.0']
```
But the one for `zlib` has:
```
== 2023-01-25 12:46:04,854 easyblock.py:2248 WARNING Loaded modules detected: ['EasyBuild/4.7.0', 'GCCcore/11.2.0', 'zlib/1.2.11-GCCcore-11.2.0', 'binutils/2.37-GCCcore-11.2.0']
```

So it looks like it didn't unload the `attr` (build) dependencies. This is with EB 4.7.0, and based on some older logs of applications built with a similar easystack (but with the old structure) and EB 4.6.2, the latter version didn't have this issue.

bedroge:
Closes #4194 by explicitly restoring the environment at the end of `build_and_install_software`. This should prevent that the dependencies of a previous (set of) installation(s) are still loaded for the next one, e.g. when using an easystack file. Also added a test that verifies this.

Test output before applying the patch to `main.py`:
```
$ python3 -O -m test.framework.easystack
.......F..
FAIL: test_easystack_restore_env_after_each_build (__main__.EasyStackTest)
Test that the build environment is reset for each easystack item
  File "/home/bob/ebdev/easybuild-framework/test/framework/easystack.py", line 148, in test_easystack_restore_env_after_each_build
    self.assertFalse(regex.search(stdout), "Pattern '%s' should not be found in: %s" % (regex.pattern, stdout))
AssertionError: <re.Match object; span=(166129, 166280), match="WARNING Loaded modules detected: ['EasyBuild-deve> is not false : Pattern 'WARNING Loaded modules detected: \[.*gompi/2018.*\]\n' should not be found in:
<LOTS OF OUTPUT>
== 2023-02-10 13:46:59,710 easyblock.py:2248 WARNING Loaded modules detected: ['EasyBuild-develop', 'GCC/6.4.0-2.28', 'hwloc/1.11.8-GCC-6.4.0-2.28', 'OpenMPI/2.1.2-GCC-6.4.0-2.28', 'gompi/2018a']
<LOTS OF OUTPUT>
Ran 10 tests in 2.169s

FAILED (failures=1)
```

After applying the patch:
```
$ python3 -O -m test.framework.easystack
..........
Ran 10 tests in 2.096s

OK
```

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -62,6 +62,7 @@ from easybuild.framework.easyconfig.tweak import obtain_ec_for, tweak
 from easybuild.tools.config import find_last_log, get_repository, get_repositorypath, build_option
 from easybuild.tools.containers.common import containerize
 from easybuild.tools.docs import list_software
+from easybuild.tools.environment import restore_env
 from easybuild.tools.filetools import adjust_permissions, cleanup, copy_files, dump_index, load_index
 from easybuild.tools.filetools import locate_files, read_file, register_lock_cleanup_signal_handlers, write_file
 from easybuild.tools.github import check_github, close_pr, find_easybuild_easyconfig
@@ -186,6 +187,10 @@ def build_and_install_software(ecs, init_session_state, exit_on_failure=True):
         update_progress_bar(STATUS_BAR, label=status_label)
 
     stop_progress_bar(STATUS_BAR)
+
+    # restore original environment for any next software builds
+    _log.info("Resetting environment")
+    restore_env(init_env)
 
     return res
 
--- a/test/framework/easystack.py
+++ b/test/framework/easystack.py
@@ -29,6 +29,7 @@ Unit tests for easystack files
 @author: Kenneth Hoste (Ghent University)
 """
 import os
+import re
 import sys
 from unittest import TextTestRunner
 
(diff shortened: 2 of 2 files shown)


bedroge:
Tests are failing because of:
```
ERROR: test_compiler_cache (test.framework.toolchain.ToolchainTest)
Test ccache
  File "/tmp/runner/736df0562413dad6f69227f3175ee31db19c4876/lib/python2.7/site-packages/test/framework/toolchain.py", line 2187, in test_compiler_cache
    self.assertTrue(os.path.samefile(os.environ['CCACHE_DIR'], ccache_dir))
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/UserDict.py", line 40, in __getitem__
    raise KeyError(key)
KeyError: 'CCACHE_DIR'
```
But I have no clue what's causing this, doesn't seem related to this PR?

boegel:
@bedroge That test is failing because in it we assume that the build environment does not get reset after the installation of easyconfigs, which the change you propose (which does make sense) is breaking...

bedroge:
> @bedroge The failing test is actually telling us that we're making a somewhat intrusive change. It's not impossible that some people are actually relying on the build environment not being reset, for example when [using EasyBuild as a library](https://tutorial.easybuild.io/2021-lust/easybuild_library/).
> 
> Here's a proposal for a different approach: [bedroge#1](https://github.com/bedroge/easybuild-framework/pull/1)

Looks good, merged!

PR #4216, opened 2023-02-15
fix typo in condition guarding installation of GitHub token in CI workflow to run unit tests + fix broken tests related to GitHub integration features

### LINKED ISSUE #4214: tests that require GitHub token are always skipped in CI due to silly typo in GitHub Actions workflow
There's a silly typo in the GitHub Actions workflow file we use to run the unit tests (`.github/workflows/unit_tests.yml`):

https://github.com/easybuilders/easybuild-framework/blob/89fc160c44eb9e43ee27243968988f77f365f384/.github/workflows/unit_tests.yml#L130

`${{matrix.modules_syntax}}` should be `${{matrix.module_syntax}}`

As a result of this, the GitHub token that is available when the tests are being run are never run, even if a GitHub token is available (which is the case when the tests are run after a PR is merged into `develop`), and we skip any tests that require a GitHub token if none is available.

This problems was introduced in #3938 (merged Jan'22)

I looked into fixing this, but then I'm seeing trouble with the permissions of GitHub token that's available when the tests are run, see for example https://github.com/boegel/easybuild-framework/actions/runs/4182110823/jobs/7244807091

Relevant part of output of `eb --check-github`:

```
* GitHub user...easybuild_test => OK
* GitHub token...ghp..FGM (len: 40) => FAIL (validation failed)
```

cc @migueldiascosta

boegel:
fixes #4214


boegel:
draft PR because for some reason the token provided by GitHub isn't recognized as valid

see for example https://github.com/boegel/easybuild-framework/actions/runs/4181752996/jobs/7244026081

```
$ eb --check-github
...
* GitHub user...easybuild_test => OK
* GitHub token...ghp..FGM (len: 40) => FAIL (validation failed)
```

boegel:
@migueldiascosta to me that we're using `${{secrets.TEST_GITHUB_TOKEN}}`, which is a self-defined environment variable, so that probably has a now expired GitHub token...

We should try using the standard `GITHUB_TOKEN` instead (see https://docs.github.com/en/actions/security-guides/automatic-token-authentication)

boegel:
Broken tests should all be fixed now, and all GitHub tests pass again when a GitHub token is available (most GitHub tests are skipped when no GitHub token is available, when the test suite is run in the scope of a PR for example, like here). See https://github.com/boegel/easybuild-framework/actions/runs/4698932377 for a run where a GitHub token *is* available and no tests are skipped.

I had to go back to provided a self-created GitHub token and make it available to the tests as an [encrypted secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets), because some tests require that the token has `gist` permissions, for example.

cc @migueldiascosta

jfgrimm (approved):
lgtm

PR #4218, opened 2023-02-16
Make test_help_long work on small terminals

Flamefire:
The output may be wrapped so allow whitespace inbetween


bartoldeman (approved):
lgtm

PR #4229, opened 2023-03-26
remove support for Python 2.7 and 3.5

branfosj:
This removes support for running EasyBuild on Python 2.x and 3.5.


easybuild/tools/py2vs3/__init__.py
```diff
@@ -22,40 +22,13 @@
...
 #
-import sys
 
-# all functionality provided by the py2 and py3 modules is made available via the easybuild.tools.py2vs3 namespace
-if sys.version_info[0] >= 3:
-    from easybuild.tools.py2vs3.py3 import *  # noqa
-else:
-    from easybuild.tools.py2vs3.py2 import *  # noqa
+# all functionality provided by the py3 modules is made available via the easybuild.tools.py2vs3 namespace
+from easybuild.tools.py2vs3.py3 import *  # noqa
```
boegel, line 27: Rather than doing this, we should remove the entire `py2vs3` package, and replace all imports accordingly?

That also implies removing things like our custom `string_type`, etc.

requirements.txt
```diff
@@ -4,13 +4,11 @@ keyring
 keyrings.alt
 
 # GitPython 3.1.15 deprecates Python 3.5
-GitPython==3.1.14; python_version >= '3.0' and python_version < '3.6'
-GitPython; python_version >= '3.6' or python_version <= '3.0'
+GitPython==3.1.14; python_version < '3.6'
```
boegel, line 7: No more need for this, since we'll only support Python 3.6+ ?

requirements.txt
```diff
@@ -20,18 +18,13 @@ PyYAML
...
 requests
 
 archspec
 
-# cryptography 3.4.0 no longer supports Python 2.7
-cryptography==3.3.2; python_version == '2.7'
 cryptography; python_version >= '3.5' and python_version < '3.11'
 
-# rich is only supported for Python 3.6+
 rich; python_version >= '3.6'
```
boegel, line 30: strip off the `; python_version >= '3.6'` ?

requirements.txt
```diff
@@ -20,18 +18,13 @@ PyYAML
...
+GC3Pie; python_version < '3.11'
 python-graph-dot
 python-hglib
 requests
 
 archspec
 
-# cryptography 3.4.0 no longer supports Python 2.7
-cryptography==3.3.2; python_version == '2.7'
 cryptography; python_version >= '3.5' and python_version < '3.11'
```
boegel, line 28: This whole line can be removed, I think, we only had this because `GC3Pie` requires it and we had to stick to `cryptography==3.3.2` for Python 2.7?

boegel (changes requested):

easybuild/tools/py2vs3/__init__.py
branfosj: While we've deprecated using EB on Python 2, we've not deprecated the use of `py2vs3`. Are we fine with making this change? Or should we deprecate `py2vs3` in EB5 and remove it in a later version.

Either way I'll switch over the imports.

requirements.txt
branfosj: f9309b8

requirements.txt
branfosj: f9309b8

easybuild/tools/py2vs3/__init__.py
boegel: I guess we could debate over whether or not we'll keep `py2vs3` to soften the blow w.r.t. dropping support for Python 2...
Deprecating imports from `py2vs3` rather than removing it entirely is perhaps a better approach, indeed.

eb
```diff
@@ -60,7 +60,7 @@ PYTHON=
 # - EB_INSTALLPYTHON is set when EasyBuild is installed as a module (by EasyBuild). It is set to the PYTHON
 #   used during that installation (for example, you could override PYTHON using EB_PYTHON at installation
 #   time, this variable preserves that choice).
-for python_cmd in "${EB_PYTHON}" "${EB_INSTALLPYTHON}" 'python' 'python3' 'python2'; do
+for python_cmd in "${EB_PYTHON}" "${EB_INSTALLPYTHON}" 'python' 'python3'; do
```
boegel, line 63: Maybe switch `python3` and `python`?

That's important especially on CentOS 7.x, which is still quite popular (~50% according to latest EasyBuild User Survey)

boegel (changes requested):

eb
branfosj: 60266d5

easybuild/tools/py2vs3/py3.py
```diff
@@ -62,6 +62,12 @@
 # string type that can be used in 'isinstance' calls
 string_type = str
 
+warning_lines = [
+    "Using py2vs3 is deprecated, since EasyBuild no longer runs on Python 2.",
+    "This module will be removed in a future EasyBuild Version.",
+]
+sys.stderr.write('\n\n' + '\n'.join(warning_lines) + '\n\n\n')
```
boegel, line 69: Let's use our `log.deprecated` for this: https://github.com/branfosj/easybuild-framework/pull/7

boegel (changes requested):

boegel (changes requested):

boegel (approved):
lgtm

A big step towards EasyBuild 5.0, thanks for the effort @branfosj!

PR #4243, opened 2023-04-15
remove deprecated is_generic_easyblock from easyconfig.py

branfosj:


test/framework/easyconfig.py
```diff
@@ -3751,22 +3751,6 @@ def test_get_paths_for(self):
         if env_eb_script_path:
             os.environ['EB_SCRIPT_PATH'] = env_eb_script_path
 
-    def test_is_generic_easyblock(self):
```
boegel, line 3754: also being tested via `test/framework/filetools.py`, so indeed OK to remove :+1:

boegel (approved):
lgtm

PR #4247, opened 2023-04-15
enforce correct extentions for patch files

branfosj:


test/framework/filetools.py
```diff
@@ -1635,36 +1635,17 @@ def test_create_patch_info(self):
         self.assertEqual(ft.create_patch_info('foo.patch'), {'name': 'foo.patch'})
         self.assertEqual(ft.create_patch_info(('foo.patch', 1)), {'name': 'foo.patch', 'level': 1})
         self.assertEqual(ft.create_patch_info(('foo.patch', 'subdir')), {'name': 'foo.patch', 'sourcepath': 'subdir'})
-        self.assertEqual(ft.create_patch_info(('foo.txt', 'subdir')), {'name': 'foo.txt', 'copy': 'subdir'})
```
boegel, line 1638: Hmm, don't we still have cases like this, where the file specified in `patches` should just be copied?

Why does this case need to be removed exactly?

branfosj: It does not. I'd misunderstood what we were testing here. I've restored those in f1b58b3

test/framework/filetools.py
```diff
@@ -1646,25 +1646,11 @@ def test_create_patch_info(self):
...
-        self.mock_stderr(False)
-        self.disallow_deprecated_behaviour()
-        expected_warning = "Use of patch file with filename that doesn't end with correct extension: foo.txt "
-        expected_warning += "(should be any of: .patch, .patch.bz2, .patch.gz, .patch.xz)"
-        fail_msg = "Warning '%s' should appear in stderr output: %s" % (expected_warning, stderr)
-        self.assertIn(expected_warning, stderr, fail_msg)
-
-        # deprecation warning is treated as an error in context of unit test suite
-        expected_error = expected_warning.replace('(', '\\(').replace(')', '\\)')
-        self.assertErrorRegex(EasyBuildError, expected_error, ft.create_patch_info, 'foo.txt')
```
boegel, line 1662: These two lines should be retained, since `create_patch_info` should still produce an error when just `foo.txt` is provided to `create_patch_info`?

boegel (changes requested):

test/framework/filetools.py
branfosj: done in 01d0cea

boegel (approved):
lgtm

PR #4248, opened 2023-04-20
change `tar` command used in `get_source_tarball_from_git` to get reproducible tarballs

Rovanion:
Hi,
This pull request contains two commits performing two changes to easybuild that were needed to build a piece of Python software that was not located on the Python Package Index but only existed in a git repository.

The first change allows sources in `exts_list` to specify a `git_config`. The second change changes the command used to produce tarballs from git archives so that they are made in a reproducible manner. Otherwise each rerun of `eb` would result in a checksum mismatch.

These two patches have been tested on CentOS 7 and no other operating system. The command used to produce the tarball from git is complicated by the old versions of software that is found on CentOS/RHEL 7. But in the comments instructions on how to drop support for such outdated operating systems is also included.

If there are any questions, feel free to ask. 
Cheerio!


easybuild/framework/easyblock.py
```diff
@@ -662,7 +662,8 @@ def collect_exts_file_info(self, fetch_files=True, verify_checksums=True):
                         if fetch_files:
                             src_path = self.obtain_file(src_fn, extension=True, urls=source_urls,
                                                         force_download=force_download,
-                                                        download_instructions=download_instructions)
+                                                        download_instructions=download_instructions,
+                                                        git_config=ext_options.get('git_config'))
```
Micket, line 666: Hm, is this for the extensions to pick up git_config from `exts_default_options`? 

Because typically, that would be specific to each individual extension anyway, and a quick test
```
exts_list = [
    ('pytorch', '1.10.1', {
        'sources': [{'filename': SOURCE_TAR_GZ, 'git_config': {
            'url': 'https://github.com/pytorch',
            'repo_name': 'pytorch',
            'tag': 'v%(version)s',
            },
        }]
    }),
]
```
seems to download just fine for me without this change. I'm not actively opposed to it, but i can't imagine anyway this would ever apply to more than 1 extension anyway?

easybuild/tools/filetools.py
```diff
@@ -2707,11 +2707,16 @@ def get_source_tarball_from_git(filename, targetdir, git_config):
...
+    # When CentOS 7 is phased out and tar>1.28 is everywhere, replace find-sort-pipe with tar-flag
+    # '--sort=name' and place LC_ALL in front of tar. Also remove flags --null, --no-recursion, and
+    # --files-from - from the flags to tar. See https://reproducible-builds.org/docs/archives/
+    tar_cmd = ['find', repo_name, '-print0', '|',
+               'LC_ALL=C', 'sort', '--zero-terminated', '|',
+               'GZIP=--no-name', 'tar', '--create', '--file', targetpath, '--gzip',
+               '--mtime="1970-01-01 00:00Z"', '--owner=0', '--group=0', '--numeric-owner',
+               '--format=gnu', '--null',
+               '--exclude=.git' if not keep_git_dir else '',
+               '--no-recursion', '--files-from -']
```
Micket, line 2719: This looks well crafted. I would prefer to have this separate from the other change as it would apply independently.

Is this stable enough with possibly changing versions of gzip as well? We could also opt to not gzip at all if that would help reproducibility across systems.

Micket (dismissed):

easybuild/framework/easyblock.py
Rovanion: No, I made the change to pick up `git_config` directly from the options map. I found no example such as yours, displaying that it was possible to put the key `git_config` under `sources`.

I have removed this commit from the Pull Request.

easybuild/tools/filetools.py
Rovanion: It should be compatible with newer versions of GNU Gzip and Tar. It is incompatible with for example FreeBSD's Tar, based on a cursory look at its manpage which does not mention a `--mtime` flag.

Dropping the gzipping of the files would not help a lot; given that all changes but the flag `--no-name` are to Tar and not Gzip.

boegel: Although this approach could clearly help, but I have a couple concerns:

1) This may not be enough to get 100% reproducible tarballs across a variety of systems. I'm fairly sure that a different version of `gzip` being used may already be sufficient, since that may do better compression, etc. Another factor could be locale, which may affect the order of paths produced with the `find` command.
   If the tarball creation not 100% reproducible across a wide variety of systems, it's basically useless, since people who get a slightly different tarball will be running into incorrect checksums all the time, which would not be good.
2) The side effect of using `--mtime` to reset time stamps and `--owner=0` is that information gets lost. Especially the timestamp info may be interesting, so it should be retained.
3) This likely breaks the creating of tarball for easyconfigs using `git_config` on macOS, which is a regression. Although you currently won't get very far with installing software on macOS with EasyBuild, features like downloading sources with `eb --fetch` should work.

I applaud your effort here @Rovanion, but I'm not sure it's wise to go forward with this approach...

akesandgren: Changing mtime is an absolute do-not-do-that.

It also has the problem of not generating a tar file that has the same content as what tar zcf does.
A small test I did shows that the file/sort/tar pipe above actually adds each file twice in the tar file.

Example:
```
mkdir q
echo hej > q/t
echo hopp > q/w
find q | sort | tar --create --file q.tar --files-from -
tar cf q2.tar q
tar tf q2.tar 
q/
q/t
q/w
tar tf q.tar 
q/
q/t
q/w
q/t
q/w
```
this is caused by directory coming first, thereby making tar copy the whole dir into the tar file, then the list of files, making tar add the files once more.

Rovanion:
Hi,
Hopefully I can adress your concerns.

>
>
> 1.a. This may not be enough to get 100% reproducible tarballs across a
> variety of systems.


This patch could be insufficient, that is always a possibility, but I have
done my best to be compatible with old versions of GNU Tar. Given that the
patch is based on documentation from reproducible-builds.org, is it not
probable that it will improve the situation compared to the code in place
today?

The code can always be improved upon if a user of easybuild comes with a
report of the code not producing reproducible tarballs.

 1.b. I'm fairly sure that a different version of gzip being used may
> already be sufficient, since that may do better compression, etc.
>

I do not think this is the case. Sorting the files, zeroing out time and
user metadata - all of this will be included by default in all versions of
gzip that I am aware of, unless we tell gzip otherwise. The efficiency of
the compression algorithm itself is irrelevant for the reproducability of
the tarball.

 1.c. Another factor could be locale, which may affect the order of paths
> produced with the find command.
>

Which is why the results of find ar piped through sort with LC_ALL set to C.


> 1.d. If the tarball creation not 100% reproducible across a wide variety
> of systems, it's basically useless, since people who get a slightly
> different tarball will be running into incorrect checksums all the time,
> which would not be good.
>

I am not sure I understand. It is already possible to create tarballs from
git archives using easybuild today, right?

This patch, at a minimum, makes the creation of tarballs from git
reproducible on some systems whereas before they were never (?)
reproducible. I say never since git does not track file creation or
modification time, so they would always be set to the time at which the
repository was cloned. [1]  But I could be totally wrong here since I'm
relatively new to easybuild.

>
>
> 2. The side effect of using --mtime to reset time stamps and --owner=0 is
> that information gets lost. Especially the timestamp info may be
> interesting, so it should be retained.


This is the very core of the patch. These variables, data that differ
between downloads of the same git repository, must be zeroed out or set to
known values so that each tarball of the same git repository and
commit/tag/hash is the same. You cannot create a reproduceable tarball of a
git repository without erasing the mtime of the files on the local
filesystem in some manner.

>
>
> 3. This likely breaks the creating of tarball for easyconfigs using
> git_config on macOS, which is a regression. Although you currently won't
> get very far with installing software on macOS with EasyBuild, features
> like downloading sources with eb --fetch should work.
>
>
> Do you want me to run some test protocol on a mac? I could probably get
hold of such a machine.

[1]: https://stackoverflow.com/a/62039285/501017

Rovanion:
Hej Åke!

> Changing mtime is an absolute do-not-do-that.
>
Could you clearify why that is? If I understand correctly Git does not
record the mtime of the original source files. This means that the mtime of
the local files on disks at the point of archive creation only reflects
when the git repository was cloned, no more no less.

> It also has the problem of not generating a tar file that has the same
> content as what tar zcf does.
> A small test I did shows that the file/sort/tar pipe above actually adds
> each file twice in the tar file.
>
> Example:
>
> mkdir q
> echo hej > q/t
> echo hopp > q/w
> find q | sort | tar --create --file q.tar --files-from -
> tar cf q2.tar q
> tar tf q2.tar
> q/
> q/t
> q/w
> tar tf q.tar
> q/
> q/t
> q/w
> q/t
> q/w
>
> this is caused by directory coming first, thereby making tar copy the
> whole dir into the tar file, then the list of files, making tar add the
> files once more.
>
Thank you for finding the bug and the clear demonstration! I had missed
attaching the flag --no-recursion to the tar-invocation. I have pushed a
modified version of the patch that rectifies this shortcomming.

Thank you also for prodding me to have a look at contents of the archive my
version of easybuild produced. Turns out .git wasn't properly pruned
either. But the contents are removed in the latest version of the patch.

Rovanion:
Yeah I can do that. I'll see if I can get it in before vacation.

boegel (dismissed):

test/framework/filetools.py
```diff
@@ -2784,7 +2784,10 @@ def run_check():
         expected = '\n'.join([
             r'  running command "git clone --depth 1 --branch tag_for_tests %(git_repo)s"',
             r"  \(in .*/tmp.*\)",
-            r'  running command "tar cfvz .*/target/test.tar.gz --exclude .git testrepository"',
+            r'  running command "find testrepository -print0 -path \'*/.git\' -prune | LC_ALL=C sort --zero-terminated'
+            r' | GZIP=--no-name tar --create --file /tmp/eb-dm7hbbde/eb-6lulalhw/eb-vqfy4wts/target/test.tar.gz'
```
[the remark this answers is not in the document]
Rovanion: Found the all to easy solution of skipping that line, opened a PR against the docs on the subject: https://github.com/easybuilders/easybuild-docs/pull/175

Back to the subject of your comment: The hardcoded path has been removed from the tests and the tests pass on my installation with Python 3.6 on CentOS 7 run by `python -m test.framework.suite test_get_source_tarball_from_git`.

Rovanion:
The only workflow errors that seem to occur are running Python 2.7 and 3.5 which seem to have been (partly?) dropped from the list of supported Pythons to run Easybuild on?

Rovanion:
The check "Tests for Apptainer container support / build (3.7, 1.0.0) (pull_request)" have failed due to a fail to write a file, looks like the machine used for testing failed in this case.

Rovanion:
Is there anything else that needs to be done before this can be merged?

boegel:
@jfgrimm Can you post a comment with your findings with this, I think you did some testing and there was a problem with the new `tar` command on macOS?

jfgrimm:
@boegel @Rovanion I did some testing across a few different systems I have access to, with the following results:

-  on Ubuntu (22.04) and Rocky (8/9):
    - I got reproducible checksums when excluding the `.git` dir.
    - when not excluding the `.git` dir, checksums aren't consistent due to differences in several files
- on macOS, the `tar` command fails with the error: `tar: Option --mtime=1970-01-01 00:00Z is not supported`. This is because it uses `bsdtar` instead of GNU `tar`, which doesn't support the `--mtime` option (or any equivalent)

jfgrimm:
when not excluding `.git`, then:
* `.git/logs/HEAD`, `.git/logs/refs/heads/main`, `.git/logs/refs/remotes/origin/HEAD` contain unix timestamps from when the repo was cloned
* `.git/index` differs
* `.git/objects/pack/pack-*.{idx,pack}` have different names, and contents

the unix timestamps are easy enough to deal with, but the metadata encoded in the index and pack files is more complicated

boegel:
@Rovanion Any idea if the issues that @jfgrimm raised can be fixed?

lexming:
I brought this PR to the last EB5 conf call and we might have a solution for it:

1. use a common set of features between GNU tar and BSD tar to guarantee compatibility
2. discard `--mtime` due to point 1 and instead manually set the timestamps of each file with `touch -t 197001010100` prior to `tar`
3. use reproducible tarballs for sources from git repos that are pulled **without** `.git` (that is the default behaviour)
4. sources from git repos explicitly setting `keep_git_dir=True` will be kept with `.git` and won;t be checksumed, because we can't as discussed here

@Rovanion is this a workable solution for you?

lexming:
A little proof of concept with `touch`:
```bash
$ git clone https://github.com/easybuilders/easybuild
Cloning into 'easybuild'...
remote: Enumerating objects: 11089, done.
remote: Counting objects: 100% (909/909), done.
remote: Compressing objects: 100% (457/457), done.
remote: Total 11089 (delta 482), reused 851 (delta 442), pack-reused 10180
Receiving objects: 100% (11089/11089), 595.66 MiB | 10.09 MiB/s, done.
Resolving deltas: 100% (7324/7324), done.
# Delete .git folder
$ rm -rf easybuild/.git
# Checksum of downloaded files
$ tar --create --file easybuild.tar easybuild/
$ sha256sum easybuild.tar
98457ba141e829641878550bcf82681c87380acd106d695235385043dc7327ad  easybuild.tar
# Checksum of downloaded files with timestamp reset
$ find easybuild/ -exec touch -t 197001010100 {} \;
$ tar --create --owner=0 --group=0 --numeric-owner --format=gnu --null --file easybuild.0.tar easybuild/
$ sha256sum easybuild.0.tar
e1c54e194764bcd89165da527c25b4f5d82fcbd132be9d5f1f75ee40d9eb6ddc  easybuild.0.tar
# Wait some time and repeat the clone + touch + tar
$ sha256sum easybuild.1.tar
e1c54e194764bcd89165da527c25b4f5d82fcbd132be9d5f1f75ee40d9eb6ddc  easybuild.1.tar
```

Rovanion:
Yes, that would work. I'm not available to work on EasyBuild at the moment, I have been directed to prioritize other projects. Feel free to implement this until I am able to work on EasyBuild again.

lexming (changes requested):
Thanks for the feedback. I made a PR to this one with the discussed changes. Please check https://github.com/Rovanion/easybuild-framework/pull/1

easybuild/tools/filetools.py
```diff
@@ -2668,17 +2667,34 @@ def get_source_tarball_from_git(filename, targetdir, git_config):
...
+        tar_cmd = ['tar', 'cfvz', archive_path, repo_name]
     else:
-        tar_cmd = ['tar', 'cfvz', targetpath, '--exclude', '.git', repo_name]
+        # create reproducible archive
+        # see https://reproducible-builds.org/docs/archives/
+        tar_cmd = [
+            # print names of all files and folders excluding .git directory
+            'find', repo_name, '-name ".git"', '-prune', '-o', '-print0',
+            # reset access and modification timestamps
+            '-exec', 'touch', '-t 197001010100', '{}', '\;', '|',
```
lexming, line 2683: I missed using a raw string for this bit:
```suggestion
            '-exec', 'touch', '-t 197001010100', '{}', r'\;', '|',
```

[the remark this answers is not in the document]
lexming: @Rovanion I missed using a raw string for this bit:
```suggestion
            '-exec', 'touch', '-t 197001010100', '{}', r'\;', '|',
```

boegel:
@Rovanion Problems with Apptainer tests was fixed in #4472, which was also applied to the `5.0.x` branch (so you'll need to sync up your PR branch with the current `5.0.x` branch)

lexming (commented):
@Rovanion I opened another PR to this one that just merges the recent changes in the `5.0.x` branch. Unfortunately, at the time of the previous merge, unit tests were broken. Can you please merge https://github.com/Rovanion/easybuild-framework/pull/2?

Rovanion:
I think all tests have passed now. All that remains are the three reviews.

lexming (approved):
@Rovanion Thanks a lot for keeping up with the updates!

LGTM

Rovanion:
Cheers!

Den mån 18 mars 2024 kl 22:57 skrev Alex Domingo ***@***.***>:

PR #4249, opened 2023-04-21
make the RPATH section check with `readelf -d` in sanity check optional 

bedroge:
See https://github.com/easybuilders/easybuild-easyblocks/pull/2913 and https://github.com/easybuilders/easybuild-easyconfigs/issues/17516, this makes it easier for easyblocks like `GoPackage` to skip this check. The check is still enabled by default.

Additionally, it would be nice to have an easyconfig parameter to disable the check for individual packages (e.g. `git-lfs`), but that can be done in a follow-up PR.


casparvl:
I haven't looked in detail, but I think this PR might need a test that intentionally installs something without `RPATH`, and checks that it passes if this RPATH skipping option is used. (I'm _assuming_ there already is a test that checks if the `RPATH` sanity check fails if the RPATH is _not_ set, `RPATH` support is used, and the RPATH sanity check fails. In case that's not there, it should also be added)

boegel (approved):
lgtm

boegel:
@casparvl You make a good point about adding a test, but all this is doing is adding support for using `check_readelf_rpath=False` in an easyblock when calling out to the parent `sanity_check_rpath`, it doesn't actually change anything at all.

Implementing a good test for this would be quite involved, and since this is blocking https://github.com/easybuilders/easybuild-easyblocks/pull/2913, I'll go ahead and merge this...

PR #4256, opened 2023-05-11
Debug bootstrap test on CI

Flamefire:
The current failure looks like

```
 == FAILED: Installation ended unsuccessfully (build directory: /tmp/tmp1ec454qx/EasyBuild/4.7.1/dummy-dummy): build failed (first 300 chars): cmd " /opt/hostedtoolcache/Python/3.7.16/x64/bin/python -m pip install --prefix=/tmp/runner/200614b91caee792396c8fd6365e015e1bb9145f/eb_bootstrap/software/EasyBuild/4.7.1  --no-deps  --ignore-installed  --no-build-isolation  ." exited with exit code 1 and output:
Processing /tmp/tmp1ec454qx/EasyBuil (took 4 secs)
== Results of the build can be found in the log file(s) /tmp/eb-u9niem1r/easybuild-EasyBuild-4.7.1-20230511.110258.MiSOu.log
```

This isn't helpful so in a first step this prints out the log file in case of error.

### DIFF
--- a/.github/workflows/bootstrap_script.yml
+++ b/.github/workflows/bootstrap_script.yml
@@ -127,7 +127,11 @@ jobs:
           # test bootstrap script
           export PREFIX=/tmp/$USER/$GITHUB_SHA/eb_bootstrap
           export EASYBUILD_BOOTSTRAP_DEPRECATED=1
-          python easybuild/scripts/bootstrap_eb.py $PREFIX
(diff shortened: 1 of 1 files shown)


branfosj:
See #4255. The bootstrap is deprecated and will be removed in EB5, so we'll just remove the workflow that tests it instead.

PR #4280, opened 2023-06-11
replace imp.load_source

### LINKED ISSUE #4264: stop relying on `imp` (will no longer be available in Python 3.12)
`imp` is deprecated since Python 3.4, replaced with `importlib`, will be removed in Python 3.12 stdlib, cfr. https://docs.python.org/3.11/library/imp.html

Currently used in:
* [ ] `easybuild/tools/filetools.py`
* [ ] `easybuild/tools/hooks.py`

branfosj:
fixes  #4264

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -44,7 +44,6 @@ import datetime
 import difflib
 import glob
 import hashlib
-import imp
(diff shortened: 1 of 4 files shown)


bartoldeman (approved):
LGTM

Issue #4252, opened 2023-05-03
implement new `run` function with cleaner API to replace `run_cmd` + `run_cmd_qa`


boegel:
The current `run_cmd` and `run_cmd_qa` functions have a very confusing API, and their implementation is ancient (dates back from a time when we still supported Python 2.4).

We should implement a new `run` function, which supports running both interactive and non-interactive commands, as well as more advanced features like running a command asynchronously, and providing support to drop into an interactive shell when a command failed.

boegel:
I had a quick brainstorm on this with @lexming, here's the plan we came up with...

### Goals

- one single `run` function that supports running (interactive) shell commands (incl. Lmod)
- sane API + defaults
- supports everything that current `run_cmd` and `run_cmd_qa` functions do
- enables new features like:
    - context-aware error parsing & reporting
    - wrapper functions for common commands like `run_make`, `run_cmake`, etc.
    - opt-in to jumping into an interactive shell session if command failed

### Current `run_cmd` and `run_cmd_qa` functions

```python
def run_cmd(cmd, log_ok=True, log_all=False, simple=False, inp=None, regexp=True, log_output=False, path=None,
            force_in_dry_run=False, verbose=True, shell=None, trace=True, stream_output=None, asynchronous=False):
    """
    Run specified command (in a subshell)
    :param cmd: command to run
    :param log_ok: only run output/exit code for failing commands (exit code non-zero)
    :param log_all: always log command output and exit code
    :param simple: if True, just return True/False to indicate success, else return a tuple: (output, exit_code)
    :param inp: the input given to the command via stdin
    :param regexp: regex used to check the output for errors;  if True it will use the default (see parse_log_for_error)
    :param log_output: indicate whether all output of command should be logged to a separate temporary logfile
    :param path: path to execute the command in; current working directory is used if unspecified
    :param force_in_dry_run: force running the command during dry run
    :param verbose: include message on running the command in dry run output
    :param shell: allow commands to not run in a shell (especially useful for cmd lists), defaults to True
    :param trace: print command being executed as part of trace output
    :param stream_output: enable streaming command output to stdout
    :param asynchronous: run command asynchronously (returns subprocess.Popen instance if set to True)
    """
```

```python
def run_cmd_qa(cmd, qa, no_qa=None, log_ok=True, log_all=False, simple=False, regexp=True, std_qa=None, path=None,
               maxhits=50, trace=True):
    """
    Run specified interactive command (in a subshell)
[...]

PR #4284, opened 2023-06-22
initial implementation of `run` function to replace `run_cmd` + `run_cmd_qa`

boegel:
cfr. #4252

Marked as draft because:

* need to check which other tests can be copied to also cover `run` with the current implementation;
* should switch from `run_cmd` to `run` where possible;

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -44,6 +44,7 @@ import subprocess
 import sys
 import tempfile
 import time
+from collections import namedtuple
 from datetime import datetime
 
 import easybuild.tools.asyncprocess as asyncprocess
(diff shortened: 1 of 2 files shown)


boegel:
@branfosj Up for taking a look at this and providing some feedback?

easybuild/tools/run.py
```diff
@@ -73,6 +74,105 @@
...
+
+    start_time = datetime.now()
+    if not hidden:
+        cmd_trace_msg(cmd_msg, start_time, work_dir, stdin, cmd_out_fp)
+
+    _log.info(f"Running command '{cmd_msg}' in {work_dir}")
+    proc = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=shell)
+
+    # return output as a regular string (UTF-8 characters get stripped out)
+    output = proc.stdout.decode('ascii', 'ignore')
```
lexming, line 145: now that we only support Python 3 we can just accept unicode strings by default

easybuild/tools/run.py
```diff
@@ -73,6 +74,105 @@
...
+        trace_msg(f"command completed: exit {res.exit_code}, ran in {time_since_start}")
+
+    return res
+
+
+def cmd_trace_msg(cmd, start_time, work_dir, stdin, cmd_out_fp):
+    """
+    Helper function to construct and print trace message for command being run
+    """
+    lines = [
```
branfosj, line 161: We should make more use of f-strings in here.

branfosj (commented):
I've gone through here and considered the discussion in https://github.com/easybuilders/easybuild-framework/issues/4252 and I'm happy with this appraoch.

easybuild/tools/run.py
```diff
@@ -73,6 +74,105 @@
 ]
 
 
+RunResult = namedtuple('RunResult', ('output', 'exit_code', 'stderr'))
+
+
+def run(cmd, fail_on_error=True, split_stderr=False, stdin=None,
+        hidden=False, in_dry_run=False, work_dir=None, shell=True,
+        output_file=False, stream_output=False, asynchronous=False,
+        qa_patterns=None, qa_wait_patterns=None):
```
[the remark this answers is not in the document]
boegel: I'm fine with putting the options in alphabetical order, but the order used now isn't random: the most used options are basically listed first, and they're more-or-less grouped by type of option (like the `qa_*` ones).

Let's get some output from others on this (@lexming?), and not block the PR over this, since this is trivial to change in a follow-up PR if we care strongly enough.

Is there a way to selectively enforce alphabetical ordering for particular functions with `flake8`?

easybuild/tools/run.py
boegel: OK, so we can use `utf-8` rather than `ascii` as encoding, but we still need to decode the `proc.stdout` bytes value to a string value, or we have to use byte sequence everywhere we play with the command output (which we really don't want to do).
In theory, the command output could contain non-UTF-8 characters (however unlikely), we still want to ignore those when decoding to avoid running into trouble.

easybuild/tools/run.py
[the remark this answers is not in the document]
boegel: I'm inclined to keep this as is for now, as long as we make sure to make no assumptions about the order of named options when calling `run`, we can reorder things later here...

easybuild/tools/run.py
boegel: done in 481883d3452c23c67341203e04e771ca233938ea

easybuild/tools/run.py
boegel: Switched to `utf-8` encoding in 1794660abf59855b22c59908dbf89ad761e03d0f

boegel:
Fix for broken tests is available in #4306

test/framework/run.py
```diff
@@ -159,6 +159,31 @@ def test_run_cmd(self):
...
+            res = run("echo hello")
+        self.assertEqual(res.output, "hello\n")
+        # no reason echo hello could fail
+        self.assertEqual(res.exit_code, 0)
+        self.assertEqual(type(res.output), str)
+
+        # test running command that emits non-UTF-8 characters
+        # this is constructed to reproduce errors like:
+        # UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2
+        # UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018'
```
ocaisa, line 175:
```suggestion
        # UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018'
        # (such errors are ignored by the 'run' implementation)
```

ocaisa: This was a bit confusing since you say they are constructed to raise errors, but there are no errors later

ocaisa (changes requested):
LGTM

Still a lot to go as regards implementation, but as long as the tests are there for each newly implemented option this should be fine.

ocaisa (approved):

Issue #4285, opened 2023-06-23
GitHub tests broken


branfosj:
In #4253 I forgot to test the GitHUb tests and they are broken - see https://github.com/easybuilders/easybuild-framework/actions/runs/5355547822/jobs/9713965558#step:9:206

PR #4287, opened 2023-06-23
suggest default title in `--review-pr`

branfosj:
Follow on from #4285. In `--review-pr` we can suggest what the default title would be for the easyconfigs found.


migueldiascosta:
lgtm; it's useful, doesn't affect any existing tests afaict, and I suppose we don't need to add an explicit test for this

boegel (approved):
lgtm

PR #4304, opened 2023-07-25
Implement `build_and_install_loop` hooks to run before and after the install loop for individual modules

### LINKED ISSUE #4303: Implement hook to run after easyconfigs are parsed but before the build starts
A hook that is run which is passed all of the easyconfigs would be very helpful. It would be passed a copy of `ordered_ecs`, so it can extract information from it.

The hook would then use the knowledge about all the easyconfigs to be built for various operations such as notifying external tools about the number of easyconfigs, and what they are, which could even be used to estimate build time.

I plan to implement this feature myself

XavierCS-dev:
- Add ~POISED~ BUILD_AND_INSTALL_LOOP hook to list of hooks
- Set ~POISED~ BUILD_AND_INSTALL_LOOP hook to run after easyconfigs are parsed + after install loop
    - Pass ~POISED~ BUILD_AND_INSTALL_LOOP hook an ordered list of the easyconfigs, to be processed by external tools.

fixes #4303


easybuild/main.py
```diff
@@ -543,6 +543,7 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
     # build software, will exit when errors occurs (except when testing)
     if not testing or (testing and do_build):
         exit_on_failure = not (options.dump_test_report or options.upload_test_report)
+        run_hook(POISED, hooks, args=[ordered_ecs])
```
boegel, line 546: Why not make this call *in* `build_and_install_software` instead?

easybuild/tools/hooks.py
```diff
@@ -61,6 +61,7 @@
 
 START = 'start'
 PARSE = 'parse'
+POISED = 'poised'
```
boegel, line 64: Not sure about `poised`, maybe we can find a better name (see also #4303)

easybuild/main.py
```diff
@@ -69,7 +69,7 @@
 from easybuild.tools.github import add_pr_labels, install_github_token, list_prs, merge_pr, new_branch_github, new_pr
 from easybuild.tools.github import new_pr_from_branch
 from easybuild.tools.github import sync_branch_with_develop, sync_pr_with_develop, update_branch, update_pr
-from easybuild.tools.hooks import START, END, load_hooks, run_hook
+from easybuild.tools.hooks import LOOP, PRE_PREF, POST_PREF, START, END, load_hooks, run_hook
```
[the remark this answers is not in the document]
XavierCS-dev: It will be used soon!

easybuild/main.py
XavierCS-dev: Looking at this, I believe it would require another argument containing the hooks to be passed to `build_and_install_software`. Could be wrong though.

test/framework/options.py
```diff
@@ -679,6 +679,7 @@ def test_avail_hooks(self):
             "List of supported hooks (in order of execution):",
             "	start_hook",
             "	parse_hook",
+            "   pre_build_and_install_loop_hook"
```
branfosj:
```suggestion
            "	pre_build_and_install_loop_hook",
```

test/framework/options.py
```diff
@@ -716,6 +717,7 @@ def test_avail_hooks(self):
             "	post_package_hook",
             "	pre_testcases_hook",
             "	post_testcases_hook",
+            "   post_build_and_install_loop_hook"
```
branfosj:
```suggestion
            "	post_build_and_install_loop_hook",
```

test/framework/options.py
```diff
@@ -679,6 +679,7 @@ def test_avail_hooks(self):
             "List of supported hooks (in order of execution):",
             "	start_hook",
             "	parse_hook",
+            "   pre_build_and_install_loop_hook",
```
branfosj, line 682:
```suggestion
            "	pre_build_and_install_loop_hook",
```

test/framework/options.py
```diff
@@ -716,6 +717,7 @@ def test_avail_hooks(self):
             "	post_package_hook",
             "	pre_testcases_hook",
             "	post_testcases_hook",
+            "   post_build_and_install_loop_hook",
```
branfosj, line 720:
```suggestion
            "	post_build_and_install_loop_hook",
```

test/framework/options.py
branfosj, line 682:
```suggestion
            "	pre_build_and_install_loop_hook",
```

branfosj (commented):
Tabs instead of spaces after the `"`.

easybuild/main.py
boegel: You can add `hooks = load_hooks(eb_go.options.hooks)` in `build_and_install_software`, which is fine since the loaded hooks are cached - see https://github.com/easybuilders/easybuild-framework/blob/b8c153df1a9299d256a482413b5bca0321e72e1b/easybuild/tools/hooks.py#L109

easybuild/tools/hooks.py
boegel: Just for future reference: as discussed in Slack: `build_and_install_loop` is clearer, so let's go with that (for lack of a better naming)

easybuild/main.py
boegel: On second thought, let's leave it as is, since the way it's done now makes the relation to `build_and_install_software` very clear, especially since there's now a `pre_build_and_install_loop` and `post_build_and_install_loop`

boegel:
@XavierCS-dev I assume this is good to go, but I'll let you mark it as "Ready for review" first...

Are you up for also updating the documentation (https://docs.easybuild.io/hooks) accordingly, via a PR to https://github.com/easybuilders/easybuild-docs ?

PR #4312, opened 2023-08-06
fix `test_add_and_remove_module_path` by replacing string comparison onf paths by checking whether they point to the same path (since symlinks may cause trouble)

boegel:


ocaisa (approved):
LGTM

PR #4314, opened 2023-08-07
implement `fail_on_error`/`in_dry_run`/`output_file` options + enable caching for `run` function, and switch from `run_cmd` to `run` function in systemtools

boegel:


easybuild/tools/run.py
```diff
@@ -77,6 +77,36 @@
...
 
+def run_cmd_cache(func):
+    """Function decorator to cache (and retrieve cached) results of running commands."""
+    cache = {}
+
+    @functools.wraps(func)
+    def cache_aware_func(cmd, *args, **kwargs):
+        """Retrieve cached result of selected commands, or run specified and collect & cache result."""
+        # cache key is combination of command and input provided via stdin
+        key = (cmd, kwargs.get('inp', None))
```
ocaisa, line 88: I can't see a `inp` kwarg anywhere, where does it come from?

ocaisa: Should it be `stdin`?

ocaisa: Ah ok, you've renamed `inp` to `stdin` in the new `run` command. You're going to have to put in the logic to do either for now
```suggestion
        key = (cmd, kwargs.get('stdin', kwargs.get('inp', None)))
```

boegel: Ah, that's specific to `run_cmd`, so this needs to be updated to also consider `stdin` used in `run`, nice catch!

I'll look into fixing it...

boegel: Fixed in c045f2388, problem is caught by additional test `test_run_cache` that was added

easybuild/tools/run.py
```diff
@@ -142,14 +177,23 @@ def run(cmd, fail_on_error=True, split_stderr=False, stdin=None,
         # 'input' value fed to subprocess.run must be a byte sequence
         stdin = stdin.encode()
 
+    # use bash as shell instead of the default /bin/sh used by subprocess.run
+    # (which could be dash instead of bash, like on Ubuntu, see https://wiki.ubuntu.com/DashAsBinSh)
+    if shell:
+        executable = '/bin/bash'
```
ocaisa, line 183: This isn't guaranteed to exist, you should check it does or use our `which()` function

ocaisa (changes requested):
Looks good, just one issue

easybuild/tools/run.py
boegel: I think it's a fair assumption that it's there, and if it's not the error message will be quite clear:

```
>>> subprocess.run("echo test", shell=True, executable='/bin/alan')
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1821, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/bin/alan'
```

Isn't that good enough, to let EasyBuild crash hard if `/bin/bash` is missing?

I don't want to check whether `/bin/bash` is *still* there every time a command is being run with `run`...
I guess I can implement a one-time check somewhere at startup, and produce a single-line error like "`ERROR: /bin/bash does not exist, EasyBuild really needs it!"`, but is that worth the effort?
In fact, if `/bin/bash` is not there, the `eb` wrapper script will fail spectacularly already, since it has `#!/bin/bash` in the shebang line:

```
bash-3.2$ eb --version
bash: eb: /bin/alan: bad interpreter: No such file or directory
```

ocaisa (approved):

PR #4315, opened 2023-08-08
Implement hooks for failure scenarios

### LINKED ISSUE #4302: Implement hooks for failure scenarios
Looking through the code base, I cannot find any hooks for scenarios where an `EasyBuildError` or `KeyboardInterrupt` occurs. Such hooks would be useful as they could provide more information on the errors to external tools.

For example, I am improving a slack bot which provides build information in a private slack channel, and having failure information posted there would be very helpful.

This is a feature I plan to add myself.

XavierCS-dev:
Implements 3 hooks for failure scenarios:

- `fail_hook`: occurs when an `EasybuildError` is encountered.
- `cancel_hook`: occurs when the user issues a keyboard interrupt.
- `crash_hook`: occurs when a generic python `Exception` is encountered.

These hooks can provide external information on which errors occurred, so users taking advantage of hooks can take automated action based on which error occurred. These hooks can also be used to export the errors to external tools.

fixes #4302


boegel (dismissed):
lgtm

boegel:
Some merge conflicts will pop up here if https://github.com/easybuilders/easybuild-framework/pull/4304 gets merged first...

boegel (changes requested):

easybuild/main.py
```diff
@@ -732,9 +733,16 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
 
 
 if __name__ == "__main__":
+    eb_go = parse_options()
```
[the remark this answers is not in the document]
XavierCS-dev: I tried updating the arguments to main, so main has the extra values available to it, however, it seems it causes a large number of tests to fail, as they seem to rely on what the arguments were previously

XavierCS-dev:
Chaning
```eb_go, cfg_settings = set_up_configuration(args=args, logfile=logfile, testing=testing)```

to be

```eb_go, cfg_settings = set_up_configuration(args=None, logfile=None, testing=None)```
in `main` may be causing a dependency issue when these arguments are set in testing.
Having `parse_options()` called twice may be the only way around it.

File: easybuild/main.py [hunk 2]
```diff
@@ -600,9 +603,11 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
     if '_' not in os.environ:
         os.environ['_'] = sys.executable
 
+    if any([args, logfile, testing]):
```
boegel, line 606: I would also do this if either `eb_go` or `cfg_settings` are `None`, so if someone calls `main()` it still works as before?

easybuild/main.py
```diff
@@ -600,9 +603,11 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+        eb_go, cfg_settings = set_up_configuration(args=args, logfile=logfile, testing=testing)
+
     # purposely session state very early, to avoid modules loaded by EasyBuild meddling in
     init_session_state = session_state()
```
boegel, line 610: move this up, do this *before* calling `set_up_configuration` as before (that may be important because `set_up_configuration` already changes stuff in the environment, like `$TMPDIR` for example)

easybuild/main.py
```diff
@@ -731,10 +736,29 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
         cleanup(logfile, eb_tmpdir, testing, silent=False)
 
 
+def prepare_main(args=None, logfile=None, testing=None):
+    """
+    Prepare Main: Set up eb_go and consequently, cfg_settings in order to obtain the hooks,
+    and prevent the configuration being set up twice.
```
boegel, line 742: This doesn't prevent anything, and it's not specific to hooks in any way.

I would just use: `Prepare for calling main function by setting up the EasyBuild configuration.`

easybuild/main.py
```diff
@@ -731,10 +736,29 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+    :param args: command line arguments to take into account when parsing the EasyBuild configuration settings
+    :param logfile: log file to use
+    :param testing: enable testing mode
```
boegel, line 745: should also specify `returns`?

easybuild/main.py
```diff
@@ -579,9 +580,11 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
     return overall_success
 
 
-def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
+def main(eb_go=None, cfg_settings=None, args=None, logfile=None, do_build=None, testing=False, modtool=None):
```
boegel, line 583: better add these at the end, to avoid breaking existing uses of `main` like `main(args)`

easybuild/main.py
```diff
@@ -731,10 +736,29 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
...
+    eb_go, cfg_settings = set_up_configuration(args=args, logfile=logfile, testing=testing)
+    return eb_go, cfg_settings
+
+
 if __name__ == "__main__":
+    eb_go, cfg_settings = prepare_main()
+    hooks = load_hooks(eb_go.options.hooks)
     try:
-        main()
+        main(eb_go, cfg_settings)
```
boegel, line 755: use named options:

```suggestion
        main(eb_go=eb_go, cfg_settings=cfg_settings)

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/main.py [hunk 8]
```diff
@@ -579,30 +580,23 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
...
-    # if $CDPATH is set, unset it, it'll only cause trouble...
-    # see https://github.com/easybuilders/easybuild-framework/issues/2944
-    if 'CDPATH' in os.environ:
-        del os.environ['CDPATH']
-
-    # When EB is run via `exec` the special bash variable $_ is not set
-    # So emulate this here to allow (module) scripts depending on that to work
-    if '_' not in os.environ:
-        os.environ['_'] = sys.executable
+    if not all([eb_go, cfg_settings, init_session_state, hooks]) or hooks is None or any([args, logfile, testing]):
```
branfosj, line 597:
```suggestion
    if not all([eb_go, cfg_settings, init_session_state]) or hooks is None or any([args, logfile, testing]):
```

easybuild/main.py
```diff
@@ -579,30 +580,23 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
     return overall_success
 
 
-def main(args=None, logfile=None, do_build=None, testing=False, modtool=None):
+def main(args=None, logfile=None, do_build=None, testing=False, modtool=None, eb_go=None, cfg_settings=None,
+         init_session_state=None, hooks=None):
```
boegel, line 584: no need to pass down `hooks`, and it's probably better/cleaner to only introduce a single new argument to `main`, like `prepared_cfg_data` which includes `init_session_state`, `eb_go`, and `cfg_settings`

fixed in https://github.com/XavierCS-dev/easybuild-framework/pull/1

boegel (changes requested):

easybuild/main.py
XavierCS-dev: Merged

boegel (approved):

PR #4316, opened 2023-08-09
add postiter hook to the list of steps so the corresponding hook can be used

branfosj:
This will allow us to use it!


branfosj:
> Does postiter count as a step? If so I imagine it would have a `pre` and `post` version. I am not entirely sure though.

Indeed, it is a step. I've updated the PR.

PR #4318, opened 2023-08-09
change openmp flag to be a string for intel_compilers

branfosj:
In https://github.com/easybuilders/easybuild-easyconfigs/pull/17707#issuecomment-1601094327 we are seeing `-['fiopenmp']` in the flags.

https://github.com/easybuilders/easybuild-easyblocks/blob/4fbf1443d22628f490e1f943a0a370c1e3e32e12/easybuild/easyblocks/q/quantumespresso.py uses `self.toolchain.get_flag('openmp')`, but
```py
    def get_flag(self, name):
        """Get compiler flag for a certain option."""
        return "-%s" % self.options.option(name)
```
Which is causing the issue:
```
>>> openmp_flag = ['fiopenmp']
>>> "-%s" % openmp_flag
"-['fiopenmp']"
```

There is a more complete change where we redo the whole toolchain and flags setup. However, that would be large change and this bugfix will be sufficient for now.


ocaisa:
@branfosj Can you open an issue on this in general so it's not forever lost that we probably need to change `get_flag` to be able to handle lists of options

PR #4321, opened 2023-08-09
implement support in `run` function for splitting stdout and stderr output + passing down environment to use, and switch to `run` function for running `module` commands

boegel:


branfosj (dismissed):

ocaisa (changes requested):
Looks good to me, just needs the changes @branfosj suggested

ocaisa:
I committed the changes from @branfosj as they are trivial and consistent with previous usage

ocaisa (approved):

PR #4323, opened 2023-08-11
add `run_shell_cmd` hook

boegel:
* 😎 : `run_cmd_shell` hook that triggers before/after running a shell command (works for both `run_cmd` and `run_cmd_qa`);
* ⚠️ some exceptions necessary for commands run in `systemtools` + `set_tmpdir`, because they are run before EasyBuild configuration is properly set up (so `build_option` doesn't work yet);
* :fire: `pre_run_cmd_shell_hook` can modify the command before running it;
* 📖 additional context is available:
  * working directory;
  * output + exit code for `post_run_cmd_shell_hook`;
  * discriminate between interactive and non-interactive commands;
* ✅  tests have been added/extended to ensure implementation works as designed;


jfgrimm (approved):
lgtm

jfgrimm:
Going in, thanks @boegel!

PR #4324, opened 2023-08-12
add `build_info_msg` easyconfig parameter

branfosj:
In some easyconfigs we have comments that document alternative or extras that may be applicable to some people, but these are hidden away from users. Adding this option gives us a way to flag such information during a build.

An example of where I think this would be useful is for the NLTK datasets - see https://github.com/easybuilders/easybuild-easyconfigs/pull/18550


boegel (approved):
lgtm

PR #4327, opened 2023-08-15
implement support in '`run`' function for running command in different working directory + switch to `run` function in filetools

boegel:


easybuild/tools/run.py
```diff
@@ -108,7 +108,7 @@ def cache_aware_func(cmd, *args, **kwargs):
 
 @run_cache
 def run(cmd, fail_on_error=True, split_stderr=False, stdin=None, env=None,
-        hidden=False, in_dry_run=False, work_dir=None, shell=True,
+        hidden=False, in_dry_run=False, verbose_dry_run=False, work_dir=None, shell=True,
```
boegel, line 111: having both `hidden` (to disable output in both trace and dry run mode) and `verbose_dry_run` is necessary evil, since there are situations where we do not want trace output, but we do want to see commands that would be executed in dry run mode (primarily in `get_source_tarball_from_git`)

test/framework/lib.py
```diff
@@ -92,8 +85,24 @@ def test_run_cmd(self):
         self.assertEqual(ec, 0)
         self.assertEqual(out, 'hello\n')
 
+    def test_run(self):
+        """Test use of runfunction in the context of using EasyBuild framework as a library."""
```
jfgrimm, line 89:
```suggestion
        """Test use of run function in the context of using EasyBuild framework as a library."""
```

jfgrimm (approved):
lgtm

PR #4331, opened 2023-08-16
only add extensions in module file if there are extensions

### LINKED ISSUE #4330: Avoid adding extension statement in module files if there are none
I'm seeing modules with:
```lua
if convertToCanonical(LmodVersion()) >= convertToCanonical("8.2.8") then
    extensions("")
end
```

If the extension list is empty, it should simply not be present.

wpoely86:
This fixes #4330.

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -622,7 +622,8 @@ class ModuleGenerator(object):
         """
         Generate a list of all extensions in name/version format
         """
-        return self.app.make_extension_string(name_version_sep='/', ext_sep=',').split(',')
+        exts_str = self.app.make_extension_string(name_version_sep='/', ext_sep=',')
+        return exts_str.split(',') if exts_str else []
 
     def _generate_help_text(self):
         """
--- a/test/framework/module_generator.py
+++ b/test/framework/module_generator.py
@@ -756,6 +756,16 @@ class ModuleGeneratorTest(EnhancedTestCase):
             regex = re.compile(pattern, re.M)
             self.assertTrue(regex.search(desc), "Pattern '%s' found in: %s" % (regex.pattern, desc))
 
+        # check if the extensions is missing if there are no extensions
+        test_ec = os.path.join(test_dir, 'easyconfigs', 'test_ecs', 't', 'toy', 'toy-0.0-test.eb')
+
+        ec = EasyConfig(test_ec)
+        eb = EasyBlock(ec)
+        modgen = self.MODULE_GENERATOR_CLASS(eb)
+        desc = modgen.get_description()
+
+        self.assertFalse(re.search(r"\s*extensions\(", desc), "No extensions found in: %s" % desc)
+
     def test_prepend_paths(self):
         """Test generating prepend-paths statements."""
         # test prepend_paths


wpoely86:
I'm unsure of the point of `_make_extension_list`? Can't we inline it?

boegel:
> I'm unsure of the point of `_make_extension_list`? Can't we inline it?

@wpoely86 Both `make_extension_string` and `_make_extension_list` were added in #3697 deliberately such that easyblocks can customize them, so they should be both retained.

wpoely86:
> > I'm unsure of the point of `_make_extension_list`? Can't we inline it?
> 
> @wpoely86 Both `make_extension_string` and `_make_extension_list` were added in #3697 deliberately such that easyblocks can customize them, so they should be both retained.

Sure, I've kept it but the point is totally unclear to me. Why would you want to override `_make_extension_list`? 

But anyway. Please merge :)

boegel:
I do admit the use case isn't entirely clear to me either, but we should be careful when changing the API, in particular that of the `EasyBlock` class.

easybuild/tools/module_generator.py
```diff
@@ -622,7 +622,7 @@ def _generate_extensions_list(self):
         """
         Generate a list of all extensions in name/version format
         """
-        return self.app.make_extension_string(name_version_sep='/', ext_sep=',').split(',')
+        return self.app.make_extension_list(name_version_sep='/')
```
boegel, line 625: Not sure this is the right approach, since it effectively bypasses `make_extension_string`, so easyblock that implement a custom `make_extension_string` won't work as expected anymore...

boegel:
@wpoely86 Rather than fiddling in `easyblock.py` and adding to/changing the API, how about this fix instead?

```diff
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -622,7 +622,8 @@ class ModuleGenerator(object):
         """
         Generate a list of all extensions in name/version format
         """
-        return self.app.make_extension_string(name_version_sep='/', ext_sep=',').split(',')
+        exts_str = self.app.make_extension_string(name_version_sep='/', ext_sep=',')
+        return exts_str.split(',') if exts_str else []

     def _generate_help_text(self):
         """
```

easybuild/tools/module_generator.py
wpoely86: 1) I can't find a single easyblock (merged or in a PR) that touches `make_extension_string`
2) I have no clue how to else do this. The original code is weird as it goes from list->string->list.
3) the only use of `make_extension_string` is for generating the EBEXTLIST variables in the module. Totally different case so it would be good to split it.

wpoely86:
> @wpoely86 Rather than fiddling in `easyblock.py` and adding to/changing the API, how about this fix instead?
> 
> ```diff
> --- a/easybuild/tools/module_generator.py
> +++ b/easybuild/tools/module_generator.py
> @@ -622,7 +622,8 @@ class ModuleGenerator(object):
>          """
>          Generate a list of all extensions in name/version format
>          """
> -        return self.app.make_extension_string(name_version_sep='/', ext_sep=',').split(',')
> +        exts_str = self.app.make_extension_string(name_version_sep='/', ext_sep=',')
> +        return exts_str.split(',') if exts_str else []
> 
>      def _generate_help_text(self):
>          """
> ```

To be honest I really dislike the list -> string -> list conversion. The design in #3697  was not great.  But it will adapt.

boegel:
> To be honest I really dislike the list -> string -> list conversion. The design in #3697 was not great. But I will adapt.

Can't disagree with you there. Maybe @Flamefire can pitch in here.
We can consider cleaning that up/improving it for EasyBuild 5.0, but for EasyBuild 4.x we should be careful in changing tactics there, since there could be easyblocks that rely on being able to customize those methods...

boegel (approved):

Flamefire:
> > To be honest I really dislike the list -> string -> list conversion. The design in #3697 was not great. But I will adapt.
> 
> Can't disagree with you there. Maybe @Flamefire can pitch in here. We can consider cleaning that up/improving it for EasyBuild 5.0, but for EasyBuild 4.x we should be careful in changing tactics there, since there could be easyblocks that rely on being able to customize those methods...

I can't remember the exact reasons from back then but yes it seems `make_extension_string` isn't really needed.   
However we use the functionality in 2 places: `EBEXTSLIST` and `modulegenerator::_generate_extension_list` and maybe we anticipated that more things might use it.

BTW:  `_generate_extension_list` vs `_generate_extensions_list` in module generator is also...confusing

PR #4334, opened 2023-08-23
trigger `run_shell_cmd` hook in `run` function

boegel:
This includes changes required for the `5.0.x` branch to fix the tests after it was synced with `develop` which includes changes from #4323 .

I had to move the `load_source` function from `tools.filetools` to `tools.hooks`, to break an import loop (was got introduced since `tools.run` now imports from `tools.hooks`)

### DIFF
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -3037,7 +3037,7 @@ class ToyBuildTest(EnhancedTestCase):
                     # 'copy_toy_file' command doesn't exist, but don't worry,
                     # this problem will be fixed in post_run_shell_cmd_hook
                     cmd += " && copy_toy_file toy copy_of_toy"
-                    return cmd
+                return cmd
 
             def post_run_shell_cmd_hook(cmd, *args, **kwargs):
                 exit_code = kwargs['exit_code']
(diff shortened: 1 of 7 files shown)


test/framework/toy_build.py
```diff
@@ -3032,28 +3032,34 @@ def end_hook():
...
 
             def pre_run_shell_cmd_hook(cmd, *args, **kwargs):
-                if cmd.strip() == TOY_COMP_CMD:
+                if isinstance(cmd, str) and cmd.strip() == TOY_COMP_CMD:
                     print("pre_run_shell_cmd_hook triggered for '%s'" % cmd)
                     # 'copy_toy_file' command doesn't exist, but don't worry,
                     # this problem will be fixed in post_run_shell_cmd_hook
                     cmd += " && copy_toy_file toy copy_of_toy"
-                return cmd
+                    return cmd
```
branfosj, line 3040:
```suggestion
                return cmd
```

Indentation has changed comparing to https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/toy_build.py#L2941

branfosj (changes requested):

test/framework/toy_build.py
boegel: I had a reason to do this, but it was probably due to a bug that's now fixed in the `run` function, so it should indeed be no longer needed, I'll check it...

PR #4355, opened 2023-10-17
minor help fix

smoors:
(created using `eb --new-pr`)
finally my OCD reached a point where i needed to fix this :)


casparvl (approved):
Looks like a good clarification to me. Code change is reasonable. Just to be sure: tested the feature branch with a `--robot` from a custom path for a dependency, everything still works like a charm (as it should, since this should only touch the description).

PR #4356, opened 2023-10-21
change `run_shell_cmd` to store command output in temporary file(s) by default + pass `RunShellCmdResult` instance to `RunShellCmdError`

boegel:
WIP because i) tests still fail, ii) it sits on top of #4351 which should get merged first


easybuild/tools/run.py
```diff
@@ -110,21 +114,13 @@ def pad_4_spaces(msg):
...
-            with open(stderr_fp, 'w') as fp:
-                fp.write(self.stderr)
-            error_info.extend([
-                pad_4_spaces(f"output (stdout)           ->  {output_fp}"),
-                pad_4_spaces(f"error/warnings (stderr)   ->  {stderr_fp}"),
-            ])
+            if self.out_file is not None:
+                error_info.append(pad_4_spaces(f"output (stdout)           ->  {self.out_file}"))
+            if self.err_file is not None:
+                error_info.append(pad_4_spaces(f"error/warnings (stderr)   ->  {self.err_file}"))
```
lexming, line 123: it took me a while to understand the logic behind this. Maybe it can be simplified a bit with something like:
```
        if self.out_file is not None:
            out_info_msg = "output (stdout)"
            if self.stderr is None:
                out_info_msg = "output (stdout + stderr)"
            error_info.append(pad_4_spaces(f"{out_info_msg}  ->  {self.out_file}"))

        if self.err_file is not None:
            error_info.append(pad_4_spaces(f"error/warnings (stderr)   ->  {self.err_file}"))
```

lexming (commented):
Looks good, passing the output and error file pointers to `RunShellCmdError` from the outside will avoid the duplicate logs. I have a minor nitpick suggestion :)

easybuild/tools/run.py
boegel: Yeah, I guess that's clearer, fixed in 8620e64b9

lexming (approved):
LGTM

PR #4357, opened 2023-10-22
take into account that `VERBOSE_VERSION` imported from `easybuild.easyblocks` is now a string value

boegel:
required because of changes in `easybuild/easyblocks/__init__.py` in https://github.com/easybuilders/easybuild-easyblocks/pull/3018


boegel:
requires https://github.com/easybuilders/easybuild-framework/pull/4358 to fix broken CI workflows

boegel:
@Micket this should be good to go now...

Micket (approved):
lgtm

PR #4358, opened 2023-10-23
use 5.0.x branch for easyblocks + easyconfigs in CI workflows (for now)

boegel:
We'll need to change this back to `develop` once the `5.0.x` branches are collapsed into `develop` after the last EasyBuild 4.x release has been published...


Micket (approved):
lgtm

PR #4359, opened 2023-10-24
add support for `%(sysroot)s` template value

xinan1911:
This improves sysroot support. For example this code line in [wget](https://github.com/easybuilders/easybuild-easyconfigs/blob/3f0b0151205bb4497a28e3c494db80e144bbbe48/easybuild/easyconfigs/w/wget/wget-1.21.3-GCCcore-11.3.0.eb#L41) can be generalized using sysroot template.
From
`preconfigopts = "export PKG_CONFIG_PATH=/usr/lib64/pkgconfig:/usr/lib/pkgconfig:/usr/lib/x86_64-linux-gnu/pkgconfig && "`
to
`preconfigopts = "export PKG_CONFIG_PATH=%(sysroot)s/usr/lib64/pkgconfig:%(sysroot)s/usr/lib/pkgconfig:%(sysroot)s/usr/lib/x86_64-linux-gnu/pkgconfig && " `

This works for both cases when --sysroot is passed and when it's not (i.e. when it defaults to None).

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -204,7 +204,6 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
     template_values['arch'] = platform.uname()[4]
 
     # set 'sysroot' template based on 'sysroot' configuration option, using empty string as fallback
- value
     template_values['sysroot'] = build_option('sysroot') or ''
 
     # step 1: add TEMPLATE_NAMES_EASYCONFIG
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -203,12 +203,9 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
     # set 'arch' for system architecture based on 'machine' (4th) element of platform.uname() return value
     template_values['arch'] = platform.uname()[4]
 
-    # set 'sysroot' based on sysroot build option
-    sysroot = build_option('sysroot')
-    if sysroot is None:
-        template_values['sysroot'] = ""
-    else:
-        template_values['sysroot'] = sysroot
+    # set 'sysroot' template based on 'sysroot' configuration option, using empty string as fallback
+ value
+    template_values['sysroot'] = build_option('sysroot') or ''
 
     # step 1: add TEMPLATE_NAMES_EASYCONFIG
     for name in TEMPLATE_NAMES_EASYCONFIG:
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -205,7 +205,7 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
 
     # set 'sysroot' based on sysroot build option
     sysroot = build_option('sysroot')
-    if sysroot == None:
+    if sysroot is None:
         template_values['sysroot'] = ""
     else:
         template_values['sysroot'] = sysroot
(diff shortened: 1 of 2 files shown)


boegel:
I was looking into this too, here's the test I came up with to test the use of the `%(sysroot)s` template:

```python
 def test_sysroot_template(self):
       """Test the %(sysroot)s template"""

       test_easyconfigs = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'easyconfigs', 'test_ecs')
       toy_ec = os.path.join(test_easyconfigs, 't', 'toy', 'toy-0.0.eb')

       test_ec_txt = read_file(toy_ec)
       test_ec_txt += '\nconfigopts = "--sysroot=%(sysroot)s/"'
       write_file(test_ec, test_ec_txt)

       ec = EasyConfig(test_ec)
       self.assertEqual(ec['configopts'], "--sysroot=/")

       update_build_option('sysroot', self.test_prefix)

       ec = EasyConfig(test_ec)
       self.assertEqual(ec['configopts'], "--sysroot=%s/" % self.test_prefix)
```

easybuild/framework/easyconfig/templates.py
```diff
@@ -201,6 +203,13 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
     # set 'arch' for system architecture based on 'machine' (4th) element of platform.uname() return value
     template_values['arch'] = platform.uname()[4]
 
+    # set 'sysroot' based on sysroot build option
+    sysroot = build_option('sysroot')
+    if sysroot is None:
+        template_values['sysroot'] = ""
+    else:
+        template_values['sysroot'] = sysroot
```
boegel, line 211:
```suggestion
    # set 'sysroot' template based on 'sysroot' configuration option, using empty string as fallback
 value
    template_values['sysroot'] = build_option('sysroot') or ''

boegel (changes requested):

easybuild/framework/easyconfig/templates.py
casparvl: Elegant and short his one. We will have to rely on None evaluating to False in this context, but I think it's safe enough to assume Python won't change that anytime soon (famous last words :D)

casparvl:
Yep, `update_build_option('sysroot', self.test_prefix)` is the thing we needed in the test, and didn't know about. 

I like that you captured it in one test, we'll update that. I also like how you just modified the `toy.eb` instead of setting the full text (we stole that from the PR you gave us as example, but it's not really needed here).

I _do_ prefer our approach to using `echo`'s, I find the `configopts = "--sysroot=%(sysroot)s/"` potentially confusing. It could be mistaken for trying to test if one can pass `--sysroot` as a configopt, whereas with the echo it's clearly about the template value. 

I also think we should stick to setting it in the three pre*opts we did now (again, stolen from the PR you gave us as example), as it verifies that template value is set in all of these stages of (which are probably the most relevant ones in terms of where it might be used) and not _just_ in the configure step.

I'll merge your approach into ours :)

casparvl:
Ah, wait, I only now see that you are actually not relying on running all steps, you just query the options again. Ok, then I understand why you did the suggestive `--sysroot`. I'll go with that, it saves some time in the CI since you don't have to actually run the steps in your case.

I'll still do it for `configure` `build` and `install` opts, just to validate that it works in all of those.

casparvl:
Hm, that's a completely different test. I guess the `update_build_option` carries over between tests, so we should probably put it back. I thought that build options were reset between tests anyway? Maybe not...

boegel:
> Hm, that's a completely different test. I guess the `update_build_option` carries over between tests, so we should probably put it back. I thought that build options were reset between tests anyway? Maybe not...

Hmm, if that's the case, then we should definitely fix that (in a separate PR), configuration options shouldn't carry over tests at all...

boegel (approved):
lgtm

PR #4360, opened 2023-10-25
add `dependency_names` method to `EasyConfig` class to get set of names of (direct) dependencies

Flamefire:
This is useful for determining if a feature or configure flag should be enabled. Especially as the pattern (getting the name out of `ec.dependencies()`) is used often already.

I considered adding it to `EasyBlock` (at least additionally) which is possible but IMO doesn't provide much benefit given that it saves only 3 chars of typing ('.ec')

I changed some tests to add testing or using this function which should be enough.

One "issue" is the handling of external modules which have a "name" of `None`. I currently filter them out. We could try reading `['external_module_metadata'].get('name', [])`   
However that would be a change in behavior making a straight replacement of current usages risky and I'm also not sure how to correctly handle that: `external_module_metadata['name']` is a list. Should all be added?

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1139,6 +1139,15 @@ class EasyConfig(object):
 
         return retained_deps
 
+    def dependency_names(self, build_only=False):
+        """
+        Return a set of names of all (direct) dependencies after filtering.
+        Iterable builddependencies are flattened when not iterating.
+
+        :param build_only: only return build dependencies, discard others
+        """
+        return {dep['name'] for dep in self.dependencies(build_only=build_only) if dep['name']}
+
     def builddependencies(self):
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm

boegel:
W.r.t. external dependencies: if we want to also get the names of those from `dependency_names`, we should just make sure that `name` is always defined for them, we don't want to add special handling for external deps all throughout the code imho.

Flamefire:
> W.r.t. external dependencies: if we want to also get the names of those from `dependency_names`, we should just make sure that `name` is always defined for them, we don't want to add special handling for external deps all throughout the code imho.

That is not possible: `name` inside a dependency is a string while it is a list of strings inside the external module info. So it only makes sense to insert that list into the result of such a function but not into the `name` property which would suddenly become either a string or a list.

PR #4361, opened 2023-10-25
deduplicate warnings & errors found in logs and add initial newline + tab in output

Flamefire:
It isn't useful to report duplicate errors multiple times. And e.g. configure may report a wrong option at the start and end. Hence deduplicate it.
Also change the logged warning/error to include a newline & tab in front of each element which makes it nicer to read. E.g.:
```
build failed (first 300 chars): Found 1 error(s) in command output:
	configure: WARNING: unrecognized options: --with-foo, --with-bar (took 2 mins 39 secs)
```

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -52,7 +52,7 @@ from easybuild.tools.build_log import EasyBuildError, dry_run_msg, print_msg, ti
 from easybuild.tools.config import ERROR, IGNORE, WARN, build_option
 from easybuild.tools.hooks import RUN_SHELL_CMD, load_hooks, run_hook
 from easybuild.tools.py2vs3 import string_type
-from easybuild.tools.utilities import trace_msg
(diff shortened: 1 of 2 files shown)


Flamefire:
Doesn't look like those test failures can be caused by the change. One is related to network, the other is... strange

boegel:
@Flamefire That test failure was a fluke, I've re-triggered the failing test runs, and they passed just fine on 2nd attempt, so do ignore.

boegel (approved):
Pretty trivial change that actually only impacts the logged warnings or raised errors for easyblocks that use actually `check_log_for_errors` (which currently only includes `OpenBLAS` and `FreeFEM`), so good to go.

In practice the lifetime of this change will probably be rather limited, since in EasyBuild 5.x we're switching to a now mechanism for running shell commands, which will probably make `check_log_for_errors` for errors pretty useless.

PR #4365, opened 2023-10-26
fix `findPythonDeps.py` script when called with an absolute path to an easyconfig

Flamefire:
`eb --missing` returns a list with filenames. So an absolute path passed to `--ec` isn't matched which leads to an error stating you should install the EasyConfig that you are trying to find the dependencies of. Fix by getting the filename/basename first.

Example error:
```
findPythonDeps.py --ec ~/.local/EasyBuildDev/easybuild-easyconfigs/easybuild/easyconfigs/n/nf-core/nf-core-2.10-foss-2022b.eb nf-core==2.10
You need to install all modules on which ~/.local/EasyBuildDev/easybuild-easyconfigs/easybuild/easyconfigs/n/nf-core/nf-core-2.10-foss-2022b.eb depends first!
Missing:
	* nf-core/2.10-foss-2022b (nf-core-2.10-foss-2022b.eb)
```


boegel (approved):
lgtm

PR #4369, opened 2023-10-29
Fix EnvironmentModules to pass unit tests

### LINKED ISSUE #4368: module function check does not support Environment Modules
When trying to use EB with Environment Modules, the module function check raises an error:

```console
$ eb --modules-tool EnvironmentModules --module-syntax Tcl 
== Temporary log file in case of crash /tmp/eb-snqwtxzg/easybuild-gvwalsbq.log
ERROR: pattern '.*modulecmd.tcl' (Environment Modules) not found in defined 'module' function.
Specify the correct modules tool to avoid weird problems due to this mismatch, see the --modules-tool and --avail-modules-tools command line options.
Or alternatively, use --allow-modules-tool-mismatch to stop treating this as an error. Obtained definition of 'module' function: module is a function
module ()  
{ 
    local _mlredir=1;
    if [ -n "${MODULES_REDIRECT_OUTPUT+x}" ]; then
    |   if [ "$MODULES_REDIRECT_OUTPUT" = '0' ]; then
    |   ¦   _mlredir=0;
    |   else
    |   ¦   if [ "$MODULES_REDIRECT_OUTPUT" = '1' ]; then
    |   ¦   ┆   _mlredir=1;
    |   ¦   fi;
    |   fi;
    fi; 
    case " $@ " in 
    |   *' --no-redirect '*)
    |   ¦   _mlredir=0
    |   ;;  
    |   *' --redirect '*)
    |   ¦   _mlredir=1
    |   ;;
    esac;
    if [ $_mlredir -eq 0 ]; then
    |   _module_raw "$@";
    else
    |   _module_raw "$@" 2>&1;
    fi
}

```

Module command cannot be found in `module` shell function definition:
* starting version 5.1.0: module command is always called from `_module_raw` shell function
* starting version 4.0: module command is called from `_module_raw` shell function when it has been initialized in an interactive shell session

xdelaruelle:
On my way to fix #4368 I have found that some unit tests (test_avail, test_load and test_get_setenv_value_from_modulefile) are failing with Environment Modules v4.2+.

Here are some changes to either update:
* a unit test (test_load) that was expecting a different behavior for Modules 4.2+
* EnvironmentModules class to better cope with newer Modules versions


boegel (changes requested):
@xdelaruelle Let's also update `.github/workflows/unit_tests.yml` to test with a more recent version of Environment Modules, to verify that these changes are sufficient to pass the tests (and to help ensure that we don't accidentally break support for recent Environment Modules later)?
Just one or two extra "key" versions is enough, maybe we should also stop testing with the now ancient v4.1.4 (we shouldn't have too many test configurations).

xdelaruelle:
> @xdelaruelle Let's also update `.github/workflows/unit_tests.yml` to test with a more recent version of Environment Modules, to verify that these changes are sufficient to pass the tests (and to help ensure that we don't accidentally break support for recent Environment Modules later)? Just one or two extra "key" versions is enough, maybe we should also stop testing with the now ancient v4.1.4 (we shouldn't have too many test configurations).

@boegel This is what I have prepared in another PR that I will push as soon as #4369 and #4371 will get merged. These two PRs are needed for the CI to pass with Environment Modules v4.2+.

EL8 is now shipping Environment Modules v4.5, so I will propose to change the v4.1.4 key by v4.5

boegel (approved):

PR #4371, opened 2023-10-29
Adapt module function check for Environment Modules v4+

### LINKED ISSUE #4368: module function check does not support Environment Modules

xdelaruelle:
Default check_module_function tests that module command is called from module shell function. With EnvironmentModules v4+, module command is usually called from the _module_raw shell function.

This commit adds a specific version of check_module_function for EnvironmentModules class. Module command is first checked within _module_raw shell function definition. If not found, default test (that checks module function) is run.

Adapt "test_module_mismatch" unit test to specifically check module command definition with EnvironmentModules.

Fixes #4368

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1328,6 +1328,31 @@ class EnvironmentModules(EnvironmentModulesTcl):
     MAX_VERSION = None
     VERSION_REGEXP = r'^Modules\s+Release\s+(?P<version>\d\S*)\s'
 
+    def check_module_function(self, allow_mismatch=False, regex=None):
+        """Check whether selected module tool matches 'module' function definition."""
+        # Modules 5.1.0+: module command is called from _module_raw shell function
+        # Modules 4.2.0..5.0.1: module command is called from _module_raw shell function if it has
+        #   been initialized in an interactive shell session (i.e., a session attached to a tty)
+        if self.testing:
+            if '_module_raw' in os.environ:
+                out, ec = os.environ['_module_raw'], 0
+            else:
+                out, ec = None, 1
+        else:
+            cmd = "type _module_raw"
+            out, ec = run_cmd(cmd, simple=False, log_ok=False, log_all=False, force_in_dry_run=True, trace=False)
+
+        if regex is None:
+            regex = r".*%s" % os.path.basename(self.cmd)
+        mod_cmd_re = re.compile(regex, re.M)
+
+        if ec == 0 and mod_cmd_re.search(out):
+            self.log.debug("Found pattern '%s' in defined '_module_raw' function." % mod_cmd_re.pattern)
+        else:
+            self.log.debug("Pattern '%s' not found in '_module_raw' function, falling back to 'module' function",
+                           mod_cmd_re.pattern)
+            super(EnvironmentModules, self).check_module_function(allow_mismatch, regex)
+
     def check_module_output(self, cmd, stdout, stderr):
         """Check output of 'module' command, see if if is potentially invalid."""
         if "_mlstatus = False" in stdout:
--- a/test/framework/modulestool.py
+++ b/test/framework/modulestool.py
@@ -39,7 +39,7 @@ from easybuild.base import fancylogger
 from easybuild.tools import modules, StrictVersion
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import read_file, which, write_file
-from easybuild.tools.modules import Lmod
+from easybuild.tools.modules import EnvironmentModules, Lmod
 from test.framework.utilities import init_config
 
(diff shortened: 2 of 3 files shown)


test/framework/modulestool.py
```diff
@@ -192,6 +192,24 @@ def test_lmod_specific(self):
...
+            os.environ['_module_raw'] = "() {  eval `/usr/share/Modules/libexec/foo.tcl' bash $*`;\n}"
+            os.environ['module'] = "() {  _module_raw \"$@\" 2>&1;\n}"
+            error_regex = ".*pattern .* not found in defined 'module' function"
+            self.assertErrorRegex(EasyBuildError, error_regex, EnvironmentModules, testing=True)
+
+            # redefine '_module_raw' function with correct module command
+            os.environ['_module_raw'] = "() {  eval `/usr/share/Modules/libexec/modulecmd.tcl' bash $*`;\n}"
+            mt = EnvironmentModules(testing=True)
+            self.assertIsInstance(mt.loaded_modules(), list)  # dummy usage
+            del os.environ['_module_raw']
```
boegel, line 211: There's no need to clean up the `_module_raw`, since the environment is reset between each test.
It's also incomplete, since `module` is not cleaned up.

easybuild/tools/modules.py
```diff
@@ -1328,6 +1328,29 @@ class EnvironmentModules(EnvironmentModulesTcl):
...
+            cmd = "type _module_raw"
+            out, ec = run_cmd(cmd, simple=False, log_ok=False, log_all=False, force_in_dry_run=True, trace=False)
+
+        if regex is None:
+            regex = r".*%s" % os.path.basename(self.cmd)
+        mod_cmd_re = re.compile(regex, re.M)
+
+        if ec == 0 and mod_cmd_re.search(out):
+            self.log.debug("Found pattern '%s' in defined 'module' function." % mod_cmd_re.pattern)
+        else:
```
boegel, line 1351: let's add an additional log statement here, something like:

```suggestion
        else:
            self.log.debug("Pattern '%s' not found in '_module_raw' function, falling back to 'module' function", mod_cmd_re.pattern)
```

easybuild/tools/modules.py
```diff
@@ -1328,6 +1328,29 @@ class EnvironmentModules(EnvironmentModulesTcl):
...
+            cmd = "type _module_raw"
+            out, ec = run_cmd(cmd, simple=False, log_ok=False, log_all=False, force_in_dry_run=True, trace=False)
+
+        if regex is None:
+            regex = r".*%s" % os.path.basename(self.cmd)
+        mod_cmd_re = re.compile(regex, re.M)
+
+        if ec == 0 and mod_cmd_re.search(out):
+            self.log.debug("Found pattern '%s' in defined 'module' function." % mod_cmd_re.pattern)
```
boegel, line 1350: log message here should mention `_module_raw` function, not `module` ?

boegel (changes requested):

xdelaruelle:
@boegel Thanks for your review. I have added your change suggestions.

boegel (approved):
lgtm

PR #4375, opened 2023-11-05
only install GitHub token when testing with Lmod 8.x + Python 3.6 or 3.9

### LINKED ISSUE #4372: GitHub tests fail in CI because we're hitting GitHub API rate limit
Since the changes proposed in #4192 by @Flamefire have been merged, we're more frequently seeing failing GitHub-related tests in the `develop` branch (when a GitHub token is available that can be used to run the GitHub tests).

We'll need to revisit the configuration, to only install the available GitHub token is less configurations, for example only when testing with `Lua` as module syntax, and maybe even only for selected Python versions (only for Python 3.6, default in RHEL8-based OSs, and Python 3.9, default for RHEL9-based OSs).

boegel:
fixes #4372


migueldiascosta (approved):
lgtm

PR #4380, opened 2023-11-13
print error message in `main_with_hooks` function when `EasyBuildError` was raised

boegel:
also deprecates `exitCode` in `print_error`, replaced with `exit_code` (very small impact, since `exitCode` is not actually used anywhere in framework or easyblocks)


smoors (approved):
lgtm

PR #4389, opened 2023-11-25
detect Fortran .mod files in GCCcore installations

### LINKED ISSUE #4203: Detect if there are .mod files for GCCcore installs
If compiling Fortran then we should not do that with GCCcore due to the incompatibility in the `.mod` files generated. It would be helpful if we searched the installed software and warn if any are found.

Possibly:
* config option to turn on the check
* config option to control if this is a warning or error
* we will need ability to say that an easyconfig produces `.mod` files but that we are sure that they are not from Fortran or to allowing us to skip the check for an easyconfig

branfosj:
fixes #4203

Targeting the `5.0.x` branch so I can use `**` in `glob` and that is not available in Python 2.7.

TODO:
* add a `bypass_mod_files_check` easyconfig parameter (see #4203 for an example of where we'll need this bypass)
* tests

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3303,6 +3303,19 @@ class EasyBlock(object):
 
         return fail_msg
 
+    def sanity_check_mod_files(self):
+        """
+        Check installation for Fortran .mod files
+        """
+        self.log.debug("Check for .mod files in install directory")
+        mod_files = glob.glob(os.path.join(self.installdir, '**', '*.mod'), recursive=True)
+
+        fail_msg = None
+        if mod_files:
+            fail_msg = ".mod files (%s) found in the installation." % ', '.join(mod_files)
+
+        return fail_msg
+
     def _sanity_check_step_common(self, custom_paths, custom_commands):
         """
         Determine sanity check paths and commands to use.
@@ -3624,6 +3637,14 @@ class EasyBlock(object):
         if linked_shared_lib_fails:
             self.log.warning("Check for required/banned linked shared libraries failed!")
             self.sanity_check_fail_msgs.append(linked_shared_lib_fails)
+
+        if self.toolchain.name in ['GCCcore']:
+            mod_files_found = self.sanity_check_mod_files()
+            if mod_files_found:
+                if build_option('fail_on_mod_files'):
+                    self.sanity_check_fail_msgs.append(mod_files_found)
+                else:
+                    print_warning(mod_files_found)
 
         # cleanup
         if self.fake_mod_data:
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -400,6 +400,7 @@ class EasyBuildOptions(GeneralOption):
                              None, 'store_true', False),
             'extra-modules': ("List of extra modules to load after setting up the build environment",
                               'strlist', 'extend', None),
+            'fail-on-mod-files': ("Fail if .mod files are detected in a GCCcore install", None, 'store_true', False),
             'fetch': ("Allow downloading sources ignoring OS and modules tool dependencies, "
                       "implies --stop=fetch, --ignore-osdeps and ignore modules tool", None, 'store_true', False),
             'filter-deps': ("List of dependencies that you do *not* want to install with EasyBuild, "
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -277,6 +277,7 @@ BUILD_OPTIONS_CMDLINE = {
         'enforce_checksums',
         'experimental',
         'extended_dry_run',
+        'fail_on_mod_files',
         'force',
         'generate_devel_module',
         'group_writable_installdir',


easybuild/framework/easyblock.py
```diff
@@ -3303,6 +3303,19 @@ def regex_for_lib(lib):
 
         return fail_msg
 
+    def sanity_check_mod_files(self):
+        """
+        Check installation for Fortran .mod files
+        """
+        self.log.debug("Check for .mod files in install directory")
```
boegel, line 3310:
```suggestion
        self.log.debug(f"Checking for .mod files in install directory {self.installdir}...")
```

easybuild/framework/easyblock.py
```diff
@@ -3625,6 +3638,14 @@ def xs2str(xs):
             self.log.warning("Check for required/banned linked shared libraries failed!")
             self.sanity_check_fail_msgs.append(linked_shared_lib_fails)
 
+        if self.toolchain.name in ['GCCcore'] and not self.cfg['skip_mod_files_check']:
+            mod_files_found = self.sanity_check_mod_files()
+            if mod_files_found:
+                if build_option('fail_on_mod_files'):
+                    self.sanity_check_fail_msgs.append(mod_files_found)
+                else:
+                    print_warning(mod_files_found)
```
boegel, line 3647:
```suggestion
            mod_files_fail_mgs = self.sanity_check_mod_files()
            if mod_files_fail_mgs:
                if build_option('fail_on_mod_files'):
                    self.sanity_check_fail_msgs.append(mod_files_fail_mgs)
                else:
                    print_warning(mod_files_fail_mgs)

File: easybuild/framework/easyconfig/default.py [hunk 3]
```diff
@@ -130,6 +130,7 @@
     'sanity_check_paths': [{}, ("List of files and directories to check "
                                 "(format: {'files':<list>, 'dirs':<list>})"), BUILD],
     'skip': [False, "Skip existing software", BUILD],
+    'skip_mod_files_check': [False, "Skip the check for .mod files in a GCCcore level install", BUILD],
```
boegel, line 133: maybe rename to `skip_mod_files_sanity_check`?

easybuild/tools/options.py
```diff
@@ -400,6 +400,7 @@ def override_options(self):
                              None, 'store_true', False),
             'extra-modules': ("List of extra modules to load after setting up the build environment",
                               'strlist', 'extend', None),
+            'fail-on-mod-files': ("Fail if .mod files are detected in a GCCcore install", None, 'store_true', False),
```
boegel, line 403: rename to `fail-on-mod-files-gcccore`?

boegel (changes requested):
@branfosj Would be good to have a dedicated test for this?

easybuild/tools/options.py
branfosj: done in 0b3daf2

easybuild/framework/easyconfig/default.py
branfosj: done in 0b3daf2

easybuild/framework/easyblock.py
branfosj: done in 0b3daf2

easybuild/framework/easyblock.py
branfosj: done in 0b3daf2

branfosj:
> @branfosj Would be good to have a dedicated test for this?

done in 809b694

easybuild/framework/easyblock.py
```diff
@@ -3303,6 +3303,19 @@ def regex_for_lib(lib):
...
+    def sanity_check_mod_files(self):
+        """
+        Check installation for Fortran .mod files
+        """
+        self.log.debug(f"Checking for .mod files in install directory {self.installdir}...")
+        mod_files = glob.glob(os.path.join(self.installdir, '**', '*.mod'), recursive=True)
+
+        fail_msg = None
+        if mod_files:
+            fail_msg = ".mod files (%s) found in the installation." % ', '.join(mod_files)
```
boegel, line 3315: @branfosj Can we try to filter out false positives here?

I was thinking about checking the output of the `file` command, if that's available.
Not sure if that makes sense...

branfosj: I'm not sure what we would want to exclude by using `file`.

```
$ file 2.3.0-GCCcore-11.2.0/lib64/kim-api/mod/kim_unit_system_module.mod
2.3.0-GCCcore-11.2.0/lib64/kim-api/mod/kim_unit_system_module.mod:                     gzip compressed data, from Unix, original size 57651

$ file Clang/15.0.5-GCCcore-11.3.0/include/flang/omp_lib.mod
Clang/15.0.5-GCCcore-11.3.0/include/flang/omp_lib.mod:                       UTF-8 Unicode (with BOM) text

$ file json-fortran/8.3.0-GCCcore-12.2.0/include/json_value_module.mod
json-fortran/8.3.0-GCCcore-12.2.0/include/json_value_module.mod:     gzip compressed data, from Unix, original size 201261

$ file CPLEX/20.10-GCCcore-10.2.0/opl/examples/opl/cplex_staticLex/staticlex3.mod
CPLEX/20.10-GCCcore-10.2.0/opl/examples/opl/cplex_staticLex/staticlex3.mod: ASCII text
```

boegel: As discussed, we can refine this later to try and weed out false positives without having to whitelist...

boegel (approved):

PR #4395, opened 2023-12-05
use source toolchain version when passing only `--try-toolchain`

Flamefire:
`process_software_build_specs` always adds a `toolchain` key even when only the name of the toolchain is available. This makes the code in `tweak` ignore the code path in using `toolchain_name` and/or `toolchain_version` with the fallback to using the values of the source toolchain.
This in turn leads to failures further down when a value of `None` is used where an actual version is expected.

Fix by only adding the `toolchain` key when we have both the name and version and hence a complete, valid toolchain spec.

CC @boegel who added the code using a default of `None` for the version in https://github.com/easybuilders/easybuild-framework/pull/810 Maybe you remember the reasoning?


ocaisa (approved):
LGTM

boegel:
> CC @boegel who added the code using a default of `None` for the version in #810 Maybe you remember the reasoning?

Not sure what the reasoning was here, doesn't make much sense indeed, your fix does make sense

PR #4398, opened 2023-12-07
Replace `parallel` EC parameter by `maxparallel`

Flamefire:
Continuation of #3842 for 5.0 to fully remove the `parallel` EC parameter

### DIFF
--- a/easybuild/base/testing.py
+++ b/easybuild/base/testing.py
@@ -215,6 +215,14 @@ class TestCase(OrigTestCase):
             if mock_stderr:
                 self.mock_stderr(False)
 
+    @contextmanager
+    def mocked_stderr(self):
+        self.mock_stderr(True)
+        try:
+            yield sys.stderr
+        finally:
+            self.mock_stderr(False)
+
     def tearDown(self):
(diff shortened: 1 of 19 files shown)


boegel:
@Flamefire Not supporting the `parallel` easyconfig parameter at all anymore is a bridge too far imho, even for EasyBuild 5.0, especially since it's a commonly used easyconfig parameter (as shown in https://github.com/easybuilders/easybuild-easyconfigs/pull/19375).

So I think we should proceed with #4580 instead, and only actually remove support for using the `parallel` easyconfig parameter in the next major EasyBuild release (6.0).

So this PR should probably be closed, I'll look into reviewing #4580 ASAP...

Flamefire:
Agreed, especially as this is basically #4580 with an extra commit which we can apply at any later point. The EC PR is still useful to promote the use of the new name.

boegel:
Hard replacing of `parallel` is a bridge too far, we'll go with the deprecation approach being implemented in #4580, so closing this...

Hard removal of support for `parallel` easyconfig parameter should only be done in EasyBuild v6.0, which is years away...

PR #4399, opened 2023-12-10
also run test_cases_step with --sanity-check-only

smoors:
the goal is to be able to run the `test_cases_step` without rebuilding/reinstalling.
it seems logical to also run it with `--sanity-check-only`?

the alternative would be to implement a `--test-cases-only` option, maybe the test cases are too big for a sanity check?

in any case, we will need to run `self.set_parallel()` before the `test_cases_step` to make sure the test cases are run with the correct number of cores. i'm not sure where would be the best place. maybe we should also run the `check_readiness_step`?

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3860,6 +3860,10 @@ class EasyBlock(object):
         """
         Run provided test cases.
         """
+        # ensure parallel is set in case check_readiness_step is skipped (e.g. with --sanity-check-only)
+        if not self.cfg['parallel']:
+            self.set_parallel()
+
         for test in self.cfg['tests']:
             change_dir(self.orig_workdir)
             if os.path.isabs(test):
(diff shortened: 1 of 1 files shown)


boegel:
@smoors Since this changes the behavior of `--sanity-check-only`, we shouldn't do this for EasyBuild 4.x

Regardless of that, I'm in favor of introducing a dedicated `--test-cases-only` for this, to avoid surprises?

smoors:
> Regardless of that, I'm in favor of introducing a dedicated `--test-cases-only` for this, to avoid surprises?

that's probably safer indeed. in that case there's no reason to wait for EB 5, right?

boegel:
> > Regardless of that, I'm in favor of introducing a dedicated `--test-cases-only` for this, to avoid surprises?
> 
> that's probably safer indeed. in that case there's no reason to wait for EB 5, right?

Indeed, since there's no breaking changes.

smoors:
you can also run `eb --skip --rebuild` to the same effect, so this is no longer an issue for me, will close for now.

PR #4400, opened 2023-12-12
rename unclear `*run*` methods to `*install_extension*` + rename `install_extensions` to `install_all_extensions`

### LINKED ISSUE #4113: rename unclear ExtensionEasyblock.run() method
The `run()` method from extensioneasyblock is the one carrying out the installation of any extension. Its name should be changed to something more meaningful that reflects its function more clearly. For instance `run_extension_steps()`.

lexming:
fixes #4113

* `run()` -> `install_extension()`
*  `prerun()` -> `pre_install_extension()`
* `postrun()` -> `post_install_extension()`
* `run_async()` -> `install_extension_async()`
* `install_extensions()` -> `install_all_extensions()`


lexming:
@boegel I made your suggested changes

easybuild/framework/extension.py
```diff
@@ -172,24 +172,64 @@ def version(self):
         return self.ext.get('version', None)
 
     def prerun(self):
+        """
+        [DEPRECATED] Stuff to do before installing a extension.
+        """
+        self.log.deprecated(
+            "Extension.prerun() is deprecated, use Extension.pre_install_extension() instead.",
+            '5.0',
```
boegel, line 180: @lexming This should be `6.0`, otherwise hitting `prerun` will raise an error, rather than only print a deprecation warning

boegel (changes requested):

boegel:
@lexming We need a companion PR in easyblocks too?

boegel:
Companion easyblocks PR:
* https://github.com/easybuilders/easybuild-easyblocks/pull/3064

easybuild/framework/extension.py
boegel: fixed in 21f1d58

easybuild/framework/easyblock.py
```diff
@@ -1914,15 +1922,15 @@ def install_extensions_sequential(self, install=True):
                 ext.toolchain.prepare(onlymod=self.cfg['onlytcmod'], silent=True, loadmod=False,
                                       rpath_filter_dirs=self.rpath_filter_dirs)
 
-            # real work
+            # actual installation of the extension
             if install:
                 try:
-                    ext.prerun()
+                    ext.pre_install_extension()
```
boegel, line 1928: This seems to make sense, but it's not actually this simple: if an easyblock (say `PythonPackage`) hasn't been adjusted to use `pre_install_extension` rather than `prerun`, then it breaks things.

For example, when installing `matplotlib`:

<details>

```
$ eb matplotlib-3.7.2-gfbf-2023a.eb -f
Installing 'kiwisolver' extension (3/5) ━━━━━━━━━━━━━━━━╺━━━━━━━━━━━━━━━━━━━━━━━ 0:00:02

    "__main__", mod_spec)
    exec(code, run_globals)
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/main.py", line 768, in <module>
    main_with_hooks()
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/main.py", line 754, in main_with_hooks
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/main.py", line 715, in main
    hooks, do_build)
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/main.py", line 552, in process_eb_args
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/main.py", line 173, in build_and_install_software
    raise ec_res['err']
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/main.py", line 134, in build_and_install_software
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/framework/easyblock.py", line 4258, in build_and_install_one
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/framework/easyblock.py", line 4132, in run_all_steps
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/framework/easyblock.py", line 3967, in run_step
    step_method(self)()
  File "/arcanine/scratch/gent/vo/000/gvo00002/vsc40023/easybuild/eb5/easybuild-easyblocks/easybuild/easyblocks/generic/pythonbundle.py", line 125, in extensions_step
    super(PythonBundle, self).extensions_step(*args, **kwargs)
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/framework/easyblock.py", line 2894, in extensions_step
    self.install_all_extensions(install=install)
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/framework/easyblock.py", line 1880, in install_all_extensions
    self.install_extensions_sequential(install=install)
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/framework/easyblock.py", line 1930, in install_extensions_sequential
    txt = ext.install_extension()
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/framework/extensioneasyblock.py", line 155, in install_extension
    EasyBlock.patch_step(self, beginpath=self.ext_dir)
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-framework/easybuild/framework/easyblock.py", line 2598, in patch_step
    beginpath = self.src[srcind]['finalpath']
TypeError: string indices must be integers
```

</details>

By not calling `prerun` anymore, the work that's implemented by `PythonPackage.prerun` is not being done anymore...

So we need to be a bit smarter here, and ensure that `prerun` is still run if the easyblock does not implement `pre_install_extension` yet...

lexming: Understood, the issue is that the main installation function in the base `EasyBlock` class has to explicitly call any deprecated methods (`prerun`, `run`, ...) found in the child EasyBlock used by the extension (if any). And that obviously won't happen if it only calls the newer methods (`pre_install_extension`, `extension_install`, ...), which probably do not exist on older extension EasyBlocks.

That was indeed a bit trickier to fix than expected. Main problem is that the base `Extension` class has to define the deprecated methods for as long as they are allowed to be used because the deprecated methods in the child EasyBlock might call the methods from the parent `Extension` class with `super`. However, it is not trivial to detect if a child class overloads one of the deprecated methods from the `Extension` class.

So I added a new method to `Extension` as a wrapper to all calls to `pre_install_extension`, `extension_install`, etc that will detect the presence of any deprecated method in the extension EasyBlock and use that one if present.

lexming:
@boegel all tests positive, this is ready

easybuild/framework/extension.py
```diff
@@ -172,29 +172,93 @@ def version(self):
...
+            'post_install_extension': 'postrun',
+        }
+
+        if substep not in deprecated:
+            raise EasyBuildError("Unknown extension installation substep: %s", substep)
+
+        try:
+            ext_substep = getattr(self, deprecated[substep])
+            parent_obj = super(self.__class__, self)
+            parent_substep = getattr(parent_obj, deprecated[substep])
```
boegel, line 244: @lexming What about the case where the easyblock being used doesn't define (for example) `run` itself, but it derives from another easyblock that does still define it.

Example: an easyblock that derives from a generic easyblock like `PythonPackage`.
If `PythonPackage` still implements `run`, we should still call it, right?

We should only skip calling `run` if the method we obtain is the one from `EasyBlock`?

lexming: That was indeed a problem. For instance, the `numpy` easyblock does not define any `prerun`, but `PythonPackage` does. So the catch on that was failing because the `prerun` method on numpy is equal to its parent class.

I figured out a simpler more robust method by just using the qualified name of the method. Then it becomes trivial to know what class defines the method of the _active_ EasyBlock. Please try again with latest commits.

lexming:
@boegel added unit tests for the new method `Extension.install_extension_substep`. I had to add various new easyblocks to test sandbox to be able to test the case where some easyblock inherits custom or deprecated `run` methods from a parent class. So there are collateral changes due to those additions.

If you think that adding those extra easyblocks is overkill, we can also just test the case where an easyblock directly defines custom or deprecated `run` methods. For such cases, the `DummyExtension` easyblock is sufficient.

Regarding the use of `__qualname__`, the qualified name is indeed also printed as part of the string representation of  the method. However, that requires parsing something like `<bound method XXXX.xxx of XXX object at 0x000000000000>`, which IMO is more prone to error than handling `__qualname__`. So I left that part unchanged.

boegel (approved):
lgtm

PR #4403, opened 2023-12-14
add `--module-cache-suffix` configuration setting to allow multiple (Lmod) caches

Flamefire:
This is useful when having multiple module trees (i.e. different `--installprefix`) to not overwrite the cache file of one with the other when using `--update-modules-tool-cache`. Inspired by how LMod itself creates the cache, e.g. `~/.cache/lmod/spiderT.rapids_x86_64_Linux.lua`

Actually the name is created by the default ["groupName" hook ](https://lmod.readthedocs.io/en/latest/170_hooks.html)

The logic is actually: `spiderT.${LMOD_SYSTEM_NAME}_.$(uname -m)_$(uname -s).lua`, see https://github.com/TACC/Lmod/blob/55d353b3a3a4649cbaaef18b2f00cdc017d7cc92/src/StandardPackage.lua#L88

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -239,6 +239,7 @@ BUILD_OPTIONS_CMDLINE = {
         'job_polling_interval',
         'job_target_resource',
         'locks_dir',
+        'module_cache_suffix',
         'modules_footer',
         'modules_header',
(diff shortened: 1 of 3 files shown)


boegel (approved):
lgtm

PR #4405, opened 2023-12-14
fix `--inject-checksums` when extension specifies patch file in tuple format

Flamefire:
`make_list_lines` had a `"'%s'" % x` part where `x` is the patch entry.

However that could be a tuple (or even dict) to designate extra options such as the subdir or level where the patch should be applied. This then failed with "TypeError: not all arguments converted during string formatting"

Fix by checking the type first and add a test for this case.


boegel (approved):
lgtm

PR #4407, opened 2023-12-19
add terse support to `--missing-modules`

Flamefire:
It is useful to have the missing modules in a format that can be processed in a script, i.e. one EC name per line.
So allow `--terse` in combination with `--missing-modules`.

@boegel I needed to adjust the early-exit to not exit for this case. Can you think of any situation where this might be an issue?

Example use case I wanted to have: `eb $(eb PyTorch-2.1.0-foss-2022a.eb -M --terse | grep -v PyTorch) -r` to install all missing dependencies.

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -684,6 +684,8 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None, pr
         options.list_prs,
         options.merge_pr,
         options.review_pr,
+        # --missing-modules is processed by process_eb_args,
+        # so we can't exit just yet here if it's used in combination with --terse
         options.terse and not options.missing_modules,
         search_query,
     ]
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -495,7 +495,7 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
     # dry_run: print all easyconfigs and dependencies, and whether they are already built
     elif dry_run_mode:
         if options.missing_modules:
-            txt = missing_deps(easyconfigs, modtool)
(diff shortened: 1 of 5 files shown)


Flamefire:
Any news here? I just used this as `eb $(eb --from-pr 19722,19723 -M --terse | grep -v TAMkin | grep -v molmod) -r` which I find very useful

boegel:
@Flamefire I took a quick look at this, and while I'm certainly all for improving the output of `--missing-modules` when combined with `--terse`, there are more (seemingly unrelated) changes in here that make reluctant to merge this.

easybuild/main.py
```diff
@@ -684,7 +684,7 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None, pr
         options.list_prs,
         options.merge_pr,
         options.review_pr,
-        options.terse,
+        options.terse and not options.missing_modules,
```
boegel, line 687: Why is early stop disabled for `--terse --missing-modules`?

Shouldn't `silent` be controlled by `options.terse` rather than being set to `True` in `clean_exit` below, as is done at the end of `main`?

boegel (changes requested):

easybuild/main.py
boegel: Oh, I see, it's because `--missing-modules` is only acted upon in `process_eb_args`, which is called after this.

OK, fine, but then we should add a comment here to explain why this is done like this, because it looks out of place:

```suggestion
        # --missing-modules is processed by process_eb_args,
        # so we can't exit just yet here if it's used in combination with --terse
        options.terse and not options.missing_modules,

boegel (approved):

PR #4409, opened 2023-12-21
clean up & speed up environment checks

Flamefire:
`os.environ` is a dict hence the check for existance and access can usually be combined into a single access of the key avoiding the repeated lookup.
Especially for `os.getenv` which already returns `None` when the key doesn't exist.
Also use `.pop(key, None)` to remove a key without checking for it's existance first to avoid C&P errors

I initially noticed this (performance/code quality) issue for the `get_software_*`-functions:
```python
    root = None
    if env_var in os.environ:
        root = os.getenv(env_var)
```


Flamefire:
Merged develop again to rerun test.

boegel (approved):
lgtm

PR #4411, opened 2023-12-21
remove superflous string formatting

Flamefire:
`"%s" % var` is the same as `str(var)` while the latter is faster due to less parsing, inserting etc.
In some places `var` was already a string so it can just be used.


easybuild/framework/easyconfig/format/format.py
```diff
@@ -370,7 +370,7 @@ def parse(self, configobj):
         for key, value in self.supported.items():
             if key not in self.VERSION_OPERATOR_VALUE_TYPES:
                 raise EasyBuildError('Unsupported key %s in %s section', key, self.SECTION_MARKER_SUPPORTED)
-            self.sections['%s' % key] = value
+            self.sections[key] = value
```
boegel, line 373: This is not fully equivalent, are we sure we don't need `str(key)` here?

easybuild/tools/docs.py
```diff
@@ -1199,7 +1199,7 @@ def avail_toolchain_opts_md(name, tc_dict):
     tc_items = sorted(tc_dict.items())
     table_values = [
         ['``%s``' % val[0] for val in tc_items],
-        ['%s' % val[1][1] for val in tc_items],
+        [val[1][1] for val in tc_items],
```
boegel, line 1202: no need for `str(val[1][1])`?

test/framework/sandbox/easybuild/tools/module_naming_scheme/test_module_naming_scheme.py
```diff
@@ -64,6 +64,6 @@ def det_module_symlink_paths(self, ec):
 
     def is_short_modname_for(self, modname, name):
         """
-        Determine whether the specified (short) module name is a module for software with the specified name.
+        Determine whether the specified (short) module name is a moduleq for software with the specified name.
```
boegel, line 67: typo introduced?

```suggestion
        Determine whether the specified (short) module name is a module for software with the specified name.

boegel (changes requested):

easybuild/framework/easyconfig/format/format.py
Flamefire: That question is actually a good reason **for** this change ;-)   
The change is correct: This line is never executed if the key is not also a key in `VERSION_OPERATOR_VALUE_TYPES` which are all strings and cannot reasonably be anything else.   
Alternatively: A non-string key doesn't make sense at this point.

easybuild/tools/docs.py
Flamefire: `tc_dict` is dict mapping toolchain options to a tuple of `(default_value, description)`. And the description can only be string (and is for all examples I could find)

easybuild/framework/easyconfig/format/format.py
boegel: OK, thanks for clarifying!

easybuild/tools/docs.py
boegel: :+1:

boegel (approved):
lgtm

PR #4412, opened 2023-12-21
clean up uses of `getattr` and `hasattr`

Flamefire:
`hasattr` is basically calling `getattr` and checking for `AttributeError`.
It also supports a default value in case the attribute is not found. Make use of that default value and catch the exception directly instead of checking `hasattr` where that makes sense.


boegel (approved):
lgtm

PR #4415, opened 2023-12-27
Run unit tests on an updated versions of Modules: v4.5.3 + v5.3.1

xdelaruelle:
Environment Modules was updated on EL8 from version 4.1.4 to version 4.5.2+patches which can be assimilated to version 4.5.3.

Seems interesting to update the modules4 version used in unit_tests CI workflow to match Environment Modules version found in EL8.

### DIFF
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -3209,11 +3209,22 @@ class ToyBuildTest(EnhancedTestCase):
                 'end',
             ])
         else:
-            expected = '\n'.join([
-                'if { ![ is-loaded GCC/4.6.3 ] && ![ is-loaded GCC/7.3.0-2.30 ] } {',
-                '    module load GCC/4.6.3',
-                '}',
-            ])
+            if not self.modtool.supports_safe_auto_load:
+                expected = '\n'.join([
+                    'if { ![ is-loaded GCC/4.6.3 ] && ![ is-loaded GCC/7.3.0-2.30 ] } {',
+                    '    module load GCC/4.6.3',
+                    '}',
+                ])
+            else:
+                expected = '\n'.join([
+                    '',
+                    "if { [ module-info mode remove ] || [ is-loaded GCC/7.3.0-2.30 ] } {",
+                    "    module load GCC",
+                    '} else {',
+                    "    module load GCC/4.6.3",
+                    '}',
+                    '',
+                ])
 
         self.assertIn(expected, toy_mod_txt)
 
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -2330,6 +2330,7 @@ class ToyBuildTest(EnhancedTestCase):
 
     def test_toy_toy(self):
         """Test building two easyconfigs in a single go, with one depending on the other."""
+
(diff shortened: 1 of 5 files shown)


xdelaruelle:
I have disabled "test_toy_toy" when Environment Modules tool is used as the test produces an inconsitent environment.

"test_toy_toy" test generates two versions of "toy" module. Both modules express a reflexive conflict (against "toy") and toy/0.0-two loads toy/0.0-one.
    
As these two modules share the same name (`toy`) and due to the reflexive conflict they express (`conflict toy`), environment obtained when loading `toy/0.0-two` is inconsistent as it tries to load `toy/0.0-one`.

Environment Modules version 4.2 (and above) detects this inconsistency and raise an error unless `--force` option is set.

I am not fully satisfied with the solution of disabling this "test_toy_toy" test for EnvironmentModules but it seemed to be the more pragmatic way to solve this. Another solution would be to rename the second software "toy_toy", but it requires some changes on the test and duplication of files in the repository. It may not be worth it.

xdelaruelle:
@boegel, @branfosj I have changed base branch for this PR to 5.0.X and rebased this PR. Please let me know if something should be changed for this PR to get merged.

test/framework/toy_build.py
```diff
@@ -2317,6 +2317,13 @@ def test_reproducibility_ext_easyblocks(self):
 
     def test_toy_toy(self):
         """Test building two easyconfigs in a single go, with one depending on the other."""
+
+        # skip when using Environment Modules tool, since this test generates an inconsistent environment
+        # (two modules with same name, expressing a reflexive conflict, one loading the other)
+        if isinstance(self.modtool, EnvironmentModules):
+            print("Skipping test_toy_toy because Environment Modules is being used")
+            return
```
boegel, line 2325: Rather than skipping this test, we should fix it: there's no explicit intention here to load two modules with the same name...

In `toy2.eb`, we could change the `name` field to `toy2`?

boegel (changes requested):

boegel:
@xdelaruelle Can you also rebase this, to make sure the recently merged PRs pass with v4.5.3?

Should we (also?) test with v5.x?

xdelaruelle:
> @xdelaruelle Can you also rebase this, to make sure the recently merged PRs pass with v4.5.3?

Of course, I will do that in the coming days, taking time to correctly handle `test_toy_toy`.

> Should we (also?) test with v5.x?

We should, especially to test to cache build part. I will add a commit on this PR to also test against Modules 5.3 (version shipped on EL9)

test/framework/toy_build.py
xdelaruelle: I have changed the name field and other properties for `toy2.eb` to correctly build and install:

```python
"name = 'toy2'",
"easyblock = 'EB_toy'",
"sources = ['toy/toy-0.0.tar.gz']",
"patches = []",
"sanity_check_paths = { 'files': ['bin/toy2'], 'dirs': ['bin']}",
"prebuildopts = 'mv toy.source toy2.c &&'",
```

boegel (approved):
lgtm

PR #4417, opened 2023-12-27
use `--all` option with EnvironmentModules v4.6+ to get available hidden modules

xdelaruelle:
Environment Modules option "--all" was introduced in version 4.6.0 to obtain hidden modules among "module avail" results.

Update EnvironmentModules class to make use of this option to return hidden modules on "available" function like done on Lmod class.

Update "test_avail" unit test to match new results obtain with Environment Modules 4.6+


boegel (changes requested):

boegel (approved):
lgtm

PR #4429, opened 2024-01-04
Add `--allow-unrecognized-configure-options` option

Flamefire:
Intended to disable the error raised by unknown configure arguments. To be used in the ConfigureMake easyblock and potentially related ones.

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -269,6 +269,7 @@ BUILD_OPTIONS_CMDLINE = {
     False: [
         'add_system_to_minimal_toolchains',
         'allow_modules_tool_mismatch',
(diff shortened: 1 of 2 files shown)


migueldiascosta:
wondering if things like this should require an additional e.g. `--unsafe` (or `--not-recommended`, or `--use-at-own-risk`...) option, similarly to how some options require `--experimental` ...

in any case, will you submit a companion PR changing the `ConfigureMake` easyblock? Merging this without the other risks forgetting about it and leaving an option that does nothing

Flamefire:
@boegel suggested this to be an EC param instead. Closing.

Flamefire:
Implemented in https://github.com/easybuilders/easybuild-easyblocks/pull/3025

PR #4435, opened 2024-01-09
create `lib` -> `lib64` symlink (or vice versa) *before* running `postinstallcmds`

ocaisa:
Not having the symlinks in place means that `postinstallcmds` that rely on using/modifying something in the library directory have to use tricks to figure out the correct location.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3019,12 +3019,6 @@ class EasyBlock(object):
         - run post install commands if any were specified
         """
 
-        self.run_post_install_commands()
-        self.apply_post_install_patches()
-        self.print_post_install_messages()
-
-        self.fix_shebang()
-
         lib_dir = os.path.join(self.installdir, 'lib')
         lib64_dir = os.path.join(self.installdir, 'lib64')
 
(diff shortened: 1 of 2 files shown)


ocaisa:
See https://github.com/easybuilders/easybuild-easyconfigs/pull/19386 and libSBML in https://github.com/easybuilders/easybuild-easyconfigs/pull/19488

smoors:
this will require fixing `Brunsli` easyconfigs due to `lib*` wildcard:
```
postinstallcmds = [ 
    "mkdir %(installdir)s/bin",
    "cp dbrunsli %(installdir)s/bin/",
    "cp libbrunsli*.a %(installdir)s/lib*/",
    "cp libbrunsli*.%s %%(installdir)s/lib*/" % SHLIB_EXT,
]
```

boegel:
@ocaisa This is a "breaking" change, so I'm re-targeting this to EasyBuild 5.0

@smoors You're right... How did you find that particular easyconfig? Any others that may be hit by this?

smoors:
> @smoors You're right... How did you find that particular easyconfig? Any others that may be hit by this?

quick and dirty regex search on `lib*` after `postinstallcmds`. did not see any other issues, but may have missed some.

PR #4442, opened 2024-01-18
treat paths that start with colon as absolute paths

smoors:


easybuild/tools/module_generator.py
```diff
@@ -981,10 +981,13 @@ def update_paths(self, key, paths, prepend=True, allow_abs=False, expand_relpath
 
         abspaths = []
         for path in paths:
-            if os.path.isabs(path) and not allow_abs:
-                raise EasyBuildError("Absolute path %s passed to update_paths which only expects relative paths.",
-                                     path)
-            elif not os.path.isabs(path):
+            if os.path.isabs(path) or path.startswith(':'):
```
boegel, line 984: Hmm, not sure about this.

That would mean that a path like `:bin` is treated like an absolute path...

We could treat `:` as a special case, but then with `== ':'`?

```python
if path == ':':
    ...
elif os.path.isabs(path) and now allow_abs:
    ...
```

smoors: yes makes more sense, as you can already provide multiple paths as a list

smoors:
`modules-4` does not accept `:` (path separator).
it seems that `""` empty string should also work, needs to be tested

PR #4444, opened 2024-01-19
implement initial support for running shell commands asynchronously using `run_shell_cmd` (WIP)

boegel:
TODO:
* [x] add support in `run_shell_cmd` for using `asynchronous=True`
* [x] use `run_shell_cmd` with `asynchronous=True` in `EasyBlock.skip_extensions_parallel`
* [ ] use `run_shell_cmd` with `asynchronous=True` in `Extension.async_cmd_start`
* [ ] test with `R` easyconfig and `--parallel-extensions-install` (and `--skip`)


boegel:
Quick test:

* `eb R-4.2.2-foss-2022b.eb --skip --force` (serial) took ~28m30s
* `eb R-4.2.2-foss-2022b.eb --skip --parallel-ext --experimental -f` took ~16m55s

And that's only the skipping part being done in parallel...

boegel:
I've tested this with `R-4.2.1-foss-2022a.eb`, and saw a significant speedup when using `--parallel-extensions-install --experimental` using 24 cores (RHEL8, AMD Rome):

* with sequential installation of extensions (default):
  * 6h25min total time
  * 5h50min for extensions step
* with `--parallel-extensions-install --experimental`:
  * 1h54min total time => 3.37x speedup
  * 1h21min for extensions step => 4.32x speedup for this specific step

The scaling is not spectacular, but I think that's mainly determine by the order in which extensions are being considered, not so much the implementation of running the extension installations in parallel in the background.

So, I think this is ready for review/merge.

Do note that these changes require a corresponding change in easyblocks, see:
* https://github.com/easybuilders/easybuild-easyblocks/pull/3163

I prefer keeping the support for `--parallel-extensions-install` experimental for now, because I would like to revisit `install_extensions_parallel` later and improve things a bit.

There's a small additional change needed to make parallel installation of extensions work for easyconfigs like `R-bundle-Bioconductor` or `R-bundle-CRAN`, but I'll make that change in a separate PR, I don't want to bury that in this one...

easybuild/framework/easyblock.py
```diff
@@ -1818,41 +1818,43 @@ def skip_extensions_parallel(self, exts_filter):
...
+                    installed_exts_ids.append(idx)
+
+                checked_exts_cnt += 1
+                exts_pbar_label = "skipping installed extensions "
+                exts_pbar_label += "(%d/%d checked)" % (checked_exts_cnt, exts_cnt)
+                self.update_exts_progress_bar(exts_pbar_label)
 
             # start additional checks asynchronously
-            while exts_queue and len(running_checks_ids) < self.cfg['parallel']:
+            while exts_queue:
```
lexming, line 1850: Is it necessary to trigger this loop for each iteration on `while exts_queue or async_shell_cmd_tasks`? This loop is just submitting the tasks for all extensions and `exts_queue` is not modified during the outer loop.

So the whole outer loop  `while exts_queue or async_shell_cmd_tasks:` could be broken down into 2 independent loops:
```
for idx, ext  in exts_queue:
    cmd, stdin = resolve_exts_filter_template(exts_filter, ext)
    task = thread_pool.submit(run_shell_cmd, cmd, ...)
    async_shell_cmd_tasks.append(task)

while async_shell_cmd_tasks:
    (done_tasks, _) = wait(async_shell_cmd_tasks, timeout=1, return_when=FIRST_COMPLETED)
    for task in done_tasks:
        async_shell_cmd_tasks.remove(task)
        res = task.result()
        [...]
        self.update_exts_progress_bar(exts_pbar_label)
```

easybuild/framework/easyblock.py
```diff
@@ -1864,6 +1866,8 @@ def skip_extensions_parallel(self, exts_filter):
 
         self.ext_instances = retained_ext_instances
 
+        thread_pool.shutdown()
```
lexming, line 1869: Maybe we can avoid an explicit shutdown by launching the executor through a `with` and use `as_completed` instead of `wait` . We can move the call to `ThreadPoolExecutor` lower to keep things compact as well:

Following the spirit of my previous comment with the 2 independent loops:
```
cmd_queue = [resolve_exts_filter_template(exts_filter, ext) for ext in exts_queue
with ThreadPoolExecutor(max_workers=self.cfg['parallel']) as thread_pool:
    async_shell_cmd_tasks = [thread_pool.submit(run_shell_cmd, cmd, ...) for cmd, stdin in cmd_queue]
    for done_task in concurrent.futures.as_completed(async_shell_cmd_tasks):
        res = done_task.result()
        idx = res.task_id
        [...]
        self.update_exts_progress_bar(exts_pbar_label)

# compose new list of extensions, skip over the ones that are already installed;
[...]

boegel: fixed in ca3828b

easybuild/framework/easyblock.py
boegel: fixed in ca3828b

lexming (approved):
LGTM

Issue #4122, opened 2022-11-04
also support using full path soname for shared libraries as alternative to using RPATH


boegel:
see https://stoppels.ch/2022/08/04/stop-searching-for-shared-libraries.html

ocaisa:
I just thought of the _perfect_ use case for this in EESSI. When using MPItrampoline, this would allow you to explicitly get back to the host system libraries that are required to make a drop in MPI replacement work without having to open the door to all of the system libraries.

PR #4448, opened 2024-02-03
enable rpath by default

branfosj:
I've marked this as draft while we discuss changing this default in EB5.

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -489,7 +489,7 @@ class EasyBuildOptions(GeneralOption):
             'required-linked-shared-libs': ("Comma-separated list of shared libraries (names, file names, or paths) "
                                             "which must be linked in all installed binaries/libraries",
                                             'strlist', 'extend', None),
-            'rpath': ("Enable use of RPATH for linking with libraries", None, 'store_true', False),
+            'rpath': ("Enable use of RPATH for linking with libraries", None, 'store_true', True),
             'rpath-filter': ("List of regex patterns to use for filtering out RPATH paths", 'strlist', 'store', None),
             'rpath-override-dirs': ("Path(s) to be prepended when linking with RPATH (string, colon-separated)",
                                     None, 'store', None),


Micket:
1. we are keeping LD_LIBRARY_PATH
2. this just affects stuff we built (no attempts to elf-patch binaries)
3. should be fine to introduce in existing software trees (it of course won't affect things retroactively, but new things will build with rpath enabled).
4. No attempts to change how users link against modules when building their own code.

Positives:
https://docs.easybuild.io/rpath-support/#rpath_support_why
though primarily i think its main advantage is to built libraries and binaries more robust against users fiddling with LD_LIBRARY_PATH.

Downsides:
1. Higher burden when making new easyconfigs due to the code that ensures things did get rpath'ed, contributions test builds will non-rpath'ed linking has occurred. e.g. https://github.com/easybuilders/easybuild-easyconfigs/issues/17516 . One could argue we would rather catch these sooner, but it would also affect every novice user of EB building local easyconfigs. 
2. Maybe we actually rather would want a different approach like: https://github.com/easybuilders/easybuild-framework/issues/4122 in which case it might hurt to set this new default if there is a better option around the corner.

jfgrimm:
I'd support this.

For reference, we enabled rpath on our new cluster (without filtering LD_LIBRARY_PATH), and have zero negative feedback from users.
EasyBuild already skips the rpath check when using the Binary easyblock, so I haven't run into many issues at all (2 failed on the rpath check out of ~1000 unique software, ~4000 total modules). 

I don't think we should worry about holding back to wait for #4122, that can always be added later (and would probably get more attention with rpath being on by default)

ocaisa:
I'm also in general supportive, but I do think there needs to be space in some cases to be able to override libraries with an ABI compatible alternative (MPI in particular). Current rpath support does make this possible (and we use that in EESSI). If we do switch to #4122 at some point, we should still be able to retain this possibility.

boegel:
I'm fine with this, as long as we add a very clear section to the documentation that explains our motivation for doing this, outline the (mostly positive) impact this has. That section should include clear instructions on how to switch back to the old behavior of not using RPATH linking, or how to disable it for specific easyconfigs (by setting the `rpath` toolchain option to `False`).

We should also only have it enabled by default on Linux, and keep it disabled on macOS (where it doesn't work at all).

branfosj:
* 612b1c0 enables / disables this for not-Mac / Mac
* https://github.com/easybuilders/easybuild-docs/pull/241 adds documentation

boegel (approved):
lgtm

PR #4453, opened 2024-02-14
implement support for running interactive commands with `run_shell_cmd`

boegel:
Currently sits on top of #4444, which should definitely get merged first - I'll sync with PR with `5.0.x` branch once that's done.

~WIP because:~ (ready for review)
* [x] ~more tests for `run_cmd_qa` should also be implemented for `run_shell_cmd` using `qa_patterns`~ (done)
* [x] ~use of `qa_wait_patterns` not supported yet;~ (now implemented)
* [x] ~probably need to add a `qa_timeout` option (equivalent to `maxhits` in `run_cmd_qa`);~ (implemented)
* [x] ~maybe some refactoring should be done, like moving code to actually answer questions to a dedicated (private) `_handle_cmd_qa` function, to avoid making `run_shell_cmd` even bigger than it already is~ (done, see `_answer_question` helper function);
* [x] ~testing with actual easyblocks not done yet~ (see https://github.com/easybuilders/easybuild-easyblocks/pull/3270)

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -360,7 +360,7 @@ def run_shell_cmd(cmd, fail_on_error=True, split_stderr=False, stdin=None, env=N
 
         exit_code = None
         stdout, stderr = b'', b''
-        check_interval_secs = 0.001
+        check_interval_secs = 0.1
         time_no_match = 0
 
         # collect output piece-wise, while checking for questions to answer (if qa_patterns is provided)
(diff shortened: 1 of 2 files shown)


easybuild/tools/run.py
```diff
@@ -226,6 +282,8 @@ def run_shell_cmd(cmd, fail_on_error=True, split_stderr=False, stdin=None, env=N
     :param qa_patterns: list of 2-tuples with patterns for questions + corresponding answers
     :param qa_wait_patterns: list of 2-tuples with patterns for non-questions
                              and number of iterations to allow these patterns to match with end out command output
```
branfosj, line 284: Is this correct? Specifically the 'and number of iterations to allow these patterns to match with end out command output' part.

boegel: No it doesn't, but that does hint towards something in `run_cmd_qa` that some easyblocks may be using but is not supported yet in `run_shell_cmd`.
I can look into implementing that when we hit it, we should get this PR merged ASAP.

Fixed docstring in 204f3287b

branfosj (approved):

easybuild/tools/run.py
boegel: Looks like that feature isn't used in the current easyblocks that use the `no_qa` option of `run_cmd_qa`, so we can reimplement it if/when the need arises...

PR #4454, opened 2024-02-14
allow only alphanumeric characters in the output filename

branfosj:
Through a combination of `preconfigopts` and `configopts`, I had a command of the form
```bash
FLAG='/path/to/include' ./configure
```

When this went through the existing code I ended up with a filename starting with `include'-`.

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -186,10 +186,10 @@ def fileprefix_from_cmd(cmd, allowed_chars=False):
     Simplify the cmd to only the allowed_chars we want in a filename
 
     :param cmd: the cmd (string)
-    :param allowed_chars: characters allowed in filename (defaults to string.ascii_letters + string.digits)
+    :param allowed_chars: characters allowed in filename (defaults to string.ascii_letters + string.digits + "_-")
     """
     if not allowed_chars:
-        allowed_chars = string.ascii_letters + string.digits
+        allowed_chars = f"{string.ascii_letters}{string.digits}_-"
 
     return ''.join([c for c in cmd if c in allowed_chars])
 
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -181,6 +181,19 @@ def run_cmd_cache(func):
 run_shell_cmd_cache = run_cmd_cache
 
 
+def fileprefix_from_cmd(cmd, allowed_chars=False):
+    """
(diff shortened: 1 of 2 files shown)


easybuild/tools/run.py
```diff
@@ -244,6 +244,9 @@ def to_cmd_str(cmd):
     if output_file:
         toptmpdir = os.path.join(tempfile.gettempdir(), 'run-shell-cmd-output')
         os.makedirs(toptmpdir, exist_ok=True)
+        # restrict the allowed characters in the name of the output_file
+        allowed_chars = string.ascii_letters + string.digits
```
boegel, line 248: @branfosj We could also allow `-` and `_`? Not sure that would make much of a difference in practice though...

We should also have a test to verify that this indeed does what we expect, let me know if you need help with this.

I'm fine with enhancing the existing `test_run_shell_cmd_basic` in `test/framework/run.py` for this.

branfosj: see faf8f14

It was eaier for me to just pull that code out into a separate function and test that directly.

test/framework/run.py
```diff
@@ -193,6 +194,24 @@ def test_run_shell_cmd_basic(self):
             self.assertTrue(isinstance(res.output, str))
             self.assertTrue(res.work_dir and isinstance(res.work_dir, str))
 
+    def test_fileprefix_from_cmd(self):
+        """test simplifications from fileprefix_from_cmd."""
+        cmds = {
+            'abd123': 'abd123',
+            'ab"a': 'aba',
+            'a{:$:S@"a': 'aSa',
```
boegel, line 202: Let's add some more test cases here, like:

* `git clone` should map to `gitclone` (or maybe we want `git-clone`?)
* `cmd-with-dash` should map to `cmd-with-dash` (requires changes in `fileprefix_from_cmd`)
* `cmd_with_underscore` should map to `cmd_with_underscore` (requires changes in `fileprefix_from_cmd`)

boegel (changes requested):

test/framework/run.py
branfosj: If the command is `git clone` then only the `git` part is passed into `fileprefix_from_cmd` and I object to adding extra code to this that increases the complexity.

branfosj: cb84bd2 (with the typo fixed in a309aae)

boegel (approved):
lgtm

PR #4456, opened 2024-02-16
fix description of `backup-modules` configuration option

Flamefire:
The documentation is wrong as it doesn't apply to `--module-only` but rather to `--rebuild` or `--force`.
Also the auto-enabling is not mentioned.


boegel (approved):
lgtm

PR #4463, opened 2024-02-22
add cuda compute capability integer format templates

jfgrimm:
some software wants the CCCs as integers, but without the sm_ prefix (so e.g. 86 instead of 8.6 or sm_86)

this adds comma-, space-, and semicolon-delimited templates of this form


boegel (approved):
lgtm

PR #4473, opened 2024-03-02
replace `log.warn` with `log.warning`

branfosj:
`log.warn` has been deprecated since Python 3.3 and will be removed in Python 3.13.


boegel (approved):
lgtm

PR #4474, opened 2024-03-03
add support for data installations

smoors:
motivation
- leverage EB to install data in a standardized way with proper versioning and checksumming
- support adding datasets as dependency for software
- easily `swap` dataset versions with `ml swap`

changes
- add cmd line option `--installpath-data` similar to `--installpath-software`
- add cmd line option `--subdir-data` (default = `data`) similar to `--subdir-software`
- add cmd line option `--sourcepath-data` similar to `--sourcepath`
- add Easyconfig parameter `data_sources` similar to `sources`

design
- the main reason for a separate `subdir_data` is reusability: in contrast to software it does not have to be rebuilt/reinstalled when for example upgrading the OS or building for a new architecture
- the reason for a separate `sourcepath_data` is that datasets can be very large, so you may want to store them in a different file system or location.


Micket (changes requested):
Tests needs adjusting
`test_prefix_option`:
https://github.com/easybuilders/easybuild-framework/blob/92d29680fb5d14086943c96a9efcd2558093f6ae/test/framework/options.py#L5032

`test_show_config`:
https://github.com/easybuilders/easybuild-framework/blob/92d29680fb5d14086943c96a9efcd2558093f6ae/test/framework/options.py#L4886

easybuild/framework/easyconfig/default.py
```diff
@@ -92,6 +92,7 @@
     'checksums': [[], "Checksums for sources and patches", BUILD],
     'configopts': ['', 'Extra options passed to configure (default already has --prefix)', BUILD],
     'cuda_compute_capabilities': [[], "List of CUDA compute capabilities to build with (if supported)", BUILD],
+    'data_sources': [[], "List of source files for data", BUILD],
```
branfosj, line 95: Is there a need to separate `data_sources` from `sources`?

smoors: data sources can be very big so i think it's good to at least have an option to separate them.

we can set `data_sources` equal to `sources` by default, would you prefer that?

smoors: done in https://github.com/easybuilders/easybuild-framework/pull/4474/commits/2e57ad18d356ab0c3d23d5bd8c9027f62699c54b https://github.com/easybuilders/easybuild-framework/pull/4474/commits/8dbf7d3c61b491dd474ea55d69ad33ce40c6d9ef

smoors: hm, i see now that i misunderstood you.
the separate parameter `data_sources` was suggested by @boegel to make it clear they are different from software sources.
there is no real need for it, it's just cosmetic. i'm not sure if it's a good idea, happy to revert if you prefer.

smoors: as discussed during the EUM, we should keep `data_sources` to allow installing software and datasets in a single easyconfig (e.g. using components) if we want to add support for this later on. this PR does support this, as it requires substantive changes and i'm unsure how useful it is. i've modified this PR to make it easier to implement support for it if/when desired.

boegel:
@smoors Let's re-target this to `5.0.x` / EasyBuild 5.0, not because it involves breaking changes, but because it can serve as a "flagship" feature of EasyBuild 5.0 (seems a bit too much to introduce in a 4.9.x bug fix release)?

boegel:
@smoors I changed to target branch in this PR from `5.0.x` to `develop`, you should synchronize your PR branch with current `develop` branch (which has received a massive update after the release of EasyBuild v5.0.0, see #4825)

easybuild/tools/config.py
```diff
@@ -514,6 +521,7 @@ class ConfigurationVariables(BaseConfigurationVariables):
         'installpath',
         'installpath_modules',
         'installpath_software',
+        'installpath_data',
```
branfosj, line 524:
```suggestion
        'installpath_data',
        'installpath_modules',
        'installpath_software',
```

easybuild/tools/config.py
```diff
@@ -526,8 +534,10 @@ class ConfigurationVariables(BaseConfigurationVariables):
         'repository',
         'repositorypath',
         'sourcepath',
+        'sourcepath_data',
         'subdir_modules',
         'subdir_software',
+        'subdir_data',
```
branfosj, line 540:
```suggestion
        'subdir_data',
        'subdir_modules',
        'subdir_software',
```

easybuild/tools/options.py
```diff
@@ -659,11 +661,15 @@ def config_options(self):
...
-            'sourcepath': ("Path(s) to where sources should be downloaded (string, colon-separated)",
+            'sourcepath': ("Path(s) to where software sources should be downloaded (string, colon-separated)",
                            None, 'store', mk_full_default_path('sourcepath')),
+            'sourcepath-data': ("Path(s) to where data sources should be downloaded (string, colon-separated) "
+                                "(same as sourcepath if not specified)", None, 'store', None),
             'subdir-modules': ("Installpath subdir for modules", None, 'store', DEFAULT_PATH_SUBDIRS['subdir_modules']),
             'subdir-software': ("Installpath subdir for software",
                                 None, 'store', DEFAULT_PATH_SUBDIRS['subdir_software']),
+            'subdir-data': ("Installpath subdir for data",
+                            None, 'store', DEFAULT_PATH_SUBDIRS['subdir_data']),
```
branfosj, line 672:
```suggestion
            'subdir-data': ("Installpath subdir for data", None, 'store', DEFAULT_PATH_SUBDIRS['subdir_data']),
            'subdir-modules': ("Installpath subdir for modules", None, 'store', DEFAULT_PATH_SUBDIRS['subdir_modules']),
            'subdir-software': ("Installpath subdir for software",
                                None, 'store', DEFAULT_PATH_SUBDIRS['subdir_software']),
```

easybuild/tools/options.py
```diff
@@ -615,6 +615,8 @@ def config_options(self):
                                     None, 'store', None),
             'installpath-software': ("Install path for software (if None, combine --installpath and --subdir-software)",
                                      None, 'store', None),
+            'installpath-data': ("Install path for data (if None, combine --installpath and --subdir-data)",
+                                 None, 'store', None),
```
branfosj, line 619: Please move this to before `installpath-modules` to maintain the alphabetical order.

branfosj (changes requested):
I'm happy with this. The changes are all to maintain alphabetical order.

easybuild/tools/options.py
branfosj: fea2ae1

easybuild/tools/options.py
branfosj: fea2ae1

easybuild/tools/config.py
branfosj: fea2ae1

easybuild/tools/config.py
branfosj: fea2ae1

branfosj (approved):

PR #4485, opened 2024-03-19
add mechanism for easyconfig template deprecation

jfgrimm:


jfgrimm:
hmm, looks like something isn't quite getting reset, resulting in deprecation warnings for subsequent tests...

boegel:
> hmm, looks like something isn't quite getting reset, resulting in deprecation warnings for subsequent tests...

should be fixed with 604a398

jfgrimm:
> @jfgrimm I guess this one can be closed, superseded by different approach being implemented in #4511 + #4514?

yes, indeed

PR #4486, opened 2024-03-19
create `env.sh` and `cmd.sh` helper scripts in `run_shell_cmd`

### LINKED ISSUE #970: add support for --record
As was pointed out during the 6th EasyBuild hackathon in Vienna (e.g. by @rtmclay), we should look into support for something like `--record`, which would keep track of the build & install procedure as performed by EasyBuild and dump a (farily) simple bash script describing the performed procedure (including executed commands (+ locations), changes to the environment, runtime, patching, ...).

Micket:
fixes #970

As discussed, I think this would be a greatly useful feature for anyone struggling to 

Rather than directly dropping users into the shell when things fail, just writing these temp files allow some advantages, primarily:
1. No need to break the flow in EB, so no need to limit this to tty sessions only or introducing any options, we can just always do this since it's cheap.
2. Allows users to revisit any of the commands (want to experiment with the configure step before going back to attempt another "make -j"? trivial with this approach)

Maybe we want to be able to limit when we generate this? Maybe, we do run a lot of shell cmds so i guess many of these are not worth ever really revisiting.

This is completely untested yet, just my somewhat formed ideas.

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -214,6 +214,8 @@ def save_cmd(cmd_str, work_dir, env, tmpdir):
         # excludes bash functions (environment variables ending with %)
         fid.write('\n'.join(f'export {key}={shlex.quote(value)}' for key, value in full_env.items()
                             if not key.endswith('%')))
+        fid.write('\n\nPS1="eb-shell> "')
+        fid.write(f'\nhistory -s {shlex.quote(cmd_str)}')
 
     # Make script that sets up bash shell with given environments set.
     cmd_fp = os.path.join(tmpdir, 'cmd.sh')
@@ -225,9 +227,7 @@ def save_cmd(cmd_str, work_dir, env, tmpdir):
             'EB_SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )',
             f'echo Shell for the command: {shlex.quote(cmd_str)}',
             'echo Use command history, exit to stop',
-            'bash --rcfile <(cat $EB_SCRIPT_DIR/env.sh; '
-            'echo \'PS1="eb-shell> "\'; '
-            f'echo history -s {shlex.quote(cmd_str)})',
+            'bash --rcfile $EB_SCRIPT_DIR/env.sh',
             ]))
     os.chmod(cmd_fp, 0o775)
 
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -204,24 +204,31 @@ def fileprefix_from_cmd(cmd, allowed_chars=False):
     return ''.join([c for c in cmd if c in allowed_chars])
 
 
-def save_cmd(cmd_str, work_dir, env, filename):
+def save_cmd(cmd_str, work_dir, env, tmpdir):
+    # Save environment variables in it's own environment file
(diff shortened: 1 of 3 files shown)


Micket:
I think this is good. I haven't actually tested using it yet (need to set up EB5 testing environment)

Micket:
I had to make some adjustments due to how exported variables, history and rcfiles work with bash. But it was all for the better. 

Now it generates a big env.sh file with the entire environment (exported)
```bash
export CC=gcc
export CXX=g++
export LMOD_CMD=/usr/share/lmod/lmod/libexec/lmod
...
```
and a cmd.sh file that sources this environment into an interactive shell + change directory + helpful text  + command history + custom prompt

```bash
#!/usr/bin/env bash

cd "/dev/shm/Bison/3.8.2/system-system/bison-3.8.2"
EB_SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
echo Shell for the command: './configure --prefix=/apps/Test2/software/Bison/3.8.2  --build=x86_64-pc-linux-gnu  --host=x86_64-pc-linux-gnu'
echo Use command history, exit to stop
bash --rcfile <(cat $EB_SCRIPT_DIR/env.sh; echo 'PS1="eb-shell> "'; echo history -s './configure --prefix=/apps/Test2/software/Bison/3.8.2  --build=x86_64-pc-linux-gnu  --host=x86_64-pc-linux-gnu')
```

I've actually tested so that it works and i already love it

Micket:
All that remains here is how to display the existence of these additional files to the user.

Currently, one would see:
```
  >> running shell command:
        ./configure --prefix=/apps/Test2/software/Bison/3.8.2  --build=x86_64-pc-linux-gnu  --host=x86_64-pc-linux-gnu
        [started at: 2024-04-07 19:48:51]
        [working dir: /dev/shm/Bison/3.8.2/system-system/bison-3.8.2]
        [output saved to /dev/shm/eb-08gn7ssh/run-shell-cmd-output/configure-rdh8fxpj/out.txt]
```
and I could add more lines like
```
        [environment saved to /dev/shm/eb-08gn7ssh/run-shell-cmd-output/configure-rdh8fxpj/env.sh]
        [command environment saved to /dev/shm/eb-08gn7ssh/run-shell-cmd-output/configure-rdh8fxpj/cmd.sh]
```
or we could consider just pointing to the directory:
```
        [environment and output in: /dev/shm/eb-08gn7ssh/run-shell-cmd/configure-rdh8fxpj/]
```

Micket:
I moved my PS1 and history to the cmd.env file instead, allowing it to be used as an rcfile directly. 

This simplifies the code but more importantly also avoid trouble with echo removing quotes around strings, so, this is now robust against quotes and spaces in the cmdstr.

boegel (changes requested):
@Micket Please take a look at the suggestions in https://github.com/Micket/easybuild-framework/pull/3

boegel (changes requested):
@Micket small fix in case custom `env` is used in `run_shell_cmd` (which should be very rare, but may happen):

* https://github.com/Micket/easybuild-framework/pull/4

boegel (approved):
I have some ideas to improve this further, but those can be implemented in a follow-up PR

PR #4487, opened 2024-03-24
Add custom delimiter support for update_paths, hack in special case for TCLLIBPATH

Micket:
TCLLIBPATH (and nothing else as far as I know) requires spaces instead of : as the path delimiter.

I opted to just hack in the special case for TCLLIBPATH, but i suppose it could be brought up even further, somehow making this an option for the easyconfigs? I don't really see how one would do this though, where we would add such a field?

```python
modextrapaths_spaces = {
    'TCLLIBPATH': 'lib/tcl8.6',
}
```

meh...

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -247,7 +247,8 @@ class ModuleGenerator(object):
         paths = self._filter_paths(key, paths)
         if paths is None:
             return ''
-        return self.update_paths(key, paths, prepend=False, allow_abs=allow_abs, expand_relpaths=expand_relpaths)
+        delim = ' ' if key == 'TCLLIBPATH' else ':'
+        return self.update_paths(key, paths, prepend=False, allow_abs=allow_abs, expand_relpaths=expand_relpaths, delim=delim)
 
     def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
         """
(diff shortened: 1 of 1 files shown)


easybuild/tools/module_generator.py
```diff
@@ -247,7 +247,9 @@ def append_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
         paths = self._filter_paths(key, paths)
         if paths is None:
             return ''
-        return self.update_paths(key, paths, prepend=False, allow_abs=allow_abs, expand_relpaths=expand_relpaths)
+        delim = ' ' if key == 'TCLLIBPATH' else ':'
```
lexming, line 250: I prefer to avoid conditionals on specific envars inside framework. It would be better to pass the delimiter as an extra option through `modextrapaths` and the code here grabs any special delimiters or defaults to `:`.
```
modextrapaths  = {
    'TCLLIBPATH': 'lib/tcl8.6',
    'TCLLIBPATH': ('lib/tcl8.6', ' '),
    'TCLLIBPATH': ['lib/tcl8.6'],
    'TCLLIBPATH': (['lib/tcl8.6'], ' '),
}
```

Micket: I hesitated with this approach since I felt that it would definitely be to ugly to expect `module_generator.append_paths(key, paths)`   to be passing in `paths` as a tuple or something that.

Instead, if not hacky approach like this, I would add the `delim` kwarg to `append_paths` and `prepend_paths` as well, and whatever login extracts the delimiter would need to happen in the easyblock instead, here:
https://github.com/easybuilders/easybuild-framework/blob/65591dd32a78850144ce39238aa577df6fab3f9f/easybuild/framework/easyblock.py#L1416-L1422


Note that currently, we allow (given the code above) for paths to be tuples as well as lists:
```python
modextrapaths = {
    'FOOPATH': ('multiple', 'paths'),  # allowed today
}
```
with a tuple as well. So, that would certainly interfere with the options here. I don't think any easyconfigs use this, so, we could say that tuples are only for use with specifying a delimiter.

Micket:
For reference, since this touches the same methods as EBPYTHONPREFIX to PYTHONPATH change, I will hold off doing any actual work here until that is merged for 5.0.x

Micket:
I switched to 5.0.x for the PR instead and make a new draft without the TCL specific hacks that I'm sure we don't want

https://github.com/easybuilders/easybuild-framework/pull/4687

PR #4492, opened 2024-03-31
switch from `ls` to `bash` in tests that are expecting this to be a binary

branfosj:
I'm running in a Rocky 8.9 container (that is based off the one from https://github.com/easybuilders/easybuild-containers) and I hit two test failures that are caused by `ls` not being as expected, which causes the two tests I am changing to fail.

In the container where I see the failure:
```
$ file /bin/ls
/bin/ls: a /usr/bin/coreutils --coreutils-prog-shebang=ls script, ASCII text executable

$ file /bin/bash
/bin/bash: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=e4abb3b0b8951a4f746530ae2e9b85559d5fb98f, stripped
```

On Ubuntu:
```
$ file /bin/ls
/bin/ls: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=9a7a491cbe4f1b92cb270406d6642d545b7d1259, for GNU/Linux 3.2.0, stripped

$ file /bin/bash
/bin/bash: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=070009c7b6202e25d64caf24bb49c6cede3a794c, for GNU/Linux 3.2.0, stripped
```

### DIFF
--- a/test/framework/filetools.py
+++ b/test/framework/filetools.py
@@ -924,7 +924,7 @@ class FileToolsTest(EnhancedTestCase):
 
         self.assertTrue(ft.is_binary(b'\00'))
         self.assertTrue(ft.is_binary(b"File is binary when it includes \00 somewhere"))
-        self.assertTrue(ft.is_binary(ft.read_file('/bin/ls', mode='rb')))
+        self.assertTrue(ft.is_binary(ft.read_file('/bin/bash', mode='rb')))
 
     def test_det_patched_files(self):
         """Test det_patched_files function."""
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm, also works fine on `macOS`

```
$ file /bin/bash
/bin/bash: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e]
/bin/bash (for architecture x86_64):	Mach-O 64-bit executable x86_64
/bin/bash (for architecture arm64e):	Mach-O 64-bit executable arm64e

PR #4496, opened 2024-04-03
Prefer `$EBPYTHONPREFIXES` over `$PYTHONPATH`

Micket:
1. I'm moving the abstraction point of update_paths to an internal only _update_paths, allowing prepend_paths and append_paths to simply call a safe update_paths directly without duplicating "__filter_paths" check. This also allows for me to add a place to add special rules for PYTHONPATH/EBPYTHONPREFIX is a safe manner without having do duplicate.

2. I will add a conditional check that rewrites recognized PYTHONPATH's to EBPYTHONPREFIX paths (work in progress)


easybuild/tools/module_generator.py
```diff
@@ -260,7 +268,18 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
+            # Special condition for PYTHONPATHs that match the standard pattern,
+            # replace with EBPYTHONPREFIX which is added to python sys path at runtime
+            ret = ''
+            for path in paths:
+                if re.match(r'lib/python\d+\.\d+/site-packages', path):
+                    self.log.info("Replaced PYTHONPATH %s with EBPYTHONPREFIX", path)
+                    ret += self._update_paths('EBPYTHONPREFIX', [''], prepend, allow_abs, expand_relpaths)
+                else:
+                    ret += self._update_paths(key, [path], prepend, allow_abs, expand_relpaths)
+            return ret
```
Flamefire, line 281: How about: 
```suggestion
            # Special condition for PYTHONPATHs that match the standard pattern,
            # replace with EBPYTHONPREFIX which is added to python sys path at runtime
            python_paths = {path for path in paths if re.match(r'lib/python\d+\.\d+/site-packages', path)}
            if python_paths:
                self.log.info("Replaced PYTHONPATH %s with EBPYTHONPREFIX", python_paths)
                paths = [path for path in paths if path not in python_paths]
                ret = self._update_paths('EBPYTHONPREFIX', [''], prepend, allow_abs, expand_relpaths)
                if paths
                    ret += self._update_paths(key, paths, prepend, allow_abs, expand_relpaths)
                return ret
```

This avoids adding `$EBPYTHONPREFIX` potentially multiple times and does only a single call to `_update_paths` for the remainder if any

Micket: Sure, though that scenario would only actually occur if there was multiple difference python versions matching (e.g. python2.7 + python3.11 etc.), which would have been very broken using PYTHONPATH already, so `python_paths` should never have more than 1 element.

Flamefire: > so python_paths should never have more than 1 element.

That is true. Might be worth to check for and show a meaningful error message

easybuild/tools/module_generator.py
```diff
@@ -981,7 +1000,7 @@ def update_paths(self, key, paths, prepend=True, allow_abs=False, expand_relpath
         abspaths = []
         for path in paths:
             if os.path.isabs(path) and not allow_abs:
-                raise EasyBuildError("Absolute path %s passed to update_paths which only expects relative paths.",
+                raise EasyBuildError("Absolute path %s passed to _update_paths which only expects relative paths.",
```
Flamefire, line 1003: Might be worth taking some freedom here to not mention the private method but "pretend" it is the public method.    
That would avoid having to change the tests.

```suggestion
                raise EasyBuildError("Absolute path %s passed to update_paths which only expects relative paths.",
```

Micket: well, i was gonna try to be precise, but I don't care either way. I can fix the tests if someone wants me to.

Micket:
I think I added the option correctly, still need to try and even use this at all, haven't gotten that far yet :)

And write tests.

easybuild/tools/module_generator.py
```diff
@@ -260,7 +268,18 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
             return ''
-        return self.update_paths(key, paths, prepend=True, allow_abs=allow_abs, expand_relpaths=expand_relpaths)
+        if key == 'PYTHONPATH' and build_option('prefer_ebpythonprefix_over_pythonpath'):
+            # Special condition for PYTHONPATHs that match the standard pattern,
+            # replace with EBPYTHONPREFIX which is added to python sys path at runtime
+            python_paths = {path for path in paths if re.match(r'lib/python\d+\.\d+/site-packages', path)}
+            if python_paths:
+                self.log.info("Replaced PYTHONPATH %s with EBPYTHONPREFIX", python_paths)
+                paths = [path for path in paths if path not in python_paths]
+                ret = self._update_paths('EBPYTHONPREFIX', [''], prepend, allow_abs, expand_relpaths)
```
Flamefire, line 278: As per your comment that there should be only a single such path:

```suggestion
            python_paths = [path for path in paths if re.match(r'lib/python\d+\.\d+/site-packages', path)]
            if python_paths:
                if len(python_paths) > 1:
                    raise EasyBuildError('Found multiple paths for PYTHONPATH: ' + ', '.join(pythonpaths))
                python_path = python_paths[0]
                self.log.info("Replaced PYTHONPATH %s with EBPYTHONPREFIX", python_path)
                ret = self._update_paths('EBPYTHONPREFIX', [''], prepend, allow_abs, expand_relpaths)
                paths = [path for path in paths if path != python_path]
```

I can't imagine a reason where multiple such paths would be valid, so that should be safer and makes e.g. the log message clearer ("Replaced PYTHONPATH lib/foo with EBPYTHONPREFIX" instead of "Replaced PYTHONPATH {/lib/foo} with EBPYTHONPREFIX")

easybuild/tools/module_generator.py
```diff
@@ -260,7 +268,21 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
         if paths is None:
             return ''
-        return self.update_paths(key, paths, prepend=True, allow_abs=allow_abs, expand_relpaths=expand_relpaths)
+        if key == 'PYTHONPATH' and build_option('prefer_ebpythonprefix_over_pythonpath'):
+            # Special condition for PYTHONPATHs that match the standard pattern,
+            # replace with EBPYTHONPREFIX which is added to python sys path at runtime
+            python_paths = [path for path in paths if re.match(r'lib/python\d+\.\d+/site-packages', path)]
+            if python_paths:
+                if len(python_paths) > 1:
+                    raise EasyBuildError('Found multiple paths for PYTHONPATH: ' + ', '.join(pythonpaths))
```
Micket, line 277:
```suggestion
                    raise EasyBuildError('Found multiple paths for PYTHONPATH: ' + ', '.join(python_paths))
```

Flamefire:
Looks good to me so far. Just confused about the commit title of 2f6b631 as it is not a warning but an error that it should/must not occur. So you might want to change either the commit title or implementation to match such that the intention is clear in the future.

Micket:
I've finally gotten around to actually test the code (and to little surprise, there was a bunch of more things to fix).

I was requested to move the duplicate PYTHONPATH check outside of the rewrite code, as it still applies. 

My biggest question is the command line option. It's long, and, will become even longer now than I realized I should have called it "ebpythonprefixes" since we use plural here. 

I'm also not sure how we ought to be dealing with default-true store_true type flags, as there is no way to override these from the command line.

easybuild/tools/module_generator.py
```diff
@@ -213,18 +213,10 @@ def _filter_paths(self, key, paths):
...
-                filtered_paths = None
-            else:
-                added_paths.add(paths)
-                filtered_paths = paths
-        else:
-            # Coerce any iterable/generator into a list
-            if not isinstance(paths, list):
-                paths = list(paths)
-            filtered_paths = [x for x in paths if x not in added_paths and not added_paths.add(x)]
+        # Coerce any iterable/generator into a list
```
Flamefire, line 216: I'm a bit worried about the change here. Why changing `_filter_paths` and even removing the ability to pass a string (which is a breaking change leading to surprising behavior)?

easybuild/tools/module_generator.py
```diff
@@ -213,18 +213,10 @@ def _filter_paths(self, key, paths):
...
+        if not isinstance(paths, list):
+            paths = list(paths)
+        filtered_paths = [x for x in paths if x not in added_paths and not added_paths.add(x)]
         if filtered_paths != paths:
             removed_paths = paths if filtered_paths is None else [x for x in paths if x not in filtered_paths]
             print_warning("Suppressed adding the following path(s) to $%s of the module as they were already added: %s",
```
Flamefire, line 222: In the below line you can use `', '.join(removed_paths)` so you don't need to change the tests and improve the output

easybuild/tools/module_generator.py
```diff
@@ -257,10 +246,41 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
+        :param key: environment variable to append paths to
+        :param paths: list of paths to append
+        :param allow_abs: allow providing of absolute paths
+        :param expand_relpaths: expand relative paths into absolute paths (by prefixing install dir)
+        """
+        if isinstance(paths, str):
+            self.log.debug("Wrapping %s into a list before using it for %s", paths, key)
+            paths = [paths]
+
         paths = self._filter_paths(key, paths)
```
Flamefire, line 264: I think this should be done in `_update_paths`, as I can imagine that an EC (somehow ends up) setting both PYTHONPATH and EBPYTHONPREFIXES and with this logic here you could end up setting the latter 2 times depending on the order as you unconditionally add EBPYTHONPREFIXES in https://github.com/easybuilders/easybuild-framework/pull/4496/files#diff-4af72e9777d353325a29df400ab4229e14defd653e413db51786dc2172d3baefR278

Flamefire (changes requested):
> I'm also not sure how we ought to be dealing with default-true store_true type flags, as there is no way to override these from the command line.
> We currently do this with the following flags:
> cleanup-builddir

We already have a way for that: `--disable-cleanup-builddir`

easybuild/tools/module_generator.py
Micket: The rest of the code (_update_paths) both converted the str case to a list. I simply moved this earlier, which also removes the need for _filter_paths to keep duplicate logic, since regardless it was always converted to a list immediately afterwards anyway.

You can pass a str to `paths`  to `append/prepend/update_paths`.

Flamefire: Yes I see that **our** usages in **this** file basically make this part dead code. However someone or something else might already use this code, so I'm not sure if we should be a bit more careful here. E.g. at least error out when a string is passed instead of an iterable of strings

On the other hand: Calling this method from outside would have been wrong (it is "private") and this PR is targetting 5.x so a breaking change for simpler logic is fine.

easybuild/tools/module_generator.py
```diff
@@ -257,11 +246,38 @@ def prepend_paths(self, key, paths, allow_abs=False, expand_relpaths=True):
...
+        :param paths: list of paths to append
+        :param allow_abs: allow providing of absolute paths
+        :param expand_relpaths: expand relative paths into absolute paths (by prefixing install dir)
+        """
+        if isinstance(paths, str):
+            self.log.debug("Wrapping %s into a list before using it for %s", paths, key)
+            paths = [paths]
+
+        if key == 'PYTHONPATH':
+            python_paths = [path for path in paths if re.match(r'lib/python\d+\.\d+/site-packages', path)]
```
Flamefire, line 265: Checking the logic in `_filter_paths` I see an issue here: You iterate `paths` multiple times: Here and at https://github.com/easybuilders/easybuild-framework/pull/4496/files#diff-4af72e9777d353325a29df400ab4229e14defd653e413db51786dc2172d3baefR275 but inside `_update_paths` you call `_filter_paths` which "coerces a generator into a list" if it isn't a list yet.

This means it this point `paths` might be a generator which can only be iterated once as otherwise the code in `_filter_paths` wasn't required.

Similar to my previous reasoning about breaking change in favor of cleaner code: I would simply disallow passing anything but a a string or a list to `update_paths` and assert that it is a list in `_filter_paths` instead of trying to convert a potentially exhausted generator to a list here or there.

Quick example:
```
paths=(i for i in range(10))
[i for i in paths if i==2] # [2]
[i for i in paths if i==2] # []
```

easybuild/tools/module_generator.py
```diff
@@ -965,6 +981,10 @@ def update_paths(self, key, paths, prepend=True, allow_abs=False, expand_relpath
         :param allow_abs: allow providing of absolute paths
         :param expand_relpaths: expand relative paths into absolute paths (by prefixing install dir)
         """
+        paths = self._filter_paths(key, paths)
+        if paths is None:
```
Flamefire, line 985: I was wondering about how the filtered paths could be `None`:   
This is inside `_filter_paths`:
```
if not filtered_paths:
    filtered_paths = None
```

I would remove both checks, neither is required and we are actually checking stuff twice. Having `_filter_paths` always return a list, even if it is empty is cleaner interface-wise. I mean otherwise you would need to document it as "returns a non-empty list of paths to add or None" instead of "returns a list of paths to add".

If we really want to exit early here, then it is better to remove the above check in `_filter_paths` and change this to:
```suggestion
        if not paths:
```

That would at least check only once

easybuild/tools/module_generator.py
```diff
@@ -976,14 +973,12 @@ def _update_paths(self, key, paths, prepend=True, allow_abs=False, expand_relpat
         Generate prepend-path or append-path statements for the given list of paths.
 
         :param key: environment variable to prepend/append paths to
-        :param paths: list of paths to prepend
+        :param paths: list of paths to prepend/append
```
Flamefire, line 976: Maybe simply:
```suggestion
        :param paths: list of paths to add
```

easybuild/tools/module_generator.py
```diff
@@ -213,25 +213,20 @@ def _filter_paths(self, key, paths):
             return paths
 
         added_paths = self.added_paths_per_key.setdefault(key, set())
-        # Coerce any iterable/generator into a list
-        if not isinstance(paths, list):
-            paths = list(paths)
```
Flamefire: I'd make this an error such that e.g. passing a string doesn't result in garbage

easybuild/tools/module_generator.py
```diff
@@ -213,55 +213,68 @@ def _filter_paths(self, key, paths):
...
+        :param expand_relpaths: expand relative paths into absolute paths (by prefixing install dir)
+        """
+        if isinstance(paths, str):
+            self.log.debug("Wrapping %s into a list before using it for %s", paths, key)
+            paths = [paths]
+        elif not isinstance(paths, list):
+            paths = list(paths)  # coerce any iterator into a list
+
+        if key == 'PYTHONPATH':
+            python_paths = [path for path in paths if re.match(r'lib/python\d+\.\d+/site-packages', path)]
```
boegel, line 262: also consider `lib64/python*/site-packages`?

easybuild/tools/options.py
```diff
@@ -480,6 +480,8 @@ def override_options(self):
                          'int', 'store', None),
             'parallel-extensions-install': ("Install list of extensions in parallel (if supported)",
                                             None, 'store_true', False),
+            'prefer-ebpythonprefix-over-pythonpath': ("Replaces PYTHONPATH with EBPYTHONPREFIX when possible",
```
boegel, line 483: `--prefer-ebpythonprefixes` is specific enough, I think

Micket: We bounced around a bunch of different names, so i'm currently thinking maybe `--replace-pythonpath` is good enough. The EBPYTHONPREFIXES name is kind of an internal implementation detail in easybuild which wouldn't mean much to any user. The comment and documentation could go into more depth to explain it and how it connects to the site customize script our python installs have.

The fact that it doesn't always replace pythonpath, exceptions really are very rare, and this wouldn't be the first option that does what is claims only when it is supported.

easybuild/tools/module_generator.py
```diff
@@ -42,6 +42,7 @@
 from contextlib import contextmanager
 from easybuild.tools import LooseVersion
 from textwrap import wrap
+from typing import Union, Sequence
```
Flamefire, line 45:
```suggestion
from typing import List, Union, Sequence
```

easybuild/tools/module_generator.py
```diff
@@ -205,8 +206,10 @@ def get_modules_path(self, fake=False, mod_path_suffix=None):
 
         return os.path.join(mod_path, mod_path_suffix)
 
-    def _filter_paths(self, key, paths):
+    def _filter_paths(self, key: str, paths: list) -> list:
```
Flamefire, line 209:
```suggestion
    def _filter_paths(self, key: str, paths: List[str]) -> List[str]:
```

easybuild/tools/module_generator.py
```diff
@@ -564,12 +568,12 @@ def unload_module(self, mod_name):
         """
         raise NotImplementedError
 
-    def _update_paths(self, key, paths, prepend=True, allow_abs=False, expand_relpaths=True):
+    def _update_paths(self, key: str, paths: list, prepend=True, allow_abs=False, expand_relpaths=True) -> str:
```
Flamefire, line 571:
```suggestion
    def _update_paths(self, key: str, paths: List[str], prepend=True, allow_abs=False, expand_relpaths=True) -> str:
```

easybuild/tools/module_generator.py
```diff
@@ -1437,12 +1441,12 @@ def modulerc(self, module_version=None, filepath=None, modulerc_txt=None):
         return super(ModuleGeneratorLua, self).modulerc(module_version=module_version, filepath=filepath,
                                                         modulerc_txt=modulerc_txt)
 
-    def _update_paths(self, key, paths, prepend=True, allow_abs=False, expand_relpaths=True):
+    def _update_paths(self, key: str, paths: list, prepend=True, allow_abs=False, expand_relpaths=True) -> str:
```
Flamefire, line 1444:
```suggestion
    def _update_paths(self, key: str, paths: List[str], prepend=True, allow_abs=False, expand_relpaths=True) -> str:
```

Flamefire (commented):
I was thinking about `typing` in EasyBuild just the other day as it makes the code so much clearer, so thanks for starting it! 🎉    
I added some suggestions about using `List[str]` instead of the generic `list`

easybuild/tools/module_generator.py
Flamefire: I guess that would be wrong to have both, so yeah:
```suggestion
            python_paths = [path for path in paths if re.match(r'lib(64)?/python\d+\.\d+/site-packages', path)]
```

Flamefire (approved):
LGTM except for that potential enhancement @boegel mentioned.

BTW: You can apply all suggestions in a single commit from the code view (there will be a 2nd button like "add to batch")

Flamefire (changes requested):
This has a **serious issue!!!**

While checking our hook we use:
```
    def module_write_hook(self, filepath, module_txt):
        if 'Python' in get_dep_names(self.cfg):
            return _replace_pythonpath(module_txt)
```

That condition there is crucial! We must not use `EBPYTHONPREFIXES` when `Python` is not a dependency. Given the possibility of transitive dependencies this gets quite hard, so a trade-off would be to disable this for SYSTEM level Easyconfigs

If we don't then e.g. the `EasyBuild` module installed with EB won't work anymore.

Micket:
Yeah I was actually just testing that out with Conda packages as well actually (I got stuck with the testing as I was having issues just getting any of the Conda easyconfigs to even install at all). Thoughts:
1. adding an option for easyconfigs/blocks to opt out of this transformation. Presumably needed for Conda packages and such, as only things that understand sitecustomize.py thing can work with these.
2. Initially I wanted to make PythonBundle and PythonPackage just generate the EBPYTHONPREFIXES directly rather than relying on this transformation, but I was requested on making it a configurable parameter + there are many easyconfigs that specify PYTHONPATH via modextrapaths (which really complicated things) I went in this direction instead. Maybe it's worth revisiting that idea
  - Maybe easyconfigs should just lay off the PYTHONPATH variable the same way we don't specify LD_LIBRARY_PATH manually either. Detecting that this path is in use and adding it automatically seems appropriate. Then one could always make that code EBPYTHONPREFIXES-aware, as well as compatible with multi-deps. In fact, maybe this already happens???
  - If so, then there is no need for some rewrite-hook here, and instead it's just if up to the blocks to pick which path they want.

boegel:
@Micket This should be closed, in favor of #4539?

Micket:
Sure, there was some cleaning up of some of the *_update functions that i wanted to keep regardless, but i'll make a clean new branch for that.

PR #4498, opened 2024-04-04
don't allow using `--pr-commit-msg` when only adding new files with `--new-pr` (unless `--force` is used)

branfosj:
We should force the generated message in these cases.

An alternative to raising an error (as implemented here) would be to generate the default PR title and replace the `commit_msg` in these cases while printing a warning.

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -1129,7 +1129,7 @@ def _easyconfigs_pr_common(paths, ecs, start_branch=None, pr_branch=None, start_
             msg += "the PR title will be automatically generated."
             if build_option('force'):
                 print_msg(msg)
-                print_msg("Using the specified --pr-commit-msg As the force build option was specified.")
+                print_msg("Using the specified --pr-commit-msg as the force build option was specified.")
             else:
                 raise EasyBuildError(msg)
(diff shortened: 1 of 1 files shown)


Micket:
@smoors We'd have to check that it wasn't modifying any file before doing that. But why not

easybuild/tools/github.py
```diff
@@ -1124,6 +1124,14 @@ def _easyconfigs_pr_common(paths, ecs, start_branch=None, pr_branch=None, start_
 
     # figure out commit message to use
     if commit_msg:
+        if pr_target_repo == GITHUB_EASYCONFIGS_REPO and all(file_info['new']) and not paths['files_to_delete']:
+            msg = "When only adding new easyconfigs usually a PR commit msg (--pr-commit-msg) should not be used, as "
+            msg += "the PR title will be automatically generated."
+            if build_option('force'):
+                print_msg(msg)
+                print_msg("Using the specified --pr-commit-msg As the force build option was specified.")
```
smoors:
```suggestion
                print_msg("Using the specified --pr-commit-msg as the force build option was specified.")
```

smoors (approved):
works as advertised, lgtm

PR #4503, opened 2024-04-06
sync with develop (20240406)

boegel:


Micket (dismissed):
lgtm

Micket:
oh, turns out we have crashing behavior

```

FAIL: test_eb_crash (test.framework.toy_build.ToyBuildTest.test_eb_crash)
Test behaviour when EasyBuild crashes, for example due to a buggy hook
  File "/tmp/runner/1dd44a9d55128d5f2b1b9646ff75e2cbe93b4cf3/lib/python3.12/site-packages/test/framework/toy_build.py", line 4175, in test_eb_crash
    self.assertFalse("This should never be reached, main function should have crashed!")
AssertionError: 'This should never be reached, main function should have crashed!' is not false
```

boegel:
@Micket No, that test is checking whether `main` crashes as it should when a broken hook is used.

That's no longer the case, I messed up something when resolving a merge conflict, I'm looking into it (yaay for tests btw!)

boegel:
I wasn’t passing down `args` in `main_with_hooks` in `easybuild/main.py`, fixed with force push in bf13328

Micket (dismissed):
lgtm

boegel:
Excessive output produced by test suite fixed by avoiding that `download_file` used in `fetch_files_from_commit` uses `trace=False` (like it already does in `fetch_files_from_pr`), see ce9a555a2

Micket (approved):
lgtm

PR #4512, opened 2024-04-18
don't delete existing environment module files when using `--dump-env-script` with `--force` or `--rebuild`

Flamefire:
This option should only require `--force` when the env file exists.

Otherwise, when the module already exists `--force` should not be required.

In any case the module file must not be deleted.


boegel (approved):
lgtm

PR #4511, opened 2024-04-17
add support for alternate easyconfig parameters/templates/constants

jfgrimm:
As discussed in slack, this is an alternative approach to introducing new parameters and deprecating old ones in one go. Instead, it's a multi-stage process:
1. Introduce new names, as alternatives
2. Do the renaming
3. Deprecate
This would allow is to do it iteratively across the different parameters / etc. over a series of EB 5.x releases.

https://github.com/easybuilders/easybuild-framework/issues/4464

To play around with this, check out the branch and modify:
* `easybuild/framework/easyconfig/parser.py`:
```
 45 # alternate easyconfig parameters, and their non-deprecated equivalents                                                 
 46 ALTERNATE_PARAMETERS = {                                                                                                
 47     # <new_param>: <equivalent_param>,                                                                                  
 48 } 
 ```
 
* `easybuild/framework/easyconfig/templates.py`:
```
164 # alternate templates, and their equivalents                                                                            
165 ALTERNATE_TEMPLATES = {                                                                                                 
166     # <new>: <equivalent_template>,                                                                                     
167 }
```
```
174 # alternate template constants, and their equivalents                                                                   
175 ALTERNATE_TEMPLATE_CONSTANTS = {                                                                                        
176     # <new_template_constant>: <equivalent_template_constant>,                                                          
177 }
```

Issue #4464, opened 2024-02-23
refactor easyconfig parameters and template constants to use a consistent naming style


jfgrimm:
for easyconfigs parameters:
| type | current name | proposed (if changed) |
| ------ | -------------------- | -------------- |
|mandatory | `description` | |
| | `docurls` | `doc_urls` |
| | `homepage` | |
| | `name` | | 
| | `software_license` | | 
| | `software_license_urls` | | 
| | `toolchain` | |
| | `version`  || 
| toolchain | `onlytcmod` | ??? | 
| | `toolchainopts` | `toolchain_opts` |
| build | `banned_linked_shared_libs` | |
| | `bin_lib_subdirs` | |
| | 
| | 
| | 
| | 
| | 
| | 
| | 
| | 
| | 
| | 
| | 
| | 
| | 
| |

jfgrimm:
templates:

|current|proposed|
| `module_name` | |
| `nameletter` | |
| `toolchain_name` | |
| `toolchain_version` | |
| `version_major_minor` | |
| `version_major` | |
| `version_minor` | |
| `cudamajver` | |
| `cudashortver` | |
| `cudaver` | |
| `cudamajver` | |
| `cudashortver` | |
| `cudaver` | |
| `javamajver` | |
| `javashortver` | |
| `javaver` | |
| `perlmajver` | |
| `perlshortver` | |
| `perlver` | |
| `pymajver` | |
| `pyshortver` | |
| `pyver` | |
| `rmajver` | |
| `rshortver` | |
| `rver` | |
| `bitbucket_account` | |
| `github_account` | |
| `name` | |
| `parallel` | |
| `version` | |
| `versionsuffix` | |
| `versionprefix` | |
| `namelower` | |
| `nameletterlower` | |
| `builddir` | |
| `installdir` | |
| `start_dir` | |
| `arch` | |
| `sysroot` | |
| `mpi_cmd_prefix` | |
| `cuda_compute_capabilities` | |
| `cuda_cc_cmake` | |
| `cuda_cc_space_sep` | |
| `cuda_cc_semicolon_sep` | |
| `cuda_sm_comma_sep` | |
| `cuda_sm_space_sep` | |

jfgrimm:
template constants:

|current|proposed|
|---|---|
| `APACHE_SOURCE` | |
| `BITBUCKET_SOURCE` | |
| `BITBUCKET_DOWNLOADS` | |
| `CRAN_SOURCE` | |
[...]

PR #4514, opened 2024-04-23
add renamed alternate constants/templates/parameters

jfgrimm:
adds support for renamed alternate:
* easyconfig parameters
* easyconfig templates
* easyconfig template constants

~~currently includes the changes from #4511; will rebase once that has been merged~~

cfr. #4464


easybuild/framework/easyconfig/parser.py
```diff
@@ -45,6 +45,59 @@
 # alternate easyconfig parameters, and their non-deprecated equivalents
 ALTERNATE_PARAMETERS = {
     # <new_param>: <equivalent_param>,
+    'build_deps': 'builddependencies',
+    'build_in_install_dir': 'buildininstalldir',
+    'build_opts': 'buildopts',
+    'build_stats': 'buildstats',
+    'clean_up_old_build': 'cleanupoldbuild',
+    'clean_up_old_install': 'cleanupoldinstall',
+    'configure_opts': 'configopts',
```
boegel, line 54: Any reason we don't abbreviate `configure` to `config` (or even `cfg`, but that's a stretch)?

```suggestion
    'config_opts': 'configopts',
```

easybuild/framework/easyconfig/parser.py
```diff
@@ -45,6 +45,59 @@
...
+    'deps': 'dependencies',
+    'doc_paths': 'docpaths',
+    'doc_urls': 'docurls',
+    'dont_create_install_dir': 'dontcreateinstalldir',
+    'exts_class_map': 'exts_classmap',
+    'exts_default_class': 'exts_default_class',
```
boegel, line 60: should be `exts_defaultclass` on right-hand side:
```suggestion
    'exts_default_class': 'exts_defaultclass',

File: easybuild/framework/easyconfig/parser.py [hunk 3]
```diff
@@ -45,6 +45,59 @@
...
+    'env_mod_aliases': 'modaliases',
+    'env_mod_alt_soft_name': 'modaltsoftname',
+    'modulepath_prepend_paths': 'moddependpaths',
+    'env_mod_extra_paths_append': 'modextrapaths_append',
+    'env_mod_extra_paths': 'modextrapaths',
+    'env_mod_extra_vars': 'modextravars',
+    'env_mod_load_msg': 'modloadmsg',
+    'env_mod_lua_footer': 'modluafooter',
+    'env_mod_tcl_footer': 'modtclfooter',
+    'env_mod_class': 'moduleclass',
```
boegel, line 77: Maybe we should avoid `class` here, and go with `env_mod_category`?
```suggestion
    'env_mod_category': 'moduleclass',
```

easybuild/framework/easyconfig/parser.py
```diff
@@ -45,6 +45,59 @@
...
+    'env_mod_depends_on': 'module_depends_on',
+    'env_mod_force_unload': 'moduleforceunload',
+    'env_mod_load_no_conflict': 'moduleloadnoconflict',
+    'env_mod_unload_msg': 'modunloadmsg',
+    'only_toolchain_mod_env': 'onlytcmod',
```
boegel, line 82: `only_toolchain_env_mod`

easybuild/framework/easyconfig/parser.py
```diff
@@ -45,6 +45,59 @@
...
+    'os_deps': 'osdependencies',
+    'post_install_cmds': 'postinstallcmds',
+    'post_install_msgs': 'postinstallmsgs',
+    'post_install_patches': 'postinstallpatches',
+    'pre_build_opts': 'prebuildopts',
+    'pre_configure_opts': 'preconfigopts',
```
boegel, line 88:
```suggestion
    'pre_config_opts': 'preconfigopts',
```

boegel (changes requested):

easybuild/framework/easyconfig/parser.py
jfgrimm: we're using `config` for "configuration" in the EB configuration options, e.g.:
```
--avail-config-file-constants
--config-files
--ignore-config-files
--job-backend-config
```

easybuild/framework/easyconfig/parser.py
jfgrimm:
```suggestion
    'only_toolchain_env_mod': 'onlytcmod',
```

easybuild/framework/easyconfig/parser.py
boegel: The context there is different, it's fine is we used `config` for both imho.

It's clear that `config_opts` in an easyconfig file specified options for the configure command (which could be `cmake`, it's definitely not always `configure`).

boegelbot: As discussed, let's stick to `configure_opts` (to align with `configure` step)

easybuild/framework/easyconfig/parser.py
boegelbot: As discussed, let's stick to `pre_configure_opts` (to align with `configure` step)

boegel: As discussed, let's stick to `pre_configure_opts` (to align with `configure` step)

easybuild/framework/easyconfig/parser.py
boegel: As discussed, let's stick to `configure_opts` (to align with `configure` step)

boegel (approved):
lgtm

PR #4519, opened 2024-04-27
code cleanup in `easyblock.py`

Flamefire:
- Remove `pass` statements where not required
- Enhance docstrings
- Use `.items` when values of dict is used (faster, clearer)
- `basename` instead of `split()[-1]`


boegel (approved):
lgtm

PR #4520, opened 2024-04-27
versions checks to avoid mixing major versions across the EasyBuild components

branfosj:
This PR will add various version checks and is EB5 prep for the EB4 branch.

* [x] check that framework and blocks major versions are the same and fails if not
* [x] check `--from-pr` / `--include-easyblocks-from-pr` branches match in use versions

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -621,7 +621,7 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None, pr
 
     # compare running Framework and EasyBlocks versions
     if 'UNKNOWN' in EASYBLOCKS_VERSION:
-        print_msg('Unable to determine EasyBlocks version, so we'll assume it is not different from Framework')
+        print_msg("Unable to determine EasyBlocks version, so we'll assume it is not different from Framework")
     elif different_major_versions(FRAMEWORK_VERSION, EASYBLOCKS_VERSION):
         raise EasyBuildError("Framework (%s) and EasyBlock (%s) major versions are different." % (FRAMEWORK_VERSION,
                                                                                                   EASYBLOCKS_VERSION))
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -621,8 +621,8 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None, pr
 
     # compare running Framework and EasyBlocks versions
     if different_major_versions(FRAMEWORK_VERSION, EASYBLOCKS_VERSION):
-        raise EasyBuildError(f"Framework ({FRAMEWORK_VERSION}) and EasyBlock ({EASYBLOCKS_VERSION}) major versions "
-                             "are different.")
+        raise EasyBuildError("Framework (%s) and EasyBlock (%s) major versions are different." % (FRAMEWORK_VERSION,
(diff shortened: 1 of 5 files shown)


easybuild/tools/version.py
```diff
@@ -103,3 +104,16 @@ def this_is_easybuild():
...
+    """Compare major versions"""
+    # Deal with version instances being either strings or LooseVersion
+    if not isinstance(v1, LooseVersion):
+        v1 = LooseVersion(v1)
+    if not isinstance(v2, LooseVersion):
+        v2 = LooseVersion(v2)
+
+    if v1.version[0] == v2.version[0]:
+        return False
+    return True
```
boegel, line 119:
```suggestion
    return v1.version[0] != v2.version[0]

File: easybuild/main.py [hunk 2]
```diff
@@ -618,6 +619,13 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None, pr
     (build_specs, _log, logfile, robot_path, search_query, eb_tmpdir, try_to_generate,
      from_pr_list, tweaked_ecs_paths) = cfg_settings
 
+    # compare running Framework and EasyBlocks versions
+    if EASYBLOCKS_VERSION == UNKNOWN_VERSION:
+        print_msg("Unable to determine EasyBlocks version, so we'll assume it is not different from Framework")
```
boegel, line 624: This should be a warning?

boegel (changes requested):

easybuild/main.py
branfosj: 0a5c99f

easybuild/tools/version.py
branfosj: 0a5c99f

easybuild/main.py
branfosj: Printing something here is problematic, so we'll log it as a warning instead.

The most likely place for this to happen is running the framework unit tests without an easyblocks installation. If you try to use EB in that case then you'll fail really quickly.

The only other case I can think of is if something breaks the version lookup of the blocks. That is a rare enough possibility that I think we should ignore.

easybuild/tools/github.py
```diff
@@ -587,9 +588,39 @@ def fetch_files_from_pr(pr, path=None, github_user=None, github_account=None, gi
         else:
             raise EasyBuildError("Couldn't find path to patched file %s", full_path)
 
+    if github_repo == GITHUB_EASYCONFIGS_REPO:
+        ver = _get_version_for_repo(os.path.join(final_path, 'setup.py'))
+    elif github_repo == GITHUB_EASYBLOCKS_REPO:
+        ver = _get_version_for_repo(os.path.join(final_path, 'easybuild', 'easyblocks', '__init__.py'))
+
+    if different_major_versions(FRAMEWORK_VERSION, ver):
+        raise EasyBuildError("Framework (%s) is a different major version than PR target (%s)." % (FRAMEWORK_VERSION,
```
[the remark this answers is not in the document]
boegel: fixed in https://github.com/branfosj/easybuild-framework/pull/10

boegel (changes requested):
@branfosj see https://github.com/branfosj/easybuild-framework/pull/10

boegel (approved):

PR #4521, opened 2024-04-29
5.0.x consideration: ruff format filetools.py

fgeorgatos:
Hi,

reference: https://github.com/astral-sh/ruff

This is only sent as proof of concept, not necessarily intending to merge anything - yet.
It is an example of the diff by using the python tool `ruff` against `filetools.py`. The intention is to have a shared example.

The supplied link has a bit of context why this could be interesting: a major version upgrade is a good moment for x-check!

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -40,6 +40,7 @@ Authors:
 * Damian Alvarez (Forschungszentrum Juelich GmbH)
 * Maxime Boissonneault (Compute Canada)
 """
+
 import datetime
 import difflib
 import glob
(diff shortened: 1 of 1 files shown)


fgeorgatos:
the produced diff has failed to pass lint for 3.10 & 3.12 here, which feels to me as a bit of a code smell (lines became slightly longer as a consequence of excessive escaping, which asks why was it is so in the first place):
* https://github.com/easybuilders/easybuild-framework/pull/4521/files#diff-0ca2799267ffd6a801a47e695fbe66c047a44775b616b26148100ec11e6ce5e5L2761

Flamefire:
> the produced diff has failed to pass lint for 3.10 & 3.12 here

One failure is due to: ` fn[len(common_prefix):]` -> ` fn[len(common_prefix) :] `, where there is some seemingly excessive line wrapping too. So I'm unsure if we want that.

A lot of the diff is changing the single to double quotes. When I introduced a code formatter into some other code base I tried to adjust the options to produce the smallest diff possible. I guess we can adjust the type of quotes used and the line length to reduce the diff a lot, can't we?

> which feels to me as a bit of a code smell (lines became slightly longer as a consequence of excessive escaping, which asks why was it is so in the first place):
> 
>     * https://github.com/easybuilders/easybuild-framework/pull/4521/files#diff-0ca2799267ffd6a801a47e695fbe66c047a44775b616b26148100ec11e6ce5e5L2761

What exactly do you mean by that?

boegel:
@fgeorgatos As discussed during the EasyBuild 5.0 sync meeting today, I don't think we want to get into this right now, since adopting `ruff` does not involve any breaking changes (so it's out-of-scope for the effort on getting EasyBuild 5.0 released).

It also seems like the Ruff configuration would have to be tweaked a bit, to be less aggressive in dealing with long lines, and to not aggressively replace `'` with `"`.

I'll go ahead and close this PR, if we do want to adopt `ruff` going forward we should come up with a clear plan on what needs to be done to introduce it more gradually, incl. necessary code changes, CI changes, docs, recommendations to contributors (like Git pre-commit hook), etc.

It would also be nice if Hound CI has support for using `ruff`, it doesn't seem to be aware of it at all currently...

fgeorgatos:
If the Hound CI comes to support ruff (which I think is likely), it may turn out to be the easy way forward. Let's hit pause.

PR #4522, opened 2024-04-29
use generic chmod command to reset permissions of reproducible archives

lexming:
As discussed, we found out that BSD tar lacks the `--mode` option. So we work around that by manually executing `chmod`.


boegel (approved):
lgtm

PR #4523, opened 2024-05-03
switch checksum default to sha256

branfosj:


Micket (dismissed):
This is a no-brainer

Micket (approved):
lgtm

PR #4525, opened 2024-05-03
add new `get_cwd` function to `tools.filetools` to retrieve current working directory

lexming:
Simple addition of a wrapper around `os.getcwd` to be able to improve error reporting whenever it fails.

This PR includes unit tests for the new method and also adapts all current uses of `os.getcwd` to the new `get_cwd`.

Motivation is to improve error reporting on issues such as those described in #4390


easybuild/tools/filetools.py
```diff
@@ -407,26 +407,41 @@ def remove(paths):
...
 
     try:
         os.chdir(path)
     except OSError as err:
-        raise EasyBuildError("Failed to change from %s to %s: %s", cwd, path, err)
+        raise EasyBuildError("Failed to change from %s to %s: %s", prev_dir, path, err)
 
-    return cwd
+    # determine final working directory: must exist
+    get_cwd()
```
boegel, line 442: I don't see the point of this, if the directory doesn't exist then `os.chdir` above would've failed?

easybuild/tools/robot.py
```diff
@@ -45,7 +45,7 @@
 from easybuild.framework.easyconfig.tools import find_resolved_modules, skip_available
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
-from easybuild.tools.filetools import det_common_path_prefix, search_file
+from easybuild.tools.filetools import get_cwd, det_common_path_prefix, search_file
```
boegel, line 48:
```suggestion
from easybuild.tools.filetools import det_common_path_prefix, get_cwd, search_file

File: easybuild/tools/run.py [hunk 3]
```diff
@@ -315,7 +315,10 @@ def to_cmd_str(cmd):
         qa_wait_patterns = []
 
     if work_dir is None:
-        work_dir = os.getcwd()
+        try:
+            work_dir = os.getcwd()
+        except FileNotFoundError:
+            raise EasyBuildError("Working directory does not exist")
```
boegel, line 321:
```suggestion
            raise EasyBuildError("Current working directory does not exist")
```

test/framework/filetools.py
```diff
@@ -2222,6 +2222,19 @@ def test_copy(self):
...
+    def test_get_cwd(self):
+        """Test get_cwd"""
+        toy_dir = os.path.join(self.test_prefix, "test_get_cwd_dir")
+        os.mkdir(toy_dir)
+        os.chdir(toy_dir)
+
+        self.assertTrue(os.path.samefile(ft.get_cwd(), toy_dir))
+
+        os.rmdir(toy_dir)
+        self.assertErrorRegex(EasyBuildError, "Working directory does not exist", ft.get_cwd)
```
boegel, line 2234:
```suggestion
        self.assertErrorRegex(EasyBuildError, "Current working directory does not exist", ft.get_cwd)

File: easybuild/tools/filetools.py [hunk 5]
```diff
@@ -407,26 +407,41 @@ def remove(paths):
...
 
+def get_cwd(must_exist=True):
+    """
+    Retrieve current working directory
+    """
+    try:
+        cwd = os.getcwd()
+    except FileNotFoundError as err:
+        if must_exist is True:
+            raise EasyBuildError("Working directory does not exist")
```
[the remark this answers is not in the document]
lexming: I made the error message even better :)

easybuild/tools/filetools.py
lexming: This is meant to catch filesystems in a faulty state, where some things work, like listing directories, but still have an undefined `CWD`. It's a rarity, but we experience such situations with the NFS mount of one of our storage systems, which can end up in a very weird state after complex builds that involve creation of many files and folders and jumping around a lot (for instance the issue with OpenFOAM in https://github.com/easybuilders/easybuild-framework/pull/4390)

lexming: I added an extra comment clarifying the purpose of this line

easybuild/tools/run.py
```diff
@@ -83,6 +83,10 @@
     "ulimit -u",  # used in det_parallelism
 )
 
+CWD_NOTFOUND_ERROR = (
```
boegel, line 86: @lexming parking this in `tools.run` doesn't make much sense to me, I propose we move it to `tools.build_log`, see https://github.com/lexming/easybuild-framework/pull/4

If you agree with that minor change, this PR is good to merge

boegel (changes requested):

easybuild/tools/run.py
lexming: `tools.build_log` makes more sense indeed, thx for the PR!

boegel (approved):
lgtm

PR #4526, opened 2024-05-05
deprecate old checksum options (incl. md5)

branfosj:
follow on to #4523 - deprecate the old checksum options (including MD5)


boegel (approved):
lgtm

PR #4530, opened 2024-05-10
do not run unit tests on Python 3.5

branfosj:
See https://github.com/easybuilders/easybuild-easyblocks/pull/3326


ocaisa (approved):
LGTM

PR #4531, opened 2024-05-13
Make generic EasyBlock usable to install software

lexming:
This is quite a big change, it makes the base `EasyBlock` class usable by adding generic implementations of the missing configure, build and install step.

This is how this works:
* easyconfigs that define an `easyblock` will use that one
* easyconfigs with a `name` that matches an existing `easyblock` will use that one (_i.e._ `EB_name`)
* easyconfigs without any specific easyblock will use `EasyBlock`

Framework will define for each step of the installation the following parameters:
* `pre_[step]_cmds`: list of commands to execute before the main command of the step (analogue in mechanics to `postinstallcmds`)
* `pre[step]opts`: string to be prepended to main command of the step
* `[step]_cmd`: main command of the step
* `[step]opts`: string to append to main command of the step

Easyconfigs can then define any of the `[step]_cmd` and then `EasyBlock` will carry out that step. Otherwise the step is skipped.

Motivation for this change is two-fold:
* **It becomes much easier to make easyconfigs for non-standard software without their own easyblock**:
    Basically `EasyBlock` becomes a _CmdCmdCmd_ easyblock. For instance, I recently had to install something with Bazel that needed a custom `./configure.py` for the configure step and `bazel build` in the build step. With this approach I can do that installation without having to write a new easyblock or hacking other easyblocks not meant for this task.
* **Provides a standard implementation on how easyblock should carry out each build step**:
    I initially only implemented the addition of `pre[step]cmds` parameters to execute commands at the beginning of each step. However, it is extremely weird to define those parameters in framework without any implementation. Custom easyblocks are then free to do whatever they like with those parameters. This PR adds a `run_main_step_action` method that serves as a reference on how all the parameters should be combined and executed.

----------------

Changelog:
* add easyconfig parameters: `pre_configure_cmds`, `pre_build_cmds`, `pre_test_cmds`, `pre_install_cmds`
* add easyconfig parameters: `configure_cmd`, `build_cmd`, `test_cmd`, `install_cmd`
* add method `easyblock._run_command_stack()`
* implement `easyblock.configure_step()` using `easyblock._run_command_stack()`
* implement `easyblock.build_step()` using `easyblock._run_command_stack()`
* implement `easyblock.install_step()` using `easyblock._run_command_stack()`
* update `easyblock.test_step()` using `easyblock._run_command_stack()`
* update `easyblock.run_post_install_commands()` using `easyblock._run_command_stack()`
* update `easyblock.get_easyblock_class()` behaviour as follows:
  *  easyconfigs that define an `easyblock` will use that one
  * easyconfigs with a `name` that matches an existing `easyblock` will use that one (_i.e._ `EB_name`)
  * easyconfigs without any specific easyblock will use `EasyBlock`
* remove unused `error_on_failed_import` attribute from `easyblock.get_easyblock_class()`
* deprecate easyconfig parameter `runtest` in favor of `test_cmd`
* standardize format of help strings of all easyconfig parameters

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -1860,11 +1860,18 @@ def det_installversion(version, toolchain_name, toolchain_version, prefix, suffi
     _log.nosupport('Use det_full_ec_version from easybuild.tools.module_generator instead of %s' % old_fn, '2.0')
 
 
-def get_easyblock_class(easyblock, name=None, error_on_missing_easyblock=True, **kwargs):
+def get_easyblock_class(easyblock, name=None, error_on_failed_import=False, error_on_missing_easyblock=True, **kwargs):
     """
     Get class for a particular easyblock (or use default)
     """
     modulepath, class_name = None, None
+
+    if error_on_failed_import:
+        _log.deprecated(
+            "Use of error_on_failed_import argument in Easyconfig.get_easyblock_class is deprecated, "
+            "use 'error_on_missing_easyblock' instead",
+            "6.0"
+        )
 
     if easyblock:
         # something was specified, lets parse it
@@ -1912,6 +1919,8 @@ def get_easyblock_class(easyblock, name=None, error_on_missing_easyblock=True, *
         if error_re.match(str(err)):
             if error_on_missing_easyblock:
                 raise EasyBuildError(f"Software-specific easyblock '{class_name}' not found for {name}")
+            elif error_on_failed_import:
+                raise EasyBuildError(f"Failed to import '{class_name}' easyblock: {err}")
             else:
                 _log.debug(f"Easyblock for {class_name} not found, but ignoring it: {err}")
         else:


boegel:
I need to give this a detailed review, but in general I quite like this idea to support "easyblock-less" installations.

It lowers the bar to entry for people since this may avoid the need to implement a custom easyblock for simple stuff.

It does perhaps also introduce a risk to encourage producing messier easyconfigs, but it's not like we don't have those already...

easybuild/framework/easyconfig/format/format.py
```diff
@@ -65,9 +65,11 @@
     ['source_urls', 'sources', 'patches', 'checksums'],
     DEPENDENCY_PARAMETERS + ['multi_deps'],
     ['osdependencies'],
-    ['preconfigopts', 'configopts'],
-    ['prebuildopts', 'buildopts'],
-    ['preinstallopts', 'installopts'],
+    ['preconfigcmds', 'preconfigopts', 'configure_cmd', 'configopts'],
+    ['prebuildcmds', 'prebuildopts', 'build_cmd', 'buildopts'],
```
boegel, line 69: make sure you take into account naming policy here (cfr. https://github.com/easybuilders/easybuild-framework/issues/4464)

boegel (changes requested):

easybuild/framework/easyconfig/format/format.py
lexming: done in https://github.com/easybuilders/easybuild-framework/pull/4531/commits/38341704c415326fb0c2e175b1b5566ba3234143

lexming:
Example easyconfig using `EasyBlock`:

```
name = 'CORSIKA'
version = '77550'

homepage = "https://www.iap.kit.edu/corsika"
description = """CORSIKA (COsmic Ray SImulations for KAscade) is a program for detailed
simulation of extensive air showers initiated by high energy cosmic ray
particles. Protons, light nuclei up to iron, photons, and many other particles
may be treated as primaries."""

toolchain = {'name': 'foss', 'version': '2023a'}
toolchainopts = {'usempi': True}

download_instructions = "Sources have to be requested to the developers"
checksums = ['fed74c144e22deb5a7c1d2dc1f04f0100eb2732cb48665a3da49ce471a3775ee']

dependencies = [
    ("ROOT", "6.30.06"),
]

# custom coconut script does not recognize -j
parallel = False

# execute ./coconut manually with your own options and extract configure command from top of config.log
configure_cmd = "./configure"
configopts = "CORDETECTOR=HORIZONTAL CORTIMELIB=TIMEAUTO CORHEMODEL=QGSJETII CORLEMODEL=URQMD "
configopts += "--enable-UPWARD --enable-SLANT --enable-THIN --enable-COREAS "
configopts += "--enable-PARALLEL --with-mpirunner-lib --enable-PARALLELIB "

build_cmd = "./coconut"
buildopts = "--batch"

preinstallopts = "mkdir -p %(installdir)s/bin && "
install_cmd = "cp"
installopts = "%(builddir)s/%(namelower)s-%(version)s/run/%(namelower)s%(version)s* %(installdir)s/bin/"

sanity_check_paths = {
    'files': ['bin/corsika77550Linux_QGSII_urqmd_thin_coreas_parallel'],
    'dirs': []
}

moduleclass = "phys"
```

Test report for this PR with this example easyconfig:
https://gist.github.com/lexming/776d52d61b8bdce7aa47c068fe092af4

easybuild/framework/easyblock.py
```diff
@@ -2720,20 +2720,66 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
...
+            return res.output
 
     def test_step(self):
         """Run unit tests provided by software (if any)."""
-        unit_test_cmd = self.cfg['runtest']
-        if unit_test_cmd:
-            self.log.debug(f"Trying to execute {unit_test_cmd} as a command for running unit tests...")
-            res = run_shell_cmd(unit_test_cmd)
+        if self.cfg['runtest'] is not None:
+            self.log.deprecated("Easyconfig parameter 'runtest' is deprecated, use 'test_cmd' instead.", "6.0")
```
boegel, line 2772: @lexming This implies updating all easyblocks/easyconfigs to use `test_cmd` instead of `runtest` before we release EasyBuild 5.0, so please open an issue on that in both repos and tag it for EasyBuild 5.0

easybuild/framework/easyblock.py
```diff
@@ -2720,20 +2720,66 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
...
+                raise EasyBuildError(f"Invalid element in '{caller}', not a string: {cmd}")
+            run_shell_cmd(cmd)
+
+    def run_step_main_action(self, action, step_name, pre_cmds=None, pre_opts="", post_opts=""):
+        """Construct main command of step and execute it"""
+        if pre_cmds:
+            self._run_command_stack(pre_cmds, f"pre_{step_name}_cmds")
+
+        step_cmd = f"{pre_opts} {action} {post_opts}"
+        return run_shell_cmd(step_cmd)
```
boegel, line 2743:
```suggestion
    def _run_cmds(self, cmds, caller):
        """Execute list of shell commands"""
        self.log.debug(f"List of commands to be executed: {cmds}")

        # make sure we have a list of commands
        if not isinstance(cmds, (list, tuple)):
            error_msg = f"Invalid argument for '{caller}', should be list or tuple of strings: {cmds}"
            raise EasyBuildError(error_msg)

        for cmd in cmds:
            if not isinstance(cmd, str):
                raise EasyBuildError(f"Invalid element in '{caller}', not a string: {cmd}")
            run_shell_cmd(cmd)

    def run_step_main_action(self, action, step_name, pre_cmds=None, pre_opts="", post_opts=""):
        """Construct main command of step and execute it"""
        if pre_cmds:
            self._run_cmds(pre_cmds, f"pre_{step_name}_cmds")

        step_cmd = f"{pre_opts} {action} {post_opts}"
        return run_shell_cmd(step_cmd)
```

easybuild/framework/easyblock.py
```diff
@@ -2720,20 +2720,66 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
...
-        """Configure build  (abstract method)."""
-        raise NotImplementedError
+        """Configure build."""
+        if self.cfg['configure_cmd'] is not None:
+            res = self.run_step_main_action(
+                self.cfg['configure_cmd'],
+                "configure",
+                self.cfg['pre_configure_cmds'],
+                self.cfg['preconfigopts'],
+                self.cfg['configopts'],
```
boegel, line 2753: This looks kind of weird without using named options, since `pre_configure_cmd` is specified after `configure_cmd`, so at the very least this should become:

```suggestion
                'configure',
                pre_cmds=self.cfg['pre_configure_cmds'],
                pre_opts=self.cfg['preconfigopts'],
                post_opts=self.cfg['configopts'],
```

Even better would be to use only named options, so also use `action='self.cfg['configure_cmd']'` and `action='configure'` (and making those both `None` by default in `run_step_main_action` and mandatory with an `is None` check in there.

Also, maybe rename `run_step_main_action` to `run_core_step`, and only pass a string like `configure`, let `run_core_step` figure out the rest (to use `configure_cmd`, `preconfigopts`, etc.)

easybuild/framework/easyblock.py
boegel: Also, what if both `test_cmd` and `runtest` are set?

Either we error out, or we let one overrule the other (with logging).

easybuild/framework/easyblock.py
```diff
@@ -2992,18 +3046,7 @@ def run_post_install_commands(self, commands=None):
...
-            # make sure we have a list of commands
-            if not isinstance(commands, (list, tuple)):
-                error_msg = f"Invalid value for 'postinstallcmds', should be list or tuple of strings: {commands}"
-                raise EasyBuildError(error_msg)
-
-            for cmd in commands:
-                if not isinstance(cmd, str):
-                    raise EasyBuildError(f"Invalid element in 'postinstallcmds', not a string: {cmd}")
-                run_shell_cmd(cmd)
+        self._run_command_stack(commands, "postinstallcmds")
```
boegel, line 3049:
```suggestion
        self._run_cmds(commands, "run_post_install_commands")

File: easybuild/framework/easyconfig/easyconfig.py [hunk 6]
```diff
@@ -1858,83 +1860,72 @@ def det_installversion(version, toolchain_name, toolchain_version, prefix, suffi
     _log.nosupport('Use det_full_ec_version from easybuild.tools.module_generator instead of %s' % old_fn, '2.0')
 
 
-def get_easyblock_class(easyblock, name=None, error_on_failed_import=True, error_on_missing_easyblock=True, **kwargs):
```
boegel, line 1861: Don't remove `error_on_failed_import`, that's API breakage.

At best deprecate it

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1858,83 +1860,72 @@ def det_installversion(version, toolchain_name, toolchain_version, prefix, suffi
...
+        # and not because of a broken import statement in the easyblock module
+        modname = modulepath.replace('easybuild.easyblocks.', '')
+        error_re = re.compile(rf"No module named '?.*/?{modname}'?")
+        _log.debug(f"Error regexp for ImportError on '{modname}' easyblock: {error_re.pattern}")
+        if error_re.match(str(err)):
+            if error_on_missing_easyblock:
+                raise EasyBuildError(f"Software-specific easyblock '{class_name}' not found for {name}")
+            else:
+                _log.debug(f"Easyblock for {class_name} not found, but ignoring it: {err}")
+        else:
```
boegel, line 1917: keep the `elif error_on_failed_import` here?

boegel (changes requested):

easybuild/framework/easyblock.py
lexming: good point, I suggest to not error out and just ignore `runtest` if both are defined (with a warning ofc). Fixed in https://github.com/easybuilders/easybuild-framework/pull/4531/commits/22d96485b45bcf7abf5e337cfb524cc9a3453d2c

easybuild/framework/easyblock.py
lexming: good idea, it is indeed easy to construct the commands from just the step name. Done in https://github.com/easybuilders/easybuild-framework/pull/4531/commits/e05a9850eca3dac2c6a7f3e0b47e08f3da870657

easybuild/framework/easyconfig/easyconfig.py
lexming: Well, we can actually break the API between major versions :wink: I re-added it with a deprecation warning nonetheless. Done in https://github.com/easybuilders/easybuild-framework/pull/4531/commits/eacd6e3b65f38c2b0505a2e5f873ad1eae72cc1a

easybuild/framework/easyconfig/easyconfig.py
lexming: Done in https://github.com/easybuilders/easybuild-framework/pull/4531/commits/eacd6e3b65f38c2b0505a2e5f873ad1eae72cc1a

lexming:
Test report with last commit and experimental easyconfig for CORSIKA in https://github.com/easybuilders/easybuild-framework/pull/4531#issuecomment-2163064181
https://gist.github.com/lexming/13af2faf00baa7f34b6923620b2015fe

```
== 2024-07-03 09:18:52,337 easyconfig.py:1906 INFO Using generic EasyBlock module for software 'CORSIKA'
== 2024-07-03 09:18:52,338 easyconfig.py:1935 INFO Successfully obtained class 'EasyBlock' for easyblock 'None' (software name 'CORSIKA')
== 2024-07-03 09:18:52,338 easyconfig.py:688 INFO Parsing easyconfig file /rhea/scratch/brussel/101/vsc10122/easybuild/easybuild-easyconfigs/easybuild/easyconfigs/CORSIKA-77550-foss-2023a.eb with rawcontent: name = 'CORSIKA'
```

easybuild/framework/easyblock.py
```diff
@@ -2720,20 +2720,72 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
...
+            "configure": ('pre_configure_cmds', 'preconfigopts', 'configure_cmd', 'configopts'),
+            "build": ('pre_build_cmds', 'prebuildopts', 'build_cmd', 'buildopts'),
+            "test": ('pre_test_cmds', 'pretestopts', 'test_cmd', 'testopts'),
+            "install": ('pre_install_cmds', 'preinstallopts', 'install_cmd', 'installopts'),
+        }
+
+        pre_step_cmds = self.cfg[step_cmd_cfgs[step][0]]
+        if pre_step_cmds:
+            try:
+                self._run_cmds(pre_step_cmds)
```
boegel, line 2757: Don't we limit the usefulness of `pre_configure_cmds` by running those commands separately?

Something like `export FOO=bar` in `pre_configure_cmds` won't work as expected because it's run in a different shell than `configure_cmd`, which limits the usefulness of `pre_configure_cmds` quite a bit I think?

I can see the value though (better errors when one of the commands in `pre_configure_cmds` fails, for example), but we should try and make it very clear that these commands are running in a different subshell?

easybuild/framework/easyblock.py
```diff
@@ -2720,20 +2720,72 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
...
-        unit_test_cmd = self.cfg['runtest']
-        if unit_test_cmd:
-            self.log.debug(f"Trying to execute {unit_test_cmd} as a command for running unit tests...")
-            res = run_shell_cmd(unit_test_cmd)
+        if self.cfg['runtest'] is not None:
+            self.log.deprecated("Easyconfig parameter 'runtest' is deprecated, use 'test_cmd' instead.", "6.0")
+            if self.cfg['test_cmd'] is None:
+                self.cfg['test_cmd'] = self.cfg['runtest']
+            else:
+                self.log.warning(
```
boegel, line 2783: Shouldn't we print a warning here (too), to avoid that this is overlooked easily?

easybuild/framework/easyconfig/default.py
```diff
@@ -88,9 +88,11 @@
                                       "to be linked in any installed binary/library", BUILD],
     'bitbucket_account': ['%(namelower)s', "Bitbucket account name to be used to resolve template values in source"
                                            " URLs", BUILD],
-    'buildopts': ['', 'Extra options passed to make step (default already has -j X)', BUILD],
+    'buildopts': ['', 'Extra options appended to build command', BUILD],
+    'build_cmd': [None, "Main shell command used in the build step", BUILD],
```
boegel, line 92: This should also mention something like "`(only used by default build step implemented by EasyBlock class)`", or something like that?
As soon as a generic/custom easyblock is used, `build_cmd` will be totally ignored.
Same goes for `configure`, `test`, `install`

easybuild/framework/easyconfig/default.py
```diff
@@ -107,18 +109,23 @@
...
                         'cores, active cpuset and restrictions in ulimit)'), BUILD],
     'patches': [[], "List of patches to apply", BUILD],
-    'prebuildopts': ['', 'Extra options pre-passed to build command.', BUILD],
-    'preconfigopts': ['', 'Extra options pre-passed to configure.', BUILD],
-    'preinstallopts': ['', 'Extra prefix options for installation.', BUILD],
-    'pretestopts': ['', 'Extra prefix options for test.', BUILD],
-    'postinstallcmds': [[], 'Commands to run after the install step.', BUILD],
-    'postinstallpatches': [[], 'Patch files to apply after running the install step.', BUILD],
-    'postinstallmsgs': [[], 'Messages to print after running the install step.', BUILD],
+    'pre_build_cmds': ['', 'Extra commands executed before main build command', BUILD],
```
boegel, line 118: Mention that these are run in a different subshell than the build command itself (no sharing of environment), and only when there's no generic/custom easyblock is used

Same for `configure`, `test`, `install`

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1858,83 +1860,81 @@ def det_installversion(version, toolchain_name, toolchain_version, prefix, suffi
...
+        # when an ImportError occurs, make sure that it's caused by not finding the easyblock module,
+        # and not because of a broken import statement in the easyblock module
+        modname = modulepath.replace('easybuild.easyblocks.', '')
+        error_re = re.compile(rf"No module named '?.*/?{modname}'?")
+        _log.debug(f"Error regexp for ImportError on '{modname}' easyblock: {error_re.pattern}")
+        if error_re.match(str(err)):
+            if error_on_missing_easyblock:
+                raise EasyBuildError(f"Software-specific easyblock '{class_name}' not found for {name}")
+            elif error_on_failed_import:
+                raise EasyBuildError(f"Failed to import '{class_name}' easyblock: {err}")
```
boegel, line 1923: This doesn't belong here, in this case the easyblock doesn't exist?

boegel (changes requested):
@lexming Ideally we also have a dedicated for this new feature, using the `toy` software we use in the tests for example, with an easyconfig where we define `configure_cmd` & co and we make sure that `EasyBlock` class is used

boegel:
@lexming I changed to target branch in this PR from 5.0.x to develop, you should synchronize your PR branch with current develop branch (which has received a massive update after the release of EasyBuild v5.0.0, see https://github.com/easybuilders/easybuild-framework/pull/4825)

PR #4516, opened 2024-04-25
don't allow unresolved templates in easyconfig parameters by default + add support for `--allow-unresolved-templates` configuration option

Flamefire:
A failure in template resolving is currently hidden in the log although
it is likely due to a bug in the code (either EasyBuild, EasyBlock or EasyConfig)
This may lead to cryptic error messages later on or misbehave silently
which is even worse.

PR #4532, opened 2024-05-14
Fix resolved (template) values in case of failure 

Flamefire:
When the value cannot be resolved usually the original value is returned.
However if it contains `%` signs there will be an escape step and that escaped value is returned.

That makes it impossible to have values which can only later be resolved, like
`cd %(startdir)s` in extensions which would become
`cd %%(startdir)s` on the first resolve-attempt.

This is impossible to "undo" in the current framework hence we must take care of this case.

I also did some related changes:
- pylint detected some cases where `dict.items` should be used
- The escaping/unescaping of template values as documented in `resolve_template` was not tested


boegel:
@Flamefire The case from https://github.com/easybuilders/easybuild-easyconfigs/pull/20119 that happens to work now should also be taken into account.
Although that happens to work by accident almost, can we somehow prevent from breaking that use case where `%(builddir)s` is being escaped to avoid trouble with the `%s` that accompanies it?

Flamefire:
> can we somehow prevent from breaking that use case where %(builddir)s is being escaped to avoid trouble with the %s that accompanies it?

I'm not sure we can and don't think we should. https://github.com/easybuilders/easybuild-framework/pull/4516 might also have avoided the issue we had as it avoids resolving where we can and allows failures at places where they might happen. Haven't checked but I guess there it won't try to resolve the extension values when reading them. And with this change it won't matter anymore how often EB tries to resolve templates until it succeeds: The value stays unchanged.

Additionally we have:
https://github.com/easybuilders/easybuild-framework/blob/ae2958838bfde991424a687782a78fa7f7703775/easybuild/framework/easyconfig/easyconfig.py#L2020

According to this no-one should expect that workaround to work, should they? And the idea was explicitly that things like `%(builddir)s %s` work, i.e. the `%s` is not tried to be resolved by EB during template steps, see the test for the single `%s` (which works now too): https://github.com/easybuilders/easybuild-framework/pull/4532/files#diff-bf1c3722aa26c6569dd59d4e4a92f16c88bf688276708beb9f09df6e40288713R3662

PR #4533, opened 2024-05-14
Use dict.items() instead of repeatedly getting the value

Flamefire:
Simple refactoring making the code faster and shorter

Targetting 5.0x because Python 2 creates a copy not a view


boegel (approved):
lgtm

PR #4537, opened 2024-05-20
add support for easyconfig parameter `module_only`

smoors:
the motivation is to support installing only the module for certain deps of a software when installing with `eb --robot`
the `module_only` parameter can then be set in a hook depending on a certain condition
use case: installing only the module for CUDA software on non-GPU nodes (e.g. the login nodes), thus making the CUDA modules visible (but not actually usable)

@boegel this helps getting the VSC-bot working in hydra by replacing the 'dummy modules' feature in our submit_build script with a hook

### DIFF
--- a/test/framework/options.py
+++ b/test/framework/options.py
@@ -366,15 +366,24 @@ class CommandLineOptionsTest(EnhancedTestCase):
 
         self.assertEqual(len(glob.glob(toy_mod_glob)), 1)
 
-        for toy_mod in glob.glob(toy_mod_glob):
-            remove_file(toy_mod)
-
         # check use of module_only parameter
+        remove_dir(os.path.join(self.test_installpath, 'modules', 'all', 'toy'))
+        remove_dir(os.path.join(self.test_installpath, 'software', 'toy', '0.0'))
+        args = [
+            test_ec,
+            '--rebuild',
+        ]
         test_ec_txt += "\nmodule_only = True\n"
         write_file(test_ec, test_ec_txt)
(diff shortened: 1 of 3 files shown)


boegel (changes requested):
changes in `easybuild/*` look OK to me, but we need a test for this...

smoors:
> changes in `easybuild/*` look OK to me, but we need a test for this...

test added in https://github.com/easybuilders/easybuild-framework/pull/4537/commits/64cde9e811ceca718fef21d58c11163abdfc81dd

test/framework/options.py
```diff
@@ -366,6 +366,16 @@ def test_skip(self):
 
         self.assertEqual(len(glob.glob(toy_mod_glob)), 1)
 
+        for toy_mod in glob.glob(toy_mod_glob):
+            remove_file(toy_mod)
+
+        # check use of module_only parameter
+        test_ec_txt += "\nmodule_only = True\n"
+        write_file(test_ec, test_ec_txt)
+        self.eb_main(args, do_build=True, raise_error=True)
```
boegel, line 375: @smoors You should tweak `args` here too, so it doesn't include `--skip` anymore (or `--force`, I guess), and also check that no software was installed?

smoors: done in https://github.com/easybuilders/easybuild-framework/pull/4537/commits/430189c2dfd39b5ea494b66b06ab6acfa347851d

boegel (approved):
lgtm

PR #4539, opened 2024-05-22
prepend to `$PYTHONPATH` or `$EBPYTHONPREFIXES` in generated module files by automatically scanning for python site package directories

Micket:
Alternative approach, don't attempt to rewrite paths, instead generate them directly. 

Will of course require some easyblocks (primarily pythonpackage and pythonbundle) to adapt by using these new options, though they should be trivial.

Note: very much a draft right now. Completely untested and probably forgot something important.

easyblock PR: https://github.com/easybuilders/easybuild-easyblocks/pull/3343
easyconfig PR: https://github.com/easybuilders/easybuild-easyconfigs/pull/20960

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1440,8 +1440,9 @@ class EasyBlock(object):
 
         # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present and python paths exist
         if not os.path.isfile(os.path.join(self.installdir, 'bin/python')):  # only needed when not a python install
-            python_paths = [path for path in glob.glob('lib/python*/site-packages', root_dir=self.installdir)
-                            if re.match(r'lib/python\d+\.\d+/site-packages', path)]
+            candidate_paths = (os.path.relpath(path, self.installdir)
+                               for path in glob.glob(f'{self.installdir}/lib/python*/site-packages'))
+            python_paths = [path for path in candidate_paths if re.match(r'lib/python\d+\.\d+/site-packages', path)]
 
             runtime_deps = [dep['name'] for dep in self.cfg.dependencies(runtime_only=True)]
             use_ebpythonprefixes = 'Python' in runtime_deps and \
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1438,6 +1438,22 @@ class EasyBlock(object):
                                      value, type(value))
             lines.append(self.module_generator.append_paths(key, value, allow_abs=self.cfg['allow_append_abs_path']))
 
+        # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present
+        if 'PYTHONPATH' not in self.module_generator.added_paths_per_key and \
+           'EBPYTHONPREFIXES' not in self.module_generator.added_paths_per_key:
+            python_paths = [path for path in glob.glob('lib*/python*/site-packages')
+                            if re.match(r'lib(64)?/python\d+\.\d+/site-packages', path)]
+            use_ebpythonprefixes = get_software_root('Python') and build_option('prefer_ebpythonprefixes') and \
+                self.cfg['prefer_ebpythonprefixes']
+
+            if len(python_paths) > 1 and not use_ebpythonprefixes:
+                raise EasyBuildError('Multiple python paths requires EBPYHONPREFIXES: ' + ', '.join(python_paths))
+            elif python_paths:
+                if use_ebpythonprefixes:
+                    lines.append(self.module_generator.append_paths('EBPYHONPREFIXES', '.'))
+                else:
+                    lines.append(self.module_generator.append_paths('PYTHONPATH', python_paths))
+
         modloadmsg = self.cfg['modloadmsg']
         if modloadmsg:
             # add trailing newline to prevent that shell prompt is 'glued' to module load message
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1441,8 +1441,9 @@ class EasyBlock(object):
         # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present
         if 'PYTHONPATH' not in self.module_generator.added_paths_per_key and \
            'EBPYTHONPREFIXES' not in self.module_generator.added_paths_per_key:
-            python_paths = [path for path in glob.glob('lib*/python*/site-packages')
-                            if re.match(r'lib(64)?/python\d+\.\d+/site-packages', path)]
+            python_paths = [path[len(self.installdir)+1:]
+                            for path in glob.glob(f'{self.installdir}/lib*/python*/site-packages')
+                            if re.match(self.installdir + r'/lib(64)?/python\d+\.\d+/site-packages', path)]
             use_ebpythonprefixes = get_software_root('Python') and build_option('prefer_ebpythonprefixes') and \
                 self.cfg['prefer_ebpythonprefixes']
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1440,7 +1440,7 @@ class EasyBlock(object):
 
         # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present and python paths exist
         if not os.path.isfile(f'{self.installdir}/bin/python'):  # only needed when not a base python installation
-            python_paths = [path[len(self.installdir)+1:]
+            python_paths = [os.path.relpath(path, self.installdir)
                             for path in glob.glob(f'{self.installdir}/lib/python*/site-packages')
                             if re.match(self.installdir + r'/lib/python\d+\.\d+/site-packages', path)]
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1439,21 +1439,20 @@ class EasyBlock(object):
             lines.append(self.module_generator.append_paths(key, value, allow_abs=self.cfg['allow_append_abs_path']))
 
         # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present
-        if 'PYTHONPATH' not in self.module_generator.added_paths_per_key and \
-           'EBPYTHONPREFIXES' not in self.module_generator.added_paths_per_key:
-            python_paths = [path[len(self.installdir)+1:]
-                            for path in glob.glob(f'{self.installdir}/lib*/python*/site-packages')
-                            if re.match(self.installdir + r'/lib(64)?/python\d+\.\d+/site-packages', path)]
-            use_ebpythonprefixes = get_software_root('Python') and build_option('prefer_ebpythonprefixes') and \
-                self.cfg['prefer_ebpythonprefixes']
+        python_paths = [path[len(self.installdir)+1:]
+                        for path in glob.glob(f'{self.installdir}/lib*/python*/site-packages')
+                        if re.match(self.installdir + r'/lib(64)?/python\d+\.\d+/site-packages', path)]
+        use_ebpythonprefixes = get_software_root('Python') and build_option('prefer_ebpythonprefixes') and \
+            self.cfg['prefer_ebpythonprefixes']
 
-            if len(python_paths) > 1 and not use_ebpythonprefixes:
-                raise EasyBuildError('Multiple python paths requires EBPYTHONPREFIXES: ' + ', '.join(python_paths))
-            elif python_paths:
-                if use_ebpythonprefixes:
-                    lines.append(self.module_generator.prepend_paths('EBPYTHONPREFIXES', ''))
-                else:
-                    lines.append(self.module_generator.prepend_paths('PYTHONPATH', python_paths))
+        if len(python_paths) > 1 and not use_ebpythonprefixes:
+            raise EasyBuildError('Multiple python paths requires EBPYTHONPREFIXES: ' + ', '.join(python_paths))
+        elif python_paths:
+            # Add paths unless they were already added
+            if use_ebpythonprefixes and '' not in self.module_generator.added_paths_per_key['EBPYTHONPREFIXES']:
+                lines.append(self.module_generator.prepend_paths('EBPYTHONPREFIXES', ''))
+            elif python_paths[0] not in self.module_generator.added_paths_per_key['PYTHONPATH']:
+                lines.append(self.module_generator.prepend_paths('PYTHONPATH', python_paths))
 
         modloadmsg = self.cfg['modloadmsg']
         if modloadmsg:
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1439,7 +1439,7 @@ class EasyBlock(object):
             lines.append(self.module_generator.append_paths(key, value, allow_abs=self.cfg['allow_append_abs_path']))
 
         # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present and python paths exist
-        if self.name not in ['Python', 'Miniconda', 'Anaconda']:  # nothing extra needed for base python installations
+        if not os.path.isfile(f'{self.installdir}/bin/python'):  # only needed when not a base python installation
             # Exclude symlinks lib -> lib64 or vice verse to avoid duplicates
             python_paths = []
             if not os.path.islink(f'{self.installdir}/lib'):
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1440,16 +1440,9 @@ class EasyBlock(object):
 
         # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present and python paths exist
         if not os.path.isfile(f'{self.installdir}/bin/python'):  # only needed when not a base python installation
-            # Exclude symlinks lib -> lib64 or vice verse to avoid duplicates
-            python_paths = []
-            if not os.path.islink(f'{self.installdir}/lib'):
-                python_paths += [path[len(self.installdir)+1:]
-                                 for path in glob.glob(f'{self.installdir}/lib/python*/site-packages')
-                                 if re.match(self.installdir + r'/lib/python\d+\.\d+/site-packages', path)]
-            if not os.path.islink(f'{self.installdir}/lib64'):
-                python_paths += [path[len(self.installdir)+1:]
-                                 for path in glob.glob(f'{self.installdir}/lib64/python*/site-packages')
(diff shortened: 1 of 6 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1423,6 +1423,22 @@ def make_module_extra(self, altroot=None, altversion=None):
...
+            use_ebpythonprefixes = get_software_root('Python') and build_option('prefer_ebpythonprefixes') and \
+                self.cfg['prefer_ebpythonprefixes']
+
+            if len(python_paths) > 1 and not use_ebpythonprefixes:
+                raise EasyBuildError('Multiple python paths requires EBPYHONPREFIXES: ' + ', '.join(python_paths))
+            elif python_paths:
+                if use_ebpythonprefixes:
+                    lines.append(self.module_generator.append_paths('EBPYHONPREFIXES', '.'))
+                else:
+                    lines.append(self.module_generator.append_paths('PYTHONPATH', python_paths))
```
boegel, line 1440: @Micket Should be `prepend_paths`, not `append_paths`?

Micket: indeed it should

easybuild/framework/easyblock.py
```diff
@@ -1421,6 +1421,32 @@ def make_module_extra(self, altroot=None, altversion=None):
...
+                python_paths += [path[len(self.installdir)+1:]
+                                 for path in glob.glob(f'{self.installdir}/lib64/python*/site-packages')
+                                 if re.match(self.installdir + r'/lib64/python\d+\.\d+/site-packages', path)]
+
+            use_ebpythonprefixes = 'Python' in self.cfg.dependencies(runtime_only=True) and \
+                build_option('prefer_ebpythonprefixes') and self.cfg['prefer_ebpythonprefixes']
+
+            if len(python_paths) > 1 and not use_ebpythonprefixes:
+                raise EasyBuildError('Multiple python paths requires EBPYTHONPREFIXES: ' + ', '.join(python_paths))
+            elif python_paths:
```
Micket, line 1442:
```suggestion
            if python_paths:
```

This is (as far as I have found, having tested just about every type of easyconfig i could find), which has an actual issue with the code right now

EasyBuild itself generates both:
```
/apps/Test2/software/EasyBuild/4.9.2/lib/python3.6/site-packages/
/apps/Test2/software/EasyBuild/4.9.2/lib64/python3.6/site-packages/
```
while the latter is empty, my glob'ing still picks it up, and having multiple PYTHONPATH's doesn't make any sense, this errors out. 
There is technically nothing wrong here, those paths are both compatible with python3.6, so they would be fine to use.

I can just remove that extra check. Simplifies the code and you simply get what you ask for. If you have multiple paths for some reason, they are both added, no problem. This is just an extra check we added that were never cared about before, and it's in no way a requirement for this PR.

On the other hand, it's literally *only* EasyBuild that does this, and I think it shouldn't duplicate the installation directories. Perhaps it should be creating a lib <-> lib64 symlink like just about everything else?

I could also make the code more robust, checking if it's really the same python version etc...

Micket: I removed this check for now, to keep this as simple as possible. This check could be introduced later if we want it (but we never checked for this in the past, so this change really has nothing to do with EBPYTHONPREFIXES, especially considering it actually only is problem when EBPYTHONPREFIXES *isn't* used)

Micket:
I've made some simplifications:

1. I just drop that extra check for python paths. We have never checked this before, it's not a requirement for this PR, and in practice it never has any real errors in anything we currently have (and it would be *very* strange if a real issue ever appeared).

2. Secondly, i dropped lib64 variant from the globbed paths, since this was never considered by our multi  python dep code (sitecustomize.py) to begin with, so I will not consider it here either for consistency. 

3. I changed the EBPYTHONPREFIXES default to false. While I think everyone really ought to use this due to the problems caused by using PYTHONPATH, there is a problem enabling this by default for existing installations: if your dependencies use PYTHONPATH, you can't shadow any libraries as PYTHONPATH will take precedence. Fixing this requires you rebuild the old easyconfigs (module-only) with `--prefer-ebpyhonprefixes`, while this is easy, i suspect many would have issues with this.
We shadow libraries like this regularly, example https://github.com/easybuilders/easybuild-easyconfigs/blob/187ba72edb9f3c71203b1bf02a2243cfd8d203af/easybuild/easyconfigs/t/TensorFlow/TensorFlow-2.9.1-foss-2022a.eb#L130
so if you had build scipy bundle before, and not try to build that tensorflow, you would pick up the wrong version of `gast` (and fail the sanity check)

Flamefire:
Compared to #4496 it looks like this doesn't handle the case where easyblocks (and also easyconfigs?) add `$PYTHONPATH` directly, does it?

Micket:
Correct, I don't override those if people have their own custom easyblocks or easyconfigs where they keep using PYTHONPATH  (my accompanying PRs remove it from easybuild-easyblocks and easybuild-easyconfigs)

Flamefire:
Is that on purpose? Because to me this looks rather accidental. I already see people using (slightly) older EasyConfigs that still have that path set.

lexming (commented):
The functionality is working well:
```bash
$ ml add SciPy-bundle/2023.07-gfbf-2023a
$ python -V
Python 3.11.3
$ which python
~/easybuild/install/software/Python/3.11.3-GCCcore-12.3.0/bin/python
$ python -m virtualenv ~/.local/venv/numpy
created virtual environment CPython3.11.3.final.0-64 in 178m
$ source ~/.local/venv/numpy/bin/activate
$ [venv] python -V
Python 3.11.3
$ [venv] pip freeze | grep numpy
numpy @ file:///home/lexming/.local/easybuild/build/SciPybundle/2023.07/gfbf-2023a/numpy/numpy-1.25.1
$ [venv] python -m pip install numpy==2.1.1
Collecting numpy==2.1.1
  Using cached numpy-2.1.1-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (16.3 MB)
Installing collected packages: numpy
  Attempting uninstall: numpy
    Found existing installation: numpy 1.25.1
    Not uninstalling numpy at /home/lexming/.local/easybuild/software/SciPy-bundle/2023.07-gfbf-2023a/lib/python3.11/site-packages, outside environment /home/lexming/.local/venv/numpy
    Can't uninstall 'numpy'. No files were found to uninstall.
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
scipy 1.11.1 requires numpy<1.28.0,>=1.21.6, but you have numpy 2.1.1 which is incompatible.
Successfully installed numpy-2.1.1
[...]
$ [venv] pip freeze | grep numpy
numpy==2.1.1
```

easybuild/tools/options.py
```diff
@@ -482,6 +482,8 @@ def override_options(self):
                                             None, 'store_true', False),
             'pre-create-installdir': ("Create installation directory before submitting build jobs",
                                       None, 'store_true', True),
+            'prefer-ebpythonprefixes': (("Prefer EBPYTHONPREFIXES over PYTHONPATH when possible"),
```
lexming, line 485: The name of this option is hard to understand, only people with previous knowledge of what `ebpythonprefixes` are will know what this does. I suggest to change it to less technical one.

There are 2 approaches, based on the perspective we pick:
1. Shall Python module have precedence over user installations?
2. Can the user override Python modules installed by EB?

Which could translate to the following options:
1. `--python-module-precedence=[high,low]`
2. `--user-override-python=[true,false]

easybuild/framework/easyblock.py
```diff
@@ -1428,6 +1428,26 @@ def make_module_extra(self, altroot=None, altversion=None):
...
+                            if re.match(self.installdir + r'/lib/python\d+\.\d+/site-packages', path)]
+
+            runtime_deps = [dep['name'] for dep in self.cfg.dependencies(runtime_only=True)]
+            use_ebpythonprefixes = 'Python' in runtime_deps and \
+                build_option('prefer_ebpythonprefixes') and self.cfg['prefer_ebpythonprefixes']
+
+            if python_paths:
+                # Add paths unless they were already added
+                if use_ebpythonprefixes:
+                    if '' not in self.module_generator.added_paths_per_key['EBPYTHONPREFIXES']:
```
boegel, line 1444: maybe worth a comment here why `''` makes sense (since it's relative to the installation prefix)

or introduce a local variable to help clarify what's going on here

easybuild/framework/easyconfig/default.py
```diff
@@ -115,6 +115,7 @@
     'patches': [[], "List of patches to apply", BUILD],
     'prebuildopts': ['', 'Extra options pre-passed to build command.', BUILD],
     'preconfigopts': ['', 'Extra options pre-passed to configure.', BUILD],
+    'prefer_ebpythonprefixes': [True, "Prefer EBPYTHONPREFIXES over PYTHONPATH when possible.", BUILD],
```
lexming, line 118: This option should have a different name than its counterpart in `tools.options` because it is set in the oposite way. Following the discussion with @Micket , it should be called `--force-<other_option>`, as it can override that option.

easybuild/framework/easyblock.py
```diff
@@ -1428,6 +1428,27 @@ def make_module_extra(self, altroot=None, altversion=None):
                                      value, type(value))
             lines.append(self.module_generator.append_paths(key, value, allow_abs=self.cfg['allow_append_abs_path']))
 
+        # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present and python paths exist
+        if not os.path.isfile(f'{self.installdir}/bin/python'):  # only needed when not a base python installation
+            python_paths = [path[len(self.installdir)+1:]
```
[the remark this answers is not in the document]
boegel: Oh, wait, we're cutting off the first part here, so my question boils down to making sure that we don't strip off `../l`, and end up with a non-sensical entry like `ib/python3.11/site-packages`...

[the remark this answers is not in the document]
Micket: fixed

easybuild/framework/easyblock.py
```diff
@@ -1428,6 +1428,27 @@ def make_module_extra(self, altroot=None, altversion=None):
                                      value, type(value))
             lines.append(self.module_generator.append_paths(key, value, allow_abs=self.cfg['allow_append_abs_path']))
 
+        # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present and python paths exist
+        if not os.path.isfile(f'{self.installdir}/bin/python'):  # only needed when not a base python installation
+            python_paths = [os.path.relpath(path, self.installdir)
+                            for path in glob.glob(f'{self.installdir}/lib/python*/site-packages')
+                            if re.match(self.installdir + r'/lib/python\d+\.\d+/site-packages', path)]
```
Flamefire, line 1435: I'd be careful with using user-supplied input in a regex. It could happen to be an invalid regex. Maybe:
```suggestion
                            if re.search(r'/lib/python\d+\.\d+/site-packages$', path)]
```

Micket: Hm, i think this would trip up on a few corner cases though, like i know fsl package does this thing
```
                      extra fsl subdir here
                          | 
                          v 
/apps/software/fsl/1.2.3/fsl/lib/python3.11/site-packages/
```

I could do 
```suggestion
                            if re.match(r'lib/python\d+\.\d+/site-packages$', os.path.relpath(self.installdir, path))]
```

Flamefire: For `/apps/software/fsl/1.2.3/fsl/lib/python3.11/site-packages/` the glob `{self.installdir}/lib/python*/site-packages'` won't match will it? So really all we need to do in the regexp is to verify the `python*` folder name which we can't do in the glob directly.

But sure this would also work. You can omit the trailing $ when using `match`. Maybe even call relpath only once:
```
[relpath
 for relpath in (os.path.relpath(path, self.installdir) for path in glob.glob(f'{self.installdir}/lib/python*/site-packages'))
 if re.match(r'/lib/python\d+\.\d+/site-packages', relpath)]
```

easybuild/framework/easyblock.py
```diff
@@ -1438,6 +1438,26 @@ def make_module_extra(self, altroot=None, altversion=None):
                                      value, type(value))
             lines.append(self.module_generator.append_paths(key, value, allow_abs=self.cfg['allow_append_abs_path']))
 
+        # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present and python paths exist
+        if not os.path.isfile(os.path.join(self.installdir, 'bin/python')):  # only needed when not a python install
+            python_paths = [path for path in glob.glob('lib/python*/site-packages', root_dir=self.installdir)
+                            if re.match(r'lib/python\d+\.\d+/site-packages', path]
```
Flamefire, line 1444:
```suggestion
                            if re.match(r'lib/python\d+\.\d+/site-packages', path)]
```

Micket: Thanks, I was really tired :) I made a force push to not mess add so much noise in the git history

easybuild/framework/easyblock.py
```diff
@@ -1438,6 +1438,26 @@ def make_module_extra(self, altroot=None, altversion=None):
                                      value, type(value))
             lines.append(self.module_generator.append_paths(key, value, allow_abs=self.cfg['allow_append_abs_path']))
 
+        # Add automatic PYTHONPATH or EBPYTHONPREFIXES if they aren't already present and python paths exist
+        if not os.path.isfile(os.path.join(self.installdir, 'bin/python')):  # only needed when not a python install
+            python_paths = [path for path in glob.glob('lib/python*/site-packages', root_dir=self.installdir)
```
Flamefire, line 1443: `root_dir` is Python 3.10: https://docs.python.org/3/library/glob.html#glob.glob

easybuild/framework/easyblock.py
```diff
@@ -1438,6 +1438,27 @@ def make_module_extra(self, altroot=None, altversion=None):
...
+                build_option('prefer_python_search_path') == EBPYTHONPREFIXES and not self.cfg['force_pythonpath']
+
+            if python_paths:
+                # Add paths unless they were already added
+                if use_ebpythonprefixes:
+                    path = ''  # EBPYTHONPREFIXES are relative to the install dir
+                    if path not in self.module_generator.added_paths_per_key[EBPYTHONPREFIXES]:
+                        lines.append(self.module_generator.prepend_paths(EBPYTHONPREFIXES, path))
+                else:
+                    for python_path in python_paths:
```
Flamefire, line 1458: I'm wondering: There can/should never be more than one here, can there? It would mean we'd have e.g. `lib/python3.8` and `lib/python2.6` and will add both to $PYTHONPATH which can't be right

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -1105,15 +1105,19 @@ def filter_deps(self, deps):
 
         return retained_deps
 
-    def dependencies(self, build_only=False):
+    def dependencies(self, build_only=False, runtime_only=False):
```
Flamefire, line 1108: Can't we have either an "enum" `dep_type=BUILD|RUNTIME|BOTH` or at least use `include_build=True, include_runtime=True` because now we have a potentially "invalid" call: When both are True

Maybe we should at least rename to `exclude_runtime, exclude_build` which would be mostly backwards compatible and setting both to True seems more reasonable to return nothing

easybuild/framework/easyblock.py
Micket: A bit of history, went through a few iterations here, at some points i had lib/ and lib64/, which if not symlinked, could be both considered without being an error (and there was a few cases like that which falsely triggered an error), so I just got tired of dealing with that as well so I opted skip that error checking for now.

Since then, I've changed to only consider lib/ since that's the only thing EBPYTHONPREFIXES could handle anyway, so, yes, it's changed slightly. So yes, if someone makes a non PythonPackage/PythonBundle, which also does multi_deps for python, and they set `force_python = True` for some reason, there would be multiple pythonpaths which maybe (possible likely) would break things (though breaking things isn't guaranteed if the actual python modules installed happens to work with both).

easybuild/framework/easyconfig/easyconfig.py
boegel: Let's follow up on that in a separate PR, I want to get this PR merged ASAP.

Maybe open an issue for it, so we can tag it with `5.0` milestone.

easybuild/framework/easyblock.py
boegel: So when using `$EBPYTHONPREFIX`, it's fine if there are multiple ones (there would be when using `multi_deps`), but not when using `$PYTHONPATH`, so we should make the latter an error.

I'll look into that, I'm working on this right now (implementing tests)

easybuild/framework/easyblock.py
```diff
@@ -1390,6 +1390,37 @@ def make_module_description(self):
...
+        if they aren't already present and the standard lib/python*/site-packages subdirectory exists
+        """
+        lines = []
+        if not os.path.isfile(os.path.join(self.installdir, 'bin', 'python')):  # only needed when not a python install
+            python_subdir_pattern = os.path.join(self.installdir, 'lib', 'python*', 'site-packages')
+            candidate_paths = (os.path.relpath(path, self.installdir) for path in glob.glob(python_subdir_pattern))
+            python_paths = [path for path in candidate_paths if re.match(r'lib/python\d+\.\d+/site-packages', path)]
+
+            runtime_deps = [dep['name'] for dep in self.cfg.dependencies(runtime_only=True)]
+            use_ebpythonprefixes = all([
```
boegel, line 1405: We also have to take into account `multi_deps` here, if that's being used we really need to use `$EBPYTHONPREFIXES` instead of `$PYTHONPATH`...

boegel (changes requested):

easybuild/framework/easyblock.py
Micket: Normal scenarios would be
1. Using Python, multiple paths -> use ebpythonprefixes
2. Using Python, preferring EBPYTHONPREFIXES and not forcing pythonpath -> use ebpythonprefixes
3. Not using Python -> use PYTHONPATH
so maybe the logic should be 
```
            use_ebpythonprefixes = 'Python' in runtime_deps and (len(python_paths) > 1 or (build_option('prefer_python_search_path') == EBPYTHONPREFIXES,
                not self.cfg['force_pythonpath']))
```

boegel: I broke that down and added logging for it too in https://github.com/Micket/easybuild-framework/pull/6, and enhanced the test to also check the `multi_deps` case

boegel (approved):
lgtm

lexming (approved):

lexming:
Merging, thanks @Micket !

PR #4544, opened 2024-05-30
trigger pre-hook in `run_shell_cmd` as early as possible so dry-run + trace output is correct

### LINKED ISSUE #4490: `run_shell_cmd` hooks should be applied earlier
The hooks for run_shell_cmd
https://github.com/easybuilders/easybuild-framework/blob/24a0f8a396d1e2b414ac8b1a579e8e0b5f0fe3db/easybuild/tools/run.py#L311-L317
are applied just before we run the subprocess, but after all of logging, dry-run etc. `cmd_str` (before hooks) is used in a bunch of places. 

I would argue it should be done as early as possible, since it would reflect the true command being run.

Micket:
Fixes #4490

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -319,6 +319,17 @@ def run_shell_cmd(cmd, fail_on_error=True, split_stderr=False, stdin=None, env=N
         except FileNotFoundError:
             raise EasyBuildError(CWD_NOTFOUND_ERROR)
 
+    if with_hooks:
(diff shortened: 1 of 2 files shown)


Micket:
I think it's right but it's untested, will do so later

boegel:
@Micket We should try and reproduce the bug in a test for `run_shell_cmd`, which will also confirm that the proposed changes are correct (they look OK to me)

PR #4548, opened 2024-06-05
use newer version of actions/checkout and actions/setup-python, to avoid relying on deprecated node.js 16

boegel:
fix for warning popping up in GitHub Actions:

```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-python@v4. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/
```

### DIFF
--- a/.github/workflows/container_tests.yml
+++ b/.github/workflows/container_tests.yml
@@ -17,10 +17,10 @@ jobs:
         python: [3.7]
       fail-fast: false
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: set up Python
-      uses: actions/setup-python@v4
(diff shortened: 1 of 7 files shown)


boegel:
The tests that use a CentOS 7.9 container are failing, because node.js 20 doesn't work there:
```
 /__e/node20/bin/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /__e/node20/bin/node)
```

So this may have to wait until we no longer need to test in CentOS 7.9 (i.e. when we've collapsed the `5.0.x` branch into `develop` in preparation for releasing EasyBuild v5.0)

PR #4550, opened 2024-06-05
set `usedforsecurity` to `False` when calling `hashlib.md5` with Python >= 3.9

### LINKED ISSUE #4317: Add usedforsecurity=False to usage of md5 for FIPS mode
When trying to install easybuild on FIPS mode system, I get the following message: 

File "/[...]/[...]/eb_tmp/lib/python3.6/site-packages/easybuild/tools/filetools.py", line 129, in <lambda>
    CHECKSUM_TYPE_MD5: lambda p: calc_block_checksum(p, hashlib.md5()),
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS

Apparently hashlib offers a named parameter called usedforsecurity for exactly this case, allowing the use of md5 even on security enabled systems: https://docs.python.org/3/library/hashlib.html

boegel:
fixes #4317

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -123,11 +123,23 @@ CHECKSUM_TYPE_MD5 = 'md5'
 CHECKSUM_TYPE_SHA256 = 'sha256'
 DEFAULT_CHECKSUM = CHECKSUM_TYPE_SHA256
 
+
+def _hashlib_md5():
+    """
+    Wrapper function for hashlib.md5,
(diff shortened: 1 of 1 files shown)


Micket (approved):
lgtm, though the bug shouldn't be triggered anymore unless you actually use md5, which we never do.

boegel:
> lgtm, though the bug shouldn't be triggered anymore unless you actually use md5, which we never do.

We don't in the central easyconfigs, but some people still might do so in their local easyconfigs.

I don't see a reason to hold this back though?

PR #4552, opened 2024-06-07
symlink downloaded repo at specified commit when using `--from-commit` so easyconfigs for dependencies are found

### LINKED ISSUE #4540: easyconfig files for dependencies not found when using `--from-commit`
When using `--from-commit`, the robot doesn't seem to be able to find easyconfig files for dependencies, despite them being available in that same commit.

Example, using EasyBuild v4.9.1 (cfr. https://github.com/EESSI/software-layer/pull/585#issuecomment-2126680129): 

```
$ eb --from-commit 04ccd901a613631b00ccbe504d6d66d6a6c2febb PyTorch-bundle-2.1.2-foss-2023a.eb -D
== Temporary log file in case of crash /tmp/eb-mqsf3w04/easybuild-_o_vpmst.log
ERROR: Missing dependencies: SentencePiece/0.2.0-foss-2023a, SoX/14.4.2-foss-2023a (no easyconfig file or existing module found)
```

Yet, an easyconfig file for `SentencePiece` v0.2.0 that can be used to resolve the dependency when using the `foss/2023a` toolchain`*is* available in that commit, see https://github.com/easybuilders/easybuild-easyconfigs/blob/04ccd901a613631b00ccbe504d6d66d6a6c2febb/easybuild/easyconfigs/s/SentencePiece/SentencePiece-0.2.0-GCC-12.3.0.eb

boegel:
fixes #4540

https://github.com/boegel/easybuild-framework/actions/runs/9416046787/job/25938331232 shows that GitHub tests, which are skipped in context of running test suite for a PR because GitHub token is not available, are (still) passing.

Without this fix, the additional test case added to `test_from_commit` fails, which shows it verifies the bug fix being fixed here.

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -612,6 +612,13 @@ def fetch_files_from_commit(commit, files=None, path=None, github_account=None, 
     if github_repo is None:
         github_repo = GITHUB_EASYCONFIGS_REPO
 
+    if github_repo == GITHUB_EASYCONFIGS_REPO:
+        easybuild_subdir = os.path.join('easybuild', 'easyconfigs')
+    elif github_repo == GITHUB_EASYBLOCKS_REPO:
+        easybuild_subdir = os.path.join('easybuild', 'easyblocks')
+    else:
+        raise EasyBuildError("Unknown repo: %s", github_repo)
+
     if path is None:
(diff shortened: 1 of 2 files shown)


ocaisa (approved):
Regex is my enemy, I guess I need to learn it. LGTM

PR #4553, opened 2024-06-07
don't try to determine repo version when file that contains version doesn't exist

boegel:
fix for a couple of broken GitHub tests in framework test suite, and for using `--from-pr` on a closed PR:
```
$ eb --from-pr 8006 -D
== Temporary log file in case of crash /tmp/eb-64svltcd/easybuild-v6yqv27h.log
WARNING: Using easyconfigs from closed PR #8006

ERROR: Couldn't determine version of PR from /tmp/eb-64svltcd/files_pr8006/setup.py
```


bartoldeman (approved):
Lgtm

PR #4554, opened 2024-06-10
Correctly evaluate result for `--dep-graph`

Flamefire:
The success message was always printed even when dumping the graph failed resulting in no file being created but a message like:

> Error: renderer for .pdf is unavailable
> Wrote dependency graph for 1 easyconfigs to graph.pdf


boegel (approved):
lgtm

PR #4555, opened 2024-06-10
Use "alternative" instead of "alternate"

jfgrimm:
* none of the uses of "alternate" in framework refer to "every other" (alternating), they all use the meaning equivalent to "alternative"
* for clarity, this renames all to "alternative" instead
* also clarifies that e.g. `ALTERNATE_PARAMETERS` pertains to easyconfigs --> `ALTERNATIVE_EASYCONFIG_PARAMETERS`


bedroge (approved):
Looks good to me.

PR #4564, opened 2024-06-19
sync with develop (20240619)

boegel:


boegel:
Hmm, CI fails with:
```
ERROR: Found printed messages in output of test suite
  >> generating module file @ /tmp/eb-ei38s7fv/eb-w9x7c6_8/eb-1kfi86qj/tmpmkhzt30r/modules/all/toy/0.0.lua
```

Need to hunt down the test that produces that message, happens because `--trace` is enabled by default...

Micket (approved):
lgtm

PR #4569, opened 2024-06-28
reuse pre-computed checksums

Flamefire:
We calculate MD5 and SHA256 checksums as part of the log output which might take considerable time.
Instead of recomputing it in `verify_checksum` we can reuse this value.

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -1242,7 +1242,7 @@ def calc_block_checksum(path, algorithm):
     return algorithm.hexdigest()
 
 
-def verify_checksum(path, checksums):
+def verify_checksum(path, checksums, computed_checksums=None):
     """
     Verify checksum of specified file.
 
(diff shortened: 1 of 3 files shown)


easybuild/tools/filetools.py
```diff
@@ -1242,7 +1242,7 @@ def calc_block_checksum(path, algorithm):
     return algorithm.hexdigest()
 
 
-def verify_checksum(path, checksums):
+def verify_checksum(path, checksums, computed_checksums=None):
```
boegel, line 1245: @Flamefire Please document this additional optional argument in the docstring below

boegel (changes requested):
Looks good, tiny request before merging

easybuild/tools/filetools.py
Flamefire: Done. Also updated the `checksums` docstring to make the difference clearer

ocaisa (approved):
LGTM

PR #4571, opened 2024-07-03
fix definition of gmpflf toolchain

TopRichard:
Fixes toolchain dependency error:
```
List of toolchain dependency modules and toolchain definition do not match 
(found 
['GCC/12.3.0', 'MPICH/4.2.1-GCC-12.3.0', 'FlexiBLAS/3.3.1-GCC-12.3.0', 'FFTW/3.3.10-GCC-12.3.0', 'FFTW.MPI/3.3.10-gmpich-2024.06', 'ScaLAPACK/2.2.0-gmpich-2024.06-fb'] 
vs expected 
{'MPICH', 'GCC', 'OpenBLAS', 'ScaLAPACK', 'FFTW'})
```

### DIFF
--- a/easybuild/toolchains/gmpflf.py
+++ b/easybuild/toolchains/gmpflf.py
@@ -36,7 +36,6 @@ from easybuild.toolchains.gmpich import Gmpich
 from easybuild.toolchains.gfbf import Gfbf
 from easybuild.toolchains.fft.fftw import Fftw
 from easybuild.toolchains.linalg.flexiblas import FlexiBLAS
-from easybuild.toolchains.linalg.openblas import OpenBLAS
 from easybuild.toolchains.linalg.scalapack import ScaLAPACK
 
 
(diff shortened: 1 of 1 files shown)


easybuild/toolchains/gmpflf.py
```diff
@@ -45,3 +45,12 @@ class Gmpflf(Gmpich, OpenBLAS, FlexiBLAS,  ScaLAPACK, Fftw):
     """Compiler toolchain with GCC, MPICH, OpenBLAS, ScaLAPACK and FFTW."""
     NAME = 'gmpflf'
     SUBTOOLCHAIN = [Gmpich.NAME, Golf.NAME, Gfbf.NAME]
```
boegel, line 47: @TopRichard The real problem is here, we shouldn't be using `golf` as a subtoolchain at all here.

We do this in `foss` for historical reasons, since for older versions of `foss` there was no FlexiBLAS yet, but here there's no need for it at all.

Just remove `Golf` from subtoolchains, and don't derive the `Gmpflf` class from `OpenBLAS` at all, and you shouldn't need the custom `__init__` at all, that should only be required in `foss` toolchain definition

boegel (changes requested):

easybuild/toolchains/gmpflf.py
```diff
@@ -34,14 +34,13 @@
 """
 from easybuild.toolchains.gmpich import Gmpich
 from easybuild.toolchains.gfbf import Gfbf
-from easybuild.toolchains.golf import Golf
 from easybuild.toolchains.fft.fftw import Fftw
 from easybuild.toolchains.linalg.flexiblas import FlexiBLAS
 from easybuild.toolchains.linalg.openblas import OpenBLAS
```
boegel, line 39: remove unused import

```suggestion
```

boegel (changes requested):

boegel (approved):

PR #4573, opened 2024-07-04
stop using CentOS 7.9 container in CI, no longer works now GitHub Actions is using node.js 20

boegel:
required because of problems with `actions/checkout` action like:
```
Run actions/checkout@v3
/usr/bin/docker exec  9843bae9dc7921cd7e422ca131698a5a550184ebb22dc5829be136c5d40196bd sh -c "cat /etc/*release | grep ^ID"
/__e/node20/bin/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by /__e/node20/bin/node)
```

see also https://github.com/actions/checkout/issues/1809 + https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default

PR #4574, opened 2024-07-04
CI: Allow using Node 16 actions

Flamefire:
Required after https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default

edit (by @boegel): see also https://github.com/actions/checkout/issues/1809

### DIFF
--- a/.github/workflows/end2end.yml
+++ b/.github/workflows/end2end.yml
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
     container:
       image: ghcr.io/easybuilders/${{ matrix.container }}-amd64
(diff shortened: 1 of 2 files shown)


jfgrimm (approved):
lgtm

boegel:
If this works, I definitely prefer it over #4573, but it's probably just postponing the inevitable...

jfgrimm:
> If this works, I definitely prefer it over #4573, but it's probably just postponing the inevitable...

definitely, but that's not that surprising since CentOS 7 is now EOL

PR #4576, opened 2024-07-12
don't use special flags for `strict`, `precise`, `loose`, `veryloose` toolchain options on RISC-V

bedroge:
While trying to build GMP (which has `precise: True` in `toolchainopts`), the configure step failed because `-mno-recip` is not supported on RISC-V, similar to Arm (see https://github.com/easybuilders/easybuild-framework/pull/3425).

@julianmorillo checked with a compiler expert, and he recommended to use `-fno-reciprocal-math`. This is a generic flag and used by default, so without a sensible alternative, I've just removed the flags for these toolchain options. It will then basically ignore the `precise: True` and use the default flags.

### DIFF
--- a/easybuild/toolchains/compiler/gcc.py
+++ b/easybuild/toolchains/compiler/gcc.py
@@ -84,6 +84,8 @@ class Gcc(Compiler):
     if systemtools.get_cpu_family() == systemtools.RISCV:
         COMPILER_UNIQUE_OPTION_MAP['strict'] = []
         COMPILER_UNIQUE_OPTION_MAP['precise'] = []
+        COMPILER_UNIQUE_OPTION_MAP['loose'] = ['fno-math-errno']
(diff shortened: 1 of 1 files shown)


bedroge:
Test report for GMP can be found here (with a modified EB version that's using the changes from this PR):
https://github.com/easybuilders/easybuild-easyconfigs/pull/20633#issuecomment-2225617963

bartoldeman (dismissed):
Did you check loose and veryloose as well? I'm surprised they'd work.

bedroge:
> Did you check loose and veryloose as well? I'm surprised they'd work.

@bartoldeman I hadn't looked into those yet (also because it seems they're not really used?), and just applied the same kind of fix that was also done for aarch64. But you're right, they will not work either, so I can remove them as well.

boegel (approved):

PR #4578, opened 2024-07-17
Fix the checksum type check

Flamefire:
This is #4164 rebased on 5.0x

Except for one conflict (`string_type` -> `str`) it was trivial. So #4164 should still be viable for develop

The None case was missed and due to the unrestricted tuple elem_type it may return valid for actually invalid entries.
So restrict that beeing overly cautious so it may wrongly return invalid.
But in that case the conversion function will be called which can do more elaborate verification.

Add test checking for None in checksums.

Also required for https://github.com/easybuilders/easybuild-framework/issues/4142 as the check now correctly handles a None value in the dict

### DIFF
--- a/easybuild/framework/easyconfig/types.py
+++ b/easybuild/framework/easyconfig/types.py
@@ -613,19 +613,33 @@ SANITY_CHECK_PATHS_DICT = (dict, as_hashable({
 }))
 # checksums is a list of checksums, one entry per file (source/patch)
 # each entry can be:
+# None
 # a single checksum value (string)
 # a single checksum value of a specified type (2-tuple, 1st element is checksum type, 2nd element is checksum)
 # a list of checksums (of different types, perhaps different formats), which should *all* be valid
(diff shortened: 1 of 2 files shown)


Micket (approved):
lgtm

Micket:
Question; this just allows the type-checking step to pass, is this still expected to fail to actually use dict+None (because it does in my test)? I assume i need another  one of your PR that fixes that right?

Flamefire:
> Question; this just allows the type-checking step to pass, is this still expected to fail to actually use dict+None (because it does in my test)? I assume i need another one of your PR that fixes that right?

I'm not sure which error but I guess you need https://github.com/easybuilders/easybuild-framework/pull/4711
This and/or https://github.com/easybuilders/easybuild-framework/pull/4579 fix the parsing of that: If the type check is wrong, `to_checksums` is invoked which is broken too. With this PR `to_checksum` will not be invoked and with #4579 but without this calling `to_checksums` at least wont crash even if it is redundantly called.

PR #4583, opened 2024-07-22
Add cuda_cc_space_sep variant that does not have periods

casparvl:
E.g. '80 90' if cuda compute capabilities is '8.0,9.0'

Can be used in e.g. Cuda-Samples as value to `SMS` argument, see the readme at https://github.com/NVIDIA/cuda-samples

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -365,7 +365,7 @@ def template_constant_dict(config, ignore=None, skip_lower=None, toolchain=None)
 
     # step 6. CUDA compute capabilities
     #         Use the commandline / easybuild config option if given, else use the value from the EC (as a default)
-    cuda_cc = build_option('cuda_cc') or config.get('cuda_cc')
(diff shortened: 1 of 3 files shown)


branfosj (approved):

easybuild/framework/easyconfig/templates.py
```diff
@@ -94,9 +94,11 @@
      "as specify by the --sysroot configuration option"),
     ('mpi_cmd_prefix', "Prefix command for running MPI programs (with default number of ranks)"),
     ('cuda_compute_capabilities', "Comma-separated list of CUDA compute capabilities, as specified via "
-     "--cuda-compute-capabilities configuration option or via cuda_compute_capabilities easyconfig parameter"),
+     "--cuda-compute-capabilities configuration option or via cuda_cc easyconfig parameter"),
```
boegel, line 97: @casparvl Maybe I'm missing something, but there's no `cuda_cc` easyconfig parameter in EasyBuild 4.x (nor in the `5.0.x` branch), so this change is incorrect?

PR #4585, opened 2024-07-23
avoid suggesting that `toolchainopts` are supported for `SYSTEM` compiler, since they're not

### LINKED ISSUE #4584: Auto-generated documentation for `SYSTEM` compiler is incorrect
The docs at https://docs.easybuild.io/version-specific/toolchain-opts/#system claim that `SYSTEM` supports `toolchainopts` but it actually doesn't

ocaisa:
Fixes #4584


ocaisa:
Whether we _should_ support `toolchainopts` is up for debate, but this PR reflects the current status

ocaisa:
Ha, life is never easy, of course the default options are expected in framework...

ocaisa:
This will require a companion PR to `SYSTEM` easyconfigs that try to set `toolchainopts` as this will no longer be allowed

branfosj:
> This will require a companion PR to `SYSTEM` easyconfigs that try to set `toolchainopts` as this will no longer be allowed

See https://github.com/easybuilders/easybuild-easyconfigs/pull/21035 - though that targets the `5.0.x` branch.

ocaisa:
@branfosj I've retargeted this to 5.0 as well

Micket:
Would it not be more appropriate to do this to SystemCompiler instead of SystemToolchain?
Does it need to set COMPILER_UNIQUE_OPTION_MAP to none as well, given that it is none from the start?

I really don't think there is any reason not to support these though, everything in COMPILER_SHARED_OPTS is expected to be shared among all the compilers, so we certainly would assume the system compiler to support them as well.

ocaisa:
This is not trivial to do, as for the system compiler there are specific things happening

https://github.com/easybuilders/easybuild-framework/blob/e913a7847c6f66325e68567dfe941dee1c524e95/easybuild/tools/toolchain/toolchain.py#L829-L856

branfosj:
> @branfosj I've retargeted this to 5.0 as well

Once https://github.com/easybuilders/easybuild-easyconfigs/pull/21035 is approved and merged then I am happy to proceed with this. It makes the processing of easyconfigs behave how the `SYSTEM` toolchain works currently - avoiding unexepected behaviour. We can always revisit making the `SYSTEM` toolchain behave more similarly to other toolchains in the future.

ocaisa:
@Micket I did give this a try this morning but it is complicated, I could not successfully extract the `CFLAGS` and `CXXFLAGS`. I've moved the changes to the parent class and left a comment there about what you would need to do if you wanted to change things

ocaisa:
@branfosj Should be good to go now

boegel (approved):
lgtm

PR #4590, opened 2024-08-01
Add `--skip-sanity-check` option

Flamefire:
For some modules the sanity check command(s) fail when run in the build environment.
This option can be used to still install it and investigate later.

This can currently be done by `eb foo.eb && eb foo.eb --module-only --force` but I'd argue this option is easier to find and use.

My current use case is testing for https://github.com/easybuilders/easybuild-easyblocks/issues/3386 by installing many modules at once and then intending to do `--sanity-check-only` for each of them.   
Currently the installation of the bulk aborts early. With this option I can install all of them in one go.

### DIFF
--- a/test/framework/options.py
+++ b/test/framework/options.py
@@ -444,7 +444,7 @@ class CommandLineOptionsTest(EnhancedTestCase):
         self.assertErrorRegex(EasyBuildError, error_pattern, self.eb_main, args, do_build=True, raise_error=True)
 
     def test_skip_sanity_check(self):
-        """Test ignore failing tests (--ignore-test-failure)."""
+        """Test skipping of sanity check step (--skip-sanity-check)."""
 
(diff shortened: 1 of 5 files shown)


test/framework/options.py
```diff
@@ -443,6 +443,27 @@ def test_ignore_test_failure(self):
         error_pattern = 'Found both ignore-test-failure and skip-test-step enabled'
         self.assertErrorRegex(EasyBuildError, error_pattern, self.eb_main, args, do_build=True, raise_error=True)
 
+    def test_skip_sanity_check(self):
+        """Test ignore failing tests (--ignore-test-failure)."""
```
boegel, line 447:
```suggestion
        """Test skipping of sanity check step (--skip-sanity-check)."""

boegel (changes requested):

Issue #4581, opened 2024-07-18
Default value of `XDG_CONFIG_DIRS` should be `/etc/xdg`, not `/etc`


runiq:
The XDG basedir spec [says](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables) that the fallback value of `XDG_CONFIG_DIRS` should be `/etc/xdg`, not `/etc`:

> If `$XDG_CONFIG_DIRS` is either not set or empty, a value equal to /etc/xdg should be used. 

I would have submitted a PR instead, but simply changing the default config location would probably break a *lot* of setups in the wild, so I decided to create an issue for deliberation.

What do?

bartoldeman:
Fixed by #4591

PR #4591, opened 2024-08-01
Use default value `$XDG_CONFIG_DIRS` from XDG basedir spec: `/etc/xdg` (instead of `/etc`)

runiq:
The [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables) mandates `/etc/xdg` instead of `/etc` for the default value of `XDG_CONFIG_DIRS`.

I'm pretty sure this requires documentation changes because this is potentially a breaking change, but I'm not entirely sure where. I hope a project member can point me to where. :no_mouth::sweat_drops:

Fixes https://github.com/easybuilders/easybuild-framework/issues/4581

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -213,6 +213,12 @@ class EasyBuildOptions(GeneralOption):
 
     DEFAULT_LOGLEVEL = 'INFO'
     DEFAULT_CONFIGFILES = DEFAULT_SYS_CFGFILES[:]
+    if 'XDG_CONFIG_DIRS' not in os.environ:
+        old_etc_location = os.path.join('/etc', 'easybuild.d')
+        if os.path.isdir(old_etc_location) and glob.glob(os.path.join(old_etc_location, '*.cfg')):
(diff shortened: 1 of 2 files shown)


boegel:
Since we're not following the spec w.r.t. the default value if `$XDG_CONFIG_DIRS` we should indeed fix this, but we should also try and limit the impact of this change.

We could for example at least print a warning message when we find one or more configuration files at `/etc/easybuild.d/*.cfg` (but only when `$XDG_CONFIG_DIRS` is *not* set?) to indicate that those configuration files will now be ignored, and that they should be moved to `/etc/xdg/easybuild.d/*.cfg`?

boegel:
We could print a warning if `$XDG_CONFIG_DIRS` is not defined and `*.cfg` files are found in `/etc/easybuild.d/`, which instructs the user to define `$XDG_CONFIG_DIRS` to `/etc` as workaround (which will affect other stuff too though), or move those configuration files to `/etc/xdg/easybuild.d/`.

bartoldeman (commented):
LGTM

bartoldeman (approved):
LGTM

PR #4594, opened 2024-08-04
fix help message for `--http-headers-fields-urlpat` configuration option

Louwrensth:
The `--help` message for `--http-headers-fields-urlpat` contained an error in the example.

Also try to make the help text read a little easier.

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -438,11 +438,11 @@ class EasyBuildOptions(GeneralOption):
                           "(e.g. --hide-deps=zlib,ncurses)", 'strlist', 'extend', None),
             'hide-toolchains': ("Comma separated list of toolchains that you want automatically hidden, "
                                 "(e.g. --hide-toolchains=GCCcore)", 'strlist', 'extend', None),
-            'http-header-fields-urlpat': ("Set extra HTTP header FIELDs when downloading files from URL PATterns. "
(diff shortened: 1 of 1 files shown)


boegel:
@Louwrensth Any reason this is a draft PR?

Louwrensth:
> @Louwrensth Any reason this is a draft PR?

I want to talk about this actually. But technically : I think this one is good to go -- there is not more stuff coming.

The reason why I made it a draft was to not burden you guys with review requests too early. Sometimes we (some people at ITER) have several PRs, and they get attention at the start of the PR, there are usually some good suggestions to improve, but after doing  those changes the back-and-forth seems to slow down and the PR seems ignored. Maybe the PR just goes down on the radar? I feel your team have a lot of reviews on your plate. 

How can we organize it best for all of us?

I wanted to hold back a bunch of PRs until I can offer a bundle of them for you. Does that help at all?

boegel:
@Louwrensth It would be better to discuss this in Slack or via email, since it's out of the context of this PR, but: I'm not sure there's an easy way to prevent that PRs get overlooked over time. We're doing what we can to keep up with contributions, but it's definitely a challenge.

Let's discuss this via another channel: maybe reach out via Slack, and we'll explore some options, we're happy to try and work out an approach that's feasible with the manpower we have available...

boegel (approved):

PR #4595, opened 2024-08-05
convert template constant lists to dicts and export the constants by name

Flamefire:
Currently access to a constant by name is a bit awkward: `next(e[1] for e in TEMPLATE_CONSTANTS if e[0] == name)`

Also having them a list of strings or string and tuple of strings isn't obvious that it is rather a mapping of constants to values/documentation and not a real list. Thus it isn't clear either, that each name must only appear once.

This is a non-functional refactoring simplifying the code and improving the expressiveness.

E.g. to get all defined constants we can simply do: `list(TEMPLATE_CONSTANTS)` (or `list(TEMPLATE_CONSTANTS.keys())`)

Additionally I included a change similar to the EasyConfig constants such that you can import them directly, e.g. `from easybuild.framework.easyconfig.templates import GITHUB_SOURCE`


boegel:
@Flamefire A merge conflict emerged here.

Also, as a reminder: `TEMPLATE_CONSTANTS` is used in the `Python` and `PythonPackage` easyblocks, so those will need to be changed accordingly in a companion PR

Flamefire:
> Also, as a reminder: `TEMPLATE_CONSTANTS` is used in the `Python` and `PythonPackage` easyblocks, so those will need to be changed accordingly in a companion PR

Already done -> https://github.com/easybuilders/easybuild-easyblocks/pull/3410 :)

Micket (approved):
lgtm

Micket:
A few selected test reports uploaded to https://github.com/easybuilders/easybuild-easyblocks/pull/3410#issuecomment-2318295385

This seems fine to me. I haven't heard any objects to this. I would like to merge it, just want to check with others first.

easybuild/framework/easyconfig/templates.py
```diff
@@ -71,112 +71,114 @@
...
+    'cuda_cc_cmake': 'List of CUDA compute capabilities suitable for use with $CUDAARCHS in CMake 3.18+',
+    'cuda_cc_space_sep': 'Space-separated list of CUDA compute capabilities',
+    'cuda_cc_space_sep_no_period':
+    "Space-separated list of CUDA compute capabilities, without periods (e.g. '80 90').",
+    'cuda_cc_semicolon_sep': 'Semicolon-separated list of CUDA compute capabilities',
+    'cuda_int_comma_sep': 'Comma-separated list of integer CUDA compute capabilities',
+    'cuda_int_space_sep': 'Space-separated list of integer CUDA compute capabilities',
+    'cuda_int_semicolon_sep': 'Semicolon-separated list of integer CUDA compute capabilities',
+    'cuda_sm_comma_sep': 'Comma-separated list of sm_* values that correspond with CUDA compute capabilities',
+    'cuda_sm_space_sep': 'Space-separated list of sm_* values that correspond with CUDA compute capabilities',
```
boegel, line 107: `software_commit` is missing here, I'll open a PR to add it back

boegel: fixed in #4621

PR #4596, opened 2024-08-05
allow specifying location for RPATH wrapper scripts via `rpath_wrappers_dir` + also pass `rpath_include_dirs` when preparing build environment for extensions

ocaisa:
Based off of https://github.com/easybuilders/easybuild-framework/pull/4003


easybuild/tools/toolchain/toolchain.py
```diff
@@ -975,7 +975,8 @@ def is_rpath_wrapper(path):
         # need to use binary mode to read the file, since it may be an actual compiler command (which is a binary file)
         return b'rpath_args.py $CMD' in read_file(path, mode='rb')
 
-    def prepare_rpath_wrappers(self, rpath_filter_dirs=None, rpath_include_dirs=None):
+    def prepare_rpath_wrappers(self, rpath_filter_dirs=None, rpath_include_dirs=None, wrappers_dir=None,
+                               add_to_path=True, disable_wrapper_log=False):
```
ocaisa, line 979: I need to rethink this slightly. I think a better approach is to have a build option `rpath_wrappers_dir` and drop these new arguments.

ocaisa:
This feature has been tested for EB 5 via https://github.com/easybuilders/easybuild-easyblocks/pull/3661

boegel:
@ocaisa I changed to target branch in this PR from 5.0.x to develop, you should synchronize your PR branch with current develop branch (which has received a massive update after the release of EasyBuild v5.0.0, see https://github.com/easybuilders/easybuild-framework/pull/4825)

easybuild/framework/easyblock.py
```diff
@@ -2103,7 +2106,9 @@ def install_extensions_sequential(self, install=True):
                 # don't reload modules for toolchain, there is no need since they will be loaded already;
                 # the (fake) module for the parent software gets loaded before installing extensions
                 ext.toolchain.prepare(onlymod=self.cfg['onlytcmod'], silent=True, loadmod=False,
-                                      rpath_filter_dirs=self.rpath_filter_dirs)
+                                      rpath_filter_dirs=self.rpath_filter_dirs,
+                                      rpath_include_dirs=self.rpath_include_dirs,
```
ocaisa, line 2110: The omission of this looks like a bug to me, but I can't be 100%

ocaisa: It probably doesn't actually matter as the wrappers will already have been prepared by the time this is called by an extension.

ocaisa: (see https://github.com/easybuilders/easybuild-framework/pull/4596/files#diff-c1bad9950fee9f74b59e255119ea04ce41e9d9ce8cfa9fa94e831c8d314d106dR1041)

easybuild/framework/easyblock.py
```diff
@@ -2894,6 +2901,15 @@ def prepare_step(self, start_dir=True, load_tc_deps_modules=True):
             '$ORIGIN/../lib64',
         ])
 
+        # Location to store RPATH wrappers
+        self.rpath_wrappers_dir = build_option('rpath_wrappers_dir')
```
ocaisa, line 2905: This build option should only be set on the command line, it should _not_ be part of the configuration or an environment variable. This is because if you did that the wrappers would never get re-created. I'm not sure how to enforce this.

Another option would be to empty the directory within the next `if` block by default.

ocaisa: Hmm, I think I might be wrong here, it looks like the default behaviour is indeed to always overwrite any existing wrappers: https://github.com/easybuilders/easybuild-framework/pull/4596/files#diff-c1bad9950fee9f74b59e255119ea04ce41e9d9ce8cfa9fa94e831c8d314d106dR1074

ocaisa: The only condition that would protect against this is https://github.com/easybuilders/easybuild-framework/pull/4596/files#diff-c1bad9950fee9f74b59e255119ea04ce41e9d9ce8cfa9fa94e831c8d314d106dR1055 and this is not triggered as the updates to `PATH` are not in place until later.

ocaisa: I can add a check there to warn or error (or just log?) that the existing wrappers will be overwritten.

easybuild/framework/easyblock.py
boegel: It has no impact for 2nd extension, where we'll pick up on wrapper created for first extension.

But the RPATH wrapper that was created for the "parent" will no longer be found (because the build environment is reset before build environment for first extension is set up), and so the installation of extensions will be done with an incorrect RPATH wrapper script in which `rpath_include_dirs` is not picked up.

That's bad since `$ORIGIN/../lib` & co are specified that way (see how `self.rpath_wrappers_dir` is defined in `EasyBlock.prepare_step`).

boegel: Hold on, the resetting of build environment before starting installation of extensions is not correct I think, I'm going through things in detail now.

boegel: The `rpath_include_dirs` mechanism was implemented in #2358, which shows that it's only picked up by the "parent" `toolchain.prepare`, the call to `toolchain.prepare` for extensions was overlooked. That confirms to me this is a genuine bug.

I'm still a bit puzzled why this doesn't cause real trouble though, or how it has flown under the radar...

easybuild/tools/options.py
```diff
@@ -524,6 +524,8 @@ def override_options(self):
             'rpath-filter': ("List of regex patterns to use for filtering out RPATH paths", 'strlist', 'store', None),
             'rpath-override-dirs': ("Path(s) to be prepended when linking with RPATH (string, colon-separated)",
                                     None, 'store', None),
+            'rpath-wrappers-dir': ("Absolute path to directory to use for RPATH wrappers creation/use",
```
boegel, line 527: Rather than adding an EasyBuild configuration option for this, we can just only give control over through an easyblock which can set `self.rpath_wrappers_dir` in the constructor, so there's no control from the outside to tweak the location of the generaed RPATH wrapper scripts.

boegel (changes requested):
@ocaisa Bunch of changes proposed for this PR in https://github.com/ocaisa/easybuild-framework/pull/47

boegel (approved):
lgtm

PR #4599, opened 2024-08-05
fix `test_compiler_cache`

Flamefire:
The test searches for the real command, e.g. `gcc`, deletes the first one found and uses the rest.
However there could be multiple ones: /usr/bin/gcc /bin/gcc In this case the second would be wrongly used as the input. Fix by converting the output of `which` into an array and using the 2nd entry.


boegel (approved):
lgtm

PR #4600, opened 2024-08-05
fix test_toy_lock_cleanup_signals

Flamefire:
The 1s timeout might not be enough for the build to create the log (e.g. when ran in a debugger)
Increase the timeout and also pass required arguments to actually create the lock.


boegel (approved):
lgtm

PR #4601, opened 2024-08-05
copy build directory and/or log file(s) if installation failed to path specified via `--failed-install-build-dirs-path` or `--failed-install-logs-path`

gkaf89:
The files can be build in some selected build path (--buildpath), and the logs of successful compilation are then concentrated to some other location for permanent storage (--logfile-format). Logs of failed builds remain in the build path location so that they can be inspected.

However, this setup is problematic when building software in HPC jobs. Quite often in HPC systems the build path is set to some fast storage local to the node, like NVME raid mounted on `/tmp` or `/dev/shm` (as suggested in the documentation: https://docs.easybuild.io/configuration/#buildpath). The node storage is often wiped out after the end of a job, so the log files and the artifacts are no longer available after the termination of the job.

This commit adds an option to accumulate errors in some more permanent location, so they can be easily inspected after a failed build.


gkaf89:
I am not sure what is the best way to select the build directory so that I can move it to a more permanent location. Are there any suggestions?

akesandgren:
> @gkaf89 The `builddir` variable that is set in each easyblock instance hold the path to the build directory for that particular easyconfig. You can determine the relative path via the `build_path()` function that is available from `easybuild.tools.config`, that should report the top directory that corresponds to the `buildpath` EasyBuild configuration option (see also https://docs.easybuild.io/configuration/#buildpath).
> 
> For, for example, for `example-1.2.3-GCC-12.3.0.eb`, the `builddir` path would be something like `/tmp/myuser/easybuild/build/example/1.2.3/GCC-12.3.0/`, with `buildpath` set to `/tmp/myuser/easybuild/build`. Not that the actual build directory in which the compilation is being done would be one level deeper, corresponding to the unpacked source tarball, so something like `/tmp/myuser/easybuild/build/example/1.2.3/GCC-12.3.0/example-1.2.3/`.
> 
> So, I think you could create a subdirectory in the permanent storage location that uses the name of the easyconfig file (to keep it simple), and copy the _contents_ of `builddir` in there. You do somehow want to make sure that the target path is unique though, because you could have multiple builds ongoing on different nodes that would all copy to the same permanent location in the end...

Yeah, the thing to copy should be `builddir` into a path with the diff of `buildpath` and `builddir` based in permanent-storage-location. Just make sure to remove old remnants of that first :-)

boegel:
@gkaf89 If you need any help with this, do let us know!

gkaf89:
@boegel The commit is ready. I won't have enough time to familiarize myself with the test framework for the `EasyBlockTest` class to prepare a test before the next release.

The commit can be tested by modifying the configuration options of some easyconfig that uses the system toolchain to cause a failure. For instance I added the option
```
configopts = '--some-invalid-option'
```
in `zlib-1.3.1.eb`. The result is that the temporary log file in the build directory and the extracted source code are copied in a permanent location.

easybuild/framework/easyblock.py
```diff
@@ -4190,6 +4197,33 @@ def print_dry_run_note(loc, silent=True):
...
+    if application_log:
+        # there may be multiple log files, or the file name may be different due to zipping
+        logs = glob.glob('%s*' % application_log)
+        print_msg(
+            "Results of the build can be found in the temporary log file(s) %s" % ', '.join(logs),
+            log=_log,
+            silent=silent
+        )
+
+        if err_log_path and not(success):
```
boegel, line 4210: code style check fails here:
```
./easybuild/framework/easyblock.py:4210:32: E275 missing whitespace after keyword
```

```suggestion
        if err_log_path and not success:

**Review by boegel — CHANGES_REQUESTED** [review]

**Comment by boegel** [comment]
@gkaf89 There's a problem with the tests, looks like `test_toy_build` was broken by the changes being made here?
See for example https://github.com/easybuilders/easybuild-framework/actions/runs/10805964835/job/29973948603

**Comment by gkaf89** [comment]
The failure is caused because the target location for permanent storage is the same as the source location. The steps I am following to resolve the issue:

- add a source/destination check to avoid a hard failure, and
- detect how the source and the destination path end with the same value in the test.

**Comment by gkaf89** [comment]
@boegel  Some edge cases where uncovered by the tests. The latest commit resolves the issue.

I leave it up you if you prefer to move it to version 5. I am not familiar enough with the tests to test the PR extensively.

File: easybuild/framework/easyblock.py (hunk 1 above)
Comment (gkaf89) [reply easybuild/framework/easyblock.py]: I have made some changes and this specific edit is no longer relevant. I am following the style and avoid parenthesis around single not arguments. Can we have the test rerun?

**Comment by boegel** [comment]
@gkaf89 As briefly discussed during the conf call today, it would be good if you could add a test (or enhance an existing one, like `test_toy_broken`) to verify that the added functionality works as intended (and keeps working).

Do let us know if you need any help with that!

**Comment by gkaf89** [comment]
Some tests fail because a warning is printed:
```
WARNING: Command ' gcc toy.c -o toy ' failed, but we'll ignore it...
```

I want the compilation to fail, is there a way to silence the warning?

**Comment by gkaf89** [comment]
> I want the compilation to fail, is there a way to silence the warning?

Using `run_test_toy_build_with_output` instead of `test_toy_build` to catch and ignore all output. I am not sure if the function `run_test_toy_build_with_output` is used as intended here.

File: easybuild/tools/config.py [hunk 2]
```diff
@@ -843,6 +845,33 @@ def log_path(ec=None):
...
 
 
+def error_log_path(ec=None):
+    """
+    Return the default error log path
+
+    This is a path where file from the build_log_path can be stored permanently
+    :param ec: dict-like value that provides values for %(name)s and %(version)s template values
+    """
+    error_log_path = ConfigurationVariables()['errorlogpath']
```
Flamefire, line 855: Not a good idea to have a variable with the same name as a function. Maybe rename the function to `get_....` or `default_...`?

easybuild/tools/config.py
```diff
@@ -843,6 +845,33 @@ def log_path(ec=None):
     return log_file_format(return_directory=True, ec=ec, date=date, timestamp=timestamp)
 
 
+def error_log_path(ec=None):
+    """
+    Return the default error log path
+
+    This is a path where file from the build_log_path can be stored permanently
```
Flamefire, line 852:
```suggestion
    This is a path where files from the build_log_path can be stored permanently
```

easybuild/tools/config.py
```diff
@@ -843,6 +845,33 @@ def log_path(ec=None):
...
+    :param ec: dict-like value that provides values for %(name)s and %(version)s template values
+    """
+    error_log_path = ConfigurationVariables()['errorlogpath']
+
+    if ec is None:
+        ec = {}
+
+    name, version = ec.get('name', '%(name)s'), ec.get('version', '%(version)s')
```
Flamefire, line 860: Why would you want `%(name)s` in the folder name?

easybuild/tools/config.py
```diff
@@ -843,6 +845,33 @@ def log_path(ec=None):
     return log_file_format(return_directory=True, ec=ec, date=date, timestamp=timestamp)
 
 
+def error_log_path(ec=None):
+    """
+    Return the default error log path
+
+    This is a path where file from the build_log_path can be stored permanently
+    :param ec: dict-like value that provides values for %(name)s and %(version)s template values
```
Flamefire, line 853: Doesn't match what is used. It expects `name` and `version` keys, not templates

easybuild/tools/config.py
```diff
@@ -843,6 +845,33 @@ def log_path(ec=None):
...
+
+    name, version = ec.get('name', '%(name)s'), ec.get('version', '%(version)s')
+    date = time.strftime("%Y%m%d")
+    timestamp = time.strftime("%H%M%S")
+
+    base_path = '/'.join([error_log_path, name + '-' + version, date + '-' + timestamp])
+
+    path = base_path
+    inc_no = 1
+    while os.path.exists(path):
```
Flamefire, line 868: Maybe rather use `filetools.create_unused_dir`? This avoids duplicating the logic and a possible conflict when another eb instance runs in parallel

test/framework/easyconfigs/test_ecs/t/toy/toy-0.0-test.eb
```diff
@@ -18,7 +18,7 @@ checksums = [[
     ('size', 273),
 ]]
 patches = [
-    'toy-0.0_fix-silly-typo-in-printf-statement.patch',
+    'toy-0.0_add-bug.patch',
```
Flamefire, line 21: was this intentional?

test/framework/toy_build.py
```diff
@@ -228,6 +233,9 @@ def test_toy_build(self, extra_args=None, ec_file=None, tmpdir=None, verify=True
                 msg = "Pattern %s found in full test report: %s" % (regex.pattern, test_report_txt)
                 self.assertTrue(regex.search(test_report_txt), msg)
 
+        if check_errorlog is not None:
```
Flamefire, line 236: Isn't this better done outside? Doesn't need to be in this function, does it?

test/framework/utilities.py
```diff
@@ -514,3 +516,15 @@ def find_full_path(base_path, trim=(lambda x: x)):
             break
 
     return full_path
+
+
+class TempDirectory:
```
Flamefire, line 521: There already is [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) which additionally allows using in a context manager

test/framework/easyconfigs/test_ecs/t/toy/toy-0.0-buggy.eb
```diff
@@ -0,0 +1,33 @@
+name = 'toy'
```
Flamefire, line 1: What we usually do instead of adding yet another test easyconfig is `read_file` it and write with modifications. See `self.contents` and `self.eb_file`

Flamefire (changes requested):
I added some suggestions inline

test/framework/easyconfigs/test_ecs/t/toy/toy-0.0-test.eb
gkaf89: No, that is a mistake, thank you for catching that.

easybuild/tools/config.py
gkaf89: I used the `%(name)s` and `%(version)s` template strings as default values. I have seen this use in other cases, `log_file_format` function on the same file (they indeed appear in file paths). It seems like a nice way to inform the user about a missing value without a hard failure.

Should we avoid using template names as default values or explain their use in a better way?

easybuild/tools/config.py
Flamefire: That "coupling" is intentional: If you don't create the directory you have a race condition for process running in parallel defeating the purpose of this function. Why don't you want that directory created? Can't you just fill it later using the created directory?

easybuild/tools/config.py
Flamefire: Ah I see. Isn't trivial to understand for me what is meant by that.   I'd change the documentation to
>     :param ec: dict-like value with at least the keys 'name' and 'version'

I would not use a default and just fail hard if the keys do not exist because that is what is documented: The function expects a dict with those keys. If you don't provide one it is the fault of the caller. If we do it right in easybuild the keys will always exist, so no need for fallbacks. Or am I missing anything?

test/framework/utilities.py
gkaf89: Thanks, nice catch! Using context manager.

easybuild/tools/config.py
gkaf89: You couple through the file system, got it.

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4206,52 @@ def print_dry_run_note(loc, silent=True):
...
+        print_msg(
+            "Results of the build can be found in the temporary log file(s) %s" % ', '.join(logs),
+            log=_log,
+            silent=silent
+        )
+
+        err_log_dir_data = get_error_log_dir(ec=easyconfig)
+        if not build_successful and err_log_dir_data is not None:
+            for log_file in logs:
+                err_log_path = get_unique_err_log_dir(err_log_dir_data)
```
Flamefire, line 4238: Why do you want a different folder for each log file?
Edit: See above. I'd suggest to check for `if logs or is_readable(builddir)` and create the folder only then create it.

Maybe early returns might make it more clear by reducing the nesting: `if not application_log: return; ... if build_successful or not log_files: return; create-folder... if folder_unset: return`

I'd put the logic for creating an unused error log dir into `get_error_log_dir` maybe named ? get_new_error_log_dir`

easybuild/tools/config.py
```diff
@@ -843,6 +845,31 @@ def log_path(ec=None):
...
+        return None
+
+    if ec is None:
+        ec = {}
+
+    name, version = ec.get('name'), ec.get('version')
+    date = time.strftime("%Y%m%d")
+    timestamp = time.strftime("%H%M%S")
+
+    parent_dir = '/'.join([errorlogpath, name + '-' + version])
```
Flamefire, line 867:
```suggestion
    parent_dir = os.path.join(errorlogpath, name + '-' + version)
```

easybuild/tools/config.py
```diff
@@ -843,6 +845,31 @@ def log_path(ec=None):
...
+def get_error_log_dir(ec=None):
+    """
+    Return a pair with the parent directory and directory name of the error log path
+
+    The error log path is a path where files from the build_log_path can be stored permanently
+    :param ec:  dict-like value with at least the keys 'name' and 'version' defined
+    """
+    errorlogpath = ConfigurationVariables()['errorlogpath']
+
+    if errorlogpath == "":
```
Flamefire, line 857:
```suggestion
    if not errorlogpath:
```

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4206,52 @@ def print_dry_run_note(loc, silent=True):
     dry_run_msg(msg, silent=silent)
 
 
+def persist_failed_compilation_log_and_artifacts(build_successful, application_log, silent, app, easyconfig):
+    err_log_dir = None
+
+    def get_unique_err_log_dir(err_log_dir_data):
+        nonlocal err_log_dir
```
Flamefire, line 4213: This makes it hard to reason about, see my question below which seems to be me missing this: It isn't getting a unique dir per call but only once which isn't obvious

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4206,52 @@ def print_dry_run_note(loc, silent=True):
...
+
+        err_log_dir_data = get_error_log_dir(ec=easyconfig)
+        if not build_successful and err_log_dir_data is not None:
+            for log_file in logs:
+                err_log_path = get_unique_err_log_dir(err_log_dir_data)
+                target_file = os.path.join(err_log_path, os.path.basename(log_file))
+                do_if_paths_distinct(copy_file, log_file, target_file)
+
+            builddir = app.builddir
+            if is_readable(builddir):
```
Flamefire, line 4243: That should likely be another option. Copying log files is one thing (size of few KBs} but the build artefacts (and actually also all extracted sources) are multiple MB or even GBs. Especially as the option is `errorlogpath` this will be surprising at least

easybuild/tools/config.py
```diff
@@ -843,6 +845,31 @@ def log_path(ec=None):
...
+        ec = {}
+
+    name, version = ec.get('name'), ec.get('version')
+    date = time.strftime("%Y%m%d")
+    timestamp = time.strftime("%H%M%S")
+
+    parent_dir = '/'.join([errorlogpath, name + '-' + version])
+    name = date + '-' + timestamp
+
+    return (parent_dir, name)
```
Flamefire, line 870: This looks odd. Using `os.path.join` and if required `os.basename/dirname` might be more obvious. Or create the folder here

Flamefire (changes requested):

test/framework/toy_build.py
```diff
@@ -271,6 +276,85 @@ def test_toy_broken(self):
...
+
+        with open(f"{log_file}", 'r') as p_log_file:
+            self.assert_error_reported(p_log_file.read(), output_regexs)
+
+    def test_toy_broken_compilation(self):
+        """Test whether log files and the build directory are copied to a permanent location after a failed
+           compilation."""
+        with tempfile.TemporaryDirectory() as tmpdir, \
+            tempfile.TemporaryDirectory() as tmp_logdir, \
+            tempfile.TemporaryDirectory() as tmp_easyconfig_dir:
```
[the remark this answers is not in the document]
gkaf89: Not sure how to resolve this issue!

test/framework/easyconfigs/test_ecs/t/toy/toy-0.0-buggy.eb
gkaf89: I used the existing convention as suggested.

Also noting here that I would prefer just adding files for integration tests.

test/framework/toy_build.py
gkaf89: I missed that one. Tests are now done outside.

easybuild/framework/easyblock.py
gkaf89: I changed the name to `get_errorlog_dir`. The 'unique' epithet is not very relevant here indeed.

easybuild/tools/config.py
gkaf89: I refactored this. This is related to the interface of `create_unused_dir` where I suggested a modification.

easybuild/framework/easyblock.py
gkaf89: I agree, this is important. I am adding an additional option about artifacts.

easybuild/tools/config.py
gkaf89: Fixed!

easybuild/tools/config.py
gkaf89: Fixed.

test/framework/toy_build.py
```diff
@@ -271,6 +278,88 @@ def test_toy_broken(self):
...
+        with open(f"{log_file}", 'r') as p_log_file:
+            self.assert_error_reported(p_log_file.read(), output_regexs)
+
+    def test_toy_broken_compilation(self):
+        """Test whether log files and the build directory are copied to a permanent location after a failed
+           compilation."""
+        with tempfile.TemporaryDirectory() as tmpdir, \
+            tempfile.TemporaryDirectory() as tmp_logdir, \
+            tempfile.TemporaryDirectory() as tmp_artifactdir, \
+            tempfile.TemporaryDirectory() as tmp_easyconfig_dir:
```
Flamefire, line 334: I think this works:
```suggestion
                tempfile.TemporaryDirectory() as tmp_logdir, \
                tempfile.TemporaryDirectory() as tmp_artifactdir, \
                tempfile.TemporaryDirectory() as tmp_easyconfig_dir:
```

Or maybe use a single parent folder instead. `self.test_prefix` is a directory created for each test case and removed afterwards

easybuild/framework/easyblock.py
Flamefire: > you have to check a few conditions that depend on the directory path

What do you need to check based on a path that just got created?

easybuild/framework/easyblock.py
Flamefire: I think you misunderstood what I meant by "option": Enabling `--error-log-path` should not also copy build artifacts. That should be another EB option to enable additionally.

> I am not sure how to cleanly separate logs from artifacts though.

Which logs exactly? You already know the EB log filenames don't you?

easybuild/tools/config.py
Flamefire: You cannot change the interface of that function as that would break every user. You can call it as `create_unused_dir(dirname, basename)` instead

If we really want the convenience the new signature needs to be `def create_unused_dir(parent_folder, name=None)` in which case the splitting is done there if the 2nd argument is not given. That needs extra tests of course.

gkaf89: I couldn't find any use of `create_unused_dir`!

Flamefire: `CMakeMake` & `MesonNinja` easyblocks use it. There might also be custom easyblocks and hooks using it that are not part of the easybuilders repositories.

test/framework/toy_build.py
gkaf89: Generated a single temporary directory, and created all other directories in the temporary directory.

easybuild/tools/config.py
gkaf89: I had to create a new function that can create arbitrary paths. Since the users can ask for any path to permanently store logs and artifacts, the function should be able to create missing parent directories.

easybuild/framework/easyblock.py
gkaf89: Yes, I created an option for build artifacts. The default is not to store the artifacts.

easybuild/tools/config.py
gkaf89: Resolved in latter discussion.

easybuild/tools/config.py
gkaf89: Fixed.

easybuild/framework/easyblock.py
gkaf89: Yes, that is an accurate description. After the directory creation I have to update the path, since we need a unique directory name to avoid race conditions when multiple instances of EassyBuild are running. The update of the path requires some extra steps. I am not satisfied with how the update is implemented, but it is at least localized in `persist_failed_compilation_log_and_artifacts` and can be refactored if needed.

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,98 @@ def print_dry_run_note(loc, silent=True):
     dry_run_msg(msg, silent=silent)
 
 
+def persist_failed_compilation_log_and_artifacts(build_successful, application_log, silent, app, easyconfig):
+    persistent_dirs = {}
+
+    def get_dir(path, dirname):
```
Flamefire, line 4213: I still don't understand why you need this and cannot just create the new folder inside `initialize_persistent_dir_and_replicate_if_paths_distinct`. See below

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,98 @@ def print_dry_run_note(loc, silent=True):
...
+    def initialize_persistent_dir_and_replicate_if_paths_distinct(replication_operation,
+                                                                  get_directory,
+                                                                  source_path,
+                                                                  destination_base_path,
+                                                                  destination_relative_path):
+        abs_source_path = os.path.abspath(source_path)
+        abs_destination_base_path = os.path.abspath(destination_base_path)
+        abs_destination_path = os.path.join(abs_destination_base_path, destination_relative_path)
+
+        if not os.path.exists(abs_source_path):
```
Flamefire, line 4236: This can never be false, can it? You glob the logs and check the builddir before calling this.

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,98 @@ def print_dry_run_note(loc, silent=True):
...
+        return get_dir(path, 'log')
+
+    def initialize_persistent_dir_and_replicate_if_paths_distinct(replication_operation,
+                                                                  get_directory,
+                                                                  source_path,
+                                                                  destination_base_path,
+                                                                  destination_relative_path):
+        abs_source_path = os.path.abspath(source_path)
+        abs_destination_base_path = os.path.abspath(destination_base_path)
+        abs_destination_path = os.path.join(abs_destination_base_path, destination_relative_path)
```
Flamefire, line 4234: It would be easier to understand if those 2 were 1 argument in the first place. Had to search whether either was used again. (Too many variables make it harder to reason in general.)

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,98 @@ def print_dry_run_note(loc, silent=True):
...
+    if errorlog_path is not None:
+        errorlog_path = os.path.join(errorlog_path, datetime_stamp)
+
+        for log_file in logs:
+            initialize_persistent_dir_and_replicate_if_paths_distinct(
+                copy_file,
+                get_log_dir,
+                log_file,
+                errorlog_path,
+                os.path.basename(log_file)
```
Flamefire, line 4272: The "if_paths_distinct" here is:
`log_file == errorlog_path / basename(log_file)` or `dirname(application_log) == errorlog_path`

In which situation can this be true? And especially as `errorlog_path` includes the timestamp with seconds I'd argue it is never the same

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,98 @@ def print_dry_run_note(loc, silent=True):
...
+        errorartifact_path = os.path.join(errorartifact_path, datetime_stamp)
+
+        builddir = app.builddir
+        if is_readable(builddir):
+            initialize_persistent_dir_and_replicate_if_paths_distinct(
+                copy_dir,
+                get_artifact_dir,
+                builddir,
+                errorartifact_path,
+                app.get_relative_builddir_base_path()
```
Flamefire, line 4292: Similar here: `builddir == errorartifact_path / get_relative_builddir_base_path` can never be true, can it?

Flamefire (changes requested):
I think you can inline `initialize_persistent_dir_and_replicate_if_paths_distinct` which makes it much easier to read. Given both conditions leading to an early return cannot be met it will always copy.

So the logic can be greatly simplified to:
`if error_log_path and logs: create(unique_error_log_path) and copy(logs, dir)`

And similar for  `errorartifact_path`

This would also allow to print the final path as currently `errorartifact_path` is not the real path as it will be modified to be unique

Or am I missing anything?

gkaf89:
> So the logic can be greatly simplified ...
> 
> [...]
> 
> Or am I missing anything?

This section needs refactoring indeed. Let me describe the requirements, and I will try to simplify it tomorrow.

1. If the target directory is a subdirectory of the build directory, I think we should print a warning and do nothing. Preserving the log files and build artifacts should never mangle the build directories, even if the user makes a mistake in the paths they provide.

2. The postfix that ensures that the log and artifact directories are unique should be the same. Something like
    ```
    logs_<date 0>_0 <--> artifacts_<date 0>_1
    logs_<date 0>_1 <--> artifacts_<date 0>_0
    ```
    should be prevented. This is mix-up occurs in test where multiple installation launch in parallel (much less probable in normal usage).

3. There is currently a single log file (.log) created for each easyconfig. I would like to keep the script more generic so that it handles multiple files in case something changes in the future (and because searching for the log file already returns a list). Artifact directory are always unique.

Flamefire:
> If the target directory is a subdirectory of the build directory,

You are currently not checking for that, are you? Only if the target and build dir are equal.   
The is-subdir can be checked easily though, so I guess it makes sense

2. I guess this situation rarely occurs and we can't really prevent that, can we? Maybe we could add a symlink in each folder to the other if both are created.

gkaf89:
1. A check was added to determine if the directories where the logs and build artifacts are stored are a inside the build directory. If they are, a warning is printing and the build continues.
2. I added a function that creates a list of unused directories, and if any directory in the list is not available, then it adds the same suffix to all of them.
3. Code is able to handle multiple log files if they are added in the future.

I need to add a test for (1). I will request a review as soon as the test is ready.

easybuild/tools/filetools.py
```diff
@@ -2980,3 +2987,81 @@ def create_unused_dir(parent_folder, name):
...
+    for p in paths:
+        shutil.rmtree(p, ignore_errors=True)
+
+
+def create_unused_dirs(paths, index_upper_bound=10000):
+    """
+    Create directories with given paths, including the parent directories
+    When a directory in the same path for any of the path  in the path list already exists, then the suffix '_<i>' is
+    appended for i=0..(index_upper_bound-1) until an index is found so that all required path as unused. All created
+    directories have the same postfix.
```
Flamefire, line 3025: To be consistent:
```suggestion
    directories have the same suffix.
```

Maybe put the "All created" on the same line because that is the most important part of the function, isn't it?

easybuild/tools/filetools.py
```diff
@@ -2980,3 +2987,81 @@ def create_unused_dir(parent_folder, name):
...
+    for path in paths:
+        for parent in set(paths) - {path}:
+            if is_predecessor(parent, path):
+                raise EasyBuildError("Path '%s' is predecesor of '%s'" % (parent, path))
+
+    final_paths = paths
+    first_nonexisting_parents = list(map(get_first_nonexisting_parent, paths))
+    paths_exist = False
+    number = -1
+    while number < index_upper_bound and not paths_exist:  # Start with no suffix and limit the number of attempts
```
Flamefire, line 3040: What about 
```suggestion
    while for number in range(-1, 10000):  # Start with no suffix and limit the number of attempts
```

Is the argument to the function for the upper bound really required? With this you can instead of `if number == index_upper_bound and not paths_exist:` just use a flag `valid_directories_created` or so and `break` in the loop

easybuild/tools/filetools.py
```diff
@@ -2980,3 +2987,81 @@ def create_unused_dir(parent_folder, name):
...
+        tried_paths = []
+        if number >= 0:
+            final_paths, first_nonexisting_parents = _apply_suffix(paths, number)
```
Flamefire, line 3043: With `first_nonexisting_parents = paths` set by this function I think this is not fully right:   
Previously it is set to a parent directory let's say for `/foo/bar/a/b` it was `/foo/bar` and now gets changed to `/foo/bar/a/b`.  
The `set_gid_sticky_bits`will not be called for `bar` or `a` although they where just created

easybuild/tools/filetools.py
```diff
@@ -2980,3 +2987,81 @@ def create_unused_dir(parent_folder, name):
...
+        try:
+            for p in final_paths:
+                tried_paths.append(p)
+                os.makedirs(p)
+            paths_exist = True
+        except OSError as err:
+            # Distinguish between error due to existing folder and anything else
+            if tried_paths != [] and not os.path.exists(tried_paths[-1]):
```
Flamefire, line 3051: It can never be empty because it checks only why the last folder added to the list couldn't be created, can it?
```suggestion
            if not os.path.exists(tried_paths[-1]):
```

Note: `x != []` can be simply `x`

test/framework/filetools.py
```diff
@@ -420,6 +420,28 @@ def test_normalize_path(self):
...
 
+    def test_is_predecessor(self):
+        """Test is_predecessor"""
+        self.assertTrue(ft.is_predecessor('/foo/bar', '/foo/bar/test0'))
+        self.assertTrue(ft.is_predecessor('/foo/bar', '/foo/bar/test0/test1'))
+        self.assertTrue(ft.is_predecessor('/foo/bar', '/foo/bar'))
+        self.assertFalse(ft.is_predecessor('/foo/bar', '/foo/test'))
+
+        # Check that trailing slashes are ignored
+        self.assertTrue(ft.is_predecessor('/foo/bar', '/foo/bar'))
```
Flamefire, line 431: This one is checked above already and doesn't match the comment
```suggestion
```

test/framework/filetools.py
```diff
@@ -3522,6 +3544,31 @@ def test_compat_makedirs(self):
         py2vs3.makedirs(name, exist_ok=True)  # No error
         self.assertExists(name)
 
+    def test_get_first_nonexisting_parent(self):
+        """Test get_first_nonexisting_parent function."""
+        test_root = os.path.join(self.test_prefix, 'a')
+
+        base_path = os.path.join(self.test_prefix, 'a')
```
Flamefire, line 3551: Can you remove one of them or be consistent in using it? It looks like `test_root` and `base_path` can always be just `base_path`

test/framework/filetools.py
```diff
@@ -3560,6 +3607,126 @@ def test_create_unused_dir(self):
...
         self.assertExists(path)
 
+    def test_create_unused_dirs(self):
+        """Test create_unused_dirs function."""
+        requested_paths = []
+        requested_paths.append(os.path.join(self.test_prefix, 'folder_a'))
+        requested_paths.append(os.path.join(self.test_prefix, 'folder_b'))
+        paths = ft.create_unused_dirs(requested_paths)
+        self.assertEqual(paths, requested_paths)
+        map(lambda p: self.assertExists(p), paths)
```
Flamefire, line 3617: `map` returns a generator, i.e. nothing gets tested here. Better to use a loop.

test/framework/filetools.py
```diff
@@ -3560,6 +3607,126 @@ def test_create_unused_dir(self):
...
+
+        # Repeat with existing folder(s) should create new ones
+        for i in range(10):
+            requested_paths = []
+            requested_paths.append(os.path.join(self.test_prefix, 'folder_a'))
+            requested_paths.append(os.path.join(self.test_prefix, 'folder_b'))
+            paths = ft.create_unused_dirs(requested_paths)
+            self.assertEqual(paths, list(map(lambda p: p + '_%s' % i, requested_paths)))
+            map(lambda p: self.assertExists(p), paths)
```
Flamefire, line 3626: Same situation as above about use of `map`

test/framework/filetools.py
```diff
@@ -3560,6 +3607,126 @@ def test_create_unused_dir(self):
...
+        self.assertEqual(paths, requested_paths)
+        map(lambda p: self.assertExists(p), paths)
+
+        # Repeat with existing folder(s) should create new ones
+        for i in range(10):
+            requested_paths = []
+            requested_paths.append(os.path.join(self.test_prefix, 'folder_a'))
+            requested_paths.append(os.path.join(self.test_prefix, 'folder_b'))
+            paths = ft.create_unused_dirs(requested_paths)
+            self.assertEqual(paths, list(map(lambda p: p + '_%s' % i, requested_paths)))
```
Flamefire, line 3625: (Especially) with a lambda there usually isn't an advantage of using map instead of a list comprehension and the latter becomes likely better:
```suggestion
            self.assertEqual(paths, [f'{p}_{i}' for p in requested_paths])
```
E.g. it avoids pitfalls like not running the generator or trying to run it twice.   
I think we also usually use that in EB

test/framework/filetools.py
```diff
@@ -3560,6 +3607,126 @@ def test_create_unused_dir(self):
...
+        os.mkdir(os.path.join(self.test_prefix, f"existing_idx_b_{existing_idx}"))
+
+        def expected_idx(n):
+            if n >= existing_idx:
+                return n + 1
+            return n
+
+        for i in range(3):
+            requested_paths = []
+            requested_paths.append(os.path.join(self.test_prefix, 'existing_idx_a'))
```
Flamefire, line 3641: Maybe use variables for `os.path.join(self.test_prefix, "existing_idx_a|b")` to make it clearer that those are the same folders?

test/framework/filetools.py
```diff
@@ -3560,6 +3607,126 @@ def test_create_unused_dir(self):
...
+            requested_paths.append(os.path.join(self.test_prefix, 'existing_idx_b'))
+            paths = ft.create_unused_dirs(requested_paths)
+            self.assertEqual(paths, list(map(lambda p: p + '_%s' % expected_idx(i), requested_paths)))
+            map(lambda p: self.assertExists(p), paths)
```
Flamefire, line 3645: See above and below. Maybe add a `def assertAllExists(self, paths)` function to the class or `def assertAllExists(paths)` to the test

test/framework/toy_build.py
```diff
@@ -271,6 +280,95 @@ def test_toy_broken(self):
...
+        self.assertTrue(len(saved_log_files) == 1, f"Unique copy of log file '{log_file}' made")
+        for saved_log_file in saved_log_files:
+            msg = f"Log file '{log_file}' copied successfully"
+            self.assertTrue(filecmp.cmp(str(log_file), str(saved_log_file), shallow=False), msg)
+
+    def assert_error_reported(self, outtxt, output_regexs):
+        for regex_pattern in output_regexs:
+            regex = re.compile(regex_pattern, re.M)
+            msg_stdout = "Pattern %s found in full test report: %s" % (regex.pattern, outtxt)
+            self.assertTrue(regex.search(outtxt), msg_stdout)
```
Flamefire, line 321: For 5.0x we can use:
```suggestion
            self.assertRegex(outtxt, regex)
```

Could also just use `regex_pattern` instead of `regex` if the flags are not required (here they are). This avoids the manually constructed message

Flamefire (changes requested):
>    I added a function that creates a list of unused directories, and if any directory in the list is not available, then it adds the same suffix to all of them.

I see. This solves the issue of potentially having different suffixes although the code becomes quite complex given for handling this rare case. But can't help that...

> I need to add a test for (1). I will request a review as soon as the test is ready.

Sorry I missed that.
Anyway: Review added. But I guess you need to target to 5.0x branch instead to avoid the python2 failures or wait until 5.0x is merged to develop.

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,90 @@ def print_dry_run_note(loc, silent=True):
...
+        return
+
+    datetime_stamp = time.strftime("%Y%m%d") + '-' + time.strftime("%H%M%S")
+    operation_args = []
+
+    log_error_path = get_log_error_path(ec=easyconfig)
+    if log_error_path is not None:
+        log_error_path = os.path.join(log_error_path, datetime_stamp)
+
+        if not is_predecessor(app.builddir, log_error_path):
```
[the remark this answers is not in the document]
gkaf89: The problem with saving the build artifacts and log files in the build directory is twofold:

- we are modifying the build directory (and in the case of artifacts we are copying a directory to itself), and 
- the build directory is not persistent, and may be deleted.

The time stamp won't help if the root path is pointing to the build directory.

We should never copy files in the build directory, but I have no strong opinion on how to handle the failure. In my opinion, if the building process has started, then we should only print a warning. I would prefer a hard failure to occur early, for instance in a sanity check of the input arguments, before starting the compilation of any software.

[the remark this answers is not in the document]
Flamefire: Oh, some part of the sentence was missing and/or mixed up, sorry. I meant:
> it can only be caused by using a `--builddir` and `--error-log-path` pointing to the same directory

I.e. as you say the inputs are already wrong. I'd rather catch this hard here, i.e. a failure to make it clear that we assume this to be "valid" at this point.

> I would prefer a hard failure to occur early, for instance in a sanity check of the input arguments, before starting the compilation of any software.

I agree. At this point I think it should even check resolved paths as due to symlinks it can still point to subfolders which we might only catch here but is still a user error.

> I have no strong opinion on how to handle the failure.

Me neither. My guideline in general is: "If we are not able to do what the user request we fail."   
So I'd make this a hard failure here and for a better UX try to detect this at the start not at the end of a long process.   
But I'm not a maintainer so this is just my personal opinion and not required to land this PR if they think otherwise which is fine for me.

Final note: I think it would be better to do the check before appending the timestamp as I think the error message not including something we added but only what the user specified is clearer

easybuild/tools/filetools.py
```diff
@@ -2980,3 +2987,81 @@ def create_unused_dir(parent_folder, name):
     set_gid_sticky_bits(path, recursive=True)
 
     return path
+
+
+def get_first_nonexisting_parent(path):
```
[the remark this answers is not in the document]
gkaf89: The function follows the same convention with `is_predecessor` (now renamed to `is_parent_path`), where a path is considered a predecessor of itself, that's why `parent = path` is possible. I find that using a reflexive relation ([partial order](https://en.wikipedia.org/wiki/Partially_ordered_set)) usually simplifies the code.

I am also renaming the function to `get_first_nonexisting_parent_path` to maintain a consistent nomenclature.

easybuild/tools/filetools.py
gkaf89: It can be `tried_paths = []` only if an empty list of paths is provided as argument. I wouldn't like the function to fail in this case, it should just do nothing.

test/framework/filetools.py
gkaf89: Noted, and removed.

easybuild/tools/filetools.py
gkaf89: You are right! I got carried away from the single file case. With multiple files `get_first_nonexisting_parent_path` should be called every time. I should probably add some arguments with default values for `set_gid_sticky_bits` and create a test.

easybuild/tools/filetools.py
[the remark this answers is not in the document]
Flamefire: I wouldn't expect that X is a parent of X. Sounds wrong to me that something can be the parent of itself.

Anyway I guess it is fine as long as it is documented. Here it could be "Get first directory that does not exist, starting at path and going up"

Similar of `is_parent_path` where something like "Check if `parent` is a prefix of `path`" or similar.

That would capture your intention unambiguously and avoids docstrings like "is_parent_path: Check iff a path is a parent of another" ;-)

easybuild/framework/easyblock.py
```diff
@@ -76,15 +76,17 @@
...
 from easybuild.tools.environment import restore_env, sanitize_env
 from easybuild.tools.filetools import CHECKSUM_TYPE_MD5, CHECKSUM_TYPE_SHA256
 from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file, change_dir, check_lock
-from easybuild.tools.filetools import compute_checksum, convert_name, copy_file, create_lock, create_patch_info
-from easybuild.tools.filetools import derive_alt_pypi_url, diff_files, dir_contains_files, download_file
-from easybuild.tools.filetools import encode_class_name, extract_file
-from easybuild.tools.filetools import find_backup_name_candidate, get_source_tarball_from_git, is_alt_pypi_url
-from easybuild.tools.filetools import is_binary, is_sha256_checksum, mkdir, move_file, move_logs, read_file, remove_dir
-from easybuild.tools.filetools import remove_file, remove_lock, verify_checksum, weld_paths, write_file, symlink
+from easybuild.tools.filetools import convert_name, copy_file, copy_dir, create_lock, create_patch_info
```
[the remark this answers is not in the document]
gkaf89: Added all new inclusions in a new line.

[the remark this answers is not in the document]
Flamefire: It looks like the original list of includes was sorted. So I think at least `compute_checksum` should be moved back.

Side note: I guess we should start using parentheses for imports, i.e.
```
from foo import (bar1,...
                          bar2,...
                          )
```

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,90 @@ def print_dry_run_note(loc, silent=True):
     dry_run_msg(msg, silent=silent)
 
 
+def create_persistence_paths(operation_args):
+    persistence_paths = []
+    for op in operation_args:
+        _, _, target_path, _ = op
+        persistence_paths.append(target_path)
```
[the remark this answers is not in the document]
gkaf89: Using list comprehension.

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,90 @@ def print_dry_run_note(loc, silent=True):
...
+
+    persistence_paths = create_unused_dirs(persistence_paths)
+
+    for i in range(len(operation_args)):
+        op = operation_args[i]
+        operation_args[i] = (op[0], op[1], persistence_paths[i], op[3])
```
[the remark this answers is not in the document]
gkaf89: Using list comprehension as suggested.

easybuild/tools/config.py
```diff
@@ -843,6 +847,58 @@ def log_path(ec=None):
...
+
+    if ec is None:
+        raise EasyBuildError("Easyconfig cannot be empty.")
+
+    name, version = ec.get('name'), ec.get('version')
+
+    if name is None:
+        raise EasyBuildError("The 'name' key is not defined.")
+    if version is None:
+        raise EasyBuildError("The 'version' key is not defined.")
```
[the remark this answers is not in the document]
gkaf89: Done!

test/framework/filetools.py
gkaf89: Added a class function `def assertAllExists(self, paths)`.

test/framework/filetools.py
gkaf89: Issue fixed. The tests were reorganized to add more cases.

test/framework/filetools.py
gkaf89: Fixed, now using a loop.

easybuild/tools/filetools.py
```diff
@@ -585,6 +585,13 @@ def normalize_path(path):
     return start_slashes + os.path.sep.join(filtered_comps)
 
 
+def is_predecessor(predecesor, path):
+    predecesor = os.path.normpath(predecesor)
+    path = os.path.normpath(path)
```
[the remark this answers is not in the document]
gkaf89: Yes, I was a bit hesitant to include symlinks, but they should be covered. I modified the code and added a test.

easybuild/tools/filetools.py
```diff
@@ -585,6 +585,13 @@ def normalize_path(path):
     return start_slashes + os.path.sep.join(filtered_comps)
 
 
+def is_predecessor(predecesor, path):
```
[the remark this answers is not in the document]
gkaf89: Done!

easybuild/tools/filetools.py
gkaf89: Issue rectified. Now the non existing directories are determined once before trying to create the descendant paths.

Flamefire:
You can use `autopep8 -i <path/to/file.py>` to automatically do the formatting. This especially helps when the indentation isn't clear from the current context.

easybuild/tools/filetools.py
```diff
@@ -2980,3 +2987,81 @@ def create_unused_dir(parent_folder, name):
...
+    appended for i=0..(index_upper_bound-1) until an index is found so that all required path as unused. All created
+    directories have the same postfix.
+
+    :param paths: list of paths where directories are created; parent path created as well
+    :param index_upper_bound: maximum index that will be tried before failing
+    """
+    paths = list(map(os.path.abspath, paths))
+    for path in paths:
+        for parent in set(paths) - {path}:
+            if is_predecessor(parent, path):
```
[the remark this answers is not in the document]
gkaf89: There was a bug in this section. Should be:
```python
for i_path, path in enumerate(paths):
  for i_parent, parent in enumerate(paths):
    if i_parent != i_path and is_parent_path(parent, path):
      raise EasyBuildError("Path '%s' is a parent path of '%s'" % (parent, path))
```

Added a test that detects the issue.

easybuild/tools/filetools.py
gkaf89: I agree that the test `tried_paths != []` is super flows, test removed.

I will leave the empty list input as is though. The empty list is effectively an identity element for the function, it may appear naturally in some application.

easybuild/tools/filetools.py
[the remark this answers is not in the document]
gkaf89: Updated the documentation strings.

easybuild/tools/filetools.py
gkaf89: I try to avoid `break`. Is it this bad to leave the while loop?

The index upper bound does not have to be exposed for most reasonable use cases. It is convenient however for testing.

easybuild/framework/easyblock.py
```diff
@@ -4198,6 +4207,90 @@ def print_dry_run_note(loc, silent=True):
...
+
+    return operation_args
+
+
+def execute_persistence_operation(operation, source_paths, target_dir, msg, silent):
```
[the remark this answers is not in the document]
gkaf89: I prefer to keep functions small. I changed the name to `execute_and_log_persistence_operation` to signify that it also logs some information.

easybuild/tools/filetools.py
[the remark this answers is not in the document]
Flamefire: Ah so you do want to detect if the same path appears multiple times too.

easybuild/tools/filetools.py
Flamefire: It's probably up to personal preference. My line of though is:
- We have essentially a counting loop, so `for` is more suitable
- `for` avoids having some code path accidentally skip the counter-increment leading to an infinite loop in some corner case
- in other languages it avoids "leaking" the counter variable to outside scope, not in Python though
- `x = False; while not x` looks a bit odd, but that is Python not having do-while loops
- My logic is `for all attempts: try-run_task; if task-succeeded: stop looping`

Maybe some middle ground if you don't want the break in a "hidden place":
```
for number in range(-1, 10000):
    if paths_exist:
        break
```
This avoids the pitfall of missing the increment and should still read similar to `while ... not paths_exist` but split into 2 logical blocks.

Many ideas, all work, you (or in the end the maintainers) decide.

I added a suggestion for your current version which to me is a bit more clear. E.g. `if cond: x = y else: x=z` and moving `tried_paths = []` right before it is filled to make it easier to see that it starts of as empty. You also don't need to check `number` after the loop as `not paths_exist` is only met when the number of tries are exceeded.

gkaf89:
@Flamefire I am designing the final details for the user interface of the new feature (latest commit). I would like the following behavior:

- The defaults are `--log-error-path='log_error_path'` and `--artifact-error-path=''`
- With an empty string input, the files are not preserved.

Is it a reasonable choice? Would the default of `--artifact-error-path` require any extra handling since it is an empty string?

Flamefire:
I'm not sure I fully understand. I would expect the defaults to be the current behavior, i.e. nothing gets copied on failure.    
We already have similar behavior, i.e. to leave the system clean. E.g. `--cleanup-build` is the default and needs to be disabled explicitly if desired. And we have `--tmp-log-dir` which has a very similar effect to your `--error-log-dir    
Do you want the default behavior that logs (on error) get copied into a subfolder of `--prefix`?

As a default for "don't consider this" we have `None`. See the `cuda-cache-dir` option which implements a path option that (only) takes effect when specified.

So I guess what you want is something like

```
            'log-error-path': ("Location where logs are copied in case of an error. "
                               "An empty value disables copying the logs.",
                               None, 'store', mk_full_default_path('log_error_path'), {'metavar': "PATH"}),
            'artifact-error-path': ("Location where artifacts are copied in case of an error, "
                                    "default is to not copy artifacts.",
                                    None, 'store', None, {'metavar': "PATH"}),
```

> Would the default of --artifact-error-path require any extra handling since it is an empty string?

The extra handling would be in the [code that uses it](https://github.com/easybuilders/easybuild-framework/pull/4601/files#diff-d0a4e88105e7ea14327d3e4cfa442197740652cd6682806bb22e070c9d5b707dR882-R883) i.e. `if not path: return None` as you already have. 

I'm not fully sure how the options handle unsetting a default. I.e. how `log-error-path=` in a config file would be interpreted. Similar options I've seen use a type of `None` so that probably works.

gkaf89:
@boegel @Flamefire I have rebased this commit on the 5.0.x, but the tests are failing locally with a complain about `LooseVersion`. Is that expected?

test/framework/toy_build.py
gkaf89: Done in the 5.0.x branch.

gkaf89:
> I'm not sure I fully understand. I would expect the defaults to be the current behavior, i.e. nothing gets copied on failure. We already have similar behavior,...

I am leaving the default behavior as is, i.e. copy nothing.

> As a default for "don't consider this" we have `None`. 

Default values for both options are thus set to `None`.

> I'm not fully sure how the options handle unsetting a default. I.e. how `log-error-path=` in a config file would be interpreted. Similar options I've seen use a type of `None` so that probably works.

Setting the default behavior to "do nothing" resolves this issue.

test/framework/options.py
```diff
@@ -1630,6 +1630,54 @@ def test_dry_run(self):
...
+                args = base_args
+                args.extend(extra_args)
+
+                if is_valid:
+                    try:
+                        self.eb_main(args, raise_error=True)
+                    except EasyBuildError:
+                        self.fail(
+                            "Should not fail with --buildpath='%s' and %s='%s'."
+                            % (build_dir, option_flag, persist_path)
```
Flamefire, line 1671:
```suggestion
                            f"Should not fail with --buildpath='{build_dir}' and {option_flag}='{persist_path}'.")
```

Flamefire (changes requested):
Just some polishing, looks good. :+1: 
I'd suggest to use f-strings consistently, IMO they make it easier to read and safer to use in most cases. Annotated those.

test/framework/options.py
```diff
@@ -1630,6 +1630,54 @@ def test_dry_run(self):
...
+
+                pattern = r"The %s (.*) " % option_flag + \
+                    r"cannot reside on a subdirectory of the --buildpath (.*)"
+
+                args = base_args
+                args.extend(extra_args)
+
+                if is_valid:
+                    try:
+                        self.eb_main(args, raise_error=True)
```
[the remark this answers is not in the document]
gkaf89: The `args` parameter list should contain the `gzip-1.4-GCC-4.6.3.eb`? Or am I reading that wrong?

Regarding the purpose of this call, yes, I want to trigger the parameter validation in a dry run to ensure that we don't fail for a valid directory. Is it excessive?

easybuild/tools/config.py
```diff
@@ -865,6 +869,48 @@ def log_path(ec=None):
...
+    """
+    log_error_path = ConfigurationVariables()['log_error_path']
+
+    if not log_error_path:
+        return None
+
+    if ec is None:
+        raise EasyBuildError("Easyconfig cannot be empty.")
+
+    name, version = ec.get('name'), ec.get('version')
```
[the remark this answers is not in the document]
gkaf89: Sorry, this seems to be a typo created while I migrated to the 5.0.x branch.

Having a `None-None` path would not work, as the `<name>-<version>` path name are needed to separate different software installations.

test/framework/filetools.py
```diff
@@ -3745,6 +3805,229 @@ def test_create_unused_dir(self):
...
+
+        def expected_idx(n):
+            if n == 0:
+                return ""
+            new_idx = n - 1
+            if n >= existing_idx:
+                new_idx += 1
+            return f"_{new_idx}"
+
+        for i in range(1):
```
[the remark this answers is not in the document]
gkaf89: A typo indeed.

I fixed it and I took the opportunity to reduced the maximum index to 3. This should be enough for the logic of the test, and there is another test that goes up to 10 directories. This should save a few ms on the tests.

easybuild/tools/config.py
```diff
@@ -865,6 +869,48 @@ def log_path(ec=None):
...
+    Return the 'log_error_path', the location where logs are copied in case of failure
+
+    :param ec:  dict-like value, cannot be None
+    """
+    log_error_path = ConfigurationVariables()['log_error_path']
+
+    if not log_error_path:
+        return None
+
+    if ec is None:
```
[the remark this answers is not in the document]
gkaf89: Yes, checking for `ec != None` explicitly is excessive, check removed.

easybuild/framework/easyblock.py
```diff
@@ -4411,6 +4413,86 @@ def print_dry_run_note(loc, silent=True):
...
+        return
+
+    datetime_stamp = time.strftime("%Y%m%d") + '-' + time.strftime("%H%M%S")
+    operation_args = []
+
+    log_error_path = get_log_error_path(easyconfig)
+    if log_error_path:
+        log_error_path = os.path.join(log_error_path, datetime_stamp)
+
+        if not is_parent_path(app.builddir, log_error_path):
```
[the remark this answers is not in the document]
gkaf89: Agree, changed the order to remove negation.

easybuild/framework/easyblock.py
```diff
@@ -4411,6 +4413,86 @@ def print_dry_run_note(loc, silent=True):
...
+            operation_args.append(
+                (
+                    copy_file,
+                    logs,
+                    log_error_path,
+                    "Logs of failed build copied to permanent storage: %s" % log_error_path
+                )
+            )
+        else:
+            print_msg(
```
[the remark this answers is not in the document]
gkaf89: Yes, waning seems the appropriate action here. Changed the code to print a warning.

test/framework/toy_build.py
```diff
@@ -281,6 +290,94 @@ def test_toy_broken(self):
...
+        self.assertTrue(len(log_files) == 1, f"Single log file expected: {log_files}")
+
+        return log_file
+
+    def assert_build_files_copied(self, buildpath, artifact_error_path):
+        build_dir = pathlib.Path(buildpath)
+        storage_dir = pathlib.Path(artifact_error_path)
+
+        app_build_dir = build_dir / r"toy/0.0/system-system"
+        iso_date_pattern = r"????????-??????"
```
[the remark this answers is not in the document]
gkaf89: No, I was just being pedantic. I converted them to normal strings as suggested.

easybuild/framework/easyblock.py
```diff
@@ -4411,6 +4413,86 @@ def print_dry_run_note(loc, silent=True):
...
+    if artifact_error_path and os.path.isdir(app.builddir):
+        artifact_error_path = os.path.join(artifact_error_path, datetime_stamp)
+
+        if not is_parent_path(app.builddir, artifact_error_path):
+            operation_args.append(
+                (
+                    lambda source, destination: copy_dir(source, destination, dirs_exist_ok=True),
+                    [app.builddir],
+                    artifact_error_path,
+                    "Artifacts of failed build copied to permanent storage: %s" % artifact_error_path
```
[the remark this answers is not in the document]
gkaf89: Used an f-string for consistency.

test/framework/toy_build.py
```diff
@@ -281,6 +290,94 @@ def test_toy_broken(self):
...
         shutil.rmtree(tmpdir)
 
+    def detect_log_file(self, tmp_log_path):
+        log_files = list(pathlib.Path(tmp_log_path).glob("**/*.log"))
+
+        self.assertTrue(len(log_files) >= 1, 'Log files generated')
+        log_file = log_files[0]
+        self.assertTrue(len(log_files) == 1, f"Single log file expected: {log_files}")
+
+        return log_file
```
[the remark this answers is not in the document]
gkaf89: I am trying to differentiate between different sources of error. If at some point multiple log files are generated, then the logic of the test will need a slight adjustment to handle a list of logs files.

easybuild/base/testing.py
```diff
@@ -124,6 +124,11 @@ def assertNotExists(self, path, msg=None):
             msg = "'%s' should not exist" % path
         self.assertFalse(os.path.exists(path), msg)
 
+    def assertAllExist(self, paths, msg=None):
+        """Assert that all paths in the given list exist"""
+        for p in paths:
+            self.assertExists(p, msg)
```
boegel, line 130: @gkaf89 Style nitpicking: please avoid using single-letter variables outside of list comprehensions (I see this popping up in a couple of places in this PR):

```suggestion
        for path in paths:
            self.assertExists(path, msg)

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/base/testing.py (hunk 56 above)
Comment (gkaf89) [reply easybuild/base/testing.py]: Fixed single character variable names and other minor style issues in a separate commit.

File: test/framework/filetools.py [hunk 57]
```diff
@@ -3745,6 +3805,231 @@ def test_create_unused_dir(self):
...
+        os.mkdir(test_root)
+        existing_idx = 1
+        requested_paths = [
+            os.path.join(test_root, "existing_idx_a"),
+            os.path.join(test_root, "existing_idx_b"),
+        ]
+
+        os.mkdir(os.path.join(test_root, f"existing_idx_b_{existing_idx}"))
+
+        def expected_idx(n_calls_to_create_unused_dirs):
```
Flamefire, line 3859: Better:
```suggestion
        def expected_suffix(n_calls_to_create_unused_dirs):
```

test/framework/toy_build.py
```diff
@@ -281,6 +290,98 @@ def test_toy_broken(self):
...
+
+            self.mock_stderr(True)
+            self.mock_stdout(True)
+            outtxt = self._test_toy_build(
+                ec_file=broken_compilation_ec, tmpdir=tmpdir,
+                verify=False, fails=True, verbose=False, raise_error=False, name='toy', versionsuffix='-buggy',
+                tmp_logdir=tmp_log_dir, log_error_dir=tmp_log_error_dir, artifact_error_dir=tmp_artifact_error_dir
+            )
+            self.mock_stderr(False)
+            self.mock_stdout(False)
```
Flamefire, line 376:
```suggestion
            with self.mocked_stdout_stderr():
                outtxt = self._test_toy_build(
                    ec_file=broken_compilation_ec, tmpdir=tmpdir,
                    verify=False, fails=True, verbose=False, raise_error=False, name='toy', versionsuffix='-buggy',
                    tmp_logdir=tmp_log_dir, log_error_dir=tmp_log_error_dir, artifact_error_dir=tmp_artifact_error_dir
                )
```

Flamefire (approved):
Just did a quick scan as I think everything has been discussed already. Just 2 minor improvements

test/framework/toy_build.py
gkaf89: I merged this change inside the [STYLE] commit.

gkaf89:
After using the commit for a while, I made a minor interface change. Now the report message mentions the actual path to the artifacts, which speeds up the navigation to the artifact location.

Flamefire:
> After using the commit for a while, I made a minor interface change. Now the report message mentions the actual path to the artifacts, which speeds up the navigation to the artifact location.

Please don't force-push in the middle of a review as I can't tell for sure what has changed. Is only the last commit new or did any of the others change as well?

gkaf89:
@Flamefire Sorry, I try to keep the commit history organized. I will try to thing of a way to limit forced pushes next time.

All changes start from the `[STYLE]` commit, older commits were not modified.

Flamefire (approved):

boegel:
@gkaf89 Finally found some time to go through this in detail...

I did some code cleanup, and renamed a bunch of things so it makes a bit more sense.
Most of that is internal, but I also renamed the configuration options to `--failed-installs-build-dirs-path` and `--failed-installs-logs-path`.

Thanks a lot for all the effort on this, especially the extensive tests that you added, that helps a lot to be confident in reviewing & refactoring this a bit!

This should be good to go now, unless I messed up something causing some tests to fail (if so, I'll get them fixed)

easybuild/tools/options.py
```diff
@@ -1218,18 +1224,18 @@ def _postprocess_config(self):
...
 
         for opt_name in path_opt_names:
             self._ensure_abs_path(opt_name)
 
         if self.options.prefix is not None:
             # prefix applies to all paths, and repository has to be reinitialised to take new repositorypath in account
             # in the legacy-style configuration, repository is initialised in configuration file itself
-            path_opts = ['buildpath', 'containerpath', 'installpath', 'packagepath', 'repository', 'repositorypath',
-                         'sourcepath']
+            path_opts = ['buildpath', 'containerpath', 'failed_installs_build_dirs_path', 'failed_installs_logs_path',
```
Flamefire, line 1237: @boegel Does this mean the options are always enabled, especially by default? Not sure how `DEFAULT_PATH_SUBDIRS` is taken into account but this part looks like
- if `--prefix` is set
- if `--failed_installs_build_dirs_path` is not set
- then set it to the default subdir under `--prefix`

I wouldn't store those by default, for some installations the build folder is huge (e.g. just NCCL alone) and even the logs can be very large

boegel: I agree, I still had this in the back of my mind to also change this: using `eb --prefix` shouldn't imply that logs & build dirs are being copied into a subdirectory of the specified prefix.

EasyBuild should be explicitly configured to copy stuff for failing installations, by setting the `failed-installs-build-dirs-path` and/or `--failed-installs-logs-path` configuration options.

I'll add a commit that takes care of that...

boegel: fixed in 480b92bec

boegel: fixed in 6a6eef670

easybuild/framework/easyblock.py
```diff
@@ -4422,6 +4426,70 @@ def print_dry_run_note(loc, silent=True):
...
+
+    operation_args = []
+
+    logs_path = get_failed_installs_logs_path(easyconfig)
+    if logs_path:
+        logs_path = os.path.join(logs_path, unique_subdir)
+
+        if is_parent_path(app.builddir, logs_path):
+            print_warning(
+                "Path to copy log files to for failed installs is subdirectory of build directory; not copying",
```
Flamefire, line 4449:
```suggestion
                "Path to copy log files of failed installs to is subdirectory of build directory; not copying",
```

easybuild/framework/easyblock.py
```diff
@@ -4422,6 +4426,70 @@ def print_dry_run_note(loc, silent=True):
...
+                )
+            )
+
+    build_dirs_path = get_failed_installs_build_dirs_path(easyconfig)
+    if build_dirs_path and os.path.isdir(app.builddir):
+        build_dirs_path = os.path.join(build_dirs_path, unique_subdir)
+
+        if is_parent_path(app.builddir, build_dirs_path):
+            print_warning(
+                "Path to copy build dirs to for failed installs is subdirectory of build directory; not copying",
```
Flamefire, line 4469:
```suggestion
                "Path to copy build dirs of failed installs to is subdirectory of build directory; not copying",
```

easybuild/framework/easyblock.py
```diff
@@ -4422,6 +4426,70 @@ def print_dry_run_note(loc, silent=True):
...
+
+    datestamp = time.strftime('%Y%m%d')
+    timestamp = time.strftime('%H%M%S')
+    salt = ''.join(random.choice(ascii_letters) for i in range(5))
+    unique_subdir = f'{datestamp}-{timestamp}-{salt}'
+
+    operation_args = []
+
+    logs_path = get_failed_installs_logs_path(easyconfig)
+    if logs_path:
```
Flamefire, line 4444: Just to be sure and clear:
```suggestion
    if logs and logs_path:
```

easybuild/framework/easyblock.py
```diff
@@ -4422,6 +4426,70 @@ def print_dry_run_note(loc, silent=True):
...
+                )
+            )
+
+    persistence_paths = [target_path for (_, _, target_path, _) in operation_args]
+    persistence_paths = create_non_existing_paths(persistence_paths)
+
+    for idx, (operation, paths, _, msg) in enumerate(operation_args):
+        for path in paths:
+            operation(path, persistence_paths[idx])
+
```
Flamefire, line 4489: Why the empty line? To me it is easier to see that the print and inner loop are on the same level without that, especially given how short the loop is
```suggestion
```

The same argument holds for the empty line though, but vertical space is more important to me

easybuild/tools/config.py
```diff
@@ -493,6 +493,8 @@ class ConfigurationVariables(BaseConfigurationVariables):
         'buildpath',
         'config',
         'containerpath',
+        'failed_installs_build_dirs_path',
```
Flamefire, line 496: Quite long name. Maybe drop the double plural which sounds a bit odd? E.g. `failed_install_build_dirs_path`
```

File: easybuild/tools/config.py [hunk 65]
```diff
@@ -865,6 +867,46 @@ def log_path(ec=None):
     return log_file_format(return_directory=True, ec=ec, date=date, timestamp=timestamp)
 
 
+def get_failed_installs_build_dirs_path(ec):
+    """
+    Return the 'failed_installs_build_dirs_path',
+    the location where build directories are copied if installation failed
```
Flamefire, line 873: I wouldn't repeat the function name here and only a single dir is returned for the `ec`:
```suggestion
    Return the location where the build directory is copied to if installation failed
```

easybuild/tools/config.py
```diff
@@ -865,6 +867,46 @@ def log_path(ec=None):
...
+    except KeyError:
+        raise EasyBuildError("The 'name' and 'version' keys are required.")
+
+    return os.path.join(failed_installs_build_dirs_path, name + '-' + version)
+
+
+def get_failed_installs_logs_path(ec):
+    """
+    Return the 'failed_installs_logs_path',
+    the location where log files are copied if installation failed
```
Flamefire, line 893:
```suggestion
    Return the location where log files are copied to if installation failed
```

easybuild/tools/config.py
```diff
@@ -865,6 +867,46 @@ def log_path(ec=None):
...
+
+    if not failed_installs_build_dirs_path:
+        return None
+
+    try:
+        name, version = ec['name'], ec['version']
+    except KeyError:
+        raise EasyBuildError("The 'name' and 'version' keys are required.")
+
+    return os.path.join(failed_installs_build_dirs_path, name + '-' + version)
```
Flamefire, line 887:
```suggestion
    return os.path.join(failed_installs_build_dirs_path, f'{name}-{version}')
```
This way it won't fail for `None` which seemingly is possible according to previous discussions

easybuild/tools/config.py
```diff
@@ -865,6 +867,46 @@ def log_path(ec=None):
...
+
+    if not failed_installs_logs_path:
+        return None
+
+    try:
+        name, version = ec['name'], ec['version']
+    except KeyError:
+        raise EasyBuildError("The 'name' and 'version' keys are required.")
+
+    return os.path.join(failed_installs_logs_path, name + '-' + version)
```
Flamefire, line 907:
```suggestion
    return os.path.join(failed_installs_logs_path, f'{name}-{version}')
```

easybuild/tools/config.py
```diff
@@ -865,6 +867,46 @@ def log_path(ec=None):
...
+def get_failed_installs_build_dirs_path(ec):
+    """
+    Return the 'failed_installs_build_dirs_path',
+    the location where build directories are copied if installation failed
+
+    :param ec:  dict-like value with 'name' and 'version' keys defined
+    """
+    failed_installs_build_dirs_path = ConfigurationVariables()['failed_installs_build_dirs_path']
+
+    if not failed_installs_build_dirs_path:
```
Flamefire, line 879: How about making it a bit shorter and clear that this is not the final path yet?
```suggestion
    base_path = ConfigurationVariables()['failed_installs_build_dirs_path']
    if not base_path:
```

easybuild/tools/filetools.py
```diff
@@ -3096,3 +3109,69 @@ def create_unused_dir(parent_folder, name):
...
+
+    first_non_existing_parent_paths = [get_first_non_existing_parent_path(p) for p in paths]
+
+    non_existing_paths = paths
+    all_paths_created = False
+    suffix = -1
+    while suffix < max_tries and not all_paths_created:
+        tried_paths = []
+        if suffix >= 0:
+            non_existing_paths = [f'{p}_{suffix}' for p in paths]
```
Flamefire, line 3153: Maybe this is clearer ?
```suggestion
        if suffix < 0:
            non_existing_paths = paths
        else:
            non_existing_paths = [f'{p}_{suffix}' for p in paths]
```

Flamefire (commented):
Some final suggestions, maybe @boegel can check which make sense.   
Looks good to me with or without.

easybuild/framework/easyblock.py
```diff
@@ -4422,6 +4426,70 @@ def print_dry_run_note(loc, silent=True):
...
+def copy_build_dirs_logs_failed_installs(application_log, silent, app, easyconfig):
+    """
+    Copy build directories and log files for failed installation (if desired)
+    """
+    # there may be multiple log files, or the file name may be different due to zipping
+    logs = glob.glob(f"{application_log}*")
+
+    datestamp = time.strftime('%Y%m%d')
+    timestamp = time.strftime('%H%M%S')
+    salt = ''.join(random.choice(ascii_letters) for i in range(5))
```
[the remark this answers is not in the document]
boegel: In theory you may have multiple jobs running the same installation (on different partitions, for example), and failing at the exact same time.
There's no coordination between these jobs, so there's no way to avoid a race condition.

This extra salt part essentially makes it impossible in practice to have a clash in target directory, and it doesn't hurt (subdirectories will still sort nicely by datestamp + timestamp)

easybuild/tools/config.py
```diff
@@ -865,6 +867,46 @@ def log_path(ec=None):
...
+def get_failed_installs_logs_path(ec):
+    """
+    Return the 'failed_installs_logs_path',
+    the location where log files are copied if installation failed
+
+    :param ec:  dict-like value with 'name' and 'version' keys defined
+    """
+    failed_installs_logs_path = ConfigurationVariables()['failed_installs_logs_path']
+
+    if not failed_installs_logs_path:
```
boegel, line 899:
```suggestion
    base_path = ConfigurationVariables()['failed_installs_logs_path']

    if not base_path:

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/tools/filetools.py (hunk 70 above)
Comment (boegel) [reply easybuild/tools/filetools.py:3153]: Nah, I think it's fine as is, the `< 0` case is handled before the while loop starts

File: easybuild/tools/config.py (hunk 64 above)
Comment (boegel) [reply easybuild/tools/config.py]: It's long, but I like that they're grouped together (in output of `--help`, etc.)

I've renamed the options to avoid double plural though, see f63a74379

File: easybuild/framework/easyblock.py (hunk 71 above)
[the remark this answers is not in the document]
Comment (boegel) [reply easybuild/framework/easyblock.py]: suggestion applied in c50dc4e6ea42ddb55f090b6c95402907eab1da03

File: easybuild/framework/easyblock.py [hunk 73]
```diff
@@ -4422,6 +4426,70 @@ def print_dry_run_note(loc, silent=True):
...
 
+def copy_build_dirs_logs_failed_installs(application_log, silent, app, easyconfig):
+    """
+    Copy build directories and log files for failed installation (if desired)
+    """
+    # there may be multiple log files, or the file name may be different due to zipping
+    logs = glob.glob(f"{application_log}*")
+
+    datestamp = time.strftime('%Y%m%d')
+    timestamp = time.strftime('%H%M%S')
```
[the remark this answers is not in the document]
boegel: fixed in c50dc4e6ea42ddb55f090b6c95402907eab1da03

easybuild/framework/easyblock.py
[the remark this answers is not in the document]
Flamefire: But the race condition is fully avoided by `create_non_existing_paths` isn't it? Isn't that the purpose of the function? That will either create the directories with that exact timestamp or with `_1/2/3` appended until one is found that doesn't exist yet

boegel (approved):

PR #4603, opened 2024-08-06
improve error messages for empty EasyConfigs

Flamefire:
Currently we get:

> ERROR: Failed to process easyconfig foo.eb: No software-specific easyblock 'EB_UNKNOWN' found for UNKNOWN

That isn't exactly clear. So improve in 2 steps:

1. Remove the UNKNOWN constant and just handle in the way we would if the class isn't found: Return `None` or error if `error_on_missing_easyblock` is set
2. Check for a fully empty file and error with a clear message

Possible improvement: Pass `error_on_missing_easyblock` in the EasyConfig constructor and use an empty dict as `extra_options` if no `easyblock_class` was found, i.e. `self.extra_options = easyblock_class.extra_options() if easyblock_class else {}`
This will error later in the parser which doesn't find the required keys like "name".   
Not sure if that error ("missing required keys: name, ...") is cleaner than "No easyblock found as neither name nor easyblock were specified" though.

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -455,6 +455,8 @@ class EasyConfig(object):
             self.path = path
             self.rawtxt = read_file(path)
             self.log.debug("Raw contents from supplied easyconfig file %s: %s", path, self.rawtxt)
+            if not self.rawtxt.strip():
+                raise EasyBuildError('Easyconfig file is empty')
         else:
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm

boegel:
I think "`No easyblock found as neither name nor easyblock were specified`" is a pretty clear error, it's more specific to me than letting it fail later because mandatory easyconfig parameters are not specified.

Thanks a lot for this enhancement @Flamefire!

PR #4604, opened 2024-08-06
Don't try to overwrite an existing file if the contents are identical

ocaisa:


hattom:
Note that file permissions are not compared as part of `filecmp.cmp`:
```
> ls -althr file?
-rwxr-xr-x 1 thay grp-tok-ipp 12 Aug 13 13:15 file1
-rw-r--r-- 1 thay grp-tok-ipp 12 Aug 13 13:15 file2
> cat test_cmp.py 
#!/usr/bin/env python3
import filecmp
print(filecmp.cmp("file1", "file2"))
> python test_cmp.py 
True
```

Could this cause problems?

boegel:
> Note that file permissions are not compared as part of `filecmp.cmp`:
> 
> ```
> > ls -althr file?
> -rwxr-xr-x 1 thay grp-tok-ipp 12 Aug 13 13:15 file1
> -rw-r--r-- 1 thay grp-tok-ipp 12 Aug 13 13:15 file2
> > cat test_cmp.py 
> #!/usr/bin/env python3
> import filecmp
> print(filecmp.cmp("file1", "file2"))
> > python test_cmp.py 
> True
> ```
> 
> Could this cause problems?

This is a very good point imho, could result in unintended side effects/breakage?

easybuild/tools/filetools.py
```diff
@@ -2385,6 +2386,9 @@ def copy_file(path, target_path, force_in_dry_run=False):
             target_exists = os.path.exists(target_path)
             if target_exists and path_exists and os.path.samefile(path, target_path):
                 _log.debug("Not copying %s to %s since files are identical", path, target_path)
+            # Don't copy them if they are already considered the same
+            elif target_exists and path_exists and os.path.isfile(target_path) and filecmp.cmp(path, target_path):
```
boegel, line 2390: At the very least we should also check whether file permissions are the same as well before skipping the actual copy operation?

boegel (changes requested):

easybuild/tools/filetools.py
akesandgren: Note that the permissions may have been changed later on by postinstallcmds or hooks or the normal EB change-permission step and thus permission differences should not be considered if the content is actually the same.

ocaisa:
Closing this in favour of #4642 which I think is the real problem

PR #4605, opened 2024-08-07
handle post-install patches in `check_checksums_for`

Flamefire:
Checksums for post-install patches come after the patch checksums, see `fetch_patches`. As those were not included in the check specifying a checksum for those patches leads to errors such as
> found 0 sources + 0 patches vs 1 checksums


boegel (approved):
lgtm

PR #4607, opened 2024-08-08
Cleanup after removing Python 2 support

Flamefire:
I cleaned up the code that supports 2.6 which means removal of OrderedDict as well as some more unittest/CI code

Rebase of #3788 for 5.0x


boegel (approved):
lgtm

PR #4608, opened 2024-08-09
Add support for `--software-commit` and an associated template `%(software_commit)s`

ocaisa:
Basic idea here is to provide minimal support to application developers. They can create a development easyconfig which leverages the `software_commit` template (most likely using a git repo as their sources). This would then allow you to use EasyBuild in CI (with the help of EESSI of course!)


boegel (approved):
lgtm

PR #4609, opened 2024-08-12
fix `copy_file` with a folder as the target

Flamefire:
This use case comes up with e.g. `--copy-ec /tmp`.

The existing code will see that `/tmp` exists and has a different UID and hence will use `shutil.copyfile` which expects the target to not be a folder. It hence fails with "[Errno 21] Is a directory"

Check for that case and make sure the target is always a file, not a directory.


boegel (approved):
lgtm

PR #4610, opened 2024-08-13
improve help string for `--dep-graph`

Flamefire:
The current one is not clear enough as to what `<ext>` is supposed to be. So provide some examples, especially `.dot` which doesn't require `graphviz-python`


boegel (approved):
lgtm

PR #4611, opened 2024-08-16
let `cmd.sh` script produced by `run_shell_cmd` define `$EB_CMD_OUT_FILE` + `$EB_CMD_ERR_FILE`

boegel:
`$EB_CMD_OUT_FILE` is useful to let people easily found the output that was generated by a shell command, since the path to the file containing it is not easy to find


Micket (approved):
lgtm

Micket:
tested it locally, seems fine

PR #4613, opened 2024-08-16
add `qa_check_interval` option to `run_shell_cmd`

boegel:
required for WIEN2k, see also https://github.com/easybuilders/easybuild-easyblocks/pull/3413


Micket (approved):
lgtm

Micket:
Not sure if the problems are due to this PR, but i'll just hold off with merging until a successfull wien2k and abaqus build.

boegel:
I'll close this, since the use case I had in mind (WIEN2k easyblock) is invalid, turns out that there's a bug in `run_shell_cmd` that's causing trouble when installing WIEN2k, which is being fixed in #4617

PR #4614, opened 2024-08-18
Use `getenv` modulefile command with `EnvironmentModules`

xdelaruelle:
Update "getenv_cmd" function from ModuleGeneratorTcl class to rely on "getenv" modulefile command on EnvironmentModules. "getenv" returns environment variable value or a fallback value if it does not exist. No error is thus raised if variable does not exist compared to using "$::env(VARNAME)" syntax.

"getenv" modulefile command also has the advantage to return "$VARNAME" rather VARNAME value if modulefile is evaluated in "display" mode. Which helps to understand that an environment variable is relying on another one.

Note that Lmod introduced a "getenv" Tcl modulefile command in version 8.7.31 but it does not handle fallback value at the moment. So the use of "getenv" with Lmod is not enabled at the moment, but proposed code in this change makes it easy to enable such support in the future.

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -863,13 +863,19 @@ class ModuleGeneratorTcl(ModuleGenerator):
         Return module-syntax specific code to get value of specific environment variable.
         """
         if default is None:
-            cmd = '$::env(%s)' % envvar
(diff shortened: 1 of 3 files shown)


boegel (approved):
lgtm

PR #4615, opened 2024-08-18
Add module cache build support on EnvironmentModules

xdelaruelle:
Environment Modules 5.3.0 introduced module cache mechanism. A cache file is generated with cachebuild command and stored under modulepath directory in a .modulecache file.

Add cache build support in "update" function of EnvironmentModules class.

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1419,6 +1419,16 @@ class EnvironmentModules(EnvironmentModulesTcl):
 
         return value
 
+    def update(self):
+        """Update after new modules were added."""
+
+        version = LooseVersion(self.version)
+        if build_option('update_modules_tool_cache') and version >= LooseVersion('5.3.0'):
+            out = self.run_module('cachebuild', return_stderr=True, check_output=False)
+
+            if self.testing:
+                return out
+
 
 class Lmod(ModulesTool):
     """Interface to Lmod."""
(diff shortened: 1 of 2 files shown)


boegel:
@xdelaruelle Can you look into resolving the merge conflict which was introduced after merging of #4614?

easybuild/tools/modules.py
```diff
@@ -1413,6 +1417,15 @@ def get_setenv_value_from_modulefile(self, mod_name, var_name):
 
         return value
 
+    def update(self):
+        """Update after new modules were added."""
+
+        if build_option('update_modules_tool_cache') and self.supports_module_cache:
```
boegel, line 1423: No need to introduce `self.supports_module_cache` or `REQ_VERSION_MODULE_CACHE`  I think, just include the version check here directly?

The `self.supports_*` mechanism is mainly useful when `module_generator` needs to figure out whether the active modules tool supports something, but there's no need for that indirection here...

```python
        if build_option('update_modules_tool_cache') and version >= LooseVersion('5.3.0'):
```

boegel (changes requested):

easybuild/tools/modules.py
xdelaruelle: Thanks for your feedback. I have revised the code to take your suggestion into account.

boegel (approved):
lgtm

PR #4617, opened 2024-08-20
fix `run_shell_cmd` to avoid that same question is answered multiple times

boegel:
The changes made in `run_shell_cmd` are required for passing the extra test case that was added

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -489,6 +489,7 @@ def run_shell_cmd(cmd, fail_on_error=True, split_stderr=False, stdin=None, env=N
         stdout, stderr = b'', b''
         check_interval_secs = 0.1
         time_no_match = 0
+        prev_stdout = ''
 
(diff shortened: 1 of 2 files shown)


boegel:
Test report coming up in https://github.com/easybuilders/easybuild-easyblocks/pull/3413 shows that this works to install WIEN2k using `run_shell_cmd`

easybuild/tools/run.py
```diff
@@ -514,8 +515,15 @@ def to_cmd_str(cmd):
                     stderr += more_stderr
 
             if qa_patterns:
-                if _answer_question(stdout, proc, qa_patterns, qa_wait_patterns):
+                # only check for question patterns if additional output is available
+                # compared to last time a question was answered;
+                # use empty list of question patterns iif no extra output (except for whitespace) is available
```
branfosj, line 520:
```suggestion
                # use empty list of question patterns if no extra output (except for whitespace) is available
```

branfosj (approved):

PR #4620, opened 2024-08-30
only call `_sanity_check_step_extensions` if `--skip-extensions` is not set

Flamefire:
Derived easyblocks might want to override this method to add own checks. By moving the check outside the call it doesn't need to be repeated in derived classes.


boegel (approved):
lgtm

PR #4621, opened 2024-08-30
restore `%(software_commit)s` in `TEMPLATE_NAMES_DYNAMIC`

boegel:
Only impact of this is that `%(software_commit)s` is not included in output of "`eb --avail-easyconfig-templates`", using the template itself actually still works as verified by `test_software_commit_template` (becuase the corresponding values is set in `template_constant_dict` regardless of whether or not it's listed in `TEMPLATE_NAMES_DYNAMIC`)


Micket (approved):
lgtm

PR #4623, opened 2024-09-04
Remove a superflous check in `EasyBlock.run_all_steps`

Flamefire:
Before checking for string equality we don't need another check on it. `None != str` for any string `str`


boegel (approved):
lgtm

PR #4624, opened 2024-09-04
move verifying of checksums from `source` to `fetch` step, to include it with `--fetch`

Flamefire:
After `--fetch` finishes it is reasonable to expect that the build won't fail due to missing or wrong sources.
However putting the checksum-step into the source-step skips over the verification and one would need to use `--stop=source` which does a lot more than necessary, e.g. creating build dirs and environment variables.

Move the checksum step into the fetch step and add a test for that.

TODO: We should rename `SOURCE_STEP` to `EXTRACT_STEP`, do that here too?

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -4038,15 +4038,6 @@ class EasyBlock(object):
             return get_step(READY_STEP, "creating build dir, resetting environment", ready_substeps, False,
                             initial=initial)
 
-        source_substeps = [
-            (False, lambda x: x.checksum_step),
-            (True, lambda x: x.extract_step),
-        ]
-
-        def source_step_spec(initial):
-            """Return source step specified."""
-            return get_step(SOURCE_STEP, "unpacking", source_substeps, True, initial=initial)
-
         install_substeps = [
             (False, lambda x: x.stage_install_step),
             (False, lambda x: x.make_installdir),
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm

boegel:
> TODO: We should rename SOURCE_STEP to EXTRACT_STEP, do that here too?

@Flamefire Not sure if we should, `source` step still makes more sense to me (extracting of *sources*), but if you want to put that up for discussion, yes let's do so in a separate PR

Flamefire:
> > TODO: We should rename SOURCE_STEP to EXTRACT_STEP, do that here too?
> 
> @Flamefire Not sure if we should, `source` step still makes more sense to me (extracting of _sources_), but if you want to put that up for discussion, yes let's do so in a separate PR

Well now we have this (`extract_step_spec` is a new name from me):

> ` extract_step_spec = (SOURCE_STEP, "unpacking", [lambda x: x.extract_step], True)`

"source" is a bit ambiguous/too general: It could be related to anything: fetching, verifying, extracting, ... and other steps are named after their called methods too.

The change is for a separate PR but IMO we can discuss/decide this here already where it becomes an "issue" due to the new logic.

boegel:
> > > TODO: We should rename SOURCE_STEP to EXTRACT_STEP, do that here too?
> > 
> > 
> > @Flamefire Not sure if we should, `source` step still makes more sense to me (extracting of _sources_), but if you want to put that up for discussion, yes let's do so in a separate PR
> 
> Well now we have this (`extract_step_spec` is a new name from me):
> 
> > ` extract_step_spec = (SOURCE_STEP, "unpacking", [lambda x: x.extract_step], True)`
> 
> "source" is a bit ambiguous/too general: It could be related to anything: fetching, verifying, extracting, ... and other steps are named after their called methods too.
> 
> The change is for a separate PR but IMO we can discuss/decide this here already where it becomes an "issue" due to the new logic.

We use source because that step was about verifying checksums of sources (+ patches) + unpacking the sources.
Now it's only about unpacking, but changing the publicly facing name of a step is quite intrusive, many people are used to using `eb --stop source`.

If we want come up with a way of deprecating the use of `source`, and come up with a non-ambiguous alternative (`extract` is perhaps too vague), then I'm happy to follow up in another PR

PR #4439, opened 2024-01-15
deprecate support for `EnvironmentModulesC` and `EnvironmentModulesTcl` module tools

branfosj:

PR #4625, opened 2024-09-07
derive `EnvironmentModules` class directly from `ModulesTool` rather than from to be deprecated `EnvironmentModulesTcl`

xdelaruelle:
As `EnvironmentModulesC` and `EnvironmentModulesTcl` are deprecated in upcoming EB 5.0, some things could be cleaned from the framework code: references to these classes or workaround when handling module tool or generating module files.

### DIFF
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -735,7 +735,7 @@ def container_path():
 
 def get_modules_tool():
     """
-    Return modules tool (EnvironmentModulesC, Lmod, ...)
+    Return modules tool (EnvironmentModules, Lmod, ...)
     """
     # 'modules_tool' key will only be present if EasyBuild config is initialized
     return ConfigurationVariables().get('modules_tool', None)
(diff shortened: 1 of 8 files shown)


xdelaruelle:
As discussed with @boegel, `EnvironmentModulesC` and `EnvironmentModulesTcl` are made deprecated on EB5, but they can still be used (with a big warning message). So I need to revise the commits of this PR not to remove code still useful for these 2 classes.

xdelaruelle:
Code has been clarified to better understand if workarounds/issues affect `EnvironmentModules` or the older classes. `EnvironmentModules` now directly inherits from `ModulesTool`. With all of that it will be easy in the future to remove the deprecated `EnvironmentModulesC` and `EnvironmentModulesTcl` classes.

boegel (approved):
lgtm

boegel:
@xdelaruelle `EnvironmentModulesC` and `EnvironmentModulesTcl` are not actually marked as deprecated yet, but we should deprecate them.

That should boil down to a well-aimed `self.log.deprecated` call in the `__init__`, are you up for looking into thar?

boegel:
> @xdelaruelle `EnvironmentModulesC` and `EnvironmentModulesTcl` are not actually marked as deprecated yet, but we should deprecate them.
> 
> That should boil down to a well-aimed `self.log.deprecated` call in the `__init__`, are you up for looking into that?

@xdelaruelle nevermind, already done ,see #4439

PR #4629, opened 2024-09-11
Rename '`source`' step to '`extract`' (affects `skipsteps` easyconfig parameter + `--stop` option)

### LINKED ISSUE #4627: Rename `source` step
This is a follow-up discussion to https://github.com/easybuilders/easybuild-framework/pull/4624

That PR moved the verification of the checksums for sources and patches to the fetch-step.

That results in the "source"-step to only extract the sources.

We have for (almost) all steps now a 1:1 correspondence of (user-facing) step names and their method in the `EasyBlock` class: https://github.com/easybuilders/easybuild-framework/blob/e0e04fb773a5cee26669211f563696da94455cfd/easybuild/framework/easyblock.py#L4043-L4048

However for the (new) "source"-step this isn't true any longer:
https://github.com/easybuilders/easybuild-framework/blob/8569344a85295f691d30fac75e2c7611c478d18d/easybuild/framework/easyblock.py#L4054

Note that even prior to the change the status (printed to stdout) was "unpacking":
```
== fetching files...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== testing...
== installing...
```

It could be argued that "source" (for use in e.g. `--stop=source`) was fitting because it verified the sources (&patches) and then extracted them, even though the former wasn't explicitly indicated.

However after the change I think "source" is no longer appropriate and we should follow the method names at least for consistency and rename this step to "extract". I'd argue that `--stop=extract` is as clear as `--stop=configure` and similar.

An alternative would be "unpack" to match the status-message although it then won't match the method name `extract_step`

This change can only be for 5.x as it is a breaking change and the above change is only for that branch.

I'm pretty sure we already have a check for valid step names, hence any use of the old name "source" would yield a clear message that it is an invalid value and to choose from the valid set of values. If not we should enhance that error message to do that.

With that the change will only cause some required change in habits for people doing `--stop=source`. This was however often just a workaround for `--fetch` not verifying the checksums, so there is a better alternative for that.

Theoretically there could be easyconfigs using `skipsteps=["source"]` although I can hardly imagine a reason for that. We currently have 4 such official ECs in the `__archive__` and I'm not sure that was ever really OK as e.g. it meant that the checksums were not verified.

No current official ECs use this.

Requesting input from @boegel, possibly putting this up on the next confcall agenda.

Flamefire:
Adjust tests to take that into account and add tests to check that the
old name causes a proper error.
Add such error handling for the `skipsteps` EC parameter.

Split up `test_skip` into more focused sub tests to allow adding the new (negative/failure) test of skipsteps to one of those.

Also a fix for `test_compiler_cache` and a bit cleanup to avoid copying the same superflous parameters in the new tests.

Fixes #4627


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -858,9 +858,21 @@ def validate(self, check_osdeps=True):
...
+        if unknown_step_names:
+            error_lines = ["Found one or more unknown step names in skipsteps:"]
+            for step in unknown_step_names:
+                error_lines.append("* %s" % step)
+                # try to find close match, may be just a typo in the hook name
+                close_matches = difflib.get_close_matches(step, STEP_NAMES, 2, 0.8)
+                if close_matches:
+                    error_lines[-1] += " (did you mean %s?)" % ', or '.join("'%s'" % s for s in close_matches)
+            raise EasyBuildError('\n'.join(error_lines))
+
```
boegel, line 875: To fix failing code style check:
```
./easybuild/framework/easyconfig/easyconfig.py:877:9: E303 too many blank lines (2)
```

```suggestion
```

boegel (changes requested):

boegel:
@Flamefire Merge conflict to fix, probably a trivial one...

Flamefire:
Jep, one of the test fixes I had opened a[ separate PR](https://github.com/easybuilders/easybuild-framework/pull/4599) for

boegel (approved):

PR #4632, opened 2024-09-12
Remove trailing dots from backup message produced by `--inject-checksums`

Flamefire:
The dots usually indicate some upcoming process or shortened information. Neither is true at this point: The backup file *has* been saved. Removing the dots makes it also easier to C&P the resulting name as the selection will no longer automatically contain the dots.

This has bothered me many times now as I do not want the backup file in the first place and a double click on the output selected the dots so I couldn't C&P the filename to the `rm ` command


boegel (approved):
lgtm

PR #4634, opened 2024-09-13
Drop support for pep8 package

Flamefire:
It is outdated for many years now and has been replaced by pycodestyle. Stop using it.


boegel:
@Flamefire Although very soon it won't matter anymore (because we'll collapse the `5.0.x` branch into `develop`), this should target the `5.0.x` currently/..

Flamefire:
Ok changed and extract #4644 for develop

PR #4641, opened 2024-09-16
Add support for GCC optimized builds in RISCV SiFive u74mc architecture

julianmorillo:


easybuild/toolchains/compiler/gcc.py
```diff
@@ -96,6 +96,7 @@ class Gcc(Compiler):
         (systemtools.POWER, systemtools.POWER_LE): 'mcpu=native',
         (systemtools.X86_64, systemtools.AMD): 'march=native',  # implies -mtune=native
         (systemtools.X86_64, systemtools.INTEL): 'march=native',  # implies -mtune=native
+        (systemtools.RISCV64, systemtools.SIFIVE): 'march=rv64gc mtune=sifive-7-series',  # flags specified in archspec
```
boegel, line 99: @julianmorillo Hmmm, is `systemtools.SIFIVE` fine-grained enough here?

julianmorillo: @boegel Probably not, if you consider all the possible `mtune `options in: https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html.
But, on the other hand, this is what it is currently being used in **archspec**: https://github.com/archspec/archspec-json/blob/b0c1872de59fe8c9902f55f97699d1b02b5cfe9c/cpu/microarchitectures.json#L3461
Also, there is no way to obtain more fine-grained information on such systems (at least not on the arriesgado machines I have been testing). 
So I believe that for the moment this granularity should suffice.

boegel: In that case, it's probably not a good idea to add this to EasyBuild by default, since if there are other SiFive models later, we would need to change/fix this somehow anyway?

Note that you can configure EasyBuild easily to use a particular set of flags instead of what it uses by default via `--optarch` (or `$EASYBUILD_OPTARCH`), see https://docs.easybuild.io/controlling-compiler-optimization-flags/#controlling_compiler_optimization_flags_optarch_flags

julianmorillo: @boegel As you requested, the detection is now more fine-grained. The compiler flags are only applied when we are in a SiFive U74-MC processor.

PR #4642, opened 2024-09-16
implement workaround for permission error when copying read-only files that have extended attributes set and using Python 3.6

jfgrimm:
Ran into this issue like this, when configuring EasyBuild with read-only-installdir:
```
== 2024-09-16 13:26:56,117 build_log.py:171 ERROR EasyBuild crashed with an error (at easybuild/base/exceptions.py:126 in __init__): Failed to copy file /mnt/scratch/users/jfg508/eb/software/EasyBuild/4.9.2/lib/python3.6/site-packages/easybuild/easyblocks/e/easybuildmeta.py to /tmp/eb-l7y97vxy/reprod_20240916132656_823857/easyblocks/easybuildmeta.py: [Errno 13] Permission denied: '/tmp/eb-l7y97vxy/reprod_20240916132656_823857/easyblocks/easybuildmeta.py' (at easybuild/tools/filetools.py:2437 in copy_file)
```

Having taken a closer look, it seems that on our system `shutil.copy2()` spits out a `PermissionError` when the file to be copied is read-only. However, the copy still succeeds (and has the correct contents).

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2435,8 +2435,15 @@ def copy_file(path, target_path, force_in_dry_run=False):
             else:
                 mkdir(os.path.dirname(target_path), parents=True)
                 if path_exists:
-                    shutil.copy2(path, target_path)
-                    _log.info("%s copied to %s", path, target_path)
+                    try:
+                        # on at least some systems, copying read-only files with shutil.copy2() spits a PermissionError
+                        # but the copy still succeeds
+                        shutil.copy2(path, target_path)
+                        _log.info("%s copied to %s", path, target_path)
+                    except PermissionError as err:
+                        # double-check whether the copy actually succeeded
+                        if not os.path.exists(target_path):
+                            raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
                 elif os.path.islink(path):
                     if os.path.isdir(target_path):
                         target_path = os.path.join(target_path, os.path.basename(path))
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2444,6 +2444,7 @@ def copy_file(path, target_path, force_in_dry_run=False):
                         # double-check whether the copy actually succeeded
                         if not os.path.exists(target_path):
                             raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
+                        _log.info("%s copied to %s ignoring permissions error: %s", path, target_path, err)
                 elif os.path.islink(path):
                     if os.path.isdir(target_path):
                         target_path = os.path.join(target_path, os.path.basename(path))
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2438,20 +2438,25 @@ def copy_file(path, target_path, force_in_dry_run=False):
                 mkdir(os.path.dirname(target_path), parents=True)
                 if path_exists:
                     try:
-                        # on at least some systems, copying read-only files with shutil.copy2() spits a PermissionError
-                        # but the copy still succeeds
+                        # on filesystems that support extended file attributes, copying read-only files with
+                        # shutil.copy2() will give a PermissionError, when using Python < 3.7
+                        # see https://bugs.python.org/issue24538
                         shutil.copy2(path, target_path)
                         _log.info("%s copied to %s", path, target_path)
                     except OSError as err:
                         # catch the more general OSError instead of PermissionError, since python2 doesn't support
                         # PermissionError
-                        if LooseVersion(platform.python_version()) >= LooseVersion('3'):
+                        pyver = LooseVersion(platform.python_version())
+                        if pyver >= LooseVersion('3.7'):
+                            raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
+                        elif LooseVersion('3.7') > pyver >= LooseVersion('3'):
                             if not isinstance(err, PermissionError):
                                 raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
                         # double-check whether the copy actually succeeded
                         if not os.path.exists(target_path) or not filecmp.cmp(path, target_path, shallow=False):
                             raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
-                        _log.info("%s copied to %s, ignoring permissions error: %s", path, target_path, err)
+                        msg = "%s copied to %s, ignoring permissions error (likely due to https://bugs.python.org/issue24538): %s"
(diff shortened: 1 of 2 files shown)


ocaisa:
I think this is the real problem that I thought I was solving with https://github.com/easybuilders/easybuild-framework/pull/4604

easybuild/tools/filetools.py
```diff
@@ -2435,8 +2435,15 @@ def copy_file(path, target_path, force_in_dry_run=False):
...
-                    _log.info("%s copied to %s", path, target_path)
+                    try:
+                        # on at least some systems, copying read-only files with shutil.copy2() spits a PermissionError
+                        # but the copy still succeeds
+                        shutil.copy2(path, target_path)
+                        _log.info("%s copied to %s", path, target_path)
+                    except PermissionError as err:
+                        # double-check whether the copy actually succeeded
+                        if not os.path.exists(target_path):
+                            raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
```
ocaisa, line 2446:
```suggestion
                        if not os.path.exists(target_path):
                            raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
                        _log.info("%s copied to %s ignoring permissions error: %s", path, target_path, err)
```

easybuild/tools/filetools.py
```diff
@@ -2435,8 +2435,16 @@ def copy_file(path, target_path, force_in_dry_run=False):
...
-                    shutil.copy2(path, target_path)
-                    _log.info("%s copied to %s", path, target_path)
+                    try:
+                        # on at least some systems, copying read-only files with shutil.copy2() spits a PermissionError
+                        # but the copy still succeeds
+                        shutil.copy2(path, target_path)
+                        _log.info("%s copied to %s", path, target_path)
+                    except PermissionError as err:
+                        # double-check whether the copy actually succeeded
+                        if not os.path.exists(target_path):
```
boegel, line 2445: Hmm, it's not really sufficient if it exists though, contents could be different?

boegel (changes requested):

easybuild/tools/filetools.py
jfgrimm: good point, I've amended that to: `if not os.path.exists(target_path) or not os.path.samefile(path, target_path):`

ocaisa: Is that the right option? That's still not checking the contents I don't believe, for that I thought you need `filecmp.cmp(a, b, shallow=False)` (after soke homework in my other PR)

jfgrimm: ugh, you're right
I'll fix that once the python2 framework test has finished, so I can see if that at least works

ocaisa:
To me this now looks good, but can we add a test for it? It may not be very thorough, but you should just be able to create a file, change the permissions to read-only and then attempt a copy. I don't know if that is guaranteed to hit the LoC, but it is worth a shot. Should be a variation on 
https://github.com/easybuilders/easybuild-framework/blob/f7035e9b2ad658ab64f9edce7915be441301c934/test/framework/filetools.py#L1805-L1810

jfgrimm:
I have traced the underlying issue to: https://bugs.python.org/issue24538#msg301493

So this should only appear in python < 3.7

jfgrimm:
since the `_copyxattr` issue is filesystem-dependent, and the issue was fixed in python 3.7+, I don't know that it's of much value adding a test?

ocaisa (approved):

ocaisa:
Like everything, this was more work than it seemed initially!

jfgrimm:
@ocaisa I've pushed another commit (or two) that actually use the upstream approach to ensure the extended attributes are truly copied as well. A cleaner fix imo

easybuild/tools/filetools.py
```diff
@@ -2435,8 +2437,42 @@ def copy_file(path, target_path, force_in_dry_run=False):
...
+
+                        # double-check whether the copy actually succeeded
+                        if not os.path.exists(target_path) or not filecmp.cmp(path, target_path, shallow=False):
+                            raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
+
+                        try:
+                            # re-enable user write permissions in target, copy xattrs, then remove write perms again
+                            adjust_permissions(target_path, stat.S_IWUSR)
+                            shutil._copyxattr(path, target_path)
+                            adjust_permissions(target_path, stat.S_IWUSR, add=False)
```
ocaisa, line 2468: Ha, this is correct but quite confusing (`add=False` means remove the permissions)

ocaisa (dismissed):

easybuild/tools/filetools.py
jfgrimm: yeah, hence the comment at the top of that block
it's not the most intuitive that `add=False` _removes_ the permissions

boegel: Should we open an issue to improve the API of `adjust_permissions` in EasyBuild 5.0?

boegel:
There really should be a dedicated test that verifies whether the changes indeed fix the problem at hand.

I puzzled one together, see https://github.com/jfgrimm/easybuild-framework/pull/1

Unless I'm misunderstanding something, the extra test shows the fix isn't fully working as designed yet, since it fails on Python 3.6.8 in RHEL8...

easybuild/tools/filetools.py
```diff
@@ -2435,8 +2437,42 @@ def copy_file(path, target_path, force_in_dry_run=False):
...
+                        # shutil.copy2() will give a PermissionError, when using Python < 3.7
+                        # see https://bugs.python.org/issue24538
+                        shutil.copy2(path, target_path)
+                        _log.info("%s copied to %s", path, target_path)
+                    except OSError as err:
+                        # catch the more general OSError instead of PermissionError, since python2 doesn't support
+                        # PermissionError
+                        if not os.stat(path).st_mode & stat.S_IWUSR:
+                            # failure not due to read-only file
+                            raise EasyBuildError("Failed to copy file %s to %s: %s", path, target_path, err)
```
boegel, line 2451: This part doesn't make sense to me, I think it should be removed...

If we give up here if the source file is read-only, then why bother with the attempt to fix the issue when using Python 3.6 via `shutil._copyxattr` below?

boegel (changes requested):

easybuild/tools/filetools.py
boegel: Discussed with @jfgrimm in Slack, the condition is wrong here, should be `if os.stat`...

Also, it makes more sense to me to check whether the *target* file is read-only or not?

Fixed in https://github.com/jfgrimm/easybuild-framework/pull/1

boegel (approved):

PR #4644, opened 2024-09-17
Ignore "ambiguous variable name" warning for EasyConfig style check

Flamefire:
'l' is considered an ambiguous name, but we use it often for 'lib'

Required for https://github.com/easybuilders/easybuild-easyconfigs/pull/21405

### DIFF
--- a/easybuild/framework/easyconfig/style.py
+++ b/easybuild/framework/easyconfig/style.py
@@ -137,6 +137,7 @@ def check_easyconfigs_style(easyconfigs, verbose=False):
     # note that W291 has been replaced by our custom W299
     options.ignore = (
         'W291',  # replaced by W299
+        'E741',  # 'l' is considered an ambiguous name, but we use it often for 'lib'
     )
     options.verbose = int(verbose)
 


boegel:
@Flamefire I actually think it would be better if we stopped using `l` as single-letter variable in easyconfig, so rather than ignoring it, we should fix it?

boegel:
> I wouldn't mind. It should be a "quick" search and replace. I see at least 104 affected easyconfigs. Additionally what should `l for l in (lib, lib64)` become? We'd need to use `local_lib` which is a bit verbose. Or `i`

I'd prefer no single letter variables at all actually...

Flamefire:
> I'd prefer no single letter variables at all actually...

In list comprehensions like `['%s/libfoo.so' % i for i in (lib, lib64)]` or `['lib/' + f for f in local_files]` I don't see how longer names would increase the readability especially when they become much longer:  `['%s/libfoo.so' % local_dir for local_dir in (lib, lib64)]` or `['lib/' + local_file for local_file in local_files]` where I find the prefix a bit confusing.
And if we do change it I guess we'd need to fix a lot of ECs. Not sure if that naming check is only run for new/changed ECs but I fear it would require some unrelated changes for otherwise small fixes to existing ECs.

boegel:
Let's then just use `x` instead of `l` (in #21405), which is non-ambiguous

boegel:
No need for this with the changes in https://github.com/easybuilders/easybuild-easyconfigs/pull/21502, so closing this...

Issue #3331, opened 2020-05-08
Problem using CPATH in modulefiles overwriting system paths


Flamefire:
TLDR: Stop setting CPATH in module files and set `C_INCLUDE_PATH, CPLUS_INCLUDE_PATH, INCLUDE`  instead

This came up for ROOT and was work-arounded in https://github.com/easybuilders/easybuild-easyblocks/pull/2047

Basically: the C/C++ preprocessor considers the paths in the following order:
- `-I`
- `CPATH`
- `-isystem`
- `CPLUS_INCLUDE_PATH`
- buildins

Now projects adding paths via `-isystem` have those ignored if the same folder/file name exists in `CPATH` which happened for ROOT and LLVM resulting in taking an incompatible, EB installed LLVM over the included LLVM

Talked to a Spack guy about the CPATH issue: They had the same: https://github.com/spack/spack/issues/11555 and solved it by using `C_INCLUDE_PATH, CPLUS_INCLUDE_PATH, INCLUDE`  instead: https://github.com/spack/spack/pull/14749   
So I guess we can and should be doing the same. For existing modules we could simply move the contents of `CPATH` to the other 3 variables after loading a module   
Obvious downside: 3 variables with duplicated content instead of 1.

From Bennet Fauber:
https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html
Lists the variables for GNU Cs.

```
Some additional environment variables affect the behavior of the preprocessor.
CPATH
C_INCLUDE_PATH
CPLUS_INCLUDE_PATH
OBJC_INCLUDE_PATH
CPATH specifies a list of directories to be searched as if specified with -I, but after any paths given with -I options on the command line. This environment variable is used regardless of which language is being preprocessed.
The remaining environment variables apply only when preprocessing the particular language indicated. Each specifies a list of directories to be searched as if specified with -isystem, but after any paths given with -isystem options on the command line.
```
CPLUS_INCLUDE_PATH is particularly useful for --std=c++99 and what have you that are specific to C++.  I am blanking on which it was, but one of the geographic libraries was using CFLAGS for that, and it wasn't working well....

boegel:
@Flamefire Do you think there's a scenario where no longer setting `$CPATH` could cause trouble?
Incorrect paths being passed to `-isystem` (like `/usr/include`), for example?

[...]

PR #4645, opened 2024-09-18
add `--search-path-cpp-headers` configuration option to control how EasyBuild sets paths to headers at build time

lexming:
Partial fix for #3331

EasyBuild (at build time) currently handles paths to headers of dependencies provided by EB through `CPPFLAGS`. The result is that the compilation command gets additional options of the form `-I/path/to/include` explicitly pointing the compiler to places with extra headers.

The GCC prepocessor also supports the following environment variables that define the search paths for headers, which have lower precedence than `-I` command line options:
* CPATH
* C_INCLUDE_PATH, CPLUS_INCLUDE_PATH, OBJC_INCLUDE_PATH

(see https://gcc.gnu.org/onlinedocs/cpp/Environment-Variables.html#Environment-Variables-1)

This PR adds a new build option and toolchain option called `cpp-headers-search-path` to control how EasyBuild will handle paths to headers **at build time**:
* _CPPFLAGS_: current approach by setting `CPPFLAGS` and injecting `-I/path/to/include` to compilation command
* _CPATH_: new approach by adding paths to include directories to `CPATH` environment variable
* _INCLUDE_PATHS_: new approach by adding paths to include directories to `C_INCLUDE_PATH`, `CPLUS_INCLUDE_PATH`, `OBJC_INCLUDE_PATH` independently

Notes:
* all these options and environment variables are also supported by LLVM and the Intel compilers.
* option _CPPFLAGS_ stays the default. So this PR does not change current behavior.
* the toolchain option has precedence over the build option as it is more specific and (probably) easyconfigs setting this option will not work with any other setting.
* FCC toolchain can now be drastically simplified by leveraging inheritance to the new `Toolchain._add_dependency_cpp_headers()` method (@migueldiascosta can you please check that this works on your side?)
* replaced hardcoded settings for `LDFLAGS` and `CPPFLAGS` with new `Toolchain` methods in ACML toolchain
* added unit test for cpp-headers-search-paths option


lexming:
@Flamefire that's out of scope of this PR. This handles all supported options by CPP, which does not apply to Fortran. For Fortran we first need to check what is supported by the different Fortran processors (gfortran for GNU and FPP for Intel) and probably it will need a different option in EB.

lexming:
Actually, after checking the docs, it turns out that the Fortran compilers in both GNU and Intel do support the same environment variables defined by CPP. So this PR already applies to Fortran as well:
* `gfortran` uses the C prepocessor (:shrug:), see https://gcc.gnu.org/onlinedocs/gfortran/Preprocessing-Options.html
* Intel Fortran compiler *only* supports `CPATH` as search path of include directories in Linux environments, see https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2024-2/supported-environment-variables.html ( :warning:  `INCLUDE` is only supported on Windows )

Flamefire:
Ok, I see. Only potential issue I see is that setting `$CPATH` here might make Intel-Fortran pick things up it didn't before.
But I guess that was already the case as we put the same paths in CPATH in the module files, don't we?

Also: How are existing entries handled? I.e. if `CPATH=/folder1:/folder2` and adding both here (again) would they be duplicated? Would the order get changed?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -1049,39 +1060,70 @@ def handle_sysroot(self):
...
+            self.variables.append_subdirs(env_var, dep_root, subdirs=header_dirs)
+
+    def _add_dependency_linker_paths(self, dep_root, extra_dirs=None):
+        """
+        Append linker paths for given dependency root directory
+        """
+        lib_dirs = ["lib64", "lib"]
+        if isinstance(extra_dirs, (tuple, list)):
+            lib_dirs.extend(extra_dirs)
+        lib_dirs = set(lib_dirs)  # remove duplicates
```
Flamefire, line 1122: Make it a set from the start? Also what happens if it is not an iterable or None? Maybe just assume it is and let Python throw if it isn't
```suggestion
        lib_dirs = {"lib64", "lib"}
        if extra_dirs:
            lib_dirs.update(extra_dirs)
```

test/framework/toolchain.py
```diff
@@ -955,6 +955,49 @@ def test_precision_flags(self):
...
+                tc = self.get_toolchain("foss", version="2018a")
+                tc.set_options({"search-path-cpp-headers": tc_opt})
+                with self.mocked_stdout_stderr():
+                    tc.prepare()
+                    for env_var in cpp_headers_mode[tc_opt]:
+                        assert_fail_msg = (
+                            f"Variable {env_var} required by search-path-cpp-headers toolchain option '{tc_opt}' "
+                            "not found in toolchain environment"
+                        )
+                        self.assertIn(env_var, [*tc.variables], assert_fail_msg)
```
Flamefire, line 991:
```suggestion
                        self.assertIn(env_var, tc.variables, assert_fail_msg)
```

test/framework/toolchain.py
```diff
@@ -955,6 +955,49 @@ def test_precision_flags(self):
...
+                self.modtool.purge()
+        # test wrong toolchain option
+        tc = self.get_toolchain("foss", version="2018a")
+        tc.set_options({"search-path-cpp-headers": "WRONG_MODE"})
+        with self.mocked_stdout_stderr():
+            error_pattern = "Unknown value selected for option search-path-cpp-headers"
+            self.assertErrorRegex(EasyBuildError, error_pattern, tc.prepare)
+        self.modtool.purge()
```
Flamefire, line 999: Do we need the purge at the end of the test?

easybuild/tools/toolchain/compiler.py
```diff
@@ -93,6 +93,7 @@ class Compiler(Toolchain):
         'vectorize': (None, "Enable compiler auto-vectorization, default except for noopt and lowopt"),
         'packed-linker-options': (False, "Pack the linker options as comma separated list"),  # ScaLAPACK mainly
         'rpath': (True, "Use RPATH wrappers when --rpath is enabled in EasyBuild configuration"),
+        'search-path-cpp-headers': (False, "Search path used at build time for include directories"),
```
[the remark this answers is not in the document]
lexming: good point, this needed a bit more changes to avoid warnings in the logs, fully fixed in https://github.com/easybuilders/easybuild-framework/pull/4645/commits/324c79c1cbef9ba200e974d5af2ff12b579f47b2

easybuild/tools/toolchain/constants.py
```diff
@@ -31,9 +31,9 @@
 * Kenneth Hoste (Ghent University)
 """
 
+from easybuild.tools.toolchain.variables import CommandFlagList, CommaSharedLibs, CommaStaticLibs, FlagList
+from easybuild.tools.toolchain.variables import IncludePaths, LibraryList, LinkLibraryPaths, SearchPaths
```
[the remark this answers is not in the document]
lexming: that's too much of a change, framework currently formats imports in this way, changing that format choice is something we should do consistently across all modules

easybuild/tools/toolchain/toolchain.py
```diff
@@ -850,7 +861,7 @@ def prepare(self, onlymod=None, deps=None, silent=False, loadmod=True,
             else:
                 self.log.debug("prepare: set additional variables onlymod=%s", onlymod)
 
-                # add LDFLAGS and CPPFLAGS from dependencies to self.vars
+                # add linker and preprocessor paths to dependencies to self.vars
```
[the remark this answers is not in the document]
lexming: That's the point, we no longer **only** use `LDFLAGS`. Now we inject paths, either through flags or through other environment variables that are not flags. That's why it is more accurate to talk about paths.
I agree with the extra _to_.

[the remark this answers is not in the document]
lexming: fixed typo in https://github.com/easybuilders/easybuild-framework/pull/4645/commits/1344b974550fe8248b5e431914ae5533c42b80d5

easybuild/tools/toolchain/toolchain.py
```diff
@@ -1049,39 +1060,70 @@ def handle_sysroot(self):
...
-        if ld is not None:
-            for p in ld:
-                if p not in ld_paths:
-                    ld_paths.append(p)
-
-        if not names:
-            deps = self.dependencies
-        else:
-            deps = [{'name': name} for name in names if name is not None]
+        Add linker and preprocessor paths to dependencies to self.variables
```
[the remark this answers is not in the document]
lexming: This docsrtring is not a repetition, this method adds paths to `self.variables`, while the other code results in this paths being added to `self.vars`. I know it is confusing, but I will not get into the `self.vars` vs `self.variables` situation in `Toolchain`. So that's why I added this docstring.

easybuild/tools/toolchain/toolchain.py
lexming: Thanks for the comment, this made realize that we should keep the order of those dirs as that translates into the order of paths in the corresponding environment variable. So I replaced going the set by `tools.utilities.nub`.

Regarding the `if instance`, I don't see a way to avoid it, because if a string is passed to this method that will be successfully taken by `extend` or `update` (as it is an iterable) but generating the wrong result.

test/framework/toolchain.py
lexming: We do it in all other toolchain tests. I think it does not hurt to clean up after the test.

Flamefire: Just looks odd to me.

My understanding is that the purge only resets the environment which we restore already in the teardown of the base testcase class. If we really want to do that at the end of all those testcases we could move it to the `teardown` of this test-class: https://github.com/easybuilders/easybuild-framework/blob/8cbb705f81f3aa9fe7e5adbee1e0e486e575f7e8/test/framework/toolchain.py#L77-L83

easybuild/tools/toolchain/toolchain.py
lexming: Good point, I added extra checks to handle unexpected types as `extra_dirs`. See https://github.com/easybuilders/easybuild-framework/pull/4645/commits/cd01b6564498618f689937af9bdd7904e56a1e73

test/framework/toolchain.py
lexming: I do not disagree with you, but this is out of scope for this PR. I will not change the workflow of tests here.

easybuild/tools/toolchain/toolchain.py
```diff
@@ -1049,39 +1060,92 @@ def handle_sysroot(self):
...
+
+    def _add_dependency_cpp_headers(self, dep_root, extra_dirs=None):
+        """
+        Append prepocessor paths for given dependency root directory
+        """
+        header_dirs = ["include"]
+
+        if extra_dirs is None:
+            extra_dirs = ()
+        elif extra_dirs and isinstance(extra_dirs, (str,)):
```
Flamefire, line 1092: Why the check for a non-empty string? Does that make sense or is it an error to pass an empty string? With this it gets ignored. Also no need for a tuple here:
```suggestion
        elif extra_dirs and isinstance(extra_dirs, str):
```

easybuild/tools/toolchain/toolchain.py
```diff
@@ -1049,39 +1060,92 @@ def handle_sysroot(self):
...
+
+    def _add_dependency_linker_paths(self, dep_root, extra_dirs=None):
+        """
+        Append linker paths for given dependency root directory
+        """
+        lib_dirs = ["lib64", "lib"]
+
+        if extra_dirs is None:
+            extra_dirs = ()
+        elif extra_dirs and isinstance(extra_dirs, (str,)):
```
Flamefire, line 1132:
```suggestion
        elif extra_dirs and isinstance(extra_dirs, str):
```

easybuild/tools/toolchain/toolchain.py
```diff
@@ -1049,39 +1060,92 @@ def handle_sysroot(self):
...
+            lib_dirs = nub(lib_dirs)  # remove duplicates
+
+        env_var = "LDFLAGS"
+        self.log.debug("Adding lib paths to toolchain variable '%s': %s", env_var, dep_root)
+        self.variables.append_subdirs(env_var, dep_root, subdirs=lib_dirs)
 
-        for root in roots:
-            self.variables.append_subdirs("CPPFLAGS", root, subdirs=cpp_paths)
-            self.variables.append_subdirs("LDFLAGS", root, subdirs=ld_paths)
+    def _unique_ordered_list_append(self, base, extra_dirs):
```
Flamefire, line 1147: Did you intend to use this but forgot? It seems to make sense to me to have the logic in a function instead of 2 places.

lexming: yeah, I did and backpedaled.  Such a method should not go into `Toolchain`, but to something more generic like `easybuild.tools.utilities`. I now had the time to deal with that and https://github.com/easybuilders/easybuild-framework/pull/4645/commits/b714e65e124949ab64b177c8e57c1d4f55c4cf7c adds that method in its proper place.

easybuild/tools/toolchain/toolchain.py
lexming: The issue with `str` is that it can be taken by `list.extend()` without error (as it is an iterable) but generates the wrong result. So instead of throwing another error in the case of a non-empty string, we do the right thing and move forward.

easybuild/tools/utilities.py
```diff
@@ -223,6 +223,25 @@ def nub(list_):
     return [x for x in list_ if x not in seen and not seen_add(x)]
 
 
+def unique_ordered_append(base, extra):
+    """Append elements of extra list to another base list keeping order and without duplicates"""
+    if extra and isinstance(extra, str):
```
Flamefire, line 228: I'm still not sure supporting a string argument is correct and omitting the empty string. It is an undocumented feature now and e.g. `unique_ordered_append(base, 'foo') == unique_ordered_append(base, ['foo'])` but `unique_ordered_append(base, '') != unique_ordered_append(base, [''])` which could be unexpected. Hence I'd just disallow strings to avoid the ambiguity

easybuild/tools/utilities.py
```diff
@@ -223,6 +223,25 @@ def nub(list_):
...
+        extra = [extra]
+
+    try:
+        base.extend(extra)
+    except TypeError as err:
+        err_msg = f"_unique_ordered_list_append: given extra list is not iterable: {extra}"
+        raise EasyBuildError(err_msg) from err
```
Flamefire, line 235: The backtrace should already contain the origin function avoiding using the wrong name here ;-)
```suggestion
        raise EasyBuildError( f"given extra list is not iterable: {extra}") from err
```

easybuild/tools/utilities.py
```diff
@@ -223,6 +223,25 @@ def nub(list_):
...
+    except AttributeError as err:
+        err_msg = f"_unique_ordered_list_append: given base cannot be extended: {base}"
+        raise EasyBuildError(err_msg) from err
```
Flamefire, line 238:
```suggestion
        raise EasyBuildError(f"given base cannot be extended: {base}") from err
```

easybuild/tools/utilities.py
```diff
@@ -223,6 +223,25 @@ def nub(list_):
...
+    else:
+        base = nub(base)  # remove duplicates
+
+    return base
```
Flamefire, line 242: I think it is clearer outside the "else" because both paths throw already and hence the `else` is superflous:
```suggestion

    return nub(base)  # remove duplicates
```

test/framework/utilities_test.py
```diff
@@ -191,6 +191,23 @@ def test_LooseVersion(self):
         self.assertEqual(LooseVersion('2.a').version, [2, 'a'])
         self.assertEqual(LooseVersion('2.a5').version, [2, 'a', 5])
 
+    def test_unique_ordered_append(self):
+        """Test unique_ordered_list_append method"""
+        base = ["potato", "tomato", "orange"]
+
+        reference = ["potato", "tomato", "orange", "apple"]
+        self.assertEqual(tu.unique_ordered_append(base, ["apple"]), reference)
```
Flamefire, line 199: Is it intended, that the original list gets modified? I.e. after this line `base == ["potato", "tomato", "orange", "apple"]`

If this is intended I'd add a test case showing it and maybe rather pass in copies and assert the base is not modified:
```suggestion
        self.assertEqual(tu.unique_ordered_append(base.copy(), ["apple"]), reference)
        # Or
        base_orig = base.copy()
        self.assertEqual(tu.unique_ordered_append(base, ["apple"]), reference)
        self.assertEqual(base, base_orig)
```

Flamefire (changes requested):
Still some refinements of the new functions which I think are required to avoid subtile issues.

easybuild/tools/utilities.py
lexming: right :slightly_smiling_face:

easybuild/tools/utilities.py
lexming: ok, I see your point. I forbid strings altogether and changed the name of the method to `unique_ordered_extend` to remove any expectation of it working like `List.append()`.

Flamefire (approved):
Looks good to me. One (very) minor suggestion where I'm not sure. See the comment.

easybuild/tools/utilities.py
```diff
@@ -223,6 +223,23 @@ def nub(list_):
...
 
 
+def unique_ordered_extend(base, affix):
+    """Extend base list with elements of affix list keeping order and without duplicates"""
+    if isinstance(affix, str):
+        # avoid extending with strings, as iterables generate wrong result without error
+        raise EasyBuildError(f"given affix list is a string: {affix}")
+
+    try:
+        ext_base = base.copy()
```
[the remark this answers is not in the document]
lexming: that's wrong, you cannot copy a `str`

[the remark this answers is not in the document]
Flamefire: I see, it gives a similar `AttributeError` and with the `raise ... from` it should be clear enough why it "cannot be extended". 👍

PR #4646, opened 2024-09-18
take into account alternate sysroot for `/bin/bash` used by `run_cmd`

casparvl:
EasyBuild's `run_cmd` runs commands in a subshell. The shell used is defined in the `exec_cmd`. However, in context where the use of a `sysroot` is configured, one would want to use `<sysroot>/bin/bash` instead. This PR implements that change. Note that since `run_cmd` is _also_ called a few times _before_ `set_up_configuration` is called, we need to be able to disable the check for the `build_option('sysroot')` for those instances (similar to how this is already don for `build_option('hooks')`).

Solves [this issue](https://github.com/EESSI/software-layer/pull/585#issuecomment-2313143191).

Edit: hmmm, I made this PR, and then realized, this is probably one of the components that is very different in EB 5.0. And it seems to be indeed... 

Someone with more EB 5.0 knowledge needs to give me a pointer of how to fix this for EB 5.0 :)

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1969,7 +1969,7 @@ def set_tmpdir(tmpdir=None, raise_error=False):
             os.close(fd)
             os.chmod(tmptest_file, 0o700)
             if not run_cmd(tmptest_file, simple=True, log_ok=False, regexp=False, force_in_dry_run=True, trace=False,
-                           stream_output=False, with_hooks=False):
+                           stream_output=False, with_hooks=False, with_sysroot=False):
                 msg = "The temporary directory (%s) does not allow to execute files. " % tempfile.gettempdir()
                 msg += "This can cause problems in the build process, consider using --tmpdir."
                 if raise_error:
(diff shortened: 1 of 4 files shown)


casparvl:
Ok, this issue actually _does not exist_ for EB 5.0: the whole `run_cmd` is now replaced by `run_shell_cmd`, and that has a different way of determining the shell command. See [this piece of code](https://github.com/easybuilders/easybuild-framework/blob/a2550eb8fab479f517badbf45925c3cebda2880c/easybuild/tools/run.py#L454):

```
    # use bash as shell instead of the default /bin/sh used by subprocess.run
    # (which could be dash instead of bash, like on Ubuntu, see https://wiki.ubuntu.com/DashAsBinSh)
    # stick to None (default value) when not running command via a shell
    if use_bash:
        bash = shutil.which('bash')
        _log.info(f"Path to bash that will be used to run shell commands: {bash}")
        executable, shell = bash, True
    else:
        executable, shell = None, False
```
_at least_ for EESSI, the result is that the bash from the prefix is used. That means our issue no longer exists. I'm closing this PR, as it is not relevant.

boegel:
@casparvl As discussed, I think it's worth considering this to be merged into `develop`, since there's a (small) window of opportunity to get this included in EasyBuild v4.9.4 (which we'll be releasing because of a bug that was introduced in the GCC easyblock, see [here](https://github.com/easybuilders/easybuild-easyblocks/pull/3449)).

I'll take a thorough look into this right now...

PR #4652, opened 2024-09-23
allow extensions to add text to the module file

### LINKED ISSUE #4647: Incomplete module for extensions
I just noticed that the `SciPy-bundle` module generated by `--module-only` is missing some paths.

Those are supposed to be added by the `numpy` extension. However the current mechanism is flawed:

- The result of `ext.run()` is collected in a variable: https://github.com/easybuilders/easybuild-framework/blob/bf0af10626644574cdf528998962b6f3c770fdc1/easybuild/framework/easyblock.py#L1950-L1952
- This is then added to the module https://github.com/easybuilders/easybuild-framework/blob/bf0af10626644574cdf528998962b6f3c770fdc1/easybuild/framework/easyblock.py#L1466

The problems with this are:

1. For `--module-only` this is not run
2. For `--skip` it might not be run
3. For `--parallel-extension` it is not used either: https://github.com/easybuilders/easybuild-framework/blob/bf0af10626644574cdf528998962b6f3c770fdc1/easybuild/framework/easyblock.py#L2014

From our "official" easyblocks only `numpy` uses that mechanism.

So we need a new method for extensions that returns the module-extra when used as an extension. The current `make_module_extra` cannot be used for that.

I'd propose `make_module_extra_extension` but we already have `make_module_extra_extensions` which would be confusing.

Flamefire:
Introduce `make_extension_module_extra` which gets called during module
file creation for every extension similar to `make_module_extra`.
This ensures it will also be called for parallel extension or --module-only builds.

Fixes https://github.com/easybuilders/easybuild-framework/issues/4647

I'm open for a better name but `make_module_extra_extensions` already exists so `make_module_extra_extension` might be confusing.

Currently the (now deprecated) feature is only used by the numpy easyblock to add the numpy includes to `$CPATH`. So that easyblock needs to be updated. I'd argue that we shouldn't add the numpy includes as a regular `pip install numpy` doesn't do that either. So if a user updates the numpy in a virtualenv using our current approach might end up using headers from one numpy and libs/python files from another.

However I'm not sure if (and why) other software we have already relies on that. One which did will be fixed by https://github.com/easybuilders/easybuild-easyconfigs/pull/21467

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1663,21 +1663,30 @@ class EasyBlock:
         Sets optional variables for extensions.
         """
         # add stuff specific to individual extensions
-        lines = [self.module_extra_extensions]
+        txt = self.module_extra_extensions
+
+        if not self.ext_instances:
+            self.prepare_for_extensions()
+            self.init_ext_instances()
+
+        for ext in self.ext_instances:
+            ext_txt = ext.make_extension_module_extra()
+            if ext_txt:
+                txt += ext_txt
 
         # set environment variable that specifies list of extensions
         # We need only name and version, so don't resolve templates
         exts_list = self.make_extension_string(ext_sep=',', sort=False)
         env_var_name = 'EBEXTSLIST' + convert_name(self.name, upper=True)
-        lines.append(self.module_generator.set_environment(env_var_name, exts_list))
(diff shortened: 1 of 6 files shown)


boegel:
@Flamefire Can you look into resolving the merge conflict?

Flamefire:
Did a full rebase to make the commits easier to read in EB 5.x

PR #4653, opened 2024-09-23
refactor generation of required environment variables in module files + deprecate `make_module_req_guess` method in `EasyBlock` class

### LINKED ISSUE #3331: Problem using CPATH in modulefiles overwriting system paths

### LINKED ISSUE #3504: directories that don't contain any library files shouldn't be added to `$LD_LIBRARY_PATH`
not a bug, just a bit confusing sometimes.

not sure what pattern th elinker looks for, but i guess files with name `lib*` or somthing like that

lexming:
Goal is to lay groundwork to be able to fix #3331 and add new option to control which search path variables are added in modules by simplifying the code injecting environment variables to module files.

This PR should not alter the resulting module files to what is currently generated by EB. If it changes the list of environment variables it should have no effect (_e.g._ because those paths are empty).

Simplification of logic handling environment variables for module files:

1. replace obscure rules defining which search paths require populated dirs with:
    * all modules require search paths that point to existing files or non-empty dirs 
      (fixes #3504)
    * fake modules can have search paths to empty dirs
2. replace obscure rules defining which search paths require files in their top level directory with:
    * by default all search paths to directories  can be populated at any level (recursive check)
    * this setting is now controlled by `top_level_file` attribute of `ModuleEnvironmentVariable`
3. replace convoluted rules to handle symlinked `lib64` that generate duplicate paths and need explicit exceptions with:
    * paths to `lib64` are always ignored if `lib64` is a symlink to `lib`
    * paths to `lib` are always ignored if `lib` is a symlink to `lib64`
    * remove `lib32` from list of library search paths
4. allow simple updates of environment variables in child EasyBlocks, replacing the clunky:
    ```
    def make_module_req_guess(self):
        guesses = super(EB_Clang, self).make_module_req_guess()
        guesses.update({'SOME_VAR': ['some/path]})
        return guesses
    ```
    with an elegant:
    ```
    self.module_load_environment.SOME_VAR = ['some/path']
    ```

Changelog:
* add `ModuleEnvironmentVariable` pseudo-dataclass to `easybuild.tools.modules` to hold definitions of environment variables for modules
* add `ModuleLoadEnvironment` singleton class to `easybuild.tools.modules` to hold environment definition for modules on load
* deprecate `EasyBlock.make_module_req_guess` in favor of `ModuleLoadEnvironment`
* add LibSymlink enum to `easybuild.framework.easyblock` to define possible linkage states of lib directories
* add `install_lib_symlink` attribute to EasyBlcok to keep track of symlink status between lib dirs
* add globals to `easybuild.tools.config` that define known directories for binaries, libraries and headers 
* avoid changing the current working directory of the main process in `easyblock.make_module_req` by working with absolute paths instead


lexming:
@boegel look all the beautiful green tests!

This seems ready on my side. I tested it with several easyblocks that add custom stuff to `make_module_req_guess` and I have hit no issues so far. Even things like `root.py` easyblock that set `PYTHONPATH` in `make_module_req_guess` instead of the more common `make_module_extra` do work. If you have anything in mind that I should test in particular let me know.

easybuild/tools/modules.py
```diff
@@ -131,6 +133,116 @@
...
+        """Shortcut to extend list of paths"""
+        self.paths.extend(*args)
+
+    def prepend(self, item):
+        """Shortcut to append to list of paths"""
+        self.paths.insert(0, item)
+
+
+# singleton metaclass: only one instance is created
+BaseModuleEnvironment = create_base_metaclass('BaseModuleEnvironment', Singleton, object)
```
boegel, line 173: @lexming Using a singleton here (so a single instance that is shared across all `EasyBlock` instances) seems like a bad idea to me...

If EasyBuild installs two things in the same session (via `--robot`, or by being fed multiple easyconfigs on the command line), you'll end up with incorrect stuff in the generated module file due to this, no?

boegel (changes requested):

easybuild/tools/modules.py
boegel: @lexming I've implemented a test that shows that using a singleton here is a bad idea, see https://github.com/lexming/easybuild-framework/pull/5

boegel:
@lexming Can you look into fixing the merge conflict, and get the test to pass?

test/framework/toy_build.py
```diff
@@ -4320,6 +4320,42 @@ def test_toy_python(self):
...
+        toy_app_ec = os.path.join(test_ecs, 't', 'toy-app', 'toy-app-0.0.eb')
+        test_toy_app_ec = os.path.join(self.test_prefix, 'test-toy-app.eb')
+        toy_ec_txt = read_file(toy_app_ec)
+        toy_ec_txt += "\npostinstallcmds += ['mkdir %(installdir)s/toy-headers']"
+        toy_ec_txt += "\npostinstallcmds += ['touch %(installdir)s/toy-headers/toy.h']"
+        write_file(test_toy_app_ec, toy_ec_txt)
+
+        self.run_test_toy_build_with_output(ec_file=test_toy_ec, extra_args=[test_toy_app_ec], raise_error=True)
+
+        toy_modtxt = read_file(os.path.join(self.test_installpath, 'modules', 'all', 'toy', '0.0.lua'))
```
boegel, line 4350: @lexming Tests are run using both `Lua` and `Tcl` module syntax, so you can't assume that the module file will be named `0.0.lua`.

To reproduce the failure:
```
TEST_EASYBUILD_MODULE_SYNTAX=Tcl python -O -m test.framework.suite test_toy_multiple_ecs_module
```

```suggestion
        toy_modtxt = read_file(os.path.join(self.test_installpath, 'modules', 'all', 'toy', '0.0'))
        if get_module_syntax() == 'Lua':
            toy_modtxt += '.lua'
```
```

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/framework/easyblock.py [hunk 3]
```diff
@@ -1605,108 +1619,113 @@ def make_module_group_check(self):
...
+        for abs_path in exp_search_paths:
+            # return relative paths
+            tentative_path = os.path.relpath(abs_path, start=self.installdir)
+            tentative_path = "" if tentative_path == "." else tentative_path  # use empty string instead of dot
+
+            # avoid duplicate entries between symlinked library dirs
+            tentative_sep = tentative_path + os.path.sep
+            if self.install_lib_symlink == LibSymlink.LIB64 and tentative_sep.startswith("lib64" + os.path.sep):
+                self.log.debug("Discarded search path to symlinked lib64 directory: %s", tentative_path)
+                break
```
boegel, line 1698: This should be `continue` rather than `break`, we should check all paths in `exp_search_paths` not jump out of the loop when we hit a symlink path?

boegel (changes requested):
@lexming All remarks fixed in https://github.com/lexming/easybuild-framework/pull/6

easybuild/framework/easyblock.py
```diff
@@ -1605,108 +1619,113 @@ def make_module_group_check(self):
...
-                            retained_paths.append(pth)
-                    if retained_paths != paths:
-                        self.log.info("Only retaining paths for %s that contain at least one file: %s -> %s",
-                                      key, paths, retained_paths)
-                        paths = retained_paths
-
-            if paths:
-                lines.append(self.module_generator.prepend_paths(key, paths))
+                mod_req_paths = []
+                top_level = getattr(self.module_load_environment, env_var).top_level_file
```
boegel, line 1658: This goes horribly wrong when using a copy of `PerlModule` that has not been migrated yet to `module_load_environment` (as is done in https://github.com/easybuilders/easybuild-easyblocks/pull/3529):

```
    "__main__", mod_spec)
    exec(code, run_globals)
  File "/tmp/vsc40023/easybuild-framework/easybuild/main.py", line 801, in <module>
    main_with_hooks()
  File "/tmp/vsc40023/easybuild-framework/easybuild/main.py", line 787, in main_with_hooks
  File "/tmp/vsc40023/easybuild-framework/easybuild/main.py", line 743, in main
    hooks, do_build)
  File "/tmp/vsc40023/easybuild-framework/easybuild/main.py", line 567, in process_eb_args
  File "/tmp/vsc40023/easybuild-framework/easybuild/main.py", line 175, in build_and_install_software
    raise ec_res['err']
  File "/tmp/vsc40023/easybuild-framework/easybuild/main.py", line 136, in build_and_install_software
    (ec_res['success'], app_log, err_msg, err_code) = build_and_install_one(ec, init_env)
  File "/tmp/vsc40023/easybuild-framework/easybuild/framework/easyblock.py", line 4432, in build_and_install_one
  File "/tmp/vsc40023/easybuild-framework/easybuild/framework/easyblock.py", line 4300, in run_all_steps
  File "/tmp/vsc40023/easybuild-framework/easybuild/framework/easyblock.py", line 4143, in run_step
    step_method(self)()
  File "/user/gent/400/vsc40023/easybuild/eb5/easybuild-easyblocks/easybuild/easyblocks/generic/perlmodule.py", line 140, in sanity_check_step
    return ExtensionEasyBlock.sanity_check_step(self, EXTS_FILTER_PERL_MODULES, *args, **kwargs)
  File "/tmp/vsc40023/easybuild-framework/easybuild/framework/extensioneasyblock.py", line 188, in sanity_check_step
    fake_mod_data = self.load_fake_module(purge=True, extra_modules=extra_modules)
  File "/tmp/vsc40023/easybuild-framework/easybuild/framework/easyblock.py", line 1796, in load_fake_module
  File "/tmp/vsc40023/easybuild-framework/easybuild/framework/easyblock.py", line 3905, in make_module_step
    txt += self.make_module_req(fake=fake)
  File "/tmp/vsc40023/easybuild-framework/easybuild/framework/easyblock.py", line 1664, in make_module_req
    requires_files = getattr(self.module_load_environment, env_var).requires_files
AttributeError: 'ModuleLoadEnvironment' object has no attribute 'PERL5LIB'
```

(note that I'm using my updated version here, see also https://github.com/lexming/easybuild-framework/pull/6)

It's getting `PERL5LIB` via `PerlModule.make_module_req_guess`, but since that has never been defined as an attribute in `module_load_environment`, this fails hard.

That tells me we're missing something above, where we deal with easyblocks that still define a custom `make_module_req_guess`?

easybuild/tools/modules.py
```diff
@@ -48,13 +48,15 @@
 from easybuild.tools.config import ERROR, IGNORE, PURGE, UNLOAD, UNSET
 from easybuild.tools.config import EBROOT_ENV_VAR_ACTIONS, LOADED_MODULES_ACTIONS
 from easybuild.tools.config import build_option, get_modules_tool, install_path
+from easybuild.tools.config import SEARCH_PATH_BIN_DIRS, SEARCH_PATH_HEADER_DIRS, SEARCH_PATH_LIB_DIRS
```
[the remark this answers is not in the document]
lexming: lgtm

easybuild/tools/modules.py
```diff
@@ -131,6 +133,141 @@
...
+        self.CPATH = SEARCH_PATH_HEADER_DIRS
+        self.MANPATH = ['man', os.path.join('share', 'man')]
+        self.PKG_CONFIG_PATH = [os.path.join(x, 'pkgconfig') for x in SEARCH_PATH_LIB_DIRS + ['share']]
+        self.ACLOCAL_PATH = [os.path.join('share', 'aclocal')]
+        self.CLASSPATH = ['*.jar']
+        self.XDG_DATA_DIRS = ['share']
+        self.GI_TYPELIB_PATH = [os.path.join(x, 'girepository-*') for x in SEARCH_PATH_LIB_DIRS]
+        self.CMAKE_PREFIX_PATH = ['']
+        # only needed for installations whith standalone lib64
+        self.CMAKE_LIBRARY_PATH = ['lib64']
```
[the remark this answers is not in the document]
lexming: lgtm

easybuild/tools/modules.py
```diff
@@ -131,6 +133,141 @@
...
+            (contents, kwargs) = value
+        except ValueError:
+            contents, kwargs = value, {}
+
+        if not isinstance(kwargs, dict):
+            contents, kwargs = value, {}
+
+        # special variables that require top level files
+        if name in ["PATH", "LD_LIBRARY_PATH"]:
+            kwargs.update({"top_level_file": True})
```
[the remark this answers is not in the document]
lexming: I see you enjoy torturing my linter

easybuild/framework/easyblock.py
lexming: Indeed, using `break` here is wrong. It just happens to work because _luckily_ `glob` puts symlinks after folders in the expanded lists of paths.

Moreover, this method is currently untested. I am adding a new unit test for it to avoid such issues.

easybuild/tools/modules.py
```diff
@@ -131,6 +133,141 @@
...
+    Contents of environment variable is a list of unique strings
+    """
+
+    def __init__(self, contents, top_level_file=False, delim=os.pathsep):
+        """
+        Initialize new environment variable
+        Actual contents of the environment variable are held in self.contents
+        """
+        self.contents = contents
+        self.top_level_file = bool(top_level_file)
```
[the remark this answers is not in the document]
lexming: That might be misleading. The current behaviour (v4.9) is that all paths require files in them to be declared in the module file. This attribute only changes the strictness of that check to require those files to be in the top directory.

Here there is an underlying design question, because `ModuleEnvironmentVariable` can be generic in purpose and used for other things than `make_module_req`. So maybe we need a ternary parameter (NO_FILES, ANY_FILES, TOP_LEVEL_FILES).

easybuild/framework/easyblock.py
lexming: I see the problem, very stupid... I changed approach and now using any deprecated `make_module_req_guess` will result in updating `module_load_environment`. Ensuring that it is always present.To this end I added an `update` method to `ModuleLoadEnvironment` in https://github.com/easybuilders/easybuild-framework/pull/4653/commits/87afba73a57d0cb1db3c61347ad52e3bdbc20680 and switch to it in https://github.com/easybuilders/easybuild-framework/pull/4653/commits/38774ac183f52f30f4960e07eeb8adf189353589

easybuild/tools/modules.py
```diff
@@ -131,6 +133,141 @@
...
+        - value = its "contents" attribute
+        """
+        for attr in self.__dict__:
+            yield attr, getattr(self, attr).contents
+
+    @property
+    def environ(self):
+        """
+        Return dict with mapping of ModuleEnvironmentVariables names with their contents
+        Equivalent in shape to os.environ
```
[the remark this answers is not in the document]
lexming: Good point, that was not my intention actually. I updated `environ` to be a proper environ equivalent with strings as values and added a new `as_dict` that returns a dict with lists as values. See https://github.com/easybuilders/easybuild-framework/pull/4653/commits/9297094a38986313fd610645eaaefc6607878fe3

easybuild/framework/easyblock.py
```diff
@@ -1605,108 +1619,113 @@ def make_module_group_check(self):
...
-            'CLASSPATH': ['*.jar'],
-            'XDG_DATA_DIRS': ['share'],
-            'GI_TYPELIB_PATH': [os.path.join(x, 'girepository-*') for x in lib_paths],
-            'CMAKE_PREFIX_PATH': [''],
-            'CMAKE_LIBRARY_PATH': ['lib64'],  # lib and lib32 are searched through the above
-        }
+        A dictionary of common search path variables to be loaded by environment modules
+        Each key contains the list of known directories related to the search path
+        """
+        return self.module_load_environment.environ
```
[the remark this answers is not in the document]
lexming: It's true that we need it here to catch the case where `make_module_req_guess()` is directly called in some arbitrary location in a custom easyblock.

However, this has the downside of duplicating deprecation warnings in the following common case:
1. EB reaches `make_module_req`
2. custom easyblock has its own `make_module_req_guess()`
3. first warning triggers
4. `make_module_req` calls custom `make_module_req_guess()`
5. custom `make_module_req_guess()` calls `super()`
6. generic `make_module_req_guess()` triggers second deprecation warning

But I don't see a way to avoid it, as custom `make_module_req_guess()` methods do not necessarily call `super`. So we need both to cover all cases.

easybuild/tools/modules.py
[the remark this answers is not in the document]
lexming: Got a nice solution for this, I generalized a bit more `ModuleEnvironmentVariable` by adding a _type_ attribute to it which can be one of `ModEnvVarType`. A new enum with `STRING`, `PATH`, `PATH_WITH_FILES` and `PATH_WITH_TOP_FILES`. Then with a single attribute we can control the behaviour of a specific variable. See https://github.com/easybuilders/easybuild-framework/pull/4653/commits/516d6e659df0482ea4ac75a4a8a379d71b2ebf60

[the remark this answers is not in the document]
lexming: Improved further the use and tests of `ModEnvVarType` in https://github.com/easybuilders/easybuild-framework/pull/4653/commits/d4a9df04bac44d7cb5067a8ffa7395f89ce54481

This is ready on my side.

easybuild/framework/easyblock.py
```diff
@@ -1614,108 +1634,128 @@ def make_module_group_check(self):
...
+            self.log.deprecated(
+                "make_module_req_guess() is deprecated, use EasyBlock.module_load_environment instead.",
+                "6.0",
+            )
+            self.module_load_environment.update(self.make_module_req_guess())
+
+        # Only inject path-like environment variables into module file
+        env_var_requirements = sorted([
+            (envar_name, envar_val) for envar_name, envar_val in self.module_load_environment.items()
+            if envar_val.is_path
```
boegel, line 1662: @lexming Shouldn't we at least log a warning or something if any entries in `self.module_load_environment` do not have `is_path` is True?

Because those cases will essentially be totally silently ignores currently?

easybuild/framework/easyblock.py
```diff
@@ -1614,108 +1634,128 @@ def make_module_group_check(self):
...
-                    if retained_paths != paths:
-                        self.log.info("Only retaining paths for %s that contain at least one file: %s -> %s",
-                                      key, paths, retained_paths)
-                        paths = retained_paths
-
-            if paths:
-                lines.append(self.module_generator.prepend_paths(key, paths))
+                path_type = search_paths.type
+                if fake:
+                    path_type = ModEnvVarType.PATH
```
boegel, line 1673: @lexming Is this only done to bypass the check in `expand_module_search_path` that ensures that this directory actually has files in them?

Why? Is it only to get some marginal speedup?

This may be important, the "fake" module is generated right before the sanity check is run, so if for some weird reason having additional paths to empty directories in environment variables causes trouble (or worse, somehow makes the sanity check pass), it sort of invalidates the sanity check...

Is there another reason to make this exception beyond a small performance optimization?

lexming: At some point this was needed, but I lost the specifics. This PR increases the amount of paths that are checked for existence before injection into the module file and I remember having issues with fake modules lacking paths that are non-existent at the time of creation of the fake, but exist in the end. However, I'm not hitting those issues anymore, so I'll revert it and see if they happen again.

easybuild/framework/easyblock.py
lexming: Yep we can log that, not sure if this deserves a warning though. I guess there might be cases where no path is added to the module file that are ok, like a `ModuleRC`.

boegel: My suggestion relates more to entries in `self.module_load_environment` for which `is_path` is `False`, those cases should be exposed clearly rather than silently ignoring them?

boegel:
@lexming What was the trigger for the change you made in 13f735c?

That broke the `test_module_environment_variable` test, and for good reason I'd say...

Can you enhance the tests to cover the case that made you make the change in 13f735c?

lexming:
@boegel actually the tests is what saved us, they properly failed. That change in `13f735c` was not needed, `args` are packed in a list and they need unpacking.

boegel (approved):

easybuild/framework/easyblock.py
boegel: @lexming Since this is a minor thing that only concerns logging, let's follow up on that in a separate PR, I didn't want to hold back this PR any longer...

lexming: Opened a new PR for this, see #4748

lexming:
updated description to match state of code at merge time

PR #4655, opened 2024-09-24
add `module-search-path-headers` configuration option to control how modules set search paths to header files

lexming:
Depends on:
* #4653 

This PR adds a new build option and easyconfig parameter called `module-search-path-headers` to control what environment variables will be added to module files with search paths to headers:
* _none_: (new setting) no variables added to module files with paths to headers
* _CPATH_: (default and current setting) modules set CPATH on load
* _INCLUDE_PATHS_: (new setting) modules set variables C_INCLUDE_PATH, CPLUS_INCLUDE_PATH, OBJC_INCLUDE_PATH on load with paths to headers

New test added to easyblock suite for this new option.

_update 1_: `easybuild.tools.modules.ModuleLoadEnvironment` is enhanced with the capability to hold private attributes that are invisible for the generation of the module environment.

_update 2_: `easybuild.tools.modules.ModuleLoadEnvironment` is enhanced with the capability to define arbitrary aliases to `ModuleEnvironmentVariable`


boegel:
@lexming please sync this with `5.0.x` now that #4653 is merged

lexming:
This is poorly designed, working on a better approach...

boegel:
> This is poorly designed, working on a better approach...

Poorly designed in what sense?

easybuild/framework/easyconfig/default.py
```diff
@@ -208,10 +208,11 @@
     'moduleloadnoconflict': [False, "Don't check for conflicts, unload other versions instead ", MODULES],
     'module_depends_on': [None, 'Use depends_on (Lmod 7.6.1+) for dependencies in generated module '
                           '(implies recursive unloading of modules) [DEPRECATED]', MODULES],
+    'module_search_path_headers': [False, "Environment variable set by modules on load with search paths "
```
boegel, line 211: @lexming `None` would be a better default here, no?

easybuild/tools/config.py
```diff
@@ -182,6 +182,14 @@
 EBPYTHONPREFIXES = 'EBPYTHONPREFIXES'
 PYTHON_SEARCH_PATH_TYPES = [PYTHONPATH, EBPYTHONPREFIXES]
 
+# options to handle header search paths in environment of modules
+MOD_SEARCH_PATH_HEADERS = {
+    "none": [],
```
boegel, line 187: @lexming Do we actually have a valid use case for `none`?

I don't see how that can work, we really want people to pick between `CPATH` or `C_INCLUDE_PATH` & co, so I would remove the `none` option here...

boegel (approved):
lgtm

easybuild/tools/config.py
lexming: very reasonable :+1:

PR #4659, opened 2024-09-26
use Slurm as the default job backend + deprecate support for GC3Pie as job backend

### LINKED ISSUE #4565: change default job backend to Slurm + deprecate support for using GC3Pie as job backend

gkaf89:
- The GC3Pie back-end for job management that has been the default so far is no longer actively supported. Use the Slurm back-end as default instead.
- Add deprecation message when the GC3Pie class is instantiated.

fixes #4565


easybuild/tools/job/gc3pie.py
```diff
@@ -104,6 +104,10 @@ def __init__(self, *args, **kwargs):
     def _check_version(self):
         """Check whether GC3Pie version complies with required version."""
 
+        deprication_msg = "The GC3Pie job back-end is no longer maintained and will be deprecated"
+        deprication_msg += ", please use a different backend"
+        _log.deprecated(deprication_msg, '5.0')
```
boegel, line 109:
```suggestion
        deprecation_msg = "The GC3Pie job backend is no longer maintained and will be removed"
        deprecation_msg += ", please use a different job backend"
        _log.deprecated(deprecation_msg, '5.0')
```

boegel (changes requested):

gkaf89:
There is a test that fails due to the deprecation message: 
```
test.framework.parallelbuild.ParallelBuildTest.test_build_easyconfigs_in_parallel_gc3pie
```

I manage to trace the issue in the deprecation message which is added in the GC3Pie constructor. During the tests the deprecation message in the log through an Exception; the test succeeds if the exception is caught. Under normal operation, logging the deprecation does not though an exception.

Is there any special handling of deprecation message logging in the tests?

boegel:
> Is there any special handling of deprecation message logging in the tests? How should I handle exceptions caused by deprecation messages in tests?

@gkaf89 Just call `self.allow_deprecated_behaviour()` in the test, there's a bunch of examples of this already in the test suite

boegel (approved):

PR #4660, opened 2024-09-27
improve portability of reproducible tarballs by replacing external `tar` command with `tarfile` module

### LINKED ISSUE #4657: reproducible tarballs have unexpected checksums on macOS
This is related to PR https://github.com/easybuilders/easybuild-easyconfigs/pull/21392

The checksum of the generated `lsp-types-0.95.1.tar.gz` with our reproducible tarball code is `455f2c12f64c2e293c72efdf2b90b96f5903d6590bd2e496bab5f12750402631` on several Linux systems (RHEL8, Rocky 8, Fedora 40 and Debian Bookworm).

However, on MacOS it has an unexpected `59869db34853933b239f1e2219cf7d431da006aa919635478511fabbfc8849d2`.

Steps to generate tarball from git repo:
1. ` git clone https://github.com/astral-sh/lsp-types.git`
2. `cd lsp-types && git checkout 3512a9f33eadc5402cfab1b8f7340824c8ca1439 && cd ..`
3. 
    ```
    find lsp-types -name ".git" -prune -o -print0 -exec touch --date=@0 {} \; -exec chmod "go+u,go-w" {} \; | 
    LC_ALL=C sort --zero-terminated | tar --create --no-recursion --owner=0 --group=0 --numeric-owner 
    --format=gnu --null --files-from - | gzip --no-name > lsp-types-0.95.1.tar.gz
    ```

We need to troubleshoot this on a mac:
* compare the checksum of the ungzip tarballs on both systems
* compare the raw contents of the tarball header, there is some metadata there that can point to differences. Basically check the outcome of `head -n 100` on the `.tar` files

lexming:
fixes #4657 

* ~~use more portable `--date` argument for `touch`~~
* ~~catch failed commands inside the pipeline~~
* ~~move generation of command to make reproducible archives intro its own method~~
* ~~replace harcoded pattern in tests of reproducible archives command for call to `filetools.reproducible_archive_cmd`~~
* ~~add `required` argument to `filetools.find_extensions()`~~
* make new implementation of `reproducible_archive_cmd` using the `tarfile` module
* added new `filetools.make_archive()` method and related unit test
* change handling of `filename` argument in `filetools.get_source_tarball_from_git()` to allow any extension and pass it to `make_archive()`
* `make_archive()` now supports uncompressed tarballs and compressed tarballs in GZIP, BZIP2 and XZ
* `make_archive()` can create reproducible tarballs in `.tar` or `.tar.gz` format
* checksums of sources from git repos will be verified in Python 3.9+


test/framework/filetools.py
```diff
@@ -2939,16 +2939,11 @@ def run_check():
...
-            r' running shell command "find {} -name \".git\" -prune -o -print0 -exec touch -t 197001010100 {{}} \; |'
-            r' LC_ALL=C sort --zero-terminated | tar --create --no-recursion --owner=0 --group=0 --numeric-owner'
-            r' --format=gnu --null --files-from - | gzip --no-name > %(test_prefix)s/target/test.tar.gz'
-        )
 
         expected = '\n'.join([
             r'  running shell command "git clone --depth 1 --branch tag_for_tests %(git_repo)s"',
             r"  \(in .*/tmp.*\)",
-            reprod_tar_cmd_pattern.format("testrepository"),
+            ft.reproducible_archive_cmd("testrepository", "%(test_prefix)s/target/test.tar.gz"),
```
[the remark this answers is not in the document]
lexming: no longer applies, replaced `reproducible_archive_cmd` with `make_archive`

easybuild/framework/easyblock.py
```diff
@@ -358,34 +359,49 @@ def get_checksum_for(self, checksums, filename=None, index=None):
...
         """
-        checksum = None
-
-        # sometimes, filename are specified as a dict
+        chksum_input = filename
+        chksum_input_git = None
+        # if filename is provided as dict, take 'filename' key
         if isinstance(filename, dict):
-            filename = filename['filename']
+            chksum_input = filename['filename']
```
boegel, line 366: @lexming Should we be more careful here, and also use `filename.get('filename', None)` to avoid crashing if the `filename` key is not set?

easybuild/framework/easyblock.py
```diff
@@ -358,34 +359,49 @@ def get_checksum_for(self, checksums, filename=None, index=None):
...
+        if chksum_input is None:
+            self.log.debug("Cannot get checksum without a file name")
+            return None
+
+        if sys.version_info[0] >= 3 and sys.version_info[1] < 9:
+            # ignore any checksum for given filename due to changes in python/cpython#90021
+            # checksums of tarballs made by EB of git repos cannot be reliably checked prior to Python 3.9
+            if chksum_input_git is not None:
+                self.log.deprecated(
+                    "Reproducible tarballs of git repos are only supported in Python 3.9+. "
```
boegel, line 378:
```suggestion
                    "Reproducible tarballs of Git repos are only possible when using Python 3.9+ to run EasyBuild. "

File: easybuild/tools/filetools.py [hunk 4]
```diff
@@ -2680,8 +2684,10 @@ def get_source_tarball_from_git(filename, target_dir, git_config):
     if not url:
         raise EasyBuildError("url not specified in git_config parameter")
 
-    if not filename.endswith('.tar.gz'):
-        raise EasyBuildError("git_config currently only supports filename ending in .tar.gz")
+    file_ext = find_extension(filename, required=False)
+    if file_ext:
+        print_warning(f"Ignoring extension of filename '{filename}' set in git_config parameter")
```
boegel, line 2689: Isn't this going to require cleaning up a whole bunch of easyconfigs?

easybuild/tools/filetools.py
```diff
@@ -2768,37 +2774,82 @@ def get_source_tarball_from_git(filename, target_dir, git_config):
...
+
+    :dir_path: string with path to directory to be archived
+    :archive_name: string with extensionless filename of archive
+    :archive_dir: string with path to directory to place the archive
+    :reproducuble: make a tarball that is reproducible accross systems
+    see https://reproducible-builds.org/docs/archives/
+
+    Archive is compressed with LZMA into a .xz because that is compatible with
+    a reproducible archive. Other formats like .gz are not reproducible due to
+    arbitrary strings and timestamps getting added into their metadata.
```
boegel, line 2799: Hmm, this is new to me, on what source is this based, do we have a reference we can point to?

easybuild/tools/filetools.py
```diff
@@ -2768,37 +2774,82 @@ def get_source_tarball_from_git(filename, target_dir, git_config):
...
+        "Filter out system-dependent data from tarball"
+        # contents of '.git' subdir are inherently system dependent
+        if "/.git/" in tarinfo.name or tarinfo.name.endswith("/.git"):
+            return None
+        # set timestamp to epoch 0
+        tarinfo.mtime = 0
+        # reset file permissions by applying go+u,go-w
+        user_mode = tarinfo.mode & stat.S_IRWXU
+        group_mode = (user_mode >> 3) & ~stat.S_IWGRP  # user mode without write
+        other_mode = group_mode >> 3  # user mode without write
```
boegel, line 2811:
```suggestion
        other_mode = group_mode >> 3  # same as group mode
```

easybuild/tools/filetools.py
```diff
@@ -2768,37 +2774,82 @@ def get_source_tarball_from_git(filename, target_dir, git_config):
...
+        tarinfo.mode = (tarinfo.mode & ~0o77) | group_mode | other_mode
+        # reset ownership numeric UID/GID 0
+        tarinfo.uid = tarinfo.gid = 0
+        tarinfo.uname = tarinfo.gname = ""
```
boegel, line 2815: Hmm, why does empty user/group name make sense, what does this imply?

easybuild/framework/easyblock.py
```diff
@@ -358,34 +359,49 @@ def get_checksum_for(self, checksums, filename=None, index=None):
...
-            filename = filename['filename']
+            chksum_input = filename['filename']
+            chksum_input_git = filename.get('git_config', None)
+        # early return if no filename given
+        if chksum_input is None:
+            self.log.debug("Cannot get checksum without a file name")
+            return None
+
+        if sys.version_info[0] >= 3 and sys.version_info[1] < 9:
+            # ignore any checksum for given filename due to changes in python/cpython#90021
```
boegel, line 374:
```suggestion
            # ignore any checksum for given filename due to changes in https://github.com/python/cpython/issues/90021

File: easybuild/tools/filetools.py [hunk 9]
```diff
@@ -2768,37 +2774,82 @@ def get_source_tarball_from_git(filename, target_dir, git_config):
...
+    archive_filename = archive_name + archive_ext
+    archive_path = archive_filename if archive_dir is None else os.path.join(archive_dir, archive_filename)
+
+    if build_option('extended_dry_run'):
+        # early return in dry run mode
+        dry_run_msg("Archiving '%s' into '%s'...", dir_path, archive_path)
+        return archive_path
+
+    # TODO: replace with TarFile.add(recursive=True) when support for Python 3.6 drops
+    # since Python v3.7 tarfile automatically orders the list of files added to the archive
```
boegel, line 2831: Do we have a reference for this?

easybuild/tools/filetools.py
```diff
@@ -2768,37 +2774,82 @@ def get_source_tarball_from_git(filename, target_dir, git_config):
...
+
+    _log.info("Archiving '%s' into '%s'...", dir_path, archive_path)
+    with tarfile.open(archive_path, "w:xz", format=tarfile.GNU_FORMAT, encoding="utf-8", preset=6) as archive:
+        for filepath in dir_files:
+            # archive with target directory in its top level, remove any prefix in path
+            file_name = os.path.relpath(filepath, start=dir_path_prefix)
+            archive.add(filepath, arcname=file_name, recursive=False, filter=archive_filter)
+            _log.debug("File/folder added to archive '%s': %s", archive_filename, filepath)
+
+    _log.info("Archive '%s' created successfully", archive_filename)
```
boegel, line 2848:
```suggestion
    _log.info(f"Archive '{archive_path}' created successfully")
```

boegel (changes requested):

easybuild/tools/filetools.py
```diff
@@ -2768,37 +2774,82 @@ def get_source_tarball_from_git(filename, target_dir, git_config):
...
+    reproducible = not keep_git_dir  # presence of .git directory renders repo unreproducible
+    archive_path = make_archive(repo_path, archive_name=filename, archive_dir=target_dir, reproducible=reproducible)
 
     # cleanup (repo_name dir does not exist in dry run mode)
     remove(tmpdir)
 
     return archive_path
 
 
+def make_archive(dir_path, archive_name=None, archive_dir=None, reproducible=False):
```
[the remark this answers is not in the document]
boegel: Or maybe even `make_reproducible_tar_xz`

boegel:
@lexming We need to make sure that `git_config` entries that still use `.tar.gz` as extension don't leading to clone+tar+xz every time, we also need to check whether the source filename with `.tar.xz` already exists?

easybuild/tools/filetools.py
boegel: Rather than printing a warning, we should enforce the use of `.tar.xz` as extension here (to "ease" the transition, in some way)

boegel (changes requested):

boegel:
TODO here:

* change so function that creates tarballs for `git_config` sources honors the provided file extension;
* only `.tar.xz` can be produced reproducibly (and only using Python >= 3.9);

easybuild/tools/filetools.py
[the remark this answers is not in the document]
lexming: as discussed, extension is no longer hardcoded and now we actually support more extensions than 4.9. See https://github.com/easybuilders/easybuild-framework/pull/4660/commits/1c44ad74338721435b272f1ccde1b58c26965045

easybuild/framework/easyblock.py
lexming: done in https://github.com/easybuilders/easybuild-framework/pull/4660/commits/fbbb63291b17d597b83d47dc0c4234ef6f70452a

easybuild/framework/easyblock.py
lexming: fixed in https://github.com/easybuilders/easybuild-framework/pull/4660/commits/4f30b16e2a60d91e1d35c87027270e993bc4d74b

easybuild/tools/filetools.py
lexming: as discussed, the archive extension is no longer hardcoded or ignored. This PR now improves `get_source_tarball_from_git` to support more extensions than what we do in 4.9. See https://github.com/easybuilders/easybuild-framework/pull/4660/commits/1c44ad74338721435b272f1ccde1b58c26965045

easybuild/tools/filetools.py
lexming: fixed in https://github.com/easybuilders/easybuild-framework/pull/4660/commits/ed2607598d3cb0327857a1f27b5e5875ad4f7911

easybuild/tools/filetools.py
lexming: I added a link to the issue upstream covering the undeterministic behaviour of gzip. See https://github.com/easybuilders/easybuild-framework/pull/4660/commits/86584970a92fa6573ffba6334af76d995469efd3

easybuild/tools/filetools.py
lexming: This results in the same archive as running `tar --owner=0 --group=0 --numeric-owner`, which are the usual options for reproducible tarballs. I added this information in the comment. See https://github.com/easybuilders/easybuild-framework/pull/4660/commits/85c96a159846403e8974aca319f7a65ebe92d7e3

easybuild/framework/easyblock.py
lexming: fixed in https://github.com/easybuilders/easybuild-framework/pull/4660/commits/fb7535373fed06bda9f3375a11d6522f39eb7baa

easybuild/tools/filetools.py
lexming: RTFM :smile: 
fixed in https://github.com/easybuilders/easybuild-framework/pull/4660/commits/d9e7682139b47ed1778e3851125d72ca1f73c6ea

easybuild/tools/filetools.py
lexming: fixed in https://github.com/easybuilders/easybuild-framework/pull/4660/commits/1c44ad74338721435b272f1ccde1b58c26965045

lexming:
@boegel this is ready on my side, summary of the final behaviour:

* (no change) `git_config` continues to provide filename with extension
* (enhancement) `get_source_tarball_from_git` will respect file extension
* (enhancement) checksums of sources in `git_config` will be verified whenever:
  1. checksum is provided in easyconfig
  2. Python 3.9+
* (enhancement) regular archives of git repos can now be created for uncompressed `.tar` or compressed tarballs in GZIP, XZ and BZIP2
* (enhancement) reproducible archives are supported for:
  1. Uncompressed tarball or compressed with XZ
  2. Python 3.9+

This means that existing easyconfigs with `.tar.gz` sources and no checksums will continue to behave in the same way (no verification) across all Python versions. Only when a checksum is manually added, it will trigger checksum versification on Python 3.9+.

We can migrate existing easyconfigs to `.tar.xz` and include a checksum to start using reproducible tarballs. Users on older Pythons will see a deprecation warning and fallback to current behaviour (no verification) and installation will proceed.

boegel (approved):
lgtm

PR #4661, opened 2024-09-27
add support for injecting checksums for cargo crates

Flamefire:
Even though the `crates` are recognized as sources for `Cargo` easyconfigs the checksums are not written, at least when neither 'source_urls', 'sources' nor 'patches' are present.

Handle `crates` like `sources` and add test.


easybuild/framework/easyblock.py
```diff
@@ -4751,8 +4751,8 @@ def make_checksum_lines(checksums, indent_level):
         if app.src:
             placeholder = '# PLACEHOLDER FOR SOURCES/PATCHES WITH CHECKSUMS'
 
-            # grab raw lines for source_urls, sources, patches
-            keys = ['patches', 'source_urls', 'sources']
+            # grab raw lines for the following params
+            keys = ['source_urls', 'sources', 'crates', 'patches']
```
boegel, line 4755: Hmm, this approach is a bit weird, as it breaks the boundary between framework and easyblocks, since `crates` is not a general easyconfig parameter...

Thoughts on this @Micket?

boegel: Maybe we come up with an approach that would also work for `components` (in `Bundle` easyblock)?

Flamefire: As discussed in the confcall I added a method `src_parameter_names` to the easyblock class.

In the `Cargo` easyblock we can override this to also return "crates".

For `components` I don't think this is useful. For now they look like:
```
components = [
    (name, version, {
        'source_urls': ['https://poppler.freedesktop.org/'],
        'sources': [SOURCE_TAR_XZ],
        'checksums': ['86b09e5a02de40081a3916ef8711c5128eaf4b1fc59d5f87d0ec66f04f595db4'],
        'configopts': "-DENABLE_BOOST=ON",
    }),
    ('poppler-data', '0.4.10', {
        'source_urls': ['https://poppler.freedesktop.org/'],
        'sources': [SOURCE_TAR_GZ],
        'checksums': ['6e2fcef66ec8c44625f94292ccf8af9f1d918b410d5aa69c274ce67387967b30'],
    }),
]
```

While the `Bundle` easyblock seemingly supports specifying `checksums` for everything at the top-level (outside the `components`) I don't think it actually does: It creates a full list of sources, patches and checksums during `__init__` but the actual checksum check (at least the contrib check) requires them in each component as above.

However with this design it would be possible to add the checksum list after the components but then we'd need to include logic to remove them from each component

The general approach is different here anyway: `crates` are just another way to specify `sources` while `components` are similar to `exts_list` which yield actual easyblock instances.

Flamefire:
Rebased to resolve conflicts with recent updates.

Flamefire:
We should exclude the rate limit error from tests like `test_from_commit`

Flamefire:
Rebased to resolve conflicts with recent updates.

boegel:
@Flamefire We need a companion PR for `Cargo` easyblock that redefines this new `src_parameter_names` static method, correct?

boegel:
> @Flamefire We need a companion PR for `Cargo` easyblock that redefines this new `src_parameter_names` static method, correct?

Ah, no, it's already there in `Cargo`, it snuck in through this PR:
- https://github.com/easybuilders/easybuild-easyblocks/pull/3665

test/framework/sandbox/easybuild/easyblocks/generic/cargo.py
```diff
@@ -0,0 +1,55 @@
...
+    def extra_options():
+        """Custom easyconfig parameters for bar."""
+        extra_vars = {
+            'crates': [[], "List of (crate, version, [repo, rev]) tuples to use", CUSTOM],
+        }
+        return EasyBlock.extra_options(extra_vars)
+
+    @staticmethod
+    def src_parameter_names():
+        return super(Cargo, Cargo).src_parameter_names() + ['crates']
```
boegel: Just making sure this matches what's used in https://github.com/easybuilders/easybuild-easyblocks/pull/4062
```suggestion
        return EasyBlock.src_parameter_names() + ['crates']
```

boegel (changes requested):
lgtm

PR #4662, opened 2024-09-29
Add bash functions to env.sh

Micket:
Adds bash functions to generated environment file.


Micket:
I did some tests and this works fine for me.

boegel (approved):
lgtm

PR #4664, opened 2024-10-01
Add flushing and possible closing of `stdin` on `run_shell_cmd`

Crivella:
When using `streaming_output` and writing directly from the STDIN process pipe, commands that attempt to read from STDIN until EOF, or attempt to read a larger size than what is being passed will block.

An example of a bash command that will result in a blocking behavior

```
cat -
```

when executed via

```
cmd = f'cat -'
run_shell_cmd(cmd, stdin='Hello\nWorld\n')
```

Flushing and closing (in case of a non-interactive run) the STDIN as proposed in the changes would allow the example code to run

A test case has been added to catch such behavior

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -414,6 +414,9 @@ def run_shell_cmd(cmd, fail_on_error=True, split_stderr=False, stdin=None, env=N
 
         if stdin:
             proc.stdin.write(stdin)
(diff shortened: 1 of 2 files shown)


Micket:
Seems important and straight forward enough to fix for 5.0

boegel (approved):
lgtm

PR #4665, opened 2024-10-01
Proposed changest to add a `timeout` to `run_shell_cmd`

Crivella:
This changes would implement the capability for `run_shell_cmd` to fail after a user-specified `timeout`.
This would be useful for commands that are known to possibly hang to fail gracefully and report meaningful logs, instead of blocking EB indefinitely.

The PR leverages the already implemented timeout in [Popen.communicate](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) (since python 3.3) for non-interactive/streamed usage of `run_shell_cmd`

In case of an interactive/streamed run the following has been added:

- 2 functions (+1 helper) in `run.py`
  - `read_pipe`: read from a pipe using a thread and raise a `TimeoutError` in case the read operation is taking longer than the specified timeout
  - `terminate_process`: attempt to terminate a process gracefully by using `Popen.terminate` first and `Popen.kill` after. Raises an EasybuildError if the process is still alive after `timeout`
- 3 checks inside `run_shell_cmd`
  - Check at the beginning of the while loop to see if the process has been alive for more than `timeout` time
  - Check when reading the stdout to ensure the operation does not block for more than what would make the process time exceed `timeout`
  - Check when reading the stderr to ensure the operation does not block for more than what would make the process time exceed `timeout`
- 2 tests to ensure that both usages succeds/fails when expected.

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -49,6 +49,7 @@ import subprocess
 import sys
 import tempfile
 import time
+import threading
 from collections import namedtuple
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


boegel:
@Crivella This will need some love, due to the changes that were made in #4755

I consider this a nice-to-have for EasyBuild 5.0 since I'm not aware of any urgent need for this (though I agree it would be nice to support this), so I won't let this block the release of EasyBuild v5.0.0

Crivella:
I've rebased this on top of the latest 5.0.x and adapted the code to use  the changes introduced by https://github.com/easybuilders/easybuild-framework/pull/4755

boegel:
@Crivella I changed to target branch in this PR from 5.0.x to develop, you should synchronize your PR branch with current develop branch (which has received a massive update after the release of EasyBuild v5.0.0, see https://github.com/easybuilders/easybuild-framework/pull/4825)

boegel:
@Crivella Please sync with `develop`, failing code style checks have been resolved in #4831

PR #4666, opened 2024-10-03
include path to `cmd.sh` script in output generated by `run_shell_cmd` when a command fails + use colors: red for ERROR line, yellow for path to output files + `cmd.sh` script

boegel:
~(WIP because tests will need some love)~

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -45,7 +45,6 @@ import shlex
 import shutil
 import string
 import subprocess
-import sys
 import tempfile
 import time
 from collections import namedtuple
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -88,6 +88,19 @@ CACHED_COMMANDS = (
 
 RunShellCmdResult = namedtuple('RunShellCmdResult', ('cmd', 'exit_code', 'output', 'stderr', 'work_dir',
                                                      'out_file', 'err_file', 'cmd_sh', 'thread_id', 'task_id'))
+RunShellCmdResult.__doc__ = """A namedtuple that represents the result of a call to run_shell_cmd,
+with the following fields:
+- cmd: the command that was executed;
+- exit_code: the exit code of the command (zero if it was succesful, non-zero if not);
+- output: output of the command (stdout+stderr combined, only stdout if stderr was caught separately);
+- stderr: stderr output produced by the command, if caught separately (None otherwise);
+- work_dir: the working directory of the command;
+- out_file: path to file with output of command (stdout+stderr combined, only stdout if stderr was caught separately);
+- err_file: path to file with stderr output of command, if caught separately (None otherwise);
+- cmd_sh: path to script to set up interactive shell with environment in which command was executed;
+- thread_id: thread ID of command that was executed (None unless asynchronous mode was enabled for running command);
+- task_id: task ID of command, if it was specified (None otherwise);
+"""
 
 
 class RunShellCmdError(BaseException):
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -90,7 +90,7 @@ RunShellCmdResult = namedtuple('RunShellCmdResult', ('cmd', 'exit_code', 'output
 RunShellCmdResult.__doc__ = """A namedtuple that represents the result of a call to run_shell_cmd,
 with the following fields:
 - cmd: the command that was executed;
-- exit_code: the exit code of the command (zero if it was succesful, non-zero if not);
+- exit_code: the exit code of the command (zero if it was successful, non-zero if not);
 - output: output of the command (stdout+stderr combined, only stdout if stderr was caught separately);
 - stderr: stderr output produced by the command, if caught separately (None otherwise);
 - work_dir: the working directory of the command;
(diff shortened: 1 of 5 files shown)


easybuild/tools/run.py
```diff
@@ -112,31 +114,37 @@ def print(self):
...
             pad_4_spaces(f"exit code                 ->  {self.exit_code}"),
+            pad_4_spaces(f"called from               ->  {called_from_info}"),
             pad_4_spaces(f"working directory         ->  {self.work_dir}"),
         ]
 
         if self.out_file is not None:
             # if there's no separate file for error/warnings, then out_file includes both stdout + stderr
             out_info_msg = "output (stdout + stderr)" if self.err_file is None else "output (stdout)         "
-            error_info.append(pad_4_spaces(f"{out_info_msg}  ->  {self.out_file}"))
+            error_info.append(pad_4_spaces(f"{out_info_msg}  ->  {self.out_file}", color=COLOR_YELLOW))
```
boegel, line 139: Hmm, the yellow here is maybe not a good idea when terminal doesn't have a dark background, should be changed to different color that works on both dark/light background?

easybuild/tools/run.py
```diff
@@ -86,7 +87,7 @@
 )
 
 RunShellCmdResult = namedtuple('RunShellCmdResult', ('cmd', 'exit_code', 'output', 'stderr', 'work_dir',
-                                                     'out_file', 'err_file', 'thread_id', 'task_id'))
+                                                     'out_file', 'err_file', 'cmd_sh', 'thread_id', 'task_id'))
```
lexming, line 90: We need a docstring somewhere defining what each of these items are.

lexming (commented):
LGTM but I'm not seeing any colors although I have `rich` installed. I see the color progress bar during installation, but then I got the following error message:
```
[bold red]ERROR: Shell command failed![/bold red]
    full command              ->  potato &&  ./configure --prefix=/home/lexming/easybuild/install/software/cURL/8.0.1-GCCcore-12.3.0  --build=x86_64-pc-linux-gnu
--host=x86_64-pc-linux-gnu --with-zlib --with-ssl=$EBROOTOPENSSL
    exit code                 ->  127
    called from               ->  'configure_step' function in /home/lexming/src/easybuild-easyblocks/easybuild/easyblocks/generic/configuremake.py (line 326)
    working directory         ->  /home/lexming/.local/easybuild/build/cURL/8.0.1/GCCcore-12.3.0/curl-8.0.1
[bold yellow]    output (stdout + stderr)  ->  /tmp/eb-5zxm3a9y/run-shell-cmd-output/potato-t2b3f6fz/out.txt[/bold yellow]
[bold yellow]    interactive shell script  ->  /tmp/eb-5zxm3a9y/run-shell-cmd-output/potato-t2b3f6fz/cmd.sh[/bold yellow]
```

easybuild/tools/run.py
lexming: Yeah, it might be safer to use blue. Although any sane color theme will have a yellow with enough contrast with the light background.

easybuild/tools/run.py
```diff
@@ -112,31 +114,37 @@ def print(self):
...
-        error_info.extend([
-            pad_4_spaces(f"called from               ->  {called_from_info}"),
-            '',
-        ])
+        if self.cmd_sh is not None:
+            error_info.append(pad_4_spaces(f"interactive shell script  ->  {self.cmd_sh}", color=COLOR_YELLOW))
+
+        error_info.append('')
 
         sys.stderr.write('\n'.join(error_info) + '\n')
```
branfosj, line 149: `sys.stderr.write` does not go through Rich.

branfosj:
```py
from rich.console import Console
error_console = Console(stderr=True)
error_console.print('[bold yellow]text[/bold yellow]')
```

easybuild/tools/run.py
boegel: Looks OK for me with yellow on a white background, and with (standard) blue it looks horrible on dark background, so sticking to yellow...

Screenshot 2024-10-14 at 09 47 14

easybuild/tools/run.py
boegel: Ah, that explains it indeed...

fixed in dd2ebf22b

easybuild/tools/run.py
boegel: fixed in 63ac9e56b

easybuild/tools/run.py
```diff
@@ -86,7 +86,20 @@
 )
 
 RunShellCmdResult = namedtuple('RunShellCmdResult', ('cmd', 'exit_code', 'output', 'stderr', 'work_dir',
-                                                     'out_file', 'err_file', 'thread_id', 'task_id'))
+                                                     'out_file', 'err_file', 'cmd_sh', 'thread_id', 'task_id'))
+RunShellCmdResult.__doc__ = """A namedtuple that represents the result of a call to run_shell_cmd,
+with the following fields:
+- cmd: the command that was executed;
+- exit_code: the exit code of the command (zero if it was succesful, non-zero if not);
```
branfosj, line 93:
```suggestion
- exit_code: the exit code of the command (zero if it was successful, non-zero if not);
```

branfosj (approved):

Issue #4643, opened 2024-09-17
Can't use multiple `--try-amend` statements in Easystack entry


cgross95:
We install MATLAB using something like

``` bash
eb MATLAB-2023b.eb \
    --try-amend="license_server=hostname" \
    --try-amend="license_server_port=12345" \
    --try-amend="key=abc-123"
```

But the equivalent Easystack will only use the last value, e.g., 

``` bash
easyconfigs:
- MATLAB-2023b.eb:
    options:
      try-amend: "license_server=hostname"
      try-amend: "license_server_port=12345"
      try-amend: "key=abc-123"
```

gives a tweaked EasyConfig that only has `key=abc123` at the bottom. `license_server` and `license_server_port` are ignored.

Crivella:
I think this is more a YAML dictionary thing than an easybuild one.
When writing a dictionary in YAML with repeated keys, the last one will be the one considered.

Though it seem that setting it as a list does not lead to the expected behavior

ocaisa:
What about trying
```yaml
easyconfigs:
- MATLAB-2023b.eb:
    options:
      try-amend: 
        - "license_server=hostname\nlicense_server_port=12345\nkey=abc-123"
```
as a hack?

We could probably do something specific for `try-amend` when doing the dict to list argument conversion, I suspect there are very few options that we allow to appear multiple times.

Crivella:
That seem to result to it being assigned as a multi-line string
```
license_server = """hostname
license_server_port=12345
key=abc-123"""
```

Crivella:
Opened a PR with a possible fix https://github.com/easybuilders/easybuild-framework/pull/4667

Crivella:
Fixed in
- #4667

PR #4667, opened 2024-10-04
allow using`amend/try-amend` multiple times in an easystack file entry

Crivella:
Related to:

- #4643 

This change checks if the option in the easystack belongs to a list of options that allows to be defined multiple times (could be defined in a more generic location) and instead of setting it's value as a list, will set it multiple times to the cli arguments.

Possible breaking change:

- Any currently existing easystack that makes use of a pattern like
  ```
    options:
      - try-amend:
        - opt=1
        - 2
        - 3
  ```
  in order to set `--try-amend=opt=1,2,3`  would not work anymore

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -2018,16 +2018,13 @@ def opts_dict_to_eb_opts(args_dict):
                 value = [value]
             for v in value:
                 args.append(option + '=' + str(v))
-            continue
-
-        if isinstance(value, (list, tuple)):
-            value = ','.join(str(x) for x in value)
-
-        if value in [True, None]:
+        elif value in [True, None]:
             args.append(option)
         elif value is False:
             args.append('--disable-' + option[2:])
         elif value is not None:
+            if isinstance(value, (list, tuple)):
+                value = ','.join(str(x) for x in value)
             args.append(option + '=' + str(value))
 
     _log.debug("Converted dictionary %s to argument list %s" % (args_dict, args))
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -2016,8 +2016,7 @@ def opts_dict_to_eb_opts(args_dict):
         if str(arg) in allow_multiple_calls:
             if not isinstance(value, (list, tuple)):
                 value = [value]
-            for v in value:
-                args.append(option + '=' + str(v))
+            args.extend(option + '=' + str(x) for x in value)
         elif value in [True, None]:
             args.append(option)
         elif value is False:
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -2002,6 +2002,7 @@ def opts_dict_to_eb_opts(args_dict):
     :return: a list of strings representing command-line options for the 'eb' command
     """
 
+    allow_multiple_calls = ['amend', 'try-amend']
     _log.debug("Converting dictionary %s to argument list" % args_dict)
     args = []
     for arg in sorted(args_dict):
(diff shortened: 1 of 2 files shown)


easybuild/tools/options.py
```diff
@@ -2011,6 +2012,14 @@ def opts_dict_to_eb_opts(args_dict):
             prefix = '--'
         option = prefix + str(arg)
         value = args_dict[arg]
+
+        if str(arg) in allow_multiple_calls:
+            if not isinstance(value, (list, tuple)):
+                value = [value]
+            for v in value:
+                args.append(option + '=' + str(v))
+            continue
```
ocaisa, line 2021: I think some people would not be happy to break out of the loop in the middle like this. What about instead leaving the list intact below for the `allow_multiple_calls` special cases, and then adding a new `elif` clause at the end of the loop to handle this new case

Crivella: Umh it would still be cleaner to check for `allow_multiple_calls` first and than check if the value is a list.
If we want to avoid `continue` I think the other 2 `if` can become `elif` as all the value checked after are not compatible anyway

ocaisa: Yeah, that was also my original thinking, move the 
```
        if isinstance(value, (list, tuple)):
            value = ','.join(str(x) for x in value)
```
into the main if clause and deal with the differences there

Crivella: Should be fixed with https://github.com/easybuilders/easybuild-framework/pull/4667/commits/550d96cb8aeeab15644d995d8dc9f4c1ca5f549f

ocaisa (approved):
LGTM but maybe @boegel wants to take a look as there is a (_very_ unlikely to be used) breaking change in here (as mentioned in https://github.com/easybuilders/easybuild-framework/pull/4667#issue-2566471841)

cgross95:
Thank you for working on this! I just tested and can verify that it works exactly as desired for our use case. If this does get merged in as is, I wrote up a documentation PR (that I am happy to edit as desired):
- https://github.com/easybuilders/easybuild-docs/pull/281

easybuild/tools/options.py
```diff
@@ -2011,14 +2012,19 @@ def opts_dict_to_eb_opts(args_dict):
...
         value = args_dict[arg]
-        if isinstance(value, (list, tuple)):
-            value = ','.join(str(x) for x in value)
 
-        if value in [True, None]:
+        if str(arg) in allow_multiple_calls:
+            if not isinstance(value, (list, tuple)):
+                value = [value]
+            for v in value:
+                args.append(option + '=' + str(v))
```
boegel, line 2020: We avoid using singe-letter variables outside of list comprehensions, so:

```suggestion
            args.extend(option + '=' + str(x) for x in value)

File: easybuild/tools/options.py [hunk 3]
```diff
@@ -2011,14 +2012,19 @@ def opts_dict_to_eb_opts(args_dict):
...
+        elif value in [True, None]:
             args.append(option)
         elif value is False:
             args.append('--disable-' + option[2:])
         elif value is not None:
+            if isinstance(value, (list, tuple)):
+                value = ','.join(str(x) for x in value)
```
boegel, line 2027: Can we also cover this with an enhanced test?

boegel (changes requested):

boegel:
> LGTM but maybe @boegel wants to take a look as there is a (_very_ unlikely to be used) breaking change in here (as mentioned in [#4667 (comment)](https://github.com/easybuilders/easybuild-framework/pull/4667#issue-2566471841))

It's OK I think, especially since easystack files are (still) an experimental feature currently...

That said, it's high time we remove the experimental tag for easystack files, definitely for EasyBuild 5.0...

easybuild/tools/options.py
Crivella: I think that is already covered with https://github.com/easybuilders/easybuild-framework/blob/3d29b0a55c803c6ac68786c77b8e65a9c90b814b/test/framework/options.py#L7239

Crivella: But i can add a separate test if needed

bartoldeman:
@boegel does this still need work or can it be merged now?

boegel (approved):
lgtm

PR #4670, opened 2024-10-06
add support for `%(rpath_enabled)s` template value

boegel:
There are situations where we should be able to detect in an easyconfig file whether or not RPATH linking is used.

This can be leveraged in easyconfigs where pre-built binaries or libraries are included, like (for example):

```python
postinstallcmds = [
    "if %(rpath_enabled)s; then patchelf %(installdir)s/...; fi",
]

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -101,7 +101,7 @@ TEMPLATE_NAMES_DYNAMIC = [
     ('cuda_sm_space_sep', "Space-separated list of sm_* values that correspond with CUDA compute capabilities"),
     ('mpi_cmd_prefix', "Prefix command for running MPI programs (with default number of ranks)"),
     # can't be a boolean (True/False), must be a string value since it's a string template
-    ('rpath', "String value indicating whether or not RPATH linking is used ('true' or 'false')"),
+    ('rpath_enabled', "String value indicating whether or not RPATH linking is used ('true' or 'false')"),
     ('software_commit', "Git commit id to use for the software as specified by --software-commit command line option"),
     ('sysroot', "Location root directory of system, prefix for standard paths like /usr/lib and /usr/include"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by boegel** [comment]
One use case for this is `NextPolish-1.4.1-GCCcore-12.3.0.eb`, which includes pre-built libraries like `nextpolish1.so` which don't have an RPATH section, so the sanity check fails on them.
Rightfully so, because if `$LD_LIBRARY_PATH` is not set, they're relying on libraries provided by the OS, like `/lib64/libbz2.so.1`, `/lib64/liblzma.so.5`, and `/lib64/libz.so.1`.

**Comment by boegel** [comment]
I've re-targeted this to `develop`, since it would be useful short term in the context of https://github.com/easybuilders/easybuild-easyconfigs/pull/21588

cc @jfgrimm

**Comment by jfgrimm** [comment]
as suggested in the conf call, this example could be shortened to:
```
postinstallcmds = [
    "if true; then patchelf %(installdir)s/...; fi",
]
```
if we set it to `true` or `false`

Alternatively, we could consider a different approach and do something more like the `ARCH` constant we have in easyconfigs? e.g. from `UCX-1.15.0-GCCcore-13.2.0.eb`
```
if ARCH == "riscv64":                                                                                                   
    patches += ['UCX-1.15.0-add_riscv_support.patch']                                                                   
    checksums += [{'UCX-1.15.0-add_riscv_support.patch':                                                                
                   '700640d469f441f3ee2c9d42e3a533f78f0c892c1d3c221aebd41279668d118e'}]                                 
    preconfigopts = 'autoreconf -fi && '
```
we could similarly have:
```
if USE_RPATH:
    postinstallcmds = ...
```

jfgrimm:
@boegel fair enough, I had thought it might be awkward with the potential that people hook on a case-by-case basis

jfgrimm (approved):

PR #4671, opened 2024-10-08
don't raise error when required extensions are not found when installing extensions in parallel

boegel:
Raising an error effectively makes it impossible to use `--parallel-extensions-install` on bundle easyconfigs like `R-bundle-CRAN-2024.06-foss-2023b.eb` and `R-bundle-Bioconductor-3.18-foss-2023a-R-4.3.2.eb`.

There's actually no reason to give up when extensions that are required dependencies for an extension that will be installed are not included in `exts_list`, since they're probably provided via a dependency like `R` or `R-bundle-CRAN`.
Note that this is exactly what we do currently when installing extensions sequentially...

If a dependency of a particular extension is totally missing, the installation of that extension will fail as it does when installing sequentially.

Note: for using `--parallel-extensions-install` on `R-bundle-CRAN` (or older `R` easyconfigs that include `Rmpi` and `Rserve`), an additional bug fix is required, see:
* https://github.com/easybuilders/easybuild-easyblocks/pull/3474

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2122,8 +2122,10 @@ class EasyBlock(object):
                     # to avoid getting stuck in an infinite loop!
                     missing_deps = [x for x in required_deps if x not in all_ext_names]
                     if missing_deps:
-                        raise EasyBuildError("Missing required dependencies for %s are not going to be installed: %s",
-                                             ext.name, ', '.join(missing_deps))
+                        msg = f"Missing required extensions for {ext.name} not found "
+                        msg += "in list of extensions being installed, let's assume they are provided by "
+                        msg += "by dependencies and proceed: " + ', '.join(missing_deps)
+                        self.log.info(msg)
                     else:
                         self.log.info("Required dependencies missing for extension %s (%s), adding it back to queue...",
                                       ext.name, ', '.join(pending_deps))
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2118,16 +2118,23 @@ class EasyBlock(object):
                 # if some of the required dependencies are not installed yet, requeue this extension
                 elif pending_deps:
 
-                    # make sure all required dependencies are actually going to be installed,
-                    # to avoid getting stuck in an infinite loop!
+                    # check whether all required dependency extensions are actually going to be installed;
+                    # if not, we assume that they are be provided by dependencies;
                     missing_deps = [x for x in required_deps if x not in all_ext_names]
                     if missing_deps:
                         msg = f"Missing required extensions for {ext.name} not found "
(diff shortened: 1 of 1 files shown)


boegel:
Hmm, this is not enough, since eventually things will get stuck because no new extensions are being picked up anymore, need to check why that's happening:
```
Not installing extensions (yet) ━━━━━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3:19:20
```

boegel:
This seems to be working well now with `eb R-bundle-Bioconductor-3.18-foss-2023a-R-4.3.2.eb --parallel-extensions-install --experimental`, but it does seem to reveal a bug:

```
ERROR: Shell command failed!
    full command              ->  tar --wildcards --extract --file /data/gent/400/vsc40023/EasyBuild/sources/r/R-bundle-Bioconductor/extensions/RcisTarget_1.22.0.tar --to-stdout '*/DESCRIPTION'
    exit code                 ->  2
    working directory         ->  /tmp/vsc40023/easybuild-framework
    output (stdout + stderr)  ->  /tmp/eb-4106xvti/run-shell-cmd-output/tar-9e1hczim/out.txt
    called from               ->  'required_deps' function in /user/gent/400/vsc40023/easybuild/eb5/easybuild-easyblocks/easybuild/easyblocks/generic/rpackage.py (line 227)
```

There's no `DESCRIPTION` file in the `RcisTarget` source tarball, and we fail hard on that, we probably shouldn't...

The issue is that the `DESCRIPTION` file is not in a subdirectory in `RcisTarget_1.22.0.tar` , so  `tar xfv RcisTarget_1.22.0.tar --to-stdout DESCRIPTION` does work just fine.

Always using `--to-stdout *DESCRIPTION` works too, so I'll strip out the `/` preceding `DESCRIPTION`...

boegel:
Extensive test report for bundles of R packages on top of this is available at https://github.com/easybuilders/easybuild-easyblocks/pull/3490#issuecomment-2419147561, which shows this is good to go...

easybuild/framework/easyblock.py
```diff
@@ -2118,20 +2118,33 @@ def update_exts_progress_bar_helper(running_exts, progress_size):
                 # if some of the required dependencies are not installed yet, requeue this extension
                 elif pending_deps:
 
-                    # make sure all required dependencies are actually going to be installed,
-                    # to avoid getting stuck in an infinite loop!
+                    # check whether all required dependency extensions are actually going to be installed;
+                    # if not, we assume that they are be provided by dependencies;
```
akesandgren, line 2122: "we assume that they are be provided" -> "we assume that they are to be provided" (or just "we assume that they are provided")

akesandgren (changes requested):

akesandgren (approved):

PR #4673, opened 2024-10-08
mark easystack support as being mature, since it's no longer experimental

boegel:


Micket (approved):
lgtm

boegel:
I really want to mark easystack support as non-experimental in EasyBuild 5.0, so I think we should already go ahead and merge this even without having #4667 merged (since that's just an enhancement)

ocaisa:
#4667 is ready to merge IMHO, it includes tests. If we make easystack support stable then that PR would introduce an unintended regression

boegel:
> #4667 is ready to merge IMHO, it includes tests. If we make easystack support stable then that PR would introduce an unintended regression

@ocaisa You're right, I've merged #4667, so I think this is good to go too (the changes in #4667 will be collapsed into `5.0.x` branch, or the other way around, soon)

ocaisa (approved):

PR #4675, opened 2024-10-09
fix color tags set by colorize when rich is available

lexming:
This has to be reversed for colorize to work as intended.

I have `rich` installed and I'm getting colorless string tags:
```
[bold red]ERROR: Shell command failed![/bold red]
```


lexming:
This is actually fixing the issue the wrong way, reversing the if uses ANSI escape codes instead of `rich`.

lexming:
This is actually fixing the issue the wrong way, reversing the `if` uses ANSI escape codes instead of `rich`.

PR #4678, opened 2024-10-11
fix dry-run output when using `multi_deps`

Flamefire:
When using `multi_deps` the dry-run sanity check step should be run multiple times just as is done in the real build.
However the decision which method to call is made before calling `_sanity_check_step_multi_deps` and that doesn't check for dry-run.

Factor out a dispatch method that can be called during the iteration of the multiple dependencies and use that in both cases.


boegel (approved):
lgtm

PR #4677, opened 2024-10-10
Add `resolve_template` method to `EasyConfig` class

Flamefire:
(created using `eb --new-pr`)

Not only does this reduce the (non-trivial) code duplication, it also allows e.g. easyblocks to resolve a template without setting an EC parameter. e.g. `self.cfg.resolve_template('lib/%(pyshortver)s')` might be useful.

PR #4679, opened 2024-10-11
Allow templates in `custom_paths` & `custom_commands` sanity-check arguments

Flamefire:
E.g. for Python packages we (almost) always want to check for `lib/python%(pyshortver)s` and potentially add additional paths from EasyConfigs.
Currently a workaround is used that sets the parameter in the easyconfig when it isn't set already which is against the semantics of `enhance_sanity_check`.

This change allows this in a trivial way.

The added test just uses `multi_deps` which covers the case without it already, so no need for 2 similar ones.
To make it "realistic" add dummy Python ECs to use as dependencies such that we can use `%(pyshortver)s`

Requires 
- [x] #4677 
- [x] #4678 

I included those PRs here though such that CI passes and either all commits can be reviewed together or just the last one which is the described change.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3362,7 +3362,7 @@ class EasyBlock(object):
             # if no sanity_check_paths are specified in easyconfig,
             # we fall back to the ones provided by the easyblock via custom_paths
             if custom_paths:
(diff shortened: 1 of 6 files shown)


boegel (approved):
lgtm

PR #4682, opened 2024-10-15
updates and fixes for findPythonDeps script

Flamefire:
I noticed that the script failed to detect e.g. `jupyter-core` because the `package_name` entry is `jupyter_core`. However that isn't consistent:

- "PyQt5-sip":    'key': 'pyqt5-sip',    'package_name': 'PyQt5-sip'
- "jupyter-core": 'key': 'jupyter-core', 'package_name': 'jupyter_core'

I fixed that and used the opportunity to generally do some updates. In total:

- Require Python ~3.3~ 3.6 (due to f-strings and `subprocess.DEVNULL` [3.3])
- Add/update documentation
- Use f-strings for formatted output
- Fix detecting packages where the name might appear as the "key" not the "package_name"
- Set PYTHONNOUSERSITE
- Use venv module instead of virtualenv to require less packages

This should make the script more robust and easier to understand and maintain. Especially the description at the top should be very helpful for those that don't know it yet.

I used PyLint to check the script. In one place there was an unused argument that should have been used but wasn't. Putting the "main" code into a function allowed to detect that as there are no more global variables.

### DIFF
--- a/easybuild/scripts/findPythonDeps.py
+++ b/easybuild/scripts/findPythonDeps.py
@@ -1,5 +1,21 @@
 #!/usr/bin/env python
 
+"""
+Find Python dependencies for a given Python package after loading dependencies specified in an EasyConfig.
+This is intended for writing or updating PythonBundle EasyConfigs:
+    1. Create a EasyConfig with at least 'Python' as a dependency.
+       When updating to a new toolchain it is a good idea to reduce the dependencies to a minimum
+       as e.g. the new "Python" module might have different packages included.
+    2. Run this script
+    3. For each dependency found by this script search existing EasyConfigs for ones providing that Python package.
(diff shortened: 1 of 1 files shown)


Micket:
> Require Python 3.3

Typo (given that you use f-strings, so it would be python3.6 at least)?

Flamefire:
> Typo (given that you use f-strings, so it would be python3.6 at least)?

Yes, wasn't aware that f-strings were so late. Also the change from virtualenv to venv wasn't complete.

The script will (also) be run with a loaded Python module. So after this change it can no longer be used to find packages for Python 2 or < 3.6. That is 2018b for Python2 and 2016b for Python 3.5. I'd say this is fine as there is likely no one adding ECs for those.

Micket (approved):
lgtm

PR #4686, opened 2024-10-20
simplify code for determining the PYTHONPATH module entries

Flamefire:
The check whether the path(s) have already been added is redundant as that is done in the module_generator class.

The whole code with all the checks is superflous if there are no python paths to add.

To avoid excessive indentation return early with an empty list in that case and same for Python installations for consistency.

`prepend_paths` can return an empty string when everything is filtered. As we return a list here (we might be better of returning a string instead) return an empty list to avoid empty lines being written if `'\n'` was used. -> TODO for later

Best to view with whitespace changes ignored

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1399,43 +1399,40 @@ class EasyBlock(object):
         Add lines for module file to update $PYTHONPATH or $EBPYTHONPREFIXES,
         if they aren't already present and the standard lib/python*/site-packages subdirectory exists
         """
-        lines = []
-        if not os.path.isfile(os.path.join(self.installdir, 'bin', 'python')):  # only needed when not a python install
-            python_subdir_pattern = os.path.join(self.installdir, 'lib', 'python*', 'site-packages')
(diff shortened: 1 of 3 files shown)


Micket:
> The check whether the path(s) have already been added is redundant as that is done in the module_generator class.

which then also prints warnings, which i wanted to avoid

https://github.com/easybuilders/easybuild-framework/blob/b5c1dd97fd34bb15855783002b2874184f1987b1/easybuild/tools/module_generator.py#L231C1-L233C40

Flamefire:
> which then also prints warnings, which i wanted to avoid

Oh I see. I'd avoid to duplicate the logic though, see the new commit(s)

Micket (approved):
lgtm

Micket (changes requested):
actually, while attempting to solve the merge conflict, I think we should also add the new warn_exists keyword to append_paths as well, to make them similar.

Flamefire:
Good point, rebased and implemented

Micket (approved):
lgtm

PR #4688, opened 2024-10-22
add `--keep-debug-symbols` configuration option to set default value of '`debug`' toolchain option, and enable it by default so `-g` is included in `$CXXFLAGS` & co

Micket:
Also defaults to "true" as was discussed for 5.0.x. 
Those concerned with disk space can consider switching this back to false. 

Disclaimer: It's getting late and I'm really not sure if i can get away with inserting the default value of "debug" in the compiler option map like that. Not sure if build_option is guaranteed to be initialized when that static dict is defined.


Micket:
OK the optparser thing won't permit setting bools in any way. Listing options they get converted to strings, which is ugly to work with, and optparse doesn't support `bool` types for selection. So it's a enabling flag instead.

Also, due to how classes are sneakily initialized by EB, the static class variables aren't always defined, so, they might be `None`. As existing codes just `if self.xxxx is not None:`  then I guess i'll just do the same.

easybuild/tools/options.py
```diff
@@ -392,6 +392,7 @@ def override_options(self):
             'debug-lmod': ("Run Lmod modules tool commands in debug module", None, 'store_true', False),
             'default-opt-level': ("Specify default optimisation level", 'choice', 'store', DEFAULT_OPT_LEVEL,
                                   Compiler.COMPILER_OPT_FLAGS),
+            'default-debug-symbols': ("Sets default value of debug toolchain option", None, 'store_true', True),
```
boegel, line 395: Maybe `retain-debug-symbols-by-default` is a better name for this configuration option?

Enabling it by default makes sense I think, there *should* be no fallout of that...

easybuild/tools/toolchain/compiler.py
```diff
@@ -82,7 +82,7 @@ class Compiler(Toolchain):
         'loose': (False, "Loose precision"),
         'veryloose': (False, "Very loose precision"),
         'verbose': (False, "Verbose output"),
-        'debug': (False, "Enable debug"),
+        'debug': (None, "Enable or disables debug symbols"),  # refers to default-debug-symbols build option by default
```
boegel, line 85:
```suggestion
        # default value for 'debug' toolchain option is determined by retain-debug-symbols-by-default configuration option
        'debug': (None, "Retain debug symbols"),
```

easybuild/tools/toolchain/compiler.py
```diff
@@ -175,6 +175,10 @@ def set_variables(self):
 
     def _set_compiler_toolchainoptions(self):
         """Set the compiler related toolchain options"""
+        # Initialize default value of debug symbols based on global build option
```
boegel, line 178:
```suggestion
        # Initialize default value of 'debug' toolchain option based on global configuration option option
```

boegel (changes requested):

easybuild/tools/options.py
Micket: Well I won't insist, though I find the wording of "retain" a bit strange, reads to me that we were intentionally stripping them out.

boegel: As agreed in sync call:
```suggestion
            'keep-debug-symbols': ("Sets default value of debug toolchain option", None, 'store_true', True),
```

boegel (changes requested):

boegel (approved):

PR #4689, opened 2024-10-23
simplify loop in `collect_exts_file_info`

Flamefire:
The loop over `exts_list` uses an excessive amount of branches and hence indentation making it hard to read.

We have 3 simple cases:

1. Extension is a string -> use it as the name and do nothing else
2. extension is neither a string, list nor dict -> Error out
3. extension is a list of only a name -> use as the name and do nothing else

Moving the type check to the top of the loop makes it instantly visible which types are accepted, the `elif`s at the bottom are hard to find after all the indents.

Using `continue` after handling the simple cases reduces the indentation by 2 levels and makes it clear, that nothing else is done for those.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -657,7 +657,12 @@ class EasyBlock:
 
         orig_github_account = self.cfg['github_account']
         for ext in exts_list:
-            if isinstance(ext, (list, tuple)) and ext:
+            if isinstance(ext, str):
+                exts_sources.append({'name': ext})
+            else:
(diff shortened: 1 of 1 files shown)


boegel:

boegel:
@Flamefire CI isn't happy with this?
```
 FAILED (failures=6, errors=2, skipped=3)
```

Flamefire:
@boegel Can look at this please? It is mostly whitespace change so should be fast to review when hiding them, but easily causes conflicts due to that.

I've changed this function recently and it would have been a bit easier to grasp if the small branches weren't so far below.
I guess that's how the missing template-resolving was missed earlier.

Issue #4340, opened 2023-08-30
(re)Design of make_extension_string & friends


wpoely86:
As pointed out in #4331 there is a bit of a weird design around the functions `make_extension_string` and its use in `_generate_extensions_list` in `module_generator.py` from #3697. The list of extensions is built (tuples of name/version), put in a list, joined to a string and then split again into a list. It can't be fixed without change the API and might break somebodies EasyBlock (no EasyBlock in the repo or in a PR touches this code right now).

It would be cleaner to keep it as a list and only convert to a string when it's really needed.

wpoely86:
From @Flamefire:

> I can't remember the exact reasons from back then but yes it seems make_extension_string isn't really needed.
> However we use the functionality in 2 places: EBEXTSLIST and modulegenerator::_generate_extension_list and maybe we anticipated that more things might use it.
> 
> BTW: _generate_extension_list vs _generate_extensions_list in module generator is also...confusing

Flamefire:
I think this is fully adressed with #4690?

lexming:
PR #4690 merged, closing.

PR #4690, opened 2024-10-23
Refactor `make_extension_string`

Flamefire:
The current usage was confusing as it isn't clear what should be overridden and what should be used.
See #4340

- Make `make_extension_string` a free function to just provide a common format
- Promote `make_extension_list` to a public method and document it
- Error when overriding the old methods
- Update the module generator and remove the confusingly named member methods

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -45,6 +45,7 @@ from easybuild.tools import LooseVersion
 from textwrap import wrap
 
 from easybuild.base import fancylogger
+from easybuild.framework.easyblock import make_extension_string
 from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.config import build_option, get_module_syntax, install_path
 from easybuild.tools.filetools import convert_name, mkdir, read_file, remove_file, resolve_path, symlink, write_file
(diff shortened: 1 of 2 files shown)


boegel:
@lexming Can you check with @wpoely86 if this is what he had in mind when he opened #4340 ?

easybuild/tools/module_generator.py
```diff
@@ -671,7 +656,7 @@ def _generate_help_text(self):
             lines.extend(self._generate_section("Compatible modules", compatible_modules_txt))
 
         # Extensions (if any)
-        extensions = self._generate_extension_list()
+        extensions = self.app.make_extension_string()
```
wpoely86, line 659: the line below uses a join on this var which is now a string instead of a list? It also needs to be adjusted?

Flamefire: I had to stare at this for a bit now but it is correct as-is:
- `_generate_extension_list` did [return a string](https://github.com/easybuilders/easybuild-framework/pull/4690/files#diff-4af72e9777d353325a29df400ab4229e14defd653e413db51786dc2172d3baefL619), so behavior is unchanged
- The `join` happens on the result of `wrap` which takes a string and returns a list of strings

wpoely86: aha, I was kinda surprised this didn't trigger a failed test.

wpoely86 (approved):

Issue #4394, opened 2023-12-04
Create an option for a developer mode 


ocaisa:
It would be nice to support a developer mode, which would allow developers to use EasyBuild as part of their own build processes. One way to do this would be make the option (`--developer`)
* Take a copy of the current directory as unpacked sources
* Jump straight to the build step
* Install software and module to a temporary location

branfosj:
A counter suggestion would be a `--start` (equivalent to [`--stop`](https://docs.easybuild.io/partial-installations/?h=st#partial_installation_stop)). I'm not sure I can think of why we'd only let this jump to the build step and not have an option to start at the configure or test steps instead.

ocaisa:
Sorry, I misspoke, I mean jump to right after the extraction step

ocaisa:
I'm not sure about `--start` as obviously the build step has a good chance of affecting the contents of the working directory, and you may not want that. At least if you make a copy, you aren't taking any risks and what you do is still reproducible (to an extent).

Crivella:
So I've opened a draft PR for a possible implementation of the feature and some points that should be discussed in term of design, eg how easy to use do we want the feature to be, vs how versatile

PR #4693, opened 2024-10-30
Feature developer mode

Crivella:
Related to:

- https://github.com/easybuilders/easybuild-framework/issues/4394

I am having a crack at implementing this feature.

Right now the user can specify a path using the `--developer` options (arguably would be better as `--source(s)-from`) taken as the current directory by default, to use with an already existing EC file.

The developer mode would append a `-dev` (maybe `.dev` could be better?) to the version originally found in the EC, and the EB will skip the fetch/patch/extension steps and, instead of extracting an existing source during the `extract_step` will copy the content of the specified source path to the build directory.
The name of the copied directory will be set to match `cfg['start_dir']` if set, or will keep its original name and configure its own start directory 

Tested by compiling the develop branch of the following code-bases using the respective 2023a EC file as a base

- QuantumESPRESSO
- zlib
- HDF5

While implementing this, i had several reflections/ideas that will requires some design choices on what we would want this feature to enable and potential problems with the current implementation

- EC with multiple sources
  - this could still work but it would require the user to have all projects in the directory pointed at with `--developer` (and possibly some minor changes to how `start_dir` is being handled.
- Bundles
  - will not work without mocking `self.src` that should be set by the fetch_step
  - will also have similar problems as the `EC with multiple sources`

The `-dev` suffix to the version will make the version appear as greater then the modified one but lower then the next numerically higher one, so it should not screw up with version checks in the easyblock (and i think having the develop version appear as greater should be the default behavior for this scenario).

I was thinking the most general way to avoid this problem might be to:

- Extend the EC source syntax to add a `develop` options to point to the dir to be copied
- Run al step as normal up to the `extract_step`
- Run the extraction (or implement a dryrun version) to get the name of the extracted folder and than remove/replace it with a copy of the folder pointed by `develop`
- Would also remove the need to handle the `-dev` prefix as the developer could just set his own `versionsuffix`

The current implementation is probably the easiest for the developer perspective (they would only need to add `--developer` to the command, but this might be limited and not work in many scenario.

The more complete approach might add a level of difficulty in that the developer will need to understand the EC file and add the relevant parts.
There is also the fact that this build process would need to download some existing sources without using them just to get the information to reproduce the expected `builddir` layout (not sure if there is a way around this).

An intermediate approach could be to allow `--developer` to define a list of paths and each of them would replace one source in the same order as they are defined in the EC (would act at the level of the `fetch_step`.
The downside would be that the developer would need to be careful about the order of the specified sources

I am not sure which would be the best way to move forward (or if there is some other solution I have not tried/thought of) so I am leaving this as a draft for discussion

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -80,10 +80,10 @@ from easybuild.tools.config import install_path, log_path, package_path, source_
 from easybuild.tools.environment import restore_env, sanitize_env
 from easybuild.tools.filetools import CHECKSUM_TYPE_SHA256
 from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file, change_dir, check_lock
-from easybuild.tools.filetools import compute_checksum, convert_name, copy_file, copy_dir, create_lock, create_patch_info
+from easybuild.tools.filetools import compute_checksum, convert_name, copy_file, create_lock, create_patch_info
 from easybuild.tools.filetools import derive_alt_pypi_url, diff_files, dir_contains_files, download_file
 from easybuild.tools.filetools import encode_class_name, extract_file, find_backup_name_candidate
-from easybuild.tools.filetools import get_cwd, get_source_tarball_from_git, is_alt_pypi_url
+from easybuild.tools.filetools import copy_dir, get_cwd, get_source_tarball_from_git, is_alt_pypi_url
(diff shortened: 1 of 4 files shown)


boegel:
@Crivella I changed to target branch in this PR from `5.0.x` to `develop`, you should synchronize your PR branch with current `develop` branch (which has received a massive update after the release of EasyBuild v5.0.0, see #4825)

boegel:
@Crivella Please sync with `develop`, failing code style checks have been resolved in #4831

PR #4697, opened 2024-11-07
add option --search-path-linker to control how linker options at build time

lexming:
Follow-up to https://github.com/easybuilders/easybuild-framework/pull/4645.

Add option `--search-path-linker` to control how linker options are handled at build time
* same approach used as for `--search-path-cpp-headers`
* default behaviour does not change, we continue to use `LDFLAGS`
* added option to use `LIBRARY_PATH` instead


lexming:
@Micket this should be easy to merge now that `--search-path-cpp-headers` is already merged

PR #4698, opened 2024-11-07
change `Toolchain.get_flag` so it doesn't automatically prepend a dash (`-`) to compiler flags, add deprecation warning for `optarch` value without leading dash, rename `Compiler.COMPILER*_FLAGS` to `Compiler.COMPILER*_OPTIONS`

### LINKED ISSUE #4269: Include entire flag in `optarch` (instead of excluding the initial dash) and maybe make it only per-compiler
The exclusion of the initial - from optarch strings is a confusing historical mistake. It looks especially weird when multiple flags are used, as it just append the initial `-`, so one ends up with stuff like
```
optarch = "march=x86-64 -mtune=generic"
```
which is just odd. If we also try to expand this into other compilers, Java, Rust, Go etc, we'd end up with things like `C target-cpu=generic` for Rust etc, which really just looks really strange.

At the same time, looking at multiple different compilers nvhpc vs gcc vs intel which don't have a unified set of flags (as well as more languages like Rust and Go where flags are completely different per compiler), we really don't have any chance of specifying general optarch; only the per-compiler version makes can work in general.

Micket:
Fixes #4269

It was extremely difficult to read this code, partly because a bunch of variables like:
```
COMPILER_F_UNIQUE_FLAGS = []
```
were, in fact, not flags, but options. I renamed these variables because it was so misleading. To add to this, we also have two COMPILER_FLAGS used for different things (neither of them are flags)

### DIFF
--- a/easybuild/toolchains/compiler/clang.py
+++ b/easybuild/toolchains/compiler/clang.py
@@ -104,7 +104,7 @@ class Clang(Compiler):
 
     COMPILER_CC = 'clang'
     COMPILER_CXX = 'clang++'
-    COMPILER_C_UNIQUE_FLAGS = []
+    COMPILER_C_UNIQUE_OPTIONS = []
 
     LIB_MULTITHREAD = ['pthread']
     LIB_MATH = ['m']
--- a/easybuild/toolchains/compiler/clang.py
+++ b/easybuild/toolchains/compiler/clang.py
@@ -55,10 +55,10 @@ class Clang(Compiler):
         'basic-block-vectorize': (False, "Basic block vectorization"),
     }
     COMPILER_UNIQUE_OPTION_MAP = {
-        'unroll': 'funroll-loops',
-        'loop-vectorize': ['fvectorize'],
-        'basic-block-vectorize': ['fslp-vectorize'],
-        'optarch': 'march=native',
+        'unroll': '-funroll-loops',
+        'loop-vectorize': ['-fvectorize'],
+        'basic-block-vectorize': ['-fslp-vectorize'],
+        'optarch': '-march=native',
         # Clang's options do not map well onto these precision modes.  The flags enable and disable certain classes of
         # optimizations.
         #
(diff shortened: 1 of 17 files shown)


Micket:
I finally found the place that actually adds the extra -

It boils down to using FlagList, and possible CommandFlagList.
Instead, these are not just plain StrList

I need to find a scenario where CommandFlagList was used withmore than a single argument, but it's all so very deeply hidden and obscured it's impossible to tell where the heck these are coming from. Tests indicate it works like
```
        cmd = CommandFlagList(["gcc", "bar", "baz"])
        self.assertEqual(str(cmd), "gcc -bar -baz")
```
so i need to find a scenario where this actually used, to ensure i did fix it correctly.

easybuild/toolchains/compiler/clang.py
```diff
@@ -80,31 +80,31 @@ class Clang(Compiler):
...
-        (systemtools.X86_64, systemtools.INTEL): 'march=x86-64 -mtune=generic',
+        (systemtools.RISCV64, systemtools.RISCV): '-march=rv64gc -mabi=lp64d',  # default for -mabi is system-dependent
+        (systemtools.X86_64, systemtools.AMD): '-march=x86-64 -mtune=generic',
+        (systemtools.X86_64, systemtools.INTEL): '-march=x86-64 -mtune=generic',
     }
 
     COMPILER_CC = 'clang'
     COMPILER_CXX = 'clang++'
-    COMPILER_C_UNIQUE_FLAGS = []
+    COMPILER_C_UNIQUE_OPTIONS = []
```
boegel, line 107: Hmm, do we really need to change this?

We have to take into account that people may have their own/custom implementations of classes that derive from `Compiler`, they will need to be adjusted accordingly (and we have no easy way to produce a clean error message there, I think)...

easybuild/tools/toolchain/compiler.py
```diff
@@ -349,6 +349,11 @@ def _set_optimal_architecture(self, default_optarch=None):
             optarch = self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(self.arch, self.cpu_family)]
 
         if optarch is not None:
+            if not optarch.startswith('-'):
+                print_warning(f'Specifying optarch "{optarch}" without initial dash is deprecated in EasyBuild 5.')
```
boegel, line 353: use `self.log.deprecated` here

test/framework/toolchain.py
```diff
@@ -1272,13 +1271,13 @@ def test_fft_env_vars_intel(self):
     def test_fosscuda(self):
         """Test whether fosscuda is handled properly."""
         tc = self.get_toolchain("fosscuda", version="2018a")
-        opts = {'cuda_gencode': ['arch=compute_35,code=sm_35', 'arch=compute_10,code=compute_10'], 'openmp': True}
+        opts = {'cuda_gencode': ['-arch=compute_35,code=sm_35', 'a-rch=compute_10,code=compute_10'], 'openmp': True}
```
boegel, line 1274: typo?

```suggestion
        opts = {'cuda_gencode': ['-arch=compute_35,code=sm_35', '-arch=compute_10,code=compute_10'], 'openmp': True}

File: test/framework/toolchain.py [hunk 4]
```diff
@@ -3144,21 +3145,6 @@ def test_env_vars_external_module(self):
         expected = {}
         self.assertEqual(res, expected)
 
-    def test_get_flag(self):
```
boegel, line 3147: Why is this test removed?

easybuild/tools/toolchain/compiler.py
```diff
@@ -124,20 +124,20 @@ class Compiler(Toolchain):
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = None
     COMPILER_GENERIC_OPTION = None
 
-    COMPILER_FLAGS = ['debug', 'ieee', 'openmp', 'pic', 'shared', 'static', 'unroll', 'verbose']  # any compiler
-    COMPILER_OPT_FLAGS = ['noopt', 'lowopt', DEFAULT_OPT_LEVEL, 'opt']  # optimisation args, ordered !
-    COMPILER_PREC_FLAGS = ['strict', 'precise', 'defaultprec', 'loose', 'veryloose']  # precision flags, ordered !
+    COMPILER_OPTIONS = ['debug', 'ieee', 'openmp', 'pic', 'shared', 'static', 'unroll', 'verbose']  # any compiler
+    COMPILER_OPT_OPTIONS = ['noopt', 'lowopt', DEFAULT_OPT_LEVEL, 'opt']  # optimisation args, ordered !
+    COMPILER_PREC_OPTIONS = ['strict', 'precise', 'defaultprec', 'loose', 'veryloose']  # precision flags, ordered !
```
boegel, line 129: Do we really need these renames from `_FLAGS` to `_OPTIONS`?

I'd say `-O3` is a compiler flag as well as an option, there's no crystal clear definition there, so let's not cause any "pain" by renaming this (since it's not really internal)

boegel (changes requested):

easybuild/tools/toolchain/compiler.py
boegel: At the very least there should be a warning if `COMPILER_PREC_FLAGS` (or any other `*_FLAGS` being renamed) is still being used...

easybuild/tools/toolchain/compiler.py
```diff
@@ -239,23 +239,34 @@ def _set_compiler_vars(self):
...
         """Collect the flags set, and add them as variables too"""
-
-        flags = [self.options.option(x) for x in self.COMPILER_FLAGS if self.options.get(x, False)]
-        cflags = [self.options.option(x) for x in self.COMPILER_C_FLAGS + self.COMPILER_C_UNIQUE_FLAGS
+        variants = ['', '_F', '_F_UNIQUE', '_C', '_C_UNIQUE', '_OPT', '_PREC']
+        for variant in variants:
+            old_var = getattr(self, f'COMPILER{variant}_FLAGS', None)
+            if old_var is not None:
+                self.log.deprecated(f'COMPILER{variant}_FLAGS has been renamed to COMPILER{variant}_OPTIONS.', '6.0')
+                setattr(self, f'COMPILER{variant}_OPTIONS', old_var)
```
Micket, line 247: Added warnings and backwards compat to handle the renamed variables to address @boegel review

test/framework/toolchain.py
bartoldeman: I think we can keep the test? It's a simpler function of course but can still be tested.

boegel: makes sense, I consider `get_flag` to be part of the public API

boegel (approved):

PR #4699, opened 2024-11-08
enhance `get_software_libdir` to return full paths if requested

Flamefire:
This avoids the need to do something like `os.path.join(get_software_root(foo), get_software_libdir(foo))`

Also removed an extra indent for readability: The `if root: <long codeblock> else: return None` gets easier to read when swapped avoiding the `else`

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1697,7 +1697,7 @@ def get_software_root(name, with_env_var=False):
     return res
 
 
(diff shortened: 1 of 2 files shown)


Micket (dismissed):
lgtm

Flamefire:
At least 5.0x can be merged into this branch, so I hope its fine the other way round later.

boegel:
@Flamefire Can you re-target this to `5.0.x` branch to make it clear it won't cause trouble/merge conflicts there?

PR #4700, opened 2024-11-08
fix robot for dry-run

smoors:


smoors:
closing in favor of https://github.com/easybuilders/easybuild-framework/pull/4704, which targets EB-5.0

PR #4704, opened 2024-11-12
change semantics of `--dry-run`, so it doesn't imply `--robot`

smoors:
`--dry-run` without `--robot` should only show info about the easyconfig itself, not the dependencies.

### DIFF
--- a/easybuild/tools/robot.py
+++ b/easybuild/tools/robot.py
@@ -242,7 +242,7 @@ def dry_run(easyconfigs, modtool, short=False):
     :param short: use short format for overview: use a variable for common prefixes
     """
     lines = []
-    if build_option('robot_path') is None:
+    if build_option('robot') is None:
         lines.append("Dry run: printing build status of easyconfigs")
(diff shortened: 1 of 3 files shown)


Micket (approved):
lgtm i guess, have we discussed this change?

boegel:
Hmm, muscle memory is strong for me for using `-D` and getting full overview of what's needed, but typing an extra `r` isn't too bad, so I don't mind.

I guess we should check with maintainers if anyone else strongly objects, but I can live with it.

@smoors I do wonder what motivated you to look into this...

smoors:
consistency! :)

this is useful when working with an easystack file, and you want to check for each easyconfig in the easystack if it has been installed and get the corresponding module name. you can of course do a grep on the output, but it’s a bit cumbersome.

it runs also much faster without the dependencies.

lexming (approved):
LGTM

Micket:
Everyone is on board. Merging

PR #4707, opened 2024-11-14
enhance `EasyBlock` class to allow passing in `logfile`

Thyre:
In the case an EasyBlock creates another instance of an EasyBlock (for example when dealing with Bundles), each instance will create new log file, which may stay open and cause issues down the line. To prevent this, add an option to the EasyBlock constructor to disable logging to files.

Marked as draft until verified with QuantumESPRESSO and Bundle changes in respective PRs. I want to really make sure that all issues are resolved first.

Related: https://github.com/easybuilders/easybuild-easyconfigs/issues/21841, https://github.com/easybuilders/easybuild-easyblocks/pull/3505

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -40,6 +40,7 @@ Authors:
 * Maxime Boissonneault (Compute Canada)
 * Davide Vanzo (Vanderbilt University)
 * Caspar van Leeuwen (SURF)
+* Jan Andre Reuter (Juelich Supercomputing Centre)
 """
 
 import copy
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -213,9 +215,13 @@ def __init__(self, ec):
         self.rpath_include_dirs = []
 
         # logging
+        # each easyblock instance creates its own logger.
+        # however, this is not necessary if a logger is passed to the easyblock
+        # in that case, reuse the logger
```
Thyre, line 220: Comment is outdated from my previous draft...

Thyre:
@boegel If you find any time for a short review, can you look over this PR?
I think the changes in the EasyBlocks are straight-forward, but the changes here (especially the tests) might require one or two revisions

boegel (approved):
lgtm

boegel:
> @boegel If you find any time for a short review, can you look over this PR? I think the changes in the EasyBlocks are straight-forward, but the changes here (especially the tests) might require one or two revisions

Looks great to me, I didn't see anything that would need to be changed...

PR #4710, opened 2024-11-22
avoid checking loaded module twice in findUpdatedEcs.sh

Flamefire:
The environment variables for the GCC and GCCcore ($EBROOT*) both point to the GCCcore directory which leads to checking that twice. This is true for all ECs using `altroot`.

To avoid outputting the same change multiple times make the folders to check unique first.

Also error if `--short` and `--diff` are both given at the same time.


Micket (approved):
lgtm

PR #4716, opened 2024-12-05
add support for using list of values in `modulename` in extensions

Flamefire:
This changes the allowed types of the `modulename` extension option:
- `False` to skip the sanity check and always install it when `--skip` is used
- `str`: Value for `%(ext_name)s` in the `exts_filter` template
- List of `str`: Multiple names to be used in the `exts_filter` template. All resulting commands must succeed.

This was [previously suggested](https://github.com/Flamefire/easybuild-easyblocks/pull/6#issuecomment-821375228) to allow Perl modules that are part of other Perl modules and hence don't need to be installed multiple times. See https://github.com/easybuilders/easybuild-easyconfigs/pull/12575

It replaces `resolve_exts_filter_template` by
`construct_exts_filter_cmds` as the method now returns a, potentially empty, list which might cause errors if used without expecting a list.

The function is only used internally in framework and not by any easyblocks.

### DIFF
--- a/easybuild/framework/extension.py
+++ b/easybuild/framework/extension.py
@@ -41,18 +41,18 @@ import os
 from easybuild.framework.easyconfig.default import get_easyconfig_parameter_default
 from easybuild.framework.easyconfig.easyconfig import resolve_template
 from easybuild.framework.easyconfig.templates import TEMPLATE_NAMES_EASYBLOCK_RUN_STEP, template_constant_dict
-from easybuild.tools.build_log import EasyBuildError, EasyBuildExit, raise_nosupport
+from easybuild.tools.build_log import EasyBuildError, EasyBuildExit
 from easybuild.tools.filetools import change_dir
 from easybuild.tools.run import run_shell_cmd
 from easybuild.tools.utilities import trace_msg
 
 
-def resolve_exts_filter_template(exts_filter, ext):
+def construct_exts_filter_cmds(exts_filter, ext):
     """
     Resolve the exts_filter tuple by replacing the template values using the extension
     :param exts_filter: Tuple of (command, input) using template values (ext_name, ext_version, src)
     :param ext: Instance of Extension or dictionary like with 'name' and optionally 'options', 'version', 'source' keys
-    :return: (cmd, input) as a tuple of strings
+    :return: (cmd, input) as a tuple of strings for each modulename. Might be empty if no filtering is intented
     """
 
     if isinstance(exts_filter, str) or len(exts_filter) != 2:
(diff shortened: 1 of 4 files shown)


boegel:

easybuild/framework/extension.py
```diff
@@ -41,18 +41,44 @@
...
+
+    try:
+        modulenames = ext['options']['modulename']
+    except KeyError:
+        modulenames = [ext['name']]
+    else:
+        if isinstance(modulenames, list):
+            if not modulenames:
+                raise EasyBuildError(f"Empty modulename list for {ext['name']} is not supported."
+                                     "Use `False` to skip checking the module!")
```
smoors:
```suggestion
                                     "Use `False` to skip the module import check!")
```

easybuild/framework/extension.py
```diff
@@ -41,18 +41,44 @@
...
+        elif modulenames is False:
+            return [ext['name']] if use_name_for_false else []
+        elif not isinstance(modulenames, str):
+            raise EasyBuildError(f"Wrong type of modulename for {ext['name']}: {type(modulenames)}: {modulenames}")
```
smoors:
```suggestion
            raise EasyBuildError(f"Invalid type for modulename {ext['name']}. Expected bool, str, or list, but got {type(modulenames).__name__}: {modulenames}")
```

Flamefire: "Invalid type for modulename scipy." sounds misleading. Also "bool" True is not allowed, only False.

Maybe:
```suggestion
            raise EasyBuildError(f"Invalid type for modulename of {ext['name']}. Expected False, str, or list, but got {type(modulenames).__name__}: {modulenames}")
```

easybuild/framework/extension.py
Flamefire: I wanted to stay generic: For Perl or R packages there is no import, examples:
```
"gem list '^%(ext_name)s$' -i"
"R -q --no-save", "library(%(ext_name)s)"
"eval `opam config env` && opam list --installed %(ext_name)s.%(ext_version)s"
"julia -e 'using %(ext_name)s'"
```

But I see where you are coming from. Maybe:
```suggestion
                                     "Use `False` to skip checking for module existence!")
```

Flamefire:
@smoors I did the changes we discussed (didn't get a notification for the thumbs up so I missed them until yesterdays confcall)
Force-push to amend the commit fixing a line length only

easybuild/framework/extension.py
```diff
@@ -314,35 +337,34 @@ def sanity_check_step(self):
...
-            modname = self.options['modulename']
-            self.log.debug("modulename found in self.options, using it: %s", modname)
-        else:
-            modname = self.name
-            self.log.debug("self.name: %s", modname)
-
+        exts_filer_cmds = construct_exts_filter_cmds(exts_filter, self)
         # allow skipping of sanity check by setting module name to False
-        if modname is False:
+        if exts_filer_cmds is None:
```
smoors: `construct_exts_filter_cmds` never returns `None`?

Flamefire: Oh, yes. It now returns an empty list
```suggestion
        if not exts_filer_cmds:
```

smoors (approved):
tested on a customized `Arrow-22.0.0-gfbf-2025a.eb`, works like a charm

```python
# multiple modulenames
'modulename': ['pyarrow', 'pyarrow.dataset', 'pyarrow.parquet', 'pyarrow.orc'],
# no modulename
'modulename': False,
```

PR #4719, opened 2024-12-11
Fix FFT entry in --list-toolchain output for Cray toolchains

Flamefire:
The entry was left uninitialized containing a "random" previous value.


boegel (approved):
lgtm

PR #4720, opened 2024-12-11
use `enumerate` where applicable + fix for `ModuleGenerator._generate_multi_deps_list`

Flamefire:
PyLint found uses of `range(len(...))` where the index was then used to access an element.

Using `enumerate` yields the value together with the index making the intention clearer (and the code likely faster).

This also led to a nice simplification possibility in make_module_dep, get_parsed_multi_deps and configobj.walk and a possible bugfix in _generate_multi_deps_list:

The code generating the load statement uses the first module of each name in `self.app.cfg.multi_deps` but the code generating the help text has an additional version check. We don't need that and it can possibly be wrong if the order of the versions was different somehow.   
I simply collect all modules with the given name using a (fast) list comprehension and mark the first as the default

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1381,8 +1381,7 @@ class EasyBlock(object):
             multi_dep_mod_names = {}
             for deplist in self.cfg.multi_deps:
                 for dep in deplist:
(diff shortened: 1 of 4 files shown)


boegel (approved):
lgtm

PR #4723, opened 2024-12-18
run sanity checks commands from an empty temporary directory (rather than the software install directory)

### LINKED ISSUE #4044: `--sanity-check-only` runs commands in installation dir
This leads to problems if the commands have output files since these will get placed in the installation

lexming:
Fixes #4044

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3739,16 +3739,16 @@ class EasyBlock(object):
         if self.toolchain.mpi_family() and self.toolchain.mpi_family() in toolchain.OPENMPI:
             env.setvar('OMPI_MCA_rmaps_base_oversubscribe', '1')
 
-        # change to install directory (better environment for running tests)
-        if os.path.isdir(self.installdir):
-            change_dir(self.installdir)
+        # run sanity checks from an empty temp directory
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -3739,16 +3739,16 @@ def xs2str(xs):
         if self.toolchain.mpi_family() and self.toolchain.mpi_family() in toolchain.OPENMPI:
             env.setvar('OMPI_MCA_rmaps_base_oversubscribe', '1')
 
-        # change to install directory (better environment for running tests)
-        if os.path.isdir(self.installdir):
-            change_dir(self.installdir)
+        # run sanity checks from an empty temp directory
+        # using the build or installation directory can produce false positives and polute them with files
+        sanity_work_dir = tempfile.mkdtemp(prefix='eb-saniy-check-')
```
boegel: @lexming Are you testing whether or not I'll catch the typo here? ;)

lexming: ohh good one :smile:

boegel: fixed in 2389fc6

boegel (approved):
lgtm

PR #4725, opened 2024-12-19
fall back to getting values with unresolved templates in `EasyBlock.check_checksums_for`

boegel:
Fix for problems that arise with for example `QuantumESPRESSO-7.3-intel-2023a.eb` in the easyconfigs test suite:
```
Failed to resolve all templates in "mkdir -p %(builddir)s/qe-%(version)s && tar xzvf %%s --strip-components=1 -C $_" using template dictionary: {...}
```

That's because the `%(builddir)s` template can not be resolved yet when checking checksums for sources/patches, but since that's part of `extract_cmd` there's no reason for concern here:

```python
sources = [
    {
        'filename': 'q-e-qe-%(version)s.tar.gz',
        'extract_cmd': 'mkdir -p %(builddir)s/qe-%(version)s && tar xzvf %s --strip-components=1 -C $_',
        'source_urls': ['https://gitlab.com/QEF/q-e/-/archive/qe-%(version)s']
    },
...
```

Falling back to grabbing values without resolving templates via `get_ref` fixes this issue, only small downside is that it leads to using `q-e-qe-%(version)s.tar.gz` in any potential error messages in case of incorrect checksum.

edit: fallout from changes in https://github.com/easybuilders/easybuild-framework/pull/4516

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2584,23 +2584,17 @@ class EasyBlock(object):
         ec_fn = os.path.basename(self.cfg.path)
         checksum_issues = []
 
-        if isinstance(ent, EasyConfig):
-            # try to get value with templates resolved, but fall back to not resolving templates;
-            # this is better for error reporting that includes names of source files
-            try:
-                sources = ent.get('sources', [])
-                patches = ent.get('patches', []) + ent.get('postinstallpatches', [])
-                checksums = ent.get('checksums', [])
-            except EasyBuildError:
-                sources = ent.get_ref('sources')
-                patches = ent.get_ref('patches') + ent.get_ref('postinstallpatches')
-                checksums = ent.get_ref('checksums')
-        elif isinstance(ent, dict):
+        # try to get value with templates resolved, but fall back to not resolving templates;
+        # this is better for error reporting that includes names of source files
+        try:
             sources = ent.get('sources', [])
             patches = ent.get('patches', []) + ent.get('postinstallpatches', [])
             checksums = ent.get('checksums', [])
-        else:
-            raise EasyBuildError(f"Unexpected value type in EasyBlock.check_checksums_for: {type(ent)}")
+        except EasyBuildError:
+            if isinstance(ent, EasyConfig):
+                sources = ent.get_ref('sources')
+                patches = ent.get_ref('patches') + ent.get_ref('postinstallpatches')
+                checksums = ent.get_ref('checksums')
 
         # Single source should be re-wrapped as a list, and checksums with it
         if isinstance(sources, dict):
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2584,9 +2584,24 @@ def check_checksums_for(self, ent, sub='', source_cnt=None):
...
+                sources = ent.get_ref('sources')
+                patches = ent.get_ref('patches') + ent.get_ref('postinstallpatches')
+                checksums = ent.get_ref('checksums')
+        elif isinstance(ent, dict):
+            sources = ent.get('sources', [])
+            patches = ent.get('patches', []) + ent.get('postinstallpatches', [])
+            checksums = ent.get('checksums', [])
+        else:
+            raise EasyBuildError(f"Unexpected value type in EasyBlock.check_checksums_for: {type(ent)}")
+
```
lexming, line 2604: Can we simplify to have a consistent behaviour across the board? Is there any downside to always fallback to not resolved templates?
```suggestion
        # try to get value with templates resolved, but fall back to not resolving templates;
        # this is better for error reporting that includes names of source files
        try:
            sources = ent.get('sources', [])
            patches = ent.get('patches', []) + ent.get('postinstallpatches', [])
            checksums = ent.get('checksums', [])
        except EasyBuildError:
            sources = ent.get_ref('sources')
            patches = ent.get_ref('patches') + ent.get_ref('postinstallpatches')
            checksums = ent.get_ref('checksums')

```

lexming (commented):
Can we just always do

easybuild/framework/easyblock.py
boegel: There's a big downside here: `ent` may be a `dict` value rather than an `EasyConfig` instance, and a `dict` value doesn't support `.get_ref`, so that would lead to a hard crash...

That's part of the reason why I enhanced the test for `check_checksums_for`

lexming: Ok, then I suggest to rewrite the same logic as follows to simplify this code:
```suggestion
        # try to get value with templates resolved, but fall back to not resolving templates;
        # this is better for error reporting that includes names of source files
        try:
            sources = ent.get('sources', [])
            patches = ent.get('patches', []) + ent.get('postinstallpatches', [])
            checksums = ent.get('checksums', [])
        except EasyBuildError as err:
            if isinstance(ent, EasyConfig):
                sources = ent.get_ref('sources')
                patches = ent.get_ref('patches') + ent.get_ref('postinstallpatches')
                checksums = ent.get_ref('checksums')

```

boegel: fixed in 85987c298

lexming (approved):
LGTM

PR #4728, opened 2024-12-22
do not attach PIPE file handle on STDIN of `run_shell_cmd` unless there are contents for it

lexming:
Fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/22094

We always attach the PIPE file handle to STDIN regardless if there is any input to be passed to the command. This seems to be ok on Python 3.9+, but it is troublesome on Python 3.6 where the resulting shell has as an attached handle without EOF even if nothing is passed to STDIN. So software like perl think that they are inside an interactive shell.


boegel (approved):
lgtm

PR #4726, opened 2024-12-19
add support for temporarily disabling requirement that all template values must resolve via `EasyConfig.expect_resolved_template_values`

boegel:
required for https://github.com/easybuilders/easybuild-easyblocks/pull/3537

PR #4735, opened 2025-01-03
Add context manager for allowing unresolved templates and make the state members private + remove support for directly setting `enable_templating` and `expect_resolved_template_values`

Flamefire:
Both (templating and disabling the enforcing) should only be disabled temporarily so make the state private and expose only the context manager and a property.

2 small fixes in adjacent code:
- Use a try-except instead of an if-else (avoids an extra lookup)
- Use the `resolve_template` member instead of the free function to remove some C&P code

I also intended to make `asdict` trivial (see https://github.com/easybuilders/easybuild-framework/commit/2a6386396c7e969d4a0c1e23ab5f880a72dedf23) but as `__getitem__` checks for deprecated parameters on each access it might impact performance too much.

Improvement after https://github.com/easybuilders/easybuild-framework/pull/4726


boegel (approved):
lgtm, but I would like to see this tested thoroughly via updated `Bundle` easyblock in https://github.com/easybuilders/easybuild-easyblocks/pull/3547 before merging it...

Flamefire:
I can do that. Any specific ECs in mind or just 3-4 ones using `easblock = 'Bundle'`?

boegel (changes requested):

boegel (approved):
lgtm

PR #4736, opened 2025-01-03
Avoid making build directory read-only

Flamefire:
When using e.g. `--stop=patch` the log directory will be inside the build directory. When `--read-only-installdir` is set the log directory and the build directory itself will be read-only which is not what the option intended: The `ensure_writable_log_dir` function changes the permissions of the folder containing the log directory to writeable without checking if this is required. When supposedly reverting it any write permissions will be removed even if they have been there before.

Similar the write permissions of the log directory are removed even when the installation was stopped.

This adds the neccessary checks.


boegel (approved):
lgtm

PR #4737, opened 2025-01-05
faster `nub` function

branfosj:
See https://www.peterbe.com/plog/fastest-way-to-uniquify-a-list-in-python-3.6


bartoldeman (approved):
lgtm

PR #4738, opened 2025-01-06
show readable error message when applying patch without (extracted) source

Flamefire:
When there is no source `self.src` is an empty list which leads to a rather generic error message. Improve that by showing that there was no source to apply the patch to.

For extensions `self.src` is set to a string as only a single source is supported.
Accessing `self.src[0].['finalpath']` leads to an error
> TypeError: string indices must be integers
This happens when the source didn't got extracted so `self.ext_dir` and hence `beginpath` will be `None`.
Make the error show that the source was not extracted.

While doing that get rid of `patch['source']` which can never be set as it would trigger a failure in `create_patch_info` as it isn't specified as an allowed key. It isn't useful anyway as only the first source will have its `finalpath` set correctly, the other, when unpacked to the same folder, are just referring to that folder instead of their subfolder.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2607,8 +2607,6 @@ class EasyBlock(object):
             self.log.info("Applying patch %s" % patch['name'])
             trace_msg("applying patch %s" % patch['name'])
 
-            # patch source at specified index (first source if not specified)
-            srcind = patch.get('source', 0)
             # if patch level is specified, use that (otherwise let apply_patch derive patch level)
             level = patch.get('level', None)
             # determine suffix of source path to apply patch in (if any)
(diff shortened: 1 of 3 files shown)


boegel:
@Flamefire Any idea why that one test keeps failing here?

Flamefire:
I think this is https://github.com/envmodules/modules/issues/82 caused by running the extension step multiple times without resetting the environment. I hope the new commit fixes this.

Flamefire:
Yes that was the issue :)

Flamefire:
Rebased

boegel:
@Flamefire Test needs love?
```
UnboundLocalError: local variable 'unpack_src' referenced before assignment
```

boegel (approved):
lgtm

PR #4739, opened 2025-01-08
Adapt `module show` command run to cope with non-zero exit code for non-existing module (required for Environment Modules v5.5+ and Lmod 8.7.56+)

xdelaruelle:
The `module show` command is used by ModulesTool `exist` method to test the existence of a modulefile. Running `module show` against a non-existing modulefile will make `modulecmd.tcl` script to produce a non-zero exit code starting Modules v5.5.

With this change, the exit code check when running `module show` is disabled. It should not harm the module existence check as it relies on the analysis of the obtained stderr output.
    
Note that Lmod (as of version 8.7.55) does not set a non-zero exit code when running `module show` against a non-existing modulefile when `LMOD_QUIET` environment variable is set. If this behavior is changed in the future, this change will also benefit to Lmod.

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -709,7 +709,8 @@ class ModulesTool(object):
             ans = MODULE_SHOW_CACHE[key]
             self.log.debug("Found cached result for 'module show %s' with key '%s': %s", mod_name, key, ans)
         else:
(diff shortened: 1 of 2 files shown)


boegel:
Looks OK, but I want to verify a bit better than we're not relying on the exit code in other places, where the `show` function is being relied on...

boegel (approved):
lgtm

PR #4753, opened 2025-02-04
Ignore unmodified files when using `--new-pr` or `--update_pr`

### LINKED ISSUE #4751: `--new-pr` with existing file leads to error even if not modified
I tried to do a `eb --new-pr *` to add new files but it failed with

> ERROR: A meaningful commit message must be specified via --pr-commit-msg when modifying/deleting files or targeting the framework repo.
> Modified: lxml-5.3.0-GCCcore-13.3.0.eb

Something must be wrong in the detection

Flamefire:
When including an unmodified EasyConfig with `--new-pr` an error is shown that a commit message is required because this EC is modified which is not the case.

Adjust the `copy_*` functions to ignore any unmodified file. This especially omits them in the `file_info` struct returned that is then used to determine commit message/title etc.

Fixes #4751

Note: While working on this I noticed that new EasyConfigs that are dependencies of added EasyConfigs are ignored in the commit message. I propose to populate `file_info` instead of the `dep_info` at https://github.com/easybuilders/easybuild-framework/blob/b37f70756960c1686875974f8381491df280cc00/easybuild/tools/github.py#L1071

This would also fix the possibly incomplete debug log message `Staging all <n> new/modified easyconfigs` that doesn't account for those ECs.

**Note**: View without whitespace changes due to changed indent.

I noticed that when you pass only unmodified Easyconfigs it will crash when trying to determine a label for the branch. This can only be caught in a test actually opening a new PR (in dry-run) so I extended the new_pr/update_pr test.
This increased the time of this (single) test from 58s to 72s.
As most of the time is spent in `git fetch` I did some mocking to bring the time down to 38s


Flamefire:
Rebased

Flamefire:
Rebased only

Flamefire:
I noticed that when you pass only unmodified Easyconfigs it will crash when trying to determine a label for the branch. This can only be caught in a test actually opening a new PR (in dry-run) so I extended the new_pr/update_pr test.
This increased the time of this (single) test from 58s to 72s.
As most of the time is spent in `git fetch` I did some mocking to bring the time down to 38s

Flamefire:
Ok, that seems to work now. I'd say we should get this in soon to be able to catch any remaining issues/corner cases I didn't think of. @boegel ?

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -2673,15 +2674,19 @@ def det_file_info(paths, target_dir):
...
-            ec_filename = file_info['ecs'][-1].filename()
+            ec = ecs[0]['ec']
+            soft_name = ec.name
+            ec_filename = ec.filename()
 
             target_path = det_location_for(path, target_dir, soft_name, ec_filename)
 
             new_file = not os.path.exists(target_path)
+            if not new_file and filecmp.cmp(path, target_path):
+                continue  # Ignore unchanged files
```
boegel: this changes the semantics of `det_file_info`, so this should be made optional?

```suggestion
            if ignore_unchanged_files and not (new_file and filecmp.cmp(path, target_path)):
                continue  # Ignore unchanged files
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -2725,6 +2730,8 @@ def copy_patch_files(patch_specs, target_dir):
     }
     for patch_path, soft_name in patch_specs:
         target_path = det_location_for(patch_path, target_dir, soft_name, os.path.basename(patch_path))
+        if os.path.exists(target_path) and filecmp.cmp(patch_path, target_path):
+            continue  # Skip copy and entry if not modified
```
boegel, line 2735: Let's add a log message when we're skipping the copy?

```suggestion
            _log.debug(f"Skipping copy for file {patch_path}, identical file already exists at {target_path}")
            continue  # Skip copy and entry if not modified

boegel (changes requested):
partial review, will try to revisit later, but changed default semantics of `det_file_info` should be fixed imho

Flamefire:
@boegel rebased again and tested locally with a custom github token. The CI tests are skipped in PRs for this.
I'm reasonably sure it works now

PR #4754, opened 2025-02-04
Small improvements to `ModuleLoadEnvironment`

Flamefire:
I just checked https://github.com/easybuilders/easybuild-framework/pull/4653 and found a few possible improvements:

1. Using an "unknown" state and requiring callers to update that is error-prone. Comparing `self.install_lib_symlink == LibSymlink.LIB64_TO_LIB` might silently fail if it is forgotten. Use a property with auto-init instead
2. Isn't `def __init__(... var_type=ModEnvVarType.PATH_WITH_FILES...)` clearer in what the default is instead of using and checking for `None`?
3. `for attr in self.__dict__: yield attr, getattr(self, attr)` can be made simpler and faster by using `yield from self.__dict__.items()` or just returning the latter
4. `environ` can use a dict-comprehension for brevity and speed to avoid repeated `update` calls

Do those make sense? @lexming @boegel

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -134,7 +134,7 @@ class LibSymlink(Enum):
     - LIB64_TO_LIB: 'lib64' is a symlink to 'lib'
     - NEITHER: neither 'lib' is a symlink to 'lib64', nor 'lib64' is a symlink to 'lib'
     - """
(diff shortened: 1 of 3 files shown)


lexming:
@Flamefire yeah, those make perfect sense, thanks for the PR!

PR #4757, opened 2025-02-07
remove duplication of logs of module commands 

lexming:
`run_shell_cmd` already prints at info level the stdout and stderr of those commands


boegel (approved):
lgtm

PR #4758, opened 2025-02-07
enhance `apply_regex_substitutions`

Flamefire:
It might be required to replace patterns with more context, e.g. content of the next or previous line to disambiguate otherwise too generic matches.

Add parameter `single_line` to enable the old behavior (default) of matching per line and otherwise match the whole text.

Add parameter `match_all` to require all patterns to match for each file not only at least one.

Also allow to pass regex `Pattern` object instead of string patterns to allow for additional flags per pattern.   
Additionally I collect all error messages for all files and report them together instead of per file for easier resolving.


boegel (approved):
lgtm

PR #4761, opened 2025-02-10
add BOTH_TO_DIR state to LibSymlink

lexming:
This covers cases where both `lib` and `lib64` are symlinks to some other directory


boegel (approved):
Tested with https://github.com/easybuilders/easybuild-easyblocks/pull/3555, this fixes the problem I was observing with old `imkl` versions (as also verified by the additional tests)

PR #4763, opened 2025-02-13
correctly deal with easyblocks that still use deprecated `make_module_req_guess` method: remove environment variables if they're not present in guesses

boegel:
Some easyblocks (like Conda, Anaconda, XALT) create an entirely new dict in `make_module_req_guess`, which we should respect. If that dict doesn't include environment variables like `PATH` and `LD_LIBRARY_PATH`, it's for a good reason.

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -351,6 +351,13 @@ class ModuleLoadEnvironment:
         except AttributeError as err:
             raise EasyBuildError("Cannot update ModuleLoadEnvironment from a non-dict variable") from err
 
+    def replace(self, new_env):
+        """Replace contents of environment with given dictionary"""
+        env_vars = [e for e in dir(self) if e.isupper()]
+        for env_var in env_vars:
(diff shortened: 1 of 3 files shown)


easybuild/tools/modules.py
```diff
@@ -351,6 +351,13 @@ def update(self, new_env):
         except AttributeError as err:
             raise EasyBuildError("Cannot update ModuleLoadEnvironment from a non-dict variable") from err
 
+    def replace(self, new_env):
+        """Replace contents of environment with given dictionary"""
+        env_vars = [e for e in dir(self) if e.isupper()]
```
lexming, line 356: you can just use `self.vars` here

lexming (changes requested):

easybuild/tools/modules.py
boegel: fixed in 55f62146d

lexming (approved):

PR #4764, opened 2025-02-17
don't enable keeping of debug symbols by default

Micket:
Switching the default because this was too painful in many cases, like LLVM in particular like to explode with `-g` present, even for just staged builds.

edit: see also https://github.com/easybuilders/easybuild-easyconfigs/issues/22247


boegel (approved):
lgtm

PR #4769, opened 2025-02-21
Ignore other classes if software specific easyblock class was found 

Flamefire:
If we have other classes next to a software specific easyblock class in an easyblock file they are likely supplemental.

So ignore them in `avail_easyblocks`.

Also 2 cleanups:
- Use the constant for `EB_` in all places
- Make the loop a bit simpler by reducing the 8-level nesting

### DIFF
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -758,7 +758,7 @@ def avail_easyblocks():
     """Return a list of all available easyblocks."""
 
     module_regexp = re.compile(r"^([^_].*)\.py$")
-    class_regex = re.compile(r"^class ([^(]*)\(", re.M)
+    class_regex = re.compile(r"^class ([^(:]*)\(", re.M)
 
     # finish initialisation of the toolchain module (ie set the TC_CONSTANT constants)
     search_toolchain('')
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -768,39 +768,37 @@ def avail_easyblocks():
         __import__(pkg)
 
         # determine paths for this package
-        paths = sys.modules[pkg].__path__
+        paths = [path for path in sys.modules[pkg].__path__ if os.path.exists(path)]
 
         # import all modules in these paths
         for path in paths:
-            if os.path.exists(path):
-                for fn in os.listdir(path):
-                    res = module_regexp.match(fn)
-                    if res:
-                        easyblock_mod_name = '%s.%s' % (pkg, res.group(1))
+            for fn in os.listdir(path):
+                res = module_regexp.match(fn)
+                if not res:
+                    continue
+                easyblock_mod_name = '%s.%s' % (pkg, res.group(1))
 
-                        if easyblock_mod_name not in easyblocks:
-                            __import__(easyblock_mod_name)
-                            easyblock_loc = os.path.join(path, fn)
+                if easyblock_mod_name in easyblocks:
+                    _log.debug("%s already imported from %s, ignoring %s",
+                               easyblock_mod_name, easyblocks[easyblock_mod_name]['loc'], path)
+                else:
+                    __import__(easyblock_mod_name)
+                    easyblock_loc = os.path.join(path, fn)
 
-                            class_names = class_regex.findall(read_file(easyblock_loc))
-                            if len(class_names) > 1:
-                                # If there is exactly one software specific easyblock we use that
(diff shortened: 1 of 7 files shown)


easybuild/framework/easyconfig/tools.py
```diff
@@ -768,33 +768,37 @@ def avail_easyblocks():
...
+                    __import__(easyblock_mod_name)
+                    easyblock_loc = os.path.join(path, fn)
+
+                    class_names = class_regex.findall(read_file(easyblock_loc))
+                    if len(class_names) > 1:
+                        # If there is exactly one software specific easyblock we use that
+                        sw_specific_class_names = [name for name in class_names
+                                                   if not is_generic_easyblock(name)]
+                        if len(sw_specific_class_names) == 1:
+                            class_names = sw_specific_class_names
```
akesandgren, line 794: This could possibly be improved to allow additional classes in a generic easyblock by checking if pkg is easybuild.easyblocks.generic and comparing the class_names with the filename

Flamefire: Good point. Implemented. By testing I also found that our (for some reason multi-line regex) would match:

```
class Foo:
    pass

class Bar(EasyBlock):
   pass
```

Resulting in the "class_names" `Foo:\n    pass\n\n\nclass Bar` because it finishes only at parentheses not the colon

Flamefire:
Through the latest error I've found a duplication of the logic: `get_easyblock_classes` and `get_easyblock_classes` do mostly the same while the former could use the latter.

`get_easyblock_classes` imports all matching modules and gets all classes that derive from `EasyBlock` (after my fix which might change behavior if one does not. Previously it returned any class defined in the file but I don't think this is an issue. I added a check if any was found, the case of multiple being found might also need handling)

`avail_easyblocks` also imports the module and then checks the paths defined by the module, then goes over each file and uses a regexp on the text content.

I'd say we should only use the logic from `get_easyblock_classes` and add the check for no or multiple found (potential) easyblocks there.    
Not fully sure how duplication is handled when `--include-easyblocks*` is in effect though.

test/framework/sandbox/easybuild/easyblocks/generic/bar.py
```diff
@@ -32,6 +32,18 @@
...
 
+class dummy1:
+    """Only to verify that unrelated classes in software specific easyblocks are ignored"""
+
+
+class dummy2(dummy1):
+    """Same but with inheritance"""
+
+
+class dummy1:
```
akesandgren, line 43: Should probably be dummy3 to not confuse anyone who reads this code.

akesandgren (changes requested):

test/framework/sandbox/easybuild/easyblocks/generic/bar.py
Flamefire:
```suggestion
class dummy3:
```

Flamefire:
> I assume you meant `get_easyblock_classes` and `avail_easyblocks` here?

Indeed. It's seemingly C&P-mistake-monday ;-) Fixed so the description makes sense

easybuild/tools/include.py
```diff
@@ -157,7 +157,7 @@ def verify_imports(pymods, pypkg, from_path):
 
 def is_software_specific_easyblock(module):
     """Determine whether Python module at specified location is a software-specific easyblock."""
-    return bool(re.search(r'^class EB_.*\(.*\):\s*$', read_file(module), re.M))
+    return bool(re.search(r'^class %s.*\(.*\):\s*$' % EASYBLOCK_CLASS_PREFIX, read_file(module), re.M))
```
akesandgren, line 160: This regex might be too loosely specified. Since this is multiline it should be
```
r"^class %s([^(:]*)\(([^)]**)\):\s*$" % EASYBLOCK_CLASS_PREFIX
```
Or at least I think so...

akesandgren: But perhaps keep that for a followup to not make this contain too many changes

Flamefire: We touched that line anyway so might as well include it. Your regex seems to be fine, just removed the parens as we don't need any capture groups. I also added a comment to describe it

boegel (approved):
lgtm

Flamefire:
> > Through the latest error I've found a duplication of the logic: `avail_easyblocks` and `get_easyblock_classes` do mostly the same while the former could use the latter.
> 
> 
> If so let's push that to a separate PR

@boegel Shall I do that PR?

PR #4774, opened 2025-02-27
add support for `ModuleLoadEnvironment` aliases in `modextrapaths` easyconfig parameter + deprecate `allow_append_abs_path`, `allow_prepend_abs_path`, `modextrapaths_append` easyconfig parameters

### LINKED ISSUE #3331: Problem using CPATH in modulefiles overwriting system paths

lexming:
## Add support for `--module-search-path-headers` to `modextrapaths`

This is the final change to fully fix #3331 . Now that we have `--module-search-path-headers` we need to allow `modextrapaths` to follow that setting. This PR achieves that by adding `MODULE_LOAD_ENV_HEADERS` as a new constant in easyconfigs. Hence, the same `MODULE_LOAD_ENV_HEADERS` global can be used in EasyBlocks and EasyConfigs.
```
modextrapaths = {
    MODULE_LOAD_ENV_HEADERS: 'include/extra_dir',
}
```

## Unify interface to add extra paths in easyconfigs

Currently, there are differing methods to handle options for extra paths. For instance:
* `modextrapaths` and `modextrapaths_append` allow to change placement of paths in the module environment
* the delimiter of those extra paths can be changed through a dictionary item passed to either of those options. e.g. `modextrapath = { 'ENV_VAR' : {'paths': "path", 'delimiter': ':'}}`
* `allow_append_abs_path` and `allow_prepend_abs_path` enable absolute paths in `modextrapaths`

This PR standardizes how extra paths can be injected in easyconfigs by making `modextrapaths` the only usable parameter. All existing options have been added to `ModuleEnvironmentVariable` (like `delimiter` and `prepend`) and all class settings can now be set through a dictionary in `modextrapaths`. Example:
```
modextrapaths = {
    'ENV_VAR': {'paths': ['foo', 'baz'], 'delimiter': ';', 'prepend': False},
}
```
Hence, `modextrapaths` becomes the interface to `ModuleLoadEnvironment` at the easyconfig level. `modextrapaths_append` is deprecated, same result can be achieved with `{'prepend': False}` in `modextrapaths`.

## Enable absolute paths in module environment

Absolute paths are allowed by default in all path variables of `ModuleLoadEnvironment`. Deprecating `allow_append_abs_path` and `allow_prepend_abs_path`.

Providing a glob pattern that is an absolute path is enough to enable absolute paths expansion and injection in module files.

Method `expand_module_search_path` is enhanced to handle absolute globs:
* absolute globs cannot turn into relative paths; and relative globs cannot turn into absolute ones.
* safety checks have added to avoid relative paths that jump outside of the install directory after expansion. Which is the only possible scenario that is dangerous.
* symlinks will be followed and resolved, ensuring that `expand_module_search_path` always returns paths avoiding symlinks. This removes the need to check for the topology of symlinks between `lib` and `lib64` for instance. However, generated modules look different in some cases, with paths to the same folder jumping through different parents (e.g. `imkl` with libraries in customs dirs and `lib` and `lib64` symlinks). 
* remove `LibSymlink` class and related methods from framework

## Enhance `ModuleLoadEnvironment` to handle environment variables starting with underscores

`ModuleLoadEnvironment` enhanced to allow handling environment variables that start with one or more underscores. This is necessary to process easyconfigs like [libglvnd-1.7.0-GCCcore-13.3.0.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/l/libglvnd/libglvnd-1.7.0-GCCcore-13.3.0.eb).

All public environment variables are now stored inside a custom dict in `ModuleLoadEnvironment._env_vars`. Getter and setter methods have been modified to be able to discriminate between private attributes of the class and public environment variables.


lexming:
Test report for `libglvnd` in https://github.com/easybuilders/easybuild-easyconfigs/pull/22416#issuecomment-2689070385

```
prepend_path("CMAKE_LIBRARY_PATH", pathJoin(root, "lib"))
prepend_path("CMAKE_PREFIX_PATH", root)
prepend_path("CPATH", pathJoin(root, "include"))
prepend_path("LD_LIBRARY_PATH", pathJoin(root, "lib"))
prepend_path("LIBRARY_PATH", pathJoin(root, "lib"))
prepend_path("PKG_CONFIG_PATH", pathJoin(root, "lib/pkgconfig"))
prepend_path("__EGL_VENDOR_LIBRARY_DIRS", "/usr/share/glvnd/egl_vendor.d")
setenv("EBROOTLIBGLVND", root)
setenv("EBVERSIONLIBGLVND", "1.7.0")
setenv("EBDEVELLIBGLVND", pathJoin(root, "easybuild/libglvnd-1.7.0-GCCcore-13.3.0-easybuild-devel"))
```

lexming:
Test report for `MODULE_LOAD_ENV_HEADERS` in `modextrapaths` for `libtirc`:
* cpath: https://github.com/easybuilders/easybuild-easyconfigs/pull/22417#issuecomment-2689198239
* include_paths:

boegel (approved):
lgtm

PR #4781, opened 2025-03-03
Add support for `--disable-robot`

### LINKED ISSUE #4780: Allow disabling `--robot`
When using `--robot` with the GitHub integration `--new-pr ---robot` it will upload ALL dependencies.

If `robot = <PATH>` is specified in the config file there is no way to turn it off:

```
$ eb --disable-robot
Usage: eb [options] easyconfig [...]

eb: error: no such option: --disable-robot
$ eb --robot ''
ERROR: Failed to parse configuration options: 'Argument passed to --robot is not an existing directory: '
```

We need some way to disable it and/or disable it for PR options unless specified on cmdline if that is possible as that behavior is likely unexpected.

Flamefire:
It might be necessary to disable dependency resolution from the commandline.
E.g. when it is enabled in the configfile and we want to upload a single
file via `--new-pr`.
Add new cmdline action `store_or_False` to generically add support for
`--disable-<name>` for such cases.

Closes #4780


easybuild/base/generaloption.py
```diff
@@ -232,7 +235,9 @@ def store_or(option, opt_str, value, parser, *args, **kwargs):  # pylint: disabl
                 """Callback for supporting options with optional values."""
                 # see http://stackoverflow.com/questions/1229146/parsing-empty-options-in-python
                 # ugly code, optparse is crap
-                if parser.rargs and not parser.rargs[0].startswith('-'):
+                if option.store_or == self.STORE_OR_FALSE and opt_str.startswith("--%s-" % self.DISABLE):
```
boegel, line 238: 2nd part of this condition feels very hard-codey...

Is that also how we can for disabling of regular boolean options?

Flamefire: Yes, that is basically copied verbatim from https://github.com/easybuilders/easybuild-framework/blob/a42e25be0300a263747c6742c06197b7fcdcddf6/easybuild/base/generaloption.py#L294-L297

PR #4782, opened 2025-03-03
use `-march=native` rather than `-xHost` for Intel oneAPI compilers >= 2025.0

### LINKED ISSUE #4744: With new clang based intel compilers (ifx, icx, icpx) we should use -march=native
https://github.com/easybuilders/easybuild-framework/blob/b37f70756960c1686875974f8381491df280cc00/easybuild/toolchains/compiler/inteliccifort.py#L62

We should set this to `-march=native` but *only* when the new ifx, icx, icpx are used. 

This will be better for AMD cpus. I think @bartoldeman said that from 2025 onwards the old compilers are gone, so, we shouldn't use that old bad -x flags that are only supported on Intel cpus.

boegel:
fixes #4744

### DIFF
--- a/easybuild/toolchains/compiler/intel_compilers.py
+++ b/easybuild/toolchains/compiler/intel_compilers.py
@@ -124,5 +124,13 @@ class IntelCompilers(IntelIccIfort):
                 (systemtools.X86_64, systemtools.INTEL): '-march=x86-64 -mtune=generic',
             }
 
+            # use -march=native for sufficient recent versions of Intel oneAPI compilers
+            if LooseVersion(comp_ver) >= LooseVersion('2025.0'):
+                self.options.options_map['optarch'] = ['-march=native']
+                self.COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
+                    (systemtools.X86_64, systemtools.AMD): '-march=native',
+                    (systemtools.X86_64, systemtools.INTEL): '-march=native',
+                }
+
         # skip IntelIccIfort.set_variables (no longer relevant for recent versions)
         Compiler.set_variables(self)


Micket (dismissed):
I have been using march=native for intel-2024a, but mostly tested on zen4, where things seem fine. I guess we should build some random crap on icelake nodes and compare the two flags.

I'll ask my colleageus to help try it out.

boegel:
This result is reason enough for me to put this on hold, and keep using `-xHOST` for now, until we learn more...

Micket:
Just some testing some more, I think the test examples i ran from shootout aren't that great for testing this. Many of them aren't really affected by advanced instructions at all, and when they are, it seems to me that xHost wins. 

Maybe the -march=... flags are just using llvms optimization passes, while xhost does the "intel specific magic"? They are definitely not doing the same code gen.
But the counter example where it fails hard is basically just some trivial dot-product loops. And -march=native there seems to be such a crappy job that it's worse than -march=generic, so something is going horrible wrong with spectralnorm example when using Fortran.

But to be clear xHost is not viable option on non-intel CPUs (the binaries wont run).
Those with mixed systems should probably consider `-march=... -ax=....`  as that would make a fat binary that dispatches on intel hardware.

I wonder if we could just set this conditionally? I looked for a nice clean way to detect if it's a amd system, but I found nothing useful.

Micket:
https://docs.amd.com/r/en-US/63857-AOCC-quick-start-guide/AMD-EPYC-9xx5-Series-Processors-Compiler-Options-Quick-Reference

At least for Zen5, and very recent one-api compilers, AMD recommends using `-axCORE-AVX512? .
Not sure we need the -ax part, -x should suffice for us (faster compile and smaller binaries maybe?)

We should make a matrix over different Zen architectures and different intel compiler versions to verify

hattom:
> https://docs.amd.com/r/en-US/63857-AOCC-quick-start-guide/AMD-EPYC-9xx5-Series-Processors-Compiler-Options-Quick-Reference
> 
> At least for Zen5, and very recent one-api compilers, AMD recommends using `-axCORE-AVX512? . Not sure we need the -ax part, -x should suffice for us (faster compile and smaller binaries maybe?)

It absolutely says this. But this goes against everything I thought I knew about `-x` and `-ax` flags. At least for the fortran compilers. I was sure the `-ax` code path was only executed if the check decides the Intel CPU can run it at runtime.
Testing would be sensible, but I had thought that it wouldn't ever take that code path on AMD.

PR #4787, opened 2025-03-04
Fix the type check after type conversion

### LINKED ISSUE #4785: convert_value_type checks for checksums has broken code
I was going to see what it would take to allow checksums to be a plain top level dict, and it turns out it's very simple. The only code preventing this is this code
https://github.com/easybuilders/easybuild-framework/blob/4702af0ed1402817f39e1c3316fb874f6466991e/easybuild/framework/easyconfig/types.py#L552-L557
which converts everything to a list (of just the keys).

First problem is that, while our documentation suggests that a simple single string is supported, this is not the case. Such a string would be converted to a list of single characters. This never worked as far as I can tell. I suppose this is good, because that's one pattern we can then just remove. It must always be a list.

Secondly, if one tries to modify `to_checksums`, to for exmaple allow a dict to be passed through
```
def to_checksums(checksums):
    """Ensure correct element types for list of checksum items: convert list elements to tuples."""
    try:
        if isinstance(checksums, dict):
            return {key: _to_checksum(checksum) for key, checksum in checksums.items()}
        else:
            return [_to_checksum(checksum) for checksum in checksums]
    except ValueError as e:
        raise EasyBuildError('Invalid checksums: %s\n\tError: %s', checksums, e)
```
or similar, the code trips up over
https://github.com/easybuilders/easybuild-framework/blob/4702af0ed1402817f39e1c3316fb874f6466991e/easybuild/framework/easyconfig/types.py#L252-L253

which is trying use
https://github.com/easybuilders/easybuild-framework/blob/4702af0ed1402817f39e1c3316fb874f6466991e/easybuild/framework/easyconfig/types.py#L683-L695

which is defined here:
https://github.com/easybuilders/easybuild-framework/blob/4702af0ed1402817f39e1c3316fb874f6466991e/easybuild/framework/easyconfig/types.py#L656-L657

but this is not a value input for `isinstance`, and the code crashes with
```
TypeError: isinstance() arg 2 must be a type or tuple of types
```

clearly, this part of the code has never even worked. Shall we just go ahead and delete this extra unnecessary check? 

I really only want to allow for checksum to a be a dict, and it almost seems from the code like it should be supported already by `CHECKSUM_DICT` via `CHECKABLE_TYPES` , but since the code always checks for `is_value_of_type(val, typ)`  anyway, where `typ == list` then this means nothing. It's always going to be forcible converted to a `list` no matter what, so it's pointless to add all of these to `CHECKABLE_TYPES`?

@Flamefire am I missing something here?

Flamefire:
We cannot use `isinstance` on non-trivial types, i.e. our tuple-specifications. Incidently it works if the result is the type in the first element of the tuple, otherwise it fails with a `TypeError` because the remainder is not a type.

The correct way is to use our `is_value_of_type`.

2 small corrections:

1. We don't need `EASY_TYPES` which had some types missing, e.g. `float`. Instead we can check if a type was passed and hence can be used with `isinstance`.
2. `convert_value_type` shouldn't be called when the type is already correct. We ensure that in all usages in framework code. So issue a warning for that case.

And improve the error message (CI failure was a bit hard to debug):
Now: `Converting [...] to type <type 'DEPENDENCIES'> failed` similar to:   
`Converting [...] to type <class 'int'> failed`

Fixes #4785

I had to adapt the tests:
- `[('foo', '1.2.3')]` was considered a wrong type for `dependencies` while it is not, as the fixed code confirms

### Discussion

1. What shall we do with the result of `to_*`? Shall we check against the expected type as defined to ensure what we pass is always "correct", shall we do a "soft" check only checking the outer type (list, tuple, dict, str, int...) or do no checks at all and assume it is correct?    
Problem is that we might want to delegate format checking to e.g. the EasyConfig instance which may print
> == FAILED: Installation ended unsuccessfully (build directory: /dev/shm/zlib/1.3.1/GCCcore-13.3.0-test-debug): build failed (first 300 chars): Invalid checksum spec '112345689': should be a string (MD5 or SHA256), 2-tuple (type, value), or tuple of alternative checksum specs. (took 0 secs)

While we'd need to do that in the type-format check code or get e.g.
>     ERROR: Failed to process easyconfig /apps/c3se-test-easyconfigs/zlib-1.3.1-GCCcore-13.3.0-test.eb: Converting type of [123456] (<class 'list'>) to <type 'CHECKSUMS'> failed: 'Invalid checksums: [112345689]\n\tError: Unexpected type of "<class 'int'>": 123456'

2. Shall we not do any checking at all and assume `to_*` does the right thing? This would mean that we have to do stricter checks in the handling code.
3. Shall we allow "wildcard" types? I.e. a type of `None` in the type-spec-definition would mean "accept any type, don't check". That would allow selectively moving checks to other places while being explicit about it. We have something similar already if we use e.g. `tuple` instead of `TUPLE_OF_STRINGS` but then we can ignore the type of e.g. specific dict entries instead of all

See discussion in #4785


boegel:

Flamefire:
For the `sanity_check_paths` something came up: Shall we require the "files" and "dirs" key in the type check or have them optional there and handle it someplace else, e.g. as currently in the CI that fails if not both are set? I think in favor of better error messages defining most things as optional is a good idea. That's what I changed here and wanted to highlight the explicit decision.

Side note: I think we should not require both keys. A missing key should be equivalent to an empty list.

akesandgren:
> Side note: I think we should not require both keys. A missing key should be equivalent to an empty list.

I agree on that part at least.

easybuild/framework/easyconfig/types.py
```diff
@@ -602,32 +630,21 @@ def ensure_iterable_license_specs(specs):
...
-STRING_OR_TUPLE_OR_DICT_LIST = (list, as_hashable({'elem_types': [str, TUPLE_OF_STRINGS, STRING_DICT]}))
 SANITY_CHECK_PATHS_ENTRY = (list, as_hashable({'elem_types': [str, TUPLE_OF_STRINGS, STRING_OR_TUPLE_DICT]}))
 SANITY_CHECK_PATHS_DICT = (dict, as_hashable({
     'elem_types': {
         SANITY_CHECK_PATHS_FILES: [SANITY_CHECK_PATHS_ENTRY],
         SANITY_CHECK_PATHS_DIRS: [SANITY_CHECK_PATHS_ENTRY],
     },
-    'opt_keys': [],
-    'req_keys': [SANITY_CHECK_PATHS_FILES, SANITY_CHECK_PATHS_DIRS],
+    'opt_keys': [SANITY_CHECK_PATHS_FILES, SANITY_CHECK_PATHS_DIRS],
```
Flamefire, line 646: This actually needs to be discussed: We currently allow missing dirs/files in sanity-check-paths and only fail it in CI.

We should decide where we want to handle this:
- Make both optional when running easybuild but fail on CI for contributions
- Make both fully optional and default them to empty lists in the code using it (sanity_check_step)
- Make them required here (as before) and fill them with empty lists in the "conversion" function if not set, then we don't need to check if the keys exist in sanity_check_step. Optionally enforce non-empty lists on CI

PR #4788, opened 2025-03-04
Allow test case filtering by class name

Flamefire:
It sometimes is useful to run either all tests from a class or a specific test from a class where the name might also be used in another class.
Use the common convention of `'<class>.<test>'` and match against that allowing e.g. "TypeCheckingTest.test_check_type" as a filter.


Flamefire:
Might be obsolete with https://github.com/easybuilders/easybuild-framework/pull/3790 which I'd prefer

boegel (approved):

PR #4789, opened 2025-03-04
Update GitHub Actions config

Flamefire:
Ubuntu 20.04 runners will be removed soon and may fail to start already (brownout)

PR #4790, opened 2025-03-04
add custom GitHub Actions workflow to run tests with Python 3.6 in Ubuntu 20.04 container

boegel:


.github/workflows/unit_tests_python36.yml
```diff
@@ -0,0 +1,82 @@
...
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+concurrency:
+  group: ${{format('{0}:{1}:{2}', github.repository, github.ref, github.workflow)}}
+  cancel-in-progress: true
+
+jobs:
+  test_python36:
+    runs-on: ubuntu-22.04
```
Flamefire, line 14: In other projects I found it a good idea to use ubuntu-latest for all jobs using containers.

```suggestion
    runs-on: ubuntu-latest
```

Flamefire:
I also started working on the before in #4789 using the Python versions available from the [action](https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json) I also opened an [issue](https://github.com/actions/setup-python/issues/1048) with them to support Python 3.6 on Ubuntu 22+

I also had to remove a job that you might want to add here:
```
          # There may be encoding errors in Python 3 which are hidden when an UTF-8 encoding is set
          # Hence run the tests (again) with LC_ALL=C and Python 3.6 (or any < 3.7)
          - python: 3.6
            modules_tool: ${{needs.setup.outputs.lmod8}}
            lc_all: C
```

PR #4795, opened 2025-03-06
Update GitHub actions workflows to use `ubuntu-24.04` where possible

Flamefire:
#4789 rebased on 5.0x

use ubuntu 24.04 or -latest whenenver possible to make updating less likely and easier

### DIFF
--- a/.github/workflows/container_tests.yml
+++ b/.github/workflows/container_tests.yml
@@ -17,10 +17,10 @@ jobs:
         python: [3.7]
       fail-fast: false
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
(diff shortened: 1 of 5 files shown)


.github/workflows/container_tests.yml
```diff
@@ -17,10 +17,10 @@ jobs:
         python: [3.7]
       fail-fast: false
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
+    - uses: actions/checkout@v4
```
boegel, line 20: Using a specific commit is deliberate, out of abundance of care.

We'll only update to newer versions/commits when really needed, so please change back.

boegel (changes requested):

boegel (approved):

PR #4799, opened 2025-03-07
Allow empty string in `modextrapaths`

appolloford:
After PR #4774, when empty string is used in `modextrapaths`, for example `modextrapaths = {'PATH': ''}` in `dotNET-Core-8.0.302.eb`, easybuild does not add `paths` key into `env_var_opts` and fails silently. This induces error when checking `paths` key existence. This PR tries to make sure `paths` is added when empty string is used.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1716,7 +1716,11 @@ class EasyBlock(object):
             }
 
             try:
-                env_var_opts.update(extra_opts)
+                # make sure paths is added for empty string
+                if extra_opts == '':
+                    env_var_opts['paths'] = extra_opts
+                else:
+                    env_var_opts.update(extra_opts)
             except ValueError:
                 # no options provided, so must be only a list of string values specifying paths
                 env_var_opts['paths'] = extra_opts
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1715,19 +1715,14 @@ class EasyBlock(object):
                 'prepend': True,
             }
 
-            try:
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1716,7 +1716,11 @@ def inject_module_extra_paths(self):
             }
 
             try:
-                env_var_opts.update(extra_opts)
+                # make sure paths is added for empty string
+                if extra_opts == '':
+                    env_var_opts['paths'] = extra_opts
+                else:
+                    env_var_opts.update(extra_opts)
```
lexming, line 1723: This works, but this `try..else` is really not appropriate given the that `update` does not raise a `ValueError` on empty strings.

Can you please replace the full try..else to:
```
if not isinstance(extra_opts, dict):
    extra_opts = {'paths': extra_opts}

env_var_opts.update(extra_opts)

if 'paths' not in env_var_opts:
    error_msg = f"'paths' key not set for ${env_var_name} in '{ec_param}' easyconfig parameter"
```

lexming (changes requested):
Thanks for reporting this issue. The case with an empty string is indeed not handled properly. See my suggested solution below

easybuild/framework/easyblock.py
appolloford: Thank you for the comment. This looks much cleaner.

lexming (approved):

boegel:
I've enhanced the tests to cover this use case in:
* https://github.com/easybuilders/easybuild-framework/pull/4803

PR #4801, opened 2025-03-07
avoid re-assigning `env_var_name` of outer for loop in inner for loop

lexming:


Micket (approved):
lgtm

branfosj (approved):

boegel:
Is this just code cleanup, or was there a genuine bug here?

If the latter, can we come up with an enhanced test that would've caught it?

lexming:
Definitely a bug fix, as it is wrong to use the same looping variable in two nested loops.

Has it cause any actual failures? Not that I know.

Some linters can catch such issues in tests. I use `ruff` and it did trigger here.

boegel:
@lexming Yeah, it's definitely not OK to re-assign that variable, no doubt about that.

Small follow-up PR to also fix the log message in that loop:
* https://github.com/easybuilders/easybuild-framework/pull/4802

PR #4810, opened 2025-03-13
make `python2_is_deprecated` function cause an exit when using Python 2

branfosj:


easybuild/tools/py2vs3/__init__.py
```diff
@@ -34,26 +32,3 @@
 
 _log = fancylogger.getLogger('py2vs3', fname=False)
 _log.deprecated("Using py2vs3 is deprecated, since EasyBuild no longer runs on Python 2.", '6.0')
-
-
-def python2_is_deprecated():
```
boegel, line 39: The functionality provided by `easybuild.tools.py2vs3` is deprecated, but why would we hard break the API just to clean up some code at this point?

`from easybuild.tools.py2vs3 import python2_is_deprecated` should keep working, but produce a deprecation warning

So I'm not in favor of this (unless I'm missing something that makes this cause trouble)

easybuild/tools/py2vs3/__init__.py
```diff
@@ -34,26 +32,3 @@
...
-            "see https://docs.easybuild.io/en/latest/Python-2-3-compatibility.html for more information.",
-        ]
-        max_len = max(len(x) for x in warning_lines)
-        for i in range(len(warning_lines)):
-            line_len = len(warning_lines[i])
-            warning_lines[i] = '!!! ' + warning_lines[i] + ' ' * (max_len - line_len) + ' !!!'
-        max_len = max(len(x) for x in warning_lines)
-        warning_lines.insert(0, '!' * max_len)
-        warning_lines.append('!' * max_len)
-        sys.stderr.write('\n\n' + '\n'.join(warning_lines) + '\n\n\n')
```
boegel, line 59: Rather than throwing this function away, we should just print an error + do `exit 1`.

That doesn't entirely match the name of the function, but that's a detail we can ignore imho.

branfosj: 380ea38

easybuild/tools/py2vs3/__init__.py
branfosj: 380ea38

easybuild/tools/py2vs3/__init__.py
```diff
@@ -38,22 +38,9 @@
...
+        sys.stderr.write('\n\nRunning EasyBuild v5.0 with Python v2 does not work. Use Python >= 3.6.\n\n\n')
```
boegel, line 45:
```suggestion
        sys.stderr.write('\n\nEasyBuild v5.0+ is not compatible with Python v2. Use Python >= 3.6.\n\n\n')

boegel (changes requested):

boegel (approved):
lgtm

PR #4818, opened 2025-03-17
Make sure cached easyconfigs don't get modifed

Flamefire:
The type that is cached is a dictionary containing the EasyConfig in the `"ec"` key. Hence we end up invoking `dict.copy` not `ec.copy`

This results in not making independent copies and hence modifications affect all references even those gotten later

Revert https://github.com/easybuilders/easybuild-framework/commit/38361401dd36985c2045f62e19df1dd93319c36e which introduced the issue

Our current `EasyConfig.copy` method effectively parses the EasyConfig text again invoking e.g. the `parse_hook` again.

I guess `__deepcopy__` needs to be more smart than this to not also miss attributes set after the construction. Hence I left that out and just used `ec.copy()` as was likely intended by the above commit. It just needs to be done on the EasyConfig instance while the dict itself can be deep-copied if it doesn't contain the easyconfig.


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -2234,6 +2234,22 @@ def resolve_template(value, tmpl_dict, expect_resolved=True):
     return value
 
 
+def _copy_ec_dict(easyconfig):
+    """Copy an easyconfig dict as (initially) parsed"""
+    ec = easyconfig.pop('ec')  # Don't copy this via deepcopy
```
boegel: Comment should be clarified: why not

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -2234,6 +2234,22 @@ def resolve_template(value, tmpl_dict, expect_resolved=True):
...
+    try:
+        new_easyconfig = copy.deepcopy(easyconfig)  # Copy the rest of the dict
```
boegel, line 2258: Can you clarify why this is in a `try`?

What if this fails? Then `new_easyconfig` will be undefined, and we'll get a nasty crash, no?

boegel (changes requested):

easybuild/framework/easyconfig/easyconfig.py
Flamefire: Done: We need `ec.copy`

easybuild/framework/easyconfig/easyconfig.py
Flamefire: This is a try-finally. So it will still propagate all errors. The only difference is that in all cases `ec` (removed above) is put back to the original easyconfig

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -2236,6 +2236,23 @@ def resolve_template(value, tmpl_dict, expect_resolved=True):
...
 
 
+def _copy_ec_dict(easyconfig):
+    """Copy an easyconfig dict as (initially) parsed"""
+    # deepcopy on the ec doesn't fully copy it, but requires the copy() method so temorarily remove it
+    ec = easyconfig.pop('ec')
+    try:
+        new_easyconfig = copy.deepcopy(easyconfig)  # Copy the rest of the dict
+    finally:
+        easyconfig['ec'] = ec  # Put back
```
```suggestion
    # deepcopy on the EasyConfig instance doesn't fully copy it,
    # but requires the copy() method, so temporarily remove it
    ec = easyconfig.pop('ec')
    try:
        new_easyconfig = copy.deepcopy(easyconfig)  # Copy the rest of the dict
    finally:
        # always put back EasyConfig instance
        easyconfig['ec'] = ec
```

boegel (changes requested):

boegel (approved):
lgtm

PR #4831, opened 2025-03-24
Fix missed message in trivial equal-asserts

Flamefire:
The default error is raised for trivial types but the message was not passed along and is missing in that case.

### DIFF
--- a/easybuild/base/testing.py
+++ b/easybuild/base/testing.py
@@ -86,7 +86,7 @@ class TestCase(OrigTestCase):
         """Make assertEqual always print useful messages"""
 
         try:
-            super(TestCase, self).assertEqual(a, b, msg)
+            super(TestCase, self).assertEqual(a, b, msg=msg)
         except AssertionError as e:
             if not self._is_diffable(a) or not self._is_diffable(b):
                 raise
--- a/easybuild/base/testing.py
+++ b/easybuild/base/testing.py
@@ -86,7 +86,7 @@ class TestCase(OrigTestCase):
         """Make assertEqual always print useful messages"""
 
         try:
-            super(TestCase, self).assertEqual(a, b)
+            super(TestCase, self).assertEqual(a, b, msg)
         except AssertionError as e:
             if not self._is_diffable(a) or not self._is_diffable(b):
                 raise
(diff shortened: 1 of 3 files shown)


easybuild/base/testing.py
```diff
@@ -86,7 +86,7 @@ def assertEqual(self, a, b, msg=None):
         """Make assertEqual always print useful messages"""
 
         try:
-            super(TestCase, self).assertEqual(a, b)
+            super(TestCase, self).assertEqual(a, b, msg)
```
ocaisa, line 89:
```suggestion
            super(TestCase, self).assertEqual(a, b, msg=msg)
```

ocaisa (approved):

Flamefire:
Looks like there was an update to flake8 or previously missed check. Fixed

easybuild/framework/easyconfig/format/version.py
```diff
@@ -499,7 +499,6 @@ def _get_all_toolchain_names(self, search_string=''):
         Initialise each search_toolchain request, save in module constant TOOLCHAIN_NAMES.
         :param search_string: passed to search_toolchain function.
         """
-        global TOOLCHAIN_NAMES
```
ocaisa, line 502: This is where my ignorance shows, but don't we need this for line 505?

boegel: There's only one `TOOLCHAIN_NAMES`, so the `global` statement is overkill.

Quick test without `global`:
```python
>>> TEST = {}
>>> class Alan(object):
...     def test(self):
...         TEST['one'] = 1
...
>>> a = Alan()
>>> TEST
{}
>>> a.test()
>>> TEST
{'one': 1}
```

ocaisa (approved):
LGTM

easybuild/framework/easyconfig/format/version.py
Flamefire: The point of "global" is that an assignment to the symbol affects the global symbol:
```
TEST = []
def foo():
    TEST.append(42)  # Access to existing symbol -> global
    TEST = [1,2,3]  # new local variable
    TEST.append(42)  # Access to local symbol shadowing global symbol
```

PyLint and others actually warn when shadowing a global name because this behavior is hard to reason about.

PR #4834, opened 2025-04-01
Remove Python2 constructs

Flamefire:
This avoids a lot of C&P by removing:
- `super(x, y)` -> `super()`
- `class Foo(object):` -> `class Foo:`


Micket (approved):
lgtm

PR #4836, opened 2025-04-02
remove useless 'global' statements

boegel:
`global` statements are only required when a function or method redefines a global variable (like in `dry_run_set_dirs` in `easybuild.tools.build_log`).

This fixes a failing code style check caught by new version of `flake8`:
```
./easybuild/framework/easyconfig/format/version.py:502:9: F824 `global TOOLCHAIN_NAMES` is unused: name is never assigned in scope
./easybuild/main.py:243:5: F824 `global _log` is unused: name is never assigned in scope
```

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -240,8 +240,6 @@ def process_easystack(easystack_path, args, logfile, testing, init_session_state
     # keep copy of original environment, so we can restore it for every easystack entry
     init_env = copy.deepcopy(os.environ)
 
-    global _log
-
     # TODO: insert fast loop that validates if all command line options are valid. If there are errors in options,
     # we want to know early on, and this loop potentially builds a lot of packages and could take very long
     # for path in orig_paths:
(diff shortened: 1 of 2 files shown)


easybuild/main.py
```diff
@@ -615,7 +611,6 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None, pr
 
     options, orig_paths = eb_go.options, eb_go.args
 
-    global _log
```
ocaisa, line 615: This one is needed isn't it as you set it below

boegel:
Part of changes originally proposed have already been made in #4831, but none of the `global _log` statements are required in `main.py`, so let's get rid of all of them...

easybuild/main.py
boegel: Oh right, that's kind of hidden...

You're right.

easybuild/main.py
```diff
@@ -308,7 +308,6 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
     """
     options = eb_go.options
 
-    global _log
```
ocaisa, line 311: I think this is needed as you set it in the following line

PR #4847, opened 2025-04-09
Avoid failure when only some passed easyconfigs exist

Flamefire:
When processing multiple EasyConfigs where one does not exist `samefile` will abort with a `FileNotFoundError`.
So move the existance check to the top of the loop.


boegel (approved):
lgtm

PR #4850, opened 2025-04-12
Add "-" before "DMKL_ILP64" CFLAGS

thanghn90:
There should be a minus sign "-" before "DMKL_ILP64" in CFLAGS. Without this, it's breaking NWChem build. I'm not sure why this is under the radar for so long. Probably people just gave up on waiting 12 hours to build NWChem. I know modifying this core framework component will require a lot of testing, so take your time to review this pull request.

Currently, these easyconfig files will be affected (based on a simple "grep" pattern of "'i8'" toolchain option in the easyconfig files, and filtering out non-intel toolchain):

NWChem-7.0.2-intel-2021a.eb
NWChem-7.0.2-intel-2022a.eb
NWChem-7.2.2-intel-2023a.eb
NWChem-7.2.3-intel-2024a.eb
DIRAC-22.0-intel-2021a-int64.eb
DIRAC-23.0-intel-2022b-int64.eb
DIRAC-23.0-intel-2023a-int64.eb

### DIFF
--- a/easybuild/toolchains/linalg/intelmkl.py
+++ b/easybuild/toolchains/linalg/intelmkl.py
@@ -136,7 +136,7 @@ class IntelMKL(LinAlg):
             # ilp64/i8
             self.BLAS_LIB_MAP.update({"lp64": '_ilp64'})
             # CPP / CFLAGS
-            self.variables.nappend_el('CFLAGS', 'DMKL_ILP64')
+            self.variables.nappend_el('CFLAGS', '-DMKL_ILP64')
 
         # exact paths/linking statements depend on imkl version
         root = self.get_software_root(self.BLAS_MODULE_NAME)[0]


Micket (approved):
lgtm

Micket:
I'm confused as to how this was missed, so I've tried to look at how this occurred, and if there are other cases where we might end up doing something wrong. I suppose i8 is just that rare, and these commercial things aren't as well tested.
This must be fallout from https://github.com/easybuilders/easybuild-framework/pull/4698 where `self.variables` ends up using a `StrList` instead of a `FlagList` (that's my only explanation)
The underlying code for these lists is so convoluted I struggle to follow it.
Maybe we also have problems with `fujitsussl.py`
```
self.variables.nappend(flags_var, ['SSL2BLAMP'])
self.variables.nappend(flags_var, ['SSL2'])
self.variables.nappend(flags_var, ['SCALAPACK'])
```
but I can't tell (or test, as I have no fujitsu system)

PR #4851, opened 2025-04-23
Remove superflous assignment in _sanity_check_step

Flamefire:
`sanity_check_load_module` returns `self.fake_mod_data` so there is no need to assign it to itself.
Remove it to avoid confusion when e.g. implementing  easyblocks using this functions


boegel (approved):
lgtm

PR #4852, opened 2025-04-23
replace deprecation warning about reproducible tarballs on Python older than 3.9 to a regular warning

lexming:
As discussed in Slack, this deprecation warning causes issues with boegelbot because the easyconfigs in our repo use reproducible tarballs, but we still need to test with Python 3.6 as it is a supported version.


lexming:
Indeed, that works too. However, this deprecation warning is still weird because the only solution is to move to Python 3.9+, so it basically translates to _Python 3.[678] are deprecated_, but those are still supported. So I would either declare Python lower than 3.9 deprecated globally, or make this warning not a deprecation.

boegel (approved):
lgtm

PR #4855, opened 2025-04-24
add support for environment variables in `set_environment` of module generators

lexming:
This PR enhances the `set_environment` methods in module generators to handle contents of variables that have references to shell environment variables (_i.e._ `$ENV_VAR`).

The effects of this PR are better seen in `test_env` in `test/framework/module_generator.py`, which is heavily enhanced to test a lot more cases. See [test/framework/module_generator.py#L956](https://github.com/easybuilders/easybuild-framework/blob/5c594ad603db8ed408a45710979538e159a27fbe/test/framework/module_generator.py#L956)

In practice, this now allows the following in `modextravars`:

```
modextravars = {
    'SOME_VARIABLE_PYTHON': '$EBROOTPYTHON/bin/python',
    'SOME_VARIABLE_CONFIG': '$HOME/.config/file',
}
```

Additionally, in the case of `ModuleGeneratorLua`, the static template `PATH_JOIN_TEMPLATE` used to construct `pathJoin` commands in Lua is replaced with a more powerful `_path_join_cmd` method that properly splits the components of the path:
```
# OLD
pathJoin(root, "foo/bar")
# NEW
pathJoin(root, "foo", "bar")
```

Companion PR in easyconfigs:
* https://github.com/easybuilders/easybuild-easyconfigs/pull/22801

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -52,7 +52,6 @@ from easybuild.tools.filetools import convert_name, mkdir, read_file, remove_fil
 from easybuild.tools.modules import ROOT_ENV_VAR_NAME_PREFIX, EnvironmentModulesC, Lmod, modules_tool
 from easybuild.tools.utilities import get_subclasses, nub, quote_str
 
-
 _log = fancylogger.getLogger('module_generator', fname=False)
 
 
@@ -1152,7 +1151,7 @@ class ModuleGeneratorLua(ModuleGenerator):
     LOAD_TEMPLATE_DEPENDS_ON = 'depends_on("%(mod_name)s")'
     IS_LOADED_TEMPLATE = 'isloaded("%s")'
 
-    PATH_JOIN_TEMPLATE = 'pathJoin(root, "%s")'
+    PATH_JOIN_TEMPLATE = 'pathJoin(root, "%s")'  # TODO: remove after 6.0, replaced by _path_join_cmd()
     UPDATE_PATH_TEMPLATE = '%s_path("%s", %s)'
     UPDATE_PATH_TEMPLATE_DELIM = '%s_path("%s", %s, "%s")'
 
@@ -1166,6 +1165,19 @@ class ModuleGeneratorLua(ModuleGenerator):
         if self.modules_tool:
             if self.modules_tool.version and LooseVersion(self.modules_tool.version) >= LooseVersion('7.7.38'):
                 self.DOT_MODULERC = '.modulerc.lua'
+
+    @staticmethod
+    def _path_join_cmd(path):
+        "Return 'pathJoin' command for given path string"
+        path_components = [quote_str(p) for p in path.split(os.path.sep) if p]
+
+        path_root = quote_str(os.path.sep) if os.path.isabs(path) else 'root'
+        path_components.insert(0, path_root)
+
+        if len(path_components) > 1:
+            return f'pathJoin({", ".join(path_components)})'
+        # no need for a pathJoin for single component paths
+        return os.path.join(*path_components)
 
     def check_version(self, minimal_version_maj, minimal_version_min, minimal_version_patch='0'):
         """
@@ -1448,7 +1460,7 @@ class ModuleGeneratorLua(ModuleGenerator):
                 # use pathJoin for (non-empty) relative paths
                 if path:
                     if expand_relpaths:
-                        abspaths.append(self.PATH_JOIN_TEMPLATE % path)
+                        abspaths.append(self._path_join_cmd(path))
                     else:
                         abspaths.append(quote_str(path))
                 else:
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -430,19 +430,26 @@ class ModuleGenerator:
 
         return res
 
-    def unpack_setenv_value(self, env_var_name, env_var_val):
+    def unpack_setenv_value(self, *args, **kwargs):
+        """
+        """
+        self.log.deprecated("...", '6.0')
+        value, use_pushenv, _ = self._unpack_setenv_value(*args, **kwargs)
+        return value, use_pushenv
+
+    def _unpack_setenv_value(self, env_var_name, env_var_val):
         """
         Unpack value that specifies how to define an environment variable with specified name.
         """
         use_pushenv = self.DEFAULT_MODEXTRAVARS_PUSHENV
-        shell_vars = self.DEFAULT_MODEXTRAVARS_SHELL_VARS
+        resolve_shell_vars = self.DEFAULT_MODEXTRAVARS_SHELL_VARS
 
         # value may be specified as a string, or as a dict for special cases
         if isinstance(env_var_val, str):
             value = env_var_val
         elif isinstance(env_var_val, dict):
             use_pushenv = env_var_val.get('pushenv', self.DEFAULT_MODEXTRAVARS_PUSHENV)
-            shell_vars = env_var_val.get('shell_vars', self.DEFAULT_MODEXTRAVARS_SHELL_VARS)
+            resolve_shell_vars = env_var_val.get('shell_vars', self.DEFAULT_MODEXTRAVARS_SHELL_VARS)
             try:
                 value = env_var_val['value']
             except KeyError as err:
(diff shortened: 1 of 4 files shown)


smoors:
after a quick discussion with @lexming he will add an option to opt out of resolving the variables

lexming:
@smoors as discussed, resolution of shell variables is enabled by default and can be disabled with an extra option in `modextravars` called `shell_vars`:

```
modextravars = {
    'VARIABLE_NAME': '$HOME/path',
    'PASSWORD_VAR': {'value': 'dijd$12h12hr', 'shell_vars': False},
}
```

Updated unit tests accordingly.

smoors (approved):
lgtm

easybuild/tools/module_generator.py
```diff
@@ -133,6 +132,15 @@ class ModuleGenerator:
     # a single level of indentation
     INDENTATION = ' ' * 4
 
+    # shell environment variable name: ${__}VAR_NAME_00_SUFFIX
+    REGEX_SHELL_VAR_PATTERN = r'[A-Z_]+[A-Z0-9_]+'
```
boegel, line 136: Just asking: strictly speaking environment variables don't have to be all caps, that's just a convention.

Are we happy with not recognizing lowercase environment variables like `$THIS_is_an_example`?

easybuild/tools/module_generator.py
```diff
@@ -426,24 +434,25 @@ def unpack_setenv_value(self, env_var_name, env_var_val):
...
+            except KeyError as err:
                 raise EasyBuildError("Required key 'value' is missing in dict that specifies how to set $%s: %s",
-                                     env_var_name, env_var_val)
+                                     env_var_name, env_var_val) from err
         else:
             raise EasyBuildError("Incorrect value type for setting $%s environment variable (%s): %s",
                                  env_var_name, type(env_var_val), env_var_val)
 
-        return value, use_pushenv
+        return value, use_pushenv, shell_vars
```
boegel, line 455: Hmm, this changes the function signature of this method, which is a breaking change to the API...

In theory (though I admit it's unlikely) there could be hooks/easyblocks/etc that use the `unpack_setenv_value` method, which would break because of this change.

easybuild/tools/module_generator.py
```diff
@@ -1152,12 +1161,13 @@ class ModuleGeneratorLua(ModuleGenerator):
     LOAD_TEMPLATE_DEPENDS_ON = 'depends_on("%(mod_name)s")'
     IS_LOADED_TEMPLATE = 'isloaded("%s")'
 
-    PATH_JOIN_TEMPLATE = 'pathJoin(root, "%s")'
+    PATH_JOIN_TEMPLATE = 'pathJoin(root, "%s")'  # TODO: remove after 6.0, replaced by _path_join_cmd()
```
boegel, line 1164: If we really want to deprecate this, we should make it a class property that emit a deprecation warning via `self.log.deprecated`?

easybuild/tools/module_generator.py
```diff
@@ -1167,6 +1177,19 @@ def __init__(self, *args, **kwargs):
...
+    @staticmethod
+    def _path_join_cmd(path):
+        "Return 'pathJoin' command for given path string"
+        path_components = [quote_str(p) for p in path.split(os.path.sep) if p]
+
+        path_root = quote_str(os.path.sep) if os.path.isabs(path) else 'root'
+        path_components.insert(0, path_root)
+
+        if len(path_components) > 1:
+            return f'pathJoin({", ".join(path_components)})'
```
boegel, line 1189: Using an f-string here makes this very hard to read, how about

```suggestion
            return 'pathJoin(' + ', '.join(path_components) + ')'

File: easybuild/tools/module_generator.py [hunk 5]
```diff
@@ -1513,19 +1536,23 @@ def set_environment(self, key, value, relpath=False):
...
-                val = self.PATH_JOIN_TEMPLATE % value
-            else:
-                val = 'root'
+            set_value = self._path_join_cmd(set_value)
+            if shell_vars:
+                set_value = self.REGEX_QUOTE_SHELL_VAR.sub(r'os.getenv("\1")', set_value)
         else:
-            val = quote_str(value)
+            if shell_vars:
+                set_value = self.REGEX_SHELL_VAR.sub(rf'{self.CONCAT_STR}os.getenv("\1"){self.CONCAT_STR}', set_value)
```
boegel, line 1547: This also deserves a comment on top with an example of what this does

easybuild/tools/module_generator.py
```diff
@@ -1513,19 +1536,23 @@ def set_environment(self, key, value, relpath=False):
...
+        set_value, use_pushenv, shell_vars = self.unpack_setenv_value(key, value)
 
         if relpath:
-            if value:
-                val = self.PATH_JOIN_TEMPLATE % value
-            else:
-                val = 'root'
+            set_value = self._path_join_cmd(set_value)
+            if shell_vars:
+                set_value = self.REGEX_QUOTE_SHELL_VAR.sub(r'os.getenv("\1")', set_value)
```
boegel, line 1544: This deserves a comment on top with an example of what this does

boegel (changes requested):

easybuild/tools/module_generator.py
boegel: For now, we'll only supports "proper" environment variables names, i.e. in all caps.

easybuild/tools/module_generator.py
boegel: I'll look into making `unpack_setenv_value` an internal method, by prefixing it with `_`, and then making the public method `unpack_setenv_value` deprecated (which allows us to avoid breakage actually)

easybuild/tools/module_generator.py
boegel: fixed in https://github.com/easybuilders/easybuild-framework/commit/510e8f0942598ec8856ffb5d7fefd0c56fac235a

easybuild/tools/module_generator.py
```diff
@@ -1513,19 +1536,23 @@ def set_environment(self, key, value, relpath=False):
...
         else:
-            val = quote_str(value)
+            if shell_vars:
+                set_value = self.REGEX_SHELL_VAR.sub(rf'{self.CONCAT_STR}os.getenv("\1"){self.CONCAT_STR}', set_value)
+            set_value = self.CONCAT_STR.join([
+                # quote any substrings that are not lua commands
+                quote_str(x) if not x.startswith('os.') else x
```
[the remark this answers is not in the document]
boegel: fixed in https://github.com/easybuilders/easybuild-framework/commit/510e8f0942598ec8856ffb5d7fefd0c56fac235a

easybuild/tools/module_generator.py
```diff
@@ -1513,19 +1536,23 @@ def set_environment(self, key, value, relpath=False):
...
-                val = 'root'
+            set_value = self._path_join_cmd(set_value)
+            if shell_vars:
+                set_value = self.REGEX_QUOTE_SHELL_VAR.sub(r'os.getenv("\1")', set_value)
         else:
-            val = quote_str(value)
+            if shell_vars:
+                set_value = self.REGEX_SHELL_VAR.sub(rf'{self.CONCAT_STR}os.getenv("\1"){self.CONCAT_STR}', set_value)
+            set_value = self.CONCAT_STR.join([
+                # quote any substrings that are not lua commands
```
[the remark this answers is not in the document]
boegel: fixed in https://github.com/easybuilders/easybuild-framework/commit/510e8f0942598ec8856ffb5d7fefd0c56fac235a

easybuild/tools/module_generator.py
boegel: fixed in https://github.com/easybuilders/easybuild-framework/commit/510e8f0942598ec8856ffb5d7fefd0c56fac235a

easybuild/tools/module_generator.py
```diff
@@ -422,28 +430,38 @@ def det_installdir(self, modfile):
...
+        value, use_pushenv, _ = self._unpack_setenv_value(*args, **kwargs)
+        return value, use_pushenv
+
+    def _unpack_setenv_value(self, env_var_name, env_var_val):
         """
         Unpack value that specifies how to define an environment variable with specified name.
         """
-        use_pushenv = False
+        use_pushenv = self.DEFAULT_MODEXTRAVARS_PUSHENV
+        resolve_shell_vars = self.DEFAULT_MODEXTRAVARS_SHELL_VARS
```
boegel, line 447: @lexming After another pre-merge review, I realized we could play our naming game better here, see https://github.com/lexming/easybuild-framework/pull/10

boegel (changes requested):

boegel (approved):
lgtm

PR #4865, opened 2025-04-30
Fix `is_patch_for` for patch dicts 

Flamefire:
The function only handled strings and lists but not the dict patch
format.
This caused a hard error because `p % templates` doesn't work when `p`
is a dict --> `TypeError`.

Factor out function to extract name of patchspec and use it.
This avoids the 2 places where the bug was and simplifies the function.


lexming (approved):
LGTM

PR #4868, opened 2025-05-05
Update fake module for each extension installed

lexming:
The extension step generates a single fake module before the first extension is installed and uses that same fake module across the installation of all extensions.

This break the installations of easyconfigs with extensions that rely on files installed by previous extensions in the same easyconfig and that make use of search paths to find those files. Since the resolution of paths in `module_load_environment` only happens before extensions are installed, no search paths are added to the fake module as those paths/globs are empty.

We added a quick workaround in https://github.com/easybuilders/easybuild-framework/pull/4809 that skips path resolution for fake modules to avoid this issue. However, this workaround has broken installation of easyconfigs that have search paths with globs that must be expanded in their sanity checks. One example is [MATSim-15.0-Java-17.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/m/MATSim/MATSim-15.0-Java-17.eb), which has `modextrapaths = {'CLASSPATH': 'libs/*'}`.

```
== sanity checking...
  >> file 'matsim-15.0.jar' found: OK
  >> (non-empty) directory 'libs' found: OK
  >> (non-empty) directory 'examples' found: OK
  >> loading modules: MATSim/15.0-Java-17...
  >> running command 'java org.matsim.run.ReleaseInfo' ...
  >> result for command 'java org.matsim.run.ReleaseInfo': FAILED
== ... (took 3 secs)
== FAILED: Installation ended unsuccessfully: Sanity check failed: sanity check command java org.matsim.run.ReleaseInfo failed with exit code 1 (output: Error: Could not find or load main class org.matsim.run.ReleaseInfo
Caused by: java.lang.ClassNotFoundException: org.matsim.run.ReleaseInfo
) (took 9 secs)
```

This PR fixes this issue for good by generating a new fake module for each extension installed. This ensures that path/glob resolution can happen and new files from previous extensions will be accounted.

Moreover, I'm adding a new context manager to handle fake modules and a new unit test to test this particular case.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -56,6 +56,7 @@ import tempfile
 import time
 import traceback
 from concurrent.futures import ThreadPoolExecutor
+from contextlib import contextmanager
 from datetime import datetime
 from string import ascii_letters
 from textwrap import indent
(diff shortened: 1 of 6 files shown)


boegel (approved):
lgtm

Issue #4869, opened 2025-05-06
use of `--from-pr` or `--include-easyblocks-from-pr` sometimes fails because PR patch file was not downloaded


boegel:
Symptom (with `--include-easyblocks-from-pr 3705`):
```
ERROR: Failed to read /tmp/eb-lot97q4h/ebs_pr3705/3705.diff: [Errno 2] No such file or directory: '/tmp/eb-lot97q4h/ebs_pr3705/3705.diff'
```

verdurin:
Further detail from another local occurrence:
```
== 2025-05-06 14:28:43,286 filetools.py:785 DEBUG Trying to download 3705.diff from https://github.com/easybuilders/easybuild-easyblocks/pull/3705.diff to /tmp/eb-c8_n7_ph/ebs_pr3705/3705.diff
== 2025-05-06 14:28:43,287 filetools.py:788 DEBUG Using timeout of 10 seconds for initiating download
== 2025-05-06 14:28:43,287 filetools.py:2026 DEBUG Not creating existing path /tmp/eb-c8_n7_ph/ebs_pr3705
== 2025-05-06 14:28:43,470 filetools.py:865 WARNING URL https://github.com/easybuilders/easybuild-easyblocks/pull/3705.diff was not found (HTTP response code 429), not trying again
```

Thyre:
This also occurs without patch files, on different systems, with different IP addresses. 
Based on the status code, some rate limit might be hit. 

I ran into this with pure EasyConfig tests, e.g. https://github.com/easybuilders/easybuild-easyconfigs/pull/22845 and with ones having requiring an changed EasyBlock as well, e.g. https://github.com/easybuilders/easybuild-easyconfigs/pull/22842. Having both an EasyBlock and at least one EasyConfig to download triggers this more often. 

The LLVM test runs on jsc-zen3 by @Crivella also trigger this issue A LOT, e.g https://github.com/easybuilders/easybuild-easyconfigs/pull/22517

Thyre:
This might help is in finding out if we hit a rate limit:

https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit

The error described when one hits it matches what we're seeing:

https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit

Thyre:
[...]

PR #4870, opened 2025-05-07
implement exponential backoff in `download_file` + allow max. 6 attempts to download PR diff in `fetch_files_from_pr` and `fetch_files_from_commit`

boegel:
workaround/fix for #4869


easybuild/tools/filetools.py
```diff
@@ -823,6 +831,8 @@ def download_file(filename, url, path, forced=False, trace=True):
     used_urllib = std_urllib
     switch_to_requests = False
 
+    wait_time_secs = 1
```
Micket, line 834:
```suggestion
    wait_time_secs = 8
```
Since the rate limiting is so aggressive, i think we can skip the first few doublings. (or make a configurable parameter, but i don't think that's necessary since the limit is due to github, and shouldn't change much from site to site.

boegel:
I tested this with an "extreme" case, and it seems to hold up well:
```
$ eb --debug --from-pr 22698,22852,22827 --disable-cleanup-tmpdir -M --include-easyblocks-from-pr 3707
== Temporary log file in case of crash /tmp/eb-o9bahsfx/easybuild-zumur0uf.log
== Waiting for 20 seconds before trying download of https://github.com/easybuilders/easybuild-easyblocks/pull/3707.diff again...
== easyblock mathematica.py included from PR #3707
== Waiting for 20 seconds before trying download of https://github.com/easybuilders/easybuild-easyconfigs/pull/22698.diff again...
== Waiting for 40 seconds before trying download of https://github.com/easybuilders/easybuild-easyconfigs/pull/22698.diff again...
== Waiting for 20 seconds before trying download of https://github.com/easybuilders/easybuild-easyconfigs/pull/22852.diff again...
== Waiting for 40 seconds before trying download of https://github.com/easybuilders/easybuild-easyconfigs/pull/22852.diff again...
== Waiting for 20 seconds before trying download of https://github.com/easybuilders/easybuild-easyconfigs/pull/22827.diff again...
== Waiting for 40 seconds before trying download of https://github.com/easybuilders/easybuild-easyconfigs/pull/22827.diff again...

143 out of 146 required modules missing:
```

The printed messages I added for debugging purposes, but it's probably wise to actually have those, so I'll update the PR accordingly.

The first wait time is also not the intended 10 seconds, it straight goes to 2*10, need to fix that too.

sassy-crick:
Just to add: I tried `eb --copy-ec --from-pr 22824` which fails for the reasons mentioned above but with the fixes it works well. Also, I tried two different IP addresses, home and work, and I had the same issue. Thus, the IP address where you are coming from does not seem to make a difference. At home I got the full `GitHub` integration setup, and that did not make a difference too.

smoors:
works well for me too

easybuild/tools/github.py
```diff
@@ -700,17 +707,21 @@ def fetch_files_from_commit(commit, files=None, path=None, github_account=None,
         diff_url = os.path.join(GITHUB_URL, github_account, github_repo, 'commit', commit + '.diff')
         diff_fn = os.path.basename(diff_url)
         diff_filepath = os.path.join(path, diff_fn)
-        if download_file(diff_fn, diff_url, diff_filepath, forced=True, trace=False):
+        # max. 6 attempts + initial wait time of 10sec -> max. 10 * (2^6) = 640sec (~10min) before giving up on download
+        # see also https://github.com/easybuilders/easybuild-framework/issues/4869
+        max_attempts = 6
```
lexming, line 712: This should be the same value as `max_attempts` in line 554. We can quickly make this a global imported from `easybuild.tools.config`. Later on this can become another option.

lexming (dismissed):
LGTM

lexming (approved):
LGTM 2

PR #4861, opened 2025-04-26
print summary after the build in trace output

smoors:
### examples:
first build failed:
```
== Summary:
   * [F] Kaleido/0.2.1-GCCcore-13.3.0
   * [-] MultiQC/1.28-foss-2024a
```
the first build succeeded, the second failed:
```
== Summary:
   * [S] Kaleido/0.2.1-GCCcore-13.3.0
   * [F] MultiQC/1.28-foss-2024a
```
both builds succeeded:
```
== Summary:
   * [S] Kaleido/0.2.1-GCCcore-13.3.0
   * [S] MultiQC/1.28-foss-2024a
```

### Note

with `rich`, in some cases a summary line like this is printed, showing the installed modules:
```
0:01:08  1 out of 2 easyconfigs done: Kaleido/0.2.1-GCCcore-13.3.0 (OK)
```
however, that is not sufficient:
- it is not printed if all builds succeeded
- it is not printed if you don’t have rich installed (or if you specified a different output style)
- it doesn’t show the module that failed and the ones that didn’t build

PR #4872, opened 2025-05-08
add a key to the summary of the build

branfosj:
Adding a summary, as per the request at https://github.com/easybuilders/easybuild-framework/pull/4861#issuecomment-2858805702

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -122,7 +122,7 @@ def summary(ecs_with_res):
     """
     summary_fmt = "   * [{}] {}"
     success_map = {True: 'S', False: 'F', None: '-'}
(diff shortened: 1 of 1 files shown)


smoors:
as discussed in the conf call, i was planning to change it to full words so we don't need a legend:
```
* [SUCCESS] M4/1.4.19-GCCcore-12.2.0
* [FAILED]  Bison/3.8.2-GCCcore-12.2.0
* [SKIPPED] flex/2.6.4-GCCcore-12.2.0
```

branfosj:
> as discussed in the conf call, i was planning to change it to full words so we don't need a legend:

I'd missed that. I'll close this then.

Issue #4110, opened 2022-10-25
add support for installing in new namespace with bwrap


smoors:
when reinstalling software with EB on an active cluster, the software is unavailable or in a broken state during the reinstallation process, which might take a long time

here we propose a 2-step install procedure to keep the time that the software is unavailable as short as possible:

  1. install the software in a new namespace in a different location with `bwrap` (bubblewrap)
  2. copy the reinstalled software to the original location, as fast and as "atomically" as possible

here is an example how `bwrap` can be used:
```
bwrap --bind / / --bind /path/to/bwrap/software/name /path/to/software/name --bind /path/to/bwrap/module/name /path/to/module/name --dev /dev --bind /dev/log /dev/log
```
the separate binds for `software` and `modules` are necessary in case of a non-default installation directory tree.
for example, in our site it is organized as follows:
```
--software
    --soft1_name
        --version1
--modules
    --toolchain_generation1
        --all
            --soft1_name
                --version1
```

here is a way to copy the installation directory that is not fully atomic, but close.
this assumes that both the bwrap and the original directory are in the same file system:
```
rsync -av --delete-after --link-dest=/path/to/bwrap/dir /path/to/bwrap/dir /path/to/original/dir
```
with `--link-dest=`, files are not copied but hardlinks are created to the bwrap dir

boegel:
I've discussed this a bit with @smoors last week, and an idea popped up during the discussion which I want to note down here...

If we want to leverage `bwrap` to make EasyBuild into installing software in a particular path while the binaries & co are actually located in another path, then the whole EasyBuild session needs to be launched with `bwrap`, which creates a chicken-egg situation: we'll somehow need to configure EasyBuild to use bwrap, so the EasyBuild configuration needs to be parsed to know whether or not `bwrap` should be used.

To circumvent that problem, we could make a (somewhat intrusive) change to the `eb` wrapper: we could implement a small Python helper script that specifies how EasyBuild should be called. That information could then be used by the `eb` wrapper to actually start the EasyBuild session.

Without EasyBuild being configured to use `bwrap`, it would produce output like:
```
${PYTHON}" -m "${EASYBUILD_MAIN}"
```

If EasyBuild is configured to use `bwrap`, it would produce output like:

```
bwrap --bind / / --bind ... ${PYTHON}" -m "${EASYBUILD_MAIN}"
```

ocaisa:
[...]

PR #4876, opened 2025-05-11
always honor keeppreviousinstall setting

smoors:
needed for installing software in a bwrap namespace, where we bind mount the software installdir (name/version) of each software that will be installed by EB.

this in turn requires setting `keeppreviousinstall` to `True`, because removing a bind mounted directory is not possible. we currently set this in the parse hook (would be useful to have a build option for this).

EB currently hard sets `keeppreviousinstall` to `False` if `buildininstalldir` is `True`, which breaks the bwrap installation. this PR prints a big warning instead of changing `keeppreviousinstall`.

see also https://github.com/easybuilders/easybuild-framework/issues/4110

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1132,8 +1132,8 @@ class EasyBlock:
                 raise EasyBuildError("self.builddir not set, make sure gen_builddir() is called first!")
             self.log.debug("Creating the build directory %s (cleanup: %s)", self.builddir, self.cfg['cleanupoldbuild'])
         else:
-            self.log.info("Overriding 'cleanupoldinstall' (to False), 'cleanupoldbuild' (to True) "
-                          "and 'keeppreviousinstall' because we're building in the installation directory.")
+            self.log.info("Overriding 'cleanupoldinstall' (to False) and 'cleanupoldbuild' (to True) "
+                          "because we're building in the installation directory.")
             # force cleanup before installation
             if build_option('module_only') or self.cfg['module_only']:
                 self.log.debug("Disabling cleanupoldbuild because we run as module-only")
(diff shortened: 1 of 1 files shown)


boegel:
@smoors Hmm, it seems rather intrusive to change this behaviour, is this really the only way?

`keepreviousinstall` and `buildininstalldir` are fundamentally incompatible, so I'm not keen on just printing a warning here and pretending it'll all work out well

boegel:
> > @smoors Hmm, it seems rather intrusive to change this behaviour, is this really the only way?
> > `keepreviousinstall` and `buildininstalldir` are fundamentally incompatible, so I'm not keen on just printing a warning here and pretending it'll all work out well
> 
> i agree that it's an intrusive change, but we really need a way to avoid cleaning up the installdir when building in the installdir. otherwise we have to modify each easyconfig that uses buildininstalldir.
> 
> the alternative would be to add another parameter to enforce this, e.g. `keeppreviousbuild`, would you prefer that?

Maybe that makes more sense, yes. That way we're at least not changing existing behavior.

Though it still feels a bit off somehow...

smoors:
another solution would be to only remove the contents of the installdir, but not the installdir itself. with bwrap, the installdir will appear anyway, so that will be very fast.

smoors:
alternative implementation with new param `forcekeeppreviousinstall` in https://github.com/easybuilders/easybuild-framework/pull/4890

smoors:
thinking more about this, i think the best solution is to add a build option to indicate that the installdir is mounted in a new namespace. that makes it clear why we are doing this, and this option can be used by the eb wrapper to install with bwrap. it will also allow fixing the Tarball easyblock, which needs `self.cfg['install_type'] = 'merge'`, again to avoid removing the installdir.

smoors:
closing in favor of https://github.com/easybuilders/easybuild-framework/pull/4894

PR #4878, opened 2025-05-15
obtain PR/commit diff via GitHub API rather than downloading `*.diff` file via github.com

boegel:
Proper fix for #4869 (way better than the workaround implemented in #4870)

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -712,13 +712,13 @@ def fetch_files_from_commit(commit, files=None, path=None, github_account=None, 
     if not files:
         github_user = build_option('github_user')
 
-        def pr_request_fn(gh):
-            return gh.repos[github_account][github_repo].pulls[pr]
+        def commit_request_fn(gh):
+            return gh.repos[github_account][github_repo].commits[commit]
 
         # see also https://docs.github.com/en/rest/commits/commits#get-a-commit,
         # in particular part about media types
         accept_diff = {'Accept': 'application/vnd.github.diff'}
-        status, data = github_api_get_request(pr_request_fn, github_user=github_user, headers=accept_diff)
+        status, data = github_api_get_request(commit_request_fn, github_user=github_user, headers=accept_diff)
         if status == HTTP_STATUS_OK:
             # decode from bytes to text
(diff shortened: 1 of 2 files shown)


easybuild/tools/github.py
```diff
@@ -703,24 +710,25 @@ def fetch_files_from_commit(commit, files=None, path=None, github_account=None,
...
-        # see also https://github.com/easybuilders/easybuild-framework/issues/4869
-        max_attempts = DEFAULT_DOWNLOAD_MAX_ATTEMPTS
-        download_file(diff_fn, diff_url, diff_filepath, forced=True, trace=False,
-                      max_attempts=max_attempts)
-        if os.path.exists(diff_filepath):
-            diff_txt = read_file(diff_filepath)
+        github_user = build_option('github_user')
+
+        def pr_request_fn(gh):
+            return gh.repos[github_account][github_repo].pulls[pr]
```
smoors, line 716: this should be `commit_request_fn` and return the commit diff

easybuild/tools/github.py
```diff
@@ -703,24 +710,25 @@ def fetch_files_from_commit(commit, files=None, path=None, github_account=None,
...
+            diff_txt = data.decode()
             _log.debug("Diff for commit %s:\n%s", commit, diff_txt)
 
             files = det_patched_files(txt=diff_txt, omit_ab_prefix=True, github=True, filter_deleted=True)
             _log.debug("List of patched files for commit %s: %s", commit, files)
         else:
-            msg = f"Failed to download diff for commit {commit} of {github_account}/{github_repo} "
-            msg += " (after {max_attempts} attempts)"
-            raise EasyBuildError(msg, exit_code=EasyBuildExit.FAIL_GITHUB)
+            error_msg = f"Failed to download diff for {github_account}/{github_repo} PR #{pr}! (HTTP status code: {status})"
```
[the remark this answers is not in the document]
smoors: should print the commit here, not the PR

easybuild/tools/github.py
boegel: last minute refactoring gone wrong... 🤦‍♂️
fixed in fce140bcc

Issue #4451, opened 2024-02-08
eb command fails outside of framework source folder


lexming:
I installed `easybuild-framework` v5.0.x in development mode `pip install -e` and the resulting installation fails to execute the `eb` command with any combination of options (or none) with the following error:
```
Traceback (most recent call last):
  File "/home/lexming/src/EB/easybuild-framework/easybuild/tools/options.py", line 1492, in parse_options
    eb_go = EasyBuildOptions(usage=usage, description=description, prog='eb', envvar_prefix=CONFIG_ENV_VAR_PREFIX,
  File "/home/lexming/src/EB/easybuild-framework/easybuild/tools/options.py", line 249, in __init__
    super(EasyBuildOptions, self).__init__(*args, **kwargs)
  File "/home/lexming/src/EB/easybuild-framework/easybuild/base/generaloption.py", line 975, in __init__
    self.main_options()
  File "/home/lexming/src/EB/easybuild-framework/easybuild/base/generaloption.py", line 1051, in main_options
    fn()
  File "/home/lexming/src/EB/easybuild-framework/easybuild/tools/options.py", line 604, in config_options
    'choice', 'store', DEFAULT_REPOSITORY, sorted(avail_repositories().keys())),
  File "/home/lexming/src/EB/easybuild-framework/easybuild/tools/repository/repository.py", line 151, in avail_repositories
    raise EasyBuildError("avail_repositories: FileRepository missing from list of repositories")
easybuild.tools.build_log.EasyBuildError: 'avail_repositories: FileRepository missing from list of repositories'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/lexming/src/EB/easybuild-framework/easybuild/main.py", line 781, in <module>
    main_with_hooks()
  File "/home/lexming/src/EB/easybuild-framework/easybuild/main.py", line 764, in main_with_hooks
    init_session_state, eb_go, cfg_settings = prepare_main(args=args)
  File "/home/lexming/src/EB/easybuild-framework/easybuild/main.py", line 758, in prepare_main
    eb_go, cfg_settings = set_up_configuration(args=args, logfile=logfile, testing=testing)
[...]

PR #4884, opened 2025-05-16
Possible fix to allow `pip install -e .` to work seamlessly with newer pip/setuptools versions

Crivella:
Possible fix for 

- #4451

### DIFF
--- a/easybuild/tools/utilities.py
+++ b/easybuild/tools/utilities.py
@@ -34,6 +34,8 @@ import glob
 import os
 import re
 import sys
+import importlib
+
 from string import ascii_letters, digits
 
 from easybuild.base import fancylogger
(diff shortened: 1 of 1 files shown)


easybuild/tools/utilities.py
```diff
@@ -132,9 +133,15 @@ def import_available_modules(namespace):
     :param namespace: The namespace to import modules from.
     """
     modules = []
-    for path in sys.path:
 
-        cand_modpath_glob = os.path.sep.join([path] + namespace.split('.') + ['*.py'])
+    try:
+        mod = importlib.import_module(namespace)
+    except ImportError as err:
+        raise EasyBuildError("import_available_modules: Failed to import %s: %s", namespace, err)
```
boegel, line 140: This changes the semantics of `import_available_modules`, since no hard error was being raised before when the specified namespace didn't exist?

boegel (changes requested):

easybuild/tools/utilities.py
Crivella: For what i can see `import_available_modules` is always used to import all sub-modules of a specific namespace similar to doing a `from NAMESPACE import *`.

In particular for #4451 the failure would be obfuscated as *something EB requires missing*

I did not see a use in framework where this is expected to fail the import with it not being a problem.

If we really wanted to preserve the original behavior we could add a `strict` parameter to the call that default to `False` and set it to true everywhere this is used

PR #4885, opened 2025-05-16
fix download progress bar

smoors:
before:
```
Fetching files: 100% (4/3) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0:00:00 (nonexisting.patch)
ERROR: Installation of test.eb failed: "Couldn't find file nonexisting.patch anywhere, ...
```

after:
```
Fetching files: 100% (3/3) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0:00:15 (nonexisting.patch)                                                                             
ERROR: Installation of test.eb failed: "Couldn't find file nonexisting.patch anywhere, ...
```

it still feels a bit weird to me that it shows 100% even though only 2/3 of the files have been obtained, but i'm not 100% sure..
i have a fix for that too, let me know if you want me to add it?

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -811,7 +811,7 @@ class EasyBlock:
             srcpaths = source_paths()
 
         # We don't account for the checksums file in the progress bar
-        if filename != 'checksum.json':
+        if filename != 'checksums.json':
             update_progress_bar(PROGRESS_BAR_DOWNLOAD_ALL, label=filename)
 
         if alt_location is None:


ocaisa (dismissed):

ocaisa:
If the other fix is as simple might as well include it

smoors:
> If the other fix is as simple might as well include it

not as simple, but not too complex: https://github.com/easybuilders/easybuild-framework/pull/4885/commits/777a00ef4351dcfb42a3ef1664fd72c97daef878

easybuild/framework/easyblock.py
```diff
@@ -811,8 +828,8 @@ def obtain_file(self, filename, extension=False, urls=None, download_filename=No
             srcpaths = source_paths()
 
         # We don't account for the checksums file in the progress bar
-        if filename != 'checksum.json':
-            update_progress_bar(PROGRESS_BAR_DOWNLOAD_ALL, label=filename)
+        if filename != 'checksums.json':
```
boegel, line 831: This typo fix is a clear indication that we should have a constant for this, fixed in https://github.com/smoors/easybuild-framework/pull/2

boegel (changes requested):

easybuild/framework/easyblock.py
```diff
@@ -127,9 +128,26 @@
 # Directory name in which to store reproducibility files
 REPROD = 'reprod'
 
+CHECKSUMS_JSON = 'checksums.json'
+
 _log = fancylogger.getLogger('easyblock')
 
 
+def update_progress_bar_on_return(func):
```
boegel, line 136: @smoors Wouldn't it be better to make this a private decorator (rename to start with `_`), and to also mention `obtain_file` in the name, since it's quite specific to that method?

boegel (changes requested):

easybuild/framework/easyblock.py
boegel: fixed in 33e0ded14

boegel (approved):

PR #4886, opened 2025-05-20
Support options for patch command

Flamefire:
Allow to pass additional arguments to the `patch` command via the `options` dict key for patch specs

This is [documented](https://docs.easybuild.io/patch-files/#including-patches-in-an-easyconfig-file) but not implemented. I couldn't find if it ever was and got removed or never was.

Related: The key checked/used is documented is wrongly: https://github.com/easybuilders/easybuild-docs/pull/322


boegel (approved):
lgtm

PR #4891, opened 2025-05-22
Fix `--dep-graph` by using `graphviz` Python package

lorisercole:
The `--dep-graph` command uses `graphviz-python` (`import gv`) to convert a dot-graph into an image file. This library seems outdated. In pip it is now called [`graphviz`](https://pypi.org/project/graphviz/) and its API is slightly different. The [docs](https://docs.easybuild.io/installation/#optional_python_packages) already refer to it.

This PR fixes the import and the function that renders the graph image.

### DIFF
--- a/easybuild/framework/easyconfig/tools.py
+++ b/easybuild/framework/easyconfig/tools.py
@@ -80,14 +80,7 @@ try:
     # https://pypi.python.org/pypi/python-graph-dot
     import pygraph.readwrite.dot as dot
     # graphviz (used for creating dependency graph images)
-    sys.path.append('..')
-    sys.path.append('/usr/lib/graphviz/python/')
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm

boegel:
Thanks a lot for the fix @lorisercole!

A next step should probably be to get rid of the `pygraph` dependency, since that's not maintained at all.
The most recent release of [`python-graph-core`](https://pypi.org/project/python-graph-dot) and [`python-graph-dot`](https://pypi.org/project/python-graph-dot) was in July 2012 🙈), though *somehow* it still works with at least Python 3.11 (doesn't with Python 3.13 though).
Perhaps `networkx` makes more sense...

PR #4894, opened 2025-05-25
add build option new-namespace-installdir

smoors:
better alternative solution for https://github.com/easybuilders/easybuild-framework/pull/4876

see also https://github.com/easybuilders/easybuild-framework/issues/4110


smoors:
solution using `empty_dir` https://github.com/easybuilders/easybuild-framework/pull/4932

PR #4904, opened 2025-05-28
prefer using `importlib.metadata` over the deprecated `pkg_resources` in `det_pypkg_version` + switch container tests to rockylinux 8.10

boegel:
fixes #4903

PR #4906, opened 2025-05-28
switch container tests to rockylinux 8.10

branfosj:
The changes in https://github.com/easybuilders/easybuild-easyconfigs/pull/22798 mean that M4/1.4.19 no longer builds on Centos 7 but that is used in the container tests. This PR switches those tests to use Rocky 8.10 instead.


boegel:
@branfosj You can actually keep this open, when we merge #4904 this will be marked as merged too (and then we won't forgot to include it in the release notes)

boegel:
Well, seems like I got that wrong.

Requires changes got merged via #4904, so closing this after all...

PR #4912, opened 2025-06-02
add EB configuration option to ignore unversioned (0.0.0) python packages

jfgrimm:
useful when your OS has some packages installed that report a version of `0.0.0` (e.g. `kiwisolver` or `mpmath` in debian 12/ubuntu 24.04)

- [] companion PR: https://github.com/easybuilders/easybuild-easyblocks/pull/3753

needs tests adding


boegel (approved):
lgtm

PR #4914, opened 2025-06-05
WIP: Implementing a new LLVM based toolchain

Crivella:
## Compilers

- New `llvm.py` compiler taken from the original `clang.py` and unifying it with `flang`
  - Now allows defining a list of unsupported `flang` compiler options based on the passed LLVM version which will be excluded, from the compiler flags

## Toolchains

- `LLVMtc`: Similar to `GCC` uses bot `GCCcore` and `SYSTEM` as a subtoolchain
- `Lfbf` equivalent of `Gfbf`
- `Lompi` equivalent of `Gompi`
- `LFoss` equivalent of `Foss`
  - Does not have the `OpenBLAS` child class as in Foss it is used only for toolchains older than `2021.0`

### DIFF
--- a/easybuild/toolchains/compiler/llvm.py
+++ b/easybuild/toolchains/compiler/llvm.py
@@ -26,7 +26,7 @@
 # along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
 ##
 """
-Support for Flang as toolchain compiler.
+EasyBuild support for Clang + Flang compiler toolchain. 
 
 Authors:
 
@@ -34,27 +34,54 @@ Authors:
 * Davide Grassano (CECAM EPFL)
 """
 
+from easybuild.tools import LooseVersion
 import easybuild.tools.systemtools as systemtools
 from easybuild.tools.toolchain.compiler import Compiler, DEFAULT_OPT_LEVEL
+from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME
+
+TC_CONSTANT_LLVM = "LLVM"
 
 
-TC_CONSTANT_FLANG = "Flang"
+class LLVM(Compiler):
+    """Compiler toolchain with Clang and GFortran compilers."""
+    # NAME = 'LLVMcore'
+    # COMPILER_MODULE_NAME = [NAME]
+    COMPILER_FAMILY = TC_CONSTANT_LLVM
+    SUBTOOLCHAIN = SYSTEM_TOOLCHAIN_NAME
 
+    # List of flags that are supported by Clang but not yet by Flang and should be filtered out
+    # The element of the list are tuples with the following structure:
+    # (min_version, max_version, [list of flags])
+    # Where min_version and max_version are strings representing a left-inclusive and right-exclusive version range,
+    # [min_version, max_version) respectively.
+    # This is used to specify general `clang`-accepted flags and remove them from `flang` compiler flags if
+    # not supported for a particular version of LLVM
+    FLANG_UNSUPPORTED_VARS = [
+        ('19', '21', [
+            '-fslp-vectorize',
+            '-fvectorize',
+            '-fno-vectorize',
+            '-fno-unsafe-math-optimizations',
+        ])
+    ]
 
-class Flang(Compiler):
-    """Clang compiler class"""
-
-    COMPILER_MODULE_NAME = ['Flang']
-    COMPILER_FAMILY = TC_CONSTANT_FLANG
-
-    # Don't set COMPILER_FAMILY in this class because Clang does not have
-    # Fortran support, and thus it is not a complete compiler as far as
-    # EasyBuild is concerned.
+    FORTRAN_FLAGS = ['FCFLAGS', 'FFLAGS', 'F90FLAGS']
 
     COMPILER_UNIQUE_OPTS = {
         'loop-vectorize': (False, "Loop vectorization"),
         'basic-block-vectorize': (False, "Basic block vectorization"),
+
+        # https://github.com/madler/zlib/issues/856
+        'lld_undefined_version': (True, "-Wl,--undefined-version - Allow unused version in version script"),
+        'no_unused_args': (
+            True,
+            (
+                "-Wno-unused-command-line-argument - Avoid some failures in CMake correctly recognizing "
+                "feature due to linker warnings"
+            )
+        ),
     }
+
     COMPILER_UNIQUE_OPTION_MAP = {
         'unroll': '-funroll-loops',
         'loop-vectorize': ['-fvectorize'],
@@ -89,7 +116,22 @@ class Flang(Compiler):
         'veryloose': ['-ffast-math'],
         'vectorize': {False: '-fno-vectorize', True: '-fvectorize'},
         DEFAULT_OPT_LEVEL: ['-O2'],
+
+        'lld_undefined_version': ['-Wl,--undefined-version',],
+        'no_unused_args': ['-Wno-unused-command-line-argument'],
     }
+
+    COMPILER_OPTIONS = [
+        'lld_undefined_version',
+    ]
+
+    # Options only available for Clang compiler
+    COMPILER_C_OPTIONS = [
+        'no_unused_args',
+    ]
+
+    # Options only available for Flang compiler
+    COMPILER_F_OPTIONS = []
 
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
@@ -105,17 +147,35 @@ class Flang(Compiler):
         (systemtools.X86_64, systemtools.INTEL): '-march=x86-64 -mtune=generic',
     }
 
-    # COMPILER_CC = 'clang'
-    # COMPILER_CXX = 'clang++'
-    # COMPILER_C_UNIQUE_OPTIONS = []
+    COMPILER_CC = 'clang'
+    COMPILER_CXX = 'clang++'
+    COMPILER_C_UNIQUE_OPTIONS = []
 
     COMPILER_F77 = 'flang'
     COMPILER_F90 = 'flang'
     COMPILER_FC = 'flang'
-    COMPILER_F_UNIQUE_OPTIONS = [
-        # 'lld_undefined_version'
-    ]
     COMPILER_F_UNIQUE_OPTIONS = []
 
     LIB_MULTITHREAD = ['pthread']
     LIB_MATH = ['m']
+
+    def _set_compiler_flags(self):
+        super()._set_compiler_flags()
+
+        unsupported_fortran_flags = None
+        for v_min, v_max, flags in self.FLANG_UNSUPPORTED_VARS:
+            if LooseVersion(self.version) >= LooseVersion(v_min) and LooseVersion(self.version) < LooseVersion(v_max):
+                unsupported_fortran_flags = flags
+                break
+        else:
+            self.log.debug("No unsupported flags found for LLVM version %s", self.version)
+
+        if unsupported_fortran_flags is not None:
+            self.log.debug(
+                f"Ensuring usupported Fortran flags `{unsupported_fortran_flags}` are removed from variables"
+            )
+            for key, lst in self.variables.items():
+                if key not in self.FORTRAN_FLAGS:
+                    continue
+                for item in lst:
+                    item.try_remove(unsupported_fortran_flags)
--- a/easybuild/toolchains/lfoss.py
+++ b/easybuild/toolchains/lfoss.py
@@ -35,7 +35,6 @@ from easybuild.toolchains.lfbf import Lfbf
 from easybuild.toolchains.lolf import Lolf
 from easybuild.toolchains.fft.fftw import Fftw
 from easybuild.toolchains.linalg.flexiblas import FlexiBLAS
-from easybuild.toolchains.linalg.openblas import OpenBLAS
 from easybuild.toolchains.linalg.scalapack import ScaLAPACK
 from easybuild.tools import LooseVersion
 
--- a/easybuild/toolchains/lfoss.py
+++ b/easybuild/toolchains/lfoss.py
@@ -1,0 +1,100 @@
+##
+# Copyright 2013-2025 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for foss compiler toolchain (includes GCC, OpenMPI, OpenBLAS, LAPACK, ScaLAPACK and FFTW).
+
+Authors:
+
+* Kenneth Hoste (Ghent University)
+* Davide Grassano (CECAM EPFL)
+"""
+from easybuild.toolchains.lompi import Lompi
+from easybuild.toolchains.lfbf import Lfbf
+from easybuild.toolchains.lolf import Lolf
+from easybuild.toolchains.fft.fftw import Fftw
+from easybuild.toolchains.linalg.flexiblas import FlexiBLAS
+from easybuild.toolchains.linalg.openblas import OpenBLAS
+from easybuild.toolchains.linalg.scalapack import ScaLAPACK
+from easybuild.tools import LooseVersion
+
+
+class LFoss(Lompi, OpenBLAS, FlexiBLAS, ScaLAPACK, Fftw):
+    """Compiler toolchain with GCC, OpenMPI, OpenBLAS, ScaLAPACK and FFTW."""
+    NAME = 'lfoss'
+    SUBTOOLCHAIN = [
+        Lompi.NAME, 
+        Lolf.NAME,
+        Lfbf.NAME
+    ]
+
+    def __init__(self, *args, **kwargs):
+        """Toolchain constructor."""
+        super(LFoss, self).__init__(*args, **kwargs)
+
+        # need to transform a version like '2018b' with something that is safe to compare with '2019'
+        # comparing subversions that include letters causes TypeErrors in Python 3
+        # 'a' is assumed to be equivalent with '.01' (January), and 'b' with '.07' (June) (good enough for this purpose)
+        version = self.version.replace('a', '.01').replace('b', '.07')
+
+        self.looseversion = LooseVersion(version)
+
+        constants = ('BLAS_MODULE_NAME', 'BLAS_LIB', 'BLAS_LIB_MT', 'BLAS_FAMILY',
+                     'LAPACK_MODULE_NAME', 'LAPACK_IS_BLAS', 'LAPACK_FAMILY')
+
+        if self.looseversion > LooseVersion('2021.0'):
+            for constant in constants:
+                setattr(self, constant, getattr(FlexiBLAS, constant))
+        else:
+            for constant in constants:
+                setattr(self, constant, getattr(OpenBLAS, constant))
+
+    def banned_linked_shared_libs(self):
+        """
+        List of shared libraries (names, file names, paths) which are
+        not allowed to be linked in any installed binary/library.
+        """
+        res = []
+        res.extend(Lompi.banned_linked_shared_libs(self))
+
+        if self.looseversion >= LooseVersion('2021.0'):
+            res.extend(FlexiBLAS.banned_linked_shared_libs(self))
+        else:
+            res.extend(OpenBLAS.banned_linked_shared_libs(self))
+
+        res.extend(ScaLAPACK.banned_linked_shared_libs(self))
+        res.extend(Fftw.banned_linked_shared_libs(self))
+
+        return res
+
+    def is_deprecated(self):
+        """Return whether or not this toolchain is deprecated."""
+
+        # foss toolchains older than foss/2019a are deprecated since EasyBuild v4.5.0;
+        if self.looseversion < LooseVersion('2019'):
+            deprecated = True
+        else:
+            deprecated = False
+
+        return deprecated
--- a/easybuild/toolchains/lfoss.py
+++ b/easybuild/toolchains/lfoss.py
@@ -40,8 +40,8 @@ from easybuild.toolchains.linalg.scalapack import ScaLAPACK
 from easybuild.tools import LooseVersion
 
 
-class LFoss(Lompi, OpenBLAS, FlexiBLAS, ScaLAPACK, Fftw):
-    """Compiler toolchain with GCC, OpenMPI, OpenBLAS, ScaLAPACK and FFTW."""
+class LFoss(Lompi, FlexiBLAS, ScaLAPACK, Fftw):
+    """Compiler toolchain with GCC, OpenMPI, FlexiBLAS, ScaLAPACK and FFTW."""
     NAME = 'lfoss'
     SUBTOOLCHAIN = [
         Lompi.NAME, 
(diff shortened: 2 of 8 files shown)


Crivella:
https://github.com/easybuilders/easybuild-framework/pull/4914/commits/d6fe4f9070e7fb3c68f42dafae26b7fe6024fdb5

Apparently some tests were failing because `xlompi` is an actual toolchain and the test was using `'x'` as an prefix to check that a toolchain with wrong name should not be found

easybuild/toolchains/compiler/llvm.py
```diff
@@ -0,0 +1,191 @@
...
+from easybuild.tools import LooseVersion
+import easybuild.tools.systemtools as systemtools
+from easybuild.tools.toolchain.compiler import Compiler, DEFAULT_OPT_LEVEL
+from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME
+
+TC_CONSTANT_LLVM = "LLVM"
+
+
+class LLVM(Compiler):
+    """Compiler toolchain with Clang and GFortran compilers."""
```
hattom: "GFortran" was copy/pasted from somewhere else?
This one should be Clang + flang, no?

easybuild/toolchains/compiler/llvm.py
```diff
@@ -0,0 +1,191 @@
...
+        for v_min, v_max, flags in self.FLANG_UNSUPPORTED_VARS:
+            if LooseVersion(self.version) >= LooseVersion(v_min) and LooseVersion(self.version) < LooseVersion(v_max):
+                unsupported_fortran_flags = flags
+                break
+        else:
+            self.log.debug("No unsupported flags found for LLVM version %s", self.version)
+
+        if unsupported_fortran_flags is not None:
+            self.log.debug(
+                f"Ensuring usupported Fortran flags `{unsupported_fortran_flags}` are removed from variables"
```
hattom: "unsupported"

easybuild/toolchains/lfoss.py
```diff
@@ -0,0 +1,90 @@
...
+from easybuild.toolchains.lfbf import Lfbf
+from easybuild.toolchains.lolf import Lolf
+from easybuild.toolchains.fft.fftw import Fftw
+from easybuild.toolchains.linalg.flexiblas import FlexiBLAS
+from easybuild.toolchains.linalg.scalapack import ScaLAPACK
+from easybuild.tools import LooseVersion
+
+
+class LFoss(Lompi, FlexiBLAS, ScaLAPACK, Fftw):
+    """Compiler toolchain with GCC, OpenMPI, FlexiBLAS, ScaLAPACK and FFTW."""
```
hattom: "GCC" -> "LLVM", right?

easybuild/toolchains/lompi.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+from easybuild.tools import LooseVersion
+import re
+
+from easybuild.toolchains.llvm import LLVMtc
+from easybuild.toolchains.mpi.openmpi import OpenMPI
+
+
+class Lompi(LLVMtc, OpenMPI):
+    """Compiler toolchain with GCC and OpenMPI."""
```
hattom: GCC -> LLVM?

easybuild/toolchains/compiler/llvm.py
Crivella: Yeah i think they were leftovers when i restarted this files from the GCC ones thanks for catching these errors, fixed in [a9080ed](https://github.com/easybuilders/easybuild-framework/pull/4914/commits/a9080ed9bf31c69147dfe7c095497b741974a30e)

Thyre:
I honestly don't really like the name `LLVMtc`. This deviates from the naming we have for GCC & Intel, and even the proposed name for the NVHPC split. In addition, people might be confused by the abbreviation, if they're not familiar with EasyBuild.

Should we go with `llvm-compilers`, similar to what we have with `intel-compilers` and `nvidia-compilers`? We could use this same naming scheme for other toolchains going forward (`rocm-compilers`, `aocc-compilers`, ...).

easybuild/toolchains/compiler/llvm.py
```diff
@@ -0,0 +1,191 @@
...
+        DEFAULT_OPT_LEVEL: ['-O2'],
+
+        'lld_undefined_version': ['-Wl,--undefined-version'],
+        'no_unused_args': ['-Wno-unused-command-line-argument'],
+        'no_int_conversion_error': ['-Wno-error=int-conversion'],
+    }
+
+    COMPILER_OPTIONS = [
+        'lld_undefined_version',
+    ]
```
Thyre: This overrides the global options, hence options like `openmp` will not work anymore.

easybuild/toolchains/compiler/llvm.py
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    COMPILER_CC = 'clang'
+    COMPILER_CXX = 'clang++'
+    COMPILER_C_UNIQUE_OPTIONS = []
+
+    COMPILER_F77 = 'flang'
+    COMPILER_F90 = 'flang'
+    COMPILER_FC = 'flang'
+    COMPILER_F_UNIQUE_OPTIONS = []
```
Thyre:
```suggestion
    COMPILER_OPTIONS = Compiler.COMPILER_OPTIONS
    COMPILER_OPTIONS += ['lld_undefined_version']

    # Options only available for Clang compiler
    COMPILER_C_UNIQUE_OPTIONS = Compiler.COMPILER_C_UNIQUE_OPTIONS
    COMPILER_C_UNIQUE_OPTIONS += [
        'no_unused_args',
        'no_int_conversion_error'
    ]

    # Options only available for Flang compiler
    COMPILER_F_UNIQUE_OPTIONS = Compiler.COMPILER_F_UNIQUE_OPTIONS

    # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
    COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
        (systemtools.POWER, systemtools.POWER): '-mcpu=native',  # no support for march=native on POWER
        (systemtools.POWER, systemtools.POWER_LE): '-mcpu=native',  # no support for march=native on POWER
        (systemtools.X86_64, systemtools.AMD): '-march=native',
        (systemtools.X86_64, systemtools.INTEL): '-march=native',
    }
    # used with --optarch=GENERIC
    COMPILER_GENERIC_OPTION = {
        (systemtools.RISCV64, systemtools.RISCV): '-march=rv64gc -mabi=lp64d',  # default for -mabi is system-dependent
        (systemtools.X86_64, systemtools.AMD): '-march=x86-64 -mtune=generic',
        (systemtools.X86_64, systemtools.INTEL): '-march=x86-64 -mtune=generic',
    }

    COMPILER_CC = 'clang'
    COMPILER_CXX = 'clang++'

    COMPILER_F77 = 'flang'
    COMPILER_F90 = 'flang'
    COMPILER_FC = 'flang'
```

I replaced it like this. OpenMP worked just fine then, but I saw all the arguments added by LLVM twice while building HPCG with `lfoss`. I don't know why and haven't checked if we encounter similar issues with e.g. `foss`.

Thyre (changes requested):

easybuild/toolchains/compiler/llvm.py
Thyre: See https://github.com/Crivella/easybuild-framework/pull/1 for a proper fix (and adding MPICH toolchains) :smile:

Crivella:
Changes in https://github.com/easybuilders/easybuild-framework/pull/4914/commits/31a27c6ad2bb08437d95808fc0155b4196cafe4a tested through

https://github.com/easybuilders/easybuild-easyconfigs/pull/23051#issuecomment-3512320478

easybuild/scripts/rpath_args.py
```diff
@@ -116,6 +116,11 @@ def add_rpath_flag(lib_path, rpath_filter, rpath_lib_paths, cmd_args_rpath, ldfl
         add_rpath_args = False
         cmd_args.append(arg)
 
+    # preprocess only mode, no linking is done
+    elif arg == '-E':
```
boegel: We need to be more careful here, since `-E` is also a valid option in the `ld` command.

Only when we're wrapping a *compiler* command does `-E` mean only pre-processing...

easybuild/scripts/rpath_args.py
```diff
@@ -80,7 +80,7 @@ def add_rpath_flag(lib_path, rpath_filter, rpath_lib_paths, cmd_args_rpath, ldfl
 args = sys.argv[4:]
 
 # determine whether or not to use -Wl to pass options to the linker based on name of command
-if cmd in ['ld', 'ld.gold', 'ld.bfd']:
+if cmd in ['ld', 'ld.gold', 'ld.bfd', 'lld', 'ld.lld', 'ld64.lld']:
```
boegel: Let's make a constant out of this list

```python
LINKER_COMMANDS = ('ld', 'ld.gold', 'ld.bfd', 'lld', 'ld.lld', 'ld64.lld')
...
if cmd in LINKER_COMMANDS:
```

easybuild/tools/toolchain/toolchain.py
```diff
@@ -830,6 +830,16 @@ def compilers(self):
 
         return (c_comps, fortran_comps)
 
+    def linkers(self):
+        """Return list of relevant linkers for this toolchain"""
+
+        if self.is_system_toolchain():
+            linkers = ['ld', 'ld.gold', 'ld.bfd']
+        else:
+            linkers = self.LINKERS or []
```
boegel: Let's take a copy here, we don't want to pass around a constant as a reference...

```suggestion
            linkers = list(self.LINKERS or [])

File: easybuild/toolchains/compiler/llvm_compilers.py [hunk 10]
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    COMPILER_CC = 'clang'
+    COMPILER_CXX = 'clang++'
+
+    COMPILER_F77 = 'flang'
+    COMPILER_F90 = 'flang'
+    COMPILER_FC = 'flang'
+
+    LINKERS = ['lld', 'ld.lld', 'ld64.lld']
```
boegel: let's make this new constant an immutable value:
```suggestion
    LINKERS = ('lld', 'ld.lld', 'ld64.lld')
```

That should also be done for other constants, but that needs a broader effort across existing supported compilers (that's out of scope for this PR)

boegel (changes requested):

easybuild/scripts/rpath_args.py
boegel: fixed in 9456d63

easybuild/scripts/rpath_args.py
boegel: fixed in 9456d63

easybuild/toolchains/compiler/llvm_compilers.py
boegel: fixed in 164df4e

easybuild/tools/toolchain/toolchain.py
boegel: fixed in 164df4e

easybuild/toolchains/compiler/llvm_compilers.py
```diff
@@ -0,0 +1,193 @@
...
+    ]
+
+    FORTRAN_FLAGS = ['FCFLAGS', 'FFLAGS', 'F90FLAGS']
+
+    COMPILER_UNIQUE_OPTS = {
+        'loop-vectorize': (False, "Loop vectorization"),
+        'basic-block-vectorize': (False, "Basic block vectorization"),
+
+        # https://github.com/madler/zlib/issues/856
+        'lld_undefined_version': (True, "-Wl,--undefined-version - Allow unused version in version script"),
```
boegel: This is only supported for sufficiently recent LLVM versions?
See failure in https://github.com/easybuilders/easybuild-easyconfigs/pull/24515#issuecomment-3567809010

Maybe we should focus on sufficiently recent versions, and disallow using this toolchain compiler for (too) old versions?

easybuild/toolchains/compiler/llvm_compilers.py
```diff
@@ -0,0 +1,193 @@
...
+            "-Wno-error=int-conversion - Avoid some failures that are normally ignored by GCC"
+        ),
+    }
+
+    COMPILER_UNIQUE_OPTION_MAP = {
+        'unroll': '-funroll-loops',
+        'loop-vectorize': ['-fvectorize'],
+        'basic-block-vectorize': ['-fslp-vectorize'],
+        'optarch': '',
+        # 'optarch': '-march=native',
```
boegel: @Crivella Should we keep it like this?

We won't to enable targeting of the host CPU...

Or is this handled via `COMPILER_OPTIMAL_ARCHITECTURE_OPTION` below?
If so, we should add a comment of the line that maps `optarch` to `''` (and remove the commented out `-march=native` line here)

easybuild/toolchains/lmpich.py
```diff
@@ -0,0 +1,41 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for gmpich compiler toolchain (includes GCC and MPICH).
```
boegel:
```suggestion
EasyBuild support for lmpich compiler toolchain (includes LLVM compilers and MPICH).

File: easybuild/toolchains/lolf.py [hunk 14]
```diff
@@ -0,0 +1,44 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for golf compiler toolchain (includes GCC, OpenBLAS, LAPACK, and FFTW).
```
boegel:
```suggestion
EasyBuild support for lolf compiler toolchain (includes LLVM compilers, OpenBLAS, LAPACK, and FFTW).

File: easybuild/toolchains/lolf.py [hunk 15]
```diff
@@ -0,0 +1,44 @@
...
+* Bart Oldeman (McGill University, Calcul Quebec, Compute Canada)
+* Davide Grassano (CECAM EPFL)
+"""
+
+from easybuild.toolchains.llvm_compilers import LLVMCompilersToolchain
+from easybuild.toolchains.fft.fftw import Fftw
+from easybuild.toolchains.linalg.openblas import OpenBLAS
+
+
+class Lolf(LLVMCompilersToolchain, OpenBLAS, Fftw):
```
boegel: Do we have a use case of a compiler toolchain that includes `OpenBLAS` directly, rather than using `FlexiBLAS`?

If not, we should remove this, we don't want to encourage direct usage of OpenBLAS?

easybuild/toolchains/compiler/llvm_compilers.py
Thyre: `--undefined-version` is not a compiler option, but a linker one (both for `ld` and `lld`).
If this option ends up in the compiler (because e.g. `-Wl` is stripped), this will cause trouble.

Do we really need this option at all @Crivella ?

easybuild/toolchains/compiler/llvm_compilers.py
Thyre: I think it is. I've seen `-march=native` in several builds, so this seems to be enabled. Agreed that we should add a proper comment for this.

easybuild/toolchains/lolf.py
Thyre: I'm guessing this is just part of the toolchains being directly ported from GCC. But I agree that we most likely do not want to offer such a toolchain.

easybuild/scripts/rpath_args.py
boegel: This is important for `Vim-9.1.1788-GCCcore-13.2.0.eb` for example, where `gcc ... -Wl,-E ... -o vim` is used.

Without the fix, RPATH sanity check fails with "`No '(RPATH)' found in 'readelf -d' output for /software/Vim/9.1.1788-GCCcore-13.2.0/bin/vim`".

easybuild/toolchains/lolf.py
boegel: then I'd say we remove this file entirely.

If a use case does pop up, we can always add it back...

bartoldeman: yes that's `golf` which was a thing I introduced years ago before we stated using FlexiBLAS and `gfbf`.

easybuild/toolchains/compiler/llvm_compilers.py
Crivella: The problem here is that `libtool` is filtering out the `-Wl,` in non-linkstep calls.

It is the reason i had added https://github.com/easybuilders/easybuild-easyconfigs/blob/b63c447dc936791aed0284dc71298a5f21c86ab4/easybuild/easyconfigs/e/ELPA/ELPA-2025.06.001-lfoss-2025b.eb#L18-L20 for `QE-7.5-lfoss-2025b`
Since i had not hit that error in the other PR for `2023b` i thought the `libtool` version included with that ELPA was not doing the filtering, but we could disable it also in that PR

> Do we really need this option at all @Crivella ?

The reason I had originally added it was for compiling `zlib` with the LLVM-only version of  the toolchain to avoid hitting the error described in https://github.com/madler/zlib/issues/856 (and since it was the default behavior when compiling with GCC to avoid hitting this error again).
I guess we could try to `remove it`/`set it to default to False` and see what happens

Thyre: I think we should keep `--undefined-version` as default to be consistent with what `ld.lld` does by default. If that causes issues, we should be able to disable this with a toolchain option.

easybuild/toolchains/compiler/llvm_compilers.py
Crivella: I think this is handled by `COMPILER_OPTIMAL_ARCHITECTURE_OPTION` which was not originally in the `Clang` TC file.
If i recall correctly i had disabled this to avoid having `-march=native` on options not present in `COMPILER_OPTIMAL_ARCHITECTURE_OPTION`, but indeed we can just remove the line

easybuild/toolchains/compiler/llvm_compilers.py
Crivella: From the testing i did i was able to rebuild everything also without the `--undefined-version` being passed to `lld`.
TBH the case in `zlib` was most likely a bug that was fixed.

I think this might be a niche enough error that we might want to only add this where needed to the `cflags` instead of everywhere?

Thyre: Fine with me.

boegel: Yes, makes sense, let's only do this when it's really needed

easybuild/toolchains/compiler/llvm_compilers.py
```diff
@@ -0,0 +1,191 @@
...
+    # Options only available for Flang compiler
+    COMPILER_F_UNIQUE_OPTIONS = Compiler.COMPILER_F_UNIQUE_OPTIONS
+
+    # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
+    COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
+        **(Compiler.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or {}),
+        (systemtools.POWER, systemtools.POWER): '-mcpu=native',  # no support for march=native on POWER
+        (systemtools.POWER, systemtools.POWER_LE): '-mcpu=native',  # no support for march=native on POWER
+        (systemtools.X86_64, systemtools.AMD): '-march=native',
+        (systemtools.X86_64, systemtools.INTEL): '-march=native',
```
boegel, line 150: @Crivella No entry for `aarch64` here now, which leads to:
```
== FAILED: Installation ended unsuccessfully: Don't know how to set optarch flags for AArch64/ARM! Use --optarch='<flags>' to override
```

Seems like that's cause by the change made in 02755c9d21324af440bf4c57e9757d1ac643f505 ?

Also `COMPILER_GENERIC_OPTION` should have an entry for `aarch64`

Thyre: Same probably applies to `COMPILER_GENERIC_OPTION` below, right?

Thyre, line 150:
```suggestion
    COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
        **(Compiler.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or {}),
        (systemtools.AARCH64, systemtools.ARM): '-march=native', 
        (systemtools.POWER, systemtools.POWER): '-mcpu=native',  # no support for march=native on POWER
        (systemtools.POWER, systemtools.POWER_LE): '-mcpu=native',  # no support for march=native on POWER
        (systemtools.X86_64, systemtools.AMD): '-march=native',
        (systemtools.X86_64, systemtools.INTEL): '-march=native',
```

easybuild/toolchains/compiler/llvm_compilers.py
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    # used with --optarch=GENERIC
+    COMPILER_GENERIC_OPTION = {
+        **(Compiler.COMPILER_GENERIC_OPTION or {}),
+        (systemtools.RISCV64, systemtools.RISCV): '-march=rv64gc -mabi=lp64d',  # default for -mabi is system-dependent
+        (systemtools.X86_64, systemtools.AMD): '-march=x86-64 -mtune=generic',
+        (systemtools.X86_64, systemtools.INTEL): '-march=x86-64 -mtune=generic',
+    }
```
Thyre, line 160:
```suggestion
    COMPILER_GENERIC_OPTION = {
        **(Compiler.COMPILER_GENERIC_OPTION or {}),
        (systemtools.AARCH64, systemtools.ARM): '-mcpu=generic -mtune=generic',  # march=generic does not exist for aarch64
        (systemtools.RISCV64, systemtools.RISCV): '-march=rv64gc -mabi=lp64d',  # default for -mabi is system-dependent
        (systemtools.X86_64, systemtools.AMD): '-march=x86-64 -mtune=generic',
        (systemtools.X86_64, systemtools.INTEL): '-march=x86-64 -mtune=generic',
    }
```

easybuild/toolchains/compiler/llvm_compilers.py
Thyre: @boegel, @Crivella See the suggestions below. Looks like `-march=generic` does not exist (at least on Neoverse V2). The suggestions were tested on LLVM 20. Maybe older LLVM versions are missing some flags, but I wouldn't expect anyone to use a toolchain based on e.g. LLVM 18.

easybuild/toolchains/compiler/llvm_compilers.py
Thyre: There's also no generic `aarch64` option for `march`, so we would need to select a fixed architecture for generic. I think using `mcpu=generic` & `mtune=generic` should be sufficient...

easybuild/toolchains/compiler/llvm_compilers.py
```diff
@@ -152,6 +153,7 @@ class LLVMCompilers(Compiler):
     # used with --optarch=GENERIC
     COMPILER_GENERIC_OPTION = {
         **(Compiler.COMPILER_GENERIC_OPTION or {}),
+        (systemtools.AARCH64, systemtools.ARM): '-march=aarch64 -mtune=generic',
```
Thyre: @boegel, this will not work:

```
[reuter1@jrc0901 ~]$ flang -march=aarch64 test.f90
flang-20: error: unsupported argument 'aarch64' to option '-march='
[reuter1@jrc0901 ~]$ clang -march=aarch64 test.c
clang: error: unsupported argument 'aarch64' to option '-march='
```

See https://github.com/easybuilders/easybuild-framework/pull/4914#discussion_r2572929043 for a suggestion that should work

Thyre: Checking the GCC documentation, there are "generic" targets one could specify. Not sure if those exist in LLVM as well: https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html#index-march

**Edit:** They do, so we could also use `-march=armv8-a` for the most generic target. I think `-mcpu=generic` should be preferred though, given that `-mcpu` was preferred over `-march` for some time (https://github.com/llvm/llvm-project/issues/98443)

boegel: fixed in d95636f57

easybuild/toolchains/compiler/llvm_compilers.py
boegel: fixed in https://github.com/easybuilders/easybuild-framework/commit/d95636f5747e0eb6cc1275197262882b7a71f978

easybuild/toolchains/compiler/llvm_compilers.py
boegel: fixed in f1e1601

boegel: fixed in f1e1601 + d95636f

PR #4915, opened 2025-06-05
Change value of architectures constants RISCV32 and RISCV64

julianmorillo:
This constant value is returned by [get_cpu_architecture](https://github.com/easybuilders/easybuild-framework/blob/87b3d9d959b5f3ec739f3db4a350982b531c58fb/easybuild/tools/systemtools.py#L341), which is called from [get_arch_prefix](https://github.com/easybuilders/easybuild-easyblocks/blob/ac2b2f9a5069927a31079219aee732e169e0b05e/easybuild/easyblocks/l/llvm.py#L154), that returns it directly (lower case) to [_get_gcc_prefix](https://github.com/easybuilders/easybuild-easyblocks/blob/ac2b2f9a5069927a31079219aee732e169e0b05e/easybuild/easyblocks/l/llvm.py#L529), causing the following error when trying to build LLVM in RISCV:
```
== FAILED: Installation ended unsuccessfully: Can't find GCC version 13.2.0 for architecture risc-v-64 in
/cvmfs/riscv.eessi.io/versions/20240402/software/linux/riscv64/generic/software/GCCcore/13.2.0/lib/gcc/risc-v-64-*/13.2.0 (took 1 min 25 secs)
== Results of the build can be found in the log file(s) /tmp/eb-0kc4ln6r/easybuild-LLVM-20.1.5-20250605.161517.VhpnA.log

ERROR: Installation of LLVM-20.1.5-GCCcore-13.2.0.eb failed: "Can't find GCC version 13.2.0 for architecture risc-v-64 in /cvmfs/riscv.eessi.io/versions/20240402/software/linux/riscv64/generic/software/GCCcore/13.2.0/lib/gcc/risc-v-64-*/13.2.0"
```
The correct path in my case would be` /cvmfs/riscv.eessi.io/versions/20240402/software/linux/riscv64/generic/software/GCCcore/13.2.0/lib/gcc/riscv64-*/13.2.0`


easybuild/tools/systemtools.py
```diff
@@ -108,7 +108,7 @@
 POWER = 'POWER'
 X86_64 = 'x86_64'
 RISCV32 = 'RISC-V-32'
-RISCV64 = 'RISC-V-64'
+RISCV64 = 'RISCV64'
```
boegel, line 111: I don't mind changes this, but we should make the same change for the `RISCV32` constant, to keep it consistent?

boegel (changes requested):

easybuild/tools/systemtools.py
julianmorillo: Sure, done.

boegel (approved):
lgtm

PR #4916, opened 2025-06-10
add `sanitycheck_dependencies` easyconfig parameter

Crivella:
Allow having a separate set of dependencies loaded only at sanity check.

NOTE:

Thinking about this more, this might not be as straight-forward, as I think the idea of sanity-checking is to test whether all commands work properly with only the build module loaded.

Introducing extra modules could make a check pass that supposedly should not...

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -4208,6 +4208,15 @@ class EasyBlock:
                 self.sanity_check_fail_msgs.append("loading fake module failed: %s" % err)
                 self.log.warning("Sanity check: %s" % self.sanity_check_fail_msgs[-1])
 
+        # This should happen after `self.load_module` or `load_fake_module` as this calls purge the
+        # current modules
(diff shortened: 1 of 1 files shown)


boegel:
Using this would indeed sort of defeat the purpose of the sanity check, but I can imagine there are cases where it would make sense, for example if you want to use an extra command in a sanity check command that is not available by default, but which is not required to use the software in the normal way.

@Crivella What use case did you have for this exactly?

Crivella:
This starting idea for this was https://github.com/easybuilders/easybuild-easyconfigs/pull/22930 where I wanted to use `pkg-config` to check that the patch we apply to the `.pc` files to add `libtiff` as a required dependency actually does not break the command.

Still i think for this to be robust we would need not only `sanity_check_deps` but also a separate sets of commands that are supposed to be ran with these extra dependencies loaded, so that all the other sanity checks can be still performed in the normal way

PR #4918, opened 2025-06-10
make EasyBuild plugin-able through entrypoints

Crivella:
The proposed changes leverage [python EntryPoints](https://docs.python.org/3/library/importlib.metadata.html#importlib.metadata.EntryPoints) to allow injecting new functionalities in easybuild without touching the main repo

For now this is a demo of what could be possible implementing the following:

- `easybuild.toolchain` entrypoint to inject a toolchain into EasyBuild from a separate python package
- `easybuild.easyblock` entrypoint to inject a new easyblock into EasyBuild from a separate python package
- `easybuild.hooks` entrypoint to inject new hooks into EasyBuild from a separate python package

For easier error detection in the plugin package, the new `easybuild.tools.entrypoints` defines several `registed_XXX` decorators that needs to be used in order to pass the entrypoint validation.

An example of a python package that makes use of the proposed functionalities can be found in https://github.com/Crivella/test_eb_ep

Using entrypoints is not enabled by default and requires the `--use-entrypoints` flag to be passed to `eb` command.

## Possible upsides

- Simplifies extending EasyBuild functionalities without touching the main repo
- Could simplify working with/maintaining a customized local stack
  - In particular allowing running multiple hooks on the same step with ordering enforced by the priority can be useful
- Could allow to have a minimal version of easybuild that is extensible via modules

## Possible downsides

- Could promote fragmentation, with different packages providing different functionalities and less upstream contributions

## TODO

- [x] Add documentation on how to create a package that uses the entrypoints
  - https://github.com/easybuilders/easybuild-docs/pull/331 
- [x] Add tests

## MAYBE

- [ ] Add new entrypoints for other functionalities
- [x] Possibly rework current (demo) functionalities for better integration (for now i took what i felt was the path of least resistance to implement the demo)

## Notes

- `importlib.metadata` has been part of python only from `3.8` before the 3rd party `importlib-metadata` package would be required
- Currently for `docs.py` CLI options like `--list-easyblocks`  `--list-toolchains` and `--show-config` that are run before the build options are initialized, the default is to assume that `use_entrypoints` is `True`. At runtime when the hooks/easyblocks/toolchains need to be injected, this will only be performed if the options is actually set to `True`

### DIFF
--- a/easybuild/tools/entrypoints.py
+++ b/easybuild/tools/entrypoints.py
@@ -47,11 +47,9 @@ def get_group_entrypoints(group: str):
                 _log.debug("`get_group_entrypoints` called before BuildOptions initialized, with python < 3.8")
         else:
             if HAVE_ENTRY_POINTS_CLS:
-                eps = entry_points(group=group)
-                res = set(eps)
+                res = set(entry_points(group=group))
             else:
-                eps = entry_points()
-                res = set(eps.get(group, []))
+                res = set(entry_points().get(group, []))
 
     return res
 
--- a/easybuild/tools/entrypoints.py
+++ b/easybuild/tools/entrypoints.py
@@ -78,7 +78,7 @@ class EasybuildEntrypoint:
             check = False
             try:
                 check = isinstance(wrap, self.expected_type) or issubclass(wrap, self.expected_type)
-            except:
+            except Exception:
                 pass
             if not check:
                 raise EasyBuildError(
@@ -176,6 +176,7 @@ class EntrypointHook(EasybuildEntrypoint):
             msg = f"Attempting to register unknown hook '{hook_name}'"
             _log.warning(msg)
             raise EasyBuildError(msg)
+
 
 class EntrypointEasyblock(EasybuildEntrypoint):
     """Class to represent an easyblock entrypoint."""
--- a/easybuild/tools/entrypoints.py
+++ b/easybuild/tools/entrypoints.py
@@ -7,7 +7,6 @@ Authors:
 import sys
 import importlib
 from easybuild.tools.config import build_option
-from typing import Callable, List
 
 from easybuild.base import fancylogger
 from easybuild.tools.build_log import EasyBuildError
@@ -23,6 +22,10 @@ if sys.version_info >= (3, 10):
     # Python >= 3.10 uses importlib.metadata.EntryPoints as a type for entry_points()
     HAVE_ENTRY_POINTS_CLS = True
 
+
+EASYBLOCK_ENTRYPOINT = "easybuild.easyblock"
+TOOLCHAIN_ENTRYPOINT = "easybuild.toolchain"
+HOOKS_ENTRYPOINT = "easybuild.hooks"
 
 _log = fancylogger.getLogger('entrypoints', fname=False)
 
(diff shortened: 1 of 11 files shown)


boegel:
I only took a (very) quick look at this, but couple of questions:
- This is basically a (more modern) alternative to `--include-easyblocks`, `--hooks`, `--include-toolchains`, correct?
- What are the main advantages of this approach compared to the mechanisms we already have?
- One downside I see is that this is more "hidden", having entrypoints available won't show up in the output of `eb --show-config`, for example. I think it's important to somehow expose the fact that entrypoints are available, perhaps by making `--show-config` print something extra in case one or more entrypoints are found?

Other random thoughts:
- The fact that this requires Python >= 3.8 (or the installation of an extra Python package) is fine, we're somewhat close to recommending Python >= 3.9 anyway, definitely not a blocker (especially since this is opt-in), as long as there's good error reporting when `importlib.metadata` is not available.
- Since this is opt-in, I have no objections on including this alongside the existing mechanisms we have (`--include-*`, `--hooks`, etc.).
- Can the entrypoints mechanisms and the existing mechanisms (`--include-*`, `--hooks`) be combined? If not, we should clearly define/document which one "wins" in case of clashes, and print a warning to make it crystal clear which easyblock/hooks/etc is being ignored (or even just error out and not allow going that route).

Crivella:
> This is basically a (more modern) alternative to `--include-easyblocks`, `--hooks`, `--include-toolchains`, correct?

This is what I've implemented as a demo as it was probably the easiest features, but i think several other features could also be done through plugins. A few ideas:

- extending the command line options
- adding custom module_tools / naming schemes
- adding custom schedulers for jobs

I think several features could be implemented, requiring varying degree of effort

> What are the main advantages of this approach compared to the mechanisms we already have

- Hooks: Right now only one hook can be ran per step, which requires the logic to run multiple *sub*-hooks to be implemented in the hook itself like we do for EESSI.
  With this, multiple hooks can be registered and ran in the order specified. Also the hooks can come from several independent sources, without needing to manually tying them together.
- Toolchains/EasyBlocks: While it is possible to specify tcs,ebs coming from multiple files, automating their loading eg by switching modules/virtualenvs is tricky.
  EG, for venvs, right now one would require manually modifying the activation script to add the environment variables, every time one wants to add a new *included by default* toolchain or easyblock.

  Using entrypoints, this would be made easier, eg one only has to install/uninstall the respective python package

> One downside I see is that this is more "hidden", having entrypoints available won't show up in the output of `eb --show-config`, for example. I think it's important to somehow expose the fact that entrypoints are available, perhaps by making `--show-config` print something extra in case one or more entrypoints are found?

Will work on it


> The fact that this requires Python >= 3.8 (or the installation of an extra Python package) is fine, we're somewhat close to recommending Python >= 3.9 anyway, definitely not a blocker (especially since this is opt-in), as long as there's good error reporting when `importlib.metadata` is not available.

Added a guard for it. It was also needed to pass the CI (still need a more robust way as the `entry_points()` itself is implemented differently for `3.8 <= Python < 3.10` vs later versions)

> Can the entrypoints mechanisms and the existing mechanisms (`--include-*`, `--hooks`) be combined? If not, we should clearly define/document which one "wins" in case of clashes, and print a warning to make it crystal clear which easyblock/hooks/etc is being ignored (or even just error out and not allow going that route).

Yes they coexist, bot for easyblocks and toolchains I've modified the functions that get them to include the one coming from entrypoints after the ones already picked up by easybuild.

- Hooks: The normal easybuild hook will be ran first, then entrypoint hooks will be ran after (if enabled and found) in the order specified by the priority (descending) and hook name in case of same prio (ascending) (could probably also log a warning if hooks with same priority are found).
- Toolchains: The entrypoint toolchains are appended or prependend (defaults to appended) to the list of toolchains found by `easybuild.tools.toolchains.utilities.search_toolchains()`, allowing an entrypoint toolchain to either just add to the list of toolchains or override an existing one (also here we could add warnings/errors for name clashes between entrypoints and possible another flag to say if a registered toolchain is allowed to override an existing one).
- EasyBlocks: For now entrypoints easyblocks are picked up before existing ones (by changing the results of `easybuild.framework.easyconfig.get_module_path()`) but similar logic could also be implemented here to change the priority

verdurin:
I like the idea - anything that makes it potentially easier to customise EB is a good thing.

easybuild/tools/entrypoints.py
```diff
@@ -0,0 +1,210 @@
+"""Python module to manage entry points for EasyBuild.
```
boegel: copyright/license header is missing

boegel (approved):
lgtm

PR #4920, opened 2025-06-13
Fix reporting of method name in `EasyBlock.run_step`

Flamefire:
The step method is a method returning the actual method. The `Bundle` easyblock uses a `getattr` call there instead of the method name.

As the reflection looks at local name(s) it will find the `getattr` instead of the actual method.

Access the `__name__` property of the returned function instead so we will always get the right name no matter how the lambda/method is defined.


boegel (approved):
lgtm

PR #4922, opened 2025-06-13
Correctly detect final path when extracting multiple tarballs 

Flamefire:
The current implementation of `extract_file` detected folders/files from the first tarball when extracting the second.
Due to the definition of `find_base_dir` it will then return the parent path (usually `builddir`)
which is used as the `finalpath` for sources.
This leads issues requiring workarounds in e.g. the Bundle easyblock
where sources from all components are extracted into the same folder.

Fix by storing the old state of the target folder and detect of
extraction resulted in a single (top-level) folder.

As part of this work I enhanced the  documentation and test of find_base_dir and added tests for the expected and (previous) error cases of `extract_file`

`guess_start_dir` detects the fixed behavior already: If there is a `prefix/<startdir>/<startdir>` which doesn't exist `prefix/<startdir>` is used instead

E.g. GTK3 has start_dir set manually for components due to this bug which would now lead to builddir/foo/foo
IMO trivial to detect/fix

Remark: It is possible to have `find_base_dir` **NOT** change into the resulting directory which means we can do this explicitely when requested instead of undoing it. Not sure if anyone outside the main repo relies on that.

Followup: https://github.com/easybuilders/easybuild-easyblocks/pull/3961

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2852,15 +2852,19 @@ class EasyBlock:
         """
         Unpack the source files.
         """
-        for src in self.src:
+        for i, src in enumerate(self.src):
             self.log.info("Unpacking source %s" % src['name'])
             srcdir = extract_file(src['path'], self.builddir, cmd=src['cmd'],
-                                  extra_options=self.cfg['unpack_options'], change_into_dir=False)
-            change_dir(srcdir)
+                                  extra_options=self.cfg['unpack_options'], change_into_dir=(i == 0))
             if srcdir:
                 self.src[self.src.index(src)]['finalpath'] = srcdir
             else:
                 raise EasyBuildError("Unpacking source %s failed", src['name'])
+        # Compatibility with previous EasyBuild versions:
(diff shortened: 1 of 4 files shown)


boegel (changes requested):
@Flamefire Rather than adding (more) binary blobs to the repo, can we create `multi-0.0.tar.gz` and `multi-1.0.tar.gz` dynamically in the test itself?

That probably even helps with exposing what's being tested, since the structure within these tarballs is now not visible.

This request is fueled by [a recent example of binary blobs being abused](https://lwn.net/Articles/967180/).

I'm aware we have other binary blobs in here already, we should look into getting rid of those too.

Flamefire:
@boegel I replaced the added binaries by creating them in the test (replacing the commit to remove them from the history)

I also replaced all uses of other binaries in that test. The test-method I added should be easily usable for other tests.

What I noticed: Our `filetools.make_archive` is lacking a way to create an archive from a list of files or contents of a folder, i.e. exclude the parent folder from the hierarchy in the archive. See the new method I added.

boegel:
@Flamefire merge conflict to resolve here...

PR #4923, opened 2025-06-14
add support for pre/post `easyblock` hook

### LINKED ISSUE #4541: Dumping of reprod directory happening after post_package_hook
Currently the dump of the `reprod` directory is happening after the last hook that is package specific. All other hooks after that are not specific to any EB file (with the exception of the hook for testcases, but that does not get executed always, and that happens anyway also before the dump happens).

This is problematic in our case, which is admittedly a bit convoluted. Bear with me:

- We install via CI once a MR is merged in our private repo
- The installations via CI are done by a single user that has rights over most of the software stack
- We want different team members to be responsible for their own packages and be able to do corrections or reinstallations on their own, without having rights over everything else
- To implement that we rely on ACLs
- The ACLs are adjusted with a script that is executed with `post_package_hook`

That script works as intended, but the fact that the `reprod` directory is dumped after it is executed, messes up that approach, since the owner of that directory is the one that has rights over the whole stack, and so the different team members can not overwrite those installations, because the ACLs are not applied to that directory.

I can think of 2 solutions:
- Dump the `reprod` directory during the packaging step
- Create a new hook for after the dump, but before the end of the `build_and_install_loop`

Opinions?

boegel:
New hooks which trigger before a single installation starts, and after an installation is fully completed.

fixes #4541


ocaisa (approved):
LGTM

test/framework/toy_build.py
```diff
@@ -3654,13 +3654,19 @@ def start_hook():
...
                print(ec['sources'])
-            # try appending to postinstallcmd to see whether the modification is actually picked up
-            # (required templating to be disabled before parse_hook is called)
+               # try appending to postinstallcmd to see whether the modification is actually picked up
+               # (required templating to be disabled before parse_hook is called)
                ec['postinstallcmds'].append('echo toy')
                print(ec['postinstallcmds'][-1])
 
+            def pre_build_and_install_loop_hook(ecs):
+                print(f'installing {len(ecs)} easyconfigs: {' '.join(ec['full_mod_name'] for ec in ecs)}')
```
ocaisa:
```suggestion
                print(f"installing {len(ecs)} easyconfigs: {' '.join(ec['full_mod_name'] for ec in ecs)}")
```

test/framework/toy_build.py
```diff
@@ -3704,6 +3710,12 @@ def post_run_shell_cmd_hook(cmd, *args, **kwargs):
                     copy_file('toy', 'copy_of_toy')
                     change_dir(cwd)
                     print("'%s' command failed (exit code %s), but I fixed it!" % (cmd, exit_code))
+
+            def post_easyblock_hook(self):
+                print(f'done with installation of {self.name} {self.version}')
+
+            def post_build_and_install_loop_hook(ecs):
+                print(f'done with installing {len(ecs)} easyconfigs: {' '.join(ec[1]['full_mod_name'] for ec in ecs)}')
```
ocaisa:
```suggestion
                print(f"done with installing {len(ecs)} easyconfigs: {' '.join(ec[1]['full_mod_name'] for ec in ecs)}")
```

test/framework/toy_build.py
```diff
@@ -3704,6 +3710,12 @@ def post_run_shell_cmd_hook(cmd, *args, **kwargs):
                     copy_file('toy', 'copy_of_toy')
                     change_dir(cwd)
                     print("'%s' command failed (exit code %s), but I fixed it!" % (cmd, exit_code))
+
+            def post_easyblock_hook(self):
+                print(f'done with installation of {self.name} {self.version}')
+
+            def post_build_and_install_loop_hook(ecs):
+                print(f"done with installing {len(ecs)} easyconfigs: {' '.join(ec[1]['full_mod_name'] for ec in ecs)}")
```
ocaisa:
```suggestion
                print(f"done with installing {len(ecs)} easyconfigs: {' '.join(ec['full_mod_name'] for ec in ecs)}")
```

test/framework/toy_build.py
```diff
@@ -3654,13 +3654,19 @@ def start_hook():
...
                print(ec['sources'])
-            # try appending to postinstallcmd to see whether the modification is actually picked up
-            # (required templating to be disabled before parse_hook is called)
+               # try appending to postinstallcmd to see whether the modification is actually picked up
+               # (required templating to be disabled before parse_hook is called)
                ec['postinstallcmds'].append('echo toy')
                print(ec['postinstallcmds'][-1])
 
+            def pre_build_and_install_loop_hook(ecs):
+                print(f"installing {len(ecs)} easyconfigs: {' '.join(ec['full_mod_name'] for ec in ecs)}")
```
ocaisa:
```suggestion
                print(f"installing {len(ecs)} easyconfigs: {' '.join(ec[1]['full_mod_name'] for ec in ecs)}")
```

test/framework/toy_build.py
```diff
@@ -3704,6 +3710,12 @@ def post_run_shell_cmd_hook(cmd, *args, **kwargs):
                     copy_file('toy', 'copy_of_toy')
                     change_dir(cwd)
                     print("'%s' command failed (exit code %s), but I fixed it!" % (cmd, exit_code))
+
+            def post_easyblock_hook(self):
+                print(f'done with installation of {self.name} {self.version}')
+
+            def post_build_and_install_loop_hook(ecs):
+                print(f"done with installing {len(ecs)} easyconfigs: {' '.join(ec['full_mod_name'] for ec in ecs)}")
```
ocaisa:
```suggestion
                print(f"done with installing {len(ecs)} easyconfigs: {' '.join(ec[1]['full_mod_name'] for ec in ecs)}")
```

ocaisa (changes requested):
I'm only guessing here...

PR #4924, opened 2025-06-17
Remove use of unsupported `source` key in patch spec

Flamefire:
- The easyconfig parser doesn't allow this key
- The documentation doesn't mention it
- No useage in easyblocks or framework are found
- `finalpath` of all but the first source is `builddir`, see https://github.com/easybuilders/easybuild-framework/pull/4922

So just remove this code.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2830,8 +2830,6 @@ class EasyBlock:
             self.log.info("Applying patch %s" % patch['name'])
             trace_msg("applying patch %s" % patch['name'])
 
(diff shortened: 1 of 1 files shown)


boegel (approved):
Makes sense, since trying to use `source` key in `patches` entry leads to clear error:

For example with:

```python
patches = [{'source': 0, 'name': 'test.patch'}]

```
Wrong patch spec '{'source': 0, 'name': 'test.patch'}', use of unknown key source in dict (valid keys are ['name', 'copy', 'level', 'sourcepath', 'alt_location', 'opts'])
```

PR #4932, opened 2025-06-21
add `clean_dir` function, use it to try to empty install dir if removing it fails

smoors:
required for installing in bwrap namespace, see also https://github.com/easybuilders/easybuild-framework/issues/4110
if install dir is bind mounted, it cannot be removed

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1251,6 +1251,7 @@ class EasyBlock:
                     # empty the installation directory, but never remove it
                     empty_dir(dir_name)
                     self.log.info("Emptied old directory %s", dir_name)
+                    set_gid_sticky_bits(dir_name, set_gid, sticky, recursive=True)
                 else:
                     remove_dir(dir_name)
                     self.log.info("Removed old directory %s", dir_name)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -90,10 +90,10 @@ from easybuild.tools.config import install_path, log_path, package_path, source_
 from easybuild.tools.config import DATA, SOFTWARE
 from easybuild.tools.environment import restore_env, sanitize_env
 from easybuild.tools.filetools import CHECKSUM_TYPE_SHA256
-from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file, change_dir, check_lock
+from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file, change_dir, check_lock, clean_dir
 from easybuild.tools.filetools import compute_checksum, convert_name, copy_dir, copy_file, create_lock
 from easybuild.tools.filetools import create_non_existing_paths, create_patch_info, derive_alt_pypi_url, diff_files
-from easybuild.tools.filetools import download_file, clean_dir, encode_class_name, extract_file
+from easybuild.tools.filetools import download_file, encode_class_name, extract_file
 from easybuild.tools.filetools import find_backup_name_candidate, get_cwd, get_source_tarball_from_git, is_alt_pypi_url
 from easybuild.tools.filetools import is_binary, is_parent_path, is_sha256_checksum, mkdir, move_file, move_logs
 from easybuild.tools.filetools import read_file, remove_dir, remove_file, remove_lock, symlink, verify_checksum
@@ -1176,7 +1176,7 @@ class EasyBlock:
             pass
         elif self.build_in_installdir:
             # building in installation directory, so clean it
-            self.make_dir(self.builddir, self.cfg['cleanupoldbuild'], isinstalldir=True)
+            self.make_dir(self.builddir, self.cfg['cleanupoldbuild'], clean_instead_of_remove=True)
         else:
             # make sure we no longer sit in the build directory before removing it.
             change_dir(self.orig_workdir)
@@ -1229,7 +1229,7 @@ class EasyBlock:
         self.make_dir(self.installdir, self.cfg['cleanupoldinstall'], dontcreateinstalldir=dontcreate,
                       isinstalldir=True)
 
-    def make_dir(self, dir_name, clean, dontcreateinstalldir=False, isinstalldir=False):
+    def make_dir(self, dir_name, clean, dontcreateinstalldir=False, clean_instead_of_remove=False):
         """
         Create the directory.
         """
@@ -1247,7 +1247,7 @@ class EasyBlock:
                     backupdir = "%s.%s" % (dir_name, timestamp)
                     copy_dir(dir_name, backupdir)
                     self.log.info("Copied old directory %s to %s", dir_name, backupdir)
-                if isinstalldir:
+                if clean_instead_of_remove:
                     # clean the installation directory: first try to remove it; if that fails, empty it
                     clean_dir(dir_name)
                     self.log.info("Emptied old directory %s", dir_name)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1175,7 +1175,7 @@ class EasyBlock:
         if self.build_in_installdir and self.iter_idx > 0:
             pass
         elif self.build_in_installdir:
-            # building in installation directory, so empty it but don't remove it
+            # building in installation directory, so clean it
             self.make_dir(self.builddir, self.cfg['cleanupoldbuild'], isinstalldir=True)
         else:
             # make sure we no longer sit in the build directory before removing it.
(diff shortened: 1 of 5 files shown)


boegel:
@smoors As discussed during today's conf call, it probably makes sense to introduce another function (`clean_dir`?) that does a two-step approach: try to remove the specified directory, fall back to emptying it when that fails for some reason.

easybuild/framework/easyblock.py
```diff
@@ -93,10 +93,11 @@
 from easybuild.tools.filetools import adjust_permissions, apply_patch, back_up_file, change_dir, check_lock
 from easybuild.tools.filetools import compute_checksum, convert_name, copy_dir, copy_file, create_lock
 from easybuild.tools.filetools import create_non_existing_paths, create_patch_info, derive_alt_pypi_url, diff_files
-from easybuild.tools.filetools import download_file, encode_class_name, extract_file, find_backup_name_candidate
-from easybuild.tools.filetools import get_cwd, get_source_tarball_from_git, is_alt_pypi_url, is_binary, is_parent_path
-from easybuild.tools.filetools import is_sha256_checksum, mkdir, move_file, move_logs, read_file, remove_dir
-from easybuild.tools.filetools import remove_file, remove_lock, symlink, verify_checksum, weld_paths, write_file
+from easybuild.tools.filetools import download_file, clean_dir, encode_class_name, extract_file
```
```suggestion
from easybuild.tools.filetools import clean_dir, download_file, encode_class_name, extract_file

File: easybuild/framework/easyblock.py [hunk 2]
```diff
@@ -1220,9 +1226,10 @@ def make_installdir(self, dontcreate=None):
         if self.build_in_installdir:
             self.cfg['keeppreviousinstall'] = True
         dontcreate = (dontcreate is None and self.cfg['dontcreateinstalldir']) or dontcreate
-        self.make_dir(self.installdir, self.cfg['cleanupoldinstall'], dontcreateinstalldir=dontcreate)
+        self.make_dir(self.installdir, self.cfg['cleanupoldinstall'], dontcreateinstalldir=dontcreate,
+                      isinstalldir=True)
 
-    def make_dir(self, dir_name, clean, dontcreateinstalldir=False):
+    def make_dir(self, dir_name, clean, dontcreateinstalldir=False, isinstalldir=False):
```
boegel: `isinstalldir` can be renamed to something that more directly specifies what it done when this is enabled/disabled:

```suggestion
    def make_dir(self, dir_name, clean, dontcreateinstalldir=False, clean_instead_of_remove=False):
```

easybuild/tools/filetools.py
```diff
@@ -406,6 +440,18 @@ def remove_dir(path):
...
 
+def clean_dir(path):
+    """
+    Try to remove directory at specified path.
+    If that fails, empty directory instead.
+    """
+    try:
+        remove_dir(path)
+    except EasyBuildError:
+        _log.debug("Failed to remove directory %s, emptying it instead", path)
```
```suggestion
    except EasyBuildError as err:
        _log.debug(f"Removing directory {path} failed, will try to empty it instead: {err}")

**Review by boegel — CHANGES_REQUESTED** [review]
@smoors Let's also add a test for the `clean_dir` function?

File: easybuild/framework/easyblock.py [hunk 4]
```diff
@@ -1171,8 +1172,13 @@ def make_builddir(self):
...
-        if not (self.build_in_installdir and self.iter_idx > 0):
-            # make sure we no longer sit in the build directory before cleaning it.
+        if self.build_in_installdir and self.iter_idx > 0:
+            pass
+        elif self.build_in_installdir:
+            # building in installation directory, so clean it
+            self.make_dir(self.builddir, self.cfg['cleanupoldbuild'], clean_instead_of_remove=True)
+        else:
+            # make sure we no longer sit in the build directory before removing it.
             change_dir(self.orig_workdir)
```
boegel, line 1179: This `change_dir` now only happens in the `else`, while before it was also done when `self.build_in_installdir` is enabled

easybuild/framework/easyblock.py
```diff
@@ -1171,8 +1172,13 @@ def make_builddir(self):
...
             self.make_dir(self.builddir, self.cfg['cleanupoldbuild'])
```
boegel: How about this:
```suggestion
        if self.build_in_installdir and self.iter_idx > 0:
            pass
        else:
            # make sure we no longer sit in the build directory before removing it.
            change_dir(self.orig_workdir)
            # if we're building in installation directory, clean it
            self.make_dir(self.builddir, self.cfg['cleanupoldbuild'], clean_instead_of_remove=self.build_in_installdir)

File: easybuild/framework/easyblock.py [hunk 6]
```diff
@@ -1233,15 +1240,20 @@ def make_dir(self, dir_name, clean, dontcreateinstalldir=False):
...
-                self.log.info("Moving existing directory %s out of the way...", dir_name)
-                timestamp = time.strftime("%Y%m%d-%H%M%S")
-                backupdir = "%s.%s" % (dir_name, timestamp)
-                move_file(dir_name, backupdir)
-                self.log.info("Moved old directory %s to %s", dir_name, backupdir)
+                if not clean:
+                    self.log.info("Creating backup of directory %s...", dir_name)
+                    timestamp = time.strftime("%Y%m%d-%H%M%S")
+                    backupdir = "%s.%s" % (dir_name, timestamp)
+                    copy_dir(dir_name, backupdir)
```
boegel: This was essentially changed from `mv` to `cp` + `rm -r`, which is a significantly more expensive operation, so it should only be done when it makes sense, i.e. when `clean_instead_of_remove` is enabled?

easybuild/tools/filetools.py
```diff
@@ -376,6 +376,40 @@ def remove_file(path):
...
+        ok = False
+        errors = []
+        # Try multiple times to cater for temporary failures on e.g. NFS mounted paths
+        max_attempts = 3
+        for i in range(0, max_attempts):
+            try:
+                for item in os.listdir(path):
+                    subpath = os.path.join(path, item)
+                    if os.path.isfile(subpath) or os.path.islink(subpath):
+                        os.remove(subpath)
```
boegel: use `remove_file` here?

easybuild/tools/filetools.py
```diff
@@ -376,6 +376,40 @@ def remove_file(path):
...
+                    elif os.path.isdir(subpath):
+                        shutil.rmtree(subpath)
```
boegel: use `remove_dir` here?

easybuild/tools/filetools.py
```diff
@@ -376,6 +376,40 @@ def remove_file(path):
...
+    # early exit in 'dry run' mode
+    if build_option('extended_dry_run'):
+        dry_run_msg("directory %s emptied" % path, silent=build_option('silent'))
+        return
+
+    if os.path.exists(path):
+        ok = False
+        errors = []
+        # Try multiple times to cater for temporary failures on e.g. NFS mounted paths
+        max_attempts = 3
```
boegel: I wouldn't do max. 3 attempts here, leave that up to `remove_dir`?

So this can be simplified *a lot* imho, no need to copy-paste most of `remove_dir` in here imho

boegel (changes requested):

easybuild/framework/easyblock.py
smoors: urgh, this is exactly why i removed the NOT logic, to avoid silly mistakes like this..

easybuild/framework/easyblock.py
```diff
@@ -1233,15 +1238,24 @@ def make_dir(self, dir_name, clean, dontcreateinstalldir=False):
...
+                    else:
+                        move_file(dir_name, backupdir)
+                        self.log.info(f"Moved old directory {dir_name} to {backupdir}")
+                if clean_instead_of_remove:
+                    # clean the installation directory: first try to remove it; if that fails, empty it
+                    clean_dir(dir_name)
+                    self.log.info(f"Cleaned old directory {dir_name}")
+                else:
+                    remove_dir(dir_name)
+                    self.log.info(f"Removed old directory {dir_name}")
```
boegel: shouldn't this whole block be under an `if clean:`?

smoors: no, only the last else block can go under `if clean` (although not strictly necessary). if clean_instead_of_remove, we should always clean it.

smoors: done in https://github.com/easybuilders/easybuild-framework/pull/4932/commits/66ae73dc1e1751cf68ad93b4aad70c0799205741

boegel (approved):
lgtm

PR #4935, opened 2025-06-24
update `XORG_*_SOURCE` constants to use new URL

jfgrimm:


boegel:
@jfgrimm Is there a public statement somewhere that announces this?

If so, we should mention it in the PR description, maybe also in the commit message?

jfgrimm:
@boegel I haven't been able to find one, still looking

boegel:
@jfgrimm As discussed in Slack, a better alternative is probably to introduce new `XORG_*` constants that provide tuples of source URLs, so both will be tried?

boegel:
@jfgrimm Can you look into failing CI?

jfgrimm:
@boegel will take a look, I think it's failing on the from-commit test

jfgrimm:
@boegel not sure how to best tackle the failing `from_commit` tests hitting the deprecation warning

boegel:
> @boegel not sure how to best tackle the failing `from_commit` tests hitting the deprecation warning

@jfgrimm We'll need to update those to using a different/newer commit.

If we already open an easyconfig PR for easyconfigs using the `SYSTEM` toolchain (like `ncurses-6.2.eb`) to stop using the deprecated constants like `GNU_SOURCE`, we can already bump the commit used in `test_from_commit` even though the easyconfig PR isn't merged yet (it can only be merged after this PR got merged)...

jfgrimm:
I've currently got https://github.com/easybuilders/easybuild-easyconfigs/pull/23299 open, which updates the use in all easyconfigs
Want me to split `ncurses` out into a separate PR, or is one large PR OK for the test?

PR #4938, opened 2025-06-27
trigger `post_easyblock_hook` later

bedroge:
The `pre/post_easyblock_hook` was added in #4923, but the `post_easyblock_hook` is triggered a bit too early: though the installation itself should be completed, some files (e.g. the log file and env file) still need to be copied to the `easybuild` subdir. This PR fixes that by triggering it a bit later, right before calling `del app`.


boegel (approved):
lgtm

Issue #4864, opened 2025-04-29
Patch files for extensions are not copied to the easybuild or reprod directory


casparvl:
We're trying to reinstall a stack for a different architecture, based on things in the `easybuild` and `easybuild/reprod` directories. However, for `SciPy-Bundle`, we ran into an issue. The original build was done using EasyBuild 4.9.4 and an EasyStack

```
easyconfigs:
  - SciPy-bundle-2023.07-gfbf-2023a.eb:
      options:
        # see https://github.com/easybuilders/easybuild-easyconfigs/pull/21693
        from-commit: 7c5144d2c1a061cd9f08b5901970b7f6ec5eb5c0
```

Now, we'd like to reinstall from the `reprod` folder, but this fails because of a missing patch file:

```
$ eb /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/SciPy-bundle/2023.07-gfbf-2023a/easybuild/reprod/SciPy-bundle-2023.07-gfbf-2023a.eb --stop patch --force --installpath /tmp/casparl/eb-test-installpath/
...
== FAILED: Installation ended unsuccessfully (build directory: /tmp/casparl/easybuild/build/SciPybundle/2023.07/gfbf-2023a): build failed (first 300 chars): Couldn't find file scipy-1.11.1_vectorization_error.patch anywhere, and downloading it didn't work either... Paths attempted (in
order): /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/SciPy-bundle/2023.07-gfbf-2023a/easybuild/reprod/s/SciPy-bundle/extensions/s (took 1 secs)
== Results of the build can be found in the log file(s) /scratch-local/casparl.11439033/eb-i9dk4swq/easybuild-SciPy-bundle-2023.07-20250429.113001.cYcuE.log
Fetching files:  77% (17/22) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╸━━━━━━━━━ 0:00:01 (scipy-1.11.1_vectorization_error.patch)
[...]

Issue #4863, opened 2025-04-29
Patch files are not copied to the reprod directory, only to the easybuild directory


casparvl:
The `reprod` folder contains EasyBlocks and EasyConfigs, but not the patch files needed to rebuild a piece of software. E.g.:

```
$ ls /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/GCCcore/12.3.0/easybuild/reprod/
easyblocks  GCCcore-12.3.0.eb  GCCcore-12.3.0.env  hooks
``` 

We originally installed this with EasyBuild 4.9.0 with the following EasyStack:
```
easyconfigs:
  - GCCcore-12.3.0.eb:
      options:
        from-pr: 20218
```

Trying to reinstall from the `reprod` folder using that same EasyBuild version fails:

```
$ eb /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/GCCcore/12.3.0/easybuild/reprod/GCCcore-12.3.0.eb --stop patch --force --installpath /tmp/casparl/eb-test-installpath/
...
== FAILED: Installation ended unsuccessfully (build directory: /tmp/casparl/easybuild/build/GCCcore/12.3.0/system-system): build failed (first 300 chars): Couldn't find file GCCcore-12.x_riscv_multiarch_support.patch anywhere, and downloading it didn't work either... Paths attempted
(in order): /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/GCCcore/12.3.0/easybuild/reprod/g/GCCcore/GCCcore-12.x_riscv_multiarch_s (took 9 secs)
== Results of the build can be found in the log file(s) /scratch-local/casparl.11439033/eb-6nf_7ddt/easybuild-GCCcore-12.3.0-20250429.112144.rOhkJ.log
Fetching files:  93% (13/14) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╺━━ 0:00:09 (GCCcore-12.x_riscv_multiarch_support.patch)
[...]

PR #4939, opened 2025-06-27
also copy patches of extensions to `easybuild` subdirectory of installation directory

bedroge:
Solves https://github.com/easybuilders/easybuild-framework/issues/4864 by adding the patches of extensions to the list of patches that are copied to `$installdir/easybuild`.

Note that there's also https://github.com/easybuilders/easybuild-framework/issues/4863: I'm not sure if the patches should go into either `easybuild` or `easybuild/reprod` or both, for now I've taken the same approach that is used for regular patches: only copy them to `easybuild`. But this can be easily changed.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -5144,7 +5144,10 @@ def build_and_install_one(ecdict, init_env):
                 _log.debug("Copied easyconfig file %s to %s", spec, newspec)
 
                 # copy patches
-                for patch in app.patches:
+                patches = app.patches
+                for ext in app.exts:
+                    patches += ext.get('patches', [])
+                for patch in patches:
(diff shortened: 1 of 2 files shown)


ocaisa (changes requested):
This looks good, but since it does indeed look pretty trivial to add the patches to the reprod directory I would also do that while we are here. I also note that the patches are also required for the repo try/except loop above (line 5122)

ocaisa (dismissed):

ocaisa (approved):

bedroge:
Note that this bug fix introduced a new bug that pops up when installing an easyconfig with an extension that has a patch. See https://github.com/easybuilders/easybuild-framework/pull/4959#issuecomment-3061245234 for more details, and a fix is implemented in https://github.com/easybuilders/easybuild-framework/pull/4960.

PR #4942, opened 2025-06-27
Don't log variables (un)set + raw output of `module` command when cleaning up fake module

### LINKED ISSUE #4937: Remove output of module commands in log files
The output of the lmod command in the log makes the test report file miss (almost) all important info. E.g.: https://gist.github.com/Flamefire/704e041f5696e98b6b8ecb0ef98ec02f

I.e. there is a large chunk of those:
```
del os.environ["_ModuleTable071_"]
os.environ["_ModuleTable072_"] = ''
del os.environ["_ModuleTable072_"]
<etc>
...
_mlstatus = True

== 2025-06-25 17:00:38,652 run.py:611 INFO Warnings and errors of '/usr/share/lmod/lmod/libexec/lmod ...' shell command (stderr only):

== 2025-06-25 17:00:38,652 run.py:616 INFO Shell command completed successfully (see output above): /usr/share/lmod/lmod/libexec/lmod python unload TensorFlow/2.18.1-CUDA-12.6.0
== 2025-06-25 17:00:38,655 environment.py:93 INFO Environment variable LD_LIBRARY_PATH set to /software/protobuf/28.0-GCCcore-13.3.0/lib:/software/Abseil/20240722.0-GCCcore-13.3.0/lib:/software/snappy/1.2.1-GCCcore-13.3.0/lib:/software/libpng/1.6.43-GCCcore-13.3.0/lib:/software/nsync/1.29.2-GCCcore-13.3.0/lib:/software/libjpeg-turbo/3.0.1-GCCcore-13.3.0/lib:/software/JsonCpp/1.9.5-GCCcore-13.3.0/lib:/software/giflib/5.2.1-GCCcore-13.3.0/lib:/software/flatbuffers/24.3.25-GCCcore-13.3.0/lib:/software/double-conversion/3.3.0-GCCcore-13.3.0/lib:/software/HDF5/1.14.5-gompi-2024a/lib:/software/Szip/2.1.1-GCCcore-13.3.0/lib:/software/SciPy-bundle/2024.05-gfbf-2024a/lib/python3.12/site-packages/numpy/core/lib:/software/genoa/r24.04/NCCL/2.22.3-GCCcore-13.3.0-CUDA-12.6.0/lib:/software/genoa/r24.04/GDRCopy/2.4.1-GCCcore-13.3.0/lib:/software/cuDNN/9.5.0.50-CUDA-12.6.0/lib:/software/CUDA/12.6.0/nvvm/lib64:/software/CUDA/12.6.0/extras/CUPTI/lib64:/software/CUDA/12.6.0/targets/x86_64-linux/lib:/software/Boost/1.85.0-GCC-13.3.0/lib:/software/ICU/75.1-GCCcore-13.3.0/lib:/software/zstd/1.5.6-GCCcore-13.3.0/lib:/software/lz4/1.9.4-GCCcore-13.3.0/lib:/software/gettext/0.22.5-GCCcore-13.3.0/lib:/software/libiconv/1.17-GCCcore-13.3.0/lib:/software/genoa/r24.04/expat/2.6.2-GCCcore-13.3.0/lib:/software/genoa/r24.04/cURL/8.7.1-GCCcore-13.3.0/lib:/software/genoa/r24.04/Java/11.0.20/lib:/software/genoa/r24.04/Python/3.12.3-GCCcore-13.3.0/lib:/software/genoa/r24.04/libffi/3.4.5-GCCcore-13.3.0/lib64:/software/genoa/r24.04/SQLite/3.45.3-GCCcore-13.3.0/lib:/software/genoa/r24.04/Tcl/8.6.14-GCCcore-13.3.0/lib:/software/genoa/r24.04/libreadline/8.2-GCCcore-13.3.0/lib:/software/genoa/r24.04/ncurses/6.5-GCCcore-13.3.0/lib:/software/genoa/r24.04/bzip2/1.0.8-GCCcore-13.3.0/lib:/software/genoa/r24.04/ScaLAPACK/2.2.0-gompi-2024a-fb/lib:/software/genoa/r24.04/FFTW.MPI/3.3.10-gompi-2024a/lib:/software/genoa/r24.04/FFTW/3.3.10-GCC-13.3.0/lib:/software/genoa/r24.04/FlexiBLAS/3.4.4-GCC-13.3.0/lib:/software/genoa/r24.04/OpenBLAS/0.3.27-GCC-13.3.0/lib:/software/genoa/r24.04/OpenMPI/5.0.3-GCC-13.3.0/lib:/software/genoa/r24.04/UCC/1.3.0-GCCcore-13.3.0/lib:/software/genoa/r24.04/PRRTE/3.0.5-GCCcore-13.3.0/lib:/software/genoa/r24.04/PMIx/5.0.2-GCCcore-13.3.0/lib:/software/genoa/r24.04/libfabric/1.21.0-GCCcore-13.3.0/lib:/software/genoa/r24.04/UCX/1.16.0-GCCcore-13.3.0/lib:/software/genoa/r24.04/libevent/2.1.12-GCCcore-13.3.0/lib:/software/genoa/r24.04/OpenSSL/3/lib:/software/genoa/r24.04/hwloc/2.10.0-GCCcore-13.3.0/lib:/software/genoa/r24.04/libpciaccess/0.18.1-GCCcore-13.3.0/lib:/software/genoa/r24.04/libxml2/2.12.7-GCCcore-13.3.0/lib:/software/genoa/r24.04/XZ/5.4.5-GCCcore-13.3.0/lib:/software/genoa/r24.04/numactl/2.0.18-GCCcore-13.3.0/lib:/software/binutils/2.42-GCCcore-13.3.0/lib:/software/zlib/1.3.1-GCCcore-13.3.0/lib:/software/genoa/r24.04/GCCcore/13.3.0/lib64 (previously undefined)
== 2025-06-25 17:00:38,657 filetools.py:403 INFO Path /tmp/easybuild-tmp/eb-gehzfysz/tmp7vzo8jyc successfully removed.
== 2025-06-25 17:00:38,657 environment.py:93 INFO Environment variable _ModuleTable002_ set to IjAyMDI0MDcyMi4qemZpbmFsIiwKfSwKQmF6ZWwgPSB7CmZuID0gIi9kYXRhL2hvcnNlL3dzL3MzMjQ4OTczLUVhc3lCdWlsZC9lYXN5YnVpbGQtZ2Vub2EvbW9kdWxlcy9hbGwvQ29tcGlsZXIvR0NDY29yZS8xMy4zLjAvQmF6ZWwvNi41LjAtSmF2YS0xMS5sdWEiLApmdWxsTmFtZSA9ICJCYXplbC82LjUuMC1KYXZhLTExIiwKbG9hZE9yZGVyID0gMjcsCnByb3BUID0ge30sCnN0YWNrRGVwdGggPSAwLApzdGF0dXMgPSAiYWN0aXZlIiwKdXNlck5hbWUgPSAiQmF6ZWwvNi41LjAtSmF2YS0xMSIsCndWID0gIjAwMDAwMDAwNi4wMDAwMDAwMDUuKmphdmEuKnpmaW5hbC0uMDAwMDAwMDExLip6ZmluYWwiLAp9LApCb29zdCA9IHsKZm4gPSAiL2RhdGEvaG9yc2Uvd3MvczMyNDg5 (previous value: 'LjAwMDAwMDAwNS4qamF2YS4qemZpbmFsLS4wMDAwMDAwMTEuKnpmaW5hbCIsCn0sCkJvb3N0ID0gewpmbiA9ICIvZGF0YS9ob3JzZS93cy9zMzI0ODk3My1FYXN5QnVpbGQvZWFzeWJ1aWxkLWdlbm9hL21vZHVsZXMvYWxsL0NvbXBpbGVyL0dDQy8xMy4zLjAvQm9vc3QvMS44NS4wLmx1YSIsCmZ1bGxOYW1lID0gIkJvb3N0LzEuODUuMCIsCmxvYWRPcmRlciA9IDQ2LApwcm9wVCA9IHt9LApyZWZfY291bnQgPSAxLApzdGFja0RlcHRoID0gMSwKc3RhdHVzID0gImFjdGl2ZSIsCnVzZXJOYW1lID0gIkJvb3N0LzEuODUuMCIsCndWID0gIjAwMDAwMDAwMS4wMDAwMDAwODUuKnpmaW5hbCIsCn0sCkN5dGhvbiA9IHsKZm4gPSAiL2RhdGEvaG9yc2Uvd3MvczMyNDg5NzMtRWFzeUJ1')
== 2025-06-25 17:00:38,657 environment.py:93 INFO Environment variable __LMOD_REF_COUNT_MODULEPATH set to /modules/all:1;/modules/all/MPI/GCC/13.3.0/OpenMPI/5.0.3:1;/software/modules/genoa/r24.04/all/MPI/GCC/13.3.0/OpenMPI/5.0.3:1;/modules/all/Compiler/GCC/13.3.0:1;/software/modules/genoa/r24.04/all/Compiler/GCC/13.3.0:1;/modules/all/Compiler/GCCcore/13.3.0:1;/software/modules/genoa/r24.04/all/Compiler/GCCcore/13.3.0:1;/modules/all/Core:1;/home/s3248973/.local/EasyBuildDev/modules:1;/software/modules/genoa/r24.04/all:1;/software/modules/genoa/r24.04/all/Core:1;/software/modules/releases/genoa:1 (previous value: '/modules/all/MPI/GCC/13.3.0/OpenMPI/5.0.3:2;/software/modules/genoa/r24.04/all/MPI/GCC/13.3.0/OpenMPI/5.0.3:2;/modules/all/Compiler/GCC/13.3.0:2;/software/modules/genoa/r24.04/all/Compiler/GCC/13.3.0:2;/modules/all/Compiler/GCCcore/13.3.0:2;/software/modules/genoa/r24.04/all/Compiler/GCCcore/13.3.0:2;/modules/all/Core:1;/tmp/easybuild-tmp/eb-gehzfysz/tmp7vzo8jyc/all/MPI/GCC/13.3.0/OpenMPI/5.0.3:1;/modules/all:1;/home/s3248973/.local/EasyBuildDev/modules:1;/software/modules/genoa/r24.04/all:1;/software/modules/genoa/r24.04/all/Core:1;/software/modules/releases/genoa:1')
<etc>
...
```

Do we really need the output of the module commands or can we silence them completely?

Flamefire:
Unloading a module (and its dependencies) changes a massive amount of variables.
As the changes done by *unloading* the fake module aren't really relevant for the remaining build (it is usually done at the end) do not log them.

This avoids filling up the part of the log uploaded with test reports with irrelevant information.

Example: https://gist.github.com/Flamefire/afae1a8e2b18a4bc2786f539d5f439d3

This change silences the output of the `module unload <fake module>` command by introducing an additional parameter to `run_shell_command` that suppresses output for successful commands.   
Another parameter for the `environment` module allows to suppress output for each set variable, e.g. 
> == 2025-06-26 19:43:42,596 environment.py:93 INFO Environment variable EBEXTSLISTMPI4PY set to mpi4py-4.0.1 (previously undefined)

Closes #4937 (mostly, we might want to suppress output of all module commands or print them only in debug logs)

I think that unloading the fake module and removing the module path is not useful. The succeeding `restore_env` already undoes all changes done by loading the module.

Note that e.g. loading the fake module unsets many environment variables. E.g. `CUDA_CACHE_DISABLE` which the `restore_env` then sets again. So directly calling `restore_env` would already reduce the amount of changes done and hence reduce the number of log lines.   
Why don't we do that?

### DIFF
--- a/easybuild/tools/environment.py
+++ b/easybuild/tools/environment.py
@@ -76,12 +76,13 @@ def get_changes():
     return _changes
 
 
-def setvar(key, value, verbose=True):
+def setvar(key, value, verbose=True, log_changes=True):
     """
     put key in the environment with value
     tracks added keys until write_changes has been called
 
     :param verbose: include message in dry run output for defining this environment variable
+    :param log_changes: show the change in the log
     """
     try:
         oldval_info = "previous value: '%s'" % os.environ[key]
@@ -90,7 +91,8 @@ def setvar(key, value, verbose=True):
     # os.putenv() is not necessary. os.environ will call this.
     os.environ[key] = value
     _changes[key] = value
-    _log.info("Environment variable %s set to %s (%s)", key, value, oldval_info)
+    if log_changes:
+        _log.info("Environment variable %s set to %s (%s)", key, value, oldval_info)
 
     if verbose and build_option('extended_dry_run'):
         quoted_value = shell_quote(value)
@@ -149,23 +151,22 @@ def read_environment(env_vars, strict=False):
     return result
 
 
-def modify_env(old, new, verbose=True):
+def modify_env(old, new, verbose=True, log_changes=True):
     """
     Compares two os.environ dumps. Adapts final environment.
+    :param log_changes: show the change in the log
     """
     old_keys = list(old.keys())
     new_keys = list(new.keys())
 
     for key in new_keys:
-        # set them all. no smart checking for changed/identical values
         if key in old_keys:
-            # hmm, smart checking with debug logging
-            if not new[key] == old[key]:
+            if new[key] != old[key]:
                 _log.debug("Key in new environment found that is different from old one: %s (%s)", key, new[key])
-                setvar(key, new[key], verbose=verbose)
+                setvar(key, new[key], verbose=verbose, log_changes=log_changes)
         else:
             _log.debug("Key in new environment found that is not in old one: %s (%s)", key, new[key])
-            setvar(key, new[key], verbose=verbose)
+            setvar(key, new[key], verbose=verbose, log_changes=log_changes)
 
     for key in old_keys:
         if key not in new_keys:
@@ -174,11 +175,12 @@ def modify_env(old, new, verbose=True):
             del os.environ[key]
 
 
-def restore_env(env):
+def restore_env(env, log_changes=True):
     """
     Restore active environment based on specified dictionary.
+    :param log_changes: show the change in the log
     """
-    modify_env(os.environ, env, verbose=False)
+    modify_env(os.environ, env, verbose=False, log_changes=log_changes)
 
 
 def sanitize_env():
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1253,7 +1253,8 @@ class ModulesTool:
         cmd = ' '.join(cmd_list)
         # note: module commands are always run in dry mode, and are kept hidden in trace and dry run output
         res = run_shell_cmd(cmd_list, env=environ, fail_on_error=False, use_bash=False, split_stderr=True,
-                            hidden=True, in_dry_run=True, output_file=False)
+                            hidden=True, in_dry_run=True, output_file=False,
+                            hide_output_on_success=silent)
 
         # stdout will contain python code (to change environment etc)
         # stderr will contain text (just like the normal module command)
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1120,12 +1120,12 @@ class ModulesTool:
             if allow_reload or mod not in loaded_modules:
                 self.run_module('load', mod)
 
-    def unload(self, modules=None):
+    def unload(self, modules, silent=False):
         """
         Unload all requested modules.
         """
         for mod in modules:
-            self.run_module('unload', mod)
+            self.run_module('unload', mod, silent=silent)
 
     def purge(self):
         """
@@ -1188,9 +1188,9 @@ class ModulesTool:
 
         return modpath
 
-    def set_path_env_var(self, key, paths):
+    def set_path_env_var(self, key, paths, silent=False):
         """Set path environment variable to the given list of paths."""
-        setvar(key, os.pathsep.join(paths), verbose=False)
+        setvar(key, os.pathsep.join(paths), verbose=False, log_changes=not silent)
 
     def check_module_output(self, cmd, stdout, stderr):
         """Check output of 'module' command, see if if is potentially invalid."""
@@ -1249,6 +1249,7 @@ class ModulesTool:
                 self.log.debug("Changing %s from '%s' to '%s' in environment for module command",
                                key, old_value, new_value)
 
+        silent = kwargs.get('silent', False)
         cmd_list = self.compose_cmd_list(args)
         cmd = ' '.join(cmd_list)
         # note: module commands are always run in dry mode, and are kept hidden in trace and dry run output
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1121,12 +1121,12 @@ class ModulesTool:
             if allow_reload or mod not in loaded_modules:
                 self.run_module('load', mod)
 
-    def unload(self, modules, silent=False):
+    def unload(self, modules, log_changes=True):
         """
         Unload all requested modules.
         """
         for mod in modules:
-            self.run_module('unload', mod, silent=silent)
+            self.run_module('unload', mod, log_changes=log_changes)
 
     def purge(self):
         """
@@ -1189,9 +1189,9 @@ class ModulesTool:
 
         return modpath
 
-    def set_path_env_var(self, key, paths, silent=False):
+    def set_path_env_var(self, key, paths, log_changes=True):
         """Set path environment variable to the given list of paths."""
-        setvar(key, os.pathsep.join(paths), verbose=False, log_changes=not silent)
+        setvar(key, os.pathsep.join(paths), verbose=False, log_changes=log_changes)
 
     def check_module_output(self, cmd, stdout, stderr):
         """Check output of 'module' command, see if if is potentially invalid."""
@@ -1250,13 +1250,13 @@ class ModulesTool:
                 self.log.debug("Changing %s from '%s' to '%s' in environment for module command",
                                key, old_value, new_value)
 
-        silent = kwargs.get('silent', False)
+        log_changes = kwargs.get('log_changes', True)
         cmd_list = self.compose_cmd_list(args)
         cmd = ' '.join(cmd_list)
         # note: module commands are always run in dry mode, and are kept hidden in trace and dry run output
         res = run_shell_cmd(cmd_list, env=environ, fail_on_error=False, use_bash=False, split_stderr=True,
                             hidden=True, in_dry_run=True, output_file=False,
-                            hide_output_on_success=silent)
+                            log_output_on_success=log_changes)
 
         # stdout will contain python code (to change environment etc)
         # stderr will contain text (just like the normal module command)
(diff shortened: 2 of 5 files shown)


easybuild/tools/environment.py
```diff
@@ -76,12 +76,13 @@ def get_changes():
     return _changes
 
 
-def setvar(key, value, verbose=True):
+def setvar(key, value, verbose=True, log_changes=True):
```
lexming, line 79: can we use the existing `verbose` for this? something like
```
if verbose:
    <print log>
if verbose and dry_run:
    <do the existing dry run prints>
```

Flamefire: `verbose` is currently solely used for the extended-dry-run and I wasn't sure of the impact.

My new idea would be:

`if verbose or debug-option`: Print setting   
Set `verbose=False` for all module command uses and the fake module unset.   
Also only log output of module commands (lmod calls) when debug-option is set

The changes by the module commands are usually not interesting and just make the log harder to read. So suppressing them by default but with an option to show them sounds like the most sensible approach.

Flamefire: Actually at least the reports from boegelbot runs include the debug flag so that wouldn't make those reports more readable :-/

lexming: Given the limited length of the failed report, we can very well disable `debug` in the bot. Most of the time debug logs are useless when a build fails anyhow, in that case you just need the output of the failing command/test/sanity check. We can discuss about this in the next call.

Flamefire: So shall we go forward with my proposal above or wait for the call too?

boegel: I strongly prefer keeping `debug` enabled in the bot, since it does result in a better log (though less easy to parse).

I think selectively making `setvar` log less as is being done here makes sense.

easybuild/tools/run.py
```diff
@@ -360,7 +360,7 @@ def _answer_question(stdout, proc, qa_patterns, qa_wait_patterns):
 def run_shell_cmd(cmd, fail_on_error=True, split_stderr=False, stdin=None, env=None,
                   hidden=False, in_dry_run=False, verbose_dry_run=False, work_dir=None, use_bash=True,
                   output_file=True, stream_output=None, asynchronous=False, task_id=None, with_hooks=True,
-                  qa_patterns=None, qa_wait_patterns=None, qa_timeout=100):
+                  qa_patterns=None, qa_wait_patterns=None, qa_timeout=100, hide_output_on_success=False):
```
boegel: `hide` is a bit ambiguous (hide from where), should be clear that this is about not logging stuff

So I would change this to `log_output_on_success=True` and adjust accordingly below and in `run_module`

easybuild/tools/modules.py
```diff
@@ -1189,9 +1189,9 @@ def modulefile_path(self, mod_name, strip_ext=False):
 
         return modpath
 
-    def set_path_env_var(self, key, paths):
+    def set_path_env_var(self, key, paths, silent=False):
```
boegel: `silent` is too ambiguous, since this only affects logging, should also be `log_changes=True` here?

boegel (changes requested):
@Flamefire Looks OK, just some suggestion w.r.t. naming of the extra options, to make it clear those impact (only) logging (and not trace output, for example).

I've implemented the suggestions myself in:
- https://github.com/Flamefire/easybuild-framework/pull/8

easybuild/tools/environment.py
Flamefire: I'd want to make a more general decision here. Basically what we usually don't want is output from a) `module commands` (proposed, not yet implemented) b) variables set by loading modules c) variables changed by unloading modules, especially the fake/real-module in the sanity check

That information might still be useful in limited contexts, e.g. during debugging strange failures, so logging it with `--debug` sounds reasonable as otherwise there won't be any way to show those.

But if those information is shown with `--debug` and the test reports use `--debug` we haven't gained anything for that common use case.    
How do we address this?

We could introduce debug levels, similar to how `-vv` is more verbose than `-v` in other programs.    
Or we could use `--debug --trace` as "log more debug info"

boegel: We have an extra level of logging (`--devel`), we could only log those details when devel log level is used?

boegel: BTW, we're sort of discussing this in the wrong place.

I'm about to merge this PR, and discussing stuff in closed PRs is something we should definitely avoid...

@Flamefire Can you open an issue to follow-up on this, beyond the changes made in this PR?

boegel (approved):
lgtm

PR #4943, opened 2025-06-28
retry fetching source file for extensions from PyPI with alternative filename (with `-` replaced with `_` in extension name)

smoors:
if downloading an extension fails, retry by replacing hyphen with underscore in the name: `abc-def-%(version).tar.gz` -> `abc_def-%(version).tar.gz`

only done if:
- it's an extension
- download from PyPI
- there's only 1 source url
- default filename is used (no `sources` or `source_tmpl` specified)

when this is merged, i would recommend to always use the exact name of the package listed on PyPI as the extension name:
- it ensures EB tries to download the 2 name formats if there are hyphens in the name
- users are more likely to search for this name when using e.g. `module spider <package>`
- it standardizes the naming of extensions in Lmod (`module spider`), avoiding that the same package is available with 2 different names, and thus easier to find all installed versions.

at some point we might consider switching to the underscore for the tarball on the first try, as that is the recommended name format, see
https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-name


easybuild/framework/easyblock.py
```diff
@@ -729,7 +733,17 @@ def collect_exts_file_info(self, fetch_files=True, verify_checksums=True):
                         if fetch_files:
                             src_path = self.obtain_file(src_fn, extension=True, urls=source_urls,
                                                         force_download=force_download,
-                                                        download_instructions=download_instructions)
+                                                        download_instructions=download_instructions,
+                                                        warning_only=warning_only)
+                            if not src_path:
```
Flamefire: I'd suggest to rename `warning_only` to something like `is_pypi_source` and change this to:
```suggestion
                            if not src_path and is_pypi_source:
```

That makes it clear when/why this is retried as the error if the file isn't found and we don't reach this code unless `is_pypi_source == True` is a bit hidden by `src_path = ...` (indicating it returns) especially with the (redundant?) check for `src_path` below

smoors: the check for src_path is indeed redundant, but i figured it might be good to keep it as a safety net for future changes to `obtain_file`, i don't know. happy to remove it if you prefer.

smoors: done in https://github.com/easybuilders/easybuild-framework/pull/4943/commits/04da4999ff4605cfc7c47d08bdd529b76e5722fa

Flamefire: I would remove it as it may lead to confusion here as it basically communicates that `None` (or empty) would be valid results of `obtain_file` which they aren't. Doesn't need to be done bow/in this PR though, I've just fallen into this "trap" during the review of this PR and had to check the code of `obtain_file` to verify.

jfgrimm (approved):
lgtm

Issue #4928, opened 2025-06-18
Wrong logic in `--sync-pr-with-develop` for fork repository name


Crivella:
Currently `--sync-pr-with-develop` relies on the `pr_target_repo` option to decide the repository name.
The same name is used bot for the the target repo and the source one.

This does not take into account that the repo opening the PR can have a different name from the original one.
See for example 

https://github.com/easybuilders/easybuild-easyconfigs/pull/22275

Where we have:

- easybuilders/easybuild-easyconfigs:develop
- julianmorillo/easybuild-easyconfigs-jmorillo:PR_BRANCH_NAME

Attempting to run the command without specifying `--pr-target-repo` will result in the wrong repo being accessed for downloading the PR branch

```
(easybuild-dev) crivella@crivella-desktop:~$ eb --sync-pr-with-develop 22275
== Temporary log file in case of crash /home/crivella/.local/easybuild/logs/easybuild-j262wq7z.log
== Determined branch name corresponding to easybuilders/easybuild-easyconfigs PR #22275: HepMC3
== fetching branch 'HepMC3' from https://github.com/julianmorillo/easybuild-easyconfigs.git...
Username for 'https://github.com':
```

While attempting to specify it will look for the PR on the specified repo which will not exist on the `easybuilders` side

```
(easybuild-dev) crivella@crivella-desktop:~$ eb --sync-pr-with-develop 22275 --pr-target-repo=easybuild-easyconfigs-jmorillo
== Temporary log file in case of crash /home/crivella/.local/easybuild/logs/easybuild-a3ba7_7u.log
ERROR: Failed to get data for PR #22275 from easybuilders/easybuild-easyconfigs-jmorillo
```

PR #4947, opened 2025-06-30
Fix logic for `pr_repo` used in `--sync-pr-with-develop`

Crivella:
Fixes
- #4928

Properly detect and use the name of the repo that is submitting the PR instead of using the same one that is receiving it

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -2190,11 +2190,12 @@ def det_account_branch_for_pr(pr_id, github_user=None, pr_target_repo=None):
 
     # branch that corresponds with PR is supplied in form <account>:<branch_label>
     account = pr_data['head']['label'].split(':')[0]
+    repo = pr_data['head']['repo']['name']
     branch = ':'.join(pr_data['head']['label'].split(':')[1:])
     github_target = '%s/%s' % (pr_target_account, pr_target_repo)
     print_msg("Determined branch name corresponding to %s PR #%s: %s" % (github_target, pr_id, branch), log=_log)
 
-    return account, branch
+    return account, repo, branch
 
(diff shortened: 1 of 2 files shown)


easybuild/tools/github.py
```diff
@@ -2190,11 +2190,12 @@ def det_account_branch_for_pr(pr_id, github_user=None, pr_target_repo=None):
 
     # branch that corresponds with PR is supplied in form <account>:<branch_label>
     account = pr_data['head']['label'].split(':')[0]
+    repo = pr_data['head']['repo']['name']
     branch = ':'.join(pr_data['head']['label'].split(':')[1:])
     github_target = '%s/%s' % (pr_target_account, pr_target_repo)
     print_msg("Determined branch name corresponding to %s PR #%s: %s" % (github_target, pr_id, branch), log=_log)
 
-    return account, branch
+    return account, repo, branch
```
boegel, line 2198: Hmm, this is a breaking change to the API...

How about introducing a `det_account_repo_branch_for_pr` (or a more sensible name, like `det_pr_target_info`), and deprecating `det_account_branch_for_pr`?

Crivella: This is only used in the 2 places modified by this PR, and nowhere else i could find, might be overkill to leave the function unused only in case some dependecies is using it?

Worst case i would have the old function call this one and return only `account, branch`

Crivella: Done in https://github.com/easybuilders/easybuild-framework/pull/4947/commits/0b9092a6f20903f6b9a3bfd3a18b46c4e36c45c5

boegel (approved):
lgtm

PR #4949, opened 2025-07-03
do not do `ext.toolchain.prepare` twice if the parallel extension install raises `NotImplementedError`

### LINKED ISSUE #4948: mpi4py/4.0.1-gompi-2024a build fails when --parallel-extensions-install is set, wrong value for MPI C compiler?
Since we enabled parallel extensions on our site, a few builds have been failing, `mpi4py-4.0.1-gompi-2024a.eb` being one example. I am aware there is no benefit in enabling parallel extensions for this easyconfig in particular (since it only has a single extension and it is Python), but @boegel said running `eb --parallel-extensions-install mpi4py-4.0.1-gompi-2024a.eb` should still work.

The shell command where it goes wrong is the following:
```
/apps/leuven/rocky8/skylake/2024a/software/Python/3.12.3-GCCcore-13.3.0/bin/python -m pip install --prefix=/data/leuven/337/vsc33716/Software/test_parallel_extensions/software/mpi4py/4.0.1-gompi-2024a  --no-deps --ignore-installed --no-build-isolation .'
```
Looking at the output from that command, the problem seems to be that the MPI C and MPI C++ compiler are set incorrectly:
```
      MPI configuration: [mpi] from 'mpi.cfg'
      MPI C compiler:    /apps/leuven/rocky8/skylake/2024a/software/OpenMPI/5.0.3-GCC-13.3.0/bin/mpicc mpicc
      MPI C++ compiler:  /apps/leuven/rocky8/skylake/2024a/software/OpenMPI/5.0.3-GCC-13.3.0/bin/mpicxx mpicxx
      checking for MPI compile and link ...
      /apps/leuven/rocky8/skylake/2024a/software/OpenMPI/5.0.3-GCC-13.3.0/bin/mpicc mpicc -fno-strict-overflow -DNDEBUG -g -O3 -Wall -O2 -ftree-vectorize -march=native -fno-math-errno -fPI      gcc: warning: gcc: linker input file unused because linking not done                                                                                                                        gcc: error: gcc: linker input file not found: No such file or directory
      gcc: warning: mpicc: linker input file unused because linking not done
      gcc: error: mpicc: linker input file not found: No such file or directory
      failure.
      removing: _configtest.c _configtest.o
      error: Cannot compile MPI programs. Check your configuration!!!
```
There is an additional `mpicc` added to `${EBROOTOPENMPI/bin/mpicc`, which makes it impossible to compile MPI programs. Any idea what is going wrong? The same build without `--parallel-extensions-install` works fine.

[easybuild-mpi4py-4.0.1-20250703.094440.fnfIq.log](https://github.com/user-attachments/files/21033609/easybuild-mpi4py-4.0.1-20250703.094440.fnfIq.log)

[out.txt](https://github.com/user-attachments/files/21033616/out.txt)

branfosj:
closes #4948

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2043,15 +2043,16 @@ class EasyBlock:
                 # If parallel extension install is not supported for this type of extension then install sequentially
                 msg = "Parallel extensions install not supported for %s - using sequential install" % self.name
                 self.log.info(msg)
-                self.install_extensions_sequential(install=install)
+                self.install_extensions_sequential(install=install, started_in_parallel=True)
         else:
             self.install_extensions_sequential(install=install)
 
-    def install_extensions_sequential(self, install=True):
+    def install_extensions_sequential(self, install=True, started_in_parallel=False):
         """
         Install extensions sequentially.
 
         :param install: actually install extensions, don't just prepare environment for installing
+        :param started_in_parallel: was install_extensions_parallel attempted before the sequential install
         """
         self.log.info("Installing extensions sequentially...")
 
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2037,10 +2037,16 @@ def install_all_extensions(self, install=True):
         self.log.debug("List of loaded modules: %s", self.modules_tool.list())
 
         if build_option('parallel_extensions_install'):
-            try:
+            # check to see if parallel extension install is supported for all extensions
+            # if it is not then we'll fallback to sequential install
+            all_exts_parallel = True
+            for ext in self.ext_instances:
+                if Extension.install_extension_async.__code__.co_code == ext.install_extension_async.__code__.co_code:
+                    all_exts_parallel = False
```
boegel, line 2048: As soon as one extension can't be installed in parallel, we don't need to check any further:

```suggestion
                    all_exts_parallel = False
                    break
```

boegel (changes requested):

easybuild/framework/easyblock.py
```diff
@@ -2039,10 +2039,17 @@ def install_all_extensions(self, install=True):
...
```
boegel, line 2048:
```suggestion
                    self.log.info(f"Extension {self.name} does not support being installed in parallel with other extensions")
                    all_exts_parallel = False
```

easybuild/framework/easyblock.py
```diff
@@ -2039,10 +2039,18 @@ def install_all_extensions(self, install=True):
         self.log.debug("List of loaded modules: %s", self.modules_tool.list())
 
         if build_option('parallel_extensions_install'):
-            try:
+            # check to see if parallel extension install is supported for all extensions
+            # if it is not then we'll fallback to sequential install
+            all_exts_parallel = True
+            for ext in self.ext_instances:
+                if Extension.install_extension_async.__code__.co_code == ext.install_extension_async.__code__.co_code:
+                    self.log.info(f"Extension {self.name} does not support being installed in parallel with other extensions")
```
```suggestion
                    self.log.info(f"Extension {ext.name} does not support asynchronous installation")

boegel (changes requested):

boegel (approved):
lgtm

testing with `matplotlib` (doesn't support installing in parallel) + `R` and `R-bundle-*` on top, just to make sure...

PR #4951, opened 2025-07-04
Ban `ligomp.so` at GCCcore level

### LINKED ISSUE #4535: The GCC OpenMP runtime (`libgomp`) should be a banned library for `GCCcore` (only)
There are different API possibilities for OpenMP and they shouldn't be mixed. For `GCCcore` level, we shouldn't allow the OpenMP runtime we create the possibility for strange behaviour if used as a dependency for an application that uses another runtime.

ocaisa:
Fixes #4535 

Will need work for easyconfigs that exist with GCCcore and use OpenMP (see https://github.com/easybuilders/easybuild-framework/issues/4535#issuecomment-3035326270)

### DIFF
--- a/easybuild/toolchains/gcccore.py
+++ b/easybuild/toolchains/gcccore.py
@@ -35,6 +35,8 @@ from easybuild.toolchains.compiler.gcc import Gcc
 from easybuild.tools import LooseVersion
 from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME
 
+# At GCCcore level we do not allow OpenMP
+GCCCORE_BANNED_LIBRARIES = ['libgomp.so']
 
 class GCCcore(Gcc):
     """Compiler-only toolchain, including only GCC and binutils."""
@@ -45,6 +47,15 @@ class GCCcore(Gcc):
     # GCCcore is only guaranteed to be present in recent toolchains
     # for old versions of some toolchains (GCC, intel) it is not part of the hierarchy and hence optional
     OPTIONAL = True
+
+    def banned_linked_shared_libs(self):
+        """
+        List of shared libraries (names, file names, paths) which are
+        not allowed to be linked in any installed binary/library.
+        """
+        banned_libs = super().banned_linked_shared_libs()
+
+        return banned_libs + GCCCORE_BANNED_LIBRARIES
 
     def is_deprecated(self):
         """Return whether or not this toolchain is deprecated."""
--- a/easybuild/toolchains/gcc.py
+++ b/easybuild/toolchains/gcc.py
@@ -31,7 +31,7 @@ Authors:
 """
 import re
 
-from easybuild.toolchains.gcccore import GCCcore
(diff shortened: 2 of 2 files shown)


easybuild/toolchains/gcccore.py
```diff
@@ -35,6 +35,9 @@
 from easybuild.tools import LooseVersion
 from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME
 
+# At GCCcore level we do not allow OpenMP
+GCCCORE_BANNED_LIBRARIES = ['libgomp.so']
```
branfosj, line 39: Has there been a discussion of what version of `GCCcore` we should add this check for? I do not think it is worthwhile to apply this to all `GCCcore` versions we have.

Probably start with one of:
* `14.2.0` (`2025a` generation)
* `14.3.0` (`2025b` generation)

ocaisa: That's true, I think it only really matters once we introduce modern LLVM (although it was probably also a potential issue for Intel compilers since we introduced `GCCcore`)

ocaisa:
For some reason this change now triggers
```
det_flexiblas_backend_libs()
```
which is causing the test failures (since the `flexiblas` command does not exist)

Micket:
```
== FAILED: Installation ended unsuccessfully: Sanity check failed: Check for banned/required shared libraries failed for
/home/ocaisa/eessi/versions/2023.06/software/linux/x86_64/intel/icelake/software/ImageMagick/7.1.1-34-GCCcore-13.2.0/bin/import,
```

This is way to strict. It's perfectly fine for binaries to link to libgomp (it's also fine if they are fortran codes).
I can't think of any reason to check any binary for any banned linking.

ocaisa:
> ```
> == FAILED: Installation ended unsuccessfully: Sanity check failed: Check for banned/required shared libraries failed for
> /home/ocaisa/eessi/versions/2023.06/software/linux/x86_64/intel/icelake/software/ImageMagick/7.1.1-34-GCCcore-13.2.0/bin/import,
> ```
> 
> This is way to strict. It's perfectly fine for binaries to link to libgomp (it's also fine if they are fortran codes). I can't think of any reason to check any binary for any banned linking.

It's not that I don't agree, but this would require a total rework of how the check is done in general in framework as there is no distinction made (and looks like you need additional requirements to check in python if something is an executable as opposed to a shared lib).

Thyre:
`gettext` uses OpenMP in `msgmerge` at one single place. One would need to check how large the impact of not using OpenMP would be. 

IIRC, UCX only uses OpenMP for a (performance) test. Could be fine to not build that test, or disable OpenMP for that...

Thyre:
Just thinking: I'm absolutely in favor of getting this forward. This might avoid headaches later on.
Some runtimes, e.g. NVHPC, now start to complain with a warning if they detect more than one runtime (during runtime).

However, we should ban more than `libgomp.so`. With LLVM, we could similarly link a `libomp.so`, or maybe even the Intel runtime (or a symlinked version of that). Listing all the names is not possible, but we should try to catch the most common ones.

PR #4919, opened 2025-06-12
add total iteration count to trace output

smoors:
also show the iteration number (starting at 1) instead of iteration index (starting at 0)

before:
```
== starting iteration #0 ...
== starting iteration #1 ...
== starting iteration #2 ...
```

after
```
== starting iteration 1/3 ...
== starting iteration 2/3 ...
== starting iteration 3/3 ...
```

PR #4952, opened 2025-07-04
do not call `det_iter_cnt()` in the `__init__`

branfosj:
#4919 calls `det_iter_cnt()` too early.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -312,7 +312,7 @@ class EasyBlock:
         self._init_log()
 
         # number of iterations
-        self.iter_cnt = self.det_iter_cnt()
+        self.iter_cnt = -1
 
         # try and use the specified group (if any)
         group_name = build_option('group')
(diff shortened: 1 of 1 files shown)


boegel:
Without this fix, the installation of `GCCcore` is broken, because not enough iterations are being done in the build.

More specifically, the additional build stages that the easyblock should do for the `nvptx` and `amdgcn` OpenMP offload targets were no longer being done, because the changes to `configopts` and `buildopts` that are done in `run_all_steps` in the custom easyblock for GCC were being done *after* `det_iter_cnt` was called.
That lead to some part of the GCC installation being missing, which was caught by the sanity check:
```
== FAILED: Installation ended unsuccessfully: Sanity check failed: no file found at 'bin/nvptx-none-as' in /tmp/software/GCCcore/13.3.0
```

@branfosj Thanks a lot for figuring this out!

boegel:
This bug also affects other software-specific easyblocks where we get creative with iterations, like `FFTW`, `GROMACS`, `ELPA`

PR #4958, opened 2025-07-09
Do not fail for read-only installation directory on `--module-only` (re)builds

Flamefire:
When only creating a module the installation directory must not be touched as it might be on a read-only file system.

Skip the code that copies reprod-files (easyconfig, easyblock, hooks) and log files.

This came up when trying to create a module-tree for EESSI reusing the installed software.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -5005,6 +5005,9 @@ def build_and_install_one(ecdict, init_env):
 
     hooks = load_hooks(build_option('hooks'))
 
+    # Don't touch installdir when doing a module-only build
+    module_only = build_option('module_only') or app.cfg['module_only']
+
     # build easyconfig
     error_msg = '(no error)'
     exit_code = None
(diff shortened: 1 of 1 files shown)


boegel:
Just producing a clear warning (when using `--module-only`) makes sense, I think

boegel (approved):
lgtm

easybuild/framework/easyblock.py
```diff
@@ -5061,7 +5064,7 @@ def build_and_install_one(ecdict, init_env):
     # make sure we're back in original directory before we finish up
     change_dir(cwd)
 
-    application_log = None
+    application_log = app.logfile
```
boegel, line 5067: This was a bit too naive, but it was easy to overlook that it could have consequences.

It introduced a regression, which was fixed in time for the EasyBuild v5.1.2 release, see:
- #5004

Flamefire: Thinking about how this happens. Previously `application_log` assignment was guarded by `if result and not dry_run:` with a fallback `if not result` otherwise it would never be assigned.

So I guess the extended-dry-run does set `app.logfile` (to what value actually and why?) and always returns success (`result = True`)

I'm wondering if extended-dry-run sets `app.logfile` to something that looks valid if we shouldn't handle it similar to other code that similarly would need to check that. Just to be sure we are not missing something similar somewhere else.

PR #4959, opened 2025-07-10
Keep list of patches unique before attempting a copy

ocaisa:
It was noted that for patches of extensions you can end up with multiple entries for a patch in the final list of patches. This change tries to make sure that each entry only appears once.

This actually can lead to a bug (observed when using `EESSI`) as the copy will fail the second time around when the target file is not writable (it was read only on the initial copy) _and_ came from a filesystem with extended attributes (like `/cvmfs`).

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -5128,7 +5128,9 @@ def build_and_install_one(ecdict, init_env):
                 repo.add_easyconfig(spec, app.name, det_full_ec_version(app.cfg), buildstats, currentbuildstats)
                 patches = app.patches
                 for ext in app.exts:
-                    patches += ext.get('patches', [])
+                    for patch in ext.get('patches', []):
+                        if patch not in patches:
+                            patches.append(patch)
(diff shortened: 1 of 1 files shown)


easybuild/framework/easyblock.py
```diff
@@ -5145,23 +5147,25 @@ def ensure_writable_log_dir(log_dir):
...
+                    for patch in ext.get('patches', []):
+                        if patch not in patches:
+                            patches.append(patch)
                 for patch in patches:
                     if 'path' in patch:
                         target = os.path.join(new_log_dir, os.path.basename(patch['path']))
                         copy_file(patch['path'], target)
                         _log.debug("Copied patch %s to %s", patch['path'], target)
 
+                application_log = os.path.join(new_log_dir, log_fn)
```
ocaisa, line 5166: log messages when trying to copy over files were disappearing into the abyss as the logs were already closed at this point. Moved the log copy to the last possible location

ocaisa:
Prefer the approach in #4960

Issue #4862, opened 2025-04-29
Trying to reproduce a stack based on the `reprod` folder fails for Bundle-type EasyConfigs


casparvl:
**Issue**

Rebuilding a `Bunlde` type EasyConfig from a reprod folder fails, e.g.

```
eb /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/OpenSSL/1.1/easybuild/reprod/OpenSSL-1.1.eb --module-only --force --installpath /tmp/casparl/eb-test-installpath/
== Temporary log file in case of crash /scratch-local/casparl.11439033/eb-t_zfg3xx/easybuild-0r740oaq.log
== found valid index for /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/icelake/software/EasyBuild/4.8.2/easybuild/easyconfigs, so using it...
== Running parse hook for OpenSSL-1.1.eb...
== processing EasyBuild easyconfig
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/OpenSSL/1.1/easybuild/reprod/OpenSSL-1.1.eb
ERROR: Failed to get application instance for OpenSSL (easyblock: EB_OpenSSL_wrapper): List of sources for bundle itself must be empty, found
[{'filename': 'openssl-1.1.1w.tar.gz', 'source_urls': ['https://www.openssl.org/source/']}]
```

A diff with the original OpenSSL from the same EasyBuild version shows:

```
$ diff /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/icelake/software/EasyBuild/4.8.2/easybuild/easyconfigs/o/OpenSSL/OpenSSL-1.1.eb /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/OpenSSL/1.1/easybuild/reprod/OpenSSL-1.1.eb
5d4
< minimum_openssl_version = '1.1.1'
12c11
< toolchain = SYSTEM
---
> toolchain = {'name': 'system', 'version': 'system'}
15c14,19
< osdependencies = ['perl']
---
> sources = [{'filename': '%(namelower)s-%(version)s.1w.tar.gz', 'source_urls': ['https://www.%(namelower)s.org/source/']}]
> checksums = ['cf3098950cb4d853ad95c0841f1f9c6d3dc102dccfcacd521d93925208b76ac8']
>
[...]

PR #4960, opened 2025-07-11
add `all_patches` variable to store all patches (incl the ones for extensions) that need to be copied to a repo/reprod dir

bedroge:
This also solves the duplication issue that https://github.com/easybuilders/easybuild-framework/pull/4959 tried to solve. 

Note that patches of components are added to `self.patches` by the Bundle easyblock: https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/bundle.py#L222. Hence, they will also be included in `self.all_patches`. 
Do note that, regardless of the change in this PR, the approach used by the Bundle may actually cause issues, because sources/patches may then be expected by the main easyblock itself, and they may also end up in the easyconfig stored in the  `reprod` folder (see https://github.com/easybuilders/easybuild-framework/issues/4862). However, I felt that solving that was out of scope for this PR, and could be handled in a follow-up PR.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -188,6 +188,7 @@ class EasyBlock:
 
         # list of patch/source files, along with checksums
         self.patches = []
+        self.all_patches = [] # also includes patches of extensions
         self.src = []
         self.data_src = []
         self.checksums = []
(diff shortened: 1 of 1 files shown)


ocaisa (approved):
This looks very clean now, if the existing tests (https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/toy_build.py#L1390) pass I think this is good to go

ocaisa (approved):
LGTM

PR #4961, opened 2025-07-17
New wrapped CLI using `click`

Crivella:
## Enabling

Using pip and optional python dependencies:

`pip install easybuild-framework[eb_click]`

## How it works

- Extract all the parameters from `EasyBuildOptions` and use them to generate a CLI with `click`
- Uses `click` + `rich_click` to have a fancier looking CLI
- `click` is only used for the nicer help and auto completion, but the parameters are still passed as is to the normal parser

### Autocomplete
From [click docs](https://click.palletsprojects.com/en/stable/shell-completion/#enabling-completion) add the following to the `.bashrc` or venv activation script.

```bash
eval "$(_EB_COMPLETE=bash_source eb)"
```

Alternatively see [aiida-core EC](https://github.com/easybuilders/easybuild-easyconfigs/blob/77a47ac1c8360c47465048d5517483137408017a/easybuild/easyconfigs/a/aiida-core/aiida-core-2.7.2-foss-2023a.eb#L108-L118) to activate it as a module

## Advantages

- Nicer looking
- Automatic integration and autocomplete with `bash`  `zsh`  `fish` shells

## TODO

- [x] Implement/convert more types for better autocompletion

## Maybe

- [x] Add autocomplete for path options (do not think there is a way that they are flagged as such in EB right now)
- [x] Better integrate logging with rich (moved in favor of https://github.com/easybuilders/easybuild-framework/pull/5125)
- [x] Rewrite the CLI logic to use `click` natively and divide the commands in subgroups to have an easier to read `help` (Discussion moved to https://github.com/easybuilders/easybuild-framework/issues/5122)

## How would this look like

<details>
<summary>eb --help</summary>
image

</details>

<details>
<summary>Error</summary>
image
</details>

<details>
<summary>Normal build run</summary>
image
</details>

<details>
<summary>Opt in/out</summary>

the new behavior will only be turned on if `click` is installed, otherwise the old one will be used.
image
</details>


boegel:
This looks kind of nice, but I'm not a big fan of introducing a required 3rd party dependency.

Is there any way we can support this opt-in, for example by defining an environment variable like `$EB_CLICK_CLI`?

Crivella:
We could make use of
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

probably not only for this but also for all the optional dependencies we specify in 
https://docs.easybuild.io/installation/#optional_python_packages

Then we add a check on top of `eb2` to try and import the required packages and print an error if they are not there suggesting to install the optional dependencies (if we keep `eb2` and `eb` separate, otherwise we could just fallback to the normal `eb` behavior if they are not there and potentially also allow to disable the behavior with the env variable)

Crivella:
@boegel 

So for now the `eb2` CLI is separate from the existing `eb` one.
I've added an optional set of install dependencies so that if one does `pip install easybuild-framework[eb2]` it also install the stuff required (and optional) for the new CLI.
The CLI will still be present if the dependencies are not there but will print an informative message instead of actually running

```
try:
    import click as original_click
except ImportError:
    def eb():
        """Placeholder function to inform the user that `click` is required."""
        print('Using `eb2` requires `click` to be installed.')
        print('Either use `eb` or install `click` with `pip install click`.')
        print('`eb2` also uses `rich` and `rich_click` as optional dependencies for enhanced CLI experience.')
        print('Exiting...')
        sys.exit(0)
```

The approach of using an ENV variable i think make sense if we decide to merge `eb` with `eb2` to only have one CLI with both behaviors, but it might be difficult to reconcile the autocomplete as they come from the environment

Crivella:
Example of a modified EC to test installing *easybuild with easybuild* with the new CLI

<details><summary>EasyConfig</summary>
<p>

```python
easyblock = 'EB_EasyBuildMeta'

name = 'EasyBuild'
version = '5.2.0'
version_suffix = '-dev'

homepage = 'https://easybuilders.github.io/easybuild'
description = """EasyBuild is a software build and installation framework
 written in Python that allows you to install software in a structured,
 repeatable and robust way."""

toolchain = SYSTEM

local_ef_commit = 'cf8bbb2e73d5468553ad83a53a52f9de8cbd6cdb'
local_eb_commit = 'develop'
local_ec_commit = 'develop'
sources = [
    {
        'filename': f'easybuild-framework-{local_ef_commit[:8]}.tar.xz',
        'git_config': {
            'url': 'https://github.com/Crivella',
            'repo_name': 'easybuild-framework',
            'commit': local_ef_commit,
        }
    },
    {
        'filename': f'easybuild-easyblocks-{local_eb_commit[:8]}.tar.xz',
        'git_config': {
            'url': 'https://github.com/easybuilders',
            'repo_name': 'easybuild-easyblocks',
            'commit': local_eb_commit,
        }
    },
    {
        'filename': f'easybuild-easyconfigs-{local_ec_commit[:8]}.tar.xz',
        'git_config': {
            'url': 'https://github.com/easybuilders',
            'repo_name': 'easybuild-easyconfigs',
            'commit': local_ec_commit,
        }
    }
]
checksums = []

click_autocomplete_bins = ['eb2']

# EasyBuild is a (set of) Python packages, so it depends on Python
# usually, we want to use the system Python, so no actual Python dependency is listed
allow_system_deps = [('Python', SYS_PYTHON_VERSION)]

local_pyshortver = '.'.join(SYS_PYTHON_VERSION.split('.')[:2])

sanity_check_paths = {
    'files': ['bin/eb'],
    'dirs': ['lib/python%s/site-packages' % local_pyshortver],
}

moduleclass = 'tools'

```

</p>
</details> 

Having the following installed on the system python (system- or user-wide) will enable:
- `click` -> the new CLI (for now actually required to pass the `click_autocomplete_bins` sanity check)
- `rich` -> new CLI with tracebacks
- `rich_click` -> new CLI with rich styling for help and errors

(Ideally this would be added as system packages?

Crivella:
The current version now does not add a new `eb2` CLI anymore but just replaces the `easybuild.main` inside the `eb` launcher with `easybuild.cli`

The function invoked there will either be a wrapper to the old one if `click` is not installed or the new `click`-wrapped CLI if click is installed.

Checks based on whether `rich` and `rich-click` are available are also added to ensure that their specific functionalities are only enabled if available


## NOTE

The multi-path autocompletion enabled by https://github.com/Crivella/easybuild-framework/blob/5439684362e867e9d78de889db1ac4aca2a52ca5/easybuild/cli/options/__init__.py#L62-L109 does not work in Bash when the path separator is`:`

boegel:
Did a high-level check and took it for a quick spin with the goal of merging this for the EasyBuild v5.3.0 release, but I'm afraid this needs more love...

1) Copyright/license headers are missing in the new files;

2) `eb --version` fails when `click` & co are available (this is a blocker):

    ```
    $ eb --version

     Usage: eb [OPTIONS] [OTHER_ARGS]...

     Try 'eb --help' for help
    ╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
    │ No such option: --version (Possible options: --software-version, --terse, --toolchain-version)                                                                                                                                              │
    ╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
    ```
    CI should have caught this... Can we enhance the existing `.github/workflows/end2end.yml` CI workflow to also runs those basic end-to-end tests in an environment where `click` is available?

3) There's no way to opt-out to the old style CLI when `click` is installed. In fact, I much prefer that the user has to do an explicit opt-in to this new (experimental?) CLI, for example by setting the `$EB_CLI_CLICK` environment variable.

4) We need an exit strategy for this, since we have plans for a serious overhaul of the `eb` CLI for EasyBuild 6.0 (ETA spring 2027)...

    What's the role of the `click`-based CLI going forward? Is it mostly a temporary "toy" to help with exploring out options for a more modern CLI?

I do like that the way in which this is implemented barely touches existing code, which makes me way more comfortable to go ahead and merge this as "alternative" CLI.

eb
```diff
@@ -44,7 +44,7 @@ trap keyboard_interrupt SIGINT
 # Python 3.6+ required
 REQ_MIN_PY3VER=6
 
-EASYBUILD_MAIN='easybuild.main'
+EASYBUILD_MAIN='easybuild.cli'
```
boegel: We should make this opt-in, and keep the default to `easybuild.main`:
```suggestion
if [ "${EB_CLI_CLICK}" == "1" ]; then
    # opt-in to using Click-based CLI (if click is available)
    EASYBUILD_MAIN='easybuild.cli'
else
    EASYBUILD_MAIN='easybuild.main'
fi
```

boegel (changes requested):

boegel:
With the above in mind, I need to lift this beyond the EasyBuild v5.3.0 release, so changing the milestone...

boegel:
@Crivella FYI, I've cancelled CI workflows here to give precedence to #5166

Crivella:
@boegel 

1. Done
2. Added it manually, it was not being picked up as in the normal parser it was not added as an `EasybuildOption` through `add_group_parser`.
3. Added the `EB_CLI_CLICK` env variable as suggested. Now the question is whether we want to hard fail if click is not available with `EB_CLI_CLICK=1` or revert to the old CLI like we do know (or a middle ground where we add a warning but still revert to the old CLI)
4. I would say it both puts in place the machinery to replace the current *wrapper* click CLI with a new independent one and also starts people to get used to the new auto-complete and colored CLI. The way we want to phase it out in favor of the new one will depend on how we decide to also play it with the old CLI (eg both in place with a deprecation warning).

Also modified the end2end tests to do
- Everything they already did
- Install deps for venv + create a virtualenv
- Install the checkout branch + the easyblock and configs downloaded for the previous test
- run the same tests + also a grep to check that `rich_click` is being used in `eb --help`
- One thing is whether we want to do the actual builds of `bzip2` in temporary dirs so we rebuild everything the second time or just run with a `--rebuild`

easybuild/cli/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+# #
+from easybuild.main import main_with_hooks
+
+try:
+    import click as original_click
+except ImportError:
+    def eb(*args, **kwargs):
+        """Placeholder function to inform the user that `click` is required."""
+        main_with_hooks()
```
boegel: I doesn't make much sense to me to specify that a Click-based CLI should be used, to then fall back to a non-Click CLI when `click` is not found...

If `click` is not found, then we should fail with an error (and print a hint about unsetting `$EB_CLI_CLICK` to opt-out of using a Click-based CLI

easybuild/cli/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+# #
```
boegel, line 24: @Crivella please add a docstring with some info on what this provides, add yourself as author

easybuild/cli/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+        from rich.traceback import install
+    except ImportError:
+        pass
+    else:
+        install(suppress=[
+            click, original_click
+        ])
+
+    from .options import EasyBuildCliOption, EasyconfigParam
+    from easybuild.tools.version import this_is_easybuild
```
boegel: Why not do these imports on top?

easybuild/cli/options/__init__.py
```diff
@@ -0,0 +1,337 @@
...
+        # If type is None assume type based on default value
+        elif self.type is None:
+            if self.default is False or self.default is True:
+                kwargs['is_flag'] = True
+                kwargs['type'] = click.BOOL
+                if self.default is True:
+                    decl = f"--{self.name}/--disable-{self.name}"
+            elif isinstance(self.default, (list, tuple)):
+                kwargs['multiple'] = True
+                kwargs['type'] = click.STRING
```
boegel, line 268: add `else` to avoid that we don't handle unknown option types?

easybuild/cli/options/__init__.py
```diff
@@ -0,0 +1,337 @@
...
+    description: str
+    type: str
+    action: str
+    default: Any
+    group: str = None
+    short: str = None
+    meta: Dict = None
+    lst: List = None
+
+    def __post_init__(self):
```
boegel, line 207: please add missing docstring

easybuild/cli/options/__init__.py
```diff
@@ -0,0 +1,337 @@
...
+        elif others == '':
+            start = self.delimiter
+        else:
+            start = f'{others}{self.delimiter}'
+        res = [CompletionItem(f"{start}{path}") for path in possibles]
+        # logging.warning(f"Shell completion for delimited path list: res={possibles}")
+        return res
+
+
+class DelimitedString(click.ParamType):
```
boegel, line 149: missing docstrings in methods below

easybuild/cli/options/__init__.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def add_group_parser(self, opt_dict, descr, *args, prefix='', **kwargs):
+        super().add_group_parser(opt_dict, descr, *args, prefix=prefix, **kwargs)
+        self._option_dicts[descr[0]] = (prefix, opt_dict)
+
+
+extracter = OptionExtracter(go_args=[])
+
+
+class DelimitedPathList(click.Path):
```
boegel, line 88: missing docstrings in methods below

easybuild/cli/options/__init__.py
```diff
@@ -0,0 +1,337 @@
...
+    'prefix',
+    'sourcepath-data',
+    'testoutput',
+    'tmp-logdir',
+    'tmpdir',
+
+    'buildpath',
+    'containerpath',
+    'installpath',
+    'sourcepath',
```
boegel, line 69: `bwrap_installpath` missing?

`packagepath`, `repositorypath` too?

Why not keep these alphabetically sorted?

Also, how will we keep this in sync as new options get added?

easybuild/cli/__init__.py
Crivella: Done

easybuild/cli/__init__.py
Crivella: Done

easybuild/cli/__init__.py
Crivella: The second one can be outside, the first i would leave inside so that the only check we need w.r.t. click is in this file.
Originally it would cause the fallback to the normal CLI to not work (since something was still trying to import click) now it would cause an `ModuleNotFound` error before our new raise that replaced the CLI fallback

easybuild/cli/options/__init__.py
Crivella: Had to check since it has been a while, added a print to check what this would hit:

```
Warning: Cannot infer type for option bwrap-installpath with default value /home/crivella/.local/easybuild/bwrap, treating as string
Warning: Cannot infer type for option module-naming-scheme with default value EasyBuildMNS, treating as string
Warning: Cannot infer type for option packagepath with default value /home/crivella/.local/easybuild/packages, treating as string
Warning: Cannot infer type for option subdir-data with default value data, treating as string
Warning: Cannot infer type for option subdir-modules with default value modules, treating as string
Warning: Cannot infer type for option subdir-software with default value software, treating as string
Warning: Cannot infer type for option subdir-user-modules with default value None, treating as string
Warning: Cannot infer type for option suffix-modules-path with default value all, treating as string
Warning: Cannot infer type for option container-image-name with default value None, treating as string
Warning: Cannot infer type for option container-tmpdir with default value None, treating as string
Warning: Cannot infer type for option create-index with default value None, treating as string
Warning: Cannot infer type for option dump-test-report with default value test_report.md, treating as string
Warning: Cannot infer type for option review-pr-filter with default value None, treating as string
Warning: Cannot infer type for option test-report-env-filter with default value None, treating as string
Warning: Cannot infer type for option dep-graph with default value None, treating as string
Warning: Cannot infer type for option search with default value None, treating as string
Warning: Cannot infer type for option search-filename with default value None, treating as string
Warning: Cannot infer type for option search-short with default value None, treating as string
Warning: Cannot infer type for option easystack with default value None, treating as string
Warning: Cannot infer type for option job-backend-config with default value None, treating as string
Warning: Cannot infer type for option job-output-dir with default value /home/crivella, treating as string
Warning: Cannot infer type for option job-target-resource with default value None, treating as string
Warning: Cannot infer type for option backup-modules with default value None, treating as string
Warning: Cannot infer type for option check-ebroot-env-vars with default value warn, treating as string
Warning: Cannot infer type for option deprecated with default value None, treating as string
Warning: Cannot infer type for option detect-loaded-modules with default value warn, treating as string
Warning: Cannot infer type for option easyblock with default value None, treating as string
Warning: Cannot infer type for option env-for-shebang with default value /usr/bin/env, treating as string
Warning: Cannot infer type for option group with default value None, treating as string
Warning: Cannot infer type for option http-header-fields-urlpat with default value None, treating as string
Warning: Cannot infer type for option minimal-build-env with default value CC:gcc,CXX:g++, treating as string
Warning: Cannot infer type for option module-cache-suffix with default value None, treating as string
Warning: Cannot infer type for option mpi-cmd-template with default value None, treating as string
Warning: Cannot infer type for option optarch with default value None, treating as string
Warning: Cannot infer type for option rpath-override-dirs with default value None, treating as string
Warning: Cannot infer type for option software-commit with default value None, treating as string
Warning: Cannot infer type for option sysroot with default value None, treating as string
Warning: Cannot infer type for option umask with default value None, treating as string
Warning: Cannot infer type for option zip-logs with default value gzip, treating as string
Warning: Cannot infer type for option package-tool with default value fpm, treating as string
Warning: Cannot infer type for option package-tool-options with default value , treating as string
Warning: Cannot infer type for option package-type with default value rpm, treating as string
Warning: Cannot infer type for option package-release with default value 1, treating as string
Warning: Cannot infer type for option aggregate-regtest with default value None, treating as string
Warning: Cannot infer type for option regtest-output-dir with default value None, treating as string
Warning: Cannot infer type for option amend with default value None, treating as string
Warning: Cannot infer type for option software-name with default value None, treating as string
Warning: Cannot infer type for option software-version with default value None, treating as string
Warning: Cannot infer type for option toolchain-name with default value None, treating as string
Warning: Cannot infer type for option toolchain-version with default value None, treating as string
Warning: Cannot infer type for option try-amend with default value None, treating as string
Warning: Cannot infer type for option try-software-name with default value None, treating as string
Warning: Cannot infer type for option try-software-version with default value None, treating as string
Warning: Cannot infer type for option try-toolchain-name with default value None, treating as string
Warning: Cannot infer type for option try-toolchain-version with default value None, treating as string
Warning: Cannot infer type for option unittest-file with default value None, treating as string
```

this are all variables with no specific type.
When we do not hit any of those if/else `type` will be left as None and click will treat the argument as a string eg:
image

This only come into play for the auto-completion (nothing will be suggested), not sure if it is better to enforce `click.STRING` as a type, but it is standard to not specify a type with click

easybuild/cli/options/__init__.py
Crivella: done

easybuild/cli/options/__init__.py
Crivella: done

easybuild/cli/options/__init__.py
Crivella: done

easybuild/cli/options/__init__.py
Crivella: This is for convenience to allow the autocompletion to treat something that should be a string for optparse as a path to a file/directory.

> Also, how will we keep this in sync as new options get added?

Right now there is no easy way to maintain this but adding stuff manually.
With a proper click CLI we will use the correct type from the outset

> Why not keep these alphabetically sorted?

I think i tried to divide them logically while i was manually adding them but fine also having them alphabetically sorted

Crivella: Also `repositorypath` is flagged as `strlist` and passes a list as a default so it is not compatible with `click.Path` the way we are setting it now
https://github.com/easybuilders/easybuild-framework/blob/daf0bac2ac2d98a95d6d6bbe74764932640e9079/easybuild/tools/options.py#L719-L722

PR #4965, opened 2025-07-30
add back `OrderedDict` import in (deprecated) `easybuild.tools.py2vs3.py3` module

Flamefire:
We still support that deprecated module and some older easyblocks do use it. E.g. the easybuildmeta easyblock in EESSI which fails with
>  cannot import name 'OrderedDict' from 'easybuild.tools.py2vs3'

### DIFF
--- a/easybuild/tools/py2vs3/py3.py
+++ b/easybuild/tools/py2vs3/py3.py
@@ -34,6 +34,7 @@ Authors:
 # these are not used here, but imported from here in other places
 import configparser  # noqa
 import urllib.request as std_urllib  # noqa
+from collections import OrderedDict  # noqa
 from collections.abc import Mapping  # noqa
 from functools import cmp_to_key
 from importlib.util import spec_from_file_location, module_from_spec


ocaisa (approved):
LGTM

boegel:
This fix makes sense, it should be merged as a bug fix (we shouldn't break the API even for deprecated modules like `easybuild.tools.py2v3.py3`), but I'm puzzled as to how/where you're hitting this...

This import was removed via https://github.com/easybuilders/easybuild-framework/pull/4607 (so included in EasyBuild 5.0.0), and the `EasyBuildMeta` easyblock was updated accordingly in https://github.com/easybuilders/easybuild-easyblocks/pull/2916 to stop assuming it's there.

Flamefire:
I'm rebuilding the modules of EESSI. Using the easyconfigs in the reprod folder failed because of changed or missing easyblocks. So I used the easyblocks from the reprod folders and hit this issue.

But then I hit an issue with a missing BLACS easyblock imported by that (older) version of the ScaLAPACK easyblock -.-

PR #4970, opened 2025-08-07
Fix RPATH wrapper by detecting & replacing `-Xlinker --enable-new-dtags`

### LINKED ISSUE #4969: RPATH wrapper does not replace `-Xlinker --enable-new-dtags` flags
I found out that the RPATH wrapper script does not detect the `-Xlinker --enabler-new-dtags` flag, which is an equivalent for `-Wl,--enable-new-dtags`:
https://github.com/easybuilders/easybuild-framework/blob/f8c94c316de871eab4959dbfaef69652b9173b7e/easybuild/scripts/rpath_args.py#L64-L68
https://github.com/easybuilders/easybuild-framework/blob/f8c94c316de871eab4959dbfaef69652b9173b7e/easybuild/scripts/rpath_args.py#L132-L141
hence something like:
```
icpx -Xlinker --enable-new-dtags -Wl,-rpath,$MKLROOT/lib/intel64 -o example example.cpp -lmkl_sequential -lmkl_core -lpthread -lm -ldl
```
will result in an executable with a RUNPATH defined instead of a RPATH.

This is particularly problematic as I discovered that the Intel MPI wrappers **always append** `-Xlinker --enable-new-dtags` to the compilation line. So even if EB's RPATH wrapper injects `--disable-new-dtags`, this flag would be overridden.
See the discussion/example here: [Intel MPI wrappers new-dtags - Intel Community](https://community.intel.com/t5/Intel-MPI-Library/Intel-MPI-wrappers-new-dtags/m-p/1707501#M12188)

lorisercole:
Fix #4969 RPATH wrapper: detect and replace `-Xlinker --enable-new-dtags`

The linker flag `-Xlinker --enable-new-dtags` was not detected by the `rpath_args.py` script, causing the use of RUNPATH instead of RPATH in some cases.
In particular, this happened with Intel MPI compiler wrappers (e.g. `mpiicpc`), that always add `--Xlinker --enable-new-dtags` to the compilation command.

I also added some minor improvements to the script.

### DIFF
--- a/easybuild/scripts/rpath_args.py
+++ b/easybuild/scripts/rpath_args.py
@@ -27,7 +27,9 @@
 Utility script used by RPATH wrapper script;
 output is statements that define the following environment variables
 * $CMD_ARGS: new list of command line arguments to pass
-* $RPATH_ARGS: command line option to specify list of paths to RPATH
+
+Usage:
+    rpath_args.py <cmd> <rpath_filter> <rpath_include> <args...>
 
 author: Kenneth Hoste (HPC-UGent)
 """
@@ -63,9 +65,9 @@ args = sys.argv[4:]
 
 # determine whether or not to use -Wl to pass options to the linker based on name of command
 if cmd in ['ld', 'ld.gold', 'ld.bfd']:
-    flag_prefix = ''
+    ldflag_prefix = ''
 else:
-    flag_prefix = '-Wl,'
+    ldflag_prefix = '-Wl,'
 
 rpath_filter = rpath_filter.split(',')
 if rpath_filter:
@@ -124,7 +126,7 @@ while idx < len(args):
             if is_new_existing_path(lib_path, rpath_lib_paths):
                 # inject -rpath flag in front for every -L with an absolute path,
                 rpath_lib_paths.append(lib_path)
-                cmd_args_rpath.append(flag_prefix + '-rpath=%s' % lib_path)
+                cmd_args_rpath.append(ldflag_prefix + '-rpath=%s' % lib_path)
 
         # always retain -L flag (without reordering!)
         cmd_args.append('-L%s' % lib_path)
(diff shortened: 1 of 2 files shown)


easybuild/scripts/rpath_args.py
```diff
@@ -38,22 +40,17 @@
...
-    if os.path.exists(new_path):
-        for path in paths:
-            if os.path.exists(path) and os.path.samefile(new_path, path):
-                res = False
-                break
-    else:
-        # path doesn't exist
-        res = False
+    for path in paths:
+        if os.path.samefile(new_path, path):
```
boegel: The `os.path.exists(path)` part should be retained, to avoid that `samefile` will fail hard:

```
>>> os.path.samefile('/tmp', '/loris')
  File "<python-input-2>", line 1, in <module>
    os.path.samefile('/tmp', '/loris')
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "<frozen genericpath>", line 137, in samefile
FileNotFoundError: [Errno 2] No such file or directory: '/loris'
```

```suggestion
        if os.path.exists(path) and os.path.samefile(new_path, path):
```

lorisercole: I removed it because I thought this check was redundant:
all the elements of the `paths` list (i.e. of `rpath_lib_paths`) were added and checked one-by-one before, so we can be sure they exist. This function is only called with `rpath_lib_paths` as second argument.
Does it make sense or am I missing something?

boegel: 1) We may be employing `is_new_existing_path` with other lists of paths in the future, so it's good to keep this separate function robust and make sure we only call `os.path.samefile` when we're sure that both paths exist.

2) We only check whether entries in `rpath_lib_paths` that is used when calling `is_new_existing_path` are i) non-empty, ii) are absolute paths, but we don't actually check whether they're an *existing* path.

So I think it's crucial to keep the `os.path.exists(path)` check, or we'll likely run into nasty surprises with non-existing paths leading to a hard crash when RPATHl linking is enabled in EasyBuild (which is the default since EasyBuild 5.0).

I don't think this check is redundant, and even if it is the tiny benefit of not doing this check doesn't outweigh the risk of running into a hard crash due to a `FileNotFoundError` being raised...

lorisercole: I did add the file-existence check in the function (line 46), so a `new_path` added to `paths` will be checked. And since `rpath_lib_paths` is built from scratch by calling this function every time, each new element will go through the existence check before being added.
But anyway, if we want a more robust function because we may reuse it somewhere else in the future, better to include this check then.

lorisercole:
Just thinking, perhaps we can extend this PR to also address this issue:
[easybuilders/easybuild-easyconfigs#23031 – RPATH wrapper with Intel compiler -c and -Werror flags leads to compilation errors](https://github.com/easybuilders/easybuild-easyconfigs/issues/23031)

As @Micket commented https://github.com/easybuilders/easybuild-easyconfigs/issues/23031#issuecomment-3076525353 , is there a reason to add RPATHs flags when the compiler is invoked with `-c`?

boegel (approved):
lgtm

PR #4972, opened 2025-08-09
Add support for "extensions" on Tcl modulefiles

xdelaruelle:
Module extensions are supported by Lmod in Tcl modulefiles since version 8.2.8. "extensions" command can be safely used in Tcl modulefiles with Environment Modules since version 5.1.0, even if it is ignored until version 5.6.0.

Check for support of module extensions is made within EasyBuild, thus "extensions" command is not included in generated Tcl modulefiles if the module tool detected by EasyBuild does not support it (Lmod<8.2.8 or EnvironmentModules<5.1.0).


boegel (approved):
lgtm

PR #4974, opened 2025-08-09
fix `test_module_extensions` failure

branfosj:
#4972 and #4971 combined broke the tests.

### DIFF
--- a/test/framework/module_generator.py
+++ b/test/framework/module_generator.py
@@ -847,10 +847,6 @@ class ModuleGeneratorTest(EnhancedTestCase):
         modgen = self.MODULE_GENERATOR_CLASS(eb)
         desc = modgen.get_description()
 
-        for pattern in patterns:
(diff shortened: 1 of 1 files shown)


branfosj:
I think this is right. There are 3 parts to this test:
1. Lines 798-824: `module_extensions = True` and there are extensions in the easyconfig
2. Lines 826-839: `module_extensions = True` and there are no extensions in the easyconfig
3. Lines 841-850: `module_extensions = False` and there are extensions in the easyconfig

For both 2 and 3 there should be no extensions in the module file and these can reuse the same `pattern`.

The last of these was more complicated - checking that none of the patterns were found - but that was more complicated than needed.

Micket (approved):
dunno how i managed to do it, but i managed to diff the PR the other way around, or i got it mixed up with the changes in the linked PRs. nevermind my previous comment.

boegel:
superseded by #4973, so closing this

PR #4975, opened 2025-08-11
Fix f-string in rpath check and Skip check for linked libs on symlinks

Flamefire:
In https://gist.github.com/boegelbot/1e87f59c5b2d74984a925e56a7adcb3d there is

> Failed to determine dynamically linked libraries for {path}, so skipping it in RPATH sanity check

There was a missed f-string causing this.

Additionally the `file` call (external tool) is unnecessary for symbolic links where it returns

>   symbolic link to libxxx

This will then not do any checks, so it can be skipped in a much cheaper way by using `islink`, This change doesn't change existing behavior it just makes it obvious.

This also means that `sanity_check_rpath` can/should be enhanced as the `is_parent_path` check for outside symlinks doesn't make sense if no check is done for any symlinks yielding an additional "Failed to determine dynamically linked libraries for..." message.

Shall we do that too?

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -1118,6 +1118,9 @@ def get_linked_libs_raw(path):
     or None for other types of files.
     """
 
+    if os.path.islink(path):
+        _log.debug(f"{path} is a symbolic link, so skipping check for linked libs")
+        return None
     res = run_shell_cmd("file %s" % path, fail_on_error=False, hidden=True, output_file=False, stream_output=False)
     if res.exit_code != EasyBuildExit.SUCCESS:
         fail_msg = "Failed to run 'file %s': %s" % (path, res.output)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3804,7 +3804,7 @@ class EasyBlock:
                     out = get_linked_libs_raw(path)
 
                     if out is None:
-                        msg = "Failed to determine dynamically linked libraries for {path}, "
+                        msg = f"Failed to determine dynamically linked libraries for {path}, "
                         msg += "so skipping it in RPATH sanity check"
                         self.log.debug(msg)
                     else:


easybuild/tools/systemtools.py
```diff
@@ -1118,6 +1118,9 @@ def get_linked_libs_raw(path):
     or None for other types of files.
     """
 
+    if os.path.islink(path):
+        _log.debug(f"{path} is a symbolic link, so skipping check for linked libs")
+        return None
```
boegel, line 1123: I think we should:
* resolve the symlink here so the `file` call below makes sense;
* do the skipping of symlinks where we're calling `get_linked_libs_raw` instead (since otherwise we're checking duplicate files);

If I call `get_linked_libs_raw` on any path that corresponds to a binary (either directly or via a symlink), I expect to get a lit of linked libraries, not `None` when it's a symlink.

boegel (changes requested):

easybuild/tools/systemtools.py
Flamefire: I didn't want to change existing behavior. Resolving the symlink here instead of at the call site might discover issues previously not seen and fail builds.

But yes skipping symlinks at the callsite makes sense. As written in the description we would then need(?) to remove the check for symlinks pointing outside the install directory as if we skip symlinks that doesn't make sense anymore.

Can do of course if the behavior change is acceptable.

boegel: We don't need to resolve the link here at all?

We just check whether the path is a symlink at the call site, rather than here.

If someone (in a different context) calls `get_linked_libs_raw` with a (single) path that corresponds to a symlink, they should should get a proper answer, not `None` if it happens to be a symlink (the function isn't called `get_linked_libs_raw_only_for_real_binaries_not_symlinks` ;) )

boegel: Hold on, I overlooked something...

Because of the `file` command that we run on the specified path, we're already getting `None` as result from calling `get_linked_libs_raw` on a symlinked path, so all you're doing here is adding a short-circuit.

Sorry for the confusion, ignore the above. :)

boegel (approved):
lgtm

PR #4977, opened 2025-08-12
fix passing command as list of strings to `run_shell_cmd`

boegel:
~Draft PR, because the test fails, signaling a problem with `run_shell_cmd`...~


smoors (approved):
lgtm

PR #4978, opened 2025-08-12
Check module extensions support within Tcl modulefiles

xdelaruelle:
Instead of making EasyBuild check if module extensions are supported by module tool to generate the proper Tcl modulefiles, move this check directly within modulefile code, like done with Lua modulefiles.

It means module extensions code is always written in modulefile, but a condition guards it, to execute this code only if the version of the module tool evaluating the modulefile supports it.

So instead of checking the version of the module tool used by EasyBuild, the check is made on the module tool that evaluates the generated modulefiles.

With this new approach, the "supports_extensions" ModuleTool state is now useless thus it is removed.

### DIFF
--- a/easybuild/tools/module_generator.py
+++ b/easybuild/tools/module_generator.py
@@ -895,10 +895,12 @@ class ModuleGeneratorTcl(ModuleGenerator):
             # - 'conflict Compiler/GCC/4.8.2/OpenMPI' for 'Compiler/GCC/4.8.2/OpenMPI/1.6.4'
             lines.extend(['', "conflict %s" % os.path.dirname(self.app.short_mod_name)])
 
-        if build_option('module_extensions'):
+        if build_option('module_extensions') and self.modules_tool.supports_extensions:
             extensions_list = self.app.make_extension_string(name_version_sep='/', ext_sep=' ')
-            if self.modules_tool.supports_extensions and extensions_list:
-                lines.extend(['', 'extensions %s' % extensions_list])
+            if extensions_list:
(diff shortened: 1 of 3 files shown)


easybuild/tools/module_generator.py
```diff
@@ -759,6 +760,34 @@ class ModuleGeneratorTcl(ModuleGenerator):
...
+        }
+
+        if isinstance(self.modules_tool, Lmod):
+            tool_version_var = "::env(LMOD_VERSION)"
+        # cannot test minimal version below 4.7.0 with Environment Modules
+        elif (isinstance(self.modules_tool, EnvironmentModules) and
+              LooseVersion(minimal_version) > LooseVersion('4.7.0')):
+            tool_version_var = "::ModuleToolVersion"
+        else:
+            raise NotImplementedError
```
boegel, line 783: This implies a breaking change for people still using Environment Modules 3.2.10, right?

Support for using that is deprecated (cfr. #4439), but maybe we shouldn't hard break it yet, and just return `None` in this case and deal with it properly below?

xdelaruelle: Though Environment Modules 3.2.10 was already totally unsupported by EB 5.

Maybe we should let `check_version` return the `NotImplementedError` exception, but avoid calling `check_version` if ModuleTool is not EnvironmentModules or Lmod?

boegel: That also works for me

xdelaruelle: done in 96be842

boegel (approved):
lgtm

PR #4980, opened 2025-08-13
Handle `post_install_patches` and `post_install_msgs` for extensions

Flamefire:
Those EasyConfig parameters were silently ignored. Run the appropriate steps in the default implementation of `post_install_extension` similar to postinstall_cmds and add a test. Using the new name for `post_install_patches` in the test reveals the need to handle that explicitely as a raw dict is used.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -48,6 +48,7 @@ import copy
 import functools
 import glob
 import inspect
+import itertools
 import json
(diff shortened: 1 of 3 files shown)


boegel (approved):
lgtm

Flamefire:
Followup for GH integration: https://github.com/easybuilders/easybuild-framework/pull/5016

PR #4984, opened 2025-08-18
introduce `PYTHON_EXE` variable in template for RPATH wrapper script (to easily replace name of `python` command to use in `buildenv` easyblock)

smoors:
we run easybuild in a venv, causing the wrappers contain the absolute path to the python executable in the venv, and thus is unusable for our users.

EDIT:
changed this PR to use a shell variable for the python executable, which makes it easier to substitute it with `python3` in the `buildenv` easyblock, see https://github.com/easybuilders/easybuild-easyblocks/pull/3910

### DIFF
--- a/easybuild/scripts/rpath_wrapper_template.sh.in
+++ b/easybuild/scripts/rpath_wrapper_template.sh.in
@@ -46,6 +46,8 @@ TOPDIR=`dirname $0`
 
 log "found CMD: $CMD | original command: %(orig_cmd)s | orig args: '$(echo \"$@\")'"
 
+PYTHON_EXE=%(python)s
+
 # rpath_args.py script spits out statement that defines $CMD_ARGS
 # options for 'python' command (see https://docs.python.org/3/using/cmdline.html#miscellaneous-options)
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm

boegel:
I'm OK with this in principle, but I do wonder if it can lead to surprises...

Maybe we should only use `python3` for the dumped RPATH wrapper scripts, and stick to using the path obtained from `sys.executable` during the EasyBuild session itself.

Would that make sense @smoors?

This is about the RPATH wrappers that are kept with `buildenv`, correct?

smoors:
> I'm OK with this in principle, but I do wonder if it can lead to surprises...
> 
> Maybe we should only use `python3` for the dumped RPATH wrapper scripts, and stick to using the path obtained from `sys.executable` during the EasyBuild session itself.
> 
> Would that make sense @smoors?
> 
> This is about the RPATH wrappers that are kept with `buildenv`, correct?

yes, that's correct. this PR only changes what's written in the RPATH wrapper scripts, it doesn't change anything else, exactly as you suggest.

PR #4988, opened 2025-08-27
Skip RPATH sanity check for symlinks explicitely

Flamefire:
`get_libs_raw` will return `None` anyway leading to a suboptimal warning.

Follow-up to https://github.com/easybuilders/easybuild-framework/pull/4975

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3791,12 +3791,10 @@ class EasyBlock:
                 self.log.debug(f"Sanity checking RPATH for files in {dirpath}")
 
                 for path in [os.path.join(dirpath, x) for x in os.listdir(dirpath)]:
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -3791,12 +3791,10 @@ def sanity_check_rpath(self, rpath_dirs=None, check_readelf_rpath=None):
                 self.log.debug(f"Sanity checking RPATH for files in {dirpath}")
 
                 for path in [os.path.join(dirpath, x) for x in os.listdir(dirpath)]:
-                    # skip the check for any symlinks that resolve to outside the installation directory
-                    if not is_parent_path(self.installdir, path):
+                    # skip the check for symlinks
```
```suggestion
                    # skip the check for symlinks (since get_linked_libs_raw will return None anyway)

boegel (changes requested):

PR #4991, opened 2025-08-31
keep `ModuleTool.mod_paths` in sync with `MODULEPATH` in `ModuleTool.load()`

### LINKED ISSUE #4986: Failing to build several HMNS modules with EnvironmentModules
Several EB files fails to build modules for EnvironmentModules in HMNS naming scheme. An environment conflict occurs between 2 versions of the `ncurses` module. The following error is obtained:

```
Module command '/usr/share/Modules/libexec/modulecmd.tcl python load GLib/2.77.1' failed with exit code 1;
stderr:
Loading ncurses/6.4
  ERROR: Conflicting ncurses/6.3 is loaded

Loading util-linux/2.39
  ERROR: Load of requirement ncurses/6.4 failed

Loading GLib/2.77.1
  ERROR: Load of requirement util-linux/2.39 failed
```

Such error happens with the following EB files for instance:

* AMOS-3.1.0-foss-2023a.eb
* cairomm-1.16.2-GCC-12.3.0.eb
* Circlator-1.5.5-foss-2023a.eb
* DjVuLibre-3.5.28-GCCcore-12.3.0.eb
* GConf-3.2.6-GCCcore-12.3.0.eb
* Geant4-11.2.2-GCC-12.3.0.eb
* GenomeTools-1.6.5-GCC-12.3.0.eb
* graphviz-python-0.20.1-GCCcore-12.3.0.eb
* gspell-1.12.2-GCCcore-12.3.0.eb
* GST-plugins-base-1.22.5-GCC-12.3.0.eb
* GStreamer-1.22.5-GCC-12.3.0.eb
* LibSoup-3.6.1-GCC-12.3.0.eb
* Octave-10.1.0-foss-2023a.eb
* OpenFOAM-5.0-20180606-foss-2023a.eb
* OpenFOAM-6-20190620-foss-2023a.eb
* OPERA-MS-0.9.0-20240703-foss-2023a.eb
* Qtconsole-5.5.1-GCCcore-12.3.0.eb
* QuPath-0.5.0-GCCcore-12.3.0-Java-17.eb
* R-tesseract-5.2.1-foss-2023a.eb
* texlive-20230313-GCC-12.3.0.eb

Command run to perform module build:

```
export EASYBUILD_MODULES_TOOL=EnvironmentModules
export EASYBUILD_MODULE_SYNTAX=Tcl
export TEST_EASYBUILD_MODULES_TOOL=${EASYBUILD_MODULES_TOOL}
export TEST_EASYBUILD_MODULE_SYNTAX="${EASYBUILD_MODULE_SYNTAX}"
eb --module-naming-scheme=HierarchicalMNS --skip-sanity-check --robot --rebuild --module-only OpenFOAM-6-20190620-foss-2023a.eb
```

A ticket has already been opened at easybuilders/easybuild-easyconfigs#23675 but I feel like this is more *framework-related*.

When looking in the eb log file for the module command run:

```
$ grep run.py /tmp/eb-1edy89w8/easybuild-GLib-2.77.1-20250820.144606.OQbdA.log | \
    grep 'Shell command' | sed 's/^.*libexec\///' | tail -n 10
modulecmd.tcl python --terse list
modulecmd.tcl python use /tmp/eb-1edy89w8/tmp4jxgp2be/all/Compiler/GCCcore/12.3.0
modulecmd.tcl python use /home/user/.local/easybuild/modules/all/Core
modulecmd.tcl python use /home/user/.local/easybuild/modules/all/Compiler/GCCcore/12.3.0
modulecmd.tcl python use /tmp/eb-1edy89w8/tmp4jxgp2be/all/Compiler/GCCcore/12.3.0
modulecmd.tcl python unuse /home/user/.local/easybuild/modules/all/Core
modulecmd.tcl python use /home/user/.local/easybuild/modules/all/Core
modulecmd.tcl python --terse list
modulecmd.tcl python load GCCcore/12.3.0
modulecmd.tcl python load GLib/2.77.1
```

We can observe that the `Core` modulepath is put on the front. `GLib/2.77.1` is requiring load of `gettext/0.21.1` and `util-linux/2.39` (among others):
* `util-linux/2.39` is available from `Compiler/GCCcore/12.3.0` modulepath and requires `ncurses/6.4`
* `gettext/0.21.1` is available from both `Core` and `Compiler/GCCcore/12.3.0`:
  * from `Compiler/GCCcore/12.3.0` it requires `ncurses/6.4`
  * from `Core` it requires ``ncurses/6.3`

Thus setting `Core` as the top priority modulepath, makes `gettext/0.21.1` load from `Core` and requires `ncurses/6.3`, which conflicts with `util-linux/2.39` that requires `ncurses/6.4`.

This error is not seen when Lmod is used as the ModuleTool. When loading `GCCcore/12.3.0`, Lmod puts the `Compiler/GCCcore/12.3.0` modulepath at the front. EnvironmentModules on its side, do not update `MODULEPATH` when loading `GCCcore/12.3.0` as `Compiler/GCCcore/12.3.0` is already seen enabled.

xdelaruelle:
Fix `ModuleTool.load()` to keep `ModuleTool.mod_paths` member in sync with `MODULEPATH` environment variable.
    
When purge mechanism was enabled in `load()`, `restore_env()` reset `MODULEPATH` but then `ModuleTool.check_module_path()` was called which set `MODULEPATH` back to the value kept in `mod_paths` member. This led to inconsistencies as loaded modules where purged but the `MODULEPATH` change made by a toolchain module was still there.
    
The proposed solution is to sync `ModuleTool.mod_paths` member with `MODULEPATH` right after environment reset. Thus `ModuleTool.set_mod_paths()` is called instead of `ModuleTool.check_mod_paths()`.
    
Such fix requires that expected `MODULEPATH` changes are specified through the `mod_paths` argument of `load()`. So this PR also contains commits that fix external `MODULEPATH` change to transmit such change via `mod_paths` argument of `load()`.
    
Fixes #4986
Fixes easybuilders/easybuild-easyconfigs#23675
Fixes easybuilders/easybuild-easyconfigs#17407
Fixes easybuilders/easybuild-easyconfigs#17368
Fixes easybuilders/easybuild-easyconfigs#23579

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1873,7 +1873,7 @@ class EasyBlock:
         fake_mod_path = self.make_module_step(fake=True)
 
         # load fake module
-        self.modules_tool.prepend_module_path(os.path.join(fake_mod_path, self.mod_subdir), priority=10000)
+        self.modules_tool.prepend_module_path(os.path.join(fake_mod_path, self.mod_subdir))
         self.load_module(purge=purge, extra_modules=extra_modules, verbose=verbose)
         return (fake_mod_path, env)
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1873,8 +1873,12 @@ class EasyBlock:
         fake_mod_path = self.make_module_step(fake=True)
         full_fake_mod_path = os.path.join(fake_mod_path, self.mod_subdir)
 
+        # indicate that path to fake module should get absolute priority
+        mod_paths = [(full_fake_mod_path, 10000)]
+
         # load fake module
-        self.load_module(purge=purge, extra_modules=extra_modules, mod_paths=[full_fake_mod_path], verbose=verbose)
+        self.load_module(purge=purge, extra_modules=extra_modules, mod_paths=mod_paths, verbose=verbose)
+
         return (fake_mod_path, env)
 
     def clean_up_fake_module(self, fake_mod_data):
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1871,10 +1871,10 @@ class EasyBlock:
 
         # create fake module
         fake_mod_path = self.make_module_step(fake=True)
+        full_fake_mod_path = os.path.join(fake_mod_path, self.mod_subdir)
 
         # load fake module
-        self.modules_tool.prepend_module_path(os.path.join(fake_mod_path, self.mod_subdir))
-        self.load_module(purge=purge, extra_modules=extra_modules, verbose=verbose)
+        self.load_module(purge=purge, extra_modules=extra_modules, mod_paths=[full_fake_mod_path], verbose=verbose)
         return (fake_mod_path, env)
 
     def clean_up_fake_module(self, fake_mod_data):
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1111,9 +1111,15 @@ class ModulesTool:
 
         # extend $MODULEPATH if needed
         for mod_path in mod_paths:
+            priority = None
+            if isinstance(mod_path, tuple) and len(mod_path) == 2:
+                mod_path, priority = mod_path
+            elif not isinstance(mod_path, str):
+                raise EasyBuildError(f"Incorrect mod_paths entry encountered when loading module(s): {mod_path}")
+
             full_mod_path = os.path.join(install_path('mod'), build_option('suffix_modules_path'), mod_path)
             if os.path.exists(full_mod_path):
-                self.prepend_module_path(full_mod_path)
+                self.prepend_module_path(full_mod_path, priority=priority)
 
         loaded_modules = self.loaded_modules()
         for mod in modules:
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1104,11 +1104,10 @@ class ModulesTool:
             # restore initial environment if provided
             if init_env is None:
                 raise EasyBuildError("Initial environment required when purging before loading, but not available")
-            else:
-                restore_env(init_env)
 
-            # make sure $MODULEPATH is set correctly after purging
-            self.check_module_path()
+            restore_env(init_env)
+            # sync mod_paths member with MODULEPATH environment variable
+            self.set_mod_paths()
 
         # extend $MODULEPATH if needed
         for mod_path in mod_paths:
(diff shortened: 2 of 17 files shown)


easybuild/framework/easyblock.py
```diff
@@ -1871,10 +1871,10 @@ def load_fake_module(self, purge=False, extra_modules=None, verbose=False):
 
         # create fake module
         fake_mod_path = self.make_module_step(fake=True)
+        full_fake_mod_path = os.path.join(fake_mod_path, self.mod_subdir)
 
         # load fake module
-        self.modules_tool.prepend_module_path(os.path.join(fake_mod_path, self.mod_subdir), priority=10000)
```
boegel, line 1876: We shouldn't just drop the `priority` part here, that's important when using Lmod at least, since that undoes a fix that was added in:
* #2506
to fix:
* #2499

boegel: @xdelaruelle fixed in https://github.com/xdelaruelle/easybuild-framework/pull/1

easybuild/tools/modules.py
```diff
@@ -1104,17 +1104,22 @@ def load(self, modules, mod_paths=None, purge=False, init_env=None, allow_reload
...
             if init_env is None:
                 raise EasyBuildError("Initial environment required when purging before loading, but not available")
-            else:
-                restore_env(init_env)
 
-            # make sure $MODULEPATH is set correctly after purging
-            self.check_module_path()
+            restore_env(init_env)
+            # sync mod_paths member with MODULEPATH environment variable
+            self.set_mod_paths()
```
boegel: `check_module_path` does more than `set_mod_paths` does, so I think we need to be careful that we're not introducing a regression here...

This mainly matters when the path in which EasyBuild is creating modules is not in the original `$MODULEPATH`.
That's fine, because EasyBuild will add this location to `$MODULEPATH` via the logic in `check_module_path` when `self.mod_paths` is set to `None`, after calling `set_mod_paths`.

So, a more correct fix would be to set `self.mod_paths` to `None`, and then call `self.check_module_paths`.

```suggestion
            self.mod_paths = None
            self.check_module_path()
```

boegel: fixed in:
- https://github.com/xdelaruelle/easybuild-framework/pull/2

I'm confident that this is the right way to fix this bug, also verified that `test_load_module` which was enhanced to capture this bug still passes when using `EnvironmentModules` as modules tool (which should be confirmed by CI)

boegel (approved):
I'm confident now that this is ready to merge.

@xdelaruelle Thanks a lot of the extensive effort on this!

PR #4992, opened 2025-09-03
also with test Python 3.14

boegel:
~~Final release of Python 3.14 is not there yet, so this is very drafty.~~

~~At least one test fails because the `rc.2` bit is unexpected, but that's mostly a non-issue (it'll disappear when we update this to use `3.14.0`)~~


boegel:
Python 3.14.0 has been tagged: https://github.com/python/cpython/releases/tag/v3.14.0

PR #5000, opened 2025-09-23
Add `return_conflicts` option to `check_conflicts` function and enhance tests

Flamefire:
This is motivated by the easyconfig tests where the result is hard to interpret:

> FAIL: test_conflicts (test.easyconfigs.easyconfigs.EasyConfigTest)
> Check whether any conflicts occur in software dependency graphs.
> ...
> AssertionError: True is not false : No conflicts detected

First the wording: It sounds like the error is that no conflicts are detected (similar in other places)

Second the actual conflicts are printed to stderr somewhere hidden above with other test output

The new parameter allows returning the text instead of printing it.

The PR where I've seen this was PyTorch: It has a build dependency on something which happens to be a runtime dependency of a dependency. So I added a test that this is detected but not if it is a build dependency for both where there would be no issue.

With that I ran into our somewhat strange dictionary duplicating a couple values that are easyconfig parameters, properties or functions: `{ 'ec': ec,  'spec': ec.path, 'short_mod_name': ec.short_mod_name, ...}`

As the robot-resolver uses that dict instead of an `EasyConfig` instance and I needed that in the test I factored out the creation of this dict into a new function. It is still strange... We should consider removing it or at least transforming it to a dataclass once we support only Python 3.7 to make it easier to reason about.

Should we report only primary dependents in the error message? See this confusing message:

> 	pybind11-2.13.6-GCC-13.3.0 as dep of: HiGHS-1.11.0-gfbf-2024a, **PyTorch-2.7.1-foss-2024a-CUDA-12.6.0**, Triton-3.1.0-foss-2024a-CUDA-12.6.0
> 	pybind11-2.12.0-GCC-13.3.0 as dep of: KMC-3.2.4-GCC-13.3.0, MACS3-3.0.3-gfbf-2024a, PyTorch-2.6.0-foss-2024a, **PyTorch-2.7.1-foss-2024a-CUDA-12.6.0**, SciPy-bundle-2024.05-gfbf-2024.05, SciPy-bundle-2024.05-gfbf-2024a, SignalP-6.0h-foss-2024a-fast, SnapATAC2-2.9.0-dev0-20250630-foss-2024a, TensorFlow-2.18.1-foss-2024a, Triton-3.3.1-foss-2024a-CUDA-12.6.0, bokeh-3.6.0-gfbf-2024a, dm-tree-0.1.9-gfbf-2024a, gemmi-0.7.1-GCC-13.3.0, heat-1.6.0-foss-2024a, hmmlearn-0.3.3-gfbf-2024a, matplotlib-3.9.2-gfbf-2024a, pybind11_abseil-202402.0-GCC-13.3.0, pyspoa-0.2.1-GCC-13.3.0, tensorstore-0.1.72-gfbf-2024a, torchvision-0.21.0-foss-2024a

I.e. it collects transitive dependents while collecting transitive dependencies in https://github.com/Flamefire/easybuild-framework/blob/9cd02c9ea03c5bdf278f0760453306e795737a5f/easybuild/tools/robot.py#L178-L181

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -48,6 +48,7 @@ import os
 import re
 from collections import OrderedDict
 from contextlib import contextmanager
+from typing import Optional
 
 import easybuild.tools.filetools as filetools
 from easybuild.base import fancylogger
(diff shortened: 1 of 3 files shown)


boegel (approved):
lgtm

PR #5001, opened 2025-09-24
prepare release notes for EasyBuild v5.1.2 + bump version to 5.1.2

boegel:


boegel:
Marked this as draft, since there's a problem with the `test_github_from_pr_x` test (which requires a GitHub token, so it's not run in CI for PRs), which I want to figure out and fix first...

PR #5010, opened 2025-09-29
Replace `assertErrorRegex` by standard `assertRaisesRegex`

Flamefire:
We no longer need this custom function with Python 2 support being dropped.

The only remaining use case was `EasyBuildError` where `str(e)` didn't return the error as a string but as an escaped string fixed since https://github.com/easybuilders/easybuild-framework/pull/5009

This makes it easier to use, especially for new devs being used to the `unittest` module and conventions for other functions like `self.assertRaises`.

Additionally this can be used as a context manager:
```
with self.assertRaises(TypeError): # similar for assertRaisesRegex
  do_this()
  do_that()
```


easybuild/tools/build_log.py
```diff
@@ -139,7 +139,7 @@ def __init__(self, msg, *args, exit_code=EasyBuildExit.ERROR, **kwargs):
 
     def __str__(self):
         """Return string representation of this EasyBuildError instance."""
-        return repr(self.msg)
```
boegel: This includes the changes from #5009 ?

Flamefire: Indeed. Otherwise tests fail due to escaping as we had special handling for that, see `convert_exception_to_str`

Flamefire:
Rebased only

Flamefire:
@boegel This looks big but except for the wrapper function and fixed indents is just a simple string search & replace to call the same function with the standard name without any other changes so I'd say this should go in in one go.

easybuild/base/testing.py
```diff
@@ -141,42 +141,9 @@ def setUp(self):
         self.orig_sys_stderr = sys.stderr
         self.stdout_std_err_buffers = []
 
-    def convert_exception_to_str(self, err):
```
boegel, line 143: We shouldn't just remove this, some people may be importing it in their own scripts...

At best, we can deprecate it, and remove it in EasyBuild 6.0

boegel (changes requested):

easybuild/base/testing.py
Flamefire: It was inside a module for testing, so I thought it won't be used outside our testing.
But sure, readded with deprecation

PR #5012, opened 2025-10-02
Honor `sanity_check_commands` & `sanity_check_paths` from extensions

Flamefire:
Currently those are silently ignored leading to success where it should fail.

The only official easyconfig affected by this so far is `MDAnalysis-2.4.2-foss-2021a.eb` but I ran into this while creating an easyconfig

Includes
- [x] https://github.com/easybuilders/easybuild-framework/pull/5015

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -4660,6 +4660,7 @@ class EasyBlock:
                     else:
                         print_warning(mod_files_found_msg)
 
+            # Do not do those checks for extensions, only in the main easyconfig
             if self.toolchain.use_rpath:
                 rpath_fails = self.sanity_check_rpath()
                 if rpath_fails:
@@ -4688,13 +4689,13 @@ class EasyBlock:
             self.fake_mod_data = None
 
         # pass or fail
-        if self.sanity_check_fail_msgs:
+        if not self.sanity_check_fail_msgs:
+            self.log.debug("Sanity check passed!")
+        elif not self.is_extension:
             raise EasyBuildError(
                 "Sanity check failed: " + '\n'.join(self.sanity_check_fail_msgs),
                 exit_code=EasyBuildExit.FAIL_SANITY_CHECK,
             )
-
-        self.log.debug("Sanity check passed!")
 
     def _set_module_as_default(self, fake=False):
         """
--- a/easybuild/framework/extensioneasyblock.py
+++ b/easybuild/framework/extensioneasyblock.py
@@ -72,8 +72,6 @@ class ExtensionEasyBlock(EasyBlock, Extension):
     def __init__(self, *args, **kwargs):
         """Initialize either as EasyBlock or as Extension."""
 
-        self.is_extension = False
-
         if isinstance(args[0], EasyBlock):
             # make sure that extra custom easyconfig parameters are known
             extra_params = self.__class__.extra_options()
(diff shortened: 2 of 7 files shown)


easybuild/framework/easyblock.py
```diff
@@ -3363,6 +3363,16 @@ def post_processing_step(self):
...
         """Decide whether to run the dry-run or the real version of the sanity-check step"""
+        if 'extension' in kwargs:
+            extension = kwargs.pop('extension')
+            self.log.deprecated(
+                "Passing `extension` to `sanity_check_step` is no longer necessary and will be ignored "
+                f"(Easyblock: {self.__class__.__name__}).",
+                '6.0',
+            )
+            if extension != self.is_extension:
+                raise EasyBuildError('Unexpected value for `extension` argument. '
```
boegel, line 3622: This isn't exactly ignoring it, as is stated above...

What's the point of raising an *error* for an incorrect value being passed that's going to be ignored?!

Flamefire: Because if we have **different** values it is an error somewhere.

So a) it doesn't *need* to be passed but b) if it is it should match the expected value

We can remove that check in EB 6 but I'd better play it safe until then

easybuild/framework/extension.py
```diff
@@ -330,9 +328,6 @@ def sanity_check_step(self):
                     fail_msg = 'command "%s" failed' % cmd
                 fail_msg += "; output:\n%s" % cmd_res.output.strip()
                 self.log.warning("Sanity check for '%s' extension failed: %s", self.name, fail_msg)
-                res = (False, fail_msg)
                 # keep track of all reasons of failure
                 # (only relevant when this extension is installed stand-alone via ExtensionEasyBlock)
                 self.sanity_check_fail_msgs.append(fail_msg)
-
-        return res
```
boegel: Drive-by partial review: why is it OK to change this, and make `Extension.sanity_check_step` not return anything anymore? This is an API change...

easybuild/framework/easyblock.py
boegel: But why care if the value is different, if we're going to ignore the value anyway?

I sort of see your point, but we can't log a deprecation message claiming we stop caring about `extension`, only to raise an error based on its value immediately afterwards...

Flamefire: > we can't log a deprecation message claiming we stop caring about `extension`, only to raise an error based on its value immediately afterwards...

Because these are 2 separate checks:
1. Passing` extension` is allowed, but **deprecated**
2. Passing a wrong `extension` is an **error**

What would you suggest for the wording? Just remove "and will be ignored"?

easybuild/framework/extension.py
Flamefire: I misunderstood part of the EasyBlock code because `ExtensionEasyBlock` got complicated as we now call `sanity_check_step` from `Extension` *and* `EasyBlock` and the most straight forward way to combine those was to just check `sanity_check_fail_msgs` because `EasyBlock.sanity_check_step` doesn't return anything.

But in any case `Extension.sanity_check_step` and hence `ExtensionEasyBlock` needs to return this tuple (or actually `True/False` would be enough)

boegel:
@Flamefire merge conflict to resolve here...

PR #5014, opened 2025-10-04
Added command line parameter `--fetch-all`

Rommelse:
Added command line parameter --fetch-all. It implies --fetch. But when a fetch fails, EasyBuild will still try to fetch all dependencies. This is handy in situations where the success of fetching is non-deterministic and the engineer wants to get all failures at once without having to restart after each failure. Non-deterministic success of fetching can for example happen if some downloads are intercepted by the CyberDefence team to scan for vulnerabilities and later retries are handled by earlier scans that have been cached.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -831,8 +831,9 @@ class EasyBlock:
                     git_config=None, no_download=False, download_instructions=None, alt_location=None,
                     warning_only=False):
         try:
-            return self.obtain_file_and_fail(filename, extension, urls, download_filename, force_download, git_config, no_download, download_instructions, alt_location, warning_only)
-        except Exception as e:
+            return self.obtain_file_and_fail(filename, extension, urls, download_filename, force_download, git_config,
+                                             no_download, download_instructions, alt_location, warning_only)
+        except Exception:
             if build_option('fetch_continue'):
                 if not urls:
                     urls = ['NO_URL']
@@ -843,8 +844,8 @@ class EasyBlock:
 
     @_obtain_file_update_progress_bar_on_return
     def obtain_file_and_fail(self, filename, extension=False, urls=None, download_filename=None, force_download=False,
-                    git_config=None, no_download=False, download_instructions=None, alt_location=None,
-                    warning_only=False):
+                             git_config=None, no_download=False, download_instructions=None, alt_location=None,
+                             warning_only=False):
         """
         Locate the file with the given name
         - searches in different subdirectories of source path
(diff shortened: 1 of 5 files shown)


easybuild/tools/options.py
```diff
@@ -464,6 +464,7 @@ def override_options(self):
                                           False),
             'fetch': ("Allow downloading sources ignoring OS and modules tool dependencies, "
                       "implies --stop=fetch, --ignore-osdeps and ignore modules tool", None, 'store_true', False),
+            'fetch-continue': ("Continue after failed fetch", None, 'store_true', False),
```
boegel: Help text should make it clear that this implies `--fetch`, and I would rename this to `--fetch-all` maybe?

```suggestion
            'fetch-all': ("Download sources (like --fetch), don't stop when downloading failed for one easyconfig",
                          None, 'store_true', False),

File: easybuild/framework/easyblock.py [hunk 2]
```diff
@@ -830,6 +830,22 @@ def collect_exts_file_info(self, fetch_files=True, verify_checksums=True):
...
+            if build_option('fetch_continue'):
+                if not urls:
+                    urls = ['NO_URL']
+                print_warning(f"FAILED: File {filename} not found from {urls[0]}. Continuing ...")
+                return 'NO_FILE_FOUND'
+            else:
+                raise
+
+    @_obtain_file_update_progress_bar_on_return
+    def obtain_file_and_fail(self, filename, extension=False, urls=None, download_filename=None, force_download=False,
```
boegel: method is a bit confusing maybe, how about:
```suggestion
    def obtain_file_raise_on_failure(self, filename, extension=False, urls=None, download_filename=None, force_download=False,

boegel (changes requested):

Rommelse:
Hi @boegel . Quite a while ago, I implemented the suggestions that came from your review. But then I might have forgotten to click the "Resolved" button in the PR in the webui. I see the milestone gets keeping postponed from 5.2.0 to 5.2.1 to 5.x. 
Is that because I forgot the "Resolved" button? Or is there another reason?

I clicked the Resolved button today. I first sync-ed the latest changes from the develop branch into this PR branch. I tested the new "--fetch-all" option again. Still work as expected, especially also in combi with --downoad-timeout.

Is there anything more I can do from my side to get this PR merged onto the develop branch?

boegel:
> Hi @boegel . Quite a while ago, I implemented the suggestions that came from your review. But then I might have forgotten to click the "Resolved" button in the PR in the webui. I see the milestone gets keeping postponed from 5.2.0 to 5.2.1 to 5.x. Is that because I forgot the "Resolved" button? Or is there another reason?
> 
> I clicked the Resolved button today. I first sync-ed the latest changes from the develop branch into this PR branch. I tested the new "--fetch-all" option again. Still work as expected, especially also in combi with --downoad-timeout.
> 
> Is there anything more I can do from my side to get this PR merged onto the develop branch?

@Rommelse I think this mostly went "out of view", I'll try and take another look at it and see if we can get it merged...

boegel (approved):
lgtm

PR #5015, opened 2025-10-06
Introduce `saved_env` context manager for restoring environment in tests

Flamefire:
Many tests modify the environment in some steps and need to restore it afterwards.
To avoid missing this the contextmanager does that automatically which avoids repeating the logic in various tests

I [recently](https://github.com/easybuilders/easybuild-framework/actions/runs/18198148973/job/51809839117) ran into this issue where a test failed with a misleading error.

Broader question: Should `EasyBlock.run_all_steps` restore the environment? When it exits it will have e.g. toolchain modules still loaded.   
We currently save the environment in `build_and_install_software` and restore it in `build_and_install_one` **before** processing/calling `run_all_steps`, although that additionally catches changes done by hooks. However the changes done to the environment when loading the hooks are only in effect for the first installation as then the hooks are cached. So I think we can ignore that case.

### DIFF
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -29,7 +29,6 @@ Toy build unit test
 @author: Kenneth Hoste (Ghent University)
 @author: Damian Alvarez (Forschungszentrum Juelich GmbH)
 """
-import copy
 import glob
 import grp
 import os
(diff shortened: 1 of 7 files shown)


test/framework/toy_build.py
```diff
@@ -3540,9 +3531,6 @@ def assert_cuda_report(missing_cc, additional_cc, missing_ptx, log, stdout=None,
         self.assertTrue(expected_result.search(outtxt), msg)
         assert_cuda_report(missing_cc=0, additional_cc=0, missing_ptx=0, log=outtxt, stdout=stdout, num_checked=0)
 
-        # Restore original environment
-        modify_env(os.environ, start_env, verbose=False)
```
boegel, line 3544: @Flamefire You're not introducing the use of the `saved_env` context manager in this test, so won't that result in leaking changes made to their environment from this test to other tests?

See the use of `setvar` in this test...

Flamefire: The teardown of EnhancedTestcase resets the environment for every test:
```
        # restore original environment
        modify_env(os.environ, self.orig_environ, verbose=False)
```

So we don't need to do this at the end of the test.

`saved_env` is only required if we need to restore the environment within a test.

boegel: Right, thanks for clarifying!

boegel (approved):
lgtm

PR #5017, opened 2025-10-07
Take into account alternative name `post_install_patches` when checking checksums

Flamefire:
Extensions/components use a dict for "options" that is passed to this function and won't automatically replace the new name to the old one.

Do this manually.


boegel (approved):
lgtm

PR #5019, opened 2025-10-07
Add decorator to ignore PR test failures due to rate limit

Flamefire:
The decorator checks if the test is run in a GitHub PR and ignores any error caused by a "HTTP 403" rate-limit exception.

### DIFF
--- a/test/framework/github.py
+++ b/test/framework/github.py
@@ -78,16 +78,26 @@ def requires_github_access():
 
     Useful when the test uses e.g. `git` commands to download from Github and would run into rate limits
     """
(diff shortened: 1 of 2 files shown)


lexming (approved):
LGTM as there is no other solution as long as EB relies on unauthenticated access to github.

I'll wait for feedback from other maintainers before merging.

Flamefire:
@boegel wrote in Slack:

> Seems sensible, but we have to be careful that this won’t accidentally hide failing things that we can/should test…

My answer:
> It is only skipped in PRs, so at least it will turn up post-merge for the unlikely case the rate limit hid the actual error

PR #5021, opened 2025-10-08
Fix `check_checksums` when using templates, patch-dicts or `checksums.json`

Flamefire:
This is a similar fix as https://github.com/easybuilders/easybuild-easyconfigs/pull/15973

`check_checksums` had various shortcomings, some touched by https://github.com/easybuilders/easybuild-framework/commit/7112e406dc8ee75fd3d5257648d23d801d77e813 (from @boegel ) which disabled templating entirely when getting the `exts_list` list.

This breaks when `checksums.json` is used: It will report missing checksums even when they are available in the checksums.json   
Additionally templates in the name/version of extension won't be resolved causing the error message harder to understand

A quick fix was to replace the above commit by `allow_unresolved_templates`-decorator to allow partially resolved extensions but don't fail as before when templates are used e.g. in `install_cmd` which fixes the issues with the name/version of the extension

However that resolves too deep as inside the extension options different template values must be used.

We have `collect_exts_file_info` which does handle all those peculiarities of the extension specs already so just use that.

This also handles the default file name of extension sources and `nosource` which we can rely on here if it fills `sources` correctly. This requires  #4054 (merged here so that tests pass)

It also fixes a failure when patches were specified as dicts as In `sources` the keyname is "filename" while in `patches` it is "name" which causes a `KeyError` in `check_checksums_for`

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -43,8 +43,10 @@ Authors:
 * Jan Andre Reuter (Juelich Supercomputing Centre)
 * Jasper Grimm (UoY)
 * Alex Domingo (Vrije Universiteit Brussel)
+* Alexander Grund (TU Dresden)
 """
 import concurrent
+import contextlib
 import copy
 import functools
 import glob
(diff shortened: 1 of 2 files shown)


boegel (approved):
lgtm

PR #5022, opened 2025-10-09
Add `--keep-going` option to fail at the end not at first failing installation + use proper exit code as determined by `main` function

Flamefire:
Allows to install any possible easyconfig without stopping if a single one fails

The main function will call `sys.exit` with the returned exit code as-if failing by an `EasyBuildError` so `eb --keep-going ec1.eb ec2-eb` will return an error code that can be used in scripts

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -795,6 +795,9 @@ def prepare_main(args=None, logfile=None, testing=None):
     :param testing: enable testing mode
     :return: 3-tuple with initial session state data, EasyBuildOptions instance, and tuple with configuration settings
     """
+    # set $___EASYBUILD___ environment variable to 'EasyBuild' to indicate that we're in an EasyBuild session
+    os.environ['___EASYBUILD___'] = 'EasyBuild'
+
     register_lock_cleanup_signal_handlers()
 
     # if $CDPATH is set, unset it, it'll only cause trouble...
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -33,6 +33,7 @@ Authors:
 * Pieter De Baets (Ghent University)
 * Jens Timmerman (Ghent University)
 """
+import inspect
 import logging
 import os
 import re
@@ -162,8 +163,7 @@ class EasyBuildLog(fancylogger.FancyLogger):
 
     def caller_info(self):
         """Return string with caller info."""
-        # findCaller returns a 3-tupe in Python 2, a 4-tuple in Python 3 (stack info as extra element)
-        (filepath, line, function_name) = self.findCaller()[:3]
+        filepath, line, function_name = self.findCaller()[:3]
         filepath_dirs = filepath.split(os.path.sep)
 
         for dirName in copy(filepath_dirs):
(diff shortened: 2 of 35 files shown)


easybuild/tools/build_log.py
```diff
@@ -397,9 +397,9 @@ def print_error(msg, *args, **kwargs):
     if exitCode is not None:
         _init_easybuildlog.deprecated("'exitCode' option in print_error function is replaced with 'exit_code'", '6.0')
 
-    # use 1 as defaut exit code
+    # default exit code
```
Thyre: I think we should keep the message that we exit with an error code by default.
Something like: `use ERROR as default exit code, if none is set`?

test/framework/options.py
```diff
@@ -6734,6 +6734,43 @@ def test_sanity_check_only(self):
         import easybuild.easyblocks.generic.toy_extension
         reload(easybuild.easyblocks.generic.toy_extension)
 
+    def test_keep_going(self):
+        """Test use of --keep-goin."""
```
Thyre:
```suggestion
        """Test use of --keep-going."""
```

easybuild/tools/build_log.py
Flamefire: I don't like including values in comments. The value is literally 2 lines down, so the only important information is that the check is to set the default, isn't it?

Thyre: I agree that the code explains this quite well already, though I would even leave out the comment entirely then.

Flamefire: Fully agreed.

```suggestion
```

easybuild/main.py
```diff
@@ -820,7 +826,7 @@ def main_with_hooks(args=None):
     hooks = load_hooks(eb_go.options.hooks)
 
     try:
-        main(args=args, prepared_cfg_data=(init_session_state, eb_go, cfg_settings))
+        sys.exit(int(main(args=args, prepared_cfg_data=(init_session_state, eb_go, cfg_settings))))
```
boegel: That's a bit too heavy on the brackets (it's Python, not Lisp ;) )

```suggestion
        exit_code = main(args=args, prepared_cfg_data=(init_session_state, eb_go, cfg_settings))
        sys.exit(int(exit_code))
```

easybuild/main.py
```diff
@@ -773,15 +775,19 @@ def main(args=None, logfile=None, do_build=None, testing=False, modtool=None, pr
...
+                                        hooks, do_build)
 
     # stop logging and cleanup tmp log file, unless one build failed (individual logs are located in eb_tmpdir)
     stop_logging(logfile, logtostdout=options.logtostdout)
-    if do_cleanup:
+    if is_successful:
         cleanup(logfile, eb_tmpdir, testing, silent=options.terse)
+    if options.dump_test_report or options.upload_test_report:
+        # Generation test reports is successful even when software failed to build
+        return EasyBuildExit.SUCCESS
```
boegel: Hmm, not sure about this...

Exit code should always indicate whether (all) installations were successful or not, regardless of whether a test report was uploaded/dumped? That seems like the least surprising behavior to me...

Flamefire: Shouldn't the exit code indicate an error in the requested operation? And a failed installation while explicitly testing for those isn't an error.

So it's a definition issue and I'm not fully sure how to solve it, so either way is fine to me. Hence I'm changing this to an error as suggested

easybuild/main.py
```diff
@@ -820,7 +823,8 @@ def main_with_hooks(args=None):
     hooks = load_hooks(eb_go.options.hooks)
 
     try:
-        main(args=args, prepared_cfg_data=(init_session_state, eb_go, cfg_settings))
+        exit_code: EasyBuildExit = main(args=args, prepared_cfg_data=(init_session_state, eb_go, cfg_settings))
+        sys.exit(init(exit_code))
```
Thyre:
```suggestion
        sys.exit(int(exit_code))
```

Probably a typo?

boegel:
@Flamefire merge conflict to fix...

PR #5025, opened 2025-10-11
Add templates for source with just version numbers (like `VERSION_TAR_GZ` and `V_VERSION_TAR_GZ`)

Micket:
Adding templates for the common `%(version)s.tar.gz` and  `v%(version)s.tar.gz`  sources.

E.g.
1. `VERSION_TAR_GZ`
2. `VVERSION_ZIP`

I didn't know a better option but to append an extra V for the second option.

I haven't tested anything yet, throwing it out there to gauge the response first.

### DIFF
--- a/easybuild/framework/easyconfig/templates.py
+++ b/easybuild/framework/easyconfig/templates.py
@@ -271,7 +271,7 @@ for ext in EXTENSIONS:
         'SOURCE_%s' % suffix: ('%(name)s-%(version)s.' + ext, "Source .%s bundle" % ext),
         'SOURCELOWER_%s' % suffix: ('%(namelower)s-%(version)s.' + ext, "Source .%s bundle with lowercase name" % ext),
         'VERSION_%s' % suffix: ('%(version)s.' + ext, "Source filename <version>.%s common at GitHub" % ext),
-        'VVERSION_%s' % suffix: ('v%(version)s.' + ext, "Source filename v<version>.%s common at GitHub" % ext),
+        'V_VERSION_%s' % suffix: ('v%(version)s.' + ext, "Source filename v<version>.%s common at GitHub" % ext),
     })
 
 for pyver in ('py2.py3', 'py2', 'py3'):


easybuild/framework/easyconfig/templates.py
```diff
@@ -271,6 +271,11 @@
         'SOURCE_%s' % suffix: ('%(name)s-%(version)s.' + ext, "Source .%s bundle" % ext),
         'SOURCELOWER_%s' % suffix: ('%(namelower)s-%(version)s.' + ext, "Source .%s bundle with lowercase name" % ext),
     })
+    TEMPLATE_CONSTANTS.update({
```
Flamefire: No need to make this 2 operations which might slow down EB startup even further.
```suggestion
```

Played around with the prior code how to do this with a single update call and dict comprehension (fastest in CPython) but not sure if people find this easier to read ;-)
```
TEMPLATE_CONSTANTS.update({
    key: value
    for ext in EXTENSIONS
    for key, value in (
        (
            f"SOURCE_{ext.replace('.', '_').upper()}",
            (f"%(name)s-%(version)s.{ext}", f"Source .{ext} bundle"),
        ),
        (
            f"SOURCELOWER_{ext.replace('.', '_').upper()}",
            (f"%(namelower)s-%(version)s.{ext}", f"Source .{ext} bundle with lowercase name"),
        ),
    )
})
```

easybuild/framework/easyconfig/templates.py
```diff
@@ -271,6 +271,11 @@
...
+        'VERSION_%s' % suffix: ('%(version)s.' + ext, "Source filename <version>.%s common at github" % ext),
+        'VVERSION_%s' % suffix: ('v%(version)s.' + ext, "Source filename v<version>.%s common at github" % ext),
```
Flamefire: I think we should be the official spelling GitHub
```suggestion
        'VERSION_%s' % suffix: ('%(version)s.' + ext, "Source filename <version>.%s common at GitHub" % ext),
        'VVERSION_%s' % suffix: ('v%(version)s.' + ext, "Source filename v<version>.%s common at GitHub" % ext),
```

easybuild/framework/easyconfig/templates.py
boegel: oh my please no

When we get to the point where updates to a trivial `dict` value or the bottleneck for EasyBuild startup, we can revisit this... ;)

boegel: Just add the lines to the existing `update` statement:
```suggestion
```

easybuild/framework/easyconfig/templates.py
```diff
@@ -270,7 +270,10 @@
     TEMPLATE_CONSTANTS.update({
         'SOURCE_%s' % suffix: ('%(name)s-%(version)s.' + ext, "Source .%s bundle" % ext),
         'SOURCELOWER_%s' % suffix: ('%(namelower)s-%(version)s.' + ext, "Source .%s bundle with lowercase name" % ext),
+        'VERSION_%s' % suffix: ('%(version)s.' + ext, "Source filename <version>.%s common at GitHub" % ext),
+        'VVERSION_%s' % suffix: ('v%(version)s.' + ext, "Source filename v<version>.%s common at GitHub" % ext),
```
boegel: I agree with @Flamefire here that an extra `_` helps a lot here:
```suggestion
        'V_VERSION_%s' % suffix: ('v%(version)s.' + ext, "Source filename v<version>.%s common at GitHub" % ext),

PR #5029, opened 2025-10-20
Pass correct `-march` option on RISC-V systems when `optarch` toolchain option is enabled

julianmorillo:
'-march=native" is not a valid value when building in RISC-V. 
Pass the ISA string with the supported extensions of the target platform instead.

### DIFF
--- a/easybuild/toolchains/compiler/gcc.py
+++ b/easybuild/toolchains/compiler/gcc.py
@@ -94,14 +94,17 @@ class Gcc(Compiler):
         # no support for -march on POWER; implies -mtune=native
         (systemtools.POWER, systemtools.POWER): '-mcpu=native',
         (systemtools.POWER, systemtools.POWER_LE): '-mcpu=native',
+        (systemtools.X86_64, systemtools.AMD): '-march=native',  # implies -mtune=native
+        (systemtools.X86_64, systemtools.INTEL): '-march=native',  # implies -mtune=native
+    }
+    if systemtools.get_cpu_family() == systemtools.RISCV:
         # documentation about the RISC-V ISA field and GCC's -march option:
         # https://github.com/riscv-non-isa/riscv-toolchain-conventions/blob/main/src/toolchain-conventions.adoc
         # https://docs.riscv.org/reference/isa/unpriv/naming.html
         # https://gcc.gnu.org/gcc-14/changes.html
-        (systemtools.RISCV64, systemtools.RISCV): '-march=' + systemtools.get_isa_riscv(),  # use all extensions
-        (systemtools.X86_64, systemtools.AMD): '-march=native',  # implies -mtune=native
-        (systemtools.X86_64, systemtools.INTEL): '-march=native',  # implies -mtune=native
-    }
+        march_opt = '-march=' + systemtools.get_isa_riscv()
+        COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(systemtools.RISCV64, systemtools.RISCV)] = march_opt
+
     # used with --optarch=GENERIC
     COMPILER_GENERIC_OPTION = {
         (systemtools.AARCH32, systemtools.ARM): '-mcpu=generic-armv7',  # implies -march=armv7 and -mtune=generic-armv7
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -629,8 +629,11 @@ def get_cpu_features():
 
 def get_isa_riscv():
     """
-    Get supported ISA string
+    Get supported ISA string on RISC-V Linux systems
     """
+    if get_cpu_family() != RISCV:
+        return None
+
     # 'rv64imafdc identifies the baseline RISC-V CPU architecture we are targetting.
     # - rv64: 64-bit RISC-V
     # - i: Integer instruction set (mandatory base ISA)
@@ -644,6 +647,7 @@ def get_isa_riscv():
     # provides all the capabilities expected by modern toolchains and runtimes,
     # and ensures broad compatibility and good performance.
     isa_string = 'rv64imafdc'
+    _log.debug(f"Default ISA string: {isa_string}")
     os_type = get_os_type()
     if os_type == LINUX:
         if is_readable(PROC_CPUINFO_FP):
@@ -653,13 +657,16 @@ def get_isa_riscv():
             res = isa_regex.search(proc_cpuinfo)
             if res:
                 isa_string = res.group(1)
-                _log.debug("Found ISA string using regex '%s': %s", isa_regex.pattern, isa_string)
+                _log.debug(f"Found ISA string using regex '{isa_regex.pattern}': {isa_string}")
+                # sort parts separated by underscore alphabetically, as required by older GCC versions (< 14)
+                isa_string = '_'.join(sorted(isa_string.split('_')))
+                _log.debug(f"Sorted ISA string {isa_string}")
             else:
-                _log.debug("Failed to determine ISA string from %s", PROC_CPUINFO_FP)
+                _log.debug(f"Failed to determine ISA string from {PROC_CPUINFO_FP}")
         else:
-            _log.debug("%s not found to determine ISA string", PROC_CPUINFO_FP)
+            _log.debug(f"{PROC_CPUINFO_FP} not found to determine ISA string")
     else:
-        _log.debug("Could not determine ISA string (OS: %s), defaulting to: %s", os_type, isa_string)
+        _log.debug(f"Could not determine ISA string (OS: {os_type}), defaulting to {isa_string}")
     return isa_string
 
 
--- a/test/framework/systemtools.py
+++ b/test/framework/systemtools.py
@@ -43,7 +43,7 @@ from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.environment import setvar
 from easybuild.tools.filetools import adjust_permissions, mkdir, read_file, symlink, which, write_file
 from easybuild.tools.run import RunShellCmdResult, run_shell_cmd
-from easybuild.tools.systemtools import CPU_ARCHITECTURES, AARCH32, AARCH64, POWER, X86_64
+from easybuild.tools.systemtools import CPU_ARCHITECTURES, AARCH32, AARCH64, POWER, RISCV, X86_64
 from easybuild.tools.systemtools import CPU_FAMILIES, POWER_LE, DARWIN, LINUX, UNKNOWN
 from easybuild.tools.systemtools import CPU_VENDORS, AMD, APM, ARM, CAVIUM, IBM, INTEL
 from easybuild.tools.systemtools import MAX_FREQ_FP, PROC_CPUINFO_FP, PROC_MEMINFO_FP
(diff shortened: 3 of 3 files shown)


easybuild/tools/systemtools.py
```diff
@@ -627,6 +627,27 @@ def get_cpu_features():
...
 
+def get_isa():
+    """
+    Get supported ISA string
+    """
+    isa_string = None
+    os_type = get_os_type()
+    if os_type == LINUX:
+        if is_readable(PROC_CPUINFO_FP):
+            cmd = "cat /proc/cpuinfo | grep isa | head -1 | cut -d':' -f2"
```
boegel: Rather than relying on a pipeline, can't we just read `/proc/cpuinfo` as a file (with `read_file`), and then use a regular expression to grab the `isa` part?

Also, `grep isa` is a bit too loose (on an x86_64 system, it matches `misalignsse` in `flags`, for example).

How portable is this across CPU families?
If it's not, we should make this function specific to RISC-V, and return `None` on non-RISC-V systems?

We should also add some tests when introducing a new function like this...

julianmorillo: Added new commits that try to address all the comments.

easybuild/tools/systemtools.py
```diff
@@ -627,6 +627,30 @@ def get_cpu_features():
...
+    """
+    Get supported ISA string
+    """
+    isa_string = ''
+    os_type = get_os_type()
+    if os_type == LINUX:
+        if is_readable(PROC_CPUINFO_FP):
+            _log.debug("Trying to determine ISA string on Linux via %s", PROC_CPUINFO_FP)
+            proc_cpuinfo = read_file(PROC_CPUINFO_FP)
+            isa_regex = re.compile(r"isa\s*:\s*(.*)")
```
Crivella: If `isa` is always at the start of the string we could do this

```suggestion
            isa_regex = re.compile(r"^isa\s*:\s*(.*)")
```

or alternatively change `re.search` to `re.match`

to address https://github.com/easybuilders/easybuild-framework/pull/5029#discussion_r2452251793

> Also, grep isa is a bit too loose (on an x86_64 system, it matches misalignsse in flags, for example).

easybuild/tools/systemtools.py
Crivella: Should probably be risc-v specific (changing the function name to something like `get_isa_riscv`)

Atleast on my intel CPU the closest thing would be the `flags : ...`


<details><summary>Details</summary>
<p>

```
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 183
model name      : 13th Gen Intel(R) Core(TM) i9-13900K
stepping        : 1
microcode       : 0x12f
cpu MHz         : 1053.547
cache size      : 36864 KB
physical id     : 0
siblings        : 32
core id         : 0
cpu cores       : 24
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 32
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpu
id aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb ssbd ibr
s ibpb stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid rdseed adx smap clflushopt clwb intel_pt sha_ni xsaveopt xsavec xgetbv1 xsaves split_lock_detect user_shstk avx_vnni dtherm i
da arat pln pts hwp hwp_notify hwp_act_window hwp_epp hwp_pkg_req hfi vnmi umip pku ospke waitpkg gfni vaes vpclmulqdq tme rdpid movdiri movdir64b fsrm md_clear serialize pconfig arch_lbr ibt flush_l1d arch_capabilities
vmx flags       : vnmi preemption_timer posted_intr invvpid ept_x_only ept_ad ept_1gb flexpriority apicv tsc_offset vtpr mtf vapic ept vpid unrestricted_guest vapic_reg vid ple shadow_vmcs ept_mode_based_exec tsc_scaling usr_wait_pause
bugs            : spectre_v1 spectre_v2 spec_store_bypass swapgs eibrs_pbrsb rfds bhi
bogomips        : 5990.40
clflush size    : 64
cache_alignment : 64
address sizes   : 46 bits physical, 48 bits virtual
power management:
```

</p>
</details>

julianmorillo: Agreed. Done.

easybuild/tools/systemtools.py
```diff
@@ -627,6 +627,30 @@ def get_cpu_features():
...
+            if res:
+                isa_string = res.group(1)
+                _log.debug("Found ISA string using regex '%s': %s", isa_regex.pattern, isa_string)
+            else:
+                _log.debug("Failed to determine ISA string from %s", PROC_CPUINFO_FP)
+        else:
+            _log.debug("%s not found to determine ISA string", PROC_CPUINFO_FP)
+    else:
+        raise SystemToolsException("Could not determine ISA string (OS: %s)" % os_type)
+    return isa_string
```
Crivella, line 670: I think raising an exception here will cause EB to fail on any non-LINUX systems, as the function calling this needs to always be resolved.

Maybe a cleaner solution would be to evaluate this at run-time instead of at the class declaration

Also i am wondering if defaulting to an empty string is the right thing to do.
This would cause `-march=` to be added causing some obfuscated failures while compiling.
Would it be better to fallback to something like the generic of `rv64gc`? The alternative would be to give a clear failure (but also would need a solution for the aforementioned problem of this function being called in the class declaration)

julianmorillo: Oh, you are right regarding the exception. I changed it by a `_log.debug` call. 
As now the function will resolve anyway to `rv64gc` as fallback, I think it is fine to evaluate this at this point (mostly to be consistent with the rest of architectures)

easybuild/tools/systemtools.py
```diff
@@ -627,6 +627,30 @@ def get_cpu_features():
     return cpu_feat
 
 
+def get_isa_riscv():
+    """
+    Get supported ISA string
+    """
+    isa_string = 'rv64imafdc'
```
boegel, line 649: @julianmorillo Please add a comment to clarify what this value is/means, and why it's a sensible default?

test/framework/systemtools.py
```diff
@@ -60,6 +61,31 @@
 
 PROC_CPUINFO_TXT = None
 
+PROC_CPUINFO_TXT_RISCV64 = """processor       : 0
+hart            : 2
+isa             : rv64imafdc
```
boegel: @julianmorillo Can we come up with a better test example, since this is exactly the same case as the default value used by `get_isa_riscv`, so if that for some reason doesn't go through the logic at all and just returns the default, it'll be overlooked by `test_isa_linux`...

easybuild/toolchains/compiler/gcc.py
```diff
@@ -94,6 +94,7 @@ class Gcc(Compiler):
         # no support for -march on POWER; implies -mtune=native
         (systemtools.POWER, systemtools.POWER): '-mcpu=native',
         (systemtools.POWER, systemtools.POWER_LE): '-mcpu=native',
+        (systemtools.RISCV64, systemtools.RISCV): '-march=' + systemtools.get_isa_riscv(),  # use all extensions
```
boegel: @julianmorillo How sure are we that the `isa` field in `/proc/cpuinfo` (which is what `get_isa_riscv` grabs) will always correspond to what GCC's `-march` option accepts as possible value?
Can we point to some GCC docs on this or something?

https://gcc.gnu.org/onlinedocs/gcc-15.2.0/gcc/RISC-V-Options.html mentions `-march=help`, maybe we should leverage that somehow (or at least mention it here)?

boegel (changes requested):

easybuild/tools/systemtools.py
julianmorillo: Done!

test/framework/systemtools.py
julianmorillo: Done!

easybuild/toolchains/compiler/gcc.py
julianmorillo: @boegel,
Regarding your first question, unfortunately, there is no certainty of that. Said that, it should usually work in most of the cases for the Base ISA (i.e. rv64imafdc, stable and long-standing) and for standard and Z-extensions (although this would depend on how old/new is the GCC version). A problem may appear with specific vendor/custom extensions that GCC used version does not (yet) support.  

Regarding your second question documentation on this can be found here https://github.com/riscv-non-isa/riscv-toolchain-conventions/blob/main/src/toolchain-conventions.adoc, and here https://docs.riscv.org/reference/isa/unpriv/naming.html

Regarding your last question, I didn't know about that `-march=help` option, thanks for pointing it out. This option exists only from GCC version 14 (see https://gcc.gnu.org/gcc-14/changes.html), so no way to leverage that for older GCC versions.

Said all of that, if it serves for something, I did an exhaustive testing in the RISC-V platforms I have available:
```
jmorillo@arriesgado-1:~/TEST$ gcc --version
gcc (Ubuntu 10.3.0-1ubuntu1) 10.3.0

jmorillo@arriesgado-1:~/TEST$ cat /proc/cpuinfo
processor       : 0
hart            : 2
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,u74-mc

jmorillo@arriesgado-1:~/TEST$ gcc -march=rv64imafdc helloworld.c
jmorillo@arriesgado-1:~/TEST$
```
```
jmorillo@pioneer-4:~/TEST$ gcc --version
gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

jmorillo@pioneer-4:~/TEST$ cat /proc/cpuinfo
processor       : 0
hart            : 1
isa             : rv64imafdcv
mmu             : sv39
mvendorid       : 0x5b7
marchid         : 0x0
mimpid          : 0x0

jmorillo@pioneer-4:~/TEST$ gcc -march=rv64imafdcv helloworld.c
jmorillo@pioneer-4:~/TEST$
```
```
jmorillo@bananaf3-3:~/TEST$ gcc --version
gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0

jmorillo@bananaf3-3:~/TEST$ cat /proc/cpuinfo
processor       : 0
hart            : 0
model name      : Spacemit(R) X60
isa             : rv64imafdcv_sscofpmf_sstc_svpbmt_zicbom_zicboz_zicbop_zihintpause
mmu             : sv39
mvendorid       : 0x710
marchid         : 0x8000000058000001
mimpid          : 0x1000000049772200

jmorillo@bananaf3-3:~/TEST$ gcc -march=rv64imafdcv_sscofpmf_sstc_svpbmt_zicbom_zicboz_zicbop_zihintpause helloworld.c
jmorillo@bananaf3-3:~/TEST$
```
```
jmorillo@premier-2:~/TEST$ gcc --version
gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0

jmorillo@premier-2:~/TEST$ cat /proc/cpuinfo
processor       : 0
hart            : 1
isa             : rv64imafdch_zicsr_zifencei_zba_zbb_sscofpmf
mmu             : sv48
uarch           : sifive,p550
mvendorid       : 0x489
marchid         : 0x8000000000000008
mimpid          : 0x6220425

jmorillo@premier-2:~/TEST$ gcc -march=rv64imafdch_zicsr_zifencei_zba_zbb_sscofpmf helloworld.c
gcc: error: ‘-march=rv64imafdch_zicsr_zifencei_zba_zbb_sscofpmf’: unexpected ISA string at end: ‘sscofpmf’
```
The only problem is on the premier board but it can be easily fixed putting `sscofpmf` at the beginning of the extensions list:
```
jmorillo@premier-2:~/TEST$ gcc -march=rv64imafdch_sscofpmf_zicsr_zifencei_zba_zbb helloworld.c
jmorillo@premier-2:~/TEST$
```
More than that, this problem, is already solved in GCC 14 (see https://gcc.gnu.org/gcc-14/changes.html): `The -march= option no longer requires the architecture string to be in canonical order, with only a few constraints remaining: the architecture string must start with rv[32|64][i|g|e], and must use an underscore as the separator after a multi-letter extension.`

What do you think? Is it worth implementing the logic for ordering the extensions just for old versions of GCC? (Apart from adding the links with the documentation to the code)

easybuild/tools/systemtools.py
boegel: FYI: I changed `get_isa_riscv` to return `None` on non-RISCV systems, mainly to avoid confusion

easybuild/toolchains/compiler/gcc.py
boegel: I've updated `get_isa_riscv` to sort the ISA parts alphabetically, so it should also work when using GCC 13.x

bedroge (approved):
Lgtm

boegel (approved):

PR #5031, opened 2025-10-24
add `search_path_vars_headers` and `search_path_vars_linker` properties to `Toolchain`

lexming:
needed to properly fix:
- https://github.com/easybuilders/easybuild-easyblocks/issues/3969
via:
- https://github.com/easybuilders/easybuild-easyblocks/pull/3973

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -259,7 +259,10 @@ class Toolchain:
 
     def _search_path_vars(self, search_object):
         """Return list of environment variables used as search paths for the given object"""
-        search_path_opt = self.search_path[search_object]
+        try:
+            search_path_opt = self.search_path[search_object]
+        except KeyError:
+            raise EasyBuildError("Failed to retrieve search path options for '%s'", search_object)
 
         # default 'flags' option does not use search paths in the build environment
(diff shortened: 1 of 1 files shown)


easybuild/tools/toolchain/toolchain.py
```diff
@@ -247,6 +247,26 @@ def __init__(self, name=None, version=None, mns=None, class_constants=None, tcde
...
+        return self._search_path_vars('cpp_headers')
+
+    @property
+    def search_path_vars_linker(self):
+        """Return list of environment variables used as search paths by the linker"""
+        return self._search_path_vars('linker')
+
+    def _search_path_vars(self, search_object):
+        """Return list of environment variables used as search paths for the given object"""
+        search_path_opt = self.search_path[search_object]
```
boegel: Produce proper error message when wrong key is used:

```python
try:
    search_path_opt = self.search_path[search_object]
except KeyError:
    raise EasyBuildError("...")

lexming: right, fixed in https://github.com/easybuilders/easybuild-framework/pull/5031/commits/19efb5e3740761bb6146088dd74b2897852e5aee

boegel (approved):
lgtm

PR #5032, opened 2025-10-27
Print total runtime of all builds

Flamefire:
We have individual runtimes like
> == COMPLETED: Installation ended successfully (took 34 secs)

But often, especially when builds are submitted as jobs and/or with many builds, I'd like to see a total build time

This PR adds that:
> == Build succeeded for 5 out of 5 in 46 mins 13 secs

I excluded the time for parsing cmdline params etc and included only the build itself, i.e. `build_and_install_software`.   
That should enough as the rest of the time should be reasonably small compared to the actual build(s)

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -604,7 +604,8 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
     success_msg = "Build succeeded "
     if build_option('ignore_test_failure'):
         success_msg += "(with --ignore-test-failure) "
-    success_msg += f"for {correct_builds_cnt} out of {len(ordered_ecs)} (total: {time2str(datetime.now() - start_time)})"
+    success_msg += f"for {correct_builds_cnt} out of {len(ordered_ecs)} "
+    success_msg += f"(total: {time2str(datetime.now() - start_time)})"
 
     repo = init_repository(get_repository(), get_repositorypath())
     repo.cleanup()


easybuild/main.py
```diff
@@ -601,7 +604,7 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
     success_msg = "Build succeeded "
     if build_option('ignore_test_failure'):
         success_msg += "(with --ignore-test-failure) "
-    success_msg += "for %s out of %s" % (correct_builds_cnt, len(ordered_ecs))
+    success_msg += f"for {correct_builds_cnt} out of {len(ordered_ecs)} in {time2str(datetime.now() - start_time)}"
```
Crivella: Maybe we could use the same style as other time reporting eg for the steps of `... (took TIME)`?

Flamefire: Like this?

> == Build succeeded for 5 out of 5
> ... (took 46 mins 13 secs)

Not sure of this format here could be confused for something else and it was easier. It would need a 2nd `print_msg` to have it match more closely the other messages:

> == Build succeeded for 5 out of 5
> == ... (took 46 mins 13 secs)

But no strong opinion, can change it if you think one is better (which one?)

Crivella: I'd be fine also having it on one line, maybe we could do something like

> Build succeeded for 5 out of 5 (total 46 mins 13 secs)

Also no strong opinion on it (could always be adjusted later)  but agree it would be nice to have the total runtime of the build as well.

Maybe for a separate PR, but related to this it might also be worth to get a total CPU/node time for parallel jobs

Flamefire: >  ==  Build succeeded for 5 out of 5 (total 46 mins 13 secs)

I was wondering how that is better than

>    == Build succeeded for 5 out of 5 in 46 mins 13 secs

but while that reads more fluently it has too many numbers next to each other to read in a quick glimpse and now I see better why you preferred the "...took" variant..

I'll add a colon too:

```suggestion
    success_msg += f"for {correct_builds_cnt} out of {len(ordered_ecs)} (total: {time2str(datetime.now() - start_time)})"
```

PR #5035, opened 2025-10-29
add example hook for updating OpenSSL 1.x dependency to 3.x

jose-d:
I needed to build a <= 2023a toolchain on EL9-like Linux systems where OpenSSLv1 is unavailable (and deprecated).

This PR proposes an example hook that addresses this scenario. I believe it could be useful to others facing the same issue.

Credits:
Thanks to @Micket for guidance on how to approach this problem.

### DIFF
--- a/contrib/hooks/bumpOpenSSL.py
+++ b/contrib/hooks/bumpOpenSSL.py
@@ -15,7 +15,7 @@ def parse_hook(ec, *args, **kwargs):
     if openssl_found:
         print(f"[openssl1->3 hook] OpenSSL found in dependencies of {ec['name']}, replacing with OpenSSL 3")
         for i, dep in enumerate(raw_deps):
-            if dep[0] == 'OpenSSL':
+            if dep[0] == 'OpenSSL' and dep[1] == '1.1':
                 raw_deps[i] = ('OpenSSL', '3', '', EASYCONFIG_CONSTANTS['SYSTEM'][0])
                 break
 


contrib/hooks/bumpOpenSSL.py
```diff
@@ -0,0 +1,26 @@
...
+    raw_deps = ec['dependencies'] if 'dependencies' in ec else []
+
+    # Check if OpenSSL is in any dependency tuple's first element
+    openssl_found = any(dep[0] == 'OpenSSL' for dep in raw_deps)
+
+    # if found, replace its version - second element in the tuple with '3'.
+    if openssl_found:
+        print(f"[openssl1->3 hook] OpenSSL found in dependencies of {ec['name']}, replacing with OpenSSL 3")
+        for i, dep in enumerate(raw_deps):
+            if dep[0] == 'OpenSSL':
```
ocaisa: Since you know _exactly_ what you are looking to replace, might as well be very specific
```suggestion
            if dep[0] == 'OpenSSL' and dep[1] == '1.1':
```

ocaisa: Otherwise this would also replace OpenSSL 4 by OpenSSL 3 (in twenty years or whatever)

Thyre: > Otherwise this would also replace OpenSSL 4 by OpenSSL 3 (in twenty years or whatever)

Development on OpenSSL 4.0 was already announced (and there's a milestone for it). Based on the discussion on the Fedora mailing list early September ago, the release could be as close as April next year.

> OpenSSL upstream has started development of the version OpenSSL 4.0
It will imply soname bump, removing ENGINE support, etc. As OpenSSL uses time-based release planning, the release would happen in April 2026. 

Just for your info.

Issue #5037, opened 2025-10-31
Potential wrong order in default include dirs for rpath wrappers


Crivella:
Currently we set the install directory of a library to take precedence over the current library location

https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/framework/easyblock.py#L2956-L2962

This can lead to rebuilds of an existing software to fail due to picking up the already installed library instead of the newly compiled one (either in multi-stage build or in running test suites)

EG:

- LLVM
  - https://github.com/easybuilders/easybuild-easyblocks/issues/3708
  - https://github.com/easybuilders/easybuild-easyblocks/blame/9706e1e3fe42cf6c0f3d1a4e98b1bbd671f63087/easybuild/easyblocks/l/llvm.py#L540-L545
- https://github.com/easybuilders/easybuild-easyconfigs/pull/24316 was picking up UCX in tests (even when explicitly removed) do to a previous installation (was working wint `--installpath=$(mktemp -d)` which took me a while to figure out why my usual test-PR command was not working

I think that picking up the `$ORIGIN` locaitons first should always be favored as, in case of an installed library, it will pick its own location, but would allow newly compiled binaries or multi library projects to pick up the correct libs when used from the build directory.

PR #5038, opened 2025-10-31
Give precedence to `$ORIGIN` locations in RPATH wrappers

Crivella:
Fixes:

- #5037

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2954,11 +2954,11 @@ class EasyBlock:
         # $ORIGIN will be resolved by the loader to be the full path to the executable or shared object
         # see also https://linux.die.net/man/8/ld-linux;
         self.rpath_include_dirs.extend([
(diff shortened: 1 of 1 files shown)


Crivella:
Builds that worked with this branch checked out (while having an existing installation of that software that would previously cause trouble in the tests)

- https://github.com/easybuilders/easybuild-easyconfigs/pull/24316#issuecomment-3473531643
- https://github.com/easybuilders/easybuild-easyconfigs/pull/24316#issuecomment-3473594867

ocaisa:
I'm pretty sure option A exists so that when you run the test step for something not yet installed it can still find the libraries that it needs (which will ultimately end up in `os.path.join(self.installdir, 'lib')` but are not there yet).

Specifically, the scenario is `temp_location/bin/something` being run which depends on `temp_location/lib/librequirement.so` _before_ the install step. My memory is running on fumes, but I believe we saw this for LAMMPS at least.

PR #5040, opened 2025-11-12
Add environment variable for profiling EasyBuild

Micket:
Intended usage:

```bash
EB_PROFILE=profile.out  eb SomeSoftware-1.2.3.eb

snakeviz profile.out  # nice tool to view profiling data
```

### DIFF
--- a/eb
+++ b/eb
@@ -127,8 +127,14 @@ if [ -z "${EB_PROFILE}" ]
 then
     EB_PROFILE_FLAGS=()
 else
-    verbose "Writing profile output to: \"${EB_PROFILE}\""
-    EB_PROFILE_FLAGS=(-m cProfile -o "${EB_PROFILE}")
+    if [ "${pyver_min}" -lt "9" ]
+    then
+        echo "EB_PROFILE not supported for Python < 3.9"
+        EB_PROFILE=""
+    else
+        verbose "Writing profile output to: \"${EB_PROFILE}\""
+        EB_PROFILE_FLAGS=(-m cProfile -o "${EB_PROFILE}")
+    fi
 fi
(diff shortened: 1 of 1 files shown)


eb
```diff
@@ -122,7 +122,16 @@ then
     export FANCYLOGGER_IGNORE_MPI4PY=1
 fi
```
branfosj, line 124: I do not consider it worth spending much time getting this working for Python 3.6.

So we could do the following:
```
if [ "${pyver_min}" -lt "9" ]
then
    echo "EB_PROFILE not supported for Python < 3.9"
    EB_PROFILE=""
fi
```

Or exit with with an error.

eb
```diff
@@ -122,7 +122,22 @@ then
...
 
+# If a profile filename is given, run easybuild via cProfile
+if [ -z "${EB_PROFILE}" ]
+then
+    EB_PROFILE_FLAGS=()
+else
+    if [ "${pyver_min}" -lt "9" ]
+    then
+        echo "EB_PROFILE not supported for Python < 3.9"
+        EB_PROFILE=""
```
boegel, line 133: This should be setting `EB_PROFILE_FLAGS`, no?

```suggestion
        EB_PROFILE_FLAGS=()
```

boegel (changes requested):

eb
boegel: @Micket ping

Flamefire: Maybe start with an empty `$EB_PROFILE_FLAGS` and append/overwrite when required? Gets rid of the first branch and avoids it here.

I'd also make it an error if EB_PROFILE can't be honored.

PR #5042, opened 2025-11-13
also pick up on toolchain components in `Toolchain._add_dependency_variables`

boegel:
Follow-up for #5041 (which was a follow-up for #5023) to fix another problem when using `--search-path-cpp-headers=include_paths`, as we're doing in EESSI.

Without this, the configure step for `` (see ) fails with:

```
configure: error: --with-mpi was given, but test for MPI failed
```

`config.log` shows:
```
conftest.c:13:10: fatal error: mpi.h: No such file or directory
   13 | #include <mpi.h>
      |          ^~~~~~~
compilation terminated.
```

### DIFF
--- a/test/framework/filetools.py
+++ b/test/framework/filetools.py
@@ -2294,8 +2294,8 @@ class FileToolsTest(EnhancedTestCase):
 
         ft.copy_dir(to_copy, target_dir, ignore=lambda src, names: [x for x in names if '6.4.0-2.28' in x])
         self.assertExists(target_dir)
-        expected = ['GCC-10.2.0.eb', 'GCC-4.6.3.eb', 'GCC-4.6.4.eb', 'GCC-4.8.2.eb', 'GCC-4.8.3.eb', 'GCC-4.9.2.eb',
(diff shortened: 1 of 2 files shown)


easybuild/tools/toolchain/toolchain.py
```diff
@@ -1140,21 +1139,21 @@ def _add_dependency_variables(self, names=None, cpp=None, ld=None):
...
-        dependencies = self.dependencies if names is None else [{"name": name} for name in names if name]
+        deps = self.dependencies if names is None else [{'name': name} for name in names if name]
 
-        # collect software install prefixes for dependencies
-        dependency_roots = []
-        for dep in dependencies:
-            if dep.get("external_module", False):
+        # collect software install prefixes for toolchain components + dependencies
+        dep_roots = []
+        for dep in deps + self.tcdeps:
```
boegel, line 1146: Just to point it out: here's the actual fix.

Everything else that was changed in `easybuild/*` is basically (very minor) code style improvement...

PR #5043, opened 2025-11-14
Enable flake8-comprehension code style check and fix issues

Flamefire:
https://pypi.org/project/flake8-comprehensions/ has various useful checks, most importantly:
> C419 Unnecessary list comprehension in <any/all>() prevents short-circuiting - rewrite as a generator.

I.e. things like `any([expensive_check(x) for x in collection])` runs the check *after* processing the full collection instead of stopping after the first "match" which costs performance.

To me the whole collection of checks look useful for performance and readability so enable them all.

### DIFF
--- a/easybuild/base/fancylogger.py
+++ b/easybuild/base/fancylogger.py
@@ -435,7 +435,7 @@ def thread_name():
     """
     returns the current threads name
     """
-    return threading.currentThread().getName()
+    return threading.current_thread().name
 
 
 def getLogger(name=None, fname=False, clsname=False, fancyrecord=None):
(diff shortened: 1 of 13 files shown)


easybuild/base/fancylogger.py
```diff
@@ -435,7 +435,7 @@ def thread_name():
     """
     returns the current threads name
     """
-    return threading.currentThread().getName()
+    return threading.current_thread().name
```
boegel, line 438: Both `currentThread()` and `getName()` are deprecated, so makes sense to stop using those since they're bound to be removed at some point, see https://docs.python.org/3/library/threading.html#threading.current_thread and https://docs.python.org/3/library/threading.html#threading.Thread.getName

PR #5045, opened 2025-11-14
hide output of module commands in log file by default + add `--debug-module-cmds` option to opt-in to keeping them

Flamefire:
This is disabled by default and hides the output of the module command.
That is usually just "noise" hiding the relevant part of the log.

With it disabled the output looks like:
> == 2025-11-14 10:58:31,197 run.py:511 INFO Running shell command '/opt/lmod/lmod/libexec/lmod python load GCC/4.6.3' in /git/easybuild-framework
> == 2025-11-14 10:58:31,261 run.py:625 INFO Shell command completed successfully: /opt/lmod/lmod/libexec/lmod python load GCC/4.6.3

I added small refactorings and fixes I've seen:
- The type hint allows auto-completing in IDEs which is quite helpful for writing the test
- `re.sub('module use [...]', modtxt, re.M)` was wrong: The 3rd parameter is `count` not `flags`

Includes:
- [x] https://github.com/easybuilders/easybuild-framework/pull/5077

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1259,12 +1259,13 @@ class ModulesTool:
                 self.log.debug("Changing %s from '%s' to '%s' in environment for module command",
                                key, old_value, new_value)
 
-        log_output = build_option('debug_module_cmds') and not kwargs.get('hide_output', False)
+        debug_module_cmds = build_option('debug_module_cmds')
+        log_output = debug_module_cmds and not kwargs.get('hide_output', False)
         cmd_list = self.compose_cmd_list(args)
         cmd = ' '.join(cmd_list)
         # note: module commands are always run in dry mode, and are kept hidden in trace and dry run output
         res = run_shell_cmd(cmd_list, env=environ, fail_on_error=False, use_bash=False, split_stderr=True,
-                            hidden=True, in_dry_run=True, output_file=False,
+                            hidden=True, in_dry_run=True, output_file=debug_module_cmds,
                             log_output_on_success=log_output)
 
         # stdout will contain python code (to change environment etc)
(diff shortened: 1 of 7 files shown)


easybuild/tools/modules.py
```diff
@@ -1130,12 +1130,12 @@ def load(self, modules, mod_paths=None, purge=False, init_env=None, allow_reload
         for mod in modules:
             self.run_module('load', mod)
 
-    def unload(self, modules, log_changes=True):
```
Crivella, line 1133: Not sure if we should still allow `log_changes` to be defined and add a deprecation warning for it and covert it to `hide_output`

Guess this depends on how much we consider this functions as a stable API of EasyBuild.

In particular here, we are inverting the logic/default value, which would mean we should also check if `log_changes` was being passed as an keyword argument or positional one, eg we would have not way to check if
`unload(..., True)` was meant for the old or new syntax

Similar comments also for the other signature changes

Crivella (changes requested):

easybuild/tools/modules.py
Flamefire: Good point, although this parameter is not critical, it only affects logging so I'd just go with it
I added the deprecation though which adds some extra code to handle cases where both are passed.

I also allow `hide_output` only by keyword. That avoids potential confusion

Crivella (approved):
LGTM

- Fix bug in usage of `re.sub` where `re.M` was being passed to `count` instead of `flags` (plus usage as positional arguments is [deprecated since 3.13 ](https://docs.python.org/3/library/re.html#re.sub))
- Re-naming of argument default value and logic properly
  - allow to keep using old one and adds a deprecation warning for it
  - keeps default behavior consistent with old one
  - Does not affect any other place than what is touched here (`.unload` is always called without specifying `log_changes` throughout the EB repositories)
- Add tests for checking that module logs are properly shown/hidden

Crivella:
Going in, thanks @Flamefire!

PR #5049, opened 2025-11-14
add support for both `$VAR` and `${VAR}` variable formats in `modextravars`

smoors:
uses `string.Template`, which also supports lowercase variables and escaping via `$$VAR` or `$${VAR}`.

### DIFF
--- a/test/framework/module_generator.py
+++ b/test/framework/module_generator.py
@@ -38,7 +38,7 @@ from unittest import TextTestRunner, TestSuite
 from easybuild.framework.easyconfig.tools import process_easyconfig
 from easybuild.tools import LooseVersion, config
 from easybuild.tools.filetools import mkdir, read_file, remove_file, write_file
-from easybuild.tools.module_generator import ModuleGeneratorLua, ModuleGeneratorTcl, dependencies_for
+from easybuild.tools.module_generator import ModuleGeneratorLua, ModuleGeneratorTcl, dependencies_for, wrap_shell_vars
 from easybuild.tools.module_naming_scheme.utilities import is_valid_module_name
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig.easyconfig import EasyConfig, ActiveMNS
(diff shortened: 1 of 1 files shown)


easybuild/tools/module_generator.py
```diff
@@ -117,6 +118,31 @@ def dependencies_for(mod_name, modtool, depth=None):
     return mods
 
 
+def wrap_shell_vars(strng, wrap_prefix, wrap_suffix):
+    """
+    Wrap variables $VAR or ${VAR} in wrap_tmpl template string
```
boegel: There's no such thing as `wrap_tmpl`?

easybuild/tools/module_generator.py
```diff
@@ -117,6 +118,31 @@ def dependencies_for(mod_name, modtool, depth=None):
     return mods
 
 
+def wrap_shell_vars(strng, wrap_prefix, wrap_suffix):
```
boegel, line 121: Let's add a dedicated unit test for this new function, along with the end-to-end tests you've already added?

easybuild/tools/module_generator.py
```diff
@@ -133,11 +159,6 @@ class ModuleGenerator:
     # a single level of indentation
     INDENTATION = ' ' * 4
 
-    # shell environment variable name: ${__}VAR_NAME_00_SUFFIX
-    REGEX_SHELL_VAR_PATTERN = r'[A-Z_]+[A-Z0-9_]+'
-    REGEX_SHELL_VAR = re.compile(rf'\$({REGEX_SHELL_VAR_PATTERN})')
-    REGEX_QUOTE_SHELL_VAR = re.compile(rf'[\"\']\$({REGEX_SHELL_VAR_PATTERN})[\"\']')
```
boegel, line 139: Strictly speaking this is part of the API, so we shouldn't just remove these constants, people may be using them (in hooks for example)

easybuild/tools/module_generator.py
```diff
@@ -1206,7 +1227,8 @@ class ModuleGeneratorLua(ModuleGenerator):
     LOAD_TEMPLATE_DEPENDS_ON = 'depends_on("%(mod_name)s")'
     IS_LOADED_TEMPLATE = 'isloaded("%s")'
 
-    OS_GETENV_TEMPLATE = r'os.getenv("%s")'
```
boegel, line 1209: Same here, this is part of the API, we shouldn't just remove this...

smoors: done in https://github.com/easybuilders/easybuild-framework/pull/5049/commits/67c8994a6f60b16a57c04b262edfadea3e2ed931

easybuild/tools/module_generator.py
smoors: done in https://github.com/easybuilders/easybuild-framework/pull/5049/commits/67c8994a6f60b16a57c04b262edfadea3e2ed931

easybuild/tools/module_generator.py
smoors: done in https://github.com/easybuilders/easybuild-framework/pull/5049/commits/7faa56c3f415fb8a548b878aaf0a5e01380b5649

boegel (approved):
lgtm

PR #5050, opened 2025-11-17
close stdout/stderr handles in `run_shell_cmd` and test teardown + remove duplication in `toy_build` tests

Flamefire:
Almost every test case starts by getting the path to the test-folder, then to `toy-0.0.eb` and then reading this file.

Not only makes this harder to read the (relevant part of) the test it also slows it down to read the same file in every test.

Hence I factored those out to file-level constants.

While testing I noticed we don't close the mocked stdout/stderr buffers and added that too

### DIFF
--- a/easybuild/base/testing.py
+++ b/easybuild/base/testing.py
@@ -139,6 +139,7 @@ class TestCase(OrigTestCase):
 
         self.orig_sys_stdout = sys.stdout
         self.orig_sys_stderr = sys.stderr
+        self.stdout_std_err_buffers = []
 
     def convert_exception_to_str(self, err):
         """Convert an Exception instance to a string."""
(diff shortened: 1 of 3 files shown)


Flamefire:
Rebased

Flamefire:
Rebased to fix

boegel (approved):
lgtm

Flamefire:
I noticed there are > 200 similar cases in other test files.
@boegel Would you merge a PR replacing those too? My idea would be to move the constants to `test/framework/__init__.py` so we can import them with `from test.framework import TEST_ECS_DIR, TOY_EC, TOY_EC_TXT` in the 18 test files that use them.
Starting many tests with this boilerplate code gets tiresome.

boegel:
> I noticed there are > 200 similar cases in other test files. @boegel Would you merge a PR replacing those too? My idea would be to move the constants to `test/framework/__init__.py` so we can import them with `from test.framework import TEST_ECS_DIR, TOY_EC, TOY_EC_TXT` in the 18 test files that use them. Starting many tests with this boilerplate code gets tiresome.

@Flamefire Sure, sounds like a nice cleanup in the tests. (sorry for the late reply, I often overlook comments in merged PRs)

Flamefire:
Followup done: https://github.com/easybuilders/easybuild-framework/pull/5203
>300 lines removed

PR #5051, opened 2025-11-18
Load dependencies of SYSTEM toolchain by full name 

Flamefire:
This allows loading e.g. `GCCcore/12.3.0/ncurses/6.1` without loading the GCCcore/12.3.0 toolchain module.
So we can use that as a dependency.

Note that this already works in the EasyBuildMNS because there the short and full module name are the same

Only in the HierarchicalMNS it fails because there 'Compiler/GCCcore/12.3.0/ncurses/6.4' needs to be used to load it.

A use of this is https://github.com/easybuilders/easybuild-easyconfigs/blob/8bca6595b2b975eb32eda0035613cf6fa17610a3/easybuild/easyconfigs/r/ripunzip/ripunzip-0.4.0.eb (from https://github.com/easybuilders/easybuild-easyconfigs/pull/17959) which uses `Rust` as a build dependency

cc @Micket

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -654,12 +654,13 @@ class Toolchain:
             self.log.debug("Loading toolchain module and dependencies...")
 
         tc_mod = None if system_toolchain else self.det_short_module_name()
-        # For SYSTEM software we allow using dependencies from any toolchain so we need to use the full
-        # module name to allow loading them in the hierarchical MNS
-        is_system_toolchain = self.is_system_toolchain()
 
         def get_module_name(dep):
-            key = 'full_mod_name' if is_system_toolchain and not dep[SYSTEM_TOOLCHAIN_NAME] else 'short_mod_name'
+            key = 'short_mod_name'
+            if system_toolchain and not dep[SYSTEM_TOOLCHAIN_NAME]:
+                # When using the SYSTEM toolchain, we should use the full module name to load dependencies NOT from the
+                # SYSTEM toolchain to allow using dependencies from any toolchain when using a hierarchical MNS
+                key = 'full_mod_name'
             return dep[key]
 
         dep_mods = [get_module_name(dep) for dep in self.dependencies]
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -656,8 +656,13 @@ class Toolchain:
         tc_mod = None if system_toolchain else self.det_short_module_name()
         # For SYSTEM software we allow using dependencies from any toolchain so we need to use the full
         # module name to allow loading them in the hierarchical MNS
-        mod_name_key = 'full_mod_name' if self.is_system_toolchain() else 'short_mod_name'
-        dep_mods = [dep[mod_name_key] for dep in self.dependencies]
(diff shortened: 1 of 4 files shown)


Flamefire:
It makes sense to merge https://github.com/easybuilders/easybuild-framework/pull/5050 first so I can rebase this such that some commits disappear.

That one should be trivial as it is a test-only change.

In any case: Do you agree with this approach at all? In Slack @ocaisa mentioned:

> It's asking for trouble to allow dependencies higher up the hierarchy, that should only be possible if they are build dependencies
> You would need very good sanity checks to ensure a runtime dependency does not slip in because of our use of rpath
> With Rust maybe it is safe, but that's the exception rather than the rule

However it currently works for EB-MNS but not HMNS

I think I can limit this change to build dependencies but that doesn't solve the MNS discrepancy.   
We'd need a check somewhere that runtime deps are same or higher in the module hierarchy and possibly a warning for build deps.

That is if we want to allow this at all. If not this PR is not required and we need to error out earlier.

boegel:
@Flamefire #5050 is merged, can you rebase this?

easybuild/tools/toolchain/toolchain.py
```diff
@@ -654,7 +654,15 @@ def _load_modules_toolchain_and_deps(self, silent=False):
             self.log.debug("Loading toolchain module and dependencies...")
 
         tc_mod = None if system_toolchain else self.det_short_module_name()
-        dep_mods = [dep['short_mod_name'] for dep in self.dependencies]
+        # For SYSTEM software we allow using dependencies from any toolchain so we need to use the full
+        # module name to allow loading them in the hierarchical MNS
+        is_system_toolchain = self.is_system_toolchain()
+
+        def get_module_name(dep):
+            key = 'full_mod_name' if is_system_toolchain and not dep[SYSTEM_TOOLCHAIN_NAME] else 'short_mod_name'
```
boegel: Are we 100% sure here that `dep` will have `system` as a key?

Maybe we should be more careful here, just in case:

```suggestion
            key = 'full_mod_name' if is_system_toolchain and not dep.get(SYSTEM_TOOLCHAIN_NAME) else 'short_mod_name'
```

and maybe this is more readable:
```suggestion
            key = 'short_mod_name'
            # when using system toolchain, we should always use full module name to load dependencies,
            # this is required to allow using dependencies from any toolchain when using a hierarchical MNS
            if self.is_system_toolchain() and not dep.get(SYSTEM_TOOLCHAIN_NAME):
                key = 'full_mod_name'

boegel (changes requested):

easybuild/tools/toolchain/toolchain.py
Flamefire: Yes, 100%: This is a parsed dependency from https://github.com/easybuilders/easybuild-framework/blob/e4bbc09fcfccd34b0009b4175a9de7d3a8f7c99c/easybuild/framework/easyconfig/easyconfig.py#L1649-L1671

The value is finally set in `finalize_dependencies`: https://github.com/easybuilders/easybuild-framework/blob/e4bbc09fcfccd34b0009b4175a9de7d3a8f7c99c/easybuild/framework/easyconfig/easyconfig.py#L1820

I applied a variation of the suggestion

PR #5054, opened 2025-11-20
pick up on `config.cfg` EasyBuild configuration file in current working directory

### LINKED ISSUE #3965: Working directory pickup of configuration files
Is there support for `eb` picking the configuration file in the folder it is executed?

j3mdamas:
Closes #3965 

I used git config as a reference: https://git-scm.com/docs/git-config#SCOPES

(there they call the easybuild user scope global scope, but we don't need to change that)

I ran some small tests locally with `eb --show-default-configfiles` and `eb --show-config` and everything seems to work as expected, but I don't know if there are some unit tests I should create.

tagging: @boegel

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -131,7 +131,7 @@ XDG_CONFIG_DIRS = os.environ.get('XDG_CONFIG_DIRS', '/etc/xdg').split(os.pathsep
 DEFAULT_SYS_CFGFILES = [sorted(glob.glob(os.path.join(d, 'easybuild.d', '*.cfg')))
                         for d in XDG_CONFIG_DIRS]
 DEFAULT_USER_CFGFILE = os.path.join(XDG_CONFIG_HOME, 'easybuild', 'config.cfg')
-DEFAULT_LOCAL_CFGFILE= os.path.join(os.path.normpath(os.getcwd()), 'config.cfg')
+DEFAULT_LOCAL_CFGFILE = os.path.join(os.path.normpath(os.getcwd()), 'config.cfg')
 
 DEFAULT_LIST_PR_STATE = GITHUB_PR_STATE_OPEN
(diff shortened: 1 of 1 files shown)


easybuild/tools/options.py
```diff
@@ -1496,6 +1500,8 @@ def show_default_configfiles(self):
             '',
             "[with $XDG_CONFIG_HOME: %s, $XDG_CONFIG_DIRS: %s]" % (xdg_config_home, xdg_config_dirs),
             '',
+            "* local-level: %s" % os.path.join('.', 'config.cfg'),
```
boegel, line 1503: I would change this to avoid confusion:
```suggestion
            "* current working directory: %s" % os.path.join('.', 'config.cfg'),

File: easybuild/tools/options.py [hunk 2]
```diff
@@ -131,6 +131,7 @@ def terminal_supports_colors(stream):
 DEFAULT_SYS_CFGFILES = [sorted(glob.glob(os.path.join(d, 'easybuild.d', '*.cfg')))
                         for d in XDG_CONFIG_DIRS]
 DEFAULT_USER_CFGFILE = os.path.join(XDG_CONFIG_HOME, 'easybuild', 'config.cfg')
+DEFAULT_LOCAL_CFGFILE = os.path.join(os.path.normpath(os.getcwd()), 'config.cfg')
```
boegel, line 134: Maybe rename to `DEFAULT_CWD_CFGFILE`, or `DEFAULT_CURRENT_WORKING_DIR_CFGFILE` if that's not clear enough?
I think "local" could be interpreted differently

boegel (changes requested):
@j3mdamas Ideally we have an extra (dedicated) unit test for this, `test/framework/config.py` is the right place for that I think.

Do let us know if that's a bridge too far for you!

j3mdamas:
@boegel just wanted to note ahead of the EB conf call that this is not forgotten, I just didn't have the opportunity to address your questions. I'll come back to it.

PR #5057, opened 2025-11-24
Don't append empty component to `start_dir` in `EasyBlock.guess_start_dir`

Flamefire:
In most cases `start_dir` is unset/empty.
When passing it to `os.path.join` it leads to `/topdir/` which could cause issues when using the template as `%(startdir)s/subdir` as a double-slash might be treated specially.


boegel (approved):
lgtm

PR #5058, opened 2025-11-28
Set `$___EASYBUILD___` environment variable to '`EasyBuild`' to indicate that you're in an EasyBuild session

ocaisa:
The use case for this is `buildenv` in EESSI (https://gitlab.com/eessi/support/-/issues/213), where you might want to auto-load a `buildenv` with your compiler, but not if you are in an EasyBuild session.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -301,7 +301,6 @@ class EasyBlock:
         self.orig_modulepath = os.getenv('MODULEPATH')
 
         # keep track of initial environment we start in, so we can restore it if needed
-        os.environ['___EASYBUILD___'] = "EasyBuild"  # set an environment variable to know we are in an EB session
         self.initial_environ = copy.deepcopy(os.environ)
         self.reset_environ = None
         self.tweaked_env_vars = {}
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -301,6 +301,7 @@ def __init__(self, ec, logfile=None):
         self.orig_modulepath = os.getenv('MODULEPATH')
 
         # keep track of initial environment we start in, so we can restore it if needed
+        os.environ['___EASYBUILD___'] = "EasyBuild"  # set an environment variable to know we are in an EB session
```
ocaisa: We need this before copying the initial environment as that environment gets restored before loading the module (which is exactly where we want to target a conditional load)

boegel: @ocaisa Let's move this up, as the very first thing we do in `__init__`, just to make sure that we never do anything that involves calling out to external tools (which could take this into account) before this is set.

boegel: Actually, this should be done way earlier, in `main.py` ?

boegel: @ocaisa see proposed change in https://github.com/ocaisa/easybuild-framework/pull/48

PR #5063, opened 2025-12-03
Vendor a copy of `tomli` in `easybuild.tools.tomllib`

Flamefire:
Include https://github.com/hukkin/tomli/tree/2.3.0 in easybuild as `easybuild.tools.tomllib` which will be replace by `tomllib` in Python 3.11+

Slightly patched to make it compatible with Python 3.6:
- Type hints (e.g. `tuple[int, int] -> tuple`)
- `/, *` in function params (end of positional args)

I copied the example from toml.io which (hopefully) triggers most code parts of the library to shake out any issues.

I decided for `tomli` instead of extracting `tomllib` from CPython because of its goal:

> A version of Tomli, the tomllib module, was added to the standard library in Python 3.11 via [PEP 680](https://www.python.org/dev/peps/pep-0680/). Tomli continues to provide a backport on PyPI for Python versions where the standard library module is not available and that have not yet reached their end-of-life.

It currently doesn't support Python 3.6 but 3.8 so we can use it when dropping Python < 3.9

Requires:
- [ ] ~~https://github.com/easybuilders/easybuild-framework/pull/5065 (simplify with `find_packages` to avoid listing the new ones)~~
- [x] https://github.com/easybuilders/easybuild-framework/pull/5064 (test with Python 3.6 to ensure compatibility)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,8 @@ easybuild_packages = [
     "easybuild.toolchains.fft", "easybuild.toolchains.linalg", "easybuild.tools", "easybuild.tools.containers",
     "easybuild.tools.deprecated", "easybuild.tools.job", "easybuild.tools.toolchain",
     "easybuild.tools.module_naming_scheme", "easybuild.tools.package", "easybuild.tools.package.package_naming_scheme",
-    "easybuild.tools.py2vs3", "easybuild.tools.repository", "test.framework", "test",
+    "easybuild.tools.py2vs3", "easybuild.tools.repository", "easybuild.tools.tomllib", "easybuild.tools.tomllib.tomli",
+    "test.framework", "test",
 ]
 
 # Verify the above list is complete, if setuptools is installed


Flamefire:
Without setuptools you'd need to list all "packages" in the `setup` call and it took me multiple tries to figure out which it actually requires.

What were those issues and are they still relevant? Our usage here might have changed enough that we are "standard" enough not to run into them (anymore). Especially if those issues were from the Python 2 days. Seems most advise using setuptools unconditionally (if not even pyproject.toml) for non-ancient Pythons

An alternative if you really want to play it safe:
- Import `setuptools.find_packages` if possible
- check output against our manual list and error on difference printing the correct list
- Have EASYBUILD_IGNORE_FINDPACKAGES=1 to skip that as a way out if it unexpectedly causes issues on some system.

But I'd really like to avoid error-prone adding (more) things to a hard-coded list if possible.

PR #5065, opened 2025-12-04
Switch to setuptools and `find_packages`

Flamefire:
Any reasonably modern Python distribution includes setuptools. So the workaround, that was useful for Python 2.x, is no longer necessary and can be removed.
This allows using `find_packages` instead of maintaining an explicit list of packages which is error-prone.

I ran into this when adding the vendored `tomli` package

see https://github.com/easybuilders/easybuild-easyblocks/pull/4006 & https://github.com/easybuilders/easybuild-easyconfigs/pull/24747

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,10 +31,7 @@ This script can be used to install easybuild-framework, e.g. using:
 import glob
 import os
 import logging
-try:
-    from distutils.core import setup
-except ImportError:
-    from setuptools import setup
+from setuptools import setup, find_packages
 
 from easybuild.tools.version import VERSION
 
@@ -68,34 +65,6 @@ def find_rel_test():
     return res
 
 
-easybuild_packages = [
-    "easybuild", "easybuild.base",
-    "easybuild.framework", "easybuild.framework.easyconfig", "easybuild.framework.easyconfig.format",
-    "easybuild.toolchains", "easybuild.toolchains.compiler", "easybuild.toolchains.mpi",
-    "easybuild.toolchains.fft", "easybuild.toolchains.linalg", "easybuild.tools", "easybuild.tools.containers",
-    "easybuild.tools.deprecated", "easybuild.tools.job", "easybuild.tools.toolchain",
-    "easybuild.tools.module_naming_scheme", "easybuild.tools.package", "easybuild.tools.package.package_naming_scheme",
-    "easybuild.tools.py2vs3", "easybuild.tools.repository",
-    "easybuild.tools.tomllib", "easybuild.tools.tomllib.tomli", "easybuild.tools._toml_writer",
-    "test.framework", "test",
-]
-
-# Verify the above list is complete, if setuptools is installed
-try:
-    import setuptools
-except ImportError:
-    pass
-else:
-    packages = set(setuptools.find_packages())
-    easybuild_packages_set = set(easybuild_packages)
-    if easybuild_packages_set != packages:
-        # Warning only
-        print("="*80 + "\n"
-              "=== WARNING: Wrong list of easybuild_packages.\n"
-              f"Missing: {packages - easybuild_packages_set}\n"
-              f"Unneccessary: {easybuild_packages_set - packages}"
-              "\n" + "="*80 + "\n")
-
 setup(
     name="easybuild-framework",
     version=str(VERSION),
@@ -106,7 +75,7 @@ implement support for installing particular (groups of) software packages.""",
     license="GPLv2",
     keywords="software build building installation installing compilation HPC scientific",
     url="https://easybuild.io",
-    packages=easybuild_packages,
+    packages=find_packages(),
     package_dir={'test.framework': 'test/framework'},
     package_data={'test.framework': find_rel_test()},
     scripts=[
--- a/.github/workflows/eb_command.yml
+++ b/.github/workflows/eb_command.yml
@@ -35,10 +35,7 @@ jobs:
         # update to latest pip, check version
         pip install --upgrade pip
         pip --version
-        if ! python -c "import distutils" 2> /dev/null; then
-            # we need setuptools for distutils in Python 3.12+, needed for python setup.py sdist
-            pip install --upgrade setuptools
-        fi
+        pip install setuptools
 
         # for modules tool
         APT_PKGS="lua5.3 liblua5.3-dev lua-filesystem lua-posix tcl tcl-dev"
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -127,10 +127,8 @@ jobs:
         pip install --upgrade pip
         pip --version
         pip install -r requirements.txt
-        if ! python -c "import distutils" 2> /dev/null; then
-            # we need setuptools for distutils in Python 3.12+, needed for python setup.py sdist
-            pip install --upgrade setuptools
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -88,7 +75,7 @@ def find_rel_test():
     license="GPLv2",
     keywords="software build building installation installing compilation HPC scientific",
     url="https://easybuild.io",
-    packages=easybuild_packages,
+    packages=find_packages(),
```
boegel, line 78: I will need to check this thoroughly, since I don't want to me up a release due to this change...

setup.py
```diff
@@ -31,10 +31,7 @@
 import glob
 import os
 import logging
-try:
-    from distutils.core import setup
-except ImportError:
-    from setuptools import setup
+from setuptools import setup, find_packages
```
boegel, line 34: I'm not happy with introducing a hard requirement on `setuptools`, since it's not part of the Python standard library, and often includes backwards-incompatible changes in new versions...

setup.py
Flamefire: I did a test locally: `find_packages()` and `easybuild_packages` were the same

setup.py
Flamefire: We already have a hard dependency on that for Python 3.12 where distutils doesn't exist anymore.

According to various sources setuptools should be available with any reasonably modern Python installation, i.e. Python 3.something. e.g. via `ensurepip`. And we don't have CI (for newer Pythons) without setuptools installed, so not sure if EB even works completely without it.

And as we don't use any advanced features I doubt that those setuptools changes affect us.

Flamefire:
What exactly were the issues? I only found https://github.com/easybuilders/easybuild-framework/pull/2988 referencing https://lists.ugent.be/wws/arc/easybuild/2019-08/msg00057.html which doesn't open for me. Maybe >6yrs later those issues are gone/rare enough?

boegel:
Copying @Flamefire's post from #5063 here to avoid splintering the discussion:

> Without setuptools you'd need to list all "packages" in the `setup` call and it took me multiple tries to figure out which it actually requires.
> 
> What were those issues and are they still relevant? Our usage here might have changed enough that we are "standard" enough not to run into them (anymore). Especially if those issues were from the Python 2 days. Seems most advise using setuptools unconditionally (if not even pyproject.toml) for non-ancient Pythons
> 
> An alternative if you really want to play it safe:
> 
>     * Import `setuptools.find_packages` if possible
> 
>     * check output against our manual list and error on difference printing the correct list
> 
>     * Have EASYBUILD_IGNORE_FINDPACKAGES=1 to skip that as a way out if it unexpectedly causes issues on some system.
> 
> 
> But I'd really like to avoid error-prone adding (more) things to a hard-coded list if possible.


> What exactly were the issues? I only found #2988 referencing https://lists.ugent.be/wws/arc/easybuild/2019-08/msg00057.html which doesn't open for me. Maybe >6yrs later those issues are gone/rare enough?

#2988 gives a hint, indeed. Unfortunately the EasyBuild mailing list isn't publicly available currently (requires UGent VPN), hopefully that can be restored soon.
Here are some links to public archives:
- https://www.mail-archive.com/easybuild@lists.ugent.be/msg04882.html
- https://www.mail-archive.com/easybuild@lists.ugent.be/msg05263.html

It's very hard to determine whether `setuptools` is now properly stable and won't cause a ton of confusing problems if we have a hard requirement on it for all Python versions we support (even if that's only in `setup.py` and nowhere else).

Generally speaking, `setuptools` has a reputation of happily making backwards-incompatible breaking changes, and expecting that people "just deal with it" by using a sufficiently recent version, or by adjusting their `setup.py` accordingly.
Somehow it didn't get an honorable mention in my [*"How To Make Package Managers Cry"* talk from FOSDEM 2018](https://archive.fosdem.org/2018/schedule/event/how_to_make_package_managers_cry/), but it really should have. I'll get an opportunity to rectify that soon...
One example is https://github.com/pypa/setuptools/issues/944, see also https://github.com/easybuilders/easybuild-easyblocks/issues/1090.

I recall frequently being frustrated with `setuptools` several years ago for this type of reasons.
I don't recall recent instances, but that could be because EasyBuild hasn't had a hard requirement for `setuptools` for many years now (see also https://github.com/easybuilders/easybuild-framework/pull/2836).

In fact, more broadly speaking we have a standing rule that EasyBuild should not have a hard dependency on any Python package outside of the Python standard library. That's a rule that was imposed a while ago, after being fed up with dealing with a whole range of weird/annoying problems with various 3rd party dependencies.
I personally try to avoid hard 3rd party dependencies in the Python tools I develop...

So, introducing a hard requirement for `setuptools.find_packages` just so we don't need to maintain a (long) hardcoded list of `easybuild.*` packages anymore is not a good enough reason imho.
Depending on `setuptools.find_packages` *may* cause quite a bit of pain for years to come, without much benefit, especially since we don't really add additional `easybuild.*` packages frequently (https://github.com/easybuilders/easybuild-framework/pull/5063 most definitely is an exception here).
In fact, `find_packages` is even a bit of a foot-gun, see for example [here](https://jwodder.github.io/kbits/posts/pypkg-mistakes).

@Flamefire Please rework https://github.com/easybuilders/easybuild-framework/pull/5063 so it doesn't require these changes.
It should boil down to adding `easybuild.tools.tomli` and maybe also `easybuild.tools.tomli.tomllib` to `easybuild_packages` in `setup.py`. Even if it somehow is more complex than that, then I still prefer *not* introducing a hard requirement for `setuptools.find_packages`.

boegel:
Only requiring `setuptools` (in `setup.py`) with Python 3.12+ is a necessary evil, because `distutils` was removed from the Python standard library.
The approach taken in https://github.com/easybuilders/easybuild-framework/pull/4478 is very deliberate: only rely on `setuptools` when we have no other choice, and see how it pans out.

Not requiring `setuptools` (cfr. https://github.com/easybuilders/easybuild-framework/pull/2836) beyond `setup.py` was a very deliberate choice that I don't want to back-pedal on, but that's not really under discussion here (since this PR only affects `setup.py`).

We'll see how things work out with Python 3.12+, where we can't avoid `setuptools`.
I don't think enough people are already using that to have a goo view on whether or not there's trouble.
The people who do use Python 3.12+ already are perhaps a bit more tech-savvy.

The default Python in RHEL10 and derivatives (Rocky Linux 10, etc.) is Python 3.12, so we'll have a better view on things soon-ish.

For Python < 3.12, I'm strongly in favor of *never* requiring `setuptools`, because for those people who can't or don't want to use Python 3.12+, there *will* be surprises with `setuptools` at some point.

*"Those Who Cannot Remember the Past Are Condemned To Repeat It"*

setup.py
boegel: Which Python/setuptools version was used here matters a lot...

I don't think we can generally assume that `find_packages` will behave the way we expect it to in all possible circumstances

Flamefire: For our standard usage I don't think it could behave differently as likely other projects would be broken too. So I'd say it is fine to assume so

Flamefire:
That doesn't affect this, does it? And we'd have a way out if anything does:

> In the unlikely case we do run into issues with a specific setuptools version, https://github.com/easybuilders/easybuild-framework/issues/5090 allows us to select/exclude them.

PR #5067, opened 2025-12-09
allow `f` suffix in CUDA compute capabilities

bedroge:
CUDA compute capabilities can now also have an `f` suffix, see https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#feature-set-compiler-targets.

Note that it's also changed in the regex that is used to extract the CUDA architectures from binaries. I wasn't sure if the `f` suffix would show up there as well, but I tested that by compiling something manually (with `-arch compute_100f`) and running `cuobjdump` on it:
```
Fatbin ptx code:
================
arch = sm_100
code version = [8,8]
host = linux
compile_size = 64bit
compressed
ptxasOptions = 

Fatbin ptx code:
================
arch = sm_100f
code version = [8,8]
host = linux
compile_size = 64bit
compressed
ptxasOptions = 
```

So, it does indeed have `sm_100f`.

### DIFF
--- a/test/framework/systemtools.py
+++ b/test/framework/systemtools.py
@@ -428,16 +428,22 @@ ptxasOptions ="""
 CUOBJDUMP_DEVICE_CODE_ONLY = """
 Fatbin elf code:
 ================
-arch = sm_90
-code version = [1,7]
+arch = sm_100f
+code version = [1,8]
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

PR #5070, opened 2025-12-09
Don't install `test.framework` package

Flamefire:
This results in a top-level Python package called "test". As this isn't namespaced to easybuild it can cause issues with other Python packages and shouldn't be installed.

Additionally IMO it isn't useful to install the tests in the first place.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -195,6 +195,9 @@ jobs:
           # run tests *outside* of checked out easybuild-framework directory,
           # to ensure we're testing installed version (see previous step)
           cd $HOME
+          # Copy test folder first as that module is NOT installed. Could be run from source tree but we want to test the installed package
+          cp -r $GITHUB_WORKSPACE/test .
+
           # initialize environment for modules tool
           if [ -f $HOME/moduleshome ]; then export MODULESHOME=$(cat $HOME/moduleshome); fi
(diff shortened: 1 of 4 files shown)


Flamefire:
> I agree that namespace-wise this isn't nice, but I think there's some value in includes the tests with the installation.

I'd say 99% of users do not need nor want the tests, even though they don't take much space

> This is a "feature" for a long time now, see also [docs.easybuild.io/unit-tests#unit_tests_running](https://docs.easybuild.io/unit-tests/#unit_tests_running)

From how I read it nothing there states that the unit tests are installed / part of the distribution package, only that there **are** unit tests, how to run them and that they are run on CI. And running the tests isn't something so important that we need a deprecation period if any process is easily adapted. We can just update the documentation.

Some guides ([1](https://blog.ionelmc.ro/2014/05/25/python-packaging/) & [[2]](https://docs.python-guide.org/writing/structure/#test-suite) use a `tests` folder (plural) in the root. Others use a namespaced `tests` package, e.g. the [PYPA group](https://github.com/pypa) (pip, setuptools, ...)

So I'd say we go with the latter: `easybuild.<pkg>.tests` (for framework, easyconfigs, easyblocks)   
Or have them collected in one folder in the installation: `easybuild.tests.<pkg>`. That would be similar to the current `python -m test.easyblocks.suite` --> `python -m easybuild.tests.easyblocks.suite`

PR #5071, opened 2025-12-11
Vendor copy of tomli-w 1.2.0, use it to implement `dump_toml` function in `easybuild.tools.filetools`

Flamefire:
As `tomllib` can only *read* TOML files we need a writer too.

I decided for [`tomli-w`](https://github.com/hukkin/tomli-w) as it is basically a single file

### DIFF
--- a/easybuild/tools/__init__.py
+++ b/easybuild/tools/__init__.py
@@ -38,11 +38,3 @@ __path__ = __import__('pkgutil').extend_path(__path__, __name__)
 
 
 from easybuild.tools.loose_version import LooseVersion  # noqa(F401)
-
-
-def dump_toml(o: dict) -> str:
-    """Convert the input directory to a string in TOML format"""
-    # Import and redirect here on purpose as the conversion method (or Python package)
-    # is an implementation detail that can be changed at any time
-    import easybuild.tools._toml_writer
-    return easybuild.tools._toml_writer.dumps(o)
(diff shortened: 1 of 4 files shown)


easybuild/tools/__init__.py
```diff
@@ -38,3 +38,11 @@
 
 
 from easybuild.tools.loose_version import LooseVersion  # noqa(F401)
+
+
+def dump_toml(o: dict) -> str:
```
boegel: Please move this to `easybuild.tools.filetools`, I strongly prefer keeping `easybuild.tools` largely empty as it is now

easybuild/tools/_toml_writer/README.md
```diff
@@ -0,0 +1,14 @@
...
+This is an implementation of a library for writing TOML files.  
+To be able to update, change and remove it without breaking backwards compatibility in EasyBuild
+it is in a "private" sub-package and should not be used directly: **It can be changed without warning!**
+
+The stable interface is `easybuild.tools.dump_toml`.
+
+## Implementation
+
+Currently, a copy of [`tomli-w`](https://github.com/hukkin/tomli-w) is vendored in this folder.
+The used version is 1.2.
```
boegel:
```suggestion
The used version is 1.2.0 (https://github.com/hukkin/tomli-w/releases/tag/1.2.0, commit a8f80172ba16fe694e37f6e07e6352ecee384c58).
```

PR #4927, opened 2025-06-18
make NVHPC a full toolchain with nvidia-compilers, NVHPCX, NVBLAS, and NVScaLAPACK

lexming:
Fixes #4853

Enhance NVHPC to be a full toolchain:
* compilers provided by new `nvidia-compilers` subtoolchain
    * equivalent to current NVHPC
    * `nvidia-compilers` will have their own easyconfigs
* mpi provided by new `NVHPCX` subtoolchain
    * `NVHPCX` is bundled in NVHPC so it will not exist with their own easyconfigs
* linalg provided by new `NVBLAS` subtoolchain
    * `NVBLAS` is bundled in NVHPC so it will not exist with their own easyconfigs

I also added some extra features related to NVHPC:
* `cuda_compute_capabilities` can be defined from environment variable `$EBNVHPCCUDACC`: this environment variable can be set by `nvidia-compilers` or `NVHPC` if they are installed with a given `cuda_compute_capabilities`, avoiding the need to manually set `cuda_compute_capabilities` for every installation in that toolchain.
* CUDA templates (_e.g._ `%(cudaver)s`) will be defined from environment variable `$EBNVHPCCUDAVER`: this environment variable will be set by `nvidia-compilers` or `NVHPC`, allowing to use those templates in all easyconfigs in the toolchain without having to depend on CUDA.
* new template `%(cuda_cc_nvhpc)s` for CUDA compute capabilities with the format expected by NVHPC

PR #5073, opened 2025-12-14
fix subtoolchains for nvompi + old NVHPC toolchains, and mark nvidia-compilers as optional subtoolchain since it's only used for recent NVHPC versions

boegel:
follow-up for:
- #4927

Required because of old toolchains like `nvompi/2022.07` (and `nvofbf/2022.07`) and `NVHPC` versions < 25.0


ocaisa (approved):
LGTM

PR #5074, opened 2025-12-16
Include full shell command that failed in raised error

Flamefire:
Allow to quickly inspect the command in the log, especially in test reports.


boegel:
The full command can be quite long, that's exactly why we don't fully include it anymore since EasyBuild v5.0

Flamefire:
Makes sense for `EasyBuildError`. How about adding a `log.error` so one can see it in the log at least?

See https://gist.github.com/boegelbot/276c79e4105f6ed3d4cb633010ef30b3 for the motivation

PR #4468, opened 2024-02-29
add support for `--from-commit` and `--include-easyblocks-from-commit`

boegel:
WIP because some things are still missing:

* [x] enhanced test for `download_repo` function to cover new `commit` option;
* [x] dedicated test for `fetch_files_from_commit` and `fetch_easyconfigs_from_commit` functions;
* [x] dedicated test for `--from-commit` option;
* [x] also implement `--include-easyblocks-from-commit` option + dedicated test;
* [x] fix combo of `--copy-ec` and `--from-commit`

PR #5077, opened 2025-12-18
fix ignoring of URLError in `test_copy_ec_from_commit`

Flamefire:
When the exception is thrown and ignored `stdout` is unassigned so the following tests cannot be done.

Fix regression from #4468.

See https://github.com/easybuilders/easybuild-framework/actions/runs/20333704730/job/58414942098?pr=5045

> UnboundLocalError: local variable 'stdout' referenced before assignment

Note that CI will still fail with
> ERROR: Found printed messages in output of test suite

And similar:
> ................................................................................................................................................................................................................Skipping online test for det_file_size (working offline)

cc @boegel


boegel (approved):
lgtm

PR #5079, opened 2025-12-19
Fix initialization order in `ModulesTool`, tests and don't set  `$LMOD_REDIRECT='no'`

Flamefire:
`self.set_mod_paths(mod_paths)` calls the modules command which isn't fully initialized at this point.   
This isn't caught as we don't pass that parameter in any usage (so far)

So move that around.

Also the `test_lmod_specific` might (almost?) always be skipped because `lmod` is not in `$PATH`, but only in `$LMOD_CMD`   
Additionally parts of the modules-specific tests can be run even if that modules tool is not available. So only skip what has to be skipped.

Finally revert https://github.com/easybuilders/easybuild-framework/pull/1755: `$LMOD_REDIRECT` only affects the `module` function which we do not use. So do not set that variable to "no" which makes working in the debug shell easier.

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -615,10 +615,6 @@ class ModulesTool:
         else:
             env_cmd_path = None
 
-        self.mod_paths = None
-        if mod_paths is not None:
-            self.set_mod_paths(mod_paths)
-
         if env_cmd_path:
             cmd_path = which(self.cmd, log_ok=False, on_error=IGNORE)
             # only use command path in environment variable if command in not available in $PATH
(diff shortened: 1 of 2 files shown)


boegel:
> Finally revert https://github.com/easybuilders/easybuild-framework/pull/1755: $LMOD_REDIRECT only affects the module function which we do not use. So do not set that variable to "no" which makes working in the debug shell easier.

@Flamefire How sure are you about this?

I see being used in https://github.com/TACC/Lmod/blob/main/shells/Rc.lua for example, that seems beyond the `module` shell function to me...

I also don't see why keeping `$LMOD_REDIRECT` set to `no` would be problematic in any way, better be cautious with this?

Flamefire:
Reasonably sure. The code you quote is actually a case **for** reverting that:

It seems to handle "echo" commands in the module file:

- `LMOD_REDIRECT=no`: print directly to stderr
- Otherwise: Convert to `echo '<msg>'` and print that to stdout

In EB we care about stderr. Not setting `LMOD_REDIRECT=no` avoids capturing `echo` output of modules

Issue #5086, opened 2026-01-09
environment is not cleaned up after `build_and_install_software`


Thyre:
Recently, I've been getting more and more spurious failures of EasyBuild when trying to use `--upload-test-report`. Error messages might include not correctly being able to access my GitHub token, or this crash:

```
== COMPLETED: Installation ended successfully (took 9 mins 17 secs)
== Results of the build can be found in the log file(s) /data/EasyBuild-develop/software/X11/20250608-GCCcore-14.3.0/easybuild/easybuild-X11-20250608-20260109.184743.log

EasyBuild crashed! Please consider reporting a bug, this should not happen...

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/data/EasyBuild/lib/python3.13/site-packages/easybuild/main.py", line 859, in <module>
    main_with_hooks()
    ~~~~~~~~~~~~~~~^^
  File "/data/EasyBuild/lib/python3.13/site-packages/easybuild/main.py", line 844, in main_with_hooks
    exit_code: EasyBuildExit = main(args=args, prepared_cfg_data=(init_session_state, eb_go, cfg_settings))
                               ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/EasyBuild/lib/python3.13/site-packages/easybuild/main.py", line 795, in main
    is_successful = process_eb_args(orig_paths, eb_go, cfg_settings, modtool, testing, init_session_state,
                                    hooks, do_build)
  File "/data/EasyBuild/lib/python3.13/site-packages/easybuild/main.py", line 629, in process_eb_args
    test_report_msg = overall_test_report(ecs_with_res, len(paths), overall_success, success_msg, init_session_state)
  File "/data/EasyBuild/lib/python3.13/site-packages/easybuild/tools/testing.py", line 459, in overall_test_report
    txt = post_pr_test_report(pr_nrs, GITHUB_EASYCONFIGS_REPO, test_report, msg, init_session_state,
                              success)
  File "/data/EasyBuild/lib/python3.13/site-packages/easybuild/tools/testing.py", line 372, in post_pr_test_report
    gpu_info = get_gpu_info()
[...]

PR #5088, opened 2026-01-09
Use initially stored environment to determine GPU info for test reports

Thyre:
Tools like `amd-smi` and `rocm-smi` rely on having a clean environment to correctly pick up information. However, the last build in `build_and_install_software` leaves a tampered environment, potentially breaking both commands.
To still correctly determine the GPU information, introduce a variable allowing to use the initial environment instead. Also use this for `nvidia-smi`, as its, by now, typically installed on the system level and not as a module.

Ref. https://github.com/easybuilders/easybuild-framework/issues/5086

### DIFF
--- a/easybuild/main.py
+++ b/easybuild/main.py
@@ -155,6 +155,8 @@ def summary(ecs_with_res):
 def build_and_install_software(ecs, init_session_state, exit_on_failure=True, testing=False):
     """
     Build and install software for all provided parsed easyconfig files.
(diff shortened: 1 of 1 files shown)


Thyre:
Successful test report with this PR e.g. here: https://github.com/easybuilders/easybuild-easyconfigs/pull/24982#issuecomment-3730128356

boegel (approved):
lgtm

PR #5089, opened 2026-01-13
Add back `easybuild.toolchains.compiler.nvhpc`

Flamefire:
`NVHPC` was replaced by `NvidiaCompilers` in EasyBuild 5.2.0 (f1761f83d9e87ad65366858fc95771faff9a6bbf) which is an accidental breaking change.
Add it back with a deprecation notice making sure both are equal for now.

### DIFF
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -3398,18 +3398,18 @@ class ToolchainTest(EnhancedTestCase):
         with self.temporarily_allow_deprecated_behaviour(), self.mocked_stdout_stderr():
             from easybuild.toolchains.compiler.nvhpc import NVHPC
             self.assertIn("nvhpc was replaced by easybuild.toolchains.compiler.nvidia_compilers", self.get_stderr())
-            
+
         # load NVidiaCompilers, deprecated NVHPC corresponds to it
         tc = NvidiaCompilers(name='NvidiaCompilers', version='2024a')  # Common usage
         # Might be checked by pre-5.2.0 users
         self.assertIsInstance(tc, NVHPC)
 
-        # load deprecated NVHPC, NvidiaCompilers corresponds to it 
+        # load deprecated NVHPC, NvidiaCompilers corresponds to it
         with self.temporarily_allow_deprecated_behaviour(), self.mocked_stdout_stderr():
             tc = NVHPC(name='NVHPC', version='2024a')  # Might be used by pre-5.2.0 users
             self.assertIn("nvhpc was replaced by easybuild.toolchains.compiler.nvidia_compilers", self.get_stderr())
         self.assertIsInstance(tc, NvidiaCompilers)
-        
+
         # load deprecated NVHPCToolchain, it corresponds to NvidiaCompilersToolchain
         with self.temporarily_allow_deprecated_behaviour(), self.mocked_stdout_stderr():
             tc = NVHPCToolchain(name='NVHPC', version='2024a')  # Might be used by pre-5.2.0 users
@@ -3420,7 +3420,7 @@ class ToolchainTest(EnhancedTestCase):
         self.assertIsInstance(tc, NvidiaCompilers)
         self.assertIsInstance(tc, NvidiaCompilersToolchain)
         self.assertIsInstance(tc, NVHPCToolchain)
-        # deprecated in 5.2.0
+        # compiler toolchain NVHPC is deprecated in 5.2.0, but might still be checked by existing code
         self.assertIsInstance(tc, NVHPC)
 
 


easybuild/toolchains/compiler/nvhpc.py
```diff
@@ -0,0 +1,36 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""Compatibility module such that compiler.nvhpc.NVHPC and compiler.compiler.nvidia_compilers.NvidiaCompilers
+can be used interchangeably
+"""
+
+from easybuild.base import fancylogger
+from easybuild.toolchains.compiler.nvidia_compilers import NVHPC  # noqa # pylint:disable=unused-import
```
lexming: I prefer to not pollute `nvidia_compilers`.

We can just import the NvidiaCompilers class in here as is. Then reassign with the NVHPC name in this deprecated module. Or am I missing something?

easybuild/toolchains/compiler/nvidia_compilers.py
```diff
@@ -45,7 +46,8 @@
 class NvidiaCompilers(Compiler):
     "NVHPC compiler class"
 
-    COMPILER_MODULE_NAME = ['nvidia-compilers']
+    # TODO EasyBuild 6.0: Remove NVHPC name
+    COMPILER_MODULE_NAME = ['nvidia-compilers', 'NVHPC']
```
boegel, line 49: @Flamefire Hmmm, is this necessary?

Doesn't this mean that the `NVHPC` module *must* be available in order to use `nvidia-compilers` as compiler component in a toolchain?

cc @lexming

Flamefire: If I understand the code correctly it is an "either". Not fully sure though.

easybuild/toolchains/compiler/nvhpc.py
Flamefire: With the ABC approach this now works yes. Prior to that I had circular imports

Done here and for `NVHPCToolchain`

test/framework/toolchain.py
```diff
@@ -3389,6 +3387,37 @@ def test_get_flag(self):
...
+        self.assertIsInstance(tc, NvidiaCompilers)
+        self.assertIsInstance(tc, NvidiaCompilersToolchain)
+        # Might be checked by pre-5.2.0 users
+        self.assertIsInstance(tc, NVHPC)
+        self.assertIsInstance(tc, NVHPCToolchain)
+
+        with self.temporarily_allow_deprecated_behaviour(), self.mocked_stdout_stderr():
+            tc = NVHPCToolchain(name='NVHPC', version='2024a')  # Might be used by pre-5.2.0 users
+            self.assertIn("NVHPCToolchain was replaced by NvidiaCompilersToolchain", self.get_stderr())
+        self.assertIsInstance(tc, NvidiaCompilersToolchain)
```
lexming:
```suggestion
        from easybuild.toolchains.nvompi import Nvompi
        from easybuild.toolchains.compiler.nvidia_compilers import NvidiaCompilers
        from easybuild.toolchains.nvhpc import NvidiaCompilersToolchain, NVHPCToolchain

        # test deprecation of NVHPC in v5.2.0
        with self.temporarily_allow_deprecated_behaviour(), self.mocked_stdout_stderr():
            from easybuild.toolchains.compiler.nvhpc import NVHPC
            self.assertIn("nvhpc was replaced by easybuild.toolchains.compiler.nvidia_compilers", self.get_stderr())
            
        # load NVidiaCompilers, deprecated NVHPC corresponds to it
        tc = NvidiaCompilers(name='NvidiaCompilers', version='2024a')  # Common usage
        # Might be checked by pre-5.2.0 users
        self.assertIsInstance(tc, NVHPC)

        # load deprecated NVHPC, NvidiaCompilers corresponds to it 
        with self.temporarily_allow_deprecated_behaviour(), self.mocked_stdout_stderr():
            tc = NVHPC(name='NVHPC', version='2024a')  # Might be used by pre-5.2.0 users
            self.assertIn("nvhpc was replaced by easybuild.toolchains.compiler.nvidia_compilers", self.get_stderr())
        self.assertIsInstance(tc, NvidiaCompilers)
        
        # load deprecated NVHPCToolchain, it corresponds to NvidiaCompilersToolchain
        with self.temporarily_allow_deprecated_behaviour(), self.mocked_stdout_stderr():
            tc = NVHPCToolchain(name='NVHPC', version='2024a')  # Might be used by pre-5.2.0 users
            self.assertIn("NVHPCToolchain was replaced by NvidiaCompilersToolchain", self.get_stderr())
        self.assertIsInstance(tc, NvidiaCompilersToolchain)

        tc = Nvompi(version='2024a')  # Common usage
        self.assertIsInstance(tc, NvidiaCompilers)
        self.assertIsInstance(tc, NvidiaCompilersToolchain)
        self.assertIsInstance(tc, NVHPCToolchain)
        # deprecated in 5.2.0
        self.assertIsInstance(tc, NVHPC)
```

easybuild/toolchains/compiler/nvidia_compilers.py
lexming: indeed, this seems more of a convenience attribute in the `Compiler` class, with no defined purpose. Only real use in framework is to determine the compiler in optarch, which is compatible with this being a list of possible compilers and not all of them: https://github.com/easybuilders/easybuild-framework/blob/fffc9a70efd784d31e36741971c4d7a7972c75bb/easybuild/tools/toolchain/compiler.py#L338

lexming:
Merging, thanks @Flamefire !

PR #1218, opened 2015-03-11
derive EasyBuildError from LoggedException, deprecate log.error in favor of raise EasyBuildError

boegel:
replaces #1183
depends on ~~https://github.com/hpcugent/vsc-base/pull/160~~ https://github.com/hpcugent/vsc-base/pull/163

update: see http://easybuild.readthedocs.org/en/latest/Deprecated-functionality.html#report-errors-by-raising-easybuilderror-rather-than-using-log-methods

PR #5092, opened 2026-01-14
Deprecate `print_error` function and replace it with new `print_error_and_exit` function

Flamefire:
There is a `print_error` function in `output` and this one additionally
exits which is not reflected in the name.
Also the handling of the parameters is confusing, e.g. using `log` to
determine whether to throw an `EasyBuildError`.
Deprecate it and replace by `print_error_and_exit`, translating the
arguments from one to the other.

CI error is due to use of `log.error` which is disallowed since https://github.com/easybuilders/easybuild-framework/pull/1218

Only `easybuild.base` seems to be exempted with a comment (which seems to be outdated)
> `# don't check Python modules in easybuild.base namespace (ingested vsc-base)`

That is also the source of the weird construct:
```
elif log is not None:
        raise EasyBuildError(message)
```

TBH I don't really like having `EasyBuildError` log an error in all cases as we sometimes catch it and do something else instead. But maybe this just means we should not raise it so often.

I removed the `log` parameter (again)

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -395,7 +395,8 @@ def print_error(msg, *args, **kwargs):
     Print error message and exit EasyBuild
     """
 
-    _init_easybuildlog.deprecated("Function 'print_error' from easybuild.tools.build_log is replaced with 'print_error_and_exit'", '6.0')
+    _init_easybuildlog.deprecated("Function 'print_error' from easybuild.tools.build_log is replaced "
+                                  "with 'print_error_and_exit'", '6.0')
 
     # grab exit code, if specified;
     # also consider deprecated 'exitCode' option


easybuild/tools/build_log.py
```diff
@@ -394,8 +394,8 @@ def print_error(msg, *args, **kwargs):
     """
     Print error message and exit EasyBuild
     """
-    if args:
-        msg = msg % args
+
+    _init_easybuildlog.deprecated("Function 'build_log.print_error' is replaced with 'print_error_and_exit'", '6.0')
```
boegel:
```suggestion
    _init_easybuildlog.deprecated("Function 'print_error' is replaced with 'print_error_and_exit'", '6.0')
```

easybuild/main.py
```diff
@@ -428,9 +428,12 @@ def process_eb_args(eb_args, eb_go, cfg_settings, modtool, testing, init_session
...
     else:
-        print_error("Please provide one or multiple easyconfig files, or use software build " +
-                    "options to make EasyBuild search for easyconfigs",
-                    log=_log, opt_parser=eb_go.parser, exit_on_error=not testing)
+        msg = ("Please provide one or multiple easyconfig files, or use software build "
+               "options to make EasyBuild search for easyconfigs")
+        if testing:
+            raise EasyBuildError(msg)
+        eb_go.parser.print_shorthelp()
+        print_error_and_exit(msg)
```
boegel, line 436: End out produced output will look like this:
```
All long option names can be passed as environment variables. Variable name is
EASYBUILD_<LONGNAME> eg. --some-opt is same as setting EASYBUILD_SOME_OPT in
the environment.
ERROR: Please provide one or multiple easyconfig files, or use software build options to make EasyBuild search for easyconfigs
```

So the error message "drowns" a bit because it blends in with the short help output.

Let's add one (or two) newlines in between?

test/framework/build_log.py
```diff
@@ -275,19 +275,26 @@ def run_check(args, silent=False, expected_stderr='', **kwargs):
         log_txt = read_file(tmp_logfile)
         self.assertIn("WARNING Test log message with a logger involved.", log_txt)
 
-    def test_print_error(self):
+    def test_print_error_and_exit(self):
         """Test print_error"""
```
boegel:
```suggestion
        """Test print_error_and_exit and (deprecated) print_error functions"""
```

easybuild/main.py
Flamefire: Note that this was the case before.

We have `output.print_error` which adds a newline before and after the message. Would be great if we could use that. A bit confusing that `print_error` are in `output` but `print_msg` and `print_warning` in `build_log`. I'd move all to one of them but we can't as we either get a cycle or break backwards compat with `build_log.print_error` (which used to exit)

I'm now calling `output.print_error` from here to have a uniform appearance of printed errors.

easybuild/tools/build_log.py
Flamefire: We also have `tools.output.print_error` which is why I wanted to disambiguate this

boegel: Then I'd go for something like:
```suggestion
    _init_easybuildlog.deprecated("Function 'print_error' from easybuild.tools.build_log is replaced with 'print_error_and_exit'", '6.0')
```

easybuild/tools/build_log.py
```diff
@@ -415,13 +413,29 @@ def print_error(msg, *args, **kwargs):
...
+            exit_code = EasyBuildExit.ERROR
+        print_error_and_exit(msg, *args, exit_code=exit_code, silent=silent)
     elif log is not None:
-        raise EasyBuildError(msg)
+        raise EasyBuildError(msg)  # Handle legacy weirdness
+
+
+def print_error_and_exit(msg, *args, exit_code=EasyBuildExit.ERROR, silent=False):
+    """
+    Print error message and exit EasyBuild, supports format strings
```
boegel, line 427: The "supports format strings" is a lie since we disable use of `rich`?

Why is that done BTW, can you clarify?
Is it mainly to avoid Rich messing things up, similar to #5098 ?

Flamefire: > The "supports format strings" is a lie since we disable use of rich?

This is related to `msg, *args` -> `msg = msg % args`

Disabling of `rich` is done so we can use this for any error, even failing to use `rich`. I wanted to make sure this always succeeds and does not cause another error.

PR #5093, opened 2026-01-15
Properly handle toolchains defining `openmp` toolchain opt as `dict` (relevant for NVHPC toolchain)

Thyre:
In 956f038cbbe646cdfe83a9151cd9029ee918958f, we've added a dict for NVHPC to not only enable OpenMP when `'openmp': True` is passed, but also _disable_ OpenMP when either `'openmp': False` or no value is passed.

This is due to NVHPC enabling OpenMP _by default_, causing issues with e.g. programs using the OpenMP Tools Interface as well.

Unfortunately, the handling of arguments in `_set_compiler_flags` didn't consider OpenMP as an argument where toolchains can pass a dict. Because of that, enabling OpenMP with `nvidia-compilers` or `NVHPC` causes the following effect, described by @pavelToman:

> If I have `toolchainopts = {'openmp': True}` in ELPA FLAGS there are `False True`:
> ```
> eb-shell> echo $CFLAGS
> -O2 -tp=host -Mflushz False True
>```
> When I have `toolchainopts = {'openmp': False}` then it seems ok with:
>```
> eb-shell> echo $CFLAGS
> -O2 -tp=host -Mflushz
>```
> but I think there is missing `-nomp` flag. When I delete the `'openmp': {False: '-nomp', True: '-mp'}` from the `COMPILER_UNIQUE_OPTION_MAP` and run ELPA with `toolchainopts = {'openmp': True}`  I've got:
>```
>eb-shell> echo $CFLAGS
>-O2 -tp=host -Mflushz -fopenmp
>```

To fix this, copy the handling of `vectorize` to also be used for `openmp`. This requires removing `openmp` from `COMPILER_OPTIONS` to avoid the default handling earlier on. Since many toolchains simply specify a string or list of strings, keep backwards compatibility by converting them to a dict, passing an empty string for `'openmp': False` (which is the default).

Also harden both `openmp` and `vectorize` against bogus passed values to avoid EasyBuild crashing with a KeyError.

### DIFF
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -528,7 +528,7 @@ class ToolchainTest(EnhancedTestCase):
 
         pass_by_value = True
         ids = []
-        for k, v in tc.variables.items():
+        for _, v in tc.variables.items():
             for x in v:
                 idx = id(x)
                 if idx not in ids:
@@ -908,8 +908,6 @@ class ToolchainTest(EnhancedTestCase):
                 flag = tc.COMPILER_UNIQUE_OPTION_MAP[opt]
                 if isinstance(flag, list):
                     flag = ' '.join(flag)
-                else:
-                    flag = flag
                 for var in flag_vars:
                     flags = tc.get_variable(var)
                     if enable:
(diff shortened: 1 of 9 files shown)


Thyre:
```
FAIL: test_independence (test.framework.toolchain.ToolchainTest.test_independence)
Test independency of toolchain instances.
  File "/tmp/runner/41bf98bba00dbba1b05ea742347ce4813c2c7ddf/lib/python3.11/site-packages/test/framework/toolchain.py", line 2338, in test_independence
    self.assertEqual(str(tc.variables['CFLAGS']), expected_cflags, msg)
  File "/tmp/runner/41bf98bba00dbba1b05ea742347ce4813c2c7ddf/lib/python3.11/site-packages/easybuild/base/testing.py", line 89, in assertEqual
    super().assertEqual(a, b, msg=msg)
AssertionError: '-O2 -ftree-vectorize -fopenmp -test -fno-math-errno' != '-O2 -ftree-vectorize -test -fno-math-errno -fopenmp'
- -O2 -ftree-vectorize -fopenmp -test -fno-math-errno
?                      ---------
+ -O2 -ftree-vectorize -test -fno-math-errno -fopenmp
?                                           +++++++++
 : Expected $CFLAGS found for toolchain GCC: -O2 -ftree-vectorize -test -fno-math-errno -fopenmp
```

That test seems to be brittle...
I'll take a look. Maybe worth making this test independent of the order of arguments.

easybuild/tools/toolchain/compiler.py
```diff
@@ -126,7 +126,7 @@ class Compiler(Toolchain):
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = None
     COMPILER_GENERIC_OPTION = None
 
-    COMPILER_OPTIONS = ['debug', 'ieee', 'openmp', 'pic', 'shared', 'static', 'unroll', 'verbose']  # any compiler
```
boegel, line 129: Hmm, removing it from this list looks a bit "scary"...

Couldn't this result in some fallout?

boegel (dismissed):
I would definitely like to see this fix covered by the tests, i.e. we should verify that `openmp` toolchain options has the correct effect for a toolchain based on `nvidia-compilers`...

easybuild/tools/toolchain/compiler.py
Thyre: Checking for `COMPILER_OPTIONS`, we're only using it [here](https://github.com/easybuilders/easybuild-framework/blob/2798877ff8221da6eb2af1b902150c35a2105e82/easybuild/tools/toolchain/compiler.py#L254) and in Cray + LLVM to extend the list.

So we should be safe here, though I understand that removing a common option is scary.

In the long run, maybe it would make sense to allow a `dict` for all the options, removing the need to implement special handling for `vectorize` and `openmp` only? We should certainly keep the option to just specify a string though, similar to the approach I've implemented for `openmp` now.

Thyre:
@boegel, tried to add a test for OpenMP & vectorize with `nvidia-compilers`. Lets see if this also passes in the CI...

test/framework/easyconfigs/test_ecs/n/nvidia-compilers/nvidia-compilers-25.9.eb
```diff
@@ -0,0 +1,45 @@
+# should be EB_NVIDIAMINCOMPILERS, but OK for testing purposes
```
Thyre:
```suggestion
# should be EB_nvidia_minus_compilers, but OK for testing purposes
```

test/framework/toolchain.py
```diff
@@ -1405,6 +1405,44 @@ def test_fosscuda(self):
         # check CUDA runtime lib
         self.assertIn("-lrt -lcudart", tc.get_variable('LIBS'))
 
+    def test_nvidia_compilers(self):
+        """Test whether nvidia-compilers is handled properly."""
+        # Test OpenMP support
+        openmp_true = "-mp"
+        openmp_false = "-nomp"
+        openmp_none = "-nomp"
```
lexming: this is a bit confusing to follow, I suggest to use a dict with all the cases and expected values:
```suggestion
        openmp_cases = {
            # (input parameter, resulting flag)
            'true': ({'openmp': True}, "-mp")
            'false': ({'openmp': False}, "-nomp")
            'none': ({}, "-nomp"),
        }
```

test/framework/toolchain.py
```diff
@@ -1405,6 +1405,44 @@ def test_fosscuda(self):
...
+        general_optflags = "-O2 %s %s -Mflushz"
+        # Create new toolchain object in each iteration to start from clean state
+        for opts, omp_flag in [({}, openmp_none), ({'openmp': True}, openmp_true), ({'openmp': False}, openmp_false)]:
```
lexming: then this can be rewritten as 
```suggestion
        for opts, omp_flag in openmp_cases.values():
```

test/framework/toolchain.py
```diff
@@ -1405,6 +1405,44 @@ def test_fosscuda(self):
...
+            tc = self.get_toolchain("nvidia-compilers", version="25.9")
+            archflags = tc.COMPILER_OPTIMAL_ARCHITECTURE_OPTION[(tc.arch, tc.cpu_family)]
+            tc.set_options(opts)
+            with self.mocked_stdout_stderr():
+                tc.prepare()
+            optflags = general_optflags % (omp_flag, archflags)
+            # strip double spaces
+            optflags = optflags.replace("  ", " ")
+            val = tc.get_variable('CFLAGS')
+            self.assertTrue(re.compile(optflags).match(val), "'%s' matches '%s'" % (val, optflags))
```
lexming: this seems too much complicated, why not just assert that `val` contains `omp_flag`?

easybuild/tools/toolchain/compiler.py
lexming: I checked and `COMPILER_OPTIONS` is indeed only used to feed compiler flags such as `CFLAGS` and co. As long as there is an alternative code path to pass those down, we are fine.

I agree that we should standardize how those options are passed to the compiler. It is complicated to have special cases such as `vectorize` and now `openmp`. Maybe open an issue on that so that we do not forget.

test/framework/toolchain.py
Thyre: Right now, we also ensure that the other flags are present.
But agreed, it's probably fine to simplify this for only checking OpenMP & vectorize.

easybuild/tools/toolchain/compiler.py
Thyre: https://github.com/easybuilders/easybuild-framework/issues/5104

PR #5101, opened 2026-01-30
Case-insensitive CLI options

vvolkl:
The shift key is stuck on my old laptop. Please consider this PR, it will instantly avoid 1.5kg of e-waste!

### DIFF
--- a/easybuild/base/generaloption.py
+++ b/easybuild/base/generaloption.py
@@ -135,6 +135,7 @@ def check_str_list_tuple(option, opt, value):  # pylint: disable=unused-argument
     else:
         return klass(split)
 
+
 def check_choice_case_insensitive(option, opt, value):
     """
     Case-insensitive check function for choice type.


easybuild/base/generaloption.py
```diff
@@ -136,6 +136,26 @@ def check_str_list_tuple(option, opt, value):  # pylint: disable=unused-argument
...
+
+    value_lower = value.lower()
+    if value_lower in choices_map:
+        # Return the properly-cased choice from the original choices list
+        return choices_map[value_lower]
+    else:
+        choices = ", ".join(map(repr, option.choices))
+        raise OptionValueError(
+            _gettext("option %s: invalid choice: %r (choose from %s)")
+            % (opt, value, choices))
```
smoors, line 156:
```suggestion
            _gettext(f"option {opt}: invalid choice: {value} (choose from {choices})")
```

vvolkl:
> @vvolkl are you up for adding a test that uses lower case values to `test/framework/options.py`?

Sure, let me look into that

PR #5105, opened 2026-02-04
change NVHPCToolchain deprecation warning to simple warning in unit tests

lexming:
CI test suite in `easybuild-easyconfigs` is currently broken due to this deprecation warning adn the recent changes in https://github.com/easybuilders/easybuild-framework/pull/5103.

Instead of mending the tens of test failing to blindly ignore all deprecation errors, we can avoid the error altogether by switching to a regular warning in the test suite.

### DIFF
--- a/easybuild/toolchains/nvhpc.py
+++ b/easybuild/toolchains/nvhpc.py
@@ -38,6 +38,7 @@ from easybuild.toolchains.linalg.nvblas import NVBLAS
 from easybuild.toolchains.linalg.nvscalapack import NVScaLAPACK
 from easybuild.toolchains.mpi.nvhpcx import NVHPCX
 from easybuild.toolchains.nvidia_compilers import NvidiaCompilersToolchain
+from easybuild.tools.build_log import print_warning
 from easybuild.tools.toolchain.toolchain import SYSTEM_TOOLCHAIN_NAME
 
 
@@ -71,6 +72,6 @@ class NVHPCToolchain(NvidiaCompilersToolchain):
         warn_msg = "NVHPCToolchain was replaced by NvidiaCompilersToolchain in EasyBuild 5.2.0"
         in_test_env = any('unittest' in frame.filename for frame in inspect.stack())
         if in_test_env:
-            self.log.warning(warn_msg)
+            print_warning(warn_msg)
         else:
             self.log.deprecated(warn_msg, '6.0')


bartoldeman:
CI fails in the check for the deprecation message now, maybe simplest to remove that check, since it's already checked elsewhere?

easybuild/toolchains/nvhpc.py
```diff
@@ -65,4 +68,10 @@ class NVHPCToolchain(NvidiaCompilersToolchain):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.log.deprecated("NVHPCToolchain was replaced by NvidiaCompilersToolchain in EasyBuild 5.2.0", '6.0')
+        # print deprecation warning (stick to warning level in CI tests)
+        warn_msg = "NVHPCToolchain was replaced by NvidiaCompilersToolchain in EasyBuild 5.2.0"
+        in_test_env = any('unittest' in frame.filename for frame in inspect.stack())
+        if in_test_env:
+            print_warning(warn_msg)
```
Flamefire, line 75: Doesn't this require adapting the easyconfig tests such that this printed message does not cause GHA CI failure?

lexming: That approach does not allow to discriminate between different sources of deprecation errors, it is either all or nothing. Since we do not want to blindly allow all deprecation errors in those tests, it is better to just adapt this specific deprecation at its origin.

Flamefire: What I meant: In GHA CI we fail if any message is printed to stdout/stderr.
With this change we print a message to stderr. So I expect the CI to fail even though tests succeed.

However I just remembered that we do this only for the framework CI so this works.

@Thyre said in Slack that only 25.x and later should use `NvidiaCompilersToolchain`. Shouldn't the deprecation then be targeted to the specific version? Then it would make this check not necessary as our official ECs won't run into this deprecation. Or is it impossible to check for a version at this point?

Thyre: It's not really that older versions cannot use `nvidia-compilers`, but we do not provide EasyConfigs for these earlier versions. Theoretically, these could be provided as well.

lexming: There is no need to explicitly check versions, the recent changes in #5103 detect easyconfigs using the older NVHPC toolchain based on the dependency tree. So this deprecation for `NVHPCToolchain` will only be show to those older versions.

Flamefire: @lexming The deprecation will also be shown for `HDF5-1.12.2-NVHPC-22.7-CUDA-11.7.0-serial.eb` which we provide. Hence the idea to deprecate only for 25+ and not show anything for e.g. 22.7 which avoids us having to check for being run from tests

lexming: indeed, you will get a deprecation warning if you ever reinstall `HDF5-1.12.2-NVHPC-22.7-CUDA-11.7.0-serial.eb` with future EB releases.
We need to pass the message that those easyconfigs will not just be archived at some point, but future EB will probably be incapable of installing those even if you keep a local copy.

PR #5108, opened 2026-02-08
don't require installation prefix for toolchain dependencies that are an external module

### LINKED ISSUE #5106: EasyBuild 5.2.0 can no longer load toolchains that have EXTERNAL modules as dependencies (issue on LUMI)
As discussed with Kenneth in a teams-call and a visit: Our Cray toolchain modules use a number of dependencies that are modules from the Cray Programming Environment hence external modules to LUMI. This causes Easybuild 5.2 to fail around the end of the prepare step when compiling packages that use the toolchains. The underlying issue is that while EBROOT variables are automatically generated for direct dependencies of a package, they are not generated for dependencies of dependencies, or in this case, the dependencies of the toolchain.

Reproducer: See https://github.com/klust/LUMI-EasyBuild-user/tree/main/easybuild/easyconfigs/k/Kenneth and the file https://github.com/klust/LUMI-EasyBuild-user/blob/main/easybuild/easyconfigs/k/Kenneth/easybuild_toolchain_issue.md in that directory that contains information on how to reproduce on LUMI and more analysis of the issue.

Note: A workaround for us that is not very elegant but workable, could be to extend our toolchain easyconfigs/modules to set the EBROOT variables for certain dependencies via a modluafooter. These toolchain modules are generated automatically anyway, so this would not be too error-prone. It might be a little tricky though to give those EBROOT variables sensible content.

boegel:
fixes #5106

### DIFF
--- a/test/framework/toolchain.py
+++ b/test/framework/toolchain.py
@@ -34,6 +34,7 @@ import shutil
 import stat
 import sys
 import tempfile
+import textwrap
 from itertools import product
 from unittest import TextTestRunner
 from test.framework.utilities import EnhancedTestCase, TestLoaderFiltered, find_full_path, init_config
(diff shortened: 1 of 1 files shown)


boegel:
converted to draft PR because I want to add a test that triggers this bug...

Issue #5110, opened 2026-02-10
Failing CI in Python >= 3.12


Crivella:
The problem stems from the removal of `pkg_resources` from Python 3.12 (moved to setuptools) and from setuptools >= 82.0.0
https://setuptools.pypa.io/en/latest/history.html#v82-0-0

See for example

https://github.com/easybuilders/easybuild-framework/actions/runs/21864109089/job/63100630517?pr=5036

### Python 3.11

```
(a) crivella@crivella-desktop:~$ python
Python 3.11.14 (main, Oct 10 2025, 08:54:03) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygraph.classes.digraph
/home/crivella/.virtualenvs/a/lib/python3.11/site-packages/pygraph/mixins/__init__.py:33: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  __import__('pkg_resources').declare_namespace(__name__)
>>>

$ pip freeze
pydot==4.0.1
pyparsing==3.3.2
python-graph-core==1.8.2
python-graph-dot==1.8.2 
```

### Python 3.12

```
(a) crivella@crivella-desktop:~$ python
Python 3.12.12 (main, Oct 10 2025, 08:52:57) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygraph.classes.digraph
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/crivella/.virtualenvs/a/lib/python3.12/site-packages/pygraph/classes/digraph.py", line 33, in <module>
    from pygraph.mixins.labeling import labeling
  File "/home/crivella/.virtualenvs/a/lib/python3.12/site-packages/pygraph/mixins/__init__.py", line 33, in <module>
    __import__('pkg_resources').declare_namespace(__name__)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'pkg_resources'
>>> 
```

Fixing the version of setuptools to 81 should fix this (or we need to move away from `python-graph-dot`

boegel:
There's little hope that this will be fixed in `python-graph-dot`, last release was in 2012...
https://pypi.org/project/python-graph-dot/#history

[...]

PR #5111, opened 2026-02-10
Fix broken CI for Python>=3.12

Crivella:
Fixes https://github.com/easybuilders/easybuild-framework/issues/5110


.github/workflows/unit_tests.yml
```diff
@@ -129,7 +129,7 @@ jobs:
         pip install -r requirements.txt
         if ! python -c "import distutils" 2> /dev/null; then
             # we need setuptools for distutils in Python 3.12+, needed for python setup.py sdist
-            pip install --upgrade setuptools
+            pip install --upgrade setuptools==81.0.0
```
boegel, line 132: For context: yes, `pkg_resources` was removed in `setuptools` 82.0.0, it has been deprecated for a while

see https://setuptools.pypa.io/en/latest/history.html#v82-0-0

Issue #5052, opened 2025-11-19
Building easyconfigs with many extensions is much slower with EB5 than with EB4


branfosj:
On my system, building `R-bundle-CRAN/2025.11-foss-2025b` (https://github.com/easybuilders/easybuild-easyconfigs/pull/24465):

* Latest develop (1oth November): 13:44:57
* EB 5.0.0 (`pip install easybuild==5.0.0`: 10:19:24
* EB 4.9.4 (`pip install easybuild==4.9.4`): 7:47:08

jsc-zen3 was 17:47:26, so it is not just my system being slow here.

I did a further test with an easyconfig that is the first 49 extensions from R-bundle-CRAN/2025.11-foss-2025:

- EB 4.9.4: 15:08
- EB 5.0.0: 22:27
- EB 5.1.2: 37:47
- EB 5.1.2 with changes from #4868 reverted: 25:10

So there is a slowdown from EB 4 -> 5, but the major slowdown is directly from #4868.

**We need a better option for resetting the environment between extensions.**

`R` / `R-bundle-CRAN` / `R-bundle-Bioconductor` / `Perl-bundle-CPAN` / `Python-bundle-PyPI` will benefit enormously on install time.

boegel:
Coming back to the idea (by @Micket) to pack `module load` commands for toolchain & dependencies together: I asked in the Lmod monthly meeting today whether `module load A B C` is perfectly equivalent to `module load A; module load B; module load C`, also in the case where `A` may add additional paths to `$MODULEPATH` (like in a hierarchical module naming scheme (HMNS)).

@rtmclay ensured me that `module load A B C` works perfectly fine, also when an HMNS is used as `A` extends the set of modules available for loading.

The only real downside of `module load A B C` versus `module load A; module load B; module load C` is that with the former you won't get *any* changes in the environment if loading `C` fails, which is obviously different when the modules are loaded one by one.
That's totally fine in the context of EasyBuild though...

boegel:
The changes being made in this PR (#5211) should be a massive help, still need to benchmark the impact a bit, but it'll be quite substantial, especially on systems with a slow shared filesystem (where the module files live) and for easyconfigs with a lot of extensions (like `R-bundle-*`).

Flamefire:
> The changes being made in this PR

Which one exactly? #5211 or #5114 ?

branfosj:
> > The changes being made in this PR
> 
> Which one exactly? #5211 or #5114 ?
[...]

PR #5114, opened 2026-02-11
load list of modules in one go as opposed to running separate `module load` commands for each of them

boegel:
This should help quite a bit with resolving the performance regression reported in #5052, since all dependencies are being loaded in one go rather than one by one, at least for the "parent" installation.

We're still doing a separate `module load` for the toolchain, but I won't look into that yet since changing that is lot more involved (there are separate private methods called from `Toolchain._load_modules`: `_load_toolchain_module` and `_load_dependencies_modules`).

As I write this (rubber ducking 🦆 ) , I realize that I'm not sure how much impact this will have for extensions, since we only load the "fake" (temporary) module file for each extension (so a single `module load`) via `EasyBlock.fake_module_environment`...

### DIFF
--- a/easybuild/tools/modules.py
+++ b/easybuild/tools/modules.py
@@ -1127,8 +1127,8 @@ class ModulesTool:
 
         if not allow_reload:
             modules = set(modules) - set(self.loaded_modules())
-        for mod in modules:
-            self.run_module('load', mod)
+
+        self.run_module('load', *modules)
 
     def unload(self, modules, log_changes=True):
         """


boegel:
I just dug into this a bit more, partially because we're planning to get rid of our SSD-based shared scratch filesystem soonish, which is making me switch to our regular GPFS scratch filesystem for my EasyBuild test environment.

This results in a *significant* slowdown of test builds, to the point where an installation of for example `Python/3.11.3-GCCcore-12.3.0` takes ~61min now, compared to the ~13min it took before (on SSD scratch)...

Looking into this a bit, a *significant* amount of the time goes to loading modules for (build) dependencies, which can be slow because EasyBuild internally sets `$LMOD_IGNORE_CACHE` (because often the modules for the dependencies are not in the Lmod cache yet, or may never be, like for my own test environment).

For `Python/3.11.3-GCCcore-12.3.0`, a grep in the log shows 39 `module load` commands being run:
- The first 12 are the initial setup of the build environment. With the change proposed in this PR, those would reduced to 2 `module load` commands: one for the toolchain + one for all dependencies combined (and with a bit more work in `Toolchain.prepare`, it could be a single one).
- For every extension (8, in this case), there are 3 `module load` commands run when setting up the build environment for the extension: one for the "fake" `Python/...` module, plus two more because there are 2 build dependencies in this case (which don't get loaded as a dependency for the fake module).
  - This can be reduced to a single `module load` command per extension relatively easily, with some work in both `EasyBlock.install_extensions_sequential` and `EasyBlock.install_extensions_parallel`.
  - With a bit more work, it can be recuded to a single `module load` command that is used to figure out the build environment for extensions, caching it, and then restoring that before starting the installation of an extension.
- At the end, there are 3 more `module load` commands run for the `Python/...` module itself. That must be part of the sanity check, and can probably be reduced to a single one as well with a bit of effort (at least in most cases).

When environment modules & the installation for the toolchain & dependencies are on slow filesystems, it's definitely worth reducing the `module load` commands being run...
In our case, on our non-SSD scratch filesystem, some of those `module load` commands take ~10 seconds, some even take ~30 sec. Very unpredictable of course, depending on the load on the filesystem, etc.
Assuming 20 seconds each, these `module load` commands alone account for ~12min of the total installation time, roughly 1/5th of the total installation time, which is obviously absurd.

Next to the `module load` commands, there's also others:
- 11 `module use --priority` commands (each taking 10-30 sec)
  - see also https://github.com/easybuilders/easybuild-framework/pull/3636
- 11 `module unload Python` commands (each taking 10-30 sec)
- 9 `module show GCCcore/12.3.0` commands (each taking 10-30 sec)
- 11 `module --terse list` commands (but these are instant)

A big chunk of these can probably also be avoided with caching the build environment and restoring it...

I also see this as a proxy for accessing other files, i.e. the actual binaries/libraries/header files etc. that are provided by the dependencies. That can't be fixed easily, of course. We could check how many stats Lmod does for a single `module load`...

Also, we're still running Lmod 8.7, significant performance improvements have been implemented in Lmod 9.x, which would also help.

Long story short: there's clearly a lot of room for improvement here...

boegel:
superseded by:
- #5211

PR #5118, opened 2026-02-12
make sure that test commands used in `test_cases` step are executable

Flamefire:
Scripts specified in the `tests` easyconfig parameter, especially from PRs, might not be executable.

Temporarily make them executable to avoid failing with permission issues.

Add test for this step (currently missing)

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -4668,13 +4668,27 @@ class EasyBlock:
                     test_cmd = os.path.join(source_path, self.name, test)
                     if os.path.exists(test_cmd):
                         break
-                if not os.path.exists(test_cmd):
-                    raise EasyBuildError(f"Test specifies invalid path: {test_cmd}")
+                else:
+                    test_cmd = test
+            if not os.path.exists(test_cmd):
+                raise EasyBuildError(f"Test specifies invalid path: {test_cmd}")
 
+            if os.path.isfile(test_cmd):
+                original_perms = os.lstat(test_cmd).st_mode
+                if original_perms & stat.S_IEXEC:
+                    original_perms = None
+                else:
+                    adjust_permissions(test_cmd, stat.S_IEXEC, add=True, recursive=False)
+            else:
+                original_perms = None
             try:
                 self.log.debug(f"Running test {test_cmd}")
-                run_shell_cmd(test_cmd)
-            except EasyBuildError as err:
+                try:
+                    run_shell_cmd(test_cmd)
+                finally:
+                    if original_perms is not None:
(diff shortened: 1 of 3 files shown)


test/framework/easyblock.py
```diff
@@ -1318,7 +1319,6 @@ def test_test_cases_step(self):
         self.assertIn('Test case success', read_file(eb.logfile))
 
         # Detect failure
-        adjust_permissions(mock_test_bin, stat.S_IWRITE)
```
Flamefire: @boegel Why did you change this? Above we have `S_IREAD` so `S_IWRITE` here makes sense too, doesn't it?

boegel: I didn't see why you were making the file writable, looks like a mistake to me, and the test passes without this.

Can you clarify why you had this there?

Flamefire: Wasn't sure, but going back to the [first commit](https://github.com/easybuilders/easybuild-framework/pull/5118/changes#diff-da513c2eb9deaf9e258f1f41b865fa8fe474273467aca7dd7c8b51dea11c1406) I had just overwritten the single test file and only in a later commit added a separate file.

So prior to that I had to make the file writable to change it for the test. Hence this was no longer necessary and the removal is fine.

easybuild/framework/easyblock.py
```diff
@@ -4668,13 +4668,27 @@ def test_cases_step(self):
                     test_cmd = os.path.join(source_path, self.name, test)
                     if os.path.exists(test_cmd):
                         break
-                if not os.path.exists(test_cmd):
-                    raise EasyBuildError(f"Test specifies invalid path: {test_cmd}")
-
+                else:
+                    test_cmd = test
+            if not os.path.exists(test_cmd):
```
boegel, line 4673: Moving this check changes the logic: if `test` was a relative command, we now require that it specified an existing path, but that wasn't the case before.

This seems to lead to trouble with PyTorch (when forcing `test_cases` step to run when using `--module-only`):
```
Test specifies non-existing path: PyTorch-check-cpp-extension.py
```

I don't think we should be requiring that the file exists if it's a relative path?

Flamefire: it should be the case before. Previously it did not require the path to exist if it was *absolute* as the check was only in the else-branch which seemed wrong

Flamefire: Issue was unrelated: The PyTorch easyblock has special handling to use `obtain_file` in the `fetch_step` resolving this path.
When skipping that step the file will not be found.

Fix for that in https://github.com/easybuilders/easybuild-framework/pull/5162

boegel (approved):
lgtm

PR #5123, opened 2026-02-19
Fix setting `$OMPI_MCA_rmaps_base_oversubscribe` in sanity check

Flamefire:
`toolchain.OPENMPI` is a string so must be checked with equals

The code is likely a C&P mistake from where it was a list of families.

On that occasion: Having the constants in `toolchain` be automatically generated is nice but breaks any linter that could detect misspellings


Crivella (approved):
LGTM

PR #5125, opened 2026-02-19
Add rich colors to `print_msg`

Crivella:
This PR is a proof of concept for adding rich coloring to the console output printed through `print_msg`

Rich will auto-recognize common patterns (eg dates, file paths) in the printed text and highlight them accordingly

Example:
image

NOTE:

Would need to adapt test to either not use `rich` or be aware of the coloring, otherwise most regex check on stdout will fail

### DIFF
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -314,6 +314,7 @@ def print_msg(msg, *args, **kwargs):
     :param newline: end message with newline
     :param stderr: print to stderr rather than stdout
     """
+    from easybuild.tools.output import use_rich  # avoid circular import
     if args:
         msg = msg % args
 
@@ -333,6 +334,16 @@ def print_msg(msg, *args, **kwargs):
 
         if newline:
             msg += '\n'
+
+        if use_rich():
+            from io import StringIO
+            from rich.markup import escape
+            from rich.console import Console
+
+            buff = StringIO()
+            console = Console(file=buff, force_terminal=True)
+            console.print(escape(msg), end='')
+            msg = buff.getvalue()
 
         if stderr:
             sys.stderr.write(msg)
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -341,7 +341,7 @@ def print_msg(msg, *args, **kwargs):
             from rich.console import Console
 
             buff = StringIO()
-            console = Console(file=buff, force_terminal=True)
+            console = Console(file=buff, force_terminal=sys.stdout.isatty())
             console.print(escape(msg), end='')
             msg = buff.getvalue()
 
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -336,14 +336,14 @@ def print_msg(msg, *args, **kwargs):
             msg += '\n'
 
         if use_rich():
-            from io import StringIO
             from rich.markup import escape
             from rich.console import Console
 
-            buff = StringIO()
-            console = Console(file=buff, force_terminal=sys.stdout.isatty())
-            console.print(escape(msg), end='')
-            msg = buff.getvalue()
+            console = Console(force_terminal=sys.stdout.isatty())
+            with console.capture() as capture:
+                console.print(escape(msg), end="")
+
+            msg = capture.get()
 
         if stderr:
             sys.stderr.write(msg)
--- a/easybuild/tools/build_log.py
+++ b/easybuild/tools/build_log.py
@@ -339,7 +339,7 @@ def print_msg(msg, *args, **kwargs):
             from rich.markup import escape
             from rich.console import Console
 
-            console = Console(force_terminal=sys.stdout.isatty())
+            console = Console()
             with console.capture() as capture:
                 console.print(escape(msg), end="")
 
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -845,7 +845,12 @@ def get_module_syntax():
 
 def get_output_style():
     """Return output style to use."""
-    output_style = build_option('output_style')
+    try:
+        output_style = build_option('output_style')
+    except EasyBuildError:
+        # If BuildOptions was not initialized yet, fall back to default and reset singleton to force re-initialization
+        BuildOptions.__class__._instances.clear()
+        output_style = OUTPUT_STYLE_BASIC
 
     if output_style == OUTPUT_STYLE_AUTO:
         if HAVE_RICH:
--- a/easybuild/tools/config.py
+++ b/easybuild/tools/config.py
@@ -677,11 +677,17 @@ def init_build_options(build_options=None, cmdline_options=None):
 def build_option(key, **kwargs):
     """Obtain value specified build option."""
 
-    build_options = BuildOptions()
-    if key in build_options:
-        return build_options[key]
-    elif 'default' in kwargs:
+    # return build option value if BuildOptions has been initialised and it's a known build option
+    if BuildOptions.__class__._instances:
+        build_options = BuildOptions()
+        if key in build_options:
+            return build_options[key]
+
+    # if above didn't return a value, return the specified default (if specified)
+    if 'default' in kwargs:
         return kwargs['default']
+
+    # if above didn't return a value, we give up
     else:
         error_msg = "Undefined build option: '%s'. " % key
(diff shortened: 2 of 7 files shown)


boegel:
```
Ran 956 tests in 718.139s

FAILED (failures=62, errors=1, skipped=3)
```

Not sure what the best approach would be to update the tests accordingly...

Maybe force disable Rich by setting some environment variable or configuring with `--color=no_color` (in `EnhancedTestCase.setUp`, for example), and then having 1-2 dedicated tests to check output with Rich enabled?

Crivella:
Yeah i think running all current tests with `--color=no_color` and add a couple that specifically test coloring would be the easiest

Crivella:
https://github.com/easybuilders/easybuild-framework/pull/5125/commits/de4f1026a12e21bc6043bdb3bdb949f62c7afd44 Should avoid breaking stuff when piping the output to a non TTY device

Possibly we might want to even move this to the `use_rich()` check

smoors:
@Crivella i tested your PR, but for some reason during the build the coloring is gone after the first line (found valid index for ...), and then comes back at the end (build succeeded for ...). adding `--output-style=rich` didn't change anything.

any idea what's wrong? is this just me?

Crivella:
For me it is working throughout all the build.
The code changes are so small that the only thing that would come to mind are:

- something is causing `use_rich` to evaluate to false in between your build
- something is altering the behavior of rich
- something is causing eb to think that its stdout is non-TTY in between the build

smoors:
it's indeed the third case: something changes during the build so that `sys.stdout.isatty()` becomes `False`.

i managed to fix it by using `sys.__stdout__.isatty()` instead. would you consider making that change?

easybuild/tools/build_log.py
```diff
@@ -334,6 +335,16 @@ def print_msg(msg, *args, **kwargs):
...
 
+        if use_rich():
+            from io import StringIO
+            from rich.markup import escape
+            from rich.console import Console
+
+            buff = StringIO()
+            console = Console(file=buff, force_terminal=sys.stdout.isatty())
+            console.print(escape(msg), end='')
+            msg = buff.getvalue()
```
smoors: you can use the rich capture helper to avoid stringio
```suggestion
            from rich.markup import escape
            from rich.console import Console

            console = Console(force_terminal=sys.stdout.isatty())
            with console.capture() as capture:
                console.print(escape(msg), end="")

            msg = capture.get()
```

Crivella:
> it's indeed the third case: something changes during the build so that sys.stdout.isatty() becomes False.
i managed to fix it by using sys.__stdout__.isatty() instead. would you consider making that change?

As far as i understand `sys.__stdout__` is supposed to be an immutable reference to the original `stdout` the program was started with, but not what the console is actually printing to.
I think checking if `sys.stdout` is TTY is the way to go, otherwise you will get colors in a stdout that, at least according to declared properties, should not support them.

Do you know what in your case is causing this behavior? Some hooks overriding stdout or is it something internal to EB that you are triggering while i am not due to configs?

smoors:
> Do you know what in your case is causing this behavior? Some hooks overriding stdout or is it something internal to EB that you are triggering while i am not due to configs?

i don't know what's causing this. i'm not doing anything special, i also disabled the hooks to be sure.

what i did find out is that with `with console.capture()` (without `StringIO`) you don't have to specify `force_terminal`, letting rich autodetect it, and that works for me.

```
force_terminal (Optional[bool], optional): Enable/disable terminal control codes, or None to auto-detect terminal. Defaults to None.
```

Crivella:
> what i did find out is that with with console.capture() (without StringIO) you don't have to specify force_terminal, letting rich autodetect it, and that works for me.

Agree we should let the auto-detect do its work seems to work for me as well, gonna commit it and lets see the CI.

easybuild/tools/build_log.py
```diff
@@ -334,6 +335,16 @@ def print_msg(msg, *args, **kwargs):
         if newline:
             msg += '\n'
 
+        if use_rich():
+            from rich.markup import escape
+            from rich.console import Console
+
+            console = Console()
```
smoors, line 342: suggestion to add some custom coloring

```suggestion
            from rich.console import Console
            from rich.theme import Theme

            theme = Theme({
                "success": "bold green",
                "fail": "bold red",
            })

            msg = escape(msg)
            for regex in [r"(\[SUCCESS\])", r"(\[PASS\])", r"( OK)$"]:
                msg = re.sub(regex, r"[success]\1[/success]", msg)
            for regex in [r"(\[FAILED\])", r"(\[FAIL\])"]:
                msg = re.sub(regex, r"[fail]\1[/fail]", msg)

            console = Console(theme=theme)
```

Crivella: I think if we want to add custom highlighting we should probably do so via a [custom highlighter](https://rich.readthedocs.io/en/stable/highlighting.html)

Would not add the extra coloring in this PR.

At that point we could even define multiple themes to be used with the same highlighter for users to choose from (that would work very well in tandem with entrypoints so it becomes very easy to define a custom theme)

boegel: I agree, we can revisit this in a subsequent PR.

I think it makes a lot of sense to use the `success` and `fail` tags that Rich supports, since I think that still allows for tweaking (through some Rich configuration).

easybuild/tools/config.py
```diff
@@ -845,7 +845,12 @@ def get_module_syntax():
 
 def get_output_style():
     """Return output style to use."""
-    output_style = build_option('output_style')
+    try:
+        output_style = build_option('output_style')
+    except EasyBuildError:
+        # If BuildOptions was not initialized yet, fall back to default and reset singleton to force re-initialization
+        BuildOptions.__class__._instances.clear()
```
boegel: @Crivella Can you clarify why this is needed?

Is it because we may be triggering `get_output_style` early through `print_msg` because we're running `use_rich()` there now?

I prefer not "exposing" internal stuff like this, maybe we should provide a function to do the reset if this is really necessary, and enhance `build_option` so it's smart to first check if `BuildOptions` have been initialized, and if not return the specified fallback value, like so:
```python
output_style = build_option('output_style', fallback_value=OUTPUT_STYLE_BASIC)

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/tools/config.py (hunk 3 above)
Comment (Crivella) [reply easybuild/tools/config.py]: Yes without that part you end up with
```
crivella@crivella-desktop:~$ eb zlib-1.3.1.eb --rebuild                                                                         

                                                                                                                                                                                                                                                                                                
ERROR: Undefined build option: 'output_style'. Make sure you have set up the EasyBuild configuration using set_up_configuration() (from easybuild.tools.options) in case you're not using EasyBuild via the 'eb' CLI.
```

we also needed to reset the `BuildOptions` instancess since `build_option` ends up initializing it before the parsing

Comment (boegel) [reply easybuild/tools/config.py]: @Crivella see https://github.com/Crivella/easybuild-framework/pull/2

File: easybuild/tools/build_log.py (hunk 2 above)
Comment (boegel) [reply easybuild/tools/build_log.py:342]: @Crivella or @smoors Can you open an issue on this, to highlight "keywords" through Rich?

I like the idea of highlighting stuff like `ERROR`/`FAIL`/`FAILED` (red), `WARNING` (yellow), `OK`/`SUCCESS`/`PASS` (green).

I wouldn't go as far as adding support for custom theming yet, but I guess it could be a step in that direction.

**Review by boegel — APPROVED** [review]
lgtm

File: easybuild/tools/config.py [hunk 4]
```diff
@@ -677,13 +677,19 @@ def init_build_options(build_options=None, cmdline_options=None):
...
+            return build_options[key]
+
+    # if above didn't return a value, return the specified default (if specified)
+    if 'default' in kwargs:
         return kwargs['default']
+
+    # if above didn't return a value, we give up
     else:
-        error_msg = "Undefined build option: '%s'. " % key
+        error_msg = "Build options are not initialized yet, or undefined build option used: '%s'. "
```
boegel, line 692: Ah, I messed that up, this was an `f"..."` string and what meant to include `{key}`, I'll open a follow-up PR to fix that...

Crivella: Right should've noticed when doing the linting MB, ping me on that PR so we get it in ASAP

boegel: fixed via https://github.com/easybuilders/easybuild-framework/pull/5149

PR #5128, opened 2026-02-20
Only make use of `graphviz` to generate dependency graphs instead of `pygraph`

Crivella:
Remove use of old/unmaintained [pygraph](https://github.com/pmatiello/python-graph/tree/master) from optional dependencies

Should be a better more robust fix for
- https://github.com/easybuilders/easybuild-framework/issues/5110

NOTE

- [X] For the CI we would want to also install `graphviz` on the machines and add `graphviz` to the requirements instead of `pygraph` to test it


boegel (approved):
lgtm

PR #5130, opened 2026-02-20
add (experimental) support for installing with bwrap

smoors:
cfr. https://github.com/easybuilders/easybuild-framework/issues/4110

UPDATE 2026-03-13

done:
- add `--bwrap-installpath` option to specify the temporary install path.
- add `--bwrap` option: if set, EB will gather all info needed to run with bwrap.
- add `rerun_with_bwrap()` function in `main.py` which reruns EB with bwrap if `--bwrap` is set (and disables `--bwrap` for a real installation).
- require `--experimental` flag
- add end-to-end CI test
- also run the interactive debug shell with bwrap
- make sure it also works with easystack files and specifying multiple easyconfigs with 1 EB cmd
- write json file with bwrap install info into bwrap installpath
- also support bwrap for `--job`

to do in follow-up PR:
- add deployment step: moving the installation to the final destination

requires the following easyblock PRs to work for all software:
- https://github.com/easybuilders/easybuild-easyblocks/pull/3935
- https://github.com/easybuilders/easybuild-easyblocks/pull/3934
- https://github.com/easybuilders/easybuild-easyblocks/pull/3933
- https://github.com/easybuilders/easybuild-easyblocks/pull/3932
- https://github.com/easybuilders/easybuild-easyblocks/pull/3931
- https://github.com/easybuilders/easybuild-easyblocks/pull/3930
- https://github.com/easybuilders/easybuild-easyblocks/pull/3929
- https://github.com/easybuilders/easybuild-easyblocks/pull/3928
- https://github.com/easybuilders/easybuild-easyblocks/pull/3927
- https://github.com/easybuilders/easybuild-easyblocks/pull/3926


boegel:
@smoors I only took a very quick look at this, but do we really need `eb.py`, or can we just update `easybuild/main.py` instead?

smoors:
> @smoors I only took a very quick look at this, but do we really need `eb.py`, or can we just update `easybuild/main.py` instead?

oh cool, i didn't think this was possible but you're right!

EDIT: done in https://github.com/easybuilders/easybuild-framework/pull/5130/commits/5595842703bf6663053dcda97444ffbdfeeb2b48

boegel:
@smoors We should cover the use of `--bwrap` in a CI workflow...

Maybe an extra step can be added at the end of `.github/workflows/end2end.yml` ?

smoors:
> @smoors We should cover the use of `--bwrap` in a CI workflow...
> 
> Maybe an extra step can be added at the end of `.github/workflows/end2end.yml` ?

done in https://github.com/easybuilders/easybuild-framework/pull/5130/commits/ab03bb83768d4d69abaca00191a77daee278e300

successful test in https://github.com/easybuilders/easybuild-framework/actions/runs/22448778136/job/65010878984?pr=5130

it runs in a separate workflow, directly in the VM instead of a container because i couldn't get bwrap to work in a container in a github workflow.

bartoldeman:
This looks like a very good start, however we have a slightly different use case where ideally we'd be able to have the installation directory on a read-only file system (e.g. cvmfs), i.e. which EESSI solves with apptainer.

but for that the
`mkdir(spath, parents=True)`
will not work, unless the directory already exists (e.g. with `eb --rebuild`). What can work is overlayfs but it'll have to do something like this:
```
while not os.path.exists(spath):
       spath = os.path.dirname(spath)
```
and then
`f"bwrap --dev-bind / / --overlay-src {spath} --overlay /bwrap/rwdir /bwrap/workdir {spath} ..."`
this way the `spath/path/to/install/dir`  can be created inside bwrap by regular eb.

It has to be mounted at that exact point (first parent dir that exists) since if you go too high you'll have permission issues inside bwrap.

This could be added in a compatible fashion, i.e. if the `mkdir` fails, use overlayfs. I'll do a few tests with this technique and will let you know.

smoors:
@bartoldeman would be nice if you can get it working with overlayfs. our bwrap version doesn't support it yet, but we currently don't have cvmfs so we don't need it (yet).

.github/workflows/end2end_bwrap.yml
```diff
@@ -0,0 +1,63 @@
...
+              echo ">>> $cmd"
+              bash -lc "source /tmp/eb_env; $cmd"
+          done
+
+      - name: End-to-end test of installing UnZip with EasyBuild in bwrap namespace
+        shell: bash
+        run: |
+          bash -l <<EOF
+            source /tmp/eb_env
+            eb UnZip-6.0.eb --trace --robot --experimental --bwrap --bwrap-installpath /tmp/bwrap-installpath --filter-deps=binutils
```
boegel, line 67: @smoors We should do a little bit of extra testing here, to make sure that the module and installation directory are found in the right place?

easybuild/tools/parallelbuild.py
```diff
@@ -154,17 +155,27 @@ def submit_jobs(ordered_ecs, cmd_line_opts, testing=False, prepare_first=True, t
...
+    pre_cmd = [
+        "unset TMPDIR",
+        f"cd {curdir}",
+    ]
+
     eb_cmd = build_option('job_eb_cmd')
+    full_eb_cmd = f"{eb_cmd} %(spec)s {opts_str} %(add_opts)s --testoutput=%(output_dir)s"
+
+    if bwrap_cmd:
+        full_eb_cmd = f'{" ".join(bwrap_cmd)} {full_eb_cmd}'
```
boegel: this is a lot easier to read for humans imho:
```suggestion
        full_eb_cmd = ' '.join(bwrap_cmd) + ' ' + full_eb_cmd
```

easybuild/tools/run.py
```diff
@@ -272,16 +272,20 @@ def create_cmd_scripts(cmd_str, work_dir, env, tmpdir, out_file, err_file):
 
     # Make script that sets up bash shell with specified environment and working directory
     cmd_fp = os.path.join(tmpdir, 'cmd.sh')
+    # using -i to force interactive shell, so env.sh is also sourced when -c is used to run commands
+    launch_cmd = 'bash --rcfile $EB_SCRIPT_DIR/env.sh -i "$@"'
+    bwrap_cmd = os.getenv('EB_BWRAP_CMD')
+    if bwrap_cmd:
+        launch_cmd = ' '.join([bwrap_cmd, launch_cmd])
```
boegel:
```suggestion
        launch_cmd = bwrap_cmd + ' ' + launch_cmd
```

boegel (changes requested):
@smoors please check https://github.com/smoors/easybuild-framework/pull/3 

If you agree with the (imho minimal) changes being made there, then please merge it, and then this PR is ready to merge imho...

easybuild/tools/bwrap.py
```diff
@@ -0,0 +1,101 @@
...
+import json
+import os
+
+from easybuild.base import fancylogger
+from easybuild.tools.build_log import print_msg
+from easybuild.tools.config import install_path
+from easybuild.tools.filetools import mkdir, write_file
+from easybuild.tools.utilities import trace_msg
+
+BWRAP_INFO = {
```
boegel: @smoors I think we should avoid this constant-which-isn't-actually-a-constant-at-all, see proposed changes in https://github.com/smoors/easybuild-framework/pull/4

boegel (changes requested):

boegel (approved):
lgtm

PR #5136, opened 2026-02-24
PoC: use isolated contextes for environment variables

Crivella:
This is related to the discussion for EB 6.0 to have thread-save handling of environment variables to allow parallel builds that modify the environment to still work.

The current changes do the following:
- hijack imports of `os` and store all env variables changes into a context (for now a `dict`) that is applied when the environment is requested.
  The hijacking is done to:
  - minimize the actual amount of code changes/refactoring for this to be implemented
  - Allow this to work also with code that touches the environment n EasyBlocks using `os` instead of `easybuild.tools.environment`
- The context is by default applied when invoking `run_shell_cmd`

## TODO

- [ ] For now only the context mechanism is implemented but the `current_context` is not thread-safe and would need to be implemented in a different way
- [x] Calling python commands that still rely on a particular environment variable to be set still need to modify the real `os.environ` eg, calls to `tempfile` with `TMPDIR` being set (if the mechanism used by the package to get the env value does not go through `os` or one of the overridden methods)
- [x] Also make behavior w.r.t. `os.chdir` and similar mocked into a context

Partially related to:
- #4681

### DIFF
--- a/easybuild/_deprecated.py
+++ b/easybuild/_deprecated.py
@@ -45,6 +45,7 @@ import easybuild.tools.asyncprocess as asyncprocess
 from easybuild.base import fancylogger
 from easybuild.tools.build_log import EasyBuildError, dry_run_msg, print_msg, time_str_since
 from easybuild.tools.config import ERROR, IGNORE, WARN, build_option
+from easybuild.tools.environment import apply_context
 from easybuild.tools.hooks import RUN_SHELL_CMD, load_hooks, run_hook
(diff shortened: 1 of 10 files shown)


Crivella:
NOTE: `tempfile` was escaping the hooking of `os` as it is imported before the install and imports `os` with a separate name (so it was not being reloaded by the hook)

https://github.com/python/cpython/blob/f8ce51a52209aa04c781d20fe0778247df5a465b/Lib/tempfile.py#L42

Installing the hook before loading the file should fix this

PR #5137, opened 2026-02-24
Replace `assertTrue`/`assertFalse` with `assertRegex` in easyblock test

Flamefire:
Cleaner and avoids mentioning the pattern and text


boegel (approved):
lgtm

PR #5138, opened 2026-02-24
Add support for non-path variables in `module_load_environment`

Flamefire:
Useful for simple values.
Using existing logic. If the value is a list elements are joined using
the delimiter.

Includes:
- [x] https://github.com/easybuilders/easybuild-framework/pull/5137

### DIFF
--- a/test/framework/easyblock.py
+++ b/test/framework/easyblock.py
@@ -654,7 +654,6 @@ class EasyBlockTest(EnhancedTestCase):
                 r'prepend_path\("PATH", pathJoin\(root, "%s"\)\)\n' % sub_path_path, re.M))
 
         # Module load environement may contain non-path variables
-        # TODO: remove whenever this is properly supported, in the meantime check warning
         eb.module_load_environment.NONPATH = {'contents': 'non_path', 'var_type': "STRING"}
         eb.module_load_environment.PATH = ['bin']
         with self.mocked_stdout_stderr():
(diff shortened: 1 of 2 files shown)


test/framework/easyblock.py
```diff
@@ -659,16 +658,16 @@ def test_make_module_req(self):
...
-            self.assertTrue(re.match(r'^\nprepend_path\("PATH", pathJoin\(root, "bin"\)\)\n$', txt, re.M))
-            self.assertFalse(re.match(r'^\nprepend_path\("NONPATH", pathJoin\(root, "non_path"\)\)\n$', txt, re.M))
+            self.assertRegex(txt, re.compile(r'^prepend_path\("PATH", pathJoin\(root, "bin"\)\)$', re.M))
+            self.assertRegex(txt, re.compile(r'^setenv\("NONPATH", "non_path"\)$', re.M))
         else:
             self.fail("Unknown module syntax: %s" % get_module_syntax())
 
         logtxt = read_file(eb.logfile)
-        self.assertRegex(logtxt, r"WARNING Non-path variables found in module load env.*NONPATH")
+        self.assertNotRegex(logtxt, r"WARNING Non-path variables found in module load env.*NONPATH")
```
boegel: @Flamefire This should be removed imho, as the code that was emiting this warning has been removed in `EasyBlock.make_module_req` ?

boegel:
@lexming Can you take a look at this?

test/framework/easyblock.py
Flamefire: Agreed
```suggestion
```

PR #5139, opened 2026-02-25
switch native flag from `-march=native` to `-mcpu=native` for LLVM compilers on Arm 64-bit (aarch64)

Thyre:
It was observed that `-march=native` does not actually provide all the CPU features on e.g. GH200, missing bf16 when building OpenBLAS 0.3.31 with LLVM 21.1.8. This causes build failures like:

```
../kernel/arm64/sbgemv_n_neon.c:146:14: error: '__builtin_neon_vld1q_bf16' needs target feature bf16,neon
  146 |         a3 = vld1q_bf16(a_ptr3);
      |              ^
```

Switching to `-mtune=native -mcpu=native` solves this issue. Therefore, apply this option generally.

### DIFF
--- a/easybuild/toolchains/compiler/llvm_compilers.py
+++ b/easybuild/toolchains/compiler/llvm_compilers.py
@@ -143,8 +143,9 @@ class LLVMCompilers(Compiler):
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
         **(Compiler.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or {}),
-        # -march=native may not include all CPU features on aarch64, e.g. missing bf16 on Grace for LLVM 21.1.8
-        (systemtools.AARCH64, systemtools.ARM): '-mcpu=native -mtune=native',
+        # -mcpu=native is recommended way to specify feature set for aarch64, see:
+        # https://github.com/easybuilders/easybuild-framework/pull/5139#issuecomment-3961654073
+        (systemtools.AARCH64, systemtools.ARM): '-mcpu=native',
         (systemtools.POWER, systemtools.POWER): '-mcpu=native',  # no support for march=native on POWER
         (systemtools.POWER, systemtools.POWER_LE): '-mcpu=native',  # no support for march=native on POWER
         (systemtools.X86_64, systemtools.AMD): '-march=native',
--- a/easybuild/toolchains/compiler/llvm_compilers.py
+++ b/easybuild/toolchains/compiler/llvm_compilers.py
@@ -143,7 +143,8 @@ class LLVMCompilers(Compiler):
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
         **(Compiler.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or {}),
-        (systemtools.AARCH64, systemtools.ARM): '-march=native',
+        # -march=native may not include all CPU features on aarch64, e.g. missing bf16 on Grace for LLVM 21.1.8
+        (systemtools.AARCH64, systemtools.ARM): '-mcpu=native -mtune=native',
         (systemtools.POWER, systemtools.POWER): '-mcpu=native',  # no support for march=native on POWER
         (systemtools.POWER, systemtools.POWER_LE): '-mcpu=native',  # no support for march=native on POWER
         (systemtools.X86_64, systemtools.AMD): '-march=native',
--- a/easybuild/toolchains/compiler/llvm_compilers.py
+++ b/easybuild/toolchains/compiler/llvm_compilers.py
@@ -144,7 +144,8 @@ class LLVMCompilers(Compiler):
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
         **(Compiler.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or {}),
         # -mcpu=native is recommended way to specify feature set for aarch64, see:
-        # https://github.com/easybuilders/easybuild-framework/pull/5139#issuecomment-3961654073
+        # https://github.com/easybuilders/easybuild-framework/pull/5139#issuecomment-3961654073 and
+        # https://developer.arm.com/community/arm-community-blogs/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu
         (systemtools.AARCH64, systemtools.ARM): '-mcpu=native',
         (systemtools.POWER, systemtools.POWER): '-mcpu=native',  # no support for march=native on POWER
         (systemtools.POWER, systemtools.POWER_LE): '-mcpu=native',  # no support for march=native on POWER


Thyre:
Failure was encountered in https://github.com/easybuilders/easybuild-easyconfigs/pull/25420#issuecomment-3958982156.

Trying a rebuild with this PR.

Thyre:
Test report for PR which caused the issue: https://github.com/easybuilders/easybuild-easyconfigs/pull/25420#issuecomment-3959994787

Will run a few more tests to make sure we don't hit any regression because of the change.

Thyre:
OpenBLAS v0.3.30: https://gist.github.com/Thyre/3c3d148fd9ac873ed4c76570f88e7e05
NumPy v2.3.2 still produces same test failure: https://gist.github.com/Thyre/f03297e66eca5b9b932ee31e4dbb1f3d

easybuild/toolchains/compiler/llvm_compilers.py
```diff
@@ -143,7 +143,8 @@ class LLVMCompilers(Compiler):
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
         **(Compiler.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or {}),
-        (systemtools.AARCH64, systemtools.ARM): '-march=native',
+        # -march=native may not include all CPU features on aarch64, e.g. missing bf16 on Grace for LLVM 21.1.8
+        (systemtools.AARCH64, systemtools.ARM): '-mcpu=native -mtune=native',
```
Crivella: Not 100% sure since the post is old(https://discourse.llvm.org/t/when-to-use-mcpu-versus-march/47953/3)  but as far as i understand `-mcpu` translates into `march + mtune`, so just that might be fine?

Thyre: I'll do a check tomorrow. If `-mcpu=native` would be enough, we should switch to that.

boegel:
Can we ask OpenBLAS maintainers for input on this?

Thyre:
> Can we ask OpenBLAS maintainers for input on this?

Sure. I'd guess that the changes described in https://github.com/OpenMathLib/OpenBLAS/pull/5643#issuecomment-3941542239 may have something to do with the new error, but generally, `-march=native` is simply missing `bf16` in the passed feature flags, which shouldn't be the case.

easybuild/toolchains/compiler/llvm_compilers.py
Thyre: `-mcpu` should be sufficient, based on the output I got. `-mtune` yields the least options of all three, `-mcpu` the most.

green-br:
Just came across this issue - https://developer.arm.com/community/arm-community-blogs/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu is usually a good reference when it comes with march/mtune/mcpu.  

I think (and happy to be corrected) for that instruction vld1q_bf16 is optional in armv9 - so march would have to disable it to make it executable on all armv9 microarchitectures. I usually refer to https://developer.arm.com/documentation/109697/2025_12/Feature-descriptions/The-Armv9-2-architecture-extension?lang=en to see what is optional vs mandatory but never easy to understand.  I previously hit similar issues and asked on Arm HPC User Group (AHUG) support channels.

Thyre:
> Just came across this issue - https://developer.arm.com/community/arm-community-blogs/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu is usually a good reference when it comes with march/mtune/mcpu.
> 
> I think (and happy to be corrected) for that instruction vld1q_bf16 is optional in armv9 - so march would have to disable it to make it executable on all armv9 microarchitectures. I usually refer to https://developer.arm.com/documentation/109697/2025_12/Feature-descriptions/The-Armv9-2-architecture-extension?lang=en to see what is optional vs mandatory but never easy to understand. I previously hit similar issues and asked on Arm HPC User Group (AHUG) support channels.

This is a very good reference, thanks a lot!
It looks like `-mcpu=native` yields the largest feature set. Interestingly, `-mcpu=native` does not include some features present with `-march=native`:

```diff
[reuter1@jrc0900 ~]$ diff -Naur mcpu.txt march.txt
--- mcpu.txt    2026-02-25 21:03:44.998971916 +0100
+++ march.txt   2026-02-25 21:03:50.711072338 +0100
@@ -1,39 +1,26 @@
 --
-"+aes"
-"+bf16"
+"+bti"
 "+ccidx"
 "+complxnum"
 "+crc"
+"+dit"
 "+dotprod"
-"+ete"
+"+flagm"
 "-fmv"
-"+fp16fml"
-"+fpac"
 "+fp-armv8"
 "+fullfp16"
-"+i8mm"
 "+jsconv"
 "+lse"
-"+mte"
 "+neon"
 "+outline-atomics"
 "+pauth"
-"+perfmon"
-"-rand"
+"+predres"
 "+ras"
 "+rcpc"
 "+rdm"
-"+sha2"
-"+sha3"
-"+sm4"
-"+spe"
+"+sb"
 "+ssbs"
 "+sve"
 "+sve2"
-"+sve-aes"
-"+sve-bitperm"
-"+sve-sha3"
-"+sve-sm4"
 "-target-feature"
-"+trbe"
 "+v9a"
```

Some of the features are reported as present when checking `/proc/cpuinfo`, so no idea why LLVM 22.1.0 does't use them with `-mcpu=native`. I'd guess they may already be included in ` "+v9a"`?

For now, I'd just go with the TL;DR:

> TL;DR: Whenever possible use only -mcpu=native. Avoid -march and -mtune.

easybuild/toolchains/compiler/llvm_compilers.py
```diff
@@ -143,7 +143,9 @@ class LLVMCompilers(Compiler):
     # used when 'optarch' toolchain option is enabled (and --optarch is not specified)
     COMPILER_OPTIMAL_ARCHITECTURE_OPTION = {
         **(Compiler.COMPILER_OPTIMAL_ARCHITECTURE_OPTION or {}),
-        (systemtools.AARCH64, systemtools.ARM): '-march=native',
+        # -mcpu=native is recommended way to specify feature set for aarch64, see:
+        # https://github.com/easybuilders/easybuild-framework/pull/5139#issuecomment-3961654073
```
boegel:
```suggestion
        # https://github.com/easybuilders/easybuild-framework/pull/5139#issuecomment-3961654073 and
        # https://developer.arm.com/community/arm-community-blogs/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu

boegel (changes requested):

boegel (approved):
This problem doesn't occur with the existing `OpenBLAS-0.3.30-llvm-compilers-20.1.8.eb`, but does occur with a copy that's updated to install `OpenBLAS` v0.3.31 instead.

I've confirmed that using `-mcpu=native` as proposed here fixes the problem, and indeed https://developer.arm.com/community/arm-community-blogs/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu makes it pretty clear that this is what we should be doing (and we're already doing so with GCC already, see also https://github.com/easybuilders/easybuild-framework/pull/1974/changes).

So, lgtm!

PR #5141, opened 2026-02-25
fix handling of NVHPC toolchain with hierarchical module naming scheme

### LINKED ISSUE #5115: NVHPC mpi directory missing in `MODULEPATH`
Using the new NVHPC toolchain with MPI and math libraries, the `mpi` directory is missing in `MODULEPATH`. We are using hierarchical modules.

```
$ ml nvhpc
$ echo $MODULEPATH
/apps/modulefiles/standard/compiler/nvidia-compilers/26.1:/apps/modulefiles/standard/compiler/gcc/14.2.0:/apps/modulefiles/standard/core:/apps/modulefiles/standard/toolchains:/apps/modulefiles/vendor
```

The expected first entry `/apps/modulefiles/standard/mpi/nvidia-compilers/26.1/nvhpc/26.1` is missing, which disables loading modules built with NVHPC. Once manually exported, I can load quantumespresso, for instance.

lexming:
fixes #5115

### DIFF
--- a/easybuild/tools/module_naming_scheme/hierarchical_mns.py
+++ b/easybuild/tools/module_naming_scheme/hierarchical_mns.py
@@ -35,6 +35,7 @@ import os
 import re
 
 from easybuild.toolchains.gcccore import GCCcore
+from easybuild.tools import LooseVersion
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.module_naming_scheme.mns import ModuleNamingScheme
 from easybuild.tools.module_naming_scheme.toolchain import det_toolchain_compilers, det_toolchain_mpi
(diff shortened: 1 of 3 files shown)


easybuild/tools/module_naming_scheme/hierarchical_mns.py
```diff
@@ -240,15 +241,18 @@ def det_modpath_extensions(self, ec):
...
+            fullver = self.det_full_version(ec)
+            paths.append(os.path.join(MPI, tc_comp_name, tc_comp_ver, ec['name'], fullver))
 
-        # special case for Cray toolchains
         elif modclass == MODULECLASS_TOOLCHAIN and tc_comp_info is None:
+            # special case for Cray toolchains
             if any(ec.name.startswith(x) for x in CRAY_TOOLCHAIN_NAME_PREFIXES):
                 paths.append(os.path.join(TOOLCHAIN, ec.name, ec.version))
+            # special case for NVHPC toolchains that lack standalone MPI component
+            elif ec.name == "NVHPC" and LooseVersion(ec.version) >= LooseVersion('25.0'):
```
bartoldeman: I wonder if we can use this instead
```
    elif ec.name == "NVHPC" and "GCCcore" not in {dep["name"] for dep in ec["dependencies"]}:
```
That follows https://github.com/easybuilders/easybuild-framework/pull/5103/changes and allows for custom easyconfigs with NVHPC >= 25.0 that are old-style (we have one of those, for historical reasons).

bartoldeman (changes requested):

bartoldeman (dismissed):
Also, would be nice to have a testcase.

easybuild/tools/module_naming_scheme/hierarchical_mns.py
lexming: good point, we can indeed replace the version check with a dependency check here.
Fixed in https://github.com/easybuilders/easybuild-framework/pull/5141/commits/18f4579b9e5925fa5fa005bdae533c6ec45fee82

lexming:
@bartoldeman I won't be adding any tests for this. We might still change this behaviour as there was some discussion to add HPCX with its own easyconfig. On top of that, such a test would need toy easyconfigs for old and new NVHPC, which is quite some work.

lexming:
@bartoldeman as discussed, I added the requested unit tests to `test_hierarchical_mns`

boegel (approved):
lgtm

PR #5142, opened 2026-03-02
Be a bit more forgiving with `copy_file`

ocaisa:
I noticed that (in https://github.com/EESSI/software-layer/pull/1405), that `jupyter-server-2.16.0-GCCcore-14.2.0.eb` builds fine with `EasyBuild/5.2.0` but errors (in EESSI) due to a copy failure for one of the easyblocks used.

I tracked this down to https://github.com/easybuilders/easybuild-easyblocks/pull/3993/changes#diff-8c64eeef13a658d3b3982cea75b316a70b272ece82de8f5d1a634dd2252071f9 which seems to trigger attempting to copy the cargo easyblock twice to the `reprod` folder. Since the permissions in EESSI are read-only, the copy fails to overwrite the previously copied file. This change makes things more forgiving (if the contents are already the same, just update the `xattrs`).

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -49,7 +49,6 @@ import inspect
 import itertools
 import os
 import pathlib
-import platform
 import re
 import shutil
 import signal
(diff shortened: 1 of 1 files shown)


bartoldeman (approved):
lgtm

Issue #5124, opened 2026-02-19
Transitive dependency `include`s missing during build step


ocaisa:
In https://github.com/easybuilders/easybuild-framework/pull/5023 we ensured that the include headers are added to `C_INCLUDE_PATH` and friends for direct dependencies. This does not cover the case though where transitive dependencies are needed. I saw this for a build of VTE:
```
In file included from /cvmfs/software.eessi.io/versions/2025.06/software/linux/aarch64/nvidia/grace/software/GTK4/4.20.2-GCCcore-14.3.0/include/gtk-4.0/gdk/wayland/gdkwayland.h:31,
                 from ../vte-0.82.2/src/app/app.cc:48:
/cvmfs/software.eessi.io/versions/2025.06/software/linux/aarch64/nvidia/grace/software/GTK4/4.20.2-GCCcore-14.3.0/include/gtk-4.0/gdk/wayland/gdkwaylanddevice.h:26:10: fatal error: wa
yland-client.h: No such file or directory
   26 | #include <wayland-client.h>
```
I entered the failing step and saw
```
eb-shell> echo $C_INCLUDE_PATH
/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/software/GTK4/4.20.2-GCCcore-14.3.0/include:
/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/software/GTK3/3.24.43-GCCcore-14.3.0/include:
/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/software/GnuTLS/3.8.10-GCCcore-14.3.0/include:
/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/software/fmt/11.2.0-GCCcore-14.3.0/include:
/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/software/lz4/1.10.0-GCCcore-14.3.0/include:
/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/software/Pango/1.57.0-GCCcore-14.3.0/include:
/cvmfs/software.eessi.io/versions/2025.06/software/linux/x86_64/amd/zen2/software/cairo/1.18.4-GCCcore-14.3.0/include:
[...]

PR #5143, opened 2026-03-09
Ensure transitive dependencies for headers and libraries are always available

ocaisa:
This addresses #5124 but is quite a major change even if just a bugfix. I believe that this was probably the original intention when introducing the related options.

Requires additional tests (and probably debating) so leaving as a draft PR for now.

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -1201,12 +1201,12 @@ class Toolchain:
                 if paths_value:
                     self.log.debug(f"Determining fallback directories to retain from ${env_var}: {paths_value}")
                     paths = paths_value.split(':')
-                    # Add the paths in reverse order (as last entry gets priority in the relevant class)
-                    paths.reverse()
                     paths_list = unique_ordered_extend(paths_list, paths)
                     # Now that we have the value, we make sure that EasyBuild will reset the environment variable later
-                    # (and only use the configured option)
-                    self.variables.append(env_var, '')
+                    # and only use the configured option
+                    # (our RPATH wrappers rely on LIBRAARY_PATH so we need an exception for that)
+                    if env_var != 'LIBRARY_PATH':
+                        self.variables.append(env_var, '')
             for var in SEARCH_PATH[search_path_var][self.search_path[search_path_var]]:
                 self.variables.append_subdirs(var, '', subdirs=paths_list)
 
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -1191,6 +1191,22 @@ class Toolchain:
             else:
                 dep_roots.extend(self.get_software_root(dep['name']))
 
+        # Exactly how the include headers or link libraries are going to be referenced is configuration dependent,
+        # gather all the options and add them as fallback options for the chosen method
+        # (allows for transitive dependencies)
+        paths_options = {'cpp_headers': [], 'linker': []}
+        for search_path_var, paths_list in paths_options.items:
+            for env_var in [y for x in SEARCH_PATH[search_path_var].values() for y in x if y.endswith('PATH')]:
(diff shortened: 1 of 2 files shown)


easybuild/tools/toolchain/toolchain.py
```diff
@@ -1191,6 +1191,25 @@ def _add_dependency_variables(self, names=None, cpp=None, ld=None):
...
+                paths_value = os.getenv(env_var)
+                if paths_value:
+                    self.log.debug(f"Determining fallback directories to retain from ${env_var}: {paths_value}")
+                    paths = paths_value.split(':')
+                    paths_list = unique_ordered_extend(paths_list, paths)
+                    # Now that we have the value, we make sure that EasyBuild will reset the environment variable later
+                    # and only use the configured option
+                    # (our RPATH wrappers rely on LIBRARY_PATH so we need an exception for that)
+                    if env_var != 'LIBRARY_PATH':
+                        self.variables.append(env_var, '')
```
ocaisa: This may seem quite aggressive, but it does seem to work in the cases I tried.

ocaisa:
I think this is good enough for a review, it passes existing CI and also fixes #5143

easybuild/tools/toolchain/toolchain.py
```diff
@@ -1191,6 +1191,25 @@ def _add_dependency_variables(self, names=None, cpp=None, ld=None):
...
+                    self.log.debug(f"Determining fallback directories to retain from ${env_var}: {paths_value}")
+                    paths = paths_value.split(':')
+                    paths_list = unique_ordered_extend(paths_list, paths)
+                    # Now that we have the value, we make sure that EasyBuild will reset the environment variable later
+                    # and only use the configured option
+                    # (our RPATH wrappers rely on LIBRARY_PATH so we need an exception for that)
+                    if env_var != 'LIBRARY_PATH':
+                        self.variables.setdefault(env_var, append_empty=True)
+            for var in SEARCH_PATH[search_path_var][self.search_path[search_path_var]]:
+                self.variables.append_subdirs(var, '', subdirs=paths_list)
```
ocaisa, line 1211: This looks odd, but the methods available here are limited. My subdirs are full paths relative to `''`

akesandgren:
Can't we just clear the relevant env variables at start of eb and then just load modules and "all will be fine"?

ocaisa:
Sure, but that is not what we say we support with the option `--search-path-cpp-headers`. @lexming can correct me, but I thought the intention there was to _exclusively_ use the indicated approach

lexming:
When I worked on `--search-path-cpp-headers` I didn't change the behaviour of EB at build time in that regard. That option only introduces the possibility to set how those paths are injected in the build environment.

The default behaviour of EB (which has not changed in ages AFAIK):
* manually inject search paths in direct dependencies with compiler/linker flags, so `include` folders are always passed with `-I` flags independently of what modules set
* the build environment is not reset, any environment variables like `CPATH` and `LIBRARY_PATH` will be present in the build environment

Now, I have not yet looked in detail the code in this PR, but it is indeed probably the case that setting `--search-path-cpp-headers` to an environment variable (instead of flags) is resetting that variable instead of prepending to it.

boegel:
We should be careful with changing the logic here imho, since I expect it's likely that we would overlook some situations here and introduce regressions.

This effort was triggered by us hitting this bug in EESSI, because we were configuring EasyBuild with `$EASYBUILD_MODULE_SEARCH_PATH_HEADERS` and `$EASYBUILD_SEARCH_PATH_CPP_HEADERS` set to `include_paths`, so `$C_INCLUDE_PATH` & co are used both in module file and by EasyBuild itself when setting up the build environment.

We revised that:
- https://github.com/EESSI/software-layer-scripts/pull/175
- https://github.com/EESSI/software-layer-scripts/pull/176

That effectively defuses the situation in EESSI, which makes me think we should take our time here...

PR #5144, opened 2026-03-10
Sort values for cuda_compute_capabilities templates 

Flamefire:
There have been instances where the order matters so sort the values.
Even if it doesn't solve some issues it will at least be consistent.

E.g. `--cuda-compute-capabilities=7.5,8.0` should be the same as `--cuda-compute-capabilities=8.0,7.5`

This is also important to avoid issues with Blackwell that introduced CUDA CC 10.x which when sorting naively will be  incorrectly the "lowest" CC.

See companion easyblock PR:
- https://github.com/easybuilders/easybuild-easyblocks/pull/4092

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3041,14 +3041,13 @@ class EasyBlock:
 
         # Set CUDA compute capabilities from default value in nvidia-compiler/NVHPC toolchains
         if get_software_root('nvidia-compilers'):
-            cuda_cc_cfg = self.cfg.get('cuda_compute_capabilities')
-            cuda_cc_opt = build_option('cuda_compute_capabilities')
+            cuda_cc = self.cfg.get_cuda_cc_template_value('cuda_compute_capabilities', required=False)
             cuda_cc_nvhpc = os.getenv('EBNVHPCCUDACC', None)
-            if not cuda_cc_cfg and not cuda_cc_opt and cuda_cc_nvhpc:
+            if not cuda_cc and cuda_cc_nvhpc:
                 self.cfg['cuda_compute_capabilities'] = cuda_cc_nvhpc.split(',')
                 self.log.info(
                     "Updated empty 'cuda_compute_capabilities' option with default CUDA compute capability "
-                    f"defined in nvidia-compilers: {self.cfg['cuda_compute_capabilities']}"
+                    f"defined in nvidia-compilers: {cuda_cc_nvhpc}"
                 )
 
         # guess directory to start configure/build/install process in, and move there
@@ -3460,7 +3459,7 @@ class EasyBlock:
         self.log.info("Checking binaries/libraries for CUDA device code...")
 
         fail_msgs = []
-        cfg_ccs = build_option('cuda_compute_capabilities') or self.cfg.get('cuda_compute_capabilities', None)
+        cfg_ccs = self.cfg.get_cuda_cc_template_value('cuda_cc_space_sep', required=False).split()
         ignore_failures = not build_option('cuda_sanity_check_error_on_failed_checks')
         strict_cc_check = build_option('cuda_sanity_check_strict')
         accept_ptx_as_devcode = build_option('cuda_sanity_check_accept_ptx_as_devcode')
@@ -3471,7 +3470,7 @@ class EasyBlock:
         ignore_file_list = [os.path.join(self.installdir, d) for d in self.cfg['cuda_sanity_ignore_files']]
 
         # If there are no CUDA compute capabilities defined, return
-        if cfg_ccs is None or len(cfg_ccs) == 0:
+        if not cfg_ccs:
             self.log.info("Skipping CUDA sanity check, as no CUDA compute capabilities were configured")
             return fail_msgs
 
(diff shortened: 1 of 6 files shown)


boegel:
Although this makes sense in general, strictly speaking this is a backwards-incompatible change.

Don't we have easyblock that make assumptions on the order of CUDA compute capabilities that are specified, where first listed is the preferred one to use (there are cases where using multiple ones isn't possible)?

Flamefire:
IIRC we only sort and take lowest or highest.
See the companion PR for easyblocks where I replaced most usages including the sorted assumption

boegel:
@casparvl Thoughts on this?

casparvl:
I like the _idea_ of this PR. Right now, it's up to the easyblock to do something sensible with this list of compute capabilities. E.g. the LAMMPS easyblock _does_ some sorting https://github.com/easybuilders/easybuild-easyblocks/blob/b01ea7afd303b0456f860588f9e1c6acc44509a6/easybuild/easyblocks/l/lammps.py#L334 but indeed seems to suffer from the issue mentioned by @Flamefire : it calls `sorted` but doesn't specify a `LooseVersion` key. Thus, a list like `["9.0a", "10.0f", "8.0", "12.0f"]` will be sorted to `['10.0f', '12.0f', '8.0', '9.0a']` (I think).

It is much nicer if the framework takes care of this. If this is documented properly (which _is_ done in this PR), the EasyBlock _knows_ what to expect, and doesn't need to do any sorting itself (reducing the potential for error on this).

That being said, @boegel is also right that this is, at least potentially, a breaking change. After merging this PR and https://github.com/easybuilders/easybuild-easyblocks/pull/4092 , Lammps would be build with `12.0f` instead of `9.0a` in the above case. That being said: using `9.0a` in the above could simply be considered a bug - in which case a change in behavior is NOT a bad thing.

I haven't made the analysis for all the other easyblocks, i.e. what would be the before vs after situation. 

We should also realize that this likely only affects software that can _only_ be build for a single CUDA CC, since in that case _some_ choice has to be made from that list. Sites that use a list of CUDA CCs probably build in some common prefix, and use that on node types with diverse GPU archs - but that usage is already broken today for this subset of software, since it simply _cant_ be build for multiple targets. So it's very well possible these sites have already made exceptions for this, and build _these_ particular packages in an architecture-specific prefix. So yes: in theory this could lead to a breaking change. But in practice, it would surprise me if anyone is really "hit" by this in practice.

In summary: _I_ would consider this acceptable, provided we document this very clearly in the release notes.

casparvl (changes requested):
We should probably also change the `help` description for `--cuda-compute-capabilities` and state that the order in which CCs are specified does not matter - just to make this behavior explicit.

boegel:
Just to be clear: it's game over for EasyBuild v5.3.0, this will need to wait until the release after that...

Flamefire:
@casparvl Done. Also replaced "List" by "Set" to emphasize this.

casparvl:
Don't have time for a full review now, but one additional remark:

> provided we document this very clearly in the release notes.

By this I meant that (ideally) we should also try to find out which software this potentially could affect. I.e. search for places in EasyBlocks & EasyConfigs in which the CUDA-related templates are used, check if they rely on certain order, and if so, report those in the release notes. This way it should be very clear to people "oh, I'm building software XYZ, behavior of how that is build may have changed in this release".

I'm not sure how many blocks/configs use these templates, and whether that is feasible to list completely. If it's O(10s) I think we should do it, if it's O(100s) it's probably intractable.

Flamefire:
For easyconfigs using templates this shouldn't change anything as the order shouldn't matter there: They are usually directly translated to `nvcc` flags through e.g. `CMAKE_CUDA_ARCHITECTURES`.
One issue with the order I've seen was caused by a bug in DeepSpeed which had undefined behavior which in case of sorted CUDA CCs wasn't visible while in the other it broke the software in mysterious ways

Our easyblocks are fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/4092 : They sorted values themselves where it mattered. Usually taking the highest or lowest value. So this PR fixes the 12.x issue for those. I doubt many will notice that as most will not have 12.x enabled or using that as the only CC

Or did you mean to report all software (easyblocks & easyconfigs) that use those templates at all to report something like: "Although unlikely the behavior of this software may change if you didn't pass a sorted list of CCCs already: ...."

casparvl:
> Or did you mean to report all software (easyblocks & easyconfigs) that use those templates at all to report something like: "Although unlikely the behavior of this software may change if you didn't pass a sorted list of CCCs already: ...."

Yeah, I meant we should put something like that in the release notes, ideally. Not sure how to ensure that this will happen other than to make @boegel aware of it _now_. I guess the list of software is basically the list from https://github.com/easybuilders/easybuild-easyblocks/pull/4092 .

boegel:
I'm still reluctant to go ahead with this, since strictly speaking it's a breaking change...

How crazy would it be to take a more careful approach: introduced `_sorted` variants, deprecate the current (non-sorted) ones?
At least that makes things more explicit...

Flamefire:
> How crazy would it be to take a more careful approach: introduced `_sorted` variants, deprecate the current (non-sorted) ones?

I think this would be overkill and makes the templates too long.

IMO assuming any order before this PR was unreliable at least, after this PR it is possible. So prior to that `%(cuda_compute_capabilities)s+PTX` couldn't be used as it may or may not do what was intended.

If you really want to avoid the breaking change, even if just through [Hyrum's Law](https://www.laws-of-software.com/laws/hyrum/), we can add `_sorted` variants for all templates, but I'd then keep them alongside the others. This way we can be explicit where it matters and people use what they want where it doesn't.
Not a big fan of the verbosity and possible inconsistency but that at least allows the unsorted option in places where people may want to control the order.
And I'd rather not do a large Search&Replace making the templates used in ECs longer when we don't need to

PR #5148, opened 2026-03-25
Add `get_environment_variable` utility function

### LINKED ISSUE #5059: Framework function to check an environment variable
In https://github.com/easybuilders/easybuild-easyblocks/issues/4003, we ran into a case where `os.getenv()` was being used a little carelessly (not providing a default value which lead to a `None` being injected into the build command). In general I think a framework function would be useful for this to reduce/remove the need for generic testing. There are a couple of cases that would be nice to cover:

- the environment variable may or may not exist (`get_environment_variable("VAR")`, returns value or `''`)
- the environment variable must be defined (`get_environment_variable("VAR", required=True)`, returns value or errors)
- the environment variable if exists should be non-empty (`get_environment_variable("VAR", empty=False)`)
- the environment variable should exist and be non-empty (`get_environment_variable("VAR", empty=False, required=True)`)

mvanhorn:
## Summary

Adds `get_environment_variable()` to `easybuild/tools/environment.py` for safe environment variable access with optional `required` and `empty` validation.

## Why this matters

Careless use of `os.getenv()` without defaults led to `None` being injected into build commands (#5059, referenced from easyblocks#4003). This function provides a single point of control for environment variable access with clear error messages.

## Changes

- `easybuild/tools/environment.py`: Added `get_environment_variable(name, required=False, empty=True)` with four modes:
  - Default: returns value or empty string
  - `required=True`: raises `EasyBuildError` if undefined
  - `empty=False`: treats empty strings as missing
  - Both: raises `EasyBuildError` if undefined OR empty

## Testing

- Python syntax verified
- Follows existing `read_environment()` and `EasyBuildError` patterns

Fixes #5059

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/easybuild/tools/environment.py
+++ b/easybuild/tools/environment.py
@@ -151,6 +151,32 @@ def read_environment(env_vars, strict=False):
     return result
 
 
+def get_environment_variable(name, required=False, empty=True):
+    """
+    Get the value of an environment variable with optional validation.
+
+    :param name: name of the environment variable
+    :param required: if True, raise EasyBuildError when the variable is not defined
+    :param empty: if False, treat empty string values as missing
+    :return: value of the environment variable, or empty string if not set
+    """
+    value = os.environ.get(name)
+
+    if value is None:
+        if required:
+            raise EasyBuildError("Required environment variable '%s' is not defined", name)
+        _log.debug("Environment variable '%s' is not defined, returning empty string", name)
+        return ''
+
+    if not empty and value == '':
+        if required:
+            raise EasyBuildError("Required environment variable '%s' is defined but empty", name)
+        _log.debug("Environment variable '%s' is empty, returning empty string", name)
+        return ''
+
+    return value
+
+
 def modify_env(old, new, verbose=True, log_changes=True):
     """
     Compares two os.environ dumps. Adapts final environment.
--- a/easybuild/tools/environment.py
+++ b/easybuild/tools/environment.py
@@ -151,13 +151,13 @@ def read_environment(env_vars, strict=False):
     return result
 
 
-def get_environment_variable(name, required=False, empty=True):
+def get_environment_variable(name, required=False, allow_empty=True):
     """
     Get the value of an environment variable with optional validation.
 
     :param name: name of the environment variable
     :param required: if True, raise EasyBuildError when the variable is not defined
-    :param empty: if False, treat empty string values as missing
+    :param allow_empty: if False, raise EasyBuildError when the variable is defined but empty/whitespace
     :return: value of the environment variable, or empty string if not set
     """
     value = os.environ.get(name)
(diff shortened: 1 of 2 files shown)


easybuild/tools/environment.py
```diff
@@ -151,6 +151,32 @@ def read_environment(env_vars, strict=False):
     return result
 
 
+def get_environment_variable(name, required=False, empty=True):
```
ocaisa: With `empty` I think what you really mean is `allow_empty`...and then if we don't `allow_empty` then it should be a straight error, not dependent on `required`.

easybuild/tools/environment.py
```diff
@@ -151,6 +151,32 @@ def read_environment(env_vars, strict=False):
...
+    """
+    value = os.environ.get(name)
+
+    if value is None:
+        if required:
+            raise EasyBuildError("Required environment variable '%s' is not defined", name)
+        _log.debug("Environment variable '%s' is not defined, returning empty string", name)
+        return ''
+
+    if not empty and value == '':
```
ocaisa: Am empty string is falsey so you can just use `value.strip()` (which also removes whitespace)
```suggestion
    # Check if we allow empty values (and we consider whitespace an empty value)
    if not empty and value.strip():
```

easybuild/tools/environment.py
```diff
@@ -151,6 +151,32 @@ def read_environment(env_vars, strict=False):
...
+        if required:
+            raise EasyBuildError("Required environment variable '%s' is defined but empty", name)
```
ocaisa:
```suggestion
       raise EasyBuildError("Required environment variable '%s' is defined but empty", name)
```

easybuild/tools/environment.py
```diff
@@ -151,6 +151,32 @@ def read_environment(env_vars, strict=False):
...
+    :param empty: if False, treat empty string values as missing
+    :return: value of the environment variable, or empty string if not set
+    """
+    value = os.environ.get(name)
+
+    if value is None:
+        if required:
+            raise EasyBuildError("Required environment variable '%s' is not defined", name)
+        _log.debug("Environment variable '%s' is not defined, returning empty string", name)
+        return ''
```
ocaisa: Don't use inline returns
```suggestion
        value = ''
```

easybuild/tools/environment.py
```diff
@@ -151,6 +151,32 @@ def read_environment(env_vars, strict=False):
...
+
+    if not empty and value == '':
+        if required:
+            raise EasyBuildError("Required environment variable '%s' is defined but empty", name)
+        _log.debug("Environment variable '%s' is empty, returning empty string", name)
+        return ''
```
ocaisa:
```suggestion
        value = ''
```

ocaisa (changes requested):
You will need to add tests for this as well.

ocaisa:
Ha, you were working directly off of https://github.com/easybuilders/easybuild-framework/issues/5059 but now I think about it, I don't think supporting `allow_empty=False` with `required=False` is a good idea in general (you either get `''` or `None` which is basically exactly what you said you didn't want).

boegel:
I also want to see the adopted, both in framework itself, and in one or more easyblocks, to show the usage of this...

mvanhorn:
Addressed all feedback:

- Renamed `empty` to `allow_empty`
- Simplified check with `value.strip()` for whitespace handling
- Removed inline returns
- When `allow_empty=False`, always raises EasyBuildError regardless of `required` (removed the confusing combination @ocaisa flagged)
- Added tests covering defined/undefined/empty/whitespace-only cases

Still need to add an adoption example in an easyblock per @boegel's request - will follow up on that.

PR #5150, opened 2026-03-26
Document where architecture dicts are allowed in easyconfig parameters & fix version templates for removed dependencies

Flamefire:
When an architecture specific `None` version is used the check happens
too late when we do a shallow/quick parsing of the easyconfig, e.g. in `list_software`.

Rework the duplication in the test and add documentation to the parameters to go with https://github.com/easybuilders/easybuild-docs/pull/366


boegel (approved):
lgtm

PR #5153, opened 2026-03-30
remove deprecated 2-element list/tuple for specifying sources

branfosj:
This was deprecated for EB4, but the removal was missed. As that is sufficiently long ago I consider it reasonable to remove at any time.


smoors (approved):
lgtm

PR #5155, opened 2026-03-31
be more careful when parsing output produced by rocm-smi command (+ add tests for `get_gpu_info` function)

### LINKED ISSUE #5154: `rocm-smi` crashes EasyBuild if no AMD GPU is found
On `jsc-zen3`, `rocm-smi` was implicitly installed by a system update. Since no AMD GPUs are present, every test build failed with:

```

  File "/usr/lib64/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/easybuild/easybuild-framework/easybuild/main.py", line 862, in <module>
    main_with_hooks()
  File "/easybuild/easybuild-framework/easybuild/main.py", line 847, in main_with_hooks
    exit_code: EasyBuildExit = main(args=args, prepared_cfg_data=(init_session_state, eb_go, cfg_settings))
  File "/easybuild/easybuild-framework/easybuild/main.py", line 798, in main
    is_successful = process_eb_args(orig_paths, eb_go, cfg_settings, modtool, testing, init_session_state,
  File "/easybuild/easybuild-framework/easybuild/main.py", line 632, in process_eb_args
    test_report_msg = overall_test_report(ecs_with_res, len(paths), overall_success, success_msg, init_session_state)
  File "/easybuild/easybuild-framework/easybuild/tools/testing.py", line 459, in overall_test_report
    txt = post_pr_test_report(pr_nrs, GITHUB_EASYCONFIGS_REPO, test_report, msg, init_session_state,
  File "/easybuild/easybuild-framework/easybuild/tools/testing.py", line 372, in post_pr_test_report
    gpu_info = get_gpu_info(init_session_state['environment'])
  File "/easybuild/easybuild-framework/easybuild/tools/systemtools.py", line 745, in get_gpu_info
    amd_driver = res.output.strip().split('\n')[1].split(',')[1]
IndexError: list index out of range
```

This failure is caused by `rocm-smi` not reporting a non-zero exit code when failing:

```
[reuter1@jsczen3l1 ~]$ rocm-smi
cat: /sys/module/amdgpu/initstate: No such file or directory
ERROR:root:Driver not initialized (amdgpu not found in modules)
[reuter1@jsczen3l1 ~]$ echo $?
0
```

See also in the [rocm-smi sources](https://github.com/ROCm/rocm-systems/blob/121fafc82c961b9a9f32eda63d2422bf7a0be817/projects/rocm-smi-lib/python_smi_tools/rocm_smi.py#L4732).

The reported error message fails to be parsed by the `strip()` and `split()` we're doing, hence causing the `IndexError`.
A simple fix could involve adding a `try/except` around that call, or extending the `try/except` around that whole block, so that we bail out when we detect a non-functional output. Any further parsing would produce invalid data anyway (see also [here](https://github.com/easybuilders/easybuild-framework/blob/7b7a9864f90af503c3352efdb5d654d9966946ef/easybuild/tools/systemtools.py#L761))

boegel:
fixes #5154

### DIFF
--- a/test/framework/systemtools.py
+++ b/test/framework/systemtools.py
@@ -518,13 +518,12 @@ ERROR:root:Driver not initialized (amdgpu not found in modules)"""
 
 ROCM_SMI_DRIVER_VERSION_BROKEN = ('', ROCM_SMI_DRIVER_VERSION_BROKEN_STDERR)
 
-ROCM_SMI_PRODUCT_NAME = """
-device,Card Series,Card Model,Card Vendor,Card SKU,Subsystem ID,Device Rev,Node ID,GUID,GFX Version
+ROCM_SMI_PRODUCT_NAME = """device,Card Series,Card Model,Card Vendor,Card SKU,Subsystem ID,Device Rev,Node ID,GUID,GFX Version
 card0,AMD Instinct MI250X/MI250,0x740c,Advanced Micro Devices Inc. [AMD/ATI],D65209,0x0b0c,0x01,8,13025,gfx90a
 card1,AMD Instinct MI250X/MI250,0x740c,Advanced Micro Devices Inc. [AMD/ATI],D65209,0x0b0c,0x01,9,24866,gfx90a
 card2,AMD Instinct MI250X/MI250,0x740c,Advanced Micro Devices Inc. [AMD/ATI],D65209,0x0b0c,0x01,10,36962,gfx90a
 card3,AMD Instinct MI250X/MI250,0x740c,Advanced Micro Devices Inc. [AMD/ATI],D65209,0x0b0c,0x01,11,51627,gfx90a
-"""
+"""  # noqa (ignore long lines)
 
 MACHINE_NAME = None
 
--- a/test/framework/systemtools.py
+++ b/test/framework/systemtools.py
@@ -51,9 +51,9 @@ from easybuild.tools.systemtools import check_linked_shared_libs, check_os_depen
 from easybuild.tools.systemtools import det_parallelism, det_pypkg_version, get_avail_core_count
 from easybuild.tools.systemtools import get_cuda_object_dump_raw, get_cuda_architectures, get_cpu_arch_name
 from easybuild.tools.systemtools import get_cpu_architecture, get_cpu_family, get_cpu_features, get_cpu_model
-from easybuild.tools.systemtools import get_cpu_speed, get_cpu_vendor, get_gcc_version, get_glibc_version, get_isa_riscv
-from easybuild.tools.systemtools import get_os_name, get_os_type, get_os_version, get_platform_name, get_shared_lib_ext
-from easybuild.tools.systemtools import get_system_info, get_total_memory, get_linked_libs_raw
+from easybuild.tools.systemtools import get_cpu_speed, get_cpu_vendor, get_gcc_version, get_glibc_version, get_gpu_info
+from easybuild.tools.systemtools import get_isa_riscv, get_os_name, get_os_type, get_os_version, get_platform_name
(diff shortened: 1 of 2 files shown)


test/framework/systemtools.py
```diff
@@ -493,6 +493,39 @@
...
+
+ROCM_SMI_DRIVER_VERSION = (ROCM_SMI_DRIVER_VERSION_STDOUT, ROCM_SMI_DRIVER_VERSION_STDERR)
+
+ROCM_SMI_DRIVER_VERSION_BROKEN_STDERR = """cat: /sys/module/amdgpu/initstate: No such file or directory
+ERROR:root:Driver not initialized (amdgpu not found in modules)"""
+
+ROCM_SMI_DRIVER_VERSION_BROKEN = ('', ROCM_SMI_DRIVER_VERSION_BROKEN_STDERR)
+
+ROCM_SMI_PRODUCT_NAME = """
+device,Card Series,Card Model,Card Vendor,Card SKU,Subsystem ID,Device Rev,Node ID,GUID,GFX Version
```
ocaisa: I suspect there should not be a newline here
```suggestion
ROCM_SMI_PRODUCT_NAME = """device,Card Series,Card Model,Card Vendor,Card SKU,Subsystem ID,Device Rev,Node ID,GUID,GFX Version
```

test/framework/systemtools.py
```diff
@@ -1537,6 +1588,57 @@ def test_get_linked_libs_raw(self):
...
+
+        expected = {
+            'AMD': {'AMD Instinct MI250X/MI250 (model: 0x740c, driver: 6.18.4)': 4},
+        }
+        self.assertEqual(get_gpu_info(), expected)
+
+        # force rocm-smi to generate broken output (no stdout, only stderr)
+        os.environ['MOCKED_BROKEN_ROCM_SMI'] = 'yes'
+        expected = {
+            'AMD': {'AMD Instinct MI250X/MI250 (model: 0x740c, driver: UNKNOWN)': 4},
```
ocaisa, line 1637: This makes me curious about what `amd-smi --showproductname --csv` would actually return in the case where it is in a broken state (@Thyre)

ocaisa (changes requested):

test/framework/systemtools.py
Thyre: `amd-smi` properly fails with a non-zero exit code. So we're never checking the output. 

So far, I've checked systems with the kernel module loaded and an AMD GPU present, and without the kernel module loaded and no AMD GPU. 

I can't tell what happens if the kernel module is loaded but no GPU is found.

Thyre: What we encountered before was a broken output due to other modules being loaded during amd-smi calls.

But even then, the exit code was non-zero, with rocm-smi being tried as the fallback IIRC.

boegel: I asked that too, and @Thyre mentioned he couldn't get `amd-smi` to produce broken output.

This PR was triggered by problems with `rocm-smi`, we haven't seen such a failure case with `amd-smi` yet. When that comes, we can deal with it (and the CI added here should make that easier than it was before these changes)

test/framework/systemtools.py
boegel: fixed in 71e5d9d29

ocaisa (approved):
LGTM

test/framework/systemtools.py
ocaisa: Sorry, I didn't notice that the output was being stripped before being split on newlines so this doesn't make a practical difference

PR #5156, opened 2026-04-02
improve 'filename' check in sources dict

smoors:
follow-up of https://github.com/easybuilders/easybuild-framework/pull/5152


boegel (approved):
lgtm

PR #5157, opened 2026-04-02
Replace `self.mock_*` pairs by `self.mocked_*` context managers

Flamefire:
Make sure cleanup happens

Followup to https://github.com/easybuilders/easybuild-framework/pull/5125 by adding the `force_tty` parameter to the context managers

- Diff can be viewed easiest by ignoring space changes (most of the diff is indents)
- Partially written by Github Copilot, partially RegEx and manual where necessary
- Kept in one place where the whole test is run with mocked outputs

Should be merged soon as this is prone to merge conflicts


easybuild/base/testing.py
```diff
@@ -209,30 +209,30 @@ def get_stderr(self):
         return sys.stderr.getvalue()
 
     @contextmanager
-    def mocked_stdout(self):
+    def mocked_stdout(self, force_tty=False):
```
Crivella, line 212: `force_tty` should be used when turning the mocked pipe ON and not when turning it OFF (enabling it on the normal stdout/stderr should be a separate operation if ever needed)

easybuild/base/testing.py
```diff
@@ -209,30 +209,30 @@ def get_stderr(self):
...
         self.mock_stdout(True)
         try:
             yield sys.stdout
         finally:
-            self.mock_stdout(False)
+            self.mock_stdout(False, force_tty=force_tty)
 
     @contextmanager
-    def mocked_stderr(self):
+    def mocked_stderr(self, force_tty=False):
```
Crivella, line 221: Same as above

test/framework/easyblock.py
```diff
@@ -2108,11 +2104,9 @@ def test_download_instructions(self):
...
 
         error_pattern = common_error_pattern + ", please follow the download instructions above"
-        self.mock_stderr(True)
-        self.mock_stdout(True)
-        self.assertErrorRegex(EasyBuildError, error_pattern, eb.fetch_step)
-        stderr = self.get_stderr().strip()
-        self.mock_stderr(False)
+        with self.mocked_stdout_stderr():
+            self.assertErrorRegex(EasyBuildError, error_pattern, eb.fetch_step)
+            stderr = self.get_stderr().strip()
```
Crivella, line 2109: I assume here it was originally an error where the stdout mocking was not being turned off?

test/framework/github.py
```diff
@@ -642,7 +628,7 @@ def test_github_download_repo(self):
         self.assertIn('easybuild', os.listdir(repodir))
         self.assertTrue(re.match('^[0-9a-f]{40}$', read_file(shafile)))
         self.assertExists(os.path.join(repodir, 'easybuild', 'easyblocks', '__init__.py'))
-        self.mock_stdout(False)
+        # stdout capture released by context manager above
```
Crivella: Do we need this comment?

Crivella (changes requested):

easybuild/base/testing.py
Flamefire: This is the function that turns the mocked pipe ON. Or do you mean something different?

test/framework/easyblock.py
Flamefire: Exactly. That's one of the reasons I want to get rid of the manual turn-on/off as soon as possible.

easybuild/base/testing.py
Crivella: I mean it should be

```
self.mock_stdout(True, force_tty=force_tty)
...
self.mock_stdout(False)
```

instead of the other way around

wanted to suggest it as a change but i think the tool does not like suggestions going over deleted lines(?) and was giving weird stuff

Crivella: @Flamefire Just a ping on this before we risk going out of sync, not sure i want to re-check 21 files of changes ;P

Flamefire: Seems I missed the notification, sorry. You are absolutely right, I somehow missed that. For good measure I'm now just passing it to both. Doesn't hurt and IMO looks consistent here. Or shall I remove it there?

Crivella: I would probably remove it since it does nothing when the mocking is being turned off (and if we ever decide to do something while turning off it is not necessarily true that we would want the same value for ON/OFF in every test)

Flamefire: done

test/framework/github.py
Flamefire: removed

test/framework/robot.py
```diff
@@ -795,11 +793,8 @@ def test_github_det_easyconfig_paths_from_pr(self):
             '--tmpdir=%s' % self.test_prefix,
         ]
 
-        self.mock_stderr(True)
-        self.mock_stdout(True)
-        outtxt = self.eb_main(args, logfile=dummylogfn, raise_error=True)
-        self.mock_stderr(False)
-        self.mock_stdout(False)
+        with self.mocked_stdout():
```
Crivella:
```suggestion
        with self.mocked_stdout_stderr():
```

I assume this should still mock both?

test/framework/tweak.py
```diff
@@ -398,12 +398,11 @@ def test_find_potential_version_mappings(self):
             'version': '0.2.8'
         }
 
-        self.mock_stderr(True)
-        potential_versions = find_potential_version_mappings(openblas_dep, tc_mapping)
-        errtxt = self.get_stderr()
-        warning_stub = "\nWARNING: There may be newer version(s) of dep 'OpenBLAS' available with a different " \
-                       "versionsuffix to '-LAPACK-3.4.2'"
-        self.mock_stderr(False)
+        with self.mocked_stdout_stderr():
```
Crivella:
```suggestion
        with self.mocked_stderr():
```

Crivella (changes requested):
Still went through the files quickly again, caught this 2 which seems to have different behavior w.r.t. to before the changes

Crivella (approved):
LGTM

PR #5159, opened 2026-04-07
Lift fallback for PyPI package names to `obtain_file`

Flamefire:
Followup to #4943

When the download URLs include a PYPI source and the download filename looks like a pypi source where the package name contains dashes then retry downloading with dashes replaced by underscores.

This is more generic than #4943 as it handles e.g. PythonPackage easyconfigs, not just extensions with defaulted names.

Done in 2 steps for easier checking:
1. Factor out a "wrapper" for `download_file` as a `EasyBlock` method from `obtain_file` without any significant changes
    Small change to dry-run logging as `"extensions typically have custom source URLs specified, only mention first"` didn't seem to match behavior so removed the comment and add the URL to the initial message instead of a second one
3. Call this again if we have a PYPI URL and at least 2 dashes in the download filename (last is assumed to be the version delimiter)

This is broader than #4943
- Handle not only extensions but any `obtain_file` calls, especially top-level easyconfig sources
- Trigger when PYPI URL appears anywhere in the URLs, not only in first position
- Trigger when there are at least 2 dashes not only when the extensions source name is defaulted and also for dots and combinations of dashes, dots and underscores. Specs say runs of those should be replaced by single underscore

Issue I had was that at this point we do not know anymore where the (download)filename came from and which template values (for name & version) we should use as extensions temporarily update the easyconfig templates.

Example easyconfig PR that makes use of this: https://github.com/easybuilders/easybuild-easyconfigs/pull/25682

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1051,20 +1051,27 @@ class EasyBlock:
             failedpaths.extend(failed_urls)
 
             # Usual PYPI sources have a format of '<name>-version.ext', with '<name>' has dashes replaced by underscores
-            num_dashes = download_filename.count('-')
-            if num_dashes >= 2 and any(PYPI_PKG_URL_PATTERN in url for url in source_urls):
+            if download_filename.count('-') >= 2 and any(PYPI_PKG_URL_PATTERN in url for url in source_urls):
                 # Last dash is the separator between name and version
-                alt_download_filename = download_filename.replace('-', '_', num_dashes - 1)
+                dl_name, dl_suffix = download_filename.rsplit('-', maxsplit=1)
+                # Guide https://packaging.python.org/en/latest/specifications/binary-distribution-format
+                # "any run of -_. characters [...] should be replaced with _".
+                # Account for different "compliance levels" and deduplicate (if only dashes were used all are the same)
+                alt_download_filenames = nub(f'{name}-{dl_suffix}' for name in (
+                    re.sub('[.-_]+', '_', dl_name),  # Fully compliant
+                    re.sub('-+', '_', dl_name),  # Only dashes
+                    re.sub(r'\.+', '_', dl_name),  # Only dots
+                ))
                 self.log.warning("Could not download %s (%s) from any of %s.\n"
-                                 "Retrying with alternative download filename '%s' as it looks like a PYPI source.",
-                                 filename, download_filename, ', '.join(source_urls), alt_download_filename)
-                # retry with alternative download_filename
-                downloaded, failed_urls = self.download_file(target_path, alt_download_filename, urls=source_urls,
-                                                             # Only log when extensions have explicit URLs
-                                                             log_url_in_dry_run=extension and urls)
-                if downloaded:
-                    return target_path
-                failedpaths.extend(failed_urls)
+                                 "Retrying with alternative download filenames '%s' as it looks like a PYPI source.",
+                                 filename, download_filename, ', '.join(source_urls), ', '.join(alt_download_filenames))
+                for alt_download_filename in alt_download_filenames:
+                    downloaded, failed_urls = self.download_file(target_path, alt_download_filename, urls=source_urls,
+                                                                 # Only log when extensions have explicit URLs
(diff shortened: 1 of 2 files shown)


boegel:
@smoors Are you up for taking a look at this?

No rush, it's a bit too late to get this in for EasyBuild v5.3.0...

easybuild/framework/easyblock.py
```diff
@@ -1054,64 +1039,34 @@ def obtain_file_raise_on_failure(self, filename, extension=False, urls=None, dow
...
+            downloaded, failed_urls = self.download_file(target_path, download_filename, urls=source_urls,
+                                                         # Only log when extensions have explicit URLs
+                                                         log_url_in_dry_run=extension and urls)
+            if downloaded:
+                return target_path
+            failedpaths.extend(failed_urls)
+
+            # Usual PYPI sources have a format of '<name>-version.ext', with '<name>' has dashes replaced by underscores
+            num_dashes = download_filename.count('-')
+            if num_dashes >= 2 and any(PYPI_PKG_URL_PATTERN in url for url in source_urls):
```
smoors: `source_urls` can also contain tuples, not only strings, see https://github.com/easybuilders/easybuild-framework/blob/632d4d24ca91f19077aaa52c52020791d82596a3/easybuild/framework/easyblock.py#L1071

Flamefire:
@smoors can you look into this now that EB 5.3 is released? The changed PyTorch easyconfigs need to be converted to PythonBundle and IIRC the `fetch_step` isn't run for extensions so the test_step will fail

easybuild/framework/easyblock.py
```diff
@@ -1054,64 +1039,34 @@ def obtain_file_raise_on_failure(self, filename, extension=False, urls=None, dow
...
+                # Last dash is the separator between name and version
+                alt_download_filename = download_filename.replace('-', '_', num_dashes - 1)
```
smoors: actually, according to the specs, we should also replace `'.'` with `'_'`
https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode

not sure if it should already be done in this PR?

smoors (commented):
now that the retry is moved to `obtain_file`, would be good to add a test? also for the download_file function?

easybuild/framework/easyblock.py
Flamefire: That should be fine to ignore here: `in url` would check if it is in such a tuple which likely won't match but then such a tuple won't be a PYPI URL, would it?

smoors: it's very unlikely, but not impossible.
i've seen easyconfigs unnecessarily using tuples like in `Bowtie2-2.5.4-GCC-14.2.0.eb`..

EDIT: oh wait, that's not actually a tuple, ok, nvm then

Flamefire:
I confused the 2 PRs: The more important one is https://github.com/easybuilders/easybuild-framework/pull/5162 to get the test file. But also getting this one ready. Writing the test might be the worst...

easybuild/framework/easyblock.py
Flamefire: Done

Flamefire:
> now that the retry is moved to `obtain_file`, would be good to add a test? also for the download_file function?

That was tricky to do. But I noticed an unrelated bug (with progress bars) which I fixed in https://github.com/easybuilders/easybuild-framework/pull/5162

Used the same approach now here too: `unittest.mock` to avoid actually downloading anything and being able to check calls.
This actually caught a few cases that were not right: E.g. `[.-_]` doesn't match `-` but is easy to overlook.

Rebased to latest develop too.

Flamefire:
@smoors Do you still want a unit test for `EasyBlock.download_file`?

It isn't too complicated and the main parts are covered by this and 2 more `obtain_file` tests.

Missing:

- Alternative URL formats
- Derived PyPI URL  usage
- dry-run
- handling IOError raised from `ft.download_file`

Flamefire:
After some testing I decided to extent this to wheels and rely on `self.name` too. Otherwise `url_requests-2.32.5-py3-none-any.whl` would fail as it will split at `-any` and replace all prior underscores

PR #5161, opened 2026-04-08
enhance `get_cuda_object_dump_raw` to not run `cuobjdump` on shell scripts to determine CUDA architecture

Flamefire:
Before checking a file for CUDA code a `file` command is used and the
output search for "executable", which wrongly triggers for Python, Bash, ... scripts

Exclude those to avoid unnecessary commands and errors logged.


boegel:
@Flamefire Do you have a concrete example of what this avoids/fixes, just for context?

boegel (approved):
lgtm

PR #5165, opened 2026-04-09
Cleanup use of `dict`s, e.g. use of `.keys()`

Flamefire:
Using `x in dict.keys()` is discouraged and should be `x in dict` directly which is potentially more performant and easier to read/shorter. Accesses to keys and their values should use `dict.items` to avoid the lookup overhead.


easybuild/tools/github.py
```diff
@@ -2239,7 +2239,7 @@ def det_pr_target_repo(paths):
 
         _log.info("Trying to derive target repository based on specified files...")
 
-        easyconfigs, files_to_delete, patch_files, py_files = [paths[key] for key in sorted(paths.keys())]
+        easyconfigs, files_to_delete, patch_files, py_files = [item[1] for item in sorted(paths.items())]
```
jfgrimm: could this not be simplified to the following?
```suggestion
        easyconfigs, files_to_delete, patch_files, py_files = sorted(paths.values())
```

Flamefire: We need to sort by key so that we get the right order.
It looks like a dataclass or NamedTuple would be better than a dict where you know all existing keys.

jfgrimm: ah yes, of course

Flamefire: TBH: I had to read the code here and the caller again to understand this. Shall I add a commit changing this to a named Tuple? Even though it is a "public" method I'd doubt anyone would be using this outside of us.

jfgrimm: I would be fine with that

Flamefire: That bubbled up the callchain more than I expected: The origin of this dict is `categorize_files_by_type`
IMO still very much worth it: Makes it much easier to reason about what is actually passed instead of having a `paths` parameter in multiple places and sometimes it is a list of paths, and sometimes a dict with only some keys which might be misspelled.

If you'd like to review that change in isolation then use https://github.com/easybuilders/easybuild-framework/pull/5190 and I'll rebase this (will cause an easy conflict to resolve)

easybuild/framework/easyconfig/tools.py
```diff
@@ -620,37 +621,30 @@ def dump_env_script(easyconfigs):
...
     Splits list of filepaths into a 4 separate lists: easyconfigs, files to delete, patch files and
     files with extension .py
     """
-    res = {
-        'easyconfigs': [],
-        'files_to_delete': [],
-        'patch_files': [],
-        'py_files': [],
-    }
+    res = CategorizedPaths([], [], [], [])
```
boegel: changing the return type of a public function (which may be used by others in their own scripts/hooks) is not a good idea imho, since the callers of this function need to adapt accordingly

see also the corresponding change in `main` you had to make...

please revert this part, it has little to do with not using `keys()` where it's not needed...

boegel (changes requested):

easybuild/framework/easyconfig/tools.py
Flamefire: I wanted to properly replace this strange-looking line:
`easyconfigs, files_to_delete, patch_files, py_files = [paths[key] for key in sorted(paths.keys())]`

IMO this is pretty much an internal class, but we can be more careful here. I factored this change out to
https://github.com/easybuilders/easybuild-framework/pull/5269 including a backwards compatible path.

PR #5164, opened 2026-04-09
Avoid infinite loop in EasyConfig cache

Flamefire:
(created using `eb --new-pr`)

I'm not fully sure why but #4818 leads to an infinite loop or at least heavy slowdown when parsing an easyconfig because of our way of implementing `EasyConfig::copy`:
We simply reparse the whole easyconfig from scratch and `finalize_dependencies` will parse the dependencies causing copies in the cache handling which parses their dependencies again leading to parsing the same EasyConfig literally thousands of times until I had to abort it.

The change of adding it to the cache before doing the copy of the returned easyconfig makes it better, seems to break the infinite loop at least for a test I had. Not sure we should keep that though.

Example:
<details><summary> `eb PyTorch-bundle-2.1.2-foss-2023a-CUDA-12.1.1.eb` </summary>

```
== Temporary log file in case of crash /tmp/easybuild-tmplog/easybuild-5ch8zcy7.log
== found valid index for /easybuild-easyconfigs/easybuild/easyconfigs, so using it...
== found valid index for /easybuild-easyconfigs/easybuild/easyconfigs, so using it...
COPY foss foss EasyConfig @ /easybuild-easyconfigs/easybuild/easyconfigs/f/foss/foss-2023a.eb
COPY foss foss EasyConfig @ /easybuild-easyconfigs/easybuild/easyconfigs/f/foss/foss-2023a.eb
COPY GCC GCC EasyConfig @ /easybuild-easyconfigs/easybuild/easyconfigs/g/GCC/GCC-12.3.0.eb
COPY GCC GCC EasyConfig @ /easybuild-easyconfigs/easybuild/easyconfigs/g/GCC/GCC-12.3.0.eb
COPY GCC GCC EasyConfig @ /easybuild-easyconfigs/easybuild/easyconfigs/g/GCC/GCC-12.3.0.eb
COPY GCCcore GCCcore EasyConfig @ /easybuild-easyconfigs/easybuild/easyconfigs/g/GCCcore/GCCcore-12.3.0.eb
COPY GCCcore GCCcore EasyConfig @ /easybuild-easyconfigs/easybuild/easyconfigs/g/GCCcore/GCCcore-12.3.0.eb
COPY GCCcore GCCcore EasyConfig @ /easybuild-easyconfigs/easybuild/easyconfigs/g/GCCcore/GCCcore-12.3.0.eb
[...]

PR #5166, opened 2026-04-09
avoid excessive creating of EasyConfig instances when copies are created

boegel:
This is a more minimal fix for the excessive re-parsing of easyconfigs that somehow got introduced via #4818, an alternative to:
- https://github.com/easybuilders/easybuild-framework/pull/5164

Grepping the log for `Creating.*EasyConfig instance` shows that when running `eb PyTorch-bundle-2.1.2-foss-2023a-CUDA-12.1.1.eb -Dr --disable-cleanup-tmpdir` the number of `EasyConfig` instances being created is reduced from 745 to 177 with this change.

It effectively rolls back the part of #4818 that introduced the problem, by not using the `_copy_ec_dict(s)` functions that were added.

### DIFF
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -2329,7 +2329,7 @@ def process_easyconfig(path, build_specs=None, validate=True, parse_only=False, 
     if not build_specs:
         cache_key = (path, validate, hidden, parse_only)
         if cache_key in _easyconfigs_cache:
-            return _copy_ec_dicts(_easyconfigs_cache[cache_key])
+            return [e.copy() for e in _easyconfigs_cache[cache_key]]
 
     easyconfigs = []
     for spec in blocks:
@@ -2354,7 +2354,7 @@ def process_easyconfig(path, build_specs=None, validate=True, parse_only=False, 
         easyconfigs.append(easyconfig)
 
     if cache_key is not None:
-        _easyconfigs_cache[cache_key] = _copy_ec_dicts(easyconfigs)
+        _easyconfigs_cache[cache_key] = [e.copy() for e in easyconfigs]
 
     return easyconfigs
 
--- a/easybuild/framework/easyconfig/easyconfig.py
+++ b/easybuild/framework/easyconfig/easyconfig.py
@@ -666,6 +666,7 @@ class EasyConfig:
             validate = self.validation
 
         # create a new EasyConfig instance
+        self.log.info(f"Creating new EasyConfig instance for {self.path} while copying")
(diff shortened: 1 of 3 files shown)


Flamefire:
If you want to go with this instead of #5164 I'd add comments that `ec.copy` is not `Easyconfig.copy` and not doing a deep copy. Possibly also rename to `ecinfo` or so because what we have there is not actually an easyconfig. Not sure how that dict is called

branfosj (dismissed):
We will go with this for the release of EB 5.3.0 and then investigate further in the future.

boegel:
> If you want to go with this instead of #5164 I'd add comments that `ec.copy` is not `Easyconfig.copy` and not doing a deep copy. Possibly also rename to `ecinfo` or so because what we have there is not actually an easyconfig. Not sure how that dict is called

As I said, I want to keep this really minimal, to avoid further delay in getting EasyBuild v5.3.0 out.

I agree that 38361401dd36985c2045f62e19df1dd93319c36e introduced a bug, but it's a long-standing one (since Aug 2015, EasyBuild v2.3.0), so it can hardly be considered an urgent problem.

Let's use your PR #5164 to come with a proper fix, a test that verifies that the problem isn't re-introduced later (that's a corner I'm not happy to cut here BTW, but I don't have time to tackle that right now), and all the necessary additional extra comments and co to go along with it.

For now, we need a quick & clean fix, and that's exactly what this PR is: reverting back to how things were, rather than trying to fix the problem that was introduced in #4818 in a rush...

Flamefire:
For future reference: The issue is that `EasyConfig.copy` does not copy the easyconfig but parses it from scratch. When it resolves dependencies it will reparse them again.
Previously we did not copy EasyConfigs when calling `ec.copy` as that `copy` is `dict.copy` not `EasyConfig.copy` but difficult to see at this point .
With that fixed (introduced in 2.3.0 and only rarely causing issues) (I think) we get for a dependency chain of A->B->C:

- Parse A
- Resolve B
- Parse B
- Resolve C
- Parse C
- Copy C
- Parse C
- Copy B
- Parse B
- Resolve C
- Copy C
- Copy A
- Resolve B
- Copy B
- Parse B
- Resolve C
- Copy C
- Parse C


> For now, we need a quick & clean fix, and that's exactly what this PR is: reverting back to how things were, rather than trying to fix the problem that was introduced in https://github.com/easybuilders/easybuild-framework/pull/4818 in a rush...

No objection. I was just suggesting to add a comment in the place which has a known issue instead of just hiding it again.

Tracking issue: https://github.com/easybuilders/easybuild-framework/issues/5167

casparvl (approved):

PR #5169, opened 2026-04-09
Add customizable Themes and Highlighters for rich colors in `print_msg`

Crivella:
This is a followup on 

- #5125
- #4918

With this PR, Easybuild ships its own highlighter and theme that can be replaced using entrypoints (for now only the theme) using the following groups

- easybuild.rich_theme: A dictionary that defines the rich theme to be used

TODO:

- [x] Expand what EB should highlight by default
- [x] Define a good color scheme to avoid too much clutter
- [x] Add to the entrypoint documentation
- [x] Tests

Example of themes via entrypoints:
- https://github.com/Crivella/test_eb_ep/commit/5ea8a589be9efcd38f453f0117dbe81a9541a76c

### DIFF
--- a/easybuild/tools/output.py
+++ b/easybuild/tools/output.py
@@ -35,11 +35,14 @@ import functools
 from collections import OrderedDict
 import sys
 
-from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.config import OUTPUT_STYLE_RICH, build_option, get_output_style
+from easybuild.tools.build_log import EasyBuildError, EB_MSG_PREFIX
+from easybuild.tools.entrypoints import EntrypointRichTheme
+from easybuild.tools.config import OUTPUT_STYLE_RICH, build_option, get_output_style, DEFAULT_THEME_NAME
 
 try:
     import rich.markup
+    from rich.theme import Theme
+    from rich.highlighter import Highlighter, RegexHighlighter, ReprHighlighter
     from rich.console import Console, Group
     from rich.live import Live
     from rich.progress import BarColumn, Progress, SpinnerColumn, TextColumn, TimeElapsedColumn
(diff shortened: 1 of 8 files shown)


Crivella:
@boegel @smoors Let me know design-wise what you think about this, if it is worth investing more time in trying to expand this more (eg also make the highlighter pluginable).

In general going to need feedback on colors and patterns, my UI design skills are not the best so i am afraid of going full rainbow with the colors

smoors:
looks good to me design-wise!
i quickly tested it and it works well, also with the entrypoints.

it would be nice if you can also add support for customizing the things that are colored automatically, like paths, numbers, ... (i'm not a fan of the default colors)

i can come up with a default color scheme, but i don't want to get into a bike-shedding match :)

Crivella:
It is already possible by overriding the defaults styles the same way you would by setting `easybuild.success` and similar

You can see all the defaults with `python -m rich.default_styles`

smoors:
> It is already possible by overriding the defaults styles the same way you would by setting `easybuild.success` and similar
> 
> You can see all the defaults with `python -m rich.default_styles`

cool, thx!

smoors:
@Crivella  i played a bit with the default theme, and settled on this PR:
https://github.com/Crivella/easybuild-framework/pull/3

see light and dark screenshots below for the result.
feel free to modify as you please.

Screenshot From 2026-04-25 19-46-04
Screenshot From 2026-04-25 19-45-36

easybuild/tools/output.py
```diff
@@ -75,6 +78,35 @@
...
+    'easybuild.prefix1': 'gray50',
+    'easybuild.prefix2': 'gray50',
+    'easybuild.installing': 'bold',
+    'easybuild.installed': 'bold',
+    'easybuild.timing': 'gray50',
+    'repr.path': 'bright_blue',
+    'repr.number': 'red',
+    'repr.ipv6': 'yellow',  # time
+    'repr.call': 'none',  # file(s)
+    'repr.ellipsis': 'gray50',
```
smoors: older rich versions apparently only support `grey`, not `gray`

```suggestion
    'easybuild.prefix1': 'grey50',
    'easybuild.prefix2': 'grey50',
    'easybuild.installing': 'bold',
    'easybuild.installed': 'bold',
    'easybuild.timing': 'grey50',
    'repr.path': 'bright_blue',
    'repr.number': 'red',
    'repr.ipv6': 'yellow',  # time
    'repr.call': 'none',  # file(s)
    'repr.ellipsis': 'grey50',
```

Crivella:
@smoors

I''ve overhauled a bit the way the Entrypoints are defined to homogenize the way they are exposed with the already existing ones.
In this case since you cannot decorate a dictionary we have to have them as functions that return the dictionary.

This allows to also easily show them as available EPs when running `--show-config`

I've' also added the option to customize the highlighters, just unsure if the `build_option` and CLI param should be `output[_/-]highlights` or `output[_/-]highlighters`

Crivella:
Another idea which might be overkill is to allow a `NONE` value to disable our custom highlighting but leave the `rich` one on

PR #5171, opened 2026-04-10
don't call `render` when using `--dep-graph` to create `.dot` file

Crivella:
Fix for 

- https://github.com/easybuilders/easybuild-easyconfigs/pull/25699
- https://github.com/easybuilders/easybuild-easyconfigs/pull/25700

Do not use the `render` methods for saving `.dot` files which also calculates the position of the elements in the graph.


boegel (approved):
lgtm

PR #5175, opened 2026-04-15
remove CI workflow for Hound style checker

branfosj:
We do `flake8` linting in https://github.com/easybuilders/easybuild-framework/blob/develop/.github/workflows/linting.yml so we no longer need the hound.

Picked up from @Flamefire's comment at https://github.com/easybuilders/easybuild-framework/pull/5174#discussion_r3086368636

From http://help.houndci.com/en/articles/2138537-flake8 it looks like we have `flake8` 3.6.0 and that was [released in 2018](https://flake8.pycqa.org/en/3.6.0/release-notes/index.html).

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,4 +1,0 @@
-# configuration for houndci, see https://houndci.com/configuration#python
-flake8:
-  enabled: true
-  config_file: setup.cfg


Flamefire:
Just noticed: The hound posts just comments but doesn't fail CI. So we could keep it and simply close/resolve/ignore the wrong ones. There doesn't seem to be too many.
Would be great if we had a commenter with a recent-ish flake8 though.

Flamefire:
This seems to be getting urgent: CI now fails with
> Hound — We've encountered an error while reviewing your code.

But no further mention on what the issue is. See e.g. https://github.com/easybuilders/easybuild-framework/pull/5185
That seems to be some internal error, so nothing we can do anything about

Crivella (approved):
LGTM

also this is an easy revert that does not break anything in case the bot gets fixed and we decide we still need it.

Also i think this should be enough to stop the flake check running, but we should also disconnect the bot from the account/repo itself to fully remove it

Crivella:
Going in, thanks @branfosj!

PR #5177, opened 2026-04-20
Improved Slurm backend for job submission

besserox:
This is an improved version of the Slurm backend for job submission (option `--job`).

## Same functionalities

- Submit each build as an individual Slurm job
- Handle dependencies between Slurm jobs

## New functionalities

- Respect the `--job-max-jobs` setting, and manage a submission queue accordingly
- Track the state of the Slurm jobs and print the current status according to the option `--job-polling-interval`
- Print a summary of all build jobs status at the end

## Change of behavior

- Terminal output if different: Any tool relies on the output generated by the `--job Slurm` option will be broken.
- Synchronous behavior: With this updated backend, EasyBuild waits for all the build jobs to be finished to exit and report the status of the whole build. This is similar to the natural behavior of EasyBuild (without `--job`) or to the behavior of other backend like `GC3Pie`. This differs from the current behavior of the Slurm backend where EasyBuild exits once all the jobs have been submitted. 

If this change of behavior is too important to be merged as it is, I can suggest two options:

- Option A: If using `--job --job-backend Slurm` without `--job-max-jobs`, then use the old behavior, ie asynchronous. If using `--job --job-backend Slurm` with `--job-max-jobs`, then use the new behavior, ie synchronous. However, I think it could be a bit confusing.
- Option B: Instead of updating the Slurm backend, make this a completely new backend with a new name, for example `SlurmQueue`.

## Use of AI tools

- I wrote the code myself, based on the existing Slurm backend, and a prototype version developed by @ekieffer.
- I used Ollama with model Qwen3.5-35b to review and improve my code.

### DIFF
--- a/easybuild/tools/job/slurm.py
+++ b/easybuild/tools/job/slurm.py
@@ -68,7 +68,7 @@ class Slurm(JobBackend):
         super().__init__(*args, **kwargs)
         # Interval between polls for status of jobs (in seconds), ie between each sacct call
         # Set default explicitly. Default value defined in tools/options.py is 30s does not seem to be used.
-        self.job_polling_interval = build_option('job_polling_interval', default = 30)
+        self.job_polling_interval = build_option('job_polling_interval', default=30)
         if self.job_polling_interval < 1:
             raise EasyBuildError("Polling interval for Slurm backend cannot be less than 1s: %s",
                                  self.job_polling_interval)
--- a/easybuild/tools/job/slurm.py
+++ b/easybuild/tools/job/slurm.py
@@ -28,13 +28,17 @@ Support for using Slurm as a backend for --job
 Authors:
 
 * Kenneth Hoste (Ghent University)
+* Emmanuel Kieffer (LuxProvide)
+* Xavier Besseron (LuxProvide)
 """
 import os
 import re
+import time
+import sys
 
 from easybuild.base import fancylogger
 from easybuild.tools import LooseVersion
-from easybuild.tools.build_log import EasyBuildError, print_msg
+from easybuild.tools.build_log import EasyBuildError, print_msg, print_warning, EasyBuildExit
 from easybuild.tools.config import JOB_DEPS_TYPE_ABORT_ON_ERROR, JOB_DEPS_TYPE_ALWAYS_RUN, build_option
 from easybuild.tools.job.backend import JobBackend
 from easybuild.tools.filetools import which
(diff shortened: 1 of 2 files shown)


easybuild/tools/job/slurm.py
```diff
@@ -56,13 +60,18 @@ def __init__(self, *args, **kwargs):
...
-        for cmd in ['sbatch', 'scontrol']:
+        for cmd in ['sbatch', 'scontrol', 'sacct']:
             path = which(cmd)
             if path is None:
                 raise EasyBuildError("Required command '%s' not found", cmd)
 
         super().__init__(*args, **kwargs)
 
+        # Add maximum jobs submitted to a queue
+        self.job_polling_interval = build_option('job_polling_interval')
```
hattom, line 71: I see no default value for polling interval, what is it? It should be impossible to run with e.g. 0, else the code will spam SLURM. Some number of minutes would seem like a sensible minimum?

The comment above polling interval refers to maximum jobs, which is actually set below.

besserox: Thanks for checking:

- The default value for `job_polling_interval` is 30s and it is set in [tools/options.py line 951](https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/options.py#L951).
- I added an additional check to make sure we don't get anything less than 1s.
- I fixed the mixed-up comments.

I have been running it many times with the default polling interval of 30s without any issue.
I believe it is fine to let the user use a low value if he wishes, even though I would not recommend it.

hattom: I see now the default value was already there -- I thought this was a new parameter, so was expecting to see it defined in this PR.
I'll resolve this thread (if I can).

boegel:
@besserox needs some work to make CI happy, `test_build_easyconfigs_in_parallel_slurm` fails with:
```
Required command 'sacct' not found
```

You'll need to add a mocked `sacct` command, like we already do for `sbatch` & `scontrol` ?

test/framework/parallelbuild.py
```diff
@@ -361,6 +382,8 @@ def test_build_easyconfigs_in_parallel_slurm(self):
             'job_cores': 3,
             'job_max_walltime': 5,
             'force': True,
+            'job_polling_interval': 30,
+            'job_max_jobs': 100,
```
besserox, line 386: The default values defined in `tools/options.py` are not set during the testing. I guess it is expected? I have `None` instead.

@boegel I put the values here instead. I guess it is the way to do it.

test/framework/parallelbuild.py
```diff
@@ -378,10 +401,10 @@ def test_build_easyconfigs_in_parallel_slurm(self):
 
         expected = {
             'dependency': 'afterok:%s' % jobs[0].jobid,
-            'hold': True,
             'job-name': 'gzip-1.5-foss-2018a',
             'nodes': 1,
-            'ntasks': 3,
+            'ntasks': 1,
+            'cpus-per-task': 3,
```
besserox, line 407: - This PR also changes the Slurm job parameters from 'N tasks with 1 CPU each' (old behavior) to '1 tasks with N CPUs' (new behavior). This seems more correct to me as there is a single task which is the python process.

- Also the hold 'trick' was removed. It does not appear to be necesary.

As a result, I fixed the expected job specifications in the testing. Does this look good to you? @boegel

besserox:
> You'll need to add a mocked `sacct` command, like we already do for `sbatch` & `scontrol` ?

Thanks for the reminder :smile: 
I found my way to add the mocked sacct command and fixed  the tests.

@boegel  Let me know if there are other issues.

PR #5180, opened 2026-04-27
Add type hints

Flamefire:
This is a draft to show how it would like with type hints

It was (almost) entirely created by an AI agent (Opencode w/ Llama 3.3) to evaluate the usefulness before doing manual work.
Take as good examples:

```
# Clear what will be put in later
_changes: Dict[str, str] = {}
KNOWN_KEYS: List[str] = []
# Precise description of the key-values which would be harder in plain English.
# Example contents: `COMPILER_CUDA_UNIQUE_OPTS['foo'] -> (['x', 'y'], 'z')`
COMPILER_CUDA_UNIQUE_OPTS: Dict[str, Tuple[List[str], str]] = {
```

Currently things like this are required:

```
from typing import Optional, List, Union

def foo(x: List[str], y: Optional[int] = None) -> Union(str, bytes):
```

In Python 3.9 types are subscriptable and `Union` is replaced by `|`:

```
# No import

def foo(x: list[str], y: int | None = None) -> str | bytes:
```

There are a static analyzers/linters to verify type hints, e.g. `mypy`
Through that we can enforce type hints and ensure correctness on changes. Type-checking the test suite should provide good coverage of our interfaces.

cc @boegel

### DIFF
--- a/easybuild/base/exceptions.py
+++ b/easybuild/base/exceptions.py
@@ -34,10 +34,12 @@ import inspect
 import logging
 import os
 
+from typing import Optional, List
+
 from easybuild.base import fancylogger
 
 
-def get_callers_logger():
+def get_callers_logger() -> Optional[logging.Logger]:
     """
     Get logger defined in caller's environment
     :return: logger instance (or None if none was found)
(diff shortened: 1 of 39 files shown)


jfgrimm:
If we target this for 6.0, we are planning to require python 3.9+ anyway

Flamefire:
The earlier we do that the better. And it is not breaking anything and through the module compatible with python 3.6

When branching of EB 6.x we can remove the imports and do a (mostly) trivial search&replace. The main work is adding those, keeping them correct and consistent with new/changed code.

jfgrimm:
that's fair

PR #5181, opened 2026-04-27
Remove uneccessary use of dunder methods

Flamefire:
Replace use of "dunder methods" (`__foo__`) by the canonical equivalents. E.g. ` value.__repr__())` -> `repr(value)`

Cleanup suggested by PyLint


boegel (approved):
lgtm

PR #5182, opened 2026-04-30
Fix `fake_module` dependencies for `ExtensionsEasyblock` with `--sanity-check-only`

Crivella:
Currently the `sanity_check_step`  of `ExtensionEasyBlock` uses

https://github.com/easybuilders/easybuild-framework/blob/632d4d24ca91f19077aaa52c52020791d82596a3/easybuild/framework/extensioneasyblock.py#L184

to run the sanity check on the main bundle before doing the sanity checks for each individual extension

At this point when running with `--sanity-check-only` the `prepare_step` is skipped and hence the call to

https://github.com/easybuilders/easybuild-framework/blob/632d4d24ca91f19077aaa52c52020791d82596a3/easybuild/framework/easyblock.py#L3054

which causes `self.toolcahain.dependencies` to be an empty list.

This causes problems when the sanity check commands of the main bundle requires one of the dependencies to be loaded, see for example https://github.com/easybuilders/easybuild-easyblocks/pull/4122#issuecomment-4351222291

As far as i can tell `toolchain.prepare_step` is always called passing `Easyconfig.dependencies()` (in general `self.cfg`) and this only does some checks on the deps passed without modifying them.

I think `self.cfg.dependencies()` should be used here directly unless there was ever a reason to not load the Bundle dependencies when running the main `sanity_check_step` of a Bundle with extensions

Tested this w.r.t. https://github.com/easybuilders/easybuild-easyblocks/pull/4122#issuecomment-4351222291 and it fixes the problem

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1461,7 +1461,7 @@ class EasyBlock:
         # include load/unload statements for dependencies
         deps = []
         self.log.debug("List of deps considered to load in generated module: %s", self.toolchain.dependencies)
-        for dep in self.toolchain.dependencies:
+        for dep in self.cfg.dependencies():
             if dep['build_only']:
                 self.log.debug("Skipping build dependency %s", dep)
             else:
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -1460,7 +1460,7 @@ class EasyBlock:
 
         # include load/unload statements for dependencies
         deps = []
-        self.log.debug("List of deps considered to load in generated module: %s", self.toolchain.dependencies)
+        self.log.debug("List of deps considered to load in generated module: %s", self.cfg.dependencies())
         for dep in self.cfg.dependencies():
             if dep['build_only']:
                 self.log.debug("Skipping build dependency %s", dep)
--- a/test/framework/easyconfig.py
+++ b/test/framework/easyconfig.py
@@ -581,6 +581,12 @@ class EasyConfigTest(EnhancedTestCase):
         toy_patch = os.path.join(self.test_sourcepath, 'toy', toy_patch_fn)
         copy_file(toy_patch, self.test_prefix)
 
+        # Creates empty module files for expected dependencies, needed for `load_fake_module`
+        tmpdir = tempfile.mkdtemp()
+        for dep_mod_name in ('Python/3.6.6', ):
+            write_file(os.path.join(tmpdir, dep_mod_name), "#%Module")
+        os.environ['MODULEPATH'] = tmpdir
+
         os.environ['EASYBUILD_SOURCEPATH'] = self.test_prefix
         init_config(build_options={'silent': True})
 
@@ -592,6 +598,7 @@ class EasyConfigTest(EnhancedTestCase):
             'homepage = "http://example.com"',
             'description = "test easyconfig"',
             'toolchain = SYSTEM',
+            # Needed to resolve %(pymajver)s template in exts_list
             'dependencies = [("Python", "3.6.6")]',
             # bogus, but useful to check whether this get resolved
             'exts_default_options = {"source_urls": [PYPI_SOURCE]}',


Crivella:
The only error I am seeing in the CI is due to `test_extensions_templates` not calling `eb.prepare_step()` which would cause the same error to pop up.
Not calling `eb.prepare_step()` i guess is fine for just testing templates, but is not realistic in an actual run where `extensions_step` is called and we should cover this case.

We should probably fix this test by also creating a fake module for `Python` if we want to test for the python templates similar to what is done in 

https://github.com/easybuilders/easybuild-easyblocks/blob/17d584071c73bbb51ce0ff9f58d17f201027e074/test/easyblocks/module.py#L454-L455

Flamefire:
IMO the actual error there is using the fake module at all: It should use the "sanity_check_module", which for module-only is the real module.

As I can't find the PR I though I made with that change I split up some changes to sanity checks I had and https://github.com/easybuilders/easybuild-framework/pull/5185 implements the fix.

However this might be relevant for `--module-only`. Did you see the same issue there too?

Crivella:
`--module-only` does go through the prepare step so it does work without this.

Still think that we should either use `self.cfg.dependencies()` or make sure that `self.toolchain.dependencies` was actually set

Flamefire:
When `--sanity-check-only` does not use this anymore and in all other uses `self.toolchain.dependencies` is set, it should be fine, shouldn't it?

I'm not sure we actually ever need `self.toolchain.dependencies`. From my understanding this is simply `self.ec.dependencies()` but hidden inside a function that cannot return anything else unless `_check_dependencies` was overwritten to have a different behavior

Crivella:
In theory yes but to avoid confusion we should either have a check on `make_module_dep` that it runs only if something that set `self.toolchain.dependencies` has ran before hand, or use this fix.

This PR also does fix the problem of `fake_module` not working in `--sanity-check-only`.
Using the real module is probably a different discussion, but one worth having. Will try to look into https://github.com/easybuilders/easybuild-framework/pull/5185 as well

Flamefire:
> In theory yes but to avoid confusion we should either have a check on `make_module_dep` that it runs only if something that set `self.toolchain.dependencies` has ran before hand, or use this fix.

I think we cannot check that: It is an empty list initially which is a valid value for a SYSTEM EC. We could potentially init it with `None` instead and watch for potential fallout.

> This PR also does fix the problem of `fake_module` not working in `--sanity-check-only`. Using the real module is probably a different discussion, but one worth having. Will try to look into #5185 as well

With `--sanity-check-only` we only want to test the real module, so we don't need to create a fake module. So we don't actually need to create a fake module and it failing would be a good thing. See
https://github.com/easybuilders/easybuild-framework/blob/632d4d24ca91f19077aaa52c52020791d82596a3/easybuild/framework/easyblock.py#L4320

So that one place in  `ExtensionEasyBlock` where it is  used is clearly a bug.

IIUC the failure though is not clear (enough) so the check mentioned would still be good, but with a clear comment that this is (likely) an internal usage error

easybuild/framework/easyblock.py
```diff
@@ -1461,7 +1461,7 @@ def make_module_dep(self, unload_info=None):
         # include load/unload statements for dependencies
         deps = []
         self.log.debug("List of deps considered to load in generated module: %s", self.toolchain.dependencies)
```
Flamefire:
```suggestion
        self.log.debug("List of deps considered to load in generated module: %s", self.cfg.dependencies())
```

Flamefire:
I'd say this is a workaround for a bug which avoids a known issue so could be merged now until #5185 fixes the underlying bug that `--sanity-check-only` doesn't (always) check the actually installed module as it's supposed to.

Only concern I have is that this hides the bug so we might miss related issues in the future

Crivella:
Even with 

- https://github.com/easybuilders/easybuild-framework/pull/5185

merged this is still relevant as it addresses a separate problem that is calling `fake_module_environment` in a situation where the `prepare_step` is skipped or overwritten in a way to skip the call to `self.toolchain.prepare`, causing the dependencies to not be added to the fake environment

Flamefire:
> a separate problem that is calling `fake_module_environment` in a situation where the `prepare_step` is skipped or overwritten in a way to skip the call to `self.toolchain.prepare`,

How exactly can this happen? Isn't the prepare step always executed except for sanity-check-only?
However I'm not opposed to that. I would even go a step further and deprecate `toolchain.dependencies` to avoid e.g. doing more to it in the prepare step that we don't account for e.g. here. Besides that it is confusing: Why does the toolchain has dependencies of an easyconfig? So either it really only has the toolchain dependencies or at least better documentation.

I'm not sure why we need it at all: It could be a property of `EasyBlock` and a separate method that does only the checks as [`_check_dependencies`](https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/toolchain/toolchain.py#L565)  runs a check only and either raises an exception or returns them all.

This also causes a small backwards-compatibility risk: Users could have their own toolchains and do something with `dependencies`

Crivella:
> How exactly can this happen?

Mostly custom easyblocks that could override the prepare step.
For that we either decide that `fake_module_environment` should give the correct result regardless or give a clear failure that it is not intended to use without having called `prepare_step` to avoid having to hunt down errors due to missing dependencies.

Flamefire:
Ok so I see 3 options:
- Remove `toolchain.dependencies`
- Make it a property that throws an error on access if not set (in prepare_step), still a minor breaking change for custom toolchains and easyblocks relying on it to be empty.
- deprecate accessing it before it is set

I'd go for the first as argued above

boegel:
> Ok so I see 3 options:
> 
>     * Remove `toolchain.dependencies`
> 
>     * Make it a property that throws an error on access if not set (in prepare_step), still a minor breaking change for custom toolchains and easyblocks relying on it to be empty.
> 
>     * deprecate accessing it before it is set
> 
> 
> I'd go for the first as argued above

We can't just blindly remove `self.toolchain.dependencies`, it's essentially part of the API, so some easyblocks are using it (`Bundle` currently, primarily), we should at least deprecate the use of it first (and only remove in EasyBuild 6).

Flamefire:
> We can't just blindly remove `self.toolchain.dependencies`, it's essentially part of the API, so some easyblocks are using it (`Bundle` currently, primarily), we should at least deprecate the use of it first (and only remove in EasyBuild 6).

Not immediately yes, that was meant for the general direction.
I would actually go for:

- Make it a property
- Deprecate accessing it
- Error when reading before initialization (prepare_step)

boegel:
@Crivella Is this still useful now that this other fix got merged?
- https://github.com/easybuilders/easybuild-framework/pull/5185

Crivella:
@boegel 

For me this change is still beneficial as it allows to `make_module_dep` to give the correct result (including dependencies) even if called without calling `self.toolchain.prepare` usually called in the prepare_step.

If we want this to not work unless `self.toolchain.prepare` was called we should otherwise give an hard failure that explains this (instead of down the line getting other failures due to missing deps)

Flamefire:
> For me this change is still beneficial as it allows to `make_module_dep` to give the correct result (including dependencies) even if called without calling `self.toolchain.prepare` usually called in the prepare_step.

When does this happen? I don't see a reasonable situation where we'd want that.

> If we want this to not work unless `self.toolchain.prepare` was called we should otherwise give an hard failure that explains this (instead of down the line getting other failures due to missing deps)

I'd say so and we can just go ahead with the deprecation route outline above.

Crivella:
In my opinion if `make_module_dep` can work without the `prepare_step` or with a custom one that does not set the toolchain dependencies i do not see why not allow that, but i am fine either way

Flamefire:
Sure it *can* work but I don't see why it *should* work.
However the change is quite trivial, so doesn't hurt I guess. Would then just leave the question if there are other parts not yet properly initialized in such an "unintended" situation.

FYI: I opened #5234 so we detect this case at least and document what that property actually is.
We do need it: The toolchain has no access to the easyconfig because the easyconfig has-a toolchain, but it needs to know the dependencies so it can set proper flags
The tests currently fail but would be fixed by this PR: In our tests we often call some later step method without ever calling the prepare step. Not sure if that is a good idea though.

PR #5183, opened 2026-04-30
fix copy-paste error in `test_toy_cuda_sanity_check` and simplify

Flamefire:
The major issue was in `assert_cuda_report` which wrongly was calling `assert_regex(num_checked_str, outtxt, stdout)`.
But the parameter to the function is `log` not `outtxt` 

This went undetected because we do not enable the Flake8 check for unused variables/parameters.
I'd argue we should, even if that means a bit initial refactoring: Most uses should be ok, so we simply silence the warning by using `_` as the name or prefix the variable with an underscore.

While doing that I noticed that there are a lot of `assertTrue` checks when testing for a regex. In https://github.com/easybuilders/easybuild-framework/pull/5137 I replaced many of those already by the better `assertRegex` which avoids us having to specify the message.
This removes almost half of the lines of the test: 1. Compile regex, 2. Assemble message, 3. do the check are replaced by just doing the check.

There were also cases where the regex was wrong, e.g. `r"Missing compute capabilities: 8.0."`
The dot would need to be escaped, but the better check is `assertIn`, we don't need a regex here.

### DIFF
--- a/test/framework/toy_build.py
+++ b/test/framework/toy_build.py
@@ -610,8 +610,8 @@ class ToyBuildTest(EnhancedTestCase):
 
         # test specifying a non-existing group
         allargs = [test_ec] + args + ['--group=thisgroupdoesnotexist']
-        outtxt, err = self.run_eb_main_capture_output(allargs, logfile=self.dummylogfn, do_build=True,
-                                                      return_error=True)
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

PR #5184, opened 2026-04-30
Do not duplicate sanity checks in each extension

Flamefire:
We have some checks that are run on the whole install folder:

- required/banned libraries
- (Fortran) module files
- RPATHs
- CUDA compute capabilities

It does not make sense to run those in the extensions which would duplicate the work. All those checks involve calling external binaries which takes time and doing that for 10s of extensions needlessly increases the time to build/check the module

Move them all into the check that the step is not ran for an extension.

Includes/Requires:
- [x] https://github.com/easybuilders/easybuild-framework/pull/5183

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -4455,32 +4455,32 @@ class EasyBlock:
                     else:
                         print_warning(mod_files_found_msg)
 
+            if self.toolchain.use_rpath:
+                rpath_fails = self.sanity_check_rpath()
+                if rpath_fails:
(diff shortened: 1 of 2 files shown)


boegel:
#5183 is merged, please sync with `develop` so we have a clear view here...

Flamefire:
Done.

boegel (approved):
lgtm

boegel:
The impact of this is quite limited I think, mainly since `sanity_check_step` is customized in `Extension`, but it doesn't hurt...

PR #5185, opened 2026-04-30
Use correct module in sanity check of extensions

Flamefire:
I identified a few things we do in the sanity check of extensions we should not:

- check for things on the whole install tree: folders, rpaths, libraries, CUDA CCs
- Create a module in `--sanity-check-only` mode

I'm fixing the first one in https://github.com/easybuilders/easybuild-framework/pull/5184 and duplicate checks for (default) paths (bin & lib) in https://github.com/easybuilders/easybuild-framework/pull/5221

The fix here is the 2nd one with this simple diff in `extensioneasyblock`:

```diff
-               with self.fake_module_environment(extra_modules=extra_modules):
+               with self.sanity_check_module_environment(extra_modules=extra_modules):
```

Not only does this avoid the complex, time-consuming logic of creating a module and manipulating MODULEPATH it might fail if module creation requires information from the build process.
See e.g. https://github.com/easybuilders/easybuild-easyblocks/pull/4122 where `get_software_version` was used, i.e. the dependency module was required to be loaded.
I added a `sanity_check_module_environment` context manager to load & unload the sanity check module (real module or fake module)

Another part was removing the `extension` parameter to those methods, which I did in a PR prior to the 5.3.0 release but took out to get that PR merged without this change. Move to https://github.com/easybuilders/easybuild-framework/pull/5220


easybuild/framework/easyblock.py
```diff
@@ -4349,6 +4391,11 @@ def _sanity_check_step(self, custom_paths=None, custom_commands=None, extension=
         :param extension: indicates whether or not sanity check is run for an extension
         :param extra_modules: extra modules to load before running sanity check commands
         """
+        if extra_modules is not None and not isinstance(extra_modules, list):
+            if isinstance(extra_modules, bool):
+                self.log.nosupport("Do not pass `extension` to `_sanity_check_step`", '5.3.1')
+            raise EasyBuildError("extra_modules should be a list of module names, got %s (type: %s)",
+                                 extra_modules, type(extra_modules))
         paths, path_keys_and_check, commands = self._sanity_check_step_common(custom_paths, custom_commands)
```
Crivella: I would argue that this can be any iterable and not just a list

easybuild/framework/easyblock.py
```diff
@@ -4349,6 +4391,11 @@ def _sanity_check_step(self, custom_paths=None, custom_commands=None, extension=
         :param extension: indicates whether or not sanity check is run for an extension
         :param extra_modules: extra modules to load before running sanity check commands
         """
+        if extra_modules is not None and not isinstance(extra_modules, list):
+            if isinstance(extra_modules, bool):
```
Crivella: If you really wanna do this, i would not alter the function signature and set `extension=None` and do the checks and deprecating here.

test/framework/toy_build.py
```diff
@@ -3544,9 +3576,10 @@ def test_toy_build_trace(self):
             r"  >> command completed: exit 0, ran in .*",
             r'^' + r'\n'.join([
                 r"== sanity checking\.\.\.",
+                r"  >> loading modules: toy/0.0...",
                 r"  >> file 'bin/yot' or 'bin/toy' found: OK",
                 r"  >> \(non-empty\) directory 'bin' found: OK",
-                r"  >> loading modules: toy/0.0\.\.\.",
+                r"  >> loading modules: toy/0.0...",
```
Crivella: `.` in regex is any char, so this is not correct if we expect exactly a dot we should keep `\.`, same above

test/framework/sandbox/easybuild/easyblocks/generic/toy_extension.py
```diff
@@ -44,6 +44,8 @@ def extra_options():
         """Custom easyconfig parameters for toy extensions."""
         extra_vars = {
             'toy_ext_param': ['', "Toy extension parameter", CUSTOM],
+            'toy_custom_cmd': [None, "Custom command to run in sanity check", CUSTOM],
```
Crivella: Not sure `toy_custom_cmd` should have a clearer name since it is used specifically for sanity checks

test/framework/toy_build.py
```diff
@@ -2494,6 +2494,37 @@ def test_toy_sanity_check_commands(self):
         regex = re.compile('^.*/eb-[^/]+/eb-sanity-check-[^/]+\n[ ]*0$')
         self.assertTrue(regex.match(out_txt), f"Pattern '{regex.pattern}' should match in: {out_txt}")
 
+    def test_extension_sanity_check_custom_commands(self):
+        """Test whether fallback in sanity check for lib64/ equivalents of library files works."""
```
Crivella: Change to describe what is being tested here

easybuild/framework/easyblock.py
```diff
@@ -4181,9 +4217,11 @@ def _sanity_check_step_common(self, custom_paths, custom_commands):
         # verify sanity_check_paths value: only known keys, correct value types, at least one non-empty value
         only_list_values = all(isinstance(x, list) for x in paths.values())
         only_empty_lists = all(not x for x in paths.values())
-        if sorted_keys != known_keys or not only_list_values or only_empty_lists:
+        if sorted_keys != known_keys or not only_list_values or (only_empty_lists and not self.is_extension):
             error_msg = "Incorrect format for sanity_check_paths: should (only) have %s keys, "
-            error_msg += "values should be lists (at least one non-empty)."
+            error_msg += "values should be lists"
+            if not self.is_extension:
```
Crivella: Sholdn't this be

```suggestion
            if only_empty_lists:
```

?

easybuild/framework/easyblock.py
```diff
@@ -4160,8 +4194,10 @@ def _sanity_check_step_common(self, custom_paths, custom_commands):
                 paths = {}
                 for key in path_keys_and_check:
                     paths.setdefault(key, [])
-                paths.update({SANITY_CHECK_PATHS_DIRS: ['bin', ('lib', 'lib64')]})
-                self.log.info("Using default sanity check paths: %s", paths)
+                # Default paths for extensions are handled in the parent easyconfig if desired
+                if not self.is_extension:
+                    paths.update({SANITY_CHECK_PATHS_DIRS: ['bin', ('lib', 'lib64')]})
+                    self.log.info("Using default sanity check paths: %s", paths)
```
Crivella: Not sure, wouldn't this risk lowering the sanity checks we do for existing ECs at the very least?

easybuild/framework/easyblock.py
```diff
@@ -3444,6 +3468,16 @@ def post_processing_step(self):
...
+        if 'extension' in kwargs:
+            extension = kwargs.pop('extension')
+            self.log.deprecated(
+                "Passing `extension` to `sanity_check_step` is no longer necessary and will be ignored "
+                f"(Easyblock: {self.__class__.__name__}).",
+                '6.0',
+            )
+            if extension != self.is_extension:
+                raise EasyBuildError('Unexpected value for `extension` argument. '
+                                     f'Should be: {self.is_extension}, got:  {extension}')
```
Crivella: See comment below, i think this should be handled in `_sanity_check_step` if you really want to do this

easybuild/framework/easyblock.py
```diff
@@ -2409,6 +2408,31 @@ def fake_module_environment(self, extra_modules=None, with_build_deps=False):
...
+
+        try:
+            yield
+        finally:
+            # cleanup (unload fake module, remove fake module dir)
+            if unload_module:
+                if self.fake_mod_data:
+                    self.clean_up_fake_module(self.fake_mod_data)
+                    self.fake_mod_data = None
+                self.sanity_check_module_loaded = False
```
Crivella, line 2433: The properly logic here would be to use an RLock or something similar but the original `fake_module_environment` would have the same problem, bu this is a minor rant since i am not not sure if we ever use this in threaded parts of the code

easybuild/framework/easyblock.py
```diff
@@ -2409,6 +2408,31 @@ def fake_module_environment(self, extra_modules=None, with_build_deps=False):
             if fake_mod_data:
                 self.clean_up_fake_module(fake_mod_data)
 
+    @contextmanager
+    def sanity_check_module_environment(self, extra_modules=None, with_build_deps=False, check_loaded=True):
```
Crivella: Unsure if `check_loaded` should be True by default as it risks breaking stuff that already exists

Crivella (changes requested):

easybuild/framework/easyblock.py
Flamefire: The intention here is to allow empty lists for extensions as the paths are usually already checked in the parent.
Hence the change in the error condition: `only_empty_lists and not self.is_extension`
So it is an error if you have an empty list in a non-extension.

easybuild/framework/easyblock.py
Flamefire: This is a new function, so we needed to check each call site. Hence I went with the safe defaults. The parameter is an escape hatch only for rare cases. FWIW: I'm building with this for 3 months now and haven't hit an issue

easybuild/framework/easyblock.py
Flamefire: Indeed: When a bundle easyconfig sets `sanity_check_paths` without `enhance_sanity_check` then we don't check those folders.
A quick check shows those often check a file in `bin` and/or e.g `lib/python*`, but e.g. VisCello only checks for a `$name` dir

I'd like to keep this change as it avoids the check for the same folders in possibly hundreds of extensions. We could go over existing easyconfigs and add `enhance_sanity_check =  True` or the relevant folder(s) where appropriate. Scrolling over `grep -A2 sanity_check_paths $(rg --files-with-matches exts_list)` shows that most do have supersets of those checks/lists in the checks

easybuild/framework/easyblock.py
Flamefire: Yes, the parameter is confusing and could be wrong due to that.
This check is just to be extra cautious for external usages as technically this is a public API (at least the caller is)
Keeping the parameter means we always need to pass `extra_modules` by name. I'd be fine with that though.

easybuild/framework/easyblock.py
Flamefire: `extra_modules`? No it needs to be a list. `load_module` concats it with another list

easybuild/framework/easyblock.py
Crivella: But you can get the message under the if even it is not true like this no?

easybuild/framework/easyblock.py
Crivella: Problem is like this people will get both the deprecation warning from `_dispatch_sanity_check_step` and than hard fail in `_sanity_check_step` (either because they were passing 3 positional arguments so the check fails or because they were passing all 4 and now they don't match the function signature)

easybuild/framework/easyblock.py
Flamefire: For extensions:
> [...] values should be lists

For main easyconfigs:
> [...] values should be lists (at least one non-empty).

That is exactly showing what the requirements are. The only change to what we had before is that it does **not** show the "(at least one non-empty)"-part for extensions as enforced by the changed condition.

Crivella: Kay now i understand i was thinking in the term of writing the message tailored to the error that triggered, but here we are going the other way of writing it to include every possible error

easybuild/framework/easyblock.py
Flamefire: I wanted to catch this for dry-run too, changed but with caveat, see comment below

easybuild/framework/easyblock.py
Flamefire: So it is only an issue when passing all after the first 2 as positional arguments. However that currently does not work for dry-run mode which is defined as `def _sanity_check_step_dry_run(self, custom_paths=None, custom_commands=None, **_):` so only accepts kwargs

IMO the removal is cleaner, but handling the extra param here is slightly safer. So done both:
Check the kwargs in the dispatcher, keep the arguments here but check (again). That keeps the check for the dry-run mode.

test/framework/sandbox/easybuild/easyblocks/generic/toy_extension.py
Flamefire: renamed to `toy_custom_sanity_check_cmd`. A bit long, but explicit.

test/framework/toy_build.py
Flamefire: Ups, C&P. Done: Renamed to `test_extension_sanity_check` and added:

> Check sanity check for extensions:
> Custom_commands from easyblocks are run.
> Paths are checked only once: The default for sanity_check_paths in extensions is empty.

test/framework/toy_build.py
Flamefire: Fixed by converting to text checked with `assertIn`.
Related: https://github.com/easybuilders/easybuild-framework/pull/5188 does a larger conversion of regex-assertions reducing error-prone duplication and doing similar replacements where plain text is checked using regex matching.

test/framework/toy_build.py
```diff
@@ -2710,12 +2744,12 @@ def test_toy_build_enhanced_sanity_check(self):
         stdout = self.get_stdout()
         self.mock_stdout(False)
 
-        pattern_lines = [
-            r"^== sanity checking\.\.\.",
-            r"  >> file 'bin/toy' found: OK",
-        ]
-        regex = re.compile(r'\n'.join(pattern_lines), re.M)
-        self.assertTrue(regex.search(stdout), "Pattern '%s' should be found in: %s" % (regex.pattern, stdout))
+        expected_out = toy_custom_sanity_check_step = textwrap.dedent("""
```
Crivella: I do not see this being used?
```suggestion
        expected_out = textwrap.dedent("""
```

test/framework/sandbox/easybuild/easyblocks/generic/toy_extension.py
```diff
@@ -106,10 +108,18 @@ def post_install_extension(self):
...
+            }
+            if self.src:
+                custom_paths['files'].extend(['bin/%s' % self.name, 'lib/lib%s.a' % self.name])
+        else:
+            custom_paths = None
+        custom_cmd = self.cfg['toy_custom_sanity_check_cmd']
+        if custom_cmd:
+            custom_cmds = [custom_cmd]
+        else:
+            custom_cmds = None
```
Crivella: Not required but i am wondering if we should make this a `toy_custom_sanity_check_cmds` which default to a list which is closer in usage to `sanity_check_commands` and would allow just doing `custom_cmds = self.cfg['toy_custom_sanity_check_cmds']`

Flamefire: at most 2 more characters per use, so yeah, makes sense

easybuild/framework/easyblock.py
Crivella: Not sure if it is worth to have double warning on normal sanity check runs just to print a deprecation warning also when running in dry run mode.

The keyword arg was already being ignored there anyway

Flamefire: There is no double-warning, is there? When you pass it as positional parameter you get the warning in `_sanity_check_step` and a failure in dry run.
When passing by keyword you get the (only) warning here in both cases.

Crivella: A missed it was being popped as a kwarg

Flamefire:
Sorry for the force-pushes, I factored out the PR to deprecate the `extension` parameter to #5220 and the duplicated directory checks to #5221 so this is much smaller now. If I didn't miss anything I don't strictly need `self.is_extension` vs `extension` here, so it should still work.

test/framework/toy_build.py
```diff
@@ -2494,6 +2494,35 @@ def test_toy_sanity_check_commands(self):
...
+        write_file(test_ec, test_ec_txt)
+        error_pattern = 'sanity check command echo "Run-Custom-Cmd for barbar" && false failed with exit code 1'
+        with self.mocked_stdout_stderr():
+            self.assertErrorRegex(EasyBuildError, error_pattern, self._test_toy_build, ec_file=test_ec,
+                                  raise_error=True, verbose=False)
+
+        test_ec_txt += (
+            "\nexts_list[0][2]['toy_custom_sanity_check_cmds'] = ['echo \"Run-Custom-Cmd for %(name)s\" && true']"
+        )
+        write_file(test_ec, test_ec_txt)
```
Crivella:
```suggestion
        test_ec_txt += '\n' + textwrap.dedent("""
            exts_list = [
                ('barbar', '0.0', {
                    'exts_filter': ('ls -l lib/lib%(ext_name)s.a', ''),
                    'toy_custom_sanity_check_cmds': ['echo "Run-Custom-Cmd for %(name)s" && {}'],
                    'sanity_check_paths': {'dirs': [], 'files': ['lib/libbarbar.a']},
                })
            ]
        """)
        test_ec = os.path.join(self.test_prefix, 'test.eb')
        write_file(test_ec, test_ec_txt.format("false"))
        error_pattern = 'sanity check command echo "Run-Custom-Cmd for barbar" && false failed with exit code 1'
        with self.mocked_stdout_stderr():
            self.assertErrorRegex(EasyBuildError, error_pattern, self._test_toy_build, ec_file=test_ec,
                                  raise_error=True, verbose=False)

        write_file(test_ec, test_ec_txt.format("true"))
```

Think this would make the 2 tests more readable

test/framework/toy_build.py
```diff
@@ -2494,6 +2494,32 @@ def test_toy_sanity_check_commands(self):
...
+            ]
+        """)
+        test_ec = os.path.join(self.test_prefix, 'test.eb')
+        write_file(test_ec, test_ec_txt.format("false"))
+        error_pattern = 'sanity check command echo "Run-Custom-Cmd for barbar" && false failed with exit code 1'
+        with self.mocked_stdout_stderr():
+            self.assertErrorRegex(EasyBuildError, error_pattern, self._test_toy_build, ec_file=test_ec,
+                                  raise_error=True, verbose=False)
+
+        write_file(test_ec, test_ec_txt.format("true"))
```
Crivella:
```suggestion
                    'toy_custom_sanity_check_cmds': ['echo "Run-Custom-Cmd for %(name)s" && PLACEHOLDER'],
                    'sanity_check_paths': {'dirs': [], 'files': ['lib/libbarbar.a']},
                })
            ]
        """)
        test_ec = os.path.join(self.test_prefix, 'test.eb')
        write_file(test_ec, test_ec_txt.replace('PLACEHOLDER', 'false'))
        error_pattern = 'sanity check command echo "Run-Custom-Cmd for barbar" && false failed with exit code 1'
        with self.mocked_stdout_stderr():
            self.assertErrorRegex(EasyBuildError, error_pattern, self._test_toy_build, ec_file=test_ec,
                                  raise_error=True, verbose=False)

        write_file(test_ec, test_ec_txt.replace('PLACEHOLDER', 'true'))
```

MB the fact that there are other non escaped curly braces causes format to fail and we cant use `%(XXX)s` + `% {'XXX': 'true'}` since there is a non escaped `%` in the text this is the closes thing

easybuild/framework/easyblock.py
```diff
@@ -2407,6 +2406,31 @@ def fake_module_environment(self, extra_modules=None, with_build_deps=False):
...
+        """Load/Unload module for performing sanity checks"""
+        if self.sanity_check_module_loaded and check_loaded:
+            raise EasyBuildError("Sanity check module is already loaded and must not be loaded again")
+
+        if self.sanity_check_module_loaded:
+            unload_module = False
+        else:
+            if with_build_deps:
+                # load modules for build dependencies as extra modules
+                extra_modules = [dep['short_mod_name'] for dep in self.cfg.dependencies(build_only=True)]
```
Crivella: I guess this was an issue also with the existing `fake_module_environment` but i am wondering if it is allright to silently ignore the passed `extra_modules` if `with_build_deps` is set to true.

At minimum we should have a `self.log.warning` here, but i am unsure why we are not just concatenating the 2

Flamefire: It **is** an issue with  the existing `fake_module_environment` yes
Though it doesn't matter, yet. The only place where we use `extra_modules` is in the sanity check of the ExtensionEasyBlock and there we naturally don't use build deps.

I would say we fix this here in both rather than introducing another pitfall: If `with_build_deps` is True then we we use all build deps and append all extra_modules that are not in there yet.

Technically that is a breaking change, but I'd say it is a bugfix only that can only affect some custom code which has likely already worked around that by passing build deps as extra_modules instead of setting both, if such code even exists.

What do you think?

Crivella: I think for now we just put a warning log and maybe open an issue about it, would not delay this PR further

Flamefire: I actually removed this parameter entirely: For the sanity checks including build-deps doesn't make much sense.
We can add it later if we ever discover a use case for that.

easybuild/framework/easyblock.py
```diff
@@ -2395,6 +2394,10 @@ def fake_module_environment(self, extra_modules=None, with_build_deps=False):
         fake_mod_data = None
 
         if with_build_deps:
+            if extra_modules:
+                print_warning("`with_build_deps` overwrites `extra_modules` in fake_module_environment",
```
Crivella:
```suggestion
                print_warning("`with_build_deps` overwrites `extra_modules` in fake_module_environment. "
```
The comma is most likely wrong, any argument other than the first is used to format the first

Crivella (changes requested):

Crivella (approved):
LGTM

Crivella:
Going in, thanks @Flamefire!

PR #5186, opened 2026-05-02
format extension names in module files with `exts_formatter` function

smoors:
will do nothing without companion easyblock PR (so can be safely merged):
- https://github.com/easybuilders/easybuild-easyblocks/pull/4125

but it needs to be merged to make the easyblock PR work.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2020,7 +2020,10 @@ class EasyBlock:
                 if not callable(formatter):
                     raise EasyBuildError("Parameter exts_formatter should be a function that accepts the extension name"
                                          "and returns the formatted name.")
+                self.log.debug(f"Using provided function to format extension names: {formatter}")
                 exts_list = [(formatter(x), y) for (x, y) in exts_list]
+            else:
+                self.log.debug("No formatter function for extension names specified")
 
         return exts_list
 
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2012,6 +2012,11 @@ class EasyBlock:
             else:
                 exts_list.append((resolve_template(ext[0], self.cfg.template_values),
                                   resolve_template(ext[1], self.cfg.template_values)))
+
+        formatter = self.cfg.get('exts_formatter')
+        if formatter:
+            exts_list = [(formatter(x), y) for (x, y) in exts_list]
+
         return exts_list
 
     def prepare_for_extensions(self):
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -2010,6 +2013,12 @@ def make_extension_list(self):
             else:
                 exts_list.append((resolve_template(ext[0], self.cfg.template_values),
                                   resolve_template(ext[1], self.cfg.template_values)))
+
+        if formatted:
+            formatter = self.cfg.get('exts_formatter')
+            if formatter:
```
boegel, line 2019: Can/should we check whether `formatter` is indeed a function, and throw a sensible error if not?

Without that, error messages produced when say `exts_formatter = True` is used will be quite nasty/confusing...

smoors: done in https://github.com/easybuilders/easybuild-framework/pull/5186/commits/610c21c0c2b20f327fa81644ba8f7059c821fbf4

easybuild/framework/easyblock.py
```diff
@@ -2010,6 +2013,15 @@ def make_extension_list(self):
             else:
                 exts_list.append((resolve_template(ext[0], self.cfg.template_values),
                                   resolve_template(ext[1], self.cfg.template_values)))
+
+        if formatted:
+            formatter = self.cfg.get('exts_formatter')
```
boegel, line 2018: As it is now, framework is making the assumption that some easyblock may define something named `exts_formatter`.

To avoid that, we should also add a default value for `exts_formatter` as a general easyconfig parameter, in `framework/easyconfig/default.py`

easybuild/framework/easyblock.py
```diff
@@ -2010,6 +2013,15 @@ def make_extension_list(self):
...
+            if formatter:
+                if not callable(formatter):
+                    raise EasyBuildError("Parameter exts_formatter should be a function that accepts the extension name"
+                                         "and returns the formatted name.")
+                exts_list = [(formatter(x), y) for (x, y) in exts_list]
```
boegel: We need to be more careful here, there may be single-element tuple entries in `exts_list` (see `if isinstance(ext, str):` above), and that would make this crash hard:
```
>>> exts_list = [('example', '1.0'), ('foo', )]
>>> formatter = lambda x: x
>>> exts_list = [(formatter(x), y) for (x, y) in exts_list]
  File "<stdin>", line 1, in <listcomp>
ValueError: not enough values to unpack (expected 2, got 1)
```

boegel (changes requested):
@smoors both issues tackled in:
- https://github.com/smoors/easybuild-framework/pull/5

boegel (approved):
lgtm

PR #5187, opened 2026-05-05
enhance `cmd.sh` so it works in non-interactive mode, like when passing `-c` to run a command

Flamefire:
When running `bash -i` with `-c` it might have unintended side effects, like modifying the parent processes state.
This manifests e.g. in the tests:
`python -m test.framework.suite test_run_shell_cmd_basic` causes the process to go to background and stall until typing `fg` on the 2nd invocation of `run_shell_cmd` with `cmd.sh -c ...`

Using `BASH_ENV` for the non-interactive use avoids this while still sourcing the env file.
To detect that the presence of any argument to `cmd.sh` is used as for interactive usage none would be present.

See https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html

### DIFF
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -273,13 +273,15 @@ def create_cmd_scripts(cmd_str, work_dir, env, tmpdir, out_file, err_file):
     # Make script that sets up bash shell with specified environment and working directory
     cmd_fp = os.path.join(tmpdir, 'cmd.sh')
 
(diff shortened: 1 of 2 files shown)


Flamefire:
@boegel Rebased. Can we get this in soon? This is very annoying when testing locally.

boegel (approved):
lgtm

Issue #5134, opened 2026-02-23
HierarchicalMNS build fails to resolve dependencies across split prefixes


FelixBangerter:
HierarchicalMNS build fails across split prefixes due to missing gateway modules (here: `GCCcore`/`pkgconf`) while flat MNS works.

## Summary

When building a base toolchain stack with EasyBuild and installing additional software on top of it in a separate prefix, the workflow works as expected with the default (flat) module naming scheme but fails with `HierarchicalMNS`.

With `HierarchicalMNS`, EasyBuild fails to locate required modules such as `GCCcore/14.3.0` and later `pkgconf/2.4.3`, even though they were successfully built in either the base or on-top prefix. The issue appears to be related to gateway modules and how `MODULEPATH` must be adjusted manually.

I would expect identical behavior independent of the module naming scheme. 

In the MWE below, I only use `GCCcore` and `libxc` as examples. I suspect this behavior is general for split prefix builds with a `HierarchicalMNS`.

## Environment

- Fresh Python virtual environment
- EasyBuild installed via `pip install easybuild`
- Lmod module system
- EasyBuild version: 5.2.1

### Setup

Create a clean environment and two separate prefixes:

```bash
python3 -m venv env
source env/bin/activate
pip3 install easybuild
mkdir -p $HOME/test/easybuild/{hierarchical,flat}/{base,on-top}
```

Build `GCC/14.3.0` in both naming schemes:

```bash
# Hierarchical MNS
nohup eb --prefix=$HOME/test/easybuild/hierarchical/base \
         --module-naming-scheme=HierarchicalMNS \
         -r GCC-14.3.0.eb > build.log 2>&1 &

# Flat MNS
nohup eb --prefix=$HOME/test/easybuild/flat/base \
         -r GCC-14.3.0.eb > build.log 2>&1 &
```

#### Dependency detection

```bash
module use $HOME/test/easybuild/flat/base/modules/all/
eb --missing --prefix=$HOME/test/easybuild/flat/on-top \
   -r libxc-7.0.0-GCC-14.3.0.eb
```

Already installed dependencies are correctly detected.

#### Build

```bash
eb --prefix=$HOME/test/easybuild/flat/on-top \
   -r libxc-7.0.0-GCC-14.3.0.eb
```

Result:

```text
[SUCCESS] libxc/7.0.0-GCC-14.3.0
```

#### Dependency detection

[...]

PR #5188, opened 2026-05-05
Use `assert(Not)Regex` instead of `assertTrue` or `assertFalse` in tests

Flamefire:
This PR is now split into smaller PRs (<= 5 modules, <= ~100 lines):

- #5240 
- #5241
- #5242
- #5243
- #5244
- #5245
- #5246
- #5247

This makes the tests shorter and hence easier to read, avoids
`re.compile` calls and manual failure message composing.
In some cases `assert_multi_regex` could be used to avoid the explicit loops.

While doing that I put the `re.compile` into the assertion instead of assigning to a variable and remove the call when there are no flags required as `assertRegex` accepts strings or compiled expressions

Examples:
```diff
-        regex = re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-[0-9]{8}\.[0-9]{6}\.log$'))
-        self.assertTrue(regex.match(res), "Pattern '%s' matches '%s'" % (regex.pattern, res))
+        self.assertRegex(res, os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-[0-9]{8}\.[0-9]{6}\.log$'))

-        regex = re.compile(r"^eb toy-0.0.eb --robot --debug -l", re.M)
-        self.assertTrue(regex.search(txt), "Pattern '%s' should be found in: %s" % (regex.pattern, txt))
+        self.assertRegex(txt, re.compile(r"^eb toy-0.0.eb --robot --debug -l", re.M))

-            fail_msg = "Pattern '%s' should be found in: %s" % (guarded_load_regex.pattern, modtxt)
-            self.assertTrue(guarded_load_regex.search(modtxt), fail_msg)
-            fail_msg = "Pattern '%s' should not be found in: %s" % (recursive_unload_regex.pattern, modtxt)
-            self.assertFalse(recursive_unload_regex.search(modtxt), fail_msg)
+            self.assertRegex(modtxt, guarded_load_regex)
+            self.assertNotRegex(modtxt, recursive_unload_regex)
```

Continuation of #4205 , #4170, #5134 to remove usage of `assertTrue` & `assertFalse` as for almost everything there is a better alternative 

Also regex assertions are over-used. In some places `assertIn` or even `assertEqual` could be used which is clearer and faster.
However the replacements done here were mostly regex-based with lot's of manual inspection to find good patterns, and finding those where `assertIn` would work is even more difficult.


boegel:
@Flamefire merge conflicts to resolve here...

Flamefire:
Rebased

boegel:
merge conflicts again...

As with #5223, I think it would help to break this into smaller PRs, which are easier to review/merge, and less likely to get blocked by merge conflicts

Flamefire:
This is a search&replace (using regex) which I think is better to do once, especially as it is only test code so mistakes would a) turn up in CI and b) are not critical to users. So I would not recommend multiple PRs but just do some quick spot checking.

If you still want separate PRs: How small do you want them? More PRs is more work but quicker to review.

boegel:
> This is a search&replace (using regex) which I think is better to do once, especially as it is only test code so mistakes would a) turn up in CI and b) are not critical to users. So I would not recommend multiple PRs but just do some quick spot checking.
> 
> If you still want separate PRs: How small do you want them? More PRs is more work but quicker to review.

Point taken. It's just trouble to get this reviewed, since it keeps "bitrotting" in between times we manage to take another look at it, and leading to merge conflicts (as is again the case now).

Naively I would say one PR per test module, but that's perhaps a step too far (that would be 25 tiny PRs).
How about PRs that change no more than say 100-ish lines, and only touch a handful of test modules (5-ish)?

Flamefire:
> How about PRs that change no more than say 100-ish lines, and only touch a handful of test modules (5-ish)?

Ok done, all PRs mentioned in the description of this one. I'd keep this open in case at some point someone likes to merge the remaining changes in one go

PR #5189, opened 2026-05-06
Allow overwriting the CUDA sanity check CLI arguments in an EasyConfig

Flamefire:
For some software it is not possible to e.g. include PTX code or has only PTX code. When that is known then the sanity check should not fail, but only for this easyconfig.
Hence allow setting/disabling those checks in an easyconfig.

The opposite is also possible:
- Disable it on command line
- Enable for specific easyconfigs, e.g. via hooks.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -67,6 +67,7 @@ from contextlib import contextmanager
 from datetime import datetime
 from string import ascii_letters
 from textwrap import indent
+from typing import Any
 
 import easybuild.tools.environment as env
 import easybuild.tools.toolchain as toolchain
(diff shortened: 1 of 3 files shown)


easybuild/framework/easyblock.py
```diff
@@ -3497,17 +3498,22 @@ def _sanity_check_step_multi_deps(self, *args, **kwargs):
         self.cfg['builddependencies'] = builddeps
         self.cfg.iterating = False
 
+    def get_from_easyconfig_or_build_option(self, option_name: str) -> Any:
+        """Return the value of this option if it is set (not None) in the easyconfig else from the CLI/config"""
+        res = self.cfg.get(option_name)
+        return res if res is not None else build_option(option_name)
```
ocaisa, line 3504: For CUDA compute capabilities what is set in the CLI/config takes precedence over what is in the easyconfig, but in this case, what is in the easyconfig is taking precedence over what is set in the CLI/config. This is likely to lead to confusion (I do not have a solution).

PR #5190, opened 2026-05-07
Replace raw `dict` by `CategorizedPaths`

Flamefire:
The current dict is pretty opaque and makes spelling mistakes easy. Avoid by using a typed, named tuple.

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -48,6 +48,7 @@ from datetime import datetime, timedelta
 from http import HTTPStatus
 from http.client import HTTPException
 from string import ascii_letters
+from typing import List, NamedTuple
 from urllib.request import HTTPError, URLError, urlopen
 
 from easybuild.base import fancylogger
@@ -88,6 +89,14 @@ try:
     from git import GitCommandError
 except ImportError as err:
     _log.warning("Failed to import 'git' Python module: %s", err)
+
+
+class CategorizedPaths(NamedTuple):
+    easyconfigs: List[str]
+    easyconfigs: List[str]
+    files_to_delete: List[str]
+    patch_files: List[str]
+    py_files: List[str]
 
 
 GITHUB_URL = 'https://github.com'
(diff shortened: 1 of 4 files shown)


easybuild/tools/github.py
```diff
@@ -90,6 +91,14 @@
     _log.warning("Failed to import 'git' Python module: %s", err)
 
 
+class CategorizedPaths(NamedTuple):
+    easyconfigs: List[str]
+    easyconfigs: List[str]
```
akesandgren: Why the duplication?

Flamefire: Oversight or merge conflict. Rebased to remove

akesandgren:
Should 
```
class CategorizedPaths(NamedTuple):
```
be in easybuild.tools.github or easybuild.framework.easyconfig.tools
you're importing it from both places right now and the def is in github.tools

the import from easybuild.gramework.easyconfig.tools is done in test/framework/easyconfig.py‎
the others use github.tools

Flamefire:
> the import from easybuild.gramework.easyconfig.tools is done in test/framework/easyconfig.py‎
> the others use github.tools

Technically the class belongs in `easybuild.framework.easyconfig.tools` next to `categorize_files_by_type` producing it.
However `github` is imported by that so it can't import the class from it -.-

I added a comment. If you prefer another solution I'd be fine with that too

akesandgren:
> > the import from easybuild.gramework.easyconfig.tools is done in test/framework/easyconfig.py‎
> > the others use github.tools
> 
> Technically the class belongs in `easybuild.framework.easyconfig.tools` next to `categorize_files_by_type` producing it. However `github` is imported by that so it can't import the class from it -.-
> 
> I added a comment. If you prefer another solution I'd be fine with that too

Ok, but shouldn't then test/framework/easyconfig.py import it from github.tools too?

akesandgren:
And we can't put it and the function to populate it in easybuild/tools/utilities.py ?
Just an idea, not sure if that would get approval by @boegel ...

Flamefire:
> Ok, but shouldn't then test/framework/easyconfig.py import it from github.tools too?

I'd rather let users import it from easyconfig together with the function, hence the comment there.

`easybuild/tools/utilities.py` would work but doesn't look like the right place. Maybe rather `filetools`?

PR #5193, opened 2026-05-21
Warn when using `get_cpu_arch_name` when archspec is not installed

Flamefire:
Some easyblocks rely on that so warn the user that this may cause issues.

### DIFF
--- a/easybuild/tools/systemtools.py
+++ b/easybuild/tools/systemtools.py
@@ -478,7 +478,8 @@ def get_cpu_arch_name(show_warning=True):
     if cpu_arch_name is None:
         if show_warning and not HAVE_ARCHSPEC:
             print_warning("Could not detect CPU architecture name which may affect functionality. "
-                          "Please install the 'archspec' Python package into the same Python environment as EasyBuild.")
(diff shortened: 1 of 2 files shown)


Micket (dismissed):
lgtm

easybuild/tools/systemtools.py
```diff
@@ -476,6 +476,9 @@ def get_cpu_arch_name():
             cpu_arch_name = str(res.name)
 
     if cpu_arch_name is None:
+        if show_warning and not HAVE_ARCHSPEC:
+            print_warning("Could not detect CPU architecture name which may affect functionality. "
+                          "Please install the 'archspec' Python package into the same Python environment as EasyBuild.")
```
smoors: maybe also add the warning to the log file?
```suggestion
                          "Please install the 'archspec' Python package into the same Python environment as EasyBuild.",
                          log=True)
```

Flamefire: That argument is an instance not a bool. But found one we have for this already, so done :)

smoors (approved):
lgtm

PR #5194, opened 2026-05-21
Cleanup duplication in `handle_include_easyblocks_from`

Flamefire:
Seen the linter warning that `included_easyblocks` may be used before assignment in the `check_included_multiple` function.
So ensure `included_easyblocks` is defined before the function.

Also the whole logging and checking code is duplicated in each branch (for the CLI options),
so move the logging from each option branch to a common function.
This can use the function that checks for duplicates as both are just logging the inclusion (the warning is also just a "log")
This makes the whole handling uniform:
- Extract filenames
- Report multiple inclusions
- Report inclusion

### DIFF
--- a/easybuild/tools/options.py
+++ b/easybuild/tools/options.py
@@ -1774,22 +1774,26 @@ def handle_include_easyblocks_from(options, log):
     """
     Handle --include-easyblocks-from-pr and --include-easyblocks-from-commit
     """
-    def check_included_multiple(included_easyblocks_from, source):
-        """Check whether easyblock is being included multiple times"""
-        included_multiple = included_easyblocks_from & included_easyblocks
-        if included_multiple:
-            warning_msg = "One or more easyblocks included from multiple locations: %s " \
-                          % ', '.join(included_multiple)
-            warning_msg += "(the one(s) from %s will be used)" % source
-            print_warning(warning_msg)
-
     if options.include_easyblocks_from_pr or options.include_easyblocks_from_commit:
         terse = build_option('terse')
-
         if options.include_easyblocks:
-            # check if you are including the same easyblock twice
-            included_paths = expand_glob_paths(options.include_easyblocks)
-            included_easyblocks = {os.path.basename(eb) for eb in included_paths}
+            included_easyblocks = {os.path.basename(eb) for eb in expand_glob_paths(options.include_easyblocks)}
+        else:
+            included_easyblocks = set()
+
+        def check_and_log_include(additional_easyblocks: list, source: str):
+            """Check whether easyblock is being included multiple times and log its inclusion"""
+            additional_easyblocks = {os.path.basename(eb) for eb in additional_easyblocks}
+            included_multiple = include_easyblocks & additional_easyblocks
+            if included_multiple:
+                warning_msg = "One or more easyblocks included from multiple locations: %s " \
+                            % ', '.join(included_multiple)
+                warning_msg += "(the one(s) from %s will be used)" % source
+                print_warning(warning_msg)
+            included_easyblocks.update(additional_easyblocks)
+            for easyblock in additional_easyblocks:
+                easyblock = os.path.basename(easyblock)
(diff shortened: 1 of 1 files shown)


easybuild/tools/options.py
```diff
@@ -1802,29 +1806,13 @@ def check_included_multiple(included_easyblocks_from, source):
 
             for easyblock_pr in easyblock_prs:
                 easyblocks_from_pr = fetch_easyblocks_from_pr(easyblock_pr)
-                included_from_pr = {os.path.basename(eb) for eb in easyblocks_from_pr}
-
-                if options.include_easyblocks:
-                    check_included_multiple(included_from_pr, "PR #%s" % easyblock_pr)
-                    included_easyblocks |= included_from_pr
```
akesandgren, line 1809: I assume this |= corresponds to the included_easyblocks.update(additional_easyblocks) in check_and_log_include
Note that the code below for easyblocks_from_commit does NOT have a corresponding |= operator
Which might be a bug in itself, but...

@boegel What do you think?

Flamefire: Below, i.e. the actual usage, we don't handle it specifically. I guess the last one "wins".
Not sure if we have a test that defines if the first or last one is used

easybuild/tools/options.py
```diff
@@ -1802,29 +1806,13 @@ def check_included_multiple(included_easyblocks_from, source):
...
+                check_and_log_include(easyblocks_from_pr, "PR #%s" % easyblock_pr)
                 include_easyblocks(options.tmpdir, easyblocks_from_pr)
 
         easyblock_commit = options.include_easyblocks_from_commit
         if easyblock_commit:
             easyblocks_from_commit = fetch_easyblocks_from_commit(easyblock_commit)
-            included_from_commit = {os.path.basename(eb) for eb in easyblocks_from_commit}
-
-            if options.include_easyblocks:
-                check_included_multiple(included_from_commit, "commit %s" % easyblock_commit)
```
akesandgren, line 1822: I meant here, there is no corresponding
included_easyblocks |= included_from_commit
in this block which may or may not be a pre-existing bug, which would in that case get solved by the 
```
included_easyblocks.update(additional_easyblocks)
```
code in check_and_log_include

akesandgren: That's what I wanted @boegel to look at...

Flamefire: I see what you mean now. As `from-pr` and `from-commit` are pretty much the same I'd say this was certainly a bug. However only a logging bug, behavior is not affected.

akesandgren: Hmmm, if I read this correctly, included_easyblocks is no longer used at all, only set.
And include_easyblocks (note diff in name here) is used but not set, so I think there is a name-typo on line 1787:  included_multiple = include_easyblocks & additional_easyblocks
that should? be 
included_multiple = included_easyblocks & additional_easyblocks

Flamefire: You are right. Not sure how that passed the tests. It worked at all by chance because there is a function with that (wrong) name. Fixed

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

Flamefire:
@akesandgren merge missed?

akesandgren:
No, i was just busy yesterday with Slurm upgrades :-)

Thus I didn't see this from my --merge-pr
```
ERROR: FAILED: {"message":"refusing to allow a Personal Access Token to create or update workflow `.github/workflows/container_tests.yml` without `workflow` scope","documentation_url":"https://docs.github.com/rest/pulls/pulls#merge-a-pull-request","status":"403"}
```

akesandgren:
Going in, thanks @Flamefire!

akesandgren:
And I still get that for some reason...

Flamefire:
I did update the workflow file in some of the PRs, basically including https://github.com/easybuilders/easybuild-framework/pull/5293 to make CI pass.
But not in this one.

Can you merge in the GUI here directly instead?

PR #5200, opened 2026-06-17
Add hardlink function in filetools

VRehnberg:
Not heavily tested. Tried to mimic behavior of symlink function, but unsure if behavior is same when link_path is a existing broken symlink (overwrite attempt will be made, but unsure if it will be successful).


akesandgren:
Note that hardlinks aren't available on every possible filesystem that could be used for installation...
And never possible across file system borders....
So in principle they should never be used, ever...

VRehnberg:
I see. On the other hand using symlinks for datasets with a lot files can have some real impact on performance.

I'll start with just adding it as an option in the EasyBlock then and move the logic into the easyblock. Possibly, would be something that you want to be able to configure globally for datasets.

PR #5201, opened 2026-06-17
Use bounded `git rev-list` for version detection

pavelToman:
EasyBuild only needs the current HEAD commit hash for version detection,
but git rev-list HEAD walks the full commit history and can be slow on
large repositories or high-latency filesystems.

Limit rev-list to one commit via git rev-list -n 1 HEAD. This preserves
the existing GitPython-based implementation while avoiding unnecessary
history traversal.

*this PR was created with help of ChatGPT5.5

easyblocks PR:
- https://github.com/easybuilders/easybuild-easyblocks/pull/4163

### DIFF
--- a/easybuild/tools/version.py
+++ b/easybuild/tools/version.py
@@ -64,7 +64,7 @@ def get_git_revision():
     try:
         path = os.path.dirname(__file__)
         gitrepo = Git(path)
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

PR #5208, opened 2026-06-19
split running test suite with Python 3.9 and Lmod 8.x with Lua/Tcl module syntax in two jobs, only use GitHub token with Lua module syntax

boegel:


bartoldeman (approved):
LGTM

PR #5213, opened 2026-06-26
prefer `bsdtar` for unpacking ISO files, keep `7z` as fallback

### LINKED ISSUE #5212: Allowing bsdtar as alternative for extracting iso files
`bsdtar` also supports easily extracting iso files.

I would say the nicest thing we can do is to detect `bsdtar` and default to that. 

```python
if not shutil.which('7z') and shutil.which('bsdtar'):
    EXTRACT_CMDS['iso'] = 'bsdtar xf %(filepath)s'
```
in `easybuild/tools/filetools.py`?

Micket:
Fixes #5212

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -175,12 +175,17 @@ EXTRACT_CMDS = {
     # zip file
     '.zip': "unzip -qq %(filepath)s",
     # iso file
-    '.iso': "7z x %(filepath)s",
+    '.iso': "bsdtar xf %(filepath)s",
     # tar.Z: using compress (LZW), but can be handled with gzip so use 'z'
     '.tar.z': "tar xzf %(filepath)s",
     # shell scripts don't need to be unpacked, just copy there
     '.sh': "cp -dR %(filepath)s .",
 }
+
+# 7z can also extract iso files, use it as a fallback
+if not shutil.which('bsdtar') and shutil.which('7z'):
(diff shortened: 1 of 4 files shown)


akesandgren:
bsdtar is available on Ubuntu in the libarchive-tools package

easybuild/tools/filetools.py
```diff
@@ -175,13 +175,18 @@ def _hashlib_md5():
...
-    '.iso': "7z x %(filepath)s",
+    '.iso': "bsdtar xf %(filepath)s",
     # tar.Z: using compress (LZW), but can be handled with gzip so use 'z'
     '.tar.z': "tar xzf %(filepath)s",
     # shell scripts don't need to be unpacked, just copy there
     '.sh': "cp -dR %(filepath)s .",
 }
 
+# 7z can also extract iso files, use it as a fallback
+if not shutil.which('bsdtar') and shutil.which('7z'):
```
```suggestion
if not shutil.which('bsdtar'):
```

I would just blindly fall back to `7z` if `bsdtar` is not available, so we always get the log message?

boegel:
@Micket For the tests, it would be nice if we could enhance `test_extract_cmd` to verify that the fallback to `7z` does *not* happen when there's a fake `bsdtar` command available.

For inspiration, see the fake `eb` command we inject in `test_find_eb_script`

Micket:
I'd squash this merge to avoid all my fumbling attempts to pass CI

boegel (approved):
lgtm

PR #5217, opened 2026-07-10
Fix `lfbf` toolchain description

Crivella:
Fix docstring in `lfbf` toolchain to also fix docs in https://docs.easybuild.io/api/easybuild/toolchains/lfbf/


Micket (dismissed):
lgtm

easybuild/toolchains/lfbf.py
Thyre: Still mentions `gfbf` and GCC

Thyre (changes requested):

branfosj (approved):

PR #5220, opened 2026-07-21
deprecate passing `extension` option to `sanity_check_step` & `sanity_check_load_module` methods in `EasyBlock` class

Flamefire:
This removes the superflous parameter which is always set to `self.extension` anyway.
It looks like an artifact from when that member didn't exist or was not set consistently.

I did this in a PR prior to the 5.3.0 release but took out to get that PR merged without this change.
Reason is to avoid potential mismatches, e.g. by forgetting to pass the parameter.
Ultimately I want to use `self.is_extension` to allow multiple extension to fail their sanity check step and having a "parallel" source of truth makes that harder.
The cause here is that the failing exts_filter allows to check other extensions but failing commands does not as that aborts directly. Easy to fix when relying on `self.is_extension`

I'll also removed it from `sanity_check_load_module` which is similarly simple but requires an [update to the easyblocks](https://github.com/easybuilders/easybuild-easyblocks/pull/4184) that use it.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -3444,6 +3444,16 @@ class EasyBlock:
 
     def _dispatch_sanity_check_step(self, *args, **kwargs):
         """Decide whether to run the dry-run or the real version of the sanity-check step"""
+        if 'extension' in kwargs:
+            extension = kwargs.pop('extension')
+            self.log.deprecated(
+                "Passing `extension` to `sanity_check_step` is no longer necessary "
+                f"(Easyblock: {self.__class__.__name__}).",
+                '6.0',
+            )
+            if extension != self.is_extension:
+                raise EasyBuildError('Unexpected value for `extension` argument. '
+                                     f'Should be: {self.is_extension}, got:  {extension}')
         if self.dry_run:
             self._sanity_check_step_dry_run(*args, **kwargs)
         else:
@@ -4316,6 +4326,8 @@ class EasyBlock:
         """
         Load module to prepare environment for sanity check
         """
+        if extension != self.is_extension:
+            raise RuntimeError(f"{self} {self.name} {extension}!={self.is_extension}")
 
         # skip loading of fake module when using --sanity-check-only, load real module instead
         if build_option('sanity_check_only') and not extension:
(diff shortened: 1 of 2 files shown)


easybuild/framework/easyblock.py
```diff
@@ -4310,23 +4320,34 @@ def _sanity_check_step_extensions(self):
...
+                raise EasyBuildError('Unexpected value for `extension` argument. '
+                                     f'Should be: {self.is_extension}, got:  {extension}')
+        del extension  # Avoid accidental use
+
+        if self.is_extension:
+            return self.fake_mod_data
 
         # skip loading of fake module when using --sanity-check-only, load real module instead
-        if build_option('sanity_check_only') and not extension:
+        if build_option('sanity_check_only'):
```
boegel, line 4343: shouldn't this still check `self.extension`?

```suggestion
        if build_option('sanity_check_only') and not self.extension:

Comment (boegel) [reply easybuild/framework/easyblock.py:4343]: ah, no, because of the inline `return` above...

File: easybuild/framework/easyblock.py [hunk 2]
```diff
@@ -4310,23 +4320,34 @@ def _sanity_check_step_extensions(self):
             self.sanity_check_fail_msgs.append(overall_fail_msg + ', '.join(x[0] for x in failed_exts))
             self.sanity_check_fail_msgs.extend(x[1] for x in failed_exts)
 
-    def sanity_check_load_module(self, extension=False, extra_modules=None):
+    def sanity_check_load_module(self, extension=None, extra_modules=None):
         """
         Load module to prepare environment for sanity check
+        extension - DEPRECATED
```
```suggestion
        :param extension: DEPRECATED: indicates whether this method is called for an extension
```

boegel (changes requested):

boegel:
Looks OK to me.

I dug into the history a bit:
- `self.is_extension` was first introduced in e68bc97467ddbca1ff80b5017c806de680f646ae (Jan'13, EasyBuild v1.11.0), but only in `ExtensionEasyBlock` class, so it could not be used generally;
- later, in 85f17b6567f97aed231387c2988b48d13193dc5c (Mar'19, EasyBuild v3.9.0), a change was made to ensure that `self.is_extension` is always defined in `EasyBlock` class, so since then we can always rely on it, but by then we had the `extension` option is various methods to effectively pass down the `is_extension` class parameter

So, makes sense to me to clean this up as proposed.

boegel (approved):
lgtm

easybuild/framework/easyblock.py
Flamefire: There is a similar case below and I agree this is not fully clear. I fixed the below case and added a short comment on the above return in https://github.com/easybuilders/easybuild-framework/pull/5227

PR #5221, opened 2026-07-21
Don't check for lib and bin paths in sanity check step for each extension

Flamefire:
We currently set defaults for paths checked in sanity check step if:
- `sanity_check_paths` is not given or `enhance_sanity_check`
- AND `custom_paths` from the easyblock is not given/empty

This is the case in general, so for a bundle with 100 extensions we will do ~ 200 redundant checks where each will walk the folders to determine if they are empty

This change might miss checking bin & lib now if the easyconfig:
- uses `exts_list`
- `enhance_sanity_check` is not set
- `sanity_check_paths` is used at the top level and does not include bin & lib or subfolders/files, like `bin/toy`, `lib/python`
- (parent) easyblock passes `custom_paths` without such folders/files
- all easyblocks used for extensions do not pass `custom_paths` or do not check for such folders/files

Reason: If e.g. the parent easyconfig/block does pass `custom_paths` then we will not check the defaults in the parent. And with the change we don't check the default in the extensions anymore and hence may not check it at all.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -4183,8 +4183,10 @@ class EasyBlock:
                 paths = {}
                 for key in path_keys_and_check:
                     paths.setdefault(key, [])
-                paths.update({SANITY_CHECK_PATHS_DIRS: ['bin', ('lib', 'lib64')]})
-                self.log.info("Using default sanity check paths: %s", paths)
+                # Default paths for extensions are handled in the parent easyconfig if desired
+                if not self.is_extension:
+                    paths.update({SANITY_CHECK_PATHS_DIRS: ['bin', ('lib', 'lib64')]})
+                    self.log.info("Using default sanity check paths: %s", paths)
 
             # if enhance_sanity_check is enabled *and* sanity_check_paths are specified in the easyconfig,
             # those paths are used to enhance the paths provided by the easyblock
@@ -4204,9 +4206,11 @@ class EasyBlock:
         # verify sanity_check_paths value: only known keys, correct value types, at least one non-empty value
         only_list_values = all(isinstance(x, list) for x in paths.values())
         only_empty_lists = all(not x for x in paths.values())
-        if sorted_keys != known_keys or not only_list_values or only_empty_lists:
+        if sorted_keys != known_keys or not only_list_values or (only_empty_lists and not self.is_extension):
             error_msg = "Incorrect format for sanity_check_paths: should (only) have %s keys, "
-            error_msg += "values should be lists (at least one non-empty)."
+            error_msg += "values should be lists"
+            if not self.is_extension:
+                error_msg += " (at least one non-empty)."
             raise EasyBuildError(error_msg % ', '.join("'%s'" % k for k in known_keys))
 
         # Resolve arch specific entries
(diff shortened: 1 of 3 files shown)


boegel:
> This change might miss checking bin & lib now if the easyconfig:
> 
>     * uses `exts_list`
> 
>     * `enhance_sanity_check` is not set
> 
>     * `sanity_check_paths` is used at the top level and does not include bin & lib or subfolders/files, like `bin/toy`, `lib/python`
> 
>     * (parent) easyblock passes `custom_paths` without such folders/files
> 
>     * all easyblocks used for extensions do not pass `custom_paths` or do not check for such folders/files
> 
> 
> Reason: If e.g. the parent easyconfig/block does pass `custom_paths` then we will not check the defaults in the parent. And with the change we don't check the default in the extensions anymore and hence may not check it at all.

I don't think that's a problem...
If `sanity_check_paths` is specified without enabling `enhance_sanity_check`, you should expect that you're in full control, i.e. only the paths listed will be checked.

I don't think people would assume that a check for a non-empty `bin` and/or` lib` directory would be done for specific extensions, generally speaking, so I don't expect anyone to actually (silently) rely on this bug .

Flamefire:
I've been discussing this with @Crivella on Slack for a while and we tried to find affected EasyConfigs but determining the exact conditions was a bit difficult which is why I opened this as a separate PR so someone else can verify with fresh eyes that a) that list of conditions is correct and complete and b) really doesn't apply to (at least) our easyconfigs.

> If `sanity_check_paths` is specified without enabling `enhance_sanity_check`, you should expect that you're in full control, i.e. only the paths listed will be checked.
> 
> I don't think people would assume that a check for a non-empty `bin` and/or` lib` directory would be done for specific extensions, generally speaking, so I don't expect anyone to actually (silently) rely on this bug .

My first suspicion would be something like PythonBundles: Assuming the packages check for bin & lib the bundle sanity_check_paths might only check for a specific path like some "data/foo"
So I guess once we are sure under which conditions this changes anything at all we can check if this applies to any EC

boegel:
> I've been discussing this with @Crivella on Slack for a while and we tried to find affected EasyConfigs but determining the exact conditions was a bit difficult which is why I opened this as a separate PR so someone else can verify with fresh eyes that a) that list of conditions is correct and complete and b) really doesn't apply to (at least) our easyconfigs.
> 
> > If `sanity_check_paths` is specified without enabling `enhance_sanity_check`, you should expect that you're in full control, i.e. only the paths listed will be checked.
> > I don't think people would assume that a check for a non-empty `bin` and/or` lib` directory would be done for specific extensions, generally speaking, so I don't expect anyone to actually (silently) rely on this bug .
> 
> My first suspicion would be something like PythonBundles: Assuming the packages check for bin & lib the bundle sanity_check_paths might only check for a specific path like some "data/foo" So I guess once we are sure under which conditions this changes anything at all we can check if this applies to any EC

I would be quite surprised if any easyconfigs would really make that kind of assumption.

And even if they would, nothing would actually break in practice: some sanity check path checks would just not happen anymore, but it wouldn't result in the installation failing. Theoretically, missing paths could be overlooked, but even that is quite unlikely to happen, I would say...

boegel (approved):
lgtm

PR #5223, opened 2026-07-30
Bring back `async_cmd_check` to check for parallel extension install completion and fix dry-run mode with parallel extensions

Flamefire:
We currently have [this documentation](https://github.com/easybuilders/easybuild-docs/blob/83797988b1a18c2b7f090c2c9a3d2ae2e27baf49/docs/installing-extensions-in-parallel.md?plain=1#L26-L29)
> and check  whether the command has completed via the `async_cmd_check` method.

It is used by rpackage which calls into the super method:
https://github.com/easybuilders/easybuild-easyblocks/blob/e53ee56b0ddce722fb4b37371096b0eabae6cf6f/easybuild/easyblocks/generic/rpackage.py#L335

However it was removed by 36febd873 so the rpackage would crash.

This PR brings it back working together with this callback and the expectations:
- When done set `self.async_cmd_output`
- return `self.async_cmd_task.done()`

This allows rpackage to work again and makes the implementation match the documentation.

Additionally I found that `--extended-dry-run` would crash:
```
                    if self.dry_run or ext.async_cmd_task.done():
                        res = ext.async_cmd_task.result()
```
In the dry-run mode the task is never started.
I resolved that by using a dummy result

And finally dry-run did query the dependencies but then didn't use them yielding misleading output.
The fix is straight forward but changes indentation, so the diff looks larger than it is.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2227,6 +2227,8 @@ class EasyBlock:
         after installing extension(s)
         """
         res = None
+        if self.dry_run:
+            return res  # No module file is written, so no changes. See _install_extensions_det_init_build_env
 
         self.log.debug(f"Checking whether contents of fake module file {fake_mod_file_path} have changed...")
 
@@ -2419,15 +2421,8 @@ class EasyBlock:
                     else:
                         pending_deps = []
 
-                if self.dry_run:
-                    tup = (ext.name, ext.version, ext.__class__.__name__)
-                    msg = "\n* installing extension %s %s using '%s' easyblock\n" % tup
-                    self.dry_run_msg(msg)
-                    running_exts.append(ext)
-
                 # if some of the required dependencies are not installed yet, requeue this extension
-                elif pending_deps:
-
+                if pending_deps:
                     # check whether all required dependency extensions are actually going to be installed;
                     # if not, we assume that they are provided by dependencies;
                     missing_deps = [x for x in required_deps if x not in all_ext_names]
@@ -2443,33 +2438,37 @@ class EasyBlock:
                         msg = f"Pending dependencies for {ext.name} after taking into account missing dependencies: "
                         self.log.debug(msg + ', '.join(pending_deps))
 
-                    if pending_deps:
-                        msg = f"Required dependencies not installed yet for extension {ext.name} ("
-                        msg += ', '.join(pending_deps)
-                        msg += "), adding it back to queue..."
-                        self.log.info(msg)
-                        # purposely adding extension back in the queue at Nth place rather than at the end,
-                        # since we assume that the required dependencies will be installed soon...
-                        exts_queue.insert(max_iter, ext)
-
                 # list of pending dependencies may be empty now after taking into account required extensions
                 # that are not being installed above, so extension may be ready to install
-                if not pending_deps:
+                if pending_deps:
+                    msg = f"Required dependencies not installed yet for extension {ext.name} ("
+                    msg += ', '.join(pending_deps)
+                    msg += "), adding it back to queue..."
+                    self.log.info(msg)
+                    # purposely adding extension back in the queue at Nth place rather than at the end,
+                    # since we assume that the required dependencies will be installed soon...
+                    exts_queue.insert(max_iter, ext)
+                else:
                     tup = (ext.name, ext.version or '')
                     print_msg("starting installation of extension %s %s..." % tup, silent=self.silent, log=self.log)
 
-                    if install and not self.dry_run:
-                        # restore build environment for this extension
-                        restore_env(build_env, log_changes=False)
+                    if install:
+                        if self.dry_run:
+                            self.dry_run_msg(f"\n* installing extension {ext.name} {ext.version} using "
+                                             f"'{ext.__class__.__name__}' easyblock\n")
+                            running_exts.append(ext)
+                        else:
+                            # restore build environment for this extension
+                            restore_env(build_env, log_changes=False)
 
-                        ext.install_extension_substep("pre_install_extension")
+                            ext.install_extension_substep("pre_install_extension")
 
-                        # note: current build environment is copied when install_extension_async is called
-                        ext.async_cmd_task = ext.install_extension_substep("install_extension_async", thread_pool)
-                        running_exts.append(ext)
+                            # note: current build environment is copied when install_extension_async is called
+                            ext.async_cmd_task = ext.install_extension_substep("install_extension_async", thread_pool)
+                            running_exts.append(ext)
 
-                        self.log.info(f"Started installation of extension {ext.name} in the background...")
-                        update_exts_progress_bar_helper(running_exts, 0)
+                            self.log.info(f"Started installation of extension {ext.name} in the background...")
+                            update_exts_progress_bar_helper(running_exts, 0)
 
             # check for extension installations that have completed
             installs_completed = False
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2471,7 +2471,7 @@ class EasyBlock:
             if running_exts:
                 self.log.info(f"Checking for completed extension installations ({len(running_exts)} running)...")
                 for ext in running_exts[:]:
-                    if self.dry_run or ext.async_cmd_task.done():
+                    if self.dry_run or ext.async_cmd_check():
                         installs_completed = True
                         res = ext.async_cmd_task.result()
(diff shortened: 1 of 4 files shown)


boegel:
> We currently have [this documentation](https://github.com/easybuilders/easybuild-docs/blob/83797988b1a18c2b7f090c2c9a3d2ae2e27baf49/docs/installing-extensions-in-parallel.md?plain=1#L26-L29)
> 
> > and check  whether the command has completed via the `async_cmd_check` method.
> 
> It is used by rpackage which calls into the super method: https://github.com/easybuilders/easybuild-easyblocks/blob/e53ee56b0ddce722fb4b37371096b0eabae6cf6f/easybuild/easyblocks/generic/rpackage.py#L335
> 
> However it was removed by [36febd8](https://github.com/easybuilders/easybuild-framework/commit/36febd8739ab37900577db8d0cdd13ea0e1f48fa) so the rpackage would crash.

It's true that `async_cmd_check` is still defined by the `RPackage` easyblock, but it serves no purpose anymore, I don't see how it could lead to a crash?

I don't see why we would reinstate `async_cmd_check`, which was removed when we switched to using `run_shell_cmd` for running shell commands asynchronously in:
- https://github.com/easybuilders/easybuild-framework/pull/4444

So instead, I think we should:
- remove the `async_cmd_check` method from `RPackage` easyblock;
- update our to correctly reflect what's required in order to be able to install extensions in parallel;

Flamefire:
> I don't see why we would reinstate `async_cmd_check`, which was removed when we switched to using `run_shell_cmd` for running shell commands asynchronously in:

The rpackage easyblock shows that this is a useful hook to check that the async command actually succeeded that simply checking the completion marker of the task doesn't allow: https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/rpackage.py#L335-L337

The base implementation of `async_cmd_check` does the simple: `if not self.async_cmd_task.done():`
But it also stores the results for use by the custom easyblocks, e.g. in the hook.

So I don't see this as a replacement but an enhancement and think we should keep it.
We could remove `self.async_cmd_output` as I guess custom easyblocks could get that via `self.async_cmd_task.result()` although it seems convenient

boegel:
@Flamefire Can you rebase this after merge of:
- https://github.com/easybuilders/easybuild-framework/pull/5248

easybuild/framework/easyblock.py
```diff
@@ -2440,49 +2439,57 @@ def update_exts_progress_bar_helper(running_exts, progress_size):
...
+                            self.log.info(f"Started installation of extension {ext.name} in the background...")
+                            update_exts_progress_bar_helper(running_exts, 0)
 
             # check for extension installations that have completed
             installs_completed = False
             if running_exts:
                 self.log.info(f"Checking for completed extension installations ({len(running_exts)} running)...")
                 for ext in running_exts[:]:
-                    if self.dry_run or ext.async_cmd_task.done():
+                    if self.dry_run or ext.async_cmd_check():
```
boegel, line 2478: I don't think replacing `async_cmd_task.done()` with `async_cmd_check()` makes sense...

Whether the command is still running (`async_cmd_task.done()` is `False`), or the check being done after an async command completed (which is what `async_cmd_check` is for), are two different things?

I would only run `async_cmd_check` if `async_cmd_task.done()` returns `True`, and it shouldn't be used in the condition here imho...

Imagine the scenario where the command was completed, but it failed.

Currently, it may not make a huge difference in practice (since only `RPackage` actually implements `async_cmd_check`), but going forward it may make a difference, depending on what `async_cmd_check` does exactly.

boegel:
> > I don't see why we would reinstate `async_cmd_check`, which was removed when we switched to using `run_shell_cmd` for running shell commands asynchronously in:
> 
> The rpackage easyblock shows that this is a useful hook to check that the async command actually succeeded that simply checking the completion marker of the task doesn't allow: https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/rpackage.py#L335-L337
> 
> The base implementation of `async_cmd_check` does the simple: `if not self.async_cmd_task.done():` But it also stores the results for use by the custom easyblocks, e.g. in the hook.
> 
> So I don't see this as a replacement but an enhancement and think we should keep it. We could remove `self.async_cmd_output` as I guess custom easyblocks could get that via `self.async_cmd_task.result()` although it seems convenient

OK, you're right, the `check_install_output` method that `RPackage` calls in `async_cmd_check` does useful things (but only when the installation of the extension *failed*, so it's pretty minor that it's being skipped currently), so it makes sense to re-instate it.

Our docs need a small fix anyways, since `run_async` is no more (it was replaced with `install_extension_async`), I'll open a PR for that aspect.

One thing we can maybe also fix while we're at it: there's a hidden assumption here that there's only one asynchronous command being launched at a time per `Extension` instance, but this is not enforced at all.
Maybe we should verify that `async_cmd_task` is `None` when we're assigning to it in `EasyBlock.install_extensions_parallel`, and throw an error when it's not `None`. Of course we also need to reset it to `None` after the completed command has been processed. `post_install_extension` is probably a good place for that...

easybuild/framework/easyblock.py
Flamefire: "async_cmd_check" = "check async command [status]"
In the default it checks if the async command has finished.
I can also imagine a subclass might run an additional command stored in its own task member. Then it could do `return super().async_cmd_check() && self.foo_task.done()`

We could support that by adding a `self.async_cmd_done()` instead of `async_cmd_check`, and then call `async_cmd_check` (ignored result) if the former passed
But I don't see the advantage of that: Users could cram their logic from async_cmd_check into async_cmd_done too as RPackage does now

Semantically it does make a difference of course. So how about `def async_cmd_done() -> bool` and `def check_async_cmd() -> None`?

I'd still keep the current solution: RPackage needs the stdout/stderr of the task, so it could as well do the checks itself.
Maybe even better: `def check_async_cmd() -> Optional[RunShellCmdResult]`, that returns None if the command is not yet completed? Then we can remove `self.async_cmd_output` and let the classes decide if they want stdout, stderr or both.

Flamefire:
> One thing we can maybe also fix while we're at it: there's a hidden assumption here that there's only one asynchronous command being launched at a time per `Extension` instance, but this is not enforced at all.
> Maybe we should verify that `async_cmd_task` is `None` when we're assigning to it in

It is: We assign to `async_cmd_task` the result of `install_extension_async` exactly once.
Subclasses shall not assign to it by means of the contract, so I'd say it's fine not to check for a violation.
As for multiple commands: That's what `async_cmd_check` could do, see above comment

test/framework/toy_build.py
```diff
@@ -2109,11 +2113,29 @@ def test_toy_exts_parallel(self):
             "toy/0.0-GCC-12.3.0",
         ]
         self.assertEqual(res, expected)
+        self.assertIn("Async toy extension build done", stdout)  # async_cmd_check of custom easyblock called
+
+        dry_run_args = args + [
+            '--extended-dry-run',
+            # Start clean, otherwise the existing dir is detected as a ghost directory
+            f'--installpath={tempfile.mkdtemp()}'
```
boegel: This causes trouble on macOS, result in test failing due to:
```
easybuild.tools.build_log.EasyBuildError: Installation of test.eb failed: Failed to create lock <very long path [Errno 63] File name too long: <very long path>
```

So I'll move this new part of the test to the end, where we can safely trash `self.test_installpath`...

easybuild/framework/easyblock.py
boegel: I like that idea, so I implemented it while reviewing this PR again:
- https://github.com/Flamefire/easybuild-framework/pull/10

companion PR to adjust `RPackage` accordingly (in https://github.com/easybuilders/easybuild-easyblocks/pull/3911/):
- https://github.com/Flamefire/easybuild-easyblocks/pull/10

test/framework/toy_build.py
boegel: fixed in 6b3a189

Flamefire:
> I like that idea, so I implemented it while reviewing this PR again:
> - https://github.com/Flamefire/easybuild-framework/pull/10

Makes sense, thanks!

boegel:
Tested with (commit 6b3a189) with `eb --parallel-extensions-install R-bundle-CRAN-2025.11-foss-2025b.eb --include-easyblocks rpackage.py` (commit `490a118` for easyblocks), works like a charm.

boegel (approved):
lgtm

PR #5227, opened 2026-08-07
Remove superflous check

Flamefire:
`self.is_extension` is already checked so cleanup condition and clarify in comments.

Followup to #5220


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #5228, opened 2026-08-07
Add `join_path_var` function to join list of paths into a colon-separated string value

Flamefire:
`os.pathsep.join` is often used but there might be empty entries.
Those should usually be omitted as they can have different semantics
than expected. E.g. `$PYTHONPATH` treats those as `$PWD`.

I ran into this because of a build failure in TensorFlow and though this method should best be in framework

### DIFF
--- a/test/framework/environment.py
+++ b/test/framework/environment.py
@@ -31,12 +31,23 @@ import os
 import sys
 from test.framework.utilities import EnhancedTestCase, TestLoaderFiltered, init_config
 from unittest import TextTestRunner
+from pathlib import Path
 
 import easybuild.tools.environment as env
 
 
 class EnvironmentTest(EnhancedTestCase):
     """ Testcase for run module """
+
+    def test_join_path_var(self):
+        """Test join_path_var function."""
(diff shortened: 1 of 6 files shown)


test/framework/environment.py
```diff
@@ -31,13 +31,24 @@
...
+from pathlib import Path
 
 import easybuild.tools.environment as env
 
 
 class EnvironmentTest(EnhancedTestCase):
     """ Testcase for run module """
 
+    def test_join_path_var(self):
+        """Test join_path_var function."""
```
boegel, line 43: I'd add one more trivial test case: when no empty paths are involved, `join_path_var` provides same result as `os.pathsep.join`:

```suggestion
        """Test join_path_var function."""
        # join_path_var is same as os.pathsep.join if there are no empty paths involved
        paths = ['/foo', '/bar', '/baz']
        self.assertEqual(env.join_path_var(paths), os.pathsep.join(paths))

```

boegel (changes requested):

boegel (approved):
lgtm

PR #5211, opened 2026-06-22
significantly reduce number of `module load` commands used to set up build environment

boegel:
- loading of modules when setting up toolchain build environment for toolchain itself (1) + dependencies (D) is not done in one go, so a single `module load` command rather than `1+D` `module load commands;
- build environment for installing extensions is determine *once*, and environment is restored right before installing an extension (without running `module` commands);
  - only implemented in `EasyBlock.install_extensions_sequential` currently;

TODO:
- [x] fix bug uncovered by @branfosj (re-regression of https://github.com/easybuilders/easybuild-framework/issues/4948);
- [x] also change `EasyBlock.install_extensions_parallel` to restore build environment for each extension installation;
- [x] fix broken tests;
- [x] look for further "easy" wins, like avoiding `module unload` commands

Essentially fixes #5052

PR #5230, opened 2026-08-09
copy `Toolchain` instance into `Extension` instances after reset & prepare

boegel:
During regression test for the EasyBuild v5.4.0, the installation of `SciPy-bundle-2023.07-gfbf-2023a.eb` failed with an error for the `numpy` extension:
```
An error was raised during test step: Time for 1000x1000 matrix dot product: 832 msec >= 500 msec => ERROR
```

This check, which is part of the test step of the custom easyblock for `numpy`, indicates that linking to the provided BLAS library didn't happen, and the (very slow) fallback library is used instead.

The problem is that the `numpy` easyblock uses `self.toolchain.get_variable` to determine the value of variables like `LIBLAPACK_MT`, and that the variables of the `Toolchain` instance used by the `numpy` extension are empty (because a `prepare` was never called on them, and hence `set_variables` was never called on them either).

The easiest fix for this is to copy the `Toolchain` instance into the `Extension` instances after a reset & prepare, as is done in `EasyBlock._install_extensions_det_init_build_env` and `EasyBlock._install_extensions_check_fake_mod_file`.

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2203,6 +2203,11 @@ class EasyBlock:
                                        rpath_include_dirs=self.rpath_include_dirs,
                                        rpath_wrappers_dir=self.rpath_wrappers_dir)
 
+                # copy Toolchain instance into Extension instances,
+                # to ensure they get the correct values for variables, etc.
+                for ext in self.ext_instances:
+                    ext.cfg._toolchain = copy.copy(self.toolchain)
+
                 # note: we must grab contents of fake module file within context of fake_module_environment
                 fake_mod_file_txt = read_file(fake_mod_file_path)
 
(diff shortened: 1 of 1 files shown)


ocaisa:
Was this an existing problem, or due to changes for this release? If it was an existing issue, the performance impact might be significant for downstream applications

ocaisa (approved):
LGTM

boegel:
> Was this an existing problem, or due to changes for this release? If it was an existing issue, the performance impact might be significant for downstream applications

This is a problem that was introduced very recently, in:
- #5211 

Copying the `Toolchain` instance is pretty fast (it's all in memory), so I doubt this would have significant performance impact? In the case of `R-bundle-CRAN`, we're talking about copying ~1k `Toolchain` instances, which are pretty small.

PR #5234, opened 2026-08-10
Raise an error if `toolchain.dependencies` is accessed before `prepare_step`

Flamefire:
Avoids silently using the uninitialized, empty list.

See https://github.com/easybuilders/easybuild-framework/pull/5182

### DIFF
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -254,6 +254,15 @@ class Toolchain:
                                  'usually in prepare_step')
         return self._dependencies
 
+    @dependencies.setter
+    def dependencies(self, deps):
+        self.log.deprecated(
+            "Setting toolchain.dependencies should only be done by the toolchain instance and "
+            "through self._dependencies.",
+            '6.0',
+        )
+        self._dependencies = deps
+
     @property
     def search_path_vars_headers(self):
         """Return list of environment variables used as search paths for headers"""
--- a/easybuild/tools/toolchain/toolchain.py
+++ b/easybuild/tools/toolchain/toolchain.py
@@ -57,6 +57,7 @@ import os
 import stat
 import sys
 import tempfile
+from typing import Any, Dict, List, Optional
 
 from easybuild.base import fancylogger
 from easybuild.tools.build_log import EasyBuildError, dry_run_msg, print_warning
@@ -196,7 +197,7 @@ class Toolchain:
         """
         self.base_init()
 
-        self.dependencies = []
+        self._dependencies: Optional[List[Dict[str, Any]]] = None
         self.toolchain_dep_mods = []
         self.cached_compilers = set()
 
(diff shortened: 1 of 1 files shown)


easybuild/tools/toolchain/toolchain.py
```diff
@@ -245,6 +246,23 @@ def __init__(self, name=None, version=None, mns=None, class_constants=None, tcde
                 self.mod_short_name = self.mns.det_short_module_name(tc_dict)
                 self.init_modpaths = self.mns.det_init_modulepaths(tc_dict)
 
+    @property
+    def dependencies(self) -> List[Dict[str, Any]]:
+        """Dependency info of the EasyConfig for which this toolchain was prepared"""
+        if self._dependencies is None:
+            raise EasyBuildError('toolchain.dependencies must not be accessed before it is prepared, '
+                                 'usually in prepare_steps')
```
boegel:
```suggestion
                                 'usually in prepare_step')

File: easybuild/tools/toolchain/toolchain.py [hunk 2]
```diff
@@ -245,6 +246,23 @@ def __init__(self, name=None, version=None, mns=None, class_constants=None, tcde
...
+        return self._dependencies
+
+    @dependencies.setter
+    def dependencies(self, deps):
+        self.log.deprecated(
+            "Setting toolchain.dependencies should only be done by the toolchain instance and "
```
boegel, line 260: Setting `toolchain.dependencies` directly is currently done in `Bundle` easyblock, so we need an associated change in there?

boegel (changes requested):

easybuild/tools/toolchain/toolchain.py
Flamefire: That is tricky... For extensions we return `self.master.toolchain`, i.e. there is only 1 toolchain
For components we instantiate a separate toolchain for each component. Not sure if this is required or just how it's done at the moment.
If they shared the toolchain instance we wouldn't need to access their dependencies.

But the toolchain is in `self.cfg._toolchain` as the chain is `self.toolchain -> self.cfg.toolchain -> self.cfg._toolchain` so not cleaner to set from `Bundle` than `comp.toolchain._dependencies`

In any case: `prepare` is never called for the component toolchain instances, so having separate instances seems to be a bad idea.

Wait: Bundle does have access to the components cfg:
```
         for _, comp in self.comp_instances: # <--- The `_` is `comp_cfg`
             comp.toolchain.dependencies = self.toolchain.dependencies
```
So we could do 
```
         for comp_cfg, _ in self.comp_instances:
             comp_cfg._toolchain = self.toolchain
```
Although that's not ideal either. Better would be in `__init__` although I'm not completely sure we can/should access `toolchain` there as that constructs the toolchain instance.

Other idea: Add `copy_dependencies(self, other_toolchain): self._dependencies = other_toolchain._dependencies` and use that in Bundle.

Or what do you think?

PR #5235, opened 2026-08-11
deprecate useless `EasyBlock.exts_all` class variable

Flamefire:
This is always equal to `self.exts` as skipping is done on `self.ext_instances`.


boegel (approved):
lgtm

PR #5236, opened 2026-08-11
Fix symlink handling in `dir_contains_files`

Flamefire:
When we check a directory that is a symlink to a folder with files or contains such a symlink but no files itself then this function would return False even though there are files reachable through that path.

Return True in this case now.

Fix also the handling of broken symlinks: `os.path.isfile(p)` returns True if `p` is a file or a symlink pointing to a file.
`os.walk` yields symlinks always as `files` so without checking for `isfile` there we would have different results for a folder that contains a broken symlink.
Uniformly return False now


boegel (approved):
lgtm

PR #5237, opened 2026-08-11
Enhance error message when creating symlink where a file is already

Flamefire:
Currently creating a symlink will check existing symlinks if they point to the same location.
This fails if the existing path is not a link: `readlink` raises `InvalidArgumentError`

As this is not user-friendly check the path for being a link first.


boegel (approved):
lgtm

PR #5238, opened 2026-08-12
Avoid crash when running extension installation in `--dry-run` mode

Flamefire:
In dry-run mode `_install_extensions_det_init_build_env` yields None values. `_install_extensions_check_fake_mod_file` will then try to read the module file which does not exist and its path is that `None` which crashes in `read_file`

Always treat the module file as unchanged in dry-run mode to avoid that.

Fix for regression since #5211

`--dry-run` for extensions is not well tested (if at all)  and has other issues. This just addresses the regression in EB 5.4.0
For further enhancements see #5223

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2221,6 +2221,8 @@ class EasyBlock:
         after installing extension(s)
         """
         res = None
+        if self.dry_run:
+            return res  # No module file is written, so no changes. See _install_extensions_det_init_build_env
 
         self.log.debug(f"Checking whether contents of fake module file {fake_mod_file_path} have changed...")
 


boegel:
@Flamefire Do you have an example to trigger the crash being fixed here? I couldn't really find a case that results in a crash (especially since with `-x` all crashes are essentially ignored and swept under the rug, except for a `WARNING: ignoring error` being printed).

You mention that `read_file` crashes in ` because it gets fed a `None` value, but that doesn't look right to me, since only the values for `build_env` and `fake_mod_file_txt` would be `None`, and `read_file` in `_install_extensions_check_fake_mod_file` is only on `fake_mod_file_path`...

I don't mind merging this, since it should do no harm, but I don't see the point if it doesn't actually fix any bug/regression?

Flamefire:
> Do you have an example to trigger the crash being fixed here? 

`eb --extended-dry-run --parallel-extensions-install CellChat-1.5.0-foss-2022a.eb`

However this runs into an earlier error fixed by #5223 

This isn't triggered in the sequential case because dry-run skips the install branch. Doesn't seem right though:
```
            if self.dry_run:
                self.dry_run_msg(msg)
            elif install:
				# actual install
```

I guess the dry-run check should be moved into the install check to be fully correct. But that's another issue, fixed in https://github.com/easybuilders/easybuild-framework/pull/5248

boegel:
This trivial change is also included in:
- #5223

So it'll get applied with the merge of that PR...

PR #5240, opened 2026-08-26
[1/8] Use `assert(Not)Regex` instead of `assertTrue` or `assertFalse` in tests

Flamefire:

### DIFF
--- a/test/framework/repository.py
+++ b/test/framework/repository.py
@@ -28,7 +28,6 @@ Unit tests for repository.py.
 @author: Toon Willems (Ghent University)
 """
 import os
-import re
 import shutil
 import sys
 import tempfile
@@ -88,7 +87,7 @@ class RepositoryTest(EnhancedTestCase):
             shutil.rmtree(repo.wc)
         except EasyBuildError as err:
             print("ignoring failed subtest in test_gitrepo, testing offline?")
-            self.assertTrue(re.search("pull in working copy .* went wrong", str(err)))
+            self.assertRegex(str(err, "pull in working copy .* went wrong"))
 
         # filepath
         tmpdir = tempfile.mkdtemp()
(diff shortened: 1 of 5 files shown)


test/framework/repository.py
```diff
@@ -88,7 +87,7 @@ def test_gitrepo(self):
             shutil.rmtree(repo.wc)
         except EasyBuildError as err:
             print("ignoring failed subtest in test_gitrepo, testing offline?")
-            self.assertTrue(re.search("pull in working copy .* went wrong", str(err)))
+            self.assertRegex(str(err, "pull in working copy .* went wrong"))
```
akesandgren: Isn't the () wrong here ? str(err), ... ?
Or even self.convert_exception_to_str(err) like it is in easybuild/base/testing.py‎

akesandgren: Also, in some cases the resulting message from the assert might be less informative using the default one, tough I don't known what the messages from assertRegex actually look like...

Flamefire: Yes, it would have aborted with a TypeError or so. Fixed.

As for the message, compare the old vs new one:
```
    self.assertTrue(re.search("pull in working copy .* went wrong", str(err)))
AssertionError: None is not true

    self.assertRegex(str(err), "pull in working copy .* went wrong")
AssertionError: Regex didn't match: 'pull in working copy .* went wrong' not found in 'this is some other error'
```

akesandgren: Yeah I meant the ones where an actual message was supplied to assertTrue/False, like the one in base/testing
```
self.assertTrue(regex.search(msg), "Pattern '%s' is found in '%s'" % (regex.pattern, msg))
```

Flamefire: Reason for the replacement is the much clearer look of `self.assertRegex` instead of `self.assertTrue` and removing the need to manually specify a message that requires some effort to mention the pattern and message again.

Also that manual message we had is more confusing to me:
```
    self.assertTrue(regex.search(msg), "Pattern '%s' is found in '%s'" % (regex.pattern, msg))
AssertionError: None is not true : Pattern 'pull in working copy .* went wrong' is found in 'this is some other error'
```

a) "None is not True" is not exactly useful here
b) I can't easily tell what the expectation is. The error says "x is found in y" when this is NOT the case

So I very much prefer to read "Regex didn't match:" in addition to avoiding the boilerplate code for the message

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #5241, opened 2026-08-26
[2/8] Use `assert(Not)Regex` instead of `assertTrue` or `assertFalse` in tests

Flamefire:

### DIFF
--- a/test/framework/config.py
+++ b/test/framework/config.py
@@ -598,33 +598,27 @@ class EasyBuildConfigTest(EnhancedTestCase):
         tmpdir = tempfile.gettempdir()
 
         res = get_log_filename('foo', '1.2.3')
-        regex = re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-[0-9]{8}\.[0-9]{6}\.log$'))
-        self.assertTrue(regex.match(res), "Pattern '%s' matches '%s'" % (regex.pattern, res))
+        self.assertRegex(res, re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-[0-9]{8}\.[0-9]{6}\.log$')))
 
         res = get_log_filename('foo', '1.2.3', date='19700101')
-        regex = re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.[0-9]{6}\.log$'))
-        self.assertTrue(regex.match(res), "Pattern '%s' matches '%s'" % (regex.pattern, res))
+        self.assertRegex(res, re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.[0-9]{6}\.log$')))
 
         res = get_log_filename('foo', '1.2.3', timestamp='094651')
-        regex = re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-[0-9]{8}\.094651\.log$'))
-        self.assertTrue(regex.match(res), "Pattern '%s' matches '%s'" % (regex.pattern, res))
+        self.assertRegex(res, re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-[0-9]{8}\.094651\.log$')))
 
         res = get_log_filename('foo', '1.2.3', date='19700101', timestamp='094651')
-        regex = re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.094651\.log$'))
-        self.assertTrue(regex.match(res), "Pattern '%s' matches '%s'" % (regex.pattern, res))
+        self.assertRegex(res, re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.094651\.log$')))
 
         # if log file already exists, numbers are added to the filename to obtain a new file path
         write_file(res, '')
         res = get_log_filename('foo', '1.2.3', date='19700101', timestamp='094651')
-        regex = re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.094651\.log\.1$'))
-        self.assertTrue(regex.match(res), "Pattern '%s' matches '%s'" % (regex.pattern, res))
+        self.assertRegex(res, re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.094651\.log\.1$')))
 
         # adding salt ensures a unique filename (pretty much)
         prev_log_filenames = []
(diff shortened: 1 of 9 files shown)


test/framework/config.py
```diff
@@ -598,33 +598,27 @@ def test_get_log_filename(self):
...
+        self.assertRegex(res, re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.094651\.log\.1$')))
 
         # adding salt ensures a unique filename (pretty much)
         prev_log_filenames = []
-        for i in range(10):
+        for _ in range(10):
             res = get_log_filename('foo', '1.2.3', date='19700101', timestamp='094651', add_salt=True)
-            regex = re.compile(os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.094651\.[a-zA-Z]{5}\.log$'))
-            self.assertTrue(regex.match(res), "Pattern '%s' matches '%s'" % (regex.pattern, res))
+            self.assertRegex(res, os.path.join(tmpdir, r'easybuild-foo-1\.2\.3-19700101\.094651\.[a-zA-Z]{5}\.log$'))
```
akesandgren, line 620: Why the lack of re.compile here compared to the others above?

akesandgren: Or perhaps the question should be why keep the re.compile above?

test/framework/module_generator.py
```diff
@@ -845,7 +843,7 @@ def test_module_extensions(self):
         else:
             pattern = r"\s*extensions\("
 
-        self.assertFalse(re.search(pattern, desc), "No extensions found in: %s" % desc)
+        self.assertNotRegex(pattern, desc, re)
```
akesandgren, line 848: Why this "re"?

test/framework/module_generator.py
```diff
@@ -856,9 +854,7 @@ def test_module_extensions(self):
         modgen = self.MODULE_GENERATOR_CLASS(eb)
         desc = modgen.get_description()
 
-        for pattern in patterns:
-            regex = re.compile(pattern, re.M)
-            self.assertFalse(regex.search(desc), "Pattern '%s' not found in: %s" % (regex.pattern, desc))
+        self.assert_multi_regex(patterns, desc, assert_true=False)
```
akesandgren: Hmmm didn't think of this in the 1/8 PR... the assert_true is a weird name, just "assert" would make more sense, at least to me. (or at least something else than assert_true...)

test/framework/config.py
Flamefire: `re.compile` was required when flags were used. E.g. when `^$` should match line begin/end. I removed the remaining ones that were no longer required now

test/framework/module_generator.py
Flamefire: I think this existed before. How about `assert_match`?

akesandgren: that feels better at least...

Flamefire: Done. Can you merge this soon so I can update the others before they fail post-merge?

test/framework/package.py
```diff
@@ -255,7 +255,7 @@ def test_package(self):
         self.assertTrue(os.path.isfile(pkgfile))
         pkgtxt = read_file(pkgfile)
         self.assertRegex(pkgtxt, r"""DESCRIPTION:.*`backticks'.*""")
-        self.assertRegex(pkgtxt, re.compile(r"""DESCRIPTION:.*\nand newlines""", re.MULTILINE))
+        self.assertRegex(pkgtxt, r"""DESCRIPTION:.*\nand newlines""")
```
akesandgren, line 258: Don't these two actually need MULTILINE? the first one does have a "$" in the middle of the regex at least and the second one includes newline in the pattern

Flamefire: The first one? `r"""DESCRIPTION:.*`backticks'.*"""` has no `$`

MULTILINE only affects `^` and `$` to match line start/end, so a `\n` always matches a newline

test/framework/package.py
```diff
@@ -232,16 +232,12 @@ def test_package(self):
...
 
         pkgtxt = read_file(pkgfile)
-        pkgtxt_regex = re.compile("STARTCONTENTS of installdir %s" % easyblock.installdir)
-        self.assertTrue(pkgtxt_regex.search(pkgtxt), "Pattern '%s' found in: %s" % (pkgtxt_regex.pattern, pkgtxt))
+        self.assertIn("STARTCONTENTS of installdir %s" % easyblock.installdir, pkgtxt)
 
-        no_logfiles_regex = re.compile(r'STARTCONTENTS.*\.(log|md)$.*ENDCONTENTS', re.DOTALL | re.MULTILINE)
-        res = no_logfiles_regex.search(pkgtxt)
-        self.assertFalse(res, "Pattern not '%s' found in: %s" % (no_logfiles_regex.pattern, pkgtxt))
+        self.assertNotRegex(pkgtxt, re.compile(r'STARTCONTENTS.*\.(log|md)$.*ENDCONTENTS', re.DOTALL | re.M))
```
akesandgren, line 240: Sorry, this is the one I meant that has $ in the middle

Flamefire: Ok, I had kept the `re.M` there.

Flamefire:
I pulled in another change to `assert_multi_regex` from one of the other PRs to avoid changing that multiple times.
It looks like for multiple regexes we usually want the multiline feature and we had used it as such

However IMO this is not a safe default: With `re.M` it may have *additional*  matches, which is ok for `assert_match=False` but for `assert_match=True` I'd say it would be better to be explicit when we need it.
Maybe we should have that as default: `multi_line = not assert_match`

test/framework/module_generator.py
```diff
@@ -847,7 +845,7 @@ def test_module_extensions(self):
         else:
             pattern = r"\s*extensions\("
 
-        self.assertFalse(re.search(pattern, desc), "No extensions found in: %s" % desc)
+        self.assertNotRegex(pattern, desc, re)
```
akesandgren: Still don't understand the "re" as third argument here. The third argument is the message string to be printed if you don't want the default.

Flamefire: Just a leftover. I thought I had fixed that, done now

akesandgren (dismissed):
LGTM

akesandgren (approved):
Looks even better now :-)

akesandgren:
Going in, thanks @Flamefire!

Flamefire:
@akesandgren Followup: #5285 as per my last comment.
Can you check that soon?

PR #5242, opened 2026-08-26
[3/8] Use `assert(Not)Regex` instead of `assertTrue` or `assertFalse` in tests

Flamefire:


test/framework/github.py
```diff
@@ -582,10 +582,10 @@ def test_github_fetch_latest_commit_sha(self):
             return
 
         sha = gh.fetch_latest_commit_sha('easybuild-framework', 'easybuilders', github_user=GITHUB_TEST_ACCOUNT)
-        self.assertTrue(re.match('^[0-9a-f]{40}$', sha))
+        self.assertRegex(sha, '^[0-9a-f]{40}$')
```
akesandgren, line 585: Would probably be better to make this a "constant", SHA_PATTERN or something

test/framework/parallelbuild.py
```diff
@@ -318,21 +314,17 @@ def test_submit_jobs(self):
             r' --testoutput=%\(output_dir\)s',
             r' --disable-job ',
         ]
-        for regex in regexs:
-            regex = re.compile(regex)
-            self.assertTrue(regex.search(cmd), "Pattern '%s' found in: %s" % (regex.pattern, cmd))
+        self.assert_multi_regex(regexs, cmd)
```
akesandgren, line 317: multi_line = False here? probably doesn't matter but...

test/framework/parallelbuild.py
```diff
@@ -318,21 +314,17 @@ def test_submit_jobs(self):
...
 
         # these patterns should NOT be found, these options get filtered out
         # (self.test_prefix was argument to --robot)
-        for regex in ['--job', '--job-cores', '--try-toolchain', '--robot=[ =]', self.test_prefix + ' ']:
-            regex = re.compile(regex)
-            self.assertFalse(regex.search(cmd), "Pattern '%s' should *not* be found in: %s" % (regex.pattern, cmd))
+        self.assert_multi_regex(['--job', '--job-cores', '--try-toolchain', '--robot=[ =]', self.test_prefix + ' '],
+                                cmd, assert_true=False)
```
akesandgren, line 322: and possibly here too (multi_line=False)

Flamefire: Depends. In the previous PR I had the idea of making the defaults different:
- multiline for non-matches (`assert_true=False`)
- non-M for matches 

This way we fail if wrongly not specifying the param but cannot unexpectedly succeed

Then possibly `self.assertMultiRegex` and `self.assertNotMultiRegex` as separate methods is better mirroring the standard methods.

Flamefire: I created #5285 with this idea so we stay on the safe side. Can you merge that first please? Then I can update the others

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #5243, opened 2026-08-26
[4/8] Use `assert(Not)Regex` instead of `assertTrue` or `assertFalse`  in tests

Flamefire:


Flamefire:
@boegel I just had to fix a test:
`self.assert_multi_regex` automatically considers the passed patterns as multi-line enabled. But the test intended to check for whitespace at the beginning and end of the whole string only.

This begs the question if the default of `multi_line=True` is correct: It works (now) for all checks, most are simple patterns not affected by that setting anyway. But it seems the better default would be `False` matching the semantics of `assert(Not)Regex`
But I simply copied the existing `_assert_regexs` function from `test/framework/options.py` which had this default.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #5244, opened 2026-08-26
[5/8] Use `assert(Not)Regex` instead of `assertTrue` or `assertFalse`  in tests

Flamefire:


test/framework/easyconfig.py
```diff
@@ -2911,10 +2909,7 @@ def test_dump_template(self):
             r"\('ext1', '1\.0\.0'\),",
             r"sanity_check_paths = {\n    'files': \['files/%\(namelower\)s/foobar', 'files/x-test'\]",
         ]
-
-        for pattern in patterns:
-            regex = re.compile(pattern, re.M)
-            self.assertTrue(regex.search(ectxt), "Pattern '%s' found in: %s" % (regex.pattern, ectxt))
+        self.assertMultiRegex(patterns, ectxt)
```
akesandgren, line 2912: Shouldn't this one also have multi_line=True? since the re.compile does

test/framework/easyconfig.py
```diff
@@ -2993,9 +2988,7 @@ def test_dump_comments(self):
             r"    'files': \['files/foobar'\],  # comment on files",
             r"    'dirs': \[\],",
         ]
-        for pattern in patterns:
-            regex = re.compile(pattern, re.M)
-            self.assertTrue(regex.search(ectxt), "Pattern '%s' found in: %s" % (regex.pattern, ectxt))
+        self.assertMultiRegex(patterns, ectxt)
```
akesandgren, line 2991: Same question here

test/framework/easyconfig.py
```diff
@@ -5374,8 +5349,7 @@ def test_easyconfigs_caches(self):
             self.eb_main(args, do_build=True, testing=False, raise_error=True, clear_caches=False)
             stdout = self.get_stdout()
 
-        regex = re.compile(r"libtoy/0\.0 is already installed", re.M)
-        self.assertTrue(regex.search(stdout), "Pattern '%s' should be found in: %s" % (regex.pattern, stdout))
+        self.assertRegex(stdout, r"libtoy/0\.0 is already installed")
```
akesandgren: multi_line here?

test/framework/easyconfig.py
Flamefire: Not required: None of the regexs use it `^$`. So the original one was C&P from somewhere else

test/framework/easyconfig.py
Flamefire: Same: `^$` not used here.
Good thing with the new defaults: It will fail if the parameter was required. So it passing is a good indication that this is correct.

test/framework/easyconfig.py
Flamefire: Actually this just checks a string. So even the regex is not required.
Replaced by `assertIn`

akesandgren: Did you intend to add this change to assertIn ?

akesandgren:
Hmmm, you might need to keep assert_multi_regex until it's all gone.... tests complaining about it missing

test/framework/easyconfig.py
Flamefire: I did. Not sure what happened to that change. Done now and also fixed similar ones (using AI, verified the diff) as it makes it easier to read without the additional backslashing

Flamefire:
> Hmmm, you might need to keep assert_multi_regex until it's all gone.... tests complaining about it missing

There was only one I likely didn't catch before. Fixed

akesandgren:
test/framework/toy_build.py line 2176 looks like it needs multi_line=true

Flamefire:
True, missed the trailing `$`. Amended

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #5246, opened 2026-08-26
[7/8] Use `assert(Not)Regex` instead of `assertTrue` or `assertFalse` in tests

Flamefire:


test/framework/options.py
```diff
@@ -329,8 +321,8 @@ def test_skip(self):
         # make sure that sanity check is *NOT* skipped under --skip
         test_ec = os.path.join(self.test_prefix, 'test.eb')
         test_ec_txt = read_file(toy_ec)
-        regex = re.compile(r"sanity_check_paths = \{(.|\n)*\}", re.M)
-        test_ec_txt = regex.sub("sanity_check_paths = {'files': ['bin/nosuchfile'], 'dirs': []}", test_ec_txt)
+        test_ec_txt = re.sub(r"sanity_check_paths = \{(.|\n)*\}",
+                             "sanity_check_paths = {'files': ['bin/nosuchfile'], 'dirs': []}", test_ec_txt)
```
akesandgren, line 325: compare this to the similar one on 375-381 below (test_skipsteps), that one is easier to read

test/framework/options.py
```diff
@@ -1392,8 +1376,7 @@ def check_copied_files():
         self.assertEqual(len(os.listdir(os.getcwd())), 0)
         args = ['--copy-ec', 'toy-0.0.eb']
         stdout = self.mocked_main(args)
-        regex = re.compile('.*/toy-0.0.eb copied to .*/%s' % os.path.basename(test_working_dir))
-        self.assertTrue(regex.match(stdout), "Pattern '%s' found in: %s" % (regex.pattern, stdout))
+        self.assertRegex(stdout, '^.*/toy-0.0.eb copied to .*/%s$' % os.path.basename(test_working_dir))
```
akesandgren, line 1379: There is no ^/$ in the original regex, so why the addition?

test/framework/options.py
```diff
@@ -1979,8 +1958,9 @@ def test_dry_run_hierarchical(self):
             ("gzip-1.5-foss-2018a.eb", "MPI/GCC/6.4.0-2.28/OpenMPI/2.1.2", "gzip/1.5", ' '),
         ]
         for ec, mod_subdir, mod_name, mark in ecs_mods:
-            regex = re.compile(r"^ \* \[%s\] \S+%s \(module: %s \| %s\)$" % (mark, ec, mod_subdir, mod_name), re.M)
-            self.assertRegex(outtxt, regex)
+            self.assertRegex(outtxt,
+                             re.compile(r"^ \* \[%s\] \S+%s \(module: %s \| %s\)$" % (mark, ec, mod_subdir, mod_name),
+                                        re.M))
```
akesandgren: I think it's cleaner to keep this as a two-liner for readability, same below

test/framework/options.py
```diff
@@ -2170,8 +2149,8 @@ def test_github_from_pr_listed_ecs(self):
             ]
             for path_prefix, module in modules:
                 ec_fn = "%s.eb" % '-'.join(module.split('/'))
-                regex = re.compile(r"^ \* \[.\] %s.*%s \(module: %s\)$" % (path_prefix, ec_fn, module), re.M)
-                self.assertRegex(outtxt, regex)
+                self.assertRegex(outtxt,
+                                 re.compile(r"^ \* \[.\] %s.*%s \(module: %s\)$" % (path_prefix, ec_fn, module), re.M))
```
akesandgren: Same here, if you need to split the command into two lines, why  not just keep the original?

test/framework/options.py
```diff
@@ -7040,8 +6981,9 @@ def test_cuda_compute_capabilities(self):
         args = ['--cuda-compute-capabilities=3.5,6.2,7.0,9.0a,10.0f', '--show-config']
         txt, _ = self._run_mock_eb(args, do_build=True, raise_error=True, testing=False, strip=True)
 
-        regex = re.compile(r"^cuda-compute-capabilities\s*\(C\)\s*=\s*3\.5, 6\.2, 7\.0, 9\.0a, 10\.0f$", re.M)
-        self.assertRegex(txt, regex)
+        self.assertRegex(txt,
+                         re.compile(r"^cuda-compute-capabilities\s*\(C\)\s*=\s*3\.5, 6\.2, 7\.0, 9\.0a, 10\.0f$",
+                                    re.M))
```
akesandgren: Same here, i suggest keeping regex = ...  and using that in assertRegex for readability

test/framework/options.py
Flamefire: Is it? I find them very similar and require less context. With the previous version you had to jump from the 2nd line up to see where `regex` came from and then down to see if `regex` is used anywhere else.

Maybe it helps to split it further:
```python
test_ec_txt = re.sub(r"sanity_check_paths = \{(.|\n)*\}",
                     "sanity_check_paths = {'files': ['bin/nosuchfile'], 'dirs': []}",
                     test_ec_txt)
```

test/framework/options.py
Flamefire: Because `re.match` is different to `re.search` used by `assertRegex`:
The former matches at the beginning of the string. Hence the addition means: Match the whole string

Flamefire:
@akesandgren I reverted some of the changes combining `re.compile` and `assertRegex`

The initial translation was done by regex replacements looking for patterns like `regex = re.compile.*\nself.assertTrue(regex.search(foo)` and then cleanup

test/framework/options.py
```diff
@@ -220,14 +219,11 @@ def test_info(self):
             with self.mocked_stdout_stderr():
                 outtxt = self.eb_main(args)
 
-            error_tmpl = "%s log messages are included when using %s ( out: %s)"
             for log_msg_type in ['INFO', 'ERROR']:
-                res = re.search(' %s ' % log_msg_type, outtxt)
-                self.assertTrue(res, error_tmpl % (log_msg_type, info_arg, outtxt))
+                self.assertIn(log_msg_type, outtxt)
```
akesandgren: Didn't notice this last round, you're actually changing what is being checked.
Previously it was checking for SPACExxxSPACE, you're just checking for xxx.

test/framework/options.py
```diff
@@ -220,14 +219,11 @@ def test_info(self):
...
 
             for log_msg_type in ['DEBUG']:
-                res = re.search(' %s ' % log_msg_type, outtxt)
-                self.assertTrue(not res, "%s log messages are *not* included when using %s" % (log_msg_type, info_arg))
+                self.assertNotIn(log_msg_type, outtxt)
```
akesandgren: Same here and the next two assertIn's

test/framework/options.py
akesandgren: yeah, that's easier

Flamefire: Adjusted and the other uses in this file too for consistency.

test/framework/options.py
Flamefire: You are right, missed that completely. The `assertIn` is correct, just fixed the string and with a f-string it looks much easier to read too

test/framework/options.py
```diff
@@ -6610,9 +6612,8 @@ def test_enforce_checksums(self):
 
         # get rid of checksums for extensions, should result in different error message
         # because of missing checksum for source of 'bar' extension
-        regex = re.compile("^.*'checksums':.*$", re.M)
-        test_ec_txt = regex.sub('', read_file(test_ec))
-        self.assertNotIn("'checksums':", test_ec_txt)
+        test_ec_txt = re.sub("^.*'checksums':.*$", '', read_file(test_ec))
+        self.assertNotIn("'checksums':", test_ec_txt, re.M)
```
akesandgren: Does assertNotIn really have re.M option? not according to the page I'm looking at...

Flamefire: That was supposed to be in the previous line. Mispasted 🙈

akesandgren:
test/framework/options.py line 6616, sub fails to remove the 'checksum': part
test/framework/options.py line 3205 (not part of this changeset, but those/that multiregex might need multi_line

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

Flamefire:
> test/framework/options.py line 3205 (not part of this changeset, but those/that multiregex might need multi_line

From quick look at usage: They could actually be replaced by assertIn/assertNotIn. Doesn't seem to use regexes, just strings.
Made a follow-up renaming the variables there that should have used snake_case: https://github.com/easybuilders/easybuild-framework/pull/5290

PR #5247, opened 2026-08-26
[8/8] Use `assert(Not)Regex` instead of `assertTrue` or `assertFalse` in tests

Flamefire:


akesandgren:
Now there is a conflict here

Flamefire:
Resolved

test/framework/toy_build.py
```diff
@@ -856,12 +847,10 @@ def test_toy_group_check(self):
                         r'    error "%s[^"]*"' % error_msg_pattern,
                         r'\}$',
                     ])
-                    regex = re.compile(pattern, re.M)
-                    self.assertTrue(regex.search(outtxt), "Pattern '%s' found in: %s" % (regex.pattern, toy_mod_txt))
+                    self.assertRegex(outtxt, re.compile(pattern, re.M))
```
akesandgren: Since this is supposed to check for code in the generated module file, this was likely originally wrong to look in outtxt and should look in toy_mod_txt, just like you do in the Lua section on line 871 ?

test/framework/toy_build.py
```diff
@@ -1123,12 +1107,10 @@ def test_toy_hierarchical_subdir_user_modules(self):
...
         # No math libs in original toolchain, --try-toolchain is too clever to upgrade it beyond necessary
         for modname in ['FFTW', 'OpenBLAS', 'ScaLAPACK']:
-            regex = re.compile('load.*' + modname, re.M)
-            self.assertFalse(regex.search(toy_modtxt), "Pattern '%s' not found in: %s" % (regex.pattern, toy_modtxt))
+            self.assertNotRegex(toy_modtxt, re.compile('load.*' + modname, re.M))
 
         for modname in ['GCC', 'OpenMPI']:
-            regex = re.compile('load.*' + modname, re.M)
-            self.assertFalse(regex.search(toy_modtxt), "Pattern '%s' not found in: %s" % (regex.pattern, toy_modtxt))
+            self.assertNotRegex(toy_modtxt, re.compile('load.*' + modname, re.M))
```
akesandgren: Merge into a single loop?

test/framework/toy_build.py
```diff
@@ -3975,16 +3917,14 @@ def check_toy_load(depends_on=False):
             "This module is compatible with the following modules, one of each line is required:",
             "* GCC/4.6.3, GCC/7.3.0-2.30",
         ])
-        error_msg_descr = "Pattern '%s' should be found in: %s" % (expected_descr_no_default, toy_mod_txt)
-        self.assertIn(expected_descr_no_default, toy_mod_txt, error_msg_descr)
+        self.assertIn(expected_descr_no_default, toy_mod_txt,)
```
akesandgren: Trailing ","

test/framework/toy_build.py
Flamefire: Yes. Fixed by unifying the checks

test/framework/toy_build.py
```diff
@@ -829,49 +828,38 @@ def test_toy_group_check(self):
...
-                    r'    LmodError\("%s[^"]*"\)' % error_msg_pattern,
+                    r'^if not \( userInGroup\("{group_name}"\) \) then',
+                    r'    LmodError\("{error_msg}[^"]*"\)',
                     r'end$',
                 ])
-                self.assertRegex(toy_mod_txt, re.compile(pattern, re.M))
             else:
                 self.fail("Unknown module syntax: %s" % get_module_syntax())
 
+            toy_mod = os.path.join(self.test_installpath, 'modules', 'all', 'toy', '0.0.lua')
```
akesandgren: Note that tcl module files don't end with .lua

akesandgren:
I have no clue as to why that test fails

Flamefire:
> I have no clue as to why that test fails

Found: `format` needs braces escaped which is bad for understanding the pattern. So I reordered it a bit to be more like the original but still less repetition

PR #5248, opened 2026-08-26
Fix dry-run message in extension installation

Flamefire:
Extensions are only installed if `install = True` is passed. The dry-run message ignored this value printing the installation message even when it wouldn't actually be done.

Move inside the condition


boegel (approved):
lgtm

PR #5249, opened 2026-08-27
[1/6] Use constants for commonly used test easyconfig folders, files and content

Flamefire:
Followup to:
- https://github.com/easybuilders/easybuild-framework/pull/5050

Split from:
- https://github.com/easybuilders/easybuild-framework/pull/5203

Requires:
- https://github.com/easybuilders/easybuild-framework/pull/5271


easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -2327,7 +2327,7 @@ def process_easyconfig(path, build_specs=None, validate=True, parse_only=False,
     # only cache when no build specifications are involved (since those can't be part of a dict key)
     cache_key = None
     if not build_specs:
-        cache_key = (path, validate, hidden, parse_only)
```
boegel, line 2330: @Flamefire Can you clarify why this change is necessary? What problem does it fix exactly?

I'm reluctant to let adjacent changes sneaking in...

What if `path` is `None` for example?

Flamefire: `path` might be a `pathlib.Path` and for the cache we do want `"/tmp/foo.eb" == Path("/tmp/foo.eb")`
Can't fully remember but it might even error when comparing or hashing `Paths`

If `path` is `None` it will be stored as `"None"` which could now clash with an easyconfig stored at that (relative) path. I'd say that means never

boegel: To be safe we could use `str(path) if path else None` at first part of cache key?

Flamefire: Well a None path is not a valid input anyway, so I don't think we need to handle it.

Just checked again: It cannot be None else it would fail earlier already in `retrieve_blocks_in_spec` at: https://github.com/easybuilders/easybuild-framework/blob/a08e2e6aedb41cd2bb09828d69bc086293b0385e/easybuild/framework/easyconfig/format/one.py#L540

boegel:
@Flamefire needs rebase after merge of https://github.com/easybuilders/easybuild-framework/pull/5255

test/framework/config.py
```diff
@@ -264,8 +265,7 @@ def test_generaloption_config_file(self):
         tmpdir = tempfile.mkdtemp(prefix='easybuild-easyconfigs-pkg-install-path')
         mkdir(os.path.join(tmpdir, 'easybuild'), parents=True)
 
-        test_ecs_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs')
-        copy_dir(test_ecs_dir, os.path.join(tmpdir, 'easybuild', 'easyconfigs'))
+        copy_dir(os.path.join(TEST_ECS_DIR), os.path.join(tmpdir, 'easybuild', 'easyconfigs'))
```
boegel: we don't need the first `os.path.join` at all?

```suggestion
        copy_dir(TEST_ECS_DIR, os.path.join(tmpdir, 'easybuild', 'easyconfigs'))

File: test/framework/config.py [hunk 3]
```diff
@@ -498,9 +498,8 @@ def test_flex_robot_paths(self):
         # to check whether easyconfigs install path is auto-included in robot path
         tmpdir = tempfile.mkdtemp(prefix='easybuild-easyconfigs-pkg-install-path')
         mkdir(os.path.join(tmpdir, 'easybuild'), parents=True)
-        test_ecs_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs')
         tmp_ecs_dir = os.path.join(tmpdir, 'easybuild', 'easyconfigs')
-        copy_dir(test_ecs_path, tmp_ecs_dir)
+        copy_dir(os.path.join(TEST_ECS_DIR), tmp_ecs_dir)
```
boegel:
```suggestion
        copy_dir(TEST_ECS_DIR, tmp_ecs_dir)
```

test/framework/lib.py
```diff
@@ -125,7 +126,7 @@ def test_modules_tool(self):
 
         self.configure()
 
-        test_mods_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'modules')
+        test_mods_path = os.path.join(TEST_MODULES_DIR)
 
         modtool = modules_tool()
         modtool.use(test_mods_path)
```
boegel:
```suggestion
        modtool = modules_tool()
        modtool.use(TEST_MODULES_DIR)

File: test/framework/style.py [hunk 5]
```diff
@@ -53,8 +53,7 @@ def test_style_conformance(self):
             return
 
         # all available easyconfig files
-        test_easyconfigs_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'test_ecs')
-        specs = glob.glob('%s/*.eb' % test_easyconfigs_path)
+        specs = glob.glob('%s/*.eb' % TEST_ECS_DIR)
```
boegel:
```suggestion
        specs = glob.glob(f'{TEST_ECS_DIR}/*.eb')
```

boegel (changes requested):

test/framework/modulestool.py
```diff
@@ -164,7 +165,7 @@ def test_lmod_specific(self):
             os.environ['PATH'] = join_path_var(new_paths)
 
             # make sure $MODULEPATH contains path that provides some modules
-            os.environ['MODULEPATH'] = os.path.abspath(os.path.join(os.path.dirname(__file__), 'modules'))
+            os.environ['MODULEPATH'] = os.path.abspath(TEST_MODULES_DIR)
```
akesandgren: Either this one can be just
```
os.environ['MODULEPATH'] = TEST_MODULES_DIR
```
or the modtool.use(TEST_MODULES_DIR) in lib.py needs an abspath around it

test/framework/config.py
```diff
@@ -263,8 +264,7 @@ def test_generaloption_config_file(self):
         tmpdir = tempfile.mkdtemp(prefix='easybuild-easyconfigs-pkg-install-path')
         mkdir(os.path.join(tmpdir, 'easybuild'), parents=True)
 
-        test_ecs_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs')
-        copy_dir(test_ecs_dir, os.path.join(tmpdir, 'easybuild', 'easyconfigs'))
+        copy_dir(TEST_ECS_DIR, os.path.join(tmpdir, 'easybuild', 'easyconfigs'))
```
akesandgren, line 267: These are not the same.
TEST_ECS_DIR is
```
TEST_ECS_DIR = TEST_DIR / 'easyconfigs' / 'test_ecs'
```
while
```
test_ecs_dir = TEST_DIR / 'easyconfigs' 
```
which means that test/framework/easyconfigs/v1.0 and v2.0 won't get copied

Same problem below.

akesandgren: Not sure that matters but thought it worthwhile to point out

Flamefire: IIRC I've seen that and considered it confusion about having multiple easyconfig folders in the test folder. I think the v1/v2 are only used in a specific test and all others do mean to use test_ecs.

test/framework/modulestool.py
Flamefire: It's based on `__file__` so abspath is not required

test/framework/style.py
```diff
@@ -53,8 +53,7 @@ def test_style_conformance(self):
             return
 
         # all available easyconfig files
-        test_easyconfigs_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'test_ecs')
-        specs = glob.glob('%s/*.eb' % test_easyconfigs_path)
+        specs = glob.glob(f'{TEST_ECS_DIR}/*.eb')
```
akesandgren: should either use os.path.join or the '/' Path op

Flamefire: That is/was even missing the recursive flag, so the test does nothing at all. Fixed it and the found issues.

PR #5253, opened 2026-08-27
[5/6] Use constants for commonly used test easyconfig folders, files and content

Flamefire:
Followup to https://github.com/easybuilders/easybuild-framework/pull/5050

Split from https://github.com/easybuilders/easybuild-framework/pull/5203

Requires:

- #5271


easybuild/tools/filetools.py
```diff
@@ -2777,7 +2778,7 @@ def copy(paths, target_path, force_in_dry_run=False, **kwargs):
     :param force_in_dry_run: force running the command during dry run
     :param kwargs: additional named arguments to pass down to copy_dir
     """
-    if isinstance(paths, str):
```
boegel: @Flamefire Does this change belong in here?

Shouldn't this be accompanied by extra test cases for the `copy` function?

Flamefire: Without it it will fail. Separate PR is at https://github.com/easybuilders/easybuild-framework/pull/5255
You can merge that first to remove the change from this PR, but it's small enough and a PR with tests etc. exists so should be safe either way

boegel:
@Flamefire needs rebase after merge of https://github.com/easybuilders/easybuild-framework/pull/5255

test/framework/easyblock.py
```diff
@@ -2878,29 +2853,25 @@ def test_extensions_sanity_check(self):
...
 
         handle, toy_ec4 = tempfile.mkstemp(prefix='easyblock_test_file_', suffix='.eb')
         os.close(handle)
-        write_file(toy_ec4, toytxt + "\nmaxparallel = 6")
+        write_file(toy_ec4, TOY_EC_TXT + "\nmaxparallel = 6")
 
         handle, toy_ec5 = tempfile.mkstemp(prefix='easyblock_test_file_', suffix='.eb')
         os.close(handle)
-        write_file(toy_ec5, toytxt + "\nmaxparallel = False")
+        write_file(toy_ec5, TOY_EC_TXT + "\nmaxparallel = False")
```
akesandgren, line 2874: Dones't this result in TOY_EC getting read multiple times?

test/framework/easyblock.py
```diff
@@ -3501,7 +3457,7 @@ def test_checksum_step(self):
 
         # create test easyconfig from which checksums have been stripped
         test_ec = os.path.join(self.test_prefix, 'test.eb')
-        ectxt = read_file(toy_ec)
+        ectxt = TOY_EC_TXT
```
akesandgren: No, toy_ec is not TOY_EC here, toy_ec is 'toy-0.0-gompi-2018a-test.eb'

test/framework/easyblock.py
Flamefire: It does not. That was one of the motivations for this PR series: Read once and reuse. `TOY_EC_TXT` is a string so doesn't cause any reads

test/framework/easyconfig.py
```diff
@@ -2484,8 +2454,7 @@ def test_update(self):
 
     def test_hide_hidden_deps(self):
         """Test use of --hide-deps on hiddendependencies."""
-        test_dir = os.path.dirname(os.path.abspath(__file__))
-        ec_file = os.path.join(test_dir, 'easyconfigs', 'test_ecs', 'g', 'gzip', 'gzip-1.4-GCC-4.6.3.eb')
+        ec_file = TEST_ECS_DIR / 'g' / 'gzip' / 'gzip-1.4-GCC-4.6.3.eb'
```
akesandgren, line 2455: Why not os.path.join like everywhere else?

test/framework/easyconfig.py
```diff
@@ -3382,62 +3344,60 @@ def test_ActiveMNS_det_full_module_name(self):
         }
 
         init_config(build_options=build_options)
-        topdir = os.path.dirname(os.path.abspath(__file__))
-        ec_file = os.path.join(topdir, 'easyconfigs', 'test_ecs', 't', 'toy', 'toy-0.0-deps.eb')
+        ec_file = TEST_ECS_DIR / 't' / 'toy' / 'toy-0.0-deps.eb'
```
akesandgren, line 3335: Same here why not os.path.join?

test/framework/easyconfig.py
```diff
@@ -3448,8 +3408,7 @@ def test_find_related_easyconfigs(self):
 
     def test_modaltsoftname(self):
         """Test specifying an alternative name for the software name, to use when determining module name."""
-        topdir = os.path.dirname(os.path.abspath(__file__))
-        ec_file = os.path.join(topdir, 'easyconfigs', 'test_ecs', 't', 'toy', 'toy-0.0-deps.eb')
+        ec_file = TEST_ECS_DIR / 't' / 'toy' / 'toy-0.0-deps.eb'
```
akesandgren, line 3399: And here too os.path.join

test/framework/easyconfig.py
```diff
@@ -3484,8 +3442,7 @@ def test_software_license(self):
 
     def test_param_value_type_checking(self):
         """Test value tupe checking of easyconfig parameters."""
-        topdir = os.path.dirname(os.path.abspath(__file__))
-        ec_file = os.path.join(topdir, 'easyconfigs', 'test_ecs', 'g', 'gzip', 'gzip-1.4-broken.eb')
+        ec_file = TEST_ECS_DIR / 'g' / 'gzip' / 'gzip-1.4-broken.eb'
```
akesandgren, line 3433: And here

test/framework/easyconfig.py
```diff
@@ -4097,16 +4044,14 @@ def test_get_paths_for(self):
                 path.append(subdir)
         os.environ['PATH'] = os.pathsep.join(path)
 
-        top_dir = os.path.dirname(os.path.abspath(__file__))
         mkdir(os.path.join(self.test_prefix, 'easybuild'))
-        test_ecs = os.path.join(top_dir, 'easyconfigs')
-        symlink(test_ecs, os.path.join(self.test_prefix, 'easybuild', 'easyconfigs'))
+        symlink(TEST_ECS_DIR, os.path.join(self.test_prefix, 'easybuild', 'easyconfigs'))
```
akesandgren, line 4034: No, test_ecs is
TEST_DIR / 'easyconfigs'
not TEST_ECS_DIR

test/framework/easyblock.py
Flamefire: thanks, fixed. The test is a bit weird: It does similar test in the block above with `test-no-checksums.eb`

test/framework/easyconfig.py
```diff
@@ -4097,16 +4044,14 @@ def test_get_paths_for(self):
...
 
         # temporarily mock stderr to avoid printed warning (because 'eb' is not available via $PATH)
         with self.mocked_stderr():
             # locations listed in 'robot_path' named argument are taken into account
             res = get_paths_for(subdir='easyconfigs', robot_path=[self.test_prefix])
-        self.assertTrue(os.path.samefile(test_ecs, res[0]))
+        self.assertTrue(os.path.samefile(TEST_ECS_DIR, res[0]))
```
akesandgren, line 4040: Same here, they are not the same

test/framework/easyconfig.py
```diff
@@ -4119,15 +4064,15 @@ def test_get_paths_for(self):
         os.environ['PATH'] = '%s:%s' % (os.path.join(self.test_prefix, 'bin'), orig_path)
 
         res = get_paths_for(subdir='easyconfigs', robot_path=None)
-        self.assertTrue(os.path.samefile(test_ecs, res[-1]))
+        self.assertTrue(os.path.samefile(TEST_ECS_DIR, res[-1]))
```
akesandgren, line 4053: Same here

test/framework/easyconfig.py
```diff
@@ -4119,15 +4064,15 @@ def test_get_paths_for(self):
...
 
         # also works when 'eb' resides in a symlinked location
         altbin = os.path.join(self.test_prefix, 'some', 'other', 'symlinked', 'bin')
         mkdir(os.path.dirname(altbin), parents=True)
         symlink(os.path.join(self.test_prefix, 'bin'), altbin)
         os.environ['PATH'] = '%s:%s' % (altbin, orig_path)
         res = get_paths_for(subdir='easyconfigs', robot_path=None)
-        self.assertTrue(os.path.samefile(test_ecs, res[-1]))
+        self.assertTrue(os.path.samefile(TEST_ECS_DIR, res[-1]))
```
akesandgren, line 4061: And here

test/framework/easyconfig.py
```diff
@@ -4137,7 +4082,7 @@ def test_get_paths_for(self):
         os.environ['PATH'] = orig_path
         sys.path.insert(0, self.test_prefix)
         res = get_paths_for(subdir='easyconfigs', robot_path=None)
-        self.assertTrue(os.path.samefile(test_ecs, res[0]))
+        self.assertTrue(os.path.samefile(TEST_ECS_DIR, res[0]))
```
akesandgren, line 4071: And here

test/framework/easyconfig.py
```diff
@@ -4197,10 +4142,9 @@ def test_get_module_path(self):
 
     def test_not_an_easyconfig(self):
         """Test error reporting when a file that's not actually an easyconfig file is provided."""
-        test_ecs_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs',)
         # run test on an easyconfig file that was downloaded using wget using a non-raw GitHub URL
         # cfr. https://github.com/easybuilders/easybuild-framework/issues/2383
-        not_an_ec = os.path.join(os.path.dirname(test_ecs_dir), 'sandbox', 'not_an_easyconfig.eb')
+        not_an_ec = TEST_DIR / 'sandbox' / 'not_an_easyconfig.eb'
```
akesandgren, line 4133: Why not os.path.join?

test/framework/easyconfig.py
```diff
@@ -4314,8 +4255,7 @@ def test_deprecated(self):
 
     def test_deprecated_toolchain(self):
         """Test use of deprecated toolchain"""
-        topdir = os.path.dirname(os.path.abspath(__file__))
-        deprecated_toolchain_ec = os.path.join(topdir, 'easyconfigs', 'test_ecs', 't', 'toy', 'toy-0.0-gompi-2018a.eb')
+        deprecated_toolchain_ec = TEST_ECS_DIR / 't' / 'toy' / 'toy-0.0-gompi-2018a.eb'
```
akesandgren, line 4243: os.path.join?

test/framework/easyconfig.py
Flamefire: `pathlib.Path` allows this shorthand which reads like using regular paths. IMO better than the verbose `os.path.join`

test/framework/easyconfig.py
Flamefire: Technically yes, but I assume this is the intention: Copy the test easyconfigs, not the v1/v2 ones used for different tests.
The below is dependent on this change/fix, so must match

test/framework/easyblock.py
akesandgren: Ok, wasn't sure how that construction would work.

test/framework/easyconfig.py
akesandgren: And does that work correctly on Windows too? (the main reason for using os.path.join in the first place)

test/framework/easyconfig.py
akesandgren: Ok, was too lazy to read the full code :-)

test/framework/easyconfig.py
```diff
@@ -3806,10 +3759,8 @@ def test_template_constant_dict(self):
 
     def test_parse_deps_templates(self):
         """Test whether handling of templates defined by dependencies is done correctly."""
-        test_ecs = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'easyconfigs', 'test_ecs')
-
         pyec = os.path.join(self.test_prefix, 'Python-2.7.10-foss-2018a.eb')
-        shutil.copy2(os.path.join(test_ecs, 'p', 'Python', 'Python-2.7.10-intel-2018a.eb'), pyec)
+        shutil.copy2(os.path.join(TEST_ECS_DIR, 'p', 'Python', 'Python-2.7.10-intel-2018a.eb'), pyec)
```
akesandgren: Should/can this use copy_file instead?

test/framework/filetools.py
```diff
@@ -734,8 +736,7 @@ def fake_urllib_open(url, *args, **kwargs):
             return self.orig_filetools_std_urllib_urlopen(url, *args, **kwargs)
 
         fn = 'toy-0.0.eb'
-        test_dir = os.path.abspath(os.path.dirname(__file__))
-        toy_dir = os.path.join(test_dir, 'easyconfigs', 'test_ecs', 't', 'toy')
+        toy_dir = os.path.join(TEST_ECS_DIR, 't', 'toy')
         url = 'file://%s/%s' % (toy_dir, fn)
```
akesandgren: How about
```
url = 'file://%s' & TOY_EC
fn = os.path.basename(TOY_EC)
```
or at least the fn= part

test/framework/filetools.py
```diff
@@ -2755,23 +2742,20 @@ def test_clean_dir(self):
...
         # load_index just returns None if there is no index in specified directory
         self.assertEqual(ft.load_index(self.test_prefix), None)
 
-        num_files = len(glob.glob(test_ecs + '/**/*.*', recursive=True))
+        num_files = len(list(TEST_ECS_DIR.rglob('*.*')))
 
         # create index for test easyconfigs;
         # test with specified path with and without trailing '/'s
-        for path in [test_ecs, test_ecs + '/', test_ecs + '//']:
+        for path in [TEST_ECS_DIR, str(TEST_ECS_DIR), str(TEST_ECS_DIR) + '/', str(TEST_ECS_DIR) + '//']:
```
akesandgren: strictly speaking this should use os.path.sep and not strings with "/" in them

test/framework/robot.py
```diff
@@ -1242,7 +1229,7 @@ def test_robot_find_subtoolchain_for_dep(self):
         write_file(barec, barec_txt)
 
         # check without --minimal-toolchains
-        init_config(build_options={'robot_path': test_easyconfigs})
+        init_config(build_options={'robot_path': TEST_ECS_DIR})
```
akesandgren, line 1232: Further below you are doing str(TEST_ECS_DIR), why not here?
Same on line 1148

test/framework/robot.py
```diff
@@ -1434,7 +1419,7 @@ def test_check_conflicts(self):
         self.assertIn("Conflict found for dependencies of foss-2018a: GCC-4.6.4 vs GCC-6.4.0-2.28", stderr)
 
         # Can also return the text
-        with self.mocked_stdout_stderr(mock_stdout=False) as mocked_stderr:
+        with self.mocked_stderr() as mocked_stderr:
```
akesandgren, line 1422: Not sure this one belongs in this PR

test/framework/easyconfig.py
Flamefire: Yes. That's one of the motivations of that class: Uniform style that works everywhere. Behind the scenes `Path` is either a `WindowsPath` or `PosixPath` doing the right thing depending on the OS

akesandgren: ok good to know

test/framework/filetools.py
Flamefire:
```
        fn = TOY_EC.name
        url = f'file://{TOY_EC.parent}/{fn}'
```
works

test/framework/robot.py
Flamefire: True, not required. Replaced all occurrences

akesandgren:
Hmmm why did the conflict pop up

Flamefire:
Rebased to resolve it.

test/framework/robot.py
Flamefire: True, just noticed that weird/old usage while changing adjacent lines and though the single-line change doesn't hurt.

Flamefire:
@akesandgren I merged #5293 to this and CI now passes again without the timeouts. If you prefer to merge the changes separately I'll rebase this one after #5293 is merged
The other comments should all be addressed too

PR #5255, opened 2026-08-27
Accept Path argument in `filetools.copy`

Flamefire:
The check for `str` before converting to list fails when being passed a `pathlib.Path` instance.
Handle that.

To make this clear I added type hints.
I wanted to create a shortcut/alias for `Union[str, Path]` but lack a good name `PathLike` is already taken in the standard [`os` module](https://docs.python.org/3/library/os.html#os.PathLike) to mean something like `Path`
Maybe just `PathOrStr`?

### DIFF
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -2784,7 +2784,7 @@ def copy(paths, target_path, force_in_dry_run=False, **kwargs):
     :param force_in_dry_run: force running the command during dry run
     :param kwargs: additional named arguments to pass down to copy_dir
     """
-    if isinstance(paths, str):
+    if isinstance(paths, (str, Path)):
         paths = [paths]
 
     _log.info("Copying %d files & directories to %s", len(paths), target_path)
--- a/easybuild/tools/filetools.py
+++ b/easybuild/tools/filetools.py
@@ -62,6 +62,7 @@ import zlib
 from functools import partial
 from html.parser import HTMLParser
 from pathlib import Path
+from typing import List, Union
 import urllib.request as std_urllib
 
 from easybuild.base import fancylogger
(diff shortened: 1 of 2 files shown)


boegel:
> To make this clear I added type hints. I wanted to create a shortcut/alias for `Union[str, Path]` but lack a good name `PathLike` is already taken in the standard [`os` module](https://docs.python.org/3/library/os.html#os.PathLike) to mean something like `Path` Maybe just `PathOrStr`?

Or just `PathStr`?

Flamefire:
> Or just `PathStr`?

Looks a bit to much like "string" to me.
Brainstorming with AI:

- PathArg
- PathInput
- StrOrPath
- PathType
- PathValue

I'd go with `PathOrStr`, `PathArg` or `PathValue`

boegel:
> > Or just `PathStr`?
> 
> Looks a bit to much like "string" to me. Brainstorming with AI:
> 
>     * PathArg
> 
>     * PathInput
> 
>     * StrOrPath
> 
>     * PathType
> 
>     * PathValue
> 
> 
> I'd go with `PathOrStr`, `PathArg` or `PathValue`

Well, but it *is* a string value, it just happens to represents a path (so `PathStr` makes sense to me). :)

If you don't like `PathStr`, I'd go with `PathOrStr`, that's the way of least surprise then imho

Flamefire:
> Well, but it _is_ a string value, it just happens to represents a path (so `PathStr` makes sense to me). :)

I'm hesitant as it sounds like `PathStr = str`.

Optimally the argument would never be a `string` instance but a `Path` instance. If we used `pathlib.Path` everywhere we expect/use a path I'd expect the code to be easier to understand. But that would be a transition and e.g. the stdlib functions accept both a `str` and a `Path` too

> If you don't like PathStr, I'd go with PathOrStr, that's the way of least surprise then imho

I think the same: Best to be explicit

boegel (approved):
lgtm

PR #5256, opened 2026-08-28
Fix line endings in `easybuild/framework/extension.py`

Flamefire:
We have a single file that is stored with CRLF line endings instead of LF

This causes (minor) headaches when modifying it as they might get mixed up.

Do a one-off conversion with `dos2unix`.

Checked with `git ls-files --eol $(find . -name '*.py') | grep crlf`

Optional: We can use a `.gitattributes` file in our repos to ensure `*.py` and `*.eb` files are uniformly committed with LF line endings.


boegel (approved):
GitHub makes this quite horrible to review, unless you use the "hide whitespace" option, then it's pretty obvious that there's no actual changes being made here

PR #5260, opened 2026-09-02
Add factory method `make_deprecated_dict_class` for creating a dict with deprecated/replaced keys

Flamefire:
This allows reuse for different use cases.

I'll use that for https://github.com/easybuilders/easybuild-framework/pull/5257 but the change looks large enough for its own PR. It mostly moves the dict-wrapper we already have into a more generic util


easybuild/framework/easyconfig/format/pyheaderconfigobj.py
```diff
@@ -103,40 +104,11 @@ def wrapper(self, key, *args, **kwargs):
...
 
 
-class DeprecatedDict(dict):
-    """Custom dictionary that handles deprecated easyconfig template constants gracefully"""
-
-    def __init__(self, *args, **kwargs):
-        self.clear()
-        self.update(*args, **kwargs)
-
-    @handle_deprecated_constants
```
boegel, line 113: should we then also deprecate this decorator, so we can get rid of it in EasyBuild 6.0?

easybuild/tools/deprecated_dict.py
```diff
@@ -0,0 +1,114 @@
...
+            return super().__setitem__(key, value)
+
+        @handle_deprecated_keys
+        def get(self, key, default=None):
+            return super().get(key, default)
+
+        def update(self, *args, **kwargs):
+            if args:
+                if len(args) > 1:
+                    raise TypeError(f"update expected at most 1 argument, got {len(args)}")
```
boegel, line 103: that's a very generic error message, maybe we can give another hint where it comes from?
```suggestion
                    raise TypeError(f"DeprecatedDict.update expected at most 1 argument, got {len(args)}")
```

test/framework/utilities_test.py
```diff
@@ -229,6 +231,110 @@ def test_unique_ordered_extend(self):
         error_pattern = "given base cannot be extended"
         self.assertErrorRegex(EasyBuildError, error_pattern, tu.unique_ordered_extend, base, reference)
 
+    def test_deprecated_dict(self):
+        """Test the DeprecatedDict class"""
```
boegel: you're actually testing `make_deprecated_dict_class` (not the `DeprecatedDict` class we have in `pyheaderconfigobj`

boegel (changes requested):

easybuild/framework/easyconfig/format/pyheaderconfigobj.py
Flamefire: Makes sense :+1:

easybuild/tools/deprecated_dict.py
Flamefire: This mirrors the error message from the base class:
```
x={1:2}
x.update(1,2,3)
TypeError: update expected at most 1 argument, got 3
```

So I'd rather keep it as the whole class is supposed to be as close to the raw dict as possible.

boegel: Well, that's exactly my gripe here... I've been bit by a `TypeError` like this multiple times not knowing where it comes from, it's a horribly unclear error. Especially your example: "What do you mean, an update is expected at the "most 1" argument? This sentence makes no sense!"

test/framework/utilities_test.py
boegel:
```suggestion
    def test_make_deprecated_dict_class(self):
        """Test for make_deprecated_dict_class"""
```

PR #5264, opened 2026-09-02
[1/5] Use `skipTest` rather than faulty copy-paste of `return` conditions

Flamefire:
Mostly preparation for use of skip markers: Contains the `requires_foo` decorators used by other PRs and related machinery

Example use for pysvn and GC3Pie

Extracted from https://github.com/easybuilders/easybuild-framework/pull/4188

### DIFF
--- a/test/framework/utilities.py
+++ b/test/framework/utilities.py
@@ -51,7 +51,7 @@ import easybuild.tools.module_naming_scheme.toolchain as mns_toolchain
 from easybuild.framework.easyconfig import easyconfig
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.main import main
-from easybuild.tools import config
+from easybuild.tools import config, LooseVersion
 from easybuild.tools.config import GENERAL_CLASS, Singleton, module_classes
 from easybuild.tools.configobj import ConfigObj
 from easybuild.tools.environment import modify_env
@@ -607,7 +607,12 @@ def requires_GC3Pie():
         ok = True
     except ImportError:
         ok = False
-    return unittest.skipUnless(ok, "GC3Pie not available")
+    if LooseVersion(sys.version) < '3.11':
+        return unittest.skipUnless(ok, "GC3Pie not available")
+    else:
+        # GC3Pie not available for Python 3.11 so silently skip:
+        # https://github.com/gc3pie/gc3pie/issues/674
+        return skip_silentCI_unless(ok, "GC3Pie not available")
 
 
 def requires_graphviz():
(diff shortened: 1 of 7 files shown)


test/framework/utilities.py
```diff
@@ -557,3 +558,87 @@ def find_full_path(base_path, trim=(lambda x: x)):
...
+        return unittest.skipUnless(ok, "GC3Pie not available")
+    else:
+        # GC3Pie not available for Python 3.11 so silently skip:
+        # https://github.com/gc3pie/gc3pie/issues/674
+        return skip_silentCI_unless(ok, "GC3Pie not available")
+
+
+def requires_graphviz():
+    try:
+        import graphwiz  # noqa # pylint:disable=unused-import
```
boegel: sneaky typo?

```suggestion
        import graphviz  # noqa # pylint:disable=unused-import
```

test/framework/utilities.py
```diff
@@ -557,3 +558,87 @@ def find_full_path(base_path, trim=(lambda x: x)):
...
+    """Decorator to skip a test if the condition is met.
+
+    On CI the test is turned into a no-op to avoid any output."""
+    if 'CI' in os.environ:
+        return skip_never if condition else skip_silently
+    else:
+        return unittest.skipUnless(condition, reason)
+
+
+def requires_pycodestyle():
```
boegel: please add a docstring to all of these `requires_*` decorators?

test/framework/utilities.py
```diff
@@ -557,3 +558,87 @@ def find_full_path(base_path, trim=(lambda x: x)):
...
+    try:
+        import pycodestyle  # noqa # pylint:disable=unused-import
+        ok = True
+    except ImportError:
+        ok = False
+    return unittest.skipUnless(ok, "no pycodestyle available")
+
+
+def requires_autopep8():
```
boegel: if you introduce them, you might as well also *use* them, I don't think that would blow up this PR a lot?

boegel (changes requested):

test/framework/utilities.py
Flamefire: I removed the repetition using a helper function: `requires_PyYAML = requires_package('yaml')`

For those I added a comment and the remaining ones with special cases I added docstrings

test/framework/utilities.py
Flamefire: It actually does blow it up. That's why I made the 5 sub-PRs in the first place.
If you want me to use them, check https://github.com/easybuilders/easybuild-framework/pull/4188

If you merge this one first, I can rebase the other 4 making them much smaller.

PR #5267, opened 2026-09-03
Initialize `name` and `version` in `EasyConfig` instance of `Extension`

Flamefire:
Do this early instead of in `ExtensionEasyBlock` as otherwise usage of `self.name` would yield the parent name due to Python MRO inside `__init__`

It also makes sense to keep those in sync with the updated template values.
I'm wondering if we could just call `self.cfg.generate_template_values()` after setting name and version

### DIFF
--- a/easybuild/framework/extension.py
+++ b/easybuild/framework/extension.py
@@ -146,8 +146,8 @@ class Extension:
             self.cfg[opt_name] = get_easyconfig_parameter_default(opt_name)
 
         # Update name and version
-        self.cfg['name'] = self.ext.get('name', None)
-        self.cfg['version'] = self.ext.get('version', None)
+        self.cfg['name'] = name
+        self.cfg['version'] = version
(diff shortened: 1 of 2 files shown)


easybuild/framework/extension.py
```diff
@@ -145,6 +145,9 @@ def __init__(self, mself, ext, extra_params=None):
         for opt_name in restore_options:
             self.cfg[opt_name] = get_easyconfig_parameter_default(opt_name)
 
+        # Update name and version
+        self.cfg['name'] = self.ext.get('name', None)
+        self.cfg['version'] = self.ext.get('version', None)
```
boegel: we should reuse the existing `name` and `version` local variables, see also below
```suggestion
        self.cfg['name'] = name
        self.cfg['version'] = version
```

boegel (changes requested):

boegel (approved):
lgtm

PR #5268, opened 2026-09-03
Remove superflous cleanup steps in tests

Flamefire:
The test tmp directory is removed by `EnhancedTestCase` so the manual cleanup can be removed to reduce the noise.

### DIFF
--- a/test/framework/easyblock.py
+++ b/test/framework/easyblock.py
@@ -164,7 +164,6 @@ class EasyBlockTest(EnhancedTestCase):
 
         # cleanup
         eb.close_log()
-        os.remove(eb.logfile)
 
     def test_load_module(self):
         """Test load_module method."""
@@ -218,7 +217,6 @@ class EasyBlockTest(EnhancedTestCase):
 
         # cleanup
         eb.close_log()
-        os.remove(eb.logfile)
 
         # test HMNS module load when conflicting dependencies are available in both Core and
         # toolchain-specific modulepaths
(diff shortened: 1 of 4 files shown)


test/framework/utilities.py
```diff
@@ -273,12 +272,6 @@ def tearDown(self):
             log_handler.close()
             log.removeHandler(log_handler)
 
-        # cleanup test tmp dir
-        try:
-            shutil.rmtree(self.test_prefix)
```
boegel, line 278: Don't we want to keep it here in `tearDown`?!

test/framework/easyblock.py
```diff
@@ -162,10 +162,6 @@ def extra_options():
         self.assertIn('options', extra_options)
         self.assertEqual(extra_options['extra_param'], [None, "help", CUSTOM])
 
-        # cleanup
-        eb.close_log()
```
boegel, line 166: hmm, not sure it's wise to not close the log, we better keep that I think

boegel (changes requested):

boegel:
@Flamefire Tests on top of Python 3.10 have been running for almost 2 hours (see [here](https://github.com/easybuilders/easybuild-framework/actions/runs/33768150608/job/100691920112?pr=5268)), I think that may be related to the removal of the `close_log` calls: when not doing those, you keep the logger active, and then you get excessive logging which slows things down massively...

test/framework/easyblock.py
Flamefire: You are right, we need to close those logs as this unregisters the handler that writes to self.logfile
I wasn't fully aware of this and it seems very brittle to me. I wonder in how many instances we do not close the log.

An option would be a context handler like:
```
with EasyBlock(foo) as eb:
  eb.dox()
# log automatically closed
```

That would be for a separate PR though. In this I'll put them back and add the accidentally remove cleanup call from the test class.

PR #5269, opened 2026-09-04
Replace raw `dict` by `CategorizedPaths` object (`NamedTuple`(-like))

Flamefire:
The current dict is pretty opaque and makes spelling mistakes easy.
Avoid by using a typed, named tuple.

There should be no other entries as this is mostly an internal class but the backwards-compat path allows it for now.

In EB 6.0 this can be a pure `NamedTuple` or a dataclass

### DIFF
--- a/easybuild/tools/github.py
+++ b/easybuild/tools/github.py
@@ -44,11 +44,12 @@ import sys
 import tempfile
 import time
 import urllib.error
+from collections.abc import MutableMapping
 from datetime import datetime, timedelta
 from http import HTTPStatus
 from http.client import HTTPException
 from string import ascii_letters
-from typing import List, NamedTuple
+from typing import ClassVar, List, Tuple
 from urllib.request import HTTPError, URLError, urlopen
 
 from easybuild.base import fancylogger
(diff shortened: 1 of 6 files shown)


boegel:
In general, I like this, it indeed leaves less room for mistakes.

However, we're changing the function signature and or return type of several functions that are part of our API here:
- `categorize_files_by_type` now returns a `CategorizedPaths` instance, instead of a `dict`;
- `new_branch_github`, `new_pr_from_branch`, `new_pr`, `det_pr_target_repo`, `update_branch`, `update_pr` now take a `CategorizedPaths` instance as a value, rather than the `dict` they used to take before;

Even if we wait with merging this for EasyBuild 6.0, we should still deprecate the old function signatures.

This could be done by:
- adding an option to `categorize_files_by_type` to determine the type of return value (`dict` by default, opt-in to `CategorizedPaths` instance);
- auto-convert a `dict` value passed to `new_pr` & co to a `CategorizedPaths` instance internally, if needed (so accept both `dict` and `CategorizedPaths` instance as input value);

I agree that it's unlikely that people would be using one of these functions, but if anyone does (for example in a Python script that leverages EasyBuild framework as a library, see also [here](https://tutorial.easybuild.io/2023-eb-eessi-uk-workshop/easybuild-python-library/)), then we would be breaking their script.

Backwards compatibility is a strong focus point of EasyBuild, and I very much want to keep it like that...

easybuild/tools/github.py
```diff
@@ -90,17 +92,99 @@
     _log.warning("Failed to import 'git' Python module: %s", err)
 
 
+class CategorizedPaths(MutableMapping):  # Todo: use NamedTuple or dataclass instead
```
boegel: Why not use `NamedTuple` already? It probably requires a newer Python version as minimal version?
If so, mention that in the comment (and put it above the `class` line while you're at it)

Flamefire: We need the dict-interface. Without that the NamedTuple (or dataclass in 3.9) from 4a794212e4633c3917be544807b1d07e0093d85e are enough:
```
class CategorizedPaths(NamedTuple):
    easyconfigs: List[str]
    easyconfigs: List[str]
    files_to_delete: List[str]
    patch_files: List[str]
    py_files: List[str]
```

Enhanced the comment and added another one below

Flamefire:
> This could be done by:
> 
>     * adding an option to `categorize_files_by_type` to determine the type of return value (`dict` by default, opt-in to `CategorizedPaths` instance);
> 
>     * auto-convert a `dict` value passed to `new_pr` & co to a `CategorizedPaths` instance internally, if needed (so accept both `dict` and `CategorizedPaths` instance as input value);

Is that an either-or or both?
I would have said the first is enough but we can be extra save and do both. Now commit for the 2nd point added

PR #5270, opened 2026-09-04
Make network ops on CI more reliable

Flamefire:
1. Use smaller file for download timeout test
	The test sometimes fails, so try a smaller file (~500B instead of some kB) and not from github.com.
1. Increase download timeout for end2end tests to 100s
2. Make apt-get more reliable
	- Do not install additional packages
	- Do not ask questions
	- Retry 3 times


boegel (approved):
lgtm

PR #5271, opened 2026-09-04
Allow `pathlib.Path` input value for `filetools.normalize_path`

Flamefire:
The side-effect of this is that we can pass `Path` instances to modules-tool which usually calls `normalize_path` very early

Note that `Path` instances are normalize using the same rules we have except for the empty string which is `Path('.')`
So we could replace the whole function by `return str(Path(path)) if path else '.'`


boegel (approved):
lgtm

PR #4766, opened 2025-02-18
add support for specifying dependencies required to obtain source files via `source_deps` easyconfig parameter

PetrKralCZ:
(created using `eb --new-pr`)

PR #5273, opened 2026-09-04
Add support for specifying dependencies required to obtain source files via `source_deps` easyconfig parameter

pavelToman:
This PR replaces https://github.com/easybuilders/easybuild-framework/pull/4766.

It adds support for the `source_deps` easyconfig parameter, which can be used to specify dependencies that are required to obtain source files.

This is useful for sources that require additional tools during the fetch step, such as `git-lfs`. Source dependencies are loaded before fetching the sources and the original environment is restored afterwards.

It also supports using `source_deps` together with `eb --fetch`, where EasyBuild normally disables the modules tool.

Created with help of ChatGPT5.6

### DIFF
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -5437,7 +5437,7 @@ def build_and_install_one(ecdict, init_env):
                     adjust_permissions(log_dir, stat.S_IWUSR, add=True, recursive=True)
                 else:
                     parent_dir = os.path.dirname(log_dir)
-                    if os.path.exists(parent_dir):
+                    if os.path.exists(parent_dir) and not (os.stat(parent_dir).st_mode & stat.S_IWUSR):
                         adjust_permissions(parent_dir, stat.S_IWUSR, add=True, recursive=False)
                         mkdir(log_dir, parents=True)
                         adjust_permissions(parent_dir, stat.S_IWUSR, add=False, recursive=False)
@@ -5530,7 +5530,7 @@ def build_and_install_one(ecdict, init_env):
                     copy_file(patch_path, target)
                     _log.debug("Copied patch %s to %s", patch_path, target)
 
-                if build_option('read_only_installdir'):
+                if build_option('read_only_installdir') and not app.cfg['stop']:
                     # take away user write permissions (again)
                     perms = stat.S_IWUSR | stat.S_IWGRP | stat.S_IWOTH
                     adjust_permissions(new_log_dir, perms, add=False, recursive=True)
--- a/easybuild/framework/easyblock.py
+++ b/easybuild/framework/easyblock.py
@@ -2862,6 +2862,14 @@ class EasyBlock:
                 raise EasyBuildError("EasyBuild-version %s is newer than the currently running one. Aborting!",
                                      easybuild_version)
 
+        source_deps = self.cfg['source_deps']
+        pre_fetch_env = None
+        # load modules for source dependencies (if any)
+        if source_deps:
+            pre_fetch_env = copy.deepcopy(os.environ)
+            source_deps_mod_names = [d['short_mod_name'] for d in source_deps]
+            self.modules_tool.load(source_deps_mod_names)
+
         start_progress_bar(PROGRESS_BAR_DOWNLOAD_ALL, self.cfg.count_files())
 
         if self.dry_run:
@@ -2948,6 +2956,9 @@ class EasyBlock:
             self.log.info("Skipped installation dirs check per user request")
 
         stop_progress_bar(PROGRESS_BAR_DOWNLOAD_ALL)
+
+        if pre_fetch_env:
+            restore_env(pre_fetch_env)
 
     def checksum_step(self):
         """Verify checksum of sources and patches, if a checksum is available."""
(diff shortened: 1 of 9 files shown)


pavelToman:
Tested here:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/26786#issuecomment-5543228789
- https://github.com/easybuilders/easybuild-easyconfigs/pull/23156#issuecomment-5569828060
- https://github.com/easybuilders/easybuild-easyconfigs/pull/23156#issuecomment-5569835187
- https://github.com/easybuilders/easybuild-easyconfigs/pull/26813#issuecomment-5586814500

easybuild/framework/easyconfig/default.py
```diff
@@ -179,6 +179,7 @@
     'multi_deps_load_default': [True, "Load module for first version listed in multi_deps by default", DEPENDENCIES],
     'osdependencies': [[], "OS dependencies that should be present on the system", DEPENDENCIES],
     'moddependpaths': [None, "Absolute path(s) to prepend to MODULEPATH before loading dependencies", DEPENDENCIES],
+    'source_deps': [[], "source dependencies that should be present befory getting the sources", DEPENDENCIES],
```
boegel:
```suggestion
    'source_deps': [[], "List of dependencies that are required to download or create the sources", DEPENDENCIES],
```

easybuild/framework/easyconfig/easyconfig.py
```diff
@@ -831,6 +831,13 @@ def remove_false_versions(deps):
                 builddeps = [self._parse_dependency(dep, build_only=True) for dep in builddeps]
             self['builddependencies'] = remove_false_versions(builddeps)
 
+            source_deps = self['source_deps']
+            if source_deps and all(isinstance(x, (list, tuple)) for b in source_deps for x in b):
```
boegel: the `b` is confusing here...
```suggestion
            if source_deps and all(isinstance(x, (list, tuple)) for y in source_deps for x in y):
```

easybuild/framework/easyblock.py
```diff
@@ -2949,6 +2962,9 @@ def fetch_step(self, skip_checksums=False):
 
         stop_progress_bar(PROGRESS_BAR_DOWNLOAD_ALL)
 
+        if pre_fetch_env:
```
```suggestion
        # revert back to environment before fetching sources, if source deps were loaded
        if pre_fetch_env:
```

easybuild/tools/config.py
```diff
@@ -467,6 +467,7 @@ def mk_full_default_path(name, prefix=DEFAULT_PREFIX):
         'external_modules_metadata',
         'extra_ec_paths',
         'mod_depends_on',  # deprecated
+        'original_modules_tool',
```
boegel:
```suggestion
        'modules_tool',
```

easybuild/framework/easyblock.py
```diff
@@ -2862,6 +2863,18 @@ def fetch_step(self, skip_checksums=False):
...
                                      easybuild_version)
 
+        source_deps = self.cfg['source_deps']
+        pre_fetch_env = None
+        # load modules for source dependencies (if any)
+        if source_deps:
+            pre_fetch_env = copy.deepcopy(os.environ)
+
+            if isinstance(self.modules_tool, NoModulesTool):
+                self.modules_tool = modules_tool(modules_tool_name=build_option('original_modules_tool'))
```
boegel: I don't see a need for the `original_` prefix, just use `modules_tool`?
```suggestion
                self.modules_tool = modules_tool(modules_tool_name=build_option('modules_tool'))
```

easybuild/tools/options.py
```diff
@@ -1398,6 +1398,7 @@ def _postprocess_config(self):
             self.options.stop = FETCH_STEP
             self.options.ignore_locks = True
             self.options.ignore_osdeps = True
+            self.original_modules_tool = self.options.modules_tool
```
boegel: I don't think this is required, since below we're adding `modules_tool` as a build option, and we only query this via `build_option('modules_tool')`?

```suggestion
```

easybuild/tools/options.py
```diff
@@ -1941,6 +1942,7 @@ def set_up_configuration(args=None, logfile=None, testing=False, silent=False, r
         'command_line': eb_cmd_line,
         'external_modules_metadata': parse_external_modules_metadata(options.external_modules_metadata),
         'extra_ec_paths': extra_ec_paths,
+        'original_modules_tool': getattr(eb_go, 'original_modules_tool', None),
```
boegel:
```suggestion
        'modules_tool': options.modules_tool,
```

test/framework/easyblock.py
```diff
@@ -1317,6 +1317,49 @@ def create_file(_filename, _url, path, *_args, **_kwargs):
         self.assertTrue(os.path.samefile(eb.src[0]['path'], expected_path_src))
         self.assertTrue(os.path.samefile(eb.patches[0]['path'], expected_path_patch))
 
+    def test_fetch_step_source_deps(self):
+        """Test fetching sources with source dependencies."""
+        init_config(
+            [f'--sourcepath={self.test_prefix}'],
+            build_options={'original_modules_tool': 'Lmod'},
+        )
```
boegel: There should be no need to set the `(original_)modules_tool` build option, we just need to tweak the `sourcepath`:
```suggestion
        init_config([f'--sourcepath={self.test_prefix}'])
```

boegel (approved):
lgtm

PR #5282, opened 2026-09-13
change RPATH wrapper to still add `-rpath` options when `-v` option is used in linker command

### LINKED ISSUE #5258: RPATH wrapper assumes that linker option `-v` only prints version, but it doesn't
Encountered originally in https://github.com/easybuilders/easybuild-easyconfigs/pull/26731
@Crivella  found that it does
```
target_link_options(aflow PRIVATE "LINKER:-v")
```

The linker still links, and prints extra information when this flag is added, but we shortcut the rpath wrapper script:

https://github.com/easybuilders/easybuild-framework/blob/a64fc639b5dd3d6ad706a985ad61607e0829a6aa/easybuild/scripts/rpath_args.py#L116-L117

so rpath linking didn't work here.

Not sure how, or if, we want to handle this. Maybe if there is no other flags but -v. Maybe it's compiler specific, haven't tested.

boegel:
fixes #5258


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

